PR #2896, opened 2020-04-01
2334 - Add ttl to get_access_token.

### LINKED ISSUE #2334: Let callers configure oauth token TTL
It can be useful for applications to be able to create shorter-lived access tokens, e.g. in order to test handling of expired tokens. For a concrete use-case see https://github.com/mozilla/application-services/issues/2279#issuecomment-561861087.

Let's add the ability for apps to request a specific TTL on their tokens.

┆Issue is synchronized with this [Jira Story](https://jira.mozilla.com/browse/SYNC-486)
┆Epic: papercuts

jhugman:
Fixes #2334.

Adds an optional `ttl` parameter to kotlin and swift interfaces, to limit the time that tokens are cached.

The FFI uses a `zero` as a default parameter.

This is not expected to be a breaking change for the iOS and Android.

- [X] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one
- ~[ ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/master/docs/dependency-management.md)~
  - ~Any new dependencies are accompanied by a summary of the due dilligence applied in selecting them.~


CHANGES_UNRELEASED.md
```diff
@@ -17,3 +17,4 @@ Please call `Viaduct.shared.useReqwestBackend()` as soon as possible before usin
 
 - A new function was added to return a list of duplicate logins, ignoring
   username. ([#2542](https://github.com/mozilla/application-services/pull/2542))
+- Added an optional `ttl` parameter to `getAccessToken`, to limit the cache length of the token.
```
eoger, line 20: This entry belongs in a different category (FxA Client), see `CHANGELOG.md` for examples.

components/fxa-client/src/oauth.rs
```diff
@@ -66,7 +68,12 @@ impl FirefoxAccount {
         let since_epoch = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .map_err(|_| ErrorKind::IllegalState("Current date before Unix Epoch."))?;
-        let expires_at = since_epoch.as_secs() + resp.expires_in;
+        let expires_in = if 0 < ttl && ttl < resp.expires_in {
+            ttl
+        } else {
+            resp.expires_in
+        };
+        let expires_at = since_epoch.as_secs() + expires_in;
```
eoger, line 76: Shouldn't we pass the TTL to the server in the request parameters instead? cc @rfk

components/fxa-client/src/oauth.rs
```diff
@@ -66,7 +68,12 @@ impl FirefoxAccount {
         let since_epoch = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .map_err(|_| ErrorKind::IllegalState("Current date before Unix Epoch."))?;
-        let expires_at = since_epoch.as_secs() + resp.expires_in;
+        let expires_in = if 0 < ttl && ttl < resp.expires_in {
```
eoger, line 71: Can we push the "magic number" logic in `components/fxa-client/ffi/src/lib.rs` instead, and take a `ttl: Option<u64>` parameter in this function?

rfk: Agreed, `Option<u64>` is the idiomatic rust way to do this, and I think that converting from "what's convenient for the FFI" to "what's idiomatic rust" is currently the job of the FFI bindings code.

components/fxa-client/src/oauth.rs
rfk: Yes we should. I expect this would need to be passed in to `access_token_with_refresh_token` and `access_token_with_session_token` in order to pass it on to the server. We should then rely on the server's calculation of `expires_in` rather than doing our own calculation here (since the server is free to ignore the requested ttl and apply e.g. its own bounds on how big or small the ttl is allowed to be).

CHANGES_UNRELEASED.md
jhugman: :+1:

PR #2897, opened 2020-04-01
Add a ./start script for contributors

vladikoff:


tools/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+        ("Run all tests for a pull request", "sh ./automation/all_tests.sh"),
+        ("Run all Rust tests", "sh ./automation/all_rust_tests.sh"),
+
+        ("Verify Desktop Environment", "sh ./libs/verify-desktop-environment.sh"),
+        ("Verify Android Environment", "sh ./libs/verify-android-environment.sh"),
+        ("Verify iOS Environment", "sh ./libs/verify-ios-environment.sh"),
+
+        ("Smoke test Android Components", "python3 ./automation/smoke-test-android-components.py"),
+        ("Smoke test Fenix Components", "python3 ./automation/smoke-test-fenix.py"),
+        ("Smoke test Firefox iOS Components", "python3 ./automation/smoke-test-firefox-ios.py"),
```
vladikoff, line 18: @rfk is this missing any scripts that were introduced as part of the recent work?

vladikoff:
> having a kitchen-sink start script listing the available commands is going to help anybody here

It helps me to work with the project, there are just too many scripts hidden in various directories that do different things. Also new scripts were introduced in the past weeks that is hard to keep track off. Without ease of access it's very hard to remember what tools we have at our disposal.
We can also probably tweak the UI a bit later, i'd prefer inputting the option number instead of using arrows

README.md
```diff
@@ -2,6 +2,10 @@
 
 _A platform for building cloud-powered applications that target Firefox users_
 
+### TL;DR
+
+Run `./start` to see the available commands.
```
vladikoff, line 7: this can also be `./dev` or something like that

eoger:
> > having a kitchen-sink start script listing the available commands is going to help anybody here
> 
> It helps me to work with the project, there are just too many scripts hidden in various directories that do different things. Also new scripts were introduced in the past weeks that is hard to keep track off. Without ease of access it's very hard to remember what tools we have at our disposal.
> We can also probably tweak the UI a bit later, i'd prefer inputting the option number instead of using arrows

Another thing that bothers me is there's no way to pass an argument to the scripts this program calls since they are hardcoded. For example `./automation/smoke-test-firefox-ios.py` is *very* likely to be used with `--use-local-repo <PATH>`.
I strongly believe this "script discovery" problem is not be solved by adding more code but by having better documentation.

vladikoff:
>  For example ./automation/smoke-test-firefox-ios.py is very likely to be used with --use-local-repo <PATH>.

It's a bit suboptimal that the script currently creates new tmp directories every time it runs without an option. We could write a new helper on top of it that clones FxiOS in the root of a-s or as a sibling of it and just works afterwards.

Also keep in mind the knowledge bias you have with most of these scripts (in /libs and /automation) because you introduced them and have a clear picture of they do.

vladikoff:
This moved to https://github.com/vladikoff/a-s-dev

PR #2901, opened 2020-04-02
Add some quality-of-life improvements for running android smoketests.

rfk:
The most important one is that we avoid publishing a new set of aars
if no code has actually changed. I'm not thrilled about the *way* we
accomplish it, but it works for me...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ credentials.json
 *-engine.json
 .cargo
 *.db
+.*.swp
 
 # Android stuff.
 *.iml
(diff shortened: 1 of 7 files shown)


rfk:
> I assume that's because gradlew is very slow to re-publish aars if contents have not changed?

(I thought I replied to this already, but I don't see it here, so trying again...)

It's not *that* slow - maybe 20-30 seconds on my new machine, previously on a minute or two on my WSL environment.  If that were the only reason to do it I probably wouldn't bother.

My main goal here was to stop generating a bunch of identical .aars with slightly different names and filling up my local maven repository, both for disks-space reasons, and to make it easier to find the one I want if I need to go poke around in the generated artifacts.

I am open to the suggestion that the complexity is not worth the benefit here :-)

PR #2917, opened 2020-04-07
Some protobuf fixes for moz-central integration

eoger:
- mozilla-central doesn't have the "full" protobuf lib but the "lite" one, which doesn't include reflection/descriptors (we don't use them anyway). `LITE_RUNTIME` is also recommended for mobile devices and should reduce the generated code size so that's a win-win.
- The package name we had (`msg_types`) is not really descriptive and I wouldn't be surprised we can't check that into moz-central, so I changed it to something more explicit. ~This change doesn't affect any of our current targets.~ It affected iOS, hence the `swift_prefix` commit.


eoger:
Updated Android protobuf lib/gradle wrapper since it was yelling at the `swift_prefix` statement.

gradle.properties
```diff
@@ -21,3 +21,5 @@ libVcsUrl=https://github.com/mozilla/application-services.git
 
 libLicense=MPL-2.0
 libLicenseUrl=https://www.mozilla.org/en-US/MPL/2.0/
+
+android.useAndroidX=true
```
thomcc, line 25: Hmm... Are we allowed to add this? IIRC androidX causes problems for some of our dependencies?

eoger: It seems that we use it already for glean in tests (`ApplicationProvider`) :/ I'll make sure it works

PR #2926, opened 2020-04-08
Check-in protobuf Rust files

eoger:
So this is blocking me from landing Viaduct on mozilla-central. `tools/protobuf-gen/README.md` in this diff explains well why we're doing this.


.circleci/config.yml
```diff
@@ -253,6 +253,13 @@ jobs:
       - run: cargo clippy --version
       - run: bash automation/all_clippy_checks.sh
       - save-sccache-cache
+  Check Protobuf files are up-to-date:
+    docker:
+      - image: circleci/rust:latest
+    resource_class: small
+    steps:
+      - checkout
+      - run: ./automation/check_protobuf_files_current.py
```
thomcc, line 262: nemesis...

components/viaduct/src/lib.rs
```diff
@@ -19,10 +19,7 @@ pub use headers::{consts as header_names, Header, HeaderName, Headers, InvalidHe
 pub use settings::GLOBAL_SETTINGS;
 
 pub(crate) mod msg_types {
-    include!(concat!(
-        env!("OUT_DIR"),
-        "/mozilla.appservices.httpconfig.protobuf.rs"
-    ));
+    include!("mozilla.appservices.httpconfig.protobuf.rs");
```
thomcc, line 22: any reason not to do these as e.g. `#[path = "mozilla.appservices.httpconfig.protobuf.rs"] pub(crate) mod msg_types;`?

components/fxa-client/src/mozilla.appservices.fxaclient.protobuf.rs
```diff
@@ -0,0 +1,181 @@
+#[derive(Clone, PartialEq, ::prost::Message)]
```
thomcc, line 1: Any chance these can get a `// DO NOT EDIT this file was automatically generated` header on these or similar?

eoger: Good idea, opened https://github.com/danburkert/prost/pull/300.

components/viaduct/src/lib.rs
eoger: Had no idea that was a thing!

eoger: Well the problem is that these files then get formatted by `cargo fmt`, and the `ignore` directive is only on the nightly channel... 😐

PR #2950, opened 2020-04-14
Fix #2902 - Allow FxA profile to refresh even when there exists a profile

### LINKED ISSUE #2902: Allow FxA profile to refresh even when there exists a profile
**Context:** 
Allow the iOS Firefox app to get notified via `.accountProfileUpdate` when we call the `refreshProfile()` in rust components. 

User opens iOS Firefox app and goes to change their picture. 
In the app we call `refreshProfile()` from rust components

**Expectation:** 
we get notified via `.accountProfileUpdate` notificaiton in iOS Firefox 

**Reality:** 
No notification is sent from FxA Rust components 

Ref issue: https://github.com/mozilla-mobile/firefox-ios/issues/6190

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1051)
┆Sprint: Backlog

nbhasin2:
Added.fetchProfile event in case .authenticatedWithProfile,
Updated case .authenticatedWithProfile with Fetch profile logic

Ref Issue: https://github.com/mozilla/application-services/issues/2902


components/fxa-client/ios/FxAClient/FxAccountState.swift
```diff
@@ -86,6 +86,7 @@ extension FxAccountManager {
             }
         case .authenticatedWithProfile:
             switch event {
+            case .fetchProfile: return .authenticatedWithProfile
```
eoger, line 89: In this case we want to stay in `.authenticatedWithProfile`

components/fxa-client/ios/FxAClient/FxAccountManager.swift
```diff
@@ -520,6 +520,15 @@ open class FxAccountManager {
                     )
                 }
             }
+            case .fetchProfile: do {
+                FxALog.info("Fetching profile...")
```
eoger, line 524: Nit: how about "Refreshing profile" instead?

components/fxa-client/ios/FxAClient/FxAccountManager.swift
```diff
@@ -520,6 +520,15 @@ open class FxAccountManager {
...
+                do {
+                    profile = try requireAccount().getProfile()
+                } catch {
+                    return Event.failedToFetchProfile
+                }
+                return Event.fetchedProfile
```
eoger, line 530: There's no transition in `FxAccountState` for this event. I guess we could add a `case .failedToFetchProfile: return .authenticatedNoProfile` OR stay in the current state (after all, we do have a profile). What do you think?

components/fxa-client/ios/FxAClient/FxAccountState.swift
```diff
@@ -86,6 +86,7 @@ extension FxAccountManager {
             }
         case .authenticatedWithProfile:
             switch event {
```
eoger, line 88: You will need the `.fetchedProfile` transition otherwise we'll log an error: 
https://github.com/mozilla/application-services/blob/536a7f8c77430d9daeb68ad79568c048aca7b8bf/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L308-L311

nbhasin2: added fetched profile below:
https://github.com/mozilla/application-services/pull/2950/commits/9b74e288a5ececb3d6e10d5f41a93bb6f073e2a6#diff-61b19ad6432d545017b667a4c94af2ceR90

eoger:
I've rolled up your commits in #3000.

PR #2951, opened 2020-04-14
Cut release v0.55.1

rfk:


rfk:
(The branch name here is incorrect, I think I may have done something wrong with the prepare-release script, but the commit and its history are correct).

PR #2954, opened 2020-04-14
Calculate new release version number from the base branch, not the current checkout [ci smoketest fenix]

rfk:
I recently cut a point-release from a previous version, using the handy `--base-branch` feature to cut from a release branch instead of master. It *mostly* worked, except that it calculated the wrong version number, based on what I had checked out rather than what was in the specified base branch.  This fixes it and also adds some brief docs to guide others who may be doing this task.

### DIFF
--- a/automation/prepare-release.py
+++ b/automation/prepare-release.py
@@ -36,7 +36,18 @@ BUILDCONFIG_VERSION_FIELD = "libraryVersion"
 UNRELEASED_CHANGES_FILE = "CHANGES_UNRELEASED.md"
 CHANGELOG_FILE = "CHANGELOG.md"
 
-# 1. Calculate new version number.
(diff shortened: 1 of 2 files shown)


automation/prepare-release.py
```diff
@@ -36,7 +36,18 @@
...
 
-# 1. Calculate new version number.
+# 1. Create a new branch based on the branch we want to release from.
+
+ensure_working_tree_clean()
+
+step_msg(f"Updating remote {remote}")
+run_cmd_checked(["git", "remote", "update", remote])
+
+release_branch = f"cut-{next_version_full}"
```
eoger, line 46: At this point `next_version_full` is still undefined, how about we create the branch first then rename it with `git branch -M <name>` after we determine the version number?

rfk: good catch; updated

PR #2968, opened 2020-04-15
Make begin_*_flow methods take AsRef<str> params

eoger:
It's making my life difficult, and I think it *should* result in less allocations (not that it matters anyway).


eoger:
> These are big functions and each call with adifferent type would cause duplication, so we should try and factor them so that the input takes AsRef but they immediately call a concrete version (example from stdlib: https://doc.rust-lang.org/src/std/fs.rs.html#266)

Thanks, I feel that would introduce too much complexity for little gains, let's close that.

Issue #2973, opened 2020-04-16
Check-in NSS bindings


eoger:
We've been talking about this for a while with @thomcc : this would make our build process leaner and less prone to errors (no more `clang-sys` compilation). I also think it *might* unblock us on the current NSS desktop upgrade issue we have.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1118)
┆Story Points: 3
┆Sprint: Backlog

PR #2974, opened 2020-04-16
Check-in NSS bindings [ci full]

eoger:
Fixes https://github.com/mozilla/application-services/issues/2973.

Most of the source code in the bindings is copy-pasted from bindgen output. We use the [ctest](https://github.com/gnzlbg/ctest) crate in `systest` to ensure the bindings are correct (and run that on CI through `automation/all_rust_tests.sh`).

My ambition for later would be to publish `nss-sys` on crates.io.


components/support/rc_crypto/nss/nss_sys/src/bindings/secmodt.rs
```diff
@@ -0,0 +1,33 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Opaque pointers as these types are giant.
+pub type PK11SlotInfo = u8;
+pub type PK11SymKey = u8;
+pub type PK11Context = u8;
+
+#[repr(u32)]
+pub enum PK11Origin {
```
thomcc, line 11: Might be worth slapping #[nonexhaustive] on these if you ever plan on publishing them.

PR #2953, opened 2020-04-14
Merge v0.55.1 to master

rfk:
This is https://github.com/mozilla/application-services/pull/2949 and the corresponding v0.55.1 point-release, for merge back to master.

PR #2979, opened 2020-04-16
Uplift v0.55.3 into master

vladikoff:
Ref: https://github.com/mozilla/application-services/pull/2953


vladikoff:
ok so we don’t need this, but we might need something from https://github.com/mozilla/application-services/pull/2953

PR #2980, opened 2020-04-17
Document our decision to encourage, but not enforce, signed commits.

rfk:
In the Kitchen Sync meeting earlier today, we discussed how the current restriction of only landing signed commits on master has proved to be an ongoing pain point without any tangible benefits. The GitHub org security checklist says that requiring signed commits is "strongly encouraged" and that if we don't do it, we should document the reasons why.

So, I'm documenting the reasons why and I will disable the config in github once this documentation change lands on master. If you strongly disagree with this course of action, now's your change to speak up.

(FWIW, I will continue to sign my commits, because I think we can get to a world where the workflow is less annoying and the benefits more tangible, and I'd encourage others to do the same.)

### DIFF
--- a/docs/build-and-publish-pipeline.md
+++ b/docs/build-and-publish-pipeline.md
@@ -37,14 +37,20 @@ For Android consumers these are the steps by which Application Services code bec
 and the integrity-protection mechanisms that apply at each step:
 
 1. Code is developed in branches and lands on `master` via pull request.
-    * GitHub branch protection prevents code being pushed to `master` without review,
-      and requires signed tags (but doesn't check *who* they're signed by).
+    * GitHub branch protection prevents code being pushed to `master` without review.
     * CircleCI and TaskCluster run automated tests against the code, but do not have
       the ability to push modified code back to GitHub thanks to the above branch protection.
       * TaskCluster jobs do not run against PRs opened by the general public,
         only for PRs from repo collaborators.
+    * Contra the [github org security guidelines](https://wiki.mozilla.org/GitHub/Repository_Security),
+      signing of individual commits is encouraged but is **not required**. Our experience in practice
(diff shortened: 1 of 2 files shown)


rfk:
(I am explicitly requesting review from the team members who were not at the meeting and thus did not have an opportunity to object in person).

rfk:
I'll also add a piece of my through process for completeness: if [neqo](https://github.com/mozilla/neqo/commits/master) can get by without requiring signed commits, then so can we.

mhammond:
I don't really object, but fail to understand how an "encourage" policy "help[s] ensure the code comes from a trusted source." in any way whatsoever. Should I infer that projects, including mozilla-central, without the same "encourage" policy, is more likely to have code that doesn't come from a trusted source? Without a clear mitigation policy for unsigned commits, it just seems like theater to me.

rfk:
I don't disagree, but IMHO the previous statement about how "Your patch must be GPG signed to ensure the commits come from a trusted source" was a similar sort of theater, in the absence of anything that actually looks at what keys were used for the signing.

Maybe we just delete that verbiage entirely?

rfk:
>  Without a clear mitigation policy for unsigned commits, it just seems like theater to me.

Yes, and that's where something like https://github.com/mozilla/application-services/issues/2981 might come in.

rfk:
FWIW I changed the contributor note to just "We encourage you to GPG sign your commits", but I wouldn't be upset to just remove it entirely.

mhammond:
Yeah, I'm inclined to agree making them mandatory is a similar kind of theater, which is why I don't object to the policy change itself. I think you can remove that last sentence entirely as you already encourage it earlier in the doc where you justify not requiring it.

rfk:
Going ahead and merging this, it's easy to undo or modify in followups.

PR #2988, opened 2020-04-20
Add mergify to repo

eoger:
We can now have an auto-merging bot in our repository thanks to the removal of the signed commits requirement. Bors is not approved at Mozilla but Mergify seems ok, so how about we give it a try?

Automatic merge will happen if all the following conditions are true:
- There's at least 1 approved review
- For non-team members: approved reviews are not stale
- The "checkin-needed" label is present
- All the CI jobs have passed

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,69 @@
+pull_request_rules:
+  - name: remove outdated reviews for non-core authors
+    conditions:
+      - base=master
+      - author!=@mozilla/application-services
+    actions:
+      dismiss_reviews:
+        message: The pull request has been modified, dismissing previous reviews.
+      label:
+        remove:
+          - checkin-needed
+  - name: automatic merge for master
+    conditions:
+      - "#approved-reviews-by>=1"
+      - base=master
+      - label=checkin-needed
+      - "status-success=android-build-pr"
+      - "status-success=lint-detekt"
+      - "status-success=lint-ktlint"
+      - "status-success=ci/circleci: Check Protobuf files are up-to-date"
+      - "status-success=ci/circleci: Check Rust dependencies"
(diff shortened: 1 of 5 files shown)


.mergify.yml
```diff
@@ -0,0 +1,69 @@
+pull_request_rules:
+  - name: remove outdated reviews for non trusted authors
```
rfk, line 2: Extremely minor nit: could we name this as "external contributors" or "non-core contributors" or something that sounds less judgmental about the level of trust involved? :-)

.mergify.yml
```diff
@@ -0,0 +1,69 @@
...
+      - "status-success=ci/circleci: Check Protobuf files are up-to-date"
+      - "status-success=ci/circleci: Check Rust dependencies"
+      - "status-success=ci/circleci: Check Rust formatting"
+      - "status-success=ci/circleci: Check Swift formatting"
+      - "status-success=ci/circleci: Lint Bash scripts"
+      - "status-success=ci/circleci: Lint Rust with clippy"
+      - "status-success=ci/circleci: Rust benchmarks"
+      - "status-success=ci/circleci: Rust tests"
+      - "status-success=ci/circleci: Sync integration tests"
+      - "status-success=ci/circleci: iOS build and test"
```
rfk, line 29: Github appears to have eaten my comment, re-posting:

If we add another suite of checks to CI, do we need to remember to come here and add it to this file as well? Seems OK to start but I'd worry about it getting out of sync over the long haul.

eoger: Yes, sadly ["There is no such thing as “every status check” in GitHub."](https://doc.mergify.io/conditions.html#validating-all-status-check). I'll add some warnings to our CI files.

PR #2991, opened 2020-04-20
Derive Serialize for AccountEvent and IncomingDeviceCommand

eoger:
Can't serialize to JSON these structs on moz-central. I've also derived `Debug` coz it's always useful.

### DIFF
--- a/components/fxa-client/src/lib.rs
+++ b/components/fxa-client/src/lib.rs
@@ -239,20 +239,29 @@ impl FirefoxAccount {
     }
 }
 
+#[derive(Debug, Serialize)]
+#[serde(tag = "eventType", content = "data")]
+#[serde(rename_all = "camelCase")]
 pub enum AccountEvent {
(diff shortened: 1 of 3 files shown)


components/fxa-client/src/lib.rs
```diff
@@ -239,6 +239,8 @@ impl FirefoxAccount {
     }
 }
 
+#[derive(Debug, Serialize)]
+#[serde(tag = "event_type", content = "data")]
```
linacambridge, line 243: Do you want to `rename_all = "camelCase"`, too, or does that only affect the variant names? (So we'll have `{ event_type: "IncomingDeviceCommand", data: { command_type: "TabReceived", data: { sender: ... } } }` instead of `incoming_device_command` and `tab_received`)?

eoger: That's a great idea! I'll add that :)

Issue #2998, opened 2020-04-21
RemoteConfig should not be persisted


eoger:
I lied in https://github.com/mozilla/application-services/blob/536a7f8c77430d9daeb68ad79568c048aca7b8bf/components/fxa-client/src/config.rs#L31-L32

I think we should still do what it says, so clients can always pick up the newest config (and it's a very cheap network call, both for the server and the clients).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1141)
┆Sprint: Backlog

rfk:
+1; I think we also have a bug on file somewhere called "discuss whether RemoteConfig should be persisted" that can also be closed by this.

PR #2999, opened 2020-04-21
Config improvements

eoger:
Fixes https://github.com/mozilla/application-services/issues/2998.
This should make the code a bit easier to maintain.


components/fxa-client/src/config.rs
```diff
@@ -19,6 +19,7 @@ struct ClientConfigurationResponse {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct OpenIdConfigurationResponse {
     authorization_endpoint: String,
+    introspection_endpoint: String,
```
rfk, line 22: Some very old self-hosted FxA servers might not report this value. Probably fine not to worry about supporting that, but we might think about whether that would show up cleanly in client logs in a way that would help us identify the source of the problem.

components/fxa-client/src/state_persistence.rs
```diff
@@ -280,7 +281,7 @@ mod tests {
     fn test_v2_ignores_unknown_fields_introduced_by_future_changes_to_the_schema() {
         // This is a snapshot of what some persisted StateV2 data would look before any backwards-compatible changes
         // were made. It's very important that you don't modify this string, which would defeat the point of the test!
```
rfk, line 283: Per the comment here, it's very important that you don't modify this string. There will be clients in the wild with a string that looks like this and we don't want to accidentally break support for them.

(I guess technically *removing* fields from the string won't affect the stated purpose of the test, which is to check whether new fields get correctly default to none. But I think it violates the spirit of the test. You could consider copying it into another test named something like `test_v2_ignores_unknown_fields_removed_in_future_changes_to_the_schema` but I don't think we should just remove it entirely)

components/fxa-client/src/state_persistence.rs
```diff
@@ -291,7 +292,7 @@ mod tests {
...
+        let state_v2_json = "{\"schema_version\":\"V2\",\"config\":{\"client_id\":\"98adfa37698f255b\",\"redirect_uri\":\"https://lockbox.firefox.com/fxa/ios-redirect.html\",\"content_url\":\"https://accounts.firefox.com\"},\"refresh_token\":{\"token\":\"bed5532f4fea7e39c5c4f609f53603ee7518fd1c103cc4034da3618f786ed188\",\"scopes\":[\"https://identity.mozilla.com/apps/oldysnc\"]},\"scoped_keys\":{\"https://identity.mozilla.com/apps/oldsync\":{\"kty\":\"oct\",\"scope\":\"https://identity.mozilla.com/apps/oldsync\",\"k\":\"kMtwpVC0ZaYFJymPza8rXK_0CgCp3KMwRStwGfBRBDtL6hXRDVJgQFaoOQ2dimw0Bko5WVv2gNTy7RX5zFYZHg\",\"kid\":\"1542236016429-Ox1FbJfFfwTe5t-xq4v2hQ\"}},\"login_state\":{\"Unknown\":null}}";
```
rfk, line 295: This one seems fine to change though :-)

components/fxa-client/src/config.rs
eoger: I believe `serde_json` will generate a helpful message saying the specific field is missing, which we'll bubble up to consumers.

rfk:
I took the liberty of adding a `checkin-needed` here, let's see how it goes.

PR #3000, opened 2020-04-21
Expose forceRefresh to refreshProfile [ci skip]

eoger:
This should allow https://github.com/mozilla/application-services/pull/2950 to make the homerun for https://github.com/mozilla/application-services/issues/2902 ⚾


eoger:
`FxAccount.swift` is not considered "public" API (I think it should not show for our users), I'll add a changelog entry tho!

PR #3007, opened 2020-04-22
Remove `nss_build_common` from extension storage.

linacambridge:
I noticed this while vendoring extension storage into m-c. Extension storage doesn't use NSS or SQLCipher, so I don't think we need this build script.


thomcc:
https://github.com/mozilla/application-services/issues/2882

The fact that we don’t use NSS in this project is why we need this hack :(

linacambridge:
Ahaaa, got it! Thanks so much for explaining, sorry to rehash it! 😅

rfk:
(I think this issue is a signal that we could use some more/better comments around this in the `build.rs` file 😁)

Issue #2625, opened 2020-02-11
Avoid fetching the device list multiple times in quick succession


rfk:
I instrumented a local Fenix build to log all network activity, then poked around in my account settings to see what would happen. I wasn't able to receive tab via push because I didn't have firebase configured in my local build, but here's the network requests involved in me receiving a sent tab via "sync now":

```
02-11 19:51:16.905 19047 19086 W GeckoViewFetchClient: GV FETCH GET https://api.accounts.firefox.com/v1/account/devices
02-11 19:51:17.545 19047 19086 W GeckoViewFetchClient: GV FETCH RESPONSE GET https://api.accounts.firefox.com/v1/account/devices 200
02-11 19:51:17.555 19047 19086 I FxaDeviceConstellation: Executing: polling for device events
02-11 19:51:17.555 19047 19086 W GeckoViewFetchClient: GV FETCH GET https://api.accounts.firefox.com/v1/account/device/commands?index=8310664
02-11 19:51:17.805 19047 19086 W GeckoViewFetchClient: GV FETCH RESPONSE GET https://api.accounts.firefox.com/v1/account/device/commands?index=8310664
200 02-11 19:51:17.805 19047 19086 W GeckoViewFetchClient: GV FETCH GET https://api.accounts.firefox.com/v1/account/devices
02-11 19:51:18.015 19047 19086 W GeckoViewFetchClient: GV FETCH RESPONSE GET https://api.accounts.firefox.com/v1/account/devices 200                       
```

Note that this requests the device list twice in quick succession - once when asked to explicitly by the app in its "sync now" logic, and again, and a second time as part of (I assume) the handling of the received tab.

We should consider some sort of caching/debouncing logic similar to what we have in Desktop, to avoid fetching the device-list multiple times in quick succession.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/SYNC-815)
┆Story Points: 3
┆Epic: FxA crate in desktop
┆Sprint: SYNC - end 2020-05-08

rfk:
A similar thing happens when sending a tab from Fenix - we fetch the devices list once when rendering the "share panel", an again a short time later when actually sending the tab to a specific device.

rfk:
There's probably a nice chunky refactor we could do here to pass the device list around rather than fetching it in multiple places, but we're unlikely to prioritize that anytime soon.

A more minimal fix for this issue would be to add some debouncing logic to the `get_devices` method here:

[...]

PR #3013, opened 2020-04-22
Add List Functionality for Attached Clients

### LINKED ISSUE #2898: Add support for "listAttachedOAuthClients” in Rust
Currently we seem to be missing the following "listAttachedOAuthClients" method https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#453 that a lot of Desktop component use in various places

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1049)
┆Story Points: 3
┆Epic: FxA crate in desktop

lougeniaC64:
Fixes #2898

### DIFF
--- a/components/fxa-client/src/http_client.rs
+++ b/components/fxa-client/src/http_client.rs
@@ -91,6 +91,11 @@ pub trait FxAClient {
         update: DeviceUpdateRequest<'_>,
     ) -> Result<UpdateDeviceResponse>;
     fn destroy_device(&self, config: &Config, refresh_token: &str, id: &str) -> Result<()>;
+    fn attached_clients(
+        &self,
+        config: &Config,
+        session_token: &str,
+    ) -> Result<Vec<GetAttachedClientResponse>>;
     fn scoped_key_data(
         &self,
         config: &Config,
@@ -341,6 +346,17 @@ impl FxAClient for Client {
 
         Self::make_request(request)?;
         Ok(())
+    }
+
+    fn attached_clients(
+        &self,
+        config: &Config,
+        session_token: &str,
+    ) -> Result<Vec<GetAttachedClientResponse>> {
+        let url = config.auth_url_path("v1/account/attached_clients")?;
+        let key = derive_auth_key_from_session_token(session_token)?;
+        let request = HawkRequestBuilder::new(Method::Get, url, &key).build()?;
+        Ok(Self::make_request(request)?.json()?)
     }
 
     fn scoped_key_data(
@@ -657,6 +673,23 @@ pub struct DeviceResponseCommon {
     pub available_commands: HashMap<String, String>,
     #[serde(rename = "pushEndpointExpired")]
     pub push_endpoint_expired: bool,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetAttachedClientResponse {
+    pub client_id: Option<String>,
+    pub session_token_id: Option<String>,
+    pub refresh_token_id: Option<String>,
+    pub device_id: Option<String>,
+    pub device_type: Option<DeviceType>,
+    pub is_current_session: bool,
+    pub name: Option<String>,
+    pub created_time: Option<u64>,
+    pub last_access_time: Option<u64>,
+    pub scope: Option<Vec<String>>,
+    pub user_agent: String,
+    pub os: Option<String>,
 }
 
 // We model the OAuthTokenRequest according to the up to date
--- a/components/fxa-client/src/lib.rs
+++ b/components/fxa-client/src/lib.rs
@@ -61,6 +61,7 @@ pub struct FirefoxAccount {
     client: Arc<FxAClient>,
     state: State,
     flow_store: HashMap<String, OAuthFlow>,
+    attached_clients_cache: Option<CachedResponse<Vec<http_client::GetAttachedClientResponse>>>,
 }
 
 impl FirefoxAccount {
@@ -69,6 +70,7 @@ impl FirefoxAccount {
             client: Arc::new(http_client::Client::new()),
             state,
             flow_store: HashMap::new(),
+            attached_clients_cache: None,
         }
     }
 
@@ -137,6 +139,7 @@ impl FirefoxAccount {
     pub fn start_over(&mut self) {
         self.state = self.state.start_over();
         self.flow_store.clear();
+        self.attached_clients_cache = None;
     }
 
     /// Get the Sync Token Server endpoint URL.
--- a/components/fxa-client/src/oauth/attached_clients.rs
+++ b/components/fxa-client/src/oauth/attached_clients.rs
@@ -1,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub use crate::http_client::GetAttachedClientResponse as AttachedClient;
+use crate::{error::*, util, CachedResponse, FirefoxAccount};
+
+// An attached clients response is considered fresh for `ATTACHED_CLIENTS_FRESHNESS_THRESHOLD` ms.
+const ATTACHED_CLIENTS_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
+
+impl FirefoxAccount {
+    pub fn get_attached_clients(&mut self, session_token: &str) -> Result<Vec<AttachedClient>> {
+        if let Some(a) = &self.attached_clients_cache {
+            if util::now() < a.cached_at + ATTACHED_CLIENTS_FRESHNESS_THRESHOLD {
+                return Ok(a.response.clone());
+            }
+        }
+        let response = self
+            .client
+            .attached_clients(&self.state.config, session_token)?;
+
+        self.attached_clients_cache = Some(CachedResponse {
+            response: response.clone(),
+            cached_at: util::now(),
+            etag: "".into(),
+        });
+
+        Ok(response)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        config::Config,
+        http_client::{DeviceType, FxAClientMock},
+    };
+    use std::sync::Arc;
+
+    #[test]
+    fn test_get_attached_clients() {
+        let config = Config::stable_dev("12345678", "https://foo.bar");
+        let mut fxa = FirefoxAccount::with_config(config);
+
+        let mut client = FxAClientMock::new();
+        client
+            .expect_attached_clients(mockiato::Argument::any, |arg| arg.partial_eq("session"))
+            .times(1)
+            .returns_once(Ok(vec![AttachedClient {
+                client_id: Some("12345678".into()),
+                session_token_id: None,
+                refresh_token_id: None,
+                device_id: None,
+                device_type: Some(DeviceType::Desktop),
+                is_current_session: true,
+                name: None,
+                created_time: None,
+                last_access_time: None,
+                scope: None,
+                user_agent: "attachedClientsUserAgent".into(),
+                os: None,
+            }]));
+
+        fxa.set_client(Arc::new(client));
+        assert!(fxa.attached_clients_cache.is_none());
+
+        let res = fxa.get_attached_clients("session");
+
+        assert!(res.is_ok());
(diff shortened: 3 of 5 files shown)


components/fxa-client/src/http_client.rs
```diff
@@ -660,6 +706,35 @@ pub struct DeviceResponseCommon {
     pub push_endpoint_expired: bool,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
vladikoff, line 709: @lougeniaC64 would https://serde.rs/attr-rename.html help here ?

components/fxa-client/src/http_client.rs
```diff
@@ -344,6 +353,43 @@ impl FxAClient for Client {
         Ok(())
     }
 
+    fn attached_clients(&self, config: &Config, refresh_token: &str) -> Result<Vec<GetAttachedClientResponse>> {
+        let url = config.auth_url_path("v1/account/attached_clients")?;
+        let request =
+            Request::get(url).header(header_names::AUTHORIZATION, bearer_token(refresh_token))?;
+        let mut attached_clients: Vec<GetAttachedClientResponse> = Self::make_request(request)?.json()?;
+
+        for mut attached_client in &mut attached_clients {
```
eoger, line 362: If possible I'd make that kind of post-processing outside of the `http_client` module

components/fxa-client/src/http_client.rs
```diff
@@ -660,6 +706,35 @@ pub struct DeviceResponseCommon {
     pub push_endpoint_expired: bool,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct GetAttachedClientResponse {
+    #[serde(rename = "clientId")]
+    pub client_id: String,
```
eoger, line 712: You can avoid renaming most of these fields with `#[serde(rename_all = "camelCase")]` on the struct.

components/fxa-client/src/http_client.rs
lougeniaC64: @eoger I moved _most_ of the post-processing logic to `util.rs` but not all because I didn't want to create circular dependencies between `util` and `http_client`. If this isn't what you had in mind, please let me know. Not sure if there's a better place to put all of this.

eoger: Thanks! Sorry to be annoying, but could we also move the whole loop in the caller?

components/fxa-client/src/http_client.rs
```diff
@@ -660,6 +695,25 @@ pub struct DeviceResponseCommon {
...
+    pub session_token_id: Option<String>,
+    pub refresh_token_id: Option<String>,
+    pub device_id: Option<String>,
+    pub device_type: Option<DeviceType>,
+    pub is_current_session: bool,
+    pub name: Option<String>,
+    pub created_time: Option<u64>,
+    pub last_access_time: Option<u64>,
+    pub last_accessed_days_ago: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
eoger, line 711: We can probably remove this directive.

components/fxa-client/src/http_client.rs
eoger: Thinking about this more, how about we let go of all that post-processing code and let the caller decide what it wants to do with the `last_access_time` time. It seems wrong to me to make assumptions here.

lougeniaC64: Yeah, I'll do that. I wasn't completely convinced that mimicking the logic that's [in desktop](https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#464) was worthwhile.

components/fxa-client/src/http_client.rs
```diff
@@ -344,6 +357,54 @@ impl FxAClient for Client {
...
+    }
+
+    fn destroy_attached_client(
+        &self,
+        config: &Config,
+        refresh_token: &str,
+        client_id: &str,
+        session_token_id: Option<String>,
+        device_id: Option<String>,
+    ) -> Result<()> {
```
lougeniaC64, line 387: According to [the docs](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/api.md#post-accountattached_clientdestroy), `refresh_token`, `client_id`, `session_token_id`, and `device_id` are optional.  I've made a call here to balance functionality and safety but it may not be the right one. Let me know if that's the case.

components/fxa-client/src/http_client.rs
```diff
@@ -91,6 +91,19 @@ pub trait FxAClient {
         update: DeviceUpdateRequest<'_>,
     ) -> Result<UpdateDeviceResponse>;
     fn destroy_device(&self, config: &Config, refresh_token: &str, id: &str) -> Result<()>;
+    fn attached_clients(
+        &self,
+        config: &Config,
+        refresh_token: &str,
+    ) -> Result<ResponseAndETag<Vec<GetAttachedClientResponse>>>;
+    fn destroy_attached_client(
```
eoger, line 99: Should we expose this function?

components/fxa-client/src/attached_client.rs
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub use crate::http_client::GetAttachedClientResponse as AttachedClient;
+use crate::{error::*, util, CachedResponse, FirefoxAccount};
+
+// An attached client response is considered fresh for `ATTACHED_CLIENTS_FRESHNESS_THRESHOLD` ms.
+const ATTACHED_CLIENTS_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
+
+impl FirefoxAccount {
```
eoger, line 11: It's a good idea putting this impl in a different module. One nit though, could we put this in the `oauth::attached_client` namespace since it is OAuth related?

components/fxa-client/src/lib.rs
```diff
@@ -61,6 +62,7 @@ pub struct FirefoxAccount {
     client: Arc<FxAClient>,
     state: State,
     flow_store: HashMap<String, OAuthFlow>,
+    attached_clients: Option<CachedResponse<Vec<http_client::GetAttachedClientResponse>>>,
```
eoger, line 65: Do we want to store this in memory or persist it?
- If in memory is fine: we should rename this to `attached_clients_cache` to make this clear.
- If we want to persist this: this needs to be moved in `V2State`

components/fxa-client/src/http_client.rs
```diff
@@ -343,6 +356,54 @@ impl FxAClient for Client {
...
+        &self,
+        config: &Config,
+        refresh_token: &str,
+    ) -> Result<ResponseAndETag<Vec<GetAttachedClientResponse>>> {
+        let url = config.auth_url_path("v1/account/attached_clients")?;
+        let request =
+            Request::get(url).header(header_names::AUTHORIZATION, bearer_token(refresh_token))?;
+        let resp = Self::make_request(request)?;
+        let attached_clients: Vec<GetAttachedClientResponse> = resp.json()?;
+        let etag = resp
```
eoger, line 369: Does this endpoint work with `etags`? I couldn't find anything in the FxA API docs.

components/fxa-client/src/lib.rs
```diff
@@ -557,4 +562,91 @@ mod tests {
             "https://firefox.com/pair"
         )
     }
+
+    #[test]
```
eoger, line 566: Nice 😎 You should probably move this in `attached_clients.rs`.

components/fxa-client/src/http_client.rs
lougeniaC64: Yikes! I made an assumption without checking. Thanks for this catch.

components/fxa-client/src/lib.rs
lougeniaC64: Ryan's notes from #2625 suggesting starting with caching in memory so I'll rename that property.

components/fxa-client/src/http_client.rs
eoger: @rfk might have an opinion here as I don't?

components/fxa-client/src/attached_client.rs
lougeniaC64: @eoger before I do anything unnecessary, did you mean to move this logic to oauth.rs as static functions or within the pre-existing FirefoxAccount implementation? I was initially thinking to move attached_client.rs to a new oauth folder but I forgot about the existing oauth.rs file and module.

components/fxa-client/src/http_client.rs
lougeniaC64: I'm going to add him as a reviewer since I'm implementing the debouncing logic he recommended for the get devices functionality.

components/fxa-client/src/attached_client.rs
eoger: > @eoger I was initially thinking to move attached_client.rs to a new oauth folder but I forgot about the existing oauth.rs file and module.

Yup that's what I was thinking too!
I believe creating an `oauth/attached_client.rs` file "should" work since there's no need for `mod.rs` files anymore in rust 2018 edition.

lougeniaC64: Oh, nice--okay, thanks!!

components/fxa-client/src/lib.rs
```diff
@@ -61,6 +61,7 @@ pub struct FirefoxAccount {
     client: Arc<FxAClient>,
     state: State,
     flow_store: HashMap<String, OAuthFlow>,
+    attached_clients_cache: Option<CachedResponse<Vec<http_client::GetAttachedClientResponse>>>,
```
rfk, line 64: Not sure how it will work put in the code, but we should consider clearing this cache in response to `DeviceConnected` and `DeviceDisconnected` push notifications.

components/fxa-client/src/http_client.rs
lougeniaC64: Thanks for this feedback. I'm okay with removing the destroy functionality. @vladikoff does that work for you? I think we discussed adding that functionality to this ticket.

vladikoff: @lougeniaC64 Yes that's fine, I double checked and we won't need it in there . Let's remove the destroy functionality

components/fxa-client/src/lib.rs
lougeniaC64: @rfk I made changes in `push.rs`. Let me know if this is what you had in mind.

eoger:
(also don't forget to squash before merging!)

components/fxa-client/src/http_client.rs
```diff
@@ -91,6 +91,11 @@ pub trait FxAClient {
         update: DeviceUpdateRequest<'_>,
     ) -> Result<UpdateDeviceResponse>;
     fn destroy_device(&self, config: &Config, refresh_token: &str, id: &str) -> Result<()>;
+    fn attached_clients(
+        &self,
+        config: &Config,
+        refresh_token: &str,
```
rfk, line 97: the signature here still says `refresh_token` (which I guess doesn't cause compilation errors because it's the correct *type* just with the wrong name).

components/fxa-client/src/oauth/attached_clients.rs
```diff
@@ -0,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub use crate::http_client::GetAttachedClientResponse as AttachedClient;
+use crate::{error::*, util, CachedResponse, FirefoxAccount};
+
+// An attached client response is considered fresh for `ATTACHED_CLIENTS_FRESHNESS_THRESHOLD` ms.
```
rfk, line 8: nit: "attached clients" plural please, for consistency

components/fxa-client/src/push.rs
```diff
@@ -35,6 +35,7 @@ impl FirefoxAccount {
                 Ok(vec![AccountEvent::ProfileUpdated])
             }
             PushPayload::DeviceConnected(DeviceConnectedPushPayload { device_name }) => {
+                self.attached_clients_cache = None;
```
rfk, line 38: 👍 ; we'll probably find we reach a point where want to abstract this away a little more, so that the "update X state on Y push notification" logic lives a little closer to the rest of the code that manages X state, but it's tractable doing it directly here for now.

PR #3031, opened 2020-04-23
Let autoPublish script run sucessfully if `local.properties` is missing.

rfk:
Like it says on the tin.


rfk:
Looking at this, I should probably mix the names of the files in the hash here as well for completeness, but we're going for "low probability of accidental hash collision" not "resistant to deliberate attack" here so I think it'll be OK.

PR #3034, opened 2020-04-24
Bug 1563169 - add Taskcluster support for master pushes

MihaiTabara:
I ran this locally via [taskgraph-diff](https://hg.mozilla.org/ci/taskgraph/file/tip/README.rst#l65) and the diffs look good. Doesn't touch at all PRs and adds the same sets of jobs that we define for master pushes.

So this should change nothing in master pushes and PR, but the way the tasks are being scheduled, via taskgraph. Once this happens, the docker-image should no longer be scheduled in the PRs, as well as the toolchains recreation since level-3 tasks will be cached in level-1 as well.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -55,6 +55,7 @@ tasks:
           else:
               $if: 'tasks_for == "github-push"'
               then: '${event.pusher.email}'
+              # Assume Pull Request
               else:
                   $if: 'tasks_for == "github-pull-request"'
                   then: '${event.pull_request.user.login}@users.noreply.github.com'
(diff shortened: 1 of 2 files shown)


.taskcluster.yml
```diff
@@ -197,142 +198,156 @@ tasks:
               taskclusterProxy: true
       in:
         $match:
-          "tasks_for == 'github-pull-request' && event['action'] in ['opened', 'reopened', 'edited', 'synchronize']":
+          "(tasks_for == 'github-pull-request' && event['action'] in ['opened', 'reopened', 'synchronize']) || (tasks_for == 'github-push' && head_branch[:10] != 'refs/tags/')":
```
MihaiTabara, line 201: I removed the `edited` action from here so that we don't retrigger the CI on Pull-requests in case we changed the title or alike.

The `github-push` section which was slightly below based on the old automation is now based on taskgraph here.

.taskcluster.yml
```diff
@@ -197,142 +198,156 @@ tasks:
...
             $let:
               level:
                 $if: 'tasks_for in ["github-push", "github-release"] && repoUrl == "https://github.com/mozilla/application-services"'
                 then: '3'
                 else: '1'
+              short_head_branch:
```
MihaiTabara, line 207: Defined as a variable because it's used in both `routes` and `scopes`. Backported from how we handle the other mobile repos actually 👍

.taskcluster.yml
```diff
@@ -197,142 +198,156 @@ tasks:
...
-              schedulerId: '${trustDomain}-level-${level}'
-              created: {$fromNow: ''}
-              deadline: {$fromNow: '1 day'}
-              expires: {$fromNow: '1 year 1 second'}  # 1 second so artifacts expire first, despite rounding errors
-              metadata:
-                $merge:
-                  - owner: "${ownerEmail}"
-                    source: '${repoUrl}/raw/${head_sha}/.taskcluster.yml'
-                  - $if: 'tasks_for in ["github-push", "github-pull-request", "github-release"]'
+              $mergeDeep:
```
MihaiTabara, line 211: Mostly changed indentation, hence all these changes below.

.taskcluster.yml
```diff
@@ -197,142 +198,156 @@ tasks:
...
+                    $flattenDeep:
+                      - checks
+                      - $if: 'level == "3"'
+                        then:
+                          - tc-treeherder.v2.${project}.${head_sha}
+                          # TODO Bug 1601928: Make this scope fork-friendly once ${project} is better defined. This will enable
+                          # staging release promotion on forks.
+                          - $if: 'tasks_for == "github-push"'
+                            then:
+                              - index.project.${project}.v2.branch.${short_head_branch}.latest.taskgraph.decision
```
MihaiTabara, line 247: @JohanLorenzo started curating these indexes in https://bugzilla.mozilla.org/show_bug.cgi?id=1631839 but I'll address that in a follow-up once releases are ready, to not add another possible source of failure in this long-overdue migration.

.taskcluster.yml
```diff
@@ -197,142 +198,156 @@ tasks:
...
-                $merge:
-                    - treeherder:
-                          $merge:
-                              - machine:
-                                    platform: gecko-decision
-                              - $if: 'tasks_for in ["github-push", "github-pull-request"]'
-                                then:
-                                    symbol: D
-                    - tasks_for: '${tasks_for}'
-          "tasks_for == 'github-push' && head_branch == 'refs/heads/master'":
```
MihaiTabara, line 328: This section dissapears in favor the `taskgraph`, as pull-requests.

PR #3040, opened 2020-04-26
Downgrade `url` version to 2.1.

linacambridge:
`url@2.1.1` contains backward-incompatible changes that break Firefox
builds. We should fix those failures and get Firefox on the new version,
but, for now, it's easier to downgrade our version in `webext-storage`.

Firefox bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1633289


linacambridge:
I've re-run `mach vendor rust` pointing it to `rev = "a5d61293a71a72a8e741ea5c666f6ea9aaf07513"`, and verified it doesn't bump `url`.

Issue #3014, opened 2020-04-23
webext-storage: Add explicit method to close the database


mhammond:
While the DB will be closed when the store is dropped, we might as well expose an explicit close method - it will also let us capture any errors during a close(), which could be useful for diagnostics.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1153)
┆Story Points: 2
┆Sprint: SYNC - end 2020-05-08

mhammond:
So `.close()` consumes the connection, which (iiuc), pretty much means it will also need to consume the store etc too (which TBH makes the most sense from an API perspective too). However, I don't quite see how this would map to the bridge in m-c, other than by swapping the OnceCell for a regular cell - @linacambridge WDYT?

mhammond:
Actually, the problem is closer to home than m-c - `Arc<Mutex<Connection>>` needs to be consumed and I'm not sure that's possible without a wrapping `Option<>`?

linacambridge:
Thanks very much, @mhammond!

I can do this tomorrow, since I've been poking around in there anyway. You're right that we need a `RefCell<Option<...>>` in there, but that can live on the m-c side of the bridge. Its `teardown` method already takes ownership of the store, so all we have to do is add the call to our `close` method.

Thinking about it even _more_, I also don't think we need an `Arc<Mutex<Connection>>`, since our `StorageSyncArea` wrapper already holds an `Arc<LazyStore>`. So we can have it be just `Mutex<Connection>`. This also means we don't have to use `Arc::try_unwrap` to get at the `Connection`, since we don't have to worry about other owning references.

But wait, there's more! 😆 Our `ConcurrentHandleMap` for the FFI _already_ wraps its contents in a mutex. So, in our FFI code, when we write `static ref STORES: ConcurrentHandleMap<Store> = ConcurrentHandleMap::new()`, it's actually defining a `RwLock<HandleMap<Mutex<Store>>>`. That means we don't need our `StoreDb` to have a mutex, either, because the handle map already makes one!

We'll need to change Desktop's `LazyStore` slightly to hold on to a `OnceCell<Mutex<Store>>`, and change its `get` method to return a `MutexGuard<'_, Store>` instead of a `&Store`...but that's a four-line change! 😄

So, in conclusion, we can:

1. Change `StorageDb` to hold a `writer: Connection`. We don't need the `Arc<Mutex<...>>` wrapping.
[...]

PR #3042, opened 2020-04-27
Simplify the extension storage DB, add docs and explicit teardown

linacambridge:
The individual commits have more details in them, but the skinny is that this is everything in https://github.com/mozilla/application-services/issues/3014#issuecomment-618844344.

### DIFF
--- a/components/webext-storage/src/db.rs
+++ b/components/webext-storage/src/db.rs
@@ -9,18 +9,19 @@ use rusqlite::Connection;
 use rusqlite::OpenFlags;
 use sql_support::ConnExt;
 use std::fs;
+use std::ops::{Deref, DerefMut};
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex};
 use url::Url;
 
-/// The entry-point to getting a database connection. No enforcement of this
-/// as a singleton is made - that's up to the caller. If you make multiple
-/// StorageDbs pointing at the same physical database, you are going to have a
-/// bad time. We only support a single writer connection - so that's the only
-/// thing we store. It's still a bit overkill, but there's only so many yaks
-/// in a day.
+/// A `StorageDb` wraps a read-write SQLite connection, and handles schema
+/// migrations and recovering from database file corruption. It can be used
+/// anywhere a `rusqlite::Connection` is expected, thanks to its `Deref{Mut}`
+/// implementations.
+///
+/// We only support a single writer connection - so that's the only thing we
+/// store. It's still a bit overkill, but there's only so many yaks in a day.
 pub struct StorageDb {
-    pub writer: Arc<Mutex<Connection>>,
+    writer: Connection,
 }
 impl StorageDb {
     /// Create a new, or fetch an already open, StorageDb backed by a file on disk.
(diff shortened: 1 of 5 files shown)


components/webext-storage/src/db.rs
```diff
@@ -62,6 +62,37 @@ impl StorageDb {
...
+        Ok(self.writer.close().map_err(|(writer, err)| {
+            // Failing to close the connection means we have unfinalized
+            // statements. Since we're likely being called as the app is
+            // shutting down, anyway, we leak the connection here, and rely on
+            // the OS to clean it up. Note that the connection's destructor will
+            // try to close it again, and panic if it can't. This is fine for
+            // the FFI, but not Desktop, which is compiled with
+            // `panic = "abort"`. To avoid a shutdown crash, we use `forget`
+            // instead of `drop`.
+            mem::forget(writer);
```
linacambridge, line 78: @thomcc, I'd love your feedback on this especially! ISTM a leak at shutdown (when resources will get reclaimed momentarily, anyway) is much better than a crash—are there consequences to forgetting a `rusqlite::Connection` like this, though? The FFI won't use this method, it'll go through the destructor, so that should panic if closing fails...but your `ConcurrentHandleMap` handles that perfectly.

I also thought about putting the `forget` in Desktop...but then I'd have to change the signature to return `Result<(), (StorageDb, Error)>`, and bubble that up in `Store`...and it feels like a lot of clunky code for (what we hope 🤞) is an edge case.

components/webext-storage/src/store.rs
```diff
@@ -9,55 +9,93 @@ use std::path::Path;
...
-        let mut conn = self.db.writer.lock().unwrap();
-        let tx = conn.transaction()?;
+        let tx = self.db.unchecked_transaction()?;
         let result = api::clear(&tx, ext_id)?;
         tx.commit()?;
         Ok(result)
     }
+
+    /// Tears down the store, closing its underlying database connection.
+    pub fn teardown(self) -> Result<()> {
```
linacambridge, line 96: If you're wondering where this will get called... 😊 [Here!](https://hg.mozilla.org/mozilla-central/file/tip/toolkit/components/extensions/storage/webext_storage_bridge/src/store.rs#l50)

components/webext-storage/src/db.rs
linacambridge: Actually, I changed my mind about this. There's no guarantee that `teardown` will be called—if Desktop drops its `StorageSyncArea` (which is holding on to a `LazyStore`, which holds on to a `webext_storage::Store`), the connection will get closed via `InnerConnection::drop`, not `teardown`, causing a panic. Let's change this to use `mem::drop` instead and make it less surprising; we can always change back if we see shutdown crashes.

linacambridge: But then, if we rely on `InnerConnection::drop` to close the connection...we don't really need an explicit `teardown` or `close` method, since dropping our last reference to the store should be enough. So we can remove the whole second commit and close #3014 as "won't fix". ✨

mhammond: Isn't it an xpcom component holding the reference? If so, "dropping the last reference" can often be more difficult than you think it should be :)

PR #3043, opened 2020-04-27
First cut at an FFI for `storage.sync`

linacambridge:
This isn't wired up to any Swift or Kotlin bindings yet, but I made a
skeleton while working on #3042, and it didn't take much more work to
finish. Hooray for small API surfaces!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,9 +11,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccc9a9dd069569f212bc4330af9f17c4afb5e8ce185e83dbb14f1349dda18b10"
+checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"
 
 [[package]]
 name = "aho-corasick"
@@ -53,9 +53,9 @@ checksum = "cff77d8686867eceff3105329d4698d96c2391c176d5d03adc90c7389162b5b8"
 
 [[package]]
 name = "assert-json-diff"
-version = "1.0.3"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c356497fd3417158bcb318266ac83c391219ca3a5fa659049f42e0041ab57d6"
+checksum = "4259cbe96513d2f1073027a259fc2ca917feb3026a5a8d984e3628e490255cc0"
 dependencies = [
  "extend",
  "serde",
@@ -1613,9 +1613,9 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "pkg-config"
-version = "0.3.17"
+version = "0.3.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05da548ad6865900e60eaba7f589cc0783590a92e940c26953ff81ddbab2d677"
+checksum = "d36492546b6af1463394d46f0c834346f31548646f6ba10849802c9c9a27ac33"
 
 [[package]]
 name = "places"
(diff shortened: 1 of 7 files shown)


components/webext-storage/ffi/src/error_codes.rs
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// An unexpected error occurred which likely cannot be meaningfully handled
```
linacambridge, line 5: The pattern we've been using for our other FFIs is a bit different—we have an `ffi` module inside our parent crate, and that has an `error_codes` submodule and conversions between that crate's errors and `ExternError`. How do you all feel about changing that slightly?

My thoughts were:

1. I'm not a big fan of submodules in general (except for `mod tests`), because it's harder to see when you're looking at the directory structure. Places in particular is a bit of a tangle: there's a top-level `bookmark_sync` module, _and_ a `mod bookmark_sync` declaration inside `storage/bookmarks.rs` (ditto for history), but you wouldn't guess that from looking at the file names. When working on that code, I get a little disoriented at first—"wait, `bookmarks` doesn't have a `bookmark_sync`, where's that coming from...oh." That said, `ffi::error_codes` is only one level of nesting, and not that bad.
2. But, we have an `ffi` module and an `ffi` "subcrate", and I do get those confused as well.
3. Having the `implement_into_ffi_by_{json, protobuf}` far away from where the types are declared gets a little confusing, too, but maybe it's my mental model that's off (I go for the type first, but grouping those into `ffi.rs` feels like a "concept first").

But I'd be happy to change this to follow the others if this is harder/clunkier/we're just not feeling it!

components/webext-storage/src/api.rs
```diff
@@ -77,33 +107,21 @@ pub struct StorageValueChange {
...
+/// A newtype for a list of changed keys and values, used to notify
+/// `storage.onChanged` listeners when an item is updated or removed. It
+/// dereferences to its contents, has a custom serializer to serialize the
+/// changes as a JSON object instead of a list, and implements
+/// `ffi_support::IntoFfi` so that it can be passed over the FFI.
 #[derive(Debug, Clone, PartialEq)]
-pub struct StorageChanges {
-    changes: Vec<StorageValueChange>,
-}
+pub struct StorageChanges(Vec<StorageValueChange>);
```
linacambridge, line 116: It was fine with `changes` before, but I liked the symmetry, and the older comment saying "it should just be a plain vec" if not for the trait implementations (`Serialize`, and now `IntoFfi`) was practically screaming "newtype!" 😁

components/webext-storage/ffi/src/error_codes.rs
mhammond: +1 from me - flat is better than nested. I agree places is too nested (and that's my fault IIRC)

components/webext-storage/src/api.rs
mhammond: At some point moving these related types into their own module probably makes sense - I'm not bothered if that time is now or not.

components/webext-storage/src/api.rs
```diff
@@ -62,6 +65,33 @@ fn remove_from_db(tx: &Transaction<'_>, ext_id: &str) -> Result<()> {
     Ok(())
 }
 
+/// A newtype for a stored value, which dereferences to its contents. This
```
mhammond, line 68: I gotta say I'm not completely loving this new type. Magic `IntoFfi` is nice and all, but the changes below with needing to do the magic `&*` dances adds a layer of complexity and opaqueness that doesn't really seem like a net win.

But needing this wrapper type so `implement_into_ffi_by_json` can be used on a JsonValue doesn't sound right, so I suspect I'm missing something?

linacambridge: I agree—I'm tempted to revert my changes to `StorageChanges` for the same reason, it feels too magical compared to what you had before. Let's wait for @thomcc to chime in, I probably missed a much easier way to do what we had before!

linacambridge: (I totally did—just call `serde_json::to_string` directly in the FFI code! 🤦‍♀️)

components/webext-storage/ffi/src/error_codes.rs
linacambridge: Nah, I definitely share that blame, too! 😆 A year and a half ago, we were all still figuring out good patterns for Rust!

thomcc: Nested in a separate file feels worse to me, I'd rather these just be constants in the parent maybe -- It's hard for me to imagine a time when you edit this file but don't also edit the other, but honestly I don't care much and it seems I'm the only one that feels this way.

rfk:
This seems like a very old PR, @linacambridge was this subsumed by something else?

linacambridge:
Uhhh...nope, I just didn't land it, oops! 😅

PR #3050, opened 2020-04-29
Avoid having serde double-serialize the values in the changes struct.

mhammond:
We used serde's `.to_string()` when created the StorageValueChange struct,
then told serde to serialize the entire StorageChanges array, so the values
ended up like `{"oldValue":"{\"sub-object\":\"sub-value\"}"}`


mhammond:
Yeah - the problem is that only *part* of it is double serialized, which doesn't end well :(

linacambridge:
That detekt check has been pending all day, I wonder what’s going on. In the meantime, I’d love to get this in to m-c, so just gonna land it manually!

linacambridge:
(And also, looks like GitHub for iOS doesn’t let you enter a custom merge commit message, whoops! 😅)

Issue #3032, opened 2020-04-24
Android networking appears broken in v0.58 release


rfk:
Ref https://github.com/mozilla-mobile/android-components/pull/6636#issuecomment-618711216

```
04-23 15:37:47.745 15466 15524 E viaduct::backend::ffi: Viaduct Ffi Error: doFetch failed: null
04-23 15:37:47.746 15466 15524 E viaduct::backend::ffi: Missing HTTP status
04-23 15:37:47.746 15466 15524 W fxa_client::ffi: Network error: Error(
04-23 15:37:47.746 15466 15524 W fxa_client::ffi:
04-23 15:37:47.746 15466 15524 W fxa_client::ffi: Network error: Backend error: Missing HTTP status)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: Error while running: complete oauth flow
04-23 15:37:47.746 15466 15524 E FirefoxAccount: mozilla.appservices.fxaclient.FxaException$Network: Network error: Backend error: Missing HTTP status
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.appservices.fxaclient.rust.RustError.intoException(RustError.kt:43)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.appservices.fxaclient.FirefoxAccount.completeOAuthFlow(FirefoxAccount.kt:687)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.components.service.fxa.FirefoxAccount$completeOAuthFlowAsync$1$1.invoke(FirefoxAccount.kt:196)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.components.service.fxa.FirefoxAccount$completeOAuthFlowAsync$1$1.invoke(FirefoxAccount.kt:29)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.components.service.fxa.UtilsKt$handleFxaExceptions$7.invokeSuspend(Utils.kt:68)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.components.service.fxa.UtilsKt$handleFxaExceptions$7.invoke(Unknown Source:10)
04-23 15:37:47.746 15466 15524 E FirefoxAccount: 	at mozilla.components.service.fxa.UtilsKt.handleFxaExceptions(Utils.kt:28)
[...]

PR #3054, opened 2020-04-29
Workaround protobuf bug

### LINKED ISSUE #3053: Protobuf crashes in v0.58.1
See https://github.com/mozilla-mobile/fenix/issues/10277 for context.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1200)
┆Sprint: Backlog

eoger:
The should fix the Android 5 issue reported in https://github.com/mozilla-mobile/fenix/issues/10277 (fixes #3053 also).
However because the fix to the above involves upgrading to 3.11.4, this would resurrect the bug we had earlier in https://github.com/mozilla/application-services/issues/3032, so I also fixed this one in a different way.

I'm going to make review comments to explain what I did here.

### DIFF
--- a/components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
+++ b/components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
@@ -8,6 +8,7 @@ import com.google.protobuf.CodedInputStream
 import com.google.protobuf.CodedOutputStream
 import com.sun.jna.Pointer
 import com.sun.jna.Structure
+import java.nio.ByteBuffer
 
 /**
  * This is a mapping for the `ffi_support::ByteBuffer` struct.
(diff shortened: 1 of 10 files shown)


components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
```diff
@@ -47,13 +48,18 @@ open class RustBuffer : Structure() {
 
     fun asCodedInputStream(): CodedInputStream? {
         return this.data?.let {
-            CodedInputStream.newInstance(it.getByteBuffer(0, this.len))
+            if (this.len < Int.MIN_VALUE || this.len > Int.MAX_VALUE) {
+                throw RuntimeException("len does not fit in a int")
+            }
+            CodedInputStream.newInstance(it.getByteArray(0, this.len.toInt()))
```
eoger, line 54: Here we avoid using the `ByteBuffer` constructor of `CodedInputStream` since it ends up creating a [buggy implementation](https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L185-L187).

components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
```diff
@@ -47,13 +48,18 @@ open class RustBuffer : Structure() {
...
         }
     }
 
     fun asCodedOutputStream(): CodedOutputStream? {
         return this.data?.let {
-            CodedOutputStream.newInstance(it.getByteBuffer(0, this.len))
+            val method = CodedOutputStream::class.java.getDeclaredMethod("newSafeInstance", ByteBuffer::class.java)
+            method.isAccessible = true
+            return method.invoke(null, it.getByteBuffer(0, this.len)) as CodedOutputStream
```
eoger, line 62: We do the same here, by [forcing the usage](https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L130-L140) of `newSafeInstance`

components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
thomcc: That's unfortunate, as this requires an additional copy. We should fix this when possible...

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -95,8 +96,8 @@ object RustHttpConfig {
                     }
                     rb
                 } catch (e: Throwable) {
-                    LibViaduct.INSTANCE.viaduct_log_error("Network error: ${e.message}")
-                    MsgTypes.Response.newBuilder().setExceptionMessage(e.message)
+                    LibViaduct.INSTANCE.viaduct_log_error("Network error: ${Log.getStackTraceString(e)}")
```
thomcc, line 99: Not opposed to this but does it have any PII concerns?

components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
eoger: Yup I opened a PR: https://github.com/protocolbuffers/protobuf/pull/7440

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
eoger: It should be fine: it logs a classic Java stack trace (no parameters or any kind of PII as far as I saw).

grigoryk: Stacktraces could easily have PII embedded in an exception message, and the new version has a higher chance of hitting that, I think? Depending on how exactly stacktrace is printed out.

components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt
grigoryk: Ouch :/

grigoryk:
Specifically, tested w/ an emulator and a FireHD tablet that runs Android 5.

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
eoger: Reverted the file.

PR #3061, opened 2020-04-30
Relax URL version

eoger:
Same as #3040 but across the board (having the same problem).


eoger:
> Also I think dependabot caused this problem. I think it's worth discussing if it's really providing us like, any value at all...

I was thinking the same thing, it would be nice if dependabot could make PRs only for minor version bumps on our dependencies and do not try to get the full version number in the .tomls.

thomcc:
I won't be there but this would be something to bring up in the kitchen sync meeting today.

PR #3063, opened 2020-05-01
Expose interrupt handles for the extension store

linacambridge:


components/webext-storage/src/db.rs
```diff
@@ -63,6 +68,29 @@ impl StorageDb {
...
+        )
+    }
+
+    /// Creates an object that knows when it's been interrupted. This should
+    /// be created at the same time as an interrupt handle, and passed to any
+    /// methods that do database work. This is the other side of the
+    /// `SqlInterruptHandle`: when a handle is interrupted, it flags all
+    /// active interrupt scopes as interrupted, too, so that they can abort
+    /// pending work.
+    #[allow(dead_code)]
```
linacambridge, line 89: This isn't used yet, but I wanted to get it in here so that we don't have to add it back when we go to use it in Mark's syncing patch.

linacambridge:
The m-c side is https://bugzilla.mozilla.org/show_bug.cgi?id=1634626.

linacambridge:
Thom, it would be wonderful if you could have a look as well, since you wrote the interrupt handle code and reviewed the m-c part! 😄

PR #2892, opened 2020-04-01
First cut at the core sync infrastructure

mhammond:
Builds on #2810. This needs the trait from Lina's patch, but is setup to work with it. It implements the 3-way merge (although doesn't quite get deletions right yet, so there's still some work to do) - but comments welcome.

PR #3064, opened 2020-05-01
Implement a skeleton `sync15::BridgedEngine` for extension storage

linacambridge:
Currently, the only supported method is `wipe`, with all other
methods returning `NotImplemented`. We can wire this up to Mark's
syncing implementation in #2892. This also implements enough of the
glue to where it can be vendored and used in Desktop.


linacambridge:
I tried to make the changes to `Cargo.toml` and `lib.rs` match up with #2892, so no matter which ones lands first, they should still rebase cleanly.

linacambridge:
m-c side: https://bugzilla.mozilla.org/show_bug.cgi?id=1634191

PR #3065, opened 2020-05-01
Reset Send Tab keys on decrypt error

### LINKED ISSUE #3095: [firefox-ios] Tab decryption is failing
This is a tracking ticket for https://github.com/mozilla-mobile/firefox-ios/issues/6460.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1254)
┆Epic: a-s consumer issues

eoger:
Fixes #3095.

For some reason I was not receiving tabs on Firefox for iOS beta v25. The following patch fixes my problem: when we get a tab decryption error delete the send-tab keys, regenerate them and re-upload the device record.

### DIFF
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -111,6 +111,18 @@ impl FirefoxAccount {
         Ok(())
     }
 
+    /// Re-register the device capabilities, this should only be used internally.
+    pub(crate) fn reregister_current_capabilities(&mut self) -> Result<()> {
+        let current_capabilities: Vec<Capability> =
+            self.state.device_capabilities.clone().into_iter().collect();
+        let commands = self.register_capabilities(&current_capabilities)?;
+        let update = DeviceUpdateRequestBuilder::new()
+            .available_commands(&commands)
+            .build();
+        self.update_device(update)?;
+        Ok(())
+    }
+
     pub(crate) fn invoke_command(
         &self,
         command: &str,
@@ -171,7 +183,7 @@ impl FirefoxAccount {
     }
 
     fn parse_commands_messages(
-        &self,
+        &mut self,
         messages: Vec<PendingCommand>,
     ) -> Result<Vec<IncomingDeviceCommand>> {
         let devices = self.get_devices()?;
@@ -189,7 +201,7 @@ impl FirefoxAccount {
     }
 
     fn parse_command(
-        &self,
+        &mut self,
         command_data: CommandData,
         devices: &[Device],
     ) -> Result<IncomingDeviceCommand> {
--- a/components/fxa-client/src/send_tab.rs
+++ b/components/fxa-client/src/send_tab.rs
@@ -4,7 +4,9 @@
 
 pub use crate::commands::send_tab::{SendTabPayload, TabHistoryEntry};
 use crate::{
-    commands::send_tab::{self, EncryptedSendTabPayload, PrivateSendTabKeys, PublicSendTabKeys},
+    commands::send_tab::{
+        self, EncryptedSendTabPayload, PrivateSendTabKeys, PublicSendTabKeys, SendTabKeysPayload,
+    },
     error::*,
     http_client::GetDeviceResponse,
     scopes, FirefoxAccount, IncomingDeviceCommand,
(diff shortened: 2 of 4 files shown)


components/fxa-client/src/send_tab.rs
```diff
@@ -64,7 +65,14 @@ impl FirefoxAccount {
                 }
             };
         let encrypted_payload: EncryptedSendTabPayload = serde_json::from_value(payload)?;
-        let payload = encrypted_payload.decrypt(&send_tab_key)?;
-        Ok(IncomingDeviceCommand::TabReceived { sender, payload })
+        match encrypted_payload.decrypt(&send_tab_key) {
+            Ok(payload) => Ok(IncomingDeviceCommand::TabReceived { sender, payload }),
+            Err(e) => {
```
rfk, line 70: How much additional work would it be to re-fetch our own device record and check whether its contents match what we expect? I have to assume that the existing device registration is somehow incorrect, since re-registering fixes the problem. So this might be a good opportunity to try to log more diagnostics about exactly *how* it was incorrect.

components/fxa-client/src/device.rs
```diff
@@ -111,6 +111,19 @@ impl FirefoxAccount {
         Ok(())
     }
 
+    /// Register a device capability against the current device.
+    /// No kind of caching is done, as this is intended to be called internally.
+    ///
+    /// **💾 This method alters the persisted account state.**
+    pub(crate) fn register_capability(&mut self, capability: Capability) -> Result<()> {
+        let commands = self.register_capabilities(&[capability])?;
```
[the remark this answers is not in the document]
eoger: Something I could do alternatively is mutate `self.state.device_capabilities` to remove `SendTab` and rely on `ensure_capabilities(copy_of_device_capabilities_before_mutation)` to do the work for us?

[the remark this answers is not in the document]
rfk: That sounds reasonable to me 👍

components/fxa-client/src/send_tab.rs
eoger: In my case everything looked all right (send tab key and push reg in order) and I never really figured why my keys were wrong. I'm not quite sure how could we compare the device contents since the random IV makes a comparison impossible :/
I believe we were hitting an error in this general area: https://searchfox.org/mozilla-mobile/rev/fd32c1b030795f42f68da92c36e9bdef97b52d29/application-services/components/fxa-client/src/send_tab.rs#56-67

rfk: I think we'd have to try to decrypt our own send-tab key bundle and check that it matches what we expect from the plaintext.

rfk:
Hrm, I was sure I wrote a big comment to go with m "request changes" here but I don't see it in the view. Maybe I clicked the wrong button and it got discarded :-/

The TL;DR was: I'm OK with doing this but a little concerned about papering over a problem, especially if it means we end up in some sort of loop of re-registering our device record every time we receive a sent tab.

When we re-register, we don't actually change the send-tab keys, right? We re-encrypt the existing send-tab keys and upload them again?

eoger:
I have added a `diagnose_remote_keys` function that should end up logging the error on Sentry.

> When we re-register, we don't actually change the send-tab keys, right? We re-encrypt the existing send-tab keys and upload them again?

Right now we actually delete them.

components/fxa-client/src/send_tab.rs
eoger: Added `diagnose_remote_keys`

components/fxa-client/src/send_tab.rs
```diff
@@ -64,7 +66,46 @@ impl FirefoxAccount {
...
+                log::error!("Could not decrypt Send Tab payload. Diagnosing then resetting the Send Tab keys.");
+                match self.diagnose_remote_keys(send_tab_key) {
+                    Ok(_) => log::error!("Could not find the cause of the Send Tab keys issue."),
+                    Err(e) => log::error!("{}", e),
+                };
+                // Reset the Send Tab keys.
+                self.state.commands_data.remove(send_tab::COMMAND_NAME);
+                self.reregister_current_capabilities()?;
```
rfk, line 79: If this network call fails, we'll be left unable to receive any tabs until app restart, which will call `ensure_capabilities` to be called and repair the situation. That seems OK from a "we already couldn't receive tabs anyway" perspective.

PR #3066, opened 2020-05-01
Add Debouncing to Get Devices Call

### LINKED ISSUE #2625: Avoid fetching the device list multiple times in quick succession

lougeniaC64:
Fixes #2625
- adds debouncing to `get_devices()`

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -9,3 +9,11 @@
 ### Breaking changes
 
 - The `include_cookies` setting is not supported anymore (was `false` by default). ([#3096](https://github.com/mozilla/application-services/pull/3096))
+
+## FxA Client
+
+- Added option boolean argument `ignoreCache` to ignore caching for `getDevices`. ([#3066](https://github.com/mozilla/application-services/pull/3066))
+
+### ⚠️ Breaking changes ⚠️
+- iOS: Renamed `fetchDevices(forceRefresh)` to `getDevices(ignoreCache)` to establish parity with Android. ([#3066](https://github.com/mozilla/application-services/pull/3066))
+- iOS: Renamed argument of `fetchProfile` from `forceRefresh` to `ignoreCache`. ([#3066](https://github.com/mozilla/application-services/pull/3066))
--- a/components/fxa-client/ffi/src/lib.rs
+++ b/components/fxa-client/ffi/src/lib.rs
@@ -91,11 +91,11 @@ pub extern "C" fn fxa_to_json(handle: u64, error: &mut ExternError) -> *mut c_ch
 #[no_mangle]
 pub extern "C" fn fxa_profile(
     handle: u64,
-    ignore_cache: bool,
+    ignore_cache: u8,
     error: &mut ExternError,
 ) -> ByteBuffer {
     log::debug!("fxa_profile");
-    ACCOUNTS.call_with_result_mut(error, handle, |fxa| fxa.get_profile(ignore_cache))
+    ACCOUNTS.call_with_result_mut(error, handle, |fxa| fxa.get_profile(ignore_cache != 0))
 }
 
 /// Get the pairing URL to navigate to on the Auth side (typically a computer).
@@ -414,15 +414,22 @@ pub extern "C" fn fxa_set_device_name(
 
 /// Fetch the devices (including the current one) in the current account.
 ///
+/// Devices might get cached in-memory and the caller might get served a cached version.
+/// To bypass this, the `ignore_cache` parameter can be set to `true`.
+///
 /// # Safety
 ///
 /// A destructor [fxa_bytebuffer_free] is provided for releasing the memory for this
 /// pointer type.
 #[no_mangle]
-pub extern "C" fn fxa_get_devices(handle: u64, error: &mut ExternError) -> ByteBuffer {
+pub extern "C" fn fxa_get_devices(
+    handle: u64,
+    ignore_cache: u8,
+    error: &mut ExternError,
+) -> ByteBuffer {
     log::debug!("fxa_get_devices");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
-        fxa.get_devices().map(|d| {
+        fxa.get_devices(ignore_cache != 0).map(|d| {
             let devices = d.into_iter().map(|device| device.into()).collect();
             fxa_client::msg_types::Devices { devices }
         })
--- a/components/fxa-client/ios/FxAClient/FxAccount.swift
+++ b/components/fxa-client/ios/FxAClient/FxAccount.swift
@@ -22,10 +22,10 @@ class FxAccount: RustFxAccount {
         persistCallback = nil
     }
 
-    override func getProfile(forceRefresh: Bool) throws -> Profile {
+    override func getProfile(ignoreCache: Bool) throws -> Profile {
         defer { tryPersistState() }
         return try notifyAuthErrors {
-            try super.getProfile(forceRefresh: forceRefresh)
+            try super.getProfile(ignoreCache: ignoreCache)
         }
     }
 
@@ -67,9 +67,9 @@ class FxAccount: RustFxAccount {
         }
     }
 
-    override func fetchDevices() throws -> [Device] {
+    override func getDevices(ignoreCache: Bool = false) throws -> [Device] {
         return try notifyAuthErrors {
-            try super.fetchDevices()
+            try super.getDevices(ignoreCache: ignoreCache)
         }
     }
 
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -12,22 +12,42 @@ use crate::{
         CommandData, DeviceUpdateRequest, DeviceUpdateRequestBuilder, PendingCommand,
         UpdateDeviceResponse,
     },
-    FirefoxAccount, IncomingDeviceCommand,
+    util, CachedResponse, FirefoxAccount, IncomingDeviceCommand,
 };
 use serde_derive::*;
 use std::collections::{HashMap, HashSet};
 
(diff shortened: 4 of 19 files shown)


components/fxa-client/src/device.rs
```diff
@@ -12,20 +12,37 @@ use crate::{
...
 use std::collections::{HashMap, HashSet};
 
+// An devices response is considered fresh for `DEVICES_FRESHNESS_THRESHOLD` ms.
+const DEVICES_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
+
 impl FirefoxAccount {
     /// Fetches the list of devices from the current account including
     /// the current one.
-    pub fn get_devices(&self) -> Result<Vec<Device>> {
+    pub fn get_devices(&mut self) -> Result<Vec<Device>> {
```
vladikoff, line 26: in profile we also have `ignore_cache`, would we need something like this here as well https://github.com/mozilla/application-services/blob/f3fe12419e64b2c77f40398c8e54ca4a30672873/components/fxa-client/src/profile.rs#L21

components/fxa-client/src/oauth/attached_clients.rs
```diff
@@ -9,6 +9,7 @@ use crate::{error::*, util, CachedResponse, FirefoxAccount};
 const ATTACHED_CLIENTS_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
 
 impl FirefoxAccount {
+    /// Fetches the list of attached clients connected to the current account.
```
lougeniaC64, line 12: FYI: This is an unrelated addition. Just adding it because I was in the neighborhood.

components/fxa-client/src/device.rs
rfk: Desktop does have a similar `ignore_cache` facility FWIW

lougeniaC64: @rfk From reading the comments in the bug I wasn't sure if this should be omitted until explicitly called for. I've added it.

CHANGES_UNRELEASED.md
```diff
@@ -3,3 +3,9 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v0.58.2...master)
+
+## FxA Client
+- iOS: Added optional boolean argument `forceRefresh` to ignore caching for `fetchDevices`.
+- Added an optional `ttl` parameter to `getAccessToken` to limit the lifetime of the token. ([#3066](https://github.com/mozilla/application-services/pull/3066))
```
rfk, line 9: There's some stuff here about `ttl` which doesn't seem quite right, possibly copy-pasta or a bag merge from a previous changelog?

components/fxa-client/ffi/src/lib.rs
```diff
@@ -414,15 +414,22 @@ pub extern "C" fn fxa_set_device_name(
...
+///
 /// # Safety
 ///
 /// A destructor [fxa_bytebuffer_free] is provided for releasing the memory for this
 /// pointer type.
 #[no_mangle]
-pub extern "C" fn fxa_get_devices(handle: u64, error: &mut ExternError) -> ByteBuffer {
+pub extern "C" fn fxa_get_devices(
+    handle: u64,
+    ignore_cache: bool,
```
rfk, line 427: For reasons that I won't pretend to understand in detail, we're not supposed to use `bool` at the FFI layer:

  https://github.com/mozilla/application-services/blob/master/docs/howtos/when-to-use-what-in-the-ffi.md#primitives

That `fxa_get_profile` currently does use it, is probably an oversight. Could you please adjust both `fxa_get_devices` and `fxa_get_profile` to use a `u8` here rather than `bool`, following the [example of the places FFI](https://github.com/mozilla/application-services/blob/5b130877dc5f5f8e4ebd8f0c1f4727d21a6c4e33/components/places/ffi/src/lib.rs#L266)? (Unfortunately this means converting it to a `u8` inside the [android](https://github.com/mozilla/application-services/blob/5b130877dc5f5f8e4ebd8f0c1f4727d21a6c4e33/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt#L318) and ios binding layers 😞)

components/fxa-client/ios/FxAClient/FxAccount.swift
```diff
@@ -67,9 +67,9 @@ class FxAccount: RustFxAccount {
         }
     }
 
-    override func fetchDevices() throws -> [Device] {
+    override func fetchDevices(forceRefresh: Bool = false) throws -> [Device] {
         return try notifyAuthErrors {
-            try super.fetchDevices()
+            try super.fetchDevices(forceRefresh: forceRefresh)
```
rfk, line 72: It pains me that we changed the name of this arg in the swift API, but 👍 for keeping it consistent with `getProfile` 😄

components/fxa-client/src/device.rs
```diff
@@ -654,4 +674,71 @@ mod tests {
...
+
+        assert!(res.is_ok());
+        assert!(fxa.devices_cache.is_some());
+
+        let cached_devices_res = fxa.devices_cache.unwrap();
+        assert!(!cached_devices_res.response.is_empty());
+        assert!(cached_devices_res.cached_at > 0);
+
+        let cached_devices = &cached_devices_res.response[0];
+        assert_eq!(cached_devices.clone().id, "device1".to_string());
```
rfk, line 717: Would it be possible to call `get_devices(false)` a second time here, and assert that it *didn't* make another network request?

components/fxa-client/ios/FxAClient/FxAccount.swift
lougeniaC64: Yeah, I'm curious why iOS uses the fetchDevices/forceRefresh naming while android uses getDevices/ignoreCache. I almost didn't catch the iOS changes because of the difference. A part of me wants to rename the iOS functions but I also want to avoid causing breaking changes and possibly disregarding a sound design decision.

components/fxa-client/ffi/src/lib.rs
lougeniaC64: Thanks! I was somewhat unsure about the FFI layer changes I made so this comment and the link are very helpful.

CHANGES_UNRELEASED.md
lougeniaC64: Copy pasta for sure. You saw nothing.

components/fxa-client/ios/FxAClient/FxAccount.swift
vladikoff: if renaming is sound then, IMHO we should clean those up at some point and report it in breaking changes

CHANGES_UNRELEASED.md
```diff
@@ -3,3 +3,7 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v0.58.2...master)
+
+## FxA Client
+- iOS: Added optional boolean argument `forceRefresh` to ignore caching for `fetchDevices`. ([#3066](https://github.com/mozilla/application-services/pull/3066))
+- Android: Added option boolean argument `ignoreCache` to ignore caching for `getDevices`. ([#3066](https://github.com/mozilla/application-services/pull/3066))
```
rfk, line 9: Oh wow it's even worse when you see the two different names next to each other in the changelog like this 😆 

I will defer to @eoger on the wisdom or safety of attempting to rename the iOS version for consistency.

components/fxa-client/ios/FxAClient/FxAccount.swift
rfk: /cc @eoger

CHANGES_UNRELEASED.md
eoger: FWIW `CHANGES_UNRELEASED.md` is empty at the moment and needs to be rebased :)
But yeah I support this breaking change, iOS only uses the `forceRefresh` argument [in 1 place](https://searchfox.org/mozilla-mobile/rev/fd32c1b030795f42f68da92c36e9bdef97b52d29/firefox-ios/RustFxA/FxAWebView.swift#163) so it won't be too bad.

PR #3067, opened 2020-05-03
Add `{Incoming, Outgoing}Envelope` types for bridged engines

linacambridge:
We need a way to pass BSO metadata, like the last modified time and
sort index, from Desktop Sync to our bridged Rust engines. We can't
pass Desktop's BSOs directly because they're encrypted, and we rely
on the existing `Weave.Crypto` machinery to decrypt them.

So instead, we introduce envelopes, which combine BSO metadata
fields with the decrypted cleartext. We intentionally use a new term,
instead of calling these "records", to avoid confusion between Desktop
Sync records (which conflate BSOs, encrypted wrappers, and payloads)
and `sync-15` records (which are just the payloads).

Envelopes are an internal concept: they are never sent to the server,
and the only methods available on them are conversions for turning
them into payloads, and building them from payloads.

These structs are shared between Desktop Sync JS code and our Rust
code. See the comments in Desktop's `bridged_engine.js`, in
[this Phabricator diff](https://phabricator.services.mozilla.com/D73581)
for more details.

### DIFF
--- a/components/support/sync15-traits/src/bridged_engine.rs
+++ b/components/support/sync15-traits/src/bridged_engine.rs
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::{error::Error, fmt};
+
+use serde::{Deserialize, Serialize};
+
+use super::{Guid, Payload, ServerTimestamp};
+
 /// A bridged Sync engine implements all the methods needed to support
 /// Desktop Sync.
 pub trait BridgedEngine {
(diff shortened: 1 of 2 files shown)


components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::{error::Error, fmt};
+
```
mhammond, line 6: lots of blank lines :)

components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -84,28 +90,110 @@ pub trait BridgedEngine {
...
+        Ok(payload)
+    }
+}
+
+/// An envelope for an outgoing item, returned from `BridgedEngine::apply`. This
+/// is similar to `IncomingEnvelope`, but omits fields that are only set by the
+/// server, like `modified`.
+#[derive(Clone, Debug, Serialize)]
+pub struct OutgoingEnvelope {
+    id: Guid,
```
mhammond, line 165: shouldn't `sortindex` be here too? (Not that we use it in rust yet, so I'd be fine with a "todo" style comment, but IIUC, conceptually it should exist here, right?)

components/support/sync15-traits/src/bridged_engine.rs
linacambridge: Haha, yeah! I've been grouping imports into sections (`std` first, then other crates, then our `crate::` imports), but it's kind of unnecessary when there's one line each! 😅

PR #3071, opened 2020-05-05
Wire up the bridged engine!

linacambridge:
Connecting @mhammond's sync methods to `BridgedEngine` so that we can get syncing working in Desktop!

This integration was a joy to write thanks to the excellent tests, clear names, and examples! I love how `store_incoming` and `apply` read especially! Thanks so much for building it out so thoughtfully, @mhammond! 🎉🚀💯

### DIFF
--- a/components/webext-storage/sql/create_sync_temp_tables.sql
+++ b/components/webext-storage/sql/create_sync_temp_tables.sql
@@ -28,7 +28,27 @@ CREATE TEMP TABLE IF NOT EXISTS storage_sync_outgoing_staging (
     guid TEXT NOT NULL PRIMARY KEY DEFAULT(generate_guid()),
     ext_id TEXT NOT NULL UNIQUE,
     data TEXT,
-    sync_change_counter INTEGER NOT NULL
+    sync_change_counter INTEGER NOT NULL,
+    was_uploaded BOOLEAN NOT NULL DEFAULT 0
 );
+
+CREATE TEMP TRIGGER IF NOT EXISTS record_uploaded
+AFTER UPDATE OF was_uploaded ON storage_sync_outgoing_staging
+WHEN NEW.was_uploaded
+BEGIN
+    -- Decrement the local change counter for uploaded items. If any local items
+    -- changed while we were uploading, their change counters will remain > 0,
+    -- and we'll merge them again on the next sync.
+    UPDATE storage_sync_data SET
+      sync_change_counter = sync_change_counter - NEW.sync_change_counter
+    WHERE ext_id = NEW.ext_id;
+
+    -- Delete uploaded tombstones entirely; they're only kept in the mirror.
+    DELETE FROM storage_sync_data WHERE data IS NULL AND sync_change_counter = 0 AND ext_id = NEW.ext_id;
+
+    -- And write the uploaded item back to the mirror.
+    INSERT OR REPLACE INTO storage_sync_mirror (guid, ext_id, data)
+    VALUES (NEW.guid, NEW.ext_id, NEW.data);
+END;
 
 DELETE FROM temp.storage_sync_outgoing_staging;
--- a/components/webext-storage/src/sync/bridge.rs
+++ b/components/webext-storage/src/sync/bridge.rs
@@ -50,6 +50,11 @@ impl<'a> sync15_traits::BridgedEngine for BridgedEngine<'a> {
         Err(ErrorKind::NotImplemented.into())
     }
 
+    fn sync_started(&self) -> Result<()> {
+        schema::create_empty_sync_temp_tables(&self.db)?;
+        Ok(())
+    }
+
     fn store_incoming(&self, incoming_envelopes: &[IncomingEnvelope]) -> Result<()> {
         let signal = self.db.begin_interrupt_scope();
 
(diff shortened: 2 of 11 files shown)


components/webext-storage/src/sync/bridge.rs
```diff
@@ -46,20 +50,61 @@ impl<'a> sync15_traits::BridgedEngine for BridgedEngine<'a> {
...
+        let outgoing = get_outgoing(&self.db, &signal)?
+            .into_iter()
+            .map(OutgoingEnvelope::from)
+            .collect::<Vec<_>>();
+        Ok(outgoing.into())
     }
 
-    fn set_uploaded(&self, _server_modified_millis: i64, _ids: &[String]) -> Result<()> {
-        Err(ErrorKind::NotImplemented.into())
+    /// TODO: This should be `SyncGuid`, not `String`.
```
linacambridge, line 88: Needs a fix to the `BridgedEngine` trait, so Desktop will need to change also.

components/webext-storage/src/sync/bridge.rs
```diff
@@ -46,20 +50,61 @@ impl<'a> sync15_traits::BridgedEngine for BridgedEngine<'a> {
...
+            .collect::<Vec<_>>();
+
+        let tx = self.db.unchecked_transaction()?;
+        record_uploaded(&tx, &guids, &signal)?;
+        tx.commit()?;
+
+        Ok(())
     }
 
+    /// TODO: Need a `sync_started`, so that we can create temp tables before we
```
linacambridge, line 103: Ditto, except we also need a new XPIDL method that we call just before we call `storeIncoming`. This gives us nice symmetry—`syncStarted` to set everything up, `syncFinished` when we're all done.

components/webext-storage/src/sync/mod.rs
```diff
@@ -24,19 +23,17 @@ pub fn is_default<T: PartialEq + Default>(v: &T) -> bool {
...
-// XXX - need to work out how to consolidate "payload" vs "bso", particularly
-// the timestamp, etc. For now, we represent this in ServerPayload, even though
-// that's never actually stored in this way.
-// XXX - this is going to be replaced with envelopes!
 #[derive(Debug, Serialize, Deserialize)]
-pub struct ServerPayload {
+// TODO: Should we camelize record properties? It's what we do for other
+// records, but, since this is the first engine where we'll use Rust
+// everywhere, it doesn't really matter.
+// #[serde(rename_all = "camelCase")]
```
linacambridge, line 30: I'm kind of tempted to say "no, don't camelize" but calling it out just in case.

components/webext-storage/src/sync/outgoing.rs
```diff
@@ -6,94 +6,76 @@
...
-/// complete and the data in the temp table will be used to update the local
-/// store.
-#[allow(dead_code)] // Kill this annotation once the bridged engine is hooked up
-pub fn get_and_record_outgoing(
-    tx: &Transaction<'_>,
-    _signal: &dyn Interruptee,
-) -> Result<Vec<ServerPayload>> {
-    // The item may not yet have a GUID (ie, it might not already be in either
-    // the mirror nor the incoming staging table.) We could probably perform
-    // some impressive sql gymnastics to handle this and arrange for the
```
linacambridge, line 57: I did the SQL gymnastics only because replacing `ServerPayload` with `Payload` was kind of annoying—we have to poke at its `.data` field directly to read out the fields, or call `into_record` when we _just_ had all the fields for the record...or use `Record`, but then it's clunky for tombstones. Since the only thing we needed to make it work was the `coalesce()`, I just went for it! 😆

I also moved writing back staged incoming into the mirror here, since we merged it correctly, and waiting for `record_uploaded` means our mirror would be out of sync with what's on the server if we're interrupted mid-sync.

components/webext-storage/sql/create_sync_temp_tables.sql
```diff
@@ -28,7 +28,27 @@ CREATE TEMP TABLE IF NOT EXISTS storage_sync_outgoing_staging (
     guid TEXT NOT NULL PRIMARY KEY DEFAULT(generate_guid()),
     ext_id TEXT NOT NULL UNIQUE,
     data TEXT,
-    sync_change_counter INTEGER NOT NULL
+    sync_change_counter INTEGER NOT NULL,
+    was_uploaded BOOLEAN NOT NULL DEFAULT 0
 );
 
+CREATE TEMP TRIGGER IF NOT EXISTS record_uploaded
```
mhammond, line 35: ah, I was wondering how you envisaged the trigger working when I updated my patch yesterday - this makes great sense.

components/webext-storage/src/sync/bridge.rs
mhammond: +1, I was just looking for that other `create_empty_sync_temp_tables()` call

components/webext-storage/src/sync/mod.rs
mhammond: Yeah, 🤷 from me. I guess I still lean towards that we should because even though we've kinda eliminated 'lastModified' and 'sortIndex' from here, they aren't going to go away so it's always going to be odd when we don't camelize.

(In this case, it's literally just 'ext_id' -> 'extId', right? Maybe we should just rename it to 'ext' ;)

rfk:
> Maybe we should just rename it to 'ext'

Think of the aggregate cost savings of not having to store all those extra bytes!!

components/webext-storage/src/sync/mod.rs
linacambridge: OK, I went ahead and added `rename_all = "camelCase"`. 😊 The others are spelled `modified` and `sortindex` (😆), so could go either way! We could also rename to `extid` to match `sortindex`, but I don't want to do that! 😂

PR #3080, opened 2020-05-05
Cargo update

eoger:
I've also removed the patch version on some deps (didn't go through all of them, next time maybe) to keep the churn in `Cargo.lock` as much as possible in the future.


eoger:
Maybe we should document a norm somewhere for new dependencies?

Issue #3083, opened 2020-05-05
The keychain state is stored with the wrong accessibility


eoger:
**TL;DR**: Send Tab on Firefox iOS v25.0 does not work if your device is locked (and hasn't received tabs succesfully recently).

The handler for Send Tab push messages tries to get the FxA state from the keychain:

https://github.com/mozilla/application-services/blob/4d6778fca195591e95b14c1b3c8d5edb056e0b39/components/fxa-client/ios/FxAClient/FxAccountStorage.swift#L17

However we're not specifying an accessibility! By default on iOS it is `kSecAttrAccessibleWhenUnlocked`, which means we can't read the FxA state till the device is unlocked and explains the problem I describe in the beginning of this ticket.

When writing the patch, we also need to ensure current keychain items are migrated to the proper accessibility.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/SYNC-1236)
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-05-08

rfk:
Excellent find 🚀

PR #3084, opened 2020-05-05
Use .afterFirstUnlock accessibility

eoger:
Fixes https://github.com/mozilla/application-services/issues/3083.

### DIFF
--- a/components/fxa-client/ios/FxAClient/FxAccountStorage.swift
+++ b/components/fxa-client/ios/FxAClient/FxAccountStorage.swift
@@ -8,13 +8,23 @@ import SwiftKeychainWrapper
 class KeyChainAccountStorage {
     internal var keychainWrapper: KeychainWrapper
     internal static var keychainKey: String = "accountJSON"
+    internal static var accessibility: KeychainItemAccessibility = .afterFirstUnlock
 
     init(keychainAccessGroup: String?) {
         keychainWrapper = KeychainWrapper.sharedAppContainerKeychain(keychainAccessGroup: keychainAccessGroup)
     }
 
     func read() -> FxAccount? {
-        if let json = keychainWrapper.string(forKey: KeyChainAccountStorage.keychainKey) {
+        // Firefox iOS v25.0 shipped with the default accessibility, which breaks Send Tab when the screen is locked.
+        // This method migrates the existing keychains to the correct accessibility.
+        keychainWrapper.ensureStringItemAccessibility(
+            KeyChainAccountStorage.accessibility,
+            forKey: KeyChainAccountStorage.keychainKey
+        )
+        if let json = keychainWrapper.string(
+            forKey: KeyChainAccountStorage.keychainKey,
+            withAccessibility: KeyChainAccountStorage.accessibility
+        ) {
             do {
                 return try FxAccount(fromJsonState: json)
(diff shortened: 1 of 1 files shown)


components/fxa-client/ios/FxAClient/FxAccountStorage.swift
```diff
@@ -8,13 +8,17 @@ import SwiftKeychainWrapper
...
 
     init(keychainAccessGroup: String?) {
         keychainWrapper = KeychainWrapper.sharedAppContainerKeychain(keychainAccessGroup: keychainAccessGroup)
     }
 
     func read() -> FxAccount? {
-        if let json = keychainWrapper.string(forKey: KeyChainAccountStorage.keychainKey) {
+        // Firefox iOS v25.0 shipped with the default accessibility, which breaks Send Tab when the screen is locked.
+        // This method migrates the existing keychains to the correct accessibility.
+        keychainWrapper.ensureStringItemAccessibility(KeyChainAccountStorage.accessibility, forKey: KeyChainAccountStorage.keychainKey)
```
eoger, line 20: This does the migration of current states.

components/fxa-client/ios/FxAClient/FxAccountStorage.swift
```diff
@@ -26,14 +30,37 @@ class KeyChainAccountStorage {
...
 
     func clear() {
-        if !keychainWrapper.removeObject(forKey: KeyChainAccountStorage.keychainKey) {
+        if !keychainWrapper.removeObject(forKey: KeyChainAccountStorage.keychainKey, withAccessibility: KeyChainAccountStorage.accessibility) {
             FxALog.error("Could not clear account state.")
         }
     }
 }
+
+public extension KeychainWrapper {
```
eoger, line 45: This was copy-pasted from firefox-ios code.

components/fxa-client/ios/FxAClient/FxAccountStorage.swift
```diff
@@ -8,13 +8,23 @@ import SwiftKeychainWrapper
 class KeyChainAccountStorage {
     internal var keychainWrapper: KeychainWrapper
     internal static var keychainKey: String = "accountJSON"
+    internal static var accessibility: KeychainItemAccessibility = .afterFirstUnlock
```
rfk, line 11: IIUC this changes *all* FxA state to be accessible after the first unlock, including the sessionToken and sync encryption keys. Does this risk weakening our security story compared to pre-v25.0 releases of Firefox iOS?

rfk:
The [docs for kSecAttrAccessibleAfterFirstUnlock](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlock) say:

```
Items with this attribute migrate to a new device when using encrypted backups.
```

Which kind of surprised me, but I guess it's fine - if you're restoring from a backup you could reasonably expect that your signed-in state would be preserved, and I don't think anything will break.

components/fxa-client/ios/FxAClient/FxAccountStorage.swift
eoger: It shouldn't, Firefox iOS used to store (and still does) Sync and FxA state under this accessibility.

PR #3096, opened 2020-05-06
Remove cookies support from Viaduct

### LINKED ISSUE #3082: Remove include_cookies from Viaduct API
From the security review we did today, cookies are tricky and would need more scrutiny to be included properly in Viaduct. Since we don't use cookies in our crates anyway, let's just remove this feature.

We can either remove `include_cookies` or hardcode the property internally to `false` no matter what if we don't want to break the API.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1237)
┆Story Points: 3
┆Epic: FxA crate in desktop

eoger:
Fixes #3082

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -2,4 +2,10 @@
 
 # Unreleased Changes
 
-[Full Changelog](https://github.com/mozilla/application-services/compare/v0.58.3...master)
+[Full Changelog](https://github.com/mozilla/application-services/compare/v0.58.2...master)
+
+## Viaduct
+
+### Breaking changes
+
+- The `include_cookies` setting is not supported anymore (was `false` by default). ([#3096](https://github.com/mozilla/application-services/pull/3096))
--- a/components/viaduct/src/fetch_msg_types.proto
+++ b/components/viaduct/src/fetch_msg_types.proto
@@ -25,10 +25,9 @@ message Request {
     optional bytes body = 3;
     map<string, string> headers = 4;
     required bool follow_redirects = 5;
-    required bool include_cookies = 6;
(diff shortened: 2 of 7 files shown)


CHANGES_UNRELEASED.md
```diff
@@ -3,3 +3,9 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v0.58.2...master)
+
+## Viaduct
+
+### Breaking changes
+
+- The `include_cookies` setting is not supported anymore. ([#3096](https://github.com/mozilla/application-services/pull/3096))
```
rfk, line 11: nit: I think it would be worth noting explicitly that they're always disabled (the current changelog implies that you can no longer control it, but doesn't say what the default is, you have to dig into the readme for that).

components/viaduct/src/fetch_msg_types.proto
```diff
@@ -25,7 +25,6 @@ message Request {
     optional bytes body = 3;
     map<string, string> headers = 4;
     required bool follow_redirects = 5;
-    required bool include_cookies = 6;
```
rfk, line 28: nit: are we able to leave some sort of comment here to explain the jump from `5` to `7` for future readers?

rfk: Alternately, would it actually be safe for us to re-number these fields? I don't think we have the usual backwards-compatibility concerns with an updated consumer received an older-format protobuf from an outdated producer.

eoger: Yup that's safe, just didn't want to pollute the git blame, but I guess it doesn't really matter.

PR #3103, opened 2020-05-06
Integrate release-v0.58.3 back to master

eoger:


rfk:
Hrm, I thought these fixes had already landed on master, so I'm lightly confused about why the `.afterFirstUnlock` change is showing up in the diff here.

Issue #1266, opened 2019-06-10
Better tests for sync_multiple


mhammond:
sync_multiple has exactly zero tests. The main problem is that to get any reasonable coverage, we need either http mocks or new traits setup specifically to help with testing.

Our sync manager is going to strike the exact same issue, but anything we do here will be able to be used by that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-679)
┆Epic: a-s live (e2e) integration tests
┆Sprint: Backlog

thomcc:
> sync_multiple has exactly zero tests

Unit tests, yeah. The integration tests exist in part for this purpose, but I agree that it's not an ideal situation.

PR #3108, opened 2020-05-10
Integration test for sync15 component

PepeRulo:
The beginning of an integration test for the sync15 component, this test currently inspects the behavior of sync_multiple(), storing records and serializing them (making sure that they're parsed correctly) and passing sync ids and metadata around (asserting that the state machine works correctly). 

There are many more aspects of sync_multiple to test, but we think that having the scaffolding will make them easier to test. The first thing would be to play with different arguments. Given the many moving parts of the function, we think that doing so in an integration test is a better start than unit tests. 

Addresses #1266 

@DawsonLeeRhodesCollege
@KrisJackson
@kwabam
@madigriffin


testing/sync-test/src/main.rs
```diff
@@ -98,7 +99,13 @@ pub fn main() {
     let opts = Opts::from_args();
     println!("### Running sync integration tests ###");
     init_testing();
-    run_test_groups(&opts, vec![crate::logins::get_test_group()]);
-    run_test_groups(&opts, vec![crate::tabs::get_test_group()]);
-    println!("### Sync integration tests passed!");
+    // Commenting these out to make the test runs shorter and less
+    // noisy. It would be nice to add an option to `Opts` to only
```
thomcc, line 103: I've definitely wanted this too. Note: Don't land this code without uncommenting these!!

testing/sync-test/src/sync15.rs
```diff
@@ -0,0 +1,315 @@
...
+        inbound: Vec<IncomingChangeset>,
+        _telem: &mut telemetry::Engine,
+    ) -> Result<OutgoingChangeset, Error> {
+        // Notice the `&mut *` and `.borrow_mut()` to extract the Vec from
+        // the RefCell.
+        let temp: Vec<TestRecord> = mem::take(&mut *self.test_records.borrow_mut());
+
+        let inbound = inbound.into_iter().next().unwrap();
+        for (payload, _timestamp) in inbound.changes {
+            // Here's an example of a magic "into" conversion that we define
```
thomcc, line 74: Hmm... I'm not sure this comment is helpful -- saying that we define it ourselves implies to me that these tests are defining it when it's actually a part of the sync15 library.

More broadly, nothing else it says is wrong, but it seems to be trying to explain `into_record`, but if `into_record` is confusing, improving the docs there is probably better than explaining it here...

Is it just the use of HRTB? They're unavoidable with serde, but if so, perhaps a link to something explainign them (rustbook or rust reference) on the into_record doc comment is worth it.

PepeRulo:
I'm a little late to the party but I don't want to miss the opportunity to thank @linacambridge and @thomcc for all your help, support, and review. Thank you so much!

PR #2820, opened 2020-03-16
Bug 1563169 - migrate PRs logic to use RelEng taskgraph instead

MihaiTabara:
This is an attempt to split the awesome work that @mitchhentges did in https://github.com/mozilla/application-services/pull/1945/ and land it in gradual phases:
* first landing PRs and iterating
* then landing the master pushes and iterating
* take care of github releases
* cleanup of old existing logic of `decisionlib`

This PRs tracks the first phase, taking care of PRs.

PR #3113, opened 2020-05-12
[ci full] Bug 1563169 - automation improvements prerequisite for Github releases

MihaiTabara:
Merely a follow-up work for https://github.com/mozilla/application-services/pull/2820 - addressing some fallouts and follow-up work from there and fixing some things that were broken in the meantime. This also lays the groundwork for upcoming Github releases taskgraph logic. 

This is now ready for review:
* works for both `ci skip` and `ci full`
* basically it doesn't add/remove any behaviors but improves the existing way of doing things to be up-to-date with the rest of mobile repos' taskgraph usage.
* mainly it gets us rid of the bashcripts --> docstrings python --> heredocs. I know this will make @eoger happy :) 

r? @JohanLorenzo CC: @badboy


taskcluster/app_services_taskgraph/job.py
```diff
@@ -0,0 +1,176 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
MihaiTabara, line 1: This file is a copy-paste from https://github.com/mozilla-mobile/fenix/blob/master/taskcluster/fenix_taskgraph/job.py with the SDK part commented out for now. I left a TODO to address later.

taskcluster/app_services_taskgraph/job.py
```diff
@@ -0,0 +1,176 @@
...
+    _set_run_task_attributes(job)
+    configure_taskdesc_for_run(config, job, taskdesc, job["worker"]["implementation"])
+
+
+@run_job_using("docker-worker", "gradlew", schema=gradlew_schema)
+def configure_gradlew(config, job, taskdesc):
+    run = job["run"]
+    worker = taskdesc["worker"] = job["worker"]
+
+    # TODO: to uncomment later when we'll port over logic from bug 1622339
```
MihaiTabara, line 75: Will address this later when I'll port over the logic from https://bugzilla.mozilla.org/show_bug.cgi?id=1622339

taskcluster/ci/lint/kind.yml
```diff
@@ -16,15 +16,14 @@ job-defaults:
...
+    description: "Running detekt over all modules"
     run:
-      command: ./gradlew --no-daemon clean detekt
+      gradlew: [detekt]
   ktlint:
-    description: "ktlint"
+    description: "Running ktlint over all modules"
     run:
-      command: ./gradlew --no-daemon clean ktlint
+      gradlew: [ktlint]
```
MihaiTabara, line 29: @JohanLorenzo Nit: we're losing the `--no-daemon clean` arguments along the way with this migration; I checked and it seems that's the case in other mobile projects too. Just wanted to double check whether this was intended or not.

taskcluster/app_services_taskgraph/build_config.py
```diff
@@ -10,20 +10,45 @@
 from taskgraph.util.memoize import memoize
 
 
+EXTENSIONS = {
```
MihaiTabara, line 13: This is partially what AC implements - https://github.com/mozilla-mobile/android-components/blob/8da8237bf9ccf1cf43088b995db5d4ca02c28a69/taskcluster/ac_taskgraph/build_config.py#L37 with some adjustments below.

taskcluster/app_services_taskgraph/build_config.py
```diff
@@ -10,20 +10,45 @@
...
-            'type': publication['type'],
-        } for publication in project['publications']]
     } for (name, project) in build_config['projects'].items()]
 
 
+def get_version():
+    return _read_build_config()["libraryVersion"]
+
+
+def get_extensions(module_name):
```
MihaiTabara, line 33: In AC we only needed the list of extensions returned here. In App-services we also needed the name of the module. Hence, I've enriched a bit the structure to a dictionary where the keys are the `publication name`. Rest of the bits are similar to https://github.com/mozilla-mobile/android-components/blob/8da8237bf9ccf1cf43088b995db5d4ca02c28a69/taskcluster/ac_taskgraph/build_config.py#L37

taskcluster/app_services_taskgraph/transforms/module_build.py
```diff
@@ -14,17 +17,16 @@ def build_task(config, tasks):
     for task in tasks:
         module_info = task["attributes"]["buildconfig"]
         name = module_info["name"]
-        version = module_info["version"]
+        version = get_version()
```
MihaiTabara, line 20: Note: Could've kept the version in the `attributes` but we're going to need a separate function for beetmover anyway so I thought it's cleaner to separate it like this.

taskcluster/app_services_taskgraph/build_config.py
```diff
@@ -10,20 +10,45 @@
...
+    'jar': ('.jar', '.pom')
+
+}
+CHECKSUMS_EXTENSIONS = ('.sha1', '.md5')
+
+
 def get_components():
     build_config = _read_build_config()
     return [{
-        'version': build_config['libraryVersion'],
```
MihaiTabara, line 16: `version`, `artifactId` and `publications` are unused as of now in the tasks' `attributes` hence I removed them to pass less data around. Will enrich the data struct once I deal with beetmover tasks, if needed.

taskcluster/app_services_taskgraph/build_config.py
```diff
@@ -10,20 +10,45 @@
...
-        'artifactId': project['artifactId'],
-        'publications': [{
-            'name': publication['name'],
-            'type': publication['type'],
-        } for publication in project['publications']]
     } for (name, project) in build_config['projects'].items()]
 
 
+def get_version():
+    return _read_build_config()["libraryVersion"]
```
MihaiTabara, line 30: Note: this is going to hit the memoized cached call of the buildconfig, so we're okay 👍

taskcluster/ci/android-build/kind.yml
```diff
@@ -30,15 +28,19 @@ jobs:
...
+        - [./taskcluster/scripts/rustup-setup.sh]
+        - [./taskcluster/scripts/cross-compile-setup.sh]
+      commands:
+        - [rsync, '-a', /builds/worker/fetches/libs/, /builds/worker/checkouts/src/libs/]
+        - [echo, 'rust.targets=linux-x86-64,x86_64\n', '>', local.properties]
+        - [./gradlew, '--no-daemon', 'clean']
+        - [./gradlew, '--no-daemon', 'assembleDebug']
+        - [./gradlew, '--no-daemon', 'testDebug']
+      using: run-commands
+      use-caches: true
```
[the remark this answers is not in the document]
tomprince: We don't have it set because the caching story with git and `run-task` is not nearly as robust as we have via `hg robustcheckout`. I recall there were issues with the wrong project being checked out. I don't recall if we ever dug into the details. It is much less of an issue given the relative sizes of the repositories that we are cloning, in any case.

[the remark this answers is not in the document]
MihaiTabara: TIL, good to know, thank you!

taskcluster/ci/module-build/kind.yml
```diff
@@ -27,16 +25,25 @@ job-defaults:
...
+      # FIXME: scripts subshell at runtime so we need to source this here
+      # to be able to access `rustup` and `rustc` from within the cc script
+      # and the gradle command. Another options could be to set those env vars
+      # here like: [export, 'PATH=$HOME/.cargo/bin:$PATH'
+      - [source, taskcluster/scripts/toolchain/rustup-setup.sh]
+      - [taskcluster/scripts/toolchain/cross-compile-setup.sh]
+      - [rsync, '-a', /builds/worker/fetches/libs/, /builds/worker/checkouts/src/libs/]
+      - [echo, 'rust.targets=arm,arm64,x86_64,x86,darwin,linux-x86-64,win32-x86-64-gnu\n', '>', local.properties]
+    gradlew:
+      - ':{module_name}:testDebug'
```
MihaiTabara, line 40: Unclear to me whether chaining the `gradlew`commands is the same as running them separately, one after another.

badboy: (Drive-by comment) It should be semantically the same to have them each on their own, but it will most likely be a bit slower, especially if `--no-daemon` is involved, as it parses the gradle task configuration again and again on each invocation

MihaiTabara: Yeah, you're totally right. My follow-up question would've been how come it now takes 9m40s for a build that use to take combined around 5mins. I can re-work this part and rely on `run-commands` instead so that I can use the `--no-daemon` under the skin. CC: @JohanLorenzo

taskcluster/ci/lint/kind.yml
JohanLorenzo: I think it's okay in this case because the docker image doesn't contain intermediary build artifacts anymore. So, I don't think there's anything to clean yet.

taskcluster/ci/module-build/kind.yml
JohanLorenzo: > Unclear to me whether chaining the gradlewcommands is the same as running them separately, one after another.

It'll be in one big `gradlew` command. So gradle will be in charge of scheduling the each step the way it wants. Like @badboy said, it's faster. In some circumstances, you may want to break the `gradlew` command into several pieces. For instance: you want to be sure that one command is finished before the next one starts. That said, in the general case you want them all in one big command.

When you have one big command, `--no-deamon` is the way to go, I think. This way, you don't have to spin a deamon that will be used once. If you have several commands, then you likely want to start the daemon.

MihaiTabara:
@eoger ready to land. I'll do it as soon as you're online, just in case.

PR #3116, opened 2020-05-14
Migrate webext data from old kinto databases

mhammond:
A bit of a WIP - a few obvious cleanups necessary, but close enough to get a first pass at a review.

### DIFF
--- a/components/webext-storage/src/migration.rs
+++ b/components/webext-storage/src/migration.rs
@@ -1,0 +1,335 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::*;
+use rusqlite::{Connection, OpenFlags, Transaction, NO_PARAMS};
+use serde_json::{Map, Value};
+use sql_support::ConnExt;
+use std::path::PathBuf;
+
+// Simple migration from the "old" kinto-with-sqlite-backing implementation
+// to ours.
+// Could almost be trivially done in JS using the regular public API if not
+// for:
+// * We don't want to enforce the same quotas when migrating.
+// * We'd rather do the entire migration in a single transaction for perf
+//   reasons.
+
+// The sqlite database we migrate from has a very simple structure:
+// * table collection_data with columns collection_name, record_id and record
+// * `collection_name` is a string of form "default/{extension_id}"
+// * `record_id` is `key-{key}`
+// * `record` is a string with json, of form: {
+//     id: {the record id repeated},
+//     key: {the key},
+//     data: {the actual data},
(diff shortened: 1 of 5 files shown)


components/webext-storage/src/migration.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+        if &self.col_name[..8] != "default/" {
+            log::trace!("collection_name of '{}' isn't ours", self.col_name);
+            return None;
+        }
+        let ext_id = &self.col_name[8..];
+        let mut record_map = match serde_json::from_str(&self.record) {
+            Ok(Value::Object(m)) => m,
+            Ok(o) => {
+                log::trace!("record value is json, but not an object: {}", o);
```
thomcc, line 64: We'll want to record this information for telemetry... Unless that happens later.

It's probably worth logging what happened (but not the data, of course) at a higher level than trace as well.

components/webext-storage/src/migration.rs
```diff
@@ -0,0 +1,335 @@
...
+                return None;
+            }
+        };
+        Some(Parsed { ext_id, key, data })
+    }
+}
+
+// The info we parse from the raw DB strings.
+struct Parsed<'a> {
+    ext_id: &'a str,
```
thomcc, line 97: It's probably worth just making this a String since we to_string it later anyway

components/webext-storage/src/migration.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+fn read_rows(filename: &PathBuf) -> Result<Vec<LegacyRow>> {
+    let flags = OpenFlags::SQLITE_OPEN_NO_MUTEX | OpenFlags::SQLITE_OPEN_READ_ONLY;
+    let src_conn = Connection::open_with_flags(&filename, flags)?;
+    let mut stmt = src_conn.prepare(
+        "SELECT collection_name, record FROM collection_data
+         ORDER BY collection_name",
+    )?;
+    let rows = stmt
```
thomcc, line 159: We'll need initial and final counts of these for telemetry... Maybe we can just get them later though. But probably not counting non-default/ would be good.

But I guess that can be done later, too.

components/webext-storage/src/store.rs
```diff
@@ -116,6 +117,16 @@ impl Store {
     pub fn close(self) -> result::Result<(), (Store, Error)> {
         self.db.close().map_err(|(db, err)| (Store { db }, err))
     }
+
+    /// Migrates data from a database in the format of the "old" kinto
+    /// implementation. Returns the count of webextensions for whom data was
+    /// migrated.
+    pub fn migrate(self, filename: &str) -> Result<usize> {
```
thomcc, line 124: I'd like this to return more information. In particular:

- number of total records we tried to migrate
- number of records successfully migrated
- if we had any recoverable internal errors
- whether not any of the records overflowed quotas.

PR #3132, opened 2020-05-19
Use thiserror and anyhow to replace failure

tarikeshaq:

- [x] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md]
- [x] **Dependencies**: This PR follows our [dependency management guidelines]

Although tons of files are changed, 99.9% of it is boilerplate changing #[derive(failure::Fail)] into deriving thiserror instead. Was hoping to start off by one component and put a draft on that, but some were dependent on each other and having a full change allowed me to uncover a few things:

- Some small (Mozilla written) dependencies that still use failure, and should probably be changed before this is merged.
  - ~~hawk~~
  - ~~ece~~
  - ~~find_places_db~~

- I had to use [backtrace](https://crates.io/crates/backtrace) to support backtracing since [std::backtrace](https://doc.rust-lang.org/std/backtrace/index.html) is only available on nightly
- Implemented my own Context to replace failure::Context

Files that have the main changes:
components/support/error/src/lib.rs
all the component's src/error.rs
Cargo.toml's to replace failure with thiserror and anyhow

~~Note: There are a couple of places (namely where the failure::Fail using dependencies are consumed) that I'll need to make a small edit in once the dependencies are updated.~~

### DIFF
--- a/components/fxa-client/Cargo.toml
+++ b/components/fxa-client/Cargo.toml
@@ -9,7 +9,6 @@ exclude = ["/android", "/ios"]
 [dependencies]
 base64 = "0.12"
 byteorder = "1.3"
-failure = "0.1"
 hex = "0.4"
 lazy_static = "1.4"
 log = "0.4"
@@ -24,6 +23,7 @@ ffi-support = "0.4"
 viaduct = { path = "../viaduct" }
 rc_crypto = { path = "../support/rc_crypto", features = ["ece", "hawk"] }
 error-support = { path = "../support/error" }
+thiserror = "1.0"
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
@@ -31,6 +31,7 @@ cli-support = { path = "../support/cli" }
 dialoguer = "0.6"
 webbrowser = "0.5"
 mockiato = "0.9"
+anyhow = "1.0"
 
 [features]
 default = []
--- a/components/places/Cargo.toml
+++ b/components/places/Cargo.toml
@@ -19,7 +19,6 @@ log = "0.4"
 lazy_static = "1.4"
 url = { version = "2.1", features = ["serde"] }
 percent-encoding = "2.1"
-failure = "0.1"
 caseless = "0.2"
 sql-support = { path = "../support/sql" }
 ffi-support = "0.4"
@@ -32,6 +31,8 @@ dogear = "0.4"
 interrupt-support = { path = "../support/interrupt" }
 error-support = { path = "../support/error" }
 sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random"]}
+thiserror = "1.0"
+anyhow = "1.0"
 
 [dependencies.rusqlite]
 version = "0.23.1"
@@ -40,7 +41,7 @@ features = ["functions", "bundled"]
 [dev-dependencies]
 more-asserts = "0.2"
 env_logger = "0.7"
-find-places-db = "0.1"
+find-places-db = "0.2"
 clap = "2.33"
 structopt = "0.3"
 tempfile = "3.1"
--- a/components/places/src/error.rs
+++ b/components/places/src/error.rs
@@ -4,78 +4,76 @@
 
 use crate::storage::bookmarks::BookmarkRootGuid;
 use crate::types::BookmarkType;
-use failure::Fail;
 use interrupt_support::Interrupted;
 use serde_json::Value as JsonValue;
-
 // Note: If you add new error types that should be returned to consumers on the other side of the
 // FFI, update `get_code` in `ffi.rs`
-#[derive(Debug, Fail)]
+#[derive(Debug, thiserror::Error)]
 pub enum ErrorKind {
-    #[fail(display = "Invalid place info: {}", _0)]
+    #[error("Invalid place info: {0}")]
     InvalidPlaceInfo(InvalidPlaceInfo),
 
-    #[fail(display = "The store is corrupt: {}", _0)]
+    #[error("The store is corrupt: {0}")]
     Corruption(Corruption),
 
-    #[fail(display = "Error synchronizing: {}", _0)]
-    SyncAdapterError(#[fail(cause)] sync15::Error),
+    #[error("Error synchronizing: {0}")]
+    SyncAdapterError(#[from] sync15::Error),
 
-    #[fail(display = "Error merging: {}", _0)]
-    MergeError(#[fail(cause)] dogear::Error),
+    #[error("Error merging: {0}")]
+    MergeError(#[from] dogear::Error),
 
-    #[fail(display = "Error parsing JSON data: {}", _0)]
-    JsonError(#[fail(cause)] serde_json::Error),
+    #[error("Error parsing JSON data: {0}")]
+    JsonError(#[from] serde_json::Error),
 
-    #[fail(display = "Error executing SQL: {}", _0)]
-    SqlError(#[fail(cause)] rusqlite::Error),
+    #[error("Error executing SQL: {0}")]
+    SqlError(#[from] rusqlite::Error),
 
-    #[fail(display = "Error parsing URL: {}", _0)]
-    UrlParseError(#[fail(cause)] url::ParseError),
+    #[error("Error parsing URL: {0}")]
+    UrlParseError(#[from] url::ParseError),
 
-    #[fail(display = "A connection of this type is already open")]
+    #[error("A connection of this type is already open")]
     ConnectionAlreadyOpen,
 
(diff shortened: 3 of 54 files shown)


components/support/error/src/lib.rs
```diff
@@ -1,60 +1,66 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 /// Define a wrapper around the the provided ErrorKind type.
 /// See also `define_error` which is more likely to be what you want.
+
 #[macro_export]
 macro_rules! define_error_wrapper {
```
eoger, line 8: @thomcc Do we still need these wrappers? I would have thought `thiserror` takes care of this for us.

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
     ($Kind:ty) => {
-        /// Re-exported, so that using crate::error::* gives you the .context()
-        /// method, which we don't use much but should *really* use more.
-        pub use failure::ResultExt;
-
         pub type Result<T, E = Error> = std::result::Result<T, E>;
+        use std::fmt;
+        use std::fmt::{Debug, Display};
+        struct Context<T: Display + Sync + 'static> {
```
thomcc, line 13: It's probably better to remove the generic parameter, rename this to ErrorData or something, and replace `context: T` with `kind: ErrorKind`.

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
+
 #[macro_export]
 macro_rules! define_error_wrapper {
     ($Kind:ty) => {
-        /// Re-exported, so that using crate::error::* gives you the .context()
-        /// method, which we don't use much but should *really* use more.
-        pub use failure::ResultExt;
-
         pub type Result<T, E = Error> = std::result::Result<T, E>;
+        use std::fmt;
```
thomcc, line 11: It's not good to have misc `use` in item-level macros, since it brings things to scope unexpectedly. The old code had `pub use failure::ResultExt`, but that's present in order to expose some useful extension trait methods.

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
+        }
 
-        #[derive(Debug)]
-        pub struct Error(Box<failure::Context<$Kind>>);
-
-        impl failure::Fail for Error {
-            fn cause(&self) -> Option<&dyn failure::Fail> {
-                self.0.cause()
+        impl<T: Display + Send + Sync + 'static> Context<T> {
+            fn new(context: T) -> Self {
```
thomcc, line 19: Should probbably be #[cold]

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
 
-            fn backtrace(&self) -> Option<&failure::Backtrace> {
-                self.0.backtrace()
+            fn backtrace(&self) -> Option<&Backtrace> {
+                Some(&self.backtrace)
             }
 
-            fn name(&self) -> Option<&str> {
-                self.0.name()
+            fn get_context(&self) -> &T {
```
thomcc, line 30: Probably removable given that we expose `fn kind`

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
-                Error(Box::new(failure::Context::new(kind)))
+            fn from(ctx: $Kind) -> Error {
+                Error(Box::new(Context::new(ctx)))
             }
         }
     };
 }
 
-/// Define a set of conversions from external error types into the provided
+/// Define a conversion from external error types into the provided
```
thomcc, line 69: It's still a set of conversions. One per external error type.

That said, I think for the most part we'd rather use thiserror instead of define_error_conversions...

components/tabs/examples/tabs_sync.rs
```diff
@@ -10,7 +10,7 @@ use cli_support::prompt::prompt_char;
 use clipboard::{ClipboardContext, ClipboardProvider};
 use tabs::{RemoteTab, TabsEngine};
 
-type Result<T> = std::result::Result<T, failure::Error>;
+type Result<T> = std::result::Result<T, anyhow::Error>;
```
thomcc, line 13: In general this and a few other places should just be `use anyhow::Result`.

Note that anyhow::Result does the trick where you can do `Result<T>` for `Result<T, anyhow::Error>` while letting Result<T, E> still work (this code predates that trick working at all, though)

components/support/error/src/lib.rs
```diff
@@ -1,79 +1,77 @@
...
-        pub struct Error(Box<failure::Context<$Kind>>);
-
-        impl failure::Fail for Error {
-            fn cause(&self) -> Option<&dyn failure::Fail> {
-                self.0.cause()
+        impl<T: Display + Send + Sync + 'static> Context<T> {
+            fn new(context: T) -> Self {
+                Context {
+                    context,
+                    backtrace: Backtrace::new(),
```
thomcc, line 22: This should be new_unresolved and defer the symbolication step until printing. It's subtle since failure does a lot of nice optimizations for us, but hides the fact that it does them.

See:
- https://github.com/rust-lang-nursery/failure/blob/master/src/backtrace/mod.rs#L7-L29
- https://github.com/rust-lang-nursery/failure/blob/master/src/backtrace/internal.rs

components/support/rc_crypto/Cargo.toml
```diff
@@ -10,13 +10,13 @@ crate-type = ["lib"]
...
-failure = "0.1"
-failure_derive = "0.1"
+thiserror = "1.0"
 error-support = { path = "../error" }
 nss = { path = "nss" }
 libsqlite3-sys = { version = "0.18.0", features = ["bundled"] }
-hawk = { version = "3.1", default-features = false, optional = true }
+hawk = { version = "3.2", default-features = false, optional = true }
 ece = { version = "1.1.2", default-features = false, features = ["serializable-keys"], optional = true }
+backtrace = "0.3.48"
```
vladikoff, line 19: make this just 0.3?

components/fxa-client/Cargo.toml
```diff
@@ -24,13 +23,17 @@ ffi-support = "0.4"
...
+backtrace = "0.3"
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
 cli-support = { path = "../support/cli" }
 dialoguer = "0.6"
 webbrowser = "0.5"
 mockiato = "0.9"
+anyhow = "1.0"
+
```
vladikoff, line 36: nit: double new line before `[features]`, not sure if there's something that will format the `.toml` file but remove the extra line somehow

components/places/src/error.rs
```diff
@@ -4,78 +4,76 @@
...
 
-    #[fail(display = "Synced bookmark has unsupported validity {}", _0)]
+    #[error("Synced bookmark has unsupported validity {0}")]
     UnsupportedSyncedBookmarkValidity(u8),
 
     // This will happen if you provide something absurd like
     // "/" or "" as your database path. For more subtley broken paths,
     // we'll likely return an IoError.
-    #[fail(display = "Illegal database path: {:?}", _0)]
+    #[error("Illegal database path: {0:?}")]
```
vladikoff, line 64: this is an interesting way to format the error arg, does that work?

tarikeshaq: According to https://github.com/dtolnay/thiserror/blob/master/README.md it translates to `write!("{:?}", self.0)`

vladikoff: cool!

components/places/Cargo.toml
```diff
@@ -32,6 +31,9 @@ dogear = "0.4"
 interrupt-support = { path = "../support/interrupt" }
 error-support = { path = "../support/error" }
 sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random"]}
+thiserror = "1.0"
+anyhow = "1.0"
+backtrace = "0.3"
```
thomcc, line 36: Nit (which will probably cause a bit of churn): We should probably just be re-exporting backtrace from error-support, since it's a bit weird that you have to have the dep when no code directly depends on it.

something like:

```rust
/// Re-export of the `backtrace` crate for use in macros and
/// to ensure the needed version is kept in sync in dependents.
pub use backtrace;
```

And inside the macros you would do e.g. `$crate::backtrace::Backtrace` to refer to the backtrace object.

tarikeshaq: Ooo that is smart!! Added it in my latest push 🕺

PR #3144, opened 2020-05-22
Add an API to get the top frecent sites

### LINKED ISSUE #2163: Add an API to fetch history pages for top sites
The Fenix team would like to fetch a list of top frecent pages for their Top Sites view (mozilla-mobile/fenix#6190). We currently have APIs for fetching visits by time range, and paging through visits, but (IIUC) nothing that orders by frecency and returns pages, not individual visits.

The API would probably look something like `fun getTopFrecentUrlsExcluding(excluded: List<String>, limit: Int): List<String>`, and the query feels like it would be similar to what [New Tab on Desktop](https://searchfox.org/mozilla-central/rev/e7c61f4a68b974d5fecd216dc7407b631a24eb8f/toolkit/modules/NewTabUtils.jsm#1200-1346) does today. We'd exclude pages below a certain threshold, and hidden pages.

We could probably just return the URL strings, especially if we're planning to let users customize titles. However, if we also need the page title, we may need to make a `data class PageInfo` with `url` and `title` properties (and maybe the frecency, too?). `VisitInfo` doesn't feel like quite the right fit here.

CC @vesta0 and @boek for product and engineering guidance about how we define top sites.

┆Issue is synchronized with this [Jira Story](https://jira.mozilla.com/browse/SYNC-359)
┆Story Points: 0
┆Epic: Fenix Migration

gabrielluong:
Fixes #2163

### DIFF
--- a/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
@@ -281,6 +281,18 @@ open class PlacesReaderConnection internal constructor(connHandle: Long) :
         }
     }
 
+    override fun getTopFrecentSiteInfos(numItems: Int): List<TopFrecentSiteInfo> {
+        val infoBuffer = rustCall { error ->
+            LibPlacesFFI.INSTANCE.places_get_top_frecent_site_infos(this.handle.get(), numItems, error)
+        }
+        try {
+            val infos = MsgTypes.TopFrecentSiteInfos.parseFrom(infoBuffer.asCodedInputStream()!!)
+            return TopFrecentSiteInfo.fromMessage(infos)
+        } finally {
+            LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(infoBuffer)
+        }
+    }
+
     override fun getVisited(urls: List<String>): List<Boolean> {
         // Note urlStrings has a potential footgun in that StringArray has a `size()` method
         // which returns the size *in bytes*. Hence us using urls.size (which is an element count)
@@ -824,7 +836,17 @@ interface ReadableHistoryConnection : InterruptibleConnection {
     fun matchUrl(query: String): String?
 
     /**
+     * Returns a list of the top frecent site infos limited by the given number of items
+     * sorted by most to least frecent.
+     *
+     * @param numItems the number of top frecent sites to return in the list.
+     * @return a list of the top frecent site infos sorted by most to least frecent.
+     */
+    fun getTopFrecentSiteInfos(numItems: Int): List<TopFrecentSiteInfo>
+
+    /**
      * Maps a list of page URLs to a list of booleans indicating if each URL was visited.
+     *
      * @param urls a list of page URLs about which "visited" information is being requested.
      * @return a list of booleans indicating visited status of each
      * corresponding page URI from [urls].
@@ -1155,6 +1177,29 @@ data class SearchResult(
         internal fun fromCollectionMessage(msg: MsgTypes.SearchResultList): List<SearchResult> {
             return msg.resultsList.map {
                 fromMessage(it)
+            }
+        }
+    }
+}
+
+/**
+ * Information about a top frecent site. Returned by `PlacesAPI.getTopFrecentSiteInfos`.
+ */
+data class TopFrecentSiteInfo(
+    /**
+     * The URL of the page that was visited.
+     */
+    val url: String,
+
+    /**
+     * The title of the page that was visited, if known.
+     */
+    val title: String?
+) {
+    companion object {
+        internal fun fromMessage(msg: MsgTypes.TopFrecentSiteInfos): List<TopFrecentSiteInfo> {
+            return msg.infosList.map {
+                TopFrecentSiteInfo(url = it.url, title = it.title)
             }
         }
     }
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -10,7 +10,10 @@ use crate::hash;
 use crate::history_sync::store::{
     COLLECTION_SYNCID_META_KEY, GLOBAL_SYNCID_META_KEY, LAST_SYNC_META_KEY,
 };
-use crate::msg_types::{HistoryVisitInfo, HistoryVisitInfos, HistoryVisitInfosWithBound};
+use crate::msg_types::{
+    HistoryVisitInfo, HistoryVisitInfos, HistoryVisitInfosWithBound, TopFrecentSiteInfo,
+    TopFrecentSiteInfos,
+};
 use crate::observation::VisitObservation;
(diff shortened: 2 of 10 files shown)


gabrielluong:
@linacambridge This should be ready for review. I don't permissions to request reviews.

mhammond:
A bit of a drive-by - I don't think the name `topSiteFrecency` is very descriptive - I had no idea what it meant when first reading this. More generally though, I'm not sure it makes sense to have this as part of the API - I don't think we want consumers of this API to try and understand what an absolute frecency value actually means - ie, the number should be seen as only relative to other frecencies (ie, more or less frecent than another site) but not that, say "anything under X isn't frecent" or "anything over X is very frecent"

gabrielluong:
> A bit of a drive-by - I don't think the name `topSiteFrecency` is very descriptive - I had no idea what it meant when first reading this. More generally though, I'm not sure it makes sense to have this as part of the API - I don't think we want consumers of this API to try and understand what an absolute frecency value actually means - ie, the number should be seen as only relative to other frecencies (ie, more or less frecent than another site) but not that, say "anything under X isn't frecent" or "anything over X is very frecent"

Totally understandable. Some of the options and naming that were used here originated from prior art of NewTabUtils.jsm#getTopFrecentSites https://searchfox.org/mozilla-central/rev/fc91a093e40dde71d10ad219946b8ae775aca9eb/toolkit/modules/NewTabUtils.jsm#1279-1284,1367-1390. 

I don't personally feel too strongly about adding this as part of the API. My main concern was that `getTopFrecentSites` would start returning the first visited site you go to on a fresh profile of  Fenix. Adding `topSiteFrecency` or perhaps better named `frecencyThreshold` provided some flexibility to avoid that.

mhammond:
> My main concern was that `getTopFrecentSites` would start returning the first visited site you go to on a fresh profile of Fenix.

My main concern is that if Fenix starts specifying a frecency of (say) 200 to avoid this problem, it means we will be largely unable to ever change the algorithm - and we certainly want to reserve the right to change it (eg, see #610)

Indeed, this would argue for never actually returning the frecency - if we return the value it's inevitable that it will end up with meaning. I wonder if we should just return the order?

While I don't see that behaviour on a fresh profile as an actual problem, if we really do want to prevent that, I guess there are a couple of options:

* Use something like (local/remote) visit count as a proxy - eg, maybe a site must be visited twice to qualify? But even this is subtle - something visited twice by following a link isn't necessarily "better" than a site visited once because the user explicitly typed the address. Similarly, the site having 1 visit but being bookmarked probably means it should be returned.

* Have the implementation handle it internally - ie, still filter, but filter internally without the app having the ability to control the threshold. Given how subtle frecency is, this seems the best option, but it does remove flexibility from the containing app.

@lina probably has opinions here and I'll be happy to defer to them, so let's see what she says tomorrow.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -1160,6 +1183,34 @@ data class SearchResult(
...
 }
 
+/**
+ * Information about a top frecent site. Returned by `PlacesAPI.getTopFrecentSiteInfos`.
+ */
+data class TopFrecentSiteInfo(
+    /**
+     * The frecency.
+     */
+    val frecency: Long,
```
linacambridge, line 1193: I think this is OK to expose to a-c, though. The cat's out of the bag because we _do_ expose frecency in autocomplete results, which a-c [renames](https://github.com/mozilla-mobile/android-components/blob/cb312ff60cc8df65b22edbf9f56e099ccd4e89ba/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt#L103) to "score" and doesn't make any guarantees beyond ["higher = more relevant"](https://github.com/mozilla-mobile/android-components/blob/cb312ff60cc8df65b22edbf9f56e099ccd4e89ba/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt#L188-L189).

components/places/src/storage/history.rs
```diff
@@ -1177,6 +1180,29 @@ pub fn get_visited_urls(
...
+           NOT hidden
+         ORDER BY frecency DESC
+         LIMIT :limit",
+        rusqlite::named_params! {
+            ":limit": num_items,
+            ":frecencyThreshold": top_site_frecency
+        },
+        TopFrecentSiteInfo::from_row,
+    )?;
+    Ok(TopFrecentSiteInfos { infos })
```
linacambridge, line 1203: Do we need to do any collapsing or deduping here, or would you like to save that for a follow-up? I'm fine with getting this landed as-is so you can build out the a-c side, and copying more of Desktop later if we decide we need it.

gabrielluong:
Thanks for the reviews and comments. I will make the changes to remove the frecency params before asking for this to land.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -823,8 +835,19 @@ interface ReadableHistoryConnection : InterruptibleConnection {
      */
     fun matchUrl(query: String): String?
 
+    /**
+     * Returns a list of the top frecent site infos and limits the number of items returned
+     * in the list based on the given number of items and minimum frecency threshold.
+     *
+     * @param numItems the number of top frecent sites to return in the list.
+     * @param topSiteFrecency: Minimum amount of frecency for a site.
+     * @return a list of the top frecent site infos.
```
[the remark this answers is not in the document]
gabrielluong: Yes, added that to the comment.

components/places/src/storage/history.rs
gabrielluong: Gonna think about this as a follow up since we can do that either here or in A-C.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
gabrielluong: I am gonna remove this for the time being since I don't feel I will need this in the short term.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -823,8 +835,19 @@ interface ReadableHistoryConnection : InterruptibleConnection {
      */
     fun matchUrl(query: String): String?
 
+    /**
+     * Returns a list of the top frecent site infos and limits the number of items returned
+     * in the list based on the given number of items and minimum frecency threshold.
+     *
+     * @param numItems the number of top frecent sites to return in the list.
+     * @param topSiteFrecency: Minimum amount of frecency for a site.
```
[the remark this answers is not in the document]
gabrielluong: Gonna remove `topSiteFrecency` and re-evaluate this for later.

PR #3156, opened 2020-05-26
Record migration telemetry for webext-storage

### LINKED ISSUE #3133: Expose information needed for migration telemetry from webext migration
In #3116 we landed support for migrating kinto data to the new webext database. However, it doesn't contain any information about the migration, and that was deferred until later.

In https://bugzilla.mozilla.org/show_bug.cgi?id=1628893#c6 I gave an overview of some of the things we might want to include for migration info as part of a "get data-reviewers on the same page" comment.

Currently, for migrations, it includes adding the following to the sync ping. 

```
        "migrations": [{
            "type": "webext-storage",
            // The number of records detected prior to migration.
            "numDetected": integer,
            // The number of records successfully migrated
            "numSucceeded": integer,
            // The number of records we failed to migrate.
            // `numFailed + numSucceeded` should equal `numDetected`.
            "numFailed": integer,
            // The amount of time the migration took in milliseconds
            "duration": integer,
            // True if any exceptions or `Result::Err`s were produced when
            // attempting to migrate.
            "anyFailures": boolean,
            // If migration failed, indicates why. Optional
            "failureReason": /* existing sync ping error instance */,
            // True if the existing table exceeded quotas (not an error
            // for migration)
            "quotasExceeded": boolean,
        }]
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1301)
┆Story Points: 3
┆Epic: Extension pref storage & sync updated to Sync1.5

thomcc:
Fixes #3133.

Need to figure out the off-by-one in the tests but other than that this is done.

### DIFF
--- a/components/webext-storage/src/migration.rs
+++ b/components/webext-storage/src/migration.rs
@@ -3,9 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::error::*;
-use rusqlite::{Connection, OpenFlags, Transaction, NO_PARAMS};
+use rusqlite::{named_params, Connection, OpenFlags, Transaction, NO_PARAMS};
 use serde_json::{Map, Value};
 use sql_support::ConnExt;
+use std::collections::HashSet;
 use std::path::Path;
 
 // Simple migration from the "old" kinto-with-sqlite-backing implementation
@@ -107,12 +108,12 @@ struct Parsed<'a> {
     data: serde_json::Value,
 }
 
-pub fn migrate(tx: &Transaction<'_>, filename: &Path) -> Result<usize> {
+pub fn migrate(tx: &Transaction<'_>, filename: &Path) -> Result<MigrationInfo> {
     // We do the grouping manually, collecting string values as we go.
     let mut last_ext_id = "".to_string();
     let mut curr_values: Vec<(String, serde_json::Value)> = Vec::new();
-    let mut num_extensions = 0;
-    for row in read_rows(filename) {
+    let (rows, mut mi) = read_rows(filename);
+    for row in rows {
         log::trace!("processing '{}' - '{}'", row.col_name, row.record);
         let parsed = match row.parse() {
             Some(p) => p,
@@ -122,8 +123,9 @@ pub fn migrate(tx: &Transaction<'_>, filename: &Path) -> Result<usize> {
         if parsed.ext_id != last_ext_id {
             if last_ext_id != "" && !curr_values.is_empty() {
                 // a different extension id - write what we have to the DB.
-                do_insert(tx, &last_ext_id, curr_values)?;
-                num_extensions += 1;
+                let entries = do_insert(tx, &last_ext_id, curr_values)?;
+                mi.extensions_successful += 1;
+                mi.entries_successful += entries;
             }
             last_ext_id = parsed.ext_id.to_string();
             curr_values = Vec::new();
@@ -141,32 +143,34 @@ pub fn migrate(tx: &Transaction<'_>, filename: &Path) -> Result<usize> {
     // and the last one
     if last_ext_id != "" && !curr_values.is_empty() {
         // a different extension id - write what we have to the DB.
-        do_insert(tx, &last_ext_id, curr_values)?;
-        num_extensions += 1;
+        let entries = do_insert(tx, &last_ext_id, curr_values)?;
+        mi.extensions_successful += 1;
+        mi.entries_successful += entries;
     }
-    log::info!("migrated {} extensions", num_extensions);
-    Ok(num_extensions)
+    log::info!("migrated {} extensions: {:?}", mi.extensions_successful, mi);
+    Ok(mi)
 }
 
-fn read_rows(filename: &Path) -> Vec<LegacyRow> {
+fn read_rows(filename: &Path) -> (Vec<LegacyRow>, MigrationInfo) {
     let flags = OpenFlags::SQLITE_OPEN_NO_MUTEX | OpenFlags::SQLITE_OPEN_READ_ONLY;
     let src_conn = match Connection::open_with_flags(&filename, flags) {
         Ok(conn) => conn,
         Err(e) => {
             log::warn!("Failed to open the source DB: {}", e);
-            return Vec::new();
+            return (Vec::new(), MigrationInfo::open_failure());
         }
     };
     // Failure to prepare the statement probably just means the source DB is
     // damaged.
     let mut stmt = match src_conn.prepare(
         "SELECT collection_name, record FROM collection_data
+         WHERE collection_name != 'default/storage-sync-crypto'
          ORDER BY collection_name",
     ) {
         Ok(stmt) => stmt,
         Err(e) => {
             log::warn!("Failed to prepare the statement: {}", e);
-            return Vec::new();
+            return (Vec::new(), MigrationInfo::open_failure());
         }
     };
     let rows = match stmt.query_and_then(NO_PARAMS, |row| -> Result<LegacyRow> {
@@ -178,18 +182,37 @@ fn read_rows(filename: &Path) -> Vec<LegacyRow> {
         Ok(r) => r,
         Err(e) => {
             log::warn!("Failed to read any rows from the source DB: {}", e);
-            return Vec::new();
+            return (Vec::new(), MigrationInfo::open_failure());
         }
     };
+    let all_rows: Vec<Result<LegacyRow>> = rows.collect();
+    let entries = all_rows.len();
+    let successful_rows: Vec<LegacyRow> = all_rows.into_iter().filter_map(Result::ok).collect();
+    let distinct_extensions: HashSet<_> = successful_rows.iter().map(|c| &c.col_name).collect();
 
-    rows.filter_map(Result::ok).collect()
+    let mi = MigrationInfo {
+        entries,
+        extensions: distinct_extensions.len(),
+        // Populated later.
+        extensions_successful: 0,
+        entries_successful: 0,
+        open_failure: false,
+    };
+
+    (successful_rows, mi)
 }
 
-fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) -> Result<()> {
+/// Insert the extension and values. If there are multiple values with the same
+/// key (which shouldn't be possible but who knows, database corruption causes
+/// strange things), chooses an arbitrary one. Returns the number of entries
+/// inserted, which could be different from `vals.len()` if multiple entries in
+/// `vals` have the same key.
+fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) -> Result<usize> {
     let mut map = Map::with_capacity(vals.len());
     for (key, val) in vals {
         map.insert(key, val);
     }
+    let num_entries = map.len();
     tx.execute_named_cached(
         "INSERT OR REPLACE INTO storage_sync_data(ext_id, data, sync_change_counter)
          VALUES (:ext_id, :data, 1)",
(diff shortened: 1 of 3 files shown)


thomcc:
(Uh, I had to do some serious git surgery in addition to mostly rewriting this patch. Hopefully nothing went wrong)

linacambridge:
Higher-level comment: I wonder if it's worth writing the migration info to a separate table (`meta`?), instead of returning it from `migrate`, and having the Sync telemetry code read and clear it later.

Returning it is a little tricky because Desktop [migrates from Kinto lazily](https://hg.mozilla.org/mozilla-central/file/2692b8d159d5c4db5589e79e2334645984a102d7/toolkit/components/extensions/storage/webext_storage_bridge/src/store.rs#l214) when you open a new database for the first time—which can happen by either an extension or Sync, and Sync might not be configured yet. It's not yet possible to use telemetry directly from rust-xpcom—the `Telemetry::Accumulate` APIs are C++-only, and `Services.telemetry` takes `jsval`s.

But what if we recorded it under a meta key, and sent and cleared it the first time we send a sync ping?

thomcc:
Hmm... That's a really good point.

I guess we have a little more time now that this has already missed 78 and will need uplift anyway, so we should do the thing that's more likely to lead to something with higher reliability...

thomcc:
@linacambridge My only real problem with this is that it's a bunch of additional complexity...

That said, I can't think of a better way. So if I'm interpreting you right, what you're suggesting is:

1. Either add a new `migration_info` table (or maybe we could do with just a few meta keys?)
2. Add a new `get_migration_info` API returning Option<MigrationInfo>
3. Add a `clear_migration_info` API to clear it -- after clearing the `get_` API would return None.

Then, telemetry code would call `get_migration_info`, and after submitting the ping, `clear_migration_info`.

Or would it be better for it to be a single API providing `take` semantics?

linacambridge:
I definitely agree about the extra complexity. I didn't think too much past storing the migration info temporarily, and it feels a little clunky...there are likely better ways to persist it. @mhammond, WDYT?

1. I was thinking just adding `#[derive(Serialize)]` to `MigrationInfo` and storing the entire struct as a JSON blob in the `meta` table, under a `migration_info` key.
2. A single API with `take` semantics sounds good to me. There's a window of time between fetching them from the database and sending the ping where quitting the browser could cause us to lose the stats...but I don't know if we can guarantee "exactly once" reporting. If we wait to delete until after we've submitted the ping, and that fails or gets interrupted, we might submit stats multiple times.

thomcc:
Oh, that's a simpler solution then.

components/webext-storage/src/migration.rs
```diff
@@ -107,29 +108,42 @@ struct Parsed<'a> {
...
+            curr_quota_bytes = 0;
         }
         // no 'else' here - must also enter this block on ext_id change.
         if parsed.ext_id == last_ext_id {
+            let quota_bytes = crate::api::get_quota_size_of(&parsed.key, &parsed.data);
+            curr_quota_bytes += quota_bytes;
+            if quota_bytes > crate::api::SYNC_QUOTA_BYTES_PER_ITEM
+                || curr_quota_bytes > crate::api::SYNC_QUOTA_BYTES
+            {
+                mi.overflowed_quotas = true;
```
linacambridge, line 145: I wonder if it's worth recording how much we overflowed the quota, using buckets and counts for how many times we overflowed in each one. (Say, 1-10 bytes, 11-20, 21-50, 51-100, 100-500, 501-1k, and 1k+, and adding one to the bucket when we spill over).

On the other hand, I don't know what we'll _do_ with that information. I'm imagining we could use it to bump the quota, and identify interesting trends—if most spillovers are 100 bytes or fewer, that's more actionable than many kilobytes. Buuuut, that would mean diverging from Chrome, and I suspect there's a long tail of spillovers that'll eventually bump up against the maximum Sync record limit, which we can't sync at all. And analyzing the data like that might be super involved...

I think `overflowed_quotas` is a good start, and bucketing overflows beyond that shouldn't block your PR. But just something I was thinking about while reviewing...

components/webext-storage/src/migration.rs
```diff
@@ -107,29 +108,42 @@ struct Parsed<'a> {
...
-    for row in read_rows(filename) {
+    let mut curr_quota_bytes: usize = 0;
+    let (rows, mut mi) = read_rows(filename);
+    for row in rows {
         log::trace!("processing '{}' - '{}'", row.col_name, row.record);
         let parsed = match row.parse() {
             Some(p) => p,
-            None => continue,
+            None => {
+                mi.non_fatal_errors_seen += 1;
```
linacambridge, line 122: How do you feel about splitting `non_fatal_errors` into `database_errors` and `parse_errors`? `open_failure` is worth calling out separately either way, but I wonder if there's room to distinguish between the database file itself being weird, and having a database that's structurally fine, but contains invalid data.

components/webext-storage/src/migration.rs
```diff
@@ -141,32 +155,52 @@ pub fn migrate(tx: &Transaction<'_>, filename: &Path) -> Result<usize> {
...
         // a different extension id - write what we have to the DB.
-        do_insert(tx, &last_ext_id, curr_values)?;
-        num_extensions += 1;
+        let entries = do_insert(tx, &last_ext_id, curr_values)?;
+        mi.extensions_successful += 1;
+        mi.entries_successful += entries;
     }
-    log::info!("migrated {} extensions", num_extensions);
-    Ok(num_extensions)
+    log::info!("migrated {} extensions: {:?}", mi.extensions_successful, mi);
```
linacambridge, line 162: This is totally fine here, given that we'll want to move to `Log.jsm` logging eventually and it's just one very small struct, but there's a dev-platform post from a while bacl cautioning about printing with `#[derive(Debug)]`, since the extra code can add up, and the majority of our users won't see it.

components/webext-storage/src/migration.rs
```diff
@@ -198,7 +286,30 @@ fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) ->
...
         },
     )?;
-    Ok(())
+    Ok(num_entries)
+}
+
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct MigrationInfo {
+    /// The number of entries (rows in the original table) we attempted to
+    /// migrate. Zero if there was some error in computing this number.
```
linacambridge, line 295: It would be great to clarify what each row in the original table means. Is it a single key-value pair for an extension (so this is all key-value pairs for all extensions), or something different?

components/webext-storage/src/migration.rs
```diff
@@ -198,7 +286,30 @@ fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) ->
...
+    pub entries: usize,
+    /// The number of records we successfully migrated (equal to `entries` for
+    /// entirely successful migrations).
```
linacambridge, line 298: Hmm, this implication of "different kinds of success" is making my Spidey-Sense tingle a little bit 🕷️ What does it mean if we attempted to migrate all entries, but `extensions` doesn't match that count?

components/webext-storage/src/migration.rs
```diff
@@ -198,7 +286,30 @@ fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) ->
...
+    /// table.
+    pub extensions: usize,
+    /// The number of extensions we successfully migrated
+    pub extensions_successful: usize,
+    /// A count of recoverable non-fatal errors seen. Generally this will cause
+    /// the failed count to be nonzero, but there are other cases that can cause
+    /// error (failing to compute # of entries or extensions).
+    pub non_fatal_errors_seen: usize,
+    /// True if we hit at least one quota error.
+    pub overflowed_quotas: bool,
```
linacambridge, line 310: So a quota error is not a `non_fatal_error`, right? But it also means it's not a successful migration (so no +1 to `extensions_successful`)—or does it? We did technically insert it into the DB, even though the extension can't do anything with it, because it won't be able to make any more updates without running into quota errors.

I guess I'd love to discuss more about how all these counts overlap, and don't, and what kind of questions we'll want to answer in analysis.

components/webext-storage/src/migration.rs
```diff
@@ -107,29 +108,42 @@ struct Parsed<'a> {
...
+                mi.extensions_successful += 1;
+                mi.entries_successful += entries;
             }
             last_ext_id = parsed.ext_id.to_string();
             curr_values = Vec::new();
+            curr_quota_bytes = 0;
         }
         // no 'else' here - must also enter this block on ext_id change.
         if parsed.ext_id == last_ext_id {
+            let quota_bytes = crate::api::get_quota_size_of(&parsed.key, &parsed.data);
```
mhammond, line 140: I don't think there's value in doing this - we should be recording when `set` fails due to being over quota.

components/webext-storage/src/migration.rs
```diff
@@ -178,18 +212,72 @@ fn read_rows(filename: &Path) -> Vec<LegacyRow> {
...
+                    non_fatal_errors_seen: 1,
+                    open_failure: true,
+                    ..Default::default()
+                },
+            );
         }
     };
+    let rows: Vec<LegacyRow> = rows.filter_map(Result::ok).collect();
+
+    let entries = src_conn.query_row::<i64, _, _>(
```
mhammond, line 227: I really think we should try to avoid doing extra reads just for telemetry - it's just adding risk without any value. Can we please try and work out how to learn what you think this will tell you without additional io? ISTM that 30% of this file is now dedicated to telemetry which doesn't seem ideal, so let's KISS where we can.

components/webext-storage/src/migration.rs
```diff
@@ -178,18 +212,72 @@ fn read_rows(filename: &Path) -> Vec<LegacyRow> {
...
+
+    (rows, mi)
 }
 
-fn do_insert(tx: &Transaction<'_>, ext_id: &str, vals: Vec<(String, Value)>) -> Result<()> {
+/// Insert the extension and values. If there are multiple values with the same
+/// key (which shouldn't be possible but who knows, database corruption causes
+/// strange things), chooses an arbitrary one. Returns the number of entries
+/// inserted, which could be different from `vals.len()` if multiple entries in
+/// `vals` have the same key.
```
mhammond, line 274: I'm a little suspect that we will want to know precisely how many corrupt databases in the wold causes sqlite to make a mistake about unique values, or what we would do even if we did. Are you sure this distinction is worthwhile?

mhammond:
(and FWIW, depending on what exactly you want to collect, telemetry about usage might be able to be collected entirely on the JS side of the world, which would probably be a bonus, and makes the case for something simple like a hard-coded meta key for stashing this easier to make)

components/webext-storage/src/migration.rs
```diff
@@ -107,23 +108,26 @@ struct Parsed<'a> {
...
     let mut curr_values: Vec<(String, serde_json::Value)> = Vec::new();
-    let mut num_extensions = 0;
-    for row in read_rows(filename) {
+    let (rows, mut mi) = read_rows(filename);
+    for row in rows {
         log::trace!("processing '{}' - '{}'", row.col_name, row.record);
         let parsed = match row.parse() {
             Some(p) => p,
-            None => continue,
+            None => {
```
mhammond, line 120: how come this change? Did a rustfmt update cause it? It seems a step backwards - I don't want to start hating rustfmt as much as I hate eslint ;)

components/webext-storage/src/migration.rs
thomcc: I think we'd want this in the sync-log tbh.

components/webext-storage/src/migration.rs
thomcc: I've killed `non_fatal_errors`, we just have `open_failure` and the counts now.

components/webext-storage/src/migration.rs
thomcc: This has been removed.

components/webext-storage/src/migration.rs
thomcc: There's no more redundancy here.

components/webext-storage/src/migration.rs
thomcc: Weird, not 100% sure. Would have expected cargo fmt to normalize that but it doesn't (an earlier verison of the patch did more here)

thomcc:
> Thanks! I still think the older comments here (from myself and Lina) are worth addressing in some way, even if just to justify.

I've added responses, and made sure nothing got missed. A lot of @linacambridge's comments were about redundancy between the different cases, which I think is gone now?

PR #3159, opened 2020-05-27
[WIP] Experiment with auto-generating FFI and language bindings

rfk:
**DO NOT MERGE. NOT WORKING. MIGHT NEVER WORK.**

I took advantage of the US holiday earlier this week, to hack around with an idea I've been  mulling over for a while: use an abstract IDL to define the interface for a component, and autogenerate a bunch of the FFI and language binding code.

It's very early stage and doesn't actually work yet, but it's far enough along that I could plausibly see it succeeding, and poking at the code could give other people an idea about how it works. So I figured I might as well make it visible :-)

You probably want to start at [the rendered README for the new `uniffi` crate](https://github.com/mozilla/application-services/blob/bindings-autogen-experiment/components/support/uniffi/README.md).


components/support/uniffi/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+Currently totally unimplemented.
+
+A great opportunity for anyone interested to
+dive in! You could try looking at the hand-written Kotlin or Swift code for
+the fxa-client component, and see if you can generate something similar from
+`fxa-client.idl`. Take a look at the way the scaffolding generator works to
+see how to get started.
+
+## What could possibly go wrong?
```
thomcc, line 191: This whole section is a big part of why I never pushed too hard here, especially the bit about "The resulting autogenerated code might be a lot harder to debug when things go wrong".

I am/was very worried that anything that's not very close to a perfect solution would only make things more likely to fail, and I still have that fear.

That said, I'm still very interested in this and am planning on diving in a little more when my schedule frees up a bit (there's a lot of telemetry work that needs to land with a bit of urgency, so I can't in good conscience dig in until that's in a better state)

rfk: Aye. I think it would have been possible for us to go too far too early down this road, before we'd really lived with writing and maintaining the code by hand and understood its failure modes. I'm lightly hopeful that we've been doing this for long enough now that it's time for some automation, but I'm still pretty suspicious of hidden costs.

mhammond: Color me skeptical too - however, a possible benefit is that we will not *need* to debug the generated code. Debugging the code generated by a compiler is tricky too, but it's so rare that we never really consider whether recent compiler changes have made this less or more difficult.

IOW, we are probably suffering from stockholm syndrome :)

again, to be clear, I believe we all agree that (a) this is far from proven but (b) the benefits are such that it's worth exploring :)

rfk:
Since this promises to be a long-lived, very experimental branch, I think it would be easier to manage in a separate repo rather than as a PR on our main repo. I've pulled it out into a standalone spot here:

  https://github.com/rfk/uniffi-rs/

It's deliberately under my personal github so that people don't spot it and think it's more important than it is; if we like where it ends up we can pull it back into this repo, or a separate repo in the mozilla org, or something else clever that we decide at a later date.

rfk:
This all now lives at https://github.com/mozilla/uniffi-rs, we don't need to keep this old PR open.

PR #3161, opened 2020-05-27
Record and return changes made during an extension storage sync (fixes #3157)

### LINKED ISSUE #3157: Record and return changes made during an extension storage sync
This ticket is the a-s side of https://bugzilla.mozilla.org/show_bug.cgi?id=1641005.

@mhammond noticed that `storage.sync` needs to emit observer notifications for changes it makes during a sync, matching the [old Kinto implementation](https://searchfox.org/mozilla-central/rev/7dafc35406b9c945189c617d427f5458933fd3fb/toolkit/components/extensions/ExtensionStorageSyncKinto.jsm#829-867). We already have the old and new values when we build our incoming plan, so we can maybe teach `apply_actions` to return storage changes for them?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1346)
┆Epic: Extension pref storage & sync updated to Sync1.5

mhammond:
If you think this is very close and just have a few minor nits, feel free to address them and land them! But otherwise another day isn't going to hurt.

### DIFF
--- a/components/webext-storage/src/sync/incoming.rs
+++ b/components/webext-storage/src/sync/incoming.rs
@@ -11,6 +11,7 @@ use sql_support::ConnExt;
 use sync15_traits::Payload;
 use sync_guid::Guid as SyncGuid;
 
+use crate::api::{StorageChanges, StorageValueChange};
 use crate::error::*;
 
 use super::{merge, remove_matching_keys, JsonMap, Record};
@@ -23,6 +24,20 @@ pub enum DataState {
     Deleted,
     /// Data exists, as stored in the map.
     Exists(JsonMap),
+}
+
+// A little helper to create a StorageChanges object when we are creating
+// a new value with multiple keys that doesn't exist locally.
+fn changes_for_new_incoming(new: &JsonMap) -> StorageChanges {
+    let mut result = StorageChanges::with_capacity(new.len());
+    for (key, val) in new.iter() {
+        result.push(StorageValueChange {
+            key: key.clone(),
+            old_value: None,
+            new_value: Some(val.clone()),
+        });
+    }
+    result
 }
 
 // This module deals exclusively with the Map inside a JsonValue::Object().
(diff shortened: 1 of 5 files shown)


components/webext-storage/src/sync/incoming.rs
```diff
@@ -236,12 +262,17 @@ pub fn plan_incoming(s: IncomingState) -> IncomingAction {
...
+                    IncomingAction::Merge {
+                        data: local_data,
+                        changes: StorageChanges::new(),
+                    }
                 }
                 (DataState::Deleted, DataState::Deleted, _) => {
                     // We agree with the remote (regardless of what we
                     // have mirrored).
-                    IncomingAction::DeleteLocally
+                    IncomingAction::DeleteLocally {
```
mhammond, line 273: Seems like this should be `DataState::Same` for clarity (although the end result should be the same)

components/webext-storage/src/sync/mod.rs
```diff
@@ -73,29 +94,81 @@ fn merge(mut other: JsonMap, mut ours: JsonMap, parent: Option<JsonMap>) -> Inco
...
+        }
     }
-    ours
+    (ours, changes)
+}
+
+// Fetches the applied changes we stashed in the storage_sync_applied table.
+// We don't convert to JSON because they need to be passed back to the browser
+// as strings anyway. However, we do return one string per extension rather than
+// a massive array.
```
mhammond, line 152: s/array/string/

components/webext-storage/src/sync/incoming.rs
linacambridge: Nice catch! Changed.

components/webext-storage/src/sync/mod.rs
linacambridge: I went ahead and changed this to return a `Vec<SyncedExtensionChanges>` with separate fields for the `ext_id` and `changes`, like we talked about—because Desktop is going to need the `ext_id`, anyway, but we can pass changes through as they are.

PR #3164, opened 2020-05-27
Remove experimental namespace for RemoteTabs

eoger:
It doesn't matter too much (and is probably not even a breaking change) as the sources themselves / the maven package were in a different namespace, but we might as well.


rfk:
I'm sad this `experimental` namespace didn't really work out for us, but thanks for giving it a try :-)

PR #3168, opened 2020-05-27
Switch releases to taskgraph

MihaiTabara:
Last PR that references https://github.com/mozilla/application-services/pull/1945/ and adds taskgraph support for releases.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -5,385 +5,278 @@
 version: 1
 reporting: checks-v1
 policy:
-  pullRequests: public
+    pullRequests: public
 tasks:
-  $let:
-    taskgraph:
-      branch: taskgraph
-      revision: a8366f88fc8b35b284ad550358c89ec10133fb42
-    trustDomain: app-services
-    decision_task_id: {$eval: as_slugid("decision_task")}
-    expires_in: {$fromNow: '1 year'}
-    # We define the following variable at the very top, because they are used in the
-    # default definition
+    - $let:
+          taskgraph:
+              branch: taskgraph
+              revision: cab4565345d0d0effa66f18fe91335dd6d744031
+          trustDomain: app-services
+      in:
+          $let:
+              # Github events have this stuff in different places
+              ownerEmail:
+                  $if: 'tasks_for in ["cron", "action"]'
+                  then: '${tasks_for}@noreply.mozilla.org'
+                  else:
+                      $if: 'tasks_for == "github-push"'
+                      then: '${event.pusher.email}'
+                      # Assume Pull Request
+                      else:
+                          $if: 'tasks_for == "github-pull-request"'
+                          then: '${event.pull_request.user.login}@users.noreply.github.com'
+                          else:
+                              $if: 'tasks_for == "github-release"'
+                              then: '${event.sender.login}@users.noreply.github.com'
 
-    head_rev:
-      $if: 'tasks_for == "github-pull-request"'
-      then: ${event.pull_request.head.sha}
-      else:
-        $if: 'tasks_for == "github-push"'
-        then: ${event.after}
-        else: ${event.release.tag_name}
-
-    repository:
-      $if: 'tasks_for == "github-pull-request"'
-      then: ${event.pull_request.head.repo.html_url}
-      else: ${event.repository.html_url}
-
-    is_repo_trusted:
-      # Pull requests on main repository can't be trusted because anybody can open a PR on it, without a review
-      $if: 'tasks_for in ["github-push", "github-release"] && event.repository.html_url == "https://github.com/mozilla/application-services"'
-      then: true
-      else: false
-
-    user:
-      $if: 'event.sender.login == "dependabot-preview[bot]"'
-      then: dependabot
-      else: ${event.sender.login}
-  in:
-    $let:
-      provisioner_id:
-        $if: 'is_repo_trusted'
-        then: 'app-services-3'
-        else: 'app-services-1'
-      # TODO: revisit once bug 1533314 is done to possibly infer better priorities
-      tasks_priority: highest
-      # Github events have this stuff in different places...
-      ownerEmail:
-          $if: 'tasks_for in ["cron", "action"]'
-          then: '${tasks_for}@noreply.mozilla.org'
-          else:
-              $if: 'tasks_for == "github-push"'
-              then: '${event.pusher.email}'
-              # Assume Pull Request
-              else:
+              baseRepoUrl:
+                  $if: 'tasks_for in ["github-push", "github-release"]'
+                  then: '${event.repository.html_url}'
+                  else:
+                      $if: 'tasks_for == "github-pull-request"'
+                      then: '${event.pull_request.base.repo.html_url}'
+                      else:
+                          $if: 'tasks_for in ["cron", "action"]'
+                          then: '${repository.url}'
+              repoUrl:
+                  $if: 'tasks_for in ["github-push", "github-release"]'
+                  then: '${event.repository.html_url}'
+                  else:
+                      $if: 'tasks_for == "github-pull-request"'
+                      then: '${event.pull_request.head.repo.html_url}'
+                      else:
+                          $if: 'tasks_for in ["cron", "action"]'
+                          then: '${repository.url}'
+              project:
+                  $if: 'tasks_for in ["github-push", "github-release"]'
+                  then: '${event.repository.name}'
+                  else:
+                      $if: 'tasks_for == "github-pull-request"'
+                      then: '${event.pull_request.head.repo.name}'
+                      else:
+                          $if: 'tasks_for in ["cron", "action"]'
+                          then: '${repository.project}'
+              head_branch:
                   $if: 'tasks_for == "github-pull-request"'
-                  then: '${event.pull_request.user.login}@users.noreply.github.com'
+                  then: ${event.pull_request.head.ref}
                   else:
-                      $if: 'tasks_for == "github-release"'
-                      then: '${event.sender.login}@users.noreply.github.com'
-
-      baseRepoUrl:
-          $if: 'tasks_for in ["github-push", "github-release"]'
-          then: '${event.repository.html_url}'
-          else:
-              $if: 'tasks_for == "github-pull-request"'
-              then: '${event.pull_request.base.repo.html_url}'
-              else:
-                  $if: 'tasks_for in ["cron", "action"]'
-                  then: '${repository.url}'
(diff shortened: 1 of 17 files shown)


.taskcluster.yml
```diff
@@ -5,385 +5,278 @@
...
+                  $if: 'tasks_for == "github-pull-request"'
+                  then: ${event.action}
+                  else: 'UNDEFINED'
+              releaseAction:
+                  $if: 'tasks_for == "github-release"'
+                  then: ${event.action}
+                  else: 'UNDEFINED'
+          in:
+              $if: >
+                tasks_for in ["action", "cron"]
```
MihaiTabara, line 103: Ditched the formerly `matched` commands to this, to match what we have in the other repos.

.taskcluster.yml
```diff
@@ -5,385 +5,278 @@
...
+                                      then:
+                                          kind: cron-task
+                          routes:
+                            $flattenDeep:
+                              - checks
+                              - $if: 'level == "3"'
+                                then:
+                                    # TODO Bug 1601928: Make this scope fork-friendly once ${project} is better defined. This will enable
+                                    # staging release promotion on forks.
+                                    - $if: 'tasks_for == "github-push"'
```
MihaiTabara, line 171: I really need to fix these routes at some point too, but will address as a follow-up. For now, let's keep things as they are (custom app-services indexes). We'll switch over to standard ones as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1631839

.taskcluster.yml
```diff
@@ -5,385 +5,278 @@
...
+                          payload:
+                              env:
+                                  # run-task uses these to check out the source; the inputs
+                                  # to `mach taskgraph decision` are all on the command line.
+                                  $merge:
+                                      - APPSERVICES_BASE_REPOSITORY: '${baseRepoUrl}'
+                                        APPSERVICES_HEAD_REPOSITORY: '${repoUrl}'
+                                        APPSERVICES_HEAD_REF: '${head_branch}'
+                                        APPSERVICES_HEAD_REV: '${head_sha}'
+                                        APPSERVICES_HEAD_TAG: '${head_tag}'
```
MihaiTabara, line 214: Added this new `HEAD_TAG` env var to pass it to taskraph below.

.taskcluster.yml
```diff
@@ -5,385 +5,278 @@
...
+                                    --repository-type="$APPSERVICES_REPOSITORY_TYPE"
+                                    --tasks-for='${tasks_for}'
+                              artifacts:
+                                  'public':
+                                      type: 'directory'
+                                      path: '/builds/worker/artifacts'
+                                      expires: {$fromNow: '1 year'}
+                          extra:
+                            $merge:
+                                - treeherder:
```
MihaiTabara, line 271: We don't (yet) have TH support for application-services so likely below is useless, for now. However, I'm keeping it here for the near future when hopefully we can migrate this project there too.

taskcluster/app_services_taskgraph/__init__.py
```diff
@@ -25,7 +29,14 @@ def _import_modules(modules):
 
 
 def get_decision_parameters(graph_config, parameters):
-    if parameters["tasks_for"] == "github-pull-request":
+    parameters["head_tag"] = None
```
MihaiTabara, line 32: Note to self: this might break if I remove the `try_task_config.json`.

taskcluster/ci/config.yml
```diff
@@ -25,6 +25,19 @@ workers:
...
+          default: app-services-t-signing
+    beetmover:
+      provisioner: scriptworker-k8s
+      implementation: scriptworker-beetmover
+      os: scriptworker
+      worker-type: 'app-services-{level}-beetmover'
 
 scriptworker:
-  scope-prefix: project:mozilla:application-services:releng
+  scope-prefix: project:mozilla:app-services:releng
```
MihaiTabara, line 43: Once we roll-out to releases, we can retire the old CoT product from everywhere: scriptworker, SOPS, beetmover, signing, cloudops-infra, etc.

try_task_config.json
```diff
@@ -0,0 +1,9 @@
+{
```
MihaiTabara, line 1: Note: I'll remove this before merging.

MihaiTabara:
I think this is ready for a first round of review r? @JohanLorenzo 

* I still need to add the `uploadSymbols` logic and maybe the `post-signing` job but I could add the later in a follow-up PR.
* Green staging release: https://firefox-ci-tc.services.mozilla.com/tasks/PBtc7bCTS9W8QXTqVI2jcQ

.taskcluster.yml
JohanLorenzo: 👍 We don't use `tc-treeherder.*` routes so taskcluster won't try to feed treeherder with these values. It's fine to keep them here, in my opinion.

automation/upload_android_symbols.sh
```diff
@@ -48,5 +48,6 @@ for i in "${!TARGET_ARCHS[@]}"; do
 done
 
 # 2. Upload them.
+TOKEN_FILE='.token'
```
JohanLorenzo: Nit: I'd call this file with a less generic name to avoid collision and to be more explicit about the type of expected token. For instance `.symbols_upload_token.

taskcluster/app_services_taskgraph/__init__.py
JohanLorenzo: Nit: I don't think you need to unset this and re-read it from the environment. I changed taskgraph to support this case in [bug 1589684](https://bugzilla.mozilla.org/show_bug.cgi?id=1589684). Example usage: https://github.com/mozilla-mobile/fenix/pull/6116

taskcluster/app_services_taskgraph/transforms/module_build.py
```diff
@@ -12,6 +12,25 @@
...
+def release_upload_symbols(config, tasks):
+    for task in tasks:
+        if config.params["tasks_for"] == u"github-release":
+            if task["attributes"]["buildconfig"]["uploadSymbols"]:
+                task["run"].setdefault("post-gradlew", [])
+                task["run"]["post-gradlew"].append(
+                    [
+                        "source",
+                        "automation/upload_android_symbols.sh",
+                        "{}".format(
```
JohanLorenzo, line 25: Nit: Unnecessary `format()`

taskcluster/app_services_taskgraph/transforms/module_build.py
```diff
@@ -12,6 +12,25 @@
 transforms = TransformSequence()
 
 
+@transforms.add
+def release_upload_symbols(config, tasks):
+    for task in tasks:
+        if config.params["tasks_for"] == u"github-release":
+            if task["attributes"]["buildconfig"]["uploadSymbols"]:
```
JohanLorenzo, line 19: Nit: these 2 `if` statements can be combined.

taskcluster/app_services_taskgraph/__init__.py
MihaiTabara: Ah, neat, I already had the changes from `.taskcluster.yml` but I fixed it to resemble Fenix in `__init__.py` as well. Thanks!

taskcluster/app_services_taskgraph/transforms/module_build.py
MihaiTabara: Good catch, done!

taskcluster/app_services_taskgraph/transforms/module_build.py
MihaiTabara: Good catch. Initially I had the last script call and the formatting part in one place but run-script bashing expects it as a list of arguments. Done!

taskcluster/scripts/get-secret.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
```
MihaiTabara, line 1: Copy-paste from Fenix.

taskcluster/scripts/write-dummy-secret.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
MihaiTabara, line 1: Copy-paste from Fenix.

MihaiTabara:
@eoger This is ready for review & merge whenever we can coordinate. 

Staging releases, `ci skip`, `ci full` and normal PRs works as expected.

.taskcluster.yml
```diff
@@ -5,385 +5,278 @@
...
+                  else: 'UNDEFINED'
+              releaseAction:
+                  $if: 'tasks_for == "github-release"'
+                  then: ${event.action}
+                  else: 'UNDEFINED'
+          in:
+              $if: >
+                tasks_for in ["action", "cron"]
+                || (tasks_for == "github-pull-request" && pullRequestAction in ["opened", "reopened", "synchronize"])
+                || (tasks_for == "github-push" && head_branch[:10] != "refs/tags/")
```
eoger, line 105: Can we remove push support for non-master branches? (see delete line 201)

automation/symbols-generation/upload_symbols.py
```diff
@@ -5,29 +5,32 @@
```
eoger, line 5: I assume symbols uploading is still OK?

MihaiTabara: Yes. I just moved the way we consume credentials. Tested locally and in staging releases. Unfortunately, I can't test production push until we actually merge but all the local graphs suggest it's working smooth.

MihaiTabara:
> Looks ok to me, let's do the migration tomorrow. Most important thing for me is that we can revert the changes as soon as we detect a problem since not being able to release quickly would leave us in a bad position.
> 
>     * I assume we'll keep the "legacy" TC scopes till we're 100% sure the transition was successful?
> 
>     * Are you able to rebase and squash these changes? I will be way easier to revert 1 commit if things go wrong.

1. Agreed to revert things. Worst comes to worst, we revert the PR and go back to what we had.
2. legacy TC scopes, CoT product and everything else are staying for at least 2-3 releases to confirm things are working fine with the new world. That's why I didn't remove the old logic either (from `automation/taskcluster`). For now, I'm keeping it there until we're safe and then cleanup in a few weeks or so.
3. Not sure what you mean re: squashing - I already did that - it's just one commit in this PR. Maybe you reviewed before I pushed my lattest changes 🤔

.taskcluster.yml
MihaiTabara: Done.

Issue #3145, opened 2020-05-22
[iOS] Receiving sent tab error: `JSON error invalid type: map, expected PushPayload`


garvankeeley:
x-ref https://github.com/mozilla-mobile/firefox-ios/issues/6671

Reports of this error receiving a sent tab. 
Error: https://user-images.githubusercontent.com/1866605/82678263-f67ea380-9c51-11ea-9c5b-82de6540f680.png

My concern is that I have not seen this reported before, it seems like something has changed on the backend.

A-S: v0.59.0

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1324)
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-06-05

rfk:
I don't think we've changed anything recently in the client code, but I wonder if this could be related to the recent changes in the autopush bridge server; /cc @eoger @jrconlin

rfk:
(Just noting bits of context as I think of them)

This `invalid type:` messaging is a serde error message, which IIUC means several things:
* The application successfully decrypted the push message contents and tried to hand it off to the FxA component
* The contents of the message were valid-enough JSON for serde to have a go at parsing it
* But the contents were not in the form expected by the fxa-client component

garvankeeley:
My 2 cents is that this message is a new thing, we are only seeing it recently, and client code is largely unchanged (as in there was no change I can recall that would correlate with this message now appearing). I can't discount the possibility that when I added more logging I covered a code path that wasn't covered previously, and thus the error started to show.

rfk:
I guess it could also be a new unexpected type of message being sent by the FxA server. Do we have e.g. sentry or other stats to tell us how often this error is happening?

garvankeeley:
It is on sentry https://sentry-fm.prod.mozaws.net/mozilla/firefox-ios-v250/?query=is%3Aunresolved+json, doesn't seem frequent.

rfk:
I wonder what the benefits vs risks would be of us returning (part of) the actual payload in that error message. It would probably help a great deal with debugging, but might risk sending PII to the logs, sentry, etc.

> v25.1 updated A-S to 0.58.3 from 0.58.0

Looking at the diff, that included and send-tab fixes from https://github.com/mozilla/application-services/pull/3101, but was otherwise a bunch of android and desktop stuff that shouldn't affect iOS.

[...]

PR #3177, opened 2020-05-28
Handle unknown push "command" values even if they have associated data

rfk:
Connects to #3145.

If we receive a push message with an unrecognized "command" field, we want to log it and ignore it. This currently works fine *unless the message also has a "data" field*, in which case serde complains about "invalid type: map, expected unit variant PushPayload::Unknown".

I've added a trivial testcase here to demonstrate the error.

Things is though...I can't figure out how to tell serde to do what we want here. The push payload struct is declared like this:

```
#[serde(tag = "command", content = "data")]
pub enum PushPayload {
    #[serde(rename = "fxaccounts:command_received")]
    CommandReceived(CommandReceivedPushPayload),
    #[serde(rename = "fxaccounts:device_connected")]
    DeviceConnected(DeviceConnectedPushPayload),
    [...various similar cases omitted...]
    #[serde(other)]
    Unknown,
}
```

From the error message, I *think* serde is complaining about the fact that it found a "data" field but doesn't have anywhere to put it in the `Unknown` variant. But the [docs](https://serde.rs/variant-attrs.html#other) insist that `#[serde(other)]` can only be applied to a unit variant, and I can't see a way to tell it to skip the "data" field in this case.

Any suggestions?

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -9,3 +9,4 @@
 ### What's new
 
 - Additional special case for China FxA in `getPairingAuthorityURL`. ([#3160](https://github.com/mozilla/application-services/pull/3160))
+- Silently ignore push messages for unrecognized commands, rather than reporting an error. ([#3177](https://github.com/mozilla/application-services/pull/3177))
--- a/components/fxa-client/src/push.rs
+++ b/components/fxa-client/src/push.rs
@@ -16,7 +16,17 @@ impl FirefoxAccount {
     ///
     /// **💾 This method alters the persisted account state.**
     pub fn handle_push_message(&mut self, payload: &str) -> Result<Vec<AccountEvent>> {
-        let payload = serde_json::from_str(payload)?;
+        let payload = serde_json::from_str(payload).or_else(|err| {
+            // Due to a limitation of serde (https://github.com/serde-rs/serde/issues/1714)
+            // we can't parse some payloads with an unknown "command" value. Try doing a
+            // less-strongly-validating parse so we can silently ignore such messages, while
+            // while reporting errors if the payload is completely unintelligible.
+            let v: serde_json::Value = serde_json::from_str(payload)?;
+            match v.get("command") {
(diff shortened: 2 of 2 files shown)


rfk:
(Alternately, we could decide to just silence any serde errors from this method; I worry this would cause us to miss actual bustage in the wild)

eoger:
> From the error message, I _think_ serde is complaining about the fact that it found a "data" field but doesn't have anywhere to put it in the `Unknown` variant. But the [docs](https://serde.rs/variant-attrs.html#other) insist that `#[serde(other)]` can only be applied to a unit variant, and I can't see a way to tell it to skip the "data" field in this case.

I believe we're being hit by https://github.com/serde-rs/serde/issues/1714. Until this is resolved we can log de-serialization errors and go on with our lives.

rfk:
@eoger thanks for tracking that down! I think there's some value in having `handle_push_message` give an outright error if the message is complete garbage (after all, that's how we found this issue) but I've added what I hope is a fairly simple work-around. What do you think?

CHANGES_UNRELEASED.md
```diff
@@ -9,3 +9,4 @@
 ### What's new
 
 - Additional special case for China FxA in `getPairingAuthorityURL`. ([#3160](https://github.com/mozilla/application-services/pull/3160))
+- Silently ignore push messages for unrecognized commands, rather than reporting an error. ([#3177](https://github.com/mozilla/application-services/pull/3177)
```
eoger, line 12:
```suggestion
- Silently ignore push messages for unrecognized commands, rather than reporting an error. ([#3177](https://github.com/mozilla/application-services/pull/3177))
```

eoger:
Just to be clear, these payloads will still show up as "Tab to begin" on iOS though, which isn't really an improvement from a user perspective. We should filter these messages [upstream on the FxA server](https://github.com/mozilla/fxa/blob/7bf15c4598ab0708cd7938878dec652120f829d1/packages/fxa-auth-server/lib/push.js#L132).

PR #3182, opened 2020-05-30
Put `backtrace` behind a Cargo feature

linacambridge:
`backtrace` brings lots of dependencies with it, some of which
(like `gimli`) we can't vendor in Firefox Desktop because they're too
big. We could work around that, but we never log backtraces on Desktop,
anyway, and pulling in the library adds bloat there. So let's put it behind
a Cargo feature flag, with a compatibility shim.

### DIFF
--- a/components/support/error/Cargo.toml
+++ b/components/support/error/Cargo.toml
@@ -7,4 +7,7 @@ license = "MPL-2.0"
 
 [dependencies]
 thiserror = "1.0"
-backtrace = "0.3"
+
+[dependencies.backtrace]
+optional = true
+version = "0.3"
--- a/components/support/error/src/lib.rs
+++ b/components/support/error/src/lib.rs
@@ -2,9 +2,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[cfg(feature = "backtrace")]
 /// Re-export of the `backtrace` crate for use in macros and
 /// to ensure the needed version is kept in sync in dependents.
 pub use backtrace;
+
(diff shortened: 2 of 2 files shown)


linacambridge:
@thomcc, if this looks alright to you, what's the right Cargo incantation to enable this by default, but disable in Gecko?

thomcc:
> @thomcc, if this looks alright to you, what's the right Cargo incantation to enable this by default, but disable in Gecko?

It's kind of a pain.

In error-support

```toml
[features]
default = ["backtrace"]
```

Then, in *every* crate that depends on error-support (if you miss even one it will pull the dep in).

```toml
[features]
# If this is a problem because it shares a name with a crate in the
# dependency graph, maybe call it `use_backtrace`
backtrace = ["error-support/backtrace"]
default = ["backtrace"]
```

Then, in gecko, the dep has to be added with `--no-default-features`. 

It's... not trivial. We could also revisit the conversation here. Backtraces in errors make test failures easier to track down, but to a large extent that could be mitigated by moving from using `fn test() -> Result<...>` to using unwraps in tests. There are other benefits too, but they're smaller.

thomcc:
It's worth noting that due to a quirk (bug/feature) in how cargo feature resolution works, I believe just *as is, without any of that work* we can get backtraces to turn on if  `cargo test --all-features` is used from the root of the workspace.

thomcc:
Ah, by --no-default-features i mean `default-features=false` in the toml.

linacambridge:
> I believe just _as is, without any of that work_ we can get backtraces to turn on if `cargo test --all-features` is used from the root of the workspace.

Oh, that would be cool! https://github.com/mozilla/application-services/pull/3151#issuecomment-634073569 suggests `in_gecko` might be a problem, though?

Issue #3122, opened 2020-05-14
authorize_code_using_session_token missing fields


eoger:
I've noticed this method, despite having bindings on Android, is not being used at all (I believe it was added for FPN).
We will definitely use it for [pairing on desktop](https://searchfox.org/mozilla-central/rev/9f074fab9bf905fad62e7cc32faf121195f4ba46/services/fxaccounts/FxAccountsPairing.jsm#315), however the HTTP request it makes is missing a few fields, namely [code_challenge, code_challenge_method and keys_jwe](https://searchfox.org/mozilla-central/rev/9f074fab9bf905fad62e7cc32faf121195f4ba46/services/fxaccounts/FxAccountsClient.jsm#274-285).

One difficulty here is the `keys_jwe` parameter: we'd like the method to accept a `keys_jwk` and handle the encryption to a `jwe` itself. We have some [prior art on mozilla-central](https://searchfox.org/mozilla-central/rev/9f074fab9bf905fad62e7cc32faf121195f4ba46/services/fxaccounts/FxAccounts.jsm#1669) but in any case it's not trivial. I'm not sure if it helps but we have the inverse operation [implemented in Rust](https://github.com/mozilla/application-services/blob/76ee86e38cbe15727a4333755b58e34722ca2aa0/components/fxa-client/src/scoped_keys.rs#L92).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1288)
┆Story Points: 5
┆Epic: Refactor Desktop FxA APIs to use Fxa-client & AccountMananger
┆Sprint: SYNC - end 2020-07-17

rfk:
Cross-linking https://github.com/mozilla/application-services/issues/3086, which will also need the ability to encrypt things into a JWE (for different reasons, but the same basic operation).

PR #3187, opened 2020-06-01
[WIP] Remove node helper + adds jwe encryption

tarikeshaq:

**WORK IN PROGRESS**

- Currently, the tests pass after removing node js helper
- Getting 500s for account destroy 🙃 (After the tests successfully pass)
- Implemented encrypt_keys_jwe (mirroring how it's done in gecko and using the decryption function as reference), hoping to also use it for #3122 but I'd need a key_fetch_token (or even better the actual keys) for that issue (Tips are welcome!!)
- Implement the retrieval of the keys and how they are formated from the code in gecko
- Still have a bunch of todos in the code I'll need to figure out
- Mostly would love comments on where to put what, especially with the stuff I have now in /components/fxa-client/src/auth.rs
- Still have a bit of redundancy to clean up, but thought I'd have this up for people to look at!


vladikoff:
* Getting 500s for account destroy 🙃 (After the tests successfully pass)

Known issue https://github.com/mozilla/fxa/issues/3629

* Implemented encrypt_keys_jwe (mirroring how it's done in gecko and using the decryption function as reference), hoping to also use it for #3122 but I'd need a key_fetch_token (or even better the actual keys) for that issue (Tips are welcome!!)

Let's figure that in a separate PR

components/fxa-client/Cargo.toml
```diff
@@ -24,6 +24,10 @@ viaduct = { path = "../viaduct" }
 rc_crypto = { path = "../support/rc_crypto", features = ["ece", "hawk"] }
 error-support = { path = "../support/error" }
 thiserror = "1.0"
+anyhow = "1.0"
+pbkdf2 = "0.3.0"
+sha2 = "0.8.1" # pbkdf2 dep
+hmac = "0.7.1" # pbkdf2 dep
```
eoger, line 30: How you would feel about implementing these operations in `rc_crypto`? (ofc I can mentor you here) I'm concerned that we'll be adding untrusted crypto code in mozilla-central.

components/fxa-client/src/auth.rs
```diff
@@ -0,0 +1,315 @@
...
+use serde_json::json;
+use std::collections::HashMap;
+use viaduct::{Method, Request};
+
+impl FirefoxAccount {
+    /// Log-in without any user interaction.
+    /// This method is only meant to be used by tests.
+    pub fn login(_email: &str, _password: &str) -> Result<()> {
+        Ok(())
+    }
```
eoger, line 20: Huh? :)

components/fxa-client/src/scoped_keys.rs
```diff
@@ -89,6 +103,94 @@ impl ScopedKeysFlow {
...
+            .unwrap();
+        let x = base64::decode_config(x, base64::URL_SAFE_NO_PAD)?;
+        let y = base64::decode_config(y, base64::URL_SAFE_NO_PAD)?;
+        let mut peer_public_key: Vec<u8> = vec![0x04];
+        peer_public_key.extend_from_slice(&x);
+        peer_public_key.extend_from_slice(&y);
+        Ok(peer_public_key)
+    }
+
+    pub fn encrypt_keys_jwe(self, jwk: &str, data: &[u8]) -> Result<String> {
```
vladikoff, line 158: @rfk is it ok to keep this here, even for https://github.com/mozilla/application-services/issues/3122#issuecomment-628965909 ?

components/fxa-client/Cargo.toml
vladikoff: issue here https://github.com/mozilla/application-services/issues/3190

tarikeshaq: I would love to 😃 I'll get started on it on a separate pr and will let you know if I have any questions as I go!

testing/sync-test/src/auth.rs
```diff
@@ -180,29 +167,51 @@ pub struct TestClient {
...
+        )?;
+        let auth_params = auth::AuthorizationParameters::new(
+            client_id.clone(),
+            code_challenge.clone(),
+            code_challenge_method.clone(),
+            scope.clone(),
+            keys_jwe,
+            state.clone(),
+        );
+        let (code, state) = auth::send_authorization_request(&acct.cfg, auth_params, &auth_key)?;
```
rfk, line 211: From a code-organization perspective, this key-fetching PKCE-code-authorizing stuff feels like it should mostly belong in the `TestAccount`, not the `TestClient`. Suggestion: make a `TestAccount::execute_oauth_flow` method that takes the URL returned by `fxa.begin_oauth_flow` here, does all the necessary authorization stuff, and returns the URI to which you would have been directed if this were a real OAuth flow. Encapsulating "all the bits that a user would normally do" inside the `TestAccount` seems like it will help us keep things straight in ours heads when we come to revisit this code in the future.

testing/sync-test/src/auth.rs
```diff
@@ -213,9 +222,23 @@ impl TestClient {
         })
     }
 
-    pub fn data_for_sync(
-        &mut self,
-    ) -> Result<(Sync15StorageClientInit, KeyBundle, String), failure::Error> {
+    fn verify_account(email: &str, config: &FxaConfig, auth_key: &[u8]) -> Result<()> {
```
rfk, line 225: As above, I think this would be better framed as a method on `TestAccount`.

testing/sync-test/src/auth.rs
```diff
@@ -109,11 +57,38 @@ impl TestAccount {
...
-    ) -> Result<Arc<TestAccount>, failure::Error> {
+    ) -> Result<Arc<TestAccount>> {
         log::info!("Creating temporary fx account");
-        // `create` doesn't return anything we care about.
-        let auth_url = cfg.auth_url()?;
-        run_helper_command("create", &[&email, &pass, auth_url.as_str()])?;
+
+        restmail::clear_mailbox(&email);
+
+        let create_endpoint = cfg.auth_url_path("v1/account/create").unwrap();
```
rfk, line 65: You may be able to simplify things here, by fetching the sync keys during account creation. If you specify `?keys=true` in the URL here, then it will return a `sessionToken` and `keyFetchToken` that you can use in the same way as the ones you get in `account_login`. You could verify the account like you do here already, then fetch the keys and store them as a property on the `TestAccount`.

Later, when you can to execute the oauth flow, you wouldn't need to log in to the account again, you could just use the keys and sessionToken obtained during account creation.

components/fxa-client/src/lib.rs
```diff
@@ -36,8 +37,9 @@ pub mod migrator;
 pub mod msg_types {
     include!("mozilla.appservices.fxaclient.protobuf.rs");
 }
-mod http_client;
-mod oauth;
+// TODO: figure out a way without making too much public....
+pub mod http_client;
```
rfk, line 41: FWIW, I think it would be fine to make the `http_client` module public, and call its methods directly from the code for `TestAccount`. The risk of mis-use seems pretty low.

For methods that should really only be used from tests, I'm a fan of giving them super explicit names, so you can imagine the test code calling something like `client.test_only_account_create(email, password)` or similar.

rfk:
> Consider committing to making the http_client class public for use in tests, and having the
> TestAccount class call named methods on it for all the places where it needs to interact with

Another alternative could be to make a new `fxa-client::TestHelpers` class that exposes both an http_client instance, and some additional methods for doing the key-derivation stuff. This would make it clear that these are only supposed to be used for tests, and we could even figure out how to avoid shipping that code at all when building for android, ios, etc.

By the way, I can't help calling them "classes" even though I know that's not what rust calls them, but I hope it's clear enough that I mean "struct" :-)

tarikeshaq:
Thanks a ton for the review and feedback @rfk !! I'll get on the comments and I'll separate the JWE stuff out

PR #3193, opened 2020-06-02
Implements pbkdf2 for rc_crypto

### LINKED ISSUE #3190: Add pbkdf2 to rc_crypto
Extracted from [https://github.com/mozilla/application-services/pull/3187/files#r433409179](https://github.com/mozilla/application-services/pull/3187/files#r433409179)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1362)
┆Story Points: 5

tarikeshaq:

fixes #3190 
Adds [pbkdf2](https://www.ietf.org/rfc/rfc2898.txt) to rc_crypto. To be used by #3187 and any future uses for extending passwords.

### DIFF
--- a/components/support/rc_crypto/nss/nss_sys/src/bindings/pk11pub.rs
+++ b/components/support/rc_crypto/nss/nss_sys/src/bindings/pk11pub.rs
@@ -110,4 +110,20 @@ extern "C" {
         attr: CK_ATTRIBUTE_TYPE,
         item: *mut SECItem,
     ) -> SECStatus;
+    pub fn PK11_CreatePBEV2AlgorithmID(
+        pbeAlgTag: u32,    /* SECOidTag */
+        cipherAlgTag: u32, /* SECOidTag */
+        prfAlgTag: u32,    /* SECOidTag */
+        keyLength: c_int,
+        iteration: c_int,
+        salt: *mut SECItem,
+    ) -> *mut SECAlgorithmID;
+
+    pub fn PK11_PBEKeyGen(
+        slot: *mut PK11SlotInfo,
+        algid: *mut SECAlgorithmID,
+        pwitem: *mut SECItem,
+        faulty3DES: PRBool,
+        wincx: *mut c_void,
+    ) -> *mut PK11SymKey;
 }
--- a/components/support/rc_crypto/nss/src/pbkdf2.rs
+++ b/components/support/rc_crypto/nss/src/pbkdf2.rs
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::util::{ensure_nss_initialized, map_nss_secstatus, sec_item_as_slice, ScopedPtr};
+use crate::{
+    error::*,
+    pk11::{
+        slot::get_internal_slot,
+        types::{AlgorithmID, SymKey},
+    },
+};
+
+// Expose for consumers to choose the hashing algorithm
+// Currently only SHA256 supported
+pub use crate::pk11::context::HashAlgorithm;
+use nss_sys::SECOidTag;
+use std::convert::TryFrom;
+
+// ***** BASED ON THE FOLLOWING IMPLEMENTATION *****
+// https://searchfox.org/mozilla-central/rev/8ccea36c4fb09412609fb738c722830d7098602b/dom/crypto/WebCryptoTask.cpp#2567
+
+pub fn pbkdf2_key_derive(
+    password: &[u8],
+    salt: &[u8],
+    iterations: u32,
+    hash_algorithm: HashAlgorithm,
+    out: &mut [u8],
+) -> Result<()> {
+    ensure_nss_initialized();
+    let oid_tag = match hash_algorithm {
+        HashAlgorithm::SHA256 => SECOidTag::SEC_OID_HMAC_SHA256 as u32,
+    };
+    let mut sec_salt = nss_sys::SECItem {
+        len: u32::try_from(salt.len())?,
+        data: salt.as_ptr() as *mut u8,
+        type_: 0,
+    };
+    let alg_id = unsafe {
+        AlgorithmID::from_ptr(nss_sys::PK11_CreatePBEV2AlgorithmID(
+            SECOidTag::SEC_OID_PKCS5_PBKDF2 as u32,
+            SECOidTag::SEC_OID_HMAC_SHA1 as u32,
+            oid_tag,
+            i32::try_from(out.len())?,
+            i32::try_from(iterations)?,
+            &mut sec_salt as *mut nss_sys::SECItem,
+        ))?
+    };
+
+    let slot = get_internal_slot()?;
+    let mut sec_pw = nss_sys::SECItem {
+        len: u32::try_from(password.len())?,
+        data: password.as_ptr() as *mut u8,
+        type_: 0,
+    };
+    let sym_key = unsafe {
+        SymKey::from_ptr(nss_sys::PK11_PBEKeyGen(
+            slot.as_mut_ptr(),
+            alg_id.as_mut_ptr(),
+            &mut sec_pw as *mut nss_sys::SECItem,
+            nss_sys::PR_FALSE,
+            std::ptr::null_mut(),
+        ))?
+    };
+    map_nss_secstatus(|| unsafe { nss_sys::PK11_ExtractKeyValue(sym_key.as_mut_ptr()) })?;
+
+    // This doesn't leak, because the SECItem* returned by PK11_GetKeyData
+    // just refers to a buffer managed by `sym_key` which we copy into `buf`
+    let mut key_data = unsafe { *nss_sys::PK11_GetKeyData(sym_key.as_mut_ptr()) };
+    let buf = unsafe { sec_item_as_slice(&mut key_data)? };
+    // Stop panic in swap_with_slice by returning an error if the sizes mismatch
+    if buf.len() != out.len() {
+        return Err(ErrorKind::InternalError.into());
+    }
+    out.swap_with_slice(buf);
+    Ok(())
+}
--- a/components/support/rc_crypto/src/pbkdf2.rs
+++ b/components/support/rc_crypto/src/pbkdf2.rs
@@ -1,0 +1,196 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::*;
+use nss::pbkdf2::pbkdf2_key_derive;
+pub use nss::pbkdf2::HashAlgorithm;
+/// Extend passwords using pbkdf2, based on the following [rfc](https://www.ietf.org/rfc/rfc2898.txt) it runs the NSS implementation
+/// # Arguments
+///
+///  * `passphrase` - The password to stretch
+///  * `salt` - A salt to use in the generation process
+///  * `iterations` - The number of iterations the hashing algorithm will run on each section of the key
+///  * `hash_algorithm` - The hash algorithm to use
+///  * `out` - The slice the algorithm will populate
+///
+/// # Examples
+///
+/// ```
+/// use rc_crypto::pbkdf2;
+/// let password = b"password";
+/// let salt = b"salt";
+/// let mut out = vec![0u8; 32];
+/// let iterations = 2; // Real code should have a MUCH higher number of iterations (Think 1000+)
+/// pbkdf2::derive(password, salt, iterations, pbkdf2::HashAlgorithm::SHA256, &mut out).unwrap(); // Oh oh should handle the error!
+/// assert_eq!(hex::encode(out), "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43");
+//
+///```
+///
+/// # Errors
+///
+/// Could possibly return an error if the HMAC algorithm fails, or if the NSS algorithm returns an error
+pub fn derive(
+    passphrase: &[u8],
+    salt: &[u8],
+    iterations: u32,
+    hash_algorithm: HashAlgorithm,
+    out: &mut [u8],
+) -> Result<()> {
+    pbkdf2_key_derive(passphrase, salt, iterations, hash_algorithm, out)?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn test_generate_correct_out() {
+        let expected = "120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b";
+        let mut out = vec![0u8; 32];
+        let password = b"password";
+        let salt = b"salt";
+        derive(password, salt, 1, HashAlgorithm::SHA256, &mut out).unwrap();
+        assert_eq!(expected, hex::encode(out));
+    }
+
+    #[test]
+    fn test_longer_key() {
+        let expected = "120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b4dbf3a2f3dad3377264bb7b8e8330d4efc7451418617dabef683735361cdc18c";
+        let password = b"password";
+        let salt = b"salt";
+        let mut out = vec![0u8; 64];
+        derive(password, salt, 1, HashAlgorithm::SHA256, &mut out).unwrap();
+        assert_eq!(expected, hex::encode(out));
+    }
+
+    #[test]
+    fn test_more_iterations() {
+        let expected = "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43";
+        let password = b"password";
+        let salt = b"salt";
+        let mut out = vec![0u8; 32];
+        derive(password, salt, 2, HashAlgorithm::SHA256, &mut out).unwrap();
+        assert_eq!(expected, hex::encode(out));
+    }
+
+    #[test]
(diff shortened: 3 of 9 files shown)


components/support/rc_crypto/src/pbkdf2.rs
```diff
@@ -0,0 +1,116 @@
...
+///```
+///
+/// # Erros
+///
+/// Could possibly return an error if the HMAC algorithm fails
+pub fn pbkdf2_generate(
+    passphrase: &[u8],
+    salt: &[u8],
+    iterations: u32,
+    key_len: u32,
```
rfk, line 33: API nit: I think it would be more idiomatic to let the caller provide an output buffer `out: &mut [u8]` and have this function will it, rather than allocate a new buffer. See e.g. the API exposed by `hkdf.rs` in this same directory.

components/support/rc_crypto/src/pbkdf2.rs
```diff
@@ -0,0 +1,116 @@
...
+            *curr_byte ^= u_curr.as_ref()[k];
+        }
+        u_prev = u_curr;
+    }
+    Ok(res)
+}
+
+fn to_32b_buf(n: u32) -> Vec<u8> {
+    let mut buf = [0; 4];
+    BigEndian::write_u32(&mut buf, n);
```
rfk, line 70: Using [`u32::to_be_bytes`](https://doc.rust-lang.org/std/primitive.u32.html#method.to_be_bytes) might help avoid an external dependency here.

rfk:
I should add: the algorithm is simple enough (and the risks of doing it wrong are low enough) that I have no objection to us shipping our own implementation, if NSS doesn't natively provide what we want or if it's a PITA to use.

tarikeshaq:
Thanks again for the comments 😄 I'll look into NSS and see what I can figure out there!

tarikeshaq:
So since I'm calling into NSS, quite a bit of unsafe code exists and since this my first time writing unsafe code (although it is just calling into NSS that's unsafe) I would love any pointers on what could change / any bad unsafe practices 😄  etc

Any advice on the best way to clean up the pointers would also be great! I'm considering just having a helper function that does it, but I was wondering if something like the stuff in nss/pk11/sym_key would be better (In fact I'm considering moving everything there similar to hkdf_expand, would that be the best approach)

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+// https://searchfox.org/mozilla-central/rev/8ccea36c4fb09412609fb738c722830d7098602b/dom/crypto/WebCryptoTask.cpp#2567
+
+pub fn nss_pbkdf2_generate(
+    password: &mut [u8],
+    salt: &mut [u8],
+    iterations: u32,
+    out: &mut [u8],
+) -> Result<()> {
+    ensure_nss_initialized();
+    // Todo: Maybe add support for different hashing algorithms?
```
eoger, line 21: Agree, how would you feel about adding an `Algorithm` param for this function?

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+        data: salt.as_mut_ptr(),
+        type_: 0,
+    };
+
+    let alg_id = unsafe {
+        nss_sys::PK11_CreatePBEV2AlgorithmID(
+            SECOidTag::SEC_OID_PKCS5_PBKDF2 as u32,
+            SECOidTag::SEC_OID_HMAC_SHA1 as u32,
+            oid_tag as u32,
+            // Should add error handling to prevent panic if (in the exteremely unlikely scenario) the cast can't be done
```
eoger, line 34: Can't we use `try_from` here? If not we should do manual bounds checking.

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Todo: Maybe add support for different hashing algorithms?
+    let oid_tag = SECOidTag::SEC_OID_HMAC_SHA256;
+    let mut sec_salt = nss_sys::SECItem {
+        len: salt.len() as u32,
+        data: salt.as_mut_ptr(),
+        type_: 0,
+    };
+
+    let alg_id = unsafe {
+        nss_sys::PK11_CreatePBEV2AlgorithmID(
```
eoger, line 30: You should wrap this call in a [scoped pointer wrapper](https://searchfox.org/mozilla-central/rev/7cadba1d8b8feaec4615f5bb98aac4b8a719793c/security/manager/ssl/ScopedNSSTypes.h#304-305). We do this using the `scoped_ptr!` macro, there's some examples in components/support/rc_crypto/nss/src/pk11/types.rs.

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+            iterations as i32,
+            &mut sec_salt as *mut nss_sys::SECItem,
+        )
+    };
+
+    if alg_id.is_null() {
+        return Err(ErrorKind::InternalError.into());
+    }
+
+    let slot = unsafe { nss_sys::PK11_GetInternalSlot() };
```
eoger, line 45: We have a safe wrapper for that `get_internal_slot`, can we use it?

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+        unsafe { nss_sys::SECOID_DestroyAlgorithmID(alg_id, nss_sys::PR_TRUE) };
+        return Err(ErrorKind::InternalError.into());
+    }
+    let mut pw_item = nss_sys::SECItem {
+        len: password.len() as u32,
+        data: password.as_mut_ptr(),
+        type_: 0,
+    };
+
+    let sym_key = unsafe {
```
eoger, line 56: I think you can use the `SymKey` wrapper struct here.

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+                nss_sys::PK11_FreeSlot(slot);
+            };
+        }
+        if !sym_key.is_null() {
+            unsafe {
+                nss_sys::PK11_FreeSymKey(sym_key);
+            };
+        }
+        return Err(ErrorKind::InternalError.into());
+    }
```
eoger, line 84: 😅

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let data = unsafe {
+        let inner = (*key_data).data;
+        std::slice::from_raw_parts_mut(inner, out.len())
+    };
```
eoger, line 109: Do you want `sec_item_as_slice`?

components/support/rc_crypto/nss/nss_sys/src/bindings/pk11pub.rs
```diff
@@ -110,4 +110,20 @@ extern "C" {
...
+        salt: *mut SECItem,
+    ) -> *mut SECAlgorithmID;
+
+    pub fn PK11_PBEKeyGen(
+        slot: *mut PK11SlotInfo,
+        algid: *mut SECAlgorithmID,
+        pwitem: *mut SECItem,
+        faulty3DES: PRBool,
+        wincx: *mut c_void,
+    ) -> *mut PK11SymKey;
```
eoger, line 128: Did you generate these with bindgen or hand-written? (in any case I assume ctest will catch a problem)

eoger: I manually checked these bindings against the nss header files and these look correct 👍

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,131 @@
...
+use crate::error::*;
+use crate::util::ensure_nss_initialized;
+use nss_sys::SECOidTag;
+
+// ***** BASED ON THE FOLLOWING IMPLEMENTATION *****
+// https://github.com/sqlcipher/sqlcipher/blob/4a81bea61e1da6fec222d713852830f1fd01aed2/src/crypto_nss.c#L182-L232
+// Which I believe is based on how it is done on WebCrypto:
+// https://searchfox.org/mozilla-central/rev/8ccea36c4fb09412609fb738c722830d7098602b/dom/crypto/WebCryptoTask.cpp#2567
+
+pub fn nss_pbkdf2_generate(
```
eoger, line 14: Nit: Can we call this `pbkdf2_key_derive`, `pbkdf2_keygen` or something else?

components/support/rc_crypto/nss/src/pbkdf2.rs
tarikeshaq: Done! (Used HashAlgorithm from context, not sure if that's the right choice, but seemed handy)

components/support/rc_crypto/nss/src/pbkdf2.rs
tarikeshaq: I had to add to the scoped_ptr! macro a second expression where the destructor has an argument (since the algorithm ID destructor requires an extra argument) wasn't sure if that is best, or if I should define a separate struct for AlgorithmID without using the macro.

eoger: `pk11_destroy_context_true` is an example where we send a param to the destructor

components/support/rc_crypto/nss/src/pk11/sym_key.rs
```diff
@@ -91,3 +95,23 @@ pub(crate) fn import_sym_key(
         ))
     }
 }
+
+/// Safe wrapper around PK11_PBEKeyGen that
+/// de-allocates memory when the key goes out of
+/// scope.
+pub(crate) fn pbe_key_gen(
```
eoger, line 102: I think that's fine to inline into the caller, since there's only one use of this.

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,71 @@
...
+    salt: &mut [u8],
+    iterations: u32,
+    hash_algorithm: HashAlgorithm,
+    out: &mut [u8],
+) -> Result<()> {
+    ensure_nss_initialized();
+    let oid_tag = match hash_algorithm {
+        HashAlgorithm::SHA256 => SECOidTag::SEC_OID_HMAC_SHA256 as u32,
+    };
+
```
eoger, line 33: Very very nit: there's a ton of line breaks in this function which makes it bit harder to reason about. Can we group instructions somehow? 1. create the alg, 2. stretch the key 3. get the vector (or any other grouping that you like really :))

components/support/rc_crypto/nss/src/pbkdf2.rs
```diff
@@ -0,0 +1,71 @@
...
+// Expose for consumers to choose the hasing algorithm
+// Currently only SHA256 supported
+pub use crate::pk11::context::HashAlgorithm;
+use nss_sys::SECOidTag;
+use std::convert::TryFrom;
+
+// ***** BASED ON THE FOLLOWING IMPLEMENTATION *****
+// https://github.com/sqlcipher/sqlcipher/blob/4a81bea61e1da6fec222d713852830f1fd01aed2/src/crypto_nss.c#L182-L232
+// Which I believe is based on how it is done on WebCrypto:
```
eoger, line 19: Nit: We can probably remove these and only keep the searchfox link

components/support/rc_crypto/nss/src/util.rs
```diff
@@ -120,6 +120,43 @@ macro_rules! scoped_ptr {
             }
         }
     };
+    // For the case the destructor requires an addtional argument
```
eoger, line 123: As mentioned previously I don't think you need this.

components/support/rc_crypto/src/pbkdf2.rs
```diff
@@ -0,0 +1,75 @@
...
+/// let mut out = vec![0u8; 32];
+/// pbkdf2::pbkdf2_generate(&mut password, &mut salt, 2, pbkdf2::HashAlgorithm::SHA256, &mut out).unwrap(); // Oh oh should handle the error!
+/// assert_eq!(hex::encode(out), "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43");
+//
+///```
+///
+/// # Erros
+///
+/// Could possibly return an error if the HMAC algorithm fails
+pub fn pbkdf2_generate(
```
eoger, line 32:
```suggestion
pub fn derive(
```

(we already are in the `pbkdf2` module)

components/support/rc_crypto/src/pbkdf2.rs
```diff
@@ -0,0 +1,75 @@
...
+/// use rc_crypto::pbkdf2;
+/// let mut password = b"password".to_vec();
+/// let mut salt = b"salt".to_vec();
+/// let mut out = vec![0u8; 32];
+/// pbkdf2::pbkdf2_generate(&mut password, &mut salt, 2, pbkdf2::HashAlgorithm::SHA256, &mut out).unwrap(); // Oh oh should handle the error!
+/// assert_eq!(hex::encode(out), "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43");
+//
+///```
+///
+/// # Erros
```
eoger, line 29:
```suggestion
/// # Errors
```

components/support/rc_crypto/nss/src/pbkdf2.rs
tarikeshaq: I can't believe I didn't see that 🤦 Thanks @eoger and I really appreciate your reviews 🙏

components/support/rc_crypto/src/pbkdf2.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[test]
+    fn test_more_iterations() {
+        let expected = "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43";
+        let mut password = b"password".to_vec();
+        let mut salt = b"salt".to_vec();
+        let mut out = vec![0u8; 32];
+        derive(&mut password, &mut salt, 2, HashAlgorithm::SHA256, &mut out).unwrap();
+        assert_eq!(expected, hex::encode(out));
+    }
+}
```
rfk, line 75: This seems to be working as expected, but I'd like to see more coverage of edge-cases in these tests, particularly around the things that might be going wrong when we transit from Rust into C. For example: do we properly handle embedded nulls in the password, and the salt? can we generate keys that aren't a full multiple of the hash blocksize? do we cleanly reject if requesting more than `(2^32 - 1) * hLen` bytes of output, assuming that NSS correctly fails in this case?

I think that `hkdf.rs` has a good sample of some of the things worth worrying about, but I also took the liberty of generating what I think would be some worthwhile extra test vectors, using the python `cryptography` library:

```
PBKDF2-SHA256(b'password', b'salt', 1, 32) = b'120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b'
PBKDF2-SHA256(b'password', b'salt', 1, 64) = b'120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b4dbf3a2f3dad3377264bb7b8e8330d4efc7451418617dabef683735361cdc18c'
PBKDF2-SHA256(b'password', b'salt', 2, 32) = b'ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43'
PBKDF2-SHA256(b'password', b'salt', 3, 33) = b'ad35240ac683febfaf3cd49d845473fbbbaa2437f5f82d5a415ae00ac76c6bfccf'
PBKDF2-SHA256(b'passw\x00rd', b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 5, 13) = b'e25d526987819f966e324faa4a'
PBKDF2-SHA256(b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', b'salt', 2, 32) = b'62384466264daadc4144018c6bd864648272b34da8980d31521ffcce92ae003b'
PBKDF2-SHA256(b'password', b'salt', 1, 0) = b''
PBKDF2-SHA256(b'', b'salt', 1, 32) = b'f135c27993baf98773c5cdb40a5706ce6a345cde61b000a67858650cd6a324d7'
PBKDF2-SHA256(b'password', b'', 1, 32) = b'c1232f10f62715fda06ae7c0a2037ca19b33cf103b727ba56d870c11f290a2ab'
PBKDF2-SHA256(b'password', b'salt', 0, 32) = b'120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b'                                                                                                     
```

tarikeshaq: Thank you for those!! I added the tests, but one thing I did notice, is that the NSS algorithm rejects 0 iterations and 0 length keyLength (out.len() ). I confirmed that via Crypto.subtle. But otherwise added the tests you mentioned and a few others to tests correct rejections.

rfk: > NSS algorithm rejects 0 iterations and 0 length keyLength

Great :-)

TBH I was kind of surprised that the python implementation allowed these, but it's worth having tests to document the specific behaviour in these cases either way.

tarikeshaq:
Thanks a lot, @rfk and @eoger !! I appreciate the reviews and feedback! 🕺

PR #3216, opened 2020-06-09
Fix bug in URL manipulation functions

thomcc:
This causes garbage to populate the origins table. We should have a solution here for how to clean it up.

One option is identify problem rows in places `o.host` not matching `get_host_and_port(p.url)` a given row in places might be enough.

Then maybe:

1. Fill in any origins we are missing.

```sql
INSERT OR IGNORE INTO moz_origins(prefix, host, rev_host)
SELECT
	get_prefix(p.url),
    get_host_and_port(p.url),
    reverse_host(get_host_and_port(p.url))
FROM moz_places p
WHERE <it was bad>;
```

2. Ensure all places rows point at the right origin
```sql
UPDATE moz_places SET
  origin_id = (SELECT id FROM moz_origins
               WHERE prefix = get_prefix(p.url) AND
                     host = get_host_and_port(p.url) AND
                     rev_host = reverse_host(get_host_and_port(p.url)))
;
```
3. Delete any leftover garbage.
```sql
DELETE FROM moz_origins o
WHERE NOT EXISTS (
    SELECT 1 FROM moz_places p WHERE p.origin_id = o.id
);
```

4. (not pictured) flag impacted rows for frecency recalc... 

I tried this locally but ran it on a DB with no triggers, so I'm not really sure how it owuld perform. Also IDK if any origins rows exist elsewhere.

I also think the way we manage moz_origin frecency seems pretty janky, and a bad fit for our profile given we usually run on mobile

CC @grigoryk, this probably makes the origin index autocomplete slowly behave worse and worse as the user visits google maps more and more.

### DIFF
--- a/components/places/src/api/matcher.rs
+++ b/components/places/src/api/matcher.rs
@@ -140,7 +140,9 @@ pub fn accept_result(conn: &PlacesDb, search_string: &str, url: &Url) -> Result<
 }
 
 pub fn split_after_prefix(href: &str) -> (&str, &str) {
-    match memchr::memchr(b':', href.as_bytes()) {
+    // Only search up to 64 bytes (matches desktop behavior)
+    let haystack = &href.as_bytes()[..href.len().min(64)];
+    match memchr::memchr(b':', haystack) {
         None => ("", href),
         Some(index) => {
             let hb = href.as_bytes();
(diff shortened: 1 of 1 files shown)


components/places/src/api/matcher.rs
```diff
@@ -140,28 +140,34 @@ pub fn accept_result(conn: &PlacesDb, search_string: &str, url: &Url) -> Result<
...
             if hb.len() >= end + 2 && hb[end] == b'/' && hb[end + 1] == b'/' {
                 end += 2;
             }
-            (&href[0..end], &href[end..])
+            href.split_at(end)
         }
     }
 }
 
 pub fn split_after_host_and_port(href: &str) -> (&str, &str) {
```
thomcc, line 158: For clarity, split_after_host_and_port is the only one here that's important, although I noticed `split_after_prefix`'s behvior differed from desktop.

PR #3217, opened 2020-06-10
Updated instructions for working with WSL and newer tools

mhammond:
Flagging Ryan for the WSL changes and Ed to see if there's anything obvious that's not going to work elsewhere...

### DIFF
--- a/docs/howtos/setup-android-build-environment.md
+++ b/docs/howtos/setup-android-build-environment.md
@@ -19,22 +19,7 @@ This process should work OK on Mac and Linux. It also works on [Windows via WSL 
 Typically, this process only needs to be run once, although periodically you
 may need to repeat some steps (eg, rust updates should be done periodically)
 
-At the end of this process you should have the following environment variables set up.
-
-- `ANDROID_NDK_ROOT`
-- `ANDROID_NDK_HOME`
-- `ANDROID_NDK_API_VERSION`
-- `ANDROID_HOME`
-- `JAVA_HOME`
-
-These variables are required every time you build, so you should add them to
-a rc file or similar so they persist between reboots etc.
-
-1. Install NDK r21 from https://developer.android.com/ndk/downloads
-    - Extract it, put it somewhere (`$HOME/.android-ndk-r21` is a reasonable
-      choice, but it doesn't matter), and set `ANDROID_NDK_ROOT` to this location.
-    - Set `ANDROID_NDK_HOME` to match `ANDROID_NDK_ROOT`, for compatibility with
-      some android grandle plugins.
+1. Prepare the [Android SDK and NDK](#setting-up-the-android-sdk-and-ndk)
 
 2. Install `rustup` from https://rustup.rs:
     - If you already have it, run `rustup update`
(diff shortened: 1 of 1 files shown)


docs/howtos/setup-android-build-environment.md
```diff
@@ -108,10 +85,106 @@ you will find:
...
+- `ANDROID_NDK_API_VERSION`
+- `ANDROID_HOME`
+- `JAVA_HOME`
+
+These variables are required every time you build, so you should add them to
+a rc file or similar so they persist between reboots etc.
+
+1. Install the SDK, largely as described above.
+
+2. Install NDK r21 from https://developer.android.com/ndk/downloads
```
rfk, line 158: I'm surprised that the instructions above seemed to install r20, but these outdated instructions call for r21. I guess it's related to the build error you mention below, but OTOH, I'm definitely using r21 and it seems to be working OK.

docs/howtos/setup-android-build-environment.md
```diff
@@ -128,43 +201,35 @@ You will need the following tools in WSL:
 
 * unzip - `sudo apt install unzip`
 
-* python 3 - `sudo apt install python3`
+* python 3 - `sudo apt install python3` (XXX - TODO - you actually need 3.6
+  and if your distro doesn't some with that, you need to google how to make it
```
rfk, line 205: s/some/come/

docs/howtos/setup-android-build-environment.md
```diff
@@ -177,3 +242,23 @@ We now want to configure maven to use the native windows maven repository - then
...
 
 \o/
+
+# Other obsolete content
+
+This content is probably outdated. It should either be removed, or saved!
+
+## Publishing to your local maven repo
+
+Note that this has been make obsolete by the `autoPublish` mechanisms above.
```
rfk, line 252: s/make/made/

docs/howtos/setup-android-build-environment.md
```diff
@@ -177,3 +242,23 @@ We now want to configure maven to use the native windows maven repository - then
 * Now you should be ready to roll - `./gradlew install` should complete and publish the components to your native maven repo!
 
 \o/
+
+# Other obsolete content
+
+This content is probably outdated. It should either be removed, or saved!
```
rfk, line 248: I would support just removing this FWIW; if you already know about `publishToMavenLocal` then you don't need these docs, and if you don't know about `publishToMavenLocal` then you're near certain to have a bad time and should instead use the auto-publish flow. I believe the detailed docs on the autopublish flow also have fallback docs describing `publishToMavenLocal`, so we wouldn't be losing much by removing them here.

mhammond:
TBH, I now have a working environment, so I'm not that interested in starting from scratch again. Given that, it seems wasteful to have 2 people work on making this document match how docker does things when 1 would work just as well - I guess you have vmware and a windows license, or can obtain one from servicenow.

I'm happy to close this if you think it's a step backwards from the old version, but that old version simply did not work - let me know if I should just close it.

mhammond:
(well, I expect you don't need windows - a clean ubuntu 18 seems fine?)

docs/howtos/setup-android-build-environment.md
```diff
@@ -108,10 +85,106 @@ you will find:
...
+
+  If not already installed, you can install the "command-line" only tools:
+
+   - Visit https://developer.android.com/studio/, at the bottom of the page locate the current SDKs for linux
+     at time of writing, this is https://dl.google.com/android/repository/commandlinetools-linux-6514223_latest.zip
+
+  Install it by executing:
+
+  > % cd ~  
+  > % mkdir -p android-sdk/cmdline-tools
```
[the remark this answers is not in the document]
eoger: Turns out this is how one installs the SDK these days

docs/howtos/setup-android-build-environment.md
```diff
@@ -108,10 +85,106 @@ you will find:
...
+
+  > `$ANDROID_HOME/cmdline-tools/tools/bin/sdkmanager --install "ndk;20.0.5594570"
+
+   - If you have Android Studio or otherwise a GUI version of the SDK, use that
+     GUI.
+
+  In either case, you don't need to select where it is installed or to set
+  further environment variables.
+
+At the end of this process, all you need is `ANDROID_HOME` set!
```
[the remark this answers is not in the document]
eoger: Turns out my NDK knowledge was out-of-date, and installing using the SDK is The Proper Way now. Thanks Mark for digging into this, I'll make a PR so we use this everywhere (it should make new contributors lives easier).

rfk:
Let's close this in favour of https://github.com/mozilla/application-services/pull/3228

PR #3218, opened 2020-06-10
adds cargo aliasing to use asdev

tarikeshaq:

The aliases as shown in the demo. Still need to add documentation, and would love to get any ideas on why you would or wouldn't use this. On that note, **any ideas where is the best place to document this?**

An alternative would be **not** to add this, and just document that users can use the asdev tool using `cargo asdev [subcommand]`

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,6 @@
 [alias]
 regen-protobufs = ["run", "--bin", "protobuf-gen", "tools/protobuf_files.toml"]
+dev-install = ["install", "asdev"]
+all_tests = ["asdev", "test"]
+rust_tests = ["asdev", "test_rust"]
+verify_env = ["asdev", "verify_env"]
--- a/docs/building.md
+++ b/docs/building.md
@@ -49,3 +49,17 @@ The following command will ensure your environment is ready to build the project
 ````
 
 The Xcode project is located at `megazords/ios/MozillaAppServices.xcodeproj`.
+
+## (Optional) Dev tool
+You can also install and use `asdev` to help you run the above commands, you can do that by running the following:
+
+```
+cargo dev-install
+```
+
+This will allow your to run all the above commands without having to remember where the scripts are. You will be able to do 
+
+```
+cargo verify_env
+``` 
+Which would run all three verification scrips. To check what other commands the dev tool supports, run `cargo asdev -h` or open the dialog using `cargo asdev`
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,8 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v61.0.5...master)
+
+
+## General
+
+- Adds cargo aliases to download and use `asdev` ([#3218](https://github.com/mozilla/application-services/pull/3218))
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -6,6 +6,28 @@ Anyone is welcome to help with the Application Services project. Feel free to ge
 - Mailing list: <https://mail.mozilla.org/listinfo/sync-dev>
 - and of course, [the issues list](https://github.com/mozilla/application-services/issues)
 
+We recommend installing the `asdev` tool, it will help you navigate a variety of tasks in this repository. You can do that by running the following:
+```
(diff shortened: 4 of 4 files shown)


docs/building.md
```diff
@@ -27,7 +27,7 @@ Roughly you need Java 8, the Android SDK, the NDK R20 and a bunch of environment
 Your best friend is the following command:
 
 ```
-./libs/verify-android-environment.sh
+cargo  verify_android
```
vladikoff, line 30: 2 spaces here

docs/building.md
```diff
@@ -12,7 +12,7 @@ This part assumes you are not interested in building the Android or iOS wrappers
 The following command will ensure your environment variables are set properly and build the dependencies needed by the project:
 
 ```
-./libs/verify-desktop-environment.sh
+cargo verify_desktop
```
eoger, line 15: -1 on that, I don't mind having a-sdev as an optional thing to the repo, however I disagree with making it mandatory (which is the de-facto the case when it replaces the canonical commands in docs)

vladikoff, line 15: if someone reads `building.md` before reading `docs/contributing.md` they might not have have `asdev` installed and get 
```
error: no such subcommand: `asdev`

	Did you mean `audit`?
```

Maybe worth repeating what you have in contributing.md  in a shorter form?

.cargo/config
```diff
@@ -1,2 +1,15 @@
 [alias]
 regen-protobufs = ["run", "--bin", "protobuf-gen", "tools/protobuf_files.toml"]
+dev-install = ["install", "asdev"]
+all_tests = ["asdev", "test"]
+rust_tests = ["asdev", "test_rust"]
+verify_desktop = ["asdev", "verify_desktop"]
+verify_android = ["asdev", "verify_android"]
+verify_ios = ["asdev", "verify_ios"]
```
vladikoff, line 8: @tarikeshaq hm looking at this list of `verify`. What do you think about squashing this into something like `cargo verify_env` (`cargo verify` might be too generic)? Then we can leave the other scripts alone in https://github.com/mozilla/application-services/pull/3218/files#r438921631 but prefix that part of the documentation to just run `cargo verify_env` and get output for all 3 platforms

tarikeshaq: That's possible! I'd have to change `asdev` to support doing all three back to back, but that's easy enough. Was wondering, would that be better, or maybe add an 'optional' section and leave the other documentation alone. Something like:
## (Optional) Dev tool
You can also install and use `asdev` to help you run the above commands, you can do that by running the following:

```
$ cargo dev-install
```

This will allow your to run the above commands without having to remember where the scripts are. For example, you would be able to run `cargo verify_desktop` which would run `./libs/verify-desktop-environment.sh`. To check the available commands, run `cargo asdev -h` or open the dialog using `cargo asdev`

vladikoff: > I'd have to change asdev to support doing all three back to back, but that's easy enough. 

Yeap, let's do that 👍

.cargo/config
```diff
@@ -1,2 +1,13 @@
 [alias]
 regen-protobufs = ["run", "--bin", "protobuf-gen", "tools/protobuf_files.toml"]
+dev-install = ["install", "asdev"]
+all_tests = ["asdev", "test"]
+rust_tests = ["asdev", "test_rust"]
+verify_env = ["asdev", "verify_env"]
+smoke_android = ["asdev", "smoke_android"]
+smoke_fenix = ["asdev", "smoke_fenix"]
+smoke_ios = ["asdev", "smoke_ios"]
```
vladikoff, line 9: let's remove the smoke testing for now and make it simpler. I filed an issue https://github.com/a-s-dev/asdev/issues/14

.cargo/config
```diff
@@ -1,2 +1,13 @@
...
+tag_minor = ["asdev", "tag_minor"]
+prepare_release = ["asdev", "prepare_release"]
+lint_bash = ["asdev", "lint_bash"]
+update_pr = ["asdev", "cargo_update"]
```
vladikoff, line 13: let's remove the above for now and fix these up a bit. I created an issue for the `prepare_release` script https://github.com/a-s-dev/asdev/issues/15

tarikeshaq: Done, got rid of those for now, until I clean up the tool more

PR #3219, opened 2020-06-10
Short circuit if /invoke_command gets rate limited, until timeout period passes

### LINKED ISSUE #3231: [rust] Don't keep trying to send tabs if the server has rate-limited us
Both Fenix and Desktop have UI that supports sending multiple tabs at once. In fact the user can easily attempt to send themselves many hundreds or thousands of tabs by right-clicking the tab bar, choosing "select all tabs", and then "send `N` tabs to device". The send-tab infrastructure only supports sending one tab at a time. So sending `N` tabs gets turned into `N` calls to our `send_tab` function.

This does not work out well for large values of `N`.

The FxA server has some anti-abuse rate-limiting on the `/invoke_command` endpoint that underlies send-tab. It's generous, designed mostly to stop really obvious abuse or really serious client bugs that might otherwise bring down the service. So the first `n < N` tabs will be sent successfully, but all the rest will be rejected by the server with a `429` error.  This error appears to be silently dropped - not only does the user not receive feedback that some tabs failed to send, but the client keeps attempting to send all the remaining tabs!

We should, instead, _not_ do that, and refuse to attempt any more tab sending according to the backoff information given by the FxA server.

I believe this issue to exist on both Desktop and the rust fxa-client, but I'm filing it here because we might end up fixing it in Desktop by replacing its send-tab implementation with the rust component.

(By the way, there is also convenient UI for sending tabs to "all devices" as a single action. With `M` devices connected to your account this produces `N x M` individual `send_tab` calls. You can imagine how well that will work out for large values of both `N` and `M`...)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1391)
┆Story Points: 8
┆Epic: Trusted Send Tab Telemetry
┆Sprint: SYNC - end 2020-06-19

tarikeshaq:

fixes #3231

### DIFF
--- a/components/fxa-client/src/http_client.rs
+++ b/components/fxa-client/src/http_client.rs
@@ -8,11 +8,14 @@ use rc_crypto::{digest, hkdf, hmac};
 use serde_derive::*;
 use serde_json::json;
 use std::collections::HashMap;
+use std::sync::Mutex;
+use std::time::{Duration, Instant};
 use url::Url;
 use viaduct::{header_names, status_codes, Method, Request, Response};
 
 const HAWK_HKDF_SALT: [u8; 32] = [0b0; 32];
 const HAWK_KEY_LENGTH: usize = 32;
+const RETRY_AFTER_DEFAULT_SECONDS: u64 = 10;
 
 #[cfg_attr(test, mockiato::mockable)]
 pub trait FxAClient {
@@ -104,7 +107,17 @@ pub trait FxAClient {
     ) -> Result<HashMap<String, ScopedKeyDataResponse>>;
 }
 
-pub struct Client;
+enum HttpClientState {
+    Ok,
+    Backoff {
+        backoff_end_duration: Duration,
+        time_since_backoff: Instant,
+    },
+}
+
+pub struct Client {
+    state: Mutex<HashMap<String, HttpClientState>>,
+}
 impl FxAClient for Client {
     fn profile(
         &self,
@@ -118,7 +131,7 @@ impl FxAClient for Client {
         if let Some(etag) = etag {
             request = request.header(header_names::IF_NONE_MATCH, format!("\"{}\"", etag))?;
         }
-        let resp = Self::make_request(request)?;
+        let resp = self.make_request(request)?;
         if resp.status == status_codes::NOT_MODIFIED {
             return Ok(None);
         }
@@ -165,7 +178,7 @@ impl FxAClient for Client {
         let request = HawkRequestBuilder::new(Method::Post, url, &key)
             .body(body)
             .build()?;
-        Ok(Self::make_request(request)?.json()?)
+        Ok(self.make_request(request)?.json()?)
     }
 
     // For the regular generation of an `access_token` from long-lived credentials.
@@ -202,7 +215,7 @@ impl FxAClient for Client {
         let request = HawkRequestBuilder::new(Method::Post, url, &key)
             .body(parameters)
             .build()?;
-        Self::make_request(request)?.json().map_err(Into::into)
+        self.make_request(request)?.json().map_err(Into::into)
     }
 
     fn authorization_code_using_session_token(
@@ -227,7 +240,7 @@ impl FxAClient for Client {
             .body(parameters)
             .build()?;
 
-        Ok(Self::make_request(request)?.json()?)
+        Ok(self.make_request(request)?.json()?)
     }
 
     fn oauth_introspect_refresh_token(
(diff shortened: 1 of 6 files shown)


components/fxa-client/src/http_client.rs
```diff
@@ -104,7 +106,17 @@ pub trait FxAClient {
     ) -> Result<HashMap<String, ScopedKeyDataResponse>>;
 }
 
-pub struct Client;
+enum FxAClientState {
```
vladikoff, line 109: maybe this should be HttpClientState? 

We already have an FxAClientState

components/fxa-client/src/http_client.rs
```diff
@@ -104,7 +106,17 @@ pub trait FxAClient {
...
+    Ok,
+    TimeOut {
+        time_out_end_duration: Duration,
```
vladikoff, line 112: It could be that different endpoints we have will have different `retryAfter` limits. Is it possible to HashMap with a { 
endpoint: time_out_end_time
}
Example draft: `/commands` : `1591892366`

components/fxa-client/src/http_client.rs
```diff
@@ -309,7 +333,36 @@ impl FxAClient for Client {
             .header(header_names::AUTHORIZATION, bearer_token(refresh_token))?
             .header(header_names::CONTENT_TYPE, "application/json")?
             .body(body.to_string());
-        Self::make_request(request)?;
+        // I am a really bad way to handle this
+        // I exist only because Tarik is experimenting :)
+        // I will be replaced with a better approach
+        match Self::make_request(request) {
```
vladikoff, line 339: I think this is on the right track, looking at the code required to just short circuit the command route, we should just upgrade our `make_request` to be ubiquitous for other FxA endpoints.

components/fxa-client/src/http_client.rs
```diff
@@ -395,25 +410,64 @@ impl Client {
...
-                Ok(json) => Err(ErrorKind::RemoteError {
-                    code: json["code"].as_u64().unwrap_or(0),
-                    errno: json["errno"].as_u64().unwrap_or(0),
-                    error: json["error"].as_str().unwrap_or("").to_string(),
-                    message: json["message"].as_str().unwrap_or("").to_string(),
-                    info: json["info"].as_str().unwrap_or("").to_string(),
-                }
-                .into()),
-                Err(_) => Err(resp.require_success().unwrap_err().into()),
+            match resp.status {
```
vladikoff, line 468: nice this looks cleaner now 👍

components/fxa-client/src/http_client.rs
```diff
@@ -482,7 +536,8 @@ pub fn send_authorization_request(
     let req = HawkRequestBuilder::new(Method::Post, auth_endpoint, auth_key)
         .body(serde_json::to_value(&auth_params)?)
         .build()?;
-    let resp: serde_json::Value = Client::make_request(req)?.json()?;
+    let client = Client::new();
```
vladikoff, line 539: We are doing a `Client::new` in each request, is there a way of not doing that?

components/fxa-client/src/http_client.rs
```diff
@@ -894,4 +951,164 @@ mod tests {
         };
         assert_eq!("{\"grant_type\":\"refresh_token\",\"client_id\":\"bar\",\"refresh_token\":\"foo\",\"scope\":\"bobo\",\"ttl\":123}", serde_json::to_string(&using_code).unwrap());
     }
+
+    #[test]
+    fn test_backoff() {
```
vladikoff, line 956: thank you for creating these tests ! 🏅

components/fxa-client/src/http_client.rs
tarikeshaq: Yup! I can move those functions to be inside the Client impl and have the integration tests worry about creating the client 😄

components/fxa-client/src/http_client.rs
tarikeshaq: Tbh was kinda surprised that we don't have tests for everything else in this file 🙃

components/fxa-client/src/http_client.rs
```diff
@@ -472,6 +526,10 @@ impl AuthorizationParameters {
     }
 }
 
+// Keeping those functions out of the FxAClient trate becouse functions in the
```
tarikeshaq, line 529: Added those because of the Client::new() 😭 Tried adding the functions into the FxAClient trait, but since they are test only they made the mockiato upset and unable to compile tests. Any ideas, or would them hanging around here be okay?

vladikoff: That's fine for now, this comment here will be helpful for the future 👍

PR #3222, opened 2020-06-11
Use Android SDK's side-by-side NDK [ci full]

eoger:
This makes it easier for contributors to install the Android NDK as it now comes bundled with the Android SDK. This PR also removes the `ANDROID_NDK_HOME` `ANDROID_NDK_ROOT` env variables requirement which only leaves `ANDROID_HOME`, just like a pretty classic Android build environment.

EDIT:
Also takes cares of the following warning we've been seeing for a while:
```
WARNING: Support for ANDROID_NDK_HOME is deprecated and will be removed in the future. Use android.ndkVersion in build.gradle instead.
Support for ANDROID_NDK_HOME is deprecated and will be removed in the future. Use android.ndkVersion in build.gradle instead.
```

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -10,5 +10,11 @@
 - Adds jwe encryption in scoped_keys. ([#3195](https://github.com/mozilla/application-services/pull/3195))
 - Adds an implementation for [pbkdf2](https://www.ietf.org/rfc/rfc2898.txt). ([#3193](https://github.com/mozilla/application-services/pull/3193))
 
+## Android
+
+- From now on the project uses the Android SDK manager side-by-side NDK. ([#3222](https://github.com/mozilla/application-services/pull/3222))
+  - Download the new NDK by running `./verify-android-environment.sh` in the `libs` directory.
+    - Alternatively, you can download the NDK in Android Studio by going in Tools > SDK Manager > SDK Tools > NDK (check Show Package Details) and check `21.3.6528147`.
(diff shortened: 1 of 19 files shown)


libs/android_defaults.sh
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# Find the NDK.
+pushd ..
+NDK_VERSION=$(./gradlew -q printNdkVersion | tail -1)
```
mhammond, line 5: I'm surprised you still need these at all! (But np if you do)

eoger: Yeah the PATH to the NDK is needed to compile NSS/SQLCipher and generate Android symbols.

CHANGES_UNRELEASED.md
```diff
@@ -10,5 +10,11 @@
 - Adds jwe encryption in scoped_keys. ([#3195](https://github.com/mozilla/application-services/pull/3195))
 - Adds an implementation for [pbkdf2](https://www.ietf.org/rfc/rfc2898.txt). ([#3193](https://github.com/mozilla/application-services/pull/3193))
 
+## Android
+
+- From now on the project uses the Android SDK manager side-by-side NDK. ([#3222](https://github.com/mozilla/application-services/pull/3222))
+  - Download the new NDK by running `./verify-android-environment.sh` in the `libs` directory.
+    - Alternatively, you can download the NDK in Android Studio by going in Tools > SDK Manager > SDK Tools > NDK (check Show Package Details) and check `21.3.6528147`.
+  - The `ANDROID_NDK_ROOT`, `ANDROID_NDK_HOME` environment variables (and the directory they point to) can be removed.
```
eoger, line 18: @mozilla/a-s-review heads up

Issue #3214, opened 2020-06-09
Protobuf decode error: failed to decode Protobuf message: unexpected end group tag


data-sync-user:
Seems like there's a problem with "Protobuf decode error: failed to decode Protobuf message: unexpected end group tag" for Synced Tabs.

Synced tabs is behind a nightly flag, cant remove the flag with this crash

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1382)
┆Story Points: 8
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-07-03

data-sync-user:
➤ Vladislav Filippov commented:

From Fenix team, they mentioned Synced Tabs is important for release

rfk:
> Synced tabs is behind a nightly flag, cant remove the flag with this crash

To clarify, is it default off in nightly (but you can enable it via a flag) or is it default on in nightly (but a flag ensures that it's default off on release)?

thomcc:
Sadly this appears to still be happening :(

We should switch from protobufs to JSON here as a last resort.

PR #3224, opened 2020-06-11
Possible fix for sync tabs issue.

thomcc:
Fix for #3214 if #3223 is the same issue.

It makes a lot of sense that would be the issue but we get a different error so hard to say.


components/tabs/android/src/test/java/mozilla/appservices/remotetabs/RemoteTabsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+class RemoteTabsTest {
+
+    @Before
+    fun init() {
+        Megazord.init()
+    }
+    @Test
+    fun doTest() {
+        RemoteTabsProvider().use { tabs ->
+            tabs.setLocalTabs(listOf(
```
thomcc, line 21: This causes an exception without the patch.

PR #3264, opened 2020-06-22
Add missing auth params [ci full]

### LINKED ISSUE #3122: authorize_code_using_session_token missing fields

tarikeshaq:
fixes #3122 

Note: The full flow isn't tested 😞 The Rust side of things is tested using the sync integration tests, but the Kotlin -> Rust protobuf flow isn't yet. Hopefully as our Kotlin -> Rust integration testing ecosystem (#3306) grows we'll be able to test this, among other integration flows 🕺

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -8,4 +8,5 @@
 
 ### ⚠️ Breaking changes ⚠️
 
-- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOauthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
+- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
+- Changes the `authorizeOAuthCode` API to now accept an `AuthorizationParams` object instead of the individual parameters. The `AuthorizationParams` also includes optional `AuthorizationPKCEParams` that contain the `codeChallenge`, `codeChallengeMethod`. `AuthorizationParams` also includes an optional `keysJwk` for requesting keys ([#3264](https://github.com/mozilla/application-services/pull/3264))
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
@@ -1,0 +1,65 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.appservices.fxaclient
+
+data class AuthorizationParams(
+    val clientId: String,
+    val scopes: Array<String>,
+    val state: String,
+    val accessType: String = "online",
+    val pkceParams: AuthorizationPKCEParams? = null,
+    val keysJwk: String? = null
+) {
+    fun intoMessage(): MsgTypes.AuthorizationParams {
+        var b = MsgTypes.AuthorizationParams.newBuilder()
+                .setClientId(this.clientId)
+                .setScope(this.scopes.joinToString(" "))
+                .setState(this.state)
+                .setAccessType(this.accessType)
+        if (this.pkceParams != null)
+            b = b.setPkceParams(this.pkceParams.intoMessage())
+        if (this.keysJwk != null)
+            b = b.setKeysJwk(this.keysJwk)
+        return b.build()
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as AuthorizationParams
+
+        if (clientId != other.clientId) return false
+        if (!scopes.contentEquals(other.scopes)) return false
+        if (state != other.state) return false
+        if (accessType != other.accessType) return false
+        if (pkceParams != other.pkceParams) return false
+        if (keysJwk != other.keysJwk) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = clientId.hashCode()
+        result = 31 * result + scopes.contentHashCode()
+        result = 31 * result + state.hashCode()
+        result = 31 * result + accessType.hashCode()
+        result = 31 * result + (pkceParams?.hashCode() ?: 0)
+        result = 31 * result + (keysJwk?.hashCode() ?: 0)
+        return result
+    }
+}
+
+data class AuthorizationPKCEParams(
+    val codeChallenge: String,
+    val codeChallengeMethod: String = "S256"
+) {
+    fun intoMessage(): MsgTypes.AuthorizationPKCEParams {
+        return MsgTypes.AuthorizationPKCEParams.newBuilder()
+                .setCodeChallenge(this.codeChallenge)
+                .setCodeChallengeMethod(this.codeChallengeMethod)
+                .build()
+    }
+}
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
@@ -308,17 +308,33 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      * @param scopes Array of scopes for the OAuth code.
      * @param state OAuth flow state.
      * @param accessType Type of access, "offline" or "online".
+     * @param codeChallenge Code challenge for Proof Key for Code Exchange (PKCE)
+     * @param codeChallengeMethod for PKCE, currently only 'S256' supported
+     * @param keysJwk Key to encrypt keys_jwe using, passed in as a base64 string
      * This performs network requests, and should not be used on the main thread.
      */
     fun authorizeOAuthCode(
         clientId: String,
         scopes: Array<String>,
         state: String,
-        accessType: String = "online"
+        accessType: String = "online",
+        codeChallenge: String,
+        codeChallengeMethod: String = "S256",
+        keysJwk: String
     ): String {
         val scope = scopes.joinToString(" ")
         return rustCallWithLock { e ->
-            LibFxAFFI.INSTANCE.fxa_authorize_auth_code(this.handle.get(), clientId, scope, state, accessType, e)
+            LibFxAFFI.INSTANCE.fxa_authorize_auth_code(
+                this.handle.get(),
+                clientId,
+                scope,
+                state,
+                accessType,
+                codeChallenge,
+                codeChallengeMethod,
+                keysJwk,
+                e
+            )
         }.getAndConsumeRustString()
     }
 
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
@@ -304,35 +304,19 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
     /**
      * Provisions an OAuth code using the session token from state
      *
-     * @param clientId OAuth client id.
-     * @param scopes Array of scopes for the OAuth code.
-     * @param state OAuth flow state.
-     * @param accessType Type of access, "offline" or "online".
-     * @param codeChallenge Code challenge for Proof Key for Code Exchange (PKCE)
-     * @param codeChallengeMethod for PKCE, currently only 'S256' supported
-     * @param keysJwk Key to encrypt keys_jwe using, passed in as a base64 string
+     * @param authParams Parameters needed for the authorization request
      * This performs network requests, and should not be used on the main thread.
      */
     fun authorizeOAuthCode(
-        clientId: String,
-        scopes: Array<String>,
-        state: String,
-        accessType: String = "online",
-        codeChallenge: String,
-        codeChallengeMethod: String = "S256",
-        keysJwk: String
+        authParams: AuthorizationParams
     ): String {
-        val scope = scopes.joinToString(" ")
+        val (nioBuf, len) = authParams.intoMessage().toNioDirectBuffer()
         return rustCallWithLock { e ->
+            val ptr = Native.getDirectBufferPointer(nioBuf)
             LibFxAFFI.INSTANCE.fxa_authorize_auth_code(
                 this.handle.get(),
-                clientId,
-                scope,
-                state,
-                accessType,
-                codeChallenge,
-                codeChallengeMethod,
-                keysJwk,
+                ptr,
+                len,
                 e
             )
         }.getAndConsumeRustString()
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
@@ -56,7 +56,7 @@ internal interface LibFxAFFI : Library {
     fun fxa_get_access_token(fxa: FxaHandle, scope: String, ttl: Long, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_get_session_token(fxa: FxaHandle, e: RustError.ByReference): Pointer?
     fun fxa_get_current_device_id(fxa: FxaHandle, e: RustError.ByReference): Pointer?
-    fun fxa_authorize_auth_code(fxa: FxaHandle, clientId: String, scope: String, state: String, accessType: String, e: RustError.ByReference): Pointer?
+    fun fxa_authorize_auth_code(fxa: FxaHandle, clientId: String, scope: String, state: String, accessType: String, codeChallenge: String, codeChallengeMethod: String, keysJwk: String, e: RustError.ByReference): Pointer?
     fun fxa_check_authorization_status(fxa: FxaHandle, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_clear_access_token_cache(fxa: FxaHandle, e: RustError.ByReference)
 
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
@@ -56,7 +56,7 @@ internal interface LibFxAFFI : Library {
     fun fxa_get_access_token(fxa: FxaHandle, scope: String, ttl: Long, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_get_session_token(fxa: FxaHandle, e: RustError.ByReference): Pointer?
     fun fxa_get_current_device_id(fxa: FxaHandle, e: RustError.ByReference): Pointer?
-    fun fxa_authorize_auth_code(fxa: FxaHandle, clientId: String, scope: String, state: String, accessType: String, codeChallenge: String, codeChallengeMethod: String, keysJwk: String, e: RustError.ByReference): Pointer?
+    fun fxa_authorize_auth_code(fxa: FxaHandle, auth_params: Pointer, auth_params_len: Int, e: RustError.ByReference): Pointer?
     fun fxa_check_authorization_status(fxa: FxaHandle, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_clear_access_token_cache(fxa: FxaHandle, e: RustError.ByReference)
 
--- a/components/fxa-client/ffi/src/lib.rs
+++ b/components/fxa-client/ffi/src/lib.rs
@@ -13,6 +13,7 @@ use ffi_support::{
 };
 use fxa_client::{
     device::{Capability as DeviceCapability, PushSubscription},
+    ffi::AuthorizationParameters,
     migrator::MigrationState,
     msg_types, FirefoxAccount,
 };
@@ -447,35 +448,16 @@ pub extern "C" fn fxa_get_devices(
 /// A destructor [fxa_bytebuffer_free] is provided for releasing the memory for this
 /// pointer type.
 #[no_mangle]
-pub extern "C" fn fxa_authorize_auth_code(
+pub unsafe extern "C" fn fxa_authorize_auth_code(
     handle: u64,
-    client_id: FfiStr<'_>,
-    scope: FfiStr<'_>,
-    state: FfiStr<'_>,
-    access_type: FfiStr<'_>,
-    code_challenge: FfiStr<'_>,
-    code_challenge_method: FfiStr<'_>,
-    keys_jwk: FfiStr<'_>,
+    auth_params: *const u8,
+    auth_params_len: i32,
     error: &mut ExternError,
 ) -> *mut c_char {
     log::debug!("fxa_authorize_auth_code");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
-        let client_id = client_id.as_str();
-        let scope = scope.as_str();
-        let state = state.as_str();
-        let access_type = access_type.as_str();
-        let code_challenge = code_challenge.as_str();
-        let code_challenge_method = code_challenge_method.as_str();
-        let keys_jwk = keys_jwk.as_str();
-        fxa.authorize_code_using_session_token(
-            client_id,
-            scope,
-            state,
-            access_type,
-            code_challenge,
-            code_challenge_method,
-            keys_jwk,
-        )
+        let auth_params = AuthorizationParameters::from_protobuf_ptr(auth_params, auth_params_len)?;
+        fxa.authorize_code_using_session_token(auth_params)
     })
 }
 
--- a/components/fxa-client/ffi/src/lib.rs
+++ b/components/fxa-client/ffi/src/lib.rs
@@ -453,6 +453,9 @@ pub extern "C" fn fxa_authorize_auth_code(
     scope: FfiStr<'_>,
     state: FfiStr<'_>,
     access_type: FfiStr<'_>,
+    code_challenge: FfiStr<'_>,
+    code_challenge_method: FfiStr<'_>,
+    keys_jwk: FfiStr<'_>,
     error: &mut ExternError,
 ) -> *mut c_char {
     log::debug!("fxa_authorize_auth_code");
@@ -461,7 +464,18 @@ pub extern "C" fn fxa_authorize_auth_code(
         let scope = scope.as_str();
         let state = state.as_str();
         let access_type = access_type.as_str();
-        fxa.authorize_code_using_session_token(client_id, scope, state, access_type)
+        let code_challenge = code_challenge.as_str();
+        let code_challenge_method = code_challenge_method.as_str();
+        let keys_jwk = keys_jwk.as_str();
+        fxa.authorize_code_using_session_token(
+            client_id,
+            scope,
+            state,
+            access_type,
+            code_challenge,
+            code_challenge_method,
+            keys_jwk,
+        )
     })
 }
 
--- a/components/fxa-client/src/ffi.rs
+++ b/components/fxa-client/src/ffi.rs
@@ -12,11 +12,12 @@
 //!
 //! None of this is that bad in practice, but much of it is not ideal.
 
+pub use crate::oauth::{AuthorizationPKCEParams, AuthorizationParameters};
 use crate::{
     commands,
     device::{Capability as DeviceCapability, Device, PushSubscription, Type as DeviceType},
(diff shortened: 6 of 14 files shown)


components/fxa-client/src/oauth.rs
```diff
@@ -163,20 +163,67 @@ impl FirefoxAccount {
...
         scope: &str,
         state: &str,
         access_type: &str,
     ) -> Result<String> {
         let session_token = self.get_session_token()?;
+        let mut code_challenge = None;
+        let mut keys_jwe = None;
+        let mut code_challenge_method = None;
+        // Is this check even neccesary? Can we just compute the keys_jwe and
+        // Send it anyway, filtering the self.state.scoped_keys on the reqeusted scopes
```
rfk, line 179: Do you mean that we would always create a `keys_jwe`, but the items that it contains would be filetered based on `scope`? (And so if the `OLD_SYNC` scope was not requested, it would be a JWE containing the string `"{}"`?). That...sounds OK to me, if it works out to be simpler to implement.

components/fxa-client/src/oauth.rs
```diff
@@ -163,20 +163,67 @@ impl FirefoxAccount {
...
+        if scope
+            .split_whitespace()
+            .any(|split_scope| split_scope == scopes::OLD_SYNC)
+        {
+            let code_verifier = util::random_base64_url_string(43)?;
+            let code_challenge_digest = digest::digest(&digest::SHA256, &code_verifier.as_bytes())?;
```
rfk, line 185: Generating these here does not feel right to me (but I could be misunderstanding or misremembering what this is trying to achieve).

In my head, we're expanding this method so that we can use it for authorizing *some other client* to access the user's account, e.g. as part of the pairing flow. In that scenario, we're conceptually acting as part of the FxA server and fulfilling an OAuth request that we've received from some other client.

If that's the case, then generating `code_challenge` here ourselves doesn't achieve anything. When we get the OAuth code from the server and send it back to the requesting client, that client won't be able to use it because it doesn't know the corresponding `code_verifier`.

I expect that we will need to accept `code_challenge`, `code_challenge_method` and `keys_jwk` as input parameters to this function, rather than generating them ourselves internally. (And you may find that doing this, actually makes the integration code easier, because you don't have to mess with `self.flow_store`).

components/fxa-client/examples/auth_code_session.rs
```diff
@@ -0,0 +1,200 @@
...
+        .truncate(true)
+        .create(true)
+        .open(CREDENTIALS_PATH)
+        .unwrap();
+    write!(file, "{}", json).unwrap();
+    file.flush().unwrap();
+}
+
+fn create_fxa_creds(cfg: Config) -> Result<FirefoxAccount> {
+    let mut acct = FirefoxAccount::with_config(cfg.clone());
```
rfk, line 55: To make this example work the way the flow would work in the real world, I suggest making two separate `FirefoxAccount` instances and pretend that you're doing a pairing flow between them. The first one would be the "authority" side of the pairing flow, and you'd sign in to it manually using the "enter password" logic that you have here now.  The second instance would be the "supplicant" side of the pairing flow, and you would:

* Call `supplicant.begin_auth_flow()` to obtain an OAuth request URL.
* Parse the OAuth parameters out of the returned OAuth URL (`client_id`, `scope`, `code_challenge`, etc)
* Call `authority.authorize_code_using_session_token()` passing it all the OAuth params that you parsed out above. This will return an OAuth code.
* Use the returned OAuth code to call `supplicant.complete_oauth_flow()` to complete the login process.

In this way, you've used an alreadys-signed-in FirefoxAccount instance to transfer the signed-in state to a new instance, much like we'd want to do with two separate apps in the wild.

components/fxa-client/src/oauth.rs
tarikeshaq: Alrighty updated it to do that. Makes a lot more sense. I also propagated the parameters all the way up the FFI so the caller can set the parameters. This would then be a breaking change 😅 (Although it seems like this API is only used in Android Components? I could make a PR for them if this ends up merging in this state)

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -302,17 +302,33 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
...
      */
     fun authorizeOAuthCode(
         clientId: String,
         scopes: Array<String>,
         state: String,
-        accessType: String = "online"
+        accessType: String = "online",
+        codeChallenge: String,
+        codeChallengeMethod: String = "S256",
+        keysJwk: String
```
rfk, line 317: /cc @vladikoff, this might be another good candidate for accepting an `OAuthParams` struct of some sort, rather than a bunch of individual parameters.

(For context: in our 1:1 earlier today @vladikoff and I were discussing how we don't really have a good example of how to pass a rich structure like that *in* to the Rust code; we mostly use the protobuf message types for passing things *out* of Rust).

thomcc:
FYI: you can add required features for an example by adding an [[examples]] section for it with `required-features = [...]`

See https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target

components/fxa-client/src/oauth.rs
```diff
@@ -167,25 +166,51 @@ impl FirefoxAccount {
...
+        } else {
+            AuthorizationRequestParameters {
+                client_id: auth_params.client_id,
+                scope: auth_params.scope,
+                state: auth_params.state,
+                access_type: auth_params.access_type,
+                code_challenge: None,
+                code_challenge_method: None,
+                keys_jwe: None,
+            }
```
eoger, line 207: Nit: you could probably factor the `AuthorizationRequestParameters` construction outside of the `if-let` block, whereas the `if-let` block would return a `Option<keys_jwe type>`, but disregard this comment if the memory checker gives you a hard time doing this.

testing/sync-test/src/auth.rs
```diff
@@ -165,19 +165,57 @@ impl TestAccount {
...
+            scope,
+            client_id,
+            &auth::derive_auth_key_from_session_token(&self.session_token)?,
+            &self.cfg,
+            (&self.k_sync, &self.xcs),
+        )?;
+        // Setup authority account that is logged in and has the appropriate scoped keys
+        let mut fxa = FirefoxAccount::with_config(self.cfg.clone());
+        fxa.set_scoped_keys(scoped_keys);
+        fxa.set_session_token(&self.session_token);
```
eoger, line 203: Is this worth introducing a new constructor?

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -316,9 +317,40 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
...
+    }
+
+    /**
+     * Provisions an OAuth code using the session token from state
+     * To be used directly by clients that require keys
+     *
+     * @param authParams Parameters needed for authorization request
+     * This performs network requests, and should not be used on the main thread.
+     */
+    fun authorizeOAuthCode(
```
eoger, line 342: As discussed on Slack, since only a-c (and no downstream customer) uses this, I think it's worth breaking the API compatibility to keep only 1 `authorizeOAuthCode` method.

CHANGES_UNRELEASED.md
```diff
@@ -8,4 +8,5 @@
 
 ### ⚠️ Breaking changes ⚠️
 
-- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOauthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
\ No newline at end of file
+- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOauthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
```
rfk, line 11: nit: `beginOAuthFlow` capitalization.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
```diff
@@ -0,0 +1,62 @@
...
+        result = 31 * result + state.hashCode()
+        result = 31 * result + accessType.hashCode()
+        result = 31 * result + (keyParams?.hashCode() ?: 0)
+        return result
+    }
+}
+
+data class AuthorizationKeyParams(
+    val codeChallenge: String,
+    val codeChallengeMethod: String = "S256",
```
rfk, line 53: We often conflate them, but `codeChallenge` and `codeChallengeMethod` don't actually have anything to do with fetching keys. These parameters are for [PKCE](https://tools.ietf.org/html/rfc7636), which allows the OAuth flow to be used by what we call "public clients"; that is, clients that cannot maintain a secret key with which to authenticate their calls to the OAuth server.

It just so happens that all of our current public clients also use the scoped-keys feature, and that anyone using the scoped-keys feature is almost certainly a public client (because we would never deliver the user's secret keys to a server-server OAuth client). So we always see these params together in practice, even though they should not be coupled together in the user's mental model of the system. One day we might add a public client that doesn't want scoped keys, but still needs to use PKCE.

So, I don't think we should bundle them into the same object here. You could make a separate `AuthorizationPKCEParams` class to hold just the PKCE-related stuff though, if that makes things easier to manage.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.appservices.fxaclient
+
+data class AuthorizationParams(
```
rfk, line 7: I like this abstraction a lot, thank you for digging in to the details here!

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
```diff
@@ -0,0 +1,62 @@
...
+    val clientId: String,
+    val scopes: Array<String>,
+    val state: String,
+    val accessType: String = "online",
+    val keyParams: AuthorizationKeyParams? = null
+) {
+    fun intoMessage(): MsgTypes.AuthorizationParams {
+        var b = MsgTypes.AuthorizationParams.newBuilder().setClientId(this.clientId)
```
rfk, line 15: nano-nit: I almost missed the call to `setClientId` here, I wonder if it's worth putting it on a newline for easier visibility (unless this is auto-formatted for kotlin guidelines, in which case no big deal).

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
```diff
@@ -0,0 +1,62 @@
...
+                .setScope(this.scopes.joinToString(" "))
+                .setState(this.state)
+                .setAccessType(this.accessType)
+        if (this.keyParams != null)
+            b = b.setKeyParams(this.keyParams.intoMessage())
+        return b.build()
+    }
+
+    // Auto Generated equals and hashCode
+    // Since AuthorizationParams holds an array of scopes
```
rfk, line 25: nit: I don't understand what this comment is trying to tell me, because it says "Auto generated" but you're adding the implementation by hand.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
```diff
@@ -56,7 +56,7 @@ internal interface LibFxAFFI : Library {
     fun fxa_get_access_token(fxa: FxaHandle, scope: String, ttl: Long, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_get_session_token(fxa: FxaHandle, e: RustError.ByReference): Pointer?
     fun fxa_get_current_device_id(fxa: FxaHandle, e: RustError.ByReference): Pointer?
-    fun fxa_authorize_auth_code(fxa: FxaHandle, clientId: String, scope: String, state: String, accessType: String, e: RustError.ByReference): Pointer?
+    fun fxa_authorize_auth_code(fxa: FxaHandle, len: Int, data: Pointer, e: RustError.ByReference): Pointer?
```
rfk, line 59: nano-nit: I believe most C APIs that pass a `(data, length)` pair tend to pass them in that order, so it took me a few moments to understand what you were doing here when I saw the `len` parameter. I think changing this to be `params_data: Pointer, params_len: Int` would be a low-cost way to make it easier for future readers to parse this API correctly.

(An alternative would be to pass a `RustBuffer` struct here, but that's probably a more invasive change).

components/fxa-client/ffi/src/lib.rs
```diff
@@ -447,21 +448,16 @@ pub extern "C" fn fxa_get_devices(
...
 #[no_mangle]
-pub extern "C" fn fxa_authorize_auth_code(
+pub unsafe extern "C" fn fxa_authorize_auth_code(
     handle: u64,
-    client_id: FfiStr<'_>,
-    scope: FfiStr<'_>,
-    state: FfiStr<'_>,
-    access_type: FfiStr<'_>,
+    auth_args_len: i32,
+    auth_args: *const u8,
```
rfk, line 454: nano-nit: ditto here re: ordering of these args, although the names here are 👍

components/fxa-client/src/ffi.rs
```diff
@@ -276,6 +277,38 @@ unsafe fn get_buffer<'a>(data: *const u8, len: i32) -> &'a [u8] {
...
+            code_challenge: proto_key_params.code_challenge,
+            code_challenge_method: proto_key_params.code_challenge_method,
+            keys_jwk: proto_key_params.keys_jwk,
+        }
+    }
+}
+
+impl AuthorizationParameters {
+    /// # Safety
+    /// Deref pointer thus unsafe
```
rfk, line 304: Heh, I know you copied this from elsewhere in our code, but I laugh when I see these comments. IIUC clippy lints on these to remind us to document the things that callers of the function need to do to ensure safety, but our comments are just like "Yep, it's unsafe! Got some pointers in here!".

(I'm not asking you to do anything about it, just noting my amusement 😄)

testing/sync-test/src/auth.rs
```diff
@@ -226,14 +262,23 @@ impl TestClient {
...
-        let query_params = final_url
-            .query_pairs()
-            .into_owned()
-            .collect::<HashMap<String, String>>();
+
+        // We either authenticate using the normal oauth_flow
+        // Or we use a pairing flow with a logged in account
+        // Both should work fine in executing the oauth flow
+        let (code, state) = if rand::random() {
+            acct.execute_oauth_pair_flow(&oauth_uri)?
```
rfk, line 270: It took me a while to understand what was happening here, because you would not ordinarily call `begin_oauth_flow` if you were doing a pairing flow. To make it clearer, I suggest reworking it like so:

```
let (code, state) = if rand::random() {
  let pairing_uri = fxa.begin_pairing_flow(...)
  acct.execute_oauth_pair_flow(&pairing_uri)
} else {
  let oauth_uri = fxa.begin_oauth_flow(...)
  ... the normal oauth stuff here ...
}
```

testing/sync-test/src/auth.rs
```diff
@@ -165,19 +165,55 @@ impl TestAccount {
...
+        let state = query_map.get("state").unwrap();
+        let code_challenge = query_map.get("code_challenge").unwrap();
+        let code_challenge_method = query_map.get("code_challenge_method").unwrap();
+        let scoped_keys = auth::get_scoped_keys(
+            scope,
+            client_id,
+            &auth::derive_auth_key_from_session_token(&self.session_token)?,
+            &self.cfg,
+            (&self.k_sync, &self.xcs),
+        )?;
```
rfk, line 199: Wow there is a lot of unpacking code here! Not a blocker, but I wonder if this would be easier to follow if you implemented `From<Url>` for `AuthorizationParameters` in order to factor out the unpacking into a separate code unit.

components/fxa-client/src/oauth.rs
```diff
@@ -167,25 +166,49 @@ impl FirefoxAccount {
...
+                .collect();
+            let data = serde_json::to_string(&data)?;
+            let client_keys_flow = ScopedKeysFlow::with_random_key()?;
+            (
+                Some(key_params.code_challenge),
+                Some(key_params.code_challenge_method),
+                Some(client_keys_flow.encrypt_keys_jwe(&keys_jwk, data.as_bytes())?),
+            )
+        } else {
+            (None, None, None)
```
rfk, line 196: It's going to be a lot of churn in your PR, but I think one benefit of separating the PKCE parms from the JWK will be that it simplifies the control-flow code here.

components/fxa-client/src/oauth.rs
```diff
@@ -167,25 +166,49 @@ impl FirefoxAccount {
...
+            let requested_scopes: HashSet<&str> = auth_params.scope.split_whitespace().collect();
+            let data: HashMap<String, ScopedKey> = self
+                .state
+                .scoped_keys
+                .iter()
+                .filter(|(scope, _)| requested_scopes.contains(&(*scope).as_ref()))
+                .map(|(key, value)| (key.clone(), value.clone()))
+                .collect();
+            let data = serde_json::to_string(&data)?;
+            let client_keys_flow = ScopedKeysFlow::with_random_key()?;
```
rfk, line 189: Hrm, this abstraction of having to create a whole "flow" just to encrypt the keys is a bit awkward, but we can live with it for now.

components/fxa-client/src/oauth.rs
```diff
@@ -167,25 +166,49 @@ impl FirefoxAccount {
...
+            auth_params.key_params
+        {
+            let keys_jwk = base64::decode_config(key_params.keys_jwk, base64::URL_SAFE_NO_PAD)?;
+            let keys_jwk = String::from_utf8(keys_jwk)?;
+            let requested_scopes: HashSet<&str> = auth_params.scope.split_whitespace().collect();
+            let data: HashMap<String, ScopedKey> = self
+                .state
+                .scoped_keys
+                .iter()
+                .filter(|(scope, _)| requested_scopes.contains(&(*scope).as_ref()))
```
rfk, line 185: Oh! This made we realize something which I don't think we've thought about explicitly yet.

There are two important questions to consider here when handing out keys to another client:
1) Do we have all the keys that the client requested?
2) Is that client actually *allowed* to request those keys?

The corresponding function in Desktop is [here](https://searchfox.org/mozilla-central/rev/31d8600b73dc85b4cdbabf45ac3f1a9c11700d8e/services/fxaccounts/FxAccountsKeys.jsm#224) and it has two parts:

1) The call to `getScopedKeyData`, which asks the FxA server what keys the given `clientId` is actually allowed to request.
2) The call to `getKeyForScope` which retrieves the key from our local store so it can be returned to the client.

I think we can defer this to a follow-up bug, but could you please add a "TODO" comment here to note that we need this extra check, and file a follow-up bug for next sprint?

(The reason I'm asking for you to file the bug, is so I can check whether what I said above made sufficient sense to enable you to file it, so please ask clarifying questions if necessary 😁)

CHANGES_UNRELEASED.md
rfk, line 11: You mention the iOS functions in this changelog entry, but I don't see any change to them in the PR itself. Do they need to be updated here as well?

tarikeshaq: Hmmm, the ios mention is from the entrypoint change #3265. Not sure why github thinks I changed it 🤷

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AuthorizationParams.kt
tarikeshaq: Ooo That makes tons of sense. Sorry for all the confusing attempts here 😅

rfk: > Sorry for all the confusing attempts here

FWIW it's really useful for me, to help me understand all the things that could be clearer, so I really appreciate your efforts to work through this!

tarikeshaq:
Thanks a TON, @rfk  for all the comments, I **REALLY** appreciate the time you put in for the reviews!

And please, do keep the nits coming (if you have the time 😅 ) Your comments force me to think about code readability and how it all fits in the large scheme of things and I love them! I'm here to learn so I'm expecting to make mistakes, and by receiving feedback, I'm learning! (So that I hopefully don't repeat them 🙃)

CHANGES_UNRELEASED.md
```diff
@@ -8,4 +8,5 @@
...
\ No newline at end of file
+- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
+- Changes the `authorizeOAuthCode` API to now accept an `AuthorizationParams` object instead of the individual parameters. The `AuthorizationParams` also includes optional `AuthorizationKeyParams` that contain the `codeChallenge`, `codeChallengeMethod` and `keysJwk` needed to request scoped keys. ([#3264](https://github.com/mozilla/application-services/pull/3264))
```
rfk, line 12: nit: this changelog entry needs to be updated for the new `AuthorizationPKCEParams`

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/LibFxAFFI.kt
```diff
@@ -56,7 +56,7 @@ internal interface LibFxAFFI : Library {
     fun fxa_get_access_token(fxa: FxaHandle, scope: String, ttl: Long, e: RustError.ByReference): RustBuffer.ByValue
     fun fxa_get_session_token(fxa: FxaHandle, e: RustError.ByReference): Pointer?
     fun fxa_get_current_device_id(fxa: FxaHandle, e: RustError.ByReference): Pointer?
-    fun fxa_authorize_auth_code(fxa: FxaHandle, clientId: String, scope: String, state: String, accessType: String, e: RustError.ByReference): Pointer?
+    fun fxa_authorize_auth_code(fxa: FxaHandle, params_data: Pointer, params_len: Int, e: RustError.ByReference): Pointer?
```
rfk, line 59: pico-nit, which I should have added last time: it probably makes sense for the names here to match the ones you use in the rust definition, so "auth_params" and "auth_params_len".

components/fxa-client/src/oauth.rs
tarikeshaq: Added the TODO in the code, and before I file the issue, lemme summarize (for my own sanity):
- Right now (in this pr) I simply run around the state and get any keys that fit the requested scopes
- However, the client calling us, may not be allowed to request those scopes, so, we have to check with the server first, and it would tell us which scopes the client is allowed to request.
- Then, for each of those scopes, we look internally and we basically do what we have here (i.e the whole keys_jwe encryption business)

One last crazy question, if we (as in the fxa-client) do not have the requested (and validated) scoped keys in our state, I'm guessing we do the whole generation using HKDF on our `kB` and the scope (or do we just error out)?

components/fxa-client/src/oauth.rs
```diff
@@ -363,6 +389,75 @@ impl FirefoxAccount {
...
+pub struct AuthorizationParameters {
+    pub client_id: String,
+    pub scope: String,
+    pub state: String,
+    pub access_type: String,
+    pub pkce_params: Option<AuthorizationPKCEParams>,
+    pub keys_jwk: Option<String>,
+}
+
+// Not impl From<url::Url> because we could get an error
```
rfk, line 422: Oh yeah, errors! I suggest implementing [`TryFrom`](https://doc.rust-lang.org/nightly/core/convert/trait.TryFrom.html) instead of leaving this comment.

testing/sync-test/src/auth.rs
```diff
@@ -165,19 +165,39 @@ impl TestAccount {
...
+            keys_jwe: Some(keys_jwe),
+            state: state.clone(),
+            access_type: "offline".to_string(),
+        };
+        auth::send_authorization_request(&self.cfg, auth_params, &auth_key)
+    }
+
+    fn execute_oauth_pair_flow(&self, oauth_uri: &str) -> Result<(String, String)> {
+        let url = Url::parse(&oauth_uri)?;
+        let auth_params = auth::AuthorizationParameters::from(url)?;
```
rfk, line 186: Thanks for trying out the use of `From` here, I think it helped make this method more accessible.

components/fxa-client/src/oauth.rs
tarikeshaq: TIL about `TryFrom` 😃

components/fxa-client/src/oauth.rs
rfk: > However, the client calling us

Re-phrasing this just in an attempt to be super clear: "the client from which we received a pairing request", which in practice is going to be a phone somewhere that scanned a QR code we were showing, and about which we know absolutely nothing other than what information is provided in the AuthorizationParameters.

> may not be allowed to request those scopes, so, we have to check with the server first, and it would tell us
> which scopes the client is allowed to request.

Correct. We'll provide the requested scopes to the server, and the server will respond with a filtered subset of those scopes (possibly empty!) along with the data required to generate the derived keys:

  * https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/oauth/api.md#response-11

We already have the derived keys so we don't need to "generate" them as such, but we can use this as permission check on whether that client is supposed to be requesting those keys. (Of course, the client could *lie* about who it is, e.g. a malicious app could tell us that it's Fenix by providing the Fenix `client_id`. But this check serves as a defense-in-depth measure so that an attacker who e.g. compromised the Firefox Send oauth flow, couldn't use it to request the sync keys).

> Then, for each of those scopes, we look internally and we basically do what we have here 

Right. We'd basically filter the current `requested_scopes` into an `allowed_scopes` and then do the same thing we have here already.

> if we (as in the fxa-client) do not have the requested (and validated) scoped keys in our state,
> I'm guessing we do the whole generation using HKDF on our kB and the scope (or do we just error out)?

If we have `kB`, we could do this, and maybe it would be useful for tests. But in practice we don't persist `kB` so we won't be able to do it in real applications. The right thing here is probably to return an error, since we cannot fulfill the authorization request as given.

components/fxa-client/src/oauth.rs
rfk: Honestly, it's my new favourite thing!

rfk:
(not merging in case @vladikoff had any further comments)

components/fxa-client/src/oauth.rs
```diff
@@ -167,25 +167,52 @@ impl FirefoxAccount {
     /// Fetch an OAuth code for a particular client using a session token from the account state.
     /// This method doesn't support OAuth public clients at this time.
     ///
-    /// * `client_id` - OAuth client id.
```
vladikoff, line 170: could we move this documentation either to `AuthorizationRequestParameters` or below under the `/// * `auth_params` Authentication parameters  ` , and possibly checking if they show up in Rustdoc?

components/fxa-client/ffi/src/lib.rs
```diff
@@ -447,21 +448,16 @@ pub extern "C" fn fxa_get_devices(
...
     error: &mut ExternError,
 ) -> *mut c_char {
     log::debug!("fxa_authorize_auth_code");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
-        let client_id = client_id.as_str();
-        let scope = scope.as_str();
-        let state = state.as_str();
-        let access_type = access_type.as_str();
-        fxa.authorize_code_using_session_token(client_id, scope, state, access_type)
+        let auth_params = AuthorizationParameters::from_protobuf_ptr(auth_params, auth_params_len)?;
```
vladikoff, line 459: @tarikeshaq Great work figuring out the process for this 👍  🎉

components/fxa-client/src/oauth.rs
tarikeshaq: Alrighty done, They show up as a nested bulliet list in rus doc now. Also removed the bit about "not supporting public clients" since we added the PKCE params

PR #3265, opened 2020-06-22
Adds entrypoint support in beginOauthFlow [ci full]

### LINKED ISSUE #3263: Support the `entrypoint` param when starting a new OAuth flow
This might be a duplicate. We need to support an `entrypoint` parameter for `beginOAuthFlow` and `beginPairingFlow`. This is required for Firefox iOS to be able to provide proper metrics to the FxA page: https://github.com/mozilla-mobile/firefox-ios/search?q=entrypoint&unscoped_q=entrypoint

Entrypoint is already supported in `getManageAccountURL` and `getManageDevicesURL`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1412)
┆Story Points: 3
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-07-03

tarikeshaq:

**This PR changes public APIs**
- Adds entrypoint to begin_oauth_flow and begin_pairing_flow
- Modifies the ffi and the Kotlin and Swift bindings

Since this has breaking changes to the APIs, I imagine I would need to open an issue/PR in andriod-components. firefox-ios already has an issue opened [mozilla-mobile/firefox-ios#6824](https://github.com/mozilla-mobile/firefox-ios/issues/6824)

But before I do that, I wanted to confirm that I'm on the right track and that in fact, the way to do this is to introduce breaking changes. 

fixes #3263

### DIFF
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
@@ -99,12 +99,13 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      * This performs network requests, and should not be used on the main thread.
      *
      * @param scopes List of OAuth scopes for which the client wants access
+     * @param entrypoint to be used for metrics
      * @return String that resolves to the flow URL when complete
      */
-    fun beginOAuthFlow(scopes: Array<String>): String {
+    fun beginOAuthFlow(scopes: Array<String>, entrypoint: String): String {
         val scope = scopes.joinToString(" ")
         return rustCallWithLock { e ->
-            LibFxAFFI.INSTANCE.fxa_begin_oauth_flow(this.handle.get(), scope, e)
+            LibFxAFFI.INSTANCE.fxa_begin_oauth_flow(this.handle.get(), scope, entrypoint, e)
         }.getAndConsumeRustString()
     }
 
@@ -112,11 +113,16 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      * Begins the pairing flow.
      *
      * This performs network requests, and should not be used on the main thread.
+     *
+     * @param pairingUrl the url to initilaize the paring flow with
+     * @param scopes List of OAuth scopes for which the client wants access
+     * @param entrypoint to be used for data collection
+     * @return String that resoles to the flow URL when complete
      */
-    fun beginPairingFlow(pairingUrl: String, scopes: Array<String>): String {
+    fun beginPairingFlow(pairingUrl: String, scopes: Array<String>, entrypoint: String): String {
         val scope = scopes.joinToString(" ")
         return rustCallWithLock { e ->
-            LibFxAFFI.INSTANCE.fxa_begin_pairing_flow(this.handle.get(), pairingUrl, scope, e)
+            LibFxAFFI.INSTANCE.fxa_begin_pairing_flow(this.handle.get(), pairingUrl, scope, entrypoint, e)
         }.getAndConsumeRustString()
     }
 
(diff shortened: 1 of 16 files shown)


components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -99,24 +99,30 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
...
     }
 
     /**
      * Begins the pairing flow.
      *
      * This performs network requests, and should not be used on the main thread.
+     *
+     * @param pairingUrl the url to initilaize the paring flow with
+     * @param scopes List of OAuth scopes for which the client wants access
+     * @param entrypoint to be used for data collection
```
vladikoff, line 119: - Let's make this an optional parameter
- change `data collection` to `metrics`

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -99,24 +99,30 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      * This performs network requests, and should not be used on the main thread.
      *
      * @param scopes List of OAuth scopes for which the client wants access
+     * @param entrypoint to be used for data collection
      * @return String that resolves to the flow URL when complete
      */
-    fun beginOAuthFlow(scopes: Array<String>): String {
+    fun beginOAuthFlow(scopes: Array<String>, entrypoint: String): String {
```
vladikoff, line 105: As I mentioned in Slack I am conflicted about making `entrypoint` optional VS enforcing consumers to pass something in here. Probably leaning towards allowing to skip it and having our logic pass in `none` or `notset` as part of the slow.

megazords/ios/MozillaAppServicesTests/FxAccountManagerTests.swift
```diff
@@ -243,7 +243,7 @@ class FxAccountManagerTests: XCTestCase {
         let beginAuthDone = expectation(description: "beginAuthDone")
         var authURL: String?
         mgr.initialize { _ in
-            mgr.beginAuthentication { url in
+            mgr.beginAuthentication(entrypoint: "testNewAccountLogIn") { url in
```
vladikoff, line 246: can these be snake case? I don't think we want to show a bad example where entrypoint format is camelCase

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
vladikoff: However our other methods already force users to set `entrypoint`, e.g in fxa_get_manage_account_url

vladikoff: In addition we need to support utm parameters here and flow id parameters here if we want parity with the desktop implementation.

Params such as 

```
    // {String} flow_id - Flow Id, such as '5445b28b8b7ba6cf71e345f8fff4bc59b2a514f78f3e2cc99b696449427fd445'
      // {Number} flow_begin_time - Flow begin timestamp, such as 1590780440325
      // {String} device_id - Device Id, such as '7e450f3337d3479b8582ea1c9bb5ba6c'
```

vladikoff: So full support for all parameters will look like, (i hope I got them all):

```
beginOAuthFlow(scopes, [entrypoint], [flow_id], [flow_begin_time], [device_id], [utm_source], [utm_content], [utm_medium], [utm_term], [utm_campaign]
```

vladikoff: After chatting with @rfk, let's split out the optional parameters in a different issue and just fix the `entrypoint` here. I filed that issue for optional parameters here: https://github.com/mozilla/application-services/issues/3278

vladikoff: And we are going ahead with making `entrypoint` a REQUIRED argument for consumers.

Issue #3254, opened 2020-06-18
Don't log expected networking failures at "error" level


rfk:
Sentry has a nice collection of reports from viaduct about things that are expected, inevitable networking-related failures, such as these [network](https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/8681110/?query=is%3Aunresolved) [timeouts](https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/8678590/?query=is%3Aunresolved). It's reasonable for us to raise an error back to the application here, but I don't think we should be logging them to sentry.

Java has a log level above error (`FATAL`) and I've seen other languages have similar (e.g. `CRITICAL`) but I don't think rust has anything higher than `ERROR` level. Proposal: log things that are "we failed to do what you asked, for totally understandable reasons" at `WARNING` level, and keep `ERROR` level for things that are bad enough they should show up in sentry, in error logs from developers, etc.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1404)
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-07-17

rfk:
(I'm partly filing this bug to try out the process of reviewing sentry during triage rotation)

thomcc:
An issue with this is that a lot of our error logging comes from automatically logging `Result::Err`s in the conversion. This isn't merely laziness, but rust makes it pretty unergonomic to log errors any other way, as you need to match on the error, and (apparently) filter out a specific subtype of error (which also Rust doesn't have good ergonomics for)...

So I don't think I disagree here, it's just easier said than done.

eoger:
From meeting, we can:
- For network errors, add a `[NETWORK-ERROR]` or `[no-sentry]` token somewhere in the error message
Then either:
- In `RustLog` (preferable since it applies to iOS too), transform the error to a warn if we match the token.
- In a-c: if we match the token, don't send a sentry error.

eoger:
Fixed by https://github.com/mozilla/application-services/pull/3273.

PR #3273, opened 2020-06-24
Downgrade network errors into warnings when passing through rc_log

thomcc:
Haven't actually tested this but it's the idea eoger sketched out in https://github.com/mozilla/application-services/issues/3254#issuecomment-648410952.

I felt compelled to do it since I have some opinions on it, and said that it was going to be a problem, which clearly ended up not being the case. Anyway, getting it done now means maybe it can be in time for the fenix release, which it could be problematic for...

The tests are superficial but at least it has some. In principal we could test this on either the iOS or the Android tests by creating a log at error level that contains the string [no-sentry], and ensuring that its actual log level comes through as a warning. This is somewhat tricky, given that both sets of logging test infrastructure aren't set up in a way to allow this sort of testing...

As a result, superficial tests. 

### DIFF
--- a/components/viaduct/src/error.rs
+++ b/components/viaduct/src/error.rs
@@ -4,13 +4,13 @@
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("Illegal characters in request header '{0}'")]
+    #[error("[no-sentry] Illegal characters in request header '{0}'")]
     RequestHeaderError(crate::HeaderName),
 
-    #[error("Backend error: {0}")]
+    #[error("[no-sentry] Backend error: {0}")]
     BackendError(String),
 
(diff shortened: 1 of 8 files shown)


components/viaduct/src/error.rs
```diff
@@ -4,27 +4,27 @@
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("Illegal characters in request header '{0}'")]
+    #[error("[no-sentry] Illegal characters in request header '{0}'")]
```
thomcc, line 7: Honestly not all of these need to be no-sentry.

I'm not sure which we're seeing in sentry. I guess we might want reporting on anything other than BackendError and NetworkError?

That said, the safe thing seemed to be to silence all of them

PR #3274, opened 2020-06-24
Remove some network error logging

eoger:
This removes most of the remaining network error logging from sentry.

### DIFF
--- a/components/support/viaduct-reqwest/src/lib.rs
+++ b/components/support/viaduct-reqwest/src/lib.rs
@@ -65,10 +65,9 @@ impl Backend for ReqwestBackend {
         viaduct::note_backend("reqwest (untrusted)");
         let request_method = request.method;
         let req = into_reqwest(request)?;
-        let mut resp = CLIENT.execute(req).map_err(|e| {
-            log::error!("Reqwest error: {:?}", e);
-            viaduct::Error::NetworkError(e.to_string())
-        })?;
+        let mut resp = CLIENT
+            .execute(req)
+            .map_err(|e| viaduct::Error::NetworkError(e.to_string()))?;
         let status = resp.status().as_u16();
         let url = resp.url().clone();
         let mut body = Vec::with_capacity(resp.content_length().unwrap_or_default() as usize);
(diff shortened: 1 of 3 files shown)


components/support/viaduct-reqwest/src/lib.rs
```diff
@@ -65,10 +65,9 @@ impl Backend for ReqwestBackend {
         viaduct::note_backend("reqwest (untrusted)");
         let request_method = request.method;
         let req = into_reqwest(request)?;
-        let mut resp = CLIENT.execute(req).map_err(|e| {
-            log::error!("Reqwest error: {:?}", e);
```
eoger, line 69: This line is responsible for 100% of the sentry network error logging on iOS.

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -91,8 +91,7 @@ object RustHttpConfig {
                     }
                     rb
                 } catch (e: Throwable) {
-                    LibViaduct.INSTANCE.viaduct_log_error("Network error: ${e.message}")
```
eoger, line 94: Same as iOS, 100% of network errors on Sentry are logged here.

components/viaduct/src/backend/ffi.rs
```diff
@@ -64,11 +64,6 @@ impl Backend for FfiBackend {
         };
 
         if let Some(exn) = response.exception_message {
-            log::error!(
-                // Well, we caught *something* java wanted to tell us about, anyway.
-                "Caught network error (presumably). Message: {:?}",
-                exn
-            );
```
eoger, line 71: Well because of the protobuf bug I uncovered this line was never called, but it will.

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -91,8 +91,7 @@ object RustHttpConfig {
...
-                    MsgTypes.Response.newBuilder().setExceptionMessage(e.message)
+                    MsgTypes.Response.newBuilder().setExceptionMessage("Network error: ${e.message}")
```
eoger, line 94: This fixes the bug I described in https://mozilla.slack.com/archives/CBBH6GU91/p1593009194368100 and ends up in `Missing HTTP status` error in Sentry.

thomcc:
Yeah, okay, I follow what's going on now, the `Throwable` we catch in Kotlin happens to have a `null` message. This will work  but seems unfortunate. We should consider logging it's name instead in the future in this case -- (I'm pretty sure I have some code somewhere in the tree that logs exception names).

Very strange that it's null, feels like an upstream bug even.

PR #3297, opened 2020-06-29
Cut release v61.0.7

thomcc:
This is both to get the synced-tabs JSON changes done as well as to check that releases from the `main`  branch work okay


thomcc:
Going to let this run merge by checkin-needed to kick the tires a bit more, even though we don't usually do that for release PRs (since we need to immediately cut a release anyway).

thomcc:
Well, I'm just going to merge rather than wait for #3299 and another rebase

PR #3298, opened 2020-06-29
Fix mergify config, hopefully

thomcc:


thomcc:
I believe `Rust benchmarks` and `Sync integration tests` are all under the `run-tests` umbrella.

thomcc:
> checkin-needed won't work in a PR that modifies the .mergify.yml file, but LGTM otherwise!

😩

PR #3302, opened 2020-06-30
WIP for collecting send-tab telemetry.

mhammond:
This is a WIP just to get feedback on the shape of this.

The main thing I'm trying to avoid here is to have the telemetry "pollute" the
public API - the public API should not be aware where and when we send
telemetry.

In this case, the "public API" in question is send-tab, where we can probably
suck up a public API change, but we shouldn't!

The whole "what telemetry API should we use/when can we use glean everywhere"
question is a good one, but one we aren't going resolve here. An ideal world
probably has our rust components directly using glean - in which case the
telemetry data also doesn't pollute the other public APIs - so let's try and
get a little closer to that world.

So the summary of how this patch hangs together is:

* There's a new `FxaTelemetry` struct which is a grab-bag of all the telemetry
  we collect (which isn't much - even desktop today doesn't gather much). This
  struct is stored in a `RefCell<>` inside `FirefoxAccount`.

* There's a new `fxa_gather_telemetry()` function in the FFI. The idea is that
  android-components *telemetry* code will call this, and translate what it
  gets back into the glean calls it needs to make (ie, android-components will
  also add a couple of new events to `sync-telemetry/metrics.yaml` - note that
  the *send-tab* code doesn't really get involved here.

* Except that the *send-tab* code will need to tell the telemetry code that
  there's probably telemetry to gather. So while it's not magic or fully
  automatic, most of the responsibilities are in the right place and there's
  only a few leaky abstractions.

* There will obviously need to be a corresponding android-components PR

Assuming the "big picture" of this is OK, some "small picture" issues:

* I'm not sure if it's OK that `fxa_gather_telemetry()` returns a JSON string.
  The idea is that it should ideally be dynamic - stuff it doesn't understand
  should be ignored. Eg, you can imagine a world where some things are only
  understood/recorded by, say, iOS. Adding new telemetry shouldn't be a breaking
  change to platforms that don't want to explicitly record it.

* No tests!

* Um - I'm sure I had more than that :) I'm sure you'll find some in the patch.

### DIFF
--- a/components/fxa-client/src/commands/send_tab.rs
+++ b/components/fxa-client/src/commands/send_tab.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use crate::telemetry::SentReceivedTab;
 /// The Send Tab functionality is backed by Firefox Accounts device commands.
 /// A device shows it can handle "Send Tab" commands by advertising the "open-uri"
 /// command in its on own device record.
@@ -18,6 +19,7 @@ use rc_crypto::ece::{self, Aes128GcmEceWebPush, EcKeyComponents, WebPushParams};
 use rc_crypto::ece_crypto::{RcCryptoLocalKeyPair, RcCryptoRemotePublicKey};
 use serde_derive::*;
 use sync15::{EncryptedPayload, KeyBundle};
+use sync_guid::Guid;
 
 pub const COMMAND_NAME: &str = "https://identity.mozilla.com/cmd/open-uri";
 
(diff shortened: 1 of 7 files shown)


components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +42,31 @@ impl EncryptedSendTabPayload {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SendTabPayload {
     pub entries: Vec<TabHistoryEntry>,
+    // XXX - tests don't fail yet even though none add these 2 new items and
+    // they aren't marked as optional for serde, which seems odd!
```
linacambridge, line 46: I marked these as optional in case we're dealing with an old Firefox that sends the `streamID` unencrypted, or a tab coming from pre-79 Firefox—so we'll still show the tab, even if we can't record telemetry for it. WDYT?

components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +42,31 @@ impl EncryptedSendTabPayload {
...
-        SendTabPayload {
-            entries: vec![TabHistoryEntry {
-                title: title.to_string(),
-                url: url.to_string(),
-            }],
-        }
+    pub fn single_tab(title: &str, url: &str) -> (Self, SentReceivedTab) {
+        let sent_telemetry = SentReceivedTab {
+            flow_id: Guid::random().to_string(),
+            stream_id: Guid::random().to_string(),
```
linacambridge, line 57: I used `util::random_base64_url_string`. Almost the same thing, just avoids pulling in another dependency.

components/fxa-client/src/send_tab.rs
```diff
@@ -38,16 +40,21 @@ impl FirefoxAccount {
     }
 
     /// Send a single tab to another device designated by its device ID.
+    /// XXX - this should change to taking a Vec of device IDs, because
+    /// collecting telemetry relies on a multi-device send happening in a single
+    /// operation.
     pub fn send_tab(&mut self, target_device_id: &str, title: &str, url: &str) -> Result<()> {
```
linacambridge, line 46: I wonder, how do you feel about this method returning a `Result<SentReceivedTab>`—maybe rename it `SendTabTelemetry`—instead of indirecting through `FxaTelemetry`? I was thinking it would be easier for the caller to get the IDs back right away, which it can then either stash in the Sync ping (iOS) or Glean (Fenix), instead of sending the tab, then turning around and calling `fxa_gather_telemetry` right away.

linacambridge: IOW, I guess I'm wondering, why shouldn't we pollute the API with telemetry? 😄 A return value means the caller has to handle it, but collecting it separately means the caller has to turn around and call `gather_telemetry` immediately after sending the tab or handling an incoming "tab received" command—and forgetting to do that means we drop the telemetry on the floor.

mhammond:
Superseded by #3308

PR #3308, opened 2020-07-01
WIP for collecting send-tab telemetry.

mhammond:
[Lina and I chatted about this and decided to collaborate - I can't work out how to change the branch in the other PR, so here we are...]

This is a WIP just to get feedback on the shape of this.

The main thing I'm trying to avoid here is to have the telemetry "pollute" the
public API - the public API should not be aware where and when we send
telemetry.

In this case, the "public API" in question is send-tab, where we can probably
suck up a public API change, but we shouldn't!

The whole "what telemetry API should we use/when can we use glean everywhere"
question is a good one, but one we aren't going resolve here. An ideal world
probably has our rust components directly using glean - in which case the
telemetry data also doesn't pollute the other public APIs - so let's try and
get a little closer to that world.

So the summuary of how this patch hangs together is:

* There's a new `FxaTelemetry` struct which is a grab-bag of all the telemetry
  we collect (which isn't much - even desktop today doesn't gather much). This
  struct is stored in a `RefCell<>` inside `FirefoxAccount`.

* There's a new `fxa_gather_telemetry()` function in the FFI. The idea is that
  android-components *telemetry* code will call this, and translate what it
  gets back into the glean calls it needs to make (ie, android-components will
  also add a couple of new events to `sync-telemetry/metrics.yaml` - note that
  the *send-tab* code doesn't really get involved here.

* Except that the *send-tab* code will need to tell the telemetry code that
  there's probably telemetry to gather. So while it's not magic or fully
  automatic, most of the responsibilities are in the right place and there's
  only a few leaky abstractions.

Assuming the "big picture" of this is OK, some "small picture" issues:

* I'm not sure if it's OK that `fxa_gather_telemetry()` returns a JSON string.
  The idea is that it should ideally be dynamic - stuff it doesn't understand
  should be ignored. Eg, you can imagine a world where some things are only
  understood/recorded by, say, iOS. Adding new telemetry shouldn't be a breaking
  change to platforms that don't want to explicitly record it.

* No tests!

* Um - I'm sure I had more than that :) I'm sure you'll find some in the patch.

### DIFF
--- a/components/fxa-client/ffi/src/lib.rs
+++ b/components/fxa-client/ffi/src/lib.rs
@@ -12,7 +12,7 @@ use ffi_support::{
     ConcurrentHandleMap, ExternError, FfiStr,
 };
 use fxa_client::{
-    device::{Capability as DeviceCapability, PushSubscription},
+    device::{Capability as DeviceCapability, CommandFetchReason, PushSubscription},
     ffi::{from_protobuf_ptr, AuthorizationParameters, MetricsParams},
     migrator::MigrationState,
     msg_types, FirefoxAccount,
@@ -506,10 +506,11 @@ pub extern "C" fn fxa_handle_session_token_change(
 pub extern "C" fn fxa_poll_device_commands(handle: u64, error: &mut ExternError) -> ByteBuffer {
     log::debug!("fxa_poll_device_commands");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
-        fxa.poll_device_commands().map(|cmds| {
-            let commands = cmds.into_iter().map(|e| e.into()).collect();
-            fxa_client::msg_types::IncomingDeviceCommands { commands }
-        })
+        fxa.poll_device_commands(CommandFetchReason::Poll)
+            .map(|cmds| {
+                let commands = cmds.into_iter().map(|e| e.into()).collect();
+                fxa_client::msg_types::IncomingDeviceCommands { commands }
+            })
     })
 }
 
@@ -615,3 +616,10 @@ pub extern "C" fn fxa_get_ecosystem_anon_id(handle: u64, error: &mut ExternError
 define_handle_map_deleter!(ACCOUNTS, fxa_free);
 define_string_destructor!(fxa_str_free);
 define_bytebuffer_destructor!(fxa_bytebuffer_free);
+
+/// Gather telemetry collected by FxA.
+#[no_mangle]
+pub extern "C" fn fxa_gather_telemetry(handle: u64, error: &mut ExternError) -> *mut c_char {
+    log::debug!("fxa_gather_telemetry");
+    ACCOUNTS.call_with_result_mut(error, handle, |fxa| fxa.gather_telemetry())
+}
--- a/components/fxa-client/src/commands/send_tab.rs
+++ b/components/fxa-client/src/commands/send_tab.rs
@@ -13,7 +13,7 @@
 /// uses the obtained public key to encrypt the `SendTabPayload` it created that
 /// contains the tab to send and finally forms the `EncryptedSendTabPayload` that is
 /// then sent to the target device.
-use crate::{device::Device, error::*, scoped_keys::ScopedKey, scopes};
+use crate::{device::Device, error::*, scoped_keys::ScopedKey, scopes, telemetry};
 use rc_crypto::ece::{self, Aes128GcmEceWebPush, EcKeyComponents, WebPushParams};
 use rc_crypto::ece_crypto::{RcCryptoLocalKeyPair, RcCryptoRemotePublicKey};
 use serde_derive::*;
@@ -40,16 +40,26 @@ impl EncryptedSendTabPayload {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SendTabPayload {
     pub entries: Vec<TabHistoryEntry>,
+    #[serde(rename = "flowID", default)]
+    pub flow_id: String,
+    #[serde(rename = "streamID", default)]
+    pub stream_id: String,
 }
 
 impl SendTabPayload {
-    pub fn single_tab(title: &str, url: &str) -> Self {
-        SendTabPayload {
-            entries: vec![TabHistoryEntry {
-                title: title.to_string(),
-                url: url.to_string(),
-            }],
-        }
+    pub fn single_tab(title: &str, url: &str) -> (Self, telemetry::SentCommand) {
+        let sent_telemetry: telemetry::SentCommand = Default::default();
+        (
+            SendTabPayload {
+                entries: vec![TabHistoryEntry {
+                    title: title.to_string(),
+                    url: url.to_string(),
+                }],
+                flow_id: sent_telemetry.flow_id.clone(),
+                stream_id: sent_telemetry.stream_id.clone(),
+            },
+            sent_telemetry,
+        )
     }
     fn encrypt(&self, keys: PublicSendTabKeys) -> Result<EncryptedSendTabPayload> {
         rc_crypto::ensure_initialized();
@@ -217,4 +227,30 @@ fn extract_oldsync_key_components(oldsync_key: &ScopedKey) -> Result<(Vec<u8>, V
     let kxcs = base64::decode_config(&kxcs, base64::URL_SAFE_NO_PAD)?;
     let ksync = oldsync_key.key_bytes()?;
     Ok((ksync, kxcs))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_minimal_parse_payload() {
+        let minimal = r#"{ "entries": []}"#;
+        let payload: SendTabPayload = serde_json::from_str(minimal).expect("should work");
+        assert_eq!(payload.flow_id, "".to_string());
+    }
+
+    #[test]
+    fn test_payload() {
+        let (payload, telem) = SendTabPayload::single_tab("title", "http://example.com");
+        let json = serde_json::to_string(&payload).expect("should work");
+        assert_eq!(telem.flow_id.len(), 12);
+        assert_eq!(telem.stream_id.len(), 12);
+        assert_ne!(telem.flow_id, telem.stream_id);
+        let p2: SendTabPayload = serde_json::from_str(&json).expect("should work");
+        // no 'PartialEq' derived so check each field individually...
+        assert_eq!(payload.entries[0].url, "http://example.com".to_string());
+        assert_eq!(payload.flow_id, p2.flow_id);
+        assert_eq!(payload.stream_id, p2.stream_id);
+    }
 }
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -9,16 +9,24 @@ use crate::{
     commands,
     error::*,
     http_client::{
-        CommandData, DeviceUpdateRequest, DeviceUpdateRequestBuilder, PendingCommand,
-        UpdateDeviceResponse,
+        DeviceUpdateRequest, DeviceUpdateRequestBuilder, PendingCommand, UpdateDeviceResponse,
     },
-    util, CachedResponse, FirefoxAccount, IncomingDeviceCommand,
+    telemetry, util, CachedResponse, FirefoxAccount, IncomingDeviceCommand,
 };
 use serde_derive::*;
 use std::collections::{HashMap, HashSet};
 
 // An devices response is considered fresh for `DEVICES_FRESHNESS_THRESHOLD` ms.
 const DEVICES_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
+
+/// The reason we are fetching commands.
+#[derive(Clone, Copy)]
+pub enum CommandFetchReason {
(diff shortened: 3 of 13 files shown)


components/fxa-client/ffi/src/lib.rs
```diff
@@ -591,3 +591,10 @@ pub extern "C" fn fxa_send_tab(
 define_handle_map_deleter!(ACCOUNTS, fxa_free);
 define_string_destructor!(fxa_str_free);
 define_bytebuffer_destructor!(fxa_bytebuffer_free);
+
+/// Gather telemetry collected by FxA.
+#[no_mangle]
+pub extern "C" fn fxa_gather_telemetry(handle: u64, error: &mut ExternError) -> *mut c_char {
```
rfk, line 597: Returning a JSON bundle here seems OK to be, for the b/w-compat reasons you describe. It's kinda weird but if we do a good job of documenting it and setting expectations that this API will eventually go away in favour of glean, then 👍

components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +42,31 @@ impl EncryptedSendTabPayload {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SendTabPayload {
     pub entries: Vec<TabHistoryEntry>,
+    // XXX - tests don't fail yet even though none add these 2 new items and
+    // they aren't marked as optional for serde, which seems odd!
```
rfk, line 46: Do you intend to mark them optional before landing though? (I agree the lack of failures seems odd in the meantime)

components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +42,31 @@ impl EncryptedSendTabPayload {
...
 
 impl SendTabPayload {
-    pub fn single_tab(title: &str, url: &str) -> Self {
-        SendTabPayload {
-            entries: vec![TabHistoryEntry {
-                title: title.to_string(),
-                url: url.to_string(),
-            }],
-        }
+    pub fn single_tab(title: &str, url: &str) -> (Self, SentReceivedTab) {
```
rfk, line 54: From a readability pespective, I wonder if it would be better to refer to this new struct by qualified name, like `telemetry::SentReceivedTab`. This might make it more obvious to the reader what the extra return value is for.

components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +42,31 @@ impl EncryptedSendTabPayload {
...
+        let sent_telemetry = SentReceivedTab {
+            flow_id: Guid::random().to_string(),
+            stream_id: Guid::random().to_string(),
```
rfk, line 57: I wonder if this guid-creation should happen inside some sort of `SentReceivedTab` constructor, perhaps even via `Default::default()`. Probably not a big deal either way.

components/fxa-client/src/send_tab.rs
```diff
@@ -38,16 +40,21 @@ impl FirefoxAccount {
     }
 
     /// Send a single tab to another device designated by its device ID.
+    /// XXX - this should change to taking a Vec of device IDs, because
+    /// collecting telemetry relies on a multi-device send happening in a single
+    /// operation.
```
rfk, line 45: I agree, but also, that's a pretty significant breaking change. It makes me wonder if we should move to supporting a `Vec` of tab data as well for the general case! But then again, desktop doesn't support that and our telemetry doesn't really make that distinction either.

components/fxa-client/src/telemetry.rs
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A somewhat mixed-bag of all telemetry we want to collect. The idea is that
+// the app will "pull" telemetry via a new API whenever it thinks there might
+// be something to record.
+// It's considered a temporary solution until either we can record it directly
+// (eg, via glean) or we come up with something better.
```
rfk, line 9: Somewhere in here, we should document out expectation that metrics will be lost if the browser closes/crashes/whatever before they are extracted by the application. I think it's a reasonable interim position for us to take, but will be good to state explicitly.

components/fxa-client/src/telemetry.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use serde_derive::*;
+
+// First, constituent parts
+#[derive(Debug, Serialize)]
+pub struct SentReceivedTab {
```
rfk, line 15: I wonder if having separate `SentTab` and `ReceivedTab` structs would help avoid the kind of awkward naming thing here (or perhaps even `SendCommand` and `ReceivedCommand` to line up more directly with the eventual event names).

components/fxa-client/src/telemetry.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct FxaTelemetry {
+    sent_tabs: Vec<SentReceivedTab>,
+    received_tabs: Vec<SentReceivedTab>,
+}
+
+impl FxaTelemetry {
+    pub fn new() -> Self {
+        FxaTelemetry {
+            sent_tabs: Vec::new(),
+            received_tabs: Vec::new(),
```
rfk, line 36: As above, the eventual event names are going to be "command-sent" and "command-received" so I wonder if there's value in using consistent terminology here. Especially if we're going to have to have multiple consumers extracting these and sending them in the sync ping, the more details we can fix here in the shared code, the better.

components/fxa-client/src/telemetry.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct SentReceivedTab {
+    pub flow_id: String,
+    pub stream_id: String,
+}
+
+// We have a naive strategy to avoid unbounded memory growth - the intention
+// is that if any platform lets things grow to hit these limits, it's probably
+// never going to consume anything - so it doesn't matter what we discard (ie,
+// there's no good reason to have a smarter circular buffer etc)
+const MAX_TAB_EVENTS: usize = 20;
```
rfk, line 24: This seems fairly low to me, and we definitely see instances of users sending more than 20 tabs in a single operation. The per-tab memory overhead here seems pretty low, what do you think about bumping this up by an order-of-magnitude or so?

components/fxa-client/src/commands/send_tab.rs
mhammond: I do - this is more a TODO for myself to work out if test coverage really is missing, and if so, add it.

components/fxa-client/src/send_tab.rs
mhammond: Yeah - we could just deprecate. But we'll need this for desktop assuming we want to keep the `flowID` semantics. OTOH, it's not clear that we do :)

components/fxa-client/src/telemetry.rs
mhammond: agreed - it's worth calling out that desktop is in the same position, and we should consider if we think that's OK or not? I'd be hugely surprised/disappointed if firefox crashing explains lost metrics on desktop 😅

components/fxa-client/src/telemetry.rs
mhammond: While Fenix makes that slightly more difficult (ie, no multi-device capability, and no multi-tab capability other than collections, I've no problem with that.

mhammond: > While Fenix makes that slightly more difficult (ie, no multi-device capability, and no multi-tab capability other than collections, I've no problem with that.

I mis-spoke - it does have "Send to all devices" - so it's going to be impossible to get desktop's flow semantics with the api we expose (ie, as mentioned above, we'll need an api which takes an array of device IDs) That probably shouldn't block getting the core version of this landed though.

mhammond:
I added capturing of the 'reason' like bug 1639843. I also renamed `fetch_device_command` to `ios_fetch_device_command` to try and discourage it being used anywhere other than iOS, because when I first found that function I assumed it was the preferred way of handling a push message, which it isn't.

There are a few comments above I still need to address and need to add some tests, but this seems to record telemetry correctly in Fenix with https://github.com/mozilla-mobile/android-components/pull/7618

components/fxa-client/src/send_tab.rs
mhammond: I've updated the comment but otherwise haven't moved towards actually doing this.

mhammond:
I think this is ready to roll (although I'll get feedback from @grigoryk on https://github.com/mozilla-mobile/android-components/pull/7618 before I actually land it.

components/fxa-client/src/device.rs
```diff
@@ -9,17 +9,24 @@ use crate::{
...
+    telemetry, util, CachedResponse, FirefoxAccount, IncomingDeviceCommand,
 };
 use serde_derive::*;
 use std::collections::{HashMap, HashSet};
 
 // An devices response is considered fresh for `DEVICES_FRESHNESS_THRESHOLD` ms.
 const DEVICES_FRESHNESS_THRESHOLD: u64 = 60_000; // 1 minute
 
+/// The reason we are fetching commands.
+pub enum CommandFetchReason {
```
linacambridge, line 23: This is a pretty small enum—might as well derive `Clone` and `Copy` on it, and pass it by value everywhere.

components/fxa-client/src/device.rs
```diff
@@ -162,19 +169,30 @@ impl FirefoxAccount {
...
     }
 
     /// Retrieve and parse a specific command designated by its index.
     ///
     /// **💾 This method alters the persisted account state.**
-    pub fn fetch_device_command(&mut self, index: u64) -> Result<IncomingDeviceCommand> {
-        let mut device_commands = self.fetch_and_parse_commands(index, Some(1))?;
+    ///
+    /// Note that this should not be used if possible, as it does not correctly
+    /// handle missed messages. It's currently used only on iOS due to platform
```
linacambridge, line 190: Out of curiosity, what does it do that's bad? 😅 Is that the "1 command per push message" limitation that the `push.rs` docs mention?

components/fxa-client/src/send_tab.rs
```diff
@@ -38,22 +38,29 @@ impl FirefoxAccount {
     }
 
     /// Send a single tab to another device designated by its device ID.
+    /// XXX - We need a new send_tabs_to_devices() so we can correctly record
+    /// telemetry for these cases.
+    /// This probably requires a new "Tab" struct with the title and url.
+    /// android-components has SendToAllUseCase(), so this isn't just theoretical.
```
linacambridge, line 44: Bug number?

components/fxa-client/src/telemetry.rs
```diff
@@ -4,10 +4,98 @@
...
+pub struct ReceivedCommand {
+    pub flow_id: String,
+    pub stream_id: String,
+    pub reason: ReceivedReason,
+}
+
+// We have a naive strategy to avoid unbounded memory growth - the intention
+// is that if any platform lets things grow to hit these limits, it's probably
+// never going to consume anything - so it doesn't matter what we discard (ie,
+// there's no good reason to have a smarter circular buffer etc)
```
linacambridge, line 74: Should we record an event for when the buffer overflows, to see how often this happens? Glean has special `invalid_value` and `invalid_overflow` buckets that it uses when the app records weird stuff (string too long, negative or overflowing counter or enum, unknown discriminant), but I don't know if it's worth doing that here.

components/fxa-client/src/commands/send_tab.rs
```diff
@@ -40,16 +41,31 @@ impl EncryptedSendTabPayload {
...
 
 impl SendTabPayload {
-    pub fn single_tab(title: &str, url: &str) -> Self {
-        SendTabPayload {
-            entries: vec![TabHistoryEntry {
-                title: title.to_string(),
-                url: url.to_string(),
-            }],
-        }
+    pub fn single_tab(title: &str, url: &str) -> (Self, telemetry::SentCommand) {
```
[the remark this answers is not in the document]
mhammond: I kinda like this better - it's one less thing to know what the fields actually are.

components/fxa-client/src/device.rs
mhammond: 👍 - but I remember going backwards and forwards about this and landed on "if it's trivial to pass by ref, then why not?" - but yeah, it probably does read a little better, so I did it.

components/fxa-client/src/device.rs
mhammond: It will not see missed messages. If I get a push for `tab-1` and a push for `tab-3`, and between them I've never explicitly polled, I'll miss `tab-2`, even if I try polling now.

mhammond: I added that to the comment :)

components/fxa-client/src/telemetry.rs
mhammond: I don't think it's worth it TBH (at least not now :) Android will grab this immediately after the tab operation, so should never happen. Is this likely on iOS? IOW, do you think we should get an issue opened for this?

PR #3309, opened 2020-07-01
Bug 1657393 - add resource-monitoring toolchain

MihaiTabara:
RelEng has developed a tool to collect CPU/memory/DiskIO/Network usage for any tasks running `runtask`. It's now live and we're turning it on in this repo. This means that from now on, in Github-pull requests, master pushes and releases, the jobs are creating a `resource-monitor.json` at the end. This is collected automatically in BigQuery for us to analyze in order to infer better-fitted (price, powerload, etC) instance types for our workers. 

Once this lands on master pushes and the level-3 toolchains are generated, we'll no longer schedule the individual fetch/toolchain tasks but cache instead.

This has been previously tested in staging releases: 
* pushed this https://treeherder.mozilla.org/#/jobs?repo=taskgraph-try&revision=8d9597fbd50c1d62a6682e61acaa98c2d170679b 
* tested in app-services via https://firefox-ci-tc.services.mozilla.com/tasks/groups/dBwQFm4pSwy2lZC3KsuDRg staging release.
* worked like a charm producing the following artifacts: https://firefoxci.taskcluster-artifacts.net/CIMEsTtIRq-VXQslIvCnbg/0/public/monitoring/resource-monitor.json

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -10,7 +10,7 @@ tasks:
     - $let:
           taskgraph:
               branch: taskgraph
-              revision: cab4565345d0d0effa66f18fe91335dd6d744031
+              revision: 0c74426138739b1cad3bb4ff169af67eef6003fc
           trustDomain: app-services
       in:
           $let:
@@ -220,7 +220,7 @@ tasks:
                                         REPOSITORIES: {$json: {appservices: "Application Services", taskgraph: "Taskgraph"}}
                                         HG_STORE_PATH: /builds/worker/checkouts/hg-store
                                         ANDROID_SDK_ROOT: /builds/worker/android-sdk
-                                        MOZ_FETCHES_DIR: /builds/worker/checkouts/src
+                                        MOZ_FETCHES_DIR: /builds/worker/fetches
                                       - $if: 'tasks_for in ["github-pull-request"]'
                                         then:
                                             APPSERVICES_PULL_REQUEST_TITLE: '${event.pull_request.title}'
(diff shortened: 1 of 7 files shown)


.taskcluster.yml
```diff
@@ -10,7 +10,7 @@ tasks:
     - $let:
           taskgraph:
               branch: taskgraph
-              revision: cab4565345d0d0effa66f18fe91335dd6d744031
+              revision: 0c74426138739b1cad3bb4ff169af67eef6003fc
```
MihaiTabara, line 13: `run-task` changes are in taskgraph now so we need to bump those.

.taskcluster.yml
```diff
@@ -220,7 +220,7 @@ tasks:
                                         REPOSITORIES: {$json: {appservices: "Application Services", taskgraph: "Taskgraph"}}
                                         HG_STORE_PATH: /builds/worker/checkouts/hg-store
                                         ANDROID_SDK_ROOT: /builds/worker/android-sdk
-                                        MOZ_FETCHES_DIR: /builds/worker/checkouts/src
+                                        MOZ_FETCHES_DIR: /builds/worker/fetches
```
MihaiTabara, line 223: Taking advantage to make this consistent with other repos.

.taskcluster.yml
```diff
@@ -266,6 +266,14 @@ tasks:
                                       type: 'directory'
                                       path: '/builds/worker/artifacts'
                                       expires: {$fromNow: '1 year'}
+                                  'public/docker-contexts':
```
MihaiTabara, line 269: This is needed since https://hg.mozilla.org/ci/taskgraph/rev/004bada4f170794a94ad43b84dd7562946305d22 was added, otherwise taskgraph breaks.

.taskcluster.yml
```diff
@@ -266,6 +266,14 @@ tasks:
...
+                                      type: 'directory'
+                                      path: '/builds/worker/checkouts/src/docker-contexts'
```
eoger, line 271: You just changed the `MOZ_FETCHES_DIR` from `checkouts/src` to `fetches` just above, is this correct?

MihaiTabara: Yes, it's unrelated. The `builds/worker/checksouts/src` is where all the artifacts are being checked out. We were initially using that for fetches as well. Which would have been fine as well, until we added `resourcemonitor`. The artifacts produced by it need to stay alive but in a normal taskgraph world, everything that's under `MOZ_FETCHES_DIR` gets wiped-off before tasks run to save space. So removing the artifacts there would've left us with no files after the run-task completed. So instead we're just downloading the artifacts in a different folder (that still gets cleaned-up. We're going to standardize this in the docker image in the near future in https://bugzilla.mozilla.org/show_bug.cgi?id=1654634 anyway so we don't even need to specify this anymore.

Issue #3306, opened 2020-06-30
Expose "no browser-flow" logins methods to Kotlin for testing


eoger:
In order to have e2e Send Tab tests on Android, we need a way to authenticate without having to spin up a browser or add fragile code that messes with UI.

In https://github.com/mozilla/application-services/pull/3187 @tarikeshaq added some methods that allow us to login without a browser flow. We should expose these methods to Kotlin test code.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1442)
┆Story Points: 13
┆Epic: a-s test coverage
┆Sprint: Backlog

eoger:
Current status: I've decided to divide the work in 2 parts, one for the "create account API" and the other one for the "authenticate without UI" parts.
I've created a PR for the former (#3326) that's waiting on feedback comments.
What's left to do is to implement an API to log-in to FxA and how to integrate it with the Kotlin account manager. Most of the Rust code is ready as we can re-use what @tarikeshaq wrote for the Rust sync integration tests. The shape on that API is dependent on what Chris needs to test Send Tab e2e, so we're blocked on this.

PR #3315, opened 2020-07-02
Factor out restmail client code

eoger:
I intend to use the restmail code for https://github.com/mozilla/application-services/issues/3306, first step is to factor out that code and clean it up a little.


components/support/restmail-client/src/lib.rs
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use error::RestmailClientError;
```
tarikeshaq, line 5: nit: We could do a `type Result<T, E = RestMailClientError> = Result<T, E>;` Then throughout the file just use `Result<T>`

PR #3323, opened 2020-07-06
Adds verifying requested scopes in authorize_code_using_session_token

### LINKED ISSUE #3318: Validate requested scopes in authorize_code_using_session_token
As per the discussion in https://github.com/mozilla/application-services/pull/3264#discussion_r449359102

We need to add validating the scopes a client requests with the `FxA` server, so that we only sever the client the scoped keys they are allowed to request. This would be as follows:
- Do what the js client does and send a [scoped-key-data](https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccountsClient.jsm#314) request to the server, where it will respond with a subset of the requested scopes.
- Look internally into the state and do [what we currently do](https://github.com/mozilla/application-services/pull/3264/files#diff-a9e1165a3ac1c1602b346c00db33b006L174) on that subset, i.e:
  - Retrieve the scoped keys from the state matching the requested and validated scopes
  - Encrypt those keys using the current mechanism used in the function

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1453)
┆Story Points: 2
┆Epic: Refactor Desktop FxA APIs to use Fxa-client & AccountMananger

tarikeshaq:
fixes #3318 

Rebased!

### DIFF
--- a/components/fxa-client/src/oauth.rs
+++ b/components/fxa-client/src/oauth.rs
@@ -177,31 +177,54 @@ impl FirefoxAccount {
         &self,
         auth_params: AuthorizationParameters,
     ) -> Result<String> {
-        // TODO: Add validation to check for which keys the client is allowed to request,
-        //       And if we have those keys stored locally.
-        //       Currently we pull ALL available keys that match the requested scopes, regardless
-        //       if the client is actually allowed to request them.
         let session_token = self.get_session_token()?;
+
+        // Validate request to ensure that the client is actually allowed to request
+        // the scopes they requested
+        let allowed_scopes = self.client.scoped_key_data(
+            &self.state.config,
+            &session_token,
+            &auth_params.client_id,
+            &auth_params.scope.join(" "),
+        )?;
+
+        if let Some(not_allowed_scope) = auth_params
+            .scope
+            .iter()
+            .find(|scope| !allowed_scopes.contains_key(*scope))
+        {
+            return Err(ErrorKind::ScopeNotAllowed(
+                auth_params.client_id.clone(),
+                not_allowed_scope.clone(),
+            )
+            .into());
+        }
+
         let keys_jwe = if let Some(keys_jwk) = auth_params.keys_jwk {
+            let mut scoped_keys = HashMap::new();
+            allowed_scopes
+                .iter()
+                .try_for_each(|(scope, _)| -> Result<()> {
+                    scoped_keys.insert(
+                        scope,
+                        self.state
+                            .scoped_keys
+                            .get(scope)
+                            .ok_or_else(|| ErrorKind::NoScopedKey(scope.clone()))?,
+                    );
+                    Ok(())
+                })?;
+            let scoped_keys = serde_json::to_string(&scoped_keys)?;
             let keys_jwk = base64::decode_config(keys_jwk, base64::URL_SAFE_NO_PAD)?;
             let keys_jwk = String::from_utf8(keys_jwk)?;
-            let requested_scopes: HashSet<&str> = auth_params.scope.split_whitespace().collect();
-            let data: HashMap<String, ScopedKey> = self
-                .state
-                .scoped_keys
-                .iter()
-                .filter(|(scope, _)| requested_scopes.contains(&(*scope).as_ref()))
-                .map(|(key, value)| (key.clone(), value.clone()))
-                .collect();
-            let data = serde_json::to_string(&data)?;
             let client_keys_flow = ScopedKeysFlow::with_random_key()?;
-            Some(client_keys_flow.encrypt_keys_jwe(&keys_jwk, data.as_bytes())?)
+            Some(client_keys_flow.encrypt_keys_jwe(&keys_jwk, scoped_keys.as_bytes())?)
         } else {
             None
         };
         let auth_request_params = AuthorizationRequestParameters {
             client_id: auth_params.client_id,
-            scope: auth_params.scope,
+            scope: auth_params.scope.join(" "),
             state: auth_params.state,
             access_type: auth_params.access_type,
             code_challenge: auth_params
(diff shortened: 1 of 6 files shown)


components/fxa-client/src/oauth.rs
```diff
@@ -172,25 +172,47 @@ impl FirefoxAccount {
...
+        let valid_scopes = self.client.scoped_key_data(
+            &self.state.config,
+            &session_token,
+            &auth_params.client_id,
+            &auth_params.scope,
+        )?;
+
+        if let Some(not_allowed_scope) = auth_params
+            .scope
+            .split_whitespace()
```
rfk, line 188: I wonder if it would be simpler/safer to do this `split_whitespace` step when you're constructing `AuthorizationParameters` struct. That is, have `AuthorizationParameters::scope` be a `Vec<String>` rather than a `String`, so that we don't accidentally forget to call `split_whitespace` if we end up using it in other places in future.

Not a blocker, just a thought.

components/fxa-client/src/oauth.rs
```diff
@@ -172,25 +172,47 @@ impl FirefoxAccount {
...
     ) -> Result<String> {
-        // TODO: Add validation to check for which keys the client is allowed to request,
-        //       And if we have those keys stored locally.
-        //       Currently we pull ALL available keys that match the requested scopes, regardless
-        //       if the client is actually allowed to request them.
         let session_token = self.get_session_token()?;
+
+        // Validate request to ensure that the client is actually allowed to request
+        // the scopes they requested
+        let valid_scopes = self.client.scoped_key_data(
```
rfk, line 179: Nit: this is begging to be called `allowed_scopes` rather than `valid_scopes`, both because it's more precise and because the surrounding comments etc use the word "allowed" :-)

components/fxa-client/src/oauth.rs
```diff
@@ -172,25 +172,47 @@ impl FirefoxAccount {
...
+            let first_mismatch = scoped_keys
+                .iter()
+                .find(|(scope, _)| !valid_scopes.contains_key(*scope));
+            if let Some(scope) = first_mismatch {
+                return Err(ErrorKind::NoScopedKey(scope.0.clone()).into());
+            }
+            let scoped_keys = serde_json::to_string(&scoped_keys)?;
             let client_keys_flow = ScopedKeysFlow::with_random_key()?;
-            Some(client_keys_flow.encrypt_keys_jwe(&keys_jwk, data.as_bytes())?)
+            Some(client_keys_flow.encrypt_keys_jwe(&keys_jwk, scoped_keys.as_bytes())?)
```
rfk, line 215: I'd like to see a couple of tests for correct behaviour here, especially given the security-sensitive nature of the checks you're doing. We should check that:

* If the server returns an error, the client receives an error.
* If the server returns a list of allowed scopes that is smaller than those requested by the client, then the client does not get the disallowed scopes.
* That the `NoScopedKey` error you're throwing above is actually throw when expected.

components/fxa-client/src/oauth.rs
```diff
@@ -172,25 +172,47 @@ impl FirefoxAccount {
...
-            let data = serde_json::to_string(&data)?;
+
+            // Return an error if we could not get the requested scopes from state
+            // Saves the client time instead of waiting for the server's response
+            let first_mismatch = scoped_keys
+                .iter()
+                .find(|(scope, _)| !valid_scopes.contains_key(*scope));
+            if let Some(scope) = first_mismatch {
+                return Err(ErrorKind::NoScopedKey(scope.0.clone()).into());
+            }
```
rfk, line 212: I don't understand this check, or the comment about waiting for the server's response. Having an explicit testcase would probably help make it clearer.

It appears to be looking for a member of `scoped_keys` that is not in `valid_scopes`, but AFAICT every member of `scoped_keys` must be in `valid_scopes` by definition, because we check that as part of constructing `scoped_keys` in the first place.

You may find that doing this as a separate check is unnecessary if you invert the way you're building `scoped_keys` above: iterate over `valid_scopes`, call `self.state.get()` for each item, and allow it to throw the `NoScopedKey` error for you if we don't have that key in our local state.

tarikeshaq: Yup that was a brain fart 🤦 Wanted to do the inverted check (i.e see if there is something in the `valid_scopes` not included) I like the inverted construction of `scoped_keys` better though! Thank you!

components/fxa-client/src/oauth.rs
tarikeshaq: Alrighty added a couple of tests in the integration tests workflow, although the position of the tests feels a little awkward 😅 

One thing I did notice however is that the server seems to return an error if we are requesting valid scopes that we are not allowed to request, and a bit of digging into the fxa server seems to align with that: [validation function in fxa](https://github.com/mozilla/fxa/blob/166d5871465334e1e2b4b237ad3f6cc00095455f/packages/fxa-auth-server/lib/oauth/grant.js#L96-L114)

Although with invalid scopes it does seem to filter them out, what I do in that case right now is error out.

rfk: > One thing I did notice however is that the server seems to return an error if we are requesting valid scopes
> that we are not allowed to request, and a bit of digging into the fxa server seems to align with that

Gotcha. That makes sense I think, and as long as we propagate the error things will be fine. It's probably the better behaviour anyway rather than just pretending those scopes were not requested :-)

rfk: > Alrighty added a couple of tests in the integration tests workflow, although the position of
> the tests feels a little awkward 😅

Heh, I see what you mean about being awkward (although FWIW I'm pleasantly surprised by how little code was required for each test).

Is this the sort of thing that might be a better fit for unittest-style tests, as part of `mod tests` further down in this file? I haven't looked into it in any detail, but it seems plausible that we could mock out `self.client.scoped_key_data` to test some of the edge-cases, without having to spin up a full-blown integration test. Thoughts?

tarikeshaq: That does indeed work better and gets rid of the awkwardness! Thanks Ryan!

components/fxa-client/src/oauth.rs
```diff
@@ -177,31 +177,54 @@ impl FirefoxAccount {
...
+            &auth_params.client_id,
+            &auth_params.scope.join(" "),
+        )?;
+
+        if let Some(not_allowed_scope) = auth_params
+            .scope
+            .iter()
+            .find(|scope| !allowed_scopes.contains_key(*scope))
+        {
+            return Err(ErrorKind::ScopeNotAllowed(
```
vladikoff, line 196: 👍  Thanks!

PR #3324, opened 2020-07-06
Adds docs on regen protobufs for kotlin

tarikeshaq:
Adds docs for regenerating local kotlin protobuf files without having to run a build

### DIFF
--- a/docs/howtos/passing-protobuf-data-over-ffi.md
+++ b/docs/howtos/passing-protobuf-data-over-ffi.md
@@ -196,6 +196,12 @@ follow the examples of the other steps it takes.
         }
     ```
 
+If you make any changes to the `.proto` file and want regenerate your local kotlin files, you can use:
+```
+./gradlew generateDebugProto
+```
+While being in the root directory of `application-services`.
+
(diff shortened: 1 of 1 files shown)


docs/howtos/passing-protobuf-data-over-ffi.md
```diff
@@ -196,6 +196,12 @@ follow the examples of the other steps it takes.
         }
     ```
 
+If you make any changes to the `.proto` file and want regenerate your local kotlin files, you can use:
+```
+./gradlew :projectname:generateDebugProto
```
eoger, line 201:
```suggestion
./gradlew generateDebugProto
```
will work!

eoger: I'd be nice if we had a gradle target that found all tasks of `GenerateProtobufTask` type (extract, debug/release builds etc.) in our project and run them. Maybe you can create a follow-up issue?

tarikeshaq: Nice!

PR #3326, opened 2020-07-06
Kotlin temporary account API

eoger:
Connects to https://github.com/mozilla/application-services/issues/3306.

This PR tackles the problem of creating test accounts:
In `TempTestAccount.create` the Rust layer creates a temporary email address and does the account verification step without showing UI. When `TempTestAccount` goes out of scope, the account is destroyed.

In a following PR, I intend to create a `TempTestAccount.login` method that returns a `FirefoxAccount`.

f? @rfk @grigoryk

### DIFF
--- a/components/fxa-client/Cargo.toml
+++ b/components/fxa-client/Cargo.toml
@@ -20,6 +20,7 @@ sync15 = { path = "../sync15" }
 url = "2.1"
 ffi-support = "0.4"
 viaduct = { path = "../viaduct" }
+restmail-client = { path = "../support/restmail-client" }
 rc_crypto = { path = "../support/rc_crypto", features = ["ece", "hawk"] }
 error-support = { path = "../support/error" }
 thiserror = "1.0"
@@ -36,4 +37,3 @@ mockito = "0.26"
 [features]
 default = []
 gecko = [ "rc_crypto/gecko" ]
-integration_test = []
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/testing/TempTestAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/testing/TempTestAccount.kt
@@ -1,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.appservices.fxaclient.testing
+
+import mozilla.appservices.fxaclient.Config
+import mozilla.appservices.fxaclient.MsgTypes
+import mozilla.appservices.fxaclient.rust.LibFxAFFI
+import mozilla.appservices.fxaclient.rustCall
+
+/**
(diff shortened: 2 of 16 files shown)


components/fxa-client/Cargo.toml
```diff
@@ -20,6 +20,7 @@ sync15 = { path = "../sync15" }
 url = "2.1"
 ffi-support = "0.4"
 viaduct = { path = "../viaduct" }
+restmail-client = { path = "../support/restmail-client" }
```
rfk, line 23: I don't think it's going to be a big deal in practice, but part of me feels bad about including code here that's only going to be used for tests, but will still take up space in the compiled library, the Fenix app, etc.

I don't have a good suggestion for what to do about it though :-(

components/fxa-client/src/lib.rs
```diff
@@ -32,8 +32,8 @@ pub mod device;
 pub mod error;
 pub mod ffi;
 pub mod migrator;
+pub mod testing;
 
-#[cfg(feature = "integration_test")]
 pub mod auth;
```
rfk, line 37: Since it's no longer gated on a feature, perhaps we should move this whole `auth` module under the `testing` namespace to make it clear that it should only be used for testing purposes.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/testing/TempTestAccount.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    override fun close() {
+        rustCall { e ->
+            LibFxAFFI.INSTANCE.fxa_testing_destroy_temp_account(
+                config.contentUrl,
+                config.clientId,
+                config.redirectUri,
+                config.tokenServerUrlOverride,
```
rfk, line 44: It feels weird to have to pass in all these arguments, as it's not obvious why a test account would need to know about e.g. `tokenServerUrlOverride`. Is this entirely because we want to re-create a `Config` on the other side? If so, I wonder if we could automatically fill in some dummy values rather than plumbing these down through all the layers of the API.

components/fxa-client/src/testing.rs
```diff
@@ -0,0 +1,112 @@
...
+        "uid": uid,
+        "code": code,
+    });
+    crate::auth::send_verification(&config, body)?.require_success()?;
+    Ok(())
+}
+
+fn rand_account_details() -> Result<TempAccountDetails> {
+    let username = rand_str()?;
+    let email = format!("{}@restmail.net", username);
```
rfk, line 101: Nit: the string "restmail.net" feels like it's an implementation detail of the `restmail_client` create, which makes me wonder whether this "generate a random email" thing should be a method on that crate.

components/fxa-client/Cargo.toml
eoger: Agree, for what it's worth it's a very small crate (<100 LOC)

rfk:
Just want to check, is this blocked on any further feedback from me, or do we need @grigoryk to weigh in here as a next step?

eoger:
I'd like to see Grisha's feedback before going on.

PR #3328, opened 2020-07-07
Adds optional metrics params

### LINKED ISSUE #3278: Support optional metrics parameters in the FxA Client
beginOAuthFlow and beginPairingFlow should support these **optional** parameters

* flow_id
* flow_begin_time
* device_id
* utm_source
* utm_content
* utm_medium
* utm_term
* utm_campaign
* entrypoint_experiment
* entrypoint_variation

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1423)
┆Story Points: 8

tarikeshaq:

fixes #3278 

Adds optional parameters for `begin_oauth_flow` and `begin_pairing_flow`. I'm gonna leave a few comments on the code about a few things I was thinking about!

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -6,7 +6,20 @@
 
 ## FxA Client
 
+### What's new ###
+- Consumers can now optionally include parameters for metrics in `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios). Those parameters can be passed in using a `MetricsParams` struct/class. `MetricsParams` is defined in both the Kotlin and Swift bindings. The parameters are the following ([#3328](https://github.com/mozilla/application-services/pull/3328)):
+  - flow_id
+  - flow_begin_time
+  - device_id
+  - utm_source
+  - utm_content
+  - utm_medium
+  - utm_term
+  - utm_campaign
+  - entrypoint_experiment
+  - entrypoint_variation
+
 ### ⚠️ Breaking changes ⚠️
 
-- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOauthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
+- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
 - Changes the `authorizeOAuthCode` API to now accept an `AuthorizationParams` object instead of the individual parameters. The `AuthorizationParams` also includes optional `AuthorizationPKCEParams` that contain the `codeChallenge`, `codeChallengeMethod`. `AuthorizationParams` also includes an optional `keysJwk` for requesting keys ([#3264](https://github.com/mozilla/application-services/pull/3264))
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
@@ -100,12 +100,26 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      *
      * @param scopes List of OAuth scopes for which the client wants access
      * @param entrypoint to be used for metrics
+     * @param metricsParams optional parameters used for metrics
      * @return String that resolves to the flow URL when complete
      */
-    fun beginOAuthFlow(scopes: Array<String>, entrypoint: String): String {
+    fun beginOAuthFlow(
+        scopes: Array<String>,
+        entrypoint: String,
+        metricsParams: MetricsParams = MetricsParams()
+    ): String {
         val scope = scopes.joinToString(" ")
+        val (nioBuf, len) = metricsParams.intoMessage().toNioDirectBuffer()
         return rustCallWithLock { e ->
-            LibFxAFFI.INSTANCE.fxa_begin_oauth_flow(this.handle.get(), scope, entrypoint, e)
+            val ptr = Native.getDirectBufferPointer(nioBuf)
+            LibFxAFFI.INSTANCE.fxa_begin_oauth_flow(
+                    this.handle.get(),
+                    scope,
+                    entrypoint,
+                    ptr,
+                    len,
+                    e
+            )
         }.getAndConsumeRustString()
     }
 
@@ -116,13 +130,29 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      *
      * @param pairingUrl the url to initilaize the paring flow with
      * @param scopes List of OAuth scopes for which the client wants access
-     * @param entrypoint to be used for data collection
+     * @param entrypoint to be used for metrics
+     * @param metricsParams optional parameters used for metrics
      * @return String that resoles to the flow URL when complete
      */
-    fun beginPairingFlow(pairingUrl: String, scopes: Array<String>, entrypoint: String): String {
+    fun beginPairingFlow(
+        pairingUrl: String,
+        scopes: Array<String>,
+        entrypoint: String,
+        metricsParams: MetricsParams = MetricsParams()
+    ): String {
         val scope = scopes.joinToString(" ")
+        val (nioBuf, len) = metricsParams.intoMessage().toNioDirectBuffer()
         return rustCallWithLock { e ->
-            LibFxAFFI.INSTANCE.fxa_begin_pairing_flow(this.handle.get(), pairingUrl, scope, entrypoint, e)
+            val ptr = Native.getDirectBufferPointer(nioBuf)
+            LibFxAFFI.INSTANCE.fxa_begin_pairing_flow(
+                    this.handle.get(),
+                    pairingUrl,
+                    scope,
+                    entrypoint,
+                    ptr,
+                    len,
+                    e
+            )
         }.getAndConsumeRustString()
     }
 
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/MetricsParams.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/MetricsParams.kt
@@ -1,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.appservices.fxaclient
+
+data class MetricsParams(
+    val flowId: String? = null,
+    val flowBeginTime: Long? = null,
+    val deviceId: String? = null,
+    val utmSource: String? = null,
+    val utmContent: String? = null,
+    val utmMedium: String? = null,
+    val utmTerm: String? = null,
+    val utmCampaign: String? = null,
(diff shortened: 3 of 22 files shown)


components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/MetricsParams.kt
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+package mozilla.appservices.fxaclient
+
+data class MetricsParams(
+    val parameters: Map<String, String>
```
tarikeshaq, line 8: Wasn't sure about my choice here. I first tried to list out the expected parameters:

 - flow_id
 - flow_begin_time
 - device_id
 - utm_source
 - utm_content
 - utm_medium
 - utm_term
  - utm_campaign
  - entrypoint_experiment
  - entrypoint_variation

But it added quite a bit of bloat (Had to null check them since they're all optional and that just explodes up)

But I do get the concern that this does not tell the consumer exactly what the parameters should be.

components/fxa-client/ffi/src/lib.rs
```diff
@@ -198,19 +199,26 @@ pub extern "C" fn fxa_get_manage_devices_url(
...
     error: &mut ExternError,
 ) -> *mut c_char {
     log::debug!("fxa_begin_pairing_flow");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
         let pairing_url = pairing_url.as_str();
         let scope = scope.as_str();
         let scopes: Vec<&str> = scope.split(' ').collect();
-        fxa.begin_pairing_flow(&pairing_url, &scopes, entrypoint.as_str())
+        let metrics_params =
+            match MetricsParams::from_protobuf_ptr(metrics_params, metrics_params_len) {
```
tarikeshaq, line 217: Wasn't sure if I should just error out if something goes wrong with the metrics. Thought I'd just keep it as an option, but I could see this coming back to bite us, since if something breaks in the metrics, we won't be able to see it, unless we explicitly check the url....

components/fxa-client/ffi/src/lib.rs
```diff
@@ -546,7 +561,7 @@ pub unsafe extern "C" fn fxa_initialize_device(
     log::debug!("fxa_initialize_device");
     ACCOUNTS.call_with_result_mut(error, handle, |fxa| {
         let capabilities =
-            DeviceCapability::from_protobuf_array_ptr(capabilities_data, capabilities_len);
+            DeviceCapability::from_protobuf_array_ptr(capabilities_data, capabilities_len)?;
```
tarikeshaq, line 564: This now errors out instead of panic'ing (Was this an okay change?)

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -100,12 +100,26 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
      *
      * @param scopes List of OAuth scopes for which the client wants access
      * @param entrypoint to be used for metrics
+     * @param metricsParams optional parameters used for metrics
      * @return String that resolves to the flow URL when complete
      */
-    fun beginOAuthFlow(scopes: Array<String>, entrypoint: String): String {
+    fun beginOAuthFlow(
+        scopes: Array<String>,
+        entrypoint: String,
```
tarikeshaq, line 108: Was thinking about making `entrypoint` a part of the `metricsParams`, but that doesn't enforce it. Should it be enforced? Otherwise, I'll just put it in the metrics params and simplify this

components/fxa-client/ios/FxAClient/FxAccountMetrics.swift
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+public struct MetricsParams {
+    public let parameters: [String: String]
```
tarikeshaq, line 8: Same thing here wasn't sure if this best fits as a dictionary or if I should spit out all the expected params

components/fxa-client/ios/FxAClient/RustFxAccount.swift
```diff
@@ -104,18 +104,48 @@ open class RustFxAccount {
...
-        let ptr = try rustCall { err in
-            fxa_begin_oauth_flow(self.raw, scope, entrypoint, err)
+        let (metricsParamsData, metricsParamsLen) = msgToBuffer(msg: metricsParams.toMsg())
+        let ptr = try metricsParamsData.withUnsafeBytes { bytes in
+            try rustCall { err in
+                fxa_begin_oauth_flow(
+                    self.raw,
+                    scope,
+                    entrypoint,
+                    bytes.bindMemory(to: UInt8.self).baseAddress!,
```
tarikeshaq, line 120: Honestly not sure what this does, but what was done in the other apis.

components/fxa-client/src/ffi.rs
```diff
@@ -36,6 +39,18 @@ pub mod error_codes {
     pub const NETWORK: i32 = 3;
 }
 
+pub trait FromProtobufPtr<T: prost::Message + Default + Into<Self>>: Sized {
```
tarikeshaq, line 42: Thought this can help, since I basically did the same thing in #3264

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
rfk: I recall discussion with @vladikoff that we thought the `entrypoint` should be required, and hence not part of the metrics params.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/MetricsParams.kt
rfk: Hrm, yeah, I see what you mean about the amount of boilerplate required. Listing them out individually still feels like the right thing to me, but I'd be curious what other folks think.

components/fxa-client/ffi/src/lib.rs
rfk: I think we should error out here - anything that goes wrong at this point will be something that went wrong inside the protobuf library, which we probably want to have show up in sentry.

components/fxa-client/ios/FxAClient/RustFxAccount.swift
rfk: me neither :-/

components/fxa-client/src/ffi.rs
rfk: +1; I also wonder if it's the sort of thing we should push all the way down into the `ffi-support` create, /cc @thomcc

components/fxa-client/examples/devices_api.rs
```diff
@@ -4,7 +4,7 @@
 
 use cli_support::prompt::prompt_string;
 use dialoguer::Select;
-use fxa_client::{device, Config, FirefoxAccount, IncomingDeviceCommand};
+use fxa_client::{device, ffi::MetricsParams, Config, FirefoxAccount, IncomingDeviceCommand};
```
tarikeshaq, line 7: Feel a little awkward access this using the ffi module 🙃 Another option is to make `oauth` public

CHANGES_UNRELEASED.md
```diff
@@ -8,5 +8,8 @@
...
\ No newline at end of file
+- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
+- Changes the `authorizeOAuthCode` API to now accept an `AuthorizationParams` object instead of the individual parameters. The `AuthorizationParams` also includes optional `AuthorizationPKCEParams` that contain the `codeChallenge`, `codeChallengeMethod`. `AuthorizationParams` also includes an optional `keysJwk` for requesting keys ([#3264](https://github.com/mozilla/application-services/pull/3264))
+
+### Non breaking changes ###
```
vladikoff, line 14: This can just be `What's new`

example: https://github.com/mozilla/application-services/blob/main/CHANGELOG.md#whats-changed-4

CHANGES_UNRELEASED.md
```diff
@@ -8,5 +8,8 @@
...
+- Consumers can now optionally include parameters for metrics in `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios). Those parameters can be passed in using a `MetricsParams` struct/class. `MetricsParams` is defined in both the Kotlin and Swift bindings and has the supported metrics parameters defined as members. ([#3328](https://github.com/mozilla/application-services/pull/3328))
```
vladikoff, line 15: Could you please expand here what new parameters we support as part of `MetricsParams`? This will be easier for publishing changelogs elsewhere later.

components/fxa-client/src/oauth.rs
```diff
@@ -110,14 +110,17 @@ impl FirefoxAccount {
     /// the pairing authority.
     /// * `scopes` - Space-separated list of requested scopes by the pairing supplicant.
     /// * `entrypoint` - The entrypoint to be used for data collection
+    /// * `metrics` - Optional parameters for metrics
     pub fn begin_pairing_flow(
         &mut self,
         pairing_url: &str,
         scopes: &[&str],
         entrypoint: &str,
+        metrics: MetricsParams,
```
vladikoff, line 119: @tarikeshaq These are all optional parameters. What are your thoughts on making this a `Option<MetricsParams>` ? this way everywhere you use it you can just do something like `acct.begin_oauth_flow(&SCOPES, "device_api_example", None)?;` This should make it easier for us whereever we use it to avoid including the `ffi` module.

tarikeshaq: Alrighty done!

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/MetricsParams.kt
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.appservices.fxaclient
+
+data class MetricsParams(
+    val flowId: String? = null,
+    val flowBeginTime: String? = null,
```
vladikoff, line 9: This is a timestamp at first and then we turn it into a "string" for the url. Not sure if we want to change this for the consumers.

components/fxa-client/src/ffi.rs
thomcc: I think this could be a generic function instead of a trait, no?

Regardless, my hesitation with putting this into ffi-support is: I don't really like adding an unsafe API like that. In particular, this shape of API (which we admittedly use elsewhere) has bitten us in the past (and each time we've said something along the lines of "lets be less aggressive about it" but then did nothing...).

As it is, I don't believe FFI support offers any non-deprecated unsafe APIs that you're expected to use directly, and I think that's a good thing -- We've had very little issue with the APIs it offers in this way.

The other thing is that we don't have a prost dependency in ffi-support atm, but I guess it could be kept optional.

Anyway, it's also not like a safe api for this can't be done, it's just not really trivial. That said, I'm pretty surprised that we haven't gotten around to it by now... So, given those things, I'd rather not add it -- it'd be the first unsafe user-facing API we add to the crate, and really we should bite the bullet and just make a safe API for it, or wait for uniffi to save us.

components/fxa-client/ffi/src/lib.rs
vladikoff: After Slack discussion, it should be fine.

components/fxa-client/src/ffi.rs
rfk: Good points, and I don't feel strongly about needing to move it with any urgency.

Is there a world where we could wrap it into some sort of safely-ecapsulated macro, like an `implement_from_ffi_by_protobuf` that is dual to `implement_into_ffi_by_protobuf`?

PR #3329, opened 2020-07-07
Exclude slow to build crates which are unlikely to break frequently from the workspace default-members

thomcc:
That is, if you do a `cargo test` from the root, this makes it so only some crates build. `cargo test --all`, `cargo test -p thing`, and `cargo test` from somewhere other than workspace root all do the same as old behavior.

This is conservative and only changes:

- nss's systest which is very slow to build and low risk to break when you don't know you're messing with it

- the megazords, which are essentially covered by the ffi, since they basically just reexport it.

With this change cargo build is about 2x faster, e.g. 71s instead of 150s, although it fluxuates a lot.

Downside: There's a bunch of duplication. The root cause of it is https://github.com/rust-lang/cargo/issues/8460, or we could do something like:

```toml
[workspace]
members = [
    "components/*",
    "components/*/ffi",
    "components/support/*",
    "components/support/rc_crypto/nss/systest",
    "megazords/full",
    "megazords/ios/rust",
    "megazords/lockbox",
    "testing/sync-test",
    "tools/protobuf-gen",
]
default-members = [
    "components/*",
    "components/*/ffi",
    "components/support/*",
    "megazords/full",
    "megazords/ios/rust",
    "megazords/lockbox",
    "testing/sync-test",
    "tools/protobuf-gen",
]
# The globs above hit a few folders which aren't crates,
# so exclude them directly
exclude = [
    "components/support",
    "components/support/android",
]
```

However, if it looks like that cargo issue will ever get fixed, we move some projects into different folders so that globbing the rust crates would be possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+# Note: Any additions here should be repeated in default-members below.
 members = [
     "components/fxa-client",
     "components/fxa-client/ffi",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -37,6 +38,64 @@ members = [
...
+# impact to how likely they are to get broken by average changes:
+#
+# - `nss_sys`'s companion test crate `systest` adds `syntex_syntax2` which is
+#   (surprisingly) by far the slowest crate to build in our dep graph, and very
+#   very rarely needs changing
+#
+# - The megazords just rexport ffi crates, which we aren't excluding, and get
+#   built freqently enough as part of gradle/xcode's build process.
+#
+# To be clear: passing the `--all` or `--workspace` arg to cargo will make use
```
rfk, line 63: nit: "make use" -> "make it use" or similar?

Issue #3314, opened 2020-07-02
Investigate verify_connection incorrectly returning empty list


jonalmeida:
Starting from https://github.com/mozilla/application-services/issues/3127#issuecomment-634557955, we were trying to find the various causes of push messages not being delivered.

In the [investigation ticket in android components](https://github.com/mozilla-mobile/android-components/issues/7143), we found some cases (that are now fixed) where we were not calling `PushManager.verify_connection` at the [expected interval](https://github.com/mozilla-mobile/android-components/issues/7236). However, we still were seeing an empty list of subscriptions returned back to us when we should have been seeing a list of the expired subscriptions.

With https://github.com/mozilla-services/autopush/issues/1397 resolved, we are now left to investigate the code path in the native Rust push manager.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1452)
┆Story Points: 5
┆Epic: Send Tab Success Rate Improvements
┆Sprint: SYNC - end 2020-11-06

eoger:
@jrconlin I've spent some time this morning auditing the set of Rust functions involved in the `verify_connection` Kotlin call and found nothing out of the ordinary. Is it possible the Autopush server is returning expired subscriptions, making the Rust code think its local subscriptions are up-to-date?

jrconlin:
@eoger The server returns the list of known subscriptions for a given user. These are subscriptions where the UA has not performed an unsubscribe action. Subscriptions do not have an expiration date and should continue to be valid until either the UA sends an unsubscribe or the bridge specifies that the associated bridge token is no longer valid. 

In terms of the server code, the verify function [fetches](https://github.com/mozilla-services/autopush/blob/master/autopush/http.py#L145) the channel list from the server, which reduces [to calling](https://github.com/mozilla-services/autopush/blob/master/autopush/web/registration.py#L393) the `all_channels()` function. 

The all_channels test is [here](https://github.com/mozilla-services/autopush/blob/master/autopush/tests/test_db.py#L342).

rfk:
> Subscriptions do not have an expiration date and should continue to be valid until either the UA sends an unsubscribe
> or the bridge specifies that the associated bridge token is no longer valid.

I wonder if the problem here is that the FxA server comes to believe a subscription has expired, while the push server itself does not.

[...]

PR #3333, opened 2020-07-09
verify_connection cleanup

eoger:
Investigating on https://github.com/mozilla/application-services/issues/3314 I did some cleanup trying to understand the code.

### DIFF
--- a/components/push/src/communications.rs
+++ b/components/push/src/communications.rs
@@ -12,7 +12,8 @@
 
 use serde_derive::*;
 use serde_json::Value;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::iter::FromIterator;
 use url::Url;
 use viaduct::{header_names, status_codes, Headers, Request};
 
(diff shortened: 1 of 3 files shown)


components/push/src/communications.rs
```diff
@@ -404,17 +402,12 @@ impl Connection for ConnectHttp {
         if &self.options.sender_id == "test" {
             return Ok(false);
         }
-        let my_chans: Vec<String> = channels
-            .to_vec()
-            .into_iter()
-            .map(|s| Store::normalize_uuid(&s))
```
eoger, line 410: The channel_ids are already normalized on the DB.

components/push/src/communications.rs
```diff
@@ -404,17 +402,12 @@ impl Connection for ConnectHttp {
...
-            .collect();
-        log::debug!("Getting Channel List");
-        let remote = self.channel_list()?;
-        // verify both lists match. Either side could have lost it's mind.
+        let local_channels: HashSet<String> = HashSet::from_iter(channels.iter().cloned());
+        let remote_channels: HashSet<String> = HashSet::from_iter(self.channel_list()?);
```
eoger, line 406: The equality of vectors was wrong here, since we were testing the content AND the order of the elements. Using a set is the proper way to do this.

PR #3335, opened 2020-07-09
Better bridged envelopes

### LINKED ISSUE #3069: Add `sortindex` to the bridged envelopes
The `sortindex` is an optional field that's used as a hint for how important an item is to sync, relative to other items in its collection. When used with `sort = index` on a collection `GET` request, the Sync server will return items with higher sort indexes first.

This bug involves adding a `sortindex` field to both [`IncomingEnvelope`](https://github.com/mozilla/application-services/blob/5c52133695116806531069ae7edd9c2d92394359/components/support/sync15-traits/src/bridged_engine.rs#L133) and [`OutgoingEnvelope`](https://github.com/mozilla/application-services/blob/5c52133695116806531069ae7edd9c2d92394359/components/support/sync15-traits/src/bridged_engine.rs#L164). It should be an `Option<i32>`.

`sortindex`, and `ttl`, in the future, is an "auto field" that should be handled similarly to how `sync15` does it for `CleartextBso`. (Technically, `IncomingEnvelope::payload(&self)` is returning a "cleartext BSO", except that type is in `sync15`, and tied to the crypto pieces, so it can't express that...but the logic is identical, and I wonder if we can lift the auto field handling into `sync15_traits`). This part makes it a little too complicated to be a good first bug.

We don't use these auto fields in our bridged engine yet, but, as @mhammond says in https://github.com/mozilla/application-services/pull/3067#discussion_r419194177, this is conceptually where they would go.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1225)
┆Epic: important not urgent
┆Sprint: Backlog

kaylagalway:
Fixes #3069 

Adds`sortIndex` and `ttl` to `OutgoingEnvelope`
Adds `ttl` to `IncomingEnvelope`

Also adds proper handling for `sortIndex` and `ttl` autofields

### DIFF
--- a/components/support/sync15-traits/src/bridged_engine.rs
+++ b/components/support/sync15-traits/src/bridged_engine.rs
@@ -125,6 +125,8 @@ pub struct IncomingEnvelope {
     pub modified: ServerTimestamp,
     #[serde(default)]
     pub sortindex: Option<i32>,
+    #[serde(default)]
+    pub ttl: Option<u32>,
     // Don't provide access to the cleartext directly. We want all callers to
     // use `IncomingEnvelope::payload`, so that we can validate the cleartext.
     cleartext: String,
@@ -142,7 +144,10 @@ impl IncomingEnvelope {
                 payload: payload.id,
             });
         }
-        Ok(payload)
+        // Remove auto field data from payload and replace with real data
+        Ok(payload
+            .with_auto_field("ttl", self.ttl)
+            .with_auto_field("sortindex", self.sortindex))
     }
 }
 
@@ -153,14 +158,22 @@ impl IncomingEnvelope {
 pub struct OutgoingEnvelope {
     id: Guid,
     cleartext: String,
+    sortindex: Option<i32>,
+    ttl: Option<u32>,
 }
 
 impl From<Payload> for OutgoingEnvelope {
     fn from(payload: Payload) -> Self {
+        let mut payload = payload;
         let id = payload.id.clone();
+        // Remove auto field data from OutgoingEnvelope payload
+        let ttl = payload.take_auto_field("ttl");
+        let sortindex = payload.take_auto_field("sortindex");
         OutgoingEnvelope {
             id,
             cleartext: payload.into_json_string(),
+            sortindex,
+            ttl,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


components/support/sync15-traits/src/payload.rs
```diff
@@ -40,6 +40,52 @@ impl Payload {
         self
     }
 
+    /// "Auto" fields are fields like 'sortindex' (and potentially 'ttl' in
+    /// the future) which are:
+    ///
+    /// - Added to the payload automatically when deserializing if present on
+    ///   the incoming BSO.
+    /// - Removed from the payload automatically and attached to the BSO if
+    ///   present on the outgoing payload.
```
linacambridge, line 49:
```suggestion
    ///   the incoming BSO or envelope.
    /// - Removed from the payload automatically and attached to the BSO or
    ///   envelope if present on the outgoing payload.
```

...Since the envelope takes the place of the BSO on Desktop.

components/support/sync15-traits/src/payload.rs
```diff
@@ -40,6 +40,52 @@ impl Payload {
         self
     }
 
+    /// "Auto" fields are fields like 'sortindex' (and potentially 'ttl' in
+    /// the future) which are:
```
linacambridge, line 44:
```suggestion
    /// "Auto" fields are fields like 'sortindex' and 'ttl', which are:
```

components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -153,14 +158,22 @@ impl IncomingEnvelope {
 pub struct OutgoingEnvelope {
     id: Guid,
     cleartext: String,
+    sortindex: Option<i32>,
+    ttl: Option<u32>,
 }
 
 impl From<Payload> for OutgoingEnvelope {
     fn from(payload: Payload) -> Self {
+        let mut payload = payload;
```
linacambridge, line 167:
```suggestion
    fn from(mut payload: Payload) -> Self {
```

I think we can declare the binding as `mut` in the argument, and avoid shadowing—though this also works!

kaylagalway: Ah yeah that makes sense, will commit this!

kaylagalway:
> Thanks again! 🎉

Thank you for the review!

PR #3339, opened 2020-07-10
[WIP] functional pairing_channel, good for demo, not good for anything else yet

tarikeshaq:
Alrighty so here's a draft of the [fxa-pairing-channel](https://github.com/mozilla/fxa-pairing-channel) in Rust. The purpose of this PR is to decide where this goes. There is a bunch of code here (albeit a big chunk is testing and demo code). This still has a long way to go to be ready for any real use case, but the skeleton is there.

Here's what still needs to be done:
- [ ] Write some proper documentation explaining the structure, to make it easier for people to understand the architecture.
- [ ] Proper error handling (currently I just `anyhow::bail!` everywhere)
- [ ] Testing. (The Record layer and Key Schedule are well tested, everything else has only one or two tests). This needs to test all the edge cases [The js implementation](https://github.com/mozilla/fxa-pairing-channel/blob/master/test/tlsconnection.js) tests
- [ ] Figure out how consumers can actually use this (i.e would they be expected to handle the WebSockets? what does an interface for this look like?)

I do not expect anyone should review this in detail at this stage (the change is too big for that :( ) But if anyone has insights on the general shape of it I would love to hear it! (I'll write some more documentation that can help with understanding the structure)

### DIFF
--- a/components/fxa-client/Cargo.toml
+++ b/components/fxa-client/Cargo.toml
@@ -24,6 +24,9 @@ rc_crypto = { path = "../support/rc_crypto", features = ["ece", "hawk"] }
 error-support = { path = "../support/error" }
 thiserror = "1.0"
 anyhow = "1.0"
+tungstenite = "0.11"
+bytes = "0.5"
+qr2term = "0.2.1"
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
--- a/components/fxa-client/src/pairing_channel/client_states.rs
+++ b/components/fxa-client/src/pairing_channel/client_states.rs
@@ -1,0 +1,388 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use bytes::{Buf, BufMut, BytesMut};
+
+use super::extension::{
+    Extension, PreSharedKeyExtension, PskKeyExchangeExtension, SupportedVersionsExtension,
+};
+use super::key_schedule::KeySchedule;
+use super::messages::{ClientHello, Finished, HandshakeMessage, HandshakeType};
+use super::record_layer::{RecordLayer, RecordType};
+use super::utils::{HASH_LENGTH, PRE_SHARED_KEY_ID, PSK_MODE_KE, VERSION_TLS_1_3};
+use anyhow::Result;
+use rc_crypto::{digest::Algorithm, hmac::SigningKey, rand};
+//
+// Client side State-machine for TLS Handshake Management.
+//
+// Internally, we manage the TLS connection by explicitly modelling the
+// client and server state-machines from RFC8446.  You can think of
+// these `State` objects as little plugins for the `Connection` class
+// that provide different behaviours of `send` and `receive` depending
+// on the state of the connection.
+//
+pub enum ClientState<F: FnMut(&[u8]) -> Result<()>> {
+    Invalid,
+    ClientWaitEe(ClientWaitEe<F>),
+    ClientWaitSh(ClientWaitSh<F>),
+    ClientWaitFinished(ClientWaitFinished<F>),
+    ClientConnected(ClientConnected<F>),
+}
+
+impl<F: FnMut(&[u8]) -> Result<()>> ClientState<F> {
+    #[allow(dead_code)]
+    pub fn client_init(psk: Vec<u8>, psk_id: Vec<u8>, callback: F) -> Result<Self> {
+        let client_start = ClientStart {
+            key_schedule: KeySchedule::new(),
+            record_layer: RecordLayer::new(callback),
+        };
+        let client_start = client_start.initialize(&psk, &psk_id)?;
+        Ok(ClientState::ClientWaitSh(client_start))
+    }
+
+    pub fn recv(&mut self, data: &[u8]) -> Result<(RecordType, Vec<u8>)> {
+        match self {
+            ClientState::ClientWaitEe(inner) => inner.record_layer.recv(data),
+            ClientState::ClientWaitSh(inner) => inner.record_layer.recv(data),
+            ClientState::ClientWaitFinished(inner) => inner.record_layer.recv(data),
+            ClientState::ClientConnected(inner) => inner.record_layer.recv(data),
+            _ => anyhow::bail!("Invalid state"),
+        }
+    }
+
+    pub fn recv_application_data(&mut self, data: &[u8]) -> Result<Vec<u8>> {
+        match self {
+            ClientState::ClientConnected(inner) => Ok(inner.recv_application_data(data)),
+            _ => anyhow::bail!("State not connected"),
+        }
+    }
+
+    pub fn recv_change_cipher_spec(&mut self, data: &[u8]) -> Result<()> {
+        match self {
+            ClientState::ClientWaitEe(inner) => inner.recv_change_cipher_spec(data),
+            _ => anyhow::bail!("Invalid state"),
+        }
+    }
+
+    pub fn add_to_transcript(&mut self, data: &[u8]) -> Result<()> {
+        match self {
+            ClientState::ClientWaitEe(inner) => inner.key_schedule.add_to_transcript(data),
+            ClientState::ClientWaitSh(inner) => inner.key_schedule.add_to_transcript(data),
+            ClientState::ClientWaitFinished(inner) => inner.key_schedule.add_to_transcript(data),
+            ClientState::ClientConnected(inner) => inner.key_schedule.add_to_transcript(data),
+            _ => anyhow::bail!("Invalid state"),
+        };
+        Ok(())
+    }
+
+    pub fn recv_handshake_message(self, message_bytes: &[u8]) -> Result<Self> {
+        let msg = HandshakeMessage::from_bytes(message_bytes)?;
+        match self {
+            ClientState::ClientWaitSh(inner) => Ok(ClientState::ClientWaitEe(
+                inner.recv_handshake_message(msg)?,
+            )),
+            ClientState::ClientWaitEe(inner) => Ok(ClientState::ClientWaitFinished(
+                inner.recv_handshake_message(msg)?,
+            )),
+            ClientState::ClientWaitFinished(inner) => Ok(ClientState::ClientConnected(
+                inner.recv_handshake_message(msg)?,
+            )),
+            _ => Ok(self),
+        }
+    }
+}
+
+// These states implement (part of) the client state-machine from
+// https://tools.ietf.org/html/rfc8446#appendix-A.1
+//
+// Since we're only implementing a small subset of TLS1.3,
+// we only need a small subset of the handshake.  It basically goes:
+//
+//   * send ClientHello
+//   * receive ServerHello
+//   * receive EncryptedExtensions
+//   * receive server Finished
+//   * send client Finished
+//
+// We include some unused states for completeness, so that it's easier
+// to check the implementation against the diagrams in the RFC.
+pub struct ClientStart<F: FnMut(&[u8]) -> Result<()>> {
+    key_schedule: KeySchedule,
+    record_layer: RecordLayer<F>,
+}
+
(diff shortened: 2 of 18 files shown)


components/support/rc_crypto/src/hkdf.rs
```diff
@@ -44,6 +44,28 @@ pub fn expand(prk: &hmac::SigningKey, info: &[u8], out: &mut [u8]) -> Result<()>
...
+pub fn expand_label(
+    secret: &hmac::SigningKey,
+    label: &str,
+    transcript: &[u8],
+    length: usize,
+) -> Result<Vec<u8>> {
+    let mut info: Vec<u8> = Vec::new();
+    let length_u16: u16 = length as u16;
+    info.extend_from_slice(&length_u16.to_be_bytes());
+    let st = format!("tls13 {}", label);
```
rfk, line 56: Given that this hard-codes some details that are particular to TLS, I think it should live inside the pairing-channel crate rather than here in the shared helper.

components/fxa-client/src/pairing_channel/client_states.rs
```diff
@@ -0,0 +1,388 @@
...
+//
+pub enum ClientState<F: FnMut(&[u8]) -> Result<()>> {
+    Invalid,
+    ClientWaitEe(ClientWaitEe<F>),
+    ClientWaitSh(ClientWaitSh<F>),
+    ClientWaitFinished(ClientWaitFinished<F>),
+    ClientConnected(ClientConnected<F>),
+}
+
+impl<F: FnMut(&[u8]) -> Result<()>> ClientState<F> {
```
rfk, line 33: You've got a lot of generics going on here with `F`, and IIUC it's entirely so you can propagate a callback down to the `RecordLayer`. Without thinking too hard about it, I wonder if you could simplify the code here by:

* Have `RecordLayer` hold a `Box<dyn FnMut(&[u8]) -> Result<()>` rather than being generic over the callback (if you haven't done any boxing of callbacks in rust before, [this stackoverflow answer](https://stackoverflow.com/questions/41081240/idiomatic-callbacks-in-rust) has a decent summary under "function trait objects" header).
* Have the `ClientState` accept an already-created `RecordLayer` instance rather than accepting a callback and creating a `RecordLayer` out of it.

At the top-level of your code, you would build up the connection in a sequence of layers: make the raw I/O callbacks, then wrap them in a record layer, then wrap that in the connection state-machine, then use it for communication.

(If you try that and hate it, or have already tried it, then no big deal - just a suggestion based on the fact that all the `<F>` in here made it a bit harder to understand what was going on).

components/fxa-client/src/pairing_channel/client_states.rs
```diff
@@ -0,0 +1,388 @@
...
+        // have to wait for the server to send the (empty) list of them.
+        if *message.get_type() != HandshakeType::EncryptedExtensions {
+            anyhow::bail!("invalid message");
+        }
+        // We do not support any EncryptedExtensions.
+        if !message.get_extensions().is_empty() {
+            anyhow::bail!("Unsupported extension")
+        }
+        let server_finished_transcript = self.key_schedule.get_transcript();
+        let client_wait_finished: ClientWaitFinished<F> = self.into();
```
rfk, line 270: This use of `From` is a neat trick! I think I like it, although it's probably worth a bit of scene-setting at the top of the file to help the reader know what to expect.

I wonder if this would read more clearly as `ClientWaitFinished::from(self)` than `self.into()`, may be worth trying to see what you think (and I really do mean "I wonder", not "I think this but want to sound modest about my own suggestions", just to be clear 😆)

components/fxa-client/src/pairing_channel/client_states.rs
```diff
@@ -0,0 +1,388 @@
...
+        let msg = HandshakeMessage::from_bytes(message_bytes)?;
+        match self {
+            ClientState::ClientWaitSh(inner) => Ok(ClientState::ClientWaitEe(
+                inner.recv_handshake_message(msg)?,
+            )),
+            ClientState::ClientWaitEe(inner) => Ok(ClientState::ClientWaitFinished(
+                inner.recv_handshake_message(msg)?,
+            )),
+            ClientState::ClientWaitFinished(inner) => Ok(ClientState::ClientConnected(
+                inner.recv_handshake_message(msg)?,
```
rfk, line 89: I'm lightly surprised that you have to repeat the fact that this produces a `ClientConnected` twice - once down in the type signature of `ClientWaitFinished::recv_handeshake_message`, and again here when you wrap it into the top-level enum. It's not a problem per se, more of a vague "I wonder if you're fighting against the abstraction layer" kind of thing. 

Does the outer layer here really need to be an `enum`, allowing calling code to inspect what the current state is?

An alternative factoring could be:

* Define a `ClientState` trait with all the methods like `recv_handshake_message` on it.
* Have each of `ClientStart`, `ClientWaitServerHello` etc independently impl the `ClientState` trait.
* Have a top-level object to encapsulate the current state, but it's opaque instead of an enum:
    * `inner` could be a `Box<dyn ClientState>` or similar to hold the current state.
    * calling `recv_handshake_message` on the top-level object would call it on `inner`, than potentially change `inner` based on the result

This would be roughtly the rust equivalent of having an abstract `ClientState` base class, and defining a concrete subclass for each of the different states.

Might make things better, might make things worse, but I think it would at least reduce a little bit of enum-matching boilerplate here. YMMV.

components/fxa-client/src/pairing_channel/client_states.rs
rfk: If you prefer to keep the generics here, you could also consider defining a custom trait like `SendCallback` or similar. Giving it a meaningful name could significantly improve the readability of the code.

components/fxa-client/src/pairing_channel/extension.rs
```diff
@@ -0,0 +1,555 @@
...
+//     ExtensionType extension_type;
+//     opaque extension_data<0..2^16-1>;
+//   } Extension;
+//
+// Extensions always appear inside of a handshake message,
+// and their internal structure may differ based on the
+// type of that message.
+#[derive(Debug)]
+pub struct Extension {
+    extension: Box<dyn ExtensionValue>,
```
rfk, line 492: Oh, interesting! This seems similar to what I was suggesting for dealing with `ClientState`, but here I would (naively) have been tempted to go the other way and deliberately expose `Extension` as an enum, because calling code seems likely to want to match on the type of extension.

components/fxa-client/src/pairing_channel/messages.rs
```diff
@@ -0,0 +1,672 @@
...
+            .find(|ext| ext.get_type_tag() == SUPPORTED_VERSIONS_ID)
+            .ok_or_else(|| anyhow::Error::msg("Missing supported versions"))?
+            .contains_version(VERSION_TLS_1_3)
+        {
+            anyhow::bail!("Does not support TLS1.3")
+        }
+
+        if extensions
+            .iter()
+            .any(|ext| ext.get_type_tag() == PRE_SHARED_KEY_ID)
```
rfk, line 311: Naively, I would have expected to see this matching on the `ExtensionType::PreSharedKey` enum rather than a numerical constant, so I'm curious if you tried that and whether it turned out to be awkward in practice. (You might have to derive `Copy` and `Eq`, or make the enum `repr(u16)` or something in order to get the nice `==` matching behaviour though).

components/fxa-client/src/pairing_channel/client_states.rs
```diff
@@ -0,0 +1,388 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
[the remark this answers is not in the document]
vladikoff: Yeah we were chatting about that in zoom. I think mainly the dependency for the new component would be rc_crypto

components/fxa-client/src/pairing_channel/client_states.rs
tarikeshaq: Wow that.... I swear I shoulda thought of that 😂 I spent a non-trivial amount of time **typing** the trait. Thanks Ryan!!!

components/fxa-client/src/pairing_channel/client_states.rs
tarikeshaq: Agree, sounds clearer! And ya the `From` trick allows us to ensure we're not doing any invalid transitions at compile time 😄

components/fxa-client/src/pairing_channel/client_states.rs
tarikeshaq: Hmmm Ya thought about that very briefly (Since that's how it's done in the js implementation), the reason I ended up this direction was an attempt to keep the transitions explicit and so that I don't need to implement functions for States that shouldn't have them. Although now that I think about it, it's possible to keep the explicit types and have the `Box<dyn ClientState>`. I'll experiment with it! Thanks Ryan!

components/fxa-client/src/pairing_channel/extension.rs
```diff
@@ -0,0 +1,555 @@
...
+                })?;
+                Ok(Self { modes })
+            }
+            _ => anyhow::bail!("Invalid Handshake message"),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct UnrecognizedExtension {}
```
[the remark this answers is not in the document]
tarikeshaq: Ya I need to audit my error handling 😆

components/fxa-client/src/pairing_channel/messages.rs
tarikeshaq: Aaah ya that's leftover from when I had the extension be a Map and the key was a u16, I'll change it up (I believe I do similar stuff with Messages as well, so I'll get that done too)

components/fxa-client/src/pairing_channel/client_states.rs
rfk: Heh, I find rust is full of "attractive nuisances" like this, where there's a very clever and zero-overhead abstraction that you can come up with if you just get all the types lined up correctly...or you could put it in a box and pay a trivial amount of runtime overhead and move on with your life. I find I have to periodically ask myself "could I just clone this?" and "could I just put this in a box?" or I end up off in the weeds 😆.

tarikeshaq: Haha I can relate to that. Like when I was starting this off, I had this crazy idea that it's technically possible to do the whole message parsing/writing without a single clone or heap allocations (or maybe just one fixed-length allocation).... That idea didn't last long (Although maybe in the future someone can attempt it 😃, pretty sure I watched a talk about someone doing something like that)

components/fxa-client/Cargo.toml
```diff
@@ -24,6 +24,9 @@ rc_crypto = { path = "../support/rc_crypto", features = ["ece", "hawk"] }
 error-support = { path = "../support/error" }
 thiserror = "1.0"
 anyhow = "1.0"
+tungstenite = "0.11"
+bytes = "0.5"
+qr2term = "0.2.1"
```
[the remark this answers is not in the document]
vladikoff: @tarikeshaq to add to this, Thom recently refactored the a-s repo structure. The 2 dependencies can go into an example of pairing channel usage into https://github.com/mozilla/application-services/tree/main/examples

components/fxa-client/src/pairing_channel/client_states.rs
```diff
@@ -0,0 +1,388 @@
...
+    fn send_handshake_mesage(&mut self, message: HandshakeMessage) -> Result<()> {
+        let bytes = message.to_bytes()?;
+        self.key_schedule.add_to_transcript(&bytes);
+        self.record_layer.send(RecordType::Handshake, &bytes)?;
+        Ok(())
+    }
+
+    fn recv_handshake_message(mut self, message: HandshakeMessage) -> Result<ClientConnected<F>> {
+        if *message.get_type() != HandshakeType::Finished {
+            anyhow::bail!("Inavlid message")
```
vladikoff, line 307: typo

vladikoff:
@tarikeshaq thanks for the work on this. I have stored this into this branch https://github.com/mozilla/application-services/tree/pairing_channel

PR #3343, opened 2020-07-13
Hack around nss-sys's --all-features situation and run far fewer redundant tests in CI

thomcc:
Previously we were building and running every package with --no-default-features. Each of these builds caused a recomputation of dependency feature sets, which in practice caused a pretty substantial number of them to require rebuilding a decent chunk of our dep graph.

We only have 1 package out of 35 that needs this, so I changed it to detect and only run the ones that need it, e.g. that one (sync-guid).

I added a hack for the situation with --default-features in NSS, and the fact that the test was just commented out is very frustrating, since it was trying to catch things like this (and we need the --all-feature test way more than the --no-default-features test.

The way nss-sys does this completely backwards (it's a mistake to have features that disable something as part of their function), but I don't have the energy right now to go into a gecko build system debugging session so I added a --cfg to hack around it.

I suspect that even if running coverage bumps up our CI bill, this will more than cover the difference.

Anyway, one last note is I added back the jq dependency to run our tests. It's ubiquitous, easy to install, and I don't have the `tr`/`sed` skill to figure the other way out. (@rfk can rewrite it if he cares to do so in the future).

### DIFF
--- a/automation/all_clippy_checks.sh
+++ b/automation/all_clippy_checks.sh
@@ -20,10 +20,9 @@ EXTRA_ARGS=( "$@" )
 cargo clippy --all --all-targets --all-features -- -D warnings ${EXTRA_ARGS[@]:+"${EXTRA_ARGS[@]}"}
 
 # Apparently --no-default-features doesn't work in the root, even with -p to select a specific package.
-# Instead we pull the list of individual package manifest files out of `cargo metadata` and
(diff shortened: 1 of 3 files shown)


rfk:
> (@rfk can rewrite it if he cares to do so in the future).

Just want to confirm that I have absolutely zero interest in doing this 😆

PR #3349, opened 2020-07-14
Optimize project layout to avoid bad build performance

thomcc:
I spent part of the weekend on a project to fix something that *really* bugged
me about our dev dependencies. specifically, as we get more/better tests and
examples, out build times get worse and IDEs get slower.

I spent this evening finishing that up (took longer than I expected, okay?)

## Quick overview

Most of what I did is the following:
    - Moved all examples to `/examples`
    - Moved benchmarks and slow-building tests to `/testing/separate`
    - Cleaned up some dep usage, fix cases where rebuild triggered for no reason.
    - ...

In between all that, a buncha research on the problem, and wrote a guideline for my findings and when to do these things, why some of the specific choices I made IMO are a good idea, ...

It's late, I haven't yet edited it, but here, test-faster.md is the best overview: https://github.com/mozilla/application-services/pull/3349/files#diff-b1d75bab56cc23929b4b20a7f920ce99.

TLDR for when to do this is:
- always for benchmarks,
- probably always for examples since it's easy,
- when warranted for tests.
    - I'd have liked to do it for the crazy fxa http+struct mocking tests but it's just not plausible. As a result, FxA saw the least benefit here, although in practice it should get benefits from both everything it depends on building faster, and from it triggering way fewer rebuilds of other crates.

## Goals

On for the defense of this major change. As background, my goals were:

- Reducing the minimum time needed to build crates, either by not building stuff we don't need, playing more nicely with cargo features, or making sccache work better.

    - This was a definite success. Some remaining avenues are open, but it went better than hped.

- Reducing the amount rebuilt when changing unrelated crates. For example, currently if `fxa-client` changes, we need to rebuild: `places`, `tabs`, and `logins` completely because of the way our code is organized and not because of a real dependency.

    - This was a definite success

- Reducing the time spent waiting on IDE results.

    - Never found a great way to measure this with any degree of reliability. It feels faster?

## Outcome

Some numbers for single-change-and-rebuild for a few crates. For repeatability:

- `$ cargo test -p $crate`
- Manually change lib.rs in a superficial way
- `$ cargo test -p $crate
- Write down $NUM in `Finished test [unoptimized + debuginfo] target(s) in $NUM`
- All done without sccache, on a Macbook Pro 2018 2.9GHz Intel Core i9 with 12 cores (6 cpu 12 hyper)

1. For logins:
    - before 7.69s
    - after 4.22s
    - 45% speedup

2. For webext-storage:
    - before 4.70s
    - after 2.61s
    - 44% speedup

3. For tabs:
    - before => 6.21s,
    - after => 1.49s,
    - 76% speedup

4. For places:
    - before 13.49s
    - after 7.24s
    - 46% speedup
    - you might point out that the same set of tests aren't running for places --
      this is fair, that moved to a separate crate, and if i time a build
      of the separate crate 10.11s, or 25% speedup.

5. For sync15:
    - before 5.14s
    - after 4.11s
    - 20% speedup

6. ...
7. The only "loser" is fxa-client:
    - Before 13.51s, after between 10.44s and 13.62s. Not very stable.
    - That said.... we know the reason for this.

    - We aren't able to separate heavyweight test dependencies (mockito or mockiato)

    - Most of the time here is spent in a proc macros, a known issue for fxa-client, esp. http_client.rs

    - On the bright side:
        - If fxa-client's dependencies have to rebuild, they all build much faster
        - When fxa-client is modified or otherwise needs rebuilding, logins, places, and tabs no longer require rebuilding.

(Aside: You might also say that moving the examples out is cheating too, but this isn't true -- previously we'd build everything needed for the examples just to be able to run the tests, but it wasn't building examples/benchmarks every time, just their deps (pointlessly), cargo can't tell the difference between an dev-dep an example needs and one a test needs. So I disagree)

Over all build time change:

- Time for `cargo clean; cargo build --test;` without sccache
    - Before: 127.4s (2m 7.4s)
    - After: 116.8s (1m 56.8s)
    - ~8% speedup

There was no reason really to expect this last benchmark to get better -- I was optimizing for other things, but 8% is nothing to sneeze at by any means. I would *guess* the reason it did is mostly because of increased build parallelism. But anyway, I listed the goals up front, this wasn't one of them, so I'm thrilled about that.

Related to full rebuild, this is work is essentially compatible with https://github.com/mozilla/application-services/pull/3329 which should have a more substantial impact for wall time, but admittedly only percieved wall time (it does this by skipping building the megazords and systest for builds that don't request them, among some other things).


docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+
+If you perform a `cargo test -p foo`, all dev-dependencies of `foo` must be
+compiled before `foo`'s tests can start. This includes dependencies only used
+non-test targets, such as examples or benchmarks.
+
+In an ideal world, cargo could run your tests as soon as it finished with the
+dependencies it needs for those tests, instead of waiting for your benchmark
+suite, or the arg-parser your examples use, or etc.
+
+Unfortunately, all cargo knows that these are `dev-dependencies`, and not which
```
rfk, line 19: nit: "knows **is** that"

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+dependencies, if `places` depends on `fxa-client`, all of `fxa-clients`
+dev-dependencies must be compiled, ready, and linked in at least to the `lib`
+target before we can even think about starting on `places`.
+
+This should all sound somewhat obvious, and I guess it is. We have not been
+careful about what shape the dependency graph ends up as when example code is
+taken into consideration (as it is by cargo during certain builds), and as a
+result, we have this problem. The problem is that this isn't really a problem we
+want to fix: Example code can and should depend on several different components,
+and use them together in intersting ways.
```
rfk, line 36: nit: "intersting" -> "interesting"

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
+# Testing faster: How to avoid making compile times worse by adding tests
```
rfk, line 1: I don't feel strongly about it, but I wonder if this doc might suit living as a "design doc" in the `./design/` subdirectory here.

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+
+So, because we don't really want to change the things our examples do, or make
+major architectural changes of the non-test code for something like this, we
+need to do something.
+
+## The Solution
+
+To fix this, we manually insert "cuts" into the dependency graph to help cargo
+out. That is, we pull some of these build targets (e.g. examples, benchmarks,
+tests if they cause a substantial compile overhead) into their dedicated crates
```
rfk, line 46: nit "their dedicated" -> "their own dedicated"?

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+so that:
+
+1. They can be built in parallel with eachother.
```
rfk, line 49: nit: "eachother" -> "each other"

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+- All rust examples should be put in `examples/*`.
+
+- All rust benchmarks should be put in `testing/separated/*`. See the section
+  below on how to set your benchmark up to avoid redundant compiles.
+
+- Rust tests which brings in heavyweight dependencies should be evaluated on an
+  ad-hoc basis. If you're concerned, measure how long compilation takes
+  with/without, and consider how many crates depend on the crate where the test
+  lives (e.g. a slow test in support/foo might be far worse than one in a leaf
+  crate), etc...
```
rfk, line 70: I'm reminded of the recent work @grigoryk did to make a [Fenix CI status dashboard](https://sql.telemetry.mozilla.org/dashboard/fenix-ci-status?p_w57980_time_period=3&p_w57980_with_tasks=OR%20job_type.symbol%20%3D%20%27gradle-dependencies%27%20OR%20job_type.symbol%20%3D%20%27ui-test-x86%27&p_w57985_time_period=3&p_w57985_with_ui_tests=OR%20job_type.symbol%20%3D%20%27gradle-dependencies%27%20OR%20job_type.symbol%20%3D%20%27ui-test-x86%27&p_w57988_time_period=3&p_w57988_with_tasks=OR%20job_type.symbol%20%3D%20%27ui-test-x86%27&p_w57990_time_period=3&p_w57990_with_tasks=OR%20job_type.symbol%20%3D%20%27gradle-dependencies%27%20OR%20job_type.symbol%20%3D%20%27ui-test-x86%27&p_w57991_time_period=3&p_w57991_with_tasks=OR%20job_type.symbol%20%3D%20%27ui-test-x86%27&p_w58003_task=T&p_w58003_time_period=3&p_w58004_task=ui-test-x86&p_w58004_time_period=3&p_w58006_task=gradle-dependencies&p_w58006_time_period=3&p_w58007_task=Bat&p_w58007_time_period=3&refresh=43200). Since you've put all this work into getting the build time down, is it worth us considering a little more work to help *keep* it down?

I don't know how involved it was to produce the above dashboard, but we might be able to borrow some or all of it from Fenix's setup.

docs/test-faster.md
```diff
@@ -0,0 +1,191 @@
...
+-- When you add a file, you then need to add it to it's parent module, and this
+is something Rust programmers are pretty used to.
+
+(In fact, we even replicated this behavior (for no reason) in the places
+integration tests, and added the `mod` declarations to a "controlling" parent
+module -- It seems weird to be in an environment where this *isn't* required)
+
+So, that's why. This way, we make it *way* less likely that you add a test file
+to some directory, and have it get ignored because you didn't realize that in
+this one folder, you need to add a `mod mytest` into a neighboring tests.rs.
```
rfk, line 189: Thank you for this excellent document!

Cargo.toml
```diff
@@ -36,6 +36,9 @@ members = [
     "megazords/lockbox",
     "testing/sync-test",
     "tools/protobuf-gen",
+
+    "examples/*/",
+    "testing/separated/*/",
```
rfk, line 41: Is this going to interact badly with the changes over in https://github.com/mozilla/application-services/pull/3329?

components/support/cli/Cargo.toml
```diff
@@ -12,3 +12,9 @@ log = "0.4"
 sync15 = { path = "../../sync15" }
 url = "2.1"
 webbrowser = "0.5"
+# disable regex feature, as it's very costly for compile time and very rarely
+# used.
+[dependencies.env_logger]
+version = "0.7"
+default-features = false
+features = ["termcolor", "atty", "humantime"]
```
rfk, line 20: I wonder if it would be worth moving this `cli` crate out of the main `components` directory and putting it somewhere under `examples/`, so then when you go to use it you're reminded that `examples/` exists.

testing/separated/README.md
```diff
@@ -0,0 +1,4 @@
+# Separated tests
+
+These are pulled out for compile-time reasons. See `docs/test-faster.md` for an
+in-depth explanation and recommended guide for how to add things.
```
rfk, line 4: 👍 cross-references++

docs/test-faster.md
thomcc: Filed a followup

Issue #2871, opened 2020-03-25
Use the new `TimingDistribution.measure` helper from glean.


rfk:
Some of our Kotlin code currently defines a custom helper method on glean's `TimingDistribution` metric to make to make it easier to measure the execution of a closure, e.g. [this definition](https://github.com/mozilla/application-services/blob/28a58972754daceda9e96382898880f1a5e9be85/components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt#L404) in this logins component.

As of [Bug 1598993](https://bugzilla.mozilla.org/show_bug.cgi?id=1598993) glean now supports such a method natively. We should use it instead of our own artisanal version.  In practice this looks like:

1. Wait until the new method is in a released version of glean and android-components
2. Find places where we implement our own `TimingDistributionMetricType.measure` method and delete them.
3. Make sure the Kotlin tests all still pass, fixing up for any compilation problems that might have been introduced by subtle API differences.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1010)
┆Sprint: Backlog

vladikoff:
The PR to add the version of glean that supports this is in https://github.com/mozilla-mobile/android-components/pull/6455

PR #3353, opened 2020-07-15
Use the Glean Gradle Plugin for the Logins component

Dexterp37:
Fixes https://github.com/mozilla/application-services/issues/2871

This allows for an improved and future-proof Glean experience. It enables "offline" builds. Note that this is a blocker for
the TOR browser. See bug 1651660 for more context.

It additionally updates the in-tree docs, which were out of date.


build.gradle
```diff
@@ -5,9 +5,10 @@ buildscript {
     ext.jna_version = '5.2.0'
     ext.android_gradle_plugin_version = '3.6.0'
 
-    // N.B. try to keep these two in sync
-    ext.android_components_version = '24.0.0'
-    ext.glean_version = '21.3.0'
+    // Since the Glean version depends on the Android components version,
+    // it is very important to use a modern version of Glean and, ideally,
+    // let this come the embedder product itself.
```
mdboom, line 10: I think you meant ... ?  If I'm wrong, maybe just clarify as you see fit?

```suggestion
    // let this embed the product itself.
```

travis79: I think it was supposed to be 
```suggestion
    // let this come from the embedding product itself.
```

Dexterp37: Whoops, my Italianenglish failed me today. @travis79 is correct :D Sorry @mdboom!

Dexterp37:
Hey @rfk ! This is now ready for review :-) Heads up: there are at least 3 things that require a closer look:

1. I bumped the required A-C version to the most recent one (3b7ccd9).
2. As a consequence of that, I had to add support for the jvm target 1.8 (a6748af).
3. Since the Glean SDK now provides a `TimingDistributionMetricType.measure` function, the one you defined was being shadowed. I removed it (2b711d2). The behaviour [is _slightly_ different](https://searchfox.org/glean/rev/258c8d2c43c37fa91b515edb216bf648d3ba6077/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt#117-130): if an exception is thrown by the measured function, the timer cancelled **but the exception is propagated** and not swallowed by the Glean SDK.

build.gradle
```diff
@@ -5,15 +5,19 @@ buildscript {
     ext.jna_version = '5.2.0'
     ext.android_gradle_plugin_version = '3.6.0'
 
-    // N.B. try to keep these two in sync
-    ext.android_components_version = '24.0.0'
-    ext.glean_version = '21.3.0'
```
Dexterp37, line 10: Note: the `glean_version` is now automatically published by the Glean Gradle plugin :-) We'll take care of keeping things in sync, no need for consuming products to do so!

build.gradle
```diff
@@ -44,6 +48,9 @@ buildscript {
 
         classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.12'
 
+        // Use the Glean Gradle plugin.
+        classpath "org.mozilla.components:tooling-glean-gradle:$android_components_version"
```
mdboom, line 52: I guess I didn't realize this would apply to other `build.gradle` files, but given that this is working, it must.  (Our docs sort of assume it's all together in the same `build.gradle` and maybe there's an update we should make there...)

components/logins/android/metrics.yaml
```diff
@@ -2,31 +2,36 @@
...
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# This file defines the metrics that will be gathered for the "logins" storage component.
-# These are emitted for all users of the component.  Additional metrics specific to the
-# *syncing* of logins are defined in a separate "sync_ping" package.
+# This file defines the metrics that will be gathered for the "logins"
+# storage component.
+# These are emitted for all users of the component.  Additional metrics
+# specific to the *syncing* of logins are defined in a separate "sync_ping"
+# package.
```
mdboom, line 9: Just a comment for visibility for the a-s team: These whitespace-only changes here are due to the fact that `glean_parser` now runs `yamllint` over the input.

build.gradle
Dexterp37: Maybe this is also due to the way A-S has split the Gradle files: in all the projects we integrated with the root build.gradle contained the majority of the directives. It is not the same here :)

rfk:
> the timer cancelled but the exception is propagated and not swallowed by the Glean SDK

Thank you, this seems like a definite improvement! 😅

build.gradle
```diff
@@ -5,15 +5,19 @@ buildscript {
     ext.jna_version = '5.2.0'
     ext.android_gradle_plugin_version = '3.6.0'
 
-    // N.B. try to keep these two in sync
-    ext.android_components_version = '24.0.0'
-    ext.glean_version = '21.3.0'
+    // Since the Glean version depends on the Android components version,
+    // it is very important to use a modern version of Glean and, ideally,
+    // let this come from the embedding product itself.
```
rfk, line 10: micro-nit: it's not clear to me if a future reader of this file will know how to interpret this comment, since there's now nothing in the surrounding code about glean. I wonder if it would be more likely to serve its intended purpose if moved down to the part where we use the glean gradle plugin.

megazords/lockbox/android/build.gradle
```diff
@@ -29,6 +29,19 @@ android {
...
+    compileOptions {
+        sourceCompatibility rootProject.ext.build.jvmTargetCompatibility
+        targetCompatibility rootProject.ext.build.jvmTargetCompatibility
+    }
+
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+        kotlinOptions {
+            jvmTarget = rootProject.ext.build.jvmTargetCompatibility
+        }
+    }
```
rfk, line 44: It's a shame you need to repeat this so often, but that's a problem with out gradle setup, not your PR. Thanks for adding all these in!

Dexterp37: My pleasure!

build.gradle
Dexterp37: Good point. I moved this down, where I declare the glean plugin dependency!

Issue #3086, opened 2020-05-06
fxa-client:  Synthesize placeholder AET identifiers for existing accounts


rfk:
(This the rust-components version of a desktop bug filed as https://bugzilla.mozilla.org/show_bug.cgi?id=1635657)

#3085 will adds support for storing `ecosystem_user_id` and `ecosystem_anon_id` values as part of the persisted account state, and for learning the value of `ecosystem_anon_id` from the FxA profile server.

However, it's only possible to calculate `ecosystem_user_id` and `ecosystem_anon_id` during an FxA login flow, when we have access to their password and can derive the necessary secrets from their FxA master key. That means users who are already signed in to Firefox can't start sending Account Ecosystem Telemetry pings until they sign in again (either on some other device, or on this device after a disconnection event).

To help bootstrap the AET data set, let's synthesize placeholder identifiers for these users. If we need the `ecosystem_anon_id` and it's not in the persisted account state, then we would:

1) Try to fetch `ecosystem_anon_id` from the FxA profile server.
2) If `ecosystem_anon_id` is not present in the FxA profile response, then synthesize a new one:
    1) Generate `ecosystem_user_id` as 32 strong random bytes.
    2) Fetch the `/.well-known/fxa-client-configuration` file from FxA and extract the `ecosystem_telemetry_public_key` field, a JWK.
    3) Encrypt `ecosystem_user_id` to the `ecosystem_telemetry_public_key` as specified in [the tech spec for AET identifiers](https://docs.google.com/document/d/1rR3uJG8sVtow4plYu6M5Jp5e5zOSrLzdkUx8BdO8kOM/).
    4) POST the resulting `ecosystem_anon_id` to the FxA profile server for storage, using concurrency-control mechanisms to help avoid races. If a concurrent update to the profile data is discovered, go back to (1).
    5) Persist the resulting `ecosystem_user_id` and `ecosystem_anon_id` values to local state.

This allows the device to start sending AET pings immediately. If and when the user does sign in again and FxA is able to calculate their real `ecosystem_user_id`, then we will discover it via the normal process of updating user profile data. The AET data pipeline will take care of correlating the temporary id with the new one, in the same way it handles `ecosystem_user_id` changing after a password reset.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1242)
┆Story Points: 8
┆Epic: Sync POC for Accounts Ecosystem Telemetry
┆Sprint: SYNC - end 2020-07-31

[...]

PR #3358, opened 2020-07-15
Expose ecosystem_user_id

### LINKED ISSUE #3085: fxa-client: Fetch and store Account Ecosystem Telemetry identifiers in FxA profile data
(This is the rust-components version of a Desktop bug filed as https://bugzilla.mozilla.org/show_bug.cgi?id=1635656)

The Account Ecosystem Telemetry project needs to include a unique user identifier in its telemetry ping, but without being able to link that identifier back to the user's actual Firefox Account. The proposal for obtaining such an identifier is described in detail in [Account Ecosystem Telemetry Identifiers](https://docs.google.com/document/d/1rR3uJG8sVtow4plYu6M5Jp5e5zOSrLzdkUx8BdO8kOM/) and is *mostly* the responsibility of the Firefox Accounts service. From the perspective of an FxA client application like Firefox the key points are:

* Each user has an `ecosystem_user_id` which is derived from their FxA master key. It's OK for client code to know this value but it needs to keep it to itself.
* For recording in AET telemetry pings, the `ecosystem_user_id` must be encrypted to a public key that's managed by the telemetry pipeline, resulting in a value called `ecosystem_anon_id`. We need this value in order to send AET telemetry pings.
* When the user signs in to their Firefox Account, the FxA services will use their password to unlock the master key, derive `ecosytem_user_id`, encrypt it to produce `ecosystem_anon_id`, and store that in their account profile data for us to fetch.

To support all this we need to add the following in fxa-client:

* Support for storing new fields `ecosystem_user_id` and `ecosystem_anon_id` as part of the persisted account state.
* Support for obtaining `ecosystem_anon_id` from the FxA server as part of the user's profile data, and refreshing it as part of refreshing the profile data.
* A public API for exposing the `ecosystem_anon_id` value so that AET can include it in telemetry pings.

The ability to learn the value of `ecosystem_user_id` is not part of this bug, but I think it's worth including the storage for it as part of this work.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1241)
┆Story Points: 2
┆Epic: Sync POC for Accounts Ecosystem Telemetry
┆Sprint: SYNC - end 2020-07-31

eoger:
Fixes #3085.

### DIFF
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
@@ -210,6 +210,15 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
         return getProfile(false)
     }
 
+    fun getEcosystemAnonId(): String? {
+        return nullableRustCallWithLock { e ->
+            LibFxAFFI.INSTANCE.fxa_get_ecosystem_anon_id(
+                this.handle.get(),
+                e
+            )
+        }?.getAndConsumeRustString()
+    }
+
     /**
      * Fetches the token server endpoint, for authenticating to Firefox Sync via OAuth.
      *
--- a/components/fxa-client/ios/FxAClient/FxAccountManager.swift
+++ b/components/fxa-client/ios/FxAClient/FxAccountManager.swift
@@ -237,6 +237,18 @@ open class FxAccountManager {
         }
     }
 
+    /// Fetch the AET (Account Ecosystem Telemetry) anonymized user ID if available.
+    public func getEcosystemAnonId(completionHandler: @escaping (Result<String?, Error>) -> Void) {
+        DispatchQueue.global().async {
+            do {
+                let ecosystemAnonId = try self.requireAccount().getEcosystemAnonId()
+                DispatchQueue.main.async { completionHandler(.success(ecosystemAnonId)) }
+            } catch {
+                DispatchQueue.main.async { completionHandler(.failure(error)) }
+            }
+        }
+    }
+
     /// Get the account management URL.
     public func getManageAccountURL(
         entrypoint: String,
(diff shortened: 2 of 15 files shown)


components/fxa-client/src/telemetry.rs
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub use crate::http_client::ProfileResponse as Profile;
+use crate::{error::*, scopes, util, CachedResponse, FirefoxAccount};
+
+impl FirefoxAccount {
+    pub fn get_ecosystem_anon_id(&mut self) -> Result<Option<String>> {
+        let profile = self.get_profile(false)?;
```
eoger, line 10: Note that I am not storing the ID in the FxA state separately, which should not matter? (well, it might for offline cases maybe). I'm happy to move it there if you think it's valuable.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt
```diff
@@ -210,6 +210,15 @@ class FirefoxAccount(handle: FxaHandle, persistCallback: PersistCallback?) : Aut
         return getProfile(false)
     }
 
+    fun getEcosystemUserId(): String? {
```
rfk, line 213: The name here should be `getEcosystemAnonId`; the `ecosystemUserId` field is different and we're not ready to expose that to consumers directly yet.

components/fxa-client/ios/FxAClient/FxAccountManager.swift
```diff
@@ -237,6 +237,18 @@ open class FxAccountManager {
         }
     }
 
+    /// Fetch the AET (Account Ecosystem Telemetry) user ID if available.
```
rfk, line 240: nit: "user id" => "anonymized user id" or similar here. (I'm being pedantic here because there are a lot of identifiers in this system with different properties and it's important that we be able to keep the straight in our heads).

components/fxa-client/src/telemetry.rs
rfk: This seems fine, we'll probably revisit state management a bit in #3086 but can cross that bridge when we get there.

PR #3374, opened 2020-07-17
Implement rc_crypto backed Rng for rand

eoger:
While working on another patch, I found myself needing to pick an element at random in an array.
Of course, I could have devised an algorithm myself, but let's be honest I went to google right away and typed "random vec element rust" and picked the first answer. Great! `rand` does that. Oh wait, we can't use it for "serious stuff" because it's not backed by NSS. Well, now it is.

```
use rand::{prelude::*, seq::SliceRandom};

let mut rng = rand_rccrypto::RcCryptoRng;
let choices = [1, 2, 4, 8, 16, 32];
let rand_item = choices.choose(&mut rng);
```

We now have a higher-level API for randomness.

### DIFF
--- a/components/support/rand_rccrypto/src/lib.rs
+++ b/components/support/rand_rccrypto/src/lib.rs
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub use rand_core;
+use rand_core::{impls, CryptoRng, Error, RngCore};
+
+pub struct RcCryptoRng;
+
+impl RngCore for RcCryptoRng {
+    fn next_u32(&mut self) -> u32 {
+        impls::next_u32_via_fill(self)
+    }
+
+    fn next_u64(&mut self) -> u64 {
+        impls::next_u64_via_fill(self)
+    }
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.try_fill_bytes(dest).unwrap()
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Error> {
+        rc_crypto::rand::fill(dest).map_err(Error::new)
+    }
+}
+
+// NSS's `PK11_GenerateRandom` is considered a CSPRNG.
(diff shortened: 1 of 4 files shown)


components/support/guid/Cargo.toml
```diff
@@ -8,11 +8,14 @@ edition = "2018"
 [dependencies]
 rusqlite = { version = "0.23.1", optional = true }
 serde = { version = "1", optional = true }
-rand = { version = "0.7", optional = true }
+# We do not need the rand default features as we provide
+# our own Rng implementation backed by rc_crypto.
+rand = { version = "0.7", optional = true, default-features = false, features = ["std"] }
+rand_rccrypto = { path = "../rand_rccrypto", optional = true }
```
rfk, line 14: From conversation in slack, it sounded like @thomcc had some hesitations about changing the dependency here. Is that a blocker?

components/support/rand_rccrypto/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn next_u32(&mut self) -> u32 {
+        impls::next_u32_via_fill(self)
+    }
+
+    fn next_u64(&mut self) -> u64 {
+        impls::next_u64_via_fill(self)
+    }
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.try_fill_bytes(dest).unwrap()
```
rfk, line 20: nit: we should at the very least panic with a descriptive message here. The `OsRng` impl is an example of more elaborate error handling, which includes a retry loop:

  * https://docs.rs/rand/0.5.0/src/rand/rngs/os.rs.html#120

But that might be overkill for our purposes.

components/support/guid/Cargo.toml
thomcc: I'd rather not land this using rc_crypto. In https://github.com/mozilla/application-services/pull/2446 we moved away from using rc_crypto, and I would like a more compelling reason than "it shows off using our impl of an RNG" to move back.

It's worth noting that while this probably doesn't matter here, the answer of "we definitely do not use rc_crypto in desktop firefox" makes the evaluation of an issue like https://github.com/mozilla/application-services/pull/3351 (which otherwise could be an exploitable vuln) much simpler.

eoger: This is not a hill I am willing to die on, I'll remove the`guid` crate change 👍

PR #3380, opened 2020-07-20
Factor-out jwcrypto bits

eoger:
I was trying to understand how the newest changes to our JWstuff worked, so as usual I refactored it :) Specifically, this removes the JWE bits from the Scoped Keys module to go on its own.

### DIFF
--- a/components/fxa-client/src/error.rs
+++ b/components/fxa-client/src/error.rs
@@ -12,12 +12,6 @@ pub enum ErrorKind {
     #[error("Unknown OAuth State")]
     UnknownOAuthState,
 
-    #[error("The client requested keys alongside the token but they were not included")]
-    TokenWithoutKeys,
-
-    #[error("Login state needs to be Married for the current operation")]
-    NotMarried,
-
     #[error("Multiple OAuth scopes requested")]
     MultipleScopesRequested,
 
@@ -39,20 +33,11 @@ pub enum ErrorKind {
     #[error("No stored current device id")]
     NoCurrentDeviceId,
 
-    #[error("Could not find a refresh token in the server response")]
-    RefreshTokenNotPresent,
-
-    #[error("Action requires a prior device registration")]
-    DeviceUnregistered,
-
     #[error("Device target is unknown (Device ID: {0})")]
     UnknownTargetDevice(String),
 
     #[error("Unrecoverable server error {0}")]
     UnrecoverableServerError(&'static str),
-
-    #[error("Invalid OAuth scope value {0}")]
-    InvalidOAuthScopeValue(String),
 
     #[error("Illegal state: {0}")]
     IllegalState(&'static str),
@@ -63,44 +48,14 @@ pub enum ErrorKind {
     #[error("Send Tab diagnosis error: {0}")]
     SendTabDiagnosisError(&'static str),
 
-    #[error("Empty names")]
-    EmptyOAuthScopeNames,
-
-    #[error("Key {0} had wrong length, got {1}, expected {2}")]
-    BadKeyLength(&'static str, usize, usize),
-
     #[error("Cannot xor arrays with different lengths: {0} and {1}")]
     XorLengthMismatch(usize, usize),
-
-    #[error("Audience URL without a host")]
-    AudienceURLWithoutHost,
 
     #[error("Origin mismatch")]
     OriginMismatch,
 
-    #[error("JWT signature validation failed")]
-    JWTSignatureValidationFailed,
-
-    #[error("ECDH key generation failed")]
-    KeyGenerationFailed,
-
-    #[error("Public key computation failed")]
-    PublicKeyComputationFailed,
-
     #[error("Remote key and local key mismatch")]
     MismatchedKeys,
-
-    #[error("Key import failed")]
-    KeyImportFailed,
-
-    #[error("AEAD open failure")]
-    AEADOpenFailure,
-
-    #[error("Random number generation failure")]
-    RngFailure,
-
-    #[error("HMAC mismatch")]
-    HmacMismatch,
 
     #[error("Client: {0} is not allowed to request scope: {1}")]
     ScopeNotAllowed(String, String),
@@ -142,6 +97,9 @@ pub enum ErrorKind {
     #[error("JSON error: {0}")]
     JsonError(#[from] serde_json::Error),
 
+    #[error("JWCrypto error: {0}")]
+    JwCryptoError(#[from] jwcrypto::JwCryptoError),
+
     #[error("UTF8 decode error: {0}")]
     UTF8DecodeError(#[from] string::FromUtf8Error),
 
@@ -171,6 +129,7 @@ error_support::define_error! {
         (HexDecodeError, hex::FromHexError),
         (Base64Decode, base64::DecodeError),
         (JsonError, serde_json::Error),
+        (JwCryptoError, jwcrypto::JwCryptoError),
         (UTF8DecodeError, std::string::FromUtf8Error),
         (RequestError, viaduct::Error),
         (UnexpectedStatus, viaduct::UnexpectedStatus),
--- a/components/fxa-client/src/http_client.rs
+++ b/components/fxa-client/src/http_client.rs
@@ -546,7 +546,7 @@ pub fn get_keys_bundle(config: &Config, hkdf_sha256_key: &[u8]) -> Result<Vec<u8
     let bundle = hex::decode(
         &resp["bundle"]
             .as_str()
-            .ok_or_else(|| ErrorKind::KeyGenerationFailed)?,
+            .ok_or_else(|| ErrorKind::UnrecoverableServerError("bundle not present"))?,
     )?;
     Ok(bundle)
 }
--- a/components/fxa-client/src/scoped_keys.rs
+++ b/components/fxa-client/src/scoped_keys.rs
@@ -74,6 +74,7 @@ impl ScopedKeysFlow {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use jwcrypto::JwkKeyParameters;
     use rc_crypto::agreement::{KeyPair, PrivateKey};
 
     #[test]
@@ -99,7 +100,7 @@ mod tests {
         let key_pair = KeyPair::from(private_key).unwrap();
         let flow = ScopedKeysFlow::from_static_key_pair(key_pair).unwrap();
         let jwk = flow.get_public_key_jwk().unwrap();
-        let Jwk::EC(ec_key_params) = jwk;
+        let JwkKeyParameters::EC(ec_key_params) = jwk.key_parameters;
         assert_eq!(ec_key_params.crv, "P-256");
         assert_eq!(
             ec_key_params.x,
--- a/components/fxa-client/src/scoped_keys.rs
+++ b/components/fxa-client/src/scoped_keys.rs
@@ -3,13 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::{error::*, FirefoxAccount};
-use rc_crypto::{
-    aead, agreement,
-    agreement::{Ephemeral, InputKeyMaterial, KeyPair},
-    digest, rand,
-};
-use serde_derive::*;
-use serde_json::{self, json};
+use jwcrypto::{self, DecryptionParameters, Jwk};
+use rc_crypto::{agreement, agreement::EphemeralKeyPair};
+use serde_derive::{Deserialize, Serialize};
 
 impl FirefoxAccount {
     pub(crate) fn get_scoped_key(&self, scope: &str) -> Result<&ScopedKey> {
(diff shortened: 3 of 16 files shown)


components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+    A256GCM,
+}
+
+impl EncryptionAlgorithm {
+    fn algorithm_id(&self) -> &'static str {
+        match self {
+            Self::A256GCM => "A256GCM",
+        }
+    }
+}
```
eoger, line 51: This is the start of a more generic JWE API, even thought we only support our use case (ECDH-ES w/ direct key agreement and A256GCM content). It kinda sucks and is probably over-engineered, but it's a start.

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+        buf.extend_from_slice(apu.as_bytes());
+        buf.extend_from_slice(&(apv.len() as u32).to_be_bytes());
+        buf.extend_from_slice(apv.as_bytes());
+        buf.extend_from_slice(&256u32.to_be_bytes());
+        digest::digest(&digest::SHA256, &buf)
+    })?;
+    Ok(secret)
+}
+
+pub fn extract_pub_key_jwk(key_pair: &KeyPair<Ephemeral>) -> Result<Jwk> {
```
eoger, line 245: Unchanged

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+    let peer_public_key = get_peer_public_from_jwk(peer_jwk)?;
+    // Note: We don't support key-wrapping, but if we did `algorithm_id` would be `alg` instead.
+    let algorithm_id = protected_header.enc.algorithm_id();
+    let ikm = private_key.agree(&agreement::ECDH_P256, &peer_public_key)?;
+    let apu = protected_header.apu.as_deref().unwrap_or_default();
+    let apv = protected_header.apv.as_deref().unwrap_or_default();
+    get_secret_from_ikm(ikm, &apu, &apv, &algorithm_id)
+}
+
+fn get_secret_from_ikm(
```
eoger, line 220: Unchanged

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+    if y.len() != (256 / 8) {
+        return Err(ErrorKind::UnrecoverableServerError("Y must be 32 bytes long.").into());
+    }
+    let mut peer_pub_key: Vec<u8> = vec![0x04];
+    peer_pub_key.extend_from_slice(&x);
+    peer_pub_key.extend_from_slice(&y);
+    Ok(peer_pub_key)
+}
+
+fn derive_shared_secret(
```
eoger, line 205: Extracted this method from encrypt/decrypt ops.

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+            let nonce = aead::Nonce::try_assume_unique_for_key(&encryption_algorithm, &iv)?;
+            let aad = aead::Aad::from(segments[0].as_bytes());
+            let plaintext = aead::open(&opening_key, nonce, aad, &ciphertext_and_tag)
+                .map_err(|_| ErrorKind::AEADOpenFailure)?;
+            Ok(String::from_utf8(plaintext.to_vec())?)
+        }
+    }
+}
+
+fn get_peer_public_from_jwk(jwk: &Jwk) -> Result<Vec<u8>> {
```
eoger, line 184: Unchanged

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+    let tag = base64::encode_config(auth_tag, base64::URL_SAFE_NO_PAD);
+    let ciphertext = base64::encode_config(ciphertext, base64::URL_SAFE_NO_PAD);
+    let iv = base64::encode_config(iv, base64::URL_SAFE_NO_PAD);
+    Ok(format!(
+        "{}.{}.{}.{}.{}",
+        header, encrypted_key, iv, ciphertext, tag
+    ))
+}
+
+pub fn decrypt_jwe(jwe: &str, encryption_params: DecryptionParameters) -> Result<String> {
```
eoger, line 124: Moved the `segments` de-serialization at the beginning, use `derive_shared_secret` and "generified" the API.

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
tarikeshaq, line 1: I'm guessing the reason this module is still in `fxa-client` as opposed to `rc_crypto` is because it's only used here right? Seems okay to me, although it doesn't seem `fxa-client`y anymore 😅

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+        base64::URL_SAFE_NO_PAD,
+    )?)?;
+    let encrypted_key = match segments[1].is_empty() {
+        true => None,
+        false => Some(base64::decode_config(
+            &segments[1],
+            base64::URL_SAFE_NO_PAD,
+        )?),
+    };
+    let iv = base64::decode_config(&segments[2], base64::URL_SAFE_NO_PAD)?;
```
tarikeshaq, line 138: This is not relevant to your changes themselves, but a general observation, what are the chances that this function is called with a malformed JWE, causing the `[]` operator to panic?  (and is that okay)

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,261 @@
...
+                base64::encode_config(additional_data.as_bytes(), base64::URL_SAFE_NO_PAD);
+            let additional_data = additional_data.as_bytes();
+            let aad = aead::Aad::from(additional_data);
+            let mut iv: Vec<u8> = vec![0; 12];
+            rand::fill(&mut iv)?;
+            let nonce = aead::Nonce::try_assume_unique_for_key(encryption_algorithm, &iv)?;
+            let mut encrypted = aead::seal(&sealing_key, nonce, aad, data)?;
+
+            let tag_idx = encrypted.len() - ((128 + 7) >> 3);
+            let tag: Vec<u8> = encrypted.drain(tag_idx..).collect();
```
tarikeshaq, line 102: TIL about drain 😄

components/fxa-client/src/jwcrypto.rs
eoger: Yeah that's a good point, alternatively we could create a `jwcrypto` crate in support/.

components/fxa-client/src/jwcrypto.rs
eoger: I'll add a check.

components/fxa-client/src/jwcrypto.rs
eoger: `Vec::split_off` is even better ;)

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    // Assemble JWE
+    let encrypted_key = encrypted_key
+        .map(|k: Vec<u8>| base64::encode_config(&k, base64::URL_SAFE_NO_PAD))
+        .unwrap_or_default();
+    let header = serde_json::to_string(&protected_header)?;
+    let header = base64::encode_config(header.as_bytes(), base64::URL_SAFE_NO_PAD);
+    let tag = base64::encode_config(auth_tag, base64::URL_SAFE_NO_PAD);
+    let ciphertext = base64::encode_config(ciphertext, base64::URL_SAFE_NO_PAD);
+    let iv = base64::encode_config(iv, base64::URL_SAFE_NO_PAD);
```
rfk, line 118: pico-nit: I was momentarily alarmed when you encoded the tag before the ciphertext, but I see you're joining them together in the correct order below. It may be worth b64-encoding these items in the same order in which they'll appear in the final JWE, for a tiny bit less mental load when scanning this code.

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,265 @@
...
+            let secret = derive_shared_secret(&protected_header, local_key_pair, peer_jwk)?;
+
+            // Part 2: decrypt the payload
+            if encrypted_key.is_some() {
+                return Err(ErrorKind::UnrecoverableServerError(
+                    "The Encrypted Key must be empty.",
+                )
+                .into());
+            }
+            if auth_tag.len() != (128 / 8) {
```
rfk, line 167: As above, you could use the `tag_len` property of the AEAD config itself rather than a constant.

components/fxa-client/src/scoped_keys.rs
```diff
@@ -86,152 +66,25 @@ impl ScopedKeysFlow {
...
-        let header = JweHeader {
-            alg: "ECDH-ES".to_string(),
-            enc: "A256GCM".to_string(),
-            epk: epk_jwk,
-            apu: None,
-            apv: None,
+        let params = EncryptionParameters::ECDH_ES {
+            enc: EncryptionAlgorithm::A256GCM,
+            peer_jwk: jwk,
+            local_key_pair: self.key_pair,
```
rfk, line 85: This seems risky! `self.key_pair` here is designed to be the jwk that we send in to a scoped-keys oauth flow, but here we're using it as the ephemeral key for generating a JWE. I think *in practice* we would never find ourselves re-using this key but there's nothing in the way the code is set up to ensure that.

Please refactor so that we don't pass in a keypair here, and instead `jwcrypto::encrypt_to_jwe` generates a fresh ephemeral keypair internally.

components/fxa-client/src/jwcrypto.rs
```diff
@@ -0,0 +1,265 @@
...
+
+/// Encrypt and serialize data in the JWE compact form.
+pub fn encrypt_to_jwe(data: &[u8], encryption_params: EncryptionParameters) -> Result<String> {
+    let (encrypted_key, protected_header, iv, ciphertext, auth_tag) = match encryption_params {
+        EncryptionParameters::ECDH_ES {
+            enc,
+            local_key_pair,
+            peer_jwk,
+        } => {
+            let local_public_key = extract_pub_key_jwk(&local_key_pair)?;
```
rfk, line 78: Rather than taking `epk` as an input parameter here, I think this method should generate a fresh keypair internally. This is both safer (since we can ensure that the key really is only used a single time) and simpler for the caller (since it doesn't have to figure out how to generate such a key).

tarikeshaq: This would also allow this https://github.com/mozilla/application-services/blob/445c58fe8fb2cb612fe4a3b50fcfdc89eb1a024b/components/fxa-client/src/oauth.rs#L234-L235 to no longer have to depend on creating a "ScopedKeysFlow" as well!

components/fxa-client/src/error.rs
```diff
@@ -12,12 +12,6 @@ pub enum ErrorKind {
     #[error("Unknown OAuth State")]
     UnknownOAuthState,
 
-    #[error("The client requested keys alongside the token but they were not included")]
```
eoger, line 15: Nit: Removed a bunch of un-used errors.

components/support/jwcrypto/src/ec.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub x: String,
+    pub y: String,
+}
+
+pub(crate) fn encrypt_to_jwe(
+    data: &[u8],
+    encryption_params: EncryptionParameters,
+) -> Result<CompactJwe> {
+    let EncryptionParameters::ECDH_ES { enc, peer_jwk } = encryption_params;
+    let local_key_pair = EphemeralKeyPair::generate(&agreement::ECDH_P256)?;
```
eoger, line 29: As agreed, we generate the ephemeral key pair in the `encrypt` operation.

components/support/jwcrypto/src/ec.rs
```diff
@@ -0,0 +1,193 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
eoger, line 1: Factored out all the EC specific code in a separate module.

components/support/jwcrypto/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "kty")]
+pub enum Jwk {
+    #[serde(alias = "ec")]
+    EC(ec::ECKeysParameters),
+}
+
+#[derive(Debug)]
+pub struct CompactJwe {
```
eoger, line 71: It took a bit of work to make it usable everywhere, but making `CompactJwe` a type was worth the trouble.

components/support/jwcrypto/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        assert!(self.jwe_segments.len() == 5);
+        self.jwe_segments.join(".")
+    }
+}
+
+/// Encrypt and serialize data in the JWE compact form.
+pub fn encrypt_to_jwe(data: &[u8], encryption_params: EncryptionParameters) -> Result<String> {
+    let jwe = match encryption_params {
+        EncryptionParameters::ECDH_ES { .. } => ec::encrypt_to_jwe(data, encryption_params)?,
+    };
```
eoger, line 169: Nice cleanup here since we defer all the crypto stuff to the `ec` module. (did I pull too much?)

components/fxa-client/src/http_client.rs
```diff
@@ -543,11 +543,7 @@ pub fn get_keys_bundle(config: &Config, hkdf_sha256_key: &[u8]) -> Result<Vec<u8
     let req = HawkRequestBuilder::new(Method::Get, keys_url, hkdf_sha256_key).build()?;
     let client = Client::new();
     let resp: serde_json::Value = client.make_request(req)?.json()?;
-    let bundle = hex::decode(
-        &resp["bundle"]
-            .as_str()
-            .ok_or_else(|| ErrorKind::KeyGenerationFailed)?,
-    )?;
+    let bundle = hex::decode(&resp["bundle"].as_str().unwrap())?;
```
rfk, line 546: Why the switch to `unwrap` here? It seems like continuing to return some sort of error (even if it's a less specific one) would be better than panicking.

components/support/jwcrypto/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+impl FromStr for CompactJwe {
+    type Err = JwCryptoError;
+    fn from_str(str: &str) -> Result<Self> {
+        let jwe_segments: Vec<String> = str.split('.').map(|s| s.to_owned()).collect();
+        if jwe_segments.len() != 5 {
+            return Err(JwCryptoError::DeserializationError);
+        }
+        Ok(Self { jwe_segments })
+    }
+}
```
rfk, line 168: Oh, yes, very nice!

components/support/jwcrypto/src/lib.rs
rfk: This looks good to me 👍

components/support/jwcrypto/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+        DecryptionParameters::ECDH_ES { .. } => ec::decrypt_jwe(&jwe, decryption_params)?,
+    })
+}
+
+#[test]
+fn test_encrypt_decrypt_jwe_ecdh_es() {
+    use rc_crypto::agreement;
+    let key_pair = EphemeralKeyPair::generate(&agreement::ECDH_P256).unwrap();
+    let jwk = ec::extract_pub_key_jwk(&key_pair).unwrap();
+    let data = b"The big brown fox jumped over... What?";
```
rfk, line 198: I assume it jumped over bobo...

PR #3385, opened 2020-07-22
Synthesize placeholder AET identifiers if missing

eoger:
Fixes https://github.com/mozilla/application-services/issues/3086. I haven't got the chance to live-test it yet though!

### DIFF
--- a/components/fxa-client/src/http_client.rs
+++ b/components/fxa-client/src/http_client.rs
@@ -3,13 +3,19 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::{config::Config, error::*};
-use rc_crypto::hawk::{Credentials, Key, PayloadHasher, RequestBuilder, SHA256};
-use rc_crypto::{digest, hkdf, hmac};
-use serde_derive::*;
+use jwcrypto::Jwk;
+use rc_crypto::{
+    digest,
+    hawk::{Credentials, Key, PayloadHasher, RequestBuilder, SHA256},
+    hkdf, hmac,
+};
+use serde_derive::{Deserialize, Serialize};
 use serde_json::json;
-use std::collections::HashMap;
-use std::sync::Mutex;
-use std::time::{Duration, Instant};
+use std::{
+    collections::HashMap,
+    sync::Mutex,
+    time::{Duration, Instant},
+};
 use url::Url;
 use viaduct::{header_names, status_codes, Method, Request, Response};
 
@@ -18,7 +24,7 @@ const HAWK_KEY_LENGTH: usize = 32;
 const RETRY_AFTER_DEFAULT_SECONDS: u64 = 10;
 
 #[cfg_attr(test, mockiato::mockable)]
-pub trait FxAClient {
+pub(crate) trait FxAClient {
     fn refresh_token_with_code(
         &self,
         config: &Config,
@@ -68,6 +74,12 @@ pub trait FxAClient {
         profile_access_token: &str,
         etag: Option<String>,
     ) -> Result<Option<ResponseAndETag<ProfileResponse>>>;
+    fn set_ecosystem_anon_id(
+        &self,
+        config: &Config,
+        access_token: &str,
+        ecosystem_anon_id: &str,
+    ) -> Result<()>;
     fn pending_commands(
         &self,
         config: &Config,
@@ -103,6 +115,8 @@ pub trait FxAClient {
         client_id: &str,
         scope: &str,
     ) -> Result<HashMap<String, ScopedKeyDataResponse>>;
+    fn fxa_client_configuration(&self, config: &Config) -> Result<ClientConfigurationResponse>;
+    fn openid_configuration(&self, config: &Config) -> Result<OpenIdConfigurationResponse>;
 }
 
 enum HttpClientState {
@@ -117,6 +131,16 @@ pub struct Client {
     state: Mutex<HashMap<String, HttpClientState>>,
 }
 impl FxAClient for Client {
+    fn fxa_client_configuration(&self, config: &Config) -> Result<ClientConfigurationResponse> {
+        // Why go through two-levels of indirection? It looks kinda dumb.
+        // Well, `config:Config` also needs to fetch the config, but does not have access
+        // to an instance of `http_client`, so it calls the helper function directly.
+        fxa_client_configuration(config.client_config_url()?)
+    }
+    fn openid_configuration(&self, config: &Config) -> Result<OpenIdConfigurationResponse> {
+        openid_configuration(config.openid_config_url()?)
+    }
+
     fn profile(
         &self,
         config: &Config,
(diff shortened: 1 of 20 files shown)


components/fxa-client/src/http_client.rs
```diff
@@ -103,6 +115,8 @@ pub trait FxAClient {
         client_id: &str,
         scope: &str,
     ) -> Result<HashMap<String, ScopedKeyDataResponse>>;
+    fn fxa_client_configuration(&self, config: &Config) -> Result<ClientConfigurationResponse>;
```
eoger, line 118: I've moved the FxA config fetching in `http_client` so it can be mocked in tests.

components/fxa-client/src/http_client.rs
```diff
@@ -68,6 +74,12 @@ pub trait FxAClient {
         profile_access_token: &str,
         etag: Option<String>,
     ) -> Result<Option<ResponseAndETag<ProfileResponse>>>;
+    fn ecosystem_anon_id(
```
rfk, line 77: API nit: this feels like it should contain a verb, like `set_ecosystem_anon_id` or similar; at first read I thought this was going to *return* the anon_id, which it doesn't.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+        // Default case: the ecosystem anon ID was generated during login.
+        if let Some(ecosystem_anon_id) = profile.ecosystem_anon_id {
+            return Ok(ecosystem_anon_id);
+        }
+        if !generate_placeholder {
+            return Err(ErrorKind::IllegalState("ecosystem_anon_id should be present").into());
+        }
+        // For older clients, we generate an ecosystem_user_id,
+        // persist it and then return ecosystem_anon_id.
+        let mut ecosystem_user_id = vec![0u8; 16];
```
rfk, line 29: The specs for this have recently changed, we need to generate 32 bytes here rather than 16.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+        // Will end up as a len 32 hex string.
+        let ecosystem_user_id = hex::encode(ecosystem_user_id);
+
+        let anon_id_key = self.fetch_random_ecosystem_anon_id_key()?;
+        let ecosystem_anon_id = jwcrypto::encrypt_to_jwe(
+            &ecosystem_user_id.as_bytes(),
+            EncryptionParameters::ECDH_ES {
+                enc: EncryptionAlgorithm::A256GCM,
+                peer_jwk: &anon_id_key,
+            },
```
rfk, line 40: Beautiful, I love when a good code-reuse plan comes together.

components/fxa-client/src/http_client.rs
```diff
@@ -143,6 +167,25 @@ impl FxAClient for Client {
         }))
     }
 
+    fn ecosystem_anon_id(
+        &self,
+        config: &Config,
+        access_token: &str,
+        ecosystem_anon_id: &str,
+    ) -> Result<()> {
+        let url = config.auth_url_path("v1/account/ecosystemAnonId")?;
```
rfk, line 176: Please use `config.profile_url_path("v1/ecosystem_anon_id")` here, we need to write via the profile-server rather than the auth-server.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
 impl FirefoxAccount {
-    pub fn get_ecosystem_anon_id(&mut self) -> Result<Option<String>> {
+    /// Get the ecosystem anon id, generating it if necessary.
+    ///
+    /// **💾 This method alters the persisted account state.**
+    pub fn get_ecosystem_anon_id(&mut self) -> Result<String> {
+        self.get_ecosystem_anon_id_helper(true)
+    }
+
+    pub fn get_ecosystem_anon_id_helper(&mut self, generate_placeholder: bool) -> Result<String> {
```
rfk, line 18: nit: does this need to be `pub`? it feels like an internal-only helper.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+                token: "profiletok".to_string(),
+                key: None,
+                expires_at: u64::max_value(),
+            },
+        );
+        fxa.add_cached_token(
+            "profile:write",
+            AccessTokenInfo {
+                scope: "profile".to_string(),
+                token: "profiletok".to_string(),
```
rfk, line 95: testing nit: if you make this `"profilewritetok"` is similar, you might get some extra checks that the correct token is being used at the correct time.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+                        crv: "P-256".to_owned(),
+                        x: "i3FM3OFSCZEoqu-jtelXwKt6AL4ODQ75NUdHbcLWQSo".to_owned(),
+                        y: "nW-S3QiHDo-9hwfBhKnGKarkt_PVqVyIPUytjutTunY".to_owned(),
+                    }),
+                }]),
+            }));
+        client
+            .expect_ecosystem_anon_id(
+                mockiato::Argument::any,
+                mockiato::Argument::any,
```
rfk, line 176: e.g. you could use `|token| token.partial_eq("profilewritetok")` on the token argument here.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+            )
+            .times(1)
+            .returns_once(Ok(()));
+        fxa.set_client(Arc::new(client));
+
+        let ecosystem_anon_id = fxa.get_ecosystem_anon_id().unwrap();
+        // Well, it looks like a jwe folks.
+        assert!(ecosystem_anon_id.chars().filter(|c| c == &'.').count() == 4);
+        assert!(fxa.state.ecosystem_user_id.unwrap().len() == 32);
```
rfk, line 186: For bonus points, you could use a known JWK public key in the mock config above, and use the corresponding known private key to decrypt it here. Not sure how much value that would provide in practice though.

components/fxa-client/src/telemetry.rs
```diff
@@ -2,12 +2,268 @@
...
+                errno: 500,
+                error: "precondition failed".to_string(),
+                message: "another user did it".to_string(),
+                info: "".to_string(),
+            }
+            .into()));
+        fxa.set_client(Arc::new(client));
+
+        assert_eq!(fxa.get_ecosystem_anon_id().unwrap(), ecosystem_anon_id);
+        assert!(fxa.state.ecosystem_user_id.is_none());
```
rfk, line 267: 👍 very nice suite of tests

components/fxa-client/src/telemetry.rs
eoger: Yeah I was going to do that then I felt like it's just re-testing the jw stack.

rfk:
One thing we should add here maybe is a changelog entry, but otherwise 👍.

PR #3400, opened 2020-07-27
Clear devices and client cache more aggressively

### LINKED ISSUE #3348: Password change doesn't immediately trigger Rust calls FxaUnauthorizedException
After user changed password on Desktop, every rust call done from mobile should get an `FxaUnauthorizedException`, so that the user re-authenticates. 

However, after password change, there is usually a ~10 seconds window(doesn't happen every time) in which performing a call to get all devices from Mozilla Android Components (`LibFxAFFI.INSTANCE.fxa_get_devices(this.handle.get(), ignoreCacheArg, e)`) returns the devices even though password is changed, and the call to get all remote tabs (`LibRemoteTabsFFI.INSTANCE.remote_tabs_get_all(this.handle.get(), error)`) sometimes returns only local tabs, and sometimes returns also the remote ones, even in those 10 seconds.

This leads to unwanted states in Fenix during that time-window. I wonder if this is an architectural decision(eventual consistency) and it should be handled by the client, or this is a problem.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1484)
┆Story Points: 3
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-07-31

vladikoff:
Adds this improvement to minimize the cases where we show cached data and the account had a password reset.

@lougeniaC64 r?

Fixes #3348

### DIFF
--- a/components/fxa-client/src/profile.rs
+++ b/components/fxa-client/src/profile.rs
@@ -27,6 +27,7 @@ impl FirefoxAccount {
                         "Access token rejected, clearing the tokens cache and trying again."
                     );
                     self.clear_access_token_cache();
+                    self.clear_devices_and_attached_clients_cache();
                     self.get_profile_helper(ignore_cache)
                 }
                 _ => Err(e),
--- a/components/fxa-client/src/oauth.rs
+++ b/components/fxa-client/src/oauth.rs
@@ -429,6 +429,7 @@ impl FirefoxAccount {
         });
         self.state.session_token = Some(session_token.to_owned());
         self.clear_access_token_cache();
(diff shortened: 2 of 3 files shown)


components/fxa-client/src/profile.rs
```diff
@@ -27,6 +27,7 @@ impl FirefoxAccount {
                         "Access token rejected, clearing the tokens cache and trying again."
                     );
                     self.clear_access_token_cache();
+                    self.clear_devices_and_attached_clients_cache();
```
vladikoff, line 30: This clear here might not be as useful as the other two, but might still be worth keeping it here.

PR #3448, opened 2020-07-31
fix swiftformat after update

tarikeshaq:
Runs updated `swiftformat` to fix lint errors that showed up after `swiftformat` got an update.

Also modifies the `swiftlint` rules to exclude `opening_braces` which contradicts with what `swiftformat` wants

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -20,6 +20,8 @@ disabled_rules:
   - todo
   # It disagrees with swiftformat on this, and thinks trailing commas are bad...
   - trailing_comma
+  # It disagrees with swiftformat on this, and wants the opening braces to be on the same line in multiline declarations
+  - opening_brace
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -248,7 +248,8 @@ jobs:
     steps:
       - checkout
       - run: brew install swiftlint swiftformat
-      - run: swiftlint --strict
+      # Commenting out swiftlint for now because it started contradictiong swiftformat (https://github.com/mozilla/application-services/pull/3448)
```
eoger, line 251: Instead of disabling swiftlint altogether, I suggest adding `opening_brace` under `disabled_rules` in `.swiftlint.yml`

tarikeshaq: Done! That sounds better lol I had no idea that existed 😆

PR #3466, opened 2020-08-05
Avoid our coverage issues by only emitting line coverage (and not branch coverage)

### LINKED ISSUE #3393: Find some way to improve our coverage accuracy
Here are some issues with the current codecov.io coverage reports:

1. Codecov considers `foo()?` to be a partial hit, where it probably be be a full hit -- after all, `?` in Rust is roughly analogous to exception propagation in other languages, and it's not like every line that could throw in a language like JS or Java is considered partial.
2. Codecov considers `log::foo!` to be a partial hit, where it should be a full hit. If rust's syntax were such that `log::foo!` could be a function call instead of a macro, it would be a full hit.
3. Codecov considers a passing `assert!`, `assert_eq!`, etc. to be a partial hit, where it should be a full hit. This drastically impacts test files.

The reason for all of these is: grcov-based coverage is computed after macro/mir-expansion, and so these lines *are* partials from it's perspective, and unfortunately, according to https://docs.codecov.io/docs/frequently-asked-questions#how-is-coverage-calculated:

> Note: Partials are considered a full miss, regardless of the number of branches hit or missed.

Ironically, locally if you generate coverage it gets this right (or, more right), which is why I saw 70+% locally, and yet we see a little over 50% here.

I'm filing this issue for documentation of the problem more than for saying we should actually fix it -- these are issues I believe are inherent in using `grcov` (instead of `tarpaulin`) to measure the coverage.

They're pretty frustrating, especially number 3, but there's nothing we can do about them at the moment.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1514)
┆Story Points: 1
┆Epic: Increase unit and integration test coverage
┆Sprint: SYNC - end 2020-08-14

thomcc:
Fixes #3393. It's a little unfortunate in theory, but in practice the branch coverage was inaccurate to the point of being unusable (for Rust code, anyway).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -282,7 +282,7 @@ jobs:
       - run: bash ./automation/emit_coverage_info.sh
       - run:
           name: Compile coverage
-          command: grcov ./target/debug/ -s . -t lcov --llvm --branch --ignore-not-existing --ignore "target/*" --ignore "/*" -o lcov.info
+          command: grcov ./target/debug/ -s . -t lcov --llvm --ignore-not-existing --ignore "target/*" --ignore "/*" -o lcov.info
       - run:
           name: Upload to codecov.io
           command: bash <(curl -s https://codecov.io/bash) -f lcov.info


thomcc:
I filed https://github.com/mozilla/application-services/issues/3468 to undo this at some point in the future if the issues resolve themselves.

In truth, it's mostly there so that the description can be referred to in the future, I don't expect us to push on it particularly hard.

PR #3479, opened 2020-08-07
Merge v61.0.11 into main

mhammond:
According to the release instructions, this is what I'm supposed to do. I'm not quite sure about how CHANGELOG.md and CHANGES_UNRELEASED.md were supposed to be resolved, but this makes the most sense to me (in particular, the stuff released in v61.0.11 is back in CHANGES_UNRELEASED.md, because it will need to be noted in the next v62.* release.)


thomcc:
Taking this so that we can cut v61.0.12

PR #3491, opened 2020-08-11
[POC] fxa-client Kotlin/Swift bindings using uniffi

eoger:
This is not really ready for production yet, but when I'm done building it it'll be a cool demo :-)

- [X] Rust scaffolding.
- [ ] Kotlin bindings + integration in build system.
- [ ] Swift bindings + integration in build system.

I left some `XXXAWKWARD` comments throughout the code to document what doesn't work/could work better in Uniffi.


components/fxa-client/src/lib.rs
```diff
@@ -162,33 +162,33 @@ impl FirefoxAccount {
...
-        Ok(self.state.config.pair_url()?)
+        Ok(self.state.config.pair_url()?.into_string())
     }
 
     /// Get the "connection succeeded" page URL.
     /// It is typically used to redirect the user after
     /// having intercepted the OAuth login-flow state/code
     /// redirection.
-    pub fn get_connection_success_url(&self) -> Result<Url> {
+    pub fn get_connection_success_url(&self) -> Result<String> {
```
rfk, line 187: Oh, interesting, is this because we don't have support for declaring "extern" types in the IDL, so we can't declare it to return a `Url`?

eoger: Yup indeed, for what it's worth the current FFI also calls `to_string()`.

PR #3492, opened 2020-08-12
Fixes broken dependencies CI

tarikeshaq:
Added the fixup, the changes on the other two files were autogenerated.

Not 100% sure yet why this was needed. There seems to be a release 3 days ago in `tinyvec`. Also, they changed to a `main` branch like we did, and deleted the master branch.

Anywhoo, on a related note, I wonder if we should do an audit of our `fixups` to make sure they are still okay. (since a decent chunk point to `master` branches, and I'm not sure if any owners made the switch)

### DIFF
--- a/megazords/full/android/dependency-licenses.xml
+++ b/megazords/full/android/dependency-licenses.xml
@@ -278,7 +278,7 @@ the details of which are reproduced below.
   </license>
   <license>
     <name>Zlib License: tinyvec</name>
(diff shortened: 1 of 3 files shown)


eoger:
Side note we probably should add `main` to https://github.com/mozilla/application-services/blob/d2123eba44ffc725bb7a3ef7a2b445604d209229/tools/dependency_summary.py#L824-L828

PR #3510, opened 2020-08-20
Ensure the syncStatus is correct after reconciling remote bookmarks.

mhammond:
This is basically a direct port of desktop's bug 1635859

Marking as draft because I want to check that unconditionally setting the change counter to zero here is OK, or whether our timed transaction hackery might bite us, and I'm out of time for the day. @linacambridge do you happen to know off the top of your head?

### DIFF
--- a/components/places/src/bookmark_sync/store.rs
+++ b/components/places/src/bookmark_sync/store.rs
@@ -507,7 +507,8 @@ impl<'a> BookmarksStore<'a> {
                title = excluded.title,
                dateAdded = excluded.dateAdded,
                lastModified = excluded.lastModified,
-               fk = excluded.fk",
+               fk = excluded.fk,
+               syncStatus = {sync_status}",
             root_guid = BookmarkRootGuid::Root.as_guid().as_str(),
             type_fragment = ItemTypeFragment("newKind"),
             sync_status = SyncStatus::Normal as u8,
@@ -1775,7 +1776,14 @@ mod tests {
         }
     }
 
-    fn apply_incoming(conn: &PlacesDb, remote_time: ServerTimestamp, records_json: Value) {
+    // Applys the incoming records, and also "finishes" the sync by pretending
+    // we uploaded the outgoing items and marks them as uploaded.
+    // Returns the GUIDs of the outgoing items.
+    fn apply_incoming(
+        conn: &PlacesDb,
+        remote_time: ServerTimestamp,
+        records_json: Value,
+    ) -> Vec<Guid> {
         // suck records into the store.
         let interrupt_scope = conn.begin_interrupt_scope();
         let store = BookmarksStore::new(&conn, &interrupt_scope);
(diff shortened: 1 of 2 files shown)


mhammond:
I think setting the change counter to zero is wrong because we fetch the local tree outside of the transaction. We also don't track the change counter for local items (we do fetch it, but only to set `needs_merge`). The easiest solution might be to move the transaction boundary - create the transaction before we call `merger.merge()`. The other alternative would be to track the counter value and decrement it by that amount, but I can't quite picture what that sql would look like. @linacambridge wdyt?

linacambridge:
Oooh, great catch, @mhammond!

I think timed transactions aren't the problem here, because `apply` [never calls](https://github.com/mozilla/application-services/blob/fe6efcc846de2c84abd529f2af3915b479493a89/components/places/src/bookmark_sync/store.rs#L1493-L1510) `maybe_commit` to yield to another transaction—it only calls `begin_transaction` at the start, and `commit` at the end.

_And_, you're completely correct, fetching the local tree outside the transaction does mean another connection (on another thread) might write to the database between `fetch_local_tree` and `apply`. Ouch!

On Desktop, I think we originally fetched the local tree inside the merge transaction for that very reason, but that ended up causing `SQLITE_BUSY` errors and deadlocks for users with large libraries. The approach Desktop uses now is to [call a SQL function](https://searchfox.org/mozilla-central/rev/ce21a13035623c1d349980057d09000e70669802/toolkit/components/places/nsPlacesTriggers.h#253,264,320,328) on every bookmarks change that bumps an in-memory ["global" change counter](https://searchfox.org/mozilla-central/rev/ce21a13035623c1d349980057d09000e70669802/toolkit/components/places/nsINavBookmarksService.idl#212-230), and checks it [before and after](https://searchfox.org/mozilla-central/rev/ce21a13035623c1d349980057d09000e70669802/toolkit/components/places/bookmark_sync/src/store.rs#34-38,58,460-462) starting the transaction.

The only mention of `note_sync_change` that I can find in Rust Places is in [this comment](https://github.com/mozilla/application-services/blob/fe6efcc846de2c84abd529f2af3915b479493a89/components/places/sql/create_shared_triggers.sql#L73-L75). I can't remember why we didn't port it over...I think I misunderstood how threading worked when we were building the component, and wrongly assumed that Rust Places would only be called from a single thread, so we'd never end up with contending writes from multiple connections because all the calls would be serialized 😕

> The easiest solution might be to move the transaction boundary - create the transaction before we call `merger.merge()`

This seems like a good short-term fix, and maybe we can port `note_sync_change` after if performance for large libraries turns out to be an issue?

mhammond:
Ok, so with #3589 and an update here to match what we agreed to on desktop (specifically, don't touch the counter, just the status) I think this is ready to roll

linacambridge (approved):
Yay! :shipit:

PR #3529, opened 2020-08-24
Run cargo update [ci full]

tarikeshaq:
Aaah looks like Friday's automated cargo-update didn't go through because of a missing `license_file`. Added the fixup.

I'm not entirely sure if it is correct though. Since the actual license file in the `tinyvec` repo is a Zlib license, but with the fixup, it's putting it under Apache 2.0 🤔 (Although in the Cargo.toml they indicate that it could be licensed under Zlib, Apache 2.0 or MIT so it might be okay)


eoger (approved):

vladikoff:
"it could be licensed under Zlib, Apache 2.0 or MIT" 

Yeap, that should work!

Issue #3536, opened 2020-08-25
UrlParseFailed crashes in every single API which deals with urls


grigoryk:
We're seeing lots of UrlParseFailed coming out of every single API which deals with external URL input (history, bookmarks).

For history, a-c swallows some of these crashes already. However, they're not limited to things we can easily ignore. For example, this crash comes up when creating a bookmark: https://crash-stats.mozilla.org/report/index/35bb5d01-8cb9-4674-b20a-136530200825.
Ignoring a failure here isn't really a feasible approach. Surfacing an error to the user saying that we failed to create a bookmark for a website they have open in a tab, since its url is invalid is also not really an option.

Can we make url parsing more robust and/or more lenient?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1702)
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-09-25

eoger:
It seems like we're hitting [this](https://searchfox.org/mozilla-mobile/rev/9cbb76f4bb290146a6f029dbd31b5affae2440cb/application-services/components/places/src/storage/bookmarks/public_node.rs#187-190), which would mean the database got populated with invalid urls (either through migration or just regular usage). Immediate action would be to ignore the error here, but maybe we should run a maintenance query that yanks invalid URLs and/or audit our URL inputs and figure out where the invalid URLs come from.

rfk:
Here's another one from a reddit thread: https://crash-stats.mozilla.org/report/index/cf7846b7-f131-49e5-9bc7-8ded60200825

mhammond:
This is a bit odd: both the fennec import process ([here](https://github.com/mozilla/application-services/blob/796e8c3a198f48919babca5abeda447f7cd70a4a/components/places/src/import/fennec/bookmarks.rs#L315) to [here](https://github.com/mozilla/application-services/blob/796e8c3a198f48919babca5abeda447f7cd70a4a/components/places/src/import/common.rs#L62)) and sync's handling of incoming records ([here](https://github.com/mozilla/application-services/blob/796e8c3a198f48919babca5abeda447f7cd70a4a/components/places/src/bookmark_sync/incoming.rs#L108) to [here](https://github.com/mozilla/application-services/blob/796e8c3a198f48919babca5abeda447f7cd70a4a/components/places/src/bookmark_sync/incoming.rs#L466)) both validate the URL in the same way these failing functions do (ie, `Url::parse()`). 

[...]

PR #3537, opened 2020-08-26
Be tolerant of invalid URLs in bookmarks.

mhammond:
This seems reasonable to me - @linacambridge WDYT?

Only a draft for now as I want to dig a little more for other places we can sanely ignore parse failures when reading from the DB.

### DIFF
--- a/components/places/src/storage/bookmarks.rs
+++ b/components/places/src/storage/bookmarks.rs
@@ -919,7 +919,14 @@ pub fn bookmarks_get_url_for_keyword(db: &PlacesDb, keyword: &str) -> Result<Opt
     )?;
 
     match bookmark_url {
-        Some(b) => Ok(Some(Url::parse(&b)?)),
+        Some(b) => match Url::parse(&b) {
+            Ok(u) => Ok(Some(u)),
+            Err(e) => {
+                // We don't have the guid to log and the keyword is PII...
+                log::warn!("ignoring invalid url: {:?}", e);
+                Ok(None)
+            }
+        },
         None => Ok(None),
     }
 }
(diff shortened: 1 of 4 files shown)


mhammond:
I think this covers everything for bookmarks - note we still throw when trying to *insert*, it's only reads we are tolerant. We probably should audit history too.

mhammond:
Note that this *helps* with #3536, but shouldn't close it.

linacambridge:
Oh, goodness—I just read the discussion in #3536, and the linked a-c and Fenix bugs. Preventing invalid URLs from being bookmarked is definitely important. We talked about a maintenance task in Desktop to clean out invalid URLs in the past, but Mak pointed out that would involve a [full table scan](https://bugzilla.mozilla.org/show_bug.cgi?id=1470043#c3), and it would be easier to clean them out at read time—which is building on what you've done here with just hiding them!

At first, I was nervous about hiding bookmarks like this, because their positions might be off, and the structure we're showing in the UI is different than what's in the database. But, those invalid bookmarks aren't going to be helpful to users, anyway, and it doesn't feel much different than deleting them outright on read. And, when repositioning a bookmark in a folder with hidden bookmarks, its position should still be correct relative to the others.

So ignoring bad bookmarks on reads (and maybe eventually moving to deleting them entirely, like Sync does now), and throwing on writes sounds great to me overall! 👍🏻

components/places/src/storage/bookmarks.rs
```diff
@@ -919,7 +919,14 @@ pub fn bookmarks_get_url_for_keyword(db: &PlacesDb, keyword: &str) -> Result<Opt
     )?;
 
     match bookmark_url {
-        Some(b) => Ok(Some(Url::parse(&b)?)),
+        Some(b) => match Url::parse(&b) {
+            Ok(u) => Ok(Some(u)),
+            Err(e) => {
+                // We don't have the guid to log and the keyword is PII...
+                log::warn!("ignoring invalid url: {:?}", e);
```
linacambridge, line 926: Can we emit some kind of counter for this, or even record part of the error in Sentry or something? It could be PII, but it would be nice to understand what kinds of syntactically invalid URLs we’re dealing with. Is it just IPs, like the Fenix ticket says, or something more?

That said, I’m not sure what we’d do with that information even if we knew—it’s interesting for my own curiosity, but it wouldn’t explain how they got opened, or written into the database in the first place. We aren’t going to be changing the URL parser or anything.

components/places/src/storage/bookmarks.rs
```diff
@@ -1241,14 +1248,31 @@ pub fn fetch_tree(
...
+                                date_added: Some(row.date_added),
+                                last_modified: Some(row.last_modified),
+                                title: row.title,
+                                url,
+                            }
+                            .into(),
+                        },
+                        // This is double-extra-invalid because various
+                        // constaints in the schema should prevent it - so make
+                        // some noise as it probably means something extra-bad?
```
linacambridge, line 1269: FWIW, on-disk database corruption could be one of the cases. On Desktop, there’s a very low but persistent occurrence of things that ought to be impossible in the schema, like `guid`s being null, types being totally nonsensical (what’s a bookmark type of -47231?), duplicate GUIDs in the tree, `moz_places` or `moz_bookmarks` not existing at all...

Anyway, I totally agree about making noise in this case, and also meaning something extra-bad! 😆

components/places/src/storage/bookmarks/public_node.rs
```diff
@@ -610,7 +657,16 @@ mod test {
...
-        let bmks = recent_bookmarks(&conns.read, 3)?;
+
+        append_invalid_bookmark(
+            &conns.write,
+            &BookmarkRootGuid::Unfiled.guid(),
+            "invalid url",
+            "badurl",
+        );
+
+        // The limit applies before we filter the invalid bookmark, so ask for 4.
```
linacambridge, line 668: This makes sense to me! 👍🏻 If invalid bookmarks are a common occurrence, something is really wrong, so assuming most of them are valid and not pushing complicated validation logic into the SQL is a great approach.

Will pagination do the right thing as well?

mhammond: > Will pagination do the right thing as well?

None of the functions I touched support pagination, so I think we'll be OK. IIRC, that's going to be more of a concern for history.

components/places/src/storage/bookmarks.rs
mhammond: yeah - given what you said and the fact that we don't have anywhere sane to record this to, I'm not going to do anything here. It's something we can revisit once glean is a little closer to us so we don't need to thread it back though android-components etc.

PR #3554, opened 2020-09-07
Bug 1661539 - Add Glean to a iOS megazord build

badboy:
This is a first try in adding Glean as a submodule to incorporate it into a iOS megazord build.
If this lands firefox iOS will also require updates (similar to the PoC in https://github.com/mozilla-mobile/firefox-ios/compare/main-dev...badboy:megazord-appservices-glean)

So far this was reasonably easy and I only need 2 small Swift changes to make it work (both of which can be later mitigated on the Glean side again).

I managed to build Firefox iOS locally from this branch.
I'm not sure if having the `git submodule update` in the `build-all-ios.sh` script is actually enough, but I think that one is run on all iOS platforms, so it's good enough?

cc @travis79, @Dexterp37, @mdboom for Glean.

### DIFF
--- a/components/logins/ios/Logins/Extensions/String+Free_Logins.swift
+++ b/components/logins/ios/Logins/Extensions/String+Free_Logins.swift
@@ -4,9 +4,17 @@
 
 import Foundation
 
-extension String {
-    public init(freeingRustString rustString: UnsafeMutablePointer<CChar>) {
-        defer { sync15_passwords_destroy_string(rustString) }
-        self.init(cString: rustString)
-    }
-}
+/*
+ * NOTE: This is now provided by Glean.
+ * Because it's all compiled into a single libmegazord.a it doesn't matter which `destroy_string` we call,
+ * internally it defers to the same string implementation.
+ * Because Glean is added as a submodule it's easier to change this occurence than it it to change the one of Glean.
+ */
+/*
+ extension String {
+     public init(freeingRustString rustString: UnsafeMutablePointer<CChar>) {
+         defer { sync15_passwords_destroy_string(rustString) }
+         self.init(cString: rustString)
+     }
+ }
+ */
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,10 @@ xcuserdata
 # Protobuf Swift
 *.pb.swift
 
+# Glean generated artifacts
+megazords/ios/MozillaAppServices/Generated/Metrics.swift
+megazords/ios/.venv
+
 # Carthage generated artifacts
 Carthage
 MozillaAppServices.framework.zip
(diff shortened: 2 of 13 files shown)


components/logins/ios/Logins/Extensions/String+Free_Logins.swift
```diff
@@ -4,9 +4,17 @@
...
-        self.init(cString: rustString)
-    }
-}
+/*
+ * NOTE: This is now provided by Glean.
+ * Because it's all compiled into a single libmegazord.a it doesn't matter which `destroy_string` we call,
+ * internally it defers to the same string implementation.
+ * Because Glean is added as a submodule it's easier to change this occurence than it it to change the one of Glean.
+ */
+/*
```
rfk, line 13: I'm curious what the underlying conflict was here - was it swift complaining about trying to add two extension methods with the same name on the `String` class?

badboy: Yes, we have the same extension to `String` with the same named parameter.
I actually propose to rename it on our side (`s/Rust/Glean/`) to avoid the conflict, but as I also want to submodule based on the release commit this has to be defered to the next release.

Dexterp37 (approved):
I like this!

travis79 (approved):
This looks reasonable enough to me, thanks @badboy for taking this on!

rfk (approved):
Great! Let's get this in a release and try it out.

@badboy do you have any other tweaks you wanted to try here, or should we merge this as-is and cut a new release?

rfk:
From conversation with @garvankeeley, let's also merge https://github.com/mozilla/application-services/pull/3556 before making a new release with this.

rfk:
Also could you please try re-running the dependency-list-generator script? It looks like something has gotten out of date.

badboy:
I rebased on `main` and then re-ran `tools/regenerate_dependency_summaries.sh`, though the files didn't actually change from the last commit.

Force-pushing again to see what CI thinks.

badboy:
I've ran the exact same commands CI runs but it doesn't fail locally.
Unfortuantely CI is not very chatty about what exactly differs.

rfk:
> I've ran the exact same commands CI runs but it doesn't fail locally.

I see similar when trying this locally :-(

I'm going to go ahead and merge this and if it's still failing, I'll be better situated to fix it up locally.

rfk (approved):

rfk:
Ref https://github.com/mozilla/application-services/pull/3565 for further investigation of the dependency issue.

garvankeeley:
https://github.com/mozilla-mobile/firefox-ios/pull/7303 is landed on iOS, locally I am building and running and it works well.

PR #3555, opened 2020-09-08
Strip `/1.0/sync/1.5` prefix if present in self-hosted tokenserver URL.

rfk:
Over in https://github.com/mozilla-services/docs/pull/89, a user noted that Firefox iOS differs from other clients when it comes to self-hosting - other clients expect you to provide the tokenserver url with `/1.0/sync/1.5` suffix, while Firefox iOS requires that you do not.

I think we should accept this known suffix and strip it if present, as a convenience for self-hosters.

### DIFF
--- a/components/fxa-client/src/config.rs
+++ b/components/fxa-client/src/config.rs
@@ -77,7 +77,12 @@ impl Config {
         &'a mut self,
         token_server_url_override: &str,
     ) -> &'a mut Self {
-        self.token_server_url_override = Some(token_server_url_override.to_owned());
+        // In self-hosting setups it is common to specify the `/1.0/sync/1.5` suffix on the
+        // tokenserver URL. Accept and strip this form as a convenience for users.
+        match token_server_url_override.strip_suffix("/1.0/sync/1.5") {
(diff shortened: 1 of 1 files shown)


mhammond:
I'm a little surprised by this because we [theoretically allow either this suffix to exist or not](https://github.com/mozilla/application-services/blob/main/components/sync15/src/token.rs#L66).

rfk:
Oh, huh. Thanks!

I was looking at the swift account manager code, which [unconditionally adds this prefix](https://github.com/mozilla/application-services/blob/0099443e9a3b7db4c246f843624d2b86f7b0a385/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L282) before returning it to the application, so maybe that difference in behavior is what's causing the user-visible effect here. Seems like it may be worth consolidating so it's stripped in one place, rather than having to remember to conditionally add it everywhere it's used.

rfk:
(Note that IIRC, the Firefox iOS app still has some parts that sync outside of the rust code, and hence does its own talking to tokenserver)

eoger (approved):
LGTM thanks

Issue #3550, opened 2020-08-31
Add ECDSA P-384 signature verification


eoger:
Blocks https://github.com/mozilla/application-services/issues/3475

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1729)
┆Story Points: 8
┆Epic: Nimbus Fenix Client SDK
┆Sprint: SYNC - end 2020-09-25

PR #3557, opened 2020-09-08
ECDSA P384-SHA384 signature verification

eoger:
Fixes https://github.com/mozilla/application-services/issues/3550

### DIFF
--- a/components/support/rc_crypto/src/agreement.rs
+++ b/components/support/rc_crypto/src/agreement.rs
@@ -124,6 +124,29 @@ impl PublicKey {
     }
 }
 
+/// An unparsed public key for key agreement.
+pub struct UnparsedPublicKey<'a> {
+    alg: &'static Algorithm,
+    bytes: &'a [u8],
+}
+
+impl<'a> UnparsedPublicKey<'a> {
+    pub fn new(algorithm: &'static Algorithm, bytes: &'a [u8]) -> Self {
+        Self {
+            alg: algorithm,
+            bytes,
+        }
+    }
+
+    pub fn algorithm(&self) -> &'static Algorithm {
+        self.alg
+    }
+
+    pub fn bytes(&self) -> &'a [u8] {
+        &self.bytes
+    }
+}
+
 /// A private key for key agreement.
 pub struct PrivateKey<U: Lifetime> {
     wrapped: ec::PrivateKey,
@@ -148,17 +171,8 @@ impl<U: Lifetime> PrivateKey<U> {
     /// Ephemeral agreement.
     /// This consumes `self`, ensuring that the private key can
     /// only be used for a single agreement operation.
-    pub fn agree(
-        self,
-        peer_public_key_alg: &Algorithm,
-        peer_public_key: &[u8],
-    ) -> Result<InputKeyMaterial> {
-        agree_(
-            &self.wrapped,
-            self.alg,
-            peer_public_key_alg,
-            peer_public_key,
-        )
+    pub fn agree(self, peer_public_key: &UnparsedPublicKey<'_>) -> Result<InputKeyMaterial> {
+        agree_(&self.wrapped, self.alg, peer_public_key)
     }
 }
 
@@ -168,15 +182,9 @@ impl PrivateKey<Static> {
     /// be used for a multiple agreement operations.
     pub fn agree_static(
         &self,
-        peer_public_key_alg: &Algorithm,
-        peer_public_key: &[u8],
+        peer_public_key: &UnparsedPublicKey<'_>,
     ) -> Result<InputKeyMaterial> {
-        agree_(
-            &self.wrapped,
-            self.alg,
-            peer_public_key_alg,
-            peer_public_key,
-        )
+        agree_(&self.wrapped, self.alg, peer_public_key)
     }
 
     pub fn import(ec_key: &EcKey) -> Result<Self> {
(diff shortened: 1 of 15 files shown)


components/support/rc_crypto/src/signature.rs
```diff
@@ -0,0 +1,89 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This file contains code that was copied from the ring crate which is under
+// the ISC license, reproduced below:
```
rfk, line 6: Is this still true for this new file?

rfk (commented):
FWIW, I don't really "get" the need for `UnparsedPublicKey` here, could you please say a bit about why it's necessary? (either as a comment in the code, or here in the issue).

components/support/rc_crypto/src/signature.rs
```diff
@@ -0,0 +1,89 @@
...
+        .unwrap();
+        let signature = base64::decode_config(
+            "XLZmtJweW4qx0u0l6EpfmB5z-S-CNj4mrl9d7U0MuftdNPhmlNacV4AKR-i4uNn0TUIycU7GsfIjIqxuiL9WdAnfq_KH_SJ95mduqXgWNKlyt8JgMLd4h-jKOllh4erh",
+            base64::URL_SAFE_NO_PAD,
+        )
+        .unwrap();
+        let public_key =
+            crate::signature::UnparsedPublicKey::new(&ECDSA_P384_SHA384, &pub_key_bytes);
+        public_key.verify(&message, &signature).unwrap();
+    }
```
rfk, line 88: Let's also add some tests for *failing* verifications of various kinds; I would hate for us to be e.g. accidentally returning a numeric error code as a boolean `true` and accidentally allowing things through that shouldn't be.

rfk (changes requested):

rfk:
Let's also make sure we get this looked over by an appropriate subject-matter expert; I've reached out to Thyla to see what her preferred process would be, but we can fall back to "ping specific people and ask them to take a look".

components/support/rc_crypto/src/signature.rs
eoger: Yes the API is very similar.

eoger:
> FWIW, I don't really "get" the need for `UnparsedPublicKey` here, could you please say a bit about why it's necessary? (either as a comment in the code, or here in the issue).

I think it's worth doing for the `signature` module at least, I like how it separates the Public Key from the `verify` operation, but I'm definitely open to suggestions!

rfk:
> I like how it separates the Public Key from the verify operation,

Yeah, I think the structure that it provides to the code is valuable, and I wonder if it's just the name that's throwing me off. I read "unparsed public key" and expect you to eventually parse it into a public key that you then use, but then it's also got this extra algorithm info attached to it which isn't used for parsing it.

This is partly a suggestion, and partly to test my own understanding - what if we invented two new similar-but-not-identical names for the two uses here:

* `KeyAgreementParams` for all the parameters (algorithm and public key) that you need in order to do a key agreement.
* `VerificationParams` for all the parameters (algorithm and public key) that you need in order to verify a signature.

Does that map to what you're looking to achieve here?

rfk:
> I've reached out to Thyla to see what her preferred process would be,

Just looping back to this, Thyla has asked @jcjones and @rmartinho to help us out with reviews here.

components/support/rc_crypto/src/agreement.rs
```diff
@@ -124,6 +124,29 @@ impl PublicKey {
...
 
+/// An unparsed public key for key agreement.
+pub struct UnparsedPublicKey<'a> {
+    alg: &'static Algorithm,
+    bytes: &'a [u8],
+}
+
+impl<'a> UnparsedPublicKey<'a> {
+    pub fn new(algorithm: &'static Algorithm, bytes: &'a [u8]) -> Self {
+        Self {
```
jcjones, line 135: It would be a good idea to assert that the length of `bytes` matches the expected length for the `algorithm`; since so many places are coercing a byte array to the key-type, it'd be good to know if we accidentally typo something in the future.

jcjones (commented):
I don't see any implementation issues. I agree that we need some explicitly failing tests -- obviously the underlying signature algorithm is well tested, but we should catch the cases where we might be misconfigured at this level where reasonable. Try the success case but with junk appended to the to-be-verified text, manipulating the hash algorithm, stuff like that. I'd still like @rmartinho to look, and he might have some suggestions for what tests to add. Consider me r+ w/ more tests.

jcjones:
> > FWIW, I don't really "get" the need for `UnparsedPublicKey` here, could you please say a bit about why it's necessary? (either as a comment in the code, or here in the issue).
> 
> I think it's worth doing for the `signature` module at least, I like how it separates the Public Key from the `verify` operation, but I'm definitely open to suggestions!

I think it's a good idea, and it's good that it explicitly calls out that no validation was attempted on the key at that point. The embedded data could be not-a-key, could be the identity point -- and it makes it clear when it's being processed, which is nice.

rmartinho (commented):
This looks alright, but yes, it should have tests for failed verification. The surface here is small, other than the failure scenarios that JC suggested, I would also check the "regular" failure case with one bit flipped in the message or something.

components/support/rc_crypto/src/agreement.rs
eoger: Yup we do that in https://github.com/mozilla/application-services/blob/b46b35d1c45088875cddd904c489aa4d4c0d65b5/components/support/rc_crypto/nss/src/ec.rs#L353-L365.

eoger:
Thanks to both of you for taking the time to look at this!

rmartinho (approved):

rfk (approved):
👍 thanks all!

PR #1720, opened 2019-09-04
Run "Check Rust dependencies" on Linux machines

eoger:
Takes half the time on average and doesn't use more expensive machines.

PR #3565, opened 2020-09-10
Run CI dependency checks on a Mac, to properly detect iOS-specific dependencies

rfk:
Trying to debug failing dependency checks in CI that we can't reproduce locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,10 +303,11 @@ jobs:
           name: Lint Bash scripts with shellcheck
           command: sh automation/lint_bash_scripts.sh
   Check Rust dependencies:
-    docker:
-      - image: circleci/rust:latest
+    # This check has to be done on a mac, to be able to detect iOS-specific dependencies.
+    macos:
+      xcode: "11.7.0"
     steps:
-      - run: sudo apt-get install python3-pip
+      - install-rust
       - setup-rust-toolchain
       - checkout
       - dependency-checks
--- a/tools/dependency_summary.py
+++ b/tools/dependency_summary.py
@@ -17,6 +17,7 @@ import subprocess
 import hashlib
 import json
 import textwrap
+import difflib
 import itertools
 import collections
 from urllib.parse import urlparse, urlunparse
@@ -1086,7 +1087,13 @@ if __name__ == "__main__":
         print_dependency_summary_markdown(deps, file=output)
 
     if args.check:
+        output.seek(0)
+        outlines = output.readlines()
         with open(args.check, 'r') as f:
-            if f.read() != output.getvalue():
+            checklines = f.readlines()
+            if outlines != checklines:
                 raise RuntimeError(
-                    "Dependency details have changed from those in {}".format(args.check))
+                    "Dependency details have changed from those in {}:\n{}".format(
+                        args.check,
+                        "".join(difflib.unified_diff(checklines, outlines))
+                    ))


rfk:
OK so, when calculating the dependency list in CI, this is discarding some but not all of the new dependencies introduced by glean:

```
RuntimeError: Dependency details have changed from those in ./DEPENDENCIES.md:
--- 
+++ 
@@ -12,7 +12,6 @@
 * [MIT License: bincode](#mit-license-bincode)
 * [MIT License: bytes](#mit-license-bytes)
 * [MIT License: caseless](#mit-license-caseless)
-* [MIT License: dashmap](#mit-license-dashmap)
 * [MIT License: h2](#mit-license-h2)
 * [MIT License: http-body](#mit-license-http-body)
 * [MIT License: hyper](#mit-license-hyper)
@@ -24,7 +23,6 @@
 * [MIT License: mio](#mit-license-mio)
 * [MIT License: openssl-sys](#mit-license-openssl-sys)
 * [MIT License: ordered-float](#mit-license-ordered-float)
-* [MIT License: oslog](#mit-license-oslog)
 * [MIT License: schannel](#mit-license-schannel)
 * [MIT License: slab](#mit-license-slab)
 * [MIT License: synstructure](#mit-license-synstructure)
@@ -434,7 +432,6 @@
 
 The following text applies to code linked from these dependencies:
 [adler](https://github.com/jonas-schievink/adler.git),
-[ahash](https://github.com/tkaitchuck/ahash),
 [android_log-sys](https://github.com/nercury/android_log-sys-rs),
 [android_logger](https://github.com/Nercury/android_logger-rs),
 [anyhow](https://github.com/dtolnay/anyhow),
@@ -444,8 +441,6 @@
 [cc](https://github.com/alexcrichton/cc-rs),
 [cfg-if](https://github.com/alexcrichton/cfg-if),
 [chrono](https://github.com/chronotope/chrono),
-[const-random-macro](https://github.com/tkaitchuck/constrandom),
-[const-random](https://github.com/tkaitchuck/constrandom),
 [core-foundation-sys](https://github.com/servo/core-foundation-rs),
 [core-foundation](https://github.com/servo/core-foundation-rs),
 [crc32fast](https://github.com/srijs/rust-crc32fast),
@@ -508,7 +503,6 @@
 [pin-utils](https://github.com/rust-lang-nursery/pin-utils),
 [pkg-config](https://github.com/rust-lang/pkg-config-rs),
 [ppv-lite86](https://github.com/cryptocorrosion/cryptocorrosion),
-[proc-macro-hack](https://github.com/dtolnay/proc-macro-hack),
 [proc-macro2](https://github.com/alexcrichton/proc-macro2),
 [prost-derive](https://github.com/danburkert/prost),
 [prost](https://github.com/danburkert/prost),
```

Following the chain of dependencies reported in `Cargo.toml`, these all seem to be unique dependencies of `oslog`. Glean itself takes a platform-specific dependency on `oslog` like this:

```
[target.'cfg(target_os = "ios")'.dependencies]
oslog = { version = "0.0.3", default-features = false, features = ["logger"] }
```

So maybe this is not being activated in the build plan on CI, but is being activated in build plans when running locally.

rfk:
Aha! As of https://github.com/mozilla/application-services/pull/1720 the dependency checks run on a linux machine rather than a Mac.

Unfortunately, this script can't properly detect iOS-specific dependencies when run on anything other than a Mac, for dumb Mac platform lockin reasons. I'm going to revert the above change.

mhammond (approved):

rfk:
@eoger I'm going to merge this to unbreak CI, but we can have a broader conversation about the CI costs involved here if necessary. (For the record, I hate this this platform-specific stuff is necessary here 🤮)

badboy:
For dependency checks we shouldn't need to compile, so running with `--target aarch64-apple-ios` should be possible.

rfk:
The specific command that we end up trying to use here would be something like:

```
cargo +nightly -Z unstable-options build --build-plan --quiet --locked --package megazord_ios --target aarch64-apple-ios
```

Which fails for me with:

```
error: Error loading target specification: failed to get iphoneos SDK path: No such file or directory
```

But it's been a while since we touched this stuff, there may be a different command to use to get the info we need here.

badboy:
Ah, I didn't actually check which command gets run.
It's not clear [if build plans will stay](https://github.com/rust-lang/cargo/issues/7614) so I guess this is a good point in time to find out if there's a more stable way to get the info we need.

rfk:
Oh, interesting, thanks!

I had an earlier version of this script that used `cargo tree` rather than the buildplan, but IIRC it had the same problem with needing to run on a Mac.

rfk:
I filed https://github.com/mozilla/application-services/issues/3572 to surface the unsupportedness of buildplans as an independent issue for us to deal with.

Issue #3522, opened 2020-08-21
Add a store for the Forms Component


lougeniaC64:
Create a store module similar to the one for [webext-storage](https://github.com/mozilla/application-services/blob/main/components/webext-storage/src/store.rs). This work will include the creation of the db module.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1690)
┆Story Points: 5
┆Epic: Form data (addresses/credit cards) shared component
┆Sprint: SYNC - end 2020-09-25

PR #3582, opened 2020-09-15
Add DB layer and API for the Autofill Component

### LINKED ISSUE #3521: Create basic Forms API
Create a basic CRUD API for the Forms component. This work will include creating a schema for credit cards (we're punting on addresses on this pass) based on the [desktop schema](https://searchfox.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#55) and an api module.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1689)
┆Story Points: 5
┆Sprint: SYNC - end 2020-09-25

lougeniaC64:
Fixes #3521 and #3522 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 # Note: Any additions here should be repeated in default-members below.
 members = [
+    "components/autofill",
     "components/fxa-client",
     "components/fxa-client/ffi",
     "components/logins",
@@ -24,6 +25,7 @@ members = [
     "components/support/rc_crypto/nss/systest",
     "components/support/sql",
     "components/support/sync15-traits",
+    "components/support/types",
     "components/support/viaduct-reqwest",
     "components/sync_manager",
     "components/sync_manager/ffi",
@@ -68,6 +70,7 @@ members = [
 # To be clear: passing the `--all` or `--workspace` arg to cargo will make it
 # use the full member set.
 default-members = [
+    "components/autofill",
     "components/fxa-client",
     "components/fxa-client/ffi",
     "components/logins",
@@ -88,6 +91,7 @@ default-members = [
     "components/support/rc_crypto/nss/nss_sys",
     "components/support/sql",
     "components/support/sync15-traits",
+    "components/support/types",
     "components/support/viaduct-reqwest",
     "components/sync_manager",
     "components/sync_manager/ffi",
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,8 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v63.0.0...main)
+
+## Autofill
+
+### What's changed ###
+- Added a basic API and database layer for the autofill component. ([#3582](https://github.com/mozilla/application-services/pull/3582))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,6 +113,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
+name = "autofill"
+version = "0.1.0"
+dependencies = [
+ "env_logger",
+ "error-support",
+ "interrupt-support",
+ "libsqlite3-sys",
+ "log 0.4.11",
+ "nss_build_common",
+ "rusqlite",
+ "serde",
+ "serde_derive",
+ "serde_json",
+ "sql-support",
+ "sync-guid",
+ "sync15-traits",
+ "thiserror",
+ "types",
+ "url",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -653,6 +675,15 @@ version = "0.1.0"
 dependencies = [
  "backtrace",
  "thiserror",
+]
+
+[[package]]
+name = "example-autofill-utils"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "cli-support",
+ "structopt",
 ]
 
 [[package]]
@@ -3222,6 +3253,16 @@ name = "try-lock"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59547bce71d9c38b83d9c0e92b6066c4253371f15005def0c30d9657f50c7642"
+
+[[package]]
+name = "types"
+version = "0.1.0"
+dependencies = [
+ "rusqlite",
+ "serde",
+ "serde_derive",
+ "serde_json",
+]
 
 [[package]]
 name = "unicase"
--- a/components/autofill/Cargo.toml
+++ b/components/autofill/Cargo.toml
@@ -1,0 +1,33 @@
+[package]
+name = "autofill"
+edition = "2018"
+version = "0.1.0"
+authors = ["application-services@mozilla.com"]
+license = "MPL-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+error-support = { path = "../support/error" }
+interrupt-support = { path = "../support/interrupt" }
+log = "0.4"
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
+sql-support = { path = "../support/sql" }
+sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random"] }
+sync15-traits = {path = "../support/sync15-traits"}
+thiserror = "1.0"
+types = { path = "../support/types" }
+url = { version = "2.1", features = ["serde"] }
+
+[dependencies.rusqlite]
+version = "0.23.1"
+features = ["functions", "bundled", "serde_json", "unlock_notify"]
+
+[dev-dependencies]
+env_logger = { version = "0.7", default-features = false }
+libsqlite3-sys = "0.18.0"
+
+[build-dependencies]
+nss_build_common = { path = "../support/rc_crypto/nss/nss_build_common" }
--- a/components/autofill/build.rs
+++ b/components/autofill/build.rs
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Work around the fact that `sqlcipher` might get enabled by a cargo feature
+//! another crate in the workspace needs, without setting up nss. (This is a
+//! gross hack).
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    // Ugh. This is really really dumb. We don't care about sqlcipher at all. really
+    if nss_build_common::env_str("DEP_SQLITE3_LINK_TARGET") == Some("sqlcipher".into()) {
+        // If NSS_DIR isn't set, we don't really care, ignore the Err case.
+        let _ = nss_build_common::link_nss();
+    }
+}
--- a/components/autofill/sql/create_shared_schema.sql
+++ b/components/autofill/sql/create_shared_schema.sql
@@ -1,0 +1,99 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+CREATE TABLE IF NOT EXISTS addresses_data (
+    guid          TEXT NOT NULL PRIMARY KEY,
+    given_name          TEXT NOT NULL,
+    additional_name     TEXT NOT NULL,
+    family_name         TEXT NOT NULL,
+    organization        TEXT NOT NULL,  -- Company
+    street_address      TEXT NOT NULL,  -- (Multiline)
+    address_level3      TEXT NOT NULL,  -- Suburb/Sublocality
+    address_level2      TEXT NOT NULL,  -- City/Town
+    address_level1      TEXT NOT NULL,  -- Province (Standardized code if possible)
+    postal_code         TEXT NOT NULL,
+    country             TEXT NOT NULL,  -- ISO 3166
+    tel                 TEXT NOT NULL,  -- Stored in E.164 format
+    email               TEXT NOT NULL,
+
+    time_created        INTEGER NOT NULL,
+    time_last_used      INTEGER,
+    time_last_modified  INTEGER NOT NULL,
+    times_used          INTEGER NOT NULL DEFAULT 0,
+
+    sync_change_counter INTEGER NOT NULL DEFAULT 1
+);
+
+CREATE TABLE IF NOT EXISTS addresses_mirror (
+   guid          TEXT NOT NULL PRIMARY KEY,
+    given_name          TEXT NOT NULL,
+    additional_name     TEXT NOT NULL,
+    family_name         TEXT NOT NULL,
+    organization        TEXT NOT NULL,  -- Company
+    street_address      TEXT NOT NULL,  -- (Multiline)
+    address_level3      TEXT NOT NULL,  -- Suburb/Sublocality
+    address_level2      TEXT NOT NULL,  -- City/Town
+    address_level1      TEXT NOT NULL,  -- Province (Standardized code if possible)
+    postal_code         TEXT NOT NULL,
+    country             TEXT NOT NULL,  -- ISO 3166
+    tel                 TEXT NOT NULL,  -- Stored in E.164 format
+    email               TEXT NOT NULL,
+
+    time_created        INTEGER NOT NULL,
+    time_last_used      INTEGER,
+    time_last_modified  INTEGER NOT NULL,
+    times_used          INTEGER NOT NULL DEFAULT 0
+);
+
+CREATE TABLE IF NOT EXISTS addresses_tombstones (
+    guid            TEXT PRIMARY KEY,
+    time_deleted    INTEGER NOT NULL
+) WITHOUT ROWID;
+
+-- NOTE: The credit card tables below are not being implemented as there are still outstanding
+-- questions around how we address security model.
+
+-- CREATE TABLE IF NOT EXISTS credit_cards_data (
+--     guid                TEXT NOT NULL PRIMARY KEY,
+--     cc_name             TEXT NOT NULL, -- full name
+--     cc_given_name       TEXT NOT NULL,
+--     cc_additonal_name   TEXT NOT NULL,
+--     cc_family_name      TEXT NOT NULL,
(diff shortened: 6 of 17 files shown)


components/forms/src/api.rs
```diff
@@ -0,0 +1,248 @@
...
+use std::time;
+use sync_guid::Guid;
+
+#[derive(Debug, Clone, Hash, PartialEq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct Address {
+    #[serde(rename = "id")]
+    pub guid: Guid,
+
+    pub given_name: String,
```
linacambridge, line 19: If the database columns are nullable, these could be `Option<String>`.

components/forms/sql/create_shared_schema.sql
```diff
@@ -0,0 +1,101 @@
...
+    time_last_used      INTEGER,
+    time_last_modified  INTEGER NOT NULL,
+    times_used          INTEGER NOT NULL DEFAULT 0,
+
+    sync_change_counter INTEGER NOT NULL DEFAULT 1
+);
+
+CREATE TABLE IF NOT EXISTS addresses_mirror (
+    guid          TEXT NOT NULL PRIMARY KEY,
+    given_name          TEXT NOT NULL,
```
linacambridge, line 32: I think these may need to be nullable, though, because we [filter out](https://searchfox.org/mozilla-central/rev/30e70f2fe80c97bfbfcd975e68538cefd7f58b2a/browser/extensions/formautofill/FormAutofillStorage.jsm#1269-1273) values with empty strings on Desktop, so they won’t show up in the incoming records at all.

components/forms/src/api.rs
```diff
@@ -0,0 +1,248 @@
...
+            ":time_created": address.time_created,
+            ":time_last_used": address.time_last_used,
+            ":time_last_modified": address.time_last_modified,
+            ":times_used": address.times_used,
+        },
+    )?;
+    Ok(address)
+}
+
+fn get_address(conn: &Connection, guid: &Guid) -> Result<Option<Address>> {
```
linacambridge, line 158: Let’s return an `AddressDoesNotExist` error if the address doesn’t exist, and simplify the return signature to just `Result<Address>`. I’m thinking, if the caller wants to distinguish between address not found and other kinds of DB errors, it can destructure the error instead of doing a double-unwrap.

components/forms/src/api.rs
```diff
@@ -0,0 +1,248 @@
...
+            times_used
+        FROM addresses_data
+        WHERE guid = :guid";
+    let address = conn.query_row(
+        sql,
+        &[guid.as_str()],
+        |row| {
+            Ok(Address {
+                guid: Guid::from_string(row.get::<_, String>("guid")?),
+                given_name: row.get::<_, String>("given_name")?,
```
linacambridge, line 186:
```suggestion
                given_name: row.get("given_name")?,
```

I’m not sure if this would compile (and please ignore me if you tried this already, and I’m talking nonsense!), but the type checker might be smart enough to infer this is a `String` without having to write out the turbofish.

Also, holy moly, so many fields! 😅

components/forms/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+pub enum QuotaReason {
+    TotalBytes,
+    ItemBytes,
+    MaxItems,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ErrorKind {
+    #[error("Quota exceeded: {0:?}")]
+    QuotaError(QuotaReason),
```
linacambridge, line 18: I know this came from `webext_storage`, but forms doesn’t have quotas, so let’s nix these (and `InvalidConnectionType`, `ConnectionAlreadyOpen`, and `WrongApiForClose`, too, I think).

linacambridge (commented):
This looks like a terrific first cut, @lougeniaC64, thank you so much! I left some comments and nits throughout, but they’re suggestions, not hard blockers. I think after @mhammond has a look, we can get this landed quickly and iterate on it with PRs instead of having a big long-running branch.

Rust forms is aliiiiive! It’s aliiiiiive! ⚡⚡

components/forms/sql/create_shared_schema.sql
```diff
@@ -0,0 +1,101 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+CREATE TABLE IF NOT EXISTS addresses_data (
+    guid          TEXT NOT NULL PRIMARY KEY,
+    given_name          TEXT NOT NULL,
+    additional_name     TEXT NOT NULL,
+    family_name         TEXT NOT NULL,
```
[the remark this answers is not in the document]
lougeniaC64: @mhammond and I discussed this early on. We decided that we'll use `""` to represent a non-value, but in a hand wavy manner. Asking for my own edification, are you making this suggestion because it's more idiomatic to use `Option` instead of checking a `String`'s value with `is_empty()`? I'm not at all opposed to this change.

components/forms/sql/create_shared_schema.sql
```diff
@@ -0,0 +1,101 @@
...
+
+    time_created        INTEGER NOT NULL,
+    time_last_used      INTEGER,
+    time_last_modified  INTEGER NOT NULL,
+    times_used          INTEGER NOT NULL DEFAULT 0,
+
+    sync_change_counter INTEGER NOT NULL DEFAULT 1
+);
+
+CREATE TABLE IF NOT EXISTS addresses_tombstones (
```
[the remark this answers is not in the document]
lougeniaC64: I've definitely seen various ways for representing deleted records so this is very useful information. Avoiding bugs is ideal so I'll make these changes. Ignore my comments above on the `NULL` vs `NOT NULL` database constraints. 😬

components/forms/src/api.rs
```diff
@@ -0,0 +1,248 @@
...
+
+/*
+    TODO:
+        - get all
+        - update
+        - delete
+        - check if deleted on create and/or update?
+*/
+
+fn add_address(tx: &Transaction<'_>, mut address: Address) -> Result<Address> {
```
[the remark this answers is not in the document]
lougeniaC64: Oh yes, I was just running into complications from this as I completed the CRUD functions for addresses. 👍🏾

components/forms/src/api.rs
lougeniaC64: It compiles! Love it, so much cleaner.

lougeniaC64:
> This looks like a terrific first cut, @lougeniaC64, thank you so much! I left some comments and nits throughout, but they’re suggestions, not hard blockers. I think after @mhammond has a look, we can get this landed quickly and iterate on it with PRs instead of having a big long-running branch.
> 
> Rust forms is aliiiiive! It’s aliiiiiive! ⚡⚡

Thanks! I have the remainder of the addresses API committed locally but I'm going to make these updates before I push them. TL;DR this is a very very WIP PR.

components/forms/sql/create_shared_schema.sql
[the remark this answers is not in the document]
mhammond: Yeah, Lougenia and I were waxing lyrical about this. I can't see a practical difference between null and an empty string in these contexts - it's not as though an empty string is a valid `given_name` where NULL would not be, or that we'd treat them differently for conflict resolution purposes, for example - and using non-nullable strings just avoids the extra `Option<>` wrapper which, in all cases I can see, we'd just end up treating as an empty string anyway.

That said though, it was more a philosophical discussion, so I'm open to making them nullable if we can identify a good reason.

components/forms/sql/create_shared_schema.sql
mhammond: Or we could default them to empty strings, and/or get serde to do the right thing, right? (again, not really advocating one way or the other, but failing to see a semantic difference, and I see potential bugs where we assume `Some(s)` means `s` actually has a value when it may not.

components/forms/sql/create_shared_schema.sql
```diff
@@ -0,0 +1,101 @@
...
+    street_address      TEXT,                -- (Multiline)
+    address_level3      TEXT NOT NULL,       -- Suburb/Sublocality
+    address_level2      TEXT NOT NULL,       -- City/Town
+    address_level1      TEXT NOT NULL,       -- Province (Standardized code if possible)
+    postal_code         TEXT NOT NULL,
+    country             TEXT NOT NULL,       -- ISO 3166
+    tel                 TEXT,                -- Stored in E.164 format
+    email               TEXT,
+
+    -- TODO: removed computed fields, need to figure out
```
mhammond, line 20: I think we should consider removing them now.

components/forms/sql/create_shared_schema.sql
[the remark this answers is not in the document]
mhammond: The other downside of that is that there's a risk that items have values *and* the deleted flag, which would be bad. It also means that almost every query must remember to have an extra condition to check the deleted flag - which would be easy to forget as testing will rarely show it up as the vast majority of time there will be no tombstones.

components/forms/src/api.rs
mhammond: But why is that a good thing? :)

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 # Note: Any additions here should be repeated in default-members below.
 members = [
+    "components/forms",
```
mhammond, line 4: As a meta comment, I wonder if "forms" really is the correct name for the component? `forms` carries a bit of baggage from desktop - there's both an engine and a collection named 'forms', but this component is completely unrelated to both of them. Maybe "autofill"?

components/forms/sql/create_shared_schema.sql
```diff
@@ -0,0 +1,101 @@
...
+
+    sync_change_counter INTEGER NOT NULL DEFAULT 1
+);
+
+CREATE TABLE IF NOT EXISTS addresses_tombstones (
+    guid            TEXT PRIMARY KEY,
+    time_deleted    INTEGER NOT NULL
+) WITHOUT ROWID;
+
+CREATE TABLE IF NOT EXISTS credit_cards_data (
```
mhammond, line 58: I think we should removing the CC table too if we don't see ourselves populating it in the short term - it might mislead the casual reader. It also allows us to side-step the requirements around client-side encryption of the CC number etc. Leaving it in as a comment saying like "we think the CC table might one day look like..." might be fine too.

components/forms/src/api.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    #[serde(default)]
+    pub tel: String,
+
+    #[serde(default)]
+    pub email: String,
+
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_timestamp")]
+    pub time_created: i64,
```
mhammond, line 49: I think I'd rather see a distinct type for timestamp fields - it's worked out quite well in places, and I wish I'd done that in webext-storage, particularly when integrating with desktop.

components/forms/src/error.rs
mhammond: yep - I had this comment queued too - remove anything that's not actually used :)

mhammond (changes requested):
This looks great Lougenia, nice work! As another meta comment, it looks like maybe rustfmt wasn't run? I see some odd formatting and missing line-ends at the end of files.

But this is awesome, and I agree with Lina that after another pass with our nits we should land it 🎉

components/forms/src/api.rs
lougeniaC64: Are you referring to [this](https://github.com/mozilla/application-services/blob/673e700194cfa44ad3c8629f74a2b01844cea7c3/components/places/src/types.rs)?

components/forms/src/api.rs
lougeniaC64: This was an oversight on my part but the [query_row](https://docs.rs/rusqlite/0.24.0/rusqlite/struct.Connection.html#method.query_row) function I'm using to execute my sql statement actually returns an error if no rows are found. I'm going to remove the `Option` from the return type but let me know if I should convert the `QueryReturnedNoRows` type to a custom error.

Cargo.toml
lougeniaC64: I actually really like this idea because even without the desktop baggage the word "forms" is still quite ambiguous. "Autofill" on the other hand is much more meaningful. I just landed all the local updates I was working on so I'll push a rename next.

components/forms/sql/create_shared_schema.sql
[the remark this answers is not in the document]
linacambridge: We decided at yesterday's meeting to keep these `NOT NULL`, and use empty strings to mean empty. Being able to use `String` directly, instead of an `Option<String>`—and then having to worry about what `Some("")` vs. `None` means—was the tipping point.

components/forms/sql/create_shared_schema.sql
linacambridge: Sold! 😁

components/forms/sql/create_shared_schema.sql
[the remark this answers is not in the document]
linacambridge: From yesterday's meeting: let's keep tombstones in their own table. If we want to make sure a GUID doesn't end up in both, we can write a pair of triggers on `addresses_data` and `addresses_tombstones` that check if the GUID we're inserting exists in the other table, and does a `RAISE(FAIL, ...)` if so. Keeping the tombstones inline means we have to `NULL` out all the fields, and it's still possible to have weird stuff like a row with a name, address, and `time_deleted`.

components/forms/src/api.rs
linacambridge: When I wrote this at first, I was thinking you'd need to handle the field not existing—so if it was `Option<String>`, you'd need to remember to check `Some` and `None`, but with a regular `String`, you might forget to check if `given_name.is_empty()`. But like you said yesterday, you could still end up with `Some("")`, and it makes the API so clunky to handle it that it's not a good thing 😅

components/forms/src/api.rs
linacambridge: Yep!

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+    pub email: String,
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct Address {
+    #[serde(rename = "id")]
+    pub guid: Guid,
+
+    pub address_fields: NewAddressFields,
```
linacambridge: Nit: Since the type is already called `Address`, how do you feel about `fields` instead of `address_fields`? Writing out, say, `address.address_fields.tel` feels a little redundant.

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+    pub time_last_used: i64,
+
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_timestamp")]
+    pub time_last_modified: i64,
+
+    #[serde(default)]
+    pub times_used: i64,
+
+    pub sync_change_counter: i64,
```
linacambridge: Let's not expose this as part of the API—and let's make it either `pub(crate)`, or have a test-only helper to read the change counter. I'm thinking because the change counter is an implementation detail that gets bumped whenever an address is created or modified, but consumers don't need to know that's how it works under the hood.

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+                street_address: row.get("street_address")?,
+                address_level3: row.get("address_level3")?,
+                address_level2: row.get("address_level2")?,
+                address_level1: row.get("address_level1")?,
+                postal_code: row.get("postal_code")?,
+                country: row.get("country")?,
+                tel: row.get("tel")?,
+                email: row.get("email")?,
+            };
+            Ok(Address {
```
linacambridge: Would you mind factoring this out into an `Address::from_row` helper function that `get_all_addresses` and `get_address` could use?

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+                    SELECT 1
+                    FROM   addresses_tombstones t
+                    WHERE  d.guid = t.guid
+                )
+        )",
+        &[guid.as_str()],
+        |row| row.get(0),
+    )?;
+
+    if exists {
```
linacambridge: Hmmm, I think this is missing a `DELETE FROM addresses_data` to drop the row?

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+        let saved_address = add_address(
+            &mut db,
+            NewAddressFields {
+                given_name: "jane".to_string(),
+                family_name: "doe".to_string(),
+                street_address: "123 Main Street".to_string(),
+                address_level2: "Seattle, WA".to_string(),
+                country: "United States".to_string(),
+
+                ..NewAddressFields::default()
```
linacambridge: Nothing to do with anything, but I love Rust's default initialization syntax! 😍🦀

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+            address_level1,
+            postal_code,
+            country,
+            tel,
+            email,
+            time_created,
+            time_last_used,
+            time_last_modified,
+            times_used,
+            sync_change_counter
```
linacambridge: This looks like another good place to use `COMMON_COLS`.

components/forms/src/api.rs
```diff
@@ -0,0 +1,553 @@
...
+pub fn get_address(conn: &mut Connection, guid: &Guid) -> Result<Address> {
+    let tx = conn.transaction()?;
+    let sql = format!(
+        "
+        SELECT
+            {common_cols},
+            sync_change_counter
+        FROM addresses_data d
+        WHERE guid = :guid
+            AND NOT EXISTS (
```
linacambridge: If `delete_address` deletes the row from `addresses_data` (and we rely on a trigger or something to make sure a GUID can't be in both `addresses_data` and `addresses_tombstones`), we can avoid the `NOT EXISTS` checks 😄

linacambridge (approved):
Great! I left a few comments, but looks wonderful on the whole, let's :shipit: and iterate!

components/forms/src/api.rs
lougeniaC64: I honestly wasn't sure how this worked so thank you for the clarification. And for sure, as you stated below, this simplifies the "exists" check.

components/forms/src/api.rs
lougeniaC64: Since this would be used in more than one component should it the types logic be moved out of the places component? Would [support](https://github.com/mozilla/application-services/tree/673e700194cfa44ad3c8629f74a2b01844cea7c3/components/support) be a good place for it?

mhammond: `support` sounds perfect!

lougeniaC64:
FYI: I didn't include updating the places component to use the Timestamp logic that I copied into the support component because this PR is big enough as is. I'll handle that in a follow-on PR. 😬

mhammond (approved):
Lovely work Lougenia - let's get it landed!

PR #3585, opened 2020-09-17
Handle retry loops on authorization status

### LINKED ISSUE #3401: Prevent endless retry loops in fxa-client
We recently had a server-side misconfiguration that causes Fenix to get into an [infinite retry loop](https://docs.google.com/document/d/1dYHO1RYFVy9O-nqxXSz7GOi1xiJweVZQY4WgaOqeAtg/), where it would:

1. Get a 401 from the sync storage node, and enter its "recover from auth error" state.
2. Call `checkAuthorizationStatus` on FxA and see that it was still connected to the account.
3. Make a new oauth token, use it to authenticate to the sync tokenserver, and try to access its sync storage node again.
4. GOTO (1).

@grigoryk landed a fix for this [in the Kotlin code](https://github.com/mozilla-mobile/android-components/pull/7580) by implementing a "circuit breaker", basically refusing to execute the `checkAuthorizationStatus` step more than a certain number of times in a certain time period.

We should push this logic down into the shared rust code so that other consumers are not vulnerable to the same issue. For example, the swift state-machine is probably harboring a similar retry loop.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1521)
┆Story Points: 3
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-09-25

vladikoff:
Fixes #3401

### DIFF
--- a/components/fxa-client/src/http_client.rs
+++ b/components/fxa-client/src/http_client.rs
@@ -1146,7 +1146,7 @@ mod tests {
             assert_eq!(*backoff_end_duration, Duration::from_secs(1_000_000));
 
             let path2 = format!("{}/{}", mockito::server_url(), "v1/account/device/commands");
-            // Hacky way to drop the mutex gaurd, so that the next call to
+            // Hacky way to drop the mutex guard, so that the next call to
             // client.make_request doesn't hang or panic
             std::mem::drop(state);
             let second_request = Request::get(Url::parse(&path2).unwrap());
--- a/components/fxa-client/src/lib.rs
+++ b/components/fxa-client/src/lib.rs
@@ -8,15 +8,14 @@
 use crate::{
     commands::send_tab::SendTabPayload,
     device::Device,
-    oauth::{OAuthFlow, OAUTH_WEBCHANNEL_REDIRECT},
+    oauth::{AuthCircuitBreaker, OAuthFlow, OAUTH_WEBCHANNEL_REDIRECT},
     scoped_keys::ScopedKey,
     state_persistence::State,
 };
 pub use crate::{
     config::Config,
     error::*,
-    oauth::IntrospectInfo,
-    oauth::{AccessTokenInfo, RefreshToken},
+    oauth::{AccessTokenInfo, IntrospectInfo, RefreshToken},
     profile::Profile,
     telemetry::FxaTelemetry,
 };
@@ -68,6 +67,7 @@ pub struct FirefoxAccount {
     flow_store: HashMap<String, OAuthFlow>,
     attached_clients_cache: Option<CachedResponse<Vec<http_client::GetAttachedClientResponse>>>,
     devices_cache: Option<CachedResponse<Vec<http_client::GetDeviceResponse>>>,
+    auth_circuit_breaker: AuthCircuitBreaker,
     // 'telemetry' is only currently used by `&mut self` functions, but that's
     // not something we want to insist on going forward, so RefCell<> it.
     telemetry: RefCell<FxaTelemetry>,
@@ -81,6 +81,7 @@ impl FirefoxAccount {
             flow_store: HashMap::new(),
             attached_clients_cache: None,
             devices_cache: None,
+            auth_circuit_breaker: Default::default(),
             telemetry: RefCell::new(FxaTelemetry::new()),
         }
     }
@@ -236,7 +237,7 @@ impl FirefoxAccount {
         }
     }
 
-    /// Disconnect from the account and optionaly destroy our device record. This will
+    /// Disconnect from the account and optionally destroy our device record. This will
     /// leave the account object in a state where it can eventually reconnect to the same user.
     /// This is a "best effort" infallible method: e.g. if the network is unreachable,
     /// the device could still be in the FxA devices manager.
--- a/components/fxa-client/src/oauth.rs
+++ b/components/fxa-client/src/oauth.rs
@@ -20,13 +20,13 @@ use std::{
     time::{SystemTime, UNIX_EPOCH},
 };
 use url::Url;
-
 // If a cached token has less than `OAUTH_MIN_TIME_LEFT` seconds left to live,
 // it will be considered already expired.
 const OAUTH_MIN_TIME_LEFT: u64 = 60;
 // Special redirect urn based on the OAuth native spec, signals that the
 // WebChannel flow is used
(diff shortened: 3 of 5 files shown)


components/fxa-client/src/oauth.rs
```diff
@@ -871,6 +915,45 @@ mod tests {
         assert_eq!(auth_status.active, true);
     }
 
+    #[test]
+    fn test_check_authorization_status_circuit_breaker() {
```
vladikoff, line 919: hm might be worth adding multiple tests for this

components/fxa-client/src/oauth.rs
```diff
@@ -93,11 +107,41 @@ impl FirefoxAccount {
...
-    pub fn check_authorization_status(&self) -> Result<IntrospectInfo> {
+    pub fn check_authorization_status(&mut self) -> Result<IntrospectInfo> {
         let resp = match self.state.refresh_token {
-            Some(ref refresh_token) => self
-                .client
-                .oauth_introspect_refresh_token(&self.state.config, &refresh_token.token)?,
+            Some(ref refresh_token) => {
+                match self
+                    .client
+                    .oauth_introspect_refresh_token(&self.state.config, &refresh_token.token)
```
rfk, line 115: In the specific bug we saw most recently in Fenix, this `oauth_introspect_refresh_token` call was actually returning `200 OK`, and it was the downstream call to the sync storage node that was giving the 401s, causing Fenix to try calling `check_authorization_status` over and over again.

Luckily, I think that makes things simpler for us here - we don't have to match on the response code, we just need to rate-limit *all* calls to this method if they occur within the configured time window. So I think you can just pull your circuit-breaker logic up to the top of the function, like this:

```
fn check_authorization_status(...) {
  // Increment `attempts` etc.
  // Return an error if we've exceeded the limit.
  // If below the limit, proceed with the body of the function as before.
}
```

components/fxa-client/src/lib.rs
```diff
@@ -68,6 +68,7 @@ pub struct FirefoxAccount {
     flow_store: HashMap<String, OAuthFlow>,
     attached_clients_cache: Option<CachedResponse<Vec<http_client::GetAttachedClientResponse>>>,
     devices_cache: Option<CachedResponse<Vec<http_client::GetDeviceResponse>>>,
+    authorization_cache: Option<oauth::AuthorizationCache>,
```
rfk, line 71: Consider implementing `Default` and initializing this field via `Default::default()`, rather than making it an `Option` and initializing to `None`. I can't think of a situation where we'd want to have a `None` in here on purpose (although as always, am open to suggestion!).

components/fxa-client/src/oauth.rs
```diff
@@ -20,13 +20,27 @@ use std::{
...
+// time allowing for a few 401s to hit the state machine in a quick succession.
+// For example, initializing the account state machine & syncing after letting our access tokens expire
+// due to long period of inactivity will trigger a few 401s, and that shouldn't be a cause for concern.
+const AUTH_CHECK_CIRCUIT_BREAKER_COUNT: u64 = 10;
 // Special redirect urn based on the OAuth native spec, signals that the
 // WebChannel flow is used
 pub const OAUTH_WEBCHANNEL_REDIRECT: &str = "urn:ietf:wg:oauth:2.0:oob:oauth-redirect-webchannel";
+
+#[derive(Clone, Copy)]
+pub(crate) struct AuthorizationCache {
```
rfk, line 39: Naming nit: for me, the name "cache" doesn't create the right mental model; caching tends to imply some transient data that's kept around for performance reasons, but that's not what we're doing here.

I suggest leaning in to the terminology and calling it an `AuthCircuitBreaker` or similar. Moving some of the increment-and-check logic into a helper method might also aid future readers in understanding what's happening here. Strawman suggestion:

```
struct AuthCircuitBreaker { ... }

impl AuthCircuitBreaker {
  fn check(&self) -> Result<()> {
    // Increment `self.attempts`, reset the timestamp, etc.
    // Return an error if `self.attempts` exceeds the threshold.
  }
}
```

And then in your `check_authorization_status` method:

```
impl FirefoxAccount {
  fn check_authorization_status(&mut self) -> Result<IntrospectInfo> {
    self.auth_circuit_breaker.check()?;
    // The rest of the method body, as before.
  }
}
```

components/fxa-client/src/oauth.rs
```diff
@@ -456,6 +458,47 @@ impl FirefoxAccount {
...
+        if self.tokens == 0 {
+            return Err(ErrorKind::AuthCircuitBreakerError.into());
+        }
+        self.tokens -= 1;
+        Ok(())
+    }
+
+    fn refill(&mut self) {
+        let now = util::now();
+        let new_tokens = ((now - self.last_refill) as f64 * self.renewal_rate as f64) as u8; // `as` is a saturing cast.
```
rfk, line 494: It's also important that it's a *truncating* cast.

components/fxa-client/src/oauth.rs
```diff
@@ -871,6 +914,57 @@ mod tests {
...
+        fxa.set_client(Arc::new(client));
+
+        for _ in 0..5 {
+            assert!(fxa.check_authorization_status().is_ok());
+        }
+        match fxa.check_authorization_status() {
+            Ok(_) => unreachable!("should not happen"),
+            Err(err) => assert!(matches!(err.kind(), ErrorKind::AuthCircuitBreakerError)),
+        }
+    }
```
rfk, line 966: Ideally, I think we'd also test that the access is allowed again after more tokens accumulate in the bucket. Are we set up to mock `util::now` in order to test this easily?

rfk: Also you wouldn't necessary have to mock out all of the request infra to this this, just make a more focused unit-test on the behaviour of `check()`.

rfk (approved):
Looks good, thanks!

Please consider adding a test that we correctly recover from the circit-breaker error state, and adding a changelog entry for completeness.

components/fxa-client/src/http_client.rs
```diff
@@ -939,7 +939,7 @@ pub struct OAuthAuthResponse {
     pub state: String,
 }
 
-#[derive(Deserialize)]
+#[derive(Clone, Deserialize)]
```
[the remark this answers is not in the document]
eoger: This was for mockito's `returns` method, but we can't use that because `Error` is not `Clone`, so I'll revert

PR #3586, opened 2020-09-17
[nobug, build] Update xcode from 11.7 to 12.0

garvankeeley:
On `main` branch of firefox-ios we are building with Xcode 12 which was just released.
This carthage issue with Xode 12 `lipo` (which creates multi-arch binaries) requires a workaround that I added https://github.com/Carthage/Carthage/issues/3019


rfk (approved):
👍

PR #3589, opened 2020-09-18
Port the sql function note_sync_change() from desktop.

mhammond:
This allows us to avoid applying the result of a bookmark sync merge
if the local tree has changed since we fetched it.

It's named `note_bookmarks_sync_change()` here to better
reflect the fact that history and bookmarks are much tighter
here than on desktop and it applies only to bookmarks, but
otherwise it is as close to the desktop implementation as
possible.

I *think* I got the `Ordering` stuff correct, but probably didn't :) `Ordering::SeqCst` doesn't seem necessary, although it's the safe choice if we aren't sure :)

Note that this is really in support of #3510, but I felt it should land separately

### DIFF
--- a/components/places/src/db/db.rs
+++ b/components/places/src/db/db.rs
@@ -5,14 +5,25 @@
 use super::schema;
 use crate::api::places_api::ConnectionType;
 use crate::error::*;
+use lazy_static::lazy_static;
 use rusqlite::Connection;
 use sql_support::{ConnExt, SqlInterruptHandle, SqlInterruptScope};
+use std::collections::HashMap;
 use std::ops::Deref;
 use std::path::Path;
 
-use std::sync::{atomic::AtomicUsize, Arc, Mutex};
+use std::sync::{
+    atomic::{AtomicI64, AtomicUsize, Ordering},
+    Arc, Mutex, RwLock,
+};
 
 pub const MAX_VARIABLE_NUMBER: usize = 999;
+
+lazy_static! {
+    // Each API has a single bookmark change counter shared across all connections.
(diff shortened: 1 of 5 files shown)


mhammond:
go home @codecov-commenter, you're clearly drunk 🍷 :)

components/places/src/db/db.rs
```diff
@@ -367,6 +426,26 @@ pub(crate) mod sql_fns {
...
+            // Because we only ever check for equality, we can use Relaxed ordering.
+            return Ok(counter.fetch_add(1, Ordering::Relaxed));
+        }
+        // Need to add the counter to the map - drop the read lock before
+        // taking the write lock.
+        drop(map);
+        let mut map = GLOBAL_BOOKMARK_CHANGE_COUNTERS
+            .write()
+            .expect("gbcc poisoned");
+        let counter = map.entry(api_id).or_insert_with(|| AtomicI64::new(0));
```
linacambridge, line 445:
```suggestion
        let counter = map.entry(api_id).or_default();
```

components/places/src/storage/bookmarks.rs
```diff
@@ -1660,6 +1660,7 @@ mod tests {
         conn.execute("UPDATE moz_bookmarks SET syncChangeCounter = 0", NO_PARAMS)
             .expect("should work");
 
+        let global_change_tracker = conn.global_bookmark_change_tracker();
```
linacambridge, line 1663: Should we check if it’s `!changed()` before, for good measure?

linacambridge (approved):
Excellent, thanks @mhammond!

PR #3590, opened 2020-09-18
Upgrade Glean to v32.4.0

badboy:
Glean changelog:

* General
  * Allow using quantity metric type outside of Gecko ([#1198](https://github.com/mozilla/glean/pull/1198))
  * Update `glean_parser` to 1.28.5
    * The `SUPERFLUOUS_NO_LINT` warning has been removed from the glinter. It likely did more harm than good, and makes it hard to make metrics.yaml files that pass across different versions of `glean_parser`.
    * Expired metrics will now produce a linter warning, `EXPIRED_METRIC`.
    * Expiry dates that are more than 730 days (~2 years) in the future will produce a linter warning, `EXPIRATION_DATE_TOO_FAR`.
    * Allow using the Quantity metric type outside of Gecko.
    * New parser configs `custom_is_expired` and `custom_validate_expires` added. These are both functions that take the expires value of the metric and return a bool. (See `Metric.is_expired` and `Metric.validate_expires`). These will allow FOG to provide custom validation for its version-based `expires` values.
  * Add a limit of 250 pending ping files. ([#1217](https://github.com/mozilla/glean/pull/1217)).

Note: This also gets rid of the 2 workarounds (removed code) in
AppService thanks to upstream changes.

### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - `automation/all_tests.sh` runs to completion and produces no failures
  - Note: For changes that need extra cross-platform testing, consider adding `[ci full]` to the PR title.
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one
  - Any breaking changes to Swift or Kotlin binding APIs are noted explicitly
- [ ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)
  - Any new dependencies are accompanied by a summary of the due dilligence applied in selecting them.

PR #3597, opened 2020-09-23
Run cargo update [ci full]

eoger:
It's been a while since we've done one of these.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,15 +35,15 @@ dependencies = [
 
 [[package]]
 name = "android_log-sys"
-version = "0.1.2"
+version = "0.2.0"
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -1074,9 +1074,9 @@ dependencies = [
 
 [[package]]
 name = "glean-ffi"
-version = "32.3.0"
+version = "32.4.0"
```
rfk, line 1077: Semver says these should be safe, but do we have anything to worry about with these being updated without a bump in the glean submodule? /cc @badboy

rfk (approved):
LGTM, although I have one raised eyebrow about the glean updates...

rfk:
The glean thing was covered in https://github.com/mozilla/application-services/pull/3590

Issue #3632, opened 2020-10-07
Update rusqlite to 0.24.1 and (including vendored version in mozilla-central)


thomcc:
Apparently one of rusqlite's dependencies causes crashes if you build Firefox with nightly Rust:

- https://github.com/rust-lang/rust/issues/77585
- https://github.com/rusqlite/rusqlite/issues/810
- https://bugzilla.mozilla.org/show_bug.cgi?id=1669722

I released v0.24.1 which avoids this dependency entirely: https://github.com/rusqlite/rusqlite/releases/tag/v0.24.1

It would be nice if someone updated rusqlite in m-c with this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1829)
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-10-23

mhammond:
Thanks heaps for this Thom - I think it's done (assuming that bugzilla patch sticks)

Issue #3627, opened 2020-10-05
Sync integration tests broken


eoger:
See https://app.circleci.com/pipelines/github/mozilla/application-services/17594/workflows/4fb4b14b-5163-42f3-9afd-da5222d1ffda/jobs/82679 for a failing run.

This is coming from FxA, see https://github.com/mozilla/fxa/issues/6609.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1824)
┆Sprint: Backlog

eoger:
Solved

PR #3633, opened 2020-10-08
Update rusqlite to 0.24.1 and libsqlite3-sys to 0.20.1

mhammond:
As @thomcc kindly flagged in #3632 - once approved I'll merge this into the vendor-mc/78 and then re-vendor that branch into m-c.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


rfk (approved):
👍

rfk:
@mhammond oh, looks like you need to regenerate the dependency list using `./tools/regenerate_dependency_summaries.sh`; LMK if it gives you any trouble and I can take a look, sometimes it only works correctly when run on a Mac, for reasons.

mhammond:
> sometimes it only works correctly when run on a Mac, for reasons.

Right - the current failure is for packages used only by the iOS megazord :/ I can fix that.

mhammond:
Test failure is #3627, so merging...

PR #3634, opened 2020-10-08
Update rusqlite to 0.24.1 and libsqlite3-sys to 0.20.1

mhammond:


rfk (approved):

mhammond:
clippy has nss failures. dependency script doesn't even run on this branch. swift linter issues aren't relevant on desktop, so I've no plan on fixing any of those failures.

PR #3637, opened 2020-10-09
Fix clippy complaining about `matches!()`.

rfk:
CI is currently broken because we treat clippy warnings as errors; this should fix it.

### DIFF
--- a/components/fxa-client/src/migrator.rs
+++ b/components/fxa-client/src/migrator.rs
@@ -214,20 +214,11 @@ mod tests {
         FirefoxAccount::with_config(config)
     }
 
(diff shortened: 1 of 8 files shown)


mhammond:
This looks fine, but I was interested to know what clippy was on about, and `main` doesn't seem to have clippy failures and I don't see any here either?

rfk:
https://github.com/mozilla/application-services/pull/3636 has some in its CI runs; I also had some locally but not as many as in CI, possibly my local rust is a bit out of date

mhammond (approved):
huh - yeah, rustup and I do see that - must have been very recent as I think I did that not long ago.

PR #3647, opened 2020-10-14
Upgrade JNA to 5.6.0 [ci full]

rfk:
What it says on the tin ^


rfk:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1668527#c11 for some discussion about using this new version, which is also used by Glean.

rfk:
The sync integration test failures here are due to FxA infra issues.

eoger (approved):

PR #3668, opened 2020-10-27
Fix how webext-storage handles incoming tombstones.

mhammond:
https://bugzilla.mozilla.org/show_bug.cgi?id=1667985

This is a large patch because of a fundamental assumption that turns out
to be invalid and had to be un-done - that we can determine which extension
a tombstone is for. Because a tombstone record has only the GUID and the
deleted flag this isn't possible (and syncing for clients with a tombstone
was broken by that assumption)

This had wide ramifications in the code. Of note:

* The mirror table in the database previously had a not-null constraint
  for the ext-id column, but that's clearly not valid for tombstones.
  Due to sqlite limitations, this means a non-trivial migration (rename
  the old table, create the new one, populate new from old, delete old)

* Same is true for the "staging" table, but that's a temp table, so no
  migration was necessary.

* As a result, some queries got a little more complex - they had to join
  via the GUID to the mirror/stage tables.

* `IncomingItem` used to have a GUID and ext-id - but it can
  no longer assume a valid ext-id. Rather than change this to
  an `Option<>` (this meaning it could be None in contexts where
  it logically could not), I took the hit of moving it to a better place.
  This left `IncomingItem` being a wrapper for a GUID, so it's
  been deleted and replaced with a plain GUID.

* The `ext-id` that was in `IncomingItem` is now in the IncomingState
  variants - every variant except the one representing an incoming tombstone.

* Similarly for `IncomingAction` - this now carries the ext-id for most
  variants, and had a number of flow-on effects - eg it's a new param
  to merge, etc, and many tests needed to be touched, although
  most in obvious ways.

* There are new tests for all of this.

### DIFF
--- a/components/webext-storage/sql/create_schema.sql
+++ b/components/webext-storage/sql/create_schema.sql
@@ -34,12 +34,17 @@ CREATE TABLE IF NOT EXISTS storage_sync_mirror (
     /* The extension_id is explicitly not the GUID used on the server.
        It can't be  a regular foreign-key relationship back to storage_sync_data
        as items with no data on the server (ie, deleted items) will not appear
-       in storage_sync_data.
+       in storage_sync_data, and the guid isn't in that table either.
+       It must allow NULL as tombstones do not carry the ext_id, so we have
+       an additional CHECK constraint.
     */
-    ext_id TEXT NOT NULL UNIQUE,
+    ext_id TEXT UNIQUE,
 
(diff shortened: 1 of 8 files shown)


mhammond:
@lougeniaC64 this is a large patch and your first real exposure to this engine, so it probably makes sense to have a chat for me to walk you through it - maybe have a first pass then we can arrange a good time.

lougeniaC64 (approved):

PR #3669, opened 2020-10-27
WIP combined Glean + Nimbus + AppServices megazord, aka The GleaNimbuZord.

rfk:
This is still at "proof of concept" stage, but I think it's ready for concrete feedback. The basic premise is as outlined in [this doc](https://docs.google.com/document/d/1oUDbQTKzJuapzHyPvtAgFgpCKe3V-dhsYSCI62Mw4Zo/):

* Import `glean` and `nimbus-sdk` repos as git submodules under `./components/external`.
* Include `glean_ffi` and `nimbus` Rust code as part of `./megazords/full` so that it all gets compiled together.
* Create wrapper components `./components/glean` and `./components/nimbus` that repackage *just* the Kotlin code, and depend on the full-megazord for the compiled Rust code:
    * `org.mozilla.appservices:glean` is a drop-in replacement for `org.mozilla.telemetry:glean`.
    * `org.mozilla.appservices:nimbus` is a drop-in replacement for `org.mozilla.experiments:nimbus`.

Things that work:

* `./gradlew publishToMavenLocal` publishes our re-packaged Glean and Nimbus packages, and they do appear to function correctly as drop-in replacements when [tried out in a-c](https://github.com/mozilla-mobile/android-components/pull/8836).
* Running `./gradlew glean:test` runs the glean testsuite against the megazorded Rust code; 8 tests currently fail.

Things that need work:

* Initializing git submodules in taskcluster
* Installing uniffi-bindgen in taskcluster (which I'm not able to test because of the above)
* Get all the Glean tests passing
* See if there's anything useful we can do to test the repackaged Nimbus
* Nimbus produces some errors in our dependency checker script.

Things I want to do before we decide whether to move forward with this:

* Compare overall compile time with the addition of the new components.
* Compare overall test runtime with the addition of the new components.
* Compare compiled artifact size relative to the size of separate AppServices and Glean libraries.


rfk:
Ref https://github.com/mozilla/glean/pull/1290 for a small glean patch that's required in order to make this work.

rfk:
Ref https://github.com/mozilla-mobile/android-components/pull/8836 for a small a-c patch that tweaks it to use the gleanimbuzord rather than the standalone glean package.

rfk:
This is still at "proof of concept" stage, but I think it's ready for concrete feedback, and I've updated the bug description with some details about what's going on here. @mhammond @eoger @badboy what do you think so far?

Cargo.toml
```diff
@@ -45,6 +45,13 @@ members = [
     "testing/separated/*/",
 ]
 
+# Don't include git submodules in our workspace, since those packages haven't
+# been written with that in mind. They might work, but they might not.
+exclude = [
+    "components/external/glean/",
+    "components/external/nimbus-sdk/",
+]
```
rfk, line 53: By excluding these from the Cargo workspace, I appear to be able to use a path dependency for the Rust code rather than fetching it from crates.io, which in theory means there's less chance of us using incompatible versions of the Native code and Kotlin code.

build.gradle
```diff
@@ -3,8 +3,8 @@
 buildscript {
     ext.kotlin_version = '1.3.71'
     ext.jna_version = '5.6.0'
-    ext.android_gradle_plugin_version = '3.6.0'
-    ext.android_components_version = '56.0.0'
+    ext.android_gradle_plugin_version = '4.0.1'
+    ext.android_components_version = '63.0.0'
```
rfk, line 7: For compatibility with the version of glean-gradle-plugin that is in the submodule.

components/glean/build.gradle
```diff
@@ -0,0 +1,201 @@
+// =============================================
+// We're doing something a bit weird here, so
+// READ THIS CAREFULLY before editing this file.
+// =============================================
+//
+// This is a lightly-modified version of the `build.gradle` file from Glean,
+// which you can find at `../external/glean/glean-core/android/build.gradle`.
```
rfk, line 7: I don't like the fact that this has copied a lot of logic from the Glean build.gradle, seems like it could easily get out of sync. But, I also don't have a better idea. We could possibly consider trying to move some of this logic into the upstream build.gradle, somehow...?

components/glean/build.gradle
```diff
@@ -0,0 +1,201 @@
...
+
+android {
+    ndkVersion rootProject.ext.build.ndkVersion
+    compileSdkVersion rootProject.ext.build.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion rootProject.ext.build['minSdkVersion']
+        targetSdkVersion rootProject.ext.build['targetSdkVersion']
+
+        buildConfigField("String", "JNA_LIBRARY_NAME", "\"megazord\"")
```
rfk, line 43: Ref https://github.com/mozilla/glean/pull/1290 for the glean-side change required to make this work.

components/glean/build.gradle
```diff
@@ -0,0 +1,201 @@
...
+            consumerProguardFiles "$rootDir/proguard-rules-consumer-jna.pro"
+        }
+    }
+
+    // We don't have any source files of our own, we're re-packaging the ones
+    // from the Glean repo (which is included as a submodule in our repo).
+    sourceSets {
+        main.java.srcDirs += "../external/glean/glean-core/android/src/main/java"
+        test.java.srcDirs += "../external/glean/glean-core/android/src/test"
+        test.resources.srcDirs += "../external/glean/glean-core/android/src/test/resources"
```
rfk, line 65: This also lets us run Glean's android-level tests as part of our own `./gradlew test` run.

components/glean/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="mozilla.telemetry.glean" />
```
rfk, line 2: This is important, it means that our re-packaged `glean` package will export the same Kotlin package as the standalone Glean.

components/nimbus/build.gradle
```diff
@@ -0,0 +1,150 @@
...
+
+android {
+    ndkVersion rootProject.ext.build.ndkVersion
+    compileSdkVersion rootProject.ext.build.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion rootProject.ext.build['minSdkVersion']
+        targetSdkVersion rootProject.ext.build['targetSdkVersion']
+
+        buildConfigField("String", "JNA_LIBRARY_NAME", "\"megazord\"")
```
rfk, line 30: TODO: we'll need to add something like https://github.com/mozilla/glean/pull/1290 in Nimbus as well in order for this to work.

components/nimbus/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="uniffi.nimbus" />
```
rfk, line 2: Also I feel like we wanted to change the name of this package, but I don't recall where we landed on that front.

megazords/ios/rust/Cargo.toml
```diff
@@ -15,4 +15,4 @@ places-ffi = { path = "../../../components/places/ffi" }
 rc_log_ffi = { path = "../../../components/rc_log" }
 viaduct = { path = "../../../components/viaduct" }
 viaduct-reqwest = { path = "../../../components/support/viaduct-reqwest" }
-glean-ffi = { version = "32.3.0" }
+glean-ffi = { path = "../../../components/external/glean/glean-core/ffi" }
```
rfk, line 18: Note that this is now a path dependency. Not sure if that will cause any problems in practice...

components/glean/build.gradle
mhammond: I'm not sure what "upstream" you refer to here, and without looking at the differences carefully I'm not sure how practical this is, but working out how to split this and use `apply from:` seems worthwhile - it scares me a little that the README above says "update `./build.gradle` to track any build changes made in Glean" and then discover you mean track stuff in this 200 line semi-complicated file and where it's not at all obvious how it relates to glean's original.

(I don't have a concrete ask here, just expressing a hope that @badboy might have ideas here :)

components/nimbus/build.gradle
```diff
@@ -0,0 +1,150 @@
+// =============================================
```
mhammond, line 1: at least this is only 150 lines? 😭

components/nimbus/src/main/AndroidManifest.xml
mhammond: I think we should - I opened https://jira.mozilla.com/browse/SDK-95

mhammond (commented):
This looks absolutely great to me and I have nothing meaningful to add!

components/glean/build.gradle
rfk: Right, I meant "upstream" as in "into some gradle files that live in the Glean repo and are maintained by the Glean team" in a very abstract sense

components/nimbus/build.gradle
rfk: For Nimbus, this is currently taken care of in the android-components wrapper, which sets a system property to tell Nimbus where to look for its shared library.

components/logins/android/build.gradle
```diff
@@ -113,6 +113,7 @@ dependencies {
     implementation 'com.google.protobuf:protobuf-javalite:3.11.4'
     api project(":full-megazord")
     implementation project(":native-support")
+    implementation project(":glean")
```
rfk, line 116: Each component that uses Glean should take a direct dependency on our re-packaged version of it. This prevents apps from trying to use the stand-alone glean in combination with an appservices component (by way of producing an inscrutable build error about duplicate class definitions).

rfk:
I tried this out in a local Fenix build by doing:

* Checkout android-components and use its `local.properties` file to point it to the local checkout of appservices.
* Checkout fenix and its its `local.properties` file to point it to the local checkouts of a-c and appservices.
* Hackily edit `app/src/main/java/org/mozilla/fenix/FenixApplication.kt` to change `enableStrictMode(true)` to `enableStrictMode(false)`, to prevent the built app from crashing out on unrelated safe-mode violations.
* `./gradlew app:assembleDebug` in the fenix repo and install the resulting APK.

AFAICT it all seemed to work as expected!

* The resulting app did not include `libglean_ffi.so`.
* The resulting `libmegazord.so` correctly included the glean symbols.
* I observed glean doing its thing in the `adb logcat` output, although the log lines were slightly differently formatted because they were now being routed through our shared `rc_log`.

I also tried simulating a local glean development flow by overriding the version of glean used by Fenix, with:

```
configurations.all {
    resolutionStrategy.dependencySubstitution {
        substitute module('org.mozilla.appservices:glean') with module("org.mozilla.telemetry:glean:${project.ext.glean_version}")
    }
}   
```

Again, this seemed to work as I expected, pulling in and using the standalone version of Glean while ignoring the version bundled with the megazord.

rfk:
When we do end up megazording with Glean on Android, I expect it's going to have to be in mozilla-central. So let's close this experiment out.

Issue #3662, opened 2020-10-21
Implement throttling and backoffs for autopush endpoints


jonalmeida:
In Android Components, we [implement our own throttling delegate](https://github.com/mozilla-mobile/android-components/blob/3f1d5499dbb7d966d3c12f127bf81065f5716b11/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt#L227-L232) to avoid hitting the Autopush and FxA servers too frequently. This would benefit from living in a-s so other clients do not have to do this as well.

I see [we have a ticket](https://github.com/mozilla/application-services/issues/1216) to do this for sync manager, but nothing that specifically calls out push servers. Please close this as a duplicate of the other if we want to merge these two together.

cc: @jrconlin

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1852)
┆Story Points: 5
┆Epic: a-s consumer issues
┆Sprint: SYNC - end 2020-11-20

PR #3673, opened 2020-10-28
Rate limit for update push endpoint

eoger:
Fixes https://github.com/mozilla/application-services/issues/3662.

This implements the rate limiting algorithm from [Android components](https://github.com/mozilla-mobile/android-components/blob/3f1d5499dbb7d966d3c12f127bf81065f5716b11/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt#L227-L232).

### DIFF
--- a/components/fxa-client/src/oauth.rs
+++ b/components/fxa-client/src/oauth.rs
@@ -11,6 +11,7 @@ use crate::{
     util, FirefoxAccount,
 };
 use jwcrypto::{EncryptionAlgorithm, EncryptionParameters};
+use rate_limiter::RateLimiter;
 use rc_crypto::digest;
 use serde_derive::*;
 use std::convert::TryFrom;
@@ -461,55 +462,28 @@ impl FirefoxAccount {
 const AUTH_CIRCUIT_BREAKER_CAPACITY: u8 = 5;
 const AUTH_CIRCUIT_BREAKER_RENEWAL_RATE: f32 = 3.0 / 60.0 / 1000.0; // 3 tokens every minute.
 
-// The auth circuit breaker rate-limits access to the `oauth_introspect_refresh_token`
-// using a fairly naively implemented token bucket algorithm.
 #[derive(Clone, Copy)]
 pub(crate) struct AuthCircuitBreaker {
-    tokens: u8,
-    last_refill: u64, // in ms.
+    rate_limiter: RateLimiter,
 }
 
 impl Default for AuthCircuitBreaker {
     fn default() -> Self {
         AuthCircuitBreaker {
-            tokens: AUTH_CIRCUIT_BREAKER_CAPACITY,
-            last_refill: Self::now(),
+            rate_limiter: RateLimiter::new(
+                AUTH_CIRCUIT_BREAKER_CAPACITY,
+                AUTH_CIRCUIT_BREAKER_RENEWAL_RATE,
+            ),
         }
     }
 }
 
 impl AuthCircuitBreaker {
     pub(crate) fn check(&mut self) -> Result<()> {
-        self.refill();
-        if self.tokens == 0 {
+        if !self.rate_limiter.check() {
             return Err(ErrorKind::AuthCircuitBreakerError.into());
         }
-        self.tokens -= 1;
         Ok(())
-    }
-
-    fn refill(&mut self) {
-        let now = Self::now();
-        let new_tokens =
-            ((now - self.last_refill) as f64 * AUTH_CIRCUIT_BREAKER_RENEWAL_RATE as f64) as u8; // `as` is a truncating/saturing cast.
-        if new_tokens > 0 {
-            self.last_refill = now;
-            self.tokens = std::cmp::min(
-                AUTH_CIRCUIT_BREAKER_CAPACITY,
-                self.tokens.saturating_add(new_tokens),
-            );
-        }
-    }
-
-    #[cfg(not(test))]
-    #[inline]
-    fn now() -> u64 {
-        util::now()
-    }
-
-    #[cfg(test)]
-    fn now() -> u64 {
-        1600000000000
     }
 }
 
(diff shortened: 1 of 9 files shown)


components/support/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+        if new_tokens > 0 {
+            self.last_refill = now;
+            self.tokens = std::cmp::min(self.capacity, self.tokens.saturating_add(new_tokens));
+        }
+    }
+
+    #[cfg(not(test))]
+    #[inline]
+    fn now() -> u64 {
+        // util::now()
```
rfk, line 46: Just putting an explicit note on this commented-out bit so we don't forget about it.

components/fxa-client/src/oauth.rs
```diff
@@ -461,56 +462,29 @@ impl FirefoxAccount {
 const AUTH_CIRCUIT_BREAKER_CAPACITY: u8 = 5;
 const AUTH_CIRCUIT_BREAKER_RENEWAL_RATE: f32 = 3.0 / 60.0 / 1000.0; // 3 tokens every minute.
 
-// The auth circuit breaker rate-limits access to the `oauth_introspect_refresh_token`
-// using a fairly naively implemented token bucket algorithm.
 #[derive(Clone, Copy)]
 pub(crate) struct AuthCircuitBreaker {
-    tokens: u8,
-    last_refill: u64, // in ms.
+    rate_limiter: RateLimiter,
```
rfk, line 467: Nice refactor 👍

components/support/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::RateLimiter;
+    #[test]
+    fn test_auth_circuit_breaker_unit_recovery() {
+        let capacity = 10;
+        let renewal_rate = 1.0 / 1000.0;
+        let mut breaker = RateLimiter::new(capacity, renewal_rate);
+        // AuthCircuitBreaker::now is fixed for tests, let's assert that for sanity.
```
rfk, line 64: nit: need to update the name in the comment here to the generic `RateLimiter`.

components/push/src/subscriber.rs
```diff
@@ -175,6 +188,108 @@ impl PushManager {
...
+        (
+            format!("ratelimit_{}_timestamp", &self.op_name),
+            format!("ratelimit_{}_count", &self.op_name),
+        )
+    }
+
+    fn get_counters(&self, store: &Store) -> (u64, u16) {
+        (
+            store
+                .get_meta(&format!("ratelimit_{}_timestamp", self.op_name))
```
rfk, line 256: Please use `db_meta_keys` here for consistency with `persist_counters` below.

components/push/src/subscriber.rs
```diff
@@ -175,6 +188,108 @@ impl PushManager {
...
+                .unwrap_or_default(),
+            store
+                .get_meta(&format!("ratelimit_{}_count", self.op_name))
+                .ok()
+                .flatten()
+                .map(|s| s.parse())
+                .transpose()
+                .ok()
+                .flatten()
+                .unwrap_or_default(),
```
rfk, line 272: FWIW, I had trouble understanding the sequence of iterator/result operations here. I *think* I have my head around it I wonder if it could be clearer.

One possibility could be to consider a generic version of `get_meta` that has the parsing logic included, something like (untested!):

```
fn get_meta_or_default<T: FromStr + Default>(&self, key: &str) -> T {
  // generic flatten-transpose-etc here.
}
```

components/push/src/subscriber.rs
```diff
@@ -175,6 +188,108 @@ impl PushManager {
...
+}
+
+impl PersistedRateLimiter {
+    pub fn new(op_name: &str, periodic_interval: u64, max_requests_in_interval: u16) -> Self {
+        Self {
+            op_name: op_name.to_owned(),
+            periodic_interval,
+            max_requests_in_interval,
+        }
+    }
```
rfk, line 207: Seeing the rate-limiter refactor above led me to expect that you'd use that newly-refactored `RateLimiter` struct here, but you aren't. Is that because the algorithm that you want is different, or because it was hard to re-use when combined with this persistence abstraction?

components/push/src/subscriber.rs
```diff
@@ -175,6 +188,108 @@ impl PushManager {
...
+
+    fn now_secs() -> u64 {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Current date before unix epoch.")
+            .as_secs()
+    }
+}
```
rfk, line 291: Since we're not re-using `RateLimiter` here, this `PersistedRateLimiter` will need its own suite of tests.

rfk (changes requested):
Thanks for diving into this. I think the refactor here is really worthwhile, and in general we should have a much lower bar for thinking "hey, there should be a circuit-breaker/rate-limiter *here*".

The "request changes" here is specifically for test coverage of the `PersistedRateLimiter` class (and also CI seems to have a test failure of some sort) but I left some broader comments as well.

components/push/src/subscriber.rs
eoger: That was my idea too, but you cannot do generics over numeric types in Rust. I can always cut the `u64` to `u16` though, I'll do that :)

components/push/src/subscriber.rs
eoger: Yup, I realized too late that we needed a persisted counter versus an in-memory one :) Still worth refactoring though!

components/push/src/communications/rate_limiter.rs
```diff
@@ -0,0 +1,163 @@
...
+            format!("ratelimit_{}_timestamp", &self.op_name),
+            format!("ratelimit_{}_count", &self.op_name),
+        )
+    }
+
+    fn get_counters(&self, store: &Store) -> (u64, u16) {
+        let (timestamp_key, count_key) = self.db_meta_keys();
+        (
+            self.get_meta_integer(store, &timestamp_key),
+            self.get_meta_integer(store, &count_key) as u16, // We know count_key is stored as a `u16`, so there's no harm doing a cast here.
```
rfk, line 75: FWIW I think the fully-generic version would have worked if you made it generic over anything implementing `FromStr` rather than over integers, e.g.:

* https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a5512d3f9fff7331e54db7c2f5897845

But I'm fine the cast here in practice.

rfk (approved):
LGTM with some trivial nits 👍

components/push/src/communications/rate_limiter.rs
eoger: Very nice!

Issue #3671, opened 2020-10-28
Dot release with #3635


gabrielluong:
Fenix needs a dot release to expose the frecency threshold in AC. https://github.com/mozilla/application-services/issues/3635

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1862)
┆Sprint: Backlog

PR #3674, opened 2020-10-28
Cut release v66.0.0

eoger:
Fixes https://github.com/mozilla/application-services/issues/3671


rfk (approved):
Not a "dot release" as called for in #3671 but I don't think that really matters :-)

PR #3677, opened 2020-10-29
Gets closer to uniffi working.

mhammond:
This is much closer, but it leaves us with 3 main problems:

* Types we use but want to hide from the ffi. Eg, our rust API uses `SyncGuid`
  and `Timestamp`, but our FFI should use `String` and `i64`. It might be
  that we can abuse traits to hide this?

* Fields in our struct that shouldn't go across the FFI - `syncChangeCounter`
  is the obvious one. This field is needed for interacting with the database,
  but probably should not be part of the API - eg, even for a rust consumer,
  we don't want to expose this field.

  I think the general answer here is that there should be 2 (or more) distinct
  types - "internal implementation" types used for reading and writing the DB,
  and a different type used for the API. This would be far more obvious if our
  database was more complex (eg, had many relations) - you might have 1 struct
  per *table* at the implementation level, but a combined struct at the API
  level. Places does exactly this - it actually goes much further - different
  structs for *reads* vs *writes* (basically better handling optional fields
  which don't need to be specified when writing, but always exist when reading)
  It's not as obvious for autofill because the 2 structs are so similar, but I
  think this is probably what we should do.

* Errors are a but wierd for reasons I need to better understand. app-services
  has this error support stuff - eg, components/support/error/src/lib.rs.
  All our existing components use this, but uniffi doesn't want us to. nimbus
  seems to avoid this complexity. I'm really not sure what to do here.

And we should also take a step back and discuss the names we are exposing
to clients here. Specifically, the main "interface" exposed to consumers is
`AutofillDb` - which doesn't seem quite right. Thom, Lina and myself had
a big discussion about this, which I can't find, but I think we landed on
"Store" as the name. IOW, I think it might make sense to:

* Add a new store struct, that takes a path in ::new()
* It creates a DB object and stores it in a private val.
* The "api like" functions on the DB move to here.
* We change the imports in lib and name in the udl to refer
  to the new struct instead of DB. This effectively makes db
  private to the crate.

### DIFF
--- a/components/autofill/src/api/addresses.rs
+++ b/components/autofill/src/api/addresses.rs
@@ -61,7 +61,7 @@ pub struct Address {
 
     #[serde(default)]
     #[serde(rename = "timeLastUsed")]
-    pub time_last_used: Timestamp,
+    pub time_last_used: Option<Timestamp>,
 
     #[serde(default)]
     #[serde(rename = "timeLastModified")]
--- a/components/autofill/src/api/addresses.rs
+++ b/components/autofill/src/api/addresses.rs
@@ -106,8 +106,8 @@ impl Address {
 }
 
 #[allow(dead_code)]
-pub fn add_address(conn: &mut Connection, new_address: NewAddressFields) -> Result<Address> {
-    let tx = conn.transaction()?;
+pub fn add_address(conn: &Connection, new_address: NewAddressFields) -> Result<Address> {
+    let tx = conn.unchecked_transaction()?;
 
     let address = Address {
         guid: Guid::random(),
@@ -172,8 +172,8 @@ pub fn add_address(conn: &mut Connection, new_address: NewAddressFields) -> Resu
 }
 
 #[allow(dead_code)]
-pub fn get_address(conn: &mut Connection, guid: &Guid) -> Result<Address> {
-    let tx = conn.transaction()?;
+pub fn get_address(conn: &Connection, guid: &Guid) -> Result<Address> {
+    let tx = conn.unchecked_transaction()?;
     let sql = format!(
         "SELECT
             {common_cols}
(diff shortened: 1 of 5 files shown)


rfk:
> I think we landed on "Store"

I also vaguely recall this outcome FWIW 👍

mhammond:
>     * I think the general answer here is that there should be 2 (or more) distinct
>       types

Just to follow up a little more on this. Consider the cred-card struct. When a-c wants to *create* or *update* a credit-card, it shouldn't need to specify `time_created` (and indeed, it would probably be wrong to even allow them to specify it)`. One option to solve that would be to have `time_created` be `Option<Timestamp>` - but that would be somewhat wrong when a credit-card is being fetched - we don't want to force consumers to deal with the fact a `time_created` may be null everywhere when it never will be. So it may be that we do want 3 structs for credit-cards: one public one used when a credit-card is created, one public when it is returned, and a private one used when we interact with the db.

(We are also going to need a function that the app can call to tell us an address or credit-card was used - `times_used` is roughly in the same category as `time_created` above)

lougeniaC64:
> > ```
> > * I think the general answer here is that there should be 2 (or more) distinct
> >   types
> > ```
> 
> Just to follow up a little more on this. Consider the credit-card struct. When a-c wants to _create_ or _update_ a credit-card, it shouldn't need to specify `time_created` (and indeed, it would probably be wrong to even allow them to specify it). One option to solve that would be to have `time_created` be `Option<Timestamp>` - but that would be somewhat wrong when a credit-card is being fetched - we don't want to force consumers to deal with the fact a `time_created` may be null everywhere when it never will be. So it may be that we do want 3 structs for credit-cards: one public one used when a credit-card is created, one public when it is returned, and a private one used when we interact with the db.
> 
> (We are also going to need a function that the app can call to tell us an address or credit-card was used - `times_used` is roughly in the same category as `time_created` above)

Thanks for this feedback, Mark. All of this makes sense however I do have a question about the function to tell if a credit card or address was used, but it'll probably be easier to discuss it on Monday. For now I'll merge this PR into my branch and push the updates to address your feedback.

PR #3694, opened 2020-11-04
Don't discard final path component from self-hosted tokenserver URLs.

### LINKED ISSUE #3693: Final path component is stripped from self-hosted tokenserver URLs
Ref https://github.com/mozilla-mobile/fenix/issues/15590 for user reports of this in Fenix.

If a user specifies a tokenserver override URL with a path component like `https://example.com/token/1.0/sync/1.5`, we incorrectly end up trying to talk to `https://example.com/1.0/sync/1.5`. This appears to be a bad interaction between the `fxa-client` and `sync15` components:

1. The `fxa-client` component will [strip the `/1.0/sync/1.5` suffix](https://github.com/mozilla/application-services/blob/3567a6d920165501897ba31a6346d1177fcdd3d3/components/fxa-client/src/config.rs#L80), leaving the tokenserver URL as `https://example.com/token`
2. The `sync15` component tries to [add the `/1.0/sync/1.5` suffix back](https://github.com/mozilla/application-services/blob/3567a6d920165501897ba31a6346d1177fcdd3d3/components/sync15/src/token.rs#L66), but there's a comment warning that:

    ```
    // Note that base_url must end in a slash, or the last path element in
    // the base url will be replaced with this suffix.
    ```

    Since the fxa-client stripped the trailing slash, this results in incorrect result of `https://example.com/1.0/sync/1.5`.

I think we should fix the `sync15` component to avoid this last-path-element-replacement behaviour, it seems like a footgun behaviour that would never be the right thing for our usecases.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1877)
┆Sprint: Backlog

rfk:
Fixes #3693.

- ~~[ ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)~~

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,10 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v66.0.0...main)
+
+## FxA Client
+
+### What's Fixed
+
+- We no longer discard the final path component from self-hosted sync tokenserver URLs.
+  ([#3694](https://github.com/mozilla/application-services/pull/3694))
--- a/components/sync15/src/token.rs
+++ b/components/sync15/src/token.rs
@@ -64,11 +64,9 @@ struct TokenServerFetcher {
 }
 
 fn fixup_server_url(mut url: Url) -> Result<Url> {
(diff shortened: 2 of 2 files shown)


mhammond (approved):
Thanks for tracking this down!

PR #3698, opened 2020-11-05
Add Autofill FFI Layer [ci full]

### LINKED ISSUE #3663: Create FFI Layer for Forms Component
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1857)
┆Story Points: 8
┆Sprint: SYNC - end 2020-11-20

lougeniaC64:
Fixes #3663


lougeniaC64:
@mhammond  is it worthwhile for there to be a `module-build-autofill` step in the full CI build at this point?

mhammond:
I think this is ready to land. It doesn't include the android bindings, but that's OK - I've got a WIP for that, but I don't see a reason to wait for that to be ready before landing this.

Ryan, do you mind having a quick look? I'm happy with it, but some of the commits are mine, so it would be nice to get new eyes on them.

components/autofill/src/db/credit_cards.rs
```diff
@@ -95,7 +95,7 @@ pub fn get_all_credit_cards(conn: &Connection) -> Result<Vec<InternalCreditCard>
         let mut stmt = tx.prepare(&sql)?;
         credit_cards = stmt
             .query_map(NO_PARAMS, InternalCreditCard::from_row)?
-            .collect::<Result<Vec<InternalCreditCard>, _>>()?;
+            .collect::<std::result::Result<Vec<InternalCreditCard>, _>>()?;
```
rfk, line 98: You might be able to make this work using `.map(Into::into).collect` rather than using `std::result::Result`, but I think the `std::result::Result` is also fine. Thanks for the explanation over in the other PR.

components/autofill/src/error.rs
```diff
@@ -26,11 +26,4 @@ pub enum ErrorKind {
...
 
-error_support::define_error! {
-    ErrorKind {
-        (SqlError, rusqlite::Error),
-        (IoError, std::io::Error),
-        (InterruptedError, Interrupted),
-        (Utf8Error, std::str::Utf8Error)
-    }
-}
+pub type Result<T> = std::result::Result<T, ErrorKind>;
```
rfk, line 29: Naming nit: now that we don't have the layer of indirection provided by `define_error!`, I wonder if it would be simpler to call this `Error` instead of `ErrorKind`. I don't think it matters too much either way, but it might simplify the API for consumers if they don't have to wonder what an "error kind" is.

rfk (approved):
LGTM! 👍

components/autofill/src/db/credit_cards.rs
mhammond: I can't make that work for reasons I kindof expected given my limited understanding of some of this.

Note that it's a vec of results - so we have the same problem - we need an `Into()` for `Result<InternalCreditCard, rusqlite::Error>` whereas IIUC, we only have it for the error.

```
98 |             .map(Into::into).collect::<Result<Vec<InternalCreditCard>>>()?;
   |                  ^^^^^^^^^^ the trait `std::convert::From<std::result::Result<db::models::credit_card::InternalCreditCard, rusqlite::Error>>` is not implemented for `std::result::Result<db::models::credit_card::InternalCreditCard, error::Error>`
```

(TBH though, I don't really understand *why* it can't do that - but IIUC, that's the same reason I mentioned in the other issue about why you need to do `fn()?; Ok(())` rather than just `fn()`

components/autofill/src/error.rs
mhammond: good idea 👍

PR #3700, opened 2020-11-06
Pin stable rust version to 1.43

mhammond:
I think this updates circle in a reasonable way. Still have to do taskcluster which seems trickier, but might as well see if I'm on the right track.

### DIFF
--- a/automation/all_clippy_checks.sh
+++ b/automation/all_clippy_checks.sh
@@ -17,7 +17,10 @@ fi
 
 EXTRA_ARGS=( "$@" )
 
-cargo clippy --all --all-targets --all-features -- -D warnings ${EXTRA_ARGS[@]:+"${EXTRA_ARGS[@]}"}
+# Later rust versions downgraded some warning to pedantic, so we allow them here.
+ALLOWS=("-Aclippy::unreadable-literal"  "-Aclippy::trivially-copy-pass-by-ref" "-Aclippy::match-bool")
+
+cargo clippy --all --all-targets --all-features -- -D warnings  "${ALLOWS[@]}" ${EXTRA_ARGS[@]:+"${EXTRA_ARGS[@]}"}
 
 # Apparently --no-default-features doesn't work in the root, even with -p to select a specific package.
 # Instead we pull the list of individual package manifest files which have default features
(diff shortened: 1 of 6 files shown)


mhammond:
I think this is everything - not sure how to test taskcluster without making a dummy release? @rfk is that the doc you had in mind?

automation/all_clippy_checks.sh
```diff
@@ -17,7 +17,10 @@ fi
 
 EXTRA_ARGS=( "$@" )
 
-cargo clippy --all --all-targets --all-features -- -D warnings ${EXTRA_ARGS[@]:+"${EXTRA_ARGS[@]}"}
+# Later rust versions downgraded some warning to pedantic, so we allow them here.
+ALLOWS=("-Aclippy::unreadable-literal"  "-Aclippy::trivially-copy-pass-by-ref" "-Aclippy::match-bool")
```
rfk, line 21: Should this list be adjusted over time as we update the version of rust, and is there any way we can leave reminders to ourselves to do this?

rfk (approved):
Looks good to me, thanks! (That wasn't the doc I had in mind, but I think it's a better place than the one I had in mind, so 👍; I was thinking of the "dependency management" one)

automation/all_clippy_checks.sh
mhammond: Good question but I can't think of a way. I guess that ideally they'd be next to the single place where we specify the rust version, but I can't see a way to avoid duplicating that either.

However, the "good" thing here is that this is listing things which later version of rust are *less* strict about - so there's no harm in keeping them there. This is better than if they were things which were made more strict in later versions.

eoger (approved):
LGTM, it should also make backports way easier. Only downside is having to Ctrl+F across the repo to bump the rust version.
Can we also modify the [verify script](https://github.com/mozilla/application-services/blob/5bbda246ec2ef14c3e4ee6dd37541b038d63d6f6/libs/verify-common.sh#L10)?

mhammond:
> 
> 
> LGTM, it should also make backports way easier. Only downside is having to Ctrl+F across the repo to bump the rust version.
> Can we also modify the [verify script](https://github.com/mozilla/application-services/blob/5bbda246ec2ef14c3e4ee6dd37541b038d63d6f6/libs/verify-common.sh#L10)?

Oops - *something* should change there but I'm not sure what. Ryan and I thought there's no need to force developers on to 1.43 - they might have good reason to want "stable" and CI will catch them if they screw up - which would imply *removing* that line. OTOH, maybe it's OK to be opinionated?

WDYT?

Issue #3699, opened 2020-11-05
Cut an official release of appservices with nimbus


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1884)
┆Story Points: 3
┆Epic: Fenix Client SDK
┆Sprint: SYNC - end 2020-11-20

rfk:
The bulk of this work is done in https://github.com/mozilla/application-services/pull/3701.

@travis79 are there pending nimbus PRs that we should merge before cutting a release here? I'm thinking specifically of [global opt-out](https://github.com/mozilla/nimbus-sdk/pull/51) and [schema version checking](https://github.com/mozilla/nimbus-sdk/pull/49), but I think we could safely publish those in a follow-up release if we want to get something out the door to unblock the [main a-c PR]9https://github.com/mozilla-mobile/android-components/pull/8261). WDYT?

travis79:
While it would be nice to have the global opt-out and the versioning, I'm fine with landing those after a release to unblock the android component landing.

jhugman:
@travis79 — I'd love to be able to land global opt out before the android-component lands.

rfk:
Version `v67.0.0` has been tagged and will include the new nimbus component. Consumers should take a dependency on `org.mozilla.appservices:nimbus`.

rfk:
>  I'd love to be able to land global opt out before the android-component lands.

Noting for completeness, it does include the global opt-out functionality.

PR #3701, opened 2020-11-06
Import nimbus-sdk as submodule, republish as part of the megazord [ci full].

rfk:
This commit pulls in https://github.com/mozilla/nimbus-sdk as a
git submodule, and adds some build integration to publish it as
part of the application-services megazord. Instead of an all-in-one
nimbus package, we produce two separate packages:

* `org.mozilla.appservices:nimbus` with the Nimbus Kotlin code
* `org.mozilla.appservices:full-megazord` udpated to include the
  Nimbus Rust code.

(This is just the nimbus-related parts of https://github.com/mozilla/application-services/pull/3669, extracted into their own PR; it's draft until at least https://github.com/mozilla/nimbus-sdk/pull/54 is resolved. Connects to #3699)

TODOs:
* [x] fix android build failures in taskcluster (proably git-submodule related)
* [x] fix iOS build failures in CircleCI (seems to be some glean-related bustage)
* [x] compare compiled artifact size to previous release
* [x] test end-to-end in a local build of samples-glean from [travis's a-c branch](https://github.com/mozilla-mobile/android-components/pull/8261)
* [x] add changelog entry


Cargo.toml
```diff
@@ -110,3 +117,8 @@ default-members = [
 opt-level = "s"
 debug = true
 lto = "thin"
+
+# We import some git submodules which in turn may depend on appservices crates.
+# We always want to use the local version of our own crates in that case.
+[patch."https://github.com/mozilla/application-services"]
+viaduct = { path = "components/viaduct" }
```
rfk, line 124: This is weird but important - without it we end up building two versions of viaduct, one for the local checkout and another for the git ref pinned by nimbus. I'm actually surprised that seemed to work at all, but it seems better to avoid it entirely rather than depending on it working correctly in future.

components/nimbus/README.md
```diff
@@ -0,0 +1,49 @@
...
+directly at the Rust level.
+
+Consumers will also need to add the following snippet to ensure that Nimbus can find the
+correct shared library:
+
+```
+    System.setProperty(
+        "uniffi.component.nimbus.libraryOverride",
+        System.getProperty("mozilla.appservices.megazord.library", "megazord")
+    )
```
rfk, line 24: Ref https://github.com/mozilla/uniffi-rs/issues/343 to figure out how to do this better.

rfk:
@mhammond, pinging you for review since it subsumes your earlier work on nimbusing the megazord. I think you've seen most of these details before in the glean-megazord PR, but some are new.

There's no rush, I won't get to the "try this out in a local app build" part until Monday, but I wanted to put it on your radar.

rfk:
>  compare compiled artifact size to previous release

TL;DR: approx a 1M increase in the size of `libmegazord.so`, which will flow through to the size of consuming apps.

---

Here's the binary sizes for the `v66.0.0` release that's currently on maven.mozilla.org:

```
[rfk] ~/Desktop/tmp % du -hs classes.jar jni/*/libmegazord.so
4.0K    classes.jar
6.5M    jni/arm64-v8a/libmegazord.so
5.1M    jni/armeabi-v7a/libmegazord.so
7.4M    jni/x86/libmegazord.so
7.5M    jni/x86_64/libmegazord.so
```

By comparison, here are the binary sizes I get for a `publishToMavenLocal` from this PR:

```
[rfk] appservices/full-megazord/tmp % du -hs classes.jar jni/*/libmegazord.so
4.0K    classes.jar
7.8M    jni/arm64-v8a/libmegazord.so
5.9M    jni/armeabi-v7a/libmegazord.so
8.8M    jni/x86/libmegazord.so
8.9M    jni/x86_64/libmegazord.so
```

That's a non-trivial size increase - over 1M on 64-bit platforms, a bit less on 32-bit.

---

We were hoping to get much of this size increase back when we complete the megazording of Glean. Here are the libraries for both Glean and the current appservices together:

```
[rfk] ~/Desktop/tmp % du -hs jni/*/*.so 
1.4M    jni/arm64-v8a/libglean_ffi.so
6.5M    jni/arm64-v8a/libmegazord.so        // 7.9M total
968K    jni/armeabi-v7a/libglean_ffi.so
5.1M    jni/armeabi-v7a/libmegazord.so      // 6.1M total
1.8M    jni/x86/libglean_ffi.so
7.4M    jni/x86/libmegazord.so              // 9.2M total
1.7M    jni/x86_64/libglean_ffi.so
7.5M    jni/x86_64/libmegazord.so           // 9.2M total
```

Here's what the binary sizes look like from a `publishToMavenLocal` of #3669 with Glean included, meaning that this includes the functionality of both `libmegazord.so` with nimbus, and `libglean_ffi.so`:

```
[rfk] appservices/full-megazord/tmpg % du -hs classes.jar jni/*/*.so                                       
4.0K    classes.jar
8.5M    jni/arm64-v8a/libmegazord.so
6.3M    jni/armeabi-v7a/libmegazord.so
9.7M    jni/x86/libmegazord.so
9.8M    jni/x86_64/libmegazord.so
```

The size of the combined megazord is still bigger than the current size of appservices+glean, but not by very much. To be honest I was hoping it would produce an overall size *saving*, so I'm going to play around with this a bit more and see what I can find.

Cargo.toml
rfk: Oh, interesting, I was messing around with LTO settings and without this fix, switching to `lto="fat"` caused the build to fail due to duplicate viaduct symbols! This reinforces my belief that it's important to ensure a single version using this gross `[patch]` behaviour.

rfk:
> Here's what the binary sizes look like from a publishToMavenLocal of #3669 with Glean included,
> meaning that this includes the functionality of both libmegazord.so with nimbus, and libglean_ffi.so:

I was able to substantially improve this by using `lto="fat'` and `opt-level="z"`:

```
[rfk] appservices/full-megazord/tmpg % du -hs jni/*/*.so                                 
7.4M    jni/arm64-v8a/libmegazord.so
5.6M    jni/armeabi-v7a/libmegazord.so
8.0M    jni/x86/libmegazord.so
8.5M    jni/x86_64/libmegazord.so
```

These sizes are comfortably smaller the sum of the two separate appservices+glean shared libraries in current release. We may not want to enable those options for other reasons, but at least it gives me some confidence that we can get back the size increase here by further investment in the megazord.

rfk:
Quick note to self for future reference, most of the size reduction in the glean megazord above seems to be due to `lto="fat"`.

mhammond (approved):
This is great 👍

rfk:
I tried a `publishToMavenLocal` of this branch against both `samples-sync`, and `samples-glean` with [@travis79's nimbus patch](https://github.com/mozilla-mobile/android-components/pull/8261), and both seemed to work as expected.

Ref https://github.com/travis79/android-components/pull/5 for the small set of compatibility tweaks I had to make on the a-c side.

rfk:
With that manual test, and with @mhammond's r+, I think this is ready to go! I'll merge once CI gives me the final ticks of approval.

PR #3704, opened 2020-11-09
Import nimbus-sdk as submodule, republish as part of the megazord [ci full].

rfk:
This commit pulls in https://github.com/mozilla/nimbus-sdk as a
git submodule, and adds some build integration to publish it as
part of the application-services megazord. Instead of an all-in-one
nimbus package, we produce two separate packages:

* `org.mozilla.appservices:nimbus` with the Nimbus Kotlin code
* `org.mozilla.appservices:full-megazord` updated to include the
  Nimbus Rust code.

(Trying again after this was backed out in https://github.com/mozilla/application-services/pull/3703)


rfk:
@mhammond r? for completeness? The only change here since the previous version, is to update `uniffi` and `nimbus-sdk` to versions that work with our pinned Rust version.

mhammond (approved):

mhammond:
(Although it looks like glean is causing detekt failures :( )

rfk:
Yeah, I *think* I figured out the right gradle blah-blah to exclude that directory from detekt, trying now...

PR #3711, opened 2020-11-10
Avoid using eprintln in tests to avoid always seeing them in circle

mhammond:
I was looking at the test outputs and noticed some strange noise...


rfk (approved):
I'm a little weirded out by landing the equivalent of a `console.log` in tests at all, but agree that `println` seems better than `eprintln`, so 👍

mhammond:
Yeah, I agree, but at least the `println`s are hidden unless the test fails!

PR #3713, opened 2020-11-11
Early WIP getting Kotlin bindings and adding it to the megazord.

mhammond:
A WIP - see https://gist.github.com/mhammond/4f9dbe0ee5a35eb31a489219b07d7d03 for how I was trying to see if I can make it with with a-c.  cc @lougeniaC64

### DIFF
--- a/components/autofill/android/build.gradle
+++ b/components/autofill/android/build.gradle
@@ -1,0 +1,142 @@
+// Copied from 'places'
+// With the following modifications:
+// * Removed glean integration
+// * Removed protobuf integration
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.26"
+}
+
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'com.google.protobuf' // obscure error is this is commented out!
+
+android {
+    ndkVersion rootProject.ext.build.ndkVersion
+    compileSdkVersion rootProject.ext.build.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion rootProject.ext.build['minSdkVersion']
+        targetSdkVersion rootProject.ext.build['targetSdkVersion']
+
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        buildConfigField("String", "LIBRARY_VERSION", "\"${rootProject.ext.library.version}\"")
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            consumerProguardFiles "$rootDir/proguard-rules-consumer-jna.pro"
+        }
+    }
+
+    sourceSets {
+        test.resources.srcDirs += "$buildDir/rustJniLibs/desktop"
+        test.resources.srcDirs += "${project(':full-megazord').buildDir}/rustJniLibs/desktop"
+
+        main {
+            proto {
+                srcDir '../src'
+            }
+        }
+    }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+    }
+
+    // This is required to support new AndroidX support libraries.
+    // See mozilla-mobile/android-components#842
+    compileOptions {
+        sourceCompatibility rootProject.ext.build.jvmTargetCompatibility
+        targetCompatibility rootProject.ext.build.jvmTargetCompatibility
+    }
+
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+        kotlinOptions {
+            jvmTarget = rootProject.ext.build.jvmTargetCompatibility
+        }
+    }
+}
+
+configurations {
+    // There's an interaction between Gradle's resolution of dependencies with different types
+    // (@jar, @aar) for `implementation` and `testImplementation` and with Android Studio's built-in
+    // JUnit test runner.  The runtime classpath in the built-in JUnit test runner gets the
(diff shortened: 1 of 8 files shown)


components/autofill/android/build.gradle
```diff
@@ -0,0 +1,142 @@
+// Copied from 'places'
+// With the following modifications:
+// * Removed glean integration
+// * Removed protobuf integration
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.26"
```
rfk, line 6: I think this python stuff is only used for glean and can be removed.

components/autofill/android/build.gradle
```diff
@@ -0,0 +1,142 @@
...
+            consumerProguardFiles "$rootDir/proguard-rules-consumer-jna.pro"
+        }
+    }
+
+    sourceSets {
+        test.resources.srcDirs += "$buildDir/rustJniLibs/desktop"
+        test.resources.srcDirs += "${project(':full-megazord').buildDir}/rustJniLibs/desktop"
+
+        main {
+            proto {
```
rfk, line 39: This is for protobuf, and might be the cause of the "obscure error" mentioned above about removing the protobuf plugin (or it might not! gradle is wild)

components/autofill/src/db/credit_cards.rs
```diff
@@ -95,7 +95,7 @@ pub fn get_all_credit_cards(conn: &Connection) -> Result<Vec<InternalCreditCard>
         let mut stmt = tx.prepare(&sql)?;
         credit_cards = stmt
             .query_map(NO_PARAMS, InternalCreditCard::from_row)?
-            .collect::<Result<Vec<InternalCreditCard>, _>>()?;
+            .collect::<std::result::Result<Vec<InternalCreditCard>, _>>()?;
```
rfk, line 98: I'm lightly surprised that this needed to use an explicit `std::result::Result`, what was the problem with the previous version?

docs/building.md
```diff
@@ -22,6 +22,11 @@ Ensure this section works for you before jumping to the Android/iOS specific sec
   - zlib
     - Using apt: `apt install zlib1g-dev`
     - Already installed on macOS by XCode.
+  - uniffi-bindgen
+    - `cargo install uniffi_bindgen`. Note that this is being updated regularly
+      and cargo isn't very clever - if you see complaints about the wrong
+      version being installed, you probably need to
+      `cargo uninstall uniffi_bindgen` and re-execute the installation command.
```
rfk, line 29: Thanks for adding this!

IIUC `cargo install -f uniffi_bindgen` should also work for upgrading to the latest version. Also this is mostly about `uniffi-bindgen` being not very clever, not cargo ;-)

examples/autofill-utils/src/autofill-utils.rs
```diff
@@ -150,7 +150,7 @@ fn run_update_address(store: &Store, filename: String) -> Result<()> {
     let guid = address_fields.guid.clone();
 
     println!("Making `update_address` api call for guid {}", guid);
-    Store::update_address(store, &address_fields)?;
+    Store::update_address(store, address_fields)?;
```
rfk, line 153: If you like, you should be able to make these take references by annotating the argument with `[ByRef]` in the UDL, ref [these docs](https://mozilla.github.io/uniffi-rs/udl/builtin_types.html#built-in-types) and [this example](https://github.com/mozilla/uniffi-rs/blob/4812fc2494d6db6e1800c0da8f7eba69b235451c/examples/sprites/src/sprites.udl#L3).

components/autofill/src/db/credit_cards.rs
lougeniaC64: Full disclosure, this was a change I made that originated in #3698. I was getting the following build error: ```a value of type `std::result::Result<std::vec::Vec<db::models::credit_card::InternalCreditCard>, error::ErrorKind>` cannot be built from an iterator over elements of type `std::result::Result<db::models::credit_card::InternalCreditCard, rusqlite::Error>` ``` I then prefixed `Result` with `std::result`, made a note to research this, and naturally forgot about it. If there's a more appropriate way to handle this I'm all ears. I'm speaking from the edge of my Rust knowledge at the moment.

rfk: Thanks. Maybe something not quite working right in the mapping from `rusqlite::Error` to `ErrorKind`. It seems like something we can safely follow up on later.

mhammond: It makes some sense to me - the specific error is `std::result::Result<Vec<InternalCreditCard>, rusqlite::Error>>`, not `std::result::Result<Vec<InternalCreditCard>, ErrorType>>` - there's no automatic conversion there.

Isn't it the same as if we have, say:

```
fn bar() -> std::result::Result<(), rusqlite::Error> {
 Ok(())
}

fn foo() -> Result<()> {
  bar() // doesn't work - wrong E type but not coerced
// but the below does.
  bar()?; // wrong E type but it's coerced here.
  Ok(())
}
```
?

(I really need to study rust's type system)

examples/autofill-utils/src/autofill-utils.rs
mhammond: Thanks! I did miss that in the docs, but I think this is OK for now - the `update` and `insert` functions are both consistent now, which really was the main goal. I don't think it makes any difference to the kotlin code, right?

rfk: Right, shouldn't make any difference to consumers.

components/autofill/src/db/credit_cards.rs
rfk: Yeah; I guess I was expecting some implicit conversion via `Into` to happen here, but really I don't have a good reason to expect that.

PR #3714, opened 2020-11-12
Generate Kotlin bindings and add them to the megazord.

mhammond:
Github forced me to make a new PR, but I made all the suggested changes and I've got a test working in a-c.


rfk (approved):
👍

PR #3735, opened 2020-12-01
Add address foreign key to credit card model

lougeniaC64:
Adding a foreign key from the addresses table in the credit cards tables. This was omitted because it wasn't clear if credit cards and addresses would be handled as two separate projects. But the relationship between these tables can be created regardless as that product decision will mainly affect the a-c and/or fenix implementation.

### DIFF
--- a/components/autofill/sql/create_shared_schema.sql
+++ b/components/autofill/sql/create_shared_schema.sql
@@ -3,7 +3,7 @@
 -- file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 CREATE TABLE IF NOT EXISTS addresses_data (
-    guid          TEXT NOT NULL PRIMARY KEY,
+    guid                TEXT NOT NULL PRIMARY KEY,
     given_name          TEXT NOT NULL,
     additional_name     TEXT NOT NULL,
     family_name         TEXT NOT NULL,
@@ -26,7 +26,7 @@ CREATE TABLE IF NOT EXISTS addresses_data (
 );
 
 CREATE TABLE IF NOT EXISTS addresses_mirror (
-   guid          TEXT NOT NULL PRIMARY KEY,
+    guid                TEXT NOT NULL PRIMARY KEY,
     given_name          TEXT NOT NULL,
     additional_name     TEXT NOT NULL,
     family_name         TEXT NOT NULL,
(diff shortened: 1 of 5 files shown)


components/autofill/sql/create_shared_schema.sql
```diff
@@ -55,36 +55,36 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
...
 CREATE TABLE IF NOT EXISTS credit_cards_mirror (
-    guid                TEXT NOT NULL PRIMARY KEY,
-    cc_name             TEXT NOT NULL, -- full name
-    cc_number           TEXT NOT NULL,
-    cc_exp_month        INTEGER,
-    cc_exp_year         INTEGER,
-    cc_type             TEXT NOT NULL,
-    -- cc_exp              TEXT NOT NULL, -- text format of the expiration date e.g. "[cc_exp_year]-[cc_exp_month]"
+    guid                    TEXT NOT NULL PRIMARY KEY,
+    billing_address_guid    TEXT NOT NULL, -- An optional GUID of an autofill address record which may or may not exist locally.
```
mhammond, line 77: I'm a bit confused by the intent here. `NOT NULL` implies it isn't optional. And the "may or may not exist" part also seems odd - does this mean this can hold a GUID that doesn't exist in the addresses table? If so, we should document better why that might happen - but if not (which I think we prefer), then a `FOREIGN KEY` constraint for this field probably makes sense.

components/autofill/src/autofill.udl
```diff
@@ -1,6 +1,7 @@
 namespace autofill {};
 
 dictionary NewCreditCardFields {
+    string billing_address_guid;
```
mhammond, line 4: Having the consumer specify the GUID will impact the API somewhat - that's not a problem, but worth thinking about, and maybe worth seeing how the desktop UI works.

What it means is that the address will need to be added separately, and in a different DB transaction. That's fine if UI asks for the address and card details in 2 separate forms, with 2 separate "save" buttons. However, if the UI shows just one form, it might be a problem and this API might be better defined as forcing the app to specify the address *fields*, and hide the GUID as an implementation details (ie, we would then add the address and CC in the same transaction).

We also need to define what happens when the CC is deleted - I assume the address remains.

(This is just food for thought rather than a specific suggestion - I guess the more meta point is that we really do need to start understanding what desktop does in a deeper way :)

mhammond (commented):
Just a drive-by...

components/autofill/src/autofill.udl
lougeniaC64: In the long run I absolutely do not think it makes sense for the consumer to provide an address guid in order to create a credit card. However as you specified doing anything else at this point would require that we have a better understanding of what desktop does or what the mobile team wants if that differs from desktop. Long story short I'm punting.

components/autofill/sql/create_shared_schema.sql
lougeniaC64: The main reason this is a WIP is because I'm honestly uncertain here. I pulled the comment from [desktop](https://searchfox.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#61) and mean to review the desktop logic to understand why the billing address guid may or may not exist.

lougeniaC64: After looking at desktop, the only reason I can find for `billing_address_guid` being null is a local address can't be found, as the comment states. While relevant for a JSON store, this shouldn't apply to our autofill component as we have a proper db that can have an FK constraint. I'm going to remove the comment and add the FK constraint.

components/autofill/sql/create_shared_schema.sql
```diff
@@ -55,36 +55,40 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
...
+    time_created            INTEGER NOT NULL,
+    time_last_used          INTEGER,
+    time_last_modified      INTEGER NOT NULL,
+    times_used              INTEGER NOT NULL DEFAULT 0,
 
     /* Same "sync change counter" strategy used by other components. */
-    sync_change_counter INTEGER NOT NULL DEFAULT 1
+    sync_change_counter INTEGER NOT NULL DEFAULT 1,
+
+    FOREIGN KEY(billing_address_guid) REFERENCES addresses_data(guid) ON DELETE SET DEFAULT
```
lougeniaC64, line 74: @mhammond  FYI, I chose `SET DEFAULT` as the `ON DELETE` action based on our conversation earlier about the one-to-many relationship between addresses and credit cards as well as [this test assertion](https://searchfox.org/mozilla-central/source/browser/extensions/formautofill/test/unit/test_creditCardRecords.js#352) in desktop. But it might make more sense to cascade.

components/autofill/sql/create_shared_schema.sql
```diff
@@ -55,36 +55,40 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
...
-    time_created        INTEGER NOT NULL,
-    time_last_used      INTEGER,
-    time_last_modified  INTEGER NOT NULL,
-    times_used          INTEGER NOT NULL DEFAULT 0
+    time_created            INTEGER NOT NULL,
+    time_last_used          INTEGER,
+    time_last_modified      INTEGER NOT NULL,
+    times_used              INTEGER NOT NULL DEFAULT 0,
+
+    FOREIGN KEY(billing_address_guid) REFERENCES addresses_data(guid) ON DELETE SET DEFAULT
```
lougeniaC64, line 91: @mhammond Also want to call out that I'm referencing the guid from the `addresses_data` table and I'm not sure if that should be happening in the `credit_cards_mirror` table. It might make more sense for the reference to be coming from the `addresses_mirror` table since the API implies a billing address should be created before a credit card. However an empty string could be passed in as the billing address guid. 🤔 Anyway, I'm leaving this comment here to raise my uncertainty.

mhammond: Yeah, the mirror table is a little trickier - given this is what's on the server and the risk that this somehow got corrupted (eg, by another old, dumb client) I think we should not enforce much on the mirror table - but this also implies we shouldn't trust it when it comes time to apply data from that mirror table back into the real table (ie, if the GUID from the mirror table doesn't exist, we will need to both ensure the local table has a valid value written and re-upload the record so the server gets fixed. That will be part of the sync work though, so I think it's fine to largely ignore here.

components/autofill/sql/create_shared_schema.sql
mhammond: How does this work given `billing_address_guid` is `NOT NULL`?

components/autofill/sql/create_shared_schema.sql
```diff
@@ -55,36 +55,38 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
...
 CREATE TABLE IF NOT EXISTS credit_cards_mirror (
-    guid                TEXT NOT NULL PRIMARY KEY,
-    cc_name             TEXT NOT NULL, -- full name
-    cc_number           TEXT NOT NULL,
-    cc_exp_month        INTEGER,
-    cc_exp_year         INTEGER,
-    cc_type             TEXT NOT NULL,
-    -- cc_exp              TEXT NOT NULL, -- text format of the expiration date e.g. "[cc_exp_year]-[cc_exp_month]"
+    guid                    TEXT NOT NULL PRIMARY KEY,
+    billing_address_guid    TEXT NOT NULL,
```
mhammond, line 79: The `NOT NULL` here is also likely to cause issues as I expect the server to have many records with this as NULL.

components/autofill/src/db/credit_cards.rs
```diff
@@ -625,9 +769,25 @@ mod tests {
...
+        )
+        .expect("saved address for touch test");
+
         let saved_credit_card = add_credit_card(
             &db,
             NewCreditCardFields {
+                billing_address_guid: saved_address.guid.to_string(),
                 cc_name: "john doe".to_string(),
                 cc_number: "5555666677778888".to_string(),
                 cc_exp_month: 5,
```
mhammond, line 793: I think we should add a new test for the `ON DELETE` behaviour.

mhammond (changes requested):
I think the `NOT NULL` constraint is going to cause some issues.

lougeniaC64:
@mhammond thanks for the feedback. I'm going to hold this work until after the sync functionality is in place. At that point I think I'll have a better understanding of how best to implement these changes.

PR #3739, opened 2020-12-06
Update dependencies list to preference `main` over `master`.

rfk:
Our periodic dependency-check task is failing in CI due to this exogenous change, this should fix it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2037,9 +2037,9 @@ dependencies = [
 
 [[package]]
 name = "ordered-float"
-version = "1.1.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3741934be594d77de1c8461ebcbbe866f585ea616a9753aa78f2bdc69f0e4579"
+checksum = "3305af35278dd29f46fcdd139e0b1fbfae2153f0e5928b39b035542dd31e37b7"
 dependencies = [
  "num-traits",
 ]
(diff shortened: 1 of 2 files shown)


rfk:
(Also picking up a fix for a brand new RUSTSEC advisory while I'm here)

megazords/full/android/dependency-licenses.xml
```diff
@@ -50,19 +50,19 @@ the details of which are reproduced below.
   </license>
   <license>
     <name>Apache License 2.0: askama</name>
-    <url>https://github.com/djc/askama/blob/main/LICENSE-APACHE</url>
+    <url>https://github.com/djc/askama/blob/master/LICENSE-APACHE</url>
```
mhammond, line 53: These changes look reversed? The 'main' branch exists in all these repos, but this patch is changing it to the old 'master'?

rfk: Oh, hrm, TBH I didn't look very hard I just re-ran the script. Let me see if I can figure out what's happening here...

rfk:
OK, what I think is happening here is that there is a (maybe newly added?) redirect from the `master` branch to the `main` branch on github. When our dependency script tries to resolve a license URL, it checks `master` *first* and then falls back to `main`. The presence of this redirect means that `master` succeeds and we choose that as the URL.

I'm going to change things to preference `main` over `master`, which should avoid this issue.

mhammond (approved):

PR #3747, opened 2020-12-07
Guide to upgrading NSS

eoger:


docs/howtos/upgrading-nss-guide.md
```diff
@@ -0,0 +1,20 @@
+# Guide to upgrading NSS
+
+Our components rely on cryptographic primitives provided by [NSS](https://developer.mozilla.org/docs/Mozilla/Projects/NSS).
+Every month or so, a new version of NSS is [published](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/NSS_Releases) and we should try to keep our version as up-to-date as possible.
+
+For various reasons, we compile NSS ourselves and link to it statically. Note that NSS is mainly used by Mozilla as a dynamic library and the NSS project is missing related CI jobs (iOS builds, windows cross-compile builds etc.) so you should expect breakage when updating the library (hence this guide).
```
rfk, line 6: Now that you mention it, I don't actually recall what these "various reasons" are. Is it worth noting them here explicitly?

docs/howtos/upgrading-nss-guide.md
```diff
@@ -0,0 +1,20 @@
...
+
+The build code is located in the [`libs/`](https://github.com/mozilla/application-services/tree/main/libs) folder.  
+The version string is located in the beggining of [`build-all.sh`](https://github.com/mozilla/application-services/blob/b0b3daa6580d04906fc53e9e479e8bebb464cf78/libs/build-all.sh#L8-L11): for most NSS upgrades, the only action needed is to bump the version number in this file and update the downloaded archive checksum.  Note that each platform has its own nss build script (e.g. [`build-nss-ios.sh`](https://github.com/mozilla/application-services/blob/b0b3daa6580d04906fc53e9e479e8bebb464cf78/libs/build-nss-ios.sh))
```
rfk, line 9: sp: "beginning".

rfk: nit: The colon at the end of the "build-all.sh" link text doesn't parse for me, should this be a plain full-stop?

rfk: nit: "nss" -> "NSS" for consistency in "has its own nss build script".

rfk: As a reader, it's not clear to me what I'm supposed to do with the knowledge that "each platform has its own nss build script". Do you mean something like "...which you might need to update on occasion if/when it breaks"?

docs/howtos/upgrading-nss-guide.md
```diff
@@ -0,0 +1,20 @@
...
+Because the NSS static build process does not output a single `.a` file (it would be great if it did), this file must [describe](https://github.com/mozilla/application-services/blob/b0b3daa6580d04906fc53e9e479e8bebb464cf78/components/support/rc_crypto/nss/nss_build_common/src/lib.rs#L85-L133) for each architecture which modules should we link against. It is mostly a duplication of logic from the [NSS gyp build files](https://searchfox.org/nss/rev/d0ca572a63597a19889611c065273f131cc09b7a/lib/freebl/freebl.gyp#385-408). Note that this logic is also duplicated in our NSS lib build steps (e.g. [build-nss-desktop.sh](https://github.com/mozilla/application-services/blob/b0b3daa6580d04906fc53e9e479e8bebb464cf78/libs/build-nss-desktop.sh#L82-L114)).
+
+One of the most common build failures we got from upgrading NSS comes from that:
+the NSS project decides to implement a vectorized/asm version of a crypto algorithm for a very specific architecture in order to improve performance. This new optimized code gets implemented as a new gyp target/module that is emitted only for the supported architectures.
```
rfk, line 17: I found this first sentence a little hard to read, let me see if I understood it correctly: The NSS team periodically add new vectorized/asm implementations of crypto algorithms in order to improve performance or support new architectures. When they do this, it breaks out build scripts because ... [the reasons you describe].

Is that accurate? If so, rewording slightly might make future readers more confident in their interpretation, something like "One of the most common build failures we when upgrading NSS comes from NSS adding new vectorized/asm versions of a crypto algorithm...".

rfk (approved):
Great, thank you! I left a couple of readibility nits but they're quite minor.

PR #3753, opened 2020-12-10
Update README for sync15 and sync_manager

mhammond:
Please be as nit-picky as possible and ask for further clarification wherever you think it makes sense!

### DIFF
--- a/components/sync15/README.md
+++ b/components/sync15/README.md
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
 It handles things like encrypting/decrypting records, obtaining and
 using storage node auth tokens, and so-on.
 
-It also needs a more complete README...
+There are 2 key concepts to understand here - the implementation itself, and
+a rust trait for a "syncable store" where component-specific logic lives - but
+before we dive into them, some preamble might help put things into context.
+
+## Introduction and History
+
+For many years Sync has worked exclusively against a "sync v1.5 server". This
+[is a REST API described here](https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html).
+The important part is that the API is conceptually quite simple - there are
+arbitrary "collections", indexed by a GUID, and lacking traditonal database
+concepts like joins. Because the payload is encrypted, there's very little
+scope for it to be much smarter. Thus it's reasonably easy to create a fairly
+generic abstraction over the API that can be easily reused.
+
+Back in the deep past, we found ourselves with 2 different components that
+needed to sync against a sync v1.5 server. The apps using these components
+didn't have schedulers or any UI for choosing what to sync - so these
+components just looked at the existing state of the engines on the server and
+synced if they were enabled.
+
+This was also pre-megazord - the idea was that apps could choose from a "menu"
+of components to include - so we didn't really want to bind these components
+together. Therefore, there was no concept of "sync all" - instead, each of the
+components had to be synced individually. So this component stated out as more
+of a "library" than a "component" which individual components could reuse - and
+each of these components was a "syncable store".
+
+Fast forward to Fenix and we needed a UI for managing all the engines supported
+there, and a single "sync now" experience etc - so we also have a sync_manager
+component - [see its README for more](../components/sync_manager/README.md).
+But even though it exists, there are still some parts of this component that
+reflect these early days - for example, it's still possible to sync just a
+single component using sync15 (ie, without going via the "sync manager"),
+although this isn't used and should be removed - the "sync manager" allows you
+to choose which engines to sync, so that should be used exclusively.
+
+## Metadata
+
+There's some metadata associated with a sync. Some of the metadata is "global"
+to the app (eg, the enabled state of engines, information about what servers to
+use, etc) and some is specific to an engine/store (eg, timestamp of the
+server's collection for this engine/store, guids for the collections, etc).
+
+We made the decision early on that no storage should be done by this
+component:
+
+* The "global" metadata should be stored by the application - but because it
+  doesn't need to interpret the data, we do this with an opaque string (that
+  is JSON, but the app should never assume or introspect that)
+
+* Each store/engine should store its own metadata, so we don't end up in the
+  situation where, say, a database is moved between profiles causing the
+  metadata to refer to a completely different data set. So each store/engine
+  stores its metadata in the same database as the data itself, so if the
+  database is moved or copied, the metadata comes with it)
+
+## Sync Implementation
+
+The core implementation does all of the interaction with things like the
+token-server, the `meta/global` and `info/collections` collections, etc. It
+does all network interaction (ie, individial stores don't need to interact with
+the network at all), tracks things like whether the server is asking us to
+"backoff" due to operational concerns, manages encryption keys and the
+encryption itself, etc. The general flow of a sync - which interacts with the
+`Store` trait - is:
+
+* Does all pre-sync setup, such as checking `meta/global`, and whether the
+  sync IDs on the server match the sync IDs we last say (ie, to check whether
+  something drastic has happened since we last synced)
+* Asks the store about how to formulate the URL query params to obtain the
+  records the store cares about. In most cases, this will simply be "records
+  since the last modified timestamp of the last sync".
+* Downloads and decrypts these records.
+* Passes these records to the store for processing, and obtains records that
+  should be uploaded to the server.
+* Encrypts these outgoing records and uploads them.
+* Tells the store about the result of the upload (ie, the last-modified
+  timestamp of the POST so it can be saved as store metadata)
+
+As above, the sync15 component really only deals with a single store at a time.
+See the "sync manager" for how multiple stores are managed (but the tl;dr is
+that the "sync manager" leans on this very heavily, but knows about multiple
+stores and manages shared state)
+
+## The `Store` trait
+
+The store trait is where all logic specific to a collection lives. For
+<handwave> reasons, it actually lives in the
+[sync-traits helper](https://github.com/mozilla/application-services/blob/main/components/support/sync15-traits/src/store.rs)
+but for the purposes if this document, you should consider it as owned by sync15.
+
+This is actually quite a simple trait - at a high level, it's really just
+concerned with:
+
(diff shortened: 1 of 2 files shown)


components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+Back in the deep past, we found ourselves with 2 different components that
+needed to sync against a sync v1.5 server. The apps using these components
+didn't have schedulers or any UI for choosing what to sync - so these
+components just looked at the existing state of the engines on the server and
+synced if they were enabled.
+
+This was also pre-megazord - the idea was that apps could choose from a "menu"
+of components to include - so we didn't really want to bind these components
+together. Therefore, there was no concept of "sync all" - instead, each of the
+components had to be synced individually. So this component stated out as more
```
lougeniaC64, line 31: nit: "started" out

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+See the "sync manager" for how multiple stores are managed (but the tl;dr is
+that the "sync manager" leans on this very heavily, but knows about multiple
+stores and manages shared state)
+
+## The `Store` trait
+
+The store trait is where all logic specific to a collection lives. For
+<handwave> reasons, it actually lives in the
+[sync-traits helper](https://github.com/mozilla/application-services/blob/main/components/support/sync15-traits/src/store.rs)
+but for the purposes if this document, you should consider it as owned by sync15.
```
lougeniaC64, line 97: nit: "of" this document

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+* Each store/engine should store its own metadata, so we don't end up in the
+  situation where, say, a database is moved between profiles causing the
+  metadata to refer to a completely different data set. So each store/engine
+  stores its metadata in the same database as the data itself, so if the
+  database is moved or copied, the metadata comes with it)
+
+## Sync Implementation
+
+The core implementation does all of the interaction with things like the
+token-server, the `meta/global` and `info/collections` collections, etc. It
```
lougeniaC64, line 67: I'm not sure how important understanding the details of these interactions are, but is it worthwhile to include links to further info on `meta/global` or `info/collections`?

components/sync_manager/README.md
```diff
@@ -1,5 +1,25 @@
...
+a single coherent framework. There are multiple, independent rust components,
+all of which know how to sync - but when an app uses multiple components,
+something must tie them together so that, eg, a single "sync now" function
+exists to sync all components, and to help the app manage the "enabled" state
+of these engines. There's other non-obvious "global state" which should also
+be shared - eg, we only need a single token from the token-server and can share
+it across all components, if the server is under load and wants clients to
+back off, that state should be shared.
+
+This crate very heavily on `sync15`, so you should read the documentation in
```
lougeniaC64, line 13: nit: I think you might be missing something like "relies" in this sentence.

lougeniaC64 (approved):
Thanks for these updates, particularly the history and the distinction between sync15, sync15-traits, and sync manager.

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+There are 2 key concepts to understand here - the implementation itself, and
+a rust trait for a "syncable store" where component-specific logic lives - but
+before we dive into them, some preamble might help put things into context.
+
+## Introduction and History
+
+For many years Sync has worked exclusively against a "sync v1.5 server". This
+[is a REST API described here](https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html).
+The important part is that the API is conceptually quite simple - there are
+arbitrary "collections", indexed by a GUID, and lacking traditonal database
```
rfk, line 17: it's not obvious to me what you mean by "indexed by a GUID", something like "containing items that and indexed by a GUID" may be a little clearer.

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+concepts like joins. Because the payload is encrypted, there's very little
```
rfk, line 18: nit: "payload" sounds like it might be a technical term that new readers may worry that they're already supposed to understand; something like "the stored items are encrypted" may be better.

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+
+This was also pre-megazord - the idea was that apps could choose from a "menu"
+of components to include - so we didn't really want to bind these components
+together. Therefore, there was no concept of "sync all" - instead, each of the
+components had to be synced individually. So this component stated out as more
+of a "library" than a "component" which individual components could reuse - and
+each of these components was a "syncable store".
+
+Fast forward to Fenix and we needed a UI for managing all the engines supported
+there, and a single "sync now" experience etc - so we also have a sync_manager
```
rfk, line 36: Is it also true that were were some bugs here when syncing just a single component, e.g. not respecting the user's data choices?

components/sync15/README.md
rfk, line 67: nit: "token-server" => "tokenserver"

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+The core implementation does all of the interaction with things like the
+token-server, the `meta/global` and `info/collections` collections, etc. It
+does all network interaction (ie, individial stores don't need to interact with
+the network at all), tracks things like whether the server is asking us to
+"backoff" due to operational concerns, manages encryption keys and the
+encryption itself, etc. The general flow of a sync - which interacts with the
+`Store` trait - is:
+
+* Does all pre-sync setup, such as checking `meta/global`, and whether the
+  sync IDs on the server match the sync IDs we last say (ie, to check whether
```
rfk, line 75: maybe nit: "say" -> "saw"? I don't think "say" makes sense in this context.

components/sync15/README.md
```diff
@@ -5,4 +5,108 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+single component using sync15 (ie, without going via the "sync manager"),
+although this isn't used and should be removed - the "sync manager" allows you
+to choose which engines to sync, so that should be used exclusively.
+
+## Metadata
+
+There's some metadata associated with a sync. Some of the metadata is "global"
+to the app (eg, the enabled state of engines, information about what servers to
+use, etc) and some is specific to an engine/store (eg, timestamp of the
+server's collection for this engine/store, guids for the collections, etc).
```
rfk, line 49: Hrm, I'm not a fan of the way that "engine/store" keeps showing up here, since we explicitly talk about a trait named `Store` but we don't really say what an "engine" is. Is this assumed knowledge from elsewhere that we should link out to? Or maybe the distinction isn't that important at this level and we could just say "store" consistently?

components/sync_manager/README.md
```diff
@@ -8,6 +28,10 @@ that `sync15` puts up.
...
-underlying crates. This seems extremely difficult to me, so this is split out
+underlying crates. This seems extremely difficult, so this is split out
 into it's own crate, which might happen to reach into the guts of `sync15` in
-some cases.
\ No newline at end of file
+some cases.
+
+Note also that the world has changed a little since then - in particular, we
+now have `sync15-traits` - there might now be an opportunity to move all the
+types into that crate, and merge `sync15` and `sync_manager`
```
rfk, line 37: Thank you for adding these encouragements to our future selves that it's fine and good and worthwhile to pursue such cleanup work.

rfk (approved):
I left a bunch-o-nits as requested, but this reads well to me 👍

components/sync15/README.md
mhammond: I would if I could find such links :/ The doc I linked to above is the closest I can find, but they are fairly vague on that.

components/sync15/README.md
mhammond: I don't recall any - before the sync manager we never attempted to update the enabled state and only used what was already on the server. I guess there would have been an edge-case when the server was empty and a desktop was yet to upload the declined state it has persisted, but I'm not sure there's value trying to explain that in detail here.

components/sync15/README.md
mhammond: That's a really good point, so I ended up adding an entire section dedicated to this - so I'm going to re-request review from both of you, specifically for 0b3ba8075b56b79a811773c46d8338562fad55ee (which I'll squash before merging, but left it separate for review)

components/sync15/README.md
```diff
@@ -5,4 +5,162 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+
+* We should reuse the term "engine" (or ideally, "sync engine") to talk about
+  the thing that actually syncs - it's not a great name, but at least it's
+  consistent with desktop.
+
+* We should rename things that are named "Store" in the old components (ie,
+  logins, places, tabs)
+
+So in this document, I'm going to use the term "engine", even though some of
+the code refers to the term "store".
```
rfk, line 58: This is excellent, and solidified my own understanding of these terms; thanks!

components/sync15/README.md
```diff
@@ -5,4 +5,162 @@ data stores that want to sync against a Firefox Sync v1.5 sync server.
...
+Before we start, a quick note on names.
+
+In Desktop Firefox, we have 2 main concepts for syncing - there's a "store"
+(which is an interface to the underlying storage) and an "engine" (the thing
+that interfaces between the "store" and the storage server to do the actual
+syncing). Note that the "store" is an interface to the store, not the store
+itself - eg, the "passwords" store exists in the "sync" directory and is an
+interface to the actual password storage code that exists externally in the
+browser. In other words, the sync functionality and how the data was actually
+stored were completely separate.
```
rfk, line 23: FWIW I'd really like to see some diagrams here, but that seems well outside the scope of this PR!

rfk (approved):

PR #3757, opened 2020-12-14
Rename 'engine' and 'store' to conform with the sync15 readme.

mhammond:
In many cases, this was literally changing 'engine' to 'store' and
'store' to 'engine', but now we have mostly consistent naming
throughout the repo, and consistency with desktop.

The sync trait is now called `SyncEngine` just to drive home the
fact that 'engine' relates to syncing.

This was largely a mechanical change. No changes were made to
the FFI functions, so this is not a breaking change.

### DIFF
--- a/components/logins/ffi/src/lib.rs
+++ b/components/logins/ffi/src/lib.rs
@@ -14,7 +14,7 @@ use ffi_support::{
     define_string_destructor, ByteBuffer, ExternError, FfiStr,
 };
 use logins::msg_types::{PasswordInfo, PasswordInfos};
-use logins::{Login, LoginDb, PasswordEngine, Result};
+use logins::{Login, LoginDb, PasswordStore, Result};
 use std::os::raw::c_char;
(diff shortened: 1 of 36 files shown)


mhammond:
Inspired by #3753 I thought I should just bite the bullet.

rfk (approved):
Looks great, thank you!

The tabs changes were a bit hard to review, but IIUC it's basically that the files named `store.rs` and `engine.rs` have swapped names, and github then makes a mess of presenting the diff. Does that sound accurate?

mhammond:
> The tabs changes were a bit hard to review, but IIUC it's basically that the files named `store.rs` and `engine.rs` have swapped names, and github then makes a mess of presenting the diff. Does that sound accurate?

Yeah - I did `git mv` them, but it looks like the subsequent changes confused git (it's not just github, git itself does the same)

lougeniaC64 (approved):
Thanks for this! The clarity that results from this change cannot be understated. 🎉🎉

PR #3770, opened 2020-12-15
Update uniffi dependency to v0.6.1

rfk:
Dependabot tried to do this in https://github.com/mozilla/application-services/pull/3768 but it doesn't know about the secret extra places where the uniffi version number if specified.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ commands:
             rustc --version
       - run:
           name: Install UniFFI
-          command: cargo install --version 0.6.0 uniffi_bindgen
+          command: cargo install --version 0.6.1 uniffi_bindgen
   build-libs:
     parameters:
       platform:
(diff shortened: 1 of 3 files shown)


rfk:
Hrm. These are failing to build due to version mismatch with uniffi in nimbus-sdk:

```
[task 2020-12-15T01:58:31.754Z] thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: The `uniffi` dependency version (0.6.0) is different than `uniffi-bindgen` own version (0.6.1). Please rectify the problem and try again.', components/external/nimbus-sdk/nimbus/build.rs:7:5
[task 2020-12-15T01:58:31.754Z] stack backtrace:
```

But nimbus only specifies `0.6` in its Cargo.toml, so it *should* be compatible with the `0.6.1` version we're using here.

If I deleted the `Cargo.lock` file in nimbus-sdk then things seem to work correctly.

rfk:
Blocked on https://github.com/mozilla/nimbus-sdk/pull/79, after which we can update the nimbus-sdk submodule.

rfk:
Going to merge this into https://github.com/mozilla/application-services/pull/3778

Issue #871, opened 2019-03-28
Review github permissions and tokens granted to CircleCI


rfk:
Per https://github.com/mozilla/application-services/pull/826#discussion_r267871825, CircleCI is currently configured with a personal token from @eoger's github account, with `repo:public_repo` scope. We should investigate whether we can generate some sort of Mozilla-owned token, and whether it can have reduced permissions.

┆Issue is synchronized with this [Jira Story](https://jira.allizom.org/browse/SYNC-141)
┆Story Points: 3

linacambridge:
FTR, I've added another SSH [deploy key](https://github.com/mozilla/application-services/settings/keys) for publishing the product portal (#982). It's scoped to just this repo, and, unlike a personal access token, I don't think it's tied to my personal account...but it is read-write.

eoger:
@hwine we've told that you are the right person to talk to for these things.
We currently expose our token to trusted jobs using the `GITHUB_TOKEN` environment variable in CircleCI for two tasks:
- Publishing to our [Github Releases](https://github.com/mozilla/application-services/releases): https://github.com/mozilla/application-services/blob/e5bafb05c5b86bc69a4ed256d5f6af51554f8bd7/.circleci/config.yml#L368-L375
- [Synchronizing Bugzilla issue onto our Github issues](https://github.com/mozilla/application-services/pull/1544/).

What would be the procedure to ensure we do not rely on my personal OAuth token to run these tasks?

hwine:
_[I do not believe a deploy key can be used for either use case, as it only grants permissions to the git repository, and not the GitHub resources associated with the repository, such as issues.]_

## Creating Releases
There are at least two sets of questions here:
1. Is Circle CI ⇾ GitHub releases an appropriate release process? (should have an RRA on that)
2. Is using a PAT a reasonable way to authenticate those operations?

I'll only focus on (2) here (y'all know the drill for (1)). PATs are OAUTH tokens, which means the permissions can not be filtered to a single repository. If the token is lost, all repositories accessible by the user are at risk. That suggests using a "service account" added as a collaborator only to the specified repository. See GitHub's terms of service on these.

Further, the token can be extracted from the CI job, so you should never run those CI jobs on pull requests from another repo. (That is, restrict automatic-ci-with-credentials to folks who can already commit directly to the repo.)

## Syncing bugs

[...]

PR #3776, opened 2020-12-15
Update build-and-publish-pipeline.md

eoger:
As of #871, we're using a token generated by appsvc-moz instead.


eoger:
(I've also transferred the login info of that account to @mhammond)

rfk (approved):
👍

PR #3782, opened 2020-12-17
Remove ecosystem_user_id and related functionality.

mhammond:
I was starting to play with telemetry.rs and noticed this stuff which, for better or worse, is now stale. The initial commits had a lot more than this, but I think this is the appropriate reversion. Ed, WDYT?


eoger (approved):
I assume with the new ecosystem telemetry proposal this is now dead code.
LGTM!

PR #3783, opened 2020-12-17
Add specific EndpointError for unhandled push error

### LINKED ISSUE #3737: Ensure push subscription endpoints are nullable or verify as non-null
@mhammond mentioned that desktop would sometimes not receive an endpoint in the subscription and something that we should check that we do as well.

In the Android Component, we expect a non-null value from the rust component, so if we did receive one, we would report that to Sentry (at this time, I don't see anything that sticks out with missing endpoints in Sentry).

We should check that missing values are not passed to us from the rust component as empty strings instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2579)
┆Epic: Important backlog

jonalmeida:
We're treating the failure to retrieve a push endpoint as a separate
error so that we can handle the exception explicitly on the consumer
end (e.g. re-try).

I looked into adding tests, but I ran into the same [issues](https://github.com/mozilla/application-services/blob/main/components/push/src/communications.rs#L441-L443) as mentioned in the test code of `communications.rs`.

Closes #3737

### DIFF
--- a/components/push/src/communications.rs
+++ b/components/push/src/communications.rs
@@ -20,7 +20,9 @@ use viaduct::{header_names, status_codes, Headers, Request};
 use crate::config::PushConfiguration;
 use crate::error::{
     self,
-    ErrorKind::{AlreadyRegisteredError, CommunicationError, CommunicationServerError},
+    ErrorKind::{
+        AlreadyRegisteredError, CommunicationError, CommunicationServerError, EndpointRegistrationError,
+    },
 };
 use crate::storage::Store;
 
@@ -200,9 +202,7 @@ impl Connection for ConnectHttp {
         {
             Ok(v) => v,
             Err(e) => {
-                return Err(
-                    CommunicationServerError(format!("Could not fetch endpoint: {}", e)).into(),
-                );
+                return Err(EndpointRegistrationError(format!("Could not fetch endpoint: {}", e)).into());
             }
         };
         if requested.is_server_error() {
--- a/components/push/src/error.rs
+++ b/components/push/src/error.rs
@@ -56,6 +56,10 @@ pub enum ErrorKind {
     /// A failure to parse a URL.
     #[error("URL parse error: {0:?}")]
     UrlParseError(#[from] url::ParseError),
+
+    /// An error if the endpoint could not be fetched
+    #[error("Endpoint Error: {0:?}")]
+    EndpointRegistrationError(String),
 }
 
 // Note, be sure to duplicate errors in the Kotlin side
@@ -74,6 +78,7 @@ impl ErrorKind {
             ErrorKind::TranscodingError(_) => 31,
             ErrorKind::RecordNotFoundError(_, _) => 32,
             ErrorKind::UrlParseError(_) => 33,
+            ErrorKind::EndpointRegistrationError(_) => 34,
         };
         ffi_support::ErrorCode::new(code)
     }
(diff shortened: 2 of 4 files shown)


components/push/src/communications.rs
```diff
@@ -200,9 +202,7 @@ impl Connection for ConnectHttp {
         {
             Ok(v) => v,
             Err(e) => {
-                return Err(
-                    CommunicationServerError(format!("Could not fetch endpoint: {}", e)).into(),
-                );
+                return Err(EndpointError(format!("Could not fetch endpoint: {}", e)).into());
```
rfk, line 205: So IIUC, this will throw `EndpointError` is there is e.g. a network-level error while subscribing, but will continue throwing a `CommunicationError` if the server successfully returns a 400-level HTTP error code.

From the PR description it sounds like you may also want to return `EndpointError` in the case of an error response from the server, is that accurate? If so, we'll also need to update the places below where we `return Err(...)`.

components/push/src/error.rs
```diff
@@ -56,6 +56,10 @@ pub enum ErrorKind {
     /// A failure to parse a URL.
     #[error("URL parse error: {0:?}")]
     UrlParseError(#[from] url::ParseError),
+
+    /// An error if the endpoint could not be fetched
+    #[error("Endpoint Error: {0:?}")]
+    EndpointError(String),
```
rfk, line 62: Minor naming feedback: when I read "endpoint error" my brain parses it as "there was an error with the endpoint", but in this case there was not yet an endpoint at all. Something like "EndpointRegistrationError" may be a better fit in amongst the other names here, although I don't bikeshed the details.

rfk (changes requested):
Sorry I didn't get to respond here before the break, and thanks for digging in! I'm not too worried about the tests, but I did have a question about the intended scope of the change...

components/push/src/error.rs
jonalmeida: I am terrible with naming things - EndpointRegistrationError sounds apt!

components/push/src/communications.rs
jonalmeida: That depends, from what I understand, the other errors are related to the client-server communication which could be a general error and not for the specific case where we can successfully establish an HTTP connection but the retrieving the endpoint fails.

`is_server_error` seems to relate to 500/600 http error codes, and `is_client_error` is for 400/500.

This line with `CommunicationServerError(format!("Could not parse response: {:?}", e))` may be a candidate, but I'm not sure what "Could not parse response" refers to. Is it a parsing error from the corrupted JSON on the server or from a network failure?

jonalmeida:
No idea what this about. 🤔 

```
    Checking example-autofill-utils v0.1.0 (/home/circleci/project/examples/autofill-utils)
error: failed to run custom build command for `nimbus-sdk v0.6.3 (/home/circleci/project/components/external/nimbus-sdk/nimbus)`

Caused by:
  process didn't exit successfully: `/home/circleci/project/target/debug/build/nimbus-sdk-94991259dc04bc24/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-changed=./src/nimbus.idl
```

components/push/src/communications.rs
rfk: > I'm not sure what "Could not parse response" refers to. Is it a parsing error from the corrupted JSON on the
> server or from a network failure?

This is if the server returns invalid JSON in its otherwise-successful response.

jonalmeida:
Closing this PR for now, when this becomes a priority again, we can come back to it with the review fixes.

PR #3789, opened 2020-12-22
Apply fixes for XCode 12 when running a local carthage build.

rfk:
Carthage + XCode 12 requires some hackery to work around a build issue.
In https://github.com/mozilla/application-services/pull/3586/ we
landed said hackery as part of our CI config. This commit pushes
it down into the build script itself, since the same issue is
present when trying to do a local Carthage build.

I lost a couple of hours to inscrutable swift-protobuf build issues this morning
until I managed to track this down :-(

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,9 +417,9 @@ jobs:
               # XCode tests build in Debug configuration, save us a full
               # Rust rebuild in Release mode by forcing Debug mode on
               # non-release builds.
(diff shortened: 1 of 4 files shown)


rfk:
Hrm, actually even with this change it's now not building for me locally, not sure what I did to get it to build before...

rfk:
Aha! Looks like I was not specifying an absolute path to the xcconfig file; updated to something that seems to work for me reliably.

eoger (approved):
Nice thanks!

PR #3800, opened 2021-01-06
Update Glean to v33.10.1

badboy:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,7 +1193,7 @@ checksum = "f6503fe142514ca4799d4c26297c4248239fe8838d827db6bd6065c6ed29a6ce"
 
 [[package]]
 name = "glean-core"
-version = "33.5.0"
+version = "33.10.1"
 dependencies = [
  "bincode",
  "chrono",
@@ -1201,7 +1201,7 @@ dependencies = [
  "flate2",
  "log 0.4.11",
  "once_cell",
- "rkv",
+ "rkv 0.16.1",
  "serde",
  "serde_json",
  "uuid",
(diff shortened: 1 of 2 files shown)


mhammond:
I'm a bit confused by the clippy failure - IIUC, `broken_intra_doc_links` is a rustdoc lint, so I guess it makes sense that it's "unknown" - but a quick google is failing to tell me how it should be resolved even though I'd expect it to be a very common problem. Any ideas @badboy?

badboy:
> I'm a bit confused by the clippy failure - IIUC, `broken_intra_doc_links` is a rustdoc lint, so I guess it makes sense that it's "unknown" - but a quick google is failing to tell me how it should be resolved even though I'd expect it to be a very common problem. Any ideas @badboy?

That's a Rust lint.
Intra-doc linking [became stable in Rust 1.48.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#rustdoc).
Because it's treated as an `unknown-lints` in older Rust we didn't really break our own stated compatibility (1.43.0) but of course that doesn't help here as it causes issues for a-s.
Let me see how I can resolve that best.

badboy:
I think the right solution here is to _not_ lint Glean code at all. Let's see if my fix does that.

Alternatively we need to add `-A unknown-lints` to the allow list.

badboy:
There we go, excluding it didn't actually work (because Glean is not part of the workspace, but then I don't even understand why it is linting that. It shouldn't lint dependencies, should it?)

So I went  with allowing `unknown-lints` instead.

badboy:
(with #3801 I'm going to remove the last commit here)

mhammond (approved):

PR #3804, opened 2021-01-11
Fix Rust Beta warning about deprecated `AtomicBool` method.

rfk:
When compiling on Rust Beta, we're generating a warning about:

  warning: use of deprecated associated function
  `std::sync::atomic::AtomicBool::compare_and_swap`:
  Use `compare_exchange` or `compare_exchange_weak` instead

This commit does as advised in the hope of eliminating the warning,
which is [currently failing some scheduled CI jobs](https://app.circleci.com/pipelines/github/mozilla/application-services/22759/workflows/0a5d643e-b7e4-4b28-ba49-81b7daabde77/jobs/91081).

### DIFF
--- a/components/places/src/api/places_api.rs
+++ b/components/places/src/api/places_api.rs
@@ -179,23 +179,19 @@ impl PlacesApi {
     }
 
     pub fn open_sync_connection(&self) -> Result<SyncConn<'_>> {
-        let prev_value = self
-            .sync_conn_active
-            .compare_and_swap(false, true, Ordering::SeqCst);
-        if prev_value {
-            Err(ErrorKind::ConnectionAlreadyOpen.into())
-        } else {
-            let db = PlacesDb::open(
-                self.db_name.clone(),
-                ConnectionType::Sync,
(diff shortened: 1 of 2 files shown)


mhammond (approved):
This looks correct to me, but I admit to not understanding why `compare_exchange` is documented as returning `current` but also saying it only succeeds when `current` == the current value - surely the return value can be implied in all cases? 

I'm also mildly surprised https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html doesn't note that `compare_and_swap` is deprecated!

Regardless, neither of those points are strictly relevant to this patch.

rfk:
> I admit to not understanding why compare_exchange is documented as returning current but also saying it only succeeds
> when current == the current value - surely the return value can be implied in all cases?

Yeah, I struggle a bit here wondering if I was missing something, but found https://github.com/rust-lang/rust/issues/31767#issuecomment-187432963 which suggests there isn't anything deep going on, it's just returning the info because it happens to be available.

PR #3811, opened 2021-01-12
Add config file to disable dependabot PRs.

rfk:
As discussed in today's team meeting, we're not really using the dependabot PRs so let's disable them pending a better policy/approach here. Ref https://github.com/mozilla/application-services/issues/3809 for broader discussion.


rfk:
The docs for this file format are [here](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#open-pull-requests-limit).

mhammond (approved):
According to the linked doc, it seems we probably *can* still expect dependabot to open PRs for security updates which is a bonus.

PR #3832, opened 2021-01-22
verify-android-environment script updates

grigoryk:
Running ANDROID_HOME/tools/bin/sdkmanager with Java 11 configured in the shell fails with an unhelpful stacktrace. So I've moved that after the java version check, removed a duplicate call, and added a hint (assuming sdkman is installed) that will switch to a required version.

Also added a call to install uniffi_bindgen which is now necessary for the builds to succeed.

Example output with these changes:
```
➜  application-services git:(verifyAndroid) ✗ ./libs/verify-android-environment.sh
rustc 1.49.0 (e1884a8e3 2020-12-29)
info: component 'rust-std' for target 'aarch64-linux-android' is up to date
info: component 'rust-std' for target 'armv7-linux-androideabi' is up to date
info: component 'rust-std' for target 'i686-linux-android' is up to date
info: component 'rust-std' for target 'x86_64-linux-android' is up to date
Incompatible java version: 11.0. JDK 8 must be installed.
Try switching versions and re-running. Using sdkman: sdk install java 8.0.282+8.hs-adpt || sdk use java 8.0.282+8.hs-adpt
➜  application-services git:(verifyAndroid) ✗ sdk install java 8.0.282+8.hs-adpt || sdk use java 8.0.282+8.hs-adpt

Stop! java 8.0.282+8.hs-adpt is already installed.

Using java version 8.0.282+8.hs-adpt in this shell.
➜  application-services git:(verifyAndroid) ✗ ./libs/verify-android-environment.sh
rustc 1.49.0 (e1884a8e3 2020-12-29)
info: component 'rust-std' for target 'aarch64-linux-android' is up to date
info: component 'rust-std' for target 'armv7-linux-androideabi' is up to date
info: component 'rust-std' for target 'i686-linux-android' is up to date
info: component 'rust-std' for target 'x86_64-linux-android' is up to date
Warning: File /Users/grishakruglov/.android/repositories.cfg could not be loaded.
[=======================================] 100% Computing updates...
Installing uniffi_bindgen if missing; it's necessary for binding generation during the builds.
    Updating crates.io index
     Ignored package `uniffi_bindgen v0.6.1` is already installed, use --force to override
Looks good! Try building with ./gradlew assembleDebug
➜  application-services git:(verifyAndroid) ✗
```

### DIFF
--- a/libs/verify-android-environment.sh
+++ b/libs/verify-android-environment.sh
@@ -68,4 +68,7 @@ if [[ -z "${CI}" ]]; then
   fi
 fi
 
+echo "Installing uniffi_bindgen if missing; it's necessary for binding generation during the builds."
+cargo install uniffi_bindgen --version 0.6.1
+
 echo "Looks good! Try building with ./gradlew assembleDebug"
--- a/libs/verify-android-environment.sh
+++ b/libs/verify-android-environment.sh
@@ -24,12 +24,6 @@ if [[ -z "${ANDROID_HOME}" ]]; then
   exit 1
 fi
 
-# NDK ez-install.
-"$ANDROID_HOME/tools/bin/sdkmanager" "ndk;$(./gradlew -q printNdkVersion | tail -1)"
-
-# NDK ez-install
-"$ANDROID_HOME/tools/bin/sdkmanager" "ndk;$(./gradlew -q printNdkVersion | tail -1)"
-
 rustup target add "${RUST_TARGETS[@]}"
 
 # Determine the Java command to use to start the JVM.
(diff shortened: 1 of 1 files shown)


libs/verify-android-environment.sh
```diff
@@ -70,4 +68,7 @@ if [[ -z "${CI}" ]]; then
   fi
 fi
 
+echo "Installing uniffi_bindgen if missing; it's necessary for binding generation during the builds."
+cargo install uniffi_bindgen
```
jonalmeida, line 72: This is great! I ran into this dependency missing as well a few weeks ago. I think we're pinned to 0.6.1 iirc.

```suggestion
cargo install -f uniffi_bindgen --version 0.6.1
```

grigoryk: Thanks, updated to use the pinned version number.

libs/verify-android-environment.sh
```diff
@@ -70,4 +68,7 @@ if [[ -z "${CI}" ]]; then
   fi
 fi
 
+echo "Installing uniffi_bindgen if missing; it's necessary for binding generation during the builds."
+cargo install uniffi_bindgen --version 0.6.1
```
rfk, line 72: nit: please move this into `verify-common.sh` since it applies to all platforms, not just Android.

rfk (approved):
Thanks!

libs/verify-android-environment.sh
gabrielluong: From https://github.com/mozilla/application-services/pull/3836/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R103, we actually need `cargo install --version 0.7.0 uniffi_bindgen` now.

PR #3833, opened 2021-01-24
Upgrade to Xcode 11.7 (0.59)

st3fan:
This patch updates Xcode from 11.4.2 to 11.7 on the 0.59 branch. This is because it is likely we need to do a Lockwise iOS update, and Xcode 11.4 is not available anymore on BuddyBuild.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,7 +230,7 @@ commands:
 jobs:
   Check Swift formatting:
     macos:
-      xcode: "11.4.1"
+      xcode: "11.7"
     steps:
       - checkout
       - run: brew install swiftlint swiftformat
@@ -322,7 +322,7 @@ jobs:
       - save-sccache-cache
   iOS build and test:
     macos:
-      xcode: "11.4.1"
+      xcode: "11.7"
     steps:
       # We do not use the ssccache cache helper commands as
       # the macOS cache is in a different folder.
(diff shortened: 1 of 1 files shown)


st3fan:
@rfk Some dependencies are flagged with security issues. Do you think those should be updated? It looks like most of them have a minor patch revision to deal with those issues. It is unclear what the risk is - but if acceptable then I am also happy to take the library as-is with the current versions.

st3fan:
Looking at https://github.com/rust-lang/futures-rs/releases, maybe all `futures-*` dependencies should go to 0.3.12? Let's see what happens.

rfk:
> Some dependencies are flagged with security issues. Do you think those should be updated? 

I lean towards not updating them TBH, as we want to minimize the risk of bustage and I don't recall any of the issues being things that could cause concrete problems for our use in practice.

rfk (commented):
This LGTM.

IIUC the various linting failures are due to using new stable versions of Rust, Swift etc that come with new linter warnings. I think they're fine to ignore.

I also think that it's fine for us to build this with a newer Rust than the previous releases in the `v0.59` series, because Rust does a pretty great job of keeping working code working across releases. But, I wanted to call it out explicitly as an additional thing that will change with the new build.

st3fan:
> I also think that it's fine for us to build this with a newer Rust than the previous releases in the `v0.59` series, because Rust does a pretty great job of keeping working code working across releases. But, I wanted to call it out explicitly as an additional thing that will change with the new build.

Great. Let's use this for a 0.59.1 (?) update then. Lockwise iOS also has a good number of tests and we will ask QA for a smoke test too. That gives me enough confidence that things work as expected.

rfk (approved):

rfk:
PR for cutting the release is here: https://github.com/mozilla/application-services/pull/3835

Issue #3826, opened 2021-01-20
Cut a release of Application Services


data-sync-user:
A release needs to be cut for bumping the xcode version #3825 and for cutting a new release of nimbus. The iOS team doesn't need this immediately but would like to have it this sprint.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2013)
┆Epic: Fenix Client SDK
┆Sprint: SYNC - 2021-01-29

PR #3834, opened 2021-01-25
Upgrade to Glean 33.10.3

badboy:
Would be good to land the Glean update before a release (#3826), so we can get that into Firefox iOS as well.


rfk (approved):
👍 thanks!

PR #3835, opened 2021-01-25
Cut release v0.59.1

rfk:
This is in service of the XCode update in https://github.com/mozilla/application-services/pull/3833

### DIFF
--- a/.buildconfig-android.yml
+++ b/.buildconfig-android.yml
@@ -1,4 +1,4 @@
-libraryVersion: 0.59.0
+libraryVersion: 0.59.1
 groupId: org.mozilla.appservices
 projects:
   fxaclient:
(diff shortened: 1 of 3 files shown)


rfk:
Sorry @lougeniaC64, I should have said like 10 times as many words for context-setting here.

@st3fan is trying to make a point-release of Lockwise, which is currently on an old release (v0.59.0) of appservices. To get it to build we need a version of appservices build with a newer XCode, which was the trigger for #3833. This is a point-release to get that change out in a build, so it's targetting the `release-v0.59` branch rather than `main`.

lougeniaC64 (approved):

rfk:
AFAIK this build worked out OK, now we just have to merge it to master for completeness.

PR #3839, opened 2021-01-28
Leverage more of serde for autofill

mhammond:
This PR:

* allows us to use serde for guids (which my tests will want)
* exposes a new `JsonError` so we can use `?` on serde-related functions.
* Moves some serde "default" attributes from the fields to the container - if we want to insist on some fields then we should probably do something other than just causing a serde error.

Most of this is really just for tests, but I thought I'd split it out to keep that other branch less painful!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,7 @@ dependencies = [
  "serde_json",
  "sql-support",
  "sync-guid",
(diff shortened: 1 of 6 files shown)


lougeniaC64 (approved):
LGTM!

PR #3842, opened 2021-01-29
WIP for exposing address metadata and rationalizing some structs

mhammond:
In the sync branch I discovered that we need to expose some metadata to
the consuming app - eg, so Fenix knows how often an address/creditcard
has been used.

As a part of this, I tried to rationalize some of the structs used.
This is a fairly large change, but I think it simplifies things.

Of special note:

* By adding a new `guid` param to `update_address()`, I've been able to 
  reuse the same struct as `add_address()`, which is good because
  `update_address()` shouldn't be able to update that metadata.

* Only addresses so far - I'll do creditcard if you think it's worth
  continuing with.

* No tests have been updated - `cargo check` works, but tests don't.
  These should also be easy.

* I replaced the `ExternalizeAddress` trait with the builtin `From/Into`
  mechanism.

Let me know what you think!

### DIFF
--- a/components/autofill/sql/create_shared_schema.sql
+++ b/components/autofill/sql/create_shared_schema.sql
@@ -3,7 +3,7 @@
 -- file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 CREATE TABLE IF NOT EXISTS addresses_data (
-    guid          TEXT NOT NULL PRIMARY KEY,
+    guid                TEXT NOT NULL PRIMARY KEY,
     given_name          TEXT NOT NULL,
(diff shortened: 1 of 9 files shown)


mhammond:
Ended up fixing the tests :)

lougeniaC64 (approved):
I really like the simplicity of the structs in this refactor!  The changes to the address `update` call may pave the way for a smarter design for the credit card `update` as well. We'll probably need to discuss our merge strategy for this PR and the [sync15 work](https://github.com/mozilla/application-services/tree/add-autofill-sync15) but otherwise LGTM!

PR #3844, opened 2021-01-29
Upgrade to Glean 34.0.0

badboy:
---

New update. Given that there just was a release of application-services it's not super critical to get that out in a release quickly.


rfk (approved):
CI says 👍, so 👍; thanks @badboy!

PR #3847, opened 2021-02-01
Update documentation to reflect that iOS now builds a dylib.

rfk:
Because:
* We recently switched from building a static library on iOS to
  building dynamic library (ref #3822).
* Several places in our docs explicitly mention that iOS builds
  are static libraries.

This commit:
* Updates the documentation to note that iOS builds are dynamic
  libraries.
* Deliberately doesn't try to say anything clever about custom megazords
  on iOS, since that's very much still to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
  - Updated nimbus-sdk to 0.7.1
  - Updated Android Components to 71.0.0
 
+## iOS
+
+### What's Changed
+ - The `MozillaAppServices.framework` artifact now contains a dynamically-linked library
+   rather than a static library.
+
 # v68.1.0 (_2020-12-17_)
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v68.0.0...v68.1.0)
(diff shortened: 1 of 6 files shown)


CHANGES_UNRELEASED.md
```diff
@@ -3,3 +3,5 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v69.0.0...main)
+
+- The bundled version of Glean has been updated to v34.0.0.
```
rfk, line 7: Yeah OK, this was an unrelated changes, but I added it because I was messing around with changelogs already...

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@
  - Updated nimbus-sdk to 0.7.1
  - Updated Android Components to 71.0.0
 
+## iOS
+
+### What's Changed
+ - The `MozillaAppServices.framework` artifact now contains a dynamiclly-linked library
```
mhammond, line 14: nit: typo in "dynamically"

mhammond (approved):

PR #3854, opened 2021-02-04
Cut release v70.0.0

rfk:


CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+
+### ⚠️ Breaking changes ⚠️
+
+- The `NewCreditCardFields` record is now called `UpdatableCreditCardFields`.
+- The `NewAddressFields` record is now called `UpdatableAddressFields`.
+
+### What's Changed
+
+- The `CreditCard` and `Address` records now exposes additional metadata around timestampes.
+- The ability to sync incoming address records has been added.
```
mhammond, line 22: That's not true :) The basis for it has landed, but I don't think that's worth calling out.

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+
+- The bundled version of Glean has been updated to v34.0.0.
+- The bundled version of the Nimbus SDK has been updated to v0.7.2.
+
+## Autofill
+
+### ⚠️ Breaking changes ⚠️
+
+- The `NewCreditCardFields` record is now called `UpdatableCreditCardFields`.
+- The `NewAddressFields` record is now called `UpdatableAddressFields`.
```
mhammond, line 17: There's actually another change I should have mentioned in the last review - `update_address` and `update_credit_card` have had their signature changed - both now take the GUID to update as a param. I don't think it matters in practice though given our zero consumers.

mhammond (approved):

CHANGELOG.md
rfk: Ah, thanks. I agree the details of breaking changes here probably aren't too important in practice yet.

PR #3855, opened 2021-02-04
Fix a-c smoketest script for upstream changes.

rfk:
The android-components repo now contains one component whose build
file is `build.gradle.kts` rather than `build.gradle`, so we need
to tweak our script to accommodate it.


mhammond (approved):
r+ all the things

rfk:
@mhammond thanks :-)

TBH I wasn't expecting anyone to get to these until tomorrow; my mental model of the world had you out for the rest of the day because of the melb lunch, but I forgot that's much less of a trip for you than it usually is for me 😆

PR #3857, opened 2021-02-05
Code cleanup for autofill incoming logic 

lougeniaC64:
Addressing the comments from PR ##3808

### DIFF
--- a/components/autofill/src/sync/address/incoming.rs
+++ b/components/autofill/src/sync/address/incoming.rs
@@ -31,16 +31,27 @@ pub fn stage_incoming(
     let mut incoming_tombstones = Vec::with_capacity(incoming_payloads.len());
 
     for payload in incoming_payloads {
+        let payload_id = payload.id.clone();
+        let payload_is_tombstone = payload.deleted;
         log::trace!(
             "incoming payload {} (deleted={})",
-            payload.id,
-            payload.deleted
+            payload_id,
+            payload_is_tombstone,
         );
-        match payload.deleted {
-            // XXX - unwraps below should be removed. We probably want to handle
-            // the error and log then ignore the error.
-            true => incoming_tombstones.push(payload.into_record::<AddressRecord>().unwrap()),
-            false => incoming_records.push(payload.into_record::<AddressRecord>().unwrap()),
+
+        match payload.into_record::<AddressRecord>() {
+            Ok(address) => {
+                if payload_is_tombstone {
+                    incoming_tombstones.push(address);
+                } else {
+                    incoming_records.push(address);
+                }
+            }
+            Err(e) => log::error!(
+                "failed to deserialize incoming payload {} into `AddressRecord`, {}",
+                payload_id,
+                e,
+            ),
         };
     }
     save_incoming_records(conn, incoming_records, signal)?;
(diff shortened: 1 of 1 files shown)


components/autofill/src/sync/address/incoming.rs
```diff
@@ -118,18 +126,18 @@ fn save_incoming_tombstones(
...
             let sql = format!(
                 "INSERT OR REPLACE INTO temp.addresses_tombstone_sync_staging (
                     guid
                 ) VALUES {}",
                 sql_support::repeat_multi_values(chunk.len(), chunk_size)
             );
-            let mut params = Vec::with_capacity(chunk.len() * chunk_size);
+            let mut params = Vec::with_capacity(chunk.len());
             for record in chunk {
                 signal.err_if_interrupted()?;
```
mhammond, line 142: While this isn't part of your patch, `err_if_interrupted` should move out of this loop, and just called once per chunk.

(and once you do that, this entire loop could be replaced with `let params: Vec<&dyn ToSql> = chunk.iter().map(|r| &r.guid as &dyn ToSql).collect();` - but I'm fine with keeping the loop explicit if you think it's clearer.

components/autofill/src/sync/address/incoming.rs
```diff
@@ -31,16 +31,24 @@ pub fn stage_incoming(
     let mut incoming_tombstones = Vec::with_capacity(incoming_payloads.len());
 
     for payload in incoming_payloads {
+        let payload_id = payload.clone().id;
+        let payload_is_tombstone = payload.clone().deleted;
```
mhammond, line 35: `.deleted` is a bool, so there should be no need to clone here at all!

mhammond (changes requested):
Thanks for taking this on - just a few changes!

components/autofill/src/sync/address/incoming.rs
```diff
@@ -266,18 +275,17 @@ fn get_incoming_record_states(conn: &Connection) -> Result<Vec<IncomingState>> {
     Ok(conn
         .conn()
         .query_rows_and_then_named(sql_query, &[], |row| -> Result<IncomingState> {
-            // XXX - change these to something like `mirror_exists` - that's how they are used.
-            let mirror_guid: Option<SyncGuid> = row.get_unwrap("m_guid");
-            let local_guid: Option<SyncGuid> = row.get_unwrap("l_guid");
-            let tombstone_guid: Option<SyncGuid> = row.get_unwrap("t_guid");
+            let mirror_exists: Option<SyncGuid> = row.get_unwrap("m_guid");
```
mhammond, line 278: Sorry, I missed this yesterday, but in that comment I meant that `mirror_exists` etc should become a bool as the guid is never actually used - ie, something like `mirror_exists = row.get_unwrap("m_guid").is_some();`- that doesn't work for the tombstone though because we actually use the guid.

components/autofill/src/sync/address/incoming.rs
```diff
@@ -118,22 +129,20 @@ fn save_incoming_tombstones(
     signal: &dyn Interruptee,
 ) -> Result<()> {
     log::info!("staging {} incoming tombstones", incoming_tombstones.len());
-    let chunk_size = 1; // XXX - chunk_size of 1 seems wrong?
     sql_support::each_sized_chunk(
         &incoming_tombstones,
-        sql_support::default_max_variable_number() / chunk_size,
+        sql_support::default_max_variable_number(),
```
mhammond, line 134: I thought I wrote this yesterday but apparently not :( I think you can change here to use `sql_support::each_sized` and drop the 2nd param, and `sql_support::repeat_multi_values` can change to `sql_support::repeat_sql_values` which also means you can drop the 2nd param, so that odd `chunk_size` can be removed entirely.

mhammond (changes requested):
Sorry about missing these yesterday :(

mhammond (approved):
Lovely!

PR #3860, opened 2021-02-09
Allow *_COMMON_COLS/VALS to be used in more places

mhammond:
Seems worthwhile to be able to use these constants in more places - WDYT?


lougeniaC64:
Yep, let's do it!

lougeniaC64 (approved):

PR #3867, opened 2021-02-12
Refactor address incoming handling into something that easier to share.

mhammond:
There's *still* no credit-card functionality here, but hopefully
it's very obvious how that will fit in without too much code
duplication.

I think this makes sense, but feel free to push back on anything that seems to cause more harm than good. Also, if you think it would be clearer to wait until CC functionality is in-place I'm happy to do that (I'm going to start on CC now anyway)


mhammond:
Superseded in #3869

PR #3869, opened 2021-02-16
Refactor to duplicate less code and get incoming syncing working with the autofill-utils example

mhammond:
I'm sending @lougeniaC64 an email with some of these details, but a summary is:

* The mirror and staging tables now store JSON (it was my suggestion to not do that in the first place, but I was wrong.) The amount of code duplication was making me sad, and I felt it was making things appear alot more complex for now real value.

* I dropped the custom structs for the sync code - eg, the address sync code now just uses InternalAddress.

* I refactored the engine code to share alot more code.

* autofill-utils is *that* close to working - we need to init viaduct


mhammond:
>     * autofill-utils is _that_ close to working - we need to init viaduct

It now works 🎉

components/autofill/sql/create_shared_schema.sql
```diff
@@ -18,37 +18,23 @@ CREATE TABLE IF NOT EXISTS addresses_data (
     email               TEXT NOT NULL,
 
     time_created        INTEGER NOT NULL,
-    time_last_used      INTEGER,
+    time_last_used      INTEGER NOT NULL DEFAULT 0,
```
lougeniaC64, line 21: Asking for my own understanding, what is the value of setting a default value here and not for the other timestamps.

mhammond: Good question: none, there's no good reason for that and it should be made consistent. In practice, we always supply the values which is why it works, so I guess it makes sense to remove the defaults? wdyt?

lougeniaC64: Yep, sounds good!

components/autofill/src/db/credit_cards.rs
```diff
@@ -138,6 +135,44 @@ pub fn update_credit_card(
...
 }
 
+/// Updates all fields including metadata - although the change counter gets
+/// slighly special treatment. Suitable for internal use (eg, by Sync)
+pub(crate) fn update_internal_credit_card(
+    tx: &Transaction<'_>,
+    card: &InternalCreditCard,
+    flag_as_changed: bool,
+) -> Result<()> {
+    let change_counter_increment = flag_as_changed as u32; // will be 1 or 0
```
lougeniaC64, line 145: I might be missing something here but under what conditions would we make an update call while not wanting to increment the `sync_change_counter`?

components/autofill/src/db/credit_cards.rs
```diff
@@ -231,7 +233,7 @@ mod tests {
         Ok(guids)
     }
 
-    fn clear_tables(conn: &Connection) -> rusqlite::Result<(), rusqlite::Error> {
+    pub fn clear_tables(conn: &Connection) -> rusqlite::Result<(), rusqlite::Error> {
```
lougeniaC64, line 236: This function may actually be called in other places, but I only see it being used in the engine code. If that's the case, how do you feel about moving the function there? Oh, and also renaming to `clear_credit_card_tables` for clarity's sake.

components/autofill/src/sync/credit_card/incoming.rs
```diff
@@ -0,0 +1,296 @@
...
+        &self,
+        tx: &Transaction<'_>,
+        old_guid: &SyncGuid,
+        new_guid: &SyncGuid,
+    ) -> Result<()> {
+        common_change_guid(tx, "credit_cards_data", old_guid, new_guid)
+    }
+
+    fn remove_record(&self, tx: &Transaction<'_>, guid: &SyncGuid) -> Result<()> {
+        common_remove_record(tx, "credit_cards_data", guid)
```
lougeniaC64, line 140: nit: It might be worth making the table names `const`s that can be used here and exported and used in place of all the string literal table names. Not a showstopper though.

components/autofill/src/sync/common.rs
```diff
@@ -0,0 +1,292 @@
...
+        assert_eq!(states.len(), 1, "1 records == 1 state!");
+        let action =
+            crate::sync::plan_incoming(ri, tx, states.pop().unwrap()).expect("plan should work");
+        // Even though the records are identical, we still merged the metadata
+        // so treat this as an Update.
+        assert!(matches!(action, crate::sync::IncomingAction::Update { .. }));
+    }
+
+    // Incoming tombstone for an existing local record.
+    pub(in crate::sync) fn do_test_incoming_tombstone<
```
lougeniaC64, line 267: nit: should this be limited to `crate::sync::tests`?

components/autofill/src/sync/credit_card/mod.rs
```diff
@@ -3,4 +3,113 @@
...
+            collection: "creditcards",
+        },
+        storage_impl: Box::new(CreditCardsEngineStorageImpl {}),
+    }
+}
+
+pub(super) struct CreditCardsEngineStorageImpl {}
+
+impl SyncEngineStorageImpl<InternalCreditCard> for CreditCardsEngineStorageImpl {
+    fn get_incoming_impl(&self) -> Box<dyn ProcessIncomingRecordImpl<Record = InternalCreditCard>> {
```
lougeniaC64, line 36: This is not a useful comment because I'm not quite sure what to suggest as an improvement, however . . . at a high level I understand what's happening in the `get_incoming_impl` function but I fear the uninitiated (or even future me) wouldn't.

lougeniaC64:
I've finished my review. This looks good. I'm approving though I'm not absolutely sure if you're finished. If you want me to review any additional changes, you can re-request review.

lougeniaC64 (approved):

components/autofill/src/db/credit_cards.rs
mhammond: Sync calls this function when updating due to an incoming record - the change counter should not increment in that case

components/autofill/src/sync/common.rs
mhammond: Sadly, these are in `crate::sync::common::tests` and are called by, `crate::sync::addresses::tests` and `crate::sync::credit_cards::tests` - so the only shared namespace is `crate::sync`

PR #3871, opened 2021-02-16
Update fxa client-id for command-line example apps from lockwise to the reference browser

mhammond:
It looks like the lockwise redirect URL no longer exists


rfk (approved):
👍

PR #3876, opened 2021-02-18
Restructure, UniFFI, and document the fxa-client crate.

rfk:
This is a substantial refactor of the fxa-client crate, intended to bring it
up to speed with our latest best-practices around developing cross-platform
Rust components, in order to ease ongoing maintenance.

~~It's still very much a work in progress, but the setup does seem to be working,
and feedback is always welcome.~~

[Edit: It's ready for review]

This introduces backwards-incompatible API changes, which I'm tracking in PRs to our downstream consumers here:

* Android Components: https://github.com/mozilla-mobile/android-components/pull/9731
* Firefox for iOS: https://github.com/mozilla-mobile/firefox-ios/pull/8044


components/fxa-client/android/build.gradle
```diff
@@ -108,6 +86,20 @@ dependencies {
...
 }
 
+// We do the uniffi binding generation here:
+android.libraryVariants.all { variant ->
+    def t = tasks.register("generate${variant.name.capitalize()}UniffiBindings", Exec) {
+        workingDir project.rootDir
+        commandLine 'cargo', 'uniffi-bindgen', 'generate', "${project.projectDir}/../src/fxa_client.udl", '--language', 'kotlin', '--out-dir', "${buildDir}/generated/source/uniffi/${variant.name}/java"
+
+        inputs.file '../src/fxa_client.udl'
+        outputs.dir "${buildDir}/generated/source/uniffi/${variant.name}/java/"
```
rfk, line 96: The `inputs`/`outputs` thing here helps gradle know when to re-run this command, and may be worth applying to other uniffied crates; /cc @lougeniaC64 if you want to try it out for autofil

components/fxa-client/src/internal/error.rs
```diff
@@ -138,12 +138,37 @@ error_support::define_error! {
...
     ErrorKind {
         (HawkError, hawk::Error),
+        (IntegerConversionError, std::num::TryFromIntError),
+    }
+}
+
+// The public FFI puts the errors into three buckets, this helps us
+// convert between them. Maybe in future we can use uniffi to expose
+// more error info to the caller?
+impl From<super::Error> for crate::FxaError {
```
rfk, line 154: I good future followup would be to lean on UniFFI to send much more fine-grained errors across the FFI.

components/fxa-client/src/internal/http_client.rs
```diff
@@ -22,58 +28,68 @@ const HAWK_HKDF_SALT: [u8; 32] = [0b0; 32];
 const HAWK_KEY_LENGTH: usize = 32;
 const RETRY_AFTER_DEFAULT_SECONDS: u64 = 10;
 
+/// Trait defining the low-level API for talking to the FxA server.
+///
+/// These are all the methods and datatypes used for interacting with
+/// the FxA server. It's defined as a trait mostly to make it mockable
+/// for testing purposes.
```
rfk, line 35: This trait, on the other hand, I just could not resist trying to clean up a little while I was in here...

components/fxa-client/src/internal/migrator.rs
```diff
@@ -2,41 +2,13 @@
...
+use super::{error::*, scoped_keys::ScopedKey, scopes, FirefoxAccount};
 use serde_derive::*;
+use std::convert::TryFrom;
 use std::time::Instant;
 
-// Values to pass back to calling code over the FFI.
-
-#[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Default)]
-pub struct FxAMigrationResult {
-    pub total_duration: u128,
```
rfk, line 14: I'm pretty sure the recipients of this field were not handling it as an unsigned 128-bit integer; I've made it an i64 in the new bindings.

components/fxa-client/src/internal/mod.rs
```diff
@@ -0,0 +1,560 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: Careful, this large diff is not rendered by default! :-)

Git thinks this is a new file, but it's actually the previous contents of `./src/lib.rs`, moved to `./src/internal/mod.rs` and only very lightly changed.

components/fxa-client/src/internal/oauth.rs
```diff
@@ -250,17 +252,12 @@ impl FirefoxAccount {
...
             access_type: auth_params.access_type,
-            code_challenge: auth_params
-                .pkce_params
-                .as_ref()
-                .map(|param| param.code_challenge.clone()),
-            code_challenge_method: auth_params
-                .pkce_params
-                .map(|param| param.code_challenge_method),
+            code_challenge: auth_params.code_challenge,
+            code_challenge_method: auth_params.code_challenge_method,
```
rfk, line 256: I didn't think that having the two PKCE params in a separate sub-struct was really pulling its weight as an abstraction, so I've just made them ordinary fields on the `AuthorizationParams` struct.

components/fxa-client/src/internal/oauth/attached_clients.rs
```diff
@@ -19,7 +21,7 @@ impl FirefoxAccount {
         let session_token = self.get_session_token()?;
         let response = self
             .client
-            .attached_clients(&self.state.config, &session_token)?;
+            .get_attached_clients(&self.state.config, &session_token)?;
```
rfk, line 24: By the way, this was not previously exposed over the FFI, but it was so trivial do so that I've gone ahead and exposed it for completeness.

components/fxa-client/src/lib.rs
```diff
@@ -2,597 +2,1268 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! # Firefox Accounts Client
```
rfk, line 5: Careful, this large diff is not rendered by default! :-)

This file is in fact brand new, and git's diff view is not going to help you here. I recommend just viewing it as a separate file.

This is the public API of the crate, from which we derived the `.udl` file and hence from which the Kotlin and Swift bindings are generated.

components/fxa-client/src/lib.rs
```diff
@@ -2,597 +2,1268 @@
...
-    include!("mozilla.appservices.fxaclient.protobuf.rs");
+use std::collections::HashMap;
+use std::convert::{TryFrom, TryInto};
+use thiserror::Error;
+
+// All the implementation details live in this "internal" module.
+// Aspirationally, I'd like to make it private, so that the public API of the crate
+// is entirely the same as the API exposed to consumers via UniFFI. That's not
+// possible right now because some of our tests/example use features that we do
+// not currently expose to consumers. But we should figure out how to expose them!
```
rfk, line 50: Just drawing attention to this aspirational comment for the future here. I think it would be good hygiene to have the UniFFIed interface be the *entire* public interface of the crate, so that e.g. our Rust examples use the same API surface as our eventual consumers. But that was a bigger refactor than I wanted to get into here, because some of our examples/tests use functionality that was not exposed over the FFI.

components/fxa-client/src/lib.rs
```diff
@@ -2,597 +2,1268 @@
...
+///
+/// Precise details of the error are hidden from consumers, mostly due to limitations of
+/// how we expose this API to other languages. The type of the error indicates how the
+/// calling code should respond.
+///
+#[derive(Debug, Error)]
+pub enum FxaError {
+    /// Thrown when there was a problem with the authentication status of the account,
+    /// such as an expired token. The application should [check its authorization status](
+    /// FirefoxAccount::check_authorization_status) to see whether it has been disconnected,
```
rfk, line 65: If you use a new-enough `cargo`, then `cargo doc` will automatically linkify references to structs and methods and what-not in this form. It's pretty nice!

components/places/src/history_sync/plan.rs
```diff
@@ -405,7 +405,10 @@ mod tests {
             visits,
         };
 
-        assert!(matches!(plan_incoming_record(&conn, record, 10), IncomingPlan::Apply{..}));
+        assert!(matches!(
+            plan_incoming_record(&conn, record, 10),
+            IncomingPlan::Apply { .. }
+        ));
```
rfk, line 411: unrelated `cargo fmt` change, I'm not sure why...

examples/fxa-client/src/devices-api.rs
```diff
@@ -4,7 +4,7 @@
 
 use cli_support::prompt::prompt_string;
 use dialoguer::Select;
-use fxa_client::{device, Config, FirefoxAccount, IncomingDeviceCommand};
+use fxa_client::internal::{device, Config, FirefoxAccount, IncomingDeviceCommand};
```
rfk, line 7: I think it would be really nice if this example used the public interface as exposed by UniFFI, but the `Config` abstraction is not part of the UniFFI API. It could probably be added, but I've gotta stop somewhere...

megazords/ios/DEPENDENCIES.md
```diff
@@ -889,6 +910,38 @@ OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
 IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 
+```
+-------------
+## MIT License: cargo_metadata
```
rfk, line 915: With sufficient effort, we might be able to relegate these to dev dependencies in a way that our dependency-script can ignore them. I haven't looked into it.

megazords/ios/MozillaAppServicesTests/FxAccountManagerTests.swift
```diff
@@ -203,12 +203,12 @@ class FxAccountManagerTests: XCTestCase {
...
             override func ensureCapabilities(supportedCapabilities _: [DeviceCapability]) throws {
                 notifyAuthError()
-                throw FirefoxAccountError.unauthorized(message: "Your token is expired yo.")
+                throw FxaError.Authentication(message: "Your token is expired yo.")
             }
 
-            override func checkAuthorizationStatus() throws -> IntrospectInfo {
+            override func checkAuthorizationStatus() throws -> AuthorizationInfo {
                 _ = try super.checkAuthorizationStatus()
-                return IntrospectInfo(active: false)
```
rfk, line 211: `IntrospectInfo` was really not a good name for the public API, where there is no operation called "introspect".

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/PersistedFirefoxAccount.kt
```diff
@@ -5,54 +5,49 @@
...
-        LibFxAFFI.INSTANCE.fxa_new(
-            config.contentUrl,
-            config.clientId,
-            config.redirectUri,
-            config.tokenServerUrlOverride,
-            e
-        )
-    }, persistCallback) {
+    constructor(config: Config, persistCallback: PersistCallback? = null) : this(FirefoxAccount(
+        config.contentUrl,
```
rfk, line 31: This is kind of dumb - we take a `Config` object on the Kotlin side, destructure it into its fields to pass over the FFI, then the Rust side turns it back into its own variant of a `Config` object! That made sense when we had to write the FFI layer by hand, but we should see whether we can nicely expose the Rust `Config` interface to Kotlin and Swift and then just accept a `Config` here in the constructor.

rfk (commented):
I'm probably going to do one more pass over the docstrings here to fill in some more implementation details, but I'm calling it ready for review. Please see the updated PR description for some high-level scene-setting and notes, and I'll leave some more detailed comments inline here.

components/fxa-client/src/fxa_client.udl
```diff
@@ -0,0 +1,1075 @@
+
+
+
```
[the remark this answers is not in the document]
rfk: Sadly, UniFFI is not capable of copying these comments over into the generated Kotlin or Swift files, because the parser we're using doesnt make them available. I expect that will change over time as we revisit the parsing story, so I've been happy to invest a little bit in the docstrings here, confident that they'll be or more use to us over time.

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
```diff
@@ -86,22 +83,31 @@
...
+			filePatterns = "*.udl";
+			fileType = pattern.proxy;
+			inputFiles = (
+			);
+			isEditable = 1;
+			outputFiles = (
+				"$(INPUT_FILE_DIR)/../ios/Generated/$(INPUT_FILE_BASE).swift",
+				"$(INPUT_FILE_DIR)/../ios/Generated/uniffi_$(INPUT_FILE_BASE)-Bridging-Header.h",
+			);
+			script = "cargo uniffi-bindgen generate $INPUT_FILE_PATH --language swift --out-dir $INPUT_FILE_DIR/../ios/Generated\n";
```
jhugman, line 109: I'm not sure what PATH weirdness I have got my bash profile into, but cargo isn't available through Xcode or xcodebuild.

```suggestion
			script = "$HOME/.cargo/bin/cargo uniffi-bindgen generate $INPUT_FILE_PATH --language swift --out-dir $INPUT_FILE_DIR/../ios/Generated\n";
```

rfk: Nice catch, thanks. I think our command-line build script does some magic to tell xcode to use the current value of `$PATH`, but we shouldn't assume things will always be built that way.

rfk:
@grigoryk I've pinged you for review here for the Kotlin-side changes, if you're happy to take a look. (Your feedback on the whole thing is of course most welcome! But don't feel obliged to dig in to all the Rust changes)

mhammond (commented):
Wow, this is looking great and is a huge improvement! It's a huge diff and I'm only 1/2 way through and need to leave for the day, so might as well submit what I have. I'll look more tomorrow.

components/fxa-client/README.md
```diff
@@ -0,0 +1,117 @@
...
+* The [core concepts](#core-concepts) for users of the component, outlined above.
+* The way we use [uniffi-rs](https://github.com/mozilla/uniffi-rs) to generate API wrappers for multiple languages, such as Kotlin and Swift.
+
+### Overview
+
+This component uses [uniffi-rs](https://github.com/mozilla/uniffi-rs) to create its
+public API surface in Rust, and then generate bindings for Kotlin and Swift. The
+code is organized as follows:
+
+* The public API surface is implemented in [`./src/lib.rs`](./src/lib/rs), with matching
```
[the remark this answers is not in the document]
rfk: Indeed. I think it would be good to pull it out into a common doc at some point, but IMHO that's premature until we get most of the components in a similar shape.

components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
```diff
@@ -0,0 +1,324 @@
...
+        inner.clearAccessTokenCache()
+    }
+
+    public func gatherTelemetry() throws -> String {
+        return try notifyAuthErrors {
+            try self.inner.gatherTelemetry()
+        }
+    }
+
+    // TODO: not sure why we switched to a bool API for the Swift wrapper here...
```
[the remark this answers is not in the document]
rfk: Heh, yeah, I guess it's not really a `TODO` so much as a `XXX I do not understand the point of this`

components/fxa-client/README.md
```diff
@@ -0,0 +1,117 @@
...
+  declarations in [`./src/fxa_client.udl`](./src/fxa_client.udl) to define how it gets
+  exposed to other languages.
+    * All the implementation details are written in Rust and can be found under
+      [`./src/internal/`](./src/internal).
```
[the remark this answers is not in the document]
rfk: I'm trying it out.

Broadly, I like having the contents of `src/lib.rs` mirror the public API exposed to foreign-language consumers, I think it makes it easier to form a consistent mental model of the component. Pushing all the non-public stuff into an `internal` crate makes that easy, but it might end up feeling a little forced.

For a crate like this, I like `internal` as a refactoring mechanism, basically push all the existing stuff in there and then we can incrementally clean it up. For crates built from the start to work with UniFFI, maybe it would be a bit pointless.

What do you think?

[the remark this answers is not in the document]
mhammond: > For a crate like this, I like `internal` as a refactoring mechanism, basically push all the existing stuff in there and then we can incrementally clean it up. For crates built from the start to work with UniFFI, maybe it would be a bit pointless.

It's a good idea in that context, but TBH I don't think it makes sense longer term (ie, it should be temporary) and that rust visibility is the right tool, particularly if fully leveraged (eg, `pub(super)` is probably underused by us).

In particular, I'm mildly concerned that pragmatic compromises might mean moving lots of code from `internal` just to expose it - or worse, marking code in `internal` as `pub`. For example, there are a number of things in places that's only `pub` because `places-utils.rs` wants to use it (and ditto autofill), which is unfortunate, but I tend to suck that up and just allow it to be marked `pub`)

That's weakly held though, particularly as it's not a "battle tested" opinion (although I'm trying to be better at it in autofill)

components/fxa-client/src/internal/commands/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::convert::TryFrom;
+
+pub mod send_tab;
+pub use send_tab::SendTabPayload;
+
+use super::device::Device;
+use super::error::*;
+
+#[derive(Clone, Debug)]
+pub enum IncomingDeviceCommand {
```
mhammond, line 14: hrm - so after writing that comment about visibility, I'm a little confused by seeing `pub` here? Maybe `pub(crate)` better reflects what's going on?

components/fxa-client/src/internal/commands/send_tab.rs
```diff
@@ -37,7 +39,7 @@ impl EncryptedSendTabPayload {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
mhammond: Yes, this makes perfect sense to me and I expect will be a common pattern for our components - eg, what we serialize to/from the sync servers is rarely going to match what we expose over the api.

Re my previous comment - an internal dir that really only had internal structs used for things like deserializing server responses does make more sense, but it makes less sense to see many `impl` blocks there other than the `Into` etc impls

components/fxa-client/src/internal/device.rs
```diff
@@ -2,19 +2,22 @@
...
 
-pub use crate::http_client::{
+use std::{
+    collections::{HashMap, HashSet},
+    convert::TryFrom,
+};
+
+use serde_derive::*;
+
+pub use super::http_client::{
```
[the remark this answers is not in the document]
mhammond: as per above comment, 100% agree

components/fxa-client/src/internal/device.rs
```diff
@@ -295,11 +293,7 @@ impl FirefoxAccount {
     // for the device because the server does not have a `PATCH commands`
     // endpoint yet.
     #[allow(dead_code)]
```
mhammond, line 295: I suspect these `dead_code` annotations are now stale, so we should probably get an issue on file to remove them, as IMO they have value in general. I understand they are often necessary during dev, but generally shouldn't be necessary longer term.

components/fxa-client/src/internal/mod.rs
mhammond: phew :)

components/fxa-client/README.md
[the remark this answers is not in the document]
rfk: Yep, I think I broadly agree with your point here - Rust has good mechanisms to manage what's public API and what's private, and we should lean on them rather than hack around them.

Apart from the contents of `internal/mod.rs`, most of what's in `internal` could quite happily live at the top level of the crate with limited visibility to prevent it from leaking out.

The goal is "don't have a bunch of `pub` stuff that's visible to Rust callers, but not available to foreign-language consumers" and I'm not too fussed about the implementation details. Which actually, that goal is pretty aligned with the recent email thread about Public API surface and mental models for our components...

components/fxa-client/src/lib.rs
mhammond: That's a great aspiration. As mentioned above, the `URL/Guid <-> String` type things might require some uniffi changes, or maybe it already has what we need there - but yeah, not sure how much we should try and insist on that.

examples/fxa-client/src/devices-api.rs
mhammond: yeah - some tension I've found here is probably phrased better as "sometimes we want an example to be more like a diagnotic/test utility than an example of real-world usage" - ie, I explicitly *want* to call internal stuff in my example while not making it public". ie2, sometimes I want my examples to have crate visibility.

mhammond (commented):
This is a massive patch but with even more massive value - I admit I skipped over some of it, so not formally approving, but will happily spend even more time to give it a formal r+ whenever you ask.

rfk:
I think there's ongoing work to be done around managing the visibility of things in this crate, but just to ensure we're on the same page: @mhammond do you think it's worth pursuing more as part of this PR, or doing it in followups over time?

mhammond:
> or doing it in followups over time?

IMO followups is fine (and TBH I'm not too bothered even then - they were more off-the-cuff comments)

rfk:
@mhammond sounds good, thanks! I'll take some time to rebase this and address your comments so far, and then let's please consider this ready for official review. It would be good to get it out into Fenix nightly while we're early in the cycle for maximal bake time.

components/fxa-client/src/internal/device.rs
rfk: I've removed a couple, and put a couple back that appeared to still be necessary.

components/fxa-client/src/internal/commands/mod.rs
rfk: This is one of the things that's currently used by some example crates.

components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
```diff
@@ -0,0 +1,324 @@
...
+
+    public func migrateFromSessionToken(
```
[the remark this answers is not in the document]
rfk: I hope to one day need it for Desktop :-)

rfk:
Just noting for completeness, I've pushed some tweaks and this is ready for formal r?.

rfk:
The failing tests here are the sync integration tests, which are failing due to `502` server errors from the FxA dev server.

components/fxa-client/src/internal/state_persistence.rs
```diff
@@ -190,15 +201,15 @@ impl From<StateV1> for Result<StateV2> {
...
-// `StateV1` was a previous state schema.
-//
-// The below is sufficient to read existing state data serialized in this form, but should not
-// be used to create new data using that schema, so it is deliberately private and deliberately
-// does not derive(Serialize).
-//
-// If you find yourself modifying this code, you're almost certainly creating a potential data-migration
-// problem and should reconsider.
-
+/// `StateV1` was a previous state schema.
```
mhammond, line 204: Does this need to exist still? It looks like it's very old and we will never hit this case in the wild. (OTOH, no need to block - maybe another ticket which could be good for onboarding?)

megazords/ios/MozillaAppServicesTests/FxAccountManagerTests.swift
```diff
@@ -173,7 +173,7 @@ class FxAccountManagerTests: XCTestCase {
     func testAccountRestorationEnsureCapabilitiesNonAuthError() {
         class MockAccount: MockFxAccount {
             override func ensureCapabilities(supportedCapabilities _: [DeviceCapability]) throws {
-                throw FirefoxAccountError.network(message: "The WiFi cable is detached.")
+                throw FxaError.Network(message: "The WiFi cable is detached.")
```
mhammond, line 176: lol @ this error message - I've never seen it before

components/fxa-client/src/fxa_client.udl
```diff
@@ -0,0 +1,1075 @@
...
+// Generic error type thrown by many [`FirefoxAccount`] operations.
+//
+// Precise details of the error are hidden from consumers, mostly due to limitations of
+// how we expose this API to other languages. The type of the error indicates how the
+// calling code should respond.
+//
+//
+//  `Authentication`:
```
mhammond, line 51: Any reason to not put the comments for each error next to the definition?

mhammond (approved):
🚢 this is great, future me thanks you!

components/fxa-client/src/fxa_client.udl
rfk: TBH I don't recall, but the answer is almost certainly "I could not convince the automagical UDL generator to do that for me"...

components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
```diff
@@ -0,0 +1,331 @@
...
+        return try notifyAuthErrors {
+            URL(string: try self.inner.beginPairingFlow(pairingUrl: pairingUrl,
+                                                        scopes: scopes,
+                                                        entrypoint: entrypoint,
+                                                        metrics: metrics))!
+        }
+    }
+
+    public func completeOAuthFlow(code: String, state: String) throws {
+        defer { tryPersistState() }
```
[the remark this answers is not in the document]
rfk: The use of `defer` here means that we'll always do it at the end of the function call, like a weird `finally` block.

components/fxa-client/src/internal/state_persistence.rs
rfk: Sounds good, filed as https://github.com/mozilla/application-services/issues/3912

components/fxa-client/src/fxa_client.udl
rfk: Aha, it was because we currently represent error enum variants in the `ComponentInterface` by just their string names, rather than as a `Variant` object, so there was nowhere for me to stash the docs on each individual variant. I'll fix it up by hand.

Issue #3874, opened 2021-02-17
Tweak the generated bindings for the autofill library


gabrielluong:
In AC, we currently have to do the following:
```
// Set the name of the native library so that we use
// the appservices megazord for compiled code.
System.setProperty(
    "uniffi.component.autofill.libraryOverride",
    System.getProperty("mozilla.appservices.megazord.library", "megazord")
)
```
similar to https://github.com/mozilla-mobile/android-components/blob/ad90b10a294353e5f963b8659fe22ffcd8c9ae2e/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/Nimbus.kt#L215-L218 in order to find the appropriate uniffi_autofill library. See https://mozilla.slack.com/archives/CKL0LGV9B/p1613585308101300 for further discussions.

Since autofill will never be published outside of the megazord, we should preconfigure the generated bindings for the autofill library so that they're always looking for `libmegazord.so`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2061)
┆Story Points: 3
┆Epic: Form data (addresses/credit cards) shared component
┆Sprint: SYNC - 2021-02-26

rfk:
To add a bit of context, the default `.so` name is set by the uniffi tool here: 

* https://github.com/mozilla/uniffi-rs/blob/4d61601bd46a2ae48858ada425c429273833fa90/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt#L2

@lougeniaC64 I'm wondering if the simplest approach here is for uniffi to grow a config option to override the details library name, and for the autofill crate to set that to be `libmegazord.so`. WDYT?

mhammond:
Ryan's proposal sounds fine to me, but as a general question - if our position is that autofill will never be published outside of the megazord, then is it also true that *none* of our components will ever be published outside the megazord, and thus we should remove/deprecate this "override" capability everywhere?

rfk:
Honestly, I've been procrastinating on trying to commit to an answer to that general question, in the hope that additional information will shake out of the better-builds/libxul/etc discussions. For example, it sounds like there's at least one situation in which we'll want to load it from `libxul.so` rather than `libmegazord.so`.

mhammond:
> For example, it sounds like there's at least one situation in which we'll want to load it from `libxul.so` rather than `libmegazord.so`.

[...]

PR #3878, opened 2021-02-20
Update uniffi for autofill

lougeniaC64:
Bumping autofill's uniffi version to 0.7.1 in order to take advantage of changes that allow the uniffi default library to be set to `libmegazord.so` as described in #3874

### DIFF
--- a/components/autofill/Cargo.toml
+++ b/components/autofill/Cargo.toml
@@ -21,7 +21,7 @@ sync15 = { path = "../sync15" }
 sync15-traits = {path = "../support/sync15-traits"}
 thiserror = "1.0"
 types = { path = "../support/types" }
-uniffi = "0.7"
+uniffi = "^0.7.1"
 url = { version = "2.1", features = ["serde"] }
 
 [dependencies.rusqlite]
@@ -35,4 +35,4 @@ libsqlite3-sys = "0.20.1"
 
 [build-dependencies]
 nss_build_common = { path = "../support/rc_crypto/nss/nss_build_common" }
-uniffi_build = "0.7"
+uniffi_build = "^0.7.1"
--- a/components/autofill/uniffi.toml
+++ b/components/autofill/uniffi.toml
@@ -1,2 +1,3 @@
 [bindings.kotlin]
-package_name = "org.mozilla.appservices.autofill"
+package_name = "mozilla.appservices.autofill"
+cdylib_name = "megazord"


lougeniaC64:
Ah, I see nimbus needs to be bumped too . . . until Monday . . .

rfk:
Cross-linking this issue for context, as it has a list of places that need the uniffi version updated:

* https://github.com/mozilla/application-services/issues/3749

rfk:
> Ah, I see nimbus needs to be bumped too . . . 

Bleh. So part of the problem here is that Nimbus ships its own `Cargo.lock` which pins a specific version of uniffi, despite the fact that we never expect anyone to consume artifacts built directly from the Nimbus repo in practice. Rather than bumping its pinned version of uniffi, I think we should *delete* all the stuff in the Nimbus repo that pretends like we might build it in a standalone fashion, meaning it will just use whatever version of uniffi is specified by the appservices megazord build. Ref https://github.com/mozilla/nimbus-sdk/pull/106.

(We'd still have to make an updated Nimbus if this were a *major* version bump for uniffi, but we shouldn't have to do it for ever minor version bump)

lougeniaC64:
> > Ah, I see nimbus needs to be bumped too . . .
> 
> Bleh. So part of the problem here is that Nimbus ships its own `Cargo.lock` which pins a specific version of uniffi, despite the fact that we never expect anyone to consume artifacts built directly from the Nimbus repo in practice. Rather than bumping its pinned version of uniffi, I think we should _delete_ all the stuff in the Nimbus repo that pretends like we might build it in a standalone fashion, meaning it will just use whatever version of uniffi is specified by the appservices megazord build. Ref [mozilla/nimbus-sdk#106](https://github.com/mozilla/nimbus-sdk/pull/106).
> 
> (We'd still have to make an updated Nimbus if this were a _major_ version bump for uniffi, but we shouldn't have to do it for ever minor version bump)

Makes sense, thanks for the explanation!

rfk:
Over in https://github.com/mozilla/application-services/pull/3880 I tried to update Nimbus to a version without `Cargo.lock`, and ended up having to also update uniffi (for reasons I don't really understand). Hopefully it helps for this PR (I think you'll still need the autofill-specific changes at the very least).

lougeniaC64:
> Over in #3880 I tried to update Nimbus to a version without `Cargo.lock`, and ended up having to also update uniffi (for reasons I don't really understand). Hopefully it helps for this PR (I think you'll still need the autofill-specific changes at the very least).

Thanks for the attempt. I'll update this PR tomorrow and hopefully get this to pass CI. 🤞🏾

components/autofill/Cargo.toml
```diff
@@ -21,7 +21,7 @@ sync15 = { path = "../sync15" }
 sync15-traits = {path = "../support/sync15-traits"}
 thiserror = "1.0"
 types = { path = "../support/types" }
-uniffi = "0.7"
+uniffi = "0.7.1"
```
rfk, line 24: nit: I think we should use `^0.7.1` here to allow semver-compatible versions higher than the current release, it will make future uniffi version bumps less painful.

rfk (approved):
👍

PR #3887, opened 2021-02-25
Document breaking API change in autofill module

rfk:
I didn't notice this when it happened in https://github.com/mozilla/application-services/pull/3878, but I think it's a good change for consistency purpose.


mhammond (approved):
huh, I didn't even know that happened :)

Issue #3463, opened 2020-08-05
Consider moving ffi-support to dedicated repo


thomcc:
Given that as of https://github.com/mozilla/application-services/pull/2894 we no longer use this as a `path` dependency (this change was required for vendoring in mozilla-central as we're not the only user of ffi-support there), it seems kind of tricky to make breaking changes to ffi-support — this is part of why I just deprecated things in #3462, rather than removing the old API.

Additionally, it's fairly confusing that we have this in our repo but are *not* using it as a `path` dependency. It also is a special case in that it uses a different license, and is the only piece that's published on crates.io...

Initially it was placed in this repo so that as we evolved our FFI story, we could easily add features, but this has firmed up a lot, so that reasoning doesn't seem to apply as much.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1612)
┆Story Points: 3
┆Epic: Nimbus: build Uniffi library to autogenerate ffi bindings
┆Sprint: SYNC - end 2020-08-14

tarikeshaq:
FWIW I agree to make it a dedicated repo, I remember a while ago the "it's not a path" bit me when I was testing something and tried a local change to `ffi_support`.

Also, `glean` and now `uniffi` both use ffi_support.

rfk:
+1

thomcc:
This is in https://github.com/mozilla/ffi-support/, and once https://bugzilla.mozilla.org/show_bug.cgi?id=1657770 is resolved I'll file a PR removing it from this repo which will close this issue.

---

### How `ffi-support` was extracted while maintaining history

For future reference, these are the steps I used to extract `components/support/ffi` (and `testing/separated/ffi-support-extra-test`) into a separate repo while preserving history, rewriting commit messages referencing issues to refer to this repo, etc. I'm not sure where else to document this, but... It seems bad to not write it down anywhere, so an issue comment is arguably than nothing...

Anyway, the steps:

First, you need to install https://github.com/newren/git-filter-repo (which is what `git filter-branch` tells you to use if you invoke it now).

Move `testing/separated/ffi-support-extra-test` to `components/support/ffi/extra-tests` in commit history
```fish
$ git filter-repo --path-rename testing/separated/ffi-support-extra-test:components/support/ffi/extra-tests
```

Move `ffi-support` to the top level (preserving history).
```fish
$ git filter-repo --subdirectory-filter components/support/ffi
```

[...]

PR #3889, opened 2021-02-25
Add `cargo uniffi-bindgen` alias for running uniffi-bindgen.

rfk:
Whelp, I've had about as many uniffi-version-mismatch-related build failures as I'm willing to tolerate. I think the requirement to have `uniffi-bindgen` installed as a separate command on the user's machine was a great way to get started quickly, but it's making it hard to iterate on our tooling, particularly when trying to co-develop features in uniffi with a consumer over here in application-services.

Inspired by how we handle protobuf files, here's a proposal for an alternative - `cargo uniffi-bindgen` alias that works in all respects like the system-level `uniffi-bindgen` command, but which always uses the version of uniffi that's currently in use by our workspace. I expect this should completely eliminate the possibility of UniFFI version mismatch errors (with the possible exception of failing to clean up files generated by a previous version).

One downside of this change, is that our compile times will get slightly longer, since you have to build `uniffi_bindgen` in the workspace rather than using the one that's already built on your system. I'll take that trade in the interests of reducing failure modes, but YMMV. (I'll add that it shouldn't really affect CI times since we `cargo install uniffi_bindgen` as part of our CI scripts anyway).

This is a draft until we get a release of UniFFI with https://github.com/mozilla/uniffi-rs/pull/406 included, but pushing it early for feedback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ members = [
     "megazords/ios/rust",
     "testing/sync-test",
     "tools/protobuf-gen",
+    "tools/embedded-uniffi-bindgen",
 
     "examples/*/",
     "testing/separated/*/",
@@ -109,6 +110,7 @@ default-members = [
     "components/webext-storage",
     "testing/sync-test",
     "tools/protobuf-gen",
+    "tools/embedded-uniffi-bindgen",
     "examples/*/",
     "testing/separated/*/",
 ]
--- a/components/autofill/android/build.gradle
+++ b/components/autofill/android/build.gradle
@@ -102,9 +102,8 @@ dependencies {
 // We do the uniffi binding generation here:
 android.libraryVariants.all { variant ->
     def t = tasks.register("generate${variant.name.capitalize()}UniffiBindings", Exec) {
-        workingDir "${project.projectDir}"
-        // Runs the bindings generation, note that you must have uniffi-bindgen installed and in your PATH environment variable
-        commandLine 'uniffi-bindgen', 'generate', '../src/autofill.udl', '--language', 'kotlin', '--out-dir', "${buildDir}/generated/source/uniffi/${variant.name}/java"
+        workingDir project.rootDir
+        commandLine 'cargo', 'uniffi-bindgen', 'generate', "${project.projectDir}/../src/autofill.udl", '--language', 'kotlin', '--out-dir', "${buildDir}/generated/source/uniffi/${variant.name}/java"
     }
     variant.javaCompileProvider.get().dependsOn(t)
     def sourceSet = variant.sourceSets.find { it.name == variant.name }
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,6 @@
 [alias]
 regen-protobufs = ["run", "--bin", "protobuf-gen", "tools/protobuf_files.toml"]
+uniffi-bindgen = ["run", "--bin", "embedded-uniffi-bindgen"]
 dev-install = ["install", "asdev"]
 all_tests = ["asdev", "test"]
 rust_tests = ["asdev", "test_rust"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -757,6 +757,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
+name = "embedded-uniffi-bindgen"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "uniffi_bindgen",
+]
+
+[[package]]
 name = "encode_unicode"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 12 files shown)


Cargo.toml
```diff
@@ -40,6 +40,7 @@ members = [
     "megazords/ios/rust",
     "testing/sync-test",
     "tools/protobuf-gen",
+    "tools/uniffi-builtin-bindgen",
```
mhammond, line 43: 'builtin' kinda jarred with me - 'embedded' sounds a bit better, but I'm not sure why as that's pretty much a synonym. I think 'builtin' made me thing 'built in to uniffi'. No strong opinion here, but I thought I'd open a bikeshed.

mhammond (commented):
+1 to the approach!

Cargo.toml
rfk: Fair point. I invested about five seconds worth of effort into the name here, and went with "builtin" entirely by analogy to `features=["builtin-bindgen"]` which you need to add to the config of each crate. I'd be fine with "embedded", and it might also help to re-order to words to something like "embedded-uniffi-bindgen" or "builtin-uniffi-bindgen" so that it's less suggestive of being built in to uniffi.

rfk:
Since this is a non-trivial workflow change, I'd like to get feedback from both current and potential future contributors to this repo; flagging @lougeniaC64 @dmose @travis79 @jhugman for visibility, please take a look and leave any thoughts at your convenience.

rfk:
After https://github.com/mozilla/application-services/pull/3890, I think we need this sooner rather than later, so I've made a new uniffi release and am moving this into "ready for review". (Still want to give everyone a change to weigh in though).

components/autofill/android/build.gradle
```diff
@@ -102,9 +102,8 @@ dependencies {
 // We do the uniffi binding generation here:
 android.libraryVariants.all { variant ->
     def t = tasks.register("generate${variant.name.capitalize()}UniffiBindings", Exec) {
-        workingDir "${project.projectDir}"
-        // Runs the bindings generation, note that you must have uniffi-bindgen installed and in your PATH environment variable
-        commandLine 'uniffi-bindgen', 'generate', '../src/autofill.udl', '--language', 'kotlin', '--out-dir', "${buildDir}/generated/source/uniffi/${variant.name}/java"
+        workingDir project.rootDir
+        commandLine 'cargo', 'uniffi-bindgen', 'generate', "${project.projectDir}/../src/autofill.udl", '--language', 'kotlin', '--out-dir', "${buildDir}/generated/source/uniffi/${variant.name}/java"
```
rfk, line 106: Unfortunately `cargo uniffi-bindgen` doesn't work within a crate directory, only in the workspace root. We can still call it from within gradle in a reasonable fashion though.

rfk:
One alternative here, which @jdragojevic asked me about the other day: we could consider moving `uniffi-rs` development into this repo like we used to do for `ffi-support`. I'm lightly -1 on the idea because I think it might discourage other projects (both internal and external) from using it, but am open to having my mind changed. And I think we'd still want to set up a `cargo uniffi-bindgen` alias for simplicity.

lougeniaC64:
> Since this is a non-trivial workflow change, I'd like to get feedback from both current and potential future contributors to this repo; flagging @lougeniaC64 @dmose @travis79 @jhugman for visibility, please take a look and leave any thoughts at your convenience.

I've had enough experience setting my local version of `uniffi-bindgen` to warmly welcome slightly longer compile times. 🚀

dmose:
Seems high-value to me.  I'm also pretty much perfectly aligned with your analysis on pulling uniffi-rs into this repo.

On a compile-time-related note, using sccache via RUSTC_WRAPPER to build this repo cut the build time after a cargo clean from ~93s to ~49s wall-clock time on my Macbook Pro.

travis79 (approved):
+1 from me on this in general.  I can't count the number of times I've used `cargo install uniffi_bindgen --force` recently.  I'm also leaning away from pulling uniffi-rs into A-S completely, but I suppose it could be set up in such a way as to publish to crates.io and still be easily consumable by outside consumers.

rfk:
> I'm also leaning away from pulling uniffi-rs into A-S completely, but I suppose it could be set up
> in such a way as to publish to crates.io and still be easily consumable by outside consumers.

Indeed. So for a bit of historical context, we used to have the `ffi-support` crate developed in this repo but published to crates.io. This is the bug in which we discussed moving it into a separate repo: https://github.com/mozilla/application-services/issues/3463

I think either approach could work, but having it in a separate repo may be slightly less friction overall.

rfk:
It's not clear to me why Travis's r+ isn't enough to unblock merging here, but if one of @mhammond or @lougeniaC64 could please r+ this as well, that would be awesome :-)

rfk:
> It's not clear to me why Travis's r+ isn't enough to unblock merging here, 

Aha! I added Travis to a team with write access to this repo and that seems to have done the trick.

rfk:
James hasn't weighed in but:

* I'm hearing strong support in the discussion here.
* This prevents an actual footgun that just shot me in the foot yesterday.
* This doesn't require any action from developers after landing.
* It's easy to undo this change if we decide we don't like it.

So I'm going to go ahead and merge.

PR #3880, opened 2021-02-23
Update Nimbus SDK to v0.8.2, uniffi to v0.7.1

rfk:
@lougeniaC64 hopefully this will relieve you of your "need to update the version of uniffi used by nimbus" problem over in https://github.com/mozilla/application-services/pull/3878

PR #3890, opened 2021-02-25
Install correct version of uniffi on taskcluster.

rfk:
Over in https://github.com/mozilla/application-services/pull/3880 I missed one of the many locations in which you need to update the uniffi version, so the `v72.0.0` release is a bust. Trying it out in conjunction with the a-c tests gives me failures like:

```
java.lang.UnsatisfiedLinkError: Error looking up function 'ffi_autofill_7af3_rustbuffer_alloc': dlsym(0x7f83cb0ad490, ffi_autofill_7af3_rustbuffer_alloc): symbol not found
```

It's actually really good that this failed with a linker error, because the bump to uniffi 0.7.1 *did* include a change to how values are passed over the FFI that could have otherwise lead to us silently doing the wrong thing! Still a PITA though.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,10 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v72.0.0...main)
+
+## General
+
+### What's Changed
+
+- This release fixes an "unsatisfied link error" problem with autofill and nimbus,
+  stemming from a misconfiguration in the Android build setup.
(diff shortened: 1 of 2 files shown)


mhammond (approved):

rfk:
I'll also add that I did try out the v72.0.0 release locally, including in a smoketest with android components, before publishing the tag. No amount of loca testing could have caught this because it's a misconfiguration in our android build script in taskcluster. Just in case anyone was on the fence about something like https://github.com/mozilla/application-services/pull/3889 ;-)

PR #3895, opened 2021-03-02
Enable auto-fixing clippy issues and fix some in prep for Rust 1.50 upgrade

dmose:

_No new tests are added because there's no new functionality here.  This is just upgrading the code to more current idioms so that we'll be able to require Rust 1.50 and not have Clippy lose its mind._
  - _No user-visible changes_
    - _There are no new dependencies in this PR; it is preparing for a future PR which will start requiring Rust 1.50_

### DIFF
--- a/automation/all_clippy_checks.sh
+++ b/automation/all_clippy_checks.sh
@@ -17,10 +17,20 @@ fi
 
 EXTRA_ARGS=( "$@" )
 
+# set this in the environment for the script when we want to use the
+# experimental automated clippy fixes from a future version of nightly
+# to ease the pain of doing an upgrade:
+if [[ -z ${CLIPPY_FIX+x} ]]
+then
+    CLIPPY_HEAD="cargo clippy"
+else
+    CLIPPY_HEAD="cargo +nightly clippy --fix -Z unstable-options"
+fi
+
 # Later rust versions downgraded some warning to pedantic, so we allow them here.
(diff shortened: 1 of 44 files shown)


automation/all_clippy_checks.sh
```diff
@@ -17,10 +17,20 @@ fi
...
 EXTRA_ARGS=( "$@" )
 
+# set this in the environment for the script when we want to use the
+# experimental automated clippy fixes from a future version of nightly
+# to ease the pain of doing an upgrade:
+if [[ -z ${CLIPPY_FIX+x} ]]
+then
+    CLIPPY_HEAD="cargo clippy"
+else
+    CLIPPY_HEAD="cargo +nightly clippy --fix -Z unstable-options"
```
rfk, line 27: Neat! I guess there's a chance that this could apply fixes from nightly clippy, that stable clippy disagrees with...but we can grapple with those if/when they come up.

components/support/ffi/src/lib.rs
```diff
@@ -573,7 +573,7 @@ impl Default for ByteBuffer {
     #[inline]
     fn default() -> Self {
         Self {
-            len: 0 as i64,
+            len: 0_i64,
```
rfk, line 576: Oh, huh, I haven't seen this syntax before and I'm not sure I like it...but I do like not having to argue about these things, so will just let clippy to its thing 😅

rfk (approved):
Not sure I like the liberal use of range syntax, but I like it more than I like bikeshedding clippy, so 👍 

Thanks @dmose!

components/support/ffi/src/lib.rs
dmose: Agreed; "You too can write Rust to be just as legible as Perl".  :-)

PR #3882, opened 2021-02-23
Extend logins and places 'counts' telemetry and delete 'timing' telemetry

mhammond:
Data review will be in https://bugzilla.mozilla.org/show_bug.cgi?id=1694316. This is a WIP to see what CI thinks about things.

PR #3896, opened 2021-03-02
Configure detekt to ignore uniffi-generated files.

rfk:
The files generated by uniffi are not the sort of files that
detekt will like and there's not really any point in having
it scan them, so let's ignore anything under `generated` for
all the components.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -253,7 +253,7 @@ ext.cargoExecWithSQLCipher = { spec, toolchain ->
 detekt {
     toolVersion = "1.7.4"
     input = files(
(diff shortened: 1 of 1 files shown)


rfk:
From a little investigation: the CI job that runs `detekt` was doing so on a fresh checkout, so the uniffi-generated files were not present to cause it any problems.

dmose (commented):
r=dmose

rfk:
The `android-build-pr` failure here appears to be due to expiring metrics, which means we'll need https://github.com/mozilla/application-services/pull/3882 in order to get CI passing.

dmose (approved):
r=dmose

PR #3898, opened 2021-03-03
Remove to_payload() and to_record() from the trait

mhammond:
And also rename them to keep clippy happy.

As discussed quickly in slack, this is to make the "encryption" work easier and PR easier to review - the CC impl of these functions needs to manage encryption and addresses do not. Even without that, I think this is still a bit clearer and avoids passing closures etc around.

Once approved and merged, I'll put up a PR for your outgoing PR.

### DIFF
--- a/components/autofill/src/sync/address/incoming.rs
+++ b/components/autofill/src/sync/address/incoming.rs
@@ -8,9 +8,13 @@ use crate::db::models::address::InternalAddress;
 use crate::db::schema::ADDRESS_COMMON_COLS;
 use crate::error::*;
 use crate::sync::common::*;
-use crate::sync::{IncomingState, Payload, ProcessIncomingRecordImpl, ServerTimestamp};
+use crate::sync::{
+    IncomingRecord, IncomingState, LocalRecordInfo, Payload, ProcessIncomingRecordImpl,
+    ServerTimestamp, SyncRecord,
+};
 use interrupt_support::Interruptee;
 use rusqlite::{named_params, Transaction};
+use sql_support::ConnExt;
 use sync_guid::Guid as SyncGuid;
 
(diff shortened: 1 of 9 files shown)


components/autofill/src/sync/credit_card/incoming.rs
```diff
@@ -61,7 +65,66 @@ impl ProcessIncomingRecordImpl for CreditCardsImpl {
...
+                                    Payload::from_json(serde_json::from_str(&m_payload)?)?;
+                                Some(InternalCreditCard::from_payload(payload)?)
+                            }
+                            None => None,
+                        }
+                    },
+                })
+            },
+        )?)
+        //}
```
lougeniaC64, line 126: nit: delete commented code

components/autofill/src/sync/address/incoming.rs
```diff
@@ -68,7 +72,64 @@ impl ProcessIncomingRecordImpl for AddressesImpl {
         LEFT JOIN addresses_data l ON s.guid = l.guid
         LEFT JOIN addresses_tombstones t ON s.guid = t.guid";
 
-        common_fetch_incoming_record_states(tx, sql, |row| Ok(InternalAddress::from_row(row)?))
+        Ok(tx.query_rows_and_then_named(
```
lougeniaC64, line 75: Oh it's sad that this logic has to exist in two places. But I'm imagining keeping this in `common.rs` would have become unruly. 😣

mhammond: Yeah, it's a bt unfortunate, but the duplication makes more sense once the encryption is taken into account.

lougeniaC64 (approved):

PR #2675, opened 2020-02-19
Fix #2673: Add BUILD_LIBRARY_FOR_DISTRIBUTION for swift module stability feature [ci skip]

garvankeeley:
Fixes #2673

This is change to iOS build system, setting flags required for Swift 5.1+ module stability, so that minor differences in Swift compiler versions will be tolerated by Carthage consumers of the library.

The problem is that this requires disabling 'Legacy Build System' in Xcode. We had this enabled to reduce parallelization of the builds as we were getting fatal locked file errors during compilation on the CI builds. My hope is that this has magically gone away in time.

PR #3900, opened 2021-03-03
Experiment: replace protobuf with uniffi

rfk:
I timedboxed a little experiment this afternoon, and it went well enough that I wanted to push it for early feedback (although it's a long way off being landable!)

Our dependency on external protobuf libraries is annoying, and it's particularly troublesome for iOS because iOS consumers need to [configure `SwiftProtoBuf` into their app as an extra dependency](https://github.com/mozilla/application-services/blob/main/docs/howtos/consuming-rust-components-on-ios.md#additional-dependencies). In addition, the presence of `SwiftProtoBuf` means [we can't build our iOS binaries with `BUILD_LIBRARY_FOR_DISTRIBUTION`](https://github.com/mozilla/application-services/pull/2675), which means iOS consumers have to be using the exact same version of Swift as our build infra, which is [regularly](https://github.com/mozilla/application-services/issues/3769) a [problem](https://github.com/mozilla/application-services/issues/3799).

Our use of protobuf seems fairly entrenched, but what if it didn't have to be?

One of the things I learned from #3876 is that it's pretty easy to have hand-written Kotlin and Swift code side-by-side with code generated by UniFFI. So I wanted to see whether UniFFI could replace our current use of protobuf without having to commit to wholesafe uniffication of all our current crates.

TL;DR I think it probably could, but I'm not sure sure if it worth the effort vs doing a full uniffication.

What we have here is the `places` component with protobuf messages removed and uniffi `dictionary` types inserted in their place, on both the Rust side and the Kotlin side (I haven't looked at the swift code yet). It builds! The Kotlin tests don't pass yet though.

I'm going to leave a couple of thoughts on the draft PR here, and I'd love to hear any feedback others might have. But just to be clear: I'm pushing this entirely as a conversation topic, not for any sort of landing at this stage.


components/places/src/api/matcher.rs
```diff
@@ -199,15 +197,15 @@ pub struct SearchResult {
 
     /// The URL to open when the user confirms a match. This is
     /// equivalent to `nsIAutoCompleteResult.getFinalCompleteValueAt`.
-    pub url: Url,
+    pub url: String,
```
rfk, line 200: UniFFI doesn't know about `Url`, so I had to make these `String` in order to expose them to the foreign language. We could have a little `From`/`Into` layer here that converted between the public string-using version and an internal Url-using version if we wanted to.

components/places/src/ffi.rs
```diff
@@ -137,13 +136,38 @@ impl From<Error> for ExternError {
...
-implement_into_ffi_by_protobuf!(msg_types::BookmarkNodeList);
+
+/// Implements [`IntoFfi`] for the provided types (more than one may be passed in) implementing
+/// `uniffi::ViaFfi` (UniFFI auto-generated serialization) by delegating to that implementation.
+///
+/// Note: for this to work, the crate it's called in must depend on `uniffi`.
+///
+/// Note: Each type passed in must implement or derive `uniffi::ViaFfi`.
+#[macro_export]
+macro_rules! implement_into_ffi_by_uniffi {
```
rfk, line 147: This is another ffi-support helper that UniFFI could generate for us pretty easily I think.

components/places/src/places.udl
```diff
@@ -1,31 +1,56 @@
...
-message HistoryVisitInfos {
-    repeated HistoryVisitInfo infos = 1;
-}
+//
+// UniFFI interface definition for the "places" component.
+//
+// The "places" component is still in the process of being converted
+// to UniFFI, so this file doesnt' define the entire interface.
+// For now it only contains the rich datatypes that we want to
+// transfer across the FFI boundary.
```
rfk, line 7: The `.proto` file actually converted into a `.udl` file by hand pretty nicely!

components/places/src/places.udl
```diff
@@ -145,63 +171,48 @@ message BookmarkNode {
...
-     * Returned by reads, and used to distinguish between the cases of
-     * "empty child_nodes because the API doesn't return children" and
-     * "empty child_nodes because this folder has no children (but
-     * we'd populate them if it had them)".
-     *
-     * Only required because you can't have `optional repeated`.
-     *
-     * Leaving this out is equivalent to false.
-     */
-    optional bool have_child_nodes = 11;
```
rfk, line 168: UniFFI can happily represent an `Option<Vec<T>>` without ambiguity, so we don't need this silliness that is only here to work around that fact that protobuf can't. (No slight against protobuf, we're using it in a way that's not exactly aligned with its core value prop).

components/places/src/places.udl
```diff
@@ -145,63 +171,48 @@ message BookmarkNode {
...
-}
-
-/** An array of bookmark nodes, since we can't represent that directly */
-message BookmarkNodeList {
-    repeated BookmarkNode nodes = 1;
-}
```
rfk, line 174: We also don't need these special "holder for a list of T" types because we can just have `sequence<T>`.

components/places/src/places.udl
```diff
@@ -145,63 +171,48 @@ message BookmarkNode {
...
+    Tags(string tags);
+};
+
+dictionary SearchResult {
+    string search_string;
+    string url;
+    string title;
+    string? icon_url;
+    i64 frecency;
+    sequence<MatchReason> reasons;
```
rfk, line 204: Some of the fields on `SearchResult` here not being returned over the FFI, such as `icon_url`. Should they be?

components/places/src/places.udl
```diff
@@ -145,63 +171,48 @@ message BookmarkNode {
...
+};
+
+
+dictionary TopFrecentSiteInfo {
+    string url;
+    string? title;
+};
+
+// This just makes sure we have some extra helper functions available,
+// to be called from our hand-written bindings.
```
rfk, line 214: Now it's UniFFI's turn for some weirdness workarounds!

UniFFI tries not to generate helper code for types that your API doesn't use. We have some functions that want to return e.g. a `sequence<SearchResult>`, so we want UniFFI's helper code for lifting that type. But UniFFI won't generate that helper code unless it sees the type somewhere in the interface. So, we fake it here, at least until the type appears organically in the signature of a uniffied function.

components/places/src/storage/bookmarks.rs
```diff
@@ -28,7 +28,7 @@ use sync_guid::Guid as SyncGuid;
 use types::Timestamp;
 use url::Url;
 
-pub use public_node::PublicNode;
+pub use public_node::InternalBookmarkNode;
```
rfk, line 31: The naming here is wierd, I don't think the `PublicNode` struct ever appears in a public API, does it? It seems more like an internal transition point between the database and the actual public data (previously the message type, now the uniffi dictionary type).

components/places/src/storage/bookmarks/conversions.rs
```diff
@@ -58,41 +57,39 @@ impl From<BookmarkTreeNode> for PublicNode {
     }
 }
 
-impl From<PublicNode> for msg_types::BookmarkNode {
-    fn from(n: PublicNode) -> Self {
-        let have_child_nodes = if n.node_type == BookmarkType::Folder {
-            Some(n.child_nodes.is_some())
-        } else {
-            None
-        };
```
rfk, line 67: `Option<Vec<T>>` FTW

rfk (commented):
Overall, I think this approach would work. But also, it's not clear to me how much effort it would save versus fully converting the FFI over to using UniFFI. Maybe it would be useful if there are some tricky bits in the FFI that UniFFI isn't well set up to handle just yet.

Still, an interesting experiment! And the promise of letting us decouple from the firefox-ios Swift version may be a nice motivator.

components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
```diff
@@ -465,78 +457,69 @@ open class CannotUpdateRoot(msg: String) : PlacesException(msg)
...
                     parentGUID = parentGUID,
-                    position = position,
-                    dateAdded = dateAdded,
-                    lastModified = lastModified,
+                    position = position!!,
+                    dateAdded = dateAdded!!,
+                    lastModified = lastModified!!,
                     title = title,
-                    url = msg.url
+                    url = node.url!!
```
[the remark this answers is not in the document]
mhammond: huh, yeah, how the "old" version works here is beyond me

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -29,6 +28,11 @@ import org.mozilla.appservices.places.GleanMetrics.PlacesManager as PlacesManage
 import mozilla.components.service.glean.private.CounterMetricType
 import mozilla.components.service.glean.private.LabeledMetricType
 
+// Different names between Rust and Kotlin, for...reasons?
+typealias VisitType = VisitTransition
+typealias VisitInfo = HistoryVisitInfo
+typealias VisitInfosWithBound = HistoryVisitInfosWithBound
```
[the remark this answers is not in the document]
mhammond: FWIW, I'm also not sure where these differences come from, but IMO they are imposing a cost we should try and avoid. I'm somewhat use to expecting camelCase etc differences, but these gratuitous naming difference just cause confusion.

components/places/ffi/src/lib.rs
```diff
@@ -10,13 +10,13 @@
 
 use ffi_support::{
     define_box_destructor, define_bytebuffer_destructor, define_handle_map_deleter,
-    define_string_destructor, ByteBuffer, ConcurrentHandleMap, ExternError, FfiStr,
+    define_string_destructor, ConcurrentHandleMap, ExternError, FfiStr,
 };
+use uniffi::ViaFfi;
 use places::error::*;
```
mhammond, line 16: It looks like `type ByteBuffer = uniffi::RustBuffer` would help keep the diff down

components/places/src/api/matcher.rs
mhammond: autofill would (maybe) like this too - there it would be more about 'guid' where we use a guid type internally but strings with uniffi, which jars a little. I haven't looked into how hard that would be though, but just a little +1 for that concept.

components/places/src/places.udl
```diff
@@ -1,31 +1,56 @@
...
+// This is useful when loading lots of history visits via pagination.
+//
+dictionary HistoryVisitInfosWithBound {
+    sequence<HistoryVisitInfo> infos;
+    i64 bound;
+    i64 offset;
+};
+
+
+enum BookmarkType {
```
mhammond, line 48: I notice alot of the diff above is just the case changing of this enum - is there a particular reason you didn't leave the old names?

components/places/src/places.udl
```diff
@@ -40,7 +65,7 @@ message HistoryVisitInfosWithBound {
  * Note that these docs comments are internal, and don't necessarily impact the actual
  * API we expose to Kotlin/Swift (this is particularly true around reads).
  */
-message BookmarkNode {
+dictionary BookmarkNode {
     /**
```
mhammond, line 69: FWIW, autofill has landed on using 2 versions of the struct - one for "incoming" (eg, "create an item" where guid might be null) and one for outgoing (eg "fetch an item" where guid is never null). I assume this is used in both directions which is why everything is nullable.

(Not sure why I'm noting this TBH :)

mhammond (commented):
This is awesome and offers real value in our longer term. I'd also note that if you can resolve the `2` bits, the patch would be a reasonable size.

> TL;DR I think it probably could, but I'm not sure sure if it worth the effort vs doing a full uniffication.

I guess I don't have any understanding of how complimentary those efforts would be, and how much extra work is involved in the "full" unification - but if it's significant, or not wasted work if we went "full" later, I'd say there's enough value here to push forward. Given the benefits are longer term, it might also be a good "onboarding" task? (Purposely ambiguous about what "it" is in that last sentence :)

components/places/src/places.udl
rfk: Right, yes, I think separate "incoming" and "outgoing" would make a big difference here.

components/places/src/places.udl
rfk: Ugh, yeah, UniFFI automagically uppercases enum variant names in the Kotlin bindings, since this seems to be the prevailing style convention for Kotlin code.

rfk:
> but if it's significant, or not wasted work if we went "full" later

I think it would mostly not be wasted work, but a concrete step towards more thorough uniffication.

> I'd say there's enough value here to push forward. Given the benefits are longer term,
> it might also be a good "onboarding" task? (Purposely ambiguous about what "it" is in that last sentence :)

Oooh, I like this idea :-)

components/places/src/api/matcher.rs
mhammond: So it was this chain of comments that distracted me and ended up with me opening https://github.com/mozilla/uniffi-rs/issues/475. The fact that autofill and logins ended up changing guids to strings is somewhat justifiable, but for our places component to start internally using Strings instead of Urls bothers me alot more.

rfk:
There is a much less "work in progress" version of this in https://github.com/mozilla/application-services/pull/4183, so I'm going to close this experimental PR in favour of that actually-working-towards-landing one.

Issue #960, opened 2019-04-11
Migrate away from SQLCipher


eoger:
Implementing an NSS crypto backend is too much work. We should have a plan to move away from SQLCipher.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-729)
┆Epic: Logins uniffi and update
┆Sprint: SYNC - 2021-05-14

thomcc:
My rough plan involves encrypting just the username and password strings in the DB, similar to the way we do on desktop.

The big issue with this is that we still actually need to keep SQLCipher around while we migrate away from it, otherwise users would lose data when they update to the version where we removed sqlcipher. I don't know how best to do this.

rfk:
One of the outcomes from today's program review, is that we should arrange a smaller "architecture review" style meeting to discuss integrating the logins store into desktop. This sounds like a great topic to tackle as part of that meeting. /cc @linuxwolf.

eoger:
Update: we have implemented an NSS backend for sqlcipher (which should be merged upstream at some point in the future).
I'm keeping the issue open as we might have reasons to move away from sqlcipher in the future anyway but feel free to close it if that last assumption is wrong.

lougeniaC64:
@mhammond I've outlined a rough ROUGH draft of the work that I think is required to remove SQLCipher from logins. However having read the comments on this ticket, I'm honestly not sure when we would be able to actually remove the SQLCipher references, but this work below will at least put us on the path to do so.

- Move the [encryption logic](https://github.com/mozilla/application-services/blob/main/components/autofill/src/encryption.rs) from autofill to [the support crate](https://github.com/mozilla/application-services/blob/main/components/support) so that it can be used in logins and autofill
  - Honestly not sure about the uniffi implications but I think if we import the encryption logic from the support crate and expose it as a public module in autofill that _should_ work
- Update the logins schema and structs replacing the `username` and `password` fields with `username_enc` and `password_enc` similar to [autofill’s cc_number_enc field](https://github.com/mozilla/application-services/blob/main/components/autofill/sql/create_shared_schema.sql#L57)
  - Update calling code and tests
- Update [the logins mirror table](https://github.com/mozilla/application-services/blob/main/components/logins/src/schema.rs#L161) so that the fields are an encrypted JSON payload similar to [autofill](https://github.com/mozilla/application-services/blob/main/components/autofill/sql/create_shared_schema.sql#L73)
  - Update calling code and tests
[...]

PR #3903, opened 2021-03-04
Remove sqlcipher linkage hack to appease Rust 1.50 (fixes SYNC-1999)

dmose:
So, the root cause of this problem is a regression between rustc 1.49 and 1.50 ([changelog](https://github.com/mozilla/application-services/blob/c0cc3c7eb6bb18c7ac8c32fef78724e5c520daf3/components/autofill/build.rs#L19)).  I think it's quite likely that rust-lang/cargo#8969 was the breaking PR.  There was a subsequent fix for some piece of the problem in rust-lang/cargo#9065, but that didn't fix the issue that we're seeing, which is still in rustc 1.52 nightly.  (BTW, I (heart) how trivial cargo makes it to test whether a given problem still exists in the nightly version of the compiler).

I haven't tried to nail down exactly what this piece is and filed a bug in rustlang because I don't have a sense of how long that's likely to take.  Furthermore, I think the cheapest way to fix this is simply to unconditionally link NSS in all cases.  It will slow down our build process a bit, but I'd be surprised if it were by an interesting amount.  My assumption is that doing this won't make any difference to the executable produced, since I wouldn't expect anything to be pulled in from NSS unless there are symbols missing at link to.

This is particularly compelling, because, as I understand it, we're considering getting rid of sqlcipher entirely in #960, which would just make the root issue go away.  Is this correct?

Note that I'm forcing everything to 1.50 in this PR to make sure that we're sure we solve the problem we're trying to.

My plan is to remove that commit before we land, just to keep things separate. I could land them all at once, though...

I have yet to test how this fix affects build in m-c, if at all...

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not.
  - _This is really just to fix stuff busted with the Rust 1.50 compiler._
  - _There are no user-facing changes._
  - _No dependency changes_

### DIFF
--- a/components/webext-storage/build.rs
+++ b/components/webext-storage/build.rs
@@ -3,14 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 //! Work around the fact that `sqlcipher` might get enabled by a cargo feature
-//! another crate in the workspace needs, without setting up nss. (This is a
+//! another crate in the workspace needs, without setting up nss.  (This is a
 //! gross hack).
-
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
-    // Ugh. This is really really dumb. We don't care about sqlcipher at all. really
-    if nss_build_common::env_str("DEP_SQLITE3_LINK_TARGET") == Some("sqlcipher".into()) {
-        // If NSS_DIR isn't set, we don't really care, ignore the Err case.
-        let _ = nss_build_common::link_nss();
-    }
+
+    // If NSS_DIR isn't set, we don't really care, ignore the Err case.
+    let _ = nss_build_common::link_nss();
 }
--- a/components/autofill/build.rs
+++ b/components/autofill/build.rs
@@ -11,9 +11,7 @@ fn main() {
     uniffi_build::generate_scaffolding("./src/autofill.udl").unwrap();
 
     println!("cargo:rerun-if-changed=build.rs");
-    // Ugh. This is really really dumb. We don't care about sqlcipher at all. really
-    if nss_build_common::env_str("DEP_SQLITE3_LINK_TARGET") == Some("sqlcipher".into()) {
-        // If NSS_DIR isn't set, we don't really care, ignore the Err case.
-        let _ = nss_build_common::link_nss();
-    }
+
+    // If NSS_DIR isn't set, we don't really care, ignore the Err case.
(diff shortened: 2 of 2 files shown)


dmose:
I assume that the reason these `cargo fmt` changes are required is because `cargo fmt` must have changed something in 1.50, otherwise, I'd expect these changes to be required on `main` as well.  Perhaps I should split them out into a separate commit...

components/webext-storage/build.rs
```diff
@@ -1,16 +1,10 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // If NSS_DIR isn't set, we don't really care, ignore the Err case.
+    let _ = nss_build_common::link_nss();
+}
```
dmose, line 10: I should probably leave the some of sqlcipher commentary here, as well as in the autofill component port, so we know why this hack is here and when we can remove it.

dmose:
I can't actually tell whether it's a problem on beta or nightly.  The reason is that because we seeming to be doing the builds that we use for tests and benchmarking with hard-errors on clippy warnings, the builds on those channels abort before they finish.  Any idea how hard this would be to fix?

dmose:
@mhammond @rfk would love your thoughts on my proposed approach here...

rfk:
> The reason is that because we seeming to be doing the builds that we use for tests and benchmarking with
> hard-errors on clippy warnings, the builds on those channels abort before they finish. Any idea how hard this would be to fix?

By "fix this" do you mean "stop the warnings from failing the build"? The magic sauce there the `-D warnings` in `all_clippy_checks.sh`.

rfk:
> I think the cheapest way to fix this is simply to unconditionally link NSS in all cases

This seems reasonable to me; if "slightly longer builds" is the only downside then it's not worth investing in any fancier, especially given longer-term plans to move away from sqlcipher.

rfk (approved):
r+ for the approach here; I would also support going ahead and changing `1.43` to `1.50` in the version pinning, on the understand that mozilla-central is also pinned to that version.

mhammond (approved):
rs=me!

dmose:
The failing sync tests are just because a live server they depend upon is down.

PR #3904, opened 2021-03-04
Cut release v72.1.1

rfk:
This is a deliberately empty release, to help test some release automation scripts.

### DIFF
--- a/.buildconfig-android.yml
+++ b/.buildconfig-android.yml
@@ -1,4 +1,4 @@
-libraryVersion: 72.1.0
+libraryVersion: 72.1.1
 groupId: org.mozilla.appservices
 projects:
   autofill:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# v72.1.1 (_2021-03-04_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v72.1.0...v72.1.1)
+
+## General
+
+- This is a deliberately empty release, to help test some release automation scripts.
+
+
 # v72.1.0 (_2021-02-25_)
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v72.0.0...v72.1.0)
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -2,4 +2,4 @@
 
 # Unreleased Changes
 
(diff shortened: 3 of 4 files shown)


mhammond (approved):

rfk:
"*How on earth does a deliberately-empty release fail CI???*", he shouts into the void...

rfk:
> "How on earth does a deliberately-empty release fail CI???"

To answer my own question, in lolsob detail:

* Nimbus depends on uniffi "0.7", that is, on the latest release in the "0.7.x" series.
* The appservices megazord depends on uniffi "^0.7.1", that is, on any release in the "0.7.x" series that descends from "0.7.1".
* The appservices workspace has a `Cargo.lock` file which pins its release of uniffi specifically to "0.7.1".
* Nimbus does not have a `Cargo.lock` file, and is not in the appservices workspace.

This was all fine and good when we cut the previous release, because `v0.7.1` was the latest release of UniFFI. But now that there is an `v0.7.2` release available, AFAICT here's what happens:

* CI installs uniffi `v0.7.1`, which is the version we expect to use and the version that is pinned in `Cargo.lock`.
* CI tries to build nimbus-sdk, invoking its `build.rs` to generate the uniffi bindings.
* UniFFI executes its [`ensure_version_compatibility`](https://github.com/mozilla/uniffi-rs/blob/56e463fe44286685276d7efd8657ad23d73639f3/uniffi_bindgen/src/lib.rs#L153) function, which calls `cargo metadata` on the target crate to check what version of uniffi it depends on.
* `cargo metadata` executes on the nimbus-sdk `Cargo.toml`. It's not part of a workspace and it doesn't have a `Cargo.lock`, so *when looked at in isolation* it appears to depend on the latest release of the uniffi "0.7.x" series, which is now `v0.7.2`.
* UniFFI correctly refuses to generate bindings for what appears to be an incompatible crate.

I think this is a bug in UniFFI, inasmuch as it isn't checking the resolved version of the `uniffi` dependency that would actually be built. The problem will go away when we merge nimbus-sdk into this repo and make it part of the workspace.

rfk:
I've filed https://github.com/mozilla/uniffi-rs/issues/413 to follow up on the behaviour of UniFFI here.

rfk:
@mhammond thanks for the view, but I'm not willing to invest the effort to fix up uniffi build issues here just for the purposes of a little experiment with build automation, let's try again atop a future release.

PR #3906, opened 2021-03-04
Encrypt credit cards numbers in the local database

mhammond:
This has everything we should need to make this work except:

* The ability to pass a key in via the sync manager.
* Actual crypto :)

Actual crypto shouldn't change the shape of this PR, nor the shape of the public API exposed. So the intention here is for me to get feedback on this shape while I work on actual crypto (which might be a bit of a can of worms, as it seems app-services has most, but not all, of what we need here)

Feel free to bikeshed - eg:

* `cc_number_enc` doesn't seem like a great field name, but I somehow like both `cc_number_encrypted` and `encrypted_cc_number` less - but I'm not going to die on any of these hills.
* Not sure passing an `EncryptorDecryptor` struct around has more value than just passing the key around. I initially started working with traits so we could have "pluggable" crypto, and it kinda makes sense in that world, but I don't think we actually live in that world, so moved back to a concrete type. OTOH, I don't *hate* the struct and means a more complex key type (eg, a JWT type) is easier to plug in.

No huge urgency - the sooner the better, but please don't drop anything just for this.


components/autofill/sql/create_shared_schema.sql
```diff
@@ -46,7 +46,12 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
 CREATE TABLE IF NOT EXISTS credit_cards_data (
     guid                TEXT NOT NULL PRIMARY KEY CHECK(length(guid) != 0),
     cc_name             TEXT NOT NULL, -- full name
-    cc_number           TEXT NOT NULL, -- TODO: consider storing this field as a hash
+    -- Encrypted card number. While this length is variable it must be > 19 chars,
```
mhammond, line 49: I should fix this comment: what I meant to say is "a real CC number is 19 chars or less, but an encrypted one will always end up being > 19 chars"

components/autofill/src/encryption.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    // to be replaced with real crypto!
+    pub(crate) fn encrypt(&self, cleartext: &str) -> Result<String> {
+        Ok(self.key.clone() + cleartext) // this is secure, right? :)
+    }
+    pub(crate) fn decrypt(&self, ciphertext: &str) -> Result<String> {
+        let mut s = ciphertext.to_string();
+        let cleartext = s.split_off(self.key.len());
+        if s != self.key {
+            Err(Error::InvalidEncryptionKey)
```
rfk, line 57: Heh, if this were real encryption, you wouldn't be able to distinguish between "wrong key" or "corrupted ciphertext", so a more generic error like `DecryptionFailure` may be more appropriate here.

components/autofill/src/sync/engine.rs
```diff
@@ -81,8 +81,10 @@ impl<T: SyncRecord + std::fmt::Debug> SyncEngine for ConfigSyncEngine<T> {
         let timestamp = inbound.timestamp;
         let num_incoming = inbound.changes.len() as u32;
         let tx = db.writer.unchecked_transaction()?;
-        let incoming_impl = self.storage_impl.get_incoming_impl();
-        let outgoing_impl = self.storage_impl.get_outgoing_impl();
+        // XXX - key will need to come in externally somehow.
+        let key = "to-be-passed-in-via-sync-manager?";
```
rfk, line 85: I had this very question - what will the process of configuring sync-manager look like from the consumer's point-of-view, when they need to provide this key? And what does it mean for our efforts to avoid this component holding on to a key in memory for any significant length of time?

rfk (commented):
FWIW, this looks reasonable to me on a first pass. I think the real test will be on the consumer side of the API.

components/autofill/src/sync/engine.rs
mhammond: > And what does it mean for our efforts to avoid this component holding on to a key in memory for any significant length of time?

That's a great question, and the only vague thoughts I have on this are:

* We will expect consumers to pass it in on every sync call - ie, there will not be a `set_key()` type API, but instead be a param to the `sync_now()` api (and I imagine that new param might be, say, an optional, generic set of key/values per engine, and this specific engine will insist on a key and fail if not supplied)
* We will manage the short lifetime in our process by similarly not "storing" it anywhere that lives longer than a single sync - eg, the "engine" in this component will be constructed each sync and isn't a long-lived object.

I really haven't thought about what we can do to ensure the key isn't sitting around in the heap even after it's been "discarded" by us, and haven't really thought about how to insist future refactoring doesn't break these assumptions, but would love ideas.

components/autofill/sql/create_shared_schema.sql
```diff
@@ -46,7 +46,12 @@ CREATE TABLE IF NOT EXISTS addresses_tombstones (
...
+    -- so CHECK that to avoid somehow storing a non-encrypted one.
+    -- XXX - is this a good idea?
+    cc_number_enc       TEXT NOT NULL CHECK(length(cc_number_enc) > 20),
```
lougeniaC64, line 52: I was wondering if rust code shouldn't be handling this with a custom error of some kind but I guess this is just a fail-safe as we expect the consumer to handle validating the number.

components/autofill/src/encryption.rs
```diff
@@ -0,0 +1,107 @@
...
+//
+// So Sync needs to know the key etc, and that needs to get passed down
+// multiple layers, from the app saying "sync now" all the way down to the
+// low level sync code.
+// To make life a little easier, we do that via a struct.
+
+use crate::error::*;
+use jwcrypto::{self};
+
+// XXX - does this struct make sense? We could just pass the string key
```
lougeniaC64, line 34: FWIW I think obfuscating this detail in the struct is better than passing it around.

mhammond:
I think this is ready. I only recently realized the numbers were not encrypted in the mirror, so there's a reasonable chance some of the changes there are uglier than they might otherwise be, so feel free to push back whereever you like in that regard.

Ryan, I'm reflagging you even though I don't expect a full review, but I welcome all comments.

components/autofill/sql/create_shared_schema.sql
```diff
@@ -62,6 +69,7 @@ CREATE TABLE IF NOT EXISTS credit_cards_data (
 
 CREATE TABLE IF NOT EXISTS credit_cards_mirror (
     guid                TEXT NOT NULL PRIMARY KEY CHECK(length(guid) != 0),
+    -- Note that the entire payload here is encrypted in the same way cc_number_enc is locally.
```
rfk, line 72: I initially read "in the same way" as meaning "using the same crypto algorithms", which IIUC is not the case; might be worth clarifying here with "(but using sync15's encryption scheme)" or similar?

rfk (commented):
This all looks very 👍 to me! Holding off on r+ because I really just skimmed the parts that don't have to do with the crypto (although they made sense from a skim!)

lougeniaC64 (approved):
I'm leaning on @rfk's crypto approval a little, but otherwise LGTM!

components/autofill/sql/create_shared_schema.sql
mhammond: No, it does mean "using the same crypto as cc_number_enc"

rfk: Oh. Well, then I guess the comment had the desired effect :-)

mhammond: > Oh. Well, then I guess the comment had the desired effect :-)

Just to be clear, this is OK, right? I did consider using sync's crypto here, but figured that added alot of complication for zero value and that having everything in the local DB use the same schema made more sense.

components/autofill/src/encryption.rs
```diff
@@ -0,0 +1,105 @@
...
+            &ciphertext,
+            jwcrypto::DecryptionParameters::Direct {
+                jwk: self.jwk.clone(),
+            },
+        )?)
+    }
+}
+
+// public functions we expose over the FFI (which is why they take `String`
+// rather than the `&str` you'd otherwise expect)
```
rfk, line 75: If you wanted these to take references, you could try adding a `[ByRef]` annotation in the UDL, but it won't buy you any performance savings or anything in practice.

components/autofill/sql/create_shared_schema.sql
rfk: Yes, seems fine; it may be worth a brief additional explanation about why, along the lines of what you say here. IIUC it's because:

* The mirror needs to have a fully-formed sync BSO payload, which includes the plaintext cc number.
* But we don't want plaintext cc numbers on disk, so we have to encrypt the whole payload for local storage.
* And the JWE encryption scheme is how we encrypt data for local storage, so that's what we use here.

Does that sound right?

rfk (approved):

mhammond:
> Does that sound right?

Exactly - thanks, I'll add those comments.

PR #3901, opened 2021-03-03
Adding Nimbus to the iOS Megazord

jhugman:
This adds Nimbus to the iOS megazord, fixing [SDK-206][0].

It depends on https://github.com/mozilla/uniffi-rs/pull/409, https://github.com/mozilla/application-services/pull/3876 and https://github.com/mozilla/uniffi-rs/issues/412, so we shouldn't consider landing this until both those PRs have landed.

[0]: https://jira.mozilla.com/browse/SDK-206

### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - `automation/all_tests.sh` runs to completion and produces no failures
  - Note: For changes that need extra cross-platform testing, consider adding `[ci full]` to the PR title.
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one
  - Any breaking changes to Swift or Kotlin binding APIs are noted explicitly
- [ ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)
  - Any new dependencies are accompanied by a summary of the due dilligence applied in selecting them.

PR #3913, opened 2021-03-09
Adding Nimbus to the iOS Megazord

rfk:
This is @jhugman's commit from #3901, re-opened and rebased atop the latest changes from #3876. Fixes https://jira.mozilla.com/browse/SDK-206.

I thought I knew how to successfully merge and land #3901, and optimistically merged it intending to do so. But I was wrong. I've disentangled it from the fxa-client uniffication PR in #3876 and I think we need a couple of additional changes in Nimbus before it can land.

It's late here so I'll add more context tomorrow, but briefly: as [CI on this PR shows](https://app.circleci.com/pipelines/github/mozilla/application-services/25999/workflows/49a55611-8a59-4bfe-b926-2b1e5dbac3d4/jobs/97712), when we try to build Nimbus as part of the iOS megazord, we start getting swift compilation failures *from glean*, such as:

```
❌  /Users/rfk/repos/mozilla/application-services/components/external/glean/glean-core/ios/Glean/Metrics/BooleanMetric.swift:92:19: thrown expression type 'String' does not conform to 'Error'

            throw "Missing value"
```

Weird, right? What I think is happening is broadly:

* Nimbus declares an error enum named [`Error`](https://github.com/mozilla/nimbus-sdk/blob/d13021e020d3a1bf3017e4de9fff1ef73837cf82/nimbus/src/nimbus.udl#L65).
* In the Swift bindings this turns into a Swift enum called `Error`, which shadows the built-in [`Error` protocol](https://developer.apple.com/documentation/swift/error).
* When compiling the iOS megazord all the swift code for all components seems to end up in a single namespace, as a single module named `MozillaAppServices`.
* The Glean Swift code thus sees the `Error` enum defined by the Nimbus SDK rather than seeing the default Swift `Error` protocol, and this interferes with the Glean Swift code's attempts to throw errors.

Short-term, I think we can work around this issue by having Nimbus rename its error class to something like `NimbusError`.

Longer-term, we need to think a bit more about namespacing and isolation between components compiled together into an iOS megazord.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -2,6 +2,15 @@
 
 # Unreleased Changes
 
+[Full Changelog](https://github.com/mozilla/application-services/compare/v72.1.0...main)
+
+## General 
+
+- The bundled version of the Nimbus SDK has been updated to v0.9.0.
+- The top-level Rust workspace now builds with Rust 1.50
+- The top-level Rust workspace is now stapled to Rust 1.50, so all developers
+  will build with 1.50, as will the continuous integration for this repo.
+
 ## FxA Client
 
 ### ⚠️ Breaking changes ⚠️
@@ -10,9 +19,3 @@
   As a result many small details of the API surface have changed, such as some
   classes changing names to be consistent between Rust, Kotlin and Swift.
   ([#3876](https://github.com/mozilla/application-services/pull/3876))
-
-[Full Changelog](https://github.com/mozilla/application-services/compare/v72.1.0...main)
-
-- The top-level Rust workspace now builds with Rust 1.50
-- The top-level Rust workspace is now stapled to Rust 1.50, so all developers
-  will build with 1.50, as will the continuous integration for this repo.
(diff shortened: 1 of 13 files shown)


rfk:
@jhugman this is your code so it's a bit weird to flag you for review on the PR here, but I'm going it as a way to communicate a broad "let's the two of us iterate on this together" intention. I hope the above description makes sense in terms of next steps.

badboy:
Looks like one can [use `Swift.Error`](https://github.com/mozilla/glean/blob/main/glean-core/ios/Glean/Utils/Sysctl.swift#L34) anyway to make it more explicit. We could fix that on the Glean side.

jhugman:
> Looks like one can use Swift.Error anyway to make it more explicit. We could fix that on the Glean side.

I think long term, I think we should adopt a practice of declaring our own errors as `MyError` rather than `Error` to avoid this sort of thing. I've update `Error` to `NimbusError`

jhugman:
Awaiting landing https://github.com/mozilla/nimbus-sdk/pull/111 to update the gitmodule hash.

rfk:
Then nimbus fix has landed and is released as `v0.9.0`. I've updated this PR to use it, let's see what CI thinks...

mhammond (approved):
rs=me

rfk:
CI is green \o/

I'm going to go ahead and merge this and work on an appservices release.

PR #3918, opened 2021-03-10
Disable building win32 release artifacts. [ci full]

rfk:
We are hitting some build errors on win32 with the latest release.
I think this may be due to the upgrade to Rust 1.50, and will see
what I can do to resolve it. But in the meantime we need to be
able to cut releases, and disabling the win32 artifacts seems like
the shortest path to that outcome.

In practise, this will mean that Android developers will not be
able to run unittests that use Application Services code if they
are on a Windows machine, since they won't have an appropriate
`.so` file from which to load the appservices code.

It's my expectation that adding `[ci full]` to the PR title here
will cause taskcluster to execute the jobs that failed in the release,
and I will manually confirm that before landing.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,9 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v73.0.0...main)
+
+# Android
+
+- The `-forUnitTest` build no longer includes code compiled for Windows, meaning that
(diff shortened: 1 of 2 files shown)


rfk:
> It's my expectation that adding [ci full] to the PR title here
> will cause taskcluster to execute the jobs that failed in the release,
> and I will manually confirm that before landing.

Yep, they do appear to be executing, see e.g. the `module-build-full-megazord` task.

mhammond (approved):

PR #3922, opened 2021-03-10
Add Direct Encryption with a Shared Symmetric Key support to jwcrypto.

mhammond:
A WIP to see if the shape feels right

Ryan, does this look like it's on the correct path?

### DIFF
--- a/components/support/jwcrypto/src/aes.rs
+++ b/components/support/jwcrypto/src/aes.rs
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Wrappers we use around rc_crypto's AES implementation. Specifically,
+// "enc=A256GCM" from RFC7518, Section 4.7 - for all the gory details, see
+// https://tools.ietf.org/html/rfc7518#section-4.7.
+
+use crate::{
+    error::{JwCryptoError, Result},
+    CompactJwe, EncryptionAlgorithm, JweHeader,
+};
+use rc_crypto::{aead, rand};
+
+/// Does the AES-encrypt heavy-lifting for the schemes supported by this crate.
+pub(crate) fn aes_gcm_encrypt(
+    data: &[u8],
+    protected_header: JweHeader,
+    content_encryption_key: &[u8],
+) -> Result<CompactJwe> {
+    assert_eq!(protected_header.enc, EncryptionAlgorithm::A256GCM);
+    let sealing_key = aead::SealingKey::new(&aead::AES_256_GCM, content_encryption_key)?;
+    let additional_data = serde_json::to_string(&protected_header)?;
+    let additional_data =
+        base64::encode_config(additional_data.as_bytes(), base64::URL_SAFE_NO_PAD);
+    let additional_data = additional_data.as_bytes();
+    let aad = aead::Aad::from(additional_data);
+    // Note that RFC7518 specifies an IV of 96 bits == 12 bytes - which means
+    // that a random IV generally isn't safe with AESGCM due to the risk of
+    // collisions in this many bits. However, for the use-cases supported by
+    // this crate, the keys are either ephemeral, or the number of encryptions
+    // for the same key is expected to be low enough to not collide in
+    // practice.
+    let mut iv: Vec<u8> = vec![0; 12];
+    rand::fill(&mut iv)?;
+    let nonce = aead::Nonce::try_assume_unique_for_key(&aead::AES_256_GCM, &iv)?;
+    let mut encrypted = aead::seal(&sealing_key, nonce, aad, data)?;
+
+    let tag_idx = encrypted.len() - aead::AES_256_GCM.tag_len();
+    let auth_tag = encrypted.split_off(tag_idx);
+    let ciphertext = encrypted;
+
+    CompactJwe::new(
+        Some(protected_header),
+        None,
+        Some(iv),
+        ciphertext,
+        Some(auth_tag),
+    )
+}
+
+/// Does the AES-decrypt heavy-lifting for the schemes supported by this crate
+pub(crate) fn aes_gcm_decrypt(jwe: &CompactJwe, content_encryption_key: &[u8]) -> Result<String> {
+    let protected_header = jwe
+        .protected_header()?
+        .ok_or(JwCryptoError::IllegalState("missing protected_header"))?;
+    assert_eq!(protected_header.enc, EncryptionAlgorithm::A256GCM);
+    let auth_tag = jwe
+        .auth_tag()?
+        .ok_or(JwCryptoError::IllegalState("auth_tag must be present."))?;
+    if auth_tag.len() != aead::AES_256_GCM.tag_len() {
+        return Err(JwCryptoError::IllegalState(
+            "The auth tag length is incorrect",
+        ));
+    }
+    let iv = jwe
+        .iv()?
+        .ok_or(JwCryptoError::IllegalState("iv must be present."))?;
+    let opening_key = aead::OpeningKey::new(&aead::AES_256_GCM, content_encryption_key)?;
+    let ciphertext_and_tag: Vec<u8> = [jwe.ciphertext()?, auth_tag].concat();
+    let nonce = aead::Nonce::try_assume_unique_for_key(&aead::AES_256_GCM, &iv)?;
+    let aad = aead::Aad::from(jwe.protected_header_raw().as_bytes());
+    let plaintext = aead::open(&opening_key, nonce, aad, &ciphertext_and_tag)?;
+    Ok(String::from_utf8(plaintext.to_vec())?)
+}
--- a/components/support/jwcrypto/src/direct.rs
+++ b/components/support/jwcrypto/src/direct.rs
@@ -1,0 +1,257 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Support for "Direct Encryption with a Shared Symmetric Key"
+//! See https://tools.ietf.org/html/rfc7518#section-4.5 for all the details.
+
+use crate::{
+    aes,
+    error::{JwCryptoError, Result},
+    Algorithm, CompactJwe, EncryptionAlgorithm, JweHeader, Jwk, JwkKeyParameters,
+};
+use rc_crypto::rand;
+
+impl Jwk {
+    /// Create a new random key suitable for `Direct` symmetric encryption.
+    /// Consumers can store this for later use, probably via serde serialization.
+    pub fn new_direct_key(kid: Option<String>) -> Result<Self> {
+        // We only support AES256 which has a 32byte key.
+        let mut bytes: Vec<u8> = vec![0; 32];
+        rand::fill(&mut bytes)?;
+        Ok(Jwk {
+            kid,
+            key_parameters: JwkKeyParameters::Direct {
+                k: base64::encode_config(&bytes, base64::URL_SAFE_NO_PAD),
+            },
+        })
+    }
+
+    // Create a new Jwk given the raw bytes of a `Direct` key. We generally
+    // prefer consumers to use serde with the entire key and the only use-case
+    // is currently tests, so let's limit this to tests for now...
+    #[cfg(test)]
+    pub fn new_direct_from_bytes(kid: Option<String>, key: &[u8]) -> Self {
+        Jwk {
+            kid,
+            key_parameters: JwkKeyParameters::Direct {
+                k: base64::encode_config(&key, base64::URL_SAFE_NO_PAD),
+            },
+        }
+    }
+}
+
+pub(crate) fn encrypt_to_jwe(
+    data: &[u8],
+    enc: EncryptionAlgorithm,
+    jwk: &Jwk,
+) -> Result<CompactJwe> {
+    // It's slightly unfortunate we need to supply a struct with ECDH specific
+    // values all specified as None, but doesn't seem likely to ever actually hurt.
+    let protected_header = JweHeader {
+        kid: jwk.kid.clone(),
+        alg: Algorithm::Direct,
+        enc,
+        epk: None,
+        apu: None,
+        apv: None,
+    };
+    let secret = match &jwk.key_parameters {
+        JwkKeyParameters::Direct { k } => base64::decode_config(&k, base64::URL_SAFE_NO_PAD)?,
+        _ => return Err(JwCryptoError::IllegalState("Not a Direct key")),
+    };
+    aes::aes_gcm_encrypt(data, protected_header, &secret)
+}
+
+pub(crate) fn decrypt_jwe(jwe: &CompactJwe, jwk: Jwk) -> Result<String> {
+    let secret = match jwk.key_parameters {
+        JwkKeyParameters::Direct { k } => base64::decode_config(&k, base64::URL_SAFE_NO_PAD)?,
+        _ => return Err(JwCryptoError::IllegalState("Not a Direct key")),
+    };
+    // `alg="dir"` mandates no encrypted key.
+    if jwe.encrypted_key()?.is_some() {
+        return Err(JwCryptoError::IllegalState(
+            "The Encrypted Key must be empty.",
+        ));
+    }
+    aes::aes_gcm_decrypt(jwe, &secret)
+}
+
+#[test]
+fn test_simple_roundtrip() {
+    // We should be able to round-trip data.
+    use super::{decrypt_jwe, encrypt_to_jwe, DecryptionParameters, EncryptionParameters};
+    let jwk = Jwk::new_direct_key(Some("my key".to_string())).unwrap();
+    let data = "to be, or not 🐝🐝";
+    let encrypted = encrypt_to_jwe(
+        data.as_bytes(),
+        EncryptionParameters::Direct {
+            jwk: &jwk,
+            enc: EncryptionAlgorithm::A256GCM,
+        },
+    )
+    .unwrap();
+    let decrypted = decrypt_jwe(&encrypted, DecryptionParameters::Direct { jwk }).unwrap();
+    assert_eq!(data, decrypted);
+}
+
+#[test]
+fn test_modified_ciphertext() {
+    // Modifying the ciphertext will fail.
+    use super::{decrypt_jwe, encrypt_to_jwe, DecryptionParameters, EncryptionParameters};
+    use std::str::FromStr;
+    let jwk = Jwk::new_direct_key(Some("my key".to_string())).unwrap();
+    let data = "to be, or not 🐝🐝";
+    let encrypted = encrypt_to_jwe(
+        data.as_bytes(),
+        EncryptionParameters::Direct {
+            jwk: &jwk,
+            enc: EncryptionAlgorithm::A256GCM,
+        },
+    )
+    .unwrap();
+    // additional text
+    assert!(matches!(
+        decrypt_jwe(
+            &(encrypted.clone() + "A"),
+            DecryptionParameters::Direct { jwk: jwk.clone() }
+        ),
+        Err(JwCryptoError::IllegalState(_))
+    ));
+    // truncated text
+    assert!(matches!(
+        decrypt_jwe(
+            &(encrypted[0..encrypted.len() - 2]),
+            DecryptionParameters::Direct { jwk: jwk.clone() }
+        ),
+        Err(JwCryptoError::IllegalState(_))
+    ));
+    // modified ciphertext - to make this test meaningful we need to
+    // reconsitute the CompactJwe and modify that, otherwise we are just going
+    // to get a base64 or json error.
+    let jwe = CompactJwe::from_str(&encrypted).unwrap();
+    let mut new_ciphertext = jwe.ciphertext().unwrap();
+    new_ciphertext[0] += 1;
+    let jwe_modified = CompactJwe::new(
+        jwe.protected_header().unwrap(),
+        jwe.encrypted_key().unwrap(),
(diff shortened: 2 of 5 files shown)


components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,41 @@
...
+};
+
+pub(crate) fn encrypt_to_jwe(data: &[u8], jwk: &Jwk) -> Result<CompactJwe> {
+    let protected_header = JweHeader {
+        kid: jwk.kid.clone(),
+        alg: Algorithm::Direct,
+        enc: EncryptionAlgorithm::A256GCM, // ??
+        epk: None,
+        apu: None,
+        apv: None,
```
rfk, line 18: It's unfortunate that you have to specify these ECDH-specific params as `None` here, but I can imagine it would be a bit of a refactor to move them into a ECDH-specific sub-struct or whatever.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,41 @@
...
+    aes_gcm_decrypt, aes_gcm_encrypt,
+    error::{JwCryptoError, Result},
+    Algorithm, CompactJwe, EncryptionAlgorithm, JweHeader, Jwk, JwkKeyParameters,
+};
+
+pub(crate) fn encrypt_to_jwe(data: &[u8], jwk: &Jwk) -> Result<CompactJwe> {
+    let protected_header = JweHeader {
+        kid: jwk.kid.clone(),
+        alg: Algorithm::Direct,
+        enc: EncryptionAlgorithm::A256GCM, // ??
```
rfk, line 15: Re: the question mark, yes, this seems right to me - you're using "direct mode" for the content-encryption key and "a256GCM" for then using that key to encrypt the contents.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub(crate) fn decrypt_jwe(jwe: &CompactJwe, jwk: Jwk) -> Result<String> {
+    let protected_header = JweHeader {
+        kid: jwk.kid.clone(),
+        alg: Algorithm::Direct,
+        enc: EncryptionAlgorithm::A256GCM, // ??
+        epk: None,
+        apu: None,
+        apv: None,
+    };
```
rfk, line 35: I'm surprised that you need to create a `protected_header` here, shouldn't the header be part of the jwe that you are trying to decrypt? This might be some factoring weirdness from the ECDH implementation that we need to clean up.

components/support/jwcrypto/src/ec.rs
```diff
@@ -23,12 +22,15 @@ pub struct ECKeysParameters {
...
+    enc: EncryptionAlgorithm,
+    peer_jwk: &Jwk,
 ) -> Result<CompactJwe> {
-    let EncryptionParameters::ECDH_ES { enc, peer_jwk } = encryption_params;
     let local_key_pair = EphemeralKeyPair::generate(&agreement::ECDH_P256)?;
     let local_public_key = extract_pub_key_jwk(&local_key_pair)?;
-    let JwkKeyParameters::EC(ref ec_key_params) = peer_jwk.key_parameters;
+    let ec_key_params = match peer_jwk.key_parameters {
+        JwkKeyParameters::EC(ref params) => params,
+        JwkKeyParameters::Direct(_) => return Err(JwCryptoError::IllegalState("Not an EC key")),
```
rfk, line 32: nit: if you use a default `_` case (or maybe an `if let`?) here you won't have to update this again later when we add more key types.

components/support/jwcrypto/src/direct.rs
rfk: Looking at the ECDH version, it seems to get the protected header via `jwe.protected_header()?` rather than constructing a new one.

components/support/jwcrypto/src/lib.rs
```diff
@@ -176,17 +196,85 @@ impl ToString for CompactJwe {
...
+) -> Result<String> {
+    // Part 2: decrypt the payload
+    if jwe.encrypted_key()?.is_some() {
+        return Err(JwCryptoError::IllegalState(
+            "The Encrypted Key must be empty.",
+        ));
+    }
+    let encryption_algorithm = match protected_header.enc {
+        EncryptionAlgorithm::A256GCM => &aead::AES_256_GCM,
+    };
```
rfk, line 260: Ah, I see. So IMHO this line should be reading the protected header from the provided `jwe` rather than accepting it as a separate parameter. Accepting it separately only creates opportunities for calling code to get confused and provide the wrong header and open itself up to cryptographic mischief. I think you should remove the `protected_header` parameter from this function entirely.

rfk (commented):
> Ryan, does this look like it's on the correct path?

Yep, it looks pretty spot-on! Some notes below.

components/support/jwcrypto/src/lib.rs
```diff
@@ -2,42 +2,67 @@
...
 pub enum EncryptionParameters<'a> {
     // ECDH-ES in Direct Key Agreement mode.
     #[allow(non_camel_case_types)]
     ECDH_ES {
         enc: EncryptionAlgorithm,
         peer_jwk: &'a Jwk,
     },
+    // Direct Encryption with a shared symmetric key.
+    Direct {
+        enc: EncryptionAlgorithm,
```
mhammond, line 39: I ended up adding support for the algo here, just for symmetry with the EC params, but happy to remove it if you think it just causes noise. In particular, the constructors for the `Direct` keys do not support this being variable, which is an argument for *something* changing here, but 🤷

mhammond:
Keeping this as draft because I'm not sure I've addressed the "please improve the comments" commitment I made - please feel free to point out where this could be improved.

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,77 @@
...
+
+// Wrappers we use around rc_crypto's AES implementation.
+
+use crate::{
+    error::{JwCryptoError, Result},
+    CompactJwe, EncryptionAlgorithm, JweHeader,
+};
+use rc_crypto::{aead, rand};
+
+/// Does the AES-encrypt heavy-lifting for the schemes supported by this crate.
```
rfk, line 13: Specifically, you are implementing "enc=A256GCM" from RFC7518, Section 4.7. I think it's worth saying that explicitly in the comments somewhere here, and perhaps even linking out to it [here](https://tools.ietf.org/html/rfc7518#section-4.7) for ease of reference.

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,77 @@
...
+    let encryption_algorithm = match protected_header.enc {
+        EncryptionAlgorithm::A256GCM => &aead::AES_256_GCM,
+    };
+    let sealing_key = aead::SealingKey::new(encryption_algorithm, secret)?;
+    let additional_data = serde_json::to_string(&protected_header)?;
+    let additional_data =
+        base64::encode_config(additional_data.as_bytes(), base64::URL_SAFE_NO_PAD);
+    let additional_data = additional_data.as_bytes();
+    let aad = aead::Aad::from(additional_data);
+    let mut iv: Vec<u8> = vec![0; 12];
```
rfk, line 28: Given our slack discussion, I think it's worth a comment that the RFC requires an IV of 96 bits == 12 bytes. See also [this comment in the corresponding desktop implementation](https://searchfox.org/mozilla-central/rev/87a8afd9f57ee4bc542ba0ec3f96a891042b6db7/services/crypto/modules/jwcrypto.jsm#106) that kind of suggests that we should check in with :mt about what we're doing here. The "But we know that..." part of the linked comment does *not* apply in this new use-case :-/

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,77 @@
...
+        Some(iv),
+        ciphertext,
+        Some(auth_tag),
+    )
+}
+
+/// Does the AES-decrypt heavy-lifting for the schemes supported by this crate
+pub(crate) fn aes_gcm_decrypt(
+    jwe: &CompactJwe,
+    encryption_algorithm: &EncryptionAlgorithm,
```
rfk, line 49: Could we read the `enc` parameter from the protected header of `jwe` rather than requiring the caller to pass it in here? Seems like an opportunity to avoid another potential footgun for the caller, where they fail to check that `enc` has the expected value.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Support for "Direct Encryption with a Shared Symmetric Key"
```
rfk, line 5: Similarly, it's probably worth mentioning and linking to [RFC7518 Section 4.5](https://tools.ietf.org/html/rfc7518#section-4.5) here for easy context for the reader.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
...
+    error::{JwCryptoError, Result},
+    Algorithm, CompactJwe, EncryptionAlgorithm, JweHeader, Jwk, JwkKeyParameters,
+};
+use rc_crypto::rand;
+
+impl Jwk {
+    /// Create a new random key suitable for `Direct` symmetric encryption.
+    /// Consumers can store this for later use, probably via serde serialization.
+    pub fn new_symmetric_key(kid: Option<String>) -> Result<Self> {
+        // We only support AES256 which as a 32byte key.
```
rfk, line 17: sp: "as" -> "has"

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
...
+    aes,
+    error::{JwCryptoError, Result},
+    Algorithm, CompactJwe, EncryptionAlgorithm, JweHeader, Jwk, JwkKeyParameters,
+};
+use rc_crypto::rand;
+
+impl Jwk {
+    /// Create a new random key suitable for `Direct` symmetric encryption.
+    /// Consumers can store this for later use, probably via serde serialization.
+    pub fn new_symmetric_key(kid: Option<String>) -> Result<Self> {
```
rfk, line 16: naming nit: `new_direct_key` would avoid introducing a new term for an existing concept.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[cfg(test)]
+    pub fn new_symmetric_from_bytes(kid: Option<String>, key: &[u8]) -> Self {
+        Jwk {
+            kid,
+            key_parameters: JwkKeyParameters::Direct(key.to_vec()),
+        }
+    }
+}
+
+pub(crate) fn encrypt_to_jwe_direct(
```
rfk, line 38: naming nit: `direct` is kind of implied by the name of the module and may not be needed in the name of the function here (although I don't feel strongly about it either way).

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
...
+    };
+    aes::aes_gcm_decrypt(jwe, &EncryptionAlgorithm::A256GCM, &secret)
+}
+
+#[test]
+fn test_simple_roundtrip() {
+    // We should be able to round-trip data.
+    use super::{decrypt_jwe, encrypt_to_jwe, DecryptionParameters, EncryptionParameters};
+    let jwk = Jwk::new_symmetric_key(Some("my key".to_string())).unwrap();
+    let data = "to be, or not 🐝🐝";
```
rfk, line 73: 😆

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,159 @@
...
+    // want to do that until we have evidence it's actually spec compliant.)
+    use super::{decrypt_jwe, DecryptionParameters};
+    let jwk = Jwk::new_symmetric_from_bytes(None, "asecret256bitkeyasecret256bitkey".as_bytes());
+    let ciphertext = "eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0..nhKdQEKqoKPzfCda.rQOj0Nfs6wO5Gj4Quw.CMJFS9YBADLLePdj1sssSg";
+    let decrypted = decrypt_jwe(ciphertext, DecryptionParameters::Direct { jwk }).unwrap();
+    assert_eq!(decrypted, "Hello, World!");
+}
+
+#[test]
+fn test_bad_key_type() {
```
rfk, line 143: It may also be worth adding some tests for things that are supposed to fail, e.g.:

* Try to decrypt with the wrong key
* Try to decrypt after modifying the ciphertext

components/support/jwcrypto/src/ec.rs
```diff
@@ -2,33 +2,39 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! Implements Elliptic-Curve Diffie-Hellman for JWE.
```
rfk, line 5: Since you're here, let's also link out to the relevant RFC sections, so it's clear exactly what we're trying to do. In this case it's [Section 4.6](https://tools.ietf.org/html/rfc7518#section-4.6) and specifically the "direct key agreement" mode.

components/support/jwcrypto/src/lib.rs
```diff
@@ -2,42 +2,67 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-//! Theorically, everything done in this crate could and should be done in a JWT library.
-//! However, none of the existing rust JWT libraries can handle ECDH-ES encryption, and API choices
-//! made by their authors make it difficult to add this feature.
-//! In the past, we chose cjose to do that job, but it added three C dependencies to build and link
-//! against: jansson, openssl and cjose itself.
+//! A library for using Javascript Web Tokens (JWK) and related concepts.
```
rfk, line 5: nit: "Javascript Web Tokens" is "JWT", not "JWK", and actually I don't think we have JWT support in here yet. I would phrase this as "A library for using JSON Object Signing and Encryption (JOSE) data formats such as JWE and JWK" or similar.

components/support/jwcrypto/src/lib.rs
rfk: The ECDH constructors shouldn't let it be variable either because we don't support any other value, but I'm fine with keeping it here, I think it conceptually makes sense.

rfk (commented):
I like how this is coming together.

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,84 @@
...
+        .ok_or(JwCryptoError::IllegalState("missing protected_header"))?;
+    let aead_algo = match protected_header.enc {
+        EncryptionAlgorithm::A256GCM => &aead::AES_256_GCM,
+    };
+    let auth_tag = jwe
+        .auth_tag()?
+        .ok_or(JwCryptoError::IllegalState("auth_tag must be present."))?;
+    if auth_tag.len() != aead_algo.tag_len() {
+        return Err(JwCryptoError::IllegalState(
+            "The auth tag must be 16 bytes long.",
```
rfk, line 72: Extremely minor nit: this error message hard-codes the magic constant `16`.

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,84 @@
...
+    error::{JwCryptoError, Result},
+    CompactJwe, EncryptionAlgorithm, JweHeader,
+};
+use rc_crypto::{aead, rand};
+
+/// Does the AES-encrypt heavy-lifting for the schemes supported by this crate.
+pub(crate) fn aes_gcm_encrypt(
+    data: &[u8],
+    protected_header: JweHeader,
+    secret: &[u8],
```
rfk, line 19: nit: here and below, if we rename `secret` to `content_encryption_key` then I think it will be slightly easier for future readers to mentally map between the spec and the code here.

components/support/jwcrypto/src/aes.rs
```diff
@@ -0,0 +1,84 @@
...
+    )
+}
+
+/// Does the AES-decrypt heavy-lifting for the schemes supported by this crate
+pub(crate) fn aes_gcm_decrypt(jwe: &CompactJwe, secret: &[u8]) -> Result<String> {
+    if jwe.encrypted_key()?.is_some() {
+        return Err(JwCryptoError::IllegalState(
+            "The Encrypted Key must be empty.",
+        ));
+    }
```
rfk, line 60: nit: technically, it's not the `enc` algorithm's job to check that this is empty. Whether or not we expect a value here depends on the `alg` parameter. It is important that we do this check *somewhere*, so we could consider moving it up into the two `alg=ECDH-ES` and `alg=dir` implementations if there's a good place for it, or I'm happy to keep it here for simplicity and add a comment saying "not really the enc algorithm's job to check this but it's here for simplicity".

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,227 @@
...
+            key_parameters: JwkKeyParameters::Direct {
+                k: base64::encode_config(&key, base64::URL_SAFE_NO_PAD),
+            },
+        }
+    }
+}
+
+pub(crate) fn encrypt_to_jwe(
+    data: &[u8],
+    enc: EncryptionAlgorithm,
```
rfk, line 46: Nit: why accept `enc` as an argument here? If we get any other value than `A256GCM` then we're going to create an invalid JWE, so it seems like we may as well just hard-code it to that unless and until we implement other `enc` schemes in future. (Alternately, we could assert that it has the expected value, to give us an obvious place to add support for other schemes in future).

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,227 @@
...
+        _ => return Err(JwCryptoError::IllegalState("Not a Direct key")),
+    };
+    aes::aes_gcm_encrypt(data, protected_header, &secret)
+}
+
+pub(crate) fn decrypt_jwe(jwe: &CompactJwe, jwk: Jwk) -> Result<String> {
+    let secret = match jwk.key_parameters {
+        JwkKeyParameters::Direct { k } => base64::decode_config(&k, base64::URL_SAFE_NO_PAD)?,
+        _ => return Err(JwCryptoError::IllegalState("Not a Direct key")),
+    };
```
rfk, line 70: This is conceptually the layer that knows what to do with the `jwe.encrypted_key()` field.

components/support/jwcrypto/src/direct.rs
```diff
@@ -0,0 +1,227 @@
...
+    .unwrap();
+    // additional text
+    assert!(matches!(
+        decrypt_jwe(
+            &(encrypted.clone() + "A"),
+            DecryptionParameters::Direct { jwk: jwk.clone() }
+        ),
+        Err(JwCryptoError::IllegalState(_))
+    ));
+    // truncated text
```
rfk, line 114: Please also test changing one of the bytes in the text without altering its length (because I expect there are length checks that would fail the add/remove cases you've got here but would allow a modified text through to lower layers of the stack).

components/support/jwcrypto/src/ec.rs
```diff
@@ -2,33 +2,41 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! Implements Elliptic-Curve Diffie-Hellman for JWE - specifically, the
+//! "direct key agreement" mode described in https://tools.ietf.org/html/rfc7518#section-4.6
```
rfk, line 6: nit: "Ephemeral-Static direct key agreement"

components/support/jwcrypto/src/ec.rs
```diff
@@ -37,42 +45,14 @@ pub(crate) fn encrypt_to_jwe(
...
-) -> Result<String> {
-    let DecryptionParameters::ECDH_ES { local_key_pair } = decryption_params;
-
+/// The ECDH helper that takes the ciphertext in the form of a CompactJwe,
+/// and the keys, creates the appropriate header, then calls the `aes` module to
+/// do the actual decrytion.
+pub(crate) fn decrypt_jwe(jwe: &CompactJwe, local_key_pair: EphemeralKeyPair) -> Result<String> {
     let protected_header = jwe.protected_header()?.ok_or(JwCryptoError::IllegalState(
         "protected_header must be present.",
     ))?;
```
rfk, line 58: This `protected_header` check is now done by `aes::aes_gcm_decrypt` so you don't need to repeat it here. (Although if you wanted to check that there was no `jwe.encrypted_key()` field, this would be the other place to do it).

components/support/jwcrypto/src/lib.rs
```diff
@@ -2,42 +2,70 @@
...
-//! made by their authors make it difficult to add this feature.
-//! In the past, we chose cjose to do that job, but it added three C dependencies to build and link
-//! against: jansson, openssl and cjose itself.
+//! A library for using JSON Object Signing and Encryption (JOSE) data formats
+//! such as JWE and JWK, as described in https://tools.ietf.org/html/rfc7518
+//! and related standards.
+//! The encryption is done by [rc_crypto] - this crate just does the JOSE
+//! wrappers around this crypto. As a result, most of the structs etc here
+//! support serialization and deserialization to and from JSON via serde in
+//! a way that's compatibile with rfc7518 etc.
```
rfk, line 11: sp: "compatible".

rfk (approved):
This looks great! I've left some very fine-grained nits with an eye to making sure that the things we're doing here map to the RFC in as straightforward a manner as possible, but I'm not willing to die on any of those hills. Good stuff 👍

components/support/jwcrypto/src/direct.rs
rfk: Concretely, if we're going to accept it as an argument, then we should have something like this to dispatch based on it at the end of the function:

```
match enc {
  EncryptionAlgorithm::A256GCM => aes::aes_gcm_encrypt(data, protected_header, &secret),
  // Other encryption algorithms would be added here in future.
}
```

Issue #3907, opened 2021-03-04
Symbolicate crashes in XCode / iOS


nbhasin2:
We see crashes in XCode in our production app but cannot debug as they aren't symbolicated. Since we have Glean in A~S and will also add Nimbus I feel having symbolicating crashes will help in better debugging when crashes happen.

Ref issue: https://github.com/mozilla-mobile/firefox-ios/issues/7588

![](https://user-images.githubusercontent.com/8919439/97471385-0ac07e00-191f-11eb-8869-3ff54de01dcd.png)

![](https://user-images.githubusercontent.com/8919439/97471456-23309880-191f-11eb-8e94-874219cb83a8.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2604)
┆Epic: Important backlog

rfk:
Unfortunately the current appservices team does not have much experience with iOS crashes or symbolication (or XCode in general to be quite honest) but we're happy to help here if we can. I'm not sure where to start though. @nbhasin2 I recall a slack conversation about `.dsym` files, so I had a quick look at the latest release published to github.

The source file [MozillaAppServices.zip](https://github.com/mozilla/application-services/releases/download/v72.1.0/MozillaAppServices.framework.zip) is, IIUC, what gets pulled in by Carthage when building firefox-ios. It contains both `Carthage/Build/iOS/MozillaAppServices.framework/` which the compiled code, and what looks like a corresponding `Carthage/Build/iOS/MozillaAppServices.framework.dSYM/Contents/Resources/DWARF/MozillaAppServices` which claims to be a "dSYM companion file":

```
[rfk] repos/mozilla-mobile/firefox-ios (git)-[master] % file Carthage/Build/iOS/MozillaAppServices.framework.dSYM/Contents/Resources/DWARF/MozillaAppServices
Carthage/Build/iOS/MozillaAppServices.framework.dSYM/Contents/Resources/DWARF/MozillaAppServices: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dSYM companion file x86_64] [arm64]
Carthage/Build/iOS/MozillaAppServices.framework.dSYM/Contents/Resources/DWARF/MozillaAppServices (for architecture x86_64):     Mach-O 64-bit dSYM companion file x86_64
[...]

PR #3923, opened 2021-03-11
Add a "crashtest" component for helping test crash-related behaviours.

rfk:
This is a definitely-for-development-purposes-only component that exposes
some functions to let you deliberately crash your application in interesting
Rust-code-related ways.

The idea is that application developers could add some developer-only UI that
deliberately triggers one of these functions, and then try out the crash or
error-handling behaviour of their app in a live setting. We hope to be able
to use this to debug symbolication issues on Firefox for iOS, ref #3907.
(/cc @nbhasin2).

As a side benefit, the very process of adding this component has let me write
a couple of nice tests that our generated bindings are working as intend around
the edges of error and panic handling.

### DIFF
--- a/components/crashtest/README.md
+++ b/components/crashtest/README.md
@@ -1,0 +1,82 @@
+# Crash Testing Helper APIs
+
+This is a little helper component to make it easier to debug
+issues with crash reporting, by letting you deliberately
+crash a consuming app.
+
+* [Features](#features)
+* [Using the Crash Testing Helper APIs](#using-the-crash-testing-helper-apis)
+* [Working on Crash Testing Helper APIs](#working-on-the-crash-testing-helper-apis)
+
+## Features
+
+The Crash Testing Helper APIs let you deliberately crash your application
+in a variety of ways:
+
+1. By triggering a hard abort inside the Rust code of the component, which
+   should surface as a crash of the application.
+1. By triggering a panic inside the Rust code of the component, which should
+   surface as an "internal error" exception to the calling code.
+
+The component does not offer any support for crash reporting, debugging etc
+itself, it's just designed to let you more easily test those things in your
+application.
+
+## Using the Crash Testing Helper APIs
+
+### Before using this component
+
+This component does not currently integrate with the [Glean SDK](https://mozilla.github.io/glean/book/index.html)
+and does not submit any telemetry, so you do not need to request a data-review before using this component.
+
+### Prerequisites
+
+To use this component, you should be familiar with how to integrate application-services components
+into an application on your target platform:
+* **Android**: Add the `mozilla.appservices.crashtest` package as a gradle dependency, but make sure
+  you're using the same version of application-services as used by [Android Components](
+  https://github.com/mozilla-mobile/android-components/tree/master/components/service/firefox-accounts/README.md).
+* **iOS**: start with the [guide to consuming rust components on
+  iOS](https://github.com/mozilla/application-services/blob/main/docs/howtos/consuming-rust-components-on-ios.md).
+* **Other Platforms**: we don't know yet; please reach out on slack to discuss!
+
+### Component API
+
+For details on how to use this component, consult the [public interface definition](./src/crashtest.udl)
+or view the generated Rust API documentation by running:
+
+```
+cargo doc --no-deps --open
+```
+
+## Working on the Crash Testing Helper APIs
+
+### Prerequisites
+
+To effectively work on the Crash Testing Helper APIs, you will need to be familiar with:
+
+* Our general [guidelines for contributors](../../docs/contributing.md).
+* The way we use [uniffi-rs](https://github.com/mozilla/uniffi-rs) to generate API wrappers for multiple languages, such as Kotlin and Swift.
+
+### Overview
+
+This component uses [uniffi-rs](https://github.com/mozilla/uniffi-rs) to create its
+public API surface in Rust, and then generate bindings for Kotlin and Swift. The
+code is organized as follows:
+
+* The public API surface is implemented in [`./src/lib.rs`](./src/lib/rs), with matching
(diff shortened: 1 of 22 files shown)


components/crashtest/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Working on the Crash Testing Helper APIs
+
+### Prerequisites
+
+To effectively work on the Crash Testing Helper APIs, you will need to be familiar with:
+
+* Our general [guidelines for contributors](../../docs/contributing.md).
+* The [core concepts](#core-concepts) for users of the component, outlined above.
```
lougeniaC64, line 59: nit: there's is no section titled `#core-concepts`

components/crashtest/src/tests.rs
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// The tests are in a separate file just to ensure that the
+// main `lib.rs` contains nothing but the public interfance.
```
lougeniaC64, line 6: nit: public "interface"

lougeniaC64 (approved):
LGTM, although I'm trusting the megazord code 😬.

components/crashtest/android/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
lougeniaC64, line 1: Newb question: I've seen this file in other components and noticed that it looks like all comments. Out of curiosity is it required? Are we adding this file in order to allow for potential configuration?

components/crashtest/src/crashtest.udl
```diff
@@ -0,0 +1,57 @@
...
+// and crash-reporting capabilities of their app.
+
+namespace crashtest {
+    
+    
+  // Trigger a hard abort inside the Rust code.
+  //
+  // This function simulates some kind of uncatchable illegal operation
+  // performed inside the Rust code. After calling this function you should
+  // expect your application to be halted with e.g. a `SIGABORT` or similar.
```
nbhasin2, line 17: nit
```suggestion
  // expect your application to be halted with e.g. a `SIGABRT` or similar.
```

components/crashtest/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+#[cfg(test)]
+mod tests;
+
+uniffi_macros::include_scaffolding!("crashtest");
+
+/// Trigger a hard abort inside the Rust code.
+///
+/// This function simulates some kind of uncatchable illegal operation
+/// performed inside the Rust code. After calling this function you should
+/// expect your application to be halted with e.g. a `SIGABORT` or similar.
```
nbhasin2, line 22: nit:
```suggestion
/// expect your application to be halted with e.g. a `SIGABRT` or similar.
```

components/crashtest/android/proguard-rules.pro
rfk: That is a great question, and I don't have a good answer TBH. I think it's a convention we inherited from android-components but you're right, they don't seem to actually *do* anything.

rfk:
> LGTM, although I'm trusting the megazord code 😬.

Yeah :-/

This is why I wanted to add at least one test in each of the Kotlin and Swift wrappers, so if it the megazord compiles cleanly and the tests run then it must be basically working as intended.

components/crashtest/src/tests.rs
rfk: My favourite mis-spelling of this is "interfarce" but this one's quite fun too; thanks!

nbhasin2:
@rfk thanks for taking this on, I am a wondering on iOS side once we add this would we just call `triggerRustAbort()` from iOS side or is there anything else we have to do. Sorry for such a silly question 😅

nbhasin2 (approved):

rfk:
> I am a wondering on iOS side once we add this would we just call triggerRustAbort() from iOS side or is there anything
> else we have to do.

Yep, in theory, you should be able to import `triggerRustAbort` etc from the `MozillaAppServices` module, call it, and crash. The signatures you'll see from Swift code are:

```

public func triggerRustAbort() {
 ...
}

public func triggerRustPanic() {
 ...
}

public func triggerRustError() throws {
  ...
}

public enum CrashTestError: LocalizedError {
    case NoError
    case ErrorFromTheRustCode(message: String)
}
```

I should also say a few more words about my motivation for each of the three functions here, for context.

Usually when we think of Rust code "crashing" that would correspond to `triggerRustPanic` - the Rust code has encountered some unexpected situation and is explicitly bailing out. The way we have this set up when compiling for Swift is that a panic unwinds the call stack, we intercept it at the FFI boundary between Rust and Swift, and turn it into a Swift `Error`. The Swift wrapper code uses `try!` for the these errors when it calls into the Rust code, so the result should ultimately surface as Swift bailing out with a runtime error.

However, maybe we're not seeing a proper Rust panic, but maybe some invalid native code that is trapping or aborting at a layer below what the Rust language lets us reason about. For example, maybe it's a mis-compilation that produces invalid instructions that only affect certain devices or platforms. I *think* that would be closer to what `triggerRustAbort` is designed to simulate, although maybe not exactly the same.

Finally, it may be the case that we're seeing failures where the Swift code in firefox-ios is not correctly catching or handling application-level errors thrown by the Rust code, e.g. errors thrown on network failure or things like that. That's what `triggerRustError` is here for, to maybe help with testing any uncaught-exception-handling code paths.

Hope that makes sense, don't hesitate to ping me if I can help push this along.

rfk:
(Also, @nbhasin2 let me know if you want this in a release, or if it works to try from a local dev build for now)

PR #3930, opened 2021-03-15
Add Nimbus.swift object

jhugman:
Fixes [SDK-208][1].

Add Nimbus.swift object and tests.

[1]: https://jira.mozilla.com/browse/SDK-208


jhugman:
Pinging @boek and @mozilla-mobile/firefox-ios for review/feedback.

.buildconfig-android.yml
```diff
@@ -86,7 +86,7 @@ projects:
       type: aar
     description: Sync 1.5 remote tabs implementation.
   nimbus:
-    path: components/nimbus
+    path: components/nimbus/android
```
jhugman, line 89: This is just to make way for the iOS code.

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -0,0 +1,141 @@
...
+ * This is the main experiments API, which is exposed through the global [Nimbus] object.
+ *
+ * Application developers are encouraged to build against this API protocol, and use the `Nimbus.create` method
+ * to create the correct implementation for them.
+ *
+ * Feature developers configuring their features shoiuld use the methods in `NimbusFeatureConfiguration`. These are safe to call from any thread.
+ * Developers building UI tools for the user or QA to modify experiment enrollment will mostly use `NimbusUserConfiguration` methods.
+ * Application developers integrating `Nimbus` into their app should use the methods in `NimbusStartup`.
+ */
+public protocol NimbusApi: NimbusStartup, NimbusFeatureConfiguration, NimbusUserConfiguration {}
```
jhugman, line 17: I like being able to split up the object into named interfaces.

I do not like the name `NimbusApi`. I'd prefer to have the protocol as `Nimbus`, and the implementations be named according to their utility (`NimbusEnabled` and `NimbusDisabled`).

I haven't renamed from `NimbusApi` because it is implemented in Android with this name.

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+private let EXPERIMENT_BUCKET_NAME = "main"
+private let EXPERIMENT_COLLECTION_NAME = "nimbus-mobile-experiments"
```
jhugman, line 8: Swift lint doesn't like these names.

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -0,0 +1,86 @@
...
+                          appVersion: info["CFBundleShortVersionString"] as? String,
+                          appBuild: info["CFBundleVersion"] as? String,
+                          architecture: nil,
+                          deviceManufacturer: Sysctl.manufacturer,
+                          deviceModel: Sysctl.model,
+                          locale: Locale.current.identifier,
+                          os: device.systemName,
+                          osVersion: device.systemVersion,
+                          androidSdkVersion: nil,
+                          debugTag: "Nimbus.rs")
```
jhugman, line 84: Attention iOS people and Glean people.

Can we get the architecture easily? Why is `androidSdkVersion` so named?

travis79: It is so named because it is an artifact of telemetry embedded in Glean that probably shouldn't be inside of Glean that is Android specific. It cargo culted its way into Nimbus from the initial copying of some of the Mako design. It should probably be removed as I don't recall it ever being used in Mako, or it could be modified to be the Nimbus SDK version or something worth targeting on, if we even need it.

travis79: Here is a link to where Glean is fetching much of this same info: https://github.com/mozilla/glean/blob/bfa108f2b8d81e8cc7e19b05421baab07301c8d5/glean-core/ios/Glean/Glean.swift#L211

travis79: `architechture` being `Sysctl.machine` and there is some massaging of the Locale going on to get it into `EN-US, EN-GB, DE-DU, etc...` format that Glean does.

components/nimbus/ios/Nimbus/NimbusCreate.swift
travis79: I would recommend following the Swift naming convention for constants (explained well here: https://github.com/raywenderlich/swift-style-guide#constants)

jhugman: More of a note-to-self than a cry-for-help 😃

.buildconfig-android.yml
rfk: Just flagging that this is going to make conflicts with @dmose's work over in https://github.com/mozilla/application-services/pull/3932, I wonder if it's worth pulling this specific change out into a separate PR that we can land ASAP and both rebase on top of?

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -0,0 +1,229 @@
...
+    }
+}
+
+// Glean integration
+private extension Nimbus {
+    func recordExposure(experimentId _: String) {
+        // TODO: https://jira.mozilla.com/browse/SDK-209
+    }
+
+    func postEnrolmentCalculation(_ events: [EnrollmentChangeEvent]?) {
```
rfk, line 45: I was about to complain about 'enrolment' vs 'enrollment' here but then I noticed that nimbus-sdk already has a heady mix of both spellings :-/

I think all of the names visible in our API layer on Android use 'enrollment', FWIW.

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
```diff
@@ -279,9 +287,12 @@
 			children = (
 				394A807525EE951D00FAF26F /* Generated */,
 				394A807325EE94C600FAF26F /* nimbus.udl */,
+				394A809725F94E1300FAF26F /* Nimbus.swift */,
+				395992B725FBE40300E3185F /* NimbusApi.swift */,
+				394D6C612600E64E008F9CE5 /* NimbusCreate.swift */,
 			);
 			name = Nimbus;
-			path = "../../components/external/nimbus-sdk/nimbus";
+			path = ../../components/nimbus/ios/Nimbus;
```
rfk, line 295: Does the path to the `nimbus.udl` child of this group need to be updated as a result of this change? It looks like the file entry for `nimbus.udl` has its path configured to be relative to its containing group, and this changes the path of its containing group.

rfk (commented):
This looks solid to me, apart from one question below about the xcode config. I'm not giving an r+ here because I don't feel qualified enough to comment on the details of the swift code, but I did look it over and it seemed to make sense to me at a high level.

components/nimbus/ios/Nimbus/NimbusCreate.swift
jhugman: This is really helpful, thank you.

components/nimbus/ios/Nimbus/Nimbus.swift
jhugman: This, I suspect is the next front in the ongoing  `en-US` v `en-ROW` _guerre des langues_.

"Enrolment" will now join "Colour" and "Centre" in one of the most poignant `grep`-based ceremonies in Computer Science.

components/nimbus/ios/Nimbus/NimbusCreate.swift
travis79: I get bitten by this one a lot, I too like the all caps constants :)

travis79 (commented):
A few nits/comments, but this is looking good!

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -0,0 +1,141 @@
...
+     * Control the opt out for all experiments at once. This is likely a user action.
+     */
+    var globalUserParticipation: Bool { get set }
+
+    /**
+     * Get the list of currently enrolled experiments
+     *
+     * @return A list of [EnrolledExperiment]s
+     */
+    func getActiveExperiments() -> [EnrolledExperiment]
```
boek, line 112: Should we have an API for getting all available experiments?

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -0,0 +1,86 @@
...
+        guard enabled else {
+            return NimbusDisabled()
+        }
+
+        let context = Nimbus.buildExperimentContext(appSettings)
+        let remoteSettings = server.map { server -> RemoteSettingsConfig in
+            let url = server.url.absoluteString
+            return RemoteSettingsConfig(
+                serverUrl: url,
+                bucketName: EXPERIMENT_BUCKET_NAME,
```
boek, line 48: Will the client ever need / want to change these values?

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -0,0 +1,86 @@
...
+                dbpath: dbPath,
+                remoteSettingsConfig: remoteSettings,
+                // The "dummy" field here is required for obscure reasons when generating code on desktop,
+                // so we just automatically set it to a dummy value.
+                availableRandomizationUnits: AvailableRandomizationUnits(clientId: nil, dummy: 0)
+            )
+
+            return Nimbus(nimbusClient: nimbusClient, errorReporter: errorReporter)
+        } catch {
+            errorReporter(error)
```
boek, line 65: Instead of catching and swallowing the error internally can we make this function `throw` to force the client to handle the error?

.buildconfig-android.yml
jhugman: Pulling this out, is a good idea.

components/nimbus/ios/Nimbus/NimbusCreate.swift
jhugman: I went with your suggestion of using wenderlich styling.

travis79 (approved):
This is looking good!  (Swift is a nifty language and I am excited to try out this implementation of Nimbus!!)

components/nimbus/ios/Nimbus/NimbusApi.swift
jhugman: The use-case for `get_all_experiments()` is to build QA tooling to exercise the branches available. It'd be nice if we didn't need to touch this tooling each time we add an experiment.

This is covered in [SDK-228][1].

[1]: https://jira.mozilla.com/browse/SDK-228

components/nimbus/ios/Nimbus/NimbusCreate.swift
jhugman: Hmm. Not sure; I'm not sure when it would make sense, TBH.

PR #3935, opened 2021-03-17
Wrap the autofill implementation in an Arc<> to help the sync manager.

mhammond:
This patch works-around the lack of the functionality described
in https://github.com/mozilla/uniffi-rs/issues/417.

In short, it splits the store into 2 parts - the
uniffi-exposed "wrapper" and the actual implementation.
This "wrapper" wraps the implementation in an `Arc<Mutex>>`, so
that it can hand a clone of this `Arc<>` to the sync manager, so
that it can construct sync engines, causing us to end up with
multiple references to the `Store` being alive but without any
being actual `&` references.

(Thinking some more about it, I probably could have done this
slightly differently, and just wrapped the DB itself, but I
ended up somewhat automatically taking this route because this
is what the shape would naturally be if uniffi helped us.
I think passing the entire store around does make more sense
in the longer term anyway (eg, there might be other state beyond
the DB) and is what the other components do, but it does make
this patch bigger than it would otherwise be)

It also makes some changes in the sync manager, really just
enough to convince me that it will work - there will be plenty
of other places that will need similar treatment. I chose to
use the `SyncEngine` trait exclusively there, which I think makes
sense because ideally we'd use that trait *everywhere* to avoid the
Sync Manager knowing about the implementation (but it would still
require it to link to all those crates, so that's just a
code-simplicity issue)

### DIFF
--- a/components/autofill/src/db/store.rs
+++ b/components/autofill/src/db/store.rs
@@ -11,35 +11,143 @@ use rusqlite::{
     Connection,
 };
 use sql_support::{self, ConnExt};
+use std::cell::RefCell;
 use std::path::Path;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, Weak};
 use sync15_traits::SyncEngine;
 use sync_guid::Guid;
 
-#[allow(dead_code)]
-pub struct Store {
-    db: Arc<Mutex<AutofillDb>>,
+// Our "sync manager" will use whatever is stashed here.
+lazy_static::lazy_static! {
+    // Mutex: just taken long enough to update the inner stuff - needed
+    //        to wrap the RefCell as they aren't `Sync`
+    // RefCell: So we can replace what it holds. Normally you'd use `get_ref()`
+    //          on the mutex and avoid the RefCell entirely, but that requires
+    //          the mutex to be declared as `mut` which is apparently
+    //          impossible in a `lazy_static`
+    // [Arc/Weak]<StoreImpl>: What the sync manager actually needs.
+    pub static ref STORE_FOR_MANAGER: Mutex<RefCell<Weak<StoreImpl>>> = Mutex::new(RefCell::new(Weak::new()));
 }
 
-#[allow(dead_code)]
+// This is the type that uniffi exposes. It holds an `Arc<>` around the
+// actual implementation, because we need to hand a clone of this `Arc<>` to
+// the sync manager and to sync engines. One day
+// https://github.com/mozilla/uniffi-rs/issues/417 will give us access to the
+// `Arc<>` uniffi owns, which means we can drop this entirely (ie, `Store` and
+// `StoreImpl` could be re-unified)
+// Sadly, this is `pub` because our `autofill-utils` example uses it.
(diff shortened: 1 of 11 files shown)


mhammond:
The other thing I meant to say is that as we discussed, this inverts how the sync manager gets engines. Currently, something (ie, Fenix code) needs a reference to both places and the sync manager and calls (something like) `manager.set_places(places)`.

However, in this world, Fenix only needs a reference to the autofill store, and can do just `autofill.offer_to_sync_manager()` - and magically, a sync manager that's created in the future (or even one already created) will pick it up. I'm obviously open to a better name.

mhammond:
cc @jhugman for context on that uniffi ticket.

components/autofill/src/db/store.rs
```diff
@@ -11,50 +11,163 @@ use rusqlite::{
...
+    //          on the mutex, but that requires the mutex to be declared as
+    //          `mut` which is apparently impossible.
+    // [Arc/Weak]<Mutex<StoreImpl>>: What the sync manager actually needs.
+    pub static ref STORE_FOR_MANAGER: Mutex<RefCell<Weak<Mutex<StoreImpl>>>> = Mutex::new(RefCell::new(Weak::new()));
+}
+
+// This is the type that uniffi exposes. It holds an `Arc<Mutex<>>` around the
+// actual implementation, because we need access to the `Arc<Mutex<>>` for
+// integration with the sync manager. Our UDL also defines us as being
+// "ThreadSafe" to we manage our own locking via the `Mutex` - but this means
```
lougeniaC64, line 35: nit: "so" we manage?

components/autofill/src/db/store.rs
```diff
@@ -11,50 +11,163 @@ use rusqlite::{
...
+// in practice we are actually held via an `Arc<Arc<Mutex<>>>` - but one day
+// https://github.com/mozilla/uniffi-rs/issues/417 will give us access to the
+// `Arc<>` uniffi owns, which means we can drop this mess entirely (ie, `Store`)
+// can then just have a mutex around the DB.
 pub struct Store {
```
lougeniaC64, line 40: Does the `Store` struct need to be public? It looks like `StoreImpl` is the type that is being imported.

mhammond: This got confusing quickly and sadly they *both* need to be pub. `Store` needs to be pub because the sync manager uses it (StoreImpl has the `Arc<Store>` which we clone and hand to the sync manager, so it gets the `Store`) and `Store` needs to be pub because autofill-utils needs it to get the sync engines.

So yeah, this sucks and I added some comments, but hopefully we can get rid of this confusion with the proposed autofill enhancements.

mhammond:
I'm about to put up a new version that doesn't wrap the entire store in an `Arc<Mutex>` - instead the store is just wrapped in an `Arc<>` and the DB remains wrapped in the `Mutex` it already has. This is both a smaller patch and what's technically needed currently. Later in the future we might need to shuffle this around a bit, but I think this is better.

lougeniaC64 (approved):
LGTM!

PR #3941, opened 2021-03-18
Compatibility with rust-ece breaking API changes

rfk:
~~WIP updates for compatibility with the work I'm doing over in https://github.com/mozilla/rust-ece/pull/52~~

The rust-ece crate has had its interface simplified to remove footguns, let's update to that release and get a few cleanups of our own while we're at it.


components/push/src/communications.rs
```diff
@@ -468,7 +468,7 @@ mod test {
             .with_body(body)
             .create();
             let mut conn = connect(config.clone(), None, None).unwrap();
-            let channel_id = hex::encode(crate::crypto::get_bytes(16).unwrap());
+            let channel_id = hex::encode(crate::crypto::get_random_bytes(16).unwrap());
```
rfk, line 471: Sorry, I just, I couldn't handle the name of this function - what *sort* of bytes are we getting??

components/push/src/crypto.rs
```diff
@@ -50,14 +48,12 @@ impl Key {
         }
     }
 
-    pub fn key_pair(&self) -> error::Result<RcCryptoLocalKeyPair> {
-        RcCryptoLocalKeyPair::from_raw_components(&self.p256key).map_err(|e| {
```
rfk, line 54: I think we were accidentally biting ourselves here with a bit of bad API design, which meant that we couldn't use the `aesgcm` scheme without knowing which specific crypto backend was in use by the `rust-ece` crate. That has been fixed now (thanks to an external contributor! ref https://github.com/mozilla/rust-ece/pull/47) so we can remove quite a bit of key parsing-and-re-serializing here.

components/push/src/subscriber.rs
```diff
@@ -232,11 +232,7 @@ mod test {
         let mut pm = PushManager::new(test_config)?;
         let (info, key) = pm.subscribe(test_channel_id, "", None)?;
         // Act like a subscription provider, so create a "local" key to encrypt the data
-        let mut auth_secret = vec![0u8; 16];
-        let mut salt = vec![0u8; 16];
-        rc_crypto::rand::fill(&mut auth_secret).unwrap();
-        rc_crypto::rand::fill(&mut salt).unwrap();
```
rfk, line 238: Not letting you pass in this salt, was the main breaking API change. The latest version of `rust-ece` just makes one for you automatically.

components/support/rc_crypto/src/ece_crypto.rs
```diff
@@ -178,260 +178,10 @@ pub(crate) fn init() {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use ece::*;
-
-    // Copy-pasta from the tests in the ECE crate.
```
rfk, line 183: Pretty happy to dump this copy-pasta...

rfk (commented):
@lougeniaC64 could you please take a look at these corresponding changes on the consumer side of the rust-ece crate?

components/push/src/crypto.rs
```diff
@@ -188,7 +184,6 @@ impl Cryptography for Crypto {
         }
     }
 
-    // IIUC: objects created on one side of FFI can't be freed on the other side, so we have to use references (or clone)
```
rfk, line 191: I wasn't able to map what this comment is saying to anything happening in the code below it, so I presumptuously deleted it.

lougeniaC64 (approved):
LGTM! 🚀

PR #3944, opened 2021-03-18
Update Nimbus to v0.10.0

travis79:
This updates Nimbus to v0.10.0, including breaking changes to the `AppContext` object which add `app_name` and `channel` as non-optional fields.


dmose (approved):
rs=dmose

PR #3945, opened 2021-03-18
Add nimbus update to cut-v74.0.0

travis79:


rfk (approved):
I'm trying to think whether we should re-order things here so that the commit that says "Cut release v74.0.0" is on top of the stack, but it probably doesn't matter.

travis79:
> I'm trying to think whether we should re-order things here so that the commit that says "Cut release v74.0.0" is on top of the stack, but it probably doesn't matter.

Maybe I can squash and force push to the cut branch rather than PR?

rfk:
> Maybe I can squash and force push to the cut branch rather than PR?

If you're happy to try that, that sounds good - I can stand by for r+ in the other PR (or, we can close that one and make a new one with your squashed commit)

travis79:
Closing this in favor of just squashing this into #3943

PR #3948, opened 2021-03-18
Revert "Automagically add rustup mobile targets via toolchain config" [ci full]

rfk:
This reverts commit 02bf1a0b7c951896e715a5b00d43d5af80bfe479.

The change from #3936 seems to be causing some issues with building releases, let's roll it back until we can debug properly.


rfk:
@mhammond just flagging this as the one that we want to get into a point-release today.

mhammond (approved):

PR #3950, opened 2021-03-18
Automagically add rustup mobile targets via toolchain config.

rfk:
We originally landed this change in #3936 but had to revert it due to release build issues. Let's bring it back, but after making sure that everything builds fine end-to-end. This PR is marked `[ci full]` for said debugging purposes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,6 @@ version: 2.1
 
 # ⚠️ If you add, rename or delete a job here, please also update .mergify.yml! ⚠️
 
-parameters:
-  # See also ../taskcluster/scripts/toolchain/rustup-setup.sh
-  stable-rust-version:
-    type: string
-    default: 1.50.0
-
 commands:
   full-checkout:
     steps:
@@ -77,8 +71,8 @@ commands:
           name: Install Rust
           command: |
             RUSTUP_PLATFORM=x86_64-apple-darwin
-            RUSTUP_VERSION=1.18.3
-            RUSTUP_SHA256=16734a9a2d87a3054bd4eea962642687e50a34b6e73e17df6f3361c2c534dc30
+            RUSTUP_VERSION=1.23.1
+            RUSTUP_SHA256=39101feb178a7e3e4443b09b36338e794a9e00385e5f44a2f7789aefb91354a9
             curl -sfSL --retry 5 --retry-delay 10 -O "https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/${RUSTUP_PLATFORM}/rustup-init"
             echo "${RUSTUP_SHA256} *rustup-init" | shasum -a 256 -c -
             chmod +x rustup-init
@@ -86,21 +80,12 @@ commands:
             rm rustup-init
             echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $BASH_ENV
   setup-rust-toolchain:
-    parameters:
-      rust-version:
-        type: string
-        default: << pipeline.parameters.stable-rust-version >>
     steps:
       - run:
           name: Set up Rust toolchain
           command: |
             rustup update
-            rustup install <<parameters.rust-version>>
-            rustup default <<parameters.rust-version>>
             rustc --version
-      - run:
-          name: Install UniFFI
-          command: cargo install --version 0.7.1 uniffi_bindgen
   build-libs:
     parameters:
       platform:
@@ -159,15 +144,10 @@ commands:
           paths:
             - Carthage
   test-setup:
-    parameters:
-      rust-version:
-        type: string
-        default: << pipeline.parameters.stable-rust-version >>
     steps:
       - full-checkout
       - build-desktop-libs
-      - setup-rust-toolchain:
-          rust-version: <<parameters.rust-version>>
+      - setup-rust-toolchain
       - setup-sccache
       # clipboard crate depends on some x11 libs.
       - run: sudo apt-get install libxcb-shape0-dev libxcb-xfixes0-dev
@@ -183,22 +163,19 @@ commands:
     parameters:
       rust-version:
         type: string
-        default: << pipeline.parameters.stable-rust-version >>
+        default: ""
     steps:
-      - test-setup:
-          rust-version: <<parameters.rust-version>>
+      - test-setup
       # Test with 1. only default features on, 2. all features on, 3. no features on.
       # This is not perfect (really we want the cartesian product), but is good enough in practice.
       - run:
           name: Test (default features, no features, and all features)
-          command: bash automation/all_rust_tests.sh --verbose
+          command: RUSTUP_TOOLCHAIN="<<parameters.rust-version>>" automation/all_rust_tests.sh --verbose
   dependency-checks:
     steps:
       - run:
           name: Check for security vulnerabilities in dependencies
           command: |
-            rustup install nightly
-            rustup default nightly
             cargo install cargo-audit
             # Explanation for ignored issues:
             #  * RUSTSEC-2021-0019:  Soundness issues in `xcb`, a clipboard library we only use for examples.
@@ -208,8 +185,6 @@ commands:
           name: Check for any unrecorded changes in our dependency trees
           command: |
             pip3 install --require-hashes -r ./tools/requirements.txt
-            rustup install nightly
-            rustup default nightly
             cargo metadata --locked > /dev/null
             python3 ./tools/dependency_summary.py --check ./DEPENDENCIES.md
             python3 ./tools/dependency_summary.py --all-ios-targets --package megazord_ios --check megazords/ios/DEPENDENCIES.md
@@ -261,7 +236,6 @@ jobs:
     steps:
       - full-checkout
       - setup-rust-toolchain
-      - run: rustup component add rustfmt
       - run: rustfmt --version
       - run: cargo fmt --all -- --check
   Lint Rust with clippy:
@@ -270,7 +244,6 @@ jobs:
     steps:
       - test-setup
       - restore-sccache-cache
-      - run: rustup component add clippy
       - run: cargo clippy --version
       - run: bash automation/all_clippy_checks.sh
       - save-sccache-cache
@@ -279,11 +252,11 @@ jobs:
       - image: circleci/rust:latest
     resource_class: large
     steps:
-      # Need to use nightly for this.
-      - test-setup:
-          rust-version: "nightly"
+      - test-setup
       - install-grcov
-      - run: bash ./automation/emit_coverage_info.sh
+      # Test coverage support in Rust has been evolving rapidly, use nightly
+      # in the hope of getting the best support/features on that front.
+      - run: RUSTUP_TOOLCHAIN="nightly" bash ./automation/emit_coverage_info.sh
       - run:
           name: Compile coverage
           command: grcov ./target/debug/ -s . -t lcov --llvm --ignore-not-existing --ignore "target/*" --ignore "/*" -o lcov.info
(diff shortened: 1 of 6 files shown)


rfk:
So this fails in taskcluster with a bunch of these errors:

```
[task 2021-03-18T23:27:17.312Z] > Task :full-megazord:cargoBuildDarwin FAILED
[task 2021-03-18T23:27:17.312Z]    Compiling cfg-if v0.1.10
[task 2021-03-18T23:27:17.312Z]    Compiling lazy_static v1.4.0
[task 2021-03-18T23:27:17.312Z]    Compiling itoa v0.4.6
[task 2021-03-18T23:27:17.312Z]    Compiling tinyvec v0.3.4
[task 2021-03-18T23:27:17.312Z]    Compiling matches v0.1.8
[task 2021-03-18T23:27:17.312Z]    Compiling percent-encoding v2.1.0
[task 2021-03-18T23:27:17.312Z]    Compiling base64 v0.12.3
[task 2021-03-18T23:27:17.312Z]    Compiling once_cell v1.5.2
[task 2021-03-18T23:27:17.312Z]    Compiling ahash v0.4.6
[task 2021-03-18T23:27:17.312Z]    Compiling ppv-lite86 v0.2.10
[task 2021-03-18T23:27:17.312Z]    Compiling fallible-iterator v0.2.0
[task 2021-03-18T23:27:17.312Z]    Compiling smallvec v1.6.1
[task 2021-03-18T23:27:17.312Z]    Compiling fallible-streaming-iterator v0.1.9
[task 2021-03-18T23:27:17.312Z]    Compiling bytes v0.5.6
[task 2021-03-18T23:27:17.312Z]    Compiling interrupt-support v0.1.0 (/builds/worker/checkouts/src/components/support/interrupt)
[task 2021-03-18T23:27:17.312Z]    Compiling base16 v0.2.1
[task 2021-03-18T23:27:17.312Z] error[E0463]: can't find crate for `core`
[task 2021-03-18T23:27:17.312Z]   |
[task 2021-03-18T23:27:17.312Z]   = note: the `x86_64-apple-darwin` target may not be installed
```

IIUC, we're trying to Darwin desktop versions of the libraries so that we can include them in the `-forUnitTests` version of the megazord, which lets downstream developers run Kotlin unittests on their Desktop machine rather than requiring an Android device.

We allegedly `rustup target add x86_64-apple-darwin` in the taskcluster setup script, which you can see in the diff of this PR. But @travis79 mentioned that he saw similar errors when using a slightly-out-of-date version of rustup.

However, earlier in the CI logs we see where I added version reporting and:

```
[task 2021-03-18T23:23:04.703Z] ++ rustup --version
[task 2021-03-18T23:23:04.705Z] rustup 1.23.1 (3df2264a9 2020-11-30)
```

Which is the same version of rustup that I have here locally, where things seem to work OK.

rfk:
So this is turning into a more general "disentagle our rust versions" bug. It occurs to me that the current use of `rust-toolchain` may be interfering with some of our CircleCI scripts, which use a build parameter to set the rustup default version. That default would be overridden by the version in our `rust-toolchain` file, meaning that e.g. the CircleCI tests that claim to be running against Rust beta would actually be running against the pinned version.

Indeed [this recent job](https://app.circleci.com/pipelines/github/mozilla/application-services/26262/workflows/2c6f1955-3016-41bd-8c77-957d4835947f/jobs/98717) reports having actually run the tests with `1.50.0` rather than with the current Rust beta (which reports itself as `1.51.0-beta.2`.

.circleci/config.yml
```diff
@@ -77,30 +71,21 @@ commands:
...
-      rust-version:
-        type: string
-        default: << pipeline.parameters.stable-rust-version >>
     steps:
       - run:
           name: Set up Rust toolchain
           command: |
             rustup update
-            rustup install <<parameters.rust-version>>
-            rustup default <<parameters.rust-version>>
```
rfk, line 99: Setting the default version like this has no effect in our repo, because we have a `rust-toolchain` file that overrides it. (This hasn't worked correctly since we landed #3910, but it hasn't cause any problems, it just hasn't been doing what it says it's doing).

.circleci/config.yml
```diff
@@ -77,30 +71,21 @@ commands:
           name: Install Rust
           command: |
             RUSTUP_PLATFORM=x86_64-apple-darwin
-            RUSTUP_VERSION=1.18.3
-            RUSTUP_SHA256=16734a9a2d87a3054bd4eea962642687e50a34b6e73e17df6f3361c2c534dc30
+            RUSTUP_VERSION=1.23.1
+            RUSTUP_SHA256=39101feb178a7e3e4443b09b36338e794a9e00385e5f44a2f7789aefb91354a9
```
rfk, line 75: I've updated the pinned versions of rustup while I'm in here.

.circleci/config.yml
```diff
@@ -183,22 +163,19 @@ commands:
...
     steps:
-      - test-setup:
-          rust-version: <<parameters.rust-version>>
+      - test-setup
       # Test with 1. only default features on, 2. all features on, 3. no features on.
       # This is not perfect (really we want the cartesian product), but is good enough in practice.
       - run:
           name: Test (default features, no features, and all features)
-          command: bash automation/all_rust_tests.sh --verbose
+          command: RUSTUP_TOOLCHAIN="<<parameters.rust-version>>" automation/all_rust_tests.sh --verbose
```
rfk, line 173: Using an environment variable like this will override the `rust-toolchain` file.

.circleci/config.yml
```diff
@@ -208,8 +185,6 @@ commands:
           name: Check for any unrecorded changes in our dependency trees
           command: |
             pip3 install --require-hashes -r ./tools/requirements.txt
-            rustup install nightly
-            rustup default nightly
```
rfk, line 212: Similarly, I don't think these have been actually running on nightly since #3910 landed. So I presume they no longer require nightly and it can be removed.

.circleci/config.yml
```diff
@@ -279,11 +252,10 @@ jobs:
...
     resource_class: large
     steps:
-      # Need to use nightly for this.
-      - test-setup:
-          rust-version: "nightly"
+      - test-setup
       - install-grcov
-      - run: bash ./automation/emit_coverage_info.sh
+      # Need to use nightly for this.
+      - run: RUSTUP_TOOLCHAIN="nightly" bash ./automation/emit_coverage_info.sh
```
rfk, line 258: I kept this as nightly, because I have this vague belief that coverage stuff "works better" on nightly. I don't really have any justification behind that though, and I expect we could remove this if we wanted, just to simplify things.

.circleci/config.yml
```diff
@@ -366,7 +338,6 @@ jobs:
       - run:
           name: Set up the build environment
           command: |
-            rustup target add aarch64-apple-ios x86_64-apple-ios
```
rfk, line 369: This now happens for free thanks to the toolchain file.

automation/all_rust_tests.sh
```diff
@@ -19,6 +19,9 @@ fi
 
 EXTRA_ARGS=( "$@" )
 
+echo "Testing with rustc version:"
+rustc --version
```
rfk, line 23: I added this here to be able to confirm that the circleci "run tests on beta" job actually does run them on beta, but it seems worth keeping for future debugging. (And I did a trial push to confirm that yes, it does).

rust-toolchain
```diff
@@ -1 +1,11 @@
...
+channel = "1.50.0"
+targets = [
+    "aarch64-linux-android",
+    "armv7-linux-androideabi",
+    "i686-linux-android",
+    "x86_64-linux-android", 
+    "aarch64-apple-ios",
+    "x86_64-apple-ios"
+]
+components = ["clippy", "rustfmt"]
```
rfk, line 11: Both CI and local developers will want these components installed, let's make them happen by default.

taskcluster/scripts/toolchain/cross-compile-setup.sh
```diff
@@ -25,7 +25,11 @@ tooltool.py \
   --manifest="/builds/worker/checkouts/src/libs/macos-cc-tools.manifest" \
   fetch
 
+popd || exit
+
+# It's important that we execute rustup in the checkout directory, so that it can see
+# any `rust-toolchain` override file that might be present.
```
rfk, line 31: This, it turns out, was the source of the problem with the release builds. I don't really understand *why*, because AFAICT we were setting the rustup default toolchain to the same toolchain that's specified in our `rust-toolchain` file. But anyway, running this from the repo root rather than from `/tmp` seems to have fixed the issue.

taskcluster/scripts/toolchain/rustup-setup.sh
```diff
@@ -15,26 +15,19 @@ export SCCACHE_IDLE_TIMEOUT='1200'
 export SCCACHE_CACHE_SIZE='40G'
 export SCCACHE_ERROR_LOG='/builds/worker/sccache.log'
 export RUST_LOG='sccache=info'
-# We are pinned at this rust stable version (see also ../.circleci/config.yml)
-export RUST_STABLE_VERSION='1.50.0'
```
rfk, line 19: Good news, we no longer need to keep this version in sync across multiple files, it's all done via `rust-toolchain` now.

taskcluster/scripts/toolchain/rustup-setup.sh
```diff
@@ -15,26 +15,19 @@ export SCCACHE_IDLE_TIMEOUT='1200'
...
     echo "${RUSTUP_SHA256} *rustup-init" | sha256sum -c -; \
     chmod +x rustup-init; \
     ./rustup-init -y --no-modify-path --default-toolchain none; \
     rm rustup-init
 export PATH=$HOME/.cargo/bin:$PATH
 
-
-rustup toolchain install "$RUST_STABLE_VERSION"
-rustup default "$RUST_STABLE_VERSION"
-rustup target add x86_64-linux-android i686-linux-android armv7-linux-androideabi aarch64-linux-android
```
rfk, line 36: Again, this all now happens for you by default when you run `cargo` or `rustup` for the first time inside the repo root, so we don't need to duplicate it here.

taskcluster/scripts/toolchain/cross-compile-setup.sh
```diff
@@ -25,7 +25,11 @@ tooltool.py \
...
+rustup --version
 rustup target add x86_64-apple-darwin
 rustup target add x86_64-pc-windows-gnu
```
rfk, line 34: These targets are really only useful in CI, so I opted to continue installing them here explicitly rather than making everyone get them by default.

.circleci/config.yml
```diff
@@ -77,30 +71,21 @@ commands:
...
       - run:
           name: Set up Rust toolchain
           command: |
             rustup update
-            rustup install <<parameters.rust-version>>
-            rustup default <<parameters.rust-version>>
             rustc --version
-      - run:
-          name: Install UniFFI
-          command: cargo install --version 0.7.1 uniffi_bindgen
```
rfk, line 103: This was leftover after a previous commit made it unnecessary, removing it while I'm here.

rfk (commented):
OK, I think I figured out what was going on here, and I'm glad to have worked it through because the result is IMHO a pretty nice cleanup of our CI config. Notes inline below.

rfk:
Also, I found https://rust-lang.github.io/rustup/overrides.html#toolchain-override-shorthand really useful for understanding how rustup decides which version of rust to run for you.

mhammond (approved):
Wow, some large yaks here, thanks for pushing through.  It might also be worth dropping a link to that toolchain-override-shorthand URL somewhere in the code too?

rfk:
Sounds good, but I'm going to take `[ci full]` off this PR before pushing those final tweaks, because I'm sure I've already expended enough taskcluster budget for one day 😬

PR #3963, opened 2021-03-22
[ci full] merge nimbus-sdk into this repo

dmose:

  - _There are no new dependencies (all the nimbus-sdk dependencies were already present, they were just previously pulled from a submodule.)_


dmose:
@rfk (and anyone else inclined), I'd love your review/approval.  I still have a few documentation tweaks to make, but the rest is done.   The current CI failure seems to be some weird infrastructure thing.

components/nimbus/CHANGELOG.md
```diff
@@ -0,0 +1,161 @@
+# Unreleased changes
```
rfk, line 1: Not for this PR, but I feel like it's going to be weird to have this nimbus-specific changelog here when future entries are going into the main changelog file. Some options:

* We could delete it.
* We could rename it to something like "OLD_CHANGELOG.md"
* We could painstakingly merge the changelog entries here into the main changelog, corresponding to the appservices version that they shipped in.

I think I'd probably lean towards renaming it. Thoughts?

(Just to reinforce, that doesn't have to block landing)

CHANGES_UNRELEASED.md
```diff
@@ -1,7 +1,21 @@
...
+  `components/nimbus-sdk`.  It is no longer a submodule.  Developers
+  may need to execute
+
+  ```bash
+  rm -fR components/external/nimbus-sdk
+  ```
+
+  This is not expected to have any ramifications for consumers, other than
+  that the `nimbus-sdk` will (in the future) have the same version numbers as
+  the rest of `application-services`.
```
rfk, line 17: Since we were already shipping it via the megazord, I don't think this will actually change from the consumer's perspective - they would already have been consuming it via a package named `org.mozilla.appservices:nimbus` with the same version number as the rest of appservices.

components/nimbus/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+rkv-safe-mode = []
+uniffi-bindings = ["uniffi", "uniffi_build"]
+
+[dependencies]
+anyhow = "1"
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
+log = "0.4"
+# Viaduct locked to version same version of app-services as mozilla-central
```
rfk, line 27: nit: this comment is now out-of-date.

components/nimbus/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+uniffi_build = { version = "^0.8.0", features = [ "builtin-bindgen" ], optional = true }
+
+[dev-dependencies]
+viaduct-reqwest = { path = "../support/viaduct-reqwest" }
+mockito = "0.27"
+env_logger = "0.7"
+clap = "2.33.3"
+tempdir = "0.3"
+
+[package.metadata.release]
```
rfk, line 49: nit: we should remove this `cargo release` config, it will no longer be used because we'll no longer be using `cargo release`.

components/nimbus/android/build.gradle
```diff
@@ -3,9 +3,6 @@
 // READ THIS CAREFULLY before editing this file.
 // =============================================
 //
-// This is a lightly-modified version of the `build.gradle` file from Nimbus,
-// which you can find at `../external/nimbus-sdk/android/build.gradle`.
-//
 // The aim here is to pre-package the Nimbus Kotlin code and publish it under
 // `org.mozilla.appservices:nimbus` rather than `org.mozilla.experiments:nimbus`,
```
rfk, line 7: nit: the comment here about using `appservices:nimbus` rather than `experiments:nimbus` and megazording can now be deleted, since there will no longer be anything published as `experiments:nimbus`.

components/nimbus/examples/README.md
```diff
@@ -0,0 +1,31 @@
+# Experiments CLI
+In this directory we have a command line interface that helps us interact with the Nimbus SDK.
```
rfk, line 2: For [reasons](https://github.com/mozilla/application-services/blob/main/docs/design/test-faster.md), we prefer to keep example crates in the top-level `./examples` subdirectory rather than inside the crate directory. Not necessary for this PR, but could you please file a follow-up issue to move this example code out into the main `./examples` directory?

components/nimbus/old-nimbus-sdk-rust-code-README.md
```diff
@@ -0,0 +1,17 @@
+# Experiments SDK
```
rfk, line 1: I reckon we could probably just delete this file.

components/nimbus/old-nimbus-sdk-toplevel-README.md
```diff
@@ -0,0 +1,42 @@
+# nimbus-sdk [![CircleCI](https://circleci.com/gh/mozilla/nimbus-sdk/tree/main.svg?style=svg)](https://circleci.com/gh/mozilla/nimbus-sdk/tree/main)
+Cross Platform Rapid Experiments "Nimbus" SDK
```
rfk, line 2: With the "cutting a release" stuff removed, I think I would be happy for this file to remain as `README.md` and serve as a component-level readme, and we can iterate on it over time to improve consistency with other components. WDYT?

components/nimbus/uniffi.toml
```diff
@@ -0,0 +1,2 @@
+[bindings.kotlin]
+package_name = "org.mozilla.experiments.nimbus"
```
rfk, line 2: Please add this, in line with other `uniffi.toml` files in this repo:

```
cdylib_name = "megazord"
```

This will cause the bindings to look for the library name in the correct `.so` by default, removing the need for [this libraryOverride setting](https://github.com/mozilla-mobile/android-components/blob/0bc7edf4a84a437c3819bd11f212d47998ba02e3/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/Nimbus.kt#L240-L245) inside of android-components.

We should also consider whether we want to change the official kotlin package name from `org.mozilla.experiments.nimbus` to `org.mozilla.appservices.nimbus` for consistency. I don't feel too strongly on that front, but also, I think it would be a pretty minor change for the android component. /cc @travis79.

rfk (approved):
This looks good to me, thanks for working through all the details here!

I've left a collection of comments below about small changes that could be made to improve consistency with the rest of the repo. I think it would be fine to do most of them inline in this PR, but also fine if you wanted to push them out to separate follow-up issues in order to get this landed.

CHANGES_UNRELEASED.md
dmose: OK, I've just shortened it to "This is not expected to have any ramifications for consumers."

components/nimbus/Cargo.toml
dmose: Fixed.

components/nimbus/android/build.gradle
dmose: The whole comment block is essentially obsolete.  I've removed it.

components/nimbus/CHANGELOG.md
dmose: Renamed and added an explanatory comment to the head of the file.

components/nimbus/Cargo.toml
dmose: Fixed.

components/nimbus/old-nimbus-sdk-rust-code-README.md
dmose: Done.

components/nimbus/uniffi.toml
dmose: Done, and filed mozilla-mobile/android-components#9969 for the removal

components/nimbus/old-nimbus-sdk-toplevel-README.md
dmose: Done.

components/nimbus/examples/README.md
dmose: SDK-235 has been filed.

PR #3981, opened 2021-03-24
Use cargo to install grcov

mhammond:
Circle is using v0.5.15, latest is v0.7.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,17 +41,7 @@ commands:
       - run:
           name: Install grcov
           command: |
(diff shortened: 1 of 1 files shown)


rfk (approved):
I guess we lose the sha-checking here, but that seems OK in this case, it's not a production dependency.

mhammond:
> I guess we lose the sha-checking here, but that seems OK in this case, it's not a production dependency.

And I couldn't find the sha checksums published anyway, and noticed we already used cargo so 🤷

PR #3987, opened 2021-03-25
Add a full "how to add a component" howto (and various other doc updates) [ci skip]

rfk:
Fixes https://jira.mozilla.com/browse/SYNC-2130.

The main point here is to add a full walkthrough of all the steps involved in adding a new component in this repo, based on my recent experience adding the small "crashtest" component. There are a lot of steps! Please [**click here**](https://github.com/mozilla/application-services/blob/uniffi-howto-updates/docs/howtos/adding-a-new-component.md) for a rendered view of that file.

While I was in there, I took the opportunity to clean up some of the existing documentation, including removing things that didn't seem very useful.


rfk:
@jhugman and @dmose, I'd appreciate your feedback on this given your recent experience adding nimbus-sdk.

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
+
+# Rust + Android FAQs
```
rfk, line 2: This used to be in a "how do I expose rust code to kotlin?" howto, but most of the good stuff was actually FAQs, so I removed the no-longer-recommented palaver about manually writing bindings and turned it into an actual FAQ.

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+all originate on the same thread, you can tell it to cache these.
+
+Calling back from Rust into Kotlin isn’t too bad so long as you ensure the
+callback can not be GCed while rust code holds onto it, and you can either
+accept the overhead of extra threads being instantiated on each call, or you can
+ensure that it only happens from a single thread.
+
+Note that the situation would be somewhat better if we used JNI directly (and
+not JNA), but this would cause us to need to write two versions of each ffi
+crate, one for iOS, and one for Android.
```
mhammond, line 65: It's not clear how this para relates to uniffi, where we don't write *any* versions of an ffi

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+
+Ultimately, in any case where you can reasonably move to making something a
```
mhammond, line 67: I think this para is trying to say something like "Often, callbacks are used to implement async functions - it's often simpler to just move to making it a blocking call and avoiding the callback entirely"?

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+completely certain that a JVM is associated with a given thread in order to call
+java code on it. The difficulty is that the JVM can GC its threads and will not
+let rust know about it. JNA can work around this for us to some extent, however
+there are difficulties.
+
+The approach it takes is essentially to spawn a thread for each callback
+invocation. If you are certain you’re going to do a lot of callbacks and they
+all originate on the same thread, you can tell it to cache these.
+
+Calling back from Rust into Kotlin isn’t too bad so long as you ensure the
```
mhammond, line 58: I don't quite get this para - the ones above basically say it is difficult because the threads might be GCs, but this says it's not bad so long as a you can avoid the GC without guidance on how to do this.

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+in Kotlin. This is in line with the Android documentation on JNI usage, and my
+own experience. It’s vastly simpler and less painful this way.
+
+(Of course, not every case is solvable like this).
+
+### Why are we using JNA rather than JNI, and what tradeoffs does that involve?
+
+We get a couple things from using JNA that we wouldn't with JNI.
+
+1. We are able to write a *single* FFI crate. If we used JNI we'd need to write
```
mhammond, line 78: same comment above re uniffi

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+   code out there that uses `bool` as a typedef for a 32-bit `int`).
+2. JNA makes it really easy to do the wrong thing and have it work but corrupt
+   memory. Several of the caveats around this are documented in the
+   [`ffi_support` docs](https://docs.rs/ffi-support/*/ffi_support/), but a
+   major one is when to use `Pointer` vs `String` (getting this wrong will
+   often work, but may corrupt memory).
+
+### How do I debug Rust code with the step-debugger in Android Studio
+
+1. Uncomment the `packagingOptions { doNotStrip "**/*.so" }` line from the
```
mhammond, line 102: TIL :)

docs/naming-conventions.md
```diff
@@ -4,6 +4,8 @@ All names in this project should adhere to the guidelines outlined in this docum
 
 ## Rust Code
 
+TL;DR: do what Rust's builtin warnings and clippy lints tell you.
```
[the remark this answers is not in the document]
mhammond: Maybe note CI will fail otherwise?

mhammond (approved):
Massive improvement, thanks. Approving because I really don't need to check the comments I made are addressed correctly, but happy to have another read if you want. Probably also worth waiting for some others to have a read too.

docs/android-faqs.md
rfk: Haha, yes, I learned this about two days ago when copy-pasting this from one markdown document to another 😆

docs/android-faqs.md
```diff
@@ -0,0 +1,119 @@
...
+
+### What challenges exist when calling back into Kotlin from Rust?
+
+There are a number of them. The issue boils down to the fact that you need to be
+completely certain that a JVM is associated with a given thread in order to call
+java code on it. The difficulty is that the JVM can GC its threads and will not
+let rust know about it. JNA can work around this for us to some extent, however
+there are difficulties.
+
+The approach it takes is essentially to spawn a thread for each callback
```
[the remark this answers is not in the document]
dmose: Uniffi is nice, since it's more explicit

docs/howtos/adding-a-new-component.md
```diff
@@ -0,0 +1,161 @@
...
+["megazord"](../design/megazords.md) crate for each target platform:
+
+* For Android, add it as a dependency in `./megazords/full/Cargo.toml` and
+  add a `pub use <your_crate_name>` to `./megazords/full/src/lib.rs`.
+* For iOS, add it as a dependency in `./megazords/ios/rust/Cargo.toml` and
+  add a `pub use <your_crate_name>` to `./megazords/ios/rust/src/lib.rs`.
+
+Run `cargo check -p <your_crate_name>` in the repository root to confirm that
+things are configured properly. This will also have the side-effect of updating
+`Cargo.lock` to contain your new crate.
```
dmose, line 44: s/crate./crate and its dependencies/.' perhaps

docs/howtos/adding-a-new-component.md
```diff
@@ -0,0 +1,161 @@
...
+
+## The Swift Bindings
+
+Make a `./components/<your_crate_name>/ios` subdirectory to contain
+Swift- and iOS-specific code. The UniFFI-generated swift bindings will
+be written to a subdirectory named `Generated`.
+
+You can include hand-written Swift code alongside the automatically
+generated bindings, by placing `.swift` files in a directory named:
+* `./ios/<your_crate_name>/`
```
dmose, line 99: It's hard for me to create the mental model from the text above.  Maybe just show a list of the directories under iOS for an Example component with a line of text describing what each one is for, rather than writing prose?

docs/howtos/adding-a-new-component.md
```diff
@@ -0,0 +1,161 @@
...
+    * `uniffi_<your_crate_name>-Bridging-Header.h`
+
+The result should look something like this:
+
+![Screenshot of XCode Project Navigator](./img/xcode_add_component_1.png)
+
+Click on the top-level "MozillaAppServices" project in the navigator,
+then go to "Build Phases" and add `<your_crate_name>.udl` to the list
+of "Compile Sources". This will trigger an XCode Build Rule that generates
+the Swift bindings automatically. Also inclue any hand-written `.swift` files
```
dmose, line 130: s/inclue/include

dmose (approved):
Looks good to me.  r=dmose

docs/android-faqs.md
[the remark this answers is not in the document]
rfk: I believe "it" here actually refers to JNA - that is, JNA works around the you-need-to-be-completely-certain-that-a-JVM-is-associated-with-the-thread problem by executing the callback on a thread that it has complete control over. I'll reword for clarity.

docs/android-faqs.md
rfk: This whole chunk of docs predates uniffi, but I'm going to replace "write" with "generate" here. Actually uniffi would probably make this more tractable if we wanted to go this route!

docs/howtos/adding-a-new-component.md
rfk: Thanks, good idea, I will add this for both the iOS and Android parts 👍

PR #3990, opened 2021-03-25
[ci full] add logging and new license names to dependency script

dmose:

  - _Infrastructure-only change_
  - _Non user-visible change._
  - _None_

### DIFF
--- a/tools/dependency_summary.py
+++ b/tools/dependency_summary.py
@@ -675,6 +675,8 @@ PACKAGE_METADATA_FIXUPS = {
 # If we can find one and only one of these files in a package, then we can be confident
 # that it's the intended license text.
 COMMON_LICENSE_FILE_NAME_ROOTS = {
+    "": ["license", "licence", "LICENSE"],
+    "Apache-2.0": ["license-apache", "licence-apache", "LICENSE-APACHE"],
     "": ["license", "licence"],
     "Apache-2.0": ["license-apache", "licence-apache"],
     "MIT": ["license-mit", "licence-mit"],
--- a/tools/dependency_summary.py
+++ b/tools/dependency_summary.py
@@ -16,6 +16,7 @@ import argparse
 import subprocess
 import hashlib
 import json
+import logging
 import textwrap
 import difflib
 import itertools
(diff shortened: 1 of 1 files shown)


tools/dependency_summary.py
```diff
@@ -671,6 +675,8 @@
 # If we can find one and only one of these files in a package, then we can be confident
 # that it's the intended license text.
 COMMON_LICENSE_FILE_NAME_ROOTS = {
+    "": ["license", "licence", "LICENSE"],
+    "Apache-2.0": ["license-apache", "licence-apache", "LICENSE-APACHE"],
```
rfk, line 679: I'm surprised this was necessary - I thought we had some use of `name.lower()` in the code that uses this in order to deal with capitalization issues, but maybe that's not working quite as intended?

tools/dependency_summary.py
```diff
@@ -924,6 +931,7 @@ def _find_license_file(self, id, license, pkgInfo):
             licenseFileNames = COMMON_LICENSE_FILE_NAMES[license]
         except KeyError:
             licenseFileNames = COMMON_LICENSE_FILE_NAMES[""]
+        logging.debug("candidate license filenames: %s", licenseFileNames)
         foundLicenseFiles = [nm for nm in os.listdir(
             pkgRoot) if nm.lower() in licenseFileNames]
```
rfk, line 936: See e.g. here where we use `nm.lower()`.

rfk (commented):
Logging is an excellent idea! A question about the other change below:

tools/dependency_summary.py
dmose: Yeah, I'd think that should do the trick too.  Right now, I'm having issues reproducing the problem I saw, so I'll just remove these particular changes and the land the logging stuff without them.

tools/dependency_summary.py
dmose: I've now removed them, and if the issue comes up again, we can debug the root cause, rather than papering over.

rfk (approved):
👍🏻 thanks!

PR #3991, opened 2021-03-25
Add extra initializers to Nimbus objects

jhugman:
This is followup work to [SDK-208][1], in order to be usable from Firefox for iOS (the work of [SDK-210][2]).

[1]: https://jira.mozilla.com/browse/SDK-208
[2]: https://jira.mozilla.com/browse/SDK-210

### DIFF
--- a/components/nimbus/ios/Nimbus/NimbusApi.swift
+++ b/components/nimbus/ios/Nimbus/NimbusApi.swift
@@ -124,14 +124,23 @@ public extension Notification.Name {
 /// This struct is used during in the `create` method to point `Nimbus` at the given `RemoteSettings` server.
 ///
 public struct NimbusServerSettings {
-    let url: URL
+    public init(url: URL) {
+        self.url = url
+    }
+
+    public let url: URL
 }
 
 /// Name and channel of the app, which should agree with what is specified in Experimenter.
 ///
 public struct NimbusAppSettings {
-    let appName: String
-    let channel: String
+    public init(appName: String, channel: String) {
+        self.appName = appName
+        self.channel = channel
+    }
+
+    public let appName: String
+    public let channel: String
 }
 
 /// This error reporter is passed to `Nimbus` and any errors that are caught are reported via this type.
--- a/components/nimbus/ios/Nimbus/Nimbus.swift
+++ b/components/nimbus/ios/Nimbus/Nimbus.swift
@@ -208,6 +208,8 @@ extension Nimbus: NimbusStartup {
 }
 
 public class NimbusDisabled: NimbusApi {
+    public static let shared = NimbusDisabled()
+
     public var globalUserParticipation: Bool = false
 }
 
--- a/components/nimbus/ios/Nimbus/NimbusCreate.swift
+++ b/components/nimbus/ios/Nimbus/NimbusCreate.swift
@@ -39,7 +39,7 @@ extension Nimbus {
                               errorReporter: @escaping NimbusErrorReporter = defaultErrorReporter) throws -> NimbusApi
     {
         guard enabled else {
-            return NimbusDisabled()
+            return NimbusDisabled.shared
         }
 
         let context = Nimbus.buildExperimentContext(appSettings)


components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -124,14 +124,23 @@ public extension Notification.Name {
...
 
 /// Name and channel of the app, which should agree with what is specified in Experimenter.
 ///
 public struct NimbusAppSettings {
-    let appName: String
-    let channel: String
+    public init(appName: String, channel: String) {
+        self.appName = appName
+        self.channel = channel
+    }
```
travis79, line 140: This `init` is provided for free by Swift and shouldn't be required, just having the public memebers defined should be enough.

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -124,14 +124,23 @@ public extension Notification.Name {
...
+
+    public let appName: String
+    public let channel: String
```
travis79, line 143: These are both public by default since the struct is `public`, so I think `public` modifiers on the struct members is redundant.

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -124,14 +124,23 @@ public extension Notification.Name {
 /// This struct is used during in the `create` method to point `Nimbus` at the given `RemoteSettings` server.
 ///
 public struct NimbusServerSettings {
-    let url: URL
+    public init(url: URL) {
+        self.url = url
+    }
```
travis79, line 129: Same comment about getting the `init` for free from Swift, I don't understand why this is necessary if the struct is already declared public?

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -124,14 +124,23 @@ public extension Notification.Name {
...
+
+    public let url: URL
```
travis79, line 131: Same comment as before, this defaults to public if the struct is public.

travis79 (changes requested):
I'm not sure why the `public init`'s were added, can you help me understand what you are trying to accomplish?

travis79:
See "Memberwise Initializers for Structure Types" here: https://docs.swift.org/swift-book/LanguageGuide/Initialization.html

components/nimbus/ios/Nimbus/NimbusApi.swift
travis79: I take this one back, the default modifier in Swift is `internal` so maybe `public` is needed.

components/nimbus/ios/Nimbus/NimbusApi.swift
travis79: Retracted, default access is `internal` so `public` here makes sense.

travis79 (approved):
👍 from me after our conversation (and trying it out without the `public init`)

jhugman:
For context for archivists:

From the Swift Book chapter on [Access Control][1]:

> The default memberwise initializer for a structure type is considered private if any of the structure’s stored properties are private. Likewise, if any of the structure’s stored properties are file private, the initializer is file private. Otherwise, the initializer has an access level of internal.
>
> As with the default initializer above, if you want a public structure type to be initializable with a memberwise initializer when used in another module, you must provide a public memberwise initializer yourself as part of the type’s definition.

We had a similar problem with this in uniffi.

[1]: https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#ID21

travis79:
> For context for archivists:
> 
> From the Swift Book chapter on [Access Control](https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#ID21):
> 
> > The default memberwise initializer for a structure type is considered private if any of the structure’s stored properties are private. Likewise, if any of the structure’s stored properties are file private, the initializer is file private. Otherwise, the initializer has an access level of internal.
> > As with the default initializer above, if you want a public structure type to be initializable with a memberwise initializer when used in another module, you must provide a public memberwise initializer yourself as part of the type’s definition.
> 
> We had a similar problem with this in uniffi.

TIL, thanks!

PR #3992, opened 2021-03-25
[SDK-209] Adds Glean integration to Nimbus for iOS

travis79:
This adds the Glean integration, to the same extent as we have in Android-Components, currently.  We can't quite use the Rust bindings yet, but moving the metrics.yaml here will be the first part of that work anyway.

### DIFF
--- a/components/nimbus/ios/Nimbus/Nimbus.swift
+++ b/components/nimbus/ios/Nimbus/Nimbus.swift
@@ -23,9 +23,7 @@ public class Nimbus: NimbusApi {
         return queue
     }()
 
-    internal init(nimbusClient: NimbusClientProtocol,
-                  errorReporter: @escaping NimbusErrorReporter)
-    {
+    internal init(nimbusClient: NimbusClientProtocol, errorReporter: @escaping NimbusErrorReporter) {
         self.errorReporter = errorReporter
         self.nimbusClient = nimbusClient
     }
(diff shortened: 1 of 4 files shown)


components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -23,9 +23,7 @@ public class Nimbus: NimbusApi {
         return queue
     }()
 
-    internal init(nimbusClient: NimbusClientProtocol,
-                  errorReporter: @escaping NimbusErrorReporter)
-    {
+    internal init(nimbusClient: NimbusClientProtocol, errorReporter: @escaping NimbusErrorReporter) {
```
travis79, line 26: The linter complained about this line, otherwise I wouldn't have touched it...

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -49,23 +47,68 @@ private extension Nimbus {
...
-        // TODO: https://jira.mozilla.com/browse/SDK-209
+        // We need to update the experiment enrollment annotations in Glean
+        // regardless of whether we recieved any events. Calling the
+        // `setExperimentActive` function multiple times with the same
+        // experiment id is safe so nothing bad should happen in case we do.
         let experiments = getActiveExperiments()
-        experiments.forEach { experiment in
-            Glean.shared.setExperimentActive(experimentId: experiment.slug, branch: experiment.branchSlug, extra: nil)
+        recordExperimentTelemetry(experiments)
+
```
jhugman, line 69: Suggest we add an empty check for events after recording experiment enrollment.

```swift
guard let events = events, !events.isEmpty else {
    return
}
```

It's an optimization for `recordExperimentEvents`, but would make `notifyOnExperimentsApplied` less chatty.

jhugman (approved):
This is lovely. Thank you.

components/nimbus/ios/Nimbus/Nimbus.swift
travis79: So we don't notify when experiments were applied if there wasn't any enrollment changes?  If so, I can just move the notify inside of the `if let events = events` block to accomplish the same thing.

travis79: As for optimizing things, it might save pushing a function call onto the stack, but the `for in` loop shouldn't have much overhead on an empty array.

PR #3995, opened 2021-03-26
Allow the sync manager to accept local encryption keys to pass to engines

mhammond:
I've tested this change on Fenix and it seems to work OK.

This is a breaking change because I can't work out how to treat the map as optional - the `optional` keyword doesn't work:

> Field labels (required/optional/repeated) are not allowed on map fields.

And [this reference](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-protobuf/kotlinx-serialization-protobuf/kotlinx.serialization.protobuf/index.html) says:

> Lists are represented as repeated fields. Because format spec says that if the list is empty, there are no elements in the stream with such tag, you must explicitly mark any field of list type with default = emptyList(). Same for maps. 

I'll reference an a-c PR to add support for unbreaking this breaking change.

cc @rfk just in-case he has a secret protobuf dossier that might make this slightly less painful.

### DIFF
--- a/components/sync_manager/android/src/main/java/mozilla/appservices/syncmanager/SyncParams.kt
+++ b/components/sync_manager/android/src/main/java/mozilla/appservices/syncmanager/SyncParams.kt
@@ -99,6 +99,25 @@ data class SyncParams(
     val enabledChanges: Map<String, Boolean>,
 
     /**
+     * A map of local encryption keys to use for engines. Only some engines
+     * use local encryption, so it's expected that many engines will not have
+     * entries. If an engine that requires a key does not find one, it will
+     * fail to sync.
+     *
+     * Note that this is for local encryption in the local databases and is not
+     * at all related to Sync encryption. The keys are needed to sync because
+     * sync will need to decrypt records from the local store before
+     * re-encrypting them using sync encryption to store on the server, and
+     * vice-versa, records read from the server will be decrypted using sync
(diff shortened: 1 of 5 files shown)


rfk:
> a secret protobuf dossier that might make this slightly less painful.

Nope, sorry, lists and optional fields in protobuf are a known pain point. See e.g. my periodic snark whenever I [port some protobuf bits to uniffi](https://github.com/mozilla/application-services/pull/3900#discussion_r586323317).

You could replace the whole protobuf interface with uniffi records I suppose :-)

components/sync_manager/android/src/main/java/mozilla/appservices/syncmanager/SyncParams.kt
```diff
@@ -98,6 +98,25 @@ data class SyncParams(
...
     val enabledChanges: Map<String, Boolean>,
 
+    /**
+     * A map of local encryption keys to use for engines. Only some engines
+     * use local encryption, so it's expected that many engines will not have
+     * entries. If an engine that requires a key does not find one, it will
+     * fail to sync.
+     *
+     * Note that this is for local encryption in the local databases and is not
+     * at all related to Sync encryption. They keys are needed to sync because
```
lougeniaC64, line 108: nit: "The"

lougeniaC64 (approved):

PR #3996, opened 2021-03-27
Add autofill to sync manager with integration tests

### LINKED ISSUE #3631: Configure the Sync Manager for Forms
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1828)
┆Story Points: 8
┆Sprint: SYNC - 2021-04-09

lougeniaC64:
Fixes #3631

### DIFF
--- a/components/autofill/src/db/store.rs
+++ b/components/autofill/src/db/store.rs
@@ -47,6 +47,12 @@ impl Store {
         })
     }
 
+    pub fn new_shared_memory(db_name: &str) -> Result<Self> {
+        Ok(Self {
+            store_impl: Arc::new(StoreImpl::new_shared_memory(db_name)?),
+        })
+    }
+
     pub fn add_credit_card(&self, fields: UpdatableCreditCardFields) -> Result<CreditCard> {
         self.store_impl.add_credit_card(fields)
     }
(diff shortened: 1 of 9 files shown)


testing/sync-test/src/auth.rs
```diff
@@ -266,12 +268,13 @@ impl TestClient {
         Ok(Self {
             fxa,
             test_acct: acct,
+            autofill_store: AutofillStore::new("./test-autofill-sync.db")?,
```
mhammond, line 271: Maybe use ` tempfile::tempfile` here? That might mean you need to store the filename so you can use it below

mhammond (approved):
This is great, thanks for the tests!

testing/sync-test/src/auth.rs
lougeniaC64: I've tried creating a temp db file with `tempfile::NamedTempFile` ([to get the path](https://docs.rs/tempfile/3.2.0/tempfile/index.html#design)) and ran into the error below which I'm assuming means the file was trashed before the first test object could be added to the database.

```
'add a1: SqlError(SqliteFailure(Error { code: ReadOnly, [extended_code: 1032](https://www.sqlite.org/rescode.html#readonly_dbmoved) }, Some("attempt to write a readonly database")))',
```
So then I tried calling [persist](https://docs.rs/tempfile/3.2.0/tempfile/struct.NamedTempFile.html#method.persist) on the temp file I created with `tempfile::NamedTempFile::new()`, which allows the tests to pass (of course, because it creates a `File` object from the `NamedTempFile`). 

Long story still long, I don't think using a temp file will work here but I'm open to other options.

mhammond: Right, yeah, that's a bit painful. How about we make memory DBs public - something like https://gist.github.com/mhammond/8b70dc0b94f93b3c25be4b723acf0467 ?

components/autofill/src/db/store.rs
```diff
@@ -142,14 +148,21 @@ impl StoreImpl {
...
+    /// Creates a store backed by an in-memory database with its own memory API (required for unit tests).
     #[cfg(test)]
     pub fn new_memory() -> Self {
         Self {
             db: Mutex::new(crate::db::test::new_mem_db()),
         }
     }
 
+    // Creates a store backed by an in-memory database that shares its memory API (required for autofill sync tests).
+    pub fn new_shared_memory(db_name: &str) -> Result<Self> {
```
lougeniaC64, line 160: @mhammond I left the above `test::new_memory` function in place and added this `new_shared_memory` function because the autofill engine unit tests need the separate memory API provided by the `new_mem_db` function while the autofill sync tests need to share the same api. Anyway, let me know if this change makes sense and I'll merge.

PR #4002, opened 2021-03-31
remote-settings-client spike

dmose:
Doesn't compile yet, but it's coming along...


dmose:
CC @leplatrem 

OK, things are now compiling.  The tests don't pass, because the remote-settings-client uses HTTP differently than our current hand-coded thing, so the mocks are wrong.  

This is fine, I think.  We've now validated that there's no architectural problems that prevent us from getting the two libraries to interface together at a code level, and I've also gotten my head around remote-settings-client, and how it would/could interact with the nimbus-sdk.  That's more or less what was hoped for in this spike.

We could conceivably clean this up and use it as a starting point when it comes time for the real code to be written, because I don't think I cut too many nasty corners in doing this.  But we can make that decision when the time comes...

PR #4003, opened 2021-03-31
autofill: add `delete_all_credit_cards`, `delete_all_addresses` and `delete_everything`

### LINKED ISSUE #3964: deleteAll methods for autofill component
Currently there are methods to delete individual addresses and credit cards, so it's impossible to perform a "wipe all credit cards/addresses" operation in a transactional way.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2114)
┆Sprint: Backlog

mhammond:
Fixes #3964

### DIFF
--- a/components/autofill/src/autofill.udl
+++ b/components/autofill/src/autofill.udl
@@ -111,6 +111,9 @@ interface Store {
     boolean delete_credit_card(string guid);
 
     [Throws=Error]
+    void delete_all_credit_cards();
+
+    [Throws=Error]
     void touch_credit_card(string guid);
 
     [Throws=Error]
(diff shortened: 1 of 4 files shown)


mhammond:
Note that when this executes, every card and address will be deleted from every device - the UX behind this will need to be very clear, and to be quite honest, I can't imagine why a user would ever want that to happen - they seem more likely to want to wipe *this* device only. So I'm going to mark this as a draft while we discuss this, because it seems like it might be a footgun. Eg, maybe we should refuse to delete if sync is configured, thereby forcing the app to disconnect first?

lougeniaC64:
> Eg, maybe we should refuse to delete if sync is configured, thereby forcing the app to disconnect first?

FWIW I wouldn't trust UI messaging with this. I think refusing to `delete all` when sync is enabled provides the necessary distinction between "delete all from this device" vs "delete all from all of my synced devices". If we don't want to or can't do that, there should be some other UI element or mechanism that prevents "delete all" from being "delete all from all of my devices" by default.

mhammond:
As discussed, we probably don't want this as written and the only identified requirement is testing.

PR #4004, opened 2021-03-31
Add "Converting an existing Component to use UniFFI" howto [ci skip]

rfk:
### [Rendered View](https://github.com/mozilla/application-services/blob/document-uniffiying-a-component/docs/howtos/converting-a-component-to-uniffi.md)

@lougeniaC64 @mhammond, pinging you both for review here because I hope you will both find yourselves either doing this or walking someone else through it in the near-to-mid-term future.


docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+and tools. The `fxa-client` crate may serve as a good example.
+
+You can view the generated documentation by running:
+
+```
+cargo doc --no-deps --open
+```
+
+In future, we intend to automatically extract documentation from the Rust code
+and make it easily available to consumers of the generated bindings.
```
rfk, line 204: What I would like to write here is "run this cool script to copy the doc comments over into your UDL file" but I'm not sure that can be run by anyone but me currently. I'll take a look at what would be required to let that be runnable by someone else...

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
+# Converting an existing Component to use UniFFI
+
+When we started building the components in this repo, exposing Rust code to
+Kotlin and Swift was a manual process and each component had its own
+hand-written FFI layer and foreign-language bindings.
+
+As we've gained more experience with building components in this way, we've
+started to automated bindings generation and capture best practices in a
+tool called [UniFFI](https://mozilla.github.io/uniffi-rs/), which is the
+current recommended approach when [adding a new component from scratch](
```
mhammond, line 10: trivial nit that's probably not even correct - 'currently'?

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+          * Higher-level hand-written Swift that wraps the FFI.
+
+The goal here is to replace much of the hand-written wrapper layers with autogenerated
+code:
+
+* The `./ffi/` crate will disappear entirely, its work is automated by UniFFI
+* The low-level `LibComponentNameFFI.kt` file will disappear entirely, as will some of the
+  code that converts it back into nice high-level Kotlin classes and interfaces.
+      * Some of the hand-written Kotlin code may remain, if it provides functionality that
+      cannot be implemented in Rust.
```
mhammond, line 63: nit: indentation (I only mention this because I'm too lazy to check whether it is rendered OK or not - ignore if so)

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+                  * Higher-level hand-written Kotlin that wraps the FFI.
+    * `ios/`
+        * `<component_name>/`
+          * `Rust<ComponentName>API.h` (low-level bindings to the C-style FFI)
+          * Higher-level hand-written Swift that wraps the FFI.
+
+The goal here is to replace much of the hand-written wrapper layers with autogenerated
+code:
+
+* The `./ffi/` crate will disappear entirely, its work is automated by UniFFI
```
mhammond, line 59: Unless I misunderstand, if you are only doing a partial conversion, this might remain, right? If that's true, I think it might be worth setting the scene for that above - there are lots of places below which assume "will be removed", so I don't think it's worth trying to qualify that everwhere, but calling it out as an intro to this section might make sense.

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+* The low-level `LibComponentNameFFI.kt` file will disappear entirely, as will some of the
+  code that converts it back into nice high-level Kotlin classes and interfaces.
+      * Some of the hand-written Kotlin code may remain, if it provides functionality that
+      cannot be implemented in Rust.
+* The low-level `RustComponentNameAPI.h` file will disappear entirely, as will some of the
+  code that converts it back into nice high-level Swift classes and interfaces.
+      * Some of the hand-written Kotlin code may remain, if it provides functionality that
```
mhammond, line 66: ditto indentation and s/Kotlin/Swift/ here?

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+* You may find it useful to move all of the existing code into a sub-module named `internal`,
+  and then make a brand new `lib.rs` that imports or re-defines just the pieces it needs
+  in order to implement the interface from the `.udl` file. The `fxa-client` crate is an
+  example of a case where this worked out well, though of course your mileage may vary.
+
+* If the existing crate contains a file named like `<component_name>_msg_types.proto`, then
+  it was using Protocol Buffers to serialize data to pass over the FFI. The message types
+  defined in the `.proto` file will need to be converted into `dictionary` or `enum` definitions
+  in your `.udl` file. See the section below for more details.
+
```
mhammond, line 149: I'd consider adding a new point along the lines of "Don't be too concerned about making small changes to the 'spelling' of the API - if there are small differences in the names generated by uniffi vs what is in the hand-written FFI, it's probably in everyones interest to take that hit and update the consumers." - I see you do make that point a bit later, but IMO it's worth introducing it here.

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+* Delete `./src/<component_name>_msg_types.proto`.
+* Delete `./src/mozilla.appservices.<component_name>.protobuf.rs`, which is generated from the `.proto` file.
+* Remote `prost` and `prost-derive` from the crate's dependencies.
+* Delete the crate from the list in `/tools/protobuf_files.toml`.
+
+If it happen to find that you've deleted the last crate from the list in `protobuf_files.toml`,
+congratulations! You've successfully removed protocol buffers from this repo entirely, and should
+file a bug to track the complete removal of protobuf from our tooling and dependency chain.
+
```
mhammond, line 190: Ditto the "Don't be concerned about small naming changes here" or similar?

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -0,0 +1,287 @@
...
+If there are existing Kotlin tests for the component, the next step is to get those passing:
+
+* `./gradlew <component_name>:test`
+
+It is normal and expected for the autogenerated bindings to be subtly different from the
+previous hand-written ones. For example, UniFFI insists on CAMEL_CASING variant names in
+Kotlin enums while the hand-written code does not to this consistently. Some components
+also have small naming differences between the Rust code and the hand-written Kotlin bindings,
+which UniFFI will not allow.
```
mhammond, line 232: If you take the previous "don't be afraid" suggestions, maybe add here "As noted above"?

mhammond (approved):
This is awesome, thanks.

docs/howtos/converting-a-component-to-uniffi.md
rfk: That's a good point, I hadn't really considered partial conversions here...I guess that would be for the case if there's something that UniFFI doesn't support yet, and you need to write your own `pub extern "C"` functions?

I think I would still argue that those functions should live in the main crate and be part of `lib.rs` and your top-level public API, rather than squirreled away in a separate `./ffi/` crate. But I could add some words to this effect if you think it would be helpful?

(In fact I don't really recall why we settled on this convention of having a separate `ffi` crate at all to be honest...)

mhammond: Yes, that would be perfect - just a small note saying "things which are unable to be converted should be moved into lib.rs" would be perfect and about all the detail needed here.

docs/howtos/converting-a-component-to-uniffi.md
rfk: @mhammond what do you think, is it worth me trying to make the automagical "copy docstrings from the Rust code into the UDL" functionality from https://github.com/mozilla/uniffi-rs/pull/416 into something more generally usable, or is that just getting ahead of ourselves?

mhammond: In one day? :) I think would be valuable, but not at all urgent. Maybe leaving a breadcrumb here would be worthwhile?

docs/howtos/converting-a-component-to-uniffi.md
rfk: it did not render ok, thanks for noticing!

docs/howtos/converting-a-component-to-uniffi.md
rfk: I'm going to add this under the earlier "write a draft of the .udl" section and then reinforce it again at several points further down.

rfk:
I'm going to go ahead and merge this (with nits addressed) so that I don't leave it hanging around while I'm on PTO, but @lougeniaC64 any feedback you may have is still very welcome.

PR #4009, opened 2021-04-02
[SDK-246] Make `channel` targeting comparision case-insensitive.

travis79:


dmose:
Looks good to me; r=dmose, and I'll merge it.

@travis79 Could you either document this somewhere (presumably DocsHub?) so users know what to expect.. or else file a ticket for that to happen later?

dmose (approved):

PR #4018, opened 2021-04-12
[SYSC-2053] More specific lookup errors

### LINKED ISSUE #3868: autofill: better errors when no record is found?
If you query for an address or credit-card using a GUID that doesn't exist, you get back a `SqlError(QueryReturnedNoRows)` - I wonder if it is worth a specific error for this case? My thinking is that most `SqlError`s are not a result of inputs to the API, but this one is.

(If we decide the status-quo isn't ideal, we then should work out whether we should return null or create a new specific error type.

cc @lougeniaC64 , WDYT?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2053)
┆Epic: a-s storage components (backlog)
┆Sprint: SYNC - 2021-04-16

bendk:
Fixes #3868

- Added NoSuchRecord error to autofill.
- Return it from get_address() and get_credit_card()
- Added tests for this case

This is my first PR, hopefully I sort-of followed the correct procedures.  A couple notes:

 - I didn't add any changelog entries, please tell me if I should do something there
 - I wasn't sure about what data to store inside NoSuchRecord.  I stored the missing GUID there, is that good?
 - Related to that, I wasn't sure how to format the error message.  Right now it's `No record with guid exists: [guid]`.  Should I have included the GUID in the message?
 - I can't run `all_tests.sh` yet, but `cargo test` does pass.


components/autofill/src/db/addresses.rs
```diff
@@ -352,6 +356,19 @@ mod tests {
         assert!(get_address(&db, &saved_address.guid).is_err());
     }
 
+    #[test]
+    fn test_address_missing_guid() {
+        let db = new_mem_db();
+        let guid = Guid::random();
+        let result = get_address(&db, &guid);
+
+        assert!(
```
lougeniaC64, line 365: This `assert` can be replaced with an `assert_eq` for brevity.

```
assert_eq!(
            result.unwrap_err().to_string(),
            Error::NoSuchRecord(guid.to_string()).to_string()
        );
```

lougeniaC64: This and the one below for credit cards.

components/autofill/src/error.rs
```diff
@@ -36,6 +36,9 @@ pub enum Error {
 
     #[error("Missing local encryption key")]
     MissingEncryptionKey,
+
+    #[error("No record with guid exists: {0}")]
```
lougeniaC64, line 40: This formatting is good.

components/autofill/src/db/addresses.rs
```diff
@@ -97,9 +97,13 @@ pub(crate) fn get_address(conn: &Connection, guid: &Guid) -> Result<InternalAddr
         WHERE guid = :guid",
         common_cols = ADDRESS_COMMON_COLS
     );
-
-    let address = tx.query_row(&sql, &[guid], |row| Ok(InternalAddress::from_row(row)?))?;
-
+    let result = tx.query_row(&sql, &[guid], |row| Ok(InternalAddress::from_row(row)?));
+    let address = match result {
+        Err(rusqlite::Error::QueryReturnedNoRows) => {
+            return Err(Error::NoSuchRecord(guid.to_string()));
```
lougeniaC64, line 103: You can abbreviate this line as follows:

```
let address = match result {
        Err(rusqlite::Error::QueryReturnedNoRows) => Err(Error::NoSuchRecord(guid.to_string()))?,
        _ => result?,
    };
```

lougeniaC64: Same for credit cards

lougeniaC64:
@bendk I left some feedback but otherwise looks good! I'll remove the line about `automation/all_tests.sh` from the PR template. Generally speaking, CI is the decider and `all_tests` is just a tool to shorten the feedback loop.

We usually add `Fixes #[Github issue number]` at the beginning of our PR descriptions if there is an associated Github issue. For this PR it's #3868. 

It's best practice to add a summary of your changes to `CHANGES_UNRELEASED.md` but it's not always necessary. Generally speaking we add entries when our patches will affect how consumers use our code. But it's also a way to communicate what we've changed even when the changes are minor.

Because this code changes how we handle the "get by GUID" functions when a record isn't found and exposes a new error type, it's worth noting. You can use [this entry](https://github.com/mozilla/application-services/blob/main/CHANGELOG.md#autofill) as a template and replace the "Breaking Changes" header with "What's Changed". I wouldn't consider this a breaking change as we are swapping error types when a specific error occurs (replacing `rusqlite::Error::QueryReturnedNoRows` occurrences with `Error::NoSuchRecord`) but not changing the signature of the functions. FWIW I'd consider this a minor change in the semver world.

bendk:
Thanks for the feedback and suggestions -- especially the rust tricks.  I think that last commit should address them.

bendk:
One more question: if there's an error, we don't call `tx.commit()` in either function.  Is this correct?

lougeniaC64:
> One more question: if there's an error, we don't call `tx.commit()` in either function. Is this correct?

@bendk That's right. That was the case before your changes and will continue to be the case after your changes **except** when a `QueryReturnedNoRows` error occurs. But your questions points to a larger issue which is the uselessness of the transactions in both functions. I think this resulted from some copypasta on my part. If you want, you can remove the commit calls and replace the transactions (via the `tx` variable) with connections (via the passed in `conn` param) in both functions. I can also do this after you merge.

Also, the swiftlint build failed in CI for unrelated reasons, but you should resolve the clippy build issues.

mhammond:
>  But your questions points to a larger issue which is the uselessness of the transactions in both functions

FWIW, I agree 100% with the value in this specific case, but we have found that read operations can be significantly faster when done inside a transaction. IOW, I think removing it probably makes sense in this case (a single query), but not in all cases.

components/autofill/src/db/addresses.rs
bendk: I wanted to use this one, but I think that was the source of the clippy error. I turned it back to a regular return, tell me if there's a method that's both compact and approved by clippy.

mhammond:
```
    let address = conn
        .query_row(&sql, &[guid], |row| Ok(InternalAddress::from_row(row)?))
        .map_err(|e| match e {
            rusqlite::Error::QueryReturnedNoRows => Error::NoSuchRecord(guid.to_string()),
            e => e.into(),
        })?;
    Ok(address)

```
Opinions on Rust are very subjective, especially for newcomers - I remember my brain hurting alot with chained `.map()` calls, so whether the code above is *better* is probably contested. TBH, I'm fine either way, so if you do happen to think it's better then feel free to take it, but if you think your existing code is clearer feel free to keep that too.

bendk: Thanks for all the help with this.  It feels kind of silly to go back and forth on such a small amount of code, but it's helpful for me to learn more rust idioms.  This version seems great to me, because map_err() describes exactly what we're doing.

lougeniaC64 (approved):
LGTM! 🎉

PR #4024, opened 2021-04-14
Execute `brew update reset` to try and fix ktlint missing

mhammond:


jdragojevic (approved):

mhammond:
CI now seems to be seeing "real" lint failures - maybe a new version is used now? cc @jhugman in case he has any insights here.

PR #4027, opened 2021-04-15
Ensure databases created before we encrypted card numbers still work.

### LINKED ISSUE #4026: Perform DB migration for encrypted CC number work
As @grigoryk mentions in https://github.com/mozilla/application-services/issues/4015#issuecomment-819879441, there's a chance some Fenix Nightly users have a DB schema pre the CC encryption done in 43c0835d374038f8c1f840a63a2f93e210e48d73. We probably should migrate the DB - we can't reasonably do that without dropping the CC numbers, but that's probably better than having those Fenix profiles forever broken for CC numbers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2164)
┆Sprint: Backlog

mhammond:
We just drop the data in the cards table because it's almost impossible
they have actual data, but we do this so we can recreate the table with
the correct schema.

Fixes #4026.

CC @grigoryk to check dropping the table is fine and @bendk because this adds the core migration framework he can leverage.

### DIFF
--- a/components/autofill/src/db/schema.rs
+++ b/components/autofill/src/db/schema.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::error::Result;
-use rusqlite::Connection;
+use rusqlite::{Connection, NO_PARAMS};
 
 pub const ADDRESS_COMMON_COLS: &str = "
     guid,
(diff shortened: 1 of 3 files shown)


components/autofill/src/db/schema.rs
```diff
@@ -90,13 +111,82 @@ fn create(db: &Connection) -> Result<()> {
...
+    }
+
+    #[test]
+    fn test_upgrade_version_0() {
+        let db = new_mem_db();
+        // Manually hack things back to where we had to migrate from
+        // version 0.
+        // We don't care what the old table actually has because we drop it
+        // without a migration.
+        db.execute_batch(
```
lougeniaC64, line 173: huge nit: the alignment of this query seems off, particularly the quotation marks (I said huge)

lougeniaC64 (approved):

components/autofill/src/db/schema.rs
mhammond: I know! Clippy insists on that first quote being there, which I don't understand. It does seem fine with the end quote being on the same line, so I guess that's something.

PR #4029, opened 2021-04-15
Support scrubbing encrypted data

bendk:
This allows scrubbing encrypted data from the local storage in order to recover from lost/corrupted encryption keys.  Right now the only piece of encrypted data is the cc_number_enc field.

It also updates the syncing logic for scrubbed data.  When we see that a local record has scrubbed data, then we always copy the remote version of that field (any any other fields that are different).

  - Note:
- [X] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -7,4 +7,7 @@
 The main credit-cards is dropped and recreated to ensure already existing
 databases will continue to work.
 
+Added support to scrub encrypted data to handle lost/corrupted client keys.
+Scrubbed data will be replaced with remote data on the next sync.
+
 [Full Changelog](https://github.com/mozilla/application-services/compare/v75.1.0...main)
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -4,10 +4,10 @@
 
 ## Autofill
 
-The main credit-cards is dropped and recreated to ensure already existing
-databases will continue to work.
+- The main credit-cards is dropped and recreated to ensure already existing
+  databases will continue to work.
 
-Added support to scrub encrypted data to handle lost/corrupted client keys.
-Scrubbed data will be replaced with remote data on the next sync.
+- Added support to scrub encrypted data to handle lost/corrupted client keys.
+  Scrubbed data will be replaced with remote data on the next sync.
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v75.1.0...main)
--- a/components/autofill/src/db/schema.rs
+++ b/components/autofill/src/db/schema.rs
@@ -170,8 +170,7 @@ fn upgrade(db: &Connection, from: i64) -> Result<()> {
             DROP TRIGGER credit_cards_tombstones_afterinsert_trigger;
             DROP TABLE credit_cards_data;
             ALTER TABLE new_credit_cards_data RENAME to credit_cards_data;
-            ",
-        )?;
+            ")?;
         // NOTE: at this point several triggers have been droped and not recreated, but that's okay
         // because they will be recreated when we execute CREATE_SHARED_TRIGGERS_SQL
     }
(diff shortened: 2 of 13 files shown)


bendk:
@mhammond @lougeniaC64 please review this when you get the chance.  I tried to add you as reviewers directly, but I don't think I have the perms yet.

components/autofill/src/db/store.rs
```diff
@@ -105,6 +105,16 @@ impl Store {
         self.store_impl.touch_address(guid)
     }
 
+    pub fn scrub_encrypted_data(&self) -> Result<()> {
+        // scrub the data on disk
+        self.store_impl.scrub_encrypted_data()?;
```
mhammond, line 110: This function should just call the impl, so please move the code below down into that impl.

mhammond (commented):
This is looking in good shape! Lougenia should give the final review - I don't need to have another look.

components/autofill/src/db/schema.rs
```diff
@@ -90,13 +111,205 @@ fn create(db: &Connection) -> Result<()> {
...
+        // a schema of zero implies "new database".
+        // These databases have a `cc_number` but we need them to have a
+        // `cc_number_enc` and `cc_number_last_4`.
+        // This was so very early in the Fenix nightly cycle, and before any
+        // real UI existed to create cards, so we don't bother trying to
+        // migrate them, we just drop the table so it's re-created with the
+        // correct schema.
+        db.execute("DROP TABLE IF EXISTS credit_cards_data", NO_PARAMS)?;
+    } else if from == 1 {
+        db.execute_batch(
```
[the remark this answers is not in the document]
bendk: I definitely would have done the simple way if I had thought that one though.  At this point, my feeling is also :shrug:, but I'm leaning towards not changing the code.

components/autofill/src/sync/engine.rs
```diff
@@ -75,6 +75,14 @@ impl<T> ConfigSyncEngine<T> {
         let key = format!("{}.{}", self.config.namespace, tail);
         crate::db::store::delete_meta(conn, &key)
     }
+    // Prepare to refetch all records from the server
+    pub fn refetch_server_records(&self) -> Result<()> {
```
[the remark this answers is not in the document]
bendk: I struggled with naming this one.  I'm going to push a change that names it `reset_local_sync_data()` but I'm definitely open to suggestions.  On this topic, when does the next fetch happen?  Does calling reset_storage() or put_meta() cause a server fetch to be scheduled early?

components/autofill/src/db/store.rs
bendk: The impl definitely seems like a better place, but `create_engine()` inputs `Arc<StoreImpl>` not `&StoreImpl` , and I couldn't figure out how to do that in the StoreImpl method. @lougeniaC64 maybe we could discuss this one some more today.

lougeniaC64: @mhammond Were you thinking something like this in the `scrub_encrypted_data` impl function?

```
crate::sync::credit_card::create_engine(Arc::new(*self)).reset_local_sync_data()?;
// also tried self.clone() with StoreImpl implementing Clone
```

This won't compile because the `writer` field of `AutofilDb` doesn't implement Clone.

components/autofill/src/db/schema.rs
```diff
@@ -90,13 +111,207 @@ fn create(db: &Connection) -> Result<()> {
...
+        db.execute_batch(select_name)
+            .expect("select should now work");
+    }
+
+    #[test]
+    fn test_upgrade_version_1() -> Result<()> {
+        let db = new_mem_db();
+        // Go back to version 0 of the credit_cards_data table and insert a row
+        db.execute_batch("
+DROP TABLE credit_cards_data;
```
lougeniaC64, line 269: nit: fix the alignment of this and the below sql queries.

CHANGES_UNRELEASED.md
```diff
@@ -2,4 +2,12 @@
 
 # Unreleased Changes
 
+## Autofill
+
+The main credit-cards is dropped and recreated to ensure already existing
+databases will continue to work.
+
+Added support to scrub encrypted data to handle lost/corrupted client keys.
+Scrubbed data will be replaced with remote data on the next sync.
```
lougeniaC64, line 11: formatting nit: Add dashes before lines 7 and 10.

components/autofill/src/db/store.rs
bendk: This is maybe out of scope for this issue, but it feels like the right way to do this would be to create a `SyncEngineStorageImpl` directly and call a method on that.  However, that's not currently possible because `put_meta()` is implemented in `ConfigSyncEngine`.  Would it make sense to move the put/get/delete meta functionality into `SyncEngineStorageImpl`?  I guess we would still need to wrap the methods in `ConfigSyncEngine` but that wouldn't be too bad.

components/autofill/src/sync/engine.rs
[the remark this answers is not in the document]
mhammond: Nope, the scheduling is very simplistic and up to the app - which isn't completely ideal - if this was desktop, we could do more to - err - "encourage" a sync very soon. This is really just changing what happens on the next sync, but doesn't change anything about when that is.

components/autofill/src/db/store.rs
mhammond: right, I missed that detail. Just leaving it where it is is fine for now.  (It does seem a little heavy-handed that we need to create an engine just for that reset, but it's not a huge deal and isn't a reliability or performance concern so can be cleaned up later)

bendk:
I'm not sure what's going on with this one exactly, but I just pushed 2 more commits to `bendk:SYNC-2158`.  For some reason github hasn't added them to this PR yet though.

lougeniaC64 (dismissed):
LGTM! 🚀

lougeniaC64 (approved):

PR #4036, opened 2021-04-21
Move Nimbus.kt from Android Components

jhugman:
[This document][1] is a good place to discuss how we go from here.

This implements Option 2 from this proposal.

This and https://github.com/mozilla-mobile/android-components/pull/10144 fix [SDK-259][2]

[1]: https://docs.google.com/document/d/1nibSjM9Vslqj6GKVxePq6fkOp4Yg4euzGeVidHBUnC4/edit?usp=sharing
[2]: https://jira.mozilla.com/browse/SDK-259


grigoryk (dismissed):
Discussed this with @csadilek, and we don't think this code should move this code into A-S, for the following reasons:

- we have an established pattern of platform-specific things (like coroutine management/threading) living in the consuming libraries, e.g. A-C or in Swift code. Other parts of A-S follow this pattern and things work out well - e.g. the A-S Kotlin/Swift code are thin shims, but this breaks away from this pattern
- this introduces a cyclical dependency on the Observer Registry. We've been trying hard to avoid doing this, and working towards removing other cyclical dependencies (e.g. the Fetch client stuff). Let's not make our lives harder here :)

It seems to me that the primary reason for moving this code is to simplify landing PRs against it (in a-c), but that doesn't seem to outweigh the above architectural arguments. What other reasons are there?

If we did want to move some of this code here, let's aim to minimize the amount of logic that's present here? E.g. at the very least keep all of the threading stuff in an A-C wrapper code. But I'm not sure what this doing all that work will give us, to be honest.

jhugman:
While this PR isn't ready to land, it may be a good idea to get a big picture view of the intent.

[This document][1] is a good place to discuss how we go from here.

[1]: https://docs.google.com/document/d/1nibSjM9Vslqj6GKVxePq6fkOp4Yg4euzGeVidHBUnC4/edit?usp=sharing

jhugman:
@grigoryk 

Ok, so taking your review comments in order.

> Discussed this with @csadilek, and we don't think this code should move this code into A-S,

It'd be good if you and Christian could have had that discussion with me in the same room; it seems you both have missed a lot of context, especially around motivations of this work.

For your situational awareness, Nimbus is going to have roughly the same number of moving parts as another Glean: handwritten platform specific code, generated app and platform specific code, a generated FFI and some Rust. It's intended to be used in all parts of the stack: from GeckoView to Fenix UI. It's to be used from the UI thread, worker threads and from different processes.

In this quarter, features are being added that necessitate handwritten Kotlin and Swift; and next quarter, generating Kotlin and Swift for app developers to use directly.

From this week, the external API of Nimbus will include non-trivial non-UI parts of it written in platform specific code.

> we have an established pattern of platform-specific things living in the consuming libraries, e.g. A-C or in Swift code.
> Other parts of A-S follow this pattern and things work out well - e.g. the A-S Kotlin/Swift code are thin shims, but this breaks away from this pattern

Well, that's kinda sorta true: there is a pattern but it has plenty of exceptions.

In iOS: there is lots of Handwritten Swift code living in FxiOS being replaced by Rust + Handwritten Swift living in Application Services. There are AS components that _are_ just the FFI, but there are also components that are not. The FFI does not define the API.

Glean does not follow this pattern. It looks after its own threading, it has handwritten platform specific code, and none of it lives in AC or Firefox for iOS.

GeckoView does not follow this pattern. For some historical reasons, Android Components maintains a wrapper layer around GeckoView but GeckoView is nothing but Android specific things not living in the consuming libraries.

The pattern only seems like a pattern for components that are (deployed on Android) AND (whose APIs are simple enough for their FFI serves as the API too).

> this introduces a cyclical dependency on the Observer Registry. We've been trying hard to avoid doing this, and working towards removing other cyclical dependencies (e.g. the Fetch client stuff). Let's not make our lives harder here :)

I'm having trouble seeing your cyclical dependency between `Nimbus` and `ObserverRegistry`. To be clear, if one exists, this isn't the insummountable problem to moving the code as presented. I can see that at worst, there may be some code duplication if two versions of the `ObserverRegistry` ships, but again not insummountable.

> It seems to me that the primary reason for moving this code is to simplify landing PRs against it (in a-c), but that doesn't seem to outweigh the above architectural arguments. What other reasons are there?

The primary reason is that Application Services are writing code on both sides of the FFI: on one platform changes to the internals of a component can happen in the same PR; but on one platform it takes a release of the megazord by Application Services, and an upgrade process of to the new megazord by A-C.

This is adding weeks to development time, increasing development cost by increasing coordination costs. This friction has many downsides, and no upsides, for either the Application Services or Android Components teams, or indeed Nimbus on iOS or Android.

jhugman:
Status update: following zoom call between @jhugman, @csadilek @grigoryk and @pocmo 👍 

 * The 'circular dependency' is actually two separate classes of identical name, with differences in APIs e.g. two versions of an evolving API. We (mozilla-mobile) have decided that the tradeoff is either to do [the work of evolving APIs][1] or not allow it, in favour of velocity. In most cases, this is not allow A/S to depend on A/C. There are unforeseen exceptions to this (e.g. Concept Fetch) where we need to be more careful about evolving its API.
 * New code in Application Services (for the purpose of this work, Nimbus) shouldn't add any dependencies on the Android Components repo. `ClassLoader` separation on Android is done by task and activity, rather than module, so some effort should be taken to pass in the `ObserverRegistry` and the `localeTag`.
 * Thread naming, and coroutine lifecycle is important to Android Components, so coroutine scopes should be passed in.

[1]: https://wiki.eclipse.org/Evolving_Java-based_APIs

jhugman:
The Nimbus object in A/S now accepts a delegate, observer and deviceInfo struct in its constructor.

components/nimbus/uniffi.toml
```diff
@@ -1,3 +1,3 @@
 [bindings.kotlin]
-package_name = "org.mozilla.experiments.nimbus"
+package_name = "org.mozilla.experiments.nimbus.internal"
```
travis79, line 2: I'm a little perplexed why you added the `internal` to this, actually, I've been wondering if we should even keep this separate from the other components who are all `org.mozilla.appservices.componentName`.  But that's not something I feel strongly enough about to block this on.

travis79 (approved):

components/nimbus/uniffi.toml
jhugman: For the dumbest reasons: uniffi generates a `nimbus.kt` which was colliding with the `Nimbus.kt` file in the same package.

jhugman:
Added https://github.com/mozilla/application-services/issues/4053 to resolve `check-dependencies`.

components/nimbus/android/build.gradle
```diff
@@ -74,10 +68,21 @@ configurations {
...
 dependencies {
     jnaForTest "net.java.dev.jna:jna:$jna_version@jar"
     implementation "net.java.dev.jna:jna:$jna_version@aar"
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    implementation "androidx.annotation:annotation:$androidx_annotation_version"
+
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlin_coroutines_version"
+
+    implementation "org.mozilla.components:service-glean:$android_components_version"
```
pocmo, line 84: I wasn't aware that there are other modules here that also depend on `services-glean`. I thought `concept-fetch` was the only dependency (#3872). Potentially also something to rethink, but out of scope for this PR.

pocmo (approved):

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -0,0 +1,547 @@
...
+ * Small struct for info derived from the device itself.
+ */
+data class NimbusDeviceInfo(
+    val localeTag: String
+)
+
+/**
+ * Provide calling apps control how Nimbus fits into it.
+ */
+data class NimbusDelegate(
```
csadilek, line 234: nit: Is there a reason this is a data class? You don't need any of the generated functions, and `logger`/`errorReporter` are essentially behaviour passed down.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -0,0 +1,547 @@
...
+    @SuppressWarnings("TooGenericExceptionCaught")
+    private fun <R> withCatchAll(thunk: () -> R) =
+        try {
+            thunk()
+        } catch (e: Throwable) {
+            try {
+                errorReporter("Error in Nimbus Rust", e)
+            } catch (e1: Throwable) {
+                logger("Exception calling rust: $e")
+                logger("Exception reporting the exception: $e1")
```
csadilek, line 335: nit: see other PR, these used to be `errors` now they're just `infos`.

csadilek (approved):
👍 Just a nit below.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
jhugman: Good spot. This was an error.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
jhugman: Everything that used to report an error via the `logger.error()` is now doing it via an `errorReporter`. 

Except here where an exception is caught coming from `errorReporter`. If `errorReporter` is itself erroring, I'm guessing nimbus won't be the only one finding out about it, and that will be visible from more than just here.

components/nimbus/android/build.gradle
rfk: FWIW, we're very motivated to take this glean dependency at the Rust level rather than the Kotlin level, and I'm hopeful that when that work comes to fruition it will break the dependency cycle here.

PR #4041, opened 2021-04-27
WIP Enforce one experiment per feature (SDK-260)

dmose:
There's more to do here, both in terms of functionality as well as dirty hacks that need to be cleaned up, so keep that in mind if you decide to look into the code as it currently stands...

  - Note:

### DIFF
--- a/components/nimbus/src/enrollment.rs
+++ b/components/nimbus/src/enrollment.rs
@@ -611,6 +611,7 @@ impl<'a> EnrollmentsEvolver<'a> {
         updated_experiments: &[Experiment],
         existing_enrollments: &[ExperimentEnrollment],
     ) -> Result<(Vec<ExperimentEnrollment>, Vec<EnrollmentChangeEvent>)> {
+
         // XXX fix up name overrides of params for clarity; check on clippy lint
         let mut enrollment_events = vec![];
         let existing_experiments = map_experiments(&existing_experiments);
@@ -622,15 +623,12 @@ impl<'a> EnrollmentsEvolver<'a> {
         all_slugs.extend(updated_experiments.keys());
         all_slugs.extend(existing_enrollment_map.keys());
 
-        // XXX go through here and make sure we're only calling
-        // evolve_enrolment once per update?
-
         // Step 1:
         // make a hash map of feature_id to enrolled_experiment.slug
         // XXX need to make sure updated feature_ids are inserted into this
         // as we update
 
-        let locally_enrolled_feature_ids =
+        let mut locally_enrolled_feature_ids =
             map_locally_enrolled_feature_ids(existing_enrollments, &existing_experiments);
 
         let mut updated_enrollments = Vec::with_capacity(all_slugs.len());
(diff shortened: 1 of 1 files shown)


components/nimbus/src/enrollment.rs
```diff
@@ -601,30 +614,116 @@ impl<'a> EnrollmentsEvolver<'a> {
...
+                if locally_enrolled_feature_ids.get(feature_id) == None {
+                    log::debug!("this feature is locally unused");
+
+                    let updated_enrollment = self.evolve_enrollment(
+                        is_user_participating,
+                        existing_experiments.get(slug).copied(),
+                        updated_experiments.get(slug).copied(),
+                        existing_enrollment_map.get(slug).copied(),
+                        &mut enrollment_events,
+                    )?;
```
jhugman, line 679: Just thinking this through: By this time, we know that this is we're in an ongoing experiment that we're not enrolled in.

1. ongoing experiment because it's in `updated_experiments`
2. if we were enrolled in it, then this feature id would be taken.

components/nimbus/src/enrollment.rs
```diff
@@ -601,30 +614,116 @@ impl<'a> EnrollmentsEvolver<'a> {
...
-                updated_enrollments.push(enrollment);
+            // if let Some(_enrollment) = updated_enrollment {
+            // What has changed? If we're still enrolled, do nothing.
+
+            // If not, remove the feature id - it's usable now.
+
+            // updated_enrollments.push(enrollment); XXX NOT NEEDED HERE, I
+            //if enrollment.status != EnrollmentStatus::Enrolled
+            // }
+        }
```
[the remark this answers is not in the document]
jhugman: i.e. if this loop is only about unenrolling from existing experiments, so we can free up `feature_id`s, then perhaps we should only loop through `existing_enrollments`.

components/nimbus/src/enrollment.rs
```diff
@@ -601,30 +614,116 @@ impl<'a> EnrollmentsEvolver<'a> {
...
+        // Step 3: now we can add new experiments that avoid the features that
+        // already being experimented upon.
+
+        // the previous calls to evolve_enrollments will have prepopulated
+        // so...
+        enrollment_events.clear();
+
+        for slug in all_slugs {
+            let updated_experiment = updated_experiments.get(slug).copied();
+            if let Some(unwrapped_experiment) = updated_experiment {
```
jhugman, line 665: This loop seems to be about enrolling into experiments from `updated_experiments`. Suggest we just loop through those.

components/nimbus/src/enrollment.rs
jhugman: (2) is important if we want want to call `evolve_enrollment` exactly once per experiment slug.

components/nimbus/src/enrollment.rs
```diff
@@ -601,30 +614,116 @@ impl<'a> EnrollmentsEvolver<'a> {
...
+                        );
+                        locally_enrolled_feature_ids.insert(feature_id.clone(), slug.clone());
+                    }
+                } else {
+                    // Does this experiment have a feature id that we've already used?
+                    log::debug!(
+                        "feature_id {} already used by experiment, marking {} NotEnrolled",
+                        feature_id,
+                        slug
+                    );
```
jhugman, line 700: Here, we only know that the `feature_id` was used by _an_ experiment. It might have been used by _this_ experiment.

We should check either `locally_enrolled_feature_ids.get(feature_id) == Some(slug)` or `existing_enrollment_map.get(slug).is_some()` before registering a non_enrollment.

jhugman: For clarity, I think I prefer the former.

jhugman (commented):
I've added some more comments. lmk if you want to pair some more.

jhugman:
This is being going round my head, so getting it out by writing it down. It may not be helpful, since it mostly re-iterates our existing understanding and you probably got the rest of the way in your own way.

Step 1 uses the enrollments that are `::Enrolled` to build up a map of `feature_id` => `experiment_slug`, i.e. `locally_enrolled_feature_ids`.

The first loop (Step 2 in our nomenclature) calls `evolve_enrollment` on all slugs in `existing_enrollments`. If the previous enrollment was `::Enrolled` AND the new enrollment isn't, then we can can count the feature id it works on as no longer active, and remove it the feature_id from the map.

The second loop (Step 3), we check if the feature_id isn't already taken, then calls `evolve_enrollment` on all slugs in `updated_experiments`, except ones that are also in `existing_enrollments`. i.e. call `evolve_enrollment` and get a second enrollment. If the first one was `None` or wasn't `::Enrolled` AND the new enrollment is, then we can add the feature_id to the map.

Lastly, we generate a map of `feature_id` => `FeatureConfig`. We can do this by looping through the `updated_enrollments` looking for `EnrollmentStatus::Enrolled`, getting the `branch` slug, then finding `Experiment` and grabbing the `FeatureConfig` for the correct branch. This is all made a lot easier by only having one `feature_id` per experiment.

For completion, we _might_ need a third loop that goes through all `existing_experiments`, except ones that are in `updated_experiments` OR `existing_enrollments`. If you go through the `evolve_enrollment` `match` expression, looking at the `(Some(_), None, None)` case I suspect that this does some cleanup/garbage collection but probably doesn't do any unenrollment. *Update:* it causes an internal error, so there's likely no point in doing this.

Further comments observations and questions: 

* I think we need to keep track of the `updated_enrollments`, and including enrollments with a `FeatureAlreadyInvolvedInAnExistingExperiment` enrollment status. This method shows that we re-generate the enrollments array each time, and every experiment is accounted for by an enrollment. 
* by thinking in terms of sets (slugs that are in existing_experiments), (slugs that are in updated_experiments) and (slugs that are in existing_enrollments), we can see the logic to ensure that we're calling `evolve_enrollment` an equivalent number of times than before we started this PR. We're just calling them in a different order.
* because of this, the `enrollment_events` doesn't need any special tweaking.
* it should be clear how we can change this algorithm in a single method to cope with multiple features per experiment. This is an explicit goal of this ticket.
* should we shuffle the ordering of `updated_experiments` for loop 3? If multiple experiments contending on the same feature(s) launch on the same day, will one of them get more of the population that the other?

dmose:
I've finished another refactor; still some stuff to sort out.

dmose:
After extensive discussion and pairing with jhugman, we've decided that this spike has done what we needed - made it clear how and how-not-to implement SDK-260, as well as which tests to write to convince ourselves that things are working well enough.  Closing this PR; ongoing work is proceeding in https://github.com/mozilla/application-services/tree/jhugman/one-experiment-per-feature which will be a PR before long.

PR #4043, opened 2021-04-28
Fix building the application services project with Apple M1 [ci full]

skhamis:
- project does not build cleanly on Apple M1 hardware
- Adjusted build scripts to correctly link the right libraries for Apple Silicon (M1)
- Added new var in the scripts to keep track of the two current macos hardware

  - Note:

### DIFF
--- a/libs/build-nss-desktop.sh
+++ b/libs/build-nss-desktop.sh
@@ -15,6 +15,8 @@ NSS_SRC_DIR=${1}
 # Whether to cross compile from Linux to a different target.  Really
 # only intended for automation.
 CROSS_COMPILE_TARGET=${2-}
+# This will stay consistent with CARGO_CFG_TARGET_ARCH
+TARGET_ARCH="unknown"
 
 if [[ -n "${CROSS_COMPILE_TARGET}" ]] && [[ "$(uname -s)" != "Linux" ]]; then
   echo "Can only cross compile from 'Linux'; 'uname -s' is $(uname -s)"
@@ -33,8 +35,8 @@ elif [[ -n "${CROSS_COMPILE_TARGET}" ]]; then
 elif [[ "$(uname -s)" == "Darwin" ]]; then
   DIST_DIR=$(abspath "desktop/darwin/nss")
   TARGET_OS="macos"
-  #We need this variable for switching libs based on different macos archs (M1 vs Intel)
-  #also, keeping the naming consistent with target_arch env on the rust side
+  # We need this variable for switching libs based on different macos archs (M1 vs Intel)
+  # Renaming uname to stay consistent with CARGO_CFG_TARGET_ARCH on the rust side
   if [[ "$(uname -m)" == "arm64" ]]; then
     TARGET_ARCH="aarch64"
   else
@@ -106,7 +108,7 @@ cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libsmime.a" "${DIST_DIR}/lib"
 cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libsoftokn_static.a" "${DIST_DIR}/lib"
 cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libssl.a" "${DIST_DIR}/lib"
 
-#Apple M1 need HW specific libs copied over to successfully build
+# Apple M1 need HW specific libs copied over to successfully build
 if [[ "${TARGET_OS}" == "macos" ]] && [[ "${TARGET_ARCH}" == "aarch64" ]]; then
   cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libarmv8_c_lib.a" "${DIST_DIR}/lib"
   cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libgcm-aes-aarch64_c_lib.a" "${DIST_DIR}/lib"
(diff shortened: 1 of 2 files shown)


libs/build-nss-desktop.sh
```diff
@@ -33,6 +33,13 @@ elif [[ -n "${CROSS_COMPILE_TARGET}" ]]; then
 elif [[ "$(uname -s)" == "Darwin" ]]; then
   DIST_DIR=$(abspath "desktop/darwin/nss")
   TARGET_OS="macos"
+  #We need this variable for switching libs based on different macos archs (M1 vs Intel)
```
mhammond, line 36: nit: here and below, please add a space between the comment char and the comment itself

libs/build-nss-desktop.sh
```diff
@@ -33,6 +33,13 @@ elif [[ -n "${CROSS_COMPILE_TARGET}" ]]; then
...
+  #also, keeping the naming consistent with target_arch env on the rust side
```
mhammond, line 37: I think we should s/target_arch/CARGO_CFG_TARGET_ARCH/ to make it extra clear what we are trying to do

mhammond (commented):
hrm - that's failing with `TARGET_ARCH: unbound variable` - so TIL bash is stricter than python :) Maybe unconditionally set it to 'unknown' above or something?

bendk (commented):
Looks good overall.  I agree with Mark's suggestions and also had a small one of my own.

libs/build-nss-desktop.sh
```diff
@@ -98,12 +105,18 @@ cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libpkcs7.a" "${DIST_DIR}/lib"
...
 cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libsoftokn_static.a" "${DIST_DIR}/lib"
 cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libssl.a" "${DIST_DIR}/lib"
-cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libhw-acc-crypto-avx.a" "${DIST_DIR}/lib"
-cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libhw-acc-crypto-avx2.a" "${DIST_DIR}/lib"
 
-# HW specific.
-# https://searchfox.org/mozilla-central/rev/1eb05019f47069172ba81a6c108a584a409a24ea/security/nss/lib/freebl/freebl.gyp#159-163
-cp -p -L "${NSS_DIST_OBJ_DIR}/lib/libgcm-aes-x86_c_lib.a" "${DIST_DIR}/lib"
+#Apple M1 need HW specific libs copied over to successfully build
+if [[ "${TARGET_OS}" == "macos" ]] && [[ "${TARGET_ARCH}" == "aarch64" ]]; then
```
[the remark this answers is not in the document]
skhamis: I think this is a good suggestion! The only issue is Linux is also used heavily in cross-compilation and from what I understand, uname is a little tricky with that kind of stuff, so this kinda errs on the side of caution to prevent any unwanted side-effects. Could be wrong though!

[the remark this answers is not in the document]
mhammond: I think Ben's correct - I don't think the cross compilation stuff would break if you did this. I do recall thinking we had a good reason for that other check yesterday but now I can't see it :)

(I think that ultimately we will want the cross compilation cruft to target this processor, which IIUC doesn't happen, but that's something we can add later if we need it (tbh, I'm not sure how these cross-compilation targets are used in practice)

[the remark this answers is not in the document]
skhamis: Since we only set target_arch in the above macos statement, this will be safe to remove though. Thanks Ben for the suggestion! Will implement shortly.

libs/build-nss-desktop.sh
```diff
@@ -15,6 +15,8 @@ NSS_SRC_DIR=${1}
 # Whether to cross compile from Linux to a different target.  Really
 # only intended for automation.
 CROSS_COMPILE_TARGET=${2-}
+# This will stay consistent with CARGO_CFG_TARGET_ARCH
```
mhammond, line 18: Hate to make more nits, but this comment might leave the impression `CARGO_CFG_TARGET_ARCH` can be `unknown` - so a bit more explicit here makes sense `We only need this in a couple of places - when it's not "unknown" it will be ...etc`

libs/build-nss-desktop.sh
```diff
@@ -33,6 +35,13 @@ elif [[ -n "${CROSS_COMPILE_TARGET}" ]]; then
 elif [[ "$(uname -s)" == "Darwin" ]]; then
   DIST_DIR=$(abspath "desktop/darwin/nss")
   TARGET_OS="macos"
+  # We need this variable for switching libs based on different macos archs (M1 vs Intel)
+  # Renaming uname to stay consistent with CARGO_CFG_TARGET_ARCH on the rust side
```
mhammond, line 39: and then delete this line (leave the first line though)

libs/build-sqlcipher-desktop.sh
```diff
@@ -13,6 +13,8 @@ SQLCIPHER_SRC_DIR=${1}
 # Whether to cross compile from Linux to a different target.  Really
 # only intended for automation.
 CROSS_COMPILE_TARGET=${2-}
+# This will mimic CARGO_CFG_TARGET_ARCH
```
mhammond, line 16: and ditto both comments in this file.

mhammond (changes requested):

libs/build-nss-desktop.sh
skhamis: all good, definitely want to make sure we bake as much context as possible so it's easy to pick-up/understand in the future!

mhammond (approved):
Great, thanks! Let's have Ben take another look and he can merge it if he approves.

bendk:
Looks great!

PR #4045, opened 2021-04-28
Add Nimbus sample, including a xcode workspace file

travis79:
This builds for the simulator, but fails on a dyld call at runtime.  I started with Carthage, but then decided to simplify it a bit by just trying to get it to work with the local code first, so I added an Xcode workspace to encompass MozillaAppServices and the NimbusSample.  Since I ran into the same issues either way, this seemed to be the simplest way to illustrate the problems I was having.


travis79:
cc @boek I couldn't flag you for review for some reason...

components/nimbus/ios/Nimbus Sample/Nimbus Sample/Experiments.swift
```diff
@@ -0,0 +1,82 @@
...
+public enum Experiments {
+    public static var shared: NimbusApi {
+        // Nimbus setup
+
+        // This constructs a `URL` representing the Remote Settings endpoint
+        // we can retrieve experiments from.
+        var nimbusEndpointUrl: URL {
+            var components = URLComponents()
+            components.scheme = "https"
+            components.host = "firefox.settings.services.mozilla.com"
```
jhugman, line 16: Should this be the live one, or a stage one?

components/nimbus/ios/Nimbus Sample/Nimbus Sample/Experiments.swift
```diff
@@ -0,0 +1,82 @@
...
+        // caught and an error message logged.
+        do {
+            return try Nimbus.create(
+                nimbusServerSettings,
+                appSettings: nimbusAppSettings,
+                dbPath: dbPath,
+                errorReporter: nimbusErrorReporter
+            )
+        } catch {
+            print("Something went wrong durning NimbusApi creation")
```
jhugman, line 63: Nit: durning.

components/nimbus/ios/Nimbus Sample/Nimbus Sample/Nimbus_SampleApp.swift
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import SwiftUI
+
+@main
+struct Nimbus_SampleApp: App {
```
jhugman, line 8: I'm not wild about the spaces in the names. In the directory paths, or here.

jhugman (changes requested):
This looks ok, but for the following:

 * Given the sample app is importing the megazord, this should go in the top level examples directory.
 * I could build it, but not run it. Do we need some docs (i.e. was I building it wrong?)
 * There are some naming nits around spaces and underscores in names which should be camel cased.

```
dyld: Library not loaded: @rpath/SwiftKeychainWrapper.framework/SwiftKeychainWrapper
```

travis79:
> This looks ok, but for the following:
> 
>     * Given the sample app is importing the megazord, this should go in the top level examples directory.
> 
>     * I could build it, but not run it. Do we need some docs (i.e. was I building it wrong?)
> 
>     * There are some naming nits around spaces and underscores in names which should be camel cased.
> 
> 
> ```
> dyld: Library not loaded: @rpath/SwiftKeychainWrapper.framework/SwiftKeychainWrapper
> ```

This is the problem I was seeking help with and which I was stuck on when I put up the PR.  

The location of the sample was just out of convenience, happy to place it in the top level samples directory, but that seemed to be all rust samples. Since there was no other iOS samples to go off of it was a shot in the dark to just put it in the component.

components/nimbus/ios/Nimbus Sample/Nimbus Sample/Experiments.swift
```diff
@@ -0,0 +1,82 @@
...
+        // Create the NimbusApi object and assign it to the local `nimbus`
+        // variable. This could potentially throw, so any errors are
+        // caught and an error message logged.
+        do {
+            return try Nimbus.create(
+                nimbusServerSettings,
+                appSettings: nimbusAppSettings,
+                dbPath: dbPath,
+                errorReporter: nimbusErrorReporter
+            )
```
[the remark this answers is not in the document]
travis79: This was my plan in the `NimbusSampleApp.swift` file, after initializing to set experiments locally and apply them.  I just got stuck trying to build and test this (see PR description) and didn't get as far as actually creating the experiment to ship with the sample.

components/nimbus/ios/Nimbus Sample/Nimbus Sample/Experiments.swift
travis79: Since we don't really have a way to target the sample unless we add it as an app to Experimenter, I'm not sure pointing this at either prod or stage gains us anything at this point.  Perhaps it would be better to point this at "example.com" and just rely on the local experiment file.  I can see where @boek's debug workflow from Firefox-iOS may be useful here, also.

travis79:
Closing this due to some bit rot, I'll open a new PR once I have the sample app tidied up.

PR #4047, opened 2021-04-29
No Bug - Fix swiftlint after update to swiftformat broke it

travis79:
Looks like there was an update to the swiftformat tool that caused the swiftlint task to start failing again.  This should hopefully fix it.

### DIFF
--- a/components/nimbus/ios/Nimbus/Nimbus.swift
+++ b/components/nimbus/ios/Nimbus/Nimbus.swift
@@ -277,37 +277,21 @@ public extension NimbusDisabled {
         return nil
     }
 
-    func initialize() {
-        return
-    }
+    func initialize() {}
 
-    func fetchExperiments() {
-        return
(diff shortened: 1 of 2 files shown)


components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -277,37 +277,21 @@ public extension NimbusDisabled {
         return nil
     }
 
-    func initialize() {
-        return
-    }
+    func initialize()
```
jhugman, line 280: TIL.

travis79: Eh, swiftformat was suggesting this (I think) but xcodebuild doesn't seem to like it so I'm tinkering on this now...   Turns out the empty braces are needed, it was the empty return that swiftformat wasn't happy with.

PR #4048, opened 2021-04-29
Introduce History Metadata storage & API for Android

grigoryk:
This introduces a new History Metadata storage along with Android bindings for it. Swift bindings to come soon.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -2,6 +2,10 @@
 
 # Unreleased Changes
 
+## History Metadata Storage
+
+- Introduced a new metadata storage API, part of libplaces. Currently only has Android bindings.
+
 ## Sync Manager
 
 - Removed support for the wipeAll command (#4006)
--- a/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
@@ -405,11 +405,11 @@ open class PlacesReaderConnection internal constructor(connHandle: Long) :
         }
     }
 
-    override suspend fun getHistoryMetadataSince(start: Long): List<HistoryMetadata> {
+    override suspend fun getHistoryMetadataSince(since: Long): List<HistoryMetadata> {
         readQueryCounters.measure {
             val rustBuffer = rustCall { error ->
                 LibPlacesFFI.INSTANCE.places_get_history_metadata_since(
-                        this.handle.get(), start, error)
+                        this.handle.get(), since, error)
             }
             try {
                 val metadata = MsgTypes.HistoryMetadataList.parseFrom(rustBuffer.asCodedInputStream()!!)
@@ -895,12 +895,12 @@ interface ReadableHistoryMetadataConnection : InterruptibleConnection {
     suspend fun getLatestHistoryMetadataForUrl(url: String): HistoryMetadata?
 
     /**
-     * Returns all [HistoryMetadata] where [HistoryMetadata.updatedAt] is greater or equal to [start].
+     * Returns all [HistoryMetadata] where [HistoryMetadata.updatedAt] is greater or equal to [since].
      *
-     * @param start Timestmap to search by.
+     * @param since Timestmap to search by.
      * @return A `List` of matching [HistoryMetadata], empty if nothing is found.
      */
-    suspend fun getHistoryMetadataSince(start: Long): List<HistoryMetadata>
+    suspend fun getHistoryMetadataSince(since: Long): List<HistoryMetadata>
 
     /**
      * Returns all [HistoryMetadata] where [HistoryMetadata.updatedAt] is between [start] and [end], inclusive.
--- a/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
@@ -261,6 +261,7 @@ open class PlacesConnection internal constructor(connHandle: Long) : Interruptib
 open class PlacesReaderConnection internal constructor(connHandle: Long) :
         PlacesConnection(connHandle),
         ReadableHistoryConnection,
+        ReadableHistoryMetadataConnection,
         ReadableBookmarksConnection {
     override fun queryAutocomplete(query: String, limit: Int): List<SearchResult> {
         val resultBuffer = rustCall { error ->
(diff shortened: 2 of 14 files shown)


grigoryk:
`swiftlint` check should be fixed by https://github.com/mozilla/application-services/pull/4047

grigoryk:
`check-dependencies` failure seems unrelated to this PR as well - we didn't touch dependencies at all.
```RuntimeError: Could not infer license URL for 'cpuid-bool'.```

File: components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt [hunk 1]
```diff
@@ -583,6 +645,43 @@ class PlacesWriterConnection internal constructor(connHandle: Long, api: PlacesA
         }
     }
 
+    override suspend fun addHistoryMetadata(metadata: HistoryMetadata): String {
+        val builder = MsgTypes.HistoryMetadata.newBuilder()
```
grigoryk, line 649: @csadilek found that this isn't friendly to our nullable values; should probably convert to defaults (e.g. empty strings) if need be.

components/places/src/storage/history_metadata.rs
```diff
@@ -0,0 +1,992 @@
...
+    // Heavy lifting around moz_places inserting (e.g. updating moz_origins, frecency, etc) is performed via triggers.
+    let places_id = match place_id {
+        Some(id) => id,
+        None => {
+            let sql = "INSERT INTO moz_places (guid, url, title, url_hash)
+               VALUES (:guid, :url, :title, hash(:url))";
+
+            let guid = SyncGuid::random();
+
+            // This normalizes urls, e.g. adding trailing slashes when they're missing.
```
mhammond, line 139: I think we should see if we can tweak the API rather than doing the sql here - I see 3 problems: (1) the query for an existing place ID is done outside the transaction, (2) we enforce a max title len (which we trim) and max URL len (which causes us to fail) that isn't done here and (3) IIUC the title for an existing moz_places record isn't updated when it doesn't match (and in that scenario you'd need to bump the change counter!)

You mentioned `new_page_info()` not taking a title param and the reason for this is that we l would need to do an update in the case where the page ID already exists. The "correct" API to use here would be something like [apply_observation](https://github.com/mozilla/application-services/blob/cd6c4e8a02f5339d87290aaa26709447b73c8ec8/components/places/src/storage/history.rs#L47) but that also adds a history record which we don't want - because presumably whatever is triggering this call to be made will also be adding a history visit too!

So I wonder if we can leverage that? If we can be certain that adding a history entry has already finished when this is called, we might be able to avoid even trying to create a new place and handling the title? IOW, can we just arrange to just bail of the place ID doesn't exist, because that should be impossible given there must also be an existing visit?

Otherwise I'll think about what shape the API should be (ie, how to split that above function in a way that splits `moz_places` from `moz_history_visits`) - probably just by adding a title param to `new_page_info()` and special-casing "might need to update title if the page already exists, but never if it doesn't" in `apply_observation()` - or maybe a new `ensure_page_info()` type function, which means you don't need to worry about any of that?

(Or maybe we just fix these problems here? I'll think about that :)

components/places/src/storage/history_metadata.rs
```diff
@@ -0,0 +1,992 @@
...
+                &[(":term", &term)],
+            )?;
+            Some(tx.conn().last_insert_rowid())
+        }
+    };
+
+    let parent_domain_id = match parent_domain_entry {
+        ParentDomainEntry::None => None,
+        ParentDomainEntry::Existing(id) => Some(id),
+        ParentDomainEntry::CreateFor(url) => {
```
mhammond, line 171: why doesn't our trigger work out here?

components/places/src/storage/history_metadata.rs
```diff
@@ -0,0 +1,992 @@
...
+                &[(":url", &parent_url.as_str())],
+                true
+            )? {
+                Some(id) => ParentDomainEntry::Existing(id),
+                None => ParentDomainEntry::CreateFor(parent_url)
+            }
+        }
+    };
+
+    let tx = db.begin_transaction()?;
```
mhammond, line 249: as implied above, I think we probably want this at the top of the function, just to ensure nothing races with us (in theory sync could - it uses its own write connection.)

mhammond (commented):
Most of this looks fine to me and quite thorough - I need to think more about that place ID and it's the end of my day, but there's nothing major I'm concerned about.

(Other than the concerns I have with syncing, which I suspect is going to need some significant schema changes, but that's tractable, especially if we are willing to drop existing data!)

components/places/src/storage/history_metadata.rs
grigoryk: For (1) and (2) I've pushed out the update. Both were oversights, I did think about doing them but promptly forgot :) Thanks for catching it.

For (3), we made a conscious choice to basically not worry about `title` updates. We're basically depending on the fact that there's a single source of these titles, and a certain sequencing in place, and an already existing title update mechanism in place. And in case we see a misalignment here, it's almost feels like a bug elsewhere (why didn't that title update persist already?) and basically doesn't matter for what we're building on top of this storage.

In practice, I doubt we'll actually end up inserting in moz_places/moz_origins (for parents) ourselves, in Fenix.

So I partially agree with "let's just not write to moz_places at all" logic (the paragraph above basically argues for that, between the lines).
But, that currently wouldn't work for iOS until they start using history APIs.

And, even in case of Fenix consuming this, not having fallback moz_places inserts starts to tie us into too many assumptions about the broader world.

So I think just ignoring title updates is a nice compromise. We've added comments in rust code and at the API boundary to make sure this behaviour is clear.

grigoryk:
Thanks, @mhammond! Addressed your feedback & responded to comments.

mhammond (commented):
I'm OK with landing this once the wipe stuff is addressed - and I'm OK with it as I don't want to block progress, but although do I need to be very clear that I think we can make significant improvements to the API. As mentioned, it's almost certainly not syncable in the current form, so we should be very clear that this data is disposable.

One of my biggest problems is that this is just dumb storage - all the logic for knowing how to coalesce must exist externally. This is probably the biggest issue for syncing.

To fix this, I think we should embrace the "observation" model and move more smarts into this Rust code. In practice, this means we drop the GUID - the consuming code doesn't get to dictate how we manage the records.

Eg, the "update" API would be something like: `places_observe_history_metadata(url, metadata)` - you are just noting that something happened - there's no need for "add" or "update" because the API abstraction should not be a DB record.

The "read" API would be something like: `places_get_history_metadata(url)` and the return value is something like:

```
struct {
  url: "http://blah",
  title: ...
  metadata: [
    // An *array* of the metadata - much like visits - but not necessarily exactly what was previously "observed"
  ]
}
```

and it is this rust code that knows how to coalesce the multiple observations into smaller groups, or even discarding them entirely. The end result is that this ends up being shaped more like history and we sync it just like history (ie, the GUID is the guid for the *place* and each record has an array)

This is something we can discuss later, but hopefully not too much later - it's a massive red flag that we are adding this data which we know will not sync, while we also know that syncing is super important to the success of the broader project.

grigoryk:
Thanks, @mhammond ! Pushed up a fix for the metadata+bookmark wipe issue. FWIW I think it's fine if we leave those records hanging in this case, but it's also fine to explicitly wipe them as well.

Filed https://github.com/mozilla/application-services/issues/4067 as a follow-up to make sure we're doing the right thing from Sync's perspective in terms of the API surface of this component. Definitely want to re-state our full intention to get this right once the dust settles a bit :) We'll keep iterating on this in the meantime.

CHANGES_UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 # Unreleased Changes
 
+## History Metadata Storage
+
+- Introduced a new metadata storage API, part of libplaces. Currently only has Android bindings.
```
mhammond, line 7: Please add "EXPERIMENTAL" in big scary characters here :)

mhammond (approved):
👍

components/places/sql/create_shared_schema.sql
```diff
@@ -237,3 +237,40 @@ CREATE TABLE IF NOT EXISTS moz_keywords(
...
+    key_presses INTEGER,
+
+    FOREIGN KEY(place_id) REFERENCES moz_places(id) ON DELETE CASCADE,
+
+    FOREIGN KEY(search_query_id) REFERENCES moz_places_metadata_search_queries(id) ON DELETE CASCADE,
+
+
+    FOREIGN KEY(parent_domain_id) REFERENCES moz_origins(id) ON DELETE CASCADE
+);
+CREATE UNIQUE INDEX IF NOT EXISTS guid_uniqueindex ON moz_places_metadata(guid);
```
mak77, line 270: the column is already defined UNIQUE, thus it already has a unique index, this is a duplicate index.

components/places/sql/create_shared_schema.sql
```diff
@@ -237,3 +237,40 @@ CREATE TABLE IF NOT EXISTS moz_keywords(
...
+
+CREATE TABLE IF NOT EXISTS moz_places_metadata_search_queries (
+    id INTEGER PRIMARY KEY,
+    term TEXT NOT NULL UNIQUE
+);
+CREATE UNIQUE INDEX IF NOT EXISTS term_uniqueindex ON moz_places_metadata_search_queries(term);
```
mak77, line 276: Same thing here, the data like this is stored 3 times

components/places/sql/create_shared_schema.sql
grigoryk: Thanks for catching this, will fix it up.

grigoryk: @mak77 fixed in https://github.com/mozilla/application-services/pull/4070

PR #4052, opened 2021-04-30
Update XCode to 12.5

mhammond:


mhammond:
12.5 failed with `xcode:12.5.0 is not supported` - so let's try 12.4 :)

st3fan:
@mhammond Thank you for putting this together. It looks like Xcode 12.5 only runs on macOS 11. Which is not generally available on CircleCI. Butm you can sign up to get it enabled on your CircleCI account. (They call it preview but a CircleCI employee says it is production ready) See this https://discuss.circleci.com/t/macos-11-big-sur-status-update/39204/7 discussion.

cc @jdragojevic in case we need approval to sign up? Also wondering if supporting macOS 11 is a bigger project.

st3fan:
See also https://github.com/CircleCI-Public/macos-big-sur-preview-docs

st3fan:
> 12.5 failed with `xcode:12.5.0 is not supported` - so let's try 12.4 :)

I think that could be useful if the product teams need to include a new A-S build in the product that they currently build with Xcode 12.4. But at least for Focus iOS there is going to be a need to switch over to Xcode 12.5 relatively soon - we simply can't lag too much behind with Apple's tooling.

st3fan:
Working with @badboy on this for Focus iOS - that only covers Glean (because Focus doesn't need more) but maybe this work results in some good findings that we can apply to the other native libraries?

https://bugzilla.mozilla.org/show_bug.cgi?id=1711447
https://github.com/mozilla-mobile/focus-ios/pull/1828

jdragojevic:
> @mhammond Thank you for putting this together. It looks like Xcode 12.5 only runs on macOS 11. Which is not generally available on CircleCI. Butm you can sign up to get it enabled on your CircleCI account. (They call it preview but a CircleCI employee says it is production ready) See this https://discuss.circleci.com/t/macos-11-big-sur-status-update/39204/7 discussion.
> 
> cc @jdragojevic in case we need approval to sign up? Also wondering if supporting macOS 11 is a bigger project.

@st3fan  - I filled out the request form on circle CI to get 12.5 access and reached out to a friend who works there to see if we can get an extra nudge on our request.

mhammond:
This gets close, but now fails with an obscure error I can't make heads or tails of when executing `brew install swift-protobuf --force`

```
==> Installing dependencies for swift-protobuf: six and protobuf
==> Installing swift-protobuf dependency: six
==> Pouring six--1.16.0_1.big_sur.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink lib/python3.9/site-packages/six.py
Target /usr/local/lib/python3.9/site-packages/six.py
already exists. You may want to remove it:
  rm '/usr/local/lib/python3.9/site-packages/six.py'

To force the link and overwrite all conflicting files:
  brew link --overwrite six

To list all files that would be deleted:
  brew link --overwrite --dry-run six

Possible conflicting files are:
/usr/local/lib/python3.9/site-packages/six.py
```

skhamis:
Sounds like it's potentially a known issue with circleci/homebrew? (link below, user has nearly the exact same issue). 
https://discuss.circleci.com/t/macos-image-users-homebrew-brownout-2021-04-26/39872/8

mhammond:
> Sounds like it's potentially a known issue with circleci/homebrew?

Nice! My google-foo failed me there :/

mhammond:
That "discuss.circleci.com" issue leads me to conclude that something is a little broken on circleci around at least the "six" package. Long story short, a hacky `rm` of that file gets us past this. Which then takes us to yet another spot where I'm well outside my comfort zone :)

* "Run XCode tests" failure:

> Running script 'Generate Glean metrics'
> error: linking with `cc` failed: exit code: 1
> error: aborting due to previous error
> error: could not compile `proc-macro2
> ...
> The following build commands failed:
> 	RuleScriptExecution /Users/distiller/project/components/nimbus/src/nimbus.udl normal x86_64

Can't see the actual compiler error yet though. Not clear if the issue has anything to do with glean. No idea what the nimbus reference is :) I pinged @badboy to see if it might be glean and I'll get back to this ASAP.

.circleci/config.yml
```diff
@@ -216,7 +216,8 @@ executors:
       - image: cimg/rust:1.50.0
   macos:
     macos:
-      xcode: "12.3.0"
+      xcode: 12.5.0
```
rfk, line 219: @st3fan if we do switch to this configuration, will firefox-ios be able to consume a library built with this version as well as focus?

rfk:
> Doesn't look like Glean, but uniffi-bindgen building code.

Agreed, I think this is failing to build the `proc-macro2` dependency of the `uniffi_bindgen` crate, which it is trying to do so that it can run it and generate the swift bindings.

However, we have other things that depend on `proc-macro2` and `raw_xcodetest.log` shows it compiling successfully earlier in the build. So I think this must be specifically something to do with the way we are trying to execute it as a cargo sub-command from inside XCode, perhaps it's not passing on the correct environment variables.

rfk:
Well the "good" news is that I appear to have XCode 12.5 and Big Sur on my mac here, and I seem to be able to reproduce the same build error.

rfk:
> Well the "good" news is that I appear to have XCode 12.5 and Big Sur on my mac here,
> and I seem to be able to reproduce the same build error.

If I manually run `cargo uniffi-bindgen` from the command-line in order to make it build all the dependencies outside of XCode, then I'm able to execute the whole iOS build and test process successfully. So I think this is definitely something from the environment from XCode that is causing the compilation to fail.

rfk:
I captured a failing command invocation from a local build, and did a binary search on the environment variables to try to find the one that's causing the problem. It seems to be due to setting `$PATH` in the environment when executing `cargo uniffi-bindgen`.

rfk:
Specifically, XCode is putting `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/local/bin` at the front of `$PATH`, so we find `cc` from there rather than the default `/usr/bin/cc`, which I guess is in turn doing something unexpected.

rfk:
Things work if I explicitly set `CARGO_BUILD_RUSTFLAGS=\"-L /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib\"` in the environment, which is not a great solution, but at least is forward progress.

rfk:
I have this building again locally now, using https://github.com/mozilla/application-services/pull/4166; hopefully that or something similar will help unstick this PR as well.

badboy:
> Things work if I explicitly set `CARGO_BUILD_RUSTFLAGS=\"-L /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib\"` in the environment, which is not a great solution, but at least is forward progress.

Ah, now I remember I had to deal with that as well recently.
Does setting [`LIBRARY_PATH`](https://github.com/mozilla/glean/blob/main/build-scripts/xc-universal-binary.sh#L26-L31) work?

rfk:
>  Does setting LIBRARY_PATH work?

I expect it will; given the fix from #4166 I haven't tried it out further yet.

rfk:
@mhammond just flagging that I reckon a rebase here has a good chance of getting things building correctly, after some other changes landed in `main`.

mhammond:
>  I reckon a rebase here has a good chance of getting things building correctly

Indeed it did :) I'll point this PR out to #firefox-ios

rfk (approved):
r+, but pending confirmation from the iOS team(s) of course 👍🏻

nbhasin2 (approved):

nbhasin2:
Thanks for getting this through 🙏🏽
Firefox iOS is moving to XCode 12.5 so we should be able to consume!!

PR #4068, opened 2021-05-05
Uniffi-ing the Logins component

skhamis:
Uniffi-ing logins get rid of all the existing hand-crafted ffi code and use the Uniffi tool to automatically generate these bindings. Want to get this out for an initial look-thru and if the team has comments on the potential API changes.

## Main Area changes
Since we're removing protobuf (as per the converting [Converting to Uniffi guide](https://mozilla.github.io/application-services/book/howtos/converting-a-component-to-uniffi.html) We needed to change quite a few areas in:
- logins/src/stores.rs
- logins/src/db.rs

There is only one struct moving forward `Login` which will be breaking changes and I have associated PRs in respective repos to accompany this PR:

> ---
> ### android-components: https://github.com/mozilla-mobile/android-components/pull/10262
> ### fenix: https://github.com/mozilla-mobile/fenix/pull/19712
> ### fxiOS: https://github.com/mozilla-mobile/firefox-ios/pull/8602
> ---

Another area is getting rid of interrupt and calling sync directly in the logins code. Mark mentioned that Sync Manager should be handling all of this and that would allow us to rip some code we'd otherwise have to make some hacky workaround to make sure they're still included (AFAIK, Uniffi doesn't let you import types like SQLInterrupt)

----

  - Note:


components/logins/src/db.rs
```diff
@@ -30,22 +30,22 @@ use url::{Host, Url};
...
 pub struct MigrationPhaseMetrics {
-    num_processed: u64,
-    num_succeeded: u64,
-    num_failed: u64,
-    total_duration: u128,
-    errors: Vec<String>,
+    pub num_processed: u64,
+    pub num_succeeded: u64,
+    pub num_failed: u64,
+    pub total_duration: u64,
```
skhamis, line 36: Unfortunately, Uniffi doesn't have i/u greater than 64 but it looks like it was chosen as u128 since that was what Rust defaults to when using the is_milis() so it seems like we might be okay

components/logins/src/db.rs
```diff
@@ -201,6 +166,41 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
     Ok(())
 }
 
+/// Opens an existing database and fetches the salt.
+/// This method is used by iOS consumers as part as the migration plan to store
+/// the salt outside of the sqlite db headers.
+///
+/// Will return an error if the database does not exist.
+pub fn open_and_get_salt(path: impl AsRef<Path>, encryption_key: &str) -> Result<String> {
```
skhamis, line 174: These don't call &self and can be taken out of the impl block (since their static) to keep things cleaner on the Uniffi side

components/logins/src/error.rs
```diff
@@ -102,7 +98,170 @@ impl Error {
...
+        }
+    }
+}
+
+// TODO: Move the above to somewhere else - everything below is our "public" error list.
+// Probably copy the pattern started by fxa-client - everything above moves to `internal/`?
+// Originally exposed manually via LoginsStorageException.kt.
+
+#[derive(Debug, thiserror::Error)]
+pub enum LoginsError {
```
skhamis, line 110: I propose we call it `LoginsRecordError` feels like almost all these issues is with some kind of record issue so makes the most sense for the final error we use.

components/logins/src/db.rs
lougeniaC64: I just git blamed that line and I actually chose `u128` when I created the `MigrationPhaseMetrics` and if memory serves it was for the exact reason you specified. Also, (this)[https://github.com/mozilla/application-services/blob/72b827c3e0f883163762857fd766df1aeb060725/components/support/ffi/src/into_ffi.rs#L48] probably explains why Uniffi doesn't support `u128`.

components/logins/src/lib.rs
```diff
@@ -9,21 +9,41 @@
...
 pub use crate::db::LoginStore;
+use crate::db::MigrationMetrics;
+use crate::db::MigrationPhaseMetrics;
 pub use crate::error::*;
 pub use crate::login::*;
 pub use crate::store::*;
 
-pub mod msg_types {
-    include!("mozilla.appservices.logins.protobuf.rs");
+#[derive(Clone, PartialEq)]
```
lougeniaC64, line 31: nit: Consider moving this into `login.rs` and exposing the `login` module as public.

components/logins/uniffi.toml
```diff
@@ -0,0 +1,3 @@
+[bindings.kotlin]
+package_name = "mozilla.appservices.logins"
+cdylib_name = "megazord"
```
lougeniaC64, line 3: nit: add a newline at the end of the file

components/logins/src/store.rs
```diff
@@ -19,8 +17,8 @@ pub struct PasswordStore {
 }
 
 impl PasswordStore {
-    pub fn new(path: impl AsRef<Path>, encryption_key: Option<&str>) -> Result<Self> {
-        let db = LoginDb::open(path, encryption_key)?;
+    pub fn new(path: impl AsRef<Path>, encryption_key: &str) -> Result<Self> {
```
lougeniaC64, line 20: For all of the below api calls that make `self.db` calls, where possible I'd move the logic to change the return type from `Login` to `PasswordInfo` to db.rs to keep these functions lean.

skhamis: I agree with this! Per our offline conversation - we'll also be changing `PasswordInfo` to `LoginRecord` since it's still needed even though the struct will be identical to `Login`

components/logins/android/build.gradle
```diff
@@ -78,39 +61,12 @@ configurations {
...
-                java {
-                    option "lite"
-                }
-            }
-        }
-    }
-}
-
-// Needs to happen before `dependencies` in order for the variables
-// exposed by the plugin to be available for this project.
```
mhammond, line 97: Some of the changes in this file don't seem strictly related?

components/logins/src/db.rs
mhammond: I also bet noone looks at these metrics these days :)

components/logins/src/db.rs
```diff
@@ -201,6 +166,41 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
     Ok(())
 }
 
+/// Opens an existing database and fetches the salt.
```
mhammond, line 169: It might be worth checking if this is a migration that's already happened? The dynamics of iOS migration might mean they are happier with killing migrations earlier than Fennec is.

components/logins/src/error.rs
mhammond: By my count, less than 1/2 of them are actually talking about the record. It might make sense to have 2 error enums - one for invalid record, one for other errors, but even that's gray - eg, is `NoSuchRecord` one of them? Probably not - so there would only be 2 errors in that enum?

components/logins/src/lib.rs
```diff
@@ -9,21 +9,41 @@
 mod error;
 mod login;
 
-mod db;
+pub mod db;
```
mhammond, line 12: I'm mildly surprised by some of the visibility changes here - I think many of them might not need to be fully public.

components/logins/src/lib.rs
mhammond: I think we also want comments about why we have the 2 nearly identical structs. IIRC, this is really just due to the Guid/String mismatch - in which case we should see how we can abuse/enhance uniffi to do the right thing in these cases (probably as a followup, but maybe not, depending on how the timing of the other parts of this end up)

Also, I think `pub(crate)` would be sufficient here - I'm mildly surprised it logins.rs actually needs to use it (which is why any `pub` is needed) - if that's impossible to unwind then I agree with Lougenia about moving it - although there could also be a case made for moving it into its own source file if we can avoid logins.rs needing to use it at all - I kinds think login.rs needing to know about both structs is a but of a smell

docs/howtos/converting-a-component-to-uniffi.md
```diff
@@ -7,8 +7,7 @@ hand-written FFI layer and foreign-language bindings.
 As we've gained more experience with building components in this way, we've
```
mhammond, line 7: Let's split the doc changes into a separate PR (but thanks for doing them!)

Another section we discovered is missing is "converting your errors" if you'd like to add even more :)

components/logins/ffi/src/lib.rs
```diff
@@ -85,39 +87,41 @@ unsafe fn bytes_to_key_string(key_bytes: *const u8, len: usize) -> Option<String
...
-    db_path: FfiStr<'_>,
-    encryption_key: FfiStr<'_>,
-    error: &mut ExternError,
-) -> *mut c_char {
-    log::debug!("sync15_passwords_get_db_salt");
-    let path = db_path.as_str();
-    let key = encryption_key.as_str();
-    ffi_support::call_with_result(error, || LoginDb::open_and_get_salt(path, key))
-}
+// #[no_mangle]
```
mhammond, line 113: I think it would be helpful to understand the state if you actually removed entirely functions which you believe have been "fully" converted in a sane way - which seems to be most of them. Then what remains in this file can be the "things we need to think about a bit more".

Alternatively, moving some comments to the UDL might make more sense, as that is where the questions will need to be answered, and will allow you to completely delete the `logins/ffi` directory, which will be quite a milestone :)

components/logins/src/error.rs
```diff
@@ -102,7 +98,170 @@ impl Error {
...
+    // In the .kt, this was actually the base-class and lots of Android code
+    // caught this exception when they want *any* exception.
+    // To confuse further, it was also thrown as a generic 'unexpected' exception.
+    // For now, let's keep those things distinct, but for parity with the old
+    // code we should see if we can replace Unexpected with whatever base-class
+    // we end up with.
+    // XXX - TODO - work out the base-class story so the above remains true.
```
mhammond, line 117: Just a reminder to look into this TODO - I really have no idea if the errors generated for kotlin by uniffi will be able to help us here - probably not a huge deal if not, but it will impact the changes in the a-c repo as that might be where the class hierarchy for exceptions is re-established)

mhammond (commented):
This is looking in great shape

components/logins/src/lib.rs
skhamis: login.rs needs to know due to the From<> impl block to convert from Login and PasswordInfo (now LoginRecord). Also a big blocker in simply removing it completely is a uniffi limitation of not being able to [have both a dictionary and object at the same time](https://mozilla.github.io/uniffi-rs/udl/interfaces.html). IMO, I think a future release should get rid of the `Login` struct and just have the LoginRecord so that we can uniffi it easily -- or potentially change uniffi to allow for this paradigm somehow

components/logins/android/build.gradle
skhamis: Interestingly enough I just ran the gradle stuff when testing things and this was autogenerated? I'll try revert changes and see when it actually removes this.

components/logins/src/db.rs
skhamis: Thanks for checking on this! According to the channel comments both iOS and android have not killed it yet (even though it's coming soon)

components/logins/src/logins.udl
```diff
@@ -0,0 +1,116 @@
...
+
+    [Throws=LoginsError]
+    void update(LoginRecord record);
+
+    [Throws=LoginsError]
+    MigrationMetrics import_multiple(sequence<LoginRecord> record);
+
+    // Methods NOT ported over from ffi/lib.rs
+
+    // Q1: sync15_passwords_sync(), do we still need to call sync in stores.rs?
```
skhamis, line 110: So I was able to delete the ffi crate completely! This and the question below were the only ones that didn't make it over to uniffi. It would simplify things a lot if we don't have to port over these two

lougeniaC64: I don't see references for either of these functions in firefox ios, android components, fenix, or mozilla central so let's leave them out for now.

components/logins/src/error.rs
skhamis: I agree here. I think splitting the enums would make things a little more confusing for the consumers of which to use and whatnot. I think going by what each of them interface with it seems like majority of them are when there is a LoginsStorage issue so new proposal would be base class of `LoginsStorageError` and `Unexpected` should be `UnexpectedLoginsStorageError` a bit verbose but is quite to the point!

mhammond (commented):
This is looking good!

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
```diff
@@ -29,6 +27,7 @@ import mozilla.components.service.glean.private.LabeledMetricType
  * LoginsStorage implementation backed by a database.
  */
 class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable, LoginsStorage {
+    private lateinit var store: PasswordStore
```
mhammond: This file can *almost* be removed - except for the locking. I made changes in b0d4c761 but `unlock()` seems wrong.

mhammond (commented):
This is looking good. Note that I pushed to this branch, so make sure you re-pull it!

components/logins/src/error.rs
```diff
@@ -10,99 +10,170 @@ macro_rules! throw {
...
 
-#[derive(Debug, thiserror::Error)]
-pub enum ErrorKind {
-    #[error("Invalid login: {0}")]
-    InvalidLogin(InvalidLogin),
-
-    #[error("The `sync_status` column in DB has an illegal value: {0}")]
-    BadSyncStatus(u8),
+/* We have some internal errors that we use that we don't want to expose that we'll keep
+    See fxa_crate for another example of using an internal module
```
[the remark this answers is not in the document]
bendk: It seems like the split here is that rust code gets one error struct and the FFI code gets another.  What's the reason for this?  Could we just have 1 struct? 

I notice that the `impl From<Error>` code has a lot of logging based on which error its converting.  Could that be moved to the base Error struct?  It seems odd to print out the logs in the `From` implementation.  What about errors that don't get converted?  What if an error gets converted twice?

components/logins/src/store.rs
bendk: Does it make sense at this point to always drop `Login` and only use `LoginRecord`?  Markh mentioned that it was out of scope for this ticket, but it seems like changing all the DB methods to convert between the two is almost as much work.

components/logins/src/logins.udl
```diff
@@ -0,0 +1,116 @@
...
+enum InvalidLoginReason {
+    "EmptyOrigin",
+    "EmptyPassword",
+    "DuplicateLogin",
+    "BothTargets",
+    "NoTarget",
+    "IllegalFieldValue",
+};
+
+interface PasswordStore {
```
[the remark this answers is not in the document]
skhamis: I agree that LoginStore name would suit this better. Unfortunately [db.rs](https://github.com/mozilla/application-services/blob/main/components/logins/src/db.rs#L1228) already has a LoginStore. Two things we can do here:

1. Rename the db.rs something like SyncLoginStore, since it seems specifically for the SyncEngine
2. Completely remove the db.rs one and add the `scope` var to the PasswordStore.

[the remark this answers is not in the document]
skhamis: I would lean towards (1) just because it'll be the safest option during this initial round of work

components/logins/android/build.gradle
```diff
@@ -57,6 +50,7 @@ android {
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
         kotlinOptions {
             jvmTarget = rootProject.ext.build.jvmTargetCompatibility
+            freeCompilerArgs += "-Xopt-in=kotlin.ExperimentalUnsignedTypes"
```
rfk, line 53: I don't think we have a Formal Policy on this, but in general I think we try not to opt in to experimental kotlin features at the library level, instead leaving that choice up to the application. I guess you were running into something like https://github.com/mozilla/application-services/issues/3957 here?

I think the plan outlined in https://github.com/mozilla/application-services/issues/3957#issuecomment-803169111 would be a tractable way to solve this in uniffi and avoid the need for us to specify the opt-in here. @skhamis if I wrote it up in a bit more detail, I wonder if you'd be interested in having a go at implementing the fix in uniffi?

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
```diff
@@ -25,333 +24,155 @@ import org.mozilla.appservices.logins.GleanMetrics.LoginsStore as LoginsStoreMet
 import mozilla.components.service.glean.private.CounterMetricType
 import mozilla.components.service.glean.private.LabeledMetricType
 
+// uniffi has unfortunate names for errors: https://github.com/mozilla/uniffi-rs/issues/442
+typealias LoginsStorageException = LoginsStorageErrorException
+
 /**
- * LoginsStorage implementation backed by a database.
+ * An artifact of the uniffi conversion - a thin-ish wrapper around a
+   PasswordStore.
```
rfk, line 32: What does the wrapper add that we couldn't implement in rust/uniffi? It's probably worth recording as part of the comment here, so that our future selves have the context if uniffi gains more capabilities in future.

components/logins/src/db.rs
```diff
@@ -30,22 +30,26 @@ use url::{Host, Url};
...
-    num_processed: u64,
-    num_succeeded: u64,
-    num_failed: u64,
-    total_duration: u128,
-    errors: Vec<String>,
+    pub(crate) num_processed: u64,
+    pub(crate) num_succeeded: u64,
+    pub(crate) num_failed: u64,
+    // Once https://github.com/mozilla/uniffi-rs/pull/434 lands we might
+    // want to make this a `Duration`?
```
rfk, line 37: 👍🏻 thanks for leaving these breadcrumb comments, they're helpful as we co-evolve the tool and its consumers.

components/logins/src/db.rs
rfk: (Yeah, the only reason anyone would want to use these methods is to work around some weirdness on iOS)

components/logins/src/error.rs
```diff
@@ -10,99 +10,163 @@ macro_rules! throw {
...
-
-    #[error("No record with guid exists (when one was required): {0:?}")]
+// Originally exposed manually via LoginsStorageException.kt.
+#[derive(Debug, thiserror::Error)]
+pub enum LoginsStorageError {
+    #[error("Unexpected error: {0}")]
+    UnexpectedLoginsStorageError(String),
+
+    // This indicates that the sync authentication is invalid, likely due to having
+    // expired.
```
rfk, line 25: nit: please consumer making these into rustdoc comments with `///`, I think they'd be useful to have show up in the generated rust docs even if you don't do any further wordsmithing on them.

rfk (commented):
Thanks for diving in to this! The `logins` component was one of the first components we created and we've learned a lot of lessons since, so I was curious how well it would map to the things we ended up capturing in uniffi.

> +1,307 −2,458 

Nice :-)

> We have two identical stucts now, Login and LoginRecord (previous PasswordInfo, was from 
> the now-gone protobuf). LoginRecord will be mainly used with Uniffi + Protobuf. The reason
> is Uniffi is not able to have both a dictionary and object at the same time (definitely
> something we should consider moving forward!).

FWIW, I'd be interested to hear more about how you would have liked uniffi to work in this case

mhammond:
> > We have two identical stucts now, 

Actually, I don't think this is quite accurate - the structs aren't identical - one has the GUID as a string and the other as a GUID. autofill worked around this by just embracing Strings more, but logins embraced GUIDs separately. I think one of the ways uniffi could help here would be to allow us to use a Guid with uniffi, which implies some way of implementing the conversions ourselves - which might be tricky given Rust rules about implementing traits in different crates - but is probably doable. (I'm going way off topic here, but I'm also thinking about this in terms of @jhugman's `JSONObject` PR and wondering if we can do something like that there too (ie, avoid having uniffi learn about the new type, but in that case somehow leverage Serialze/Deserialize in a way that uniffi still knows just enough to allow objects implementing them to work)

components/logins/android/build.gradle
mhammond: This all looks correct now 👍

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
mhammond: yes, good call. IIRC, there are 2 main things:

* metrics :/
* The lock/unlock semantics - when we "unlock" we actually create a new `PasswordStore` and when we lock it, we drop it. However, the `DatabaseLoginsStorage` stays alive.

So I think we can drop the former once we get a glean API in rust and the latter once we drop sqlcipher.

components/logins/src/internal/error.rs
```diff
@@ -0,0 +1,91 @@
+#[derive(Debug, thiserror::Error)]
```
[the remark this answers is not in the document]
mhammond: still not super-keen on this lonely single file in `internal` and I'm not sure the refactor necessary to make is sane should be done in this already huge PR. My gut reaction is to move these back into the top-level `error.rs` with plenty of block comments about how that refector should be done. I wonder how @rfk feels about this.

components/logins/src/lib.rs
```diff
@@ -10,20 +10,21 @@ mod error;
...
 
 mod db;
-pub mod schema;
+mod schema;
 mod store;
 mod update_plan;
 mod util;
 
-mod ffi;
+// Following how fxa-crate hides some of the implementation details we don't want to expose
```
mhammond, line 18: (Although I think that's largely defeated by the `pub use crate::error::*` here and the `pub use crate::internal::error::*;` in `error.rs`)

components/logins/src/lib.rs
```diff
@@ -10,20 +10,21 @@ mod error;
...
+mod internal;
+
+uniffi_macros::include_scaffolding!("logins");
 
 // Mostly exposed for the sync manager.
```
mhammond, line 23: nit: let's kill this comment - it's stale and a little confusing now.

components/logins/src/logins.udl
```diff
@@ -0,0 +1,116 @@
+namespace logins {
```
mhammond, line 1: we need an MPL license here too.

components/logins/src/logins.udl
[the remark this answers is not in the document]
mhammond: I agree re (1) - maybe `RustLoginStore` is a better name? I actually owe you a PR for the sync manager work, so leave that with me.

components/logins/src/lib.rs
skhamis: I agree it adds much more complexity for no real gain. Will move it back in to the top-level!

components/logins/android/build.gradle
skhamis: I definitely would! The only reason I had it do it in the build.gradle was I couldn't get the @Expiremental... annotations in the generated files. Also good that it lets the consumer know that kotlin deems it experimental

components/logins/src/lib.rs
rfk: FWIW pushing it back up to the top-level sounds fine to me, I think we're still learning about the best patterns to use for these things.

components/logins/android/build.gradle
rfk: Ref https://github.com/mozilla/uniffi-rs/issues/444 for the corresponding UniFFI bug.

components/logins/src/error.rs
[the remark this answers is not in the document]
mhammond: > It seems like the split here is that rust code gets one error struct and the FFI code gets another. What's the reason for this? Could we just have 1 struct?

Yes, I think we would do that if we were starting from scratch. But for reasons lost to history, error.rs was always our "internal" ones and the FFI invented its own. I think that's an anti-pattern and something we could consider consolidating in the future.

components/sync_manager/ffi/src/lib.rs
```diff
@@ -25,15 +25,16 @@ pub extern "C" fn sync_manager_set_places(_places_api_handle: u64, error: &mut E
...
-pub extern "C" fn sync_manager_set_logins(_logins_handle: u64, error: &mut ExternError) {
-    ffi_support::call_with_result(error, || -> MgrResult<()> {
-        log::debug!("sync_manager_set_logins");
-        let api = logins_ffi::STORES.get_u64(_logins_handle, |api| -> Result<_, HandleError> {
-            Ok(std::sync::Arc::clone(api))
-        })?;
-        sync_manager::set_logins(api);
-        Ok(())
-    })
+pub extern "C" fn sync_manager_set_logins(_logins_handle: u64, _error: &mut ExternError) {
```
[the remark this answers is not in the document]
mhammond: this can die now :) But we need to track down the callers :/

mhammond (commented):
This is looking sweeet!

components/sync_manager/src/manager.rs
```diff
@@ -64,44 +61,42 @@ impl SyncManager {
...
-        match r.upgrade() {
-            None => None,
-            Some(arc) => match engine {
-                "addresses" => Some(Box::new(autofill::sync::address::create_engine(arc))),
-                "creditcards" => Some(Box::new(autofill::sync::credit_card::create_engine(arc))),
-                _ => unreachable!("can't process unknown engine: {}", engine),
-            },
+    // Newer engines have a singleton "imported" via crate interdependencies,
+    // all which follow the same basic pattern, so handle them all here.
+    fn imported_engine(engine: &str) -> Option<Box<dyn SyncEngine>> {
```
skhamis, line 70: nice :)

mhammond:
>  I wonder if we add register/deregister methods to the lock/unlock flow until we can get rid of SQLCipher.

Yeah, I think that's roughly what we need to do and I pushed a patch that kinda does this. I say "kinda" because the `unlock` *does not* automatically register, although `lock` does unregister. This means that the burden for registering still remains with the consumer - which is always going to be the case even after we remove sqlcipher. That burden now means "you must call `registerWithSyncManager()` after every `unlock()` but in practice, I think that currently still means "exactly once".

(I think that must also be true currently - because unlock destroys the handle it would still be necessary to call `set_logins()` after every unlock - but in practice `unlock()` is called exactly once so it's all good)

components/logins/src/db.rs
```diff
@@ -24,22 +24,26 @@ use url::{Host, Url};
...
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Default)]
 pub struct MigrationMetrics {
-    fixup_phase: MigrationPhaseMetrics,
-    insert_phase: MigrationPhaseMetrics,
-    num_processed: u64,
-    num_succeeded: u64,
-    num_failed: u64,
-    total_duration: u128,
-    errors: Vec<String>,
+    pub(crate) fixup_phase: MigrationPhaseMetrics,
```
mhammond, line 38: ditto here - probably don't need to be pub?

components/logins/src/login.rs
```diff
@@ -300,13 +297,12 @@ fn string_or_default(row: &Row<'_>, col: &str) -> Result<String> {
...
 impl Login {
     #[inline]
-    pub fn guid(&self) -> &Guid {
-        &self.guid
+    pub fn guid(&self) -> Guid {
+        Guid::from_string(self.id.clone())
     }
-
     #[inline]
     pub fn guid_str(&self) -> &str {
```
mhammond, line 304: let's get an issue on file to remove this - we now have `&self.id` available

components/logins/src/store.rs
```diff
@@ -74,63 +84,85 @@ impl PasswordStore {
...
     }
 
-    pub fn new_interrupt_handle(&self) -> sql_support::SqlInterruptHandle {
-        self.db.new_interrupt_handle()
+    pub fn check_valid_with_no_dupes(&self, login: &Login) -> Result<()> {
+        self.db.lock().unwrap().check_valid_with_no_dupes(&login)
     }
 
     /// A convenience wrapper around sync_multiple.
+    // Unforunately, iOS still uses this until they implement sync engine
```
mhammond, line 142: typo: missing "t" in "Unfortunately" - and s/implement sync engine/use the sync manager/ at the end.

examples/sync-pass/src/sync-pass.rs
```diff
@@ -289,7 +289,7 @@ fn main() -> Result<()> {
     }
 
     loop {
-        match prompt_char("[A]dd, [D]elete, [U]pdate, [S]ync, [V]iew, [B]ase-domain search, [R]eset, [W]ipe, [T]ouch, E[x]ecute SQL Query, or [Q]uit").unwrap_or('?') {
+        match prompt_char("[A]dd, [D]elete, [U]pdate, [V]iew, [B]ase-domain search, [R]eset, [W]ipe, [T]ouch, E[x]ecute SQL Query, or [Q]uit").unwrap_or('?') {
```
mhammond, line 292: The `[S]ync` has been removed from here, but we still support syncing, so I think it should go back?

mhammond (approved):
Great work Sammy!

components/logins/android/src/test/java/mozilla/appservices/logins/DatabaseLoginsStorageTest.kt
```diff
@@ -39,114 +40,49 @@ class DatabaseLoginsStorageTest {
...
 
         store.lock()
+
         return store
     }
 
-    protected fun finishAndClose(store: LoginsStorage) {
+    protected fun finishAndClose(store: DatabaseLoginsStorage) {
         store.ensureLocked()
         assertEquals(store.isLocked(), true)
```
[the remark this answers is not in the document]
skhamis: Added back!

CHANGES_UNRELEASED.md
```diff
@@ -22,4 +22,20 @@ Use the template below to make assigning a version number during the release cut
...
+
+## Logins
+
+### ⚠️ Breaking changes ⚠️
+
+Logins now Uniffi-ed!
+
+API Changes for Logins components:
+
+- Login is the main struct moving forward
```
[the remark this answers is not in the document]
skhamis: Nice, definitely needed to update that and I think you're proposed changes does that succinctly.

examples/sync-pass/src/sync-pass.rs
skhamis: Ah forgot add that back, you're correct!

CHANGES_UNRELEASED.md
```diff
@@ -22,4 +22,14 @@ Use the template below to make assigning a version number during the release cut
 ## [viaduct-reqwest]
 
 ### What's Changed
-  - Update viaduct-reqwest to use reqwest 0.11. ([#4146](https://github.com/mozilla/application-services/pull/4146))
+
+- Update viaduct-reqwest to use reqwest 0.11. ([#4146](https://github.com/mozilla/application-services/pull/4146))
+
+## Logins
+
+### ⚠️ Breaking changes ⚠️
```
rfk, line 30: nit: please put the breaking changes at the *top* of the changelog for easy visibility :-)

rfk: Sorry, to be more clear: I think we should move the whole "logins" section of the changelog up to the top, so that its breaking changes are the most obvious aspect of the changelog.

rfk (approved):
Net negative >2000 lines of code, this looks so good! Thanks for working patiently through the many details here @skhamis.

(Mostly I just really wanted to r+ this one because I'm excited, I don't have any substantial feedback to offer 🎉)

PR #4071, opened 2021-05-06
[WIP] Migrate logins from sqlcipher databases to plaintext

bendk:
This part of a long-running effort, but I figured it was worth it to make a PR so we could discuss the code.  The base branch is the `logins-per-field-encryption` just to emphasize that it's definitely not ready to be merged into main.

This commit changes the logins encryption handling form using sqlcipher to encrypt the whole database to having a plaintext database.  Eventually the plaintext database will have the username/password fields encrypted, but that's still work to do.  The plan is to add that code to `schema::upgrade()` which gets called by the function to migrate to a plaintext DB.  The schema upgrade happens before the plaintext DB migration, which seems important since otherwise the app could crash in the middle of the process and expose the sensitive data.

I'm not sure how to expose the `migrate_sqlcipher_db_to_plaintext` to the FFI layer.  Right now there's a static function in LoginDb that does that.  Maybe this should be moved to LoginStore or some other place though, I'm not sure on that one.

  - Note:


bendk:
After talking to @lougeniaC64 a bit, I'm going to refactor the schema code.  I'll break upgrade() into 2 parts:
  - `upgrade_sqlcipher_db(db, encryption_key)` -- this will handle be used when migrating a sqlcipher database.  It will handle all the current upgrades as well as the upgrade to encrypt the username/password field
  - `upgrade(db)` -- will handle all future upgrades

The main reason to do this is that we need the encryption key when upgrading from sqlcipher, but we won't need it after that.  `LoginDB::open()` calls `schema::init()` and this split avoids having to pass in the encryption key to open() just for the one schema migration.

mhammond:
>     * `upgrade_sqlcipher_db(db, encryption_key)` -- this will handle be used when migrating a sqlcipher database.  It will handle all the current upgrades as well as the upgrade to encrypt the username/password field

IIUC, this will need 2 keys - the sqlcipher key and the per-field key.

bendk:
Okay, I think this one is maybe ready to be merged into the feature branch.  Still a long ways from being ready to merge to main though.

bendk:
Something really weird is happening with this PR.  It seems like it hasn't kept up with the changes in my branch.  Maybe I did one too many rebases or something, but I'm going to close this one and open a fresh one.

Issue #4062, opened 2021-05-04
cargo audit reports security issues


bendk:
While working on  #3994 I noticed that the CI step that runs `cargo audit` was failing:

https://app.circleci.com/pipelines/github/mozilla/application-services/26673/workflows/1cf1fc48-ff25-49e6-9025-cd5cb3873146/jobs/101778

Looks like we need to update a few dependencies

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2181)
┆Story Points: 1
┆Sprint: SYNC - 2021-05-14

PR #4072, opened 2021-05-06
Updating openssl-src (#4062)

bendk:

  - Note:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2033,9 +2033,9 @@ checksum = "77af24da69f9d9341038eba93a073b1fdaaa1b788221b00a69bce9e762cb32de"
 
 [[package]]
 name = "openssl-src"
-version = "111.12.0+1.1.1h"
+version = "111.15.0+1.1.1k"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "858a4132194f8570a7ee9eb8629e85b23cbc4565f2d4a162e87556e5956abf61"
+checksum = "b1a5f6ae2ac04393b217ea9f700cd04fa9bf3d93fae2872069f3d15d908af70a"
 dependencies = [
  "cc",
 ]


grigoryk:
I don't think `Cargo.lock` should be edited manually - try to update Cargo.toml instead.

However, I'm not sure what we can change ourselves. Looking from the audit failure:
```
openssl-src 111.12.0+1.1.1h
└── openssl-sys 0.9.58
    ├── openssl 0.10.30
    │   └── native-tls 0.2.4
    │       ├── tokio-tls 0.3.1
    │       │   ├── reqwest 0.10.8
    │       │   │   └── viaduct-reqwest 0.1.0
...
```
we see that there isn't a direct dependency on openssl-src, it's getting pulled in via tokio-tls & reqwest.
Looking at `tokio-native-tls`''s dependencies, https://github.com/tokio-rs/tls/blob/master/tokio-native-tls/Cargo.toml#L40 - it depends on `openssl 0.10`, which for some reason doesn't seem like it's not picking up the latest 10.34 (https://github.com/sfackler/rust-openssl/releases) - I assume that would have pulled in the latest version of `openssl-src`...

Can we overwrite parts of the dependency graph w/ cargo? I'm not sure actually, can't find that in the docs.

mhammond:
Yeah, I had a quick look earlier in the week and came to the same conclusion - it's a transitive dependency and hopefully the intermediate crates will have an update soon. I'm also guessing that is why we don't have a dependabot PR for this yet.

(It's certainly true that just updating Cargo.lock isn't the right thing, and I believe Cargo will end up reverting that at some point)

bendk:
I'm not sure if this is the proper way to update sub-dependencies, but this changeset came from `cargo update -p openssl-src`.  I'm way to scared to update Cargo.lock manually.

mhammond:
huh. I did know this but I forgot it. openssl-sys specifies openssl-src as version "111.0.1" - the thing that I forgot is that this is the same as specifying "^111.0.1" which means anything >= "111.0.1" but < "112".

So I think this is good - we should probably update all openssl related packages, but best I can tell, that's not a requirement here.

mhammond (approved):

PR #4074, opened 2021-05-07
Introduce FeatureConfig.value to foreign languages

jhugman:
Fixes [SDK-261].

This introduces a `value` property to `FeatureConfig`.

In this state it builds with the current version of uniffi, and expects value to be a `String`.

Thus: it is not ready to land, since `value` is supposed to be a `serde_json::Value`.

It is ready to review, however.

[1]: https://jira.mozilla.com/browse/SDK-261

  - Note:

### DIFF
--- a/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTest.kt
+++ b/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTest.kt
@@ -184,10 +184,19 @@ class NimbusTest {
                   "schemaVersion": "1.0.0",
                   "slug": "test-experiment",
                   "endDate": null,
+                  "featureIds": ["about_welcome"],
                   "branches": [
                     {
                       "slug": "test-branch",
-                      "ratio": 1
+                      "ratio": 1,
+                      "feature": {
+                          "featureId": "about_welcome",
+                          "enabled": false,
+                          "value": {
+                            "text": "OK then",
+                            "number": 42
+                          }
+                      }
                     }
                   ],
                   "probeSets": [],
@@ -221,6 +230,18 @@ class NimbusTest {
         assertEquals(appInfo.appName, expContext.appName)
         assertEquals(appInfo.channel, expContext.channel)
         // If we could control more of the context here we might be able to better test it
+    }
+
+    @Test
+    fun `Receiving JSON features`() {
+        // Load the experiment in nimbus so and optIn so that it will be active. This is necessary
+        // because recordExposure checks for active experiments before recording.
+        nimbus.setUpTestExperiments(packageName, appInfo)
+
+        val json = nimbus.getFeatureConfigVariablesJson("about_welcome")
+        assertNotNull(json)
+        assertEquals(42, json!!["number"])
+        assertEquals("OK then", json["text"])
     }
 
     @Test
--- a/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTest.kt
+++ b/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTest.kt
@@ -17,6 +17,7 @@ import mozilla.components.service.glean.net.ConceptFetchHttpUploader
 import mozilla.components.service.glean.testing.GleanTestRule
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Rule
@@ -233,15 +234,18 @@ class NimbusTest {
     }
 
     @Test
-    fun `Receiving JSON features`() {
-        // Load the experiment in nimbus so and optIn so that it will be active. This is necessary
-        // because recordExposure checks for active experiments before recording.
+    fun `Smoke test— receiving JSON features`() {
         nimbus.setUpTestExperiments(packageName, appInfo)
-
+        // The test experiment has exactly one branch with 100% enrollment
+        // We should be able to get feature variables for the feature in this
+        // experiment.
         val json = nimbus.getFeatureConfigVariablesJson("about_welcome")
         assertNotNull(json)
         assertEquals(42, json!!["number"])
         assertEquals("OK then", json["text"])
+
+        val json2 = nimbus.getFeatureConfigVariablesJson("non-existent-feature")
+        assertNull(json2)
     }
 
     @Test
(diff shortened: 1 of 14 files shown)


components/nimbus/src/dbcache.rs
```diff
@@ -98,9 +110,19 @@ impl DatabaseCache {
...
     pub fn get_experiment_branch(&self, id: &str) -> Result<Option<String>> {
-        self.get_data(|data| match data.branches_by_feature.get(id) {
+        self.get_data(|data| match data.features_by_feature_id.get(id) {
             None => data.branches_by_experiment.get(id).cloned(),
-            Some(branch_slug) => Some(branch_slug.to_owned()),
+            Some(feature) => Some(feature.branch.clone()),
+        })
+    }
+
+    pub fn get_feature_config_variables(&self, feature_id: &str) -> Result<Option<String>> {
```
jhugman, line 119: This will change to the `EnrolledExperiment` or the `FeatureConfig.value` when uniffi gets these capabilities.

components/nimbus/src/lib.rs
```diff
@@ -373,8 +378,20 @@ pub struct FeatureConfig {
...
+    pub value: String,
+}
+
+fn default_value() -> String {
+    "{}".into()
+}
+
+#[derive(Deserialize, Serialize, Debug, Default, Clone, PartialEq)]
+pub struct FeatureVariables {
+    #[serde(flatten)]
```
dmose, line 393: What's the motivation for flattening here?

dmose (commented):
Looking good so far...

components/nimbus/src/dbcache.rs
jhugman: I'm sorry, I meant `EnrolledFeature`. Although now it won't.

We may consider a further struct to mimic `EnrolledFeature` to go over the FFI with the experiment slug in it.

components/nimbus/src/lib.rs
```diff
@@ -820,3 +837,45 @@ mod test_schema_bw_compat {
...
+    #[test]
+    fn test_deserialize_untyped_json() -> Result<()> {
+        let without_value = serde_json::from_value::<FeatureConfigProposed>(json!(
+            {
+                "featureId": "some_control",
+                "enabled": true,
+            }
+        ))?;
+
+        log::debug!("default value: {:?}", without_value);
```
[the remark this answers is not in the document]
jhugman: Oh yes, I could feasibly delete this test altogether— it was a scratch pad for us using serde_json.

components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTest.kt
```diff
@@ -223,6 +232,18 @@ class NimbusTest {
         // If we could control more of the context here we might be able to better test it
     }
 
+    @Test
+    fun `Receiving JSON features`() {
+        // Load the experiment in nimbus so and optIn so that it will be active. This is necessary
```
travis79, line 237:
```suggestion
        // Load the experiment in nimbus and optIn so that it will be active. This is necessary
```

travis79 (approved):
I'm good with this, provided the checks pass.

components/nimbus/src/dbcache.rs
```diff
@@ -43,25 +44,36 @@ impl DatabaseCache {
...
 
         // Build the new hashmaps.  Note that this is somewhat temporary, is
         // likely to change when the full FeatureConfig stuff is implemented.
         // Further, note that, for the moment, we only (currently) support
         // one feature_id per experiment, meaning that we ignore everything
         // except the first feature_id in the array.  Some of the multi-feature
         // code may want to live in the EnrollmentEvolver.
-        let mut branches_by_experiment = HashMap::with_capacity(experiments.len());
-        let mut branches_by_feature = HashMap::with_capacity(experiments.len());
+        let mut branches_by_experiment = HashMap::with_capacity(enrollments.len());
```
dmose, line 55: I'm wondering if this wants to be called branches_by_enrolled_experiment for clarity as to what's actually going on.

components/nimbus/src/dbcache.rs
```diff
@@ -98,9 +110,20 @@ impl DatabaseCache {
     }
 
     pub fn get_experiment_branch(&self, id: &str) -> Result<Option<String>> {
```
dmose, line 112: Similarly, I'm guessing this wants to be get_enrolled_experiment_branch.  I also suspect we don't really want to make that change now since it would be a breaking change, but at least some comments could be helpful, and maybe a spinoff ticket if we think the method is likely to survive long enough to be worth renaming.

components/nimbus/src/dbcache.rs
dmose: I can't find an `EnrolledFeature`.  Perhaps `EnrolledFeatureConfig` is what you mean?

I'm assuming you're talking about at some future time, and not as part of this PR.  Correct?

components/nimbus/src/nimbus.udl
```diff
@@ -110,9 +104,12 @@ interface NimbusClient {
     [Throws=NimbusError]
     string? get_experiment_branch(string id);
 
+    [Throws=NimbusError]
+    string? get_feature_config_variables(string feature_id);
```
dmose, line 108: Is this waiting on a new uniffi dependency in order to switch to a JSONValue?  If so, will that happen in this PR or a different one?

components/nimbus/src/dbcache.rs
dmose: Similarly, I suppose features_by_feature_id might want be to enrolled_features_by_feature_id

dmose (approved):
r=dmose on the rust stuff, with the comments addressed as you feel appropriate.

components/nimbus/src/dbcache.rs
jhugman: Yes.

components/nimbus/src/dbcache.rs
jhugman: If the user has a branch for an experiment, then they're by definition enrolled.

components/nimbus/src/dbcache.rs
jhugman: Yes, you're right: this would be a breaking change :( Having said that, I think this method will becoming less important, though likely not going away.

components/nimbus/src/nimbus.udl
jhugman: I don't think this is going to get a JSON Value in time, so we're dropping down to using a `String`. :(

PR #4076, opened 2021-05-07
Enable consuming applications to change the server `collection` being used

travis79:
This removes the unnecessary `bucket` from the server config, as well as adds `collection` to the Swift `NimbusServerSettings` to match the Kotlin `NimbusServerSettings`.  

This should allow a consuming application to start-up the Nimbus-SDK pointing at the `nimbus-preview` collection for QA testing purposes.

### DIFF
--- a/components/nimbus/ios/Nimbus/NimbusApi.swift
+++ b/components/nimbus/ios/Nimbus/NimbusApi.swift
@@ -132,12 +132,16 @@ public extension Notification.Name {
 /// This struct is used during in the `create` method to point `Nimbus` at the given `RemoteSettings` server.
 ///
 public struct NimbusServerSettings {
-    public init(url: URL) {
+    public init(url: URL, collection: String = remoteSettingsCollection) {
         self.url = url
+        self.collection = collection
     }
 
     public let url: URL
+    public let collection: String
 }
+
+public let remoteSettingsCollection = "nimbus-mobile-experiments"
 
 /// Name and channel of the app, which should agree with what is specified in Experimenter.
 ///
--- a/components/nimbus/ios/Nimbus/NimbusCreate.swift
+++ b/components/nimbus/ios/Nimbus/NimbusCreate.swift
@@ -4,9 +4,6 @@
 
 import Foundation
 
-private let remoteSettingsBucket = "main"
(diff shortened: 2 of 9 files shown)


components/nimbus/src/client/http_client.rs
```diff
@@ -112,8 +110,8 @@ impl SettingsClient for Client {
 
     fn fetch_experiments(&self) -> Result<Vec<Experiment>> {
         let path = format!(
-            "buckets/{}/collections/{}/records",
-            &self.bucket_name, &self.collection_name
+            "buckets/main/collections/{}/records",
```
jhugman, line 113: Is bucket _always_ going to be `main`?

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -132,13 +132,17 @@ public extension Notification.Name {
 /// This struct is used during in the `create` method to point `Nimbus` at the given `RemoteSettings` server.
 ///
 public struct NimbusServerSettings {
-    public init(url: URL) {
+    public init(url: URL, collection: String = remoteSettingsCollection) {
         self.url = url
+        self.collection = collection
```
jhugman, line 137: Can we unify the parameter names between kotlin and swift at the same time?

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -45,11 +42,9 @@ public extension Nimbus {
 
         let context = Nimbus.buildExperimentContext(appSettings)
         let remoteSettings = server.map { server -> RemoteSettingsConfig in
-            let url = server.url.absoluteString
             return RemoteSettingsConfig(
-                serverUrl: url,
-                bucketName: remoteSettingsBucket,
-                collectionName: remoteSettingsCollection
+                serverUrl: server.url.absoluteString,
+                collectionName: server.collection
```
jhugman, line 47: Can we unify the property names?

jhugman (approved):
I have a couple of nits, but this is looking sweet.

I believe this is the first 4 language PR in the Nimbus project. 😻

PR #4078, opened 2021-05-10
migrate the database to drop experiments, enrollments, and maybe updates w/o features or featureIds 

dmose:
This PR has a bunch of tests related to Experiments and ExperimentEnrollments that have malformed feature-related stuff and a database, along with a database migrator that discards experiments & enrollments that have these problems.

  - Note:
  - *None added*


components/nimbus/src/persistence.rs
```diff
@@ -176,6 +179,38 @@ impl SingleStore {
...
+                match unserialized {
+                    Ok(value) => result.push(value),
+                    Err(e) => {
+                        // If there is an error, we won't push this onto the
+                        // result Vec, but we won't blow up the entire
+                        // deserialization either.
+                        log::warn!(
+                            "try_collect_all: discarded a record while deserializing with: {:?}",
+                            e
+                        );
```
travis79, line 206: What do you think about dropping a `log::debug!` with the contents that we couldn't deserialize?

components/nimbus/src/persistence.rs
```diff
@@ -227,13 +262,93 @@ impl Database {
...
+        log::debug!("entered maybe upgrade");
         let mut writer = self.rkv.write()?;
         let db_version = self.meta_store.get::<u16, _>(&writer, DB_KEY_DB_VERSION)?;
         match db_version {
             Some(DB_VERSION) => {
                 // Already at the current version, no migration required.
                 return Ok(());
             }
+            Some(1) => {
+                log::debug!("Upgrading from version 1 to version 2");
```
travis79, line 274: This seems like it might be worth upgrading to an `info` to me.  It shouldn't be happening enough to be too chatty, I hope.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +559,503 @@ mod tests {
...
+    //         },
+    //         "userFacingName":"Diagnostic test experiment",
+    //         "referenceBranch":"control",
+    //         "isEnrollmentPaused":false,
+    //         "proposedEnrollment":7,
+    //         "userFacingDescription":"This is a test experiment for diagnostic purposes.",
+    //         "id":"secure-gold", // change when cloning
+    //         "last_modified":1_602_197_324_372i64
+    //     })]
+    // }
```
travis79, line 608: If this isn't being used, let's get rid of it.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +559,503 @@ mod tests {
...
+    //                         {
+    //                             "enrollment_id": "801ee64b-0b1b-44a7-be47-5f1b5c189083", // change on cloning
+    //                             "reason": "Qualified",
+    //                             "branch": "control",
+    //                             "feature_id": "about_welcome" // change on cloning
+    //                         }
+    //                     }
+    //                 }
+    //     )]
+    // }
```
travis79, line 965: nit: Let's get rid of this if it isn't needed

travis79 (commented):
The test is great! Just what I imagined.

components/nimbus/src/persistence.rs
dmose: Done.

components/nimbus/src/persistence.rs
dmose: All the stuff that's commented out will either be morphed into something that is used or removed before landing.

components/nimbus/src/persistence.rs
dmose: Similarly, this will all morphed into something that is used or dropped before landing.

components/nimbus/src/persistence.rs
dmose: Assuming I stick with the current migrator strategy; I like it!

dmose: I still need to look into this.

CHANGES_UNRELEASED.md
```diff
@@ -13,4 +13,12 @@
...
  - Android and iOS `Branch` objects no longer have access to a `FeatureConfig` object.
 
+### ⚠️ Breaking changes ⚠️
+- The experiment database will be migrating from version 1 to version 2 on
+  first run.  Various kinds of incorrectly specified feature and featureId
+  related fields will be detected, and any related experiments/enrollments will
+  be discarded.  If there is an error during the database upgrade, the database
+  will be wiped, since losing existing enrollments is still less bad than
+  having the database in an unknown inconsistent state.
+
```
dmose, line 23: Is there an easy way to ensure that whenever a database gets wiped, an event gets sent to sentry?  That would be wonderful to monitor in Nightly and Beta as a way to be sure that we don't need to uplift anything.

components/nimbus/src/lib.rs
```diff
@@ -841,6 +841,184 @@ mod test_schema_bw_compat {
         assert_eq!(exp.app_id, Some("org.mozilla.fenix".to_string()));
         assert_eq!(exp.channel, Some("nightly".to_string()));
     }
+
+    #[test]
+    /// Migrating v1 to v2 involves finding enrollments and experiments that
+    /// don't contain all the feature_id stuff they should and discarding.
+    fn test_migrate_v1_to_v2() -> Result<()> {
```
dmose, line 848: This a first attempt before I realized I needed to be doing these tests in bulk.  I'm inclined to get rid of it, since it's pretty duplicative, however it is a bit of extra coverage which uses slightly different APIs and further more tests valid and invalid stuff in the same batch.  Thoughts?

components/nimbus/src/persistence.rs
```diff
@@ -257,6 +312,104 @@ impl Database {
...
+        let slugs_without_enrollment_feature_ids: HashSet<String> = enrollments
+            .iter()
+            .filter_map(
+                    |e| {
+                if matches!(e.status, EnrollmentStatus::Enrolled {ref feature_id, ..} if feature_id.is_empty()) {
+                    log::warn!("Enrollment for {:?} missing feature_ids; experiment & enrollment will be discarded", &e.slug);
+                    Some(e.slug.to_owned())
+                } else {
+                    None
+                }
```
jhugman, line 347: I am nervous about this for future migrations. Thinking aloud:

1. `ExperimentEnrollment` is an internal data structure, so we can change it if we want.
2. Enrollments are becoming increasingly detached from feature id (our recent trip into `evolve_enrollments` showed us that features and experiments are deliberately orthogonal.
3. We will want to support multiple `feature_ids`, so it may be an idea to remove `feature_id` and replace it with`feature_ids` in enrollments, or remove it from enrollments altogether.
4. This code will not work as a transitory state between from 1 to (a currently futuristic) 3.

I'm not sure there's anything to do right now, but we should be aware of this if/when we remove `feature_id` from enrollment.

components/nimbus/src/lib.rs
jhugman: I agree. 

I've suggested a `create_old_database` method below, to help recreate an old version of the database. Perhaps when you write that you could revisit this, refactoring it in terms of that method.

jhugman (commented):
This PR has some thorough testing, good stuff. You've gone the extra mile to make sure that the migration is done in the appropriate layer, too.

I've made some suggestions for you below. I've tried to make them as easily actionable as possible; it might seem a lot, but the largest things are extracting things into methods.

Comments: I'd like to try and get rid of some of the `XXX` and conversational style comment, i.e. asking questions in comments. I think self-review is a better place to ask the questions, and the comments to say what the code _is_ doing.

I would also be really helpful if you could grab a version of the JSON as it is coming from experimenter now. Having an archive of different versions of the actual JSON as multi-line strings in a `get_v2_experiments()` will make it writing future migrations a lot easier.

I don't think this is quite ready to land, but we're getting close. Please ping me if things aren't clear.

CHANGES_UNRELEASED.md
dmose: @jhugman I seem to recall you mentioning having context around how the errors are plumbed through outside of Rust to Sentry (or not) in the embedding apps.  Do you know the answer to this offhand?

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+        let rkv = Database::open_rkv(&tmp_dir)?;
+        let meta_store = SingleStore::new(rkv.open_single("meta", StoreOptions::create())?);
+        let experiment_store =
+            SingleStore::new(rkv.open_single("experiments", StoreOptions::create())?);
+        let mut writer = rkv.write()?;
+
+        meta_store.put(&mut writer, "db_version", &1)?;
+
+        // write a bunch of invalid experiments
+        let invalid_feature_experiments = &get_invalid_feature_experiments();
```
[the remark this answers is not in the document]
dmose: I'm thinking I'll go with v1_db_schema_experiments, since it's otherwise easy to confuse the db schema with the DTO schema (which would maybe ideally be the same thing, but I'm concerned that coupling them too tightly is not the thing to take an now, if we want to do it at all).

dmose:
> This PR has some thorough testing, good stuff. You've gone the extra mile to make sure that the migration is done in the appropriate layer, too.

Thanks!

> I've made some suggestions for you below. I've tried to make them as easily actionable as possible; it might seem a lot, but the largest things are extracting things into methods.

Those are great -- I wanted to be sure to get something up for review, so I didn't go through all the method extraction stuff once it started getting late. Happy to do them now.

> Comments: I'd like to try and get rid of some of the `XXX` and conversational style comment, i.e. asking questions in comments. I think self-review is a better place to ask the questions, and the comments to say what the code _is_ doing.

Some of those comments were really just intended as self-review, not something that would land; agreed that the PR is a better place for them going forward.

> I would also be really helpful if you could grab a version of the JSON as it is coming from experimenter now. Having an archive of different versions of the actual JSON as multi-line strings in a `get_v2_experiments()` will make it writing future migrations a lot easier.

That does sound like a fine thing.  I'm curious as to your thoughts about db schema version vs experimenter schema version.  I'm inclined to version by the experimenter schema number (ie schemaVersion) and then map schema version to db version in the code somewhere, maybe in the migrator.

components/nimbus/src/persistence.rs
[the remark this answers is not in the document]
dmose: (The 1 has already been chosen as the existing DB_VERSION)

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+                "branches": [
+                    {
+                    "slug": "default_browser_newtab_banner",
+                    "ratio": 100,
+                    "feature": {
+                        "featureId": "fenix-default-browser",
+                        "enabled": true,
+                        "value": {}
+                    }
+                    },
```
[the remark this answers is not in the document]
dmose: Fixed.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+                "schemaVersion": "1.0.0",
+                "slug": "branch-feature-feature-id-missing", // change when cloning
+                "endDate": null,
+                "featureIds": ["ccc"], // change when cloning
+                "branches":[
+                    {
+                        "slug": "control",
+                        "ratio": 1,
+                        "feature": {
+                            "featureId": "ccc", // change when cloning
```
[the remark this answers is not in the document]
dmose: Good idea; done.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
         assert_ne!(fs::metadata(&db_file)?.len(), garbage_len);
         Ok(())
     }
+
+    // XXX secure-gold has some fields. Ideally, we would also have an
+    // experiment with all current fields set, and another with almost no
+    // optional fields set
+    fn get_valid_feature_experiments() -> Vec<serde_json::Value> {
```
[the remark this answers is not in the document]
dmose: Done.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+                "referenceBranch": "control",
+                "featureIds": [
+                    "fenix-default-browser"
+                ]
+            }),
+        ]
+    }
+    /// Each of this should uniquely reference a single experiment returned
+    /// from get_valid_feature_experiments
+    fn get_valid_feature_enrollments() -> Vec<serde_json::Value> {
```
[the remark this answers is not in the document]
dmose: Done

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+                                "reason": "Qualified",
+                                "branch": "control",
+                                "feature_id": "abc" // change on cloning
+                            }
+                        }
+                    }
+        )]
+    }
+
+    fn get_invalid_feature_experiments() -> Vec<serde_json::Value> {
```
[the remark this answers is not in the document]
dmose: Done.

components/nimbus/src/persistence.rs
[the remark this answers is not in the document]
dmose: Went with `db_v1_experiments_with_missing_feature_fields`.  The version numbers have already been set in them migration code since before this patch.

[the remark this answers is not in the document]
dmose: Fixed.

components/nimbus/src/persistence.rs
```diff
@@ -227,13 +260,35 @@ impl Database {
...
+
+                        // XXX I feel like I must be missing something obvious
+                        // here, but how to test this clause doesn't jump out
+                        // at me.  The ways I've thought of so far to cause
+                        // errors would either be caught in open_rkv or one of
+                        // the other clauses in this function, or discarded
+                        // with a warning in try_collect_all.
+                        self.meta_store.clear(&mut writer)?;
+                        self.experiment_store.clear(&mut writer)?;
+                        self.enrollment_store.clear(&mut writer)?;
```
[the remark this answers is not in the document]
dmose: self.clear_all_db_stores_except_updates() has been extracted.

components/nimbus/src/persistence.rs
dmose: Done!

CHANGES_UNRELEASED.md
```diff
@@ -13,4 +13,12 @@
 
  - Android and iOS `Branch` objects no longer have access to a `FeatureConfig` object.
 
+### ⚠️ Breaking changes ⚠️
+- The experiment database will be migrating from version 1 to version 2 on
+  first run.  Various kinds of incorrectly specified feature and featureId
+  related fields will be detected, and any related experiments/enrollments will
+  be discarded.  If there is an error during the database upgrade, the database
+  will be wiped, since losing existing enrollments is still less bad than
+  having the database in an unknown inconsistent state.
```
dmose, line 22: Reminder: should add that other things with missing required fields (eg schemaVersion) will also be discard.

components/nimbus/src/persistence.rs
```diff
@@ -257,6 +312,104 @@ impl Database {
...
+                }
+
+                let feature_ids_arrays_with_empty_strings =
+                    e.feature_ids.iter().find(|id| id.is_empty());
+                if e.feature_ids.is_empty() || feature_ids_arrays_with_empty_strings != None {
+                    log::warn!("{:?} experiment has invalid feature_ids array; experiment & enrollment will be discarded", &e.slug);
+                    Some(e.slug.to_owned())
+                } else {
+                    None
+                }
```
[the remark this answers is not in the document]
dmose: Looks rustier and less repetitive.  r=dmose; thanks!

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+                    "namespace":"branch-feature-feature-id-empty", // change when cloning
+                    "randomizationUnit":"nimbus_id"
+                },
+                "userFacingName":"Diagnostic test experiment",
+                "referenceBranch":"control",
+                "isEnrollmentPaused":false,
+                "proposedEnrollment":7,
+                "userFacingDescription":"This is a test experiment for diagnostic purposes.",
+            }),
+            // XXX We should probably just allow branch-feature-values to be
```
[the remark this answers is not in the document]
dmose: Nope; as per your second comment, I've removed it.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +597,729 @@ mod tests {
...
+            // non-existent and delete this clause, because the value field
+            // itself can be {}, and the FeatureConfig definition of `value`
+            // (see enrollment.rs) uses #[serde(default)] on (presumably to
+            // cope with the {}).  I believe this strategy should be fine,
+            // regardless of what branch-feature-values are out there in the
+            // wild.
+            //
+            // It's also not clear how one would easily tell the two cases
+            // apart in the migrator code anyhow without removing the
+            // #[serde(default)].  @jhugman, your thoughts?
```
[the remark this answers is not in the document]
dmose: Agreed; removed.

components/nimbus/src/persistence.rs
dmose: Good point.

CHANGES_UNRELEASED.md
dmose: https://jira.mozilla.com/browse/SDK-287 filed for this.

dmose:
This should be more or less ready to go.  I'll probably file a few spin-off tickets tomorrow (Thursday).

components/nimbus/src/lib.rs
travis79: I like the idea of being able to recreate old versions of the schema for testing migrations. At one point I was hoping that enum variants in Rust would be the way to go with this, but just having a test method to create an old version of the database is probably simpler to implement and understand.

components/nimbus/src/persistence.rs
```diff
@@ -257,6 +311,113 @@ impl Database {
...
+
+    fn clear_all_db_stores_except_updates(
+        &self,
+        writer: &mut rkv::Writer<rkv::backend::SafeModeRwTransaction>,
+    ) -> Result<(), NimbusError> {
+        self.meta_store.clear(writer)?;
+        self.experiment_store.clear(writer)?;
+        self.enrollment_store.clear(writer)?;
+        Ok(())
+    }
```
travis79, line 326: I'd like to point out that clearing the meta_store is wiping the database version, the nimbus_id, and whether or not the user has opted out of experiments globally, which I don't know if it's entirely necessary or desired.

travis79: Also, it is a design requirement of Nimbus to try and report experiments we have been enrolled in for some time after the enrollment has ended (30 days, I think), and this would potentially wipe away that information.  Currently I don't think that there are any real experiments that have the potential for this to happen in, so maybe it's just more of a "let's don't forget".

dmose:
The test and clippy failures are actually the same thing -- need different cfg stuff from lmdb and safe-mode database backends.

dmose:
Should be straightforward.

components/nimbus/src/persistence.rs
```diff
@@ -442,6 +604,668 @@ mod tests {
...
+                "isEnrollmentPaused": false,
+                "bucketConfig": {
+                    "randomizationUnit": "nimbus_id",
+                    "namespace": "fenix-default-browser-4",
+                    "start": 0,
+                    "count": 10000,
+                    "total": 10000
+                },
+                "probeSets": [],
+                // "outcomes": [], NOT CURRENTLY (YET?) IMPLEMENTED
```
travis79, line 678: outcomes is an analysis specific piece of the schema and not of interest to the SDK.

travis79 (changes requested):
I'm withholding final r+ due to my concerns about wiping the meta_store from the database.  I'm concerned that just wiping it completely is deleting more than we necessarily need to for the purpose of cleaning up a bad migration.

components/nimbus/src/persistence.rs
travis79: I'm not as concerned about the historical reporting of experiments as I am about losing the global opt out setting specifically, and less specifically the nimbus_id.  If we dump this database store then could we potentially opt a user back into experiments if they had previously opted out, is my fear.

Rotating the nimbus_id isn't the end of the world, but we should avoid doing it unless we have a need as it could potentially affect the user's enrollment bucketing.

components/nimbus/src/persistence.rs
dmose: OK, update the coment.

components/nimbus/src/persistence.rs
dmose: Fixed.

travis79 (approved):
With my concerns addressed, I'm onboard with landing this!  Congrats and thanks for the hard work on this!

components/nimbus/src/lib.rs
dmose: Refactored this test to use create_old_database.  I'll file a separate ticket where we can discuss how best to manage old database schemas.

PR #4093, opened 2021-05-12
add osx branch to protobuf artifacts to get only x86_64

### LINKED ISSUE #4092: ./gradlew test fails on Apple M1 due to not finding correct protobuf binaries
In any component that uses protobuf, there is a gradle plugin that's used that looks like:

{{rotobuf {
    protoc {            artifact = 'com.google.protobuf:protoc:3.11.4'
        }
    }
    generateProtoTasks {
        all().each { task ->
            task.builtins {
                java {
                    option "lite"
                }
            }
        }
    }
}}}

produces an error trying to run the gradle files:

{{   > Could not find protoc-3.11.4-osx-aarch_64.exe (com.google.protobuf:protoc:3.11.4).
     Searched in the following locations:
         https://jcenter.bintray.com/com/google/protobuf/protoc/3.11.4/protoc-3.11.4-osx-aarch_64.exe}}

While protobuf can compile successfully on Apple M1, protobuf doesn't have compiled binaries pushed to maven which causes the error above.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2201)
┆Story Points: 2
┆Sprint: SYNC - 2021-05-28

skhamis:
Fixes #4092 

Apple M1 can use the x86_64 binaries fine but it's difficult to have a locally built version of protobuf and try connect it to gradle. This allows all macos systems to just grab the x86_64 version instead of trying to look for the ARM. We can remove this code once Protobuf starts pushing osx-aarch-64 to maven or we get rid of all protobufs from uniffi.

- [ X] **Quality**: This PR builds and tests run cleanly
  - Note:
- [] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one
- [] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)

### DIFF
--- a/components/logins/android/build.gradle
+++ b/components/logins/android/build.gradle
@@ -80,7 +80,13 @@ configurations {
 
 protobuf {
     protoc {
-        artifact = 'com.google.protobuf:protoc:3.11.4'
+        // This is a hack for Apple M1 since protobuf does not compile for M1s yet
+        // However, M1s can just use the x86_64 just fine
+        if (osdetector.os == "osx") {
+            artifact = 'com.google.protobuf:protoc:3.11.4:osx-x86_64'
+        } else {
+            artifact = 'com.google.protobuf:protoc:3.11.4'
+        }
     }
     generateProtoTasks {
         all().each { task ->
--- a/components/places/android/build.gradle
+++ b/components/places/android/build.gradle
@@ -77,7 +77,13 @@ configurations {
 }
 protobuf {
     protoc {
-        artifact = 'com.google.protobuf:protoc:3.11.4'
+        // This is a hack for Apple M1 since protobuf does not compile for M1s yet
+        // However, M1s can just use the x86_64 just fine
+        if (osdetector.os == "osx") {
+            artifact = 'com.google.protobuf:protoc:3.11.4:osx-x86_64'
+        } else {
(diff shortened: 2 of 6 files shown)


mhammond:
This looks like a reasonable work-around to me, but Ryan might have better ideas!

rfk (approved):
Looks good to me! Given our ambition for protobuf to go away over time, we should optimize for "keep it running" rather than "keep it nice", so no objections to a bit of hackery here.

(That we have to repeat this in so many places, is a bigger problem that I've been giving a bit of through to addressing, but certainly not for this PR)

skhamis:
> Looks good to me! Given our ambition for protobuf to go away over time, we should optimize for "keep it running" rather than "keep it nice", so no objections to a bit of hackery here.
> 
> (That we have to repeat this in so many places, is a bigger problem that I've been giving a bit of through to addressing, but certainly not for this PR)

I agree I was trying to think of an elegant solution but it started to get crazy....given that this only affects Macos/Apple silicon and can actually be removed if/when protobuf starts pushing M1 binaries as well. Thanks for the review both!

Issue #4085, opened 2021-05-12
`./gradlew publishToMavenLocal` failing with nimbus errors


mhammond:
I'm out of time to dig more tonight, but @skhamis and I struck the error below running `./gradlew publishToMavenLocal`.

It might just be something local - but if it's not, I think we have 2 issues here: the error below, and that CI didn't grab it. Can anyone else repro this?

```
> Task :nimbus:compileReleaseKotlin FAILED
e: /mnt/c/src/moz/application-services/components/nimbus/android/build/generated/source/uniffi/release/java/org/mozilla/experiments/nimbus/nimbus.kt: (1028, 12): Redeclaration: EnrolledExperiment
e: /mnt/c/src/moz/application-services/components/nimbus/android/build/generated/source/uniffi/release/java/org/mozilla/experiments/nimbus/nimbus.kt: (1068, 12): Redeclaration: Branch
e: /mnt/c/src/moz/application-services/components/nimbus/android/build/generated/source/uniffi/release/java/org/mozilla/experiments/nimbus/nimbus.kt: (1099, 12): Redeclaration: FeatureConfig
e: /mnt/c/src/moz/application-services/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt: (42, 11): Redeclaration: Branch
e: /mnt/c/src/moz/application-services/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt: (44, 11): Redeclaration: EnrolledExperiment
e: /mnt/c/src/moz/application-services/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt: (45, 11): Redeclaration: FeatureConfig

FAILURE: Build failed with an exception.
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2196)
┆Sprint: Backlog

skhamis:
Interestingly enough I cannot repo this on the latest main

mhammond:
A different checkout doesn't repro for me either. I'll try and work out why but this can be closed.

PR #4094, opened 2021-05-13
Add a clean.py helper script

mhammond:
I think if I had this I would not have opened #4085 - so here it is :)

I stole some of this from Ben's recent .py script. Having it in `automation` seems a little wrong as it's never going to be run by automation, but `libs` seems wrong for different reasons and I can't see a sane alternative.

(There are more things we could do in the future, such as a "clobber" mode which might kill things under libs, and I know it's not going to clean stuff generated for iOS, but this seems like an OK start?)

CC @dmose who IIRC was asking for one of these in the past too.

### DIFF
--- a/tools/clean.py
+++ b/tools/clean.py
@@ -1,0 +1,69 @@
+#!/usr/bin/python3
+"""
+Clean the world.
+
+Note that this is probably only a partial clean - in particular, no attempt
+is made to clean the iOS/Swift world, because markh doesn't know what that
+involves.
+
+Please make a PR with anything you notice that should be cleaned but isn't!
+"""
+
+import argparse
+import subprocess
+from pathlib import Path
+import shlex
+import shutil
+
+PROJECT_ROOT = Path(__file__).parent.parent
+
+def run_command(dry_run, cmdline, **kwargs):
+    print('Executing:', ' '.join(shlex.quote(str(part)) for part in cmdline))
+    if not dry_run:
+        subprocess.check_call(cmdline, **kwargs)
+
+def find_generated_directories(look_dir):
+    for child in look_dir.iterdir():
+        if child.name == 'support':
+            for sub in find_generated_directories(child):
+                yield sub
+        else:
+            # `android/build` directories should be removed.
+            sub = child / 'android' / 'build'
+            if sub.is_dir():
+                yield sub
+            # TODO: ios/swift?
+
+def clean_android(dry_run):
+    # pathlib.Path will join "." and "gradlew" as "gradlew", which doesn't
+    # work as "." is not on the path!
+    gradlew = (PROJECT_ROOT / "gradlew").resolve()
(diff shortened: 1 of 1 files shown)


mhammond:
>  and I can't see a sane alternative.

Far out, I didn't look very hard, did I? :) ./tools exists so I put it there.

mhammond:
in slack, @jhugman points out `./gradlew clean` exists 👍

docs/metrics/logins/metrics.md
```diff
@@ -1,32 +1,33 @@
 <!-- AUTOGENERATED BY glean_parser.  DO NOT EDIT. -->
 
 # Metrics
+
```
mhammond, line 4: oh ffs :(

bendk (approved):
Looks good to me.  Makes me want to move tests.py to tools too.

mhammond:
Ryan, do you mind having a quick look at this and seeing if there's anything obvious I missed?

rfk:
> Having it in automation seems a little wrong as it's never going to be run by automation,
> but libs seems wrong for different reasons and I can't see a sane alternative.

FWIW, my mental model is that `automation` is for things that will be run by a computer, and `tools` is for things that will be run by a human. I think we have some scripts that violate this in both directions at the moment, but would be happy to help move them around if consistency here seems valuable.

tools/clean.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/python3
+"""
+Clean the world.
+
+Note that this is probably only a partial clean - in particular, no attempt
+is made to clean the iOS/Swift world, because markh doesn't know what that
+involves.
```
rfk, line 7: It's possible in theory to clean an xcode project from the command-line using `xcodebuild`, see e.g. [here](https://stackoverflow.com/questions/19091769/clean-ios-xcode-target-in-command-line). May be worth a followup issue for someone to poke at.

tools/clean.py
```diff
@@ -0,0 +1,58 @@
...
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('-n', '--dry-run', dest="dry_run", action="store_true",
+                        help='show what would be executed/removed without actually doing it.')
+    return parser.parse_args()
+
+def main():
+    args = parse_args()
+    run_command(args.dry_run, ["cargo", "clean"])
```
rfk, line 53: You mentioned in a comment about `./gradlew clean` and it seems worth adding here. In fact if there's android-specific stuff that *doesn't* get cleaned by `./gradlew clean`, it may be worth our time to try to fix that command to do it rather than working around it here.  (Or it may not, gradle's always an adventure...)

rfk (approved):
Looks like a great start, I'm strongly in favour of merging and iterating as we discover more stuff.

PR #4098, opened 2021-05-13
Removed auto-generated metrics.md files, fixed error in publishToLocalMaven with directories

skhamis:
This PR has two parts (they might seem disparate, but not so much!)

Using auto-publishing workflow in a-c or fenix with a-s was failing when the nimbus metrics markdown file was generated because it was a new file rather than updating the a previous generated file. Thus when `./gradlew publishToMavenLocal` tries to call `with open(nm, "rb") as f:` on the changes it found in `git status` but fails because it's a directory.

1. Catching that error prevents this kind of mishap in the future
2. Added in the auto-gen glean metrics to prevent people from having to discard those changes locally and/or add it to an unrelated commit

  - Note:


docs/metrics/nimbus/metrics.md
```diff
@@ -0,0 +1,33 @@
+<!-- AUTOGENERATED BY glean_parser.  DO NOT EDIT. -->
```
mhammond, line 1: I suspect the changes to all the other metrics.md files are due to glean being updated - but I can't explain why this new file!

skhamis: It's definitely a little weird how it would pop up as a brand new file. Especially since it generates even on a clean branch as well

mhammond:
Good and bad news. Good in general, bad for this PR :) #glean tells me these generated .md files are, basically, deprecated, so I think instead we should disable their generation and remove the .md files.

[logins is enabled here](https://github.com/mozilla/application-services/blob/f1871d904fd7d96105dc5b2944dcc03abfb3abbc/components/logins/android/build.gradle#L98) etc.

cc @travis79; does nuking `metrics.md` files and their generation sound ok?

travis79:
> Good and bad news. Good in general, bad for this PR :) #glean tells me these generated .md files are, basically, deprecated, so I think instead we should disable their generation and remove the .md files.
> 
> [logins is enabled here](https://github.com/mozilla/application-services/blob/f1871d904fd7d96105dc5b2944dcc03abfb3abbc/components/logins/android/build.gradle#L98) etc.
> 
> cc @travis79; does nuking `metrics.md` files and their generation sound ok?

Yep, that is the recommended thing to do now that the Glean Dictionary covers all the same info.  Should be safe to remove the generated metrics.md files and disable their generation.

skhamis:
Thanks both! I think it's much clearer now why there is some inconsistencies with these files. In an effort to keep the same context I'll go ahead and update this PR (but let me know if I should make a new one completely).

skhamis:
@travis79 or even @mhammond  Does it make sense to remove the metrics.yaml files in these respective directories as well [logins example](https://github.com/mozilla/application-services/blob/f1871d904fd7d96105dc5b2944dcc03abfb3abbc/components/logins/android/metrics.yaml)? Looking at more updated components they don't have even the metrics.yaml and I tried looking into the Glean Dictionary and seems to concur but never know if there is something I missed.

mhammond:
> @travis79 or even @mhammond Does it make sense to remove the metrics.yaml files in these respective directories as well 

Nope, they must stay - they define the metrics we emit.

> Looking at more updated components they don't have even the metrics.yaml

That's because they don't emit metrics themselves. Now we have uniffi, we tend to have more light-weight Android code next to the component (if we have any at all!) and currently it's really only the mobile code that can emit metrics directly. We soon hope to have Glean available directly to our Rust code, which will mean all these newer components will be able to emit metrics without going via Android, so will get these .yaml files then.

mhammond (approved):
Thanks!

PR #4103, opened 2021-05-14
pin more python package versions to satisfy --require-hashes

mhammond:
welp, that will teach me to merge dependabot PRs :/ Let's see if this makes things green again...


mhammond:
Let's fight this yak another day.

PR #4105, opened 2021-05-14
Run cargo update and fix related license/dependency bustage

mhammond:
I suspect this will also fix code coverage.

Narrator: it did not. The reason I thought it might is that our coverage task is failing with:

```
error[E0557]: feature has been removed
  --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/extprim-1.7.1/src/lib.rs:58:81
   |
58 | #![cfg_attr(extprim_channel="unstable", feature(llvm_asm, test, specialization, const_fn))]
   |                                                                                 ^^^^^^^^ feature has been removed
   |
```
and thought I saw an update that might address this, but I misread (even though a package `const_fn` has been removed, I think that's just a coincidence and extprim isn't touched)

And the dependency checker seems to hate me for glean-shaped reasons I'm still trying to work out.

### DIFF
--- a/tools/dependency_summary.py
+++ b/tools/dependency_summary.py
@@ -260,6 +260,15 @@ PACKAGE_METADATA_FIXUPS = {
             "fixup": "https://raw.githubusercontent.com/rust-lang/futures-rs/master/LICENSE-APACHE",
         }
     },
+    "typenum": {
+        "license": {
+            "check": "MIT OR Apache-2.0"
+        },
+        "license_file": {
+            "check": None,
+            "fixup": "LICENSE-APACHE",
+        }
+    },
     # These packages do not include their license file in their release distributions,
     # so we have to fetch it over the network. Each has been manually checked and resolved
     # to a final URL from which the file can be fetched (typically based on the *name* of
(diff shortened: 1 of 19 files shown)


mhammond:
The `into_string()` changes are unfortunate, but sadly that was deprecated in the same version that the new technique was introduced - so we can't land that part before the cargo update (as the `String::from()` changes do not work) and can't land it after (as clippy will fail). I guess we could `allow` that lint failure, but I figured we might as well fix it.

tools/dependency_summary.py
```diff
@@ -669,19 +723,6 @@
             "fixup": "https://raw.githubusercontent.com/LeopoldArkham/humansize/master/LICENSE-APACHE",
         }
     },
-    # cpuid-bool has been renamed to cpufeatures - it's a transitive dependency, so
-    # we expect this to become unnecessary as the dependent crate updates.
```
rfk, line 673: Oh hey, nice when a little wish from a code comment actually comes to pass 😁

rfk (approved):
This LGTM. Thanks for digging through the thorny details here @mhammond!

PR #4111, opened 2021-05-17
Enable the `bundled` feature of `rusqlite` when building `support/guid`.

rfk:
Most of our other crates enable either `bundled` or `sqlite` when building
`rusqlite`, but the `support/guid` crate does neither. This seems to result
in build errors when the user doesn't have a system-level build of sqlite
available. Let's be consistent across all the crates.

Unless this is going to subtly break something in other parts of the build..?

### DIFF
--- a/components/support/guid/Cargo.toml
+++ b/components/support/guid/Cargo.toml
@@ -6,7 +6,7 @@ license = "MPL-2.0"
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


mhammond (approved):

rfk:
I'm going to go ahead and merge this as it blocks some other PRs, but @dmose I would still welcome you weighing in if this sets off any alarm bells about sqlite+rust build errors.

PR #4115, opened 2021-05-17
Use `&mut self` instead of a RefCell for the autofill local encryption key.

mhammond:
Ben asked why we used a RefCell and there is no good answer to that
question :)

  - Note:


bendk (approved):
LGTM

mhammond:
I thought I wrote this test failure up: I tried to fix the failure here but ended up giving up. It *should* be possible as there's no shared reference when we want to call that function, but Rust being Rust liked making it difficult. Feel free to pick this up!

bendk:
I think I have this working in #4126

PR #4120, opened 2021-05-18
Add manual feature exposure recording

travis79:


components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -48,9 +48,9 @@ private extension Nimbus {
 
 // Glean integration
 internal extension Nimbus {
-    func recordExposure(experimentId: String) {
+    func recordExposure(_ featureId: String) {
         let activeExperiments = getActiveExperiments()
-        if let experiment = activeExperiments.first(where: { $0.slug == experimentId }) {
+        if let experiment = activeExperiments.first(where: { $0.featureIds.contains(featureId) }) {
```
jhugman, line 53: It took me a minute to persuade myself that this worked.

My reasoning: 

1. `getActiveExperiments` gets a list of experiments that are enrolled.
2. Features are only enrolled in zero or one experiment at a time.
3. The first experiment we find where the `featureIds` contains our `featureId` will therefore be the only experiment enrolled involving this experiment.

For the purposes of this PR: let's add a comment persuading the reader that this works.

Moving forward: Adding Rust FFI surface area in the same PR should be fine — previously it wasn't. We should probably keep this, for the sake of shipping, but start the habit of treating the FFI as internals, not an API itself.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -156,6 +156,26 @@ interface NimbusInterface {
...
+     * Records the `exposure` event in telemetry.
+     *
+     * This is a manual function to accomplish the same purpose as passing `true` as the
+     * `recordExposureEvent` property of the `getVariables` function. It is intended to be used
+     * when requesting feature variables must occur at a different time than the actual user's
+     * exposure to the feature within the app.
+     *
+     * In the case where the use of this function is required, then the `getVariables` function
+     * should be called with `false` so that the exposure event is not recorded when the variables
+     * are fetched.
```
jhugman, line 169: This PR doesn't seem to call this method from `getVariables`. We should probably fix that before landing.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -156,6 +156,26 @@ interface NimbusInterface {
      */
     fun resetTelemetryIdentifiers() = Unit
 
+    /**
+     * Records the `exposure` event in telemetry.
+     *
+     * This is a manual function to accomplish the same purpose as passing `true` as the
+     * `recordExposureEvent` property of the `getVariables` function. It is intended to be used
+     * when requesting feature variables must occur at a different time than the actual user's
+     * exposure to the feature within the app.
```
jhugman, line 165: e.g. experimenting on a menu, where menu construction happens at a different time to the user seeing the menu. 

Or `getVariables` is being asked for multiple times, with the same feature, and we want to record exposure only once.

e.g. when `getVariables` is being asked for every keystroke.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -510,19 +534,19 @@ open class Nimbus(
...
 
     // The exposure event should be recorded when the expected treatment (or no-treatment, such as
     // for a "control" branch) is applied or shown to the user.
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     @WorkerThread
-    internal fun recordExposureOnThisThread(experimentId: String) = withCatchAll {
+    internal fun recordExposureOnThisThread(featureId: String) = withCatchAll {
         val activeExperiments = getActiveExperiments()
-        activeExperiments.find { it.slug == experimentId }?.also { experiment ->
+        activeExperiments.find { it.featureIds.contains(featureId) }?.also { experiment ->
```
jhugman, line 549: This means:
1. `getExperimentBranch` by experiment slug is now not recording exposure events.
2. `getExperimentBranch` by feature id did not record exposure events prior to this change?

jhugman (commented):
A few comments. I think we should wait until the swift variables land first.

components/nimbus/ios/Nimbus/Nimbus.swift
travis79: I foresee where this entire function will be moved to Rust and can just be called across the FFI and recorded in Glean there (once I can get to it). So, rather than touching the Rust now and potentially making even more surface area that might need refactored when that happened, it seemed useful for me to try and keep this isolated in the foreign language bindings where the Glean integration is currently.

I agree definitely about adding a comment to clarify what exactly we are expecting from this line.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
travis79: Maybe a merge snafu, I will investigate.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
travis79: Correct. My thinking was that get branch by experiment slug was deprecated and this would help to support that. yes, we may have been missing exposure events for a couple of experiments since it was trying to match feature ids against the slug.

megazords/ios/MozillaAppServicesTests/NimbusTests.swift
```diff
@@ -293,8 +299,12 @@ class NimbusTests: XCTestCase {
...
         try nimbus.applyPendingExperimentsOnThisThread()
 
-        // Record the exposure event in Glean
-        nimbus.recordExposure(experimentId: "test-experiment")
+        // Record the valid exposure event in Glean
+        nimbus.recordExposureEvent(featureId: "test-feature")
+
+        // We will also attempt to record an invalid feature exposure to ensure it doesn't
+        // get recorded
+        nimbus.recordExposureEvent(featureId: "not-a-feature")
```
jhugman, line 307: Should this be accompanied with a negative test, called before the positive test?

jhugman (approved):
A couple of additional changes needed, but doesn't need to be held up by another review cycle. Thanks for jumping on this one!

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -184,12 +192,21 @@ extension Nimbus: NimbusFeatureConfiguration {
...
+
+        if recordExposureEvent {
+            recordExposure(featureId)
+        }
+
         return JSONVariables(with: json)
     }
+
+    public func recordExposureEvent(featureId: String) {
+        recordExposure(featureId)
```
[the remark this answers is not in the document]
travis79: I'd be happy to do that, actually.  The only thing that worries me is that it dilutes down the `GleanIntegration` piece if part of the integration will be defined in another extension. But, let me tinker with it and see if we can simplify things a bit.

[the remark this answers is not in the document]
travis79: Per our convo on Slack, I was able to satisfy this by refactoring a bit so that everything Glean still lives inside the same extension definition.

travis79:
Change requests addressed, awaiting checks to pass

PR #4123, opened 2021-05-19
Observation API for History Metadata Storage

grigoryk:
TODO: ~~db migration (drop&recreate moz_places_metadata)~~, ~~missing Swift bindings update~~

This PR gets us closer to https://github.com/mozilla/application-services/issues/4067 - removes an explicit metadata `guid`, switching to an observation API (vs a CRUD api that was in place originally).

  - Note:

### DIFF
--- a/components/places/ffi/src/lib.rs
+++ b/components/places/ffi/src/lib.rs
@@ -13,7 +13,7 @@ use ffi_support::{
     define_string_destructor, ByteBuffer, ConcurrentHandleMap, ExternError, FfiStr,
 };
 use places::error::*;
-use places::msg_types::{BookmarkNodeList, HistoryMetadata, SearchResultList};
+use places::msg_types::{BookmarkNodeList, HistoryMetadataObservation, SearchResultList};
 use places::storage::bookmarks;
 use places::types::VisitTransitionSet;
 use places::{storage, ConnectionType, PlacesApi, PlacesDb};
@@ -486,32 +486,18 @@ pub extern "C" fn places_get_latest_history_metadata_for_url(
 /// # Safety
 /// Deref pointer, thus unsafe
 #[no_mangle]
-pub unsafe extern "C" fn places_add_history_metadata(
+pub unsafe extern "C" fn places_note_history_metadata_observation(
     handle: u64,
     data: *const u8,
     len: i32,
     error: &mut ExternError,
-) -> *mut c_char {
-    log::debug!("places_add_history_metadata");
+) {
+    log::debug!("places_note_history_metadata_observation");
     CONNECTIONS.call_with_result(error, handle, |conn| -> places::Result<_> {
         let buffer = get_buffer(data, len);
-        let metadata: HistoryMetadata = prost::Message::decode(buffer)?;
-        let guid = storage::history_metadata::add_metadata(conn, metadata)?;
-        Ok(guid.into_string())
-    })
-}
-
-#[no_mangle]
-pub extern "C" fn places_update_history_metadata(
-    handle: u64,
-    guid: FfiStr<'_>,
-    total_view_time: i32,
-    error: &mut ExternError,
-) {
-    log::debug!("places_update_history_metadata");
-    CONNECTIONS.call_with_result_mut(error, handle, |conn| {
-        let guid = SyncGuid::from_string(guid.into_string());
-        storage::history_metadata::update_metadata(conn, &guid, total_view_time)
+        let observation: HistoryMetadataObservation = prost::Message::decode(buffer)?;
+        storage::history_metadata::apply_metadata_observation(conn, observation)?;
+        Ok(())
     })
 }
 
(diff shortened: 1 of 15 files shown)


components/places/ffi/src/lib.rs
```diff
@@ -486,32 +486,18 @@ pub extern "C" fn places_get_latest_history_metadata_for_url(
...
-    guid: FfiStr<'_>,
-    total_view_time: i32,
-    error: &mut ExternError,
-) {
-    log::debug!("places_update_history_metadata");
-    CONNECTIONS.call_with_result_mut(error, handle, |conn| {
-        let guid = SyncGuid::from_string(guid.into_string());
-        storage::history_metadata::update_metadata(conn, &guid, total_view_time)
+        let observation: HistoryMetadataObservation = prost::Message::decode(buffer)?;
+        storage::history_metadata::note_observation(conn, observation)?;
```
mhammond, line 499: `storage::history::` already has a `apply_observation`, so maybe `storage::history_metadata::apply_observation` or similar here too? or maybe `apply_metadata_observation` if having the same name smells wrong.

mhammond (commented):
I only had a quick look, but this is looking great and is exactly what I had in mind when talking about this - thanks! I'll dig deeper next week.

components/places/ffi/src/lib.rs
grigoryk: Sure, `apply_metadata_observation` works.

grigoryk:
@mhammond I've updated this PR a bit - generally cleaned things up, fixed up tests, but mainly introduced a split ("key" vs "observation") at the api boundary level and in the guts, as well. I've left the protobuf layer mainly as-is, since it doesn't really buy us much to introduce the type split there as well.

Splitting things into `key` and `observation` helped quite a bit conceptually at the higher levels, but also helped clean up the queries as well. Not happy with the name `key`, of course, but neither @csadilek nor myself could come up with something better. Open to suggestions, but it's also not a terrible name and works for us.

Need to update the Swift bindings before landing this, but wanted to officially tag you for a review first :)

grigoryk:
Corresponding A-C PR in https://github.com/mozilla-mobile/android-components/pull/10346

components/places/src/storage/history_metadata.rs
```diff
@@ -2,47 +2,209 @@
...
+                true,
+            )? {
+                Some(id) => SearchQueryEntry::Existing(id),
+                None => SearchQueryEntry::CreateFor(lowercase_term),
+            },
+        )
+    }
+}
+
+trait CanDebounce {
```
mhammond, line 118: nit: you don't need a trait for this, because no where do you use a generic type - ie, IIUC, you could just impl that function in each type and it would work fine. I'm fine if you think it adds some value though, but it also does add some cognitive overhead

components/places/src/storage/history_metadata.rs
```diff
@@ -402,395 +504,516 @@ mod tests {
...
-                    meta.parent_url.is_none(),
-                    "parent_url expected to be None"
+                    meta.referrer_url.is_none(),
+                    "referrer_url expected to be None"
                 ),
             };
         };
     }
 
+    #[macro_use]
```
mhammond, line 522: fancy :)

mhammond (approved):
Just a few nits, and there might be some cleanups worthwhile after we feel the "experimentation" phase is over, but there's nothing to block moving forward on this,

mhammond:
(and both CI failures seem real and related to this patch)

components/places/src/storage/history_metadata.rs
grigoryk: Cleaned this up, no more trait!

components/places/src/storage/history_metadata.rs
```diff
@@ -2,47 +2,209 @@
...
+    }
+}
+
+enum PlaceEntry {
+    Existing(i64),
+    CreateFor(Url, Option<String>),
+}
+
+impl PlaceEntry {
+    fn from(url: &str, tx: &PlacesTransaction<'_>, title: Option<String>) -> Result<Self> {
```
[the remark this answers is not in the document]
grigoryk: Switched to `fetch`.

components/places/src/storage/history_metadata.rs
```diff
@@ -2,47 +2,209 @@
...
+};
 use sql_support::{self, ConnExt};
 use sync_guid::Guid as SyncGuid;
 use types::Timestamp;
 use url::Url;
 
 use lazy_static::lazy_static;
 
+impl HistoryMetadataObservation {
+    fn from(observation_with_title: HistoryMetadataObservationWithTitle) -> (Self, Option<String>) {
```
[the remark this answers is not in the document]
grigoryk: Removed `HistoryMetadataObservationWithTitle` entirely and removed this as well.

grigoryk:
All right, just waiting for a green CI run now. Should be ready to land.

PR #4124, opened 2021-05-19
Rename LoginStore to LoginsSyncEngine.

mhammond:
The logins component currently has a PasswordStore and a LoginStore - the
latter is actually the "sync engine" and renaming this now makes the
uniffi work we are doing that little bit easier to understand.

The more we can do to keep stuff out of Sammy's PR , the better :)

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -1225,12 +1225,12 @@ impl LoginDb {
     }
 }
 
-pub struct LoginStore<'a> {
+pub struct LoginsSyncEngine<'a> {
     pub db: &'a LoginDb,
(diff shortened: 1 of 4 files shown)


components/logins/src/db.rs
```diff
@@ -1225,12 +1225,12 @@ impl LoginDb {
     }
 }
 
-pub struct LoginStore<'a> {
+pub struct LoginsSyncEngine<'a> {
```
skhamis, line 1228: Perfect name!

skhamis (approved):
Thanks for making these changes. Always nice to do things to reduce that massive PR :)

lougeniaC64 (approved):
LGTM! 🚀

PR #4110, opened 2021-05-17
Update rust and rustup config in CircleCI.

rfk:
This fixes #4108, getting our CircleCI tests passing again with
RUSTUP_TOOLCHAIN="beta". It's a bit of a circuitous fix but I
think it's worthwhile taking this approach.

As noted over in #4108, the `circleci/rust` docker image configures
rustup to use its "minimal" profile by default, which means it doesn't
automatically install `rustup` into non-default toolchains, which was
causing problems for our tests.

According to a Google search for "CircleCI Rust" and to the note at
the top of [1], the `circleci/rust` image is considered legacy and has
been replaced by `cimg/rust`. This new image uses the "default" rustup
profile, meaning it will more closely match the experience of executing
Rust tooling on your local machine, and in particular will install
`rustfmt` by default in all toolchains.

This new `cimg/rust` image does not support a `:latest` tag, and instead
requires you to specify an explicit version of the Rust toolchain to
use. To avoid duplicating this information in many places, I've used the
"executors" feature of CircleCI to consolidate the details of the docker
image in a single place.

Since I was defining some executors anyway, I also took the opportunity
to consolidate the details of the "macos" executor, so if we want to
bump the version of xcode in future it can be done in a single place.

Finally, I updated the version of `rustup-init` that is used to boostrap
the Rust tooling on the macos executor and in taskcluster. This isn't
necessary, but I did it as part of investigating the fix here and it
seemed worthwhile to keep.

[1] https://circleci.com/developer/images/image/cimg/rust

PR #4128, opened 2021-05-20
CircleCI must install Python in order to build system libs.

rfk:
The docker image change in #4110 seems to mean that `python` is no
longer available in the Docker image by default, which means that
the build scripts in `./libs` no longer run successfully on CircleCI.

We did not notice this change because CircleCI usually restores
these from a cache. But if we do something that invalidates that
cache (like modifying the files in `./libs`) then CircleCI starts
failing.

This commit fixes the issue by explicily installing `python` as
part of the dependencies for the system lib builds. It also includes
a cosmetic change to `./libs/README.md` to clear the CircleCI cache
and test that the fix actually works.

  - Note:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ commands:
   build-desktop-libs:
     steps:
       - run: sudo apt-get update
-      - run: sudo apt-get install tcl
+      - run: sudo apt-get install python tcl
       - run:
           name: Install NSS build system dependencies
           command: sudo apt-get install ninja-build gyp zlib1g-dev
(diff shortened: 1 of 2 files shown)


mhammond (approved):
I guess I'm mildly surprised it's not `python3`, but OTOH I guess I'm glad the world has moved on such that it's no longer ambiguous :)

rfk:
> I guess I'm mildly surprised it's not python3, but OTOH I guess I'm glad the world has moved on such that it's no longer ambiguous

Hahahahahaha, oh @mhammond, you optimist.

What's actually happening is that the docker image only ships with `python3`, but the NSS build scripts need Python 2 and expect to find it on the system as `python`, which is what the `python` package installs.

(Actually that may not be fair to NSS, its scripts mostly look like they try to work with either python2 or python3...but I do remember having to do a downgrade to build that thing at some point in the past).

PR #4129, opened 2021-05-20
logins: split the DB from the sync engine

mhammond:
This is a bit painful, but will be necessary for the uniffi work - in short, our `LoginDb` has been serving as both our database abstraction and our `SyncEngine` abstraction. This PR splits this up so that the `LoginsSyncEngine` does only sync-specific stuff, and the `LoginDb` is the more generic abstraction. Obviously the sync engine uses the DB - but now it can have a much shorter lifetime, and later, will actually use a `Mutex<LoginDb>`. This is the same pattern used by autofill.

I thought about doing the `Mutex` now, which in theory would work, but is much much riskier - attempting to take the mutex twice would deadlock and there's a risk we'd do that on the same thread (eg, take the mutex, then call another function which takes the mutex). There are a couple of things there that will make our future lives easier though - eg, lots of `let db = &self.store.db` - in the future that's going to change to `let db = self.store.db.lock().unwrap()` etc.

This is going to conflict with @bendk and @skhamis work. To help Ben out, I took the liberty of applying this to https://github.com/mhammond/application-services/tree/logins-per-field-encryption-with-engine-split - this is Ben's fork (which is ahead of the main fork), and has had applied f01f9bafa55f56425c3281dca2122c6fb6bee897 (which I added yesterday) and this PR. Feel free to do with it what you like. Obviously that only makes sense if this PR is accepted and merged.

I *nearly* got this fully working on the uniffi branch but ran out of time. I'll finish that and the sync manager work for uniffi asap.

### DIFF
--- a/components/logins/src/store.rs
+++ b/components/logins/src/store.rs
@@ -1,15 +1,13 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-use crate::db::{LoginDb, LoginsSyncEngine, MigrationMetrics};
+use crate::db::{LoginDb, MigrationMetrics};
 use crate::error::*;
 use crate::login::Login;
+use crate::LoginsSyncEngine;
 use std::cell::Cell;
 use std::path::Path;
-use sync15::{
-    sync_multiple, telemetry, EngineSyncAssociation, KeyBundle, MemoryCachedState,
-    Sync15StorageClientInit,
-};
+use sync15::{sync_multiple, telemetry, KeyBundle, MemoryCachedState, Sync15StorageClientInit};
 
 // This store is a bundle of state to manage the login DB and to help the
 // SyncEngine.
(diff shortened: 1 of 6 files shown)


components/logins/src/store.rs
```diff
@@ -154,7 +153,7 @@ impl PasswordStore {
     }
 
     pub fn check_valid_with_no_dupes(&self, login: &Login) -> Result<()> {
-        self.db.check_valid_with_no_dupes(login)
+        self.db.check_valid_with_no_dupes(&login)
```
bendk, line 156: Maybe I'm reading the rust wrong, but is this change needed?  It seems like one more level of referencing than required.

bendk (approved):
Looks good to me.  I fully support moving the code around to make the abstractions more logical.  It'll be annoying to rebase, but worth it IMO.  Just the fact that this divides the code db.rs into 2 modules feels like a win to me.

rfk:
@mhammond looks like this might have broken something in the Android build on taskcluster, ref [this failed run](https://github.com/mozilla/application-services/pull/4129/checks?check_run_id=2634315078).

PR #4131, opened 2021-05-21
Update uniffi dependency to v0.9.0

rfk:
There was a bit of churn here in the `fxa-client` component to make it `[Threadsafe]` and avoid the new deprecation warning, but otherwise this seems straightforward. Pinging both Sync and Nimbus folks for review from a risk-mitigation perspective, given any upcoming releases.

### DIFF
--- a/components/fxa-client/src/lib.rs
+++ b/components/fxa-client/src/lib.rs
@@ -104,7 +104,9 @@ pub enum FxaError {
 /// account and accessing other services on behalf of the user.
 ///
 pub struct FirefoxAccount {
-    internal: internal::FirefoxAccount,
+    // For now, we serialize all access on a single `Mutex` for thread safety across
+    // the FFI. We should make the locking more granular in future.
+    internal: std::sync::Mutex<internal::FirefoxAccount>,
 }
 
 /// # Constructors and state management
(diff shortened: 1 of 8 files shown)


components/fxa-client/src/lib.rs
```diff
@@ -104,7 +104,9 @@ pub enum FxaError {
 /// account and accessing other services on behalf of the user.
 ///
 pub struct FirefoxAccount {
-    internal: internal::FirefoxAccount,
+    // For now, we serialize all access on a single `Mutex` for thread safety across
+    // the FFI. We should make the locking more granular in future.
+    internal: std::sync::Mutex<internal::FirefoxAccount>,
```
rfk, line 109: Right now I am very happy that I went with this `internal` abstraction that could be easily wrapped in a single Mutex 🎉

rfk: (For context: the previous version of UniFFI was adding a Mutex like this automagically, but that behaviour is deprecated and will be removed in a future version)

rfk:
This is currently failing CI on an Android test issue, but I think that may have been introduced by https://github.com/mozilla/application-services/pull/4129.

mhammond (approved):
🚀 🚢

rfk:
Since we just cut a new release for Nimbus, I'm going to go ahead and merge this once CI is green.

PR #4133, opened 2021-05-21
SDK-265 SDK-266 Variables with resources

jhugman:
[1]: https://jira.mozilla.com/browse/SDK-265
[2]: https://jira.mozilla.com/browse/SDK-266
[3]: https://jira.mozilla.com/browse/SDK-285

This PR adds [localized strings][1] and [images][2] provided by the app, but usable from nimbus.

Unfortunately, these can't be tested in place (they need an app with resources), so testing is done as part of making [our own validation experiment][3].

It also adds some cleanup, and fixes some paper cuts around developing our own experiment.

I'll add a changelog after self-review.

  - Note:

### DIFF
--- a/components/nimbus/ios/Nimbus/NimbusApi.swift
+++ b/components/nimbus/ios/Nimbus/NimbusApi.swift
@@ -36,12 +36,20 @@ public protocol NimbusFeatureConfiguration {
     /// See `recordExposureEvent` for more information on manually recording the event.
     ///
     /// - Returns a `Variables` object used to configure the feature.
-    func getVariables(featureId: String, recordExposureEvent: Bool) -> Variables
+    func getVariables(featureId: String, sendExposureEvent: Bool) -> Variables
 }
 
 public extension NimbusFeatureConfiguration {
+    /// Get the variables needed to configure the feature given by `featureId`.
+    ///
+    /// By default this sends an exposure event.
+    ///
+    /// - Parameters:
+    ///     - featureId The string feature id that identifies to the feature under experiment.
+    ///
+    /// - Returns a `Variables` object used to configure the feature.
     func getVariables(featureId: String) -> Variables {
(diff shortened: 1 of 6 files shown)


components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -39,6 +39,12 @@ public protocol NimbusFeatureConfiguration {
     func getVariables(featureId: String, recordExposureEvent: Bool) -> Variables
 }
 
+public extension NimbusFeatureConfiguration {
+    func getVariables(featureId: String) -> Variables {
+        return getVariables(featureId: featureId, recordExposureEvent: true)
+    }
+}
```
jhugman, line 46: Added this extension method because protocols don't allow default arguments.

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -22,7 +22,7 @@ public protocol NimbusFeatureConfiguration {
     /// - Parameter featureId The string feature id that applies to the feature under experiment.
     /// - Returns A String representing the branch-id or "slug"; or `nil` if not enrolled in this experiment.
     ///
-    func getExperimentBranch(featureId: String) -> String?
+    func getExperimentBranch(experimentId: String) -> String?
```
jhugman, line 25: Renamed now that we're moving away from branch names.

This method is still used, but only for tooling.

components/nimbus/ios/Nimbus/NimbusApi.swift
travis79: The implementation of this in Nimbus.swift also has a default value, I don't think we need it in both places if you are going to define this as an extension here (nifty solution, by the way).

travis79 (approved):
One little nit, but other than that, this is awesome!

Issue #4090, opened 2021-05-12
Make sync-pass example work again


data-sync-user:
I don't think the sync-pass example is going to work after the encryption changes.  I think we could fix it without too much work.  I think the main issue is how to input the encryption key.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/SYNC-2291)
┆Epic: Logins update and sqlcipher migration
┆Fix Versions: Release 92

mhammond:
> I think the main issue is how to input the encryption key.

probably the same way as the autofill example - we just store it as a meta value in the DB? Obviously this isn't secure, but it doesn't matter in this case.

Issue #4065, opened 2021-05-04
Add encryption to logins sync logic


lougeniaC64:
* Implement [EncryptorDecryptor](https://github.com/mozilla/application-services/blob/main/components/autofill/src/encryption.rs#L40) logins sync logic so the encrypted fields in the new per-field encrypted database can be read
  * ~Keep in mind that we need to support both encryption approaches during the migration period~
* Update the sync manager logic to register the logins store
* Add tests and make sure existing tests still pass

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2632)
┆Epic: Logins update and sqlcipher migration
┆Fix Versions: Release 92
┆Sprint End Date: 2021-07-23

data-sync-user:
➤ Janet Dragojevic commented:

This is done.

PR #4135, opened 2021-05-21
Logins per field encryption

bendk:

  - Note:


bendk:
Finally have my branch in decent shape.  `sync-pass` seems to be working for me and I rebased this to include the `SyncEngine` refactoring.  There's definitely more work left to do, but I tried to leave TODO breadcrumbs for that.

components/logins/src/login.rs
```diff
@@ -299,6 +322,45 @@ fn string_or_default(row: &Row<'_>, col: &str) -> Result<String> {
 }
 
 impl Login {
```
bendk, line 324: This code feels kind of yucky to me with all the `clone()` calls, but I wasn't sure how to improve it.  Other options I thought about, but didn't go with:

  - Manually implement deserialize (seems like a lot of work)
  - Skip `from_record()` and `into_record()` and get things from the `data` field directly
 
Is there a better way?

skhamis: If I'm understanding right, another possibility is doing a From/Into: https://github.com/mozilla/application-services/blob/main/components/logins/src/login.rs#L549 for your impl and this will actually prevent needing to call `self.guid, etc` (and consequently prevent needing to call clone)

bendk: We can't use From, since the method also inputs the `EncryptorDecryptor`.  But thanks for the suggestion because it made me realized that Into() consumes the caller, which makes it possible to avoid the cloning.

components/logins/src/store.rs
```diff
@@ -126,11 +141,16 @@ impl PasswordStore {
...
     ) -> Result<telemetry::SyncTelemetryPing> {
-        let engine = LoginsSyncEngine::new(&self);
-
+        let mut engine = LoginsSyncEngine::new(&self);
         let mut disk_cached_state = engine.get_global_state()?;
         let mut mem_cached_state = self.mem_cached_state.take();
+        // TODO-sqlcipher: unwrap() is not right here, but the above comment makes me think we
+        // should not worry about this too much yet.
+        engine
+            .set_local_encryption_key(local_encryption_key)
```
lougeniaC64, line 152: @mhammond Just out of curiosity, what prevents us from removing this `sync` function and relying on sync manager for logins. If it's not too much of a lift, it might be worth removing this function so this `set_local_encryption_key` function isn't necessary.

components/logins/src/db.rs
```diff
@@ -2,15 +2,36 @@
...
 
+/// Logins DB handling
+///
+/// The logins database works differently than other components because "mirror" and "local" mean
+/// different things.  At some point we should probably refactor to make it match them, but here's
+/// how it works for now:
+///
+///   - loginsM is the mirror table, which means it stores what we believe is on the server.  This
+///     means either the last record we fetched from the server or the last record we uploaded.
+///   - loginsL is the local table, which means it stores local changes have not been sent to the
```
lougeniaC64, line 13: nit: local changes _that_ have not

components/logins/src/db.rs
```diff
@@ -88,71 +91,73 @@ impl LoginDb {
...
+                    }
+                    if path.exists() {
+                        std::fs::remove_file(&path)?;
+                    }
+                    MigrationMetrics { ..Default::default()}
+                }
+                Ok(metrics) => {
+                    log::info!("Deleting old sqlcipher DB after migration");
+                    if sqlcipher_path.exists() {
+                        std::fs::remove_file(sqlcipher_path)?;
```
lougeniaC64, line 146: I'm wondering if a failed `remove_file` call should be collected in the `MigrationMetrics`. I _think_ it makes sense to do that so all migration-related errors can be reported in telemetry, but I'm unsure so I'm leaving this comment to spur discussion.

components/logins/src/db.rs
```diff
@@ -88,71 +91,73 @@ impl LoginDb {
...
-        Self::with_connection(Connection::open(path)?, encryption_key, None)
+    pub fn open(path: impl AsRef<Path>) -> Result<Self> {
+        Self::with_connection(Connection::open(path)?)
     }
 
-    pub fn open_with_salt(
+    // Open a dabase, after potentially migrating from a sqlcipher database.  This method handles
+    // the migration process:
+    //
+    //    - If there's not a file at sqlcipher_path, then we skip the migratnion
```
lougeniaC64, line 101: nit: migration

lougeniaC64:
So far so good! 🎉 As expected, some tests are broken, some functionality isn't working as expected, and there will be additional changes but . . . this serves as a good foundation for the remainder of the work. I've broken out my feedback by the tickets on the board below.

[Collect success/failure metrics work](#4102): I think this can be considered done. There will need to additional work to actually report the metrics to glean, but that's outside of the scope of this issue. I have an outstanding question about whether errors stemming from failing to delete the database file should be included in the migration metrics but I don't have strong opinions on that either way.

[Storage per-field encryption and migration](#4064): Since this issue represents the bulk of the storage changes and they are blocked by the api design, this is as done as it can be. Also want us to do as much as we can to exercise the migration code.

[Add encryption to logins sync](#4065): I have an open question around whether we should keep the [sync function](https://github.com/mozilla/application-services/blob/fac0e27c78f37286f695369e41f29003f127a727/components/logins/src/store.rs#L125) or do the work to use sync manager instead. That may be too heavy of a lift but it would prevent us from having another function that requires the per-field encryption key.

[Sync-pass update](#4090): I'd be okay with calling this one done but we're probably going to have to update it as the uniffi changes come in and the api is updated.

[Handle logins sqlcipher migration failure](https://jira.mozilla.com/browse/SYNC-2203): I think we can consider this one done as well.

@mhammond I'd be interested in your feedback on the migration code and the open questions I added.

components/logins/src/db.rs
```diff
@@ -2,15 +2,36 @@
...
+///   - When we want to fetch a record, we need to look in both loginsL and loginsM for the data.
+///     If a record is in both tables, then we prefer the loginsL data.  GET_BY_GUID_SQL contains a
+///     clever UNION query to accomplish this.
+///   - If a record is in both the local and mirror tables, we call the local record the "overlay"
+///     and set the is_overridden flag on the mirror record.
+///   - When we sync, the presence of a record in loginsL means that there was a local change that
+///     we need to send to the the server and/or reconcile it with incoming changes from the
+///     server.
+///   - After we sync, we move all records from loginsL to loginsM, overwriting any previous data.
+///     loginsL will be an empty table after this.  See mark_as_synchronized() for the details.
```
mhammond, line 24: thanks heaps for writing this up!

components/logins/src/db.rs
```diff
@@ -24,22 +45,22 @@ use url::{Host, Url};
 
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Default)]
 pub struct MigrationPhaseMetrics {
-    num_processed: u64,
-    num_succeeded: u64,
-    num_failed: u64,
-    total_duration: u128,
-    errors: Vec<String>,
+    pub num_processed: u64,
```
mhammond, line 48: It would be good if these could be `pub(crate)` instead - that's what the uniffi branch has ended up with, so we might as well avoid future conflicts early.

...

Actually, that branch no longer needs them to be public in any way! I asked Sammy to revert that change in his branch.

I also see that sync-pass uses them. TBH, I'd be inclined to have sync-pass just `{:?}` the entire struct - it's really just for us. But I'll leave that up to you!

components/logins/src/db.rs
```diff
@@ -88,71 +91,73 @@ impl LoginDb {
...
     }
 
-    pub fn open(path: impl AsRef<Path>, encryption_key: Option<&str>) -> Result<Self> {
-        Self::with_connection(Connection::open(path)?, encryption_key, None)
+    pub fn open(path: impl AsRef<Path>) -> Result<Self> {
+        Self::with_connection(Connection::open(path)?)
     }
 
-    pub fn open_with_salt(
+    // Open a dabase, after potentially migrating from a sqlcipher database.  This method handles
```
mhammond, line 98: typo: "dabase"

components/logins/src/db.rs
```diff
@@ -88,71 +91,73 @@ impl LoginDb {
...
+    // the migration process:
+    //
+    //    - If there's not a file at sqlcipher_path, then we skip the migratnion
+    //    - If there is a file, then we attempt the migration and delete the file afterwards.
+    //
+    //  The salt arg is for IOS and other systems where the salt is stored externally.
```
mhammond, line 104: Not sure "and other systems" is accurate :) (Not that it matters though :)

components/logins/src/db.rs
mhammond: yeah, failure to remove the files is an edge case, and probably something we just want to make noise about, but not fail? I haven't thought this through though

components/logins/src/db.rs
```diff
@@ -88,71 +91,73 @@ impl LoginDb {
...
+                    log::info!("Deleting old sqlcipher DB after migration");
+                    if sqlcipher_path.exists() {
+                        std::fs::remove_file(sqlcipher_path)?;
+                    }
+                    metrics
+                }
+            }
+        } else {
+            log::debug!("SQLCipher DB not found, skipping migration");
+            MigrationMetrics { ..Default::default()}
```
mhammond, line 153: I think `cargo fmt` is going to have something to say about this :)

components/logins/src/schema.rs
```diff
@@ -88,14 +88,19 @@
...
 use lazy_static::lazy_static;
-use rusqlite::Connection;
+use rusqlite::{Connection, NO_PARAMS};
 use sql_support::ConnExt;
+use std::time::Instant;
 
 /// Note that firefox-ios is currently on version 3. Version 4 is this version,
 /// which adds a metadata table and changes timestamps to be in milliseconds
-pub const VERSION: i64 = 4;
+pub const VERSION: i64 = 5;
```
mhammond, line 101: I don't quite get why this isn't being reset? The comment above probably also needs updating (and I hope iOS isn't on version 3 still!?)

mhammond (commented):
This is looking great! I'm slightly confused by some of the migration code, but only spent a little time looking at at and I'm out of time, but I'll get back to it next week!

components/logins/src/db.rs
bendk: I think failure to remove the file means opening the DB should fail.  Otherwise, the next time the app starts up, we're going to try to migrate the old DB file again.  Otherwise, we'd need a way to distinguish between a successful migration, but a failure to remove the file, and a migration that started, but failed before the end.  I think there is a way to do that, but is it worth it?  I can't picture a situation where you could successfully migrate, but not delete the old file.

In terms of metrics, I think it should be reported, but I'm not sure how to do it with `MigrationMetrics`.  Is there any other telemetry related to start up failures?  I'm wondering if it could be grouped with things like failing to create a new database file.

components/logins/src/db.rs
bendk: Sounds good.  I'll update sync-pass to use `{:?}` and add a note to update the field visibility once the uniffi branch gets merged in.

components/logins/src/db.rs
bendk: It complained, but maybe not enough.  I switched this line to `MigrationMetrics::default()`, along with a bunch of other `default()` calls.

components/logins/src/schema.rs
bendk: To me, it feels more natural to think of the schema version as separate whether the database was using sqlcipher encryption.  So "version 4" specifies a particular schema, rather than also needing to say if it was plaintext or sqlcipher DB.  But I'm happy to change this if others like the resetting the numbers better.

mhammond: I don't feel too strongly either, but having it at "version 4" implies it's possible we will find a "version 3" out there. A comment was also added to explain why there are no migrations for numbers lower than this, because otherwise it would seem wrong.  IOW, I left this more as a note-to-self to say "hrm - maybe I don't actually understand the migration and what's going on here, I better come back and look more closely" - which is exactly the red-flag I'm thinking it might be worth avoiding

bendk: I think I'm missing what the red-flag is, but resetting the schema sounds good to me.  I just pushed some code that does that.  in fact, I realized that the `sqlcipher_export` function doesn't export PRAGMA values, so we should be manually setting a version number anyways.  Are there any other PRAGMA values that we should be copying over?

PR #4138, opened 2021-05-24
Add formatting guidance to `CHANGES_UNRELEASED.MD`

lougeniaC64:
Our [prepare release](https://github.com/mozilla/application-services/blob/main/automation/prepare-release.py) script expects that the `Full Changelog` line will be at the top of the `CHANGES_UNRELEASED` file. Lately we've been adding new entries above that line, which moves the changelog line to the bottom of the file and prevents the script from copying unreleased change entries to `CHANGELOG.md`. 

I've added a comment to the unreleased changes file that makes note of this. On a somewhat unrelated note, I've also provided some guidance on entry structure to make it easier for whoever's cutting the next release to gauge the correct version number. 

My guess is that the placement of the comment, let alone the verbiage, will provide a visual cue that additions should be entered below, but this may be too light-handed. Either way, I'm open to alternatives.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -3,3 +3,18 @@
 # Unreleased Changes
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v77.0.1...main)
+
+<!-- WARNING: New entries should be added below this comment to ensure the `./automation/prepare-release.py` script works as expected.
+
+Use the template below to make assigning a version number during the release cutting process easier.
+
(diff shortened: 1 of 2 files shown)


rfk (approved):
Great, thank you!

We could consider having the prepare-release script be a bit more heavy-handed, such as freaking out if it finds the any text preceding the expected "Full Changelog" blah-blah. But let's try this lighter touch and see how it goes, I expect it will make a big difference 👍🏻

PR #4142, opened 2021-05-25
Refactor and document autofill store

bendk:
This PR comes out of me trying to understand the autofill syncing / DB code better, especially ownership/locking of the Store and StoreImpl.  Along the way, it seemed like there were some parts could be simpler so I tried coding them up.  It seems to work: the tests pass and autofill-utils runs.  That said, it's very possible some of these changes are a bad idea and feel free to tell me that.  It'll be a good learning experience at least.

  - Note:

### DIFF
--- a/components/autofill/src/db/store.rs
+++ b/components/autofill/src/db/store.rs
@@ -2,10 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::db::models::address::{Address, UpdatableAddressFields};
-use crate::db::models::credit_card::{CreditCard, UpdatableCreditCardFields};
+use crate::db::models::address::{Address, InternalAddress, UpdatableAddressFields};
+use crate::db::models::credit_card::{CreditCard, InternalCreditCard, UpdatableCreditCardFields};
 use crate::db::{addresses, credit_cards, AutofillDb};
 use crate::error::*;
+use crate::sync::engine::ConfigSyncEngine;
 use rusqlite::{
     types::{FromSql, ToSql},
     Connection,
@@ -13,7 +14,6 @@ use rusqlite::{
 use sql_support::{self, ConnExt};
 use std::path::Path;
 use std::sync::{Arc, Mutex};
-use sync15_traits::SyncEngine;
 use sync_guid::Guid;
 
 // Our "sync manager" will use whatever is stashed here.
@@ -102,7 +102,8 @@ impl Store {
         credit_cards::scrub_encrypted_credit_card_data(&self.db.lock().unwrap().writer)?;
         // Force the sync engine to refetch data (only need to do this for the credit cards, since the
         // addresses engine doesn't store encrypted data).
-        crate::sync::credit_card::create_engine(self.clone()).reset_local_sync_data()?;
+        self.create_credit_cards_sync_engine()
+            .reset_local_sync_data()?;
         Ok(())
     }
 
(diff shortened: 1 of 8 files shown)


mhammond (dismissed):
Nice - this makes perfect sense, thanks!

components/autofill/src/db/store.rs
```diff
@@ -2,178 +2,60 @@
...
-// `StoreImpl` could be re-unified)
-// Sadly, this is `pub` because our `autofill-utils` example uses it.
+// Get a Store for the SyncManager to use
+pub fn get_store_for_manager() -> Option<Store> {
+    STORE_FOR_MANAGER.lock().unwrap().as_ref().cloned()
+}
+
+// This is the type that uniffi exposes. It has `Arc<>` around the db mutex because
+// register_with_sync_manager() needs to clone ourself.  This is redundant though, because uniffi
+// also has an Arc<>.  One day https://github.com/mozilla/uniffi-rs/issues/419 will give us access
```
rfk, line 33: This piece of the UniFFI `Arc` work is ready on `main` as of yesterday, I think I'll go ahead and get it into a release so we can try it out.

components/autofill/src/db/store.rs
```diff
@@ -2,178 +2,60 @@
...
     // Mutex: just taken long enough to update the inner stuff - needed
-    //        to wrap the RefCell as they aren't `Sync`
-    // RefCell: So we can replace what it holds. Normally you'd use `get_ref()`
-    //          on the mutex and avoid the RefCell entirely, but that requires
-    //          the mutex to be declared as `mut` which is apparently
-    //          impossible in a `lazy_static`
-    // [Arc/Weak]<StoreImpl>: What the sync manager actually needs.
-    pub static ref STORE_FOR_MANAGER: Mutex<RefCell<Weak<StoreImpl>>> = Mutex::new(RefCell::new(Weak::new()));
+    //        to wrap the Option<Store> as they aren't `Sync`
+    static ref STORE_FOR_MANAGER: Mutex<Option<Store>> = Mutex::new(None);
```
mhammond, line 23: Actually, this change bothers me a little - it's no longer a weak reference. IIUC, once `register_with_sync_manager()` is called the instance will become immortal?

mhammond:
Let's hold off from merging this until that uniffi feature becomes available - we might find that simplifies almost as much as this but does allow for a more natural expression of the weak reference - we can then compare how the 2 approaches play out.

components/autofill/src/db/store.rs
rfk: One interesting thing I recently learned about `Weak` references in Rust is that, while they don't prevent the `Drop` impl from being called, they *do* prevent freeing of the underlying memory allocation (because the weakref count itself is stored in the allocation). So the allocation here will be immortal regardless, but if we have interesting `Drop` behaviours that we'd expect to be run, then this change would prevent those from running.

mhammond: huh - that *is* surprising. However, I'm thinking less of the allocation itself but the chain of structs. Eg, the IIUC, `AutofillDb` instance would be immortal (with an open connection to the db file) as I didn't see anything setting the `Option<Store>` to `None`, so an `Arc<>` to the DB remains.

components/autofill/src/db/store.rs
rfk: If you update to UniFFI `v0.10.0` (like [this](https://github.com/mozilla/application-services/pull/4131) then you should be able to use the new `[Self=ByArc]` annotation to directly obtain the underlying `Arc`.

components/autofill/src/db/store.rs
bendk: Yeah, this was the one thing I wasn't sure about.  It's not going to be dropped, unless you register a different store.  I guess I was thinking that if a consumer did create a store, call `register_with_sync_manager()`, and dropped their reference, maybe it makes sense for it to stay registered.

Still, it's definitely a behavior change and could be risky, so I can just change it back to a Weak.

mhammond:
We are going to take #4145 instead (which has taken a lot from this - thanks Ben!)

PR #4144, opened 2021-05-26
Update uniffi to version 0.10

mhammond:
Yet again, WCGW? :)

Some `Cargo.lock` changes surprised me, but they seem to be coming from an updated `cargo_metadata` in Uniffi, so 🤷

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,12 +309,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b700ce4376041dcd0a327fd0097c41095743c4c8af8887265942faf1100bd040"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


rfk (approved):
Definitely nothing will go wrong 👍🏻

mhammond:
doh - I should have known the deps would fail

mhammond:
I'm mildly surprised I had to update the license script in 6599cf7aee0794ac0a2c21024a39446b24e67543 - @rfk does that seem correct or should I dig deeper?

rfk:
@mhammond the ones that were removed feel like they're Mac-specific dependencies, grr...

mhammond:
>  feel like they're Mac-specific dependencies, grr...

yep :)

PR #4145, opened 2021-05-26
Use uniffi's [Self=ByArc] and kill the Store/StoreImpl split

mhammond:
Here's a partial alternative to @bendk's #4142 which uses the new `[Self=ByArc]` annotation in uniffi. I say "partial" as it doesn't include some of the other cleanups and doc changes, which I think are great, but I ran out of time and wanted this to focus mainly on the Arc stuff. I also stole a test I added in the logins-uniffi work (and bad me for not adding it here before) - it also checks the weak reference implementation.

It's still not *quite* as clean in terms of still having `STORE_FOR_MANAGER: Mutex<RefCell<Weak<Store>>>` but I think the ability to have weak references is important. I'd be totally fine if you'd like to take this over or integrate it into your other PR, or happy to discuss this more.

### DIFF
--- a/components/autofill/src/db/store.rs
+++ b/components/autofill/src/db/store.rs
@@ -25,136 +25,16 @@ lazy_static::lazy_static! {
     //          on the mutex and avoid the RefCell entirely, but that requires
     //          the mutex to be declared as `mut` which is apparently
     //          impossible in a `lazy_static`
-    // [Arc/Weak]<StoreImpl>: What the sync manager actually needs.
-    pub static ref STORE_FOR_MANAGER: Mutex<RefCell<Weak<StoreImpl>>> = Mutex::new(RefCell::new(Weak::new()));
+    // [Arc/Weak]<Store>: What the sync manager actually needs.
+    pub static ref STORE_FOR_MANAGER: Mutex<RefCell<Weak<Store>>> = Mutex::new(RefCell::new(Weak::new()));
 }
 
-// This is the type that uniffi exposes. It holds an `Arc<>` around the
-// actual implementation, because we need to hand a clone of this `Arc<>` to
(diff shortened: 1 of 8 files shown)


mhammond:
> It's still not quite as clean in terms of still having STORE_FOR_MANAGER: Mutex<RefCell<Weak<Store>>>

I clearly didn't try hard enough :) The `RefCell` clearly isn't needed (thanks @bendk!)

components/autofill/src/db/store.rs
```diff
@@ -248,6 +117,44 @@ impl StoreImpl {
...
         addresses::touch(&self.db.lock().unwrap().writer, &Guid::new(&guid))
     }
+
+    pub fn scrub_encrypted_data(self: Arc<Self>) -> Result<()> {
+        // scrub the data on disk
+        // Currently only credit cards have encrypted data
+        credit_cards::scrub_encrypted_credit_card_data(&self.db.lock().unwrap().writer)?;
+        // Force the sync engine to refetch data (only need to do this for the credit cards, since the
+        // addresses engine doesn't store encrypted data).
+        //
```
bendk, line 127: No need for this comment anymore, the code is looking nice and clean now.

bendk (approved):
Looks great.  I think using `Mutex<Weak<>>` to hold the Store is the way to go.

I think that this has all the important changes from my branch.  Maybe some of the comments would be nice to port over, but nothing important.

mhammond:
Thanks Ben - I took more inspiration from your PR and moved the engine-fetching back into this crate and also grabbed more of your comments. So I think this is ready to roll.

mhammond:
(oh, and what the hell, I'll take your r+ even with the latest changes!)

PR #4154, opened 2021-05-31
Specify the incoming GUID when handling a local dupe, avoiding assert…

mhammond:
…ion errors syncing with dupes

### DIFF
--- a/components/autofill/src/sync/address/incoming.rs
+++ b/components/autofill/src/sync/address/incoming.rs
@@ -150,7 +150,7 @@ impl ProcessIncomingRecordImpl for IncomingAddressesImpl {
         &self,
         tx: &Transaction<'_>,
         incoming: &Self::Record,
-    ) -> Result<Option<(SyncGuid, Self::Record)>> {
+    ) -> Result<Option<Self::Record>> {
         let sql = format!("
             SELECT
                 {common_cols},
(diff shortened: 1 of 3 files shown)


components/autofill/src/sync/address/incoming.rs
```diff
@@ -199,7 +199,7 @@ impl ProcessIncomingRecordImpl for IncomingAddressesImpl {
         });
 
         match result {
-            Ok(r) => Ok(Some((incoming.guid.clone(), r))),
```
mhammond, line 202: This is actually where the bug was - it should be the local record ID - but we are returning that local record, so I thought I'd simplify and just drop the GUID entirely as it made the error less obvious.

components/autofill/src/sync/mod.rs
```diff
@@ -342,7 +342,7 @@ fn plan_incoming<T: std::fmt::Debug + SyncRecord>(
                                 mirror.as_ref().map(|m| m.metadata()),
                             );
                             IncomingAction::UpdateLocalGuid {
-                                old_guid,
```
mhammond, line 345: so in effect here, we were passing `old_guid: incoming_record.id().clone(), record: incoming_record` - and it would have been obvious why the assertion was failing. Hence my decision to remove that from the enum.

components/autofill/src/sync/credit_card/incoming.rs
```diff
@@ -194,7 +194,7 @@ impl ProcessIncomingRecordImpl for IncomingCreditCardsImpl {
         let incoming_cc_number = self.encdec.decrypt(&incoming.cc_number_enc)?;
         for record in records {
             if self.encdec.decrypt(&record.cc_number_enc)? == incoming_cc_number {
-                return Ok(Some((incoming.guid.clone(), record)));
```
mhammond, line 197: and yeah, copy-pasta

rfk (approved):
Makes sense, and I agree your cleanup makes things more obvious to follow 👍🏻

PR #4161, opened 2021-05-31
fix nightly clippy nits

mhammond:
So looking into other issues with `+nightly`, I saw one or 2 minor warnings, then I ended up falling down a rabbit hole, then I ended up with a few bruises, plus this.

Clippy has become even more opinionated, but overall I don't hate it, so thought I'd throw it out there. Could even be split? Or we can ignore it for now - much of it was done via `--fix`

### DIFF
--- a/components/support/guid/src/lib.rs
+++ b/components/support/guid/src/lib.rs
@@ -418,6 +418,7 @@ mod test {
         assert!(!Guid::from("aaaabbbbccc=").is_valid_for_places()); // right length, bad character
     }
 
+    #[allow(clippy::cmp_owned)] // See clippy note below.
     #[test]
     fn test_comparison() {
         assert_eq!(Guid::from("abcdabcdabcd"), "abcdabcdabcd");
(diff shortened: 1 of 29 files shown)


components/support/guid/src/lib.rs
```diff
@@ -442,9 +442,12 @@ mod test {
         );
 
         // order by data instead of length
-        assert!(Guid::from("zzz") > Guid::from("aaaaaa"));
-        assert!(Guid::from("ThisIsASolowGuid") < Guid::from("zzz"));
-        assert!(Guid::from("ThisIsASolowGuid") > Guid::from("AnotherSlowGuid"));
+        // hrmph - clippy in 1.54-nightly wanted the original, below:
+        // > assert!(Guid::from("zzz") > Guid::from("aaaaaa"));
+        // changed to below: what magic is this? :/
+        assert!(*"zzz" > *"aaaaaa");
```
mhammond, line 448: this freaks me out though!

rfk: This PR looks like a really nice collection of cleanups and simplifications...and then WHAT THE WHAT SORCERY IS THIS I DONT EVEN UNDERSTAND WHAT THIS IS SUPPOSED TO BE DOING. This can't possibly be right can it?

rfk: I think clippy has mis-fired here. If I change the `impl Ord for Guid` to switch the order of comparison, then the old-style tests correctly started failing, while the new-style tests continue to pass. We should revert this IMHO.

mhammond: yeah, agreed. Even if it was correct I was going to suggest we just allow that lint

rfk: The alleged failing lint here is [`cmp_owned`](https://rust-lang.github.io/rust-clippy/master/#cmp_owned) which seems to have [a little bit of a history of false positives](https://github.com/rust-lang/rust-clippy/search?q=cmp_owned&type=issues).

rfk (approved):
👍🏻

PR #4164, opened 2021-06-01
explicitly rm six.py to work around unexplained iOS build issues

mhammond:


rfk (approved):
Thanks for investigating this

PR #4165, opened 2021-06-01
Update rust to comply with our new rust version policy

mhammond:
This also means updating to the new ctest2 package. It has the side effect of helping our coverage.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,18 @@ version: 2.1
 # ⚠️ If you add, rename or delete a job here, please also update .mergify.yml! ⚠️
 
 commands:
+  # Our policy for updating rust versions is at https://github.com/mozilla/application-services/tree/main/docs/rust-versions.md
+  # See also rust-toolchain in the root of this repo, which is used to specify our official target version.
+  setup-rust-target-version:
+    steps:
+      # So long as this is executed after the checkout it will use the version specified in rust-toolchain.yaml
+      - run: rustup update
+  # Our minimum supported rust version is specified here.
+  setup-rust-min-version:
+    steps:
+      # Technically should be on 1.47 here, but that boat sailed - 1.50 has been our effective min for some time now.
+      - run: rustup override set 1.50.0
+
   full-checkout:
     steps:
       - checkout
@@ -69,13 +81,6 @@ commands:
             ./rustup-init -y --no-modify-path
             rm rustup-init
             echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $BASH_ENV
-  setup-rust-toolchain:
-    steps:
-      - run:
-          name: Set up Rust toolchain
-          command: |
-            rustup update
-            rustc --version
   build-libs:
     parameters:
       platform:
@@ -138,7 +143,7 @@ commands:
     steps:
       - full-checkout
       - build-desktop-libs
-      - setup-rust-toolchain
+      - setup-rust-target-version
       - setup-sccache
       # clipboard crate depends on some x11 libs.
       - run: sudo apt-get install libxcb-shape0-dev libxcb-xfixes0-dev
@@ -150,15 +155,11 @@ commands:
       - run:
           name: Verify the build environment
           command: ./libs/verify-desktop-environment.sh
-  rust-tests:
-    parameters:
-      rust-version:
-        type: string
-        default: ""
+  run-tests:
     steps:
-      - test-setup
-      - run: RUSTUP_TOOLCHAIN="<<parameters.rust-version>>" automation/tests.py rust-tests
-      - run: RUSTUP_TOOLCHAIN="<<parameters.rust-version>>" automation/tests.py nss-bindings
+      - run: automation/tests.py rust-tests
+      - run: automation/tests.py nss-bindings
+
   dependency-checks:
     steps:
       - run:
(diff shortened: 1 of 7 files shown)


mhammond:
On one hand:

> Merging #4165 (918b701) into main (5465f0f) will increase coverage by 16.98%.

Yay! On the other:

> error: aborting due to 454 previous errors

something something rust versions PK11_CreateContextBySymKey something something

components/support/rc_crypto/nss/systest/build.rs
```diff
@@ -58,6 +58,7 @@ fn main() {
     cfg.skip_struct(|s| {
         s == "SECKEYPublicKeyStr_u" // inline union
     });
+    cfg.skip_fn(|s| s == "PK11_CreateContextBySymKey");
```
rfk, line 61: If I comment out this line, the tests still seem to pass for me locally - what errors were you seeing that led to this being skipped?

rfk: Heh - want I actually want to do is try *breaking* something in the `nss_sys` crate and understand how this test helps me detect that, so I was poking around in here to see what I needed to do to start.

mhammond:
locally:
```
  running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "/mnt/c/src/moz/application-services/libs/desktop/linux-x86-64/nss/include/nss" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-unknown-warning-option" "-Wno-deprecated-declarations" "-o" "/mnt/c/src/moz/application-services/target/debug/build/systest-a56f6eea10a19998/out/all.o" "-c" "/mnt/c/src/moz/application-services/target/debug/build/systest-a56f6eea10a19998/out/all.c"
  cargo:warning=/mnt/c/src/moz/application-services/target/debug/build/systest-a56f6eea10a19998/out/all.c: In function ‘__test_fn_PK11_CreateContextBySymKey’:
  cargo:warning=/mnt/c/src/moz/application-services/target/debug/build/systest-a56f6eea10a19998/out/all.c:2008:24: error: return from incompatible pointer type [-Werror=incompatible-pointer-types]
  cargo:warning=                 return PK11_CreateContextBySymKey;
  cargo:warning=                        ^~~~~~~~~~~~~~~~~~~~~~~~~~
  cargo:warning=/mnt/c/src/moz/application-services/target/debug/build/systest-a56f6eea10a19998/out/all.c: At top level:
  cargo:warning=cc1: error: unrecognized command line option ‘-Wno-unknown-warning-option’ [-Werror]
  cargo:warning=cc1: error: unrecognized command line option ‘-Wno-address-of-packed-member’ [-Werror]
  cargo:warning=cc1: all warnings being treated as errors
  exit code: 1

  --- stderr
  error: expected identifier, found keyword `crate`
   --> ../nss_sys/src/bindings/keyhi.rs:5:5
    |
  5 | use crate::*;
    |     ^^^^^

```

(and `use crate` error repeated about a billion times :)

rfk:
Gotcha thanks. I can indeed reproduce this failure on WSL, although distressingly it works fine on my Mac :-/

mhammond:
yaks down rabbit holes. [Here](https://github.com/mozilla/application-services/blob/main/Cargo.toml#L71) suggests we aren't running `nss_sys` tests by default, but [just a few lines later](https://github.com/mozilla/application-services/blob/main/Cargo.toml#L99) it's explicitly listed. I touch that comment in my rustup patch, so I think I should just nuke it

mhammond:
Not needed on CI either. I think I might keep the change with a suitable comment that keeping WSL users happy is OK and nothing of value is lost :)

.circleci/config.yml
```diff
@@ -3,6 +3,18 @@ version: 2.1
...
+  # Our policy for updating rust versions is at https://github.com/mozilla/application-services/tree/main/docs/rust-versions.md
+  # See also rust-toolchain in the root of this repo, which is used to specify our official target version.
+  setup-rust-target-version:
+    steps:
+      # So long as this is executed after the checkout it will use the version specified in rust-toolchain.yaml
+      - run: rustup update
+  # Our minimum supported rust version is specified here.
+  setup-rust-min-version:
+    steps:
+      # Technically should be on 1.47 here, but that boat sailed - 1.51 is min.
```
rfk, line 15: Since our policy is "must be buildable with the min version" and the problem is in a test crate, an alternative here could be to use `1.47` and only have CI execute a `cargo build` or equivalent, rather than actually running the tests. I think this would skip trying to build `ctest2` because it's not in the workspace default members.

rfk:
(Put another way: I think we're better off getting whatever assurances we can about the min-supported version, rather than raising the number in order to get more assurances that aren't necessarily in the scope of our policy)

mhammond:
We are currently on 1.50 - going back to 1.47 fails with:
```
error[E0658]: use of unstable library feature 'slice_fill'
   --> /home/skip/.cargo/registry/src/github.com-1ecc6299db9ec823/ece-2.1.0/src/aes128gcm.rs:340:64
    |
340 |         output[self.plaintext.len() + 1..padded_plaintext_len].fill(0);
    |                                                                ^^^^
    |
```
So I think the best we can do is 1.50. Sadly though, tests.py still insists on running that systest, so I now just run `cargo test`. yakitty yak :)

rfk (approved):
A valiant effort, thanks @mhammond!

PR #4166, opened 2021-06-01
Add script for running `cargo` from inside XCode.

rfk:
Directly executing `cargo` from inside of an XCode build doesn't
seem to work quite right, e.g. on Big Sur it isn't able to properly
build any `build.rs` due to problems with library paths in the linker

This copies the same technique that we already use for running `cargo`
to build a universal binary, but as a re-useable cargo wrapper rather
than a specific command.

### DIFF
--- a/build-scripts/xc-cargo.sh
+++ b/build-scripts/xc-cargo.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# This is a small wrapper for running `cargo` inside of an XCode build,
+# which unfortunately doesn't seem to work quite right out-of-the-box.
+set -eEuvx
+
+
+# XCode tries to be helpful and overwrites the PATH. Reset that.
+PATH="$(bash -l -c 'echo $PATH')"
+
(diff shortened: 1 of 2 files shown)


build-scripts/xc-cargo.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# This is a small wrapper for running `cargo` inside of an XCode build,
+# which unfortunately doesn't seem to work quite right out-of-the-box.
+set -eEuvx
+
+
+# XCode tries to be helpful and overwrites the PATH. Reset that.
+PATH="$(bash -l -c 'echo $PATH')"
```
rfk, line 9: This is what our existing `xc-universal-binary.sh` script does, and as noted [here](https://github.com/mozilla/application-services/pull/4052#issuecomment-851755959) the XCode build issues do seem to be related to `$PATH`..so 🤷🏻‍♂️

mhammond (approved):
:stamp

PR #4172, opened 2021-06-01
More lint failures in history_metadata.rs

mhammond:
Later rusts complain about some of the tests


grigoryk (approved):
Thanks! Somehow i didn't hit these myself.

mhammond:
> Somehow i didn't hit these myself.

Only in later rust versions!

PR #4174, opened 2021-06-02
Document our rust version policy

mhammond:
This defines our policy, but doesn't actually implement it. Bumping the rust version is going be a bit of a yak-fest, and I don't need to put everyone through this. I'll hold off on landing this until I have a matching PR for the implementation though.

This is basically a copy-paste of the uniffi policy.

Flagging @rfk to review on behalf of most components, + @travis79 and @jhugman to review on behalf of nimbus.

### DIFF
--- a/docs/rust-versions.md
+++ b/docs/rust-versions.md
@@ -1,0 +1,57 @@
+# Rust Versions
+
+Like almost all Rust projects, the entire point of the application-services
+components is that they be used by external projects. If these components
+use Rust features available in only the very latest Rust version, this will
+cause problems for projects which aren't always able to be on that latest
+version.
+
+Given application-services is currently developed and maintained by Mozilla
+staff, it should be no surprise that an important consideration is
+mozilla-central (aka, the main Firefox repository).
+
+## Mozilla-central Rust policies.
+
+It should also come as no surprise that the Rust policy for mozilla-central
+is somewhat flexible. There is an official [Rust Update Policy Document
+](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html])
+but everything in the future is documented as "estimated".
+
+Ultimately though, that page defines 2 Rust versions - "Uses" and "Requires",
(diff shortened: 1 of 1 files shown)


travis79 (approved):
Sounds good to me! My only concerns, neither of which block landing this documentation are:

1. Is there any way to automate the "uses" and "requires" version pinning, i.e. can we scrape this from M-C somehow in our CI to verify that we are both on the right versions, and that we can build with them?
2. What does this mean for building Rust on an M1? Doesn't that currently require a "nightly" version of rustc?  Will pinning to a version of Rust affect engineers with an M1 rig?

skhamis:
> Sounds good to me! My only concerns, neither of which block landing this documentation are:
> 
>     1. Is there any way to automate the "uses" and "requires" version pinning, i.e. can we scrape this from M-C somehow in our CI to verify that we are both on the right versions, and that we can build with them?
> 
>     2. What does this mean for building Rust on an M1? Doesn't that currently require a "nightly" version of rustc?  Will pinning to a version of Rust affect engineers with an M1 rig?

M1 user here, I can answer (2). We discovered that actually even running `cargo +stable test` successfully runs. Which I believe is at least rust 1.52 and above. So actually, if we're able to update from 1.47 pin M1 users can run tests without the `+channel` option!

mhammond:
Thanks Travis!

> Is there any way to automate the "uses" and "requires" version pinning, i.e. can we scrape this from M-C somehow in our CI

We probably could work out how to so this for the min version - "target" version seems more difficult to scrape. However, in the general case a bump of version (particularly the "target" version) will require patches, particularly for new warnings. But we could at least notice the change and take some warning-like action.

While I'd support having this, because this is just a document and doesn't change the status quo in any meaningful way, it hasn't personally pushed me over a line where I'm actually going to try and implement anything like that though :)

mhammond:
James is on PTO and I think we have enough stakeholders with r+, so I'm merging this!

PR #4179, opened 2021-06-03
Database migration system

bendk:
My initial attempt to try to write some generic code to handle migrations.  I expanded the scope slightly and made it handle the entire process of opening a database file.  The difference is that it optionally handles migration failures by deleting the DB file and starting from scratch.

I tested this with the autofill-utils example and it seemed to work in these cases:
  - v0 database (it throws out the CC data in this case)
  - v1 database
  - v2 database (no migrations, just register the functions and setup the temp tables)

  - Note:

### DIFF
--- a/components/autofill/src/db/schema.rs
+++ b/components/autofill/src/db/schema.rs
@@ -5,7 +5,7 @@
 use crate::db::sql_fns;
 use rusqlite::functions::FunctionFlags;
 use rusqlite::Connection;
-use sql_support::open_database::{ErrorHandling, MigrationLogic, Result};
+use sql_support::open_database::{MigrationLogic, Result};
 
 pub const ADDRESS_COMMON_COLS: &str = "
     guid,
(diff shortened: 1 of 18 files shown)


bendk:
Is the dependency check failing because I added `tempfile` to the `sql_support` crate?  If so, what should I do to fix this?

bendk:
I notice that several components currently try to optimistically open newer schema versions.  I didn't implement this one for a couple reasons:

  - The most likely situation is seems like failure a little bit down the line.  For example if any column names changed.
  - If all of the column/table names stay the same, then we can try to update the DB, but do we know that we're writing the correct data to it?  For example if we change a `timestamp` field from being seconds to milliseconds.

That said, one thing we should also definitely not do with a newer schema version is recreate the database.  I'm going to update the code to handle this now.

rfk:
> Is the dependency check failing because I added tempfile to the sql_support crate? If so, what should I do to fix this?

Likely yes, and the way to fix it is to run `./tools/regenerate_dependency_summaries.sh`. Unfortunately this script sometimes doesn't work correctly if you're not on a Mac (because our build includes some Mac-specific dependencies, and `cargo` won't tell us about them unless we're actually running on a Mac) so the alternative approach is to copy-paste the diff from the failing CircleCI check in to your local repo. I'd love to find a way to do better there :-/

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,544 @@
...
+/// The code handles some common cases:
+///
+///   - Opening new databases.  If this is the first time opening the database, then initialize it
+///     to the current schema.
+///
+///   - Migrating existing databases.  If this is an existing database, then run a series of
+///     upgrade functions to migrate it to the current schema.
+///
+///   - Handling migration failures.   If opening or migrating a database results in an error,
+///     we can optionally delete the database file and create a new one.
```
rfk, line 16: It would be great to have a high-level sketch of how to use the component as part of the module-level doc here. We haven't been great at this in the past, but it's worth us trying to keep getting better over time.

I'm thinking specifically about something narrative to set expectations, along the lines of:

* "You'll need to create a `MigrationLogic` struct, that handles A, B, C."
* "Then you pass this to `open_database` which does X, Y, Z."

components/autofill/src/db/schema.rs
```diff
@@ -73,120 +75,132 @@ const CREATE_SHARED_SCHEMA_SQL: &str = include_str!("../../sql/create_shared_sch
 const CREATE_SHARED_TRIGGERS_SQL: &str = include_str!("../../sql/create_shared_triggers.sql");
 const CREATE_SYNC_TEMP_TABLES_SQL: &str = include_str!("../../sql/create_sync_temp_tables.sql");
 
-// The schema version - changes to this typically require a custom
-// migration code.
-const VERSION: i64 = 2;
+pub(super) fn migration_logic() -> MigrationLogic {
+    MigrationLogic {
+        name: "autofill db".to_string(),
+        start_version: 0,
```
rfk, line 81: The docs say that `start_version` is "usually 1", so it may be worth a comment here about why it's zero.

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,544 @@
...
+    pub prepare: Option<DatabaseFunc>,
+    // Initialize a newly created database to <end_version>
+    pub init: DatabaseFunc,
+    // Upgrade functions.  upgrades[n] will migrate version n to n+1
+    pub upgrades: Vec<DatabaseFunc>,
+    // Runs after the init/upgrade functions
+    pub finish: Option<DatabaseFunc>,
+    // How to handle migration errors
+    pub error_handling: ErrorHandling,
+}
```
rfk, line 109: Did you consider using a `trait` for `MigrationLogic` rather than a `struct`? I don't have my head around all the details here yet, but it *might* be more convenient for consumers than using the `DatabaseFunc`-pointer approach. As a sketch, you could have a trait that provides slots for customization by the consumer, like this:

```
pub trait MigrationLogic: Clone {
    //
    // These are the parts the consumer has to fill in.
    //

    // Name to display in the logs
    const NAME: &'static str;
    // The first version that this migration applies to (usually 1)
    const START_VERSION: u32 = 1;
    // The version that the last upgrade function upgrades to.
    // Note: this is intentionally redundant so that it can act as a sanity check on the length of
    // the upgrades vec.
    const END_VERSION: u32;
    // How to handle migration errors
    const ERROR_HANDLING: ErrorHandling;

    // Initialize a newly created database to <end_version>
    fn init(&self, conn: &Connection) -> Result<()>;
    // Upgrade dispatcher. Calling upgrade_from(n) will migrate version n to n+1
    fn upgrade_from(&self, conn: &Connection, n: usize) -> Result<()>;

    //
    // These are optional extra hooks the consumer can override.
    //

    // Runs before the init/upgrade functions
    fn prepare(&self, conn: &Connection) -> Result<()> {
        Ok(())
    }
    // Runs after the init/upgrade functions
    fn finish(&self, conn: &Connection) -> Result<()> {
        Ok(())
    }    

    //
    // And these are default implementations for any `MigrationLogic`.
    //

    fn sanity_check(&self) -> Result<()> {
        // ... logic here ...
        Ok(())
    }

    fn run(&self, conn: &Connection, init: bool) -> Result<()> {
        // ... logic here ...
        Ok(())
    }

    // ... etc ...
}
```

And the consumer implementation could look like:

```
#[derive(Clone)]
pub struct AutofillMigrationLogic { }

impl MigrationLogic for AutofillMigrationLogic {
    const NAME: &'static str = "autofill db";
    const START_VERSION: u32 = 0;
    const END_VERSION: u32 = 2;
    const ERROR_HANDLING: ErrorHandling = ErrorHandling::DeleteAndRecreate;
    
    
    fn prepare(&self, conn: &Connection) -> Result<()> {
        // ... custom logic here ...
        Ok(())
    }

    
    fn init(&self, conn: &Connection) -> Result<()> {
        // ... custom logic here ...
        Ok(())
    }
    
    fn upgrade_from(&self, conn: &Connection, n: usize) -> Result<()> {
        match n {
            0 => upgrade_to_v1(conn),
            1 => upgrade_to_v2(conn),
            _ => panic!("unexpected migration version {}", n),
        }
    }

    fn finish(&self, conn: &Connection) -> Result<()> {
        // ... custom logic here ...
        Ok(())
    }  
}
```

This might help you avoid the need for declaring a specific type of `DatabaseFunc`, and make the process feel a little bit more declarative for consumers.

There are of course likely to be tradeoffs with that approach as well, and I don't feel strongly about it. Just throwing it out there as a suggestion for overall code organization.

rfk (commented):
I'm not familiar enough with the details of the current autofill migrations so I haven't dug in there yet, but wanted to share some initial comments from the perspective of a potential new consumer of this functionality. I'll dig in more again soon.

components/support/sql/src/open_database.rs
bendk: The one thing I wanted to avoid was having a single upgrade function than runs N upgrades.  It seemed too easy to mess up the if branches -- which is what I did with the autofill upgrade when going from v0 to v2, but it ended up working out because would always run `CREATE_SHARED_SCHEMA_SQL` at the end and we decided it was okay to lose data from the v0 databases.

But having a function that runs 1 upgrade, combined with the match seems great.  I'll refactor to use this approach.

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,574 @@
...
+
+pub fn open_database_with_flags<ML: MigrationLogic>(
+    location: DatabaseLocation,
+    open_flags: OpenFlags,
+    migration_logic: &ML,
+    error_handling: ErrorHandling,
+) -> Result<Connection> {
+    // Try running the migration logic with an existing file
+    let initializing = !location.exists();
+    let mut conn = location.open(open_flags)?;
```
mhammond, line 192: Errors here should probably have the `DeleteAndRecreate` treatment? I'm thinking of corrupted databases mainly.

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,574 @@
...
+    let first_result = run_migration_logic(migration_logic, &conn, initializing);
+
+    let final_result = match (&first_result, error_handling) {
+        // Ignore DeleteAndRecreate if the version is too new
+        (Err(Error::VersionTooNew(_)), _) => first_result,
+        // If DeleteAndRecreate was specified, try deleting the file and initializing
+        (Err(e), ErrorHandling::DeleteAndRecreate) => {
+            log::warn!(
```
mhammond, line 200: and here I'm slightly worried that a simple logic error in some edge-case in our migrations might cause us to delete the DB, when a more appropriate action might be to make noise and have an app update fix the issue without deleting any data.

In some ways this is a bit of a philosophical question: would a user prefer to have their (say) logins fail to work for some period, but an app update fix it in the future, or would they prefer to lose all their logins but have the component continue to work?

(It's a genuine question and I'm not sure of the answer, but I do need to think more about this "any error deletes" or whether we need a more nuanced approach?)

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,574 @@
...
+// our other crates.
+pub mod test_utils {
+    use super::*;
+    use std::path::Path;
+    use tempfile::TempDir;
+
+    // Database file that we can programatically run upgrades on
+    //
+    // We purposefully don't keep a connection to the database around to force upgrades to always
+    // run against a newly opened DB, like they would in the real world.  See SYNC-2209 for
```
mhammond, line 236: as above, I think links to github issues are more appropriate.

components/autofill/src/db/schema.rs
```diff
@@ -173,20 +152,62 @@ fn upgrade(db: &Connection, from: i64) -> Result<()> {
...
+        AutofillMigrationLogic::define_functions(&conn)?;
+        conn.set_prepared_statement_cache_capacity(128);
+        Ok(())
+    }
+
+    fn init(&self, db: &Connection) -> Result<()> {
+        Ok(db.execute_batch(CREATE_SHARED_SCHEMA_SQL)?)
+    }
+
+    fn upgrade_from(&self, db: &Connection, version: u32) -> Result<()> {
```
mhammond, line 176: What's the reason for keeping the loop in sqlsupport? I'm wondering if this might deny the ability for some optimizations (eg, you could imagine a hypothetical scenario where it going from v5 to v7 could be done much more efficiently than going via v6.

components/autofill/src/db/schema.rs
```diff
@@ -203,21 +224,61 @@ mod tests {
...
-                cc_exp_month: row.get("cc_exp_month")?,
-                cc_exp_year: row.get("cc_exp_year")?,
-                cc_type: row.get("cc_type")?,
-                time_created: row.get("time_created")?,
-                time_last_used: row.get("time_last_used")?,
-                time_last_modified: row.get("time_last_modified")?,
-                times_used: row.get("times_used")?,
-                sync_change_counter: row.get("sync_change_counter")?,
-            })
+        fn from_row(row: &Row<'_>) -> Self {
```
mhammond, line 227: there's a fair bit of boilerplate there that seems like it's not going to scale that well once we get to V7. If we are leaning into the .sql files with old versions of the schemas, maybe we should also have that same .sql setup the test data? Eg, in the `_v0` file, just below the 'create table' have a few 'insert' clauses?

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,574 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Clone, PartialEq)]
+pub enum ErrorHandling {
+    // Try to start over with a fresh database.  This is the behavior that most components should
+    // use (See SYNC-937, SYNC-816)
```
mhammond, line 88: minor nit, but IMO we shouldn't link to a jira that's only accessible to staff. IOW, please link to the public github issue - just #xxxx is OK.

mhammond (commented):
Thanks Ben,
  One meta-issue I have with this is that I'm not completely sure what problem it is solving - it seems quite a heavy handed way to avoid future versions of sync-2209, and it's not immediately clear to me that it really does make migration errors that much less likely. As I mention below, it also seems to make it more likely that a future migration error might be catastrophic, in terms of nuking all a user's data! Unless I'm missing something, it doesn't seem to make it much easier to test for actual databases on disk to be upgraded - which is a tricky problem, but what would have caught that bug.
  
The diff stat is `+883 −304` which is quite a lot of new code - alot of that is test code, but still - the migration for autofill was entirely encapsulated in a single rust source module, whereas now we have multiple components at play, with new traits and enums.
  
I need to think a little more about this, but I think it might become clearer if places was ported to it too - it must be capable of not only *handling* places, but it must obviously *improve* it IMO. Maybe we should schedule a time to chat about it?

components/support/sql/src/open_database.rs
bendk: Agreed on thinking this through.  The suggestion to delete and recreate by default was based on the fact that it seemed like that was where the past discussions ended up, but there's definitely risk involved.  The one thing I would push for is to make a policy and be consistent and explicit about it.  It seems like a lot of the current migration code is going for the `ReturnError` mode just because that's what naturally happens if you don't make a decision.

components/autofill/src/db/schema.rs
bendk: The main reason is that it made the upgrade function feel simpler if it always just handled one upgrade.  One thing weighing on my mind was the if statement in the autofill upgrade function, which made it so if you were on v0 then you didn't run the v1 upgrade code.  This ended up being fine, since the v0 upgrade deleted the table then we would recreate it at the end, but that was pure luck on my part.

I was thinking about the scenario where you wanted to skip versions.  We could make it so the upgrade function could return something to indicate that.  But the code was already getting complex enough, so I didn't implement it.

components/autofill/src/db/schema.rs
```diff
@@ -73,76 +75,53 @@ const CREATE_SHARED_SCHEMA_SQL: &str = include_str!("../../sql/create_shared_sch
...
         // real UI existed to create cards, so we don't bother trying to
-        // migrate them, we just drop the table so it's re-created with the
+        // migrate them, we just drop the table and re-create it with the
         // correct schema.
-        db.execute("DROP TABLE IF EXISTS credit_cards_data", NO_PARAMS)?;
-    } else if from == 1 {
+        db.execute_batch(
+            "
+            DROP TABLE IF EXISTS credit_cards_data;
+            CREATE TABLE credit_cards_data (
```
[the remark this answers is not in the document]
bendk: The thing I'd like to avoid is re-executing the schema then running more upgrade functions, since the schema that those upgrades work on shouldn't change.   Maybe upgrade functions could return a value that indicates that the database should be deleted and recreated to handle situations like this one.

bendk:
I definitely understand the meta-issue of extra complexity that you bring up Mark.  I thought this was going to be a lot less code when I started, but of course it wasn't :smile: .  Here's what I see as the main benefits:

- It makes it so that each migration follows the same flow.  For example in logins/places we only call `create()` for version 0.  But in autofill `create()` is always called the end of `init()`.
- It makes writing tests easier.
- It reduces amount of boilerplace.  I think this is especially important for migrations, when one logic error can ruin a database (even if we don't do `DeleteAndRecreate`)
- It makes things more explicit.  I like that you have to state the earliest version you support upgrading from and how you're going to deal with errors.

However, a lot of that depends on the system working for other components without needed extra special cases.  If each time we add a new component we need to add new enum values, then we can't really say that we have a consistent flow.  I think your suggestion of porting places would be a great next step.  I'm not sure it needs to improve that code though, for me having things be more consistent is enough of a win.

I'm going to put this on pause until we're sure that it's a good direction.  Chat sounds great, I'll message you about that.

components/autofill/src/db/schema.rs
rfk: FWIW, I get a strong YAGNI feeling on the idea of running multiple migrations at a time, especially since it would multiply the number of potential code paths to test. If we find ourselves wanting to do that at some point in the future, it seems like the sort of thing that we'll want to spend some careful thought and testing on.

components/support/sql/src/open_database.rs
bendk: Maybe there's a better way to solve this one:
  - If there's an error outside of the upgrade logic (opening the DB, getting the current version, etc), then I think we can assume the database is corrupt and should be thrown away
  - If there's an error inside the upgrade logic, what if we still continue on with a fresh DB, but instead of deleting the file we rename it to something like `[db-path].[app-services-version].backup`?
  - At the end of `open_database`, we look for backups from previous versions.  If we see one, we call `MigrationLogic.recover_from(connection_to_db, connection_to_backup, backup_schema_version)`, then delete the backup
  - Most of the time, `recover_from()` would be a no-op.  But if we are ever in the disaster scenario where we realize that a migration bug caused data loss, we can release a new version with recovery code.


Side note: For any system where we hold on to the data in order to try to recover later, we should probably give the user a way to delete the data.  For the above suggestion, what if we made `wipe()` also delete the backup file?

bendk: Closing this one because we decided not to try to delete and recreate the files.

bendk:
This is the updated open_database() code.  The main changes from last week are:
  - Removed the ErrorHandling/DatabaseLocation enum and generally tried to simplify the code
  - Updated the schema testing code.  The new pattern is to create an SQL file that initializes an old version and populates some data.  Then we can test runing all the upgrades checking that the data is still there.
  - Ported webext_storage to using the system
  - Tweaked a couple things based on trying to get places to use the system

On that note, I created a [changeset that gets places to use the new system](https://github.com/bendk/application-services/commit/16a75fc8c7ef10004a32c798dad64e2d380d4905) but I wasn't confident enough to check it in here.  Some notes about that one:
   - It has an ugly panic when trying to run upgrades against a `ReadOnly` database.  I'm not sure what the best thing to do is there.  The old code ignored that case, but that seems to open up the possibility of trying to read from an old database version.
   - The schema code needs to wrap `create_bookmark_roots()` to make the exception handling work.  I think it would be better to move that function into the schema code.  IMO, callling an external function from the upgrade code as an anti-pattern in general.
   - Overall, it seems like this could work but it's a lot of changes and probably should be done as a separate project.

The tests are failing because of a NSS linker error.  It seems to be related to the sql-support tests.  If I remove those than it seems to work.  Not sure what's going on there, if there's an easy way to fix this I can do that.  Otherwise maybe I'll remove them.  Interestingly, the tests work if you run `cargo test -p sql-support`, but not `cargo test`.

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,432 @@
...
+#[derive(Error, Debug)]
+pub enum Error {
+    // Generic error meaning that something went wrong during the migration.  You can return this
+    // from your upgrade functions to signal that the database is beyond repair and can't be
+    // migrated.
+    #[error("MigrationError: {0}")]
+    MigrationError(String),
+    // Error with the migration logic struct, for example the number upgrade functions doesn't
+    // line up with start_version and end_version
+    #[error("MigrationLogicError: {0}")]
```
mhammond, line 37: This doesn't seem used either? In fact, it looks like almost none of them are. How would you feel about avoiding the new error enum, having the only "real" error be the rusqlite one, and `VersionTooNew` could even just panic - there's nothing a consumer (or us!) could do to handle this in practice, and "less is more" WRT errors I think. I'm a bit on the fence about this though, so if you feel strongly let me know.

mhammond (approved):
This looks great - I don't need to see it again based on the error decision, but I'm happy to chat more about it if you want.

components/support/sql/src/open_database.rs
bendk: I decided to leave in the error struct for 2 reasons:
  - It makes it easy to distinguish errors while opening the database vs errors during normal operations.  Maybe this will help us with tracking down where the corruption errors are happening
  - It feels a bit scary to me to panic while in the middle of a transaction.  This doesn't really apply to `VersionTooNew`, but it does apply to the current panics in the upgrade_from() function when the version is too old.

I'm also on the fence though.  1) could probably be handled a different way and 2) might be more paranoia than a real issue.

components/support/sql/src/open_database.rs
```diff
@@ -0,0 +1,432 @@
...
+use rusqlite::{Connection, OpenFlags, NO_PARAMS};
+use std::path::Path;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    // Generic error meaning that something went wrong during the migration.  You can return this
+    // from your upgrade functions to signal that the database is beyond repair and can't be
+    // migrated.
+    #[error("MigrationError: {0}")]
```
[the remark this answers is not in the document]
bendk: I though it could be used by the upgrade_from() function if it detected some issue with the data it couldn't recover from.  But I'm thinking that's pretty unlikely.  I deleted that variant and a few others.

Issue #4054, opened 2021-04-30
Failures when running gradle locally


bendk:
The `./gradlew test` is failing for me when I run it locally.  The failing tasks are `:crashtest:testDebugUnitTest` and `:crashtest:testReleaseUnitTest` .  I believe markh said that the tests are failing for him as well.  We don't run those tests in CI.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2174)
┆Story Points: 2
┆Epic: local development, ci and release (backlog)
┆Sprint: SYNC - 2021-06-11

bendk:
I just looked through the `.cargo/config.yml` and I don't see the gradle tests being run there either.  So maybe it's not a local issue, maybe the tests aren't passing in general.

rfk:
FWIW I can also reproduce these failures locally.

> I just looked through the .cargo/config.yml and I don't see the gradle tests being run there either. 

IIRC the gradle/android tests get run in taskcluster (from under `./taskcluster/`) rather than in Circle (which I assume it what you meant here with the `.yml` file).

rfk:
Hrm, doing a `./gradlew crashtest:clean` seems to have fixed the issue for me. @bendk can you confirm whether this is still failing for you locally, even after a clean?

rfk:
The errors @bendk is getting are different from mine:

```
> Task :crashtest:testDebugUnitTest FAILED
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.robolectric.util.ReflectionHelpers$7 (file:/home/bdk/.gradle/caches/modules-2/files-2.1/org.robolectric/shadowapi/3.8/d638f001e81ef737bb35db2964312360cc996a94/shadowapi-3.8.jar) to method java.lang.ClassLoader.getPackage(java.lang.String)
WARNING: Please consider reporting this to the maintainers of org.robolectric.util.ReflectionHelpers$7
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
mozilla.appservices.crashtest.CrashTestHelpersTest > testErrorsAreThrown FAILED
    java.lang.NoClassDefFoundError at Class.java:-2
        Caused by: java.lang.ClassNotFoundException at SandboxClassLoader.java:132
            Caused by: java.lang.RuntimeException at ClassInstrumentor.java:125
                Caused by: java.lang.IllegalArgumentException at null:-1
[...]

PR #4180, opened 2021-06-04
Use a shared helper for gradle uniffi-bindgen tasks.

rfk:
We currently have four copies of the logic for calling UniFFI
from gradle, several with their own little tweaks. Let's take
the best lessons from each and consolidate them into a single
shared helper.

This version is, I think, an improvement over all the existing
ones:

* Like the nimbus version, it registers the task as generating
  java source files, which should help things integrate cleanly
  into the broader build.
* Like the fxa-client version, it declares explicit inputs so
  that we can avoid re-running the task when nothing has changed.
* New in this version, it declares an input dependency on anything
  that would affect the version of `uniffi-bindgen` that is in use,
  since that too will change the generated outputs.

That last change should really help with smooth updates to the
version of UniFFI in use. Previously, if you changed the UniFFI
version it would not trigger a cleanup of old generated bindings,
and you'd get `java.lang.UnsatisfiedLinkError` errors like in #4054.
(This is UniFFI's last-line-of-defence measure to prevent you from
using bindings generated with one version of UniFFI on a library
compiled with another, which is very useful, but an extremely
confusing failure mode).

With this change, an updated version of UniFFI will invalidate
the gradle task that generates the bindings, causing them to
be regenerated with the new version.

Fixes some of the errors from #4054, and takes a step towards
the goal of #3988.

### DIFF
--- a/components/nimbus/src/nimbus.udl
+++ b/components/nimbus/src/nimbus.udl
@@ -74,7 +74,6 @@ enum NimbusError {
     "DatabaseNotReady",
 };
 
-[Threadsafe]
 interface NimbusClient {
     [Throws=NimbusError]
     constructor(
--- a/publish.gradle
+++ b/publish.gradle
@@ -200,3 +200,27 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
         checkMavenArtifacts.dependsOn(checkFileSizeTask)
     }
 }
+
+
+
+// A convenience function for configuring a `uniffi-bindgen` task,
+// with appropriate dependency info. This fill call `uniffi-bindgen`
+// on the provided `.udl` file in order to generate Kotlin language
+// bindings and include them in the source set for the project.
+ext.configureUniFFIBindgen = { udlFilePath ->
+    android.libraryVariants.all { variant ->
+        def uniffiGeneratedPath = "generated/source/uniffi/${variant.name}/java"
+        def t = tasks.register("generate${variant.name.capitalize()}UniFFIBindings", Exec) {
+            workingDir project.rootDir
+            commandLine 'cargo', 'uniffi-bindgen', 'generate', "${project.projectDir}/${udlFilePath}", '--language', 'kotlin', '--out-dir', "${buildDir}/${uniffiGeneratedPath}"
+            outputs.dir "${buildDir}/${uniffiGeneratedPath}"
+            // Re-generate if the interface definition changes.
+            inputs.file "${project.projectDir}/${udlFilePath}"
+            // Re-generate if our uniffi-bindgen tooling changes.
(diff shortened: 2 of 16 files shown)


rfk:
@travis79 could you please r? this from the Nimbus perspective?

components/nimbus/src/nimbus.udl
```diff
@@ -74,7 +74,6 @@ enum NimbusError {
     "DatabaseNotReady",
 };
 
-[Threadsafe]
```
rfk, line 77: In UniFFI v0.11.0 the `[Threadsafe]` attribute is now the default and only way to safely expose objects, and as such is deprecated.

publish.gradle
```diff
@@ -200,3 +200,27 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
...
     }
 }
+
+
+
+// A convenience function for configuring a `uniffi-bindgen` task,
+// with appropriate dependency info. This fill call `uniffi-bindgen`
+// on the provided `.udl` file in order to generate Kotlin language
+// bindings and include them in the source set for the project.
+ext.configureUniFFIBindgen = { udlFilePath ->
```
skhamis, line 210: really nice!

skhamis (approved):
I really like this change and think it's definitely needed as having a common build task will allow less unexpected things to happen (ex. new logins uniffi kotlin test couldn't find the generated .kt file successfully). 

Will make sure logins conforms to this once I rebase after this PR lands

travis79 (approved):
My only question is the dropping of adding the generated code to the source sets, is this no longer necessary?  Not worth blocking on, I just noticed that this PR eliminated that step.

skhamis:
> My only question is the dropping of adding the generated code to the source sets, is this no longer necessary? Not worth blocking on, I just noticed that this PR eliminated that step.

@travis79 That was one concern I had too but I _think_ the new consolidated task has [registerJavaGeneratingTask](https://www.javadoc.io/doc/com.android.tools.build/gradle/0.8.2/com/android/build/gradle/api/BaseVariant.html#registerJavaGeneratingTask) which states:

`Adds to the variant a task that generates Java source code. This will make the compileJava task depend on this task and add the new source folders as compilation inputs. The new source folders are also added to the model. ` 

Which is pretty nifty (looks like it was used in the nimbus gradle too!)

rfk:
> My only question is the dropping of adding the generated code to the source sets, is this no longer necessary? 

Right, my understanding is basically what @skhamis said: the `registerJavaGeneratingTask` task is a nicer and cleaner way to achieve the same affect. Nimbus switched over to using it already in #4036 and this will bring the other projects into alignment.

PR #4182, opened 2021-06-05
Add Logins API Update ADR

### LINKED ISSUE #4101: Create proposal for logins api changes
* Investigate the impact of not exposing the migration function
* See if we can refactor/optimize the handling of empty usernames in a way that makes `potential_dupes_ignoring_usernames` obsolete
* Identify where add/update is used and see if it can be replaced by save/update

┆Issue is synchronized with this [Jira Spike](https://jira.mozilla.com/browse/SYNC-2206)
┆Story Points: 8
┆Epic: Update logins api
┆Sprint: SYNC - 2021-06-25

lougeniaC64:
Fixes #4101 

This is a proposal for addressing the issue of moving logins to per-field encryption while maintaining the validation and de-duping logic. Please consider this merely a proposal. Feel free to make comments, suggestions, or make changes. I aimed to be thorough at a high level but also concise, so if there are places where clarifications or additional information would be good, please let me know.

  - Note:


docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,77 @@
...
+
+  The proposed changes will reduce the above list of functions to the following:
+    - `save_or_update`
+      - This function will replace `add` and `update`
+    - `import_multiple`
+      - This function will be removed once the Fennec to Fenix migration period ends
+
+  Both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` functions will removed from the API. Neither function is called in Firefox iOS. Android Components uses both to provide validation and de-duping before logins are added or updated. We can eliminate the need to externalize these functions by replicating this logic in the new `save_or_update` function.
+
+  If we exclude the encrypted `username` and `password` fields from [the Nul check](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/login.rs#L446) in the validate and fixup logic, we can eliminate the need for the `get_all`, `get_by_base_domain`, and `get_by_id` functions to have the encryption key.
```
lougeniaC64, line 58: I just want to highlight this part of the proposal because this is where I am uncertain. At present, the three "get" API functions I mentioned above call `Login::from_row` which calls `maybe_fixup` which calls `validate_and_fixup`, which is where the `Nul` check occurs. My understanding is that we're fixing up logins upon db retrieval in order to gradually clean up our login records. If we want to continue to do this, I'm choosing to exclude the encrypted password and username fields so that the get functions don't need the encryption key. But if we no longer need to fixup retrieved records, that simplifies things. But for certain if we want to leave the validation logic unchanged and continue fixing up retrieved records, all three get functions will need the encryption key.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,77 @@
...
+  * Uses the same encryption model as autofill so there is consistency in our approaches
+* Cons
+  * Potentially weakens the fixup and validation logic to either cover fewer logins fields or fewer functions
+  * Breaks the logins API and requires consumer code changes
+  * Requires consumer code to both encrypt logins fields _and_ pass the encryption key when calling either `save_or_update` and `import_multiple`
+
+
+### Option 2 - Implement a different encryption approach
+
+(Judging from the comments in our tickets I don't think we've identified a viable alternative but I'm adding this section here in case I'm mistaken and there are other options to document.)
```
lougeniaC64, line 72: If I'm wrong here, please let me know and/or feel free to add other options.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+## Decision Outcome
+
+(I'm purposely leaving this blank until we arrive at a decision.)
+
+
+## Pros and Cons of the Options
+
+### Option 1 - Reduce the API functions that require the encryption key and pass the key the remaining functions
```
rfk, line 34: sp?: "pass the key *to* the"

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,77 @@
...
+    - [update](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L611)
+    - [get_all](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L273)
+    - [get_by_base_domain](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L279)
+    - [get_by_id](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L330)
+    - [check_valid_with_no_dupes](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L663)
+    - [potential_dupes_ignoring_username](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L721)
+    - [import_multiple](https://github.com/mozilla/application-services/blob/1248a352cb2701b92395f2783bee8a88d18de455/components/logins/src/db.rs#L455)
+
+      **Note:** Functions related to sync have been omitted as it’s assumed they’ll have access to decrypted data.
```
rfk, line 48: nit: the apostrophes here seem to be `U+2019 : RIGHT SINGLE QUOTATION MARK`, which, not sure how I feel about that in an markdown doc 😅

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,77 @@
...
+
+  The proposed changes will reduce the above list of functions to the following:
+    - `save_or_update`
+      - This function will replace `add` and `update`
+    - `import_multiple`
+      - This function will be removed once the Fennec to Fenix migration period ends
+
+  Both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` functions will removed from the API. Neither function is called in Firefox iOS. Android Components uses both to provide validation and de-duping before logins are added or updated. We can eliminate the need to externalize these functions by replicating this logic in the new `save_or_update` function.
```
rfk, line 56: Being able to push this logic further down into the shared code also seems like an independent win, aside from the benefits it has in enabling the encryption stuff.

docs/adr/0000-use-markdown-architectural-decision-records.md
```diff
@@ -0,0 +1,28 @@
...
+* [Michael Nygard's template](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) – The first incarnation of the term "ADR"
+* [Sustainable Architectural Decisions](https://www.infoq.com/articles/sustainable-architectural-design-decisions) – The Y-Statements
+* Other templates listed at <https://github.com/joelparkerhenderson/architecture_decision_record>
+* Formless – No conventions for file format and structure
+
+## Decision Outcome
+
+Chosen option: "MADR 2.1.2", because
+
+* Implicit assumptions should be made explicit.
```
rfk, line 20: A meta note that I support being explicit about this, thank you for adding this 👍🏻

docs/adr/0001-update-logins-api.md
rfk: I don't have all the context from design discussions in the autofill component, so the below may not be tractable, but I wanted to suggest it for completeness.

IIUC, the proposed API exposes `encrypt_string` and `decrypt_string` functions that consumers need to use in order to access the individual encrypted fields. Would it be useful to instead have a higher-level `decrypt_login_record` function that takes a whole login record with encrypted fields, and:

* If the encrypted fields are all valid, returns some sort of `PlainTextLogin` object will all fields decrypted
* If the encrypted fields contained invalid data, repairs the data in the db and then returns an error

I kind of feel like we already decided not to do this in the autofill component, in which case, perhaps it's worth us recording the reasons why as part of this ADR.

rfk (commented):
I don't feel sufficiently well-versed on the problem to click "approve" or "request changes" here, but I like ADRs and I have opinions so I figured I may as well share them :-)

docs/adr/0001-update-logins-api.md
mhammond: > Would it be useful to instead have a higher-level `decrypt_login_record`

Possibly, yeah - especially here where we have 2 encrypted fields. However, unless I misunderstand, this doesn't actually solve the problems here - it would still require the "get" functions to have the key available, which I think we should try and avoid. None of our other components do these kinds of fixups at "get" time. I'd be fine with a new API designed expressly for fixups that did require the key, but I'm inclined to suggest the API shape is wrong where getters end up writing fixed up records to the DB.

docs/adr/0001-update-logins-api.md
mhammond: I think this would be something like "let the component hold on to the key and therefore make the api unavailable if the key isn't held" - which is roughly what we have now. There's no reason we couldn't stick with the same fundamental model we have now even with per-field encryption and sqlite.

(I'm inclined to think this model defeats alot of the purpose in an encrypted store, but doesn't defeat all of it, so I guess we should keep it on the table)

mhammond (commented):
This looks great, but doesn't really seem to have a specific API proposal - eg, how will dupes be found etc?

docs/adr/0001-update-logins-api.md
rfk: In a little more detail, was I was thinking here was something like the following consumer experience:

```
let encryptedLogin = store.get_by_id("foobar")
try:
  let decryptedLogin = store.decrypt_login_record(key, encryptedLogin)
except CorruptedLoginRecordError:
  print("oh no what do we do?")
```

That is, guard against invalid data at the point where you decrypt the fields.

Having written it out like that though, I'm not sure I like it. On one hand, it seems like it could spread the "fixup at read time" logic to lots of parts of the *consumer* code, which architecturally seems like the opposite of what we want. On the other hand, it seems better to refuse to decrypt corrupted data such as a login with a nul in the username, rather than hand out known-invalid data to the consumer.

lougeniaC64: Actually, I think this may work?? I mean if these three "get" functions didn't call the fixup logic, the consumer would retrieve the encrypted data and then be responsible for decrypting it. That's the pattern that was established in autofill anyway. So if we have this higher level `decrypt_login_record` function, we'd be adhering to the same pattern. And then we could remove the fixup logic from the get calls.

docs/adr/0001-update-logins-api.md
lougeniaC64: I'm assuming this would mean keeping the encryption key in the db "open" functions , require a code review, and still allow the same consolidation of API functions.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,104 @@
...
+    - Pass the per-field encryption key to the `import_multiple` function
+      - This function will be removed once the Fennec to Fenix migration period ends
+    - Remove both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` from the API
+      - Neither function is called in Firefox iOS
+      - Android Components uses both to provide validation and de-duping before logins are added or updated so we can eliminate the need to externalize these functions by replicating this logic in the new `add_or_update` function
+    - Create a `decrypt_and_fixup_login` function that both decrypts a login _and_ performs the validate and fixup logic
+      - This will eliminate the need for the `get_all`, `get_by_base_domain`, and `get_by_id` API functions to perform the fixup logic
+
+    Making the above changes will reduce the API functions requiring the encryption key to the following:
+    - `add_or_update`
```
lougeniaC64, line 62: If we approve of combining the decryption and fixup logic and exposing that in our API, I think we should also have a similar function encryption for consistency's sake. That would mean `add_or_update` would no longer require the encryption key and a new `encrypt_and_fixup_login` would require it.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,104 @@
...
+  Propsed changes:
+    - Combine the `add` and `update` functions into a new `add_or_update` function
+      - This will allow the removal of consumer code that distinguishes when a login record should be created or updated
+    - Pass the per-field encryption key to the `import_multiple` function
+      - This function will be removed once the Fennec to Fenix migration period ends
+    - Remove both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` from the API
+      - Neither function is called in Firefox iOS
+      - Android Components uses both to provide validation and de-duping before logins are added or updated so we can eliminate the need to externalize these functions by replicating this logic in the new `add_or_update` function
+    - Create a `decrypt_and_fixup_login` function that both decrypts a login _and_ performs the validate and fixup logic
+      - This will eliminate the need for the `get_all`, `get_by_base_domain`, and `get_by_id` API functions to perform the fixup logic
```
bendk, line 59: I like this idea a lot, but how would it work?  Would `decrypt_and_fixup_login()` return the plaintext username/password alongside the fixed up login?  Would it store the decrypted values inside the `Login` struct?  And how would `encrypt_and_fixup_login()` work?

What about having 2 types: `Login` and `EncryptedLogin`:
- The `get_*` methods return `EncryptedLogin` instances
- We define `EncryptedLogin.decrypt_and_fixup(key) -> Login` and `Login.encrypt_and_fixup(key) -> EncryptedLogin`
- `add_or_update()` inputs an `EncryptedLogin`

But then I also wonder if we need expose `EncryptedLogin` at all.  Are consumers going to want to use logins without accessing the username/password and without going through validation?  If not, then the API can be simplified by always giving the consumer decrypted logins and requiring the key in the `get_*` methods and `save_or_update()`.   It doesn't simplify the key management for the consumer if can get logins without a key but then they always decrypt them afterwards.

Another option would be to make the main API calls return `Logins`, but then have some specialized functions that return `EncryptedLogin`.  This would be good if the consumers sometimes, but rarely, needed to use logins without a key.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,104 @@
...
+  - Update the database layer logic to prevent accessing the database without an encryption key
+  - Combine the `add` and `update` functions into `add_or_update`
+  - Remove both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` from the API
+
+  Making the above changes will reduce the list of API functions requiring the encryption key to the following:
+    - `add_or_update`
+    - `import_multiple`
+    - `get_all`
+    - `get_by_base_domain`
+    - `get_by_id`
```
bendk, line 91: Is this a copy-and-paste error?  I thought the idea was that we would store the key so the consumer code would never need to pass it in.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,104 @@
...
+    - Remove both the `potential_dupes_ignoring_username` and `check_valid_with_no_dupes` from the API
+      - Neither function is called in Firefox iOS
+      - Android Components uses both to provide validation and de-duping before logins are added or updated so we can eliminate the need to externalize these functions by replicating this logic in the new `add_or_update` function
+    - Create a `decrypt_and_fixup_login` function that both decrypts a login _and_ performs the validate and fixup logic
+      - This will eliminate the need for the `get_all`, `get_by_base_domain`, and `get_by_id` API functions to perform the fixup logic
+
+    Making the above changes will reduce the API functions requiring the encryption key to the following:
+    - `add_or_update`
+    - `decrypt_and_fixup_login`
+    - `import_multiple`
```
bendk, line 64: Another change is how the username/password fields on the Login object works.  It seems minor, but there's definitely some edge cases.  I think the main questions are:

  - How do you get/set values the fields?
  - How do you set the username field to missing?
  - How do we know when the time_password_changed field should be bumped?

bendk (commented):
Looks great overall and I plan on stealing that design template for the DB migration ADR.

docs/adr/0001-update-logins-api.md
rfk: > We define EncryptedLogin.decrypt_and_fixup(key) -> Login and Login.encrypt_and_fixup(key) -> EncryptedLogin

As a purely technical note, uniffi won't currently let you define methods like that on a `dictionary` type, it would have to be either a free-standing function or a method on the `interface` type for `Store`.

IIRC, the reason we want to let consumers work with `EncryptedLogin` records is because prompting for the encryption key is a disruptive user event. So for example, if the user is on a page, we want to be able to check if we have any saved logins for that page before we prompt the user to unlock their keystore and actually decrypt the record.

docs/adr/0001-update-logins-api.md
rfk: I like the symmetry of this, which I think will be helpful for consumers to keep a coherent mental model of the operations. So on the way out of storage it would be like:

```
enc_login = store.get_by_base_domain("foo.com") // Or similar, I actually don't remember the precise API details...
if do_we_want_to_use_this(enc_login):
    key = get_the_key_from_somewhere()
    login = store.decrypt_and_fixup(key, enc_login)
    do_stuff_with_the(login)
```

And on the way in to storage it would be like:

```
enc_login = store.encrypt_and_fixup_login(key, login)
store.add_or_update(enc_login)
```

Hrm...on the other hand, it seems like the `encrypt_and_fixup_login` is going to want to do the dupe checking and what-not anyway, and from a pure implementation point-of-view it probably wants to just write the new record in there if it's not a dupe. (In fact, would there be a chance of creating dupes if two threads were running this procedure at the same time, because we do the check non-atomically with saving the data?).

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Pros
+  * Improves the logins API for consumers by combining add/update functionality (see [#3899](https://github.com/mozilla/application-services/issues/3899) for details)
+  * Removes redundant validation and de-duping logic in consumer code
+  * Uses the same encryption model as autofill so there is consistency in our approaches
+* Cons
+  * Breaks the logins API and requires consumer code changes
```
rfk, line 71: In practice I expect we're going to be the ones making the consumer code changes anyway, so maybe no that much of a con :-/

docs/adr/0001-update-logins-api.md
bendk: Mark explained that to me, and I'm finally 100% convinced.  From an ADR-perspective, I'll just note that you could have methods for those checks that didn't return logins.  Something like `get_potential_logins_count()`.  The reason for doing that is to avoid any potential confusion on the consumer code side with mixing up the 2 types of logins.  Maybe there would be exceptions when trying to get a plaintext data from an encrypted login or maybe the consumers would decide to hold on to the keys longer than they needed to just in case they ever needed to decrypt a login.

However, I'm realizing that the snippet you pasted above feels very natural and avoids all the issues.  You can also extend it to handling multiple logins:

```
enc_logins = store.get_by_base_domain("foo.com")
if user_wants_to_use_any_login(enc_logins):
    key = get_the_key_from_somewhere()
    logins = [store.decrypt_and_fixup(key, enc_login) for enc_login in enc_logins]
    present_login_selector(logins)
```

docs/adr/0001-update-logins-api.md
bendk: I'm liking the symmetry too and having `add_or_update` input an encrypted login would also give us the ability to update logins without ever decrypting them.  For example we could have a `last_seen` field that gets updated even if the user doesn't unlock their key.  This is bordering on YAGNI, but it still seems nice to me.

The dupe checking issue seems real to me, but it also seems like even with 1 function you would have issues.  Could we leverage the database for this?

  - Add a `dupe_hash` column that's a secure hash of the all of the fields that need to be unique together (username, hostname, realm, etc.). Add a unique constraint for this one.
  - `encypt_and_fixup_login()` calculates the hash
  - `add_or_update()` uses the hash to figure out what it should do

bendk: This also brings up another question:  Is the login in your second snippet the same object as the login in the first?  For example, in this scenario:
  - User selects an existing login and we auto-fill the form fields
  - User changes the username field value
  - User submits the form

Do we expect and/or require the client to send us the login dict used it step 1 or will they create a new dict with login data and give us that?

docs/adr/0001-update-logins-api.md
lougeniaC64: Yeah, I think I'm going to remove this. IIRC I added it to distinguish from an option (that doesn't exist in this doc btw) which would have maintained our current logins API. But I think regardless of how we approach maintaining the fixup and de-duping issues we're going to want to replace `add` and `update` with `add_or_update`, which will always require consumer code changes.

docs/adr/0001-update-logins-api.md
lougeniaC64: Not a copy paste error but perhaps a bit of unintentional misdirection. These functions _would_ require the encryption key but that fact is immaterial since the consumer wouldn't need to pass it in. I'll remove this list to prevent confusion.

docs/adr/0001-update-logins-api.md
lougeniaC64: > Hrm...on the other hand, it seems like the `encrypt_and_fixup_login` is going to want to do the dupe checking and what-not anyway, and from a pure implementation point-of-view it probably wants to just write the new record in there if it's not a dupe.

Yeaaaaaah not only that, `add_or_update` not having the key means that we'd have no way of supporting `time_password_changed` because we need to compare the old and new decrypted password values before, you know, adds or updates. 

In conclusion I vote no to my own idea of adding `encrypt_and_fixup_login` to the API.

rfk: > In conclusion I vote no to my own idea of adding encrypt_and_fixup_login to the API.

👍🏻, and let's please record the reasoning as part of this ADR because there's no way I'm going to remember it three months from now

docs/adr/0001-update-logins-api.md
lougeniaC64: @bendk Sorry for the delayed response, I somehow missed this.


>     * How do you get/set values the fields?
What edge cases were you thinking of specifically. I'm assuming you're referring to temporarily holding the the decrypted username of password values to make the deduping or fixup work cleaner. If that's the case, I assumed we might need internal structs however I'll readily admit to not thinking too much about this as this doc is a bit higher level. Were there specific cases you were concerned about?

>     * How do you set the username field to missing?
So we've already set the precedent that an empty string value indicates that the record has been scrubbed, which is a meaningful distinction that we probably don't want to overload. So maybe a missing username would be NULL value in the database and perhaps a None Option in the Logins struct. @mhammond How are you feeling about `Option<String>` usernames? 😬


>     * How do we know when the time_password_changed field should be bumped?

At present this only happens during db updates. So in the new `add_or_update` field I'm thinking we'd continue updating that field if we determine that the login already exists in the database. Since that function will have the encryption key, this shouldn't be a problem. Let me know if there are edge cases here that you're thinking of that I may not have considered.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,96 @@
...
+## Decision Drivers
+
+* Avoiding the implementation of new security approaches that may require additional time and security resources
+* Establishing a standard encyrption approach across components
+
+
+## Considered Options
+
+* Option 1 - Reduce the API functions that require the encryption key and pass the key to the remaining functions
+* Option 2 - Implement a different key management approach
```
mhammond, line 24: I think "Implement a different" is misleading - it's actually more like "Implement the same"! I'd reword this as something like "Keep the general shape of the API that's in place now - the app can pass the encryption key at any time to "unlock" the API, and re-lock it at any time, but the API in its entirety is only available when unlocked."

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,96 @@
...
+(I'm purposely leaving this blank until we arrive at a decision.)
+
+
+## Pros and Cons of the Options
+
+### Option 1 - Reduce the API functions that require the encryption key and pass the key to the remaining functions
+
+* Description
+
+  Currently the below logins API functions would require the per-field encryption key:
```
mhammond, line 38: I think we should expand on why the `get` functions need the key, as it's not obvious that it's not because we want to return the decrypted values from our API, but because they call the `validate` and/or `fixup` functions which do need access the the decrypted values to operate correctly. Maybe as a "note" where you mention you are ignoring sync?

docs/adr/0001-update-logins-api.md
mhammond: All the above sounds fine, but as a meta-comment, I don't think it's necessary for this ADR to get into that level of detail as that doesn't really drive the higher-level decision about what scheme to use. (That's also *partially* true for some other points, but there is a fine line because we do want to thrash out things not *directly* related but that still might cause problems with the scheme we agree on)

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,96 @@
...
+  * Uses the same encryption model as autofill so there is consistency in our approaches
+* Cons
+  * Requires consumer code to both encrypt login fields _and_ pass the encryption key when calling either `add_or_update` and `import_multiple`
+
+
+### Option 2 - Implement a different key management approach
+
+* Description
+
+  With our current SQLCipher implementation, if the encryption key is not available when opening the database, the logins API is not accessible. We could replicate this logic with per-field encryption allowing us to store the encryption key for use with any calls that require it.
```
mhammond, line 79: Just a couple of tweaks here:
* Unlike the option above, this would mean the publicly exposed API deals only with decrypted login records - all encryption is internal (which works because we always have the key)

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,96 @@
...
+
+  Proposed changes:
+  - Update the database layer logic to prevent accessing the database without an encryption key
```
mhammond, line 82: And here, we could tweak it to say "Any attempt to use the API will fail as the login records are encrypted or decrypted as the key isn't available" - I guess I'm kinda saying an explicit check probably isn't really necessary as the failures will happen implicitly.

mhammond (approved):
This is great Lougenia, nice work! I think there's some wordsmithing still to do, but I'm hitting the big green "approve" button because I agree with and approve the key architectural decisions here which is the entire point of an ADR.

docs/adr/0001-update-logins-api.md
bendk: Part of the question was really aim aimed at lower-level details that I agree the ADR doesn't need to address.  The one thing that I'd like to see expanded on is the general flow of fetch login -> decrypt login -> save login back to the db.

What `decrypt_and_fixup()` do?
  - Adds plaintext username/password data to the existing dictionary
  - Returns a new dictionary with the plaintext data instead of encrypted data (converts `EncryptedLogin` to `Login`)

My suggestion is that `decrypt_and_fixup()` returns a new type.  Otherwise we can get into ambiguous situations, like if the decrypted password and encrypted password fields don't match up.

What get passed to `add_or_update()`
  - The dictionary that was fetched from the DB
  - A new dictionary that the consumer code creates
  - Is the data encrypted or plaintext?

I kind of think that `add_or_update()` should input a newly created dictionary with plaintext data, but I'm not sure.

docs/adr/0001-update-logins-api.md
lougeniaC64: Since this isn't a feature that needs to be implemented, I'm going to remove the first bullet point and add, "Any attempt to use the API will fail as the login records aren't encrypted or decrypted if the key isn't available", to the description.

docs/adr/0001-update-logins-api.md
```diff
@@ -0,0 +1,96 @@
...
+* Status: proposed
+* Date: 2021-05-27
+
+Technical Story: [#4101](https://github.com/mozilla/application-services/issues/4101)
+
+## Context and Problem Statement
+
+The per-field encryption in autofill, which we are planning to replicate in logins, separates the storage and encryption logic by limiting the storage layer to the management of encrypted data.
+
+Applying this approach in logins will break the existing validation and deduping code so we need a way to implement per-field encryption while supporting the validation and de-duping behavior.
```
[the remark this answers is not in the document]
lougeniaC64: FYI I added these points between the "Context and Problem Statement" and "Decision Drivers" sections in order to keep the former section as brief as possible.

lougeniaC64:
> This is great Lougenia, nice work! I think there's some wordsmithing still to do, but I'm hitting the big green "approve" button because I agree with and approve the key architectural decisions here which is the entire point of an ADR.

Thanks, Mark! I've addressed all our feedback but if you see anything that still isn't quite right, let me know and I'll update the adr.

PR #4183, opened 2021-06-07
Replace places history_metadata's protobuf implementations with uniffi.

mhammond:
This is intended to be a small but significant step on the road to uniffi-ing
places. It takes all the hard stuff from @rfk 's #3900, but limits the scope to just
the new-ish history_metadata module.

The hope is that because @grigoryk is actively working on this code, the transition to uniffi might be less disruptive. IMO, it's close to landable and lays a good foundation for us to take on the rest of #3900.

<del>No attempt at swift, no attempt at android-components.</del>Swift and android also work

@grigoryk, how disruptive does this look to you? @rfk, any concerns with this?


components/autofill/Cargo.toml
```diff
@@ -23,7 +23,7 @@ sync15 = { path = "../sync15" }
 sync15-traits = {path = "../support/sync15-traits"}
 thiserror = "1.0"
 types = { path = "../support/types" }
-uniffi = "^0.10"
+uniffi = "^0.11"
```
mhammond, line 26: doh! Not sure what went wrong here, but I didn't mean to upgrade uniffi, and nothing here depends on it.

components/places/src/storage/history_metadata.rs
```diff
@@ -4,14 +4,82 @@
...
+}
+
+impl FromSql for DocumentType {
+    #[inline]
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        Ok(match value.as_i64()? {
+            0 => DocumentType::Regular,
+            1 => DocumentType::Media,
+            other => {
+                // seems safe to ignore?
```
grigoryk, line 29: Stepping maybe too far ahead, I guess that can happen if other clients implement document types we do not, in which case we _may_ rewrite them to `regular` as a result of this conversion - depending on how these results are used, although not likely currently (e.g. these are never written back by the api consumers). And we don't want that kind of misalignment to happen at all, really, so perhaps this is moot.

components/places/src/places.udl
```diff
@@ -0,0 +1,42 @@
...
+
+enum DocumentType {
+     // A page that isn't described by any other more specific types.
+    "Regular",
+    // A media page.
+    "Media",
+};
+
+// This is used as an "input" to the api.
+dictionary HistoryMetadataObservation {
```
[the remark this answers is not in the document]
grigoryk: fwiw `apply_metadata_observation_impl` is basically already written against a `HistoryMetadataCompoundKey` type, except for the first few lines

grigoryk (commented):
That basically looks good to me, thanks for pushing it out!

Not very disruptive at all, now would be a great time for these sort of things.

components/places/android/src/main/java/mozilla/appservices/places/LibPlacesFFI.kt
```diff
@@ -9,10 +9,13 @@ import com.sun.jna.Library
 import com.sun.jna.Pointer
 import com.sun.jna.PointerType
 import com.sun.jna.StringArray
-import mozilla.appservices.support.native.RustBuffer
 import mozilla.appservices.support.native.loadIndirect
 import org.mozilla.appservices.places.BuildConfig
 
+// yay - 2 RustBuffers :)
```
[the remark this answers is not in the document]
rfk: (But it's probably worth expanding this comment to talk about why)

components/places/src/ffi.rs
```diff
@@ -143,9 +144,41 @@ implement_into_ffi_by_protobuf!(msg_types::HistoryVisitInfos);
...
 implement_into_ffi_by_protobuf!(msg_types::BookmarkNodeList);
-implement_into_ffi_by_protobuf!(msg_types::HistoryMetadata);
-implement_into_ffi_by_protobuf!(msg_types::HistoryMetadataList);
 implement_into_ffi_by_delegation!(
     crate::storage::bookmarks::PublicNode,
     msg_types::BookmarkNode
 );
+
+// XXX - this should be in some other crate?
+// uniffi!
```
rfk, line 153: I think it would be reasonable to include this in UniFFI. One thing to watch out for would be that UniFFI depends on `ffi-support` and exposes the depended-on version as `uniffi::deps::ffi_support`. If you want to move this macro into UniFFI, it may be worth doing `use uniffi::deps::ffi_support` rather than taking our own separate dependency on it, in order to ensure that we use the same version.

components/places/uniffi.toml
```diff
@@ -0,0 +1,5 @@
+[bindings.kotlin]
+# Typically there'd be no `.uniffi` tail, but without it we see obscure
+# kotlin errors about being unable to resolve which `rustCall` to use.
```
rfk, line 3: Please leave an additional note here along the lines of "this can go away once we UniFFI the entire places crate".

rfk (approved):
> @rfk, any concerns with this?

None at all, this is exciting to see and is definitely a mode we want to support!

components/places/src/places.udl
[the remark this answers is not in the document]
mhammond: > I'd consider lifting (url, referrer, search_term) tuple into a `key` type, otherwise `url` as the only required item appears to act as an "anchor" here.

My take is that the key is an implementation detail but not strictly relevant to consumers. I'm also not convinced it is an implementation detail that will survive - it's seems like a reasonable implementation choice for the very first experiments, but it's not a fundamental property of anything. Future, more nuanced implementations might make different choices at different times - that it may end up turning into a "kind of a key, except when ..."

> I'd consider lifting (url, referrer, search_term) tuple into a `key` type, otherwise `url` as the only required item appears to act as an "anchor" here.

But that's entirely true. Url is the only required item. Url can not be blank, the other items can. When making an observation you should supply what you know. There's no natural concept of a "key" when visiting a page - it's just an artificial construct supporting the current implementation. The existing `VisitObservation` API doesn't specify the key, even though it's ending up in a DB and certainly will use a key. When querying history you also don't get to specify a "key" - you can query by various things, not just the database's primary key. I think the same thing applies here.

So if it's OK with you, I think I'd still rather leave `key` out at this time.

components/places/src/places.udl
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+namespace places {};
+
+// Some kind of namespacing for uniffi would be ideal. Multiple udl/macro defns?
+// Everything below is from the crate::storage::history_metadata module...
```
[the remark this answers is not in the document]
mhammond: > An `interface` here wouldn't work in combination w/ the rest of this component being non-uniffied?

No, I can't quite see how that would work in the current setup :( It would be close though and maybe we can arrange for something, but I'm still thinking we should charge ahead while we work that out.

components/places/src/storage/history_metadata.rs
mhammond: FTR, I think ignoring it is the wrong approach - it's not what other enums in this crate do, so there's no good reason to set a new precedent. Getting unknown values in via sync is an interesting issue, but one Sync should explicitly deal with rather than writing "invalid" enum values. We could even consider using a `CHECK` constraint in the sql?

components/places/src/ffi.rs
mhammond: I actually backed away from that, and replaced that comment with:
```
/// This is only necessary because we have a kinda "hybrid" FFI situation -
/// some things generated via UniFFI, others by hand. Because this macro only
/// makes sense in this Frankenstein world it's not in UniFFI itself.
```

components/places/src/places.udl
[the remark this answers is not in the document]
grigoryk: Let's leave it out for now. I'm inclined to agree with your assessment of how this may play out in the future, but also inclined to think that the timelines will be quite stretched out (meaning, the current "key" shape will persist for quite a while). Let's see how it plays out!

mhammond:
So this got a bit messier dealing with both hand-written and uniffi generated FFI functions and structures, particularly around errors. The patch also got a bit bigger because the kotlin linter was run which adjusted stuff not otherwise touched by this patch - but all in all I think this is worth going with, especially as we expect more metadata changes in the shortist term. I'm going to re-flag Grisha and Ryan even though Ryan is on PTO this week.

components/places/src/ffi.rs
```diff
@@ -110,27 +209,60 @@ fn get_code(err: &Error) -> ErrorCode {
...
+}
+
+/// This is very very hacky - we somehow need to ensure the same error hierarchy
+/// exists for both hand-written FFI functions and those generated by uniffi,
+/// and there doesn't seem to be a clean way of doing that. So our .udl defines
+/// a single error type - ErrorWrapper::Wrapped(). The `String` message there
+/// is, roughly, `format!("{}|{}", extern_error.code, extern_error.message)`.
+/// There then exists code on the Swift and Kotlin side of the world which
+/// unpacks this and returns the exact same error objects as if it was an
+/// `ExternError` in the first place.
```
rfk, line 240: FWIW, I agree with your implied assessment that this hackery should be short-lived enough that it's not worth trying to do anything nicer here 👍🏻

components/places/src/storage/history_metadata.rs
```diff
@@ -223,39 +291,36 @@ pub fn get_latest_for_url(db: &PlacesDb, url: &Url) -> Result<Option<HistoryMeta
     Ok(metadata)
 }
 
-pub fn get_between(db: &PlacesDb, start: i64, end: i64) -> Result<HistoryMetadataList> {
-    let metadata = db.query_rows_and_then_named_cached(
+pub fn get_between(db: &PlacesDb, start: i64, end: i64) -> Result<Vec<HistoryMetadata>> {
```
rfk, line 294: I still smile every time I see one of these flip from `FooList` to `Vec<Foo>`

rfk (approved):
I can't comment on the feature-specific stuff, but the UniFFI-related stuff here looks like a solid improvement even with a few (temporary) rough edges 👍🏻

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -386,66 +423,37 @@ open class PlacesReaderConnection internal constructor(connHandle: Long) :
...
-            } finally {
-                LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuffer)
+            return rustCallUniffi(this) {
+                mozilla.appservices.places.uniffi.placesGetHistoryMetadataBetween(this.handle.get(), start, end)
             }
         }
     }
 
-    override suspend fun queryHistoryMetadata(query: String, limit: Int): List<HistoryMetadata> {
+    override suspend fun queryHistoryMetadata(query: String, limit: Long): List<HistoryMetadata> {
```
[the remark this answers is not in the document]
grigoryk: Yup, it was i32 changed to be on the Rust side - https://github.com/mozilla/application-services/pull/4330/files to match the `Int` here.

But I see that this PR changes both sides to be Long/i64, which will work. Though, I _think_ we do want Int/i32 for the `limit` param, that's what they are in the rest of the API surface (and that's the type that matches our usage of it, as well).

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -929,7 +943,12 @@ interface WritableHistoryMetadataConnection : ReadableHistoryMetadataConnection
     /**
      * Record or update metadata information about a URL. See [HistoryMetadataObservation].
      */
-    suspend fun noteHistoryMetadataObservation(key: HistoryMetadataKey, observation: HistoryMetadataObservation)
+    suspend fun noteHistoryMetadataObservation(observation: HistoryMetadataObservation)
```
grigoryk, line 946: Not sure we need to expose the generic version, we don't have anything that'll use it afaik. But I suppose it doesn't hurt.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -929,7 +943,12 @@ interface WritableHistoryMetadataConnection : ReadableHistoryMetadataConnection
...
+    // There's a bit of an impedance mismatch here; `HistoryMetadataKey` is
+    // a concept that only exists here and not in the rust. We can iterate on
+    // this as the entire "history metadata" requirement evolves.
```
grigoryk, line 949: Thanks :)

components/places/src/ffi.rs
```diff
@@ -6,9 +6,105 @@
...
+fn places_metadata_delete_older_than(handle: i64, older_than: i64) -> Result<(), ErrorWrapper> {
+    log::debug!("places_note_history_metadata_observation");
+    CONNECTIONS.get(
+        Handle::from_u64(handle as u64)?,
+        |conn| -> Result<_, ErrorWrapper> {
+            crate::storage::history_metadata::delete_older_than(conn, older_than)?;
+            Ok(())
+        },
+    )
+}
```
grigoryk, line 107: I kinda expected that these bindings would go away in this PR and there will be some include of them out of generated uniffied rs files, but i guess not?

grigoryk (approved):
That's great, @mhammond ! I've mostly looked through the API and impl changes (less so, all the glue), but it looks good!

components/places/src/ffi.rs
mhammond: Because this is just a partial conversion, these bindings remain because they need to adapt from a "handle" model still used by the most of the FFI. They will go away once the handle goes away.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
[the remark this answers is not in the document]
mhammond: 👍 - it's back to an int

mhammond:
Hey @nbhasin2, do you mind having a quick look at the swift parts of this? I'd hate to break what you just recently fixed :)

nbhasin2:
@mhammond This looks good. I wasn't able to fully test it locally as glean version was outdated but don't see any issues in swift code. 

I did see `noteHistoryMetadataObservation()` with no key as param. Should I be using this or the other 3 new methods you added?

https://github.com/mozilla/application-services/pull/4183/files?diff=split&w=1#diff-7062352bc04f4b3b923a816873fbabf5a4e0922c721962c05b299073e2f462b2R874

nbhasin2 (approved):

mhammond:
Thanks!

> Should I be using this or the other 3 new methods you added?

I think one of the 3 makes the most sense, but @grigoryk is probably who you should listen to here more than me :)

PR #4187, opened 2021-06-08
Consolidate megazord-related gradle logic into a shared helper. [ci full]

rfk:
Prior to this commit, each of our individual components had its
own copy of gradle logic for depending on the megazord, including
a couple of lengthy comments about fiddling around with JNA for
the Android Studio test runner.

I've moved all that logic into a helper function in `publish.gradle`
so that we can maintain it in a single location going forward.
There may be a cleaner or more idiomatic way to achieve this kind
of code-sharing in gradle, but I figure this *has* to be better
than our current copy-paste-based approach.

I've manually confirmed that a local `./gradlew test` completes
successfully, as well as testing via Android Studio.

Connects to #https://github.com/mozilla/application-services/issues/3988
(although doesn't go as far as I'd like; one step at a time!)

### DIFF
--- a/components/autofill/android/build.gradle
+++ b/components/autofill/android/build.gradle
@@ -27,11 +27,6 @@ android {
         }
     }
 
-    sourceSets {
-        test.resources.srcDirs += "$buildDir/rustJniLibs/desktop"
-        test.resources.srcDirs += "${project(':full-megazord').buildDir}/rustJniLibs/desktop"
-    }
-
     testOptions {
         unitTests {
             includeAndroidResources = true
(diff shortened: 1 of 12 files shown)


rfk:
@grigoryk can you think of any potential downsides from doing this?

mhammond (approved):
`+78 −580` 🚀 

(This is really a rubber stamp, so waiting for Grisha is advised :)

rfk:
This is starting to get stale, and it really seems like an easy-to-reverse-if-we-hate-it change to me, so I'm going to rebase this and get it merged.

Issue #3389, opened 2020-07-23
Only build megazords once in Taskcluster CI


thomcc:
Apparently we're still doing this probably around 10x per push in taskcluster... Maybe more.

I remember from when I set up the megazords that this was avoidable, but a pain -- apparently we never got to doing it.

I just heard from @eoger how much Taskcluster CI costs us a month and... This is worth fixing, since it should reduce this number substantially (potentially around 10x -- there should also be a CI speedup as a result).

That said, tragically, I don't remember the details beyond that it requires adjusting how our android dependencies are set up, and possibly requires a patch or two in https://github.com/mozilla/rust-android-gradle.

So, perhaps step 1 would be investigation into how to do this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1512)
┆Story Points: 8
┆Epic: local development, ci and release (backlog)
┆Sprint: SYNC - 2021-06-25

eoger:
FWIW the last time I tried to tackle this long-lived issue, I tried the simple solution of building all the android modules in the same run, but I cargo kept re-building archs it built before (never figured out why).

data-sync-user:
➤ Ryan Kelly commented:

I'm planning to spend a bit of quality time with our gradle scripts this sprint, optimistically pulling this in as part of that work.

rfk:
> We could adjust the dependency in the gradle scripts so that running the tests depends on building full-megazord, but just
> building and publishing the android package does not.

I took a stab at this approach in #4188. Even if it does nothing for our CI costs, it's a nice improvement to the local test-running experience!

rfk:
> But ideally we'd get it down to just 6, so I'm not calling this successful just yet.

In #4189 it's down to 7: 6 builds for the actual publication of the megazord on 6 platforms, and 1 build for running all the tests in a single TaskCluster task. I'm happy to stop there I think; it may be possible to invest significantly more energy and figure out how to test all the modules using the built megazord artifact, but that can be a problem for another day.

PR #4188, opened 2021-06-08
Reduce gradle task dependencies on megazord build [ci full].

rfk:
Connects to #3389 (and is hopefully a big step in the right direction,
but not a full fix).

Prior to this commit, the gradle tasks for building and for testing
each component would depend on the "cargoBuild" gradle task of the
megazord, which means they would built it for all configured targets.
In TaskCluster we configure *seven* targets, so each `module-build-X`
TaskCluster job will build the full megazord seven times.

This commit makes two changes:

* Building the gradle package of a component no longer depends on
  building the megazord at all. We doesn't actually need to the
  Rust code built in order to package up the Kotlin files, and I
  think this dependency was a hangover from when each component
  used to build its own individual Rust library.
* Running the gradle tests of a component now depends only on
  building the megazord for the native host platform, since that's
  the only one that is actually needed for running the tests.

Given that we have 11 components that depend on the megazord,
each of which will now avoid building 6 useless copies of it
for non-native platforms, I expect this will save a total of
66 expensive `cargo build` invocations in a full TaskCluster run.

I've also included a few other tweaks to do less unnecessary
work in TaskCluster, so let's see how it works out in practice...

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -251,7 +251,6 @@ ext.dependsOnTheMegazord = {
     // has their own build dir)
     android {
         sourceSets {
-            test.resources.srcDirs += "$buildDir/rustJniLibs/desktop"
             test.resources.srcDirs += "${project(':full-megazord').buildDir}/rustJniLibs/desktop"
         }
     }
@@ -271,8 +270,8 @@ ext.dependsOnTheMegazord = {
         // avoiding other configurations from being resolved.  Tricky!
         testImplementation files(configurations.jnaForTest.copyRecursive().files)
     }
-    // For running local tests, depend on a local `cargo build` of the megazord.
-    // Unfortunately the `cargoBuild` task isn't available until after evaluation.
+    // For running local tests, depend on a local native `cargo build` of the megazord.
+    // Unfortunately the `cargoBuild` tasks aren't available until after evaluation.
     evaluationDependsOn(":full-megazord")
     afterEvaluate {
         android.libraryVariants.all { variant ->
@@ -281,10 +280,8 @@ ext.dependsOnTheMegazord = {
                 productFlavor += "${it.name.capitalize()}"
             }
             def buildType = "${variant.buildType.name.capitalize()}"
-            tasks["generate${productFlavor}${buildType}Assets"].dependsOn(project(':full-megazord').tasks["cargoBuild"])
-
-            // For unit tests.
-            tasks["process${productFlavor}${buildType}UnitTestJavaRes"].dependsOn(project(':full-megazord').tasks["cargoBuild"])
+            def nativeCargoBuildTask = "cargoBuild${rootProject.ext.nativeRustTarget.capitalize()}"
+            tasks["process${productFlavor}${buildType}UnitTestJavaRes"].dependsOn(project(':full-megazord').tasks[nativeCargoBuildTask])
         }
     }
 }
--- a/taskcluster/ci/android-build/kind.yml
+++ b/taskcluster/ci/android-build/kind.yml
@@ -37,7 +37,7 @@ jobs:
         - [source, taskcluster/scripts/toolchain/rustup-setup.sh]
         - [source, taskcluster/scripts/toolchain/cross-compile-setup.sh]
         - [rsync, '-a', /builds/worker/fetches/libs/, /builds/worker/checkouts/src/libs/]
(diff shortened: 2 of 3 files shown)


taskcluster/ci/android-build/kind.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         - [source, taskcluster/scripts/toolchain/rustup-setup.sh]
         - [source, taskcluster/scripts/toolchain/cross-compile-setup.sh]
         - [rsync, '-a', /builds/worker/fetches/libs/, /builds/worker/checkouts/src/libs/]
-        - [bash, '-c', 'echo "rust.targets=linux-x86-64,x86_64\n" > local.properties']
+        - [bash, '-c', 'echo "rust.targets=linux-x86-64\n" > local.properties']
```
rfk, line 40: This is the task that gets run on an ordinary PR, and thankfully it was only building the megazord twice rather than 70+ times. Still, I think we can save a bit of time here. We only need to build for the native platform in order to run the tests, so I've removed `x86_64` from the target list here.

You could argue that there's some value in ensuring that the megazord *builds* correctly on the mobile target platforms on every PR, but in that case I think we should have all the target platforms here rather than just this one. Unless it was here for some other reason that I don't anticipate?

taskcluster/ci/android-build/kind.yml
```diff
@@ -49,5 +49,3 @@ jobs:
       toolchain:
         - android-libs
         - desktop-linux-libs
-        - desktop-macos-libs
-        - desktop-win32-x86-64-libs
```
rfk, line 53: Since this task doesn't build for windows or darwin, I don't see why it should have to download the pre-built libs for those platforms.

publish.gradle
```diff
@@ -251,7 +251,6 @@ ext.dependsOnTheMegazord = {
     // has their own build dir)
     android {
         sourceSets {
-            test.resources.srcDirs += "$buildDir/rustJniLibs/desktop"
```
rfk, line 254: This line wasn't doing anything, because none of the component packages actually build any rust code into their own `$buildDir`; it's all in the megazord one below.

rfk:
First, taking a look at taskcluster in its default mode, without the special `[ci full]` flag in the PR title. This configuration runs a single task named `android-build-pr` that builds and tests all the gradle components in one go.

[Here](https://firefox-ci-tc.services.mozilla.com/tasks/VVG-r0B0Szu6jQ41tL35mw) is a pretty representative run of the task prior to this commit. It completes in 8 minutes and if you click through to the log, you can see it running two `cargo build`s of the megazord before reporting `496 actionable tasks: 481 executed, 15 up-to-date`.

[Here](https://firefox-ci-tc.services.mozilla.com/tasks/NZgL2oyQS4OcsTBxf82vXA) is the run from this PR. It completed in 6 minutes and if you click through to the log, you can see it running a single `cargo build` before reporting `494 actionable tasks: 479 executed, 15 up-to-date`. So we've eliminated a couple of tasks and shaved around a minute off the CI run. Not bad!

Now, to try it with `[ci full]`...

rfk:
> Now, to try it with [ci full]...

[Here](https://firefox-ci-tc.services.mozilla.com/tasks/ZWHsndY3QKeHy4bWHGnQtw) is the `module-build-crashtest` task run from the commit prior to this one. It took 11 minutes to run and the log shows it running 6 different `cargoBuild` tasks (my earlier figure of 7 was inaccurate because we no longer build for windows desktop).

[Here](https://firefox-ci-tc.services.mozilla.com/tasks/DLtUbooWSL6NNb7QZL-URA) is the corresponding `module-build-crashtest` task run from this commit. It took 5 minutes to run and the log shows it running only a single `cargoBuild` task. Much better! The other 10 `module-build-X` tasks seem to show similar improvement, and I reckon this could save a full taskcluster-hour of combined execution time on a `[ci full]` run.

As far as I can tell the resulting artifacts are still being build correctly, but I'll try to download a couple and QA them locally to make sure.

MihaiTabara (approved):
LGTM from releng standpoint. I'm not knowledgeable on the Rust bits, but the changes make sense.

MihaiTabara:
CC @bhearsum - one of the PRs I had mentioned a while back.

PR #4189, opened 2021-06-08
Run all android tests at once when cutting a release [ci full]

### LINKED ISSUE #3389: Only build megazords once in Taskcluster CI

rfk:
Prior to the commit, the Taskcluster job for each individual component
would execute `./gradlew component:test` and would thus need to build
the Rust code for the megazord.

With this commit, each of those tasks now depend on the existing
`android-build` task which knows how to run all the tests in a single
build. This should reduce the number of times we have to build the
megazord from once-per-components to just once. (Plus the independent
builds we do for actually publishin the megazord).

This is separate from #4188 because I'm not sure if it'll work, and I didn't
want to hold up the already-substantial improvements in that PR.

Fixes #3389.

### DIFF
--- a/taskcluster/ci/module-build/kind.yml
+++ b/taskcluster/ci/module-build/kind.yml
@@ -42,7 +42,6 @@ job-defaults:
       # add `win32-x86-64-gnu` back in to the list of targets here.
       - [bash, '-c', 'echo "rust.targets=arm,arm64,x86_64,x86,darwin,linux-x86-64\n" > local.properties']
     gradlew:
-      - ':{module_name}:testDebug'
       - ':{module_name}:assembleRelease'
       - ':{module_name}:publish'
       - ':{module_name}:checkMavenArtifacts'
--- a/taskcluster/ci/android-build/kind.yml
+++ b/taskcluster/ci/android-build/kind.yml
@@ -16,10 +16,10 @@ kind-dependencies:
 jobs:
   pr:
     attributes:
-      run-on-pr-type: normal-ci
+      run-on-pr-type: all
       resource-monitor: true
     needs-sccache: false # TODO: Bug 1623426 deal with this once we're in prod
-    run-on-tasks-for: [github-pull-request]
+    run-on-tasks-for: [github-pull-request, github-release]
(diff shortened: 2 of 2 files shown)


taskcluster/ci/module-build/kind.yml
```diff
@@ -42,7 +42,6 @@ job-defaults:
       # add `win32-x86-64-gnu` back in to the list of targets here.
       - [bash, '-c', 'echo "rust.targets=arm,arm64,x86_64,x86,darwin,linux-x86-64\n" > local.properties']
     gradlew:
-      - ':{module_name}:testDebug'
```
rfk, line 45: Ordinarily I would be a little concerned about not testing the actual code that will actually end up in the release artifacts, but we already weren't doing that here - `testDebug` will do a separate debug build of the android code in order to run the tests, and will run it on top of a separately-compiled megazord lib that then gets thrown away. So I don't see any downside to splitting this out into a separate task in this case.

rfk:
Well this *seems* to work...the individual `module-build` tasks now complete in just three minutes, the `android-build-pr` runs like it does on normal PRs, and the `module-build-full-megazord` still does its thing and produces a sensible-looking artifact.

rfk:
@lougeniaC64 and @mhammond, pinging you both for review on these in case there's anything I've missed.

lougeniaC64 (approved):
LGTM! 🚢

PR #4190, opened 2021-06-08
SDK-278 Remove app_id matching from `is_experiment_available`

travis79:
This removes the app_id matching from `is_experiment_available` so we no longer need to fake the app_id while testing and debugging.  The top-level targeting only considers the `app_name` and `channel` now.

### DIFF
--- a/components/nimbus/src/evaluator.rs
+++ b/components/nimbus/src/evaluator.rs
@@ -464,11 +464,10 @@ mod tests {
             ..Default::default()
         };
 
-        // Application context for matching the above experiment.  If any of the `app_name`, `app_id`,
-        // or `channel` doesn't match the experiment, then the client won't be enrolled.
+        // Application context for matching the above experiment.  If the `app_name` or
+        // `channel` doesn't match the experiment, then the client won't be enrolled.
         let mut context = AppContext {
             app_name: "NimbusTest".to_string(),
-            app_id: "org.example.app".to_string(),
             channel: "nightly".to_string(),
             ..Default::default()
         };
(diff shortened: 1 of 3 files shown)


components/nimbus/src/evaluator.rs
```diff
@@ -109,9 +111,15 @@ pub fn evaluate_enrollment(
...
 ///
-/// `is_release` supports two modes:
-/// if `true`, available means available for enrollment: i.e. does the `app_name`, `app_id` and `channel` match.
-/// if `false`, available means available for testing: i.e. does only the `app_name` match.
+/// # Arguments:
+/// - `app_context` The application parameters to use for targeting purposes
+/// - `exp` The `Experiment` to evaluate
+/// - `is_release` Supports two modes:
+///     if `true`, available means available for enrollment: i.e. does the `app_name` and `channel` match.
+///     if `false`, available means available for testing: i.e. does only the `app_name` match.
```
jhugman, line 119: I'm still not sure about the naming of this: you correctly point out that `true` is for enrollment, `false` is for testing, so perhaps this is a simple `mode`, and we provide two further functions which call this one: 

```rust
fn is_experiment_available_for_testing(…) -> bool

fn is_experiment_available_for_enrollment(…) -> bool
```

jhugman (approved):
The code looks good! The docs look great!

One last thing: given the `app_id` is linked to an id that is immutably linked to the app (i.e. hard for one app to pretend to be another), we now lean quite heavily on the `NIMBUS_URL` being a secret to keep drive-by forkers from enrolling in experiments that may be harmful to them.

components/nimbus/src/evaluator.rs
travis79: 😬 I just fixed up the comment format to be compatible with rustdocs, the comment was pre-existing, as was the parameter and its usage.  I agree that the naming is a big confusing, fortunately this is internal to Nimbus and I'll file an issue to remind us to clear up this tech-debt.

travis79:
I agree that we now lean heavily on app_name and app_id, both of which are supplied by the application and could be faked.  It is still possible to filter on app_id through the targeting string/jexl though, so if it is a concern we could rely on that method.

PR #4192, opened 2021-06-09
Uniffi-ing the Tabs component

### LINKED ISSUE #4035: RemoteTabsProvider.setLocalTabs fails with `PoisonError` after unrelated sync failures
(sorry @Amejia481, I'm going to edit your initial bug comment here to add ongoing context; -@rfk)

## Current Status

As described in [this comment](https://github.com/mozilla/application-services/issues/4035#issuecomment-851102193), we think the issue here is that `SyncManager::do_sync` holds a `MutexGuard` for the tabs-store and logins-store for the duration of a sync. Any uncaught panic in the course of a sync will poison these mutexes, rendering the stores unusable even if the original error had nothing to do with tabs or logins.

In addition to fixing the underlying issues causing a panic during a sync (e.g. ref #4153) we should refactor things inside `SyncManager` so that it only holds the mutex for the tab store *while it is syncing tabs*, not for the whole duration of the sync.

------

The issue was addressed on AS `61.0.7` but it's happening again on AS `74.0.1`.

Sentry Fenix nightly https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/8825532/
Fenix issue https://github.com/mozilla-mobile/fenix/issues/12162

image

**Stacktrace:**

```
mozilla.appservices.remotetabs.RemoteTabProviderException: called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"
    at mozilla.appservices.remotetabs.rust.RustError.intoException(RustError.kt:4)
    at mozilla.appservices.remotetabs.RemoteTabsProvider.setLocalTabs(RemoteTabsProvider.kt:11)
    at mozilla.components.browser.storage.sync.RemoteTabsStorage$store$2.invokeSuspend(RemoteTabsStorage.kt:15)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:15)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:11)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2624)
┆Epic: Uniffi-cation of logins, tabs and history metadata
┆Fix Versions: Release 92
┆Sprint End Date: 2021-07-23

skhamis:
(@rfk edit - Fixes #4035)

Thanks to @grigoryk for starting this work effort in #4151!

Accompanying PRs:
**android-components: https://github.com/mozilla-mobile/android-components/pull/10475**

---

  - Note:


components/tabs/src/sync/engine.rs
```diff
@@ -144,9 +148,19 @@ impl<'a> SyncEngine for TabsEngine<'a> {
             };
             remote_tabs.push(tab);
         }
-        self.storage.replace_remote_tabs(remote_tabs);
+        self.store
+            .storage
+            .lock()
+            .unwrap()
+            .replace_remote_tabs(remote_tabs);
         let mut outgoing = OutgoingChangeset::new("tabs", inbound.timestamp);
```
mhammond, line 156: I think we should move the creation of outgoing up a statement, then just take the mutex once to give it the remote ones and to fetch the local ones.

components/tabs/src/sync/engine.rs
```diff
@@ -161,7 +175,7 @@ impl<'a> SyncEngine for TabsEngine<'a> {
             let local_record = ClientRemoteTabs {
                 client_id: local_id,
                 client_name,
-                device_type,
+                device_type: device_type.into(),
```
mhammond, line 178: using the clone of `DeviceType` in storage.rs will avoid this into (and I think is just generally cleaner - there doesn't seem to be a good reason for this crate to have both versions). That might even mean you can avoid that `From` impl too.

components/tabs/src/tabs.udl
```diff
@@ -0,0 +1,33 @@
+namespace tabs {
+
+};
+
+// TODO should this be [Threadsafe]? this runs into problems, e.g. mem_cached_state: Cell<MemoryCachedState> on TabsStorage
+interface TabsStore {
+    constructor();
+
+    [Self=ByArc]
```
mhammond, line 9: I think only `register_with_sync_manager` needs `Self=ByArc`

components/tabs/src/tabs.udl
```diff
@@ -0,0 +1,33 @@
+namespace tabs {
+
+};
+
+// TODO should this be [Threadsafe]? this runs into problems, e.g. mem_cached_state: Cell<MemoryCachedState> on TabsStorage
```
mhammond, line 5: This comment can die

examples/tabs-sync/src/tabs-sync.rs
```diff
@@ -70,7 +75,11 @@ fn main() -> Result<()> {
...
                 log::info!("Syncing!");
-                match store.sync(&cli_fxa.client_init, &cli_fxa.root_sync_key, &device_id) {
+                match Arc::clone(&store).sync(
+                    &cli_fxa.client_init,
+                    &cli_fxa.root_sync_key,
+                    &device_id,
+                ) {
                     Err(e) => {
                         log::warn!("Sync failed! {}", e);
                     }
```
mhammond, line 85: I'm fairly sure we don't want the -shm and -wal files below :)

components/tabs/src/sync/engine.rs
skhamis: Per our conversation, this won't be possible since we get `remote_client.device_type` from Sync15 and cannot directly use the local `DeviceType`

skhamis:
@mhammond Thanks for the review Mark! I got most of your changes in & this should be close to the "completed" PR (I was able to remove pretty much all kotlin code).

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,11 @@ Use the template below to make assigning a version number during the release cut
...
 
 -->
+
+## Tabs
+
+### ⚠️ Breaking changes ⚠️
+
+Tabs has been Uniffi-ed.
+
+`setLocalTabs` is now `updateLocalState`.
```
mhammond, line 28: As we discussed, we will just rename the function locally to match what android expects.

components/tabs/src/sync/engine.rs
```diff
@@ -144,9 +148,10 @@ impl<'a> SyncEngine for TabsEngine<'a> {
             };
             remote_tabs.push(tab);
         }
-        self.storage.replace_remote_tabs(remote_tabs);
         let mut outgoing = OutgoingChangeset::new("tabs", inbound.timestamp);
-        if let Some(local_tabs) = self.storage.prepare_local_tabs_for_upload() {
+        let storage = self.store.storage.lock().unwrap();
```
mhammond, line 152: I don't think this matters much, but it might help avoid foot-guns in the future if we held the mutex for as short as possible - something like:

```
let local_tabs = {
  // keep the mutex for as short as possible
  let storage = self.store.storage.lock().unwrap();
  storage.replace_remote_tabs(remote_tabs);
  storage.prepare_local_tabs_for_upload()
};
if let Some(local_tabs) = local_tabs { ...
```

(in practice the statements below are going to be quick, but it seems worth being more explicit about mutex lifetimes where we can)

mhammond (approved):
This looks great 🚢

skhamis:
Hmmm so I might've jumped the gun a little on removing all the android code. So we have two options:
1. Continue without kotlin code but will need to do things similar to Logins where there will be some breaking changes and will need to have an accompanying PR + time the lands properly:
   -- Removing `getHandle()` and adding `registerWithSyncManager` is an unavoidable breaking change
   -- Uniffi adds "Exception" to the end of an error: So to retro-fit I'd have to name a new rust the enum error `RemoteTabProvider` which is a little confusing
      --- **Note: This is only used in the explicit `sync()` call so we'd also have to remove it eventually**
2. Update the existing kotlin code to be a thin wrapper and add all new functionality -> update android components to use all the new features -> then silently remove the old/unused code

@grigoryk, @rfk thoughts?

rfk:
> Continue without kotlin code but will need to do things similar to Logins where there will
> be some breaking changes and will need to have an accompanying PR + time the lands properly

IMHO we're better off eating the breaking changes rather than carrying around compatibility shims indefinitely. (Which IIUC is that you decided to do here, so I'm mostly sharing this as an explicit 👍🏻)

skhamis:
Agreed! We should move ahead with what's best in the long run. @grigoryk I have made an accompanying PR https://github.com/mozilla-mobile/android-components/pull/10475 to help ease this transition similar to logins

PR #4193, opened 2021-06-09
Remove incorrect serde implementation for one copy of the DeviceType …

mhammond:
Due to [this uniffi issue](https://github.com/mozilla/uniffi-rs/issues/478) we have a few duplicates of the `DeviceType` enum, and one of them has incorrect serde serialize and deserialize implementations. Fortunately it turns out they are unused. While I was there I added tests for the one in fxa-client which does implement it correctly.

### DIFF
--- a/components/fxa-client/src/internal/http_client.rs
+++ b/components/fxa-client/src/internal/http_client.rs
@@ -794,6 +794,53 @@ pub enum DeviceType {
     Unknown,
 }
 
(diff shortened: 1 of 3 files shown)


skhamis (approved):
Nice! Looks good to me and will definitely help in the uniffi-ing Tabs PR -- at least until we can share types eventually via uniffi.

PR #4194, opened 2021-06-09
Extract the bulk of shared gradle logic into script plugins [ci full].

rfk:
Prior to this commit, the `build.gradle` for each individual
project contained copy-pasted logic for configuring Android
and protobuf. Some of them were even using slightly different
versions of the same dependencies.

This this commit, there are two new gradle scripts that encapsulate
this shared logic:

* ./build-scripts/component-common.gradle, for basic Android
  and Kotlin setup.
* ./build-scripts/protobof-common.gradle, for configuration specific
  to the protobuf plugin.

Hopefully this will make the logic easier to maintain going forward.

There may be gradle-related reasons that we shouldn't do this, so
I will reach out to some Android experts for a review. But even if this
makes our builds slightly slower, I think its a win overall for a team
of non-gradle-experts who need tomaintain this stuff.


build-scripts/component-common.gradle
```diff
@@ -0,0 +1,73 @@
...
+    // We're told the following is required to support new AndroidX support libraries.
+    // See mozilla-mobile/android-components#842
+    //
+    // tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    //    kotlinOptions {
+    //         jvmTarget = rootProject.ext.build.jvmTargetCompatibility
+    //     }
+    // }
+    //
+    // However, it doesn't work when factored out into this script.
```
rfk, line 51: Specifically, if I try to do the previous snippet in this build script, gradle tries to look up `org.jetbrains....` as a property on `$project`, which fails. Maybe there's some trick I don't know about to have it looked up as a global package name?

build-scripts/component-common.gradle
```diff
@@ -0,0 +1,73 @@
...
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    testImplementation "junit:junit:$junit_version"
+    testImplementation "org.robolectric:robolectric:$roboelectric_core_version"
+    testImplementation "org.mockito:mockito-core:$mockito_core_version"
+
+    androidTestImplementation "com.android.support.test:runner:$android_test_runner_version"
+    androidTestImplementation "com.android.support.test.espresso:$espresso_core_version"
```
rfk, line 72: A handful of our current components to not currently depend on all of these test dependencies, so this change will introduce a few potentially-spurious test dependencies in a few of our packages. That seems fine to me, but that may be because I wouldn't be the one paying the cost of those dependencies. Thoughts?

build-scripts/component-common.gradle
rfk: More concerningly, I don't know how to test whether this change has broken whatever AndroidX thing it was introduced to address; our `./gradlew test` still passes and an a-c test run seems to be going OK so far. I'll try a Fenix build with locally-published components as well for completeness.

rfk:
@jhugman how do you feel about this from a Nimbus PoV?

build-scripts/component-common.gradle
rfk: > I'll try a Fenix build with locally-published components as well for completeness.

I can't build Fenix against latest a-c right now, looks like there might be a refactor in progress that needs updates in Fenix. Will try again later, but definitely want to do a full test run on Fenix before merging this.

mhammond (approved):
This looks great to me, although I suggest you still wait for a review from someone who actually knows something about gradle

jhugman (approved):
FWIW: I'm very happy to see some order brought to these files. Thank you @rfk .

rfk:
This is starting to get stale, and it really seems like an easy-to-reverse-if-we-hate-it change to me, so I'm going to rebase this and get it merged.

rfk:
(Actually, I still need to do a final test and build with Fenix before merging this, which I will do shortly)

build-scripts/component-common.gradle
rfk: It took me a bit of fiddling around to find the right combo of branches, but I have successfully build Fenix against a local publish of appservices and a-c with these changes in place, and confirmed that sync- and accounts-related functionality is working as expected

PR #4195, opened 2021-06-09
[ci full] Fix #3766: high level content signature chain of trust verification

### LINKED ISSUE #3766: Expose NSS VerifyCodeSigningCertificateChain in `rc_crypto` (requires NSS 3.66)
Periodic update.  As part of the week, need to review the existing documentation.

The rc_crypto crate is probably our scariest crate, both because it's crypto-related and because it's got some weird licensing stuff going on. It's been a while since any of us had to deal with said scariness, so it's a good time to revisit that crate with fresh eyes and see if there are any docs that need to be updated (or brought into existence).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2586)
┆Epic: a-s supporting tools (backlog)

leplatrem:
Fix #3766

  - Note:


leplatrem:
I'm stuck on this link error:
```
/usr/bin/ld: /home/mathieu/Code/Mozilla/application-services/target/debug/deps/systest-5615b05b28b8b37e.4ehwytlossh7jkgj.rcgu.o: in function `systest::fn_VerifyCodeSigningCertificateChain':
          /home/mathieu/Code/Mozilla/application-services/target/debug/build/systest-6de8ac3438657825/out/all.rs:(.text._ZN7systest36fn_VerifyCodeSigningCertificateChain17h06fc7a56f4e12328E+0xe): undefined reference to `VerifyCodeSigningCertificateChain'

```

Whereas I can see that `"-L" "/home/mathieu/Code/Mozilla/application-services/libs/desktop/linux-x86-64/nss/lib"` is passed to the compilation command, and that `nm libs/desktop/linux-x86-64/nss/lib/libmozpkix.a` shows `VerifyCodeSigningCertificateChain`.

@rfk do you have any idea?

leplatrem:
>  @rfk do you have any idea?

Dana unblocked me :) I also added a flag to `stdc++`, I don't know what it implies elsewhere in the bootstrapping area.

Basically, my next steps are:
- expose `CERT_GetDefaultCertDB()`
- expose `CERT_NewTempCertificate()` and `CERT_ExtractPublicKey()`
- use them to parse the leaf certificate, and extract its public key, to verify the content signature, using the code that I had in remote-settings-client already

components/support/rc_crypto/nss/nss_build_common/src/lib.rs
```diff
@@ -78,6 +78,8 @@ fn link_nss_libs(kind: LinkingKind) {
     for lib in libs {
         println!("cargo:rustc-link-lib={}={}", kind_str, lib);
     }
+    // Link against C++ stdlib (for mozpkix)
+    println!("cargo:rustc-link-lib=stdc++");
```
rfk, line 82: Nice find. I would like us to be using a more structured approach to the build here such as the [`cc` crate](https://github.com/alexcrichton/cc-rs) but that's well beyond the scope of this PR.

One thing to watch out for here, is apparently the name of this may change on different platforms, as described [here](https://github.com/alexcrichton/cc-rs#c-support). We'll have to do some manual testing on Android to see if any tweaks are needed.

rfk: On my Mac, I had to change this to `println!("cargo:rustc-link-lib=c++");` to get it to compile successfully. (The tests fail in something about splitting the pem, but that seems unlikely to be related to c++ linking issues).

components/support/rc_crypto/nss/src/pkixc.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    let mut out: PRErrorCode = 0;
+
+    let result = unsafe {
+        nss_sys::VerifyCodeSigningCertificateChain(
+            p_certificates.as_mut_ptr(),
+            cert_lens.as_ptr(),
+            certificates.len(),
+            seconds_since_epoch,
+            root_sha256_hash.as_ptr(),
```
rfk, line 37: Careful, you haven't checked the length of `root_sha256_hash` anywhere here. For safety, we should either check it here in the function body, or insist on it in the type of the parameter, which IIRC would be something like `root_sha256_hash: &[u8; 32]`. Which approach is better probably depends on the details of how it's being called.

(I see that you check for this in the higher-level code, but it's good practice to put these checks next to the `unsafe` code that depends on them, otherwise we're pretending that `verify_code_signing_certificate_chain` is safe when it isn't really).

leplatrem:
I'm not sure how to fix the failing tests, but at last the code is ready to be reviewed :)

Thanks :pray:

rfk:
@leplatrem thanks! I'm keen to dive in here, but also I have a week of PTO after the wellness week so I may not get a chance to look before the week of 12th July. Do you need this looked at before then?

leplatrem:
@rfk Thanks! Don't worry, more or less the same here :)

rfk:
Sorry for the ongoing delay here, I've been caught up in a few other things after turning from break. Please hassle me if I haven't commented substantially here by the end of the week.

components/support/rc_crypto/nss/src/pkixc.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use nss_sys::PRErrorCode;
+
+// NSS error codes.
+// https://searchfox.org/mozilla-central/source/security/nss/lib/util/secerr.h#29
+// https://searchfox.org/mozilla-central/source/security/nss/lib/mozpkix/include/pkix/Result.h
+const SEC_ERROR_UNKNOWN_ISSUER: i32 = -8179; // -8192 + 13
+const SEC_ERROR_EXPIRED_ISSUER_CERTIFICATE: i32 = -8162; // -8192 + 30
+const SEC_ERROR_SUBJECT_MISMATCH: i32 = -12276; // ???
+const SEC_ERROR_EXPIRED_CERTIFICATE: i32 = -16378; // ???
```
mozkeeler, line 16: https://searchfox.org/mozilla-central/rev/04b41051dd2b08df2a06f86f8bf9230c0848fc51/security/nss/lib/mozpkix/include/pkix/pkixnss.h#81

components/support/rc_crypto/nss/src/pkixc.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::util::ensure_nss_initialized;
+
+use nss_sys::PRErrorCode;
+
+// NSS error codes.
+// https://searchfox.org/mozilla-central/source/security/nss/lib/util/secerr.h#29
+// https://searchfox.org/mozilla-central/source/security/nss/lib/mozpkix/include/pkix/Result.h
+const SEC_ERROR_UNKNOWN_ISSUER: i32 = -8179; // -8192 + 13
+const SEC_ERROR_EXPIRED_ISSUER_CERTIFICATE: i32 = -8162; // -8192 + 30
+const SEC_ERROR_SUBJECT_MISMATCH: i32 = -12276; // ???
```
mozkeeler, line 15: https://searchfox.org/mozilla-central/rev/04b41051dd2b08df2a06f86f8bf9230c0848fc51/security/nss/lib/ssl/sslerr.h#42

rfk:
@leplatrem I think rebasing should fix the failing "rust dependencies" test. I'll see if I can reproduce the carthage build failures locally on my mac.

components/support/rc_crypto/nss/nss_sys/src/bindings/pkixc.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    // bool VerifyCodeSigningCertificateChain(
+    //     const uint8_t** certificates,
+    //     const uint16_t* certificateLengths,
+    //     size_t numCertificates,
+    //     uint64_t secondsSinceEpoch,
+    //     const uint8_t* rootSHA256Hash,
+    //     const uint8_t* hostname,
+    //     size_t hostnameLength,
+    //     /* out */ PRErrorCode* error);
```
rfk, line 27: nit: please move this to above the function that it documents, for consistency with the other file.

components/support/rc_crypto/nss/src/cert.rs
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::*;
+use crate::util::{ensure_nss_initialized, sec_item_as_slice};
+// use crate::error::ErrorKind;
```
rfk, line 7: nit: remove this if unneeded

components/support/rc_crypto/nss/src/cert.rs
```diff
@@ -0,0 +1,39 @@
...
+            nss_sys::PR_FALSE,
+            nss_sys::PR_TRUE,
+        )
+    };
+
+    let pub_key = unsafe { CERT_ExtractPublicKey(cert) };
+    let mut pub_key_data = unsafe { (*pub_key).u.ec.publicValue };
+    let pub_key_data_raw = unsafe { sec_item_as_slice(&mut pub_key_data)? };
+
+    // TODO: Free resources!
```
rfk, line 36: Yes please :-)

components/support/rc_crypto/nss/src/pkixc.rs
```diff
@@ -0,0 +1,73 @@
...
+    certificates: Vec<&[u8]>,
+    seconds_since_epoch: u64,
+    root_sha256_hash: &[u8],
+    hostname: &str,
+) -> Result<()> {
+    ensure_nss_initialized();
+
+    let mut cert_lens: Vec<u16> = vec![];
+    for certificate in &certificates {
+        cert_lens.push(certificate.len() as u16);
```
rfk, line 30: Is it worth erroring out here if the length is too large to fit in a `u16`? IIUC Rust will just truncate the bits when casting to a smaller integer size, which will always produce a smaller length, which will not risk memory unsafety...so I guess it's fine? May be worth a comment to that effect at least.

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
rfk, line 4: Please include some brief high-level module docs here, even if just some links out to further reading. I fear future maintainers of this crate may not have all the context to remember what this is for.

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+use std::str;
+
+use crate::error::*;
+use crate::signature;
+
+fn decode_root_hash(input: &str) -> Result<Vec<u8>> {
+    let bytes_hex = input.split(':');
```
rfk, line 11: An example of the sort of string we're trying to pass here would be useful to help understanding, I am interpreting it as something like "ab:12:ef:...".

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    let mut result: Vec<u8> = vec![];
+    for byte_hex in bytes_hex {
+        let byte = match hex::decode(&byte_hex) {
+            Ok(v) => v,
+            Err(_) => return Err(ErrorKind::RootHashFormatError(input.to_string()).into()),
+        };
+        result.extend(byte);
+    }
```
rfk, line 20: If you wanted to mess around with Rust iterators, this would be a good opportunity to use `collect`. It's possible to [`collect` into a `Result<Vec<T>>`](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect) and get the early-termination-on-error thing for free, something along the lines of:

```
fn decode_root_hash(input: &str) -> Result<Vec<u8>> {
    input.split(':').map(|byte_hex| {
        hex::decode(&byte_hex).map_err(|_| {
            ErrorKind::RootHashFormatError(input.to_string()).into()
        })
    }).collect()
}
```

(The more explicit version is also fine if you *don't* want to mess around with Rust iterators).

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    Ok(result)
+}
+
+fn split_pem(pem_content: &[u8]) -> Result<Vec<Vec<u8>>> {
```
rfk, line 25: Please add docs here to help understand the `Vec<Vec<u8>>` situation

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+    input: &[u8],
+    signature: &[u8],
+    pem_bytes: &[u8],
+    seconds_since_epoch: u64,
+    root_sha256_hash: &str,
+    hostname: &str,
+) -> Result<()> {
+    let certificates = split_pem(&pem_bytes)?;
+
+    let mut certificates_slices: Vec<&[u8]> = vec![];
```
rfk, line 73: You might be able to use something like `let mut certificate_slices = certificates.iter().collect();` here for succinctness, letting Rust infer the type based on its future use as a parameter to `verify_code_signing_certificate_chain`.

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+        Ok(bytes) => bytes,
+        Err(err) => return Err(ErrorKind::CertificateContentError(err.to_string()).into()),
+    };
+
+    let signature_bytes = match base64::decode_config(&signature, base64::URL_SAFE) {
+        Ok(b) => b,
+        Err(err) => return Err(ErrorKind::SignatureContentError(err.to_string()).into()),
+    };
+
+    // Since signature is NIST384p / secp384r1, we can perform a few safety checks.
```
rfk, line 105: Out of curiosity, what would happen if we *didn't* do these checks?

components/support/rc_crypto/src/contentsignature.rs
```diff
@@ -0,0 +1,305 @@
...
+/f7/
+-----END CERTIFICATE-----",
+        )
+        .unwrap();
+        assert_eq!(result, vec![vec![1, 2, 3, 4, 5, 6], vec![253, 254, 255]]);
+    }
+
+    #[test]
+    fn test_verify() {
+        assert!(verify(
```
rfk, line 283: It would be great to give each of these tests a comment to indicate the specific reason why the verification fails, e.g. "Fails with invalid hostname", "Fails with expired signature", etc. Also it would be great to have tests for a few different ways it can fail - not to check that NSS is doing what we expect, but to ensure that we're not e.g. accidentally always passing an empty string or a zero timestamp or something.

rfk (commented):
This looks pretty good to me, although fiddling around in the low-level NSS bits always makes me nervous. Next steps here:

* I have the tests passing locally on my mac; I need to try to reproduce the CI *failures* and help you debug from there.
* I left a couple of comments of places where we could use some of Rust's more advanced features to make the code more succinct, but up to you if you want to take them.
* I think we need to add a few more tests and some docs before final merge.

I'm pretty excited to see this coming together!

rfk:
> I have the tests passing locally on my mac; I need to try to reproduce the CI failures and help you debug from there.

A quick google suggests this is *probably* to do with linking against the C++ stdlib from within an XCode build, e.g. https://github.com/auth0/react-native-lock/issues/88. So hopefully it won't be too thorny to get to the bottom of...

components/support/rc_crypto/src/contentsignature.rs
leplatrem: I think I've tried and got out of compilation errors with the long version...

```rust
79 |     let mut certificates_slices = certificates.iter().collect();
   |                                                       ^^^^^^^ value of type `Vec<&[u8]>` cannot be built from `std::iter::Iterator<Item=&Vec<u8>>`
```

components/support/rc_crypto/src/contentsignature.rs
leplatrem: I could not figure out how to flatten items and collect results, so I ended up writing an explicit loop :man_shrugging:

leplatrem:
Thanks @rfk @mozkeeler for your review and help! Tests now pass :tada: 

I added a few commits, I'd be happy to rebase/reorganize them once reviewed. 

Please let me know if you expect more test cases for the verification. If we want to recreate invalid certificates chains for different fined-grained reasons, I may have to look for tools or solutions to alter existing valid certs...
I also added 2 new kinds of errors (SSL and PKIX) that would help find out the original NSS error from the integer code more easily (IMO).

components/support/rc_crypto/src/contentsignature.rs
leplatrem: I added a comment below, but basically the signature verification would fail for a number of different reasons. Doing these checks here help us being more explicit about the reason of failure.

mozkeeler (changes requested):
This looks really good. I mostly looked at the new NSS parts and noted a few changes that should be made.

components/support/rc_crypto/nss/src/cert.rs
```diff
@@ -0,0 +1,43 @@
...
+        )
+    };
+
+    let pub_key = unsafe { CERT_ExtractPublicKey(cert) };
+    let mut pub_key_data = unsafe { (*pub_key).u.ec.publicValue };
+    let pub_key_data_raw = unsafe { sec_item_as_slice(&mut pub_key_data)? };
+
+    unsafe {
+        SECKEY_DestroyPublicKey(pub_key);
+        CERT_DestroyCertificate(cert);
```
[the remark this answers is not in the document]
rfk: Good call, I believe we have a `ScopedPtr` wrapper struct that is used to help with this in other functions. See `src/pk11/types.rs` for where it's used to define RAII wrappers on some other pointer types.

In fact, at least `SECKEYPublicKey` seems to already have a `ScopedPtr` wrapper struct. You may be able to achieve both the null-check and the automatic cleanup by doing something like:

```
use crate::pk11::types::PublicKey;
[...]
let pub_key = unsafe { PublicKey::from_ptr(CERT_ExtractPublicKey(cert)) }
```

The `CERTCertificate` types will need to be added to the list in `types.rs` to achieve a similar affect.

components/support/rc_crypto/src/contentsignature.rs
rfk: ah well, it was worth a try :-)

rfk (changes requested):
The higher-level `rc_crypto` bits here look good to me, thanks!

@mozkeeler's suggestion of an RAII guard on the low-level NSS types is spot on and hopefully fairly straightforward using the `scoped_ptr!` macro that I mention below, and I think it's worth me taking another look at this once that's in place, just for completeness.

components/support/rc_crypto/nss/src/pkixc.rs
```diff
@@ -0,0 +1,103 @@
...
+            ROOT_HASH_LENGTH
+        ))
+        .into());
+    }
+
+    let mut out: PRErrorCode = 0;
+
+    let result = unsafe {
+        nss_sys::VerifyCodeSigningCertificateChain(
+            p_certificates.as_mut_ptr(),
```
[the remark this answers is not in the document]
leplatrem: I'm not really aware of the implications of this here, since Rust input parameters are not mutable anyway. But since it's a mistake on the exposed API I'll add a comment

rfk (approved):
Nice, the scoped-ptr stuff looks 👍🏻

rfk:
I think this is good to go when you're happy with it @leplatrem. Please consider a bit of a rebase and history cleanup, then add the `checkin-needed` label when it's ready to land. (I'm not sure if the bot will successfully do its thing in that case, but it will at the very least signal to us that it's ready for merge).

PR #4198, opened 2021-06-10
Don't use the codecov bash uploader

bendk:

  - Note:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,7 +258,7 @@ jobs:
           command: grcov ./target/debug/ -s . -t lcov --llvm --ignore-not-existing --ignore "target/*" --ignore "/*" -o lcov.info
       - run:
           name: Upload to codecov.io
-          command: bash <(curl -s https://codecov.io/bash) -f lcov.info
+          command: ./automation/upload-to-codecov.sh
   Check Protobuf files are up-to-date:
     executor: docker
     resource_class: small
--- a/automation/upload-to-codecov.sh
+++ b/automation/upload-to-codecov.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+set -ex
+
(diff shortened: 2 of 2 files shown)


bendk:
Interestingly, I just got an email from codecov introducing their new uploader: https://about.codecov.io/blog/introducing-codecovs-new-uploader/.

We could use that, but it seems simple enough to use this bash script.  I guess it comes down to how much you trust codecov at this point.

mhammond (commented):
This seems extremely simple - the existing bash loader seems to also run `sed` over the reports and does various other things, like checking the coverage file actually exists, etc - are we sure they are not necessary?

It also seems this PR apparently failed to generate coverage info, which seems odd.

I also got their security email and it had [a link](https://about.codecov.io/blog/validating-the-bash-script-on-ci) to a better way to validate the bash script - which still isn't ideal. I think other teams here also just vendored the bash script in, which is safe as long as the script is audited every time it is updated, which seems unlikely tbh :)

But I've no objections to running with this if it works, but I'd like to understand why there's no coverage for this PR before hitting the "approve" button.

mhammond:
> I'd like to understand why there's no coverage for this PR before hitting the "approve" button.

hrm - I see that for some other PRs too :(

bendk:
That uploader script does so much, it's really hard to understand it.  My general approach was to extract what seemed like the core functionality and if the reports worked then assume the rest wasn't needed.  I see a bunch of sed/awk code around line 1556, but that seems to only run for `.gcov` files.

I'm not exactly sure what happened with the coverage diff.  The commit that it's trying to compare to seems correct though.  When I go to the commit list (https://app.codecov.io/gh/mozilla/application-services/commits) I can see the commit (`367280a`) and it's just listed as processing.  Seems like that's a separate, hopefully temporary, issue.

mhammond (approved):
SGTM and it's easy to change our mind later.

PR #4201, opened 2021-06-11
Added an MADR document for database migration errors

bendk:
Hopefully this captures the discussion we had on Tuesday.

  - Note:

### DIFF
--- a/docs/adr/0002-database-corruption.md
+++ b/docs/adr/0002-database-corruption.md
@@ -1,0 +1,108 @@
+# Handling Database Corruption
+
+* Status: accepted
+* Date: 2021-06-08
+
+## Context and Problem Statement
+
+Some of our users have corrupt SQLite databases and this makes the related
+component unusable.  The best way to deal with corrupt databases is to simply
+delete the database and start fresh (#2628).  However, we only want to do this
+for persistent errors, not transient errors like programming logic errors, disk
+full, etc.  This ADR deals with 2 related questions:
+
+  * A) When and how do we identify corrupted databases?
+  * B) What do we do when we identify corrupted databases?
+
+## Decision Drivers
+
+* Deleting valid user data should be avoided at almost any cost
+* Keeping a corrupted database around is almost as bad.  It currently prevents
+  the component from working at all.
+* We don't currently have a good way to distinguish between persistent and
+  transient errors, but this can be improved by reviewing telemetry and sentry
+  data.
+
+## Considered Options
+
+* A) When and how do we identify corrupted databases?
+  * 1: Assume all errors when opening a database are from corrupt databases
+  * 2: Check for errors when opening a database and compare against known corruption error types
+  * 3: Check for errors for all database operations and compare against known corruption error types
+* B) What do we do when we identify corrupted databases?
+  * 1: Delete the database file and recreate the database
+  * 2: Move the database file and recreate the database
+  * 3: Have the component fail
+
+## Decision Outcome
+
+* A2: Check for errors when opening a database and compare against known corruption error types
+* B1: Delete the database file and recreate the database
+
+Decision B follows from the choice of A.  Since we're being conservative in
+identifying errors, we can delete the database file with relative confidence.
+
+"Check for errors for all database operations and compare against known
+corruption error types" also seems like a reasonable solution that we may
+pursue in the future, but we decided to wait for now.  Checking for errors
+during opening time is the simpler solution to implement and should fix the
+issue in many cases.  The plan is to implement that first, then monitor
+sentry/telemetry to decide what to do next.
+
+# Pros and Cons of the Options
+
+### A1: Assume all errors when opening a database are from corrupt databases
+* Good, because the sentry data indicates that many errors happen during opening time
+* Good, because migrations are especially likely to trigger corruption errors
+* Good, because it's a natural time to delete the database -- the consumer code
+  hasn't run any queries yet and doesn't have any open connections.
+* Bad, because it will delete valid user data in several situations that are
+  relatively common: migration logic errors, OOM errors, Disk full.
+
+### A2: Check for errors when opening a database and compare against known corruption error types (Decided)
+* Good, because should eliminate the possibility of deleting valid user data.
+* Good, because the sentry data indicates that many errors happen during opening time
+* Good, because it's a natural time to delete the database -- the consumer code
+  hasn't run any queries yet and doesn't have any open connections.
+* Bad, because we don't currently have a good list corruption errors
+
+### A3: Check for errors for all database operations and compare against known corruption error types
+* Good, because the sentry data indicates that many errors happen outside of opening time
+* Good, because should eliminate the possibility of deleting valid user data.
+* Bad, because the consumer code probably doesn't expect the database to be
+  deleted and recreated in the middle of a query.  However, this is just an
+  extreme case of normal database behavior -- for example any given row can be
+  deleted during a sync.
+* Bad, because we don't currently have a good list corruption errors
+
+### B1: Delete the database file and recreate the database (Decided)
+* Good, because it would allow users with corrupted databases to use the
+  affected components again
+* Bad, because any misidentification will lead to data loss.
+
+### B2: Move the database file and recreate the database
+
+This option would be similar to 1, but instead of deleting the file we would
+move it to a backup location.  When we started up, we could look for backup
+files and try to import lost data.
+
+* Good, because if we misidentify corrupt databases, then we have the
+  possibility of recovering the data
+* Good, because it allows a way for users to delete their data (in theory).
+  If the consumer code executed a `wipe()` on the database, we could also
+  delete any backup data.
+* Bad, because it's very difficult to write a recovery function that merged
+  deleted data with any new data.  This function would be fairly hard to test
+  and it would be easy to introduce a new logic error.
+* Bad, because it adds significant complexity to the database opening code
+* Bad, because the user experience would be strange.  A user would open the
+  app, discover that their data was gone, then sometime later discover that the
+  data is back again.
+
+### B3: Return a failure code
+
+* Good, because this option leaves no chance of user data being deleted
+* Good, because it's the simplest to implement
+* Bad, because the component will not be usable if the database is corrupt
+* Bad, because the user's data is potentially exposed in the corrupted database
+  file and we don't provide any way for them to delete it.


docs/0002-database-migration-errors.md
```diff
@@ -0,0 +1,81 @@
...
+* Deleting valid user data should be avoided at almost any cost
+* Keeping a corrupted database around is almost as bad.  It currently prevents
+  the component from working at all.
+* There's no failproof way to distinguish the two error cases.
+
+## Considered Options
+
+* Delete the database file and recreate the database
+* Move the database file and recreate the database
+* Return a failure code
```
bendk, line 26: Two things that I didn't include:
  - Looking at the sentry errors and telemetry to see how many users are impacted by corrupt databases.
  - Trying to develop a system for detecting corrupt databases during normal operations

We discussed both, although it didn't feel like completely committed to either.  I'm also not sure if that would be part of this ADR or a separate one, so I just left it out.

docs/0002-database-migration-errors.md
```diff
@@ -0,0 +1,81 @@
...
+issue #2628), however migration logic errors should be handled by returning an
+error code and releasing fixed code.  What should we do if we see an error
+during the migration process?
+
+## Decision Drivers
+
+* Deleting valid user data should be avoided at almost any cost
+* Keeping a corrupted database around is almost as bad.  It currently prevents
+  the component from working at all.
+* There's no failproof way to distinguish the two error cases.
```
mhammond, line 20: While this is true, I think there are *some* cases that can be distinguished - eg, `SQLITE_CORRUPT`.

mhammond (commented):
Thanks Ben, it's very useful for us to get some key decisions documented.

However, IMO, a problem with this ADR is what it doesn't say - it basically says "we will not take action on corrupt databases" - which I think we agreed is true for the specific case of migration, although not true in general (well, it's true *today* in general, but we need to make it not true :)

I guess I see 2 paths forward:
* Keep this ADR specific to migrations, but add text along the lines of "we recognize the need to handle corrupt databases, but that's a general problem and not specific to migrations"
* Turn this ADR into a generic "handling database corruption" - and in that document migrations would just get a side-note saying there's no need to have migrations treat corrupt databases differently than the rest of the world.

My preference would be for the second (because just doing the first begs the question of where the general strategy is documented), but then the document needs to change fairly drastically - it will then need to explicitly state our strategy for - which I would summarize as "delete databases only when an error is certain to mean the DB is corrupt, and monitor telemetry/sentry/whatever"

bendk:
> My preference would be for the second (because just doing the first begs the question of where the general strategy is documented), but then the document needs to change fairly drastically - it will then need to explicitly state our strategy for - which I would summarize as "delete databases only when an error is certain to mean the DB is corrupt, and monitor telemetry/sentry/whatever"

That's my preference too, but I feel like there are some questions that need to be answered first:

  - How would database deletion work if we're catch the error during regular database operations?  At this point the consumer code may have already been using the DB.  Is it safe to delete the db and start over?
  - How do we know which errors are definitely from SQLite corruption?
  - What's the plan for implementing the new strategy?

I think we touched on some of these during the conversation but it didn't feel like we made a decision.  Maybe I'll try to write them into the ADR and we can discuss from there.

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+* Good, because it's a natural time to delete the database -- since the user hasn't run any queries yet.
+* Bad, migrations happen infrequently.  Errors that manifest during migrations
+  were probably also causing issues before.
+
+### A: Check for corruption whenever we perform a database operation (Decided)
+* Good, because this gives us the opportunity to fix corruption errors sooner.
+* Bad, because the consumer code probably doesn't expect the database to be
+  deleted and recreated in the middle of a query.  However, this is just an
+  extreme case of normal database behavior -- for example any given row can be
+  deleted during a sync.
```
rfk, line 84: In the case of a corrupted database, I presume the caller was already going to receive *some sort of error* here anyway, so perhaps it's not really that bad for them in practice. We're making a problem that they already had more obvious, but not introducing a new problem.

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+* Bad, because logic errors are most likely during migrations.
+
+### B: Compare error values against known corruption errors for all operations
+* Good, because should eliminate the possibility of deleting valid user data.
+* Bad, because we don't currently have a good list of these.
+* Bad, because it's tricky to try to repair a database that the consumer code
+  is currently using.
+
+### B: Compare error values against known corruption errors when opening the database (Decided)
+* Good, because should eliminate the possibility of deleting valid user data.
```
rfk, line 97: Not sure I understand what you mean here - is the premise that checking during db open is likely to avoid false-positives, where we think the issue was db corruption but actually it was some logic error causing a bad query?

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+* Good, because from what we've seen on sentry it seems like this is a very common case.
+* Good, because inside the open() method is a very convenient time to try to repair the DB.
+* Bad, because we don't currently have a good list of these.
+
+### C: Delete the database file and recreate the database (Decided)
+* Good, because it would allow user's with corrupted databases to use the
```
rfk, line 103: nit: no apostrophe in "users".

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+  deleted data with any new data.  This function would be fairly hard to test
+  and it would be easy to introduce a new logic error.
+* Bad, because it adds significant complexity to the migration code
+* Bad, because the user experience would be strange.  A user would open the
+  app, discover that their data was gone, then sometime later discover that the
+  data is back again.
+
+### C: Return a failure code
+
+* Good, because this option leaves no chance of of user data being deleted
```
rfk, line 128: nit: "of of"

docs/adr/0002-database-corruption.md
rfk: Actually, I'm not sure I fully understand what's being proposed here in the combination of decided options for (A) and (B). The decided option here for (A) says we will check for corruption whenever we perform a database operation, but the decided option for (B) says we will only do it when opening the database.

Are we checking on every operation, or only on every open?

rfk:
(FWIW, I do like the more general framing as a "how to deal with corruption" document rather than as being specific to migrations).

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+* Status: accepted
+* Date: 2021-06-08
+
+## Context and Problem Statement
+
+Some of our users have corrupt SQLite databases and this makes the related
+component unusable.  The best way to deal with corrupt databases is to simply
+delete the database and start fresh (see #2628).  However, it's not so easy
+to know if a database operation error is from a corrupt database or from bad
+code.  We want a system that identifies and deals with database corruption but
```
mhammond, line 12: There are a few places below where you talk about "both" cases, so I'd like to suggest reframing this a little here - make the distinction between "permanent" and "transient" errors, and explicitly say we are considering issues in the code as transient because we can push a fix?

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+* Keeping a corrupted database around is almost as bad.  It currently prevents
+  the component from working at all.
+* We don't currently have a good way to distinguish between the two error
+  cases, but this can be improved by reviewing telemetry and sentry data.
+
+## Considered Options
+
+* A) When do we identify corrupted databases?
+  * At migration time only
+  * Whenever we perform a database operation
```
mhammond, line 32: I think A) and B) here should be merged as they contradicts a little - eg "when opening" seems to be missing for A). "When and how do we...?"

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+  * Have the component fail
+
+## Decision Outcome
+
+* A: Whenever we perform a database operation
+
+This was chosen because it can be a long time between migrations and we want
+to de able to fix corrupt database during that period.
+
+* B: Compare error values against known corruption errors when opening the database
```
mhammond, line 49: Eg, A and B are contradictory here - A) says we will identify whenever we perform a database operation and B) says only when opening

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+only target opening the database, which is the easiest time to fix issues.
+
+* C: Delete the database file and recreate the database
+
+This follows the choice of B.  Since we're being conservative in identifying
+errors, we can delete the database file with relative confidence.
+
+### Implementation notes:
+
+This can be implemented by updating the `Store` code to wrap each database
```
mhammond, line 63: and here we are back to "each database operation". I think this is going to be difficult in practice as (a) the code detecting the error doesn't have access to all connections to close and (b) there's a risk apps will behave badly in that scenario - eg, the strategy here might be to (say) somehow flag the detection and delete at next open?

However, given this should be long-lived, I don't think we are actually making the decision that we will never check on each operation - it's closer to "we want to be cautious, so will *initially* start with just on open, but will improve this as we learn.

Or to put it another way, if we do end up implementing the "every operation" strategy later, this adr should remain accurate without changes.

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+operation with a check for errors that we can positively identify as
+corruption-related.  If we see one of those errors, then we can delete the
+database file and start from scratch.
+
+Also, we will start tracking Sentry errors and telemetry to identify errors
+that we know are from SQLite corruption.
```
mhammond, line 69: maybe add something like: "telemetry might also be able to help us determine which errors are permanent and which later resolve themselves"? Going a bit off-topic, but I'm particularly thinking about obscure ones like "Device IO Error".

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Pros and Cons of the Options
+
+### A: Check for corruption at migration time only
```
mhammond, line 73: I think s/migration/open/ here - and probably other places in the doc. (I think that's fine because in practice, a migration is going to be the very first thing after opening)

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,132 @@
...
+  app, discover that their data was gone, then sometime later discover that the
+  data is back again.
+
+### C: Return a failure code
+
+* Good, because this option leaves no chance of of user data being deleted
+* Good, because it's the simplest to implement
+* Bad, because the component will not be usable if the database is corrupt
+* Bad, because the user's data is potentially exposed in the corrupted database
+  file and we don't provide any way for them to delete it.
```
mhammond, line 132: I think this last point was a hangover from the migration focus and can probably be deleted - in the context of this document, it is the status-quo.

mhammond (commented):
This is looking great and heading in the right direction - it just needs a bit of cleanup I think.

(The slightly bad-news from sentry is that the error code on open *is not* the top crasher, so we probably will need to move to the "any operation" option, but I'm still 100% on-board with the "initially just on open, later we will see" approach)

docs/adr/0002-database-corruption.md
bendk: Oops, I meant to delete this one with my last edit.

I agree that there's definitely some risk, but the more I think about it the more it seems manageable.  For (a) we would make it so only the `Store` has access to the raw connection and the other components access the DB through the store.  Then the store should be free to drop the connection and create a new one.  For (b) it feels like apps will behave badly if you drop all the data on them, but I'm struggling to come up with something concrete.  For example: I fetch a set of logins, do something with them, then run another query later on.  Even today, I can't assume that any of the logins are still in the database -- they may have been deleted by a sync.  Also, the current process is to return an error, which seems about as likely to me to cause issues.

In any case, I definitely agree with the decision to initially target with opening the DB, learn more and then move to other operations.

docs/adr/0002-database-corruption.md
bendk: That represents the fact that I was going back and forth between the two options, missed some edits, and the doc ended up with some confusion.  I think the decision was to start by only checking at open time, see how that affects the sentry errors/telemetry, and then make a decision about how to tackle the rest.

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,106 @@
...
+issue in many cases.  The plan is to implement that first, then monitor
+sentry/telemetry to decide what to do next.
+
+# Pros and Cons of the Options
+
+### A1: Assume all errors when opening a database are from corrupt databases
+* Good, because the sentry data indicates that many errors happen during opening time
+* Good, because migrations are especially likely to trigger corruption errors
+* Good, because it's a natural time to delete the database -- the consumer code hasn't run any queries yet.
+* Bad, because creates the possibility of deleting valid user data.
```
mhammond, line 58: I'd even upgrade this to "likely to delete valid user data" - eg, I could imagine OOM etc causing this too.

docs/adr/0002-database-corruption.md
```diff
@@ -0,0 +1,106 @@
...
+* Bad, because migrations likely sources of logic errors.
+
+### A2: Check for errors when opening a database and compare against known corruption error types (Decided)
+* Good, because should eliminate the possibility of deleting valid user data.
+* Good, because the sentry data indicates that many errors happen during opening time
+* Good, because it's a natural time to delete the database -- the consumer code hasn't run any queries yet.
```
mhammond, line 64: add "nor have outstanding connections"?

mhammond (approved):
This looks great, thanks!

mhammond:
welp, I also just rediscovered [this doc](https://github.com/mozilla/application-services/blob/main/docs/design/db-fatal-error-handling.md) I wrote last year (which I had forgotten about :( There's not a huge amount of overlap with this adr, but it might be useful the the actual implementation of this ADR and steps to help better classify them in sentry.

PR #4202, opened 2021-06-14
Update UniFFI to v0.12.0 [ci full]

rfk:
The latest release of UniFFI includes a significant change to the
way Object instances are managed, switching from handlemaps to
raw Arc<T> pointers. We should get this into consumers for testing
at some point early in a Nightly cycle.


mhammond (approved):

rfk:
The `module-build-nimbus` failure here looks like some sort of spurious network issue.

PR #4207, opened 2021-06-15
Allow getDBSaltForKey error to propagate for fxiOS

skhamis:
The way iOS uses getDBSaltForKey, they have a do/catch wrapper if there is no salt and generate a new one [Here](https://github.com/mozilla-mobile/firefox-ios/blob/041c85483b1aa4a4e357d8bd9419550e4f20347a/Storage/Rust/RustLogins.swift#L108). The uniffi logins effort added a `try!` which hard fails and causes error propagation to be disabled.

I have also updated the iOS testing on the a-s side to better mimic what fxiOS is doing and testy the APIs the way they are using it.

  - Note:

### DIFF
--- a/megazords/ios/MozillaAppServicesTests/LoginsTests.swift
+++ b/megazords/ios/MozillaAppServicesTests/LoginsTests.swift
@@ -10,9 +10,8 @@ class LoginsTests: XCTestCase {
     var encryptionKey: String!
     var salt: String!
 
-    //This tests setup mimics closer to what fxiOS is doing
+    // This test setup mimics as close as we can to how fxiOS consumes our API
     override func setUp() {
-
         let directory = NSTemporaryDirectory()
         let filename = "testdb-\(UUID().uuidString).db"
         let fileURL = URL(fileURLWithPath: directory).appendingPathComponent(filename)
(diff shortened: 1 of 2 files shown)


megazords/ios/MozillaAppServicesTests/LoginsTests.swift
```diff
@@ -6,36 +6,60 @@ import XCTest
 @testable import MozillaAppServices
 
 class LoginsTests: XCTestCase {
-    func getTestStorage() -> LoginsStorage {
+    var storage: LoginsStorage!
+    var encryptionKey: String!
+    var salt: String!
+
+    //This tests setup mimics closer to what fxiOS is doing
```
rfk, line 13: nit: "closer" is good context for reviewing the PR, but in a few months time someone reading this is just going to think "closer than what?"; something like "as closely as we can" or similar might be better for a long-lived comment.

rfk (approved):
Makes sense to me 👍🏻

rfk:
@skhamis I was about to cut a release for unrelated reasons, do you want to get this fix into a new release as well? (ref https://github.com/mozilla/application-services/pull/4206)

skhamis:
> @skhamis I was about to cut a release for unrelated reasons, do you want to get this fix into a new release as well? (ref #4206)

@rfk Actually that'd be awesome! let me make that quick nit and swiftlint error and get this merged in ASAP

PR #4211, opened 2021-06-15
Ensure JNA always encodes strings as UTF-8.

### LINKED ISSUE #4209: PlacesHistoryStorageTest.kt test failures depending on locale
This is a test failure I found when running the android-components test suite, but I suspect it may be a bug in the underlying appservices layers somewhere.

The android-components `PlacesHistoryStorageTest.kt` suite has some tests that check page titles in non-English locales, such as [this one](https://github.com/mozilla-mobile/android-components/blob/d7b1b62da50cf224fce1eec7a163d858fb1a2334/components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt#L302). I have regularly observed intermittent test failures here when trying to smoketest a-c against a new a-s release, and finally decided to figure out why.

It appears to be due to my `$LANG` environment variable 😬 

If I run them like this, which is my default `$LANG` when using the native terminal on my Mac:

```
LANG="en_AU.UTF-8" ./gradlew browser-storage-sync:testDebugUnitTest
```

Then everything works fine and all tests pass. If I instead run them like this, which is my default `$LANG` when sshing in to my Mac from my Windows machine:

```
LANG="C.UTF-8" ./gradlew browser-storage-sync:testDebugUnitTest
```

Then this test fails:

```
  TEST: store can be used to track page meta information - title changes
    [Robolectric] mozilla.components.browser.storage.sync.PlacesHistoryStorageTest.store can be used to track page meta information - title changes: sdk=28; resources=binary
  FAILURE
org.junit.ComparisonFailure: expected:<[?????????]> but was:<[?????????]>
    at org.junit.Assert.assertEquals(Assert.java:115)
    at org.junit.Assert.assertEquals(Assert.java:144)
    at mozilla.components.browser.storage.sync.PlacesHistoryStorageTest$store can be used to track page meta information - title changes$1.invokeSuspend(PlacesHistoryStorageTest.kt:302)
```

I'm lightly alarmed that setting `$LANG` can affect the tests like this, and it makes me wonder if we may have some bug in the handoff of non-ascii strings between Kotlin and Rust.

Does this reproduce for anyone else?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2251)
┆Story Points: 3
┆Sprint: SYNC - 2021-06-25

rfk:
JNA defaults to encoding strings in the system default encoding,
which is usually UTF-8 but not always. Rust *always* expects strings
encoded in UTF-8, so this commit adds logic to enforce that encoding
when we detect that JNA won't do it for us for free.
    
I was able to test this locally by setting `LANG=C` in my environment
and then running the Places Kotlin tests. Prior to this commit they
fail on some Unicode test data; with this commit they succeed.
    
Luckily for us in practice, Android devices *always* use UTF-8 as
the default encoding, so this issue won't have caused any problems
for production users. Still, we shouldn't depend on that in the general
case.

Fixes #4209.

### DIFF
--- a/components/support/android/src/main/java/mozilla/appservices/support/native/Helpers.kt
+++ b/components/support/android/src/main/java/mozilla/appservices/support/native/Helpers.kt
@@ -11,8 +11,14 @@ package mozilla.appservices.support.native
 import android.util.Log
 import com.google.protobuf.CodedOutputStream
 import com.google.protobuf.MessageLite
+import com.sun.jna.DefaultTypeMapper
+import com.sun.jna.FromNativeContext
 import com.sun.jna.Library
+import com.sun.jna.Memory
 import com.sun.jna.Native
+import com.sun.jna.Pointer
+import com.sun.jna.ToNativeContext
+import com.sun.jna.TypeConverter
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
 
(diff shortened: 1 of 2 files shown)


rfk:
(This will stop being a problem if we move everything to UniFFI, because it does explicit UTF-8 encoding rather than relying on JNA's default behaviour)

rfk:
Ref #4220, I have put `checkin-needed` on this PR but have deliberately *not* rebased it atop latest main. I want to confirm the behaviour of mergify when all the other checks pass but the branch is not up to date.

mhammond (approved):
Far. Out. I was not expecting this to be as complicated - but it LGTM!

rfk:
> I want to confirm the behaviour of mergify when all the other checks pass but the branch is not up to date.

Capture

Good bot!

rfk:
@Mergifyio refresh

rfk:
Hrm, when I updated the branch, mergify seems to have added a cancelled check for "the rule doesn't match anymore" and that cancelled check is causing mergifying to think that there are failed status checks on the PR, so not it won't merge it. I'm going to merge this one by hand, but it looks like we'll need to do some more config tweaking here :-/

PR #4213, opened 2021-06-15
Allow multiple features per experiment (without DTO change)

jhugman:
Fixes [SDK-298](https://jira.mozilla.com/browse/SDK-298).

This PR performs quite a bit of cleanup from the Feature API work.

 * Remove `feature_id` from enrollments, to break the 1-1 link between experiment and enrollment.
 * Changes `get_feature_config()` to `get_enrolled_features()`, and to return a *list* of features configs, not just one.
 * Change `evolve_enrollments()` to iterate over a list of feature configs per experiment, instead of one (i.e. remove `get_first_feature_id()`))
 * All access to the feature in `Branch` is now done via `get_feature_configs()`.
 * Derive the `feature_ids` from the `Branch`es themselves, not from the top level property of `Experiment`. This can be removed from the DTO if needed.
 * Fixed up tests to make sure the `Branch` feature_configs are correctly formed.
 * Added new tests to show multifeature experiment collisions that prevent enrollment.

Without further changes to the DTO (i.e. migrating from a single `feature` to a list of `features`), this is allows for each branch having a different feature.

The feature that doesn't get specified by a given branch is given a default (i.e. enrollment exclusion happens, but a default is given to the application).

In order to take advantage of this:

1. Experimenter needs to be able unlock the branches' feature id.
2. Change the DTO to allow `features` instead of `feature`.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -25,5 +25,9 @@ Use the template below to make assigning a version number during the release cut
     Users on very old versions of this component will no longer be able to cleanly update to this version. Instead, the consumer code
     will recieve an error indicating that the schema was not correctly formated.
 
+## Nimbus
+### What's Changed
+  - Nimbus SDK now supports different branches having different Feature Configs ([#4213](https://github.com/mozilla/application-services/pull/4213))
+
 ## Other
   - `./libs/build-all.sh` now displays a more helpful error message when a file fails checksum integrity test.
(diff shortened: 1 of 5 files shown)


travis79 (approved):
This is looking good to me, I do have one question: Would it be beneficial to add the whole `Branch` to the `EnrollmentStatus` rather than just the slug?  I'm thinking of reducing the need to perform the lookup, but that's not a blocker for this, just thinking of keeping things simple.

jhugman:
> Would it be beneficial to add the whole Branch to the EnrollmentStatus rather than just the slug? I'm thinking of reducing the need to perform the lookup,

I… don't… think… so. I think you're thinking about the lookup in `get_enrolled_feature_configs()`? This is only done at `apply_pending_experiments`, i.e. at startup. I expect any performance gains coming from removing O(1) hash table lookups will be largely removed by the increase in I/O deserializing then serializing the larger `EnrollmentStatus`es.

Issue #4212, opened 2021-06-15
Prototype Swift Package with nimbus-sdk + glean


data-sync-user:
Let's learn about of Swift packaging will work for us, by using the work Jan-Erik did to package Glean as a swift package ([https://github.com/mozilla/glean-swift),](https://github.com/mozilla/glean-swift),) and make a prototype package that combines Nimbus, Glean and Viaduct.

It won't be for shipping, it will just be to learn about how it works and be able to make a concrete proposal for what a future shipping version should look like.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2253)
┆Story Points: 8
┆Epic: Unified a-s swift packages for iOS
┆Sprint: SYNC - 2021-06-25

data-sync-user:
➤ Ryan Kelly commented:

This prototyping process has been completed, with the following artifacts:

  * A demo video of the proposed consumer experience for iOS apps ( https://drive.google.com/file/d/12xsOMoFkxHZAEZ8tL5gBaiTYTYoQN1OO/view )
  * An ADR proposal for how to publish appservices as a multi-product Swift Package ( https://github.com/mozilla/application-services/pull/4304 )
  * A concrete work breakdown in the corresponding epic ( https://mozilla-hub.atlassian.net/browse/SYNC-2243 )

PR #4216, opened 2021-06-16
WIP experiment with standalone Nimbus iOS XCFramework [ci skip]

rfk:
(Draft PR for early visibility)

Leaning heavily on @badboy's excellent work on [building Glean as an XCFramework](https://github.com/mozilla/glean/pull/1663), I've started experimenting with what it would look like to do the same for Nimbus. My short-term goal is to see if I can make a Nimbus equivalent of the [glean-swift repo](https://github.com/mozilla/glean-swift) and try consuming Nimbus as a Swift Package. Not because we have consumers that want to use it that way yet, but to help me get my head around all the moving parts.

Connects to #4212.

### DIFF
--- a/build-scripts/xc-universal-er-binary.sh
+++ b/build-scripts/xc-universal-er-binary.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# This should be invoked from inside xcode, not manually
+if [ "$#" -ne 3 ]
+then
+    echo "Usage (note: only call inside xcode!):"
+    echo "Args: $*"
+    echo "path/to/build-scripts/xc-universal-binary.sh <FFI_TARGET> <APPSVC_ROOT_PATH> <buildvariant>"
+    exit 1
+fi
+
+# what to pass to cargo build -p, e.g. glean_ffi
+FFI_TARGET=$1
+# path to app services root
+APPSVC_ROOT_ROOT=$2# buildvariant from our xcconfigs
+BUILDVARIANT=$3
+
+RELFLAG=
+if [[ "$BUILDVARIANT" != "debug" ]]; then
+    RELFLAG=--release
+fi
+
+set -euvx
+
+if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
+  # Assume we're in Xcode, which means we're probably cross-compiling.
+  # In this case, we need to add an extra library search path for build scripts and proc-macros,
+  # which run on the host instead of the target.
+  # (macOS Big Sur does not have linkable libraries in /usr/lib/.)
+  export LIBRARY_PATH="${DEVELOPER_SDK_DIR}/MacOSX.sdk/usr/lib:${LIBRARY_PATH:-}"
+fi
+
+IS_SIMULATOR=0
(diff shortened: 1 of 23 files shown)


rfk:
One early observation: I was initially hoping to do this for *just* Nimbus, but I think we have to do it for a little Nimbus + Glean megazord, because the Nimbus Swift bindings directly use the Glean Swift bindings, and I don't know how well it would work to have Nimbus instead try to depend on the Glean Swift Package. We'll see...

build-scripts/xc-universal-er-binary.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
rfk, line 1: This script is also ~shamelessly copied from~ based heavily on the work done to compile Glean for the M1. We probably want to push the changes here up into our existing `xc-universal-binary.sh` script, even without any of the new XCFramework/Swift-Package stuff.

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 51;
```
rfk, line 6: Eh? XCode doing XCode things maybe?

megazords/nimbus-ios/NimbusStandalone.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,587 @@
+// !$*UTF8*$!
```
rfk, line 1: This and its supporting `.swift` files were copied from the iOS megazord and heavily trimmed down. Depending on how the build process shapes up, we might want to consider moving e.g. the Swift RustLog and Viaduct stuff out of the iOS megazord and into some shared location.

st3fan:
@rfk This is awesome progress. I'll be the first consumer with Focus iOS. Let me know if you have either a swift package or just the xcframework and i'll integrate that into a test branch of Focus.

rfk:
Some more progress here:

* I've made https://github.com/rfk/nimbus-swift as a swift package that publishes the pre-built Nimbus+Glean binary artifact
* I've hacked up focus-ios so that it pulls this as a dependency rather than the standalone glean, ref https://github.com/rfk/focus-ios/pull/1
* I've further hacked up focus-ios to make some calls into the Nimbus API
* When I build this hacked-up focus-ios, it seems to run successfully and to actually make calls to Nimbus!

focus-nimbus-hackery

From this I interpret that the XCFramework stuff is actually compiling correctly. @st3fan if you're happy to check out the above and see if it seems sensible, that would be great.

Next, I need to see if I can understand what it would look like for Nimbus to consume Glean as a swift package, rather than requiring consumers like Focus to know about megazording. Megazording may turn out to be the only tractable approach in the short term, but it'll be worth understanding the landscape.

rfk:
This was a really useful experiment, but I'm going to close it in favour of something *even better* which I'll share more about shortly 😁

PR #4218, opened 2021-06-16
Removes old StateV1 schema from fxa-client

### LINKED ISSUE #3912: Remove the old `StateV1` code from fxa-client
Ref https://github.com/mozilla/application-services/pull/3876#discussion_r589897355

The fxa-client code has support for a very old `StateV1` storage format, which is how we stored data on disk for the first couple of iterations of the client. We've been using the newer `StateV2` format for quite some time now, and anybody still stuck with old-format data is probably never going to update. We could reduce ongoing maintenance complexity by removing the old `StateV1` code from [state_persistence.rs](https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/state_persistence.rs).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2096)
┆Sprint: SYNC - 2021-06-25

tarikeshaq:
fixes #3912 

Removes the old StateV1 from the fxa-client.

Still not 100% sure of the implication of this, but based on the issue it seems more or less harmless.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,4 +18,12 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
-- `./libs/build-all.sh` now displays a more helpful error message when a file fails checksum integrity test.
+
+## fxa-client
+### ⚠️ Breaking Changes ⚠️
+  - The old StateV1 persisted state schema is now removed. ([#4218](https://github.com/mozilla/application-services/pull/4218))
+    Users on very old versions of this component will no longer be able to cleanly update to this version. Instead, the consumer code
+    will recieve an error indicating that the schema was not correctly formated.
+
+## Other
+  - `./libs/build-all.sh` now displays a more helpful error message when a file fails checksum integrity test.
--- a/components/fxa-client/src/internal/config.rs
+++ b/components/fxa-client/src/internal/config.rs
@@ -90,46 +90,6 @@ impl Config {
         self
     }
 
-    // FIXME
-    #[allow(clippy::too_many_arguments)]
-    pub(crate) fn init(
-        content_url: String,
-        auth_url: String,
-        oauth_url: String,
-        profile_url: String,
-        token_server_endpoint_url: String,
-        authorization_endpoint: String,
-        issuer: String,
-        jwks_uri: String,
-        token_endpoint: String,
-        userinfo_endpoint: String,
-        introspection_endpoint: String,
-        client_id: String,
-        redirect_uri: String,
-        token_server_url_override: Option<String>,
-    ) -> Self {
-        let remote_config = RemoteConfig {
-            auth_url,
-            oauth_url,
-            profile_url,
-            token_server_endpoint_url,
-            authorization_endpoint,
-            issuer,
-            jwks_uri,
-            token_endpoint,
-            userinfo_endpoint,
-            introspection_endpoint,
-        };
-
-        Config {
-            content_url,
-            remote_config: RefCell::new(Some(Arc::new(remote_config))),
-            client_id,
-            redirect_uri,
-            token_server_url_override,
-        }
-    }
-
     fn remote_config(&self) -> Result<Arc<RemoteConfig>> {
         if let Some(remote_config) = self.remote_config.borrow().clone() {
             return Ok(remote_config);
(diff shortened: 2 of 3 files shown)


CHANGES_UNRELEASED.md
```diff
@@ -18,4 +18,9 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
-- `./libs/build-all.sh` now displays a more helpful error message when a file fails checksum integrity test.
+
+## fxa-client
+  - The old StateV1 persisted state schema is now removed. ([#4218](https://github.com/mozilla/application-services/pull/4218))
```
rfk, line 23: Let's add a brief comment here about what this change means for consumers, both because it's useful info for consumers and because it helps us reason through the implications.

In practice I expect it's something like "Users on very old versions of this component will no longer be able to cleanly update to this new version. They will...???" <- actually I'm not 100% sure what will happen to those users if they do try to update, let's see if we can work it out :-)

components/fxa-client/src/internal/config.rs
```diff
@@ -90,46 +90,6 @@ impl Config {
         self
     }
 
-    // FIXME
-    #[allow(clippy::too_many_arguments)]
-    pub(crate) fn init(
```
rfk, line 95: Took me a while to understand what was happening here - was `StateV1` the only consumer of this method, and hence it can now be removed as dead code?

components/fxa-client/src/internal/state_persistence.rs
```diff
@@ -135,156 +132,10 @@ impl StateV2 {
     }
 }
 
-/// Migration from `StateV1` to `StateV2`.
-/// There was a lot of changing of structs and renaming of fields,
-/// but the key change is that we went from supporting multiple active
-/// refresh_tokens to only supporting a single one.
-///
```
rfk, line 142: In place of the implementation here, could you please leave a modified version of this comment, explaining that there *was* a `StateV1` that looked different, and it was deliberately removed? (Otherwise future readers of this code may *infer* the existence of a `StateV1` and spend needless cycles trying to work out where the code for it lives).

components/fxa-client/src/internal/state_persistence.rs
```diff
@@ -135,156 +132,10 @@ impl StateV2 {
...
-        let state = state_from_json(state_v1_json).unwrap();
```
rfk, line 253: I think we should keep this test data, and turn it into a test for what happens when we encounter the no-longer-supported `StateV1` data in the wild.

The simplest possible behaviour is "we throw an error, and it's up to consumers of the library to catch this and deal with it". I think that's fine (we don't really expect it to happen in practice, otherwise we wouldn't be deleting this code) but it's worth making sure that we do actually throw such an error, rather than e.g. silently mis-interpreting the old-format data as some newer format.

rfk (changes requested):
Great! The code here all looks good, but I had a couple of higher-level expectation-management comments that I think are worth pursuing for completeness.

components/fxa-client/src/internal/config.rs
tarikeshaq: Yup! Clippy was complaining that it was dead code, so it went

components/fxa-client/src/internal/state_persistence.rs
tarikeshaq: Ended up putting this up in the module docs at the top - seemed more appropriate there 😄

components/fxa-client/src/internal/state_persistence.rs
```diff
@@ -26,6 +26,8 @@
 //! For backwards-incompatible changes to the data (such as removing or significantly refactoring
 //! fields) we define a new `StateV{X+1}` struct, and use the `From` trait to define how to update
 //! from older struct versions.
+//! For an example how the conversion works, [we can look at `StateV1` which was deliberately removed](https://github.com/mozilla/application-services/issues/3912)
+//! The code that was deleted demonstrates how we can implement the migration
```
rfk, line 30: 👍🏻

CHANGES_UNRELEASED.md
```diff
@@ -18,4 +18,11 @@ Use the template below to make assigning a version number during the release cut
...
+    Users on very old versions of this component will no longer be able to cleanly update to this version. Instead, the consumer code
+    will recieve an error indicating that the schema was not correctly formated.
```
rfk, line 25: Semantically, I think we should consider this a breaking change; could you please flag it as such such using the template from the comment at the top of the doc?

rfk (approved):
Thanks! One final thought occurred to me about the changelog, but otherwise this looks great 👍🏻.

PR #4220, opened 2021-06-18
Update and simplify Mergify bot config.

rfk:
The current mergify bot config has it rebasing active PRs and
impersonating random org members in order to do so This is a known
limitation of how mergify has to interact with the GitHub API, but
it's still weird and confusing.

This commit, along with an accompanying change to GitHub branch
protection rules, is an attempt to simplify that behaviour. We
now expect mergify to only merge PRs that are up-to-date with their
base branch, and to only do it be directly adding the existing commits
to the base branch.

While I was in here, I also took the opportunity to clean up some of
the status-check-checking logic, which in its previous incarnation
looked like it could very easily bitrot.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -15,72 +15,32 @@ pull_request_rules:
       - "#approved-reviews-by>=1"
       - base=main
       - label=checkin-needed
-      - "status-success=android-build-pr"
-      - "status-success=lint-detekt"
-      - "status-success=lint-ktlint"
-      - "status-success=check-protobuf-uptodate"
-      - "status-success=check-dependencies"
-      - "status-success=check-formatting"
-      - "status-success=swiftlint"
-      - "status-success=bash-lint"
-      - "status-success=clippy"
-      - "status-success=run-tests"
-      - "status-success=carthage-framework"
+      # What we want to say here is "all checks passed", but that's not a concept
+      # that makes sense in GitHub:
+      #
+      #    https://docs.mergify.io/conditions/#validating-all-status-checks
+      #
+      # But we don't want to just list out all the checks, because we have many of
+      # them and some of them are added dynamically and we might add or remove checks
+      # without updating the list here.
+      #
+      # So instead we insist that a small set of known checks has completed,
+      # and also that there are no other checks in unexpected states.
+      - "check-success=Decision Task"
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -15,72 +15,32 @@ pull_request_rules:
...
+      #
+      # So instead we insist that a small set of known checks has completed,
+      # and also that there are no other checks in unexpected states.
+      - "check-success=Decision Task"
+      - "check-success=run-tests"
+      - "check-success=check-formatting"
+      - "#check-success-or-neutral>=5" # we should always have at least this many checks
+      - "#check-pending=0"
+      - "#check-stale=0"
+      - "#check-failure=0"
```
skhamis, line 35: I might be missing something, but at the bottom of the linked doc it says not to use the `#check-failure=0` since that will fail immediately https://docs.mergify.io/conditions/#about-status-checks

.mergify.yml
```diff
@@ -15,72 +15,32 @@ pull_request_rules:
...
-      - "status-success=bash-lint"
-      - "status-success=clippy"
-      - "status-success=run-tests"
-      - "status-success=carthage-framework"
-    actions:
-      merge:
-        method: merge
-        strict: smart
-        strict_method: rebase
+        method: rebase
```
skhamis, line 44: nice! Looks simpler and definitely like the rebase pattern more as we can keep the way the PR writer intends

skhamis (approved):
Looks good to me! Thanks for diving into this and definitely understand this process a lot more. Just one question regarding the check-failure but otherwise looks good.

.mergify.yml
rfk: Right. The intention here is that *all* of these conditions have to be true simultaneously:

* At initial creation, `#check-failure=0` is true (nothing has failed, yet!) but e.g. `#check-success-or-neutral>=5` will be false because the checks haven't started to run yet.
* After the checks start running and the other conditions become true, `#check-failure=0` will be false if any of the checks have failed (even if it's one we forgot to name explicitly in the list here)

My read of the advice in the linked comment is to guard against the naive use of `#check-failure=0` by itself as a way to say "no checks failed" without any further qualification. (Which to be fair, I 100% would have done if they didn't have that warning in there).

PR #4221, opened 2021-06-18
Make `cargo uniffi-bindgen` alias work when executed from a crate.

rfk:
Previously, our special `cargo uniffi-bindgen` alias would work correctly
when executed from the workspace root directory, but would not work when
executed in the root directory of a crate (because from that location,
`cargo run` tries to run a binary from the crate you're in rather than
from the containing workspace).

This tweaks the alias to explicitly specify the package containing
the command, so it should work from any directory.

(I encountered this while working on #4216, but to be honest my
main motivation here is to see if the new mergify config will successfully
auto-land this PR).

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,5 @@
 [alias]
 regen-protobufs = ["run", "--bin", "protobuf-gen", "tools/protobuf_files.toml"]
-uniffi-bindgen = ["run", "--bin", "embedded-uniffi-bindgen"]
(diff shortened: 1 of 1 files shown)


mhammond (approved):

mhammond:
good bot!

rfk:
Heh, it still needlessly re-wrote my commit to include itself as the committer, but at least it did it under its own name!

PR #4224, opened 2021-06-21
Fix parse tabs bug

### LINKED ISSUE #3892: Unable to parse tab records uploaded by firefox-ios
I happened to notice that my Fenix is not able to display synced tabs from my Firefox iOS device, even though the iOS device sees the Fenix tabs just fine. In the debug log for Fenix is see:

```
Error deserializing incoming record: Error parsing JSON data: invalid string "1614239376.80", expected u64
```

If I use about:sync to inspect the synced tab records from my iOS device, indeed I see records with floats for the timestamp, e.g:

```
[{"title":"Hacker News","lastUsed":"1614239376.80",
```

So I think this is a simple schema mis-match between the Rust code and the firefox-ios code. The [schema docs](https://mozilla-services.readthedocs.io/en/latest/sync/objectformats.html#id9) for the tab format are not clear on the precise semantics of this field:

```
lastUsed string or integer: string representation of Unix epoch (in seconds) at which the tab was last accessed.
Or the integer 0. Your code should accept either. This is ghastly; we apologize.
```

Probably the right thing for us to do is always write an integer value here, but accept and truncate floats if they're present..?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2602)
┆Epic: a-s storage components (backlog)
┆Sprint End Date: 2021-06-25

lougeniaC64:
Fixes #3892. This is also related to the firefox iOS bug [#8272](https://github.com/mozilla-mobile/firefox-ios/pull/8727), which updates the tab sync code converting the `lastUsed` type from a stringified float to an integer. The docs have also been updated [here](https://github.com/mozilla-services/docs/pull/92).

  - Note:

### DIFF
--- a/components/tabs/src/sync/record.rs
+++ b/components/tabs/src/sync/record.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::error::*;
+use serde::de::Error;
 use serde_derive::{Deserialize, Serialize};
 use serde_json::{Map, Value as JsonValue};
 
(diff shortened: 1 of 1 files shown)


components/tabs/src/sync/record.rs
```diff
@@ -27,7 +28,163 @@ pub struct TabsRecord {
 impl TabsRecord {
     #[inline]
     pub fn from_payload(payload: sync15::Payload) -> Result<Self> {
-        let record: TabsRecord = payload.into_record()?;
-        Ok(record)
+        let client_name: String = payload.data["clientName"]
```
rfk, line 31: Please leave a brief inline comment here about why we have to parse this by hand rather than letting serde do its magic.

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +28,163 @@ pub struct TabsRecord {
...
+            .as_str()
+            .unwrap_or_default()
```
rfk, line 33: Hrm, so I think there's a subtle change in behaviour here. If `clientName` is present but is not a string, IIUC the previous code would have given an error, while the new code will default it to the empty string. I don't have a super strong opinion about which is the better behaviour, but it may be useful to opt for consistency with other engines and continue to error out in this case. @lougeniaC64 WDYT?

(If we do want to error out here, [`Option::ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or) may be helpful, and we should add some tests to cover missing or badly-typed data)

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +28,163 @@ pub struct TabsRecord {
...
+                    "test2.com"
+                ],
+                "icon": "test2.png",
+                "lastUsed": "1623745018.99" // test with stringified float value
+            },
+        ]);
+
+        data.insert("tabs".to_string(), tabs);
+        data.insert("clientName".to_string(), JsonValue::from("Nightly"));
+        data.insert("ttl".to_string(), JsonValue::from(0));
```
rfk, line 145: I don't feel strongly about it, but I wonder if this would be more readable if you constructed the data wrapper as part of the JSON macro, along the lines of:

```
let data = json!({
  "clientName": ...,
  "tabs": [
     ...
   ],
}).as_map().unwrap().clone();
```

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +28,163 @@ pub struct TabsRecord {
...
+        let l = last_used_val.as_str().unwrap_or_default();
+
+        match l.parse::<f64>() {
+            Ok(f) => f.trunc() as u64,
+            Err(_) => l.parse::<u64>().unwrap_or_default(),
+        }
+    } else {
+        match last_used_val.as_f64() {
+            Some(f) => f.trunc() as u64,
+            None => last_used_val.as_u64().unwrap_or_default(),
```
rfk, line 90: As above, this will now ignore invalid `lastUsed` fields and replace with the default value of zero.

rfk (commented):
@lougeniaC64 thanks, this looks good to me, but I did have a question about an apparent change in how this handles invalid tab records. I'm not opposed to the change but wanted to note it explicitly and make sure we give it some consideration - thoughts?

lougeniaC64:
> @lougeniaC64 thanks, this looks good to me, but I did have a question about an apparent change in how this handles invalid tab records. I'm not opposed to the change but wanted to note it explicitly and make sure we give it some consideration - thoughts?

Ironically, when unwrapping I tend to default to `unwrap_or_default`. I'd much rather maintain consistency with the previous behavior, so I'll make the necessary changes.

lougeniaC64:
@rfk I got (perhaps) a little overzealous, wrote tests around the previous behavior and then amended my code to pass those tests. The error messages aren't exactly the same of course, but the behavior should be. Please feel free to take your red pen out. Also, this isn't a rush as I'll be out tomorrow.

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +29,379 @@ pub struct TabsRecord {
 impl TabsRecord {
     #[inline]
     pub fn from_payload(payload: sync15::Payload) -> Result<Self> {
-        let record: TabsRecord = payload.into_record()?;
-        Ok(record)
+        // Note: We are hand-parsing the tabs payload so that we can provide support for
+        // older clients that may have a stringified float or integer `last_used` value.
+        let client_name: String =
```
rfk, line 34: Stylistically, it feels like we shouldn't need the explicit `: String` here, unless I'm missing something? (Put another way, the presence of `: String` here made me wonder if I was missing something non-obvious about the behaviour of the called function that made the explicit type necessary, but I couldn't see anything)

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +29,379 @@ pub struct TabsRecord {
...
+
+                Ok(TabsRecordTab {
+                    title,
+                    url_history,
+                    icon: icon?,
+                    last_used,
+                })
+            })
+            .into_iter()
+            .collect::<Result<_>>();
```
rfk, line 90: Stylistically, I think it would be cleaner to apply the question-mark here like `.collect::<Result<_>>()?;` rather than later on when you're constructing the `TabsRecord`, then you don't have to repeat the `Result` type when you're defining the `tabs` variable.

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +29,379 @@ pub struct TabsRecord {
...
+    // are either floats, integers, or stringified floats or integers. We attempt to
+    // represent `last used` as a float before converting it to an integer. If that
+    // operation isn't successful, we try converting `last used` to an integer directly.
+    // If that isn't successful, the returned value will be zero.
+
+    let invalid_err_msg = "invalid `lastUsed`, expected u64";
+    let last_used = if last_used_val.is_string() {
+        let l = last_used_val.as_str().unwrap_or_default();
+
+        match l.parse::<f64>() {
```
rfk, line 114: (This is a comment on Rust, not on your code)

I simultaneously love and hate this `parse` syntax...like, it's so powerful and so clever and at the same time, so darn *weird* for what seems like a really basic operation 😬. Anyways...

components/tabs/src/sync/record.rs
```diff
@@ -27,7 +29,379 @@ pub struct TabsRecord {
...
+                            .ok_or_else(|| {
+                                serde_json::Error::custom(
+                                    "invalid `urlHistory` value, expected string",
+                                )
+                            })?
+                            .to_string())
+                    })
+                    .into_iter()
+                    .collect::<Result<_>>()?;
+                let icon: Result<Option<String>, serde_json::Error> =
```
rfk, line 65: Could you use your `parse_string_from_json` helper for `icon` here?

rfk (approved):
> I got (perhaps) a little overzealous

Heh, you will never hear me accuse anyone of that! 😆 

This looks great, thanks for being so thorough with it. Since you invited the Red Pen I left a couple of very hand-wavy comments about places in the code where my reading hit a little pause to check I understood what was going on, but really, this is all 👍🏻, good stuff!

Also: my word serde gives us a lot for free by default!

PR #4225, opened 2021-06-22
Build iOS Rust code as an XCFramework.

rfk:
(Opening as a draft for early visibility)

This is an experiment in making our iOS build process more compatible
with M1 Mac build environments, and easier to consume via Swift Packages.

In our existing iOS build setup, we compile together all of the Rust
code and all of the wrapping Swift code into a single module named
"MozillaAppServices". This is a bit of a violation of clean package
separation, because it means that we risk name conflicts in the Swift
code between different components.

In this new setup, we don't pre-compile any of the Swift code. Instead
we build only the Rust code, package it along with its associated headers,
tell Swift to treat this as a module named "MozillaRustComponents".
Each individual Swift wrapper can then `import MozillaRustComponents`
and get access to the low-level FFI functions that it needs.

An additional benefit of this setup is that we can distribute it as
a `.xcframework` bundle for easy consumption on different platforms,
and we can build it without using XCode.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,6 +177,7 @@ commands:
             cargo metadata --locked > /dev/null
             python3 ./tools/dependency_summary.py --check ./DEPENDENCIES.md
             python3 ./tools/dependency_summary.py --all-ios-targets --package megazord_ios --check megazords/ios/DEPENDENCIES.md
+            python3 ./tools/dependency_summary.py --all-ios-targets --package ios_rust --check megazords/ios-rust/DEPENDENCIES.md
             python3 ./tools/dependency_summary.py --all-android-targets --package megazord --check megazords/full/DEPENDENCIES.md
             python3 ./tools/dependency_summary.py --all-android-targets --package megazord --format pom --check megazords/full/android/dependency-licenses.xml
   bench-all:
@@ -380,6 +381,21 @@ jobs:
       - store_artifacts:
           path: raw_xcodebuild.log
           destination: logs/raw_xcodebuild.log
+      - run:
+          name: Build XCFramework archive
+          command: |
+            # For now we need the nightly toolchain to build for the M1 simulator.
+            rustup install nightly
+            # For now we need to build the Rust stdlib from source for the M1 simulator.
+            rustup component add rust-src --toolchain nightly-x86_64-apple-darwin
+            if [ -z "${CIRCLE_TAG}" ]; then
+              # XCode tests build in Debug configuration, save us a full
+              # Rust rebuild in Release mode by forcing Debug mode on
+              # non-release builds.
+              bash megazords/ios-rust/build-xcframework.sh --build-profile debug
+            else
+              bash megazords/ios-rust/build-xcframework.sh --build-profile release
+            fi
       - save_cache:
           name: Save sccache cache
           key: sccache-cache-macos-{{ arch }}-{{ epoch }}
@@ -392,21 +408,33 @@ jobs:
             ZIP_URL=https://circleci.com/api/v1.1/project/github/mozilla/application-services/$CIRCLE_BUILD_NUM/artifacts/0/dist/MozillaAppServices.framework.zip
             echo "{\"0.0.1\":\"$ZIP_URL\"}" > mozilla.app-services.json
       - store_artifacts:
+          name: Store Carthage framework in workspace
           path: MozillaAppServices.framework.zip
           destination: dist/MozillaAppServices.framework.zip
       - store_artifacts:
+          name: Store Carthage bin-only project specification in workspace
           path: mozilla.app-services.json
           destination: dist/mozilla.app-services.json
+      - store_artifacts:
+          name: Store XCFramework bundle in workspace
+          path: megazords/ios-rust/MozillaRustComponents.xcframework.zip
+          destination: dist/MozillaRustComponents.xcframework.zip
       - run:
           name: "Carthage binary snapshot URL"
           command: |
             JSON_URL=https://circleci.com/api/v1.1/project/github/mozilla/application-services/$CIRCLE_BUILD_NUM/artifacts/0/dist/mozilla.app-services.json
             echo "Add the following line to your Cartfile:"
             echo "binary \"$JSON_URL\" ~> 0.0.1-snapshot # mozilla/application-services@$CIRCLE_SHA1"
+      - run:
+          name: "XCFramework bundle checksum"
+          command: |
+            shasum -a 256 ./megazords/ios-rust/MozillaRustComponents.xcframework.zip
+            echo "Use the above checksum to depend on MozillaRustComponents.xcframework.zip as a Swift Package binary target"
       - persist_to_workspace:
           root: .
           paths:
             - MozillaAppServices.framework.zip
+            - megazords/ios-rust/MozillaRustComponents.xcframework.zip
   Carthage release:
     executor: macos
     steps:
@@ -422,6 +450,21 @@ jobs:
             echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
             unzip "${GHR}.zip"
             ./${GHR}/ghr -replace "${CIRCLE_TAG}" MozillaAppServices.framework.zip
+  XCFramework release:
+    executor: macos
+    steps:
+      - full-checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Release XCFramework archive on GitHub
+          command: |
+            GHR=ghr_v0.12.0_darwin_amd64
+            GHR_SHA256=c868ef9fc5dd8c8a397b74d84051d83693c42dd59041cb17b66f90f563477249
+            curl -sfSL --retry 5 --retry-delay 10 -O "https://github.com/tcnksm/ghr/releases/download/v0.12.0/${GHR}.zip"
+            echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
+            unzip "${GHR}.zip"
+            ./${GHR}/ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
 
 workflows:
   version: 2
@@ -473,13 +516,21 @@ workflows:
   coverage:
     jobs:
       - Generate code coverage
-  carthage-framework:
+  ios-artifacts:
     jobs:
       - iOS build and test:
           filters: # required since `Release` has tag filters AND requires `Build`
             tags:
               only: /.*/
       - Carthage release:
+          requires:
+            - iOS build and test
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
+      - XCFramework release:
           requires:
             - iOS build and test
           filters:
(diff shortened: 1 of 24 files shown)


.circleci/config.yml
```diff
@@ -377,6 +378,17 @@ jobs:
...
+          name: Build XCFramework archive
+          command: |
+            if [ -z "${CIRCLE_TAG}" ]; then
+              # XCode tests build in Debug configuration, save us a full
+              # Rust rebuild in Release mode by forcing Debug mode on
+              # non-release builds.
+              bash megazords/ios-rust/build-xcframework.sh --build-profile debug
+            else
+              bash megazords/ios-rust/build-xcframework.sh --build-profile release
+            fi
```
rfk, line 391: It's not clear to me that we actually want to build this on every single PR, especially if it's expensive to build. But it's useful to me for testing purposes to start out doing it that way, and also it's what we already do for the carthage archive.

rfk:
OK! I seem to have managed to get CircleCI to build a `.zip` artifact using the new script and upload it:

* https://111888-129966583-gh.circle-artifacts.com/0/dist/MozillaRustComponents.xcframework.zip

Tomorrow, I will try pulling this in as a binary target from a Swift package and see if it'll work end-to-end.

rfk:
> Tomorrow, I will try pulling this in as a binary target from a Swift package and see if it'll work end-to-end.

It works! The Swift Package repo over at https://github.com/mozilla/rust-components-swift/ is referencing this artifact from CircleCI and I can successfully build against it for a local simulator build.

The PR here is still draft because it's blocked by the UniFFI update in https://github.com/mozilla/application-services/pull/4377, but I believe it's ready for review in the meantime.

rfk:
For testing whether this PR actually works, I recommend:

* Following the instructions in https://mozilla-hub.atlassian.net/browse/SYNC-2675 to test out a build with a published package, which as mentioned above is using an artifact produced by a CircleCI run on this PR.
* Follow the instructions in the README in this PR to try doing a local build and testing it in a consuming app, using [this fork of focus-ios](https://github.com/rfk/focus-ios/pull/2) as the consuming app.

I'm deliberately not saying any more detail about how to perform those tasks, because if there are parts of the procedure that aren't clear based on the docs in this PR and the rust-components-swift repo, then that's useful signal about things that need more docs :-)

.circleci/config.yml
```diff
@@ -422,6 +450,21 @@ jobs:
...
+          at: .
+      - run:
+          name: Release XCFramework archive on GitHub
+          command: |
+            GHR=ghr_v0.12.0_darwin_amd64
+            GHR_SHA256=c868ef9fc5dd8c8a397b74d84051d83693c42dd59041cb17b66f90f563477249
+            curl -sfSL --retry 5 --retry-delay 10 -O "https://github.com/tcnksm/ghr/releases/download/v0.12.0/${GHR}.zip"
+            echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
+            unzip "${GHR}.zip"
+            ./${GHR}/ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
```
tarikeshaq, line 467: Side comment: this is identical to the Carthage flow, which makes sense, I imagine we should probably briefly update https://mozilla.github.io/application-services/book/build-and-publish-pipeline.html?highlight=pipeline#application-services-build-and-publish-pipeline once those changes land (for the sake of time, we can get to it later)

.gitignore
```diff
@@ -49,4 +50,5 @@ build/docs/*
 docs/book
 
 # Generated rustdocs
-docs/rust-docs
\ No newline at end of file
+docs/rust-docs
+
```
tarikeshaq, line 54: 🙏

components/nimbus/ios/Nimbus/FeatureVariables.swift
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+import UIKit
```
tarikeshaq, line 6: Just out of curiosity, I imagine something broke, it didn't compile and we needed to add this?

components/fxa-client/uniffi.toml
```diff
@@ -3,4 +3,6 @@ package_name = "mozilla.appservices.fxaclient"
 cdylib_name = "megazord"
 
 [bindings.swift]
+ffi_module_name = "MozillaRustComponents"
+ffi_module_filename = "fxa_clientFFI"
```
tarikeshaq, line 7: Not sure, but do we still want to keep this if we're not adding the `fxa_client` yet? (until we can get NSS built for M1)

megazords/ios-rust/README.md
```diff
@@ -0,0 +1,108 @@
...
+1. Update its `uniffi.toml`, if any, to include the following settings:
+    ```
+    [bindings.swift]
+    ffi_module_name = "MozillaRustComponents"
+    ffi_module_filename = "<crate_name>FFI"
+    ```
+1. Add it as a dependency in `Cargo.toml`.
+1. Add a `pub use` declaration for it in `./src/lib.rs`.
+1. Add logic to `build-xcframework.sh` to copy or generate its header file into the build.
+1. Add a `#import` for its header file to `MozillaRustComponents.h`
```
tarikeshaq, line 60: Do we want to be explicit that the `#import` should follow the `#if canImport(...)` pattern? I think it's easy enough to figure out, especially if someone looks at one of the already existing components, but just thinking out loud

megazords/ios-rust/README.md
```diff
@@ -0,0 +1,108 @@
...
+and edit its `Swift.package` to use the above URL and checksum for the `MozillaRustComponents` binary target.
+* Commit the result to your local checkout, and make a git tag in `MAJOR.MINOR.PATCH` format.
+* Add your local `rust-components-swift` repo as a Swift Package dependency in a consuming app,
+  by specifying `file://path/to/rust-components-swift` as the git repo.
+
+Note that your changes *must* be included in a local git tag. Swift Package Manager is very
+opinionated and will only pull dependencies from a git tag that it can parse as a semver
+version number.
+
+For local testing without relying on CircleCI, you can also:
```
tarikeshaq, line 83: A super minor organizational thought (feel free to completely discarding this, I'm just making sure I note down everything I think of)

I read through this section and was first thinking "wait you can't do this locally without a PR?" until I hit this line, could be worth it to change the titles, maybe have this whole  section be a `## Testing` section, and have `### Testing Locally`, `### Testing a PR` and `### Testing a release` as separate sections - again for the sake of time, we can do this later ourselves 😄

megazords/ios-rust/build-xcframework.sh
```diff
@@ -0,0 +1,173 @@
...
+# Hardware iOS targets
+cargo_build aarch64-apple-ios
+
+# M1 iOS simulator.
+# It's currently in Nightly only and requires to build `libstd`.
+# We hope this will be available by default in Rust 1.56.0.
+BUILD_ARGS=(+nightly "${BUILD_ARGS[@]}" -Z build-std)
+cargo_build aarch64-apple-ios-sim
+
+# TODO: would it be useful to also include desktop builds here?
```
tarikeshaq, line 105: 🤔 Could be a good spike for a later time (to explore running tests via `swift test` like you mentioned in our chat yesterday)

megazords/ios-rust/build-xcframework.sh
```diff
@@ -0,0 +1,173 @@
...
+mkdir -p "$COMMON/Modules"
+cp "$THIS_DIR/module.modulemap" "$COMMON/Modules/"
+
+cp "$THIS_DIR/DEPENDENCIES.md" "$COMMON/DEPENDENCIES.md"
+
+mkdir -p "$COMMON/Headers"
+cp "$THIS_DIR/MozillaRustComponents.h" "$COMMON/Headers"
+cp "$REPO_ROOT/components/rc_log/ios/RustLogFFI.h" "$COMMON/Headers"
+cp "$REPO_ROOT/components/viaduct/ios/RustViaductFFI.h" "$COMMON/Headers"
+# TODO: it would be neat if there was a single UniFFI command that would spit out
```
tarikeshaq, line 136: This could be a good ticket/spike on the Uniffi side for later 😄

megazords/ios-rust/build-xcframework.sh
```diff
@@ -0,0 +1,173 @@
...
+# For now we generate the Swift bindings to get the headers as a side effect,
+# then delete the generated Swift code. Bleh.
+$CARGO uniffi-bindgen generate "$REPO_ROOT/components/nimbus/src/nimbus.udl" -l swift -o "$COMMON/Headers"
+$CARGO uniffi-bindgen generate "$REPO_ROOT/components/crashtest/src/crashtest.udl" -l swift -o "$COMMON/Headers"
+rm -rf "$COMMON"/Headers/*.swift
+
+# Flesh out the framework for each architecture based on the common files.
+# It's a little fiddly, because we apparently need to put all the simulator targets
+# together into a single fat binary, but keep the hardware target separate.
+# (TODO: we should try harder to see if we can avoid using `lipo` here, eliminating it
```
tarikeshaq, line 147: This could be another follow up nice-to-have ticket!

tarikeshaq (commented):
Took a look at the changes, looks good to me!

I'll spend some time getting this built locally and tested, then will come back and approve it!

Dropped a couple of thoughts, @rfk most of them are just reminders to myself to come back and create follow-up tickets (if they aren't already covered elsewhere)

components/fxa-client/uniffi.toml
tarikeshaq: Actually never mind, this more of a `uniffi` thing that it is a `Swift Packages` thing right?

megazords/ios-rust/README.md
tarikeshaq: Never mind! I have this the other way around 🤦 🤦

megazords/ios-rust/build-xcframework.sh
rfk: Right, exactly! I actually prototyped this a little bit and it seemed like it would work, but:

* Glean doesn't publish desktop builds, so we'd have to bring Glean into our megazord before we could do this
* It makes the compiled bundle even bigger, which may be a problem for production releases

megazords/ios-rust/README.md
rfk: Good idea, thanks.

components/fxa-client/uniffi.toml
rfk: It won't do any harm here if we leave it, but also it's only required because of how we do the XCFramework build, so it might make sense to take it out just to avoid confusion.

components/nimbus/ios/Nimbus/FeatureVariables.swift
rfk: Ah yes, thanks for the reminder. The code in this file is using some classes from `UIKit` (I think something about image-handling) but wasn't importing it explicitly. That was fine in our old iOS build setup because something else in the megazord was import it, but it broke when trying to compile this as its own package.

skhamis (approved):
This is awesome! Really easy to understand and tested with both a "ci published" version as well as local ( built on an M1) and both worked flawlessly. Great foundation for an eventual switchover to full xcframeworks!

components/fxa-client/uniffi.toml
rfk: (I'm going to go ahead and remove it to avoid confusion)

rfk:
Thanks both, I've updated to address a few of the docs-related comments.

PR #4226, opened 2021-06-22
Migrate DB to use origin and formActionOrigin

skhamis:
Since we're going to be migrating the DB anyways we might as well get rid of any mention of hostname/formSubmitUrl since desktop is now using origin/formActionOrigin and that is the "proper" name for these values anyways. I didn't need to actually/procure are migration code since when we create the tables it'll have the new names anyways.

  - Note:


components/logins/src/schema.rs
```diff
@@ -325,8 +325,8 @@ pub fn upgrade_sqlcipher_db(db: &mut Connection, encryption_key: &str) -> Result
     if user_version < 3 {
         // These indices were added in v3 (apparently)
         tx.execute_all(&[
-            CREATE_OVERRIDE_HOSTNAME_INDEX_SQL,
-            CREATE_DELETED_HOSTNAME_INDEX_SQL,
+            CREATE_OVERRIDE_ORIGIN_INDEX_SQL,
+            CREATE_DELETED_ORIGIN_INDEX_SQL,
         ])?;
```
bendk, line 330: I think these should stay the same.  If a user actually had a v2 database, then they would be using the old column names at this point.

bendk (changes requested):
The new code looks good overall and should work for new DBs, but I do think you need code to upgrade old databases.

The test I used a bunch was running the `sync-pass` example  with a checkout from `main`, adding a login or two, then checking out this branch and running `sync-pass` with the same database.  That should trigger the schema upgrade code and I think you'll see issues.

I think you could get away without writing a full new upgrade though.  The v4 -> v5 upgrade hasn't been released yet, so you could maybe sneak in the extra rename column statements around line 397.

components/logins/src/schema.rs
skhamis: Good catch, was a bit over zealous in the renames. Turns out I need to have a new version of that function for the `create` while keeping that one for any v3. Tests pass and migrations work now!

mhammond:
I can't see the PR for logins-per-field-encryption - where does that live? I'm not yet convinced `sqlcipher_export()` is the correct approach here given the other schema and encryption changes, but I'm not sure where to add those comments :)

bendk:
> I can't see the PR for logins-per-field-encryption - where does that live? I'm not yet convinced `sqlcipher_export()` is the correct approach here given the other schema and encryption changes, but I'm not sure where to add those comments :)

That one was https://github.com/mozilla/application-services/pull/4135, but I closed it when I merged the code from my fork into the mozilla branch.  It would be great to talk about that approach more, maybe we could just re-open the PR and continue there.

bendk:
Looks good to me after the latest changes.

PR #4294, opened 2021-06-23
Fix uniffi logins crash when there is no DB

skhamis:
This is related to the uniffi effort for the logins crate

fxiOS currently crashes when testing locally on https://github.com/mozilla-mobile/firefox-ios/pull/8602 due to creating a new DB and trying to migrate https://github.com/mozilla-mobile/firefox-ios/blob/041c85483b1aa4a4e357d8bd9419550e4f20347a/Storage/Rust/RustLogins.swift#L113

**Note: This will require a package update to be able to use in the PR above**

---

  - Note:


rfk (approved):
Seems reasonable to me - is the expected behaviour that the caller will catch this and fall back to some default behaviour?

skhamis:
Yes, they catch and run some logic off of that from what I understand!

Issue #3695, opened 2020-11-05
Push component should automatically recover from the server losing its uaid


rfk:
As noted in https://github.com/mozilla/application-services/issues/3314#issuecomment-714571731 and subsequent comments, if the autopush server receives a `404` or `410` error when trying to delivery a push via one of the mobile bridges, it will completely drop the registration record for that uaid. AFAICT any attempts by the client to operate on that uaid will produce an error.

I don't think the push component currently handles this case well. In fact I don't see any codepaths that would case us to detect that `self.uaid` has become invalid and is being rejected by the server.

This seems like something that could be handled in `verify_connection` by:

* Detecting the "uaid does not exist" error.
* Registering a new uaid.
* Reporting all existing subscriptions as expired.

We should also add some handling of this case in the `update` method, which clients may call when in this state in an attempt to repair this push subscriptions.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/SYNC-2570)
┆Epic: FxA Ecosystem (backlog)
┆Fix Versions: Release 92
┆Sprint End Date: 2021-08-06

rfk:
The docs and structure of the push component and overdue for a bit of an overhaul, we sadly haven't really kept up very good maintenance of it since it was initially written by the push team. However, that's an undertaking far beyond the scope of this bug! (But if you happen to find opportunities to clean up the docs or clarify things as you go here, that's also valuable).

I've added some high-level context-setting in https://github.com/mozilla/application-services/pull/4303 which I hope will help a bit.

Specifically for this bug, we want to deal with the unhandled error being reported from Fenix users [here](https://github.com/mozilla-mobile/fenix/issues/16721#issuecomment-737467963) and [here](https://github.com/mozilla-mobile/fenix/issues/17272):

```

W/App     (29404): mozilla.appservices.push.CommunicationError: Error(Communication Error: "Unhandled client error 410 : \"{\\\"code\\\": 410, \\\"errno\\\": 103, \\\"error\\\": \\\"\\\", \\\"more_info\\\": \\\"http://autopush.readthedocs.io/en/latest/http.html#error-codes\\\", \\\"message\\\": \\\"Request did not validate UAID not found\\\"}\"")
W/App     (29404):   at mozilla.appservices.push.RustError.intoException(RustError.kt:13)
W/App     (29404):   at mozilla.appservices.push.PushManager.subscribe(PushManager.kt:13)
[...]

PR #4303, opened 2021-06-25
Add some context-setting docs for the push component.

rfk:
There's a lot if background info you need to know if you're going
to effectively work on the push component, this tries to add some
pointers to it in the README.

I've also removed what appeared to be an early prototype API doc
since it mostly didn't match what is actually in the code, and
removed the `LICENSE` file for consistency with other components.

This is in service of being able to work on https://github.com/mozilla/application-services/issues/3695
so @tarikeshaq, pinging you for r? from a "new reader" perspective.
@jrconlin or @pjenvey if you have a moment to glance over these
for technical correctly that would be appreciated, but I know you've
got a lot going on so we won't block on you.


tarikeshaq (approved):
Thanks @rfk! This is good context - Gave it a quick read and will be looking around `autopush`'s docs for more information.

I'll leave this unmerged for now in case anyone else has anything to add

PR #4304, opened 2021-06-25
Add an ADR for plans around Swift Packages.

rfk:
This summarizes the results of my prototyping work on Swift Packages
with a concrete proposal for how we should proceed. Given preliminary
feedback I've had so far I have gone ahead and selected a recommended
option, but that is of course up for change if further discussion leads
us in that direction.

Here is the [**Rendered View**](https://github.com/mozilla/application-services/blob/swift-package-adr/docs/adr/0003-swift-packaging.md) for your convenience.

### DIFF
--- a/docs/adr/0003-swift-packaging.md
+++ b/docs/adr/0003-swift-packaging.md
@@ -1,0 +1,245 @@
+# Distributing Swift Packages
+
+* Status: proposed
+* Deciders: rfkelly
+* Date: 2021-06-25
+
+## Context and Problem Statement
+
+Our iOS consumers currently obtain application-services as a pre-compiled `.framework` bundle
+distributed via [Carthage](https://github.com/Carthage/Carthage). The current setup is not
+compatible with building on new M1 Apple Silicon machines and has a number of other problems.
+As part of a broader effort to modernize the build process of iOS applications at Mozilla,
+we have been asked to re-evaluate how application-services components are dsitributed for iOS.
+
+See [Problems with the current setup](#problems-with-the-current-setup) for more details.
+
+## Decision Drivers
+
+* Ease-of-use for iOS consumers.
+* Compatibility with M1 Apple Silicon machines.
+* Consistency with other iOS components being developed at Mozilla.
+* Ability for the Nimbus Swift bindings to easily depend on Glean.
+* Ease of maintainability for application-services developers.
+
+## Considered Options
+
+* **(A) Do Nothing**
+  * Keep our current build and distribution setup as-is.
+* **(B) Use Carthage to build XCFramework bundles**
+  * Make a minimal change to our Carthage setup so that it builds
+    the newer XCFramework format, which can support M1 Apple Silicon.
+* **(C) Distribute a single pre-compiled Swift Package**
+  * Convert the all-in-one `MozillaAppServices` Carthage build to a similar
+    all-in-one Swift Package, distributed as a binary artifact.
+* **(D) Distribute multiple source-based Swift Package targets, with pre-compiled Rust code**
+  * Split the all-in-one `MozillaAppServices` Carthage build into a separate
+    Swift Package target for each component, with a shared dependency on pre-compiled
+    Rust code as a binary artiact.
+
+## Decision Outcome
+
+Chosen option: **(D) Distribute multiple source-based Swift Packages, with pre-compiled Rust code**.
+
+This option will provide the best long-term consumer experience for iOS developers, and
+has the potential to simplify maintenance for application-services developers after an
+initial investment of effort.
+
+### Positive Consequences
+
+* Swift packages are very convenient to consume in newer versions of Xcode.
+* Different iOS apps can choose to import a different subset of the available components,
+  potentiallying helping keep application size down.
+* Avoids issues with mis-matched Swift version between application-services build
+  and consumers, since Swift files are distributed in source form.
+* Encourages better conceptual separation between Swift code for different components;
+  e.g. it will make it possible for two Swift components to define an item of the same
+  name without conflicts.
+* Reduces the need to use Xcode as part of application-services build process, in favour
+  of command-line tools.
+
+### Negative Consequences
+
+* More up-front work to move to this new setup.
+* We may be less likely to notice if our build setup breaks when used from within Xcode,
+  because we're not exercising that code path ourselves.
+* May be harder to concurrently publish a Carthage framework for current consumers who aren't
+  able to move to Swift packages.
+* There is likely to be some amount of API breakage for existing consumers, if only in having
+  to replace a single `import MozillaAppServices` with independent imports of each component.
+
+### Implementation Sketch
+
+We will maintain the existing Carthage build infrastructure in the application-services repo and continue publishing a pre-built Carthage framework,
+to support firefox-ios until they migrate to Swift Packages.
+
+We will add an additional iOS build task in the application-services repo, that builds *just the Rust code* as a `.xcframework` bundle.
+An initial prototype shows that this can be achieved using a relatively straightforward shell script, rather than requiring a second Xcode project.
+It will be published as a `.zip` artifact on each release in the same way as the current Carthage framework.
+The Rust code will be built as a static library, so that the linking process of the consuming application can pull in
+just the subset of the Rust code that is needed for the components it consumes.
+
+We will initially include only Nimbus and its dependencies in the `.xcframework` bundle,
+but will eventually expand it to include all Rust components (including Glean, which will continue
+to be included in the `application-services` repo as a git submodule)
+
+We will create a new repository `rust-components-swift` to serve as the root of the new Swift Package distribution.
+It will import the `application-services` repository as a git submodule. This will let us iterate quickly on the
+Swift packaging setup without impacting existing consumers.
+
+We will initially include only Nimbus and its dependencies in this new repository, and the Nimbus swift code
+it will depend on Glean via the external `glean-swift` package. In the future we will publish all application-services
+components that have a Swift interface through this repository, as well as Glean and any future Rust components.
+(That's why the repository is being given a deliberately generic name).
+
+The `rust-components-swift` repo will contain a `Package.swift` file that defines:
+
+* A single binary target that references the pre-built `.xcframework` bundle of Rust code.
+* One Swift target for each component, that references the Swift code from the git submodule
+  and depends on the pre-built Rust code.
+
+We will add automation to the `rust-components-swift` repo so that it automatically tracks
+releases made in the `application-services` repo and creates a corresponding git tag for
+the Swift package.
+
+At some future date when all consumers have migrated to using Swift packages, we will remove
+the Carthage build setup from the application-services repo.
+
+At some future date, we will consider whether to move the `Package.swift` definition in to the `application-services` repo,
+or whether it's better to keep it separate. (Attempting to move it into the `application-services` will involve non-trivial
+changes to the release process, because the checksum of the released `.xcframework` bundle needs to be included in
+the release taged version of the `Package.swift` file.)
+
+# Pros and Cons of the Options
+
+### **(A) Do Nothing**
+
+In this option, we would make no changes to our iOS build and publishing process.
+
+* Good, because it's the least amount of work.
+* Neutral, because it doesn't change the maintainability of the system for appservices
+  developers.
+* Neutral, because it doesn't change the amount of separation between Swift code
+  for our various components.
+* Neutral, because it doesn't address the Swift version incompatibility issues around
+  binary artifacts.
+* Bad, because it will frustrate consumers who want to develop on M1 Apple Silicon.
+* Bad, because it may prevent consumers from migrating to a more modern build setup.
+* Bad, because it would prevent consumers from consuming Glean as a Swift package;
+  we would require them to use the Glean that is bundled in our build.
+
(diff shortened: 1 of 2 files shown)


jhugman (approved):
💯 I approve of this decision.

travis79 (approved):
This is awesome, I'm 100% on-board with the multiple Swift packages option from the Nimbus perspective!  One question, since Glean already supports Swift packages, what does this mean for the Glean submodule that is currently included in A-S? Will this complicate the build process for Glean if we are using the submodule for Android builds and the Glean-provided (external) Swift package for iOS?

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+* Deciders: rfkelly
+* Date: 2021-06-25
+
+## Context and Problem Statement
+
+Our iOS consumers currently obtain application-services as a pre-compiled `.framework` bundle
+distributed via [Carthage](https://github.com/Carthage/Carthage). The current setup is not
+compatible with building on new M1 Apple Silicon machines and has a number of other problems.
+As part of a broader effort to modernize the build process of iOS applications at Mozilla,
+we have been asked to re-evaluate how application-services components are dsitributed for iOS.
```
tarikeshaq, line 13: nit: typo
```suggestion
we have been asked to re-evaluate how application-services components are distributed for iOS.
```

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+
+This option will provide the best long-term consumer experience for iOS developers, and
+has the potential to simplify maintenance for application-services developers after an
+initial investment of effort.
+
+### Positive Consequences
+
+* Swift packages are very convenient to consume in newer versions of XCode.
+* Different iOS apps can choose to import a different subset of the available components,
+  potentiallying helping keep application size down.
```
tarikeshaq, line 52:
```suggestion
  potentially helping keep application size down.
```

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+We will add an additional iOS build task in the application-services repo, that builds *just the Rust code* as a `.xcframework` bundle. An initial prototype shows that this can be achieved using a relatively straightforward shell script, rather than requiring a second XCode project. It will be
+published as a `.zip` artifact on each release in the same way as the current Carthage framework.
+
+We will create a new repository `rust-components-swift` to serve as the root of the new Swift Package
+distribution. It will import the `application-services` repository as a git submodule. This will let
+us iterate quickly on the Swift packaging setup without impacting existing consumers.
+
+The `rust-components-swift` repo will contain a `Package.swift` file that defines:
+
+* A single binary target that references the pre-built `.xcframework` bundle of Rust code.
```
tarikeshaq, line 82: I think I need to go learn some more about Swift packages - but I'm assuming this means that if our consumer is consuming multiple component packages, it's still exactly 1 `.xcframework` (i.e we're not gonna be duping our compiled rust code)? I think that's the case, but just wanted to say it out loud 😄

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+* May be harder to concurrently publish a Carthage framework for current consumers who aren't
+  able to move to Swift packages.
+* There is likely to be some amount of API breakage for existing consumers, if only in having
+  to replace a single `import MozillaAppServices` with independent imports of each component.
+
+### Implementation Sketch
+
+We will maintain the existing Carthage build infrastructure in the application-services repo and continue publishing a pre-built Carthage framework, to support firefox-ios until they migrate to Swift Packages.
+
+We will add an additional iOS build task in the application-services repo, that builds *just the Rust code* as a `.xcframework` bundle. An initial prototype shows that this can be achieved using a relatively straightforward shell script, rather than requiring a second XCode project. It will be
```
tarikeshaq, line 73: (very naive thought incoming): when we talk about the `Rust Code` here, we mean _all_ the rust code right? like, the fxa_client component's rust code will still include the nimbus's rust code?

If not, ignore me 😬 - if it does include all the code, is there a world where each component only has a separate chunk of the Rust code that it needs? as I'm writing this, I imagine there's an easy answer to why it's `no`, but wanted to double check

tarikeshaq (approved):
I have little experience with iOS tooling, but this seems like a big improvement to me! 🚀

st3fan:
@rfk I think it would also be good to mention the `BUILD_LIBRARY_FOR_DISTRIBUTION` build flag as this is what enables _Library Evolution_ and _Module Stability_ in Swift. Because @badboy 's current Glean experiment uses this, we were able to use that framework without any changes in Xcode 12.5, 12.5.1 and ... 13.0Beta!

More technical details in https://swift.org/blog/library-evolution/

rfk:
@st3fan thanks! In my ideal world here I don't think this flag will be necessary because we ship the Swift sources rather than pre-compiled Swift binaries, but I will dig a little deeper and factor it in.

docs/adr/0003-swift-packaging.md
badboy: To answer this: yep, an xcframework is basically a directory with one library per target in correctly named directories, so that the final application knows where to look.

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+
+Chosen option: **(D) Distribute multiple source-based Swift Packages, with pre-compiled Rust code**.
+
+This option will provide the best long-term consumer experience for iOS developers, and
+has the potential to simplify maintenance for application-services developers after an
+initial investment of effort.
+
+### Positive Consequences
+
+* Swift packages are very convenient to consume in newer versions of XCode.
```
badboy, line 50: https://youtu.be/Cj8n4MfhjUc?t=41

```suggestion
* Swift packages are very convenient to consume in newer versions of Xcode.
```

(sorry, I had to.)

docs/adr/0003-swift-packaging.md
badboy: While a valid thing here, I think it's probably negligible for the Swift code compared to shipping the prebuilt Rust library that has code of all components anyway.

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+* Different iOS apps can choose to import a different subset of the available components,
+  potentiallying helping keep application size down.
+* Avoids issues with mis-matched Swift version between application-services build
+  and consumers, since Swift files are distributed in source form.
+* Encourages better conceptual separation between Swift code for different components;
+  e.g. it will make it possible for two Swift components to define an item of the same
+  name without conflicts.
+* Reduces the need to use XCode as part of application-services build process, in favour
+  of command-line tools.
```
badboy, line 59: As long as it's tested that it continues to work within Xcode. But I guess the consumers still working is the test.

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+### Implementation Sketch
+
+We will maintain the existing Carthage build infrastructure in the application-services repo and continue publishing a pre-built Carthage framework, to support firefox-ios until they migrate to Swift Packages.
+
+We will add an additional iOS build task in the application-services repo, that builds *just the Rust code* as a `.xcframework` bundle. An initial prototype shows that this can be achieved using a relatively straightforward shell script, rather than requiring a second XCode project. It will be
+published as a `.zip` artifact on each release in the same way as the current Carthage framework.
+
+We will create a new repository `rust-components-swift` to serve as the root of the new Swift Package
+distribution. It will import the `application-services` repository as a git submodule. This will let
+us iterate quickly on the Swift packaging setup without impacting existing consumers.
```
badboy, line 78: Might be worth calling out how Glean factors in here.
While not strictly part of application-services, we of course want to continue shipping bundled with it.

Probably the submodule of glean in a-s would be removed and become a submodule of `rust-components-swift` directly.

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+
+The `rust-components-swift` repo will contain a `Package.swift` file that defines:
+
+* A single binary target that references the pre-built `.xcframework` bundle of Rust code.
+* One Swift target for each component, that references the Swift code from the git submodule
+  and depends on the pre-built Rust code.
+
+We will add automation to the `rust-components-swift` repo so that it automatically tracks
+releases made in the `application-services` repo and creates a corresponding git tag for
+the Swift package.
```
badboy, line 88: For now it's probably fine if we still follow the a-s releases. Letting Glean releases trigger new packages can be considered in the future.

docs/adr/0003-swift-packaging.md
```diff
@@ -0,0 +1,225 @@
...
+
+At some future date when all consumers have migrated to using Swift packages, we will remove
+the Carthage build setup from the application-services repo.
+
+At some future date, we will consider whether to move the `Package.swift` definition in to the application-services repo, or whether it's better to keep it separate.
```
badboy, line 93: Note that this will probably require changing how releases work anyway, because the `Package.swift` needs the checksum before tagging it.

badboy (commented):
Left some comments.
I think it would be good if this spells out that Glean will be eventually included in such a build.

> This is awesome, I'm 100% on-board with the multiple Swift packages option from the Nimbus perspective! One question, since Glean already supports Swift packages, what does this mean for the Glean submodule that is currently included in A-S? Will this complicate the build process for Glean if we are using the submodule for Android builds and the Glean-provided (external) Swift package for iOS?

Glean's Swift package support was built for that exact purpose: to support the new consumers.
Eventually I want Glean & a-s to ship bundled again.

We don't use a submodule for Android builds. Right now Glean Kotlin is shipped separately. Eventually this will be shipped together with a-s through m-c.

For iOS see inline comment.

docs/adr/0003-swift-packaging.md
badboy: Oh one more thing I forgot to mention:
Building for the iOS simulator on Arm machines (M1) currently requires Nightly.
I got approval for that target to become Tier 2, thus shipping the precompiled standard library and therefore making it possible on stable to compile for it, but that hasn't landed yet.

docs/adr/0003-swift-packaging.md
rfk: 😆 such fixes are always welcome!

docs/adr/0003-swift-packaging.md
rfk: Yes, here "the Rust code" means "the Rust code for all components, compiled together into a single big blob". We expect this will actually be OK in practice because we're building it as a staticlib, so when it gets incorporated into the final application build, the compiler will only pull in the parts of that lib that it needs.

We've had quite a few discussions about being able to distribute different components with separately-compiled chunks of the Rust code, but it's tricky because the components sometimes want to talk to each other directly, and in order to do that they need to be compiled into the same artifact.

I have an imaginary world in my head where the components can each be distributed as little `.so` files that call each other through their C FFI layer, but that world is a long way from being reality and I don't think we have any concrete plans to head in that direction.

docs/adr/0003-swift-packaging.md
rfk: I will call this out explicitly under "negative consequences" as a risk, thanks.

docs/adr/0003-swift-packaging.md
rfk: For Glean, I think we would still want to keep the submodule in `application-services` in order to include it in the `.xcframework` bundle, and then *also* include it in the `rust-components-swift` repo for publishing the Swift code. I also think we can safely defer that decision, because for now we can consume Glean via its standalone Swift package.

rfk: (but I'm adding a comment to this section about future possibilities, for completeness)

rfk:
Thanks folks! I've added some more notes to the doc to capture the points raised in discussion here, and will mark this as "accepted".

rfk:
>  One question, since Glean already supports Swift packages, what does this mean for the Glean submodule that is currently included in A-S?

I think we can and should consume Glean as an external Swift module *at the Swift level*, if only from a code hygiene perspective. Right now the Nimbus Swift bindings are using a bunch of stuff from Glean that are "private" as far as the Swift module declarations are concerned, and are only visible to Nimbus because we compile them both together in the appservices iOS megazord. I filed https://mozilla-hub.atlassian.net/browse/SYNC-2677 for followup on that front.

Orthogonally, we can include the Glean Rust code as part of the `.xcframework` build proposed in this ADR, and can publish a Glean Swift module as part of the proposed `rust-components-swift` package rather than (or in addition to) a standalone `glean-swift` package. I expect the timing of whether and when we make that change can be safely driven by the Glean team and/or our desire to finally start calling Glean directly from Rust, without changing the shape of anything else here.

So I suggest concretely:

* We keep the Glean submodule that is in A-S right now, because we need to continue publishing it for firefox-ios.
* Initially, we do not include Glean in the new `.xcframework` bundle, and have our Swift consumers depend on it as an external Swift package.
* Once we're happy that that's all working as intended, we can take a look at publishing Glean as a product in the `rust-components-swift` repo rather than as a standalone Swift package.

rfk:
(I should add, I am deliberately proposing a generic `rust-components-swift` name for the new repo in order to make it feel welcoming to any Rust component, not just the things that happen to publish through application-services).

PR #4312, opened 2021-06-25
Added a doc that describes the general syncing process

bendk:
This one contains my current picture of how syncing works in general.  I think it could be useful for someone starting up.  Is there anything incorrect here?  Could it be presented in a clearer way?

  - Note:

### DIFF
--- a/docs/sync-overview.md
+++ b/docs/sync-overview.md
@@ -43,7 +43,7 @@ This document provides a high-level overview of how syncing works.  **Note**: ea
    - The mirror table stores the last known record from the server
    - The staging temporary table stores the incoming records that we're currently processing
    - The local/mirror/staging tables contains a `guid` as its primary key.  A record will share the same `guid` for the local/mirror/staging table.
-   - The metadata table stores the GUID for the collection as a whole and the timestamp of the last sync
+   - The metadata table stores the GUID for the collection as a whole and the the last-known server timestamp of the collection.
 
 ### `apply_incoming` stages
   - **stage incoming**: write out each incoming server record to the staging table
--- a/docs/sync-overview.md
+++ b/docs/sync-overview.md
@@ -1,0 +1,57 @@
+# Sync Overview
+
+This document provides a high-level overview of how syncing works.  **Note**: each component has its own quirks and will handle sync slightly differently than the general process described here.
+
+## General flow and architecture
+
+- **Crates involved**:
+  - The `sync15` and `support/sync15-traits` handle the general syncing logic and define the `SyncEngine` trait
+  - Individual component crates (`logins`, `places`, `autofill`, etc).  These implement `SyncEngine`.
+  - `sync_manager` manages the overall syncing process.
+- **High level sync flow**:
+  - Sync is initiated by the application that embeds application-services.
+  - The application calls `SyncManager.sync()` to start the sync process.
+  - `SyncManager` creates `SyncEngine` instances to sync the individual components
+
+### Sync manager
+
+[`SyncManager`](https://github.com/mozilla/application-services/blob/main/components/sync_manager/src/manager.rs) is responsible for performing the high-level parts of the sync process:
+  - The consumer code calls it's `sync()` function to start the sync, passing
+    in a [`SyncParams`](https://mozilla.github.io/application-services/rust-docs/sync_manager/msg_types/struct.SyncParams.html) object in, which describes what should be synced.
+  - `SyncManager` checks if we are currently in a backoff period and should wait before contacting the server again.
+  - `SyncManager` creates invididual `SyncEngine` instances for each component that should be synced.  We currently use 2 different methods for this:
+    - The older method is for the `SyncManager` to hold a weakref to a `Store` use that to create the `SyncEngine` (tabs and places).  The `SyncEngine` uses the `Store` for database access, see the [`TabsStore`](https://mozilla.github.io/application-services/rust-docs/tabs/struct.TabsStore.html) for an example.
+    - The newer method is for the components to provide a function to create the `SyncEngine`, hiding the details of how that engine gets created (autofill/logins).  These components also define a `Store` instance for the `SyncEngine` to use, but it's all transparent to the `SyncManager`.  (See [`autofill::get_registered_sync_engine()`](https://mozilla.github.io/application-services/rust-docs/autofill/db/store/fn.get_registered_sync_engine.html) and [`autofill::db::store::Store`](https://mozilla.github.io/application-services/rust-docs/autofill/db/store/struct.Store.html))
+  - For components that use local encryption, `SyncManager` passes the local encryption key to their `SyncEngine`
+  - Finally, calls `sync_multiple()` function from the `sync15` crate, sending it the `SyncEngine` instances.  `sync_multiple()` then calls the `sync()` function for each individual `SyncEngine`
+
+### Sync engines
+  - [`SyncEngine`](https://github.com/mozilla/application-services/blob/main/components/support/sync15-traits/src/engine.rs) is defined in the `support/sync15-traits` crate and defines the interface for syncing a component.
+  - A new `SyncEngine` instance is created for each sync
+  - `SyncEngine.apply_incoming()` does the main work.  It is responsible for processing incoming records from the server in order to update the local records and calculating which local records should be synced back.
+
+## The `apply_incoming` pattern
+
+`SyncEngine` instances are free to implement `apply_incoming()` any way they want, but the most components follow a general pattern.
+
+### Database Tables
+
+   - The local table stores records for the local application
+   - The mirror table stores the last known record from the server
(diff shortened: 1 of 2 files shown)


docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+  - `sync_manager` manages the overall syncing process.
+- **High level sync flow**:
+  - Sync is initiated by the application that embeds application-services.
+  - The application calls `SyncManager.sync()` to start the sync process.
+  - `SyncManager` creates `SyncEngine` instances to sync the individual components
+
+### Sync manager
+
+[`SyncManager`](https://github.com/mozilla/application-services/blob/main/components/sync_manager/src/manager.rs) is responsible for performing the high-level parts of the sync process:
+  - The consumer code calls it's `sync()` function to start the sync, passing
```
tarikeshaq, line 19: Can we link to the `SyncParams` somewhere? I think it would be useful to see what passes into the sync function

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+   - The local table stores records for the local application
+   - The mirror table stores the last known record from the server
+   - The staging temporary table stores the incoming records that we're currently processing
+   - Each table contains a `guid` as its primary key.  A record will share the same `guid` for the local/mirror/staging table.
+
+### `apply_incoming` stages
+  - **stage incoming**: write out each incoming server record to the staging table
+  - **fetch states**: take the rows from all 3 tables and combine them into a single struct containing `Option`s for the local/mirror/staging records.
+  - **iterate states**: loop through each state, decide how to do change the local records, then execute that plan.
+    - **reconcile/plan**: For each state we create an action plan for it.  The action plan is a low-level description of what to change (add this record, delete this one, modify this field, etc).  Here are some common situations:
```
tarikeshaq, line 48: I'm super curious about this, I imagine each component has its own magic that it does for figuring out how to sync in different cases...

Maybe for the future (or maybe this already exists somewhere!), It would be awesome to have short descriptions of how different components handle different states... It's a lotta work, not sure where it would fit and possibly just looking at the code is suffienct so we might be okay without it 🤷

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+  - `SyncManager` creates `SyncEngine` instances to sync the individual components
+
+### Sync manager
+
+[`SyncManager`](https://github.com/mozilla/application-services/blob/main/components/sync_manager/src/manager.rs) is responsible for performing the high-level parts of the sync process:
+  - The consumer code calls it's `sync()` function to start the sync, passing
+    in a `SyncParams` object in, which describes what should be synced.
+  - `SyncManager` checks if we are currently in a backoff period and should wait before contacting the server again.
+  - `SyncManager` creates invididual `SyncEngine` instances for each component that should be synced.  We currently use 2 different methods for this:
+    - The older method is for the `SyncManager` to store a weakref to the `Store` for each component and use that to create a `SyncEngine` (tabs and places)
```
tarikeshaq, line 23: I would love to see some context of what a `Store` is! Maybe a link to somewhere?

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
+# Sync Overview
+
+This document provides a high-level overview of how syncing works.  **Note**: each component has its own quirks and will handle sync slightly differently than the general process described here.
+
+## General flow and architecture
+
+- **Crates involved**:
+  - The `sync15` and `support/sync15-traits` handle the general syncing logic and define the `SyncEngine` trait
+  - `SyncEngine` is implemented by each component crate (login, places, autofill, etc).
```
tarikeshaq, line 9: super nit: I think this can be a secondary point under the first point since `SyncEngine` is not a crate

tarikeshaq (commented):
This was a good reading for me!

I haven't really interacted with Sync to this was a very beginner-friendly doc.

had a couple of comments about more context, but it generally LGTM.

(I'll keep the approval to someone who knows more about Sync 😆)

docs/sync-overview.md
bendk: Yeah this would be very nice to have.  If you look through the engines code it's usually pretty readable, but a couple rustdoc strings could go a long way and would also be linkable from this doc.

docs/sync-overview.md
bendk: I was thinking of the login/places/autofill crates, but it wasn't very clear.  I'll try to reverse it, maybe that will help.

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+  - **fetch states**: take the rows from all 3 tables and combine them into a single struct containing `Option`s for the local/mirror/staging records.
+  - **iterate states**: loop through each state, decide how to do change the local records, then execute that plan.
+    - **reconcile/plan**: For each state we create an action plan for it.  The action plan is a low-level description of what to change (add this record, delete this one, modify this field, etc).  Here are some common situations:
+       - **A record only appears in the staging table**.  It's a new record from the server and should be added to the local DB
+       - **A record only appears in the local table**.  It's a new record on the local instance and should be synced back to the serve
+       - **Identical records appear in the local/mirror tables and a newer record is in the staging table**.  The record was updated remotely and the changes should be propagated to the local DB.
+       - **A record appears in the mirror table and newer records appear in both the local and staging tables**.  The record was updated both locally and remotely and we should perform a 3-way merge.
```
mhammond, line 52: Note that "newer" in the above few sentences is slightly misleading - we try very hard to avoid comparing local and remote timestamps. Eg, the change counter is used exclusively to say whether a local record has changed or not. I'm not sure this is a big deal though.

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+    - **apply plan**: After we create the action plan, then we execute it.
+  - **fetch outgoing**:
+     - Calculate which records need to be sent back to the server
+     - Update the mirror table
+     - Return those records back to the `sync15` code so that it can sync them with the server.
```
mhammond, line 57: s/sync them with/upload them to/ - "sync them" kinda implies some additional logic, but there's none - we just POST exactly what we have been given.

Then a final bullet: * The sync15 code returns the timestamp reported by the server in the POST response and hands it back to the engine. The engine persists this timestamp in the `metadata` table - the next sync will then use this timestamp to fetch records that have since been changed by other devices

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+## The `apply_incoming` pattern
+
+`SyncEngine` instances are free to implement `apply_incoming()` any way they want, but the most components follow a general pattern.
+
+### Database Tables
+
+   - The local table stores records for the local application
+   - The mirror table stores the last known record from the server
+   - The staging temporary table stores the incoming records that we're currently processing
+   - Each table contains a `guid` as its primary key.  A record will share the same `guid` for the local/mirror/staging table.
```
mhammond, line 42: There's also a "metadata" table that stores the collection's guid and timestamp.

mhammond (approved):
This is great, thanks Ben. Feel free to take my suggestions literally or bash the words into something better, or even ignore them if you feel they might detract from the main points.

docs/sync-overview.md
```diff
@@ -0,0 +1,57 @@
...
+
```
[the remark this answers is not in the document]
mhammond: (implied in the above is that we don't hold a DB transaction open while doing network IO, so we can't rely on transactions and a simple bool to track this changed state - not sure if it's worth mentioning that or not :)

bendk:
@mhammond thanks for all the info.  I tried to integrate it in as best I could.

This doc is looking really nice to me at this point, should we merge it?

docs/sync-overview.md
```diff
@@ -0,0 +1,65 @@
...
+`SyncEngine` instances are free to implement `apply_incoming()` any way they want, but the most components follow a general pattern.
+
+### Database Tables
+
+   - The local table stores records for the local application
+   - The mirror table stores the last known record from the server
+   - The staging temporary table stores the incoming records that we're currently processing
+   - The local/mirror/staging tables contains a `guid` as its primary key.  A record will share the same `guid` for the local/mirror/staging table.
+   - The metadata table stores the GUID for the collection as a whole and the timestamp of the last sync
```
mhammond, line 46: This is probably my fault, but I think for clarity, change "timestamp of the last sync" to something like "the last-known server timestamp of the collection". The current wording implies that the timestamp will advance on every sync, which isn't true. This timestamp is used in, eg, `if-unmodified-since` headers and as a url param to fetch new records created since we were last here (and on most syncs we will find neither us nor anything else wrote to the server since the last sync, so this timestamp remains unchanged)

mhammond (approved):
This is great, thanks.

skhamis (approved):
Love the document! Definitely a good thing to have as almost all our components will soon use the sync manager. LGTM!

PR #4314, opened 2021-06-25
[SDK-239] Create a Nimbus iOS sample app

travis79:
This adds a standalone example app, currently relying on the Swift Package for Nimbus residing at https://github.com/rfk/rust-components-swift, which may need to be updated once this build output is landed in tree in A-S


travis79:
This is still a draft as I would like to 
- add instrumentation tests
- add test UI elements similar to Firefox iOS
- add some Glean instrumentation
- etc.

jhugman:
This is > 12 months out of date. What say we close this, @travis79 and open it if we get back to it?

jhugman (changes requested):
This is somewhat out of date. Requesting changes so it drops out of my review queue.

PR #4322, opened 2021-07-06
Updating the prepare-release script

bendk:
Fixed the PR URL when using a forked repo

  - Note:

### DIFF
--- a/automation/prepare-release.py
+++ b/automation/prepare-release.py
@@ -40,6 +40,22 @@ CHANGELOG_FILE = "CHANGELOG.md"
 
 ensure_working_tree_clean()
 
+step_msg(f"Calculating remote owner {remote}")
+result = run_cmd_checked(["git", "remote", "get-url", remote], stdout=subprocess.PIPE)
+repo_url = result.stdout.decode('utf8').strip()
+repo_owner_patterns = [
+    r"git@github.com:(.*?)/application-services.git",
+    r'https://github.com/(.*?)/application-services.git',
+]
+for pattern in repo_owner_patterns:
+    match = re.match(pattern, repo_url)
+    if match:
+        repo_owner = match.group(1)
+        break
+else:
+    fatal_err(f"Error parsing remote URL: {repo_url}")
+print(f"Parsed github owner {repo_owner} for {repo_url}")
+
 step_msg(f"Updating remote {remote}")
 run_cmd_checked(["git", "remote", "update", remote])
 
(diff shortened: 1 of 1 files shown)


automation/prepare-release.py
```diff
@@ -40,6 +40,16 @@
 
 ensure_working_tree_clean()
 
+step_msg(f"Calculating remote owner {remote}")
+result = run_cmd_checked(["git", "remote", "get-url", remote], stdout=subprocess.PIPE)
+repo_url = result.stdout.decode('utf8').strip()
+print(repo_url)
```
tarikeshaq, line 46: super nit: something like: `f"Remote repo url is: {repo_url}"` can help in case someone's URL is off

automation/prepare-release.py
```diff
@@ -40,6 +40,16 @@
...
+match = re.match(r"git@github.com:(.*?)/application-services.git", repo_url)
```
tarikeshaq, line 47: Do we want to also support `https` here?  Not sure if it even works now, so ignore this if it already doesn't

On that note, if we decide that we only support `ssh` remote url's we might wanna add a quick note about it in the `cutting a release doc` - again if that already exists, ignore this 😬

tarikeshaq (commented):
Looks mostly good to me, a couple of tiny clarifications - but I do think this is useful for anyone working on a fork 😄

rfk:
I don't object to this, but I'm curious under what circumstances we would *want* to cut a release against a forked repo rather than the canonical one. For testing purposes?

bendk:
> I don't object to this, but I'm curious under what circumstances we would _want_ to cut a release against a forked repo rather than the canonical one. For testing purposes?

It's for regular releases and the idea is just to save keystrokes by not having to type in the `--remote` option.  Maybe this is a bad idea though.  The difference is the `cut.X.Y.Z` branch gets stored on the forked repo, not `mozilla/application-services`.  I was assuming that didn't matter, but that could definitely be wrong.

bendk:
> I don't object to this, but I'm curious under what circumstances we would _want_ to cut a release against a forked repo rather than the canonical one. For testing purposes?

It's for regular releases and the idea is just to save keystrokes by not having to type in the `--remote` option.  Maybe this is a bad idea though.  The difference is the `cut.X.Y.Z` branch gets stored on the forked repo, not `mozilla/application-services`.  I was assuming that didn't matter, but that could definitely be wrong.

rfk:
Ah, that makes sense, thanks! Making the PR from a forked repo seems fine to me if it fits into folks default workflow 👍🏻

rfk (approved):

bendk:
Cool, I just updated the commit based on the suggestions from @tarikeshaq.  Do they seem okay to you?

tarikeshaq (approved):
🚀

PR #4324, opened 2021-07-07
Avoids evaluating enrollment when opted in experiment

tarikeshaq:
fixes https://mozilla-hub.atlassian.net/browse/SDK-317

Still a work in progress, mostly to verify that the issue is fixed on a local build of android/ios.

I'll drop a couple of comments in the code for areas that I'm a little confused about too

  - Note:

### DIFF
--- a/components/nimbus/src/enrollment.rs
+++ b/components/nimbus/src/enrollment.rs
@@ -169,7 +169,11 @@ impl ExperimentEnrollment {
                     updated_enrollment
                 }
             }
-            EnrollmentStatus::Enrolled { ref branch, .. } => {
+            EnrollmentStatus::Enrolled {
+                ref branch,
+                ref reason,
+                ..
+            } => {
                 if !is_user_participating {
                     log::debug!(
                         "Existing experiment enrollment '{}' is now disqualified (global opt-out)",
@@ -185,6 +189,10 @@ impl ExperimentEnrollment {
                         self.disqualify_from_enrolled(DisqualifiedReason::Error);
                     out_enrollment_events.push(updated_enrollment.get_change_event());
                     updated_enrollment
+                } else if matches!(reason, EnrolledReason::OptIn) {
+                    // we check if we opted-in an experiment, if so
+                    // we don't need to update our enrollment
+                    self.clone()
                 } else {
                     let evaluated_enrollment = evaluate_enrollment(
                         nimbus_id,
--- a/components/nimbus/tests/test_restart.rs
+++ b/components/nimbus/tests/test_restart.rs
@@ -1,0 +1,126 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+mod common;
+#[cfg(feature = "rkv-safe-mode")]
+#[cfg(test)]
+mod test {
+
+    use super::common::new_test_client_with_db;
+    #[cfg(feature = "rkv-safe-mode")]
+    use nimbus::error::Result;
+    use serde_json::json;
+    use tempdir::TempDir;
+
+    #[test]
+    fn test_restart_opt_in() -> Result<()> {
+        let temp_dir = TempDir::new("test_restart_opt_in")?;
+        let client = new_test_client_with_db(&temp_dir)?;
+        client.initialize()?;
+        let experiment_json = serde_json::to_string(&json!({
+            "data": [{
+                "schemaVersion": "1.0.0",
+                "slug": "secure-gold",
+                "endDate": null,
+                "featureIds": ["some-feature"],
+                "branches": [
+                    {
+                    "slug": "control",
+                    "ratio": 1
+                    },
+                    {
+                    "slug": "treatment",
+                    "ratio": 1
+                    }
+                ],
+                "channel": "nightly",
+                "probeSets": [],
+                "startDate": null,
+                "appName": "fenix",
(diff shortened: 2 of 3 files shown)


components/nimbus/src/enrollment.rs
```diff
@@ -185,6 +189,10 @@ impl ExperimentEnrollment {
                         self.disqualify_from_enrolled(DisqualifiedReason::Error);
                     out_enrollment_events.push(updated_enrollment.get_change_event());
                     updated_enrollment
+                } else if matches!(reason, EnrolledReason::OptIn) {
+                    // we check if we opted-in an experiment, if so
+                    // we don't need to update our enrollment
+                    self.clone()
```
tarikeshaq, line 195: Hmmm I'm still slightly unsure about this, because if we go through the `evaluate_enrollment` codepath, shouldn't we still end up on the `self.clone()` in line 213?  Which means that the enrollment status and reason would have been preserved.

I still need to get familiarized a bit more with the code too, so I could be missing something

components/nimbus/src/enrollment.rs
```diff
@@ -2472,6 +2480,69 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_evolver_opt_in() -> Result<()> {
+        let existing_enrollments = vec![
```
tarikeshaq, line 2485: The tests pass even without the change, which is mildly concerning. This is because of what I noted above, we end up returning the `OptIn` enrollment even though we evaluate the enrollment.

components/nimbus/tests/test_restart.rs
```diff
@@ -0,0 +1,40 @@
...
+#[cfg(test)]
+mod test {
+
+    use super::common::new_test_client_with_db;
+    #[cfg(feature = "rkv-safe-mode")]
+    use nimbus::error::Result;
+    use tempdir::TempDir;
+
+    #[test]
+    fn test_restart_opt_in() -> Result<()> {
```
tarikeshaq, line 16: This tests the restart scenario as close as possible - I'm still a bit fuzzy about the details of `apply_pending_experiments`, but I think I get the big idea. 

This also passes without the change because of the same reason.

tarikeshaq (commented):
I'll take this off draft once I get a fenix build with this code and see if I still get the same issue, noted down a couple of notes.

My biggest confusion is what happens after we `evaluate_enrollment` on the case where the user is already enrolled, it seems that instead of returning the new `evaluated_enrollment` we return `self.clone`, I'm still a bit all over the place with my knowledge here though 😬

tarikeshaq:
Ran a quick test with a local Fenix build on this, seems that it is preserving the opt in after restart. Haven't tested IOS yet, but I'll get on that - in the mean time made this "ready for review"

components/nimbus/tests/test_restart.rs
jhugman: > This also passes without the change because of the same reason.

🧙 If you can't reproduce the problem, how do you know you've fixed it? 🧙‍♂️

components/nimbus/tests/test_restart.rs
```diff
@@ -0,0 +1,40 @@
...
+        let temp_dir = TempDir::new("test_restart_opt_in")?;
+        let client = new_test_client_with_db(&temp_dir)?;
+        client.initialize()?;
+        client.fetch_experiments()?;
```
jhugman, line 20: I think I'd use `set_experiments_locally(payload: string)` to get couple of experiments which you can cause trigger the issue.

I think the `secure-gold` experiment in the `test/experiments` folder is set to always enroll: you need an experiment that won't enroll naturally.

You _might_ be able add another file to the `tests/experiments` directory, but this might be more trouble than it's worth.

components/nimbus/tests/test_restart.rs
```diff
@@ -0,0 +1,40 @@
...
+        client.apply_pending_experiments()?;
+        client.opt_in_with_branch("secure-gold".into(), "treatment".into())?;
+        let before_restart_experiments = client.get_active_experiments()?;
+        assert_eq!(before_restart_experiments.len(), 1);
+        assert_eq!(before_restart_experiments[0].branch_slug, "treatment");
+        // we drop the NimbusClient to terminate the underlying database connection
+        drop(client);
+        let client = new_test_client_with_db(&temp_dir)?;
+        client.initialize()?;
+        client.fetch_experiments()?;
```
jhugman, line 30: Use the `set_experiments_locally()` as you will above.

jhugman (commented):
Like you, I'm not convince you've fixed the bug.

I've given you some hints on how you might reproduce the bug. Once you've done that we'll be able to say whether or not your fix works or not.

components/nimbus/tests/test_restart.rs
jhugman: I think you'll need one experiment that has `"targeting": "false"` and one with a `bucketConfig` as:

```json
  "bucketConfig": {
    "count": 0,
    "start": 0,
    "total": 10000,
    "namespace": "secure-gold",
    "randomizationUnit": "nimbus_id"
  },
```

i.e. 0 out of 10000 targeted devices will be enrolled into the experiment.

The `targeting` *string* is a JEXL string which evaluates to a boolean. If it evaluates to false then the client is not enrolled into the experiment.

jhugman (approved):
This is really good. Well done. 🎉 

Please find the big green button ✅ , and proceed to land. 🛩️

tarikeshaq:
Our dependency check CI is failing for an unrelated reason to this change, opened an issue to address that #4328 - will merge this once the other CI's pass after the rebase!

PR #4325, opened 2021-07-08
Eliminate outdated UAID's from push component

tarikeshaq:
Relates to #3695 

For context: https://github.com/mozilla/application-services/issues/3695#issuecomment-876639532

Basically, whenever we unsubscribe from **all** channels, we also want to make sure we dump the UAID on the component and all associated records, the server does that anyways. The theory behind the problem we were getting is that we keep an invalid `UAID` in memory, then send it over to the server, the server rightfully so complains because we already told it that the UAID is invalid.

This change will hopefully force us to re-generate a new UAID on the next subscribe instead of using a stale one.

Keeping as a draft until I add a quick test in the component and give it a quick run to make sure nothing big broke

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -1,10 +1,6 @@
 **See [the release process docs](docs/howtos/cut-a-new-release.md) for the steps to take when cutting a new release.**
 
 # Unreleased Changes
-## Nimbus
-### What's changed
-  - Fixed a bug where opt-in enrollments in experiments were not preserved when the application is restarted ([#4324](https://github.com/mozilla/application-services/pull/4324))
-  - The nimbus component now specifies the version of the server's api - currently V1. That was done to avoid redirects. ([#4319](https://github.com/mozilla/application-services/pull/4319))
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v80.0.1...main)
 
@@ -22,7 +18,15 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+## Nimbus
+### What's changed
+  - Fixed a bug where opt-in enrollments in experiments were not preserved when the application is restarted ([#4324](https://github.com/mozilla/application-services/pull/4324))
+  - The nimbus component now specifies the version of the server's api - currently V1. That was done to avoid redirects. ([#4319](https://github.com/mozilla/application-services/pull/4319))
+
+## Push
+### What's changed
+  - Fixed a bug where we don't delete a client's UAID locally when it's deleted on the server ([#4325](https://github.com/mozilla/application-services/pull/4325))
 
 ## Tabs
-
-Tabs has been Uniffi-ed!
+### ⚠️ Breaking Changes ⚠️
+ - Tabs has been Uniffi-ed! ([#4192](https://github.com/mozilla/application-services/pull/4192))
--- a/components/push/src/communications.rs
+++ b/components/push/src/communications.rs
@@ -65,7 +65,7 @@ pub trait Connection {
     ) -> error::Result<RegisterResponse>;
 
     /// Drop an endpoint
-    fn unsubscribe(&self, channel_id: Option<&str>) -> error::Result<bool>;
+    fn unsubscribe(&mut self, channel_id: Option<&str>) -> error::Result<bool>;
 
     /// Update the autopush server with the new native OS Messaging authorization token
     fn update(&mut self, new_token: &str) -> error::Result<bool>;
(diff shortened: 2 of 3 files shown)


jrconlin:
Ah, YES! Excellent catch!

The server does, indeed, toss the UAID as being invalid when an unsubscribe event happens for a UAID (This is differentiated from a CHID unsubscribe, which preserves the UAID). Trying to resubscribe with the prior UAID would fail in that case. 

(Also, this feels like it's a bug that I probably introduced when I first wrote the code. I'll figure out what form of penance I'll take later. Thanks for fixing this.)

components/push/src/subscriber.rs
```diff
@@ -101,19 +100,19 @@ impl PushManager {
     }
 
     // XXX: maybe -> Result<()> instead
-    pub fn unsubscribe(&self, channel_id: Option<&str>) -> Result<bool> {
+    pub fn unsubscribe(&mut self, channel_id: Option<&str>) -> Result<bool> {
         if self.conn.uaid.is_none() {
             return Err(ErrorKind::GeneralError("No subscriptions created yet.".into()).into());
         }
-        let uaid = self.conn.uaid.as_ref().unwrap();
+        let uaid = self.conn.uaid.as_ref().unwrap().to_owned();
```
rfk, line 107: Out of curiosity, why is the `to_owned()` necessary here?

rfk (commented):
Nice catch @tarikeshaq! What you're proposing here makes sense to me, although I think we should probably add a couple of tests for completeness.

components/push/src/subscriber.rs
tarikeshaq: Oh It's because I'm changing `ConnectHttp.unsubscribe` to take a mutable reference, so the borrow checker will complain because otherwise the `uaid` would be an immutable reference used in `delete_record`

In retrospect though, I can avoid the clone by moving this whole line into the `self.store.delete_record` so I'll do that 😄

tarikeshaq:
@rfk if this looks good to you, I'll have it as a part of the release I'm cutting  😄 - I added a few tests on both the `PushManager` and `Connection` layers, also gave send-tab a quick run on a local fenix build and it seems OK, but if there are any red flags we can delay this

CHANGES_UNRELEASED.md
```diff
@@ -22,7 +18,15 @@ Use the template below to make assigning a version number during the release cut
...
 
-## Tabs
+## Push
+### What's changed
+  - Fixed a bug where we don't delete a client's UAID locally when it's deleted on the server ([#4325](https://github.com/mozilla/application-services/pull/4325))
 
-Tabs has been Uniffi-ed!
+## Tabs
+### ⚠️ Breaking Changes ⚠️
+ - Tabs has been Uniffi-ed! ([#4192](https://github.com/mozilla/application-services/pull/4192))
```
rfk, line 32: I know this isn't your change, but I want to flag it here for maybe fixing up when we cut the release. Changelogs are for consumers, and *especially* in the case of a breaking change we should put our "consumer" had on and think about what the consumer would like to have called out in the changelog. Specifically in this case, I think we should say more words that consumers should expect e.g. some name changes in the API surface because the bindings are now auto-generated. We don't have to list out all the individual changes, but it's useful to equip consumers with some high-level expectations of what they need to watch out for.

components/push/src/communications.rs
```diff
@@ -274,7 +274,12 @@ impl Connection for ConnectHttp {
             .headers(self.headers()?)
             .send()
         {
-            Ok(_) => Ok(true),
+            Ok(_) => {
+                if channel_id.is_none() {
+                    self.uaid = None;
+                }
```
rfk, line 280: Not for this PR, but something to link about more architecturally for this component - if the `DELETE` request succeeds but then we fail to persist this `self.uaid` change to disk due to e.g. a crash or unrelated error, then we'll still be in the same position of disagreement between client and server and will need to be prepared to deal with it.

This is definitely still a big improvement over the previous behaviour!

components/push/src/communications.rs
```diff
@@ -596,5 +604,98 @@ mod test {
...
+            .with_status(200)
+            .with_header("content-type", "application/json")
+            .with_body(body)
+            .create();
+            conn.subscribe(&channel_id, None).unwrap();
+            ap_mock.assert();
+            // we verify that the UAID is the new one we got from
+            // the server
+            assert_eq!(conn.uaid.unwrap(), DUMMY_UAID2);
+        }
```
rfk, line 699: 👍🏻 thanks for adding this!

rfk (approved):
🚢 it 👍🏻

CHANGES_UNRELEASED.md
rfk: /cc @skhamis just for visibility if there was anything specific you wanted to add in the changelog entry here.

PR #4327, opened 2021-07-09
PrivateSendTabKeys should encode commands_data as base64

### LINKED ISSUE #3999: Sendtab keys are serialized as stringified integer arrays
I happened to by poking around at the fxa-client example app (`cargo run --example fxa-devices-api`) and inspected the `credentials.json` that it creates after you sign in. Here's the (redacted) contents:

```
{
   "last_handled_command" : null,
   "scoped_keys" : {
      "https://identity.mozilla.com/apps/oldsync" : {
         "kid" : "1600128535160-CX62z6WifHPIUK8UKGRk3A",
         "scope" : "https://identity.mozilla.com/apps/oldsync",
         "k" : "<redacted>",
         "kty" : "oct"
      }
   },
   "current_device_id" : "7da9369643c8962d8361ae191b8289d4",
   "in_flight_migration" : null,
   "access_token_cache" : {},
   "schema_version" : "V2",
   "session_token" : null,
   "refresh_token" : {
      "token" : "<redacted>",
      "scopes" : [
         "https://identity.mozilla.com/apps/oldsync",
         "profile"
      ]
   },
   "last_seen_profile" : null,
   "config" : {
      "content_url" : "https://accounts.firefox.com",
      "token_server_url_override" : null,
      "redirect_uri" : "https://accounts.firefox.com/oauth/success/a2270f727f45f648",
      "client_id" : "a2270f727f45f648"
   },
   "device_capabilities" : [
      "SendTab"
   ],
   "commands_data" : {
      "https://identity.mozilla.com/cmd/open-uri" : "{\"V1\":{\"p256key\":{\"curve\":\"P256\",\"private_key\":[228,95,185,23,134,90,245,21,155,232,21,93,120,243,55,203,86,175,250,218,73,187,200,203,124,55,63,139,234,255,20,38],\"public_key\":[4,63,70,241,111,95,19,127,97,249,238,254,41,153,68,242,72,46,207,191,249,14,2,188,129,232,195,251,229,61,196,32,246,232,218,81,202,195,185,172,231,42,156,70,196,13,177,59,148,212,70,91,7,106,80,145,254,172,27,34,133,12,153,28,222]},\"auth_secret\":[155,176,194,179,53,253,78,159,250,92,40,145,1,105,154,64]}}"
   }
}
```

That "commands_data" blob looks unfortunate - the binary data for the "private_key", "public_key" and "auth_secret" fields seems to have been encoded as a JSON array literal for the raw bytes rather than, say, as a base64 string.

I don't thing anything outright _bad_ will happen as a result of this, but it's likely to be inefficient to load/save and generally makes it look like we haven't been very careful with the serialization code here :-/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2619)
┆Story Points: 3
┆Epic: Important backlog
┆Sprint End Date: 2021-07-09

skhamis:
fixes #3999

After looking through the code and the example, it seems like the general trend is to use base64 crate to encode/decode the keys when necessary. I took this approach and can validate and send tabs successfully so it seems to be the right thing to do? Definitely would like feedback if this is the right approach.

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/commands/send_tab.rs
+++ b/components/fxa-client/src/internal/commands/send_tab.rs
@@ -115,13 +115,15 @@ impl PrivateSendTabKeys {
     // `serde_json::to_string` would compile because both types derive
     // `Serialize`.
     pub(crate) fn serialize(&self) -> Result<String> {
-        Ok(serde_json::to_string(&VersionnedPrivateSendTabKeys::V1(
-            self.clone(),
-        ))?)
+        Ok(base64::encode_config(
+            serde_json::to_string(&VersionnedPrivateSendTabKeys::V1(self.clone()))?,
+            base64::URL_SAFE_NO_PAD,
+        ))
     }
 
     pub(crate) fn deserialize(s: &str) -> Result<Self> {
-        let versionned: VersionnedPrivateSendTabKeys = serde_json::from_str(s)?;
+        let decoded = base64::decode(s)?;
+        let versionned: VersionnedPrivateSendTabKeys = serde_json::from_slice(&decoded)?;
         match versionned {
             VersionnedPrivateSendTabKeys::V1(prv_key) => Ok(prv_key),
         }


rfk (commented):
Using Base64 definitely sounds like the way to go, but I think we'll need to take some care here to migrate any existing data from the old format to the new, otherwise send-tab might break for already-signed-in-clients. In practice that probably means:

* Introducing a `VersionnedPrivateSendTabKeys::V2` variant holds the same data as the `V1` variant, but knows how to write out the keys as base64 rather than an integer array.
* Adding code to forward-migrate from `V1` to `V2` in the `deserialize` method, to handle loading data from already-signed-in clients.

The migration code from the FxA [state_persistence.rs](https://github.com/mozilla/application-services/blob/3f125205fc20d3260e98e0f3073595f58cd333cf%5E1/components/fxa-client/src/internal/state_persistence.rs#L29) may be useful as a reference.

rfk:
(In particular the FxA state-persistence code has some good examples of the kind of [tests](https://github.com/mozilla/application-services/blob/fea3d593dda942dd1a9fc77dfdae6b48533379cc/components/fxa-client/src/internal/state_persistence.rs#L249) we'll want to write to ensure that the migration works against old data).

skhamis:
> Using Base64 definitely sounds like the way to go, but I think we'll need to take some care here to migrate any existing data from the old format to the new, otherwise send-tab might break for already-signed-in-clients. In practice that probably means:
> 
>     * Introducing a `VersionnedPrivateSendTabKeys::V2` variant holds the same data as the `V1` variant, but knows how to write out the keys as base64 rather than an integer array.
> 
>     * Adding code to forward-migrate from `V1` to `V2` in the `deserialize` method, to handle loading data from already-signed-in clients.
> 
> 
> The migration code from the FxA [state_persistence.rs](https://github.com/mozilla/application-services/blob/3f125205fc20d3260e98e0f3073595f58cd333cf%5E1/components/fxa-client/src/internal/state_persistence.rs#L29) may be useful as a reference.

I wonder if maybe I'm thinking about this wrong. Since we're not changing the fields/types or schema and changing only the internals of the serialize/deserialize. Would we still have to do a migrate? I do agree in the deserialize we'll need to catch if it's incoming as a non base64 and parse accordingly.

I also probably had a misunderstanding that credentials.json doesn't get re-built and gets modified when running the `cargo run --example fxa-devices-api`

rfk:
> I also probably had a misunderstanding that credentials.json doesn't get re-built and gets modified when running the cargo
> run --example fxa-devices-api

Oh! I'm sorry, I did not properly surface an important piece of context in the original issue. While the easiest way to produce this data locally is using the example app, this data is not specific to the examples - it's the same JSON format that is used for persisting account data in our clients in the wild, via the `to_json` and `from_json` methods on the `FirefoxAccount` struct. So any change to the data format here will also be a change to the handling of persisted data from signed-in clients in the wild.

You're right that it may be possible to handle the backwards-compatibility by hand when deserializing, rather than adding a full-blown migration, that's definitely worth a try. But we'll still need explicit tests for handling loading of the old format.

rfk:
Actually, what would be useful here, is to generate a new "state v2" test vector [like this one](https://github.com/mozilla/application-services/blob/65c88a630a5d95a1580a67d1e3a6d5b4915fd2ad/components/fxa-client/src/internal/state_persistence.rs#L153) that has a snapshot of how we're currently storing device-related data, to make sure we don't regress it with future changes.  IIUC, the change made in this PR would have risked deserializing errors on existing clients, so the tests in that file *should* have caught it. They didn't, because the JSON snapshots used in the tests don't currently include device information.

tarikeshaq:
@skhamis this has gotten stale so I'll close it - the ticket is still open if we come back to it

Issue #4200, opened 2021-06-10
Resolve Fenix breaking changes for sqlcipher remove


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2246)
┆Story Points: 8
┆Epic: Migrate logins away from SQLCipher
┆Sprint: SYNC - 2021-06-25

Issue #4199, opened 2021-06-10
Resolve ios breaking changes for sqlcipher removal


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2247)
┆Story Points: 8
┆Epic: Migrate logins away from SQLCipher
┆Sprint: SYNC - 2021-06-25

Issue #4091, opened 2021-05-12
Update logins API to maintain validation/dupe handling


data-sync-user:
Update the logins API in order to maintain support for the validation/dupe handling for per-field encryption. As specified in [ADR-0001](https://github.com/mozilla/application-services/blob/main/docs/adr/0001-update-logins-api.md), this work will include the following:

- Replacing login `add` and `update` functions with a new `add_or_update` function. The following should be true of this new function:
  - Will take the encryption key as a parameter
  - Needs to support accurate population of the `time_password_changed` field for login updates
  - Should call `fixup_and_check_for_dupes` as both `add` and `update` currently do
  - Should replicate a call to `potential_dupes_ignoring_username` to distinguish add and update login operations [as occurs in a-c](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt#L37)
- Removing `check_valid_with_no_dupes` and `potential_dupes_ignoring_username` functions
  - The logic in these functions should be moved to the new `add_or_update` function
- Adding an encryption key parameter to `import_multiple` calls 
- Fixing all broken tests and adding tests where sensible

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2290)
┆Epic: Logins update and sqlcipher migration
┆Fix Versions: Release 93
┆Sprint End Date: 2021-09-03

data-sync-user:
➤ Lougenia Bailey commented:

Status update: Momentarily on hold for the uniffi ticket (UNIFFI-46). Once the uniffi ticket is complete, I’ll start implementing the logins api changes proposed in api shape I’ve solicited feedback for.

PR #4332, opened 2021-07-12
[Will Not Merge] Logins API Shape

lougeniaC64:
TL;DR This branch is a tool, it will not be merged.

I created this PR as a means to socialize [the updated Logins API](https://github.com/mozilla/application-services/blob/8c5950031b256ad44ad6e66d833dd114f83f6349/components/logins/src/logins.udl) and provide a stable branch for the work required for issues #4199 and #4200. There are a lot of dubious code changes here (not to mention commented out tests 😵) that were made solely to allow parallelizing the logins migration, API, and consumer code updates. I will make [the necessary logins API changes](#4091) in earnest in a forthcoming PR. Once I receive initial feedback on this API shape, I'll close this PR leaving the `logins-api-shape` branch available for the aforementioned consumer work.

  - Note:

### DIFF
--- a/components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
+++ b/components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
@@ -49,9 +49,9 @@ class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable {
 
     @Synchronized
     @Throws(LoginsStorageException::class)
-    fun unlock(encryptionKey: String) {
+    fun unlock(@Suppress("UNUSED_PARAMETER") encryptionKey: String) {
         return unlockCounters.measure {
-            val store = LoginStore(dbPath, encryptionKey)
+            val store = LoginStore(dbPath)
             if (this.store.getAndSet(store) != null) {
                 // this seems wrong?
                 throw LoginsStorageErrorException.MismatchedLock("Unlock called when we are already unlocked")
@@ -111,6 +111,13 @@ class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable {
     }
 
     @Throws(LoginsStorageException::class)
+    fun decryptAndFixupLogin(encKey: String, login: Login): DecryptedLogin {
+        return readQueryCounters.measure {
+            checkUnlocked().decryptAndFixupLogin(encKey, login)
+        }
+    }
+
+    @Throws(LoginsStorageException::class)
     fun list(): List<Login> {
         return readQueryCounters.measure {
             checkUnlocked().list()
@@ -125,44 +132,17 @@ class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable {
     }
 
     @Throws(LoginsStorageException::class)
-    fun add(login: Login): String {
+    fun addOrUpdate(encKey: String, login: LoginFields): String {
         return writeQueryCounters.measure {
-            checkUnlocked().add(login)
+            checkUnlocked().addOrUpdate(encKey, login)
         }
     }
 
     @Throws(LoginsStorageException::class)
-    fun importLogins(logins: List<Login>): String {
+    fun importLogins(encKey: String, logins: List<Login>): String {
         return writeQueryCounters.measure {
-            checkUnlocked().importMultiple(logins)
+            checkUnlocked().importMultiple(encKey, logins)
         }
-    }
-
-    @Throws(LoginsStorageException::class)
-    fun update(login: Login) {
-        return writeQueryCounters.measure {
-            checkUnlocked().update(login)
-        }
-    }
-
-    @Synchronized
-    @Throws(LoginsStorageException::class)
-    fun potentialDupesIgnoringUsername(login: Login): List<Login> {
-        return readQueryCounters.measure {
-            checkUnlocked().potentialDupesIgnoringUsername(login)
-        }
-    }
-
-    @Throws(LoginsStorageErrorException.InvalidRecord::class)
-    fun ensureValid(login: Login) {
-        readQueryCounters.measureIgnoring({ e -> e is LoginsStorageErrorException.InvalidRecord }) {
-            checkUnlocked().checkValidWithNoDupes(login)
-        }
-    }
-
-    @Throws(LoginsStorageException::class)
-    fun rekeyDatabase(newEncryptionKey: String) {
-        return checkUnlocked().rekeyDatabase(newEncryptionKey)
     }
 
     fun registerWithSyncManager() {
--- a/components/logins/src/logins.udl
+++ b/components/logins/src/logins.udl
@@ -2,30 +2,50 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-namespace logins {
-
-    [Throws=LoginsStorageError]
-    string open_and_get_salt(string path, [ByRef] string encryption_key);
-
-    [Throws=LoginsStorageError]
-    void open_and_migrate_to_plaintext_header(string path, [ByRef] string encryption_key, [ByRef] string salt);
-};
-
+namespace logins { };
 
 // Converted from mozilla.appservices.logins.protobuf.rs
-dictionary Login {
+dictionary DecryptedLogin {
     string id;
-    string hostname;
+    string origin;
     string password;
     string username;
     string? http_realm;
-    string? form_submit_url;
+    string? form_action_origin;
     string username_field;
     string password_field;
     i64 times_used;
     i64 time_created;
     i64 time_last_used;
     i64 time_password_changed;
+};
+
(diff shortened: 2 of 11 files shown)


components/logins/src/logins.udl
```diff
@@ -2,18 +2,26 @@
...
+    i64 times_used;
+    i64 time_created;
+    i64 time_last_used;
+    i64 time_password_changed;
 };
 
-
 // Converted from mozilla.appservices.logins.protobuf.rs
-dictionary Login {
+dictionary LoginPayload {
```
skhamis, line 24: I like the idea of `Login` and `LoginPayload` given that we can have the payload be a reduced set of params that is needed to be sent over by the consumer, however I see that `LoginPayload` still contains all the fields -- will this be going away? I would imagine it would like something like:

```
LoginPayload {
    string id;
    string hostname;
    string password;
    string username;
    string? http_realm;
    string? form_submit_url;
    string username_field;
    string password_field;
}
```

but I'm probably missing something here, mainly my initial thoughts from my current mental map!

lougeniaC64: If you look in the login.rs file of the per-field-encryption branch, [the Login struct](https://github.com/mozilla/application-services/blob/18e48dd16b07272a6519a4fe0844d3d3a98eb918/components/logins/src/login.rs#L241) has the encrypted username and password fields and the `LoginPayload` struct is for the unencrypted fields. I do like separating those out so the consumers understand which data they're dealing with. But also, I think that the `username_enc` and `password_enc` fields will end up being optional strings, or diverging from `Logins` in some other way (particularly when the scrubbing logic is added). Having said that, it may be that using one struct is fine for now or that `LoginPayload` isn't the correct nomenclature. Either way, open to suggestions

bendk: `LoginPayload` was designed to be the data we got from the server, which was why the username/password were plaintext.  I think we need to keep that struct as-is (or maybe someone who knows serde more can implement that code in a batter way).

I do like idea of a `Login`-like struct with reduced fields that's intended for the consumer code.  It would just need to be separate from `LoginPayload`

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
```diff
@@ -111,60 +111,40 @@ class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable {
     }
 
     @Throws(LoginsStorageException::class)
-    fun list(): List<Login> {
+    fun decryptAndFixupLogin(encKey: String, login: Login): LoginPayload {
```
bendk, line 114: Maybe we can commit to this just being `decrypt()` since Sammy is going to rewrite the migration to validate it?  I'm not sure on that at all, just wanted to raise the question.

components/logins/src/logins.udl
lougeniaC64: I'm not sure which fields are expected and which we can hide, I know that `time_password_changed` needs to be exposed for iOS and I think one of the other timestamp fields might be used for ordering.

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
```diff
@@ -111,60 +111,40 @@ class DatabaseLoginsStorage(private val dbPath: String) : AutoCloseable {
...
-            checkUnlocked().add(login)
+    fun getByBaseDomain(baseDomain: String): List<Login> {
+        return readQueryCounters.measure {
+            checkUnlocked().getByBaseDomain(baseDomain)
         }
     }
 
     @Throws(LoginsStorageException::class)
-    fun importLogins(logins: List<Login>): String {
+    fun addOrUpdate(encKey: String, login: Login): String {
```
bendk, line 135: `addOrUpdate()` is the biggest question for me.  I'm going to copy/rephrase my thoughts from the google doc here.  BTW, thanks for making this PR, it's nice to have the design more in the open.

Based on the way the API is currently shaped, it seems like you should use the API like this:

 - When the user clicks on the username/password field, call getByBaseDomain() and present some of them to the user
 - If the user clicks on one, then autofil the form and store that `Login` somewhere
 - If the user changes the username/password field values, then update the username/password fields for that stored `Login`
 - If the user submits the form and clicks "Save this login", send the stored `Login` to `addOrUpdate()`.  `addOrUpdate()` will then update that `Login` since it already existed in the DB.

However, depending on what the user entered into the username field, we aren't going to update that login.  If they enter a new username then we should insert a new record.  If they enter an existing username, then we should update a different record. 

I think the issue is that `addOrUpdate()` seems like it would work like a `save()` method in an ORM: it will do an `INSERT` if the guid is not set and an `UPDATE ... WHERE guid={guid}` if the guid it set.  But actually, the GUID is more like a hint.  It only matters as a tiebreaker when there are duplicate records with the same username.   (At least this how I'm reading  [shouldUpdateOrCreateAsync()](https://github.com/mozilla-mobile/android-components/blob/d7b1b62da50cf224fce1eec7a163d858fb1a2334/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt#L29) and [ensureValidLogin()](https://github.com/mozilla-mobile/android-components/blob/d7b1b62da50cf224fce1eec7a163d858fb1a2334/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt#L74))

So I think the signature should change to better indicate what's happening:
  - Define a `LoginData` dictionary, which is the same as `Login`, but without the GUID field
  - Make the signature `addOrUpdate(encKey: String, login: LoginData, selectedGuid: String?)`

Also, on the consumer side we need to store the `Login` the user selects, but we shouldn't be mutating it.  Instead, we should mutate a separate `LoginData` value.  This is because we don't really know which `Login` is going to be updated until the entire process is complete.

bendk:
This is great overall, it was really nice for me to have something more concrete to discuss.  I think the only thing that needs serious thought is addOrUpdate(), my other comments address fairly minor things.

components/logins/src/logins.udl
lougeniaC64: I made some updates to the exposed login structs. I removed `LoginPayload`, added a `LoginFields` struct with the meta fields removed for the `add_or_update` function, and then added a `DecryptedLogin` struct for the `decrypt_and_fixup_login` function. The meta fields are still needed for the fennec to fenix migration code, but we can consider removing those when we stop supporting that logic.

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
lougeniaC64: We should postpone this decision until after the current effort so shipping the migration isn't delayed and any fixup-related regressions are isolated. But I hope that we are able to eventually remove the fixup logic here.

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
lougeniaC64: IIUC the login returned from ` shouldUpdateOrCreateAsync` is merged [here](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt#L76) with the selected login (the one passed into the [onLoginSave function](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt#L68) where `shouldUpdateOrCreateAsync` is called via `ensureValidAsync`) and the resulting merged login is then passed to the [update function](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt#L73). So while I agree the login guid is used as more of a tool to assess the precedence of the potential dupes in the `shouldUpdateOrCreateAsync` function, the login guid dictates which login is updated thereafter making it more consequential. 

Outside of the potential delays the necessary product and/or external team dependencies required to make this update, I question whether this change should be included in an effort that should be invisible to the user. Having said that, I think it's worth capturing this in a ticket for future consideration.

bendk: My thinking was that we were already going to have to change the consumer code, so it wouldn't be much extra effort.  But if that's not true then I definitely agree that we should try to limit the scope.  In that case, what about adding some comments to the `addOrUpdate()` that describe how it's used and what the GUID field is doing?

lougeniaC64:
I think the discussion we're about to have makes this PR obsolete so I'm closing it. cc: @bendk @mhammond

PR #4335, opened 2021-07-13
Fixes panic when indexing hashmap in tabs

tarikeshaq:
Fixes a problem that I got locally where sync manager panics when syncing tabs and poisons the mutex for future syncs

I think we should also have two todos beyond this PR:
- Figure out _why_ I had a payload without a `ttl`, there could be a very intuitive answer to this
- Figure out if there is an easy way to test stuff like this, it seems **very** hard to do so though 😞 

  - Note:

### DIFF
--- a/components/tabs/src/sync/record.rs
+++ b/components/tabs/src/sync/record.rs
@@ -82,7 +82,10 @@ impl TabsRecord {
             .into_iter()
             .collect::<Result<_>>()?;
 
-        let ttl: u32 = payload.data["ttl"].as_u64().unwrap_or_default() as u32;
+        let ttl: u32 = match payload.data.get("ttl") {
+            Some(v) => v.as_u64().unwrap_or_default() as u32,
+            None => u32::default(),
+        };
 
         Ok(TabsRecord {
             id,


tarikeshaq:
The change didn't make it into any releases as far as I can tell, so I don't think we need to make any new releases for this

components/tabs/src/sync/record.rs
```diff
@@ -82,7 +82,10 @@ impl TabsRecord {
             .into_iter()
             .collect::<Result<_>>()?;
 
-        let ttl: u32 = payload.data["ttl"].as_u64().unwrap_or_default() as u32;
+        let ttl: u32 = match payload.data.get("ttl") {
+            Some(v) => v.as_u64().unwrap_or_default() as u32,
+            None => u32::default(),
```
tarikeshaq, line 87: I just wanted to note that I have no idea if this is the appropriate value to use here, but followed what was already set there with the `unwrap_or_default()` I imagine a ttl of 0 is a special value that's ok to use?

rfk: Yep, zero seems as good any to me.

rfk (approved):
👍🏻 nice find!

mhammond:
I was chatting to @tarikeshaq about this and we were wondering how tabs uses a hashmap indexed via the TTL, but it turns out this is actually indexing into a `serde_json::Map` (ie, a JSON "object") just to get the field value.

Interestingly though, according to both https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html and my observations, the `ttl` field is never returned in the payload - so in practice we will never hit the `Some` arm here. That clearly doesn't matter in practice, although it might be worth another issue to avoid future confusion?

(Indeed, I believe it's wrong for that TTL to be on the record at all, but I haven't dug into how difficult it would be to fix that - we do want the TTL in the record when we are sending them, but this shouldn't escape into the public API like it did here all that time ago)

PR #4336, opened 2021-07-13
Add test for `ttl` parsing during Tabs syncing

lougeniaC64:
Just adding test cases to cover the fix @tarikeshaq created in #4335 for the `ttl` population bug I introduced in the refactor of the `from_record` function. I'm stunned with all the tests I created this slipped through the cracks. 😅

  - Note:

### DIFF
--- a/components/tabs/src/sync/record.rs
+++ b/components/tabs/src/sync/record.rs
@@ -400,4 +400,103 @@ pub(crate) mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_tabs_record_from_payload_ttl_values() -> Result<()> {
+        struct TestCase {
+            case_description: String,
+            data: JsonValue,
+            expected_ttl: u32,
+        }
+
+        let test_cases = vec![
+            TestCase {
+                case_description: "test missing ttl".to_string(),
+                data: json!({
+                    "clientName": "Nightly",
+                    "tabs": [
+                        {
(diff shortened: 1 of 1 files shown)


components/tabs/src/sync/record.rs
```diff
@@ -400,4 +400,83 @@ pub(crate) mod tests {
...
+    #[test]
+    fn test_tabs_record_from_payload_ttl_values() -> Result<()> {
+        struct TestCase {
+            case_description: String,
+            data: JsonValue,
+        }
+
+        let test_cases = vec![
+            TestCase {
+                case_description: "test invalid ttl".to_string(),
```
tarikeshaq, line 413: do we want to be specific here about the test case, like "test no ttl provided" etc? Same goes for the other test cases

tarikeshaq (approved):
Added a quick comment about the `case_description`, and do want to also add a test case where the `ttl` is anything other than `0` and verify that it gets parsed correctly? Not sure if this is already covered though

otherwise looks good to me!

components/tabs/src/sync/record.rs
lougeniaC64: This is a copy/paste omission. I'll make these more specific--thanks.

lougeniaC64:
Closing as these `from_record` changes are being reverted in #4337.

PR #4340, opened 2021-07-14
Adds push component diagram and cargo docs

tarikeshaq:
- adds the push component diagram ([rendered here](https://github.com/mozilla/application-services/blob/adds-push-diagram/components/push/README.md))
- adds all the contents of the `README.md` to the high level rust docs of the component
  (this should probably wait till we have the conversation about who the audience is for the docs, I can adapt what docs I put here as a part of that conversation)
- moves over the `cargo doc` part from our github action to the `build_book` script, because currently, it fails unless you manually copy over the rust docs into the docs folder
- Final note, I'm 90% sure that the link I put in the rust docs to the diagram will work once the book is published, but we'll have to wait and confirm - unfortunately it can't be a relative path, but must be hosted somewhere  ref: https://github.com/rust-lang/rust/issues/32104 (and lucky for us, it'll be hosted once it's pushed up)


components/push/README.md
```diff
@@ -29,6 +29,9 @@ the application.
 The HTTP endpoint is provided by Mozilla's [autopush service](https://autopush.readthedocs.io/en/latest/),
 and we use the [rust-ece](https://github.com/mozilla/rust-ece) to manage encryption with the cryptographic keys.
 
+Here's a helpful diagram of how the *subscription* flow works at a high level across the moving parts:
+![A Sequence diagram showing how the different parts of push interact](../../docs/diagrams/Push-Component-Subscription-flow.png "Sequence diagram")
```
rfk, line 33: One tiny suggestion for this diagram: it may be helpful to have concrete examples of "Sender" and "Receiving App", just to help the consumer orient to diagram to their own mental model of the world. For example having "Sender (e.g. FxA server)" and "Receiving App (e.g. Fenix)".

components/push/src/lib.rs
```diff
@@ -4,6 +4,181 @@
...
+//! bridge service; see [this google doc](https://docs.google.com/document/d/18L_g2hIj_1mncF978A_SHXN4udDQLut5P_ZHYZEwGP8) for details.
+//!
+//! ## API
+//!
+//! ## Initialization
+//!
+//! Calls are handled by the `PushManager`, which provides a handle for future calls.
+//!
+//! example:
+//! ```kotlin
```
rfk, line 78: Heh, interesting...I think it's lightly weird to have Kotlin example code in a `.rs` source file, but also, this *does* seem inline with the discussion of how might consume these docs, so I'm happy to run with it!

components/push/src/lib.rs
```diff
@@ -4,6 +4,181 @@
 
 #![allow(unknown_lints)]
 #![warn(rust_2018_idioms)]
+//! # Rust Push Component
```
rfk, line 7: If having these live here in the Rustdoc works out well, we could consider removing them from the README and replacing it with a more succinct overview and a link to the generated docs. That's for a future PR though.

rfk (approved):
Looks good to me! Thanks for accepting the challenge to go the extra mile and have it published in the docs site :-)

PR #4341, opened 2021-07-15
Rewrite logins migration

skhamis:
Initial draft of rewriting the logins migration to a manual approach rather than using export

Most of the new functionality is in `migrate_from_sqlcipher_db`

Still left to discuss:
(1) needing to convert timestamps to millis?
(2) Using raw SQL vs exposed APIs

  - Note:

### DIFF
--- a/components/logins/src/migrate_sqlcipher_db.rs
+++ b/components/logins/src/migrate_sqlcipher_db.rs
@@ -115,12 +115,16 @@ mod tests {
             ALTER TABLE loginsL RENAME passwordEnc to password;
             ALTER TABLE loginsM RENAME usernameEnc to username;
             ALTER TABLE loginsM RENAME passwordEnc to password;
-            INSERT INTO loginsL(guid, username, password, origin,
-            httpRealm, formActionOrigin, usernameField, passwordField, timeCreated, timeLastUsed,
+            ALTER TABLE loginsL RENAME origin to hostname;
+            ALTER TABLE loginsL RENAME formActionOrigin to formSubmitURL;
+            ALTER TABLE loginsM RENAME origin to hostname;
+            ALTER TABLE loginsM RENAME formActionOrigin to formSubmitURL;
+            INSERT INTO loginsL(guid, username, password, hostname,
+            httpRealm, formSubmitURL, usernameField, passwordField, timeCreated, timeLastUsed,
             timePasswordChanged, timesUsed, local_modified, is_deleted, sync_status)
             VALUES ('a', 'test', 'password', 'https://www.example.com', NULL, 'https://www.example.com',
             'username', 'password', 1000, 1000, 1, 10, 1000, 0, 2);
-            INSERT INTO loginsM(guid, username, password, origin, httpRealm, formActionOrigin,
+            INSERT INTO loginsM(guid, username, password, hostname, httpRealm, formSubmitURL,
             usernameField, passwordField, timeCreated, timeLastUsed, timePasswordChanged, timesUsed,
             is_overridden, server_modified)
             VALUES ('b', 'test', 'password', 'https://www.example.com', 'Test Realm', NULL,
--- a/components/logins/src/migrate_sqlcipher_db.rs
+++ b/components/logins/src/migrate_sqlcipher_db.rs
@@ -7,11 +7,45 @@
 use crate::db::MigrationMetrics;
 use crate::encryption::EncryptorDecryptor;
 use crate::error::*;
-use crate::schema;
+use crate::Login;
+use crate::LoginStore;
+use lazy_static::lazy_static;
 use rusqlite::{Connection, NO_PARAMS};
 use sql_support::ConnExt;
 use std::path::Path;
 use std::time::Instant;
+
+// Keep it around to be easy to reference the old schema
+pub const SQL_CIPHER_COLS: &str = "
+    guid,
+    username,
+    password,
+    hostname,
+    httpRealm,
+    formSubmitURL,
+    usernameField,
+    passwordField,
+    timeCreated,
+    timeLastUsed,
+    timePasswordChanged,
+    timesUsed
+";
+
+lazy_static! {
+    pub static ref GET_BY_GUID_SQL: String = format!(
+        "SELECT {common_cols}
+         FROM loginsL
+         WHERE is_deleted = 0
+
+         UNION ALL
+    
+         SELECT {common_cols}
+         FROM loginsM
+         WHERE is_overridden IS NOT 1
+        ",
+        common_cols = SQL_CIPHER_COLS,
+    );
+}
 
 pub fn migrate_sqlcipher_db_to_plaintext(
     old_db_path: impl AsRef<Path>,
@@ -22,14 +56,13 @@ pub fn migrate_sqlcipher_db_to_plaintext(
 ) -> Result<MigrationMetrics> {
     let mut db = Connection::open(old_db_path)?;
     init_sqlcipher_db(&mut db, old_encryption_key, salt)?;
-    // We want the new db to have the latest changes
-    let mut new_db = Connection::open(new_db_path)?;
-    schema::create(&new_db)?;
+
+    // Init the new plaintext db as we would a regular client
+    let plaintext_store = LoginStore::new(new_db_path)?;
 
     //Here we need to call a function that goes row-by-row and copies to the new DB
-    let metrics = migrate_from_sqlcipher_db(&mut db, &mut new_db, new_encryption_key)?;
+    let metrics = migrate_from_sqlcipher_db(&mut db, plaintext_store, new_encryption_key)?;
 
-    new_db.execute("PRAGMA user_version = 1", NO_PARAMS)?;
     Ok(metrics)
 }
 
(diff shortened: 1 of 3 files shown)


components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -19,19 +55,18 @@ pub fn migrate_sqlcipher_db_to_plaintext(
...
 ) -> Result<MigrationMetrics> {
     let mut db = Connection::open(old_db_path)?;
-    let metrics = init_sqlcipher_db(&mut db, old_encryption_key, salt, new_encryption_key)?;
-    sqlcipher_export(&mut db, &new_db_path)?;
-    let new_db = Connection::open(new_db_path)?;
-    new_db.execute("PRAGMA user_version = 1", NO_PARAMS)?;
+    init_sqlcipher_db(&mut db, old_encryption_key, salt)?;
+
+    // Init the new plaintext db as we would a regular client
+    let plaintext_store = LoginStore::new(new_db_path)?;
```
lougeniaC64, line 61: nit: I would suggest renaming 'plaintext_store'. It sounds like we're storing unencrypted login records.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,27 +101,80 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
-        .ok_or_else(|| ErrorKind::InvalidPath(path.to_os_string()))?;
-
-    conn.execute("ATTACH DATABASE ? AS plaintext key ''", &[path])?;
-    // this one is a bit weird because it's a SELECT statement that we know will return 0 rows.
-    // rusqlite will return an Error if we use execute(), so we use query_row with a dummy closure
-    conn.query_row("SELECT sqlcipher_export('plaintext')", NO_PARAMS, |_| Ok(0))?;
-    conn.execute("DETACH DATABASE plaintext", NO_PARAMS)?;
-    Ok(())
+//Manually copy over row by row from sqlcipher db to a plaintext db
+pub fn migrate_from_sqlcipher_db(
```
lougeniaC64, line 105: Is this (or will this be) the `migrate_single_row` you were referring to in the PR description?

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,27 +101,80 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+            username_field: username_field.unwrap_or_default(),
+            password_field: password_field.unwrap_or_default(),
+            // TO_DO: Do we need to convert from microsecond timestamps to milliseconds??
+            time_created,
+            time_last_used,
+            time_password_changed,
+            times_used,
+        };
+
+        // This should be updated with whatever new API we're using to validate, fixup and add
```
lougeniaC64, line 155: If the plan is to use `add_or_update` here (there won't be an `add` store function after the API changes) keep in mind that validate and fixup logic will be occurring in that new function. However using that function here may confuse what's happening here. After all we should only ever be using the add logic in `add_or_update`.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,40 +99,95 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+
+    let mut metrics: MigrationMetrics = MigrationMetrics::default();
+
+    // Select From both LoginsL and LoginsM with a union to ensure we're covering our
+    // migration cases (one table has but not the other, etc)
+    let mut select_stmt = cipher_conn.prepare(&GET_BY_GUID_SQL)?;
+
+    // Use raw rows to avoid extra copying since we're looping over an entire table
+    let mut rows = select_stmt.query(NO_PARAMS)?;
+    while let Some(row) = rows.next()? {
```
bendk, line 118: Maybe wrap everything inside the brace in a function so that we can handle errors and continue on.  For example if one of the required fields were missing or some field had the wrong type.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -5,11 +5,47 @@
...
+    pub static ref GET_BY_GUID_SQL: String = format!(
+        "SELECT {common_cols}
+         FROM loginsL
+         WHERE is_deleted = 0
+
+         UNION ALL
+    
+         SELECT {common_cols}
+         FROM loginsM
+         WHERE is_overridden IS NOT 1
```
bendk, line 44: I think you might need to remove the WHERE clauses and loop over each row from both tables to handle syncing correctly.  Some examples:

  - If there's a row in both tables, then loginsM contains the last known record from the server and loginsL contains the overridden value.  We want to copy both rows to make the merging logic correct.
  - If there's row in LoginsL with is_deleted=0, then I think that means the user deleted the login on this device.  We want to copy that row so that when we sync we treat that as a newly deleted login, rather than a newly added login from another device.

The way the schema is setup is pretty confusing, especially since it's different than most other components.  But I think copying/converting all the rows for both tables should work.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,40 +99,95 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+        let time_last_used: i64 = row.get("timeLastUsed").unwrap_or_default();
+        let time_password_changed: i64 = row.get("timePasswordChanged")?;
+        let times_used: i64 = row.get("timesUsed")?;
+
+        // TODO: Discuss
+        // Need to handle in loginsL: local_modified, is_deleted
+        // Feels like we potentially shouldn't migrate these fields?
+
+        // Need to handle in loginsM: is_overridden, server_modified
+        // Similar to the other: I only see server_modified potentially being needed
```
bendk, line 140: I think we need to copy all of the columns since they're all used with the current code.  We discussed updating the schema to make it match the other components.  If we did that, these columns might not be needed anymore but it would require changing the syncing code.  And I'd like to avoid adding another large change in the branch if we can help it.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,40 +99,95 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+            password_field: password_field.unwrap_or_default(),
+            // TO_DO: Do we need to convert from microsecond timestamps to milliseconds??
+            time_created,
+            time_last_used,
+            time_password_changed,
+            times_used,
+        };
+
+        // Leveraging the add_or_update to get free fixup
+        match new_db_store.add_or_update(login) {
```
bendk, line 161: I like the idea of leveraging `add_or_update()`, but I'm not sure that this will work if my comment above is correct and you need to insert data to both the loginsL and loginsM tables.

I'm also always paranoid of using shared code for migrations.  This means that if we ever want to change `add_or_update()` or the `Login` struct, then it's going to change how the migration works and it could be pretty easy to introduce a bug since we're no longer thinking about migrating these old databases.

bendk (changes requested):
I think the general restructuring looks good but there's a bunch of small details to iron out.  Probably the main issue is how to save the rows to the new database.  I think you may need to write new functions to directly save `loginsL` and `loginsM` rows.

components/logins/src/migrate_sqlcipher_db.rs
skhamis: I see, that indeed is a little confusing. I lifted this SQL to from the [db.rs](https://github.com/mozilla/application-services/blob/b43d90441d7dd160341cb22d8817aa146be50edb/components/logins/src/db.rs#L909) to closely match what we're currently doing in code.

But it seems like we might need to get those rows as well to ensure sync knows how to properly work with the tables

components/logins/src/migrate_sqlcipher_db.rs
skhamis: ah I refactored a bit so `migrate_single_row` doesn't exist anymore. However this will still be the main entrypoint

components/logins/src/migrate_sqlcipher_db.rs
skhamis: I definitely go back and forth on this:

On one hand - writing code once ensures that any new fixups/ways we add to the database stays consistent for anyone migrating from an old version but does potentially impede new functionality if we always have to mold to "legacy" DBs. However in having them completely separate means we still need to make sure any new changes need to be included in the APIs as well as the migration logic else we accidentally put in bad data....definitely need to think about this a little more!

mhammond: > I like the idea of leveraging `add_or_update()`, but I'm not sure

We might be able to leverage the fact that we know it's an empty DB - so it might mean that `add_or_update()` gets a little refactored and we still call the underlying implementation - the main thing I'm keen on is ensuring our real fixup/validation stuff is called and not using sql here for any inserts that might be necessary.

(I'm still leaning towards the nuclear option of changing/migrating the mirror TBH - I'm usually actively looking for incremental approaches to refactoring, but this is a bit of a rare case - but I don't think we need to make that call yet)

mhammond:
`loginsSyncMeta` needs to be migrated too, otherwise we will do a `reset` anyway (so need not have migrated the mirror in the first place)

PR #4343, opened 2021-07-19
Updates to uniffi tabs 

skhamis:
Two adjustments to uniffi tabs based on the feedback in this discussion: https://github.com/mozilla-mobile/android-components/pull/10475

Convert lastUsed `i64` to stay consistent with other components.

  - Note:


tarikeshaq (changes requested):
Only putting `Requesting changes` because the `RemoteTabError` won't ever be returned - otherwise I know little about the work here 😬

components/tabs/src/tabs.udl
```diff
@@ -14,12 +14,17 @@ interface TabsStore {
     void register_with_sync_manager();
 };
 
+[Error]
```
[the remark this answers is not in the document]
skhamis: Good catch, I initially had that in my updates but didn't update to only use internalexception (which we can already use in a-c). I'll remove that and fix some of the clippy errors.

components/tabs/src/tabs.udl
```diff
@@ -19,7 +19,7 @@ dictionary RemoteTab {
     string title;
     sequence<string> url_history;
     string? icon;
-    u64 last_used;
```
rfk, line 22: Technically a breaking API change, so please also add a note to the changelog.

rfk (approved):
LGTM with a note about the changelog 👍🏻 

I saw that there was an update to the Kotlin version in a-c that would make unsigned types supported by default, but it seems reasonable for us to change away from them for consistency and then introduce them again in a more holistic fashion.

components/tabs/src/tabs.udl
skhamis: Will do! I do wonder if this will mark a major upgrade when cutting a build given that this is more of a patch since v81 turned the API to be u64 instead of i64 being passed over the wire.

mhammond: I think it was i64 before this uniffi work, right?

PR #4346, opened 2021-07-20
fixes github actions failure

tarikeshaq:
Since I moved the `cargo doc` to the `build-book.sh` script, on #4340  which is a different CD step, it expects that the `NSS` environment variables are set correctly, otherwise `cargo doc` will fail. (as is seen in https://github.com/mozilla/application-services/actions/runs/1047252137)

This basically just sets the correct env variables on the `Build mdbook` step so cargo doc doesn't fail


bendk (approved):
Looks great to me

PR #4347, opened 2021-07-20
Adds improvements to push component error handling

### LINKED ISSUE #3695: Push component should automatically recover from the server losing its uaid

tarikeshaq:
fixes #3695 

A couple of TODOs for myself before this is ready:
- [x] Add tests to cover the new errors added and the changes to the error handling of client errors
- [x] Make sure that removing the extra context in the `RequestError`s (i.e errors where `viaduct` complains) is worth the conversion and the reduction in code size
- [x] I'm not sure if I need to add any other logic similar to what I added in `verify_connection` to catch the `UAIDNotRecognizedError`, (Possibly in `subscribe`?), but I'll decide if I should do that, or if we should just let the consumer deal with those

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,11 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Push
+### What's changed
+  - The push component will now attempt to auto-recover from the server losing its UAID ([#4347](https://github.com/mozilla/application-services/pull/4347))
+    - The push component will return a new kotlin Error `UAIDNotRecognizedError` in cases where auto-recovering isn't possible (when subscribing)
+    - Two other new errors were defined that were used to be reported under a generic error:
+      - `JSONDeserializeError` for errors in deserialization
+      - `RequestError` for errors in sending a network request
--- a/components/push/src/communications.rs
+++ b/components/push/src/communications.rs
@@ -20,13 +20,17 @@ use viaduct::{header_names, status_codes, Headers, Request};
 use crate::config::PushConfiguration;
 use crate::error::{
     self,
-    ErrorKind::{AlreadyRegisteredError, CommunicationError, CommunicationServerError},
+    ErrorKind::{
+        AlreadyRegisteredError, CommunicationError, CommunicationServerError,
+        UAIDNotRecognizedError,
+    },
 };
 use crate::storage::Store;
 
 mod rate_limiter;
 pub use rate_limiter::PersistedRateLimiter;
 
+const UAID_NOT_FOUND_ERRNO: u32 = 103;
 #[derive(Debug)]
 pub struct RegisterResponse {
     /// The UAID associated with the request
@@ -140,6 +144,38 @@ impl ConnectHttp {
         };
         Ok(headers)
     }
+
+    fn handle_response_error(&self, response: &viaduct::Response) -> error::Result<()> {
+        // An error response, the extended object structure is retrieved from
+        // https://autopush.readthedocs.io/en/latest/http.html#response
+        #[derive(Deserialize)]
+        struct ResponseError {
+            pub errno: u32,
+            pub message: String,
+        }
+        if response.is_server_error() {
+            let response_error = response.json::<ResponseError>()?;
+            return Err(CommunicationServerError(format!(
+                "General Server Error: {}",
+                response_error.message
+            ))
+            .into());
+        }
+        if response.is_client_error() {
+            let response_error = response.json::<ResponseError>()?;
+            if response.status == status_codes::CONFLICT {
+                return Err(AlreadyRegisteredError.into());
+            }
+            if response.status == status_codes::GONE && response_error.errno == UAID_NOT_FOUND_ERRNO
+            {
+                return Err(UAIDNotRecognizedError(response_error.message).into());
+            }
+            return Err(
+                CommunicationError(format!("Unhandled client error {:?}", response)).into(),
+            );
+        }
+        Ok(())
+    }
 }
 
 impl Connection for ConnectHttp {
@@ -193,40 +229,12 @@ impl Connection for ConnectHttp {
             });
         }
         let url = Url::parse(&url)?;
-        let requested = match Request::post(url)
+        let response = Request::post(url)
             .headers(self.headers()?)
             .json(&body)
-            .send()
-        {
-            Ok(v) => v,
-            Err(e) => {
-                return Err(
-                    CommunicationServerError(format!("Could not fetch endpoint: {}", e)).into(),
-                );
-            }
-        };
-        if requested.is_server_error() {
-            return Err(CommunicationServerError("General Server error".to_string()).into());
-        }
-        if requested.is_client_error() {
-            if requested.status == status_codes::CONFLICT {
-                return Err(AlreadyRegisteredError.into());
-            }
-            return Err(CommunicationError(format!(
-                "Unhandled client error {} : {:?}",
-                requested.status,
-                String::from_utf8_lossy(&requested.body)
-            ))
-            .into());
-        }
-        let response: Value = match requested.json() {
-            Ok(v) => v,
-            Err(e) => {
-                return Err(
-                    CommunicationServerError(format!("Could not parse response: {:?}", e)).into(),
-                );
-            }
-        };
+            .send()?;
+        self.handle_response_error(&response)?;
+        let response: Value = response.json()?;
 
         if self.uaid.is_none() {
             self.uaid = response["uaid"].as_str().map(ToString::to_string);
@@ -270,18 +278,14 @@ impl Connection for ConnectHttp {
         if &self.options.sender_id == "test" {
             return Ok(true);
         }
-        match Request::delete(Url::parse(&url)?)
+        let response = Request::delete(Url::parse(&url)?)
             .headers(self.headers()?)
-            .send()
-        {
-            Ok(_) => {
-                if channel_id.is_none() {
-                    self.uaid = None;
-                }
-                Ok(true)
-            }
-            Err(e) => Err(CommunicationServerError(format!("Could not unsubscribe: {}", e)).into()),
+            .send()?;
+        self.handle_response_error(&response)?;
+        if channel_id.is_none() {
+            self.uaid = None;
         }
+        Ok(true)
     }
 
     /// Update the push server with the new OS push authorization token
@@ -309,16 +313,12 @@ impl Connection for ConnectHttp {
         );
         let mut body = HashMap::new();
         body.insert("token", new_token);
-        match Request::put(Url::parse(&url)?)
+        let response = Request::put(Url::parse(&url)?)
             .json(&body)
             .headers(self.headers()?)
-            .send()
-        {
-            Ok(_) => Ok(true),
-            Err(e) => {
-                Err(CommunicationServerError(format!("Could not update token: {}", e)).into())
-            }
-        }
+            .send()?;
+        self.handle_response_error(&response)?;
+        Ok(true)
     }
 
     /// Get a list of server known channels. If it differs from what we have, reset the UAID, and refresh channels.
@@ -349,7 +349,7 @@ impl Connection for ConnectHttp {
             &options.sender_id,
             &self.uaid.clone().unwrap(),
         );
-        let request = match Request::get(Url::parse(&url)?)
+        let response = match Request::get(Url::parse(&url)?)
             .headers(self.headers()?)
             .send()
         {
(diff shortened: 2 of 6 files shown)


rfk:
@bendk let's jointly review this one as a knowledge-sharing exercise - maybe we can each take a look independently and then meet briefly later in the week to discuss?

components/push/src/communications.rs
```diff
@@ -140,6 +144,38 @@ impl ConnectHttp {
...
+            }
+            if response.status == status_codes::GONE && response_error.errno == UAID_NOT_FOUND_ERRNO
+            {
+                return Err(UAIDNotRecognizedError(response_error.message).into());
+            }
+            return Err(
+                CommunicationError(format!("Unhandled client error {:?}", response)).into(),
+            );
+        }
+        Ok(())
```
rfk, line 177: For a moment here, I thought you were silently ignoring other kinds of errors by returning `Ok(())`, but IIUC you will *always* call this `handle_response_error` method and the fall-through case here is the case that no error occurred. That seems fine, but it may be worth considering a name of the method that makes this behaviour more clear.

components/push/src/communications.rs
```diff
@@ -411,7 +397,16 @@ impl Connection for ConnectHttp {
             return Ok(false);
         }
         let local_channels: HashSet<String> = channels.iter().cloned().collect();
-        let remote_channels: HashSet<String> = HashSet::from_iter(self.channel_list()?);
+        let remote_channels: HashSet<String> = match self.channel_list() {
+            Ok(v) => HashSet::from_iter(v),
+            Err(e) => match e.kind() {
+                UAIDNotRecognizedError(_) => {
+                    self.uaid = None;
```
rfk, line 404: Do we also want to call `unsubscribe` in this case, like we do for when there's a mismatch in the channel lists? (I note that `unsubscribe` contains its own internal version of setting `uaid = None`).

tarikeshaq: Hmmm in this case I don't think we should call `unsubscribe`, because we only get this error if the server lost our UAID for one reason or the other, and the `unsubscribe` network request would just fail with the same error...

That said, I think it's begging for a comment to describe what I just said

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,10 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+## Push
+### What's changed
+  - The push component will not attempt to auto-recover from the server losing its UAID ([#4347](https://github.com/mozilla/application-services/pull/4347))
+    - The push component will return a new kotlin Error `UAIDNotRecognizedError` in cases where auto-recovering isn't possible (when subscribing)
```
tarikeshaq, line 24: Some context on why we don't try to auto-recover on `subscribe` calls, but do so on `verifyConnection`, (I'm open to being convinced with other suggestions!).

If the consumer (A-C) is calling `verifyConnection` they're expecting to get a list of the channels they should re-subscribe to if something went wrong

but if they are calling `subscribe`, that expectation isn't there. So if we attempt to wipe the UAID and the channels, the consumer would not be notified to resubscribe.

The way I'm thinking of making this better is to let the consumer (A-C) catch this error and run `verifyConnection` fixing its worldview.

components/push/src/communications.rs
```diff
@@ -697,5 +694,228 @@ mod test {
...
+                server_host: server_address().to_string(),
+                sender_id: SENDER_ID.to_owned(),
+                bridge_type: Some("test".to_owned()),
+                registration_id: Some("SomeRegistrationValue".to_owned()),
+                ..Default::default()
+            };
+            // We mock that the server thinks
+            // we already registered!
+            let body = json!({
+                "code": status_codes::CONFLICT,
```
tarikeshaq, line 896: Looking at https://autopush.readthedocs.io/en/latest/http.html#error-codes, `409` isn't listed on the error codes....

I'm a little suspicious that the whole `AlreadyRegistered` error might never be hit. I'll take a little time to see if I can confirm that and get rid of the `AlreadyRegistered` error all together, but that's probably better for another clean-up PR

components/push/src/communications.rs
tarikeshaq: Good point! I renamed it to `check_response_error` 😄

components/push/src/communications.rs
tarikeshaq: this raises an interesting question that probably requires a decent bit of high-level figuring out on my end....

if the server loses our UAID without us explicitly `unsubscribe`ing, how bad is it on the server, and is it really OK for the client to request another UAID as opposed to erroring out? I'm assuming the client will eventually request another UAID after the verifyConnection is returned to the consumer, and the consumer decides to resubscribe.

Like, will we have dangling connections to firebase/etc, and if we create a second connection, how bad is it? I'll need to look more into what specific cases do we end up losing the UAID (I'm assuming there are more cases than the fix I did on #4325). if we lose the UAID because we called `unsubscribe(None)` this is fine. If we lose it for some other reason, we might want to check and confirm that requesting another UAID doesn't leave our overall system in a bad state

anyways, I might be waaay overthinking things 😅

rfk: >  and the unsubscribe network request would just fail with the same error...

Ah yes, I see - in my head the `unsubscribe` call was also deleting the local records but that's not the case, that happens at a higher level of the call stack. Thanks!

rfk: > I'm assuming the client will eventually request another UAID after the verifyConnection is returned to the consumer,
> and the consumer decides to resubscribe.

I believe that's the intention, yes.

IIUC what's supposed to happen if there is a server-side error, is that the native push layer signals to the app that it needs to reconnect, and the app [registers a new token](https://github.com/mozilla/application-services/blob/eae2a401008e155bb6ebcb3375b77db70a19fd9d/components/push/src/subscriber.rs#L128) with the autopush server. However, I don't think we have a good shared understanding of all the possible combinations of (client state, autopush state, native push infra state) and the possible transitions between them. A valuable longer-term effort here would be to construct a state transition diagram for each of the three systems interacting here, it will probably surface some interactions we haven't thought of yet.

components/push/src/communications.rs
```diff
@@ -411,7 +397,18 @@ impl Connection for ConnectHttp {
...
         }
         let local_channels: HashSet<String> = channels.iter().cloned().collect();
-        let remote_channels: HashSet<String> = HashSet::from_iter(self.channel_list()?);
+        let remote_channels: HashSet<String> = match self.channel_list() {
+            Ok(v) => HashSet::from_iter(v),
+            Err(e) => match e.kind() {
+                UAIDNotRecognizedError(_) => {
+                    // We do not unsubscribe, because the
+                    // server already lost our UAID
+                    self.uaid = None;
```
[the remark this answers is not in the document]
tarikeshaq: Hmmm that's a good point.. I think you're right. https://autopush.readthedocs.io/en/latest/http.html#registration seems to hint that it'll always return a new UAID **and** `secret`.

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,10 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+## Push
+### What's changed
+  - The push component will now attempt to auto-recover from the server losing its UAID ([#4347](https://github.com/mozilla/application-services/pull/4347))
+    - The push component will return a new kotlin Error `UAIDNotRecognizedError` in cases where auto-recovering isn't possible (when subscribing)
```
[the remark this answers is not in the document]
rfk: We don't yet have iOS bindings for this component (we've always intended to, but the work to migrate from the existing push infra in firefox-ios hasn't gotten to the top of anyone's stack)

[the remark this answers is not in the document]
tarikeshaq: The push component is currently only in android :( hopefully when we uniffi it and if there is interest on the iOS side to use the component it'll end up there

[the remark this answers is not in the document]
tarikeshaq: Ooops, @rfk beat me to the comment 😄

components/push/src/communications.rs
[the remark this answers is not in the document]
tarikeshaq: Okay, added setting `self.auth = None` for now, but this is begging to be refactored. I'll file a follow-up for refactors (including this and the `channel_id` you pointed above!

components/push/src/communications.rs
```diff
@@ -270,18 +278,14 @@ impl Connection for ConnectHttp {
...
-                if channel_id.is_none() {
-                    self.uaid = None;
-                }
-                Ok(true)
-            }
-            Err(e) => Err(CommunicationServerError(format!("Could not unsubscribe: {}", e)).into()),
+            .send()?;
+        self.check_response_error(&response)?;
+        if channel_id.is_none() {
+            self.uaid = None;
```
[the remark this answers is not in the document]
tarikeshaq: Added a comment for the function, it is confusing 😞 I'll add this refactoring as a part of another clean-up issue I'll create though!

components/push/src/communications.rs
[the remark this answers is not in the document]
bendk: A follow-up for refactors sounds good to me.  I'll just note one issue to tackle: now if `verify_connection` is called twice in a row with a lost UAID, the first one will return `Ok(false)` and the second will be `Err(CommunicationError)`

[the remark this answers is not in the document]
tarikeshaq: Good point!! Should of caught that - I wonder if we need to fix this right away (happy to do so as a part of this PR!)
a fix would either:
- Keep a new list of (to be re-subscribed) channels in the component, save the state of the component (probably at the `PushManager` level, and return that list directly if any subsequent `verify_connections` get called
- or, keep the persisted state until the consumer calls `subscribe` then wipe the old state. Things can go wrong here if the consumer decides to call something other than `subscribe` so I prefer the first option.

Either way, it depends if we want to support being consistent on two consecutive `verify_connection` @jonalmeida do you happen to know if it's possible for the consumer on android to call a failed `verify_connection` twice before `subscribe`ing? If so I'll add it as a part of this PR, otherwise, I'll do a follow-up PR of that once this merges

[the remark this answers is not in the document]
bendk: I'm not sure if this is needed, I don't know enough about how it's used.  But if you want to make that return value consistent, what about reworking how `uaid` and `secret` are stored?  Instead of 2 `Option`s, have a single enum:

```
enum ConnectionState {
  NotConnected,
  NeedsReconnection,
  Connected(uaid: String, auth: String),
}
```

That way `verify_connection` could return the error for `NotConnected` and return `false` for NeedsReconnection.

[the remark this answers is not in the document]
jonalmeida: > @jonalmeida do you happen to know if it's possible for the consumer on android to call a failed `verify_connection` twice before `subscribe`ing? If so I'll add it as a part of this PR, otherwise, I'll do a follow-up PR of that once this merges

That shouldn't happen from our current design. Calls to verify_connection, should start the flow that leads to a subscribe event.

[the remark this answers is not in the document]
tarikeshaq: Thanks Jonathan! @bendk I created https://github.com/mozilla/application-services/issues/4372 to track the reactors you pointed out, since this isn't possible for now we can tackle it there

tarikeshaq:
Mildly concerned about the test failure here.... it seems like the issue is that `rc_crypto::ensure_initialized` was just never called in the test, but it is concerning that it's only showing up in this PR where we are modifying some of the subscribe behavior. I'll investigate

components/push/src/communications.rs
tarikeshaq: filed #4373!

tarikeshaq:
@rfk, @bendk  re-requesting review - it should be ready considering the follow-ups I filed (unless you catch anything else!) Couldn't figure out why an `ece not initialized` test failure popped up earlier, to be extra safe I added a `rc_crypto::ensure_initialized` on that test. I wasn't able to repro locally

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,11 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Push
+### What's changed
+  - The push component will now attempt to auto-recover from the server losing its UAID ([#4347](https://github.com/mozilla/application-services/pull/4347))
+    - The push component will return a new kotlin Error `UAIDNotRecognizedError` in cases where auto-recovering isn't possible (when subscribing)
+    - Two other new errors were defined that were used to be reported under a generic error:
```
rfk, line 26: nit: "were used" -> "used" or "were previously" or similar?

components/push/src/subscriber.rs
```diff
@@ -228,7 +228,7 @@ mod test {
     #[test]
     fn full() -> Result<()> {
         use rc_crypto::ece;
-
+        rc_crypto::ensure_initialized();
```
rfk, line 231: Hrm, so it looks like the `push` crate only calls this in the `decrypt` method in `./src/crypto.rs`. Probably it should also call it in the `generate_key` method as well (which would result in it being initialized by the call to `pm.subscribe` in the test here).

rfk (approved):
LGTM, I left one suggestion for something to try about initializing rc_crypto.

bendk (approved):
Changes look great to me!

PR #4354, opened 2021-07-23
Split logins sync functionality into separate modules

mhammond:
login.rs was wearing many many hats, so I split the sync specific stuff out into its own module - I went from one extreme to the other - there's now "payload.rs", "merge.rs" etc, but I think it looks good. OTOH, there are other refactors that become clearer (eg, the fact that `sync::payload` adds an impl for `Login`) but I thought it better to leave this as literally just moving stuff rather than a larger refactor.

I did this mainly to help future us on this per-fields work, making it a bit clearer what is and isn't sync specific.


components/logins/src/sync/engine.rs
```diff
@@ -640,8 +676,8 @@ mod tests {
                 (
                     sync15::Payload::from_json(serde_json::json!({
                         "id": "dummy_000003",
-                        "formActionOrigin": "https://www.example.com/submit",
```
mhammond, line 643: This also fixes a regression on this branch where the names of the items in the payload were changed from `hostname` to `origin` and `formSubmitURL` to `formActionOrigin` - the names in the payload must match other clients.

bendk (approved):
I didn't go through all the details here, under the assumption that the only functional change is the `LoginPayload` field names.  I really like the idea of splitting up the code more.

skhamis:
Late to this refactor party but wanted to say this is an awesome refactor and will absolutely help with future logins work!

PR #4356, opened 2021-07-26
Update Glean submodule to v39.0.4

### LINKED ISSUE #4355: Xcode build warnings due to expired Glean metrics
**Edit**: in my initial testing I seemed able to consistently reproduce a build failure due to these warnings, but now I no longer can, so perhaps there was something else at play. I think it's still worth updating Glean to remove the build warnings.

---

Attempting to build our latest `main` via `./build-carthage.sh` produces some Glean warnings:

```
warning: Sorry, Glean found some glinter nits:
warning: WARNING: EXPIRED: glean.time.invalid_timezone_offset: Metric has expired. Please consider removing it.

warning: Please fix the above nits to continue.
warning: To disable a check, add a `no_lint` parameter with a list of check names to disable.
warning: This parameter can appear with each individual metric, or at the top-level to affect the entire file.
```

~~These say they're warnings, but AFAICT they fail the build, and if I apply the fix from https://github.com/mozilla/glean/pull/1697 manually then the build succeeds.~~

~~I wasn't able to find a way to ignore this lint at build time, and the above fix is not yet in a release of Glean. I'm going to try pinning out submodule to that specific commit in order to unbreak the build, but I think we'll need some help from the Glean team on a "proper" fix here.~~

Let's update to a version of Glean that doesn't produce these warnings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2708)

rfk:
**Edit**: I had originally thought I was getting outright build failures as a result of this issue, but CI seems to build consistently despite the Glean warnings, so maybe there was something else at play in my local environment. Still, let's get rid of the warnings as a matter of general build hygiene.

---

This fixes a build issue on iOS, where we ~fail to generate the Glean core metrics because one of them is expired~ get warnings about expired metrics from Glean. It is currently pinned to a specific commit of Glean but we should see about getting a tagged release.

It also includes an update to the Nimbus code on iOS to avoid using deprecated APIs.

Fixes #4355, but draft until we get a release or advice to the contrary from the Glean folks.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,9 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## General
+
+### What's Changed
+
+  - The bundled version of Glean has been updated to v39.0.4.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "addr2line"
-version = "0.15.2"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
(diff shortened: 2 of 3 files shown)


travis79 (approved):

badboy:
As said in the other bug, this should pull in [v39.0.4](https://github.com/mozilla/glean/releases/tag/v39.0.4) now.

PR #4357, opened 2021-07-26
Verify desktop build environment when verifying for mobile.

rfk:
It's reasonable for mobile-focused consumers to want to be able
to do a quick `cargo build` to test their local checkout, and it's
not reasonable for us to expect them to understand how or why this
is different from building inside Xcode or Android Studio. Let's
help make this Just Work (tm) by verifying the Desktop development
environment whenever we verify a Mobile development environment.

What I've done here is essentially:

* Move the bulk of the android/ios verify logic into new scripts named `verify-[platform]-ci-environment.sh`
* Make the existing `verify-[platform]-environment.sh` scripts be wrappers around this logic that also:
    * Verify the desktop build environment
    * Print human-facing next steps

See [this slack thread](https://mozilla.slack.com/archives/CKL0LGV9B/p1627078154471100) for a situation in which doing this by default would have helped provide a smoother consumer experience.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,7 +351,7 @@ jobs:
       - carthage-bootstrap
       - run:
           name: Verify the build environment
-          command: ./libs/verify-ios-environment.sh
+          command: ./libs/verify-ios-ci-environment.sh
       - run:
           name: Run XCode tests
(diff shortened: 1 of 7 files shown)


lougeniaC64 (approved):
LGTM

PR #4359, opened 2021-07-26
Allows nimbus to accept context defined in the consumer

tarikeshaq:
I'll need to add more thorough testing, but this is an initial stab at the problem to see if we wanna go through this route.

fixes https://mozilla-hub.atlassian.net/browse/SDK-345

A couple of thoughts about this, and I'm not sure if this warrants an ADR...

Pros:
  - Consumers can now define their own `targeting` fields, and as long as those align with what the settings server sends us, then the nimbus SDK will be able to evaluate it correctly, without the SDK needing to know about the specific fields. This saves us developer time of having to go on and add more fields to the `AppContext` struct.
  - As non-browser consumers (or consumers who simply don't have some of the fields other consumers have, I'm looking at you `android_sdk_version`) it wouldn't make sense for the SDK to include those fields. 

Cons:
  - Biggest con I can think of is that there's less type checking. The consumers would have to be extra careful to get their fields right, when right now the uniffi-generated struct make it impossible to get wrong.
  - Minor con is that if we want to access the `AppContext.specific_context` fields for non-targeting-related reasons, we'll probably need to be a bit more careful on the Rust side to add error handling

  - Note:

### DIFF
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
@@ -261,7 +261,13 @@ data class NimbusAppInfo(
      *
      * Examples: "nightly", "beta", "release"
      */
-    val channel: String
+    val channel: String,
+    /**
+     * Application derived attributes measured by the application, but useful for targeting of experiments.
+     *
+     * Example: mapOf("userType": "casual", "isFirstTime": "true")
+     */
+    val customTargetingAttributes: Map<String, String> = mapOf()
 )
 
 /**
@@ -609,6 +615,7 @@ open class Nimbus(
             deviceModel = Build.MODEL,
             locale = deviceInfo.localeTag,
             os = "Android",
-            osVersion = Build.VERSION.RELEASE)
+            osVersion = Build.VERSION.RELEASE,
+            customTargetingAttributes = appInfo.customTargetingAttributes)
     }
 }
--- a/components/nimbus/ios/Nimbus/NimbusApi.swift
+++ b/components/nimbus/ios/Nimbus/NimbusApi.swift
@@ -202,16 +202,19 @@ public struct NimbusServerSettings {
 
 public let remoteSettingsCollection = "nimbus-mobile-experiments"
 
-/// Name and channel of the app, which should agree with what is specified in Experimenter.
+/// Name, channel and specific context of the app which should agree with what is specified in Experimenter.
+/// The specifc context is there to capture any context that the SDK doesn't need to be explictly aware of.
 ///
 public struct NimbusAppSettings {
-    public init(appName: String, channel: String) {
+    public init(appName: String, channel: String, customTargetingAttributes: [String: String] = [String: String]()) {
         self.appName = appName
         self.channel = channel
+        self.customTargetingAttributes = customTargetingAttributes
     }
 
     public let appName: String
     public let channel: String
+    public let customTargetingAttributes: [String: String]
 }
 
 /// This error reporter is passed to `Nimbus` and any errors that are caught are reported via this type.
--- a/components/nimbus/src/evaluator.rs
+++ b/components/nimbus/src/evaluator.rs
@@ -249,6 +249,7 @@ mod tests {
             os_version: Some("10".to_string()),
             android_sdk_version: Some("29".to_string()),
             debug_tag: None,
+            custom_targeting_attributes: None,
         };
         assert_eq!(targeting(expression_statement, &ctx), None);
 
@@ -267,6 +268,7 @@ mod tests {
             os_version: Some("10".to_string()),
             android_sdk_version: Some("29".to_string()),
             debug_tag: None,
+            custom_targeting_attributes: None,
         };
         assert_eq!(targeting(expression_statement, &ctx), None);
 
@@ -285,6 +287,7 @@ mod tests {
             os_version: Some("10".to_string()),
             android_sdk_version: Some("29".to_string()),
             debug_tag: None,
+            custom_targeting_attributes: None,
         };
         assert!(matches!(
             targeting(expression_statement, &non_matching_ctx),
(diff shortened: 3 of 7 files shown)


components/nimbus/src/matcher.rs
```diff
@@ -67,4 +69,6 @@ pub struct AppContext {
     pub os_version: Option<String>,
     pub android_sdk_version: Option<String>,
     pub debug_tag: Option<String>,
+    #[serde(flatten)]
+    pub specific_context: Option<HashMap<String, String>>,
```
tarikeshaq, line 73: Just to be 1000% sure, the fact that this is optional, it doesn't break backward compatibility right?

components/nimbus/src/nimbus.udl
```diff
@@ -13,6 +13,7 @@ dictionary AppContext {
     string? os_version;
     string? android_sdk_version;
     string? debug_tag;
+    record<DOMString, string>? specific_context;
```
tarikeshaq, line 16: `DOMString` here is a awkward, but I think that's a uniffi limitation? I'll double-check, saw it in the docs

tarikeshaq: Not too sure about the name `specific_context`, happy to take suggestions for something more representative

components/nimbus/src/matcher.rs
travis79: Yeah, as long as this is optional, then it shouldn't fail serializing/deserializing if it isn't there.

components/nimbus/src/nimbus.udl
travis79: I don't have any issues with `specific_context`, maybe adding `app_` to it might make it more clear?  So `app_specific_context` or even `app_defined_context`?  Again, not blocking on this (naming things is hard).

components/nimbus/src/matcher.rs
```diff
@@ -67,4 +69,6 @@ pub struct AppContext {
     pub os_version: Option<String>,
     pub android_sdk_version: Option<String>,
     pub debug_tag: Option<String>,
+    #[serde(flatten)]
+    pub app_specific_context: Option<HashMap<String, String>>,
```
jhugman, line 73: This is a really elegant way of solving this!

components/nimbus/ios/Nimbus/NimbusApi.swift
```diff
@@ -202,16 +202,19 @@ public struct NimbusServerSettings {
...
-    public init(appName: String, channel: String) {
+    public init(appName: String, channel: String, appSpecificContext: [String: String]? = nil) {
         self.appName = appName
         self.channel = channel
+        self.appSpecificContext = appSpecificContext
     }
 
     public let appName: String
     public let channel: String
+    public let appSpecificContext: [String: String]?
```
jhugman, line 217: I'm not wild about the naming, although I prefer `appSpecificContext` to `specificContext`. FWIW [`Context` has a very specific meaning][1] in Android development.

I prefer `customAttributes` or `appSpecificTargetingAttributes`, or some such. Everyone likes custom things.

[1]: https://developer.android.com/reference/android/content/Context

I appreciate you've already renamed once, so please take this as optional.

components/nimbus/src/matcher.rs
```diff
@@ -52,6 +53,7 @@ pub struct Matcher {
 /// - `os_version`: The user-visible version of the operating system (e.g. "1.2.3")
 /// - `android_sdk_version`: Android specific for targeting specific sdk versions
 /// - `debug_tag`: Used for debug purposes as a way to match only developer builds, etc.
+/// - `app_specific_context`: Context that is defined by consumers that the SDK doesn't need to know about
```
jhugman, line 56:
```suggestion
/// - `app_specific_context`: Context contains attributes specific to the application, derived by the application.
```

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -261,7 +261,13 @@ data class NimbusAppInfo(
...
      * Examples: "nightly", "beta", "release"
      */
-    val channel: String
+    val channel: String,
+    /**
+     * Some specific context that the SDK doesn't need to be explicitly aware of.
+     * this is used to allow targeting on fields that are plaform or app specific
+     * that other consumers of the SDK shouldn't consume
+     */
+    val appSpecificContext: Map<String, String>? = null
```
jhugman, line 270: We might also simplify a lot of code here by making all this non-optional.

```suggestion
    val appSpecificContext: Map<String, String> = mapOf()
```

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -261,7 +261,13 @@ data class NimbusAppInfo(
      *
      * Examples: "nightly", "beta", "release"
      */
-    val channel: String
+    val channel: String,
+    /**
+     * Some specific context that the SDK doesn't need to be explicitly aware of.
+     * this is used to allow targeting on fields that are plaform or app specific
+     * that other consumers of the SDK shouldn't consume
```
jhugman, line 268: By this point, this is app developer focused documentation.

I suggest the docs should be written for them: e.g. application derived attributes measured by the application, but useful for targeting of experiments. e.g. `mapOf("userType" to "casual", "isFirstTime": "true")`.

components/nimbus/src/evaluator.rs
```diff
@@ -316,6 +320,58 @@ mod tests {
             })
         ));
     }
+    use std::collections::HashMap;
+
+    #[test]
+    fn test_targeting_app_specific_context() {
+        // Here's our valid jexl statement
+        let expression_statement =
+            "app_id == '1010' && (app_version == '4.4' || locale == \"en-US\") && is_first_run == 'true' && ios_version == '8.8'";
```
jhugman, line 329: I had only just noticed that all the existing attributes were `String`. Question: does JEXL allows us to get numbers from strings? If so, this should definitely be documented/tested here.

jhugman (approved):
This is lovely. A few minor minor things, and a couple of questions, but nothing to hold back with another review cycle. 

I just noticed this is a Draft, so please ask as many questions as you want before asking for another review.

components/nimbus/src/evaluator.rs
travis79: I'm pretty sure it does, since we can use comparison operators..   

See:
https://github.com/mozilla/jexl-rs/blob/351aaeb6eb172b2538a46a6aa52cc14bd1b0d909/jexl-parser/src/ast.rs#L47

and looking at some of the tests in jexl-rs

components/nimbus/src/matcher.rs
```diff
@@ -67,4 +69,6 @@ pub struct AppContext {
     pub os_version: Option<String>,
     pub android_sdk_version: Option<String>,
     pub debug_tag: Option<String>,
+    #[serde(flatten)]
+    pub custom_targeting_attributes: Option<HashMap<String, String>>,
```
travis79, line 73: Best name yet! ❤️

travis79 (approved):
This is awesome!  Thanks @tarikeshaq

components/nimbus/src/evaluator.rs
tarikeshaq: I'll create a follow-up ticket in the testing epic to add more comprehensive jexl tests!

PR #4361, opened 2021-07-28
Large refactor of the core structs and encryption model

mhammond:
As discussed a little in person - this is a large patch that changes the shape of the structs etc, but does *not* radically change the shape of the API (other than to require the encryption key where necessary) - those API changes should be done ASAP, but probably in a different PR.

I'm open to all bike-shedding - some of the struct names don't seem ideal, but better alternative aren't completely obvious. YMMV, so if you think better names are obvious, then please share!

Except for migration, all tests work, and all `TODO-sqlcipher` comments have been removed. Android compiles. I didn't merge Sammy's migration PR yet because this PR adds some primitives that could be used (and if not for needing to migrate the mirror, that migration could almost entirely be implemented via `import_multiple`.


components/logins/src/logins.udl
```diff
@@ -2,32 +2,72 @@
...
+    [Throws=LoginsStorageError]
+    string create_key();
 
-// Converted from mozilla.appservices.logins.protobuf.rs
-dictionary Login {
-    string id;
+    // Decrypt a string into an EncryptedFields struct. This will be used to get the credentials
+    // from a `Login` dict.
+    [Throws=LoginsStorageError]
+    EncryptedFields decrypt_fields([ByRef]string ciphertext, [ByRef]string encryption_key);
```
mhammond, line 15: actually, this can probably die as `Login` has a method that does the same thing.

components/logins/src/db.rs
```diff
@@ -306,43 +307,18 @@ impl LoginDb {
...
-        }
-        if login.times_used == 0 {
-            login.times_used = 1;
-        }
-
+    // The single place we insert new rows or update existing local rows.
+    // just the SQL - no validation or anything.
+    fn insert_new_login(&self, login: &Login) -> Result<()> {
         let sql = format!(
-            "INSERT OR IGNORE INTO loginsL (
```
mhammond, line 337: The "OR IGNORE" previously here seems like a footgun, so I removed it.

components/logins/src/db.rs
```diff
@@ -373,33 +348,93 @@ impl LoginDb {
...
+            rows_changed, 1,
+            "adding a new record must have changed 1 row"
+        );
+        Ok(())
+    }
+
+    fn update_existing_login(&self, login: &Login) -> Result<()> {
+        // assumes the "local overlay" exists, so the guid must too.
+        let sql = format!(
+            "UPDATE loginsL
```
mhammond, line 375: ditto here - no "OR" clauses because all callers of this should be 100% confident there's no existing record with the same guid.

components/logins/src/db.rs
```diff
@@ -373,33 +348,93 @@ impl LoginDb {
...
+        self.insert_new_login(&result)?;
         tx.commit()?;
-        Ok(login)
+        Ok(result)
     }
 
-    pub fn import_multiple(&self, logins: &[Login]) -> Result<MigrationMetrics> {
+    pub fn import_multiple(
+        &self,
+        logins: Vec<Login>,
```
mhammond, line 435: stuff got a but messy because this public function wants to add new records via a `Login` struct rather than an `UpdatableLogin` (eg, `validate_and_fixup` for a `Login` struct is only needed because this exists). This was the thing that pushed me over the edge of creating the `LoginFields` struct rather than flattening them.

components/logins/src/login.rs
```diff
@@ -490,37 +509,138 @@ impl Login {
...
+            .map(|new_fields| Self {
+                id: self.id.clone(),
+                fields: new_fields,
+                enc_fields: self.enc_fields.clone(),
+                time_created: self.time_created,
+                time_password_changed: self.time_password_changed,
+                time_last_used: self.time_last_used,
+                times_used: self.times_used,
+            }))
+    }
```
bendk, line 536: I'm not sure if this should be implemented on the `Login` struct since there's no way to validate the encrypted fields.  It seems better to validate the `LoginFields`, then decrypt `enc_fields` and use the `EncryptedLogin.validate_and_fixup()` code.

bendk (commented):
I like the overall approach here.  The only question I had was the validation code

I've gone forth in my mind on nested vs non-nested structs and at this point I think a) there are a bunch of minor differences that mostly balance each other out and b) I think I slightly prefer the nested approach.

components/logins/src/db.rs
```diff
@@ -555,126 +536,131 @@ impl LoginDb {
...
+        login: UpdatableLogin,
+        encdec: &EncryptorDecryptor,
+    ) -> Result<Login> {
+        let guid = Guid::new(sguid);
+        let now_ms = util::system_time_ms_i64(SystemTime::now());
         let tx = self.unchecked_transaction()?;
+
+        // XXX - it's not clear that throwing here on a dupe is the correct thing to do - eg, a
+        // user updated the username to one that already exists - the better thing to do is
+        // probably just remove the dupe.
```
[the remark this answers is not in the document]
lougeniaC64: I agree about removing the dupe. The only caveat I'd have is considering what would happen if the dupe has an older `time_password_changed` timestamp. In that case we probably want to keep the dupe or at minimum the dupe's metadata.

examples/sync-pass/src/sync-pass.rs
```diff
@@ -89,24 +89,13 @@ fn update_string(field_name: &str, field: &mut String, extra: &str) -> bool {
...
-        encdec.decrypt(field).unwrap(),
-        extra
-    ));
-    if let Some(s) = opt_s {
-        *field = encdec.encrypt(&s).unwrap();
-        true
-    } else {
-        false
-    }
+fn update_encrypted_fields(fields: &mut EncryptedFields, extra: &str) {
```
lougeniaC64, line 92: nit: Seeing `EncryptedFields` outside of the `Login` struct makes me want it to be renamed `EncryptedLoginFields` so this struct isn't construed as a generic struct not associated with logins. Although perhaps it would read as redundant nested inside of the `Login` struct but maybe worth it? Do what you will with this.

components/logins/src/db.rs
```diff
@@ -246,13 +247,13 @@ impl LoginDb {
...
                     .ok()
-                    .and_then(|login| Url::parse(&login.origin).ok());
+                    .and_then(|login| Url::parse(&login.fields.origin).ok());
                 let this_host = login.as_ref().and_then(|url| url.host());
                 match (&base_host, this_host) {
                     (Host::Domain(base), Some(Host::Domain(look))) => {
                         // a fairly long-winded way of saying
-                        // `login.origin == base_domain ||
-                        //  login.origin.ends_with('.' + base_domain);`
+                        // `fields.origin == base_domain ||
```
lougeniaC64, line 255: nit: `login.fields.origin` instead of `fields.origin`

components/logins/src/logins.udl
mhammond: well, it does in rust, but it's a dictionary in UDL, so this needs to remain.

components/logins/src/db.rs
mhammond: I'm not sure what you are asking here, but to restate what you said, validating a `Login` does just validate the `fields` and `enc_fields` members. This means there's no validation on the metadata, which possibly should happen, but the existing code didn't check that either IIUC.

components/logins/src/db.rs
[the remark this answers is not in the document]
mhammond: I think an important consideration here will be the existing Fenix code - eg, will the UI update correctly if we do that deletion? It would clearly be bad if the login we removed stayed in the list of logins. Sadly I can't build Fenix currently for reasons I don't understand, so can't check.

The metadata is an interesting question - maybe we should delete the one being updated as that has entirely wrong metadata now that one of the keys has changed.

components/logins/src/login.rs
```diff
@@ -490,37 +509,138 @@ impl Login {
...
+}
+
+/// The encrypted fields.
+#[derive(Debug, Clone, Hash, PartialEq, Serialize, Deserialize, Default)]
+pub struct EncryptedFields {
```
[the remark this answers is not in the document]
mhammond: I like `SecureFields` - I'll make that change tomorrow!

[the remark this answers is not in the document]
mhammond: Or as Lougenia comments below, `SecureLoginFields`? Or `LoginSecureFields` (doesn't read as well, but is more consistent!)? WDYT?

components/logins/src/login.rs
```diff
@@ -490,37 +509,138 @@ impl Login {
...
+    // prior to per-field encryption the `username` was allowed to be null.
+    // We could make it `Option<String>` but that doesn't seem valuable, as
+    // an empty string and None mean the same thing anyway.
+    // XXX - although this might end up being important in the sync payload (as
+    // in, the item missing entirely from the payload?)
+    // We should check desktop's semantics.
+    pub username: String,
+    pub password: String,
+}
```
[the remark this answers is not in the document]
mhammond: hrm, interesting.

components/logins/src/db.rs
```diff
@@ -555,126 +536,131 @@ impl LoginDb {
...
-        self.check_for_dupes(&login)?;
-        Ok(login)
+    pub fn fixup_and_check_for_dupes(
+        &self,
+        guid: &Guid,
+        fields: LoginFields,
+        enc_fields: EncryptedFields,
+        encdec: &EncryptorDecryptor,
+    ) -> Result<(LoginFields, EncryptedFields)> {
+        let fields = fields.fixup()?;
```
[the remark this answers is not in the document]
mhammond: yes, nice catch, thanks

components/logins/src/login.rs
mhammond: hrm, yeah, that's probably correct.

components/logins/src/login.rs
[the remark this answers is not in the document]
bendk: I like the readability of `SecureLoginFields`

components/logins/src/login.rs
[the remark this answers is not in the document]
mhammond: I really low-cost thing was to just `[#serde(rename="...")]` the fields, so it is `{"u": "username", "p":"password"}` - not as tight as a custom serializer could do, but probably good enough - WDYT?

components/logins/src/login.rs
```diff
@@ -490,37 +509,142 @@ impl Login {
...
+    }
+}
+
+/// The encrypted fields.
+#[derive(Debug, Clone, Hash, PartialEq, Serialize, Deserialize, Default)]
+pub struct EncryptedFields {
+    // prior to per-field encryption the `username` was allowed to be null.
+    // We could make it `Option<String>` but that doesn't seem valuable, as
+    // an empty string and None mean the same thing anyway.
+    // XXX - although this might end up being important in the sync payload (as
```
mhammond, line 545: Note to self: actually check this!

mhammond:
I pushed up the rename to `SecureLoginFields` and a few other changes. I've a few other comments to address before I'll re-request review, but I think this is in good enough shape to unblock the Fenix and iOS efforts?

components/logins/src/login.rs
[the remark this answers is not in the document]
bendk: Yeah, I think that's a good balance between simplicity and reducing the field size.

mhammond:
After doing all review comments I merged main into logins-per-field-encryption, then rebased this branch against that. I think this is ready to roll now, so re-flagging reviews.

bendk (approved):
Looks good to me, we can keep iterating in the main branch.

bendk:
I think we should consider renaming the `UpdatableLogin` struct.  We use it for both `add()` and `update()` which is a bit confusing.  What able `SavableLogin` or `MutableLogin`?

lougeniaC64 (approved):
LGTM!

mhammond:
> I think we should consider renaming the `UpdatableLogin` struct. We use it for both `add()` and `update()` which is a bit confusing. What able `SavableLogin` or `MutableLogin`?

Not a huge fan of either of those names TBH and very lightly prefer `UpdatableLogin` over them. How about `Editable`?

mhammond:
> How about `Editable`?

`Writable`?

(Don't really care, but worth trying to get something we all like)

bendk:
`Editable` or `Writable` are both good with me.

bendk:
I'm for merging this one into `logins-per-field-encryption` ASAP and maybe also merging the extra commits here: https://github.com/bendk/application-services/tree/logins-per-field-encryption.  I feel like merging frequently is the best way to deal with the API being in flux.  Definitely don't let my comments abut the struct names slow this down.

PR #4363, opened 2021-07-28
Split out some Swift code from the iOS megazord

rfk:
We currently have many little bits of Swift code living in the `./megazords/ios` directory that should logically "belong to" a particular component, such as the Swift API for Viaduct and some keychain helpers for FxAClient. This PR pulls them out of the megazord directory and puts them in individual component directories alongside the code that they relate to.

I think it's an improvement purely for consistency, but the real goal here is to make it easier to distribute all this code as independent Swift package products rather than as one big monolith.

For reviewers: the interesting parts here are the renamed files, each of which is an individual commit with some commentary on the move. The changes in `project.pbxproj` and just tracking the renames.


rfk:
@tarikeshaq r? since we discussed working together on the Swift packaging stuff.

@mhammond r? specifically for the move of some of the sync-related helpers into the `sync15` component, does that make sense to you?

tarikeshaq (approved):
This makes sense to me from the perspective of keeping the swift code closer to the component (for easier packaging)

I don't completely understand the implications in terms of building the megazord (while the swift packaging work is still in progress) - but CI is green so I'm not worried!

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 51;
```
skhamis, line 6: I know it's XCode doing this but feels a little weird we're going down a version rather than up

skhamis (commented):
Nothing huge to add other than really like these changes! I also think it'll be better for future refactoring of these components to not have to touch megazord and cause some unintended side-effects. Since there's only one target can't imagine this disrupting anything -- LGTM!

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
rfk: Yeah, it's weird, I think I will just manually revert this change

PR #4368, opened 2021-07-29
Have Nimbus optionally import Glean as an external Swift module.

### LINKED ISSUE #4308: Make Nimbus (optionally) consume Glean as a Swift Package dependency
The Nimbus swift bindings currently use various functions exposed by Glean, and broadly assume that they will be built together with Glean in a single combined Swift module. When we move to Swift Packages that’s no longer going to work, since consumers will want to depend on Glean as its own separate Swift package.

Swift as a neat `#if canImport()` feature that I hope to use to help with this, described [here](https://www.hackingwithswift.com/example-code/language/how-to-check-whether-a-module-is-available-using-canimport). Broadly:

* If `#if canImport(Glean)` to check whether Glean is available for import; if it is then we’re consuming Glean via an external package, if it isn’t then we’re building together with glean in a megazord bundle.
* If we’re using Glean as an external package, add `import Glean` in the relevant places.

This is a little complicated by the fact that Nimbus currently uses several Glean APIs that are not part of its official public API, such as the `Sysctl` and `Logger` helpers. We’ll need to either ask the Glean team to expose them as public APIs, or replace them with something else (perhaps by copying the relevant bits of code directly from Glean).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2677)
┆Story Points: 8
┆Epic: Unified a-s swift packages for iOS
┆Sprint End Date: 2021-08-06

rfk:
Currently on iOS, Nimbus and Glean are always compiled together as
part of the same Swift module in the iOS megazord. This means that
Nimbus can see and use many of Glean's internal helper functions,
which in fact it does.

When we switch to distributing via Swift packages, we'll need to
be more careful about inter-module hygiene. I think ideally the
Nimbus Swift code would consume Glean as an ordinary Swift dependency
via `import Glean`.

To make that work while also preserving the current behaviour for
current consumers, I have:

* Added a conditional import of Glean, which gets ignored in the
  existing megazord but makes Glean available in a Swift package build.
* Copied the internal helper modules that we were using from Glean
  into the `nimbus/ios` directory, so that this directory can be
  compiled as a standalone Swift module using only public Glean
  APIs.

Importantly, I have *not* included the copied utility files in the
iOS megazord Xcode project. That build already uses the ones from
Glean and would error out if we included duplicates.

All in all, this doesn't seem like a sustainable setup, but perhaps
it can be the first step towards something cleaner. Maybe we could
make a shared utility module that both Glean and Nimbus can depend
on? Maybe Glean could commit to making these utilities part of its
public API surface?

Suggestions welcome! What I know is that this helps unblock the work
on publishing as a Swift package in the meantime. Fixes #4308.

### DIFF
--- a/components/nimbus/ios/Nimbus/Nimbus.swift
+++ b/components/nimbus/ios/Nimbus/Nimbus.swift
@@ -4,6 +4,17 @@
 
 import Foundation
 
+// Depending on build setup, we may be importing Glean as a Swift module
+// or we may be compiled together with it. This detects whether Glean is
+// an external module and makes it available to our entire package if so.
+//
+// Note that the files under `./Utils` are copies of internal files from
+// Glean, and it's very important they they be excluded from any build
+// that is compiling us together with Glean.
+#if canImport(Glean)
+    @_exported import Glean
+#endif
+
 public class Nimbus: NimbusApi {
     private let nimbusClient: NimbusClientProtocol
 
--- a/components/nimbus/ios/Nimbus/Utils/Logger.swift
+++ b/components/nimbus/ios/Nimbus/Utils/Logger.swift
@@ -1,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import os.log
+
+class Logger {
+    private let log: OSLog
+
+    /// Creates a new logger instance with the specified tag value
+    ///
+    /// - parameters:
+    ///     * tag: `String` value used to tag log messages
+    init(tag: String) {
+        self.log = OSLog(
+            subsystem: Bundle.main.bundleIdentifier!,
+            category: tag
+        )
+    }
+
+    /// Output a debug log message
+    ///
+    /// - parameters:
+    ///     * message: The message to log
+    func debug(_ message: String) {
+        log(message, type: .debug)
+    }
+
+    /// Output an info log message
+    ///
+    /// - parameters:
+    ///     * message: The message to log
+    func info(_ message: String) {
+        log(message, type: .info)
+    }
+
+    /// Output an error log message
+    ///
+    /// - parameters:
+    ///     * message: The message to log
+    func error(_ message: String) {
+        log(message, type: .error)
+    }
+
+    /// Private function that calls os_log with the proper parameters
+    ///
+    /// - parameters:
+    ///     * message: The message to log
+    ///     * level: The `LogLevel` at which to output the message
+    private func log(_ message: String, type: OSLogType) {
+        os_log("%@", log: self.log, type: type, message)
+    }
+}
--- a/automation/tests.py
+++ b/automation/tests.py
@@ -286,7 +286,8 @@ def swift_format():
     if on_darwin():
         run_command([
             'swiftformat', 'megazords', 'components/*/ios',
-            '--exclude', '**/Generated', '--lint', '--swiftversion', '4',
+            '--exclude', '**/Generated', '--exclude', 'components/nimbus/ios/Nimbus/Utils',
+            '--lint', '--swiftversion', '4',
         ])
     else:
         print("WARNING: skipping swiftformat on non-Darwin host")
--- a/components/nimbus/ios/Nimbus/Utils/Sysctl.swift
+++ b/components/nimbus/ios/Nimbus/Utils/Sysctl.swift
@@ -1,0 +1,162 @@
+// swiftlint:disable line_length
+// REASON: URLs and doc strings
+
+// Copyright © 2017 Matt Gallagher ( http://cocoawithlove.com ). All rights reserved.
+//
+// Original: https://github.com/mattgallagher/CwlUtils/blob/0e08b0194bf95861e5aac27e8857a972983315d7/Sources/CwlUtils/CwlSysctl.swift
+// Modified:
+//   * iOS only
+//   * removed unused functions
+//   * reformatted
+//
+// ISC License
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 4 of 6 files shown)


components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -4,6 +4,17 @@
...
 
+// Depending on build setup, we may be importing Glean as a Swift module
+// or we may be compiled together with it. This detects whether Glean is
+// an external module and makes it available to our entire package if so.
+//
+// Note that the files under `./Utils` are copies of internal files from
+// Glean, and it's very important they they be excluded from any build
+// that is compiling us together with Glean.
+#if canImport(Glean)
+@_exported import Glean
```
rfk, line 15: The use of `@_exported` here means we can avoid having to `import Glean` in every file that wants it, and was the most expedient option. I'm open to the suggestion that we should avoid it given that the leading underscore indicates it is unstable, but on the other hand, firefox-ios imports the existing megazord using `@_exported import MozillaAppServices` so I don't feel too bad about it...

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 51;
```
rfk, line 6: This, I don't know what it is, some Xcode mystery...

components/nimbus/ios/Nimbus/Utils/Logger.swift
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rfk, line 1: The rest of the files here are copy-pastes of stuff from Glean, which I don't feel great about doing, but in order to build as a Swift package we need to have all the necessary Swift code here under the `nimbus/ios` directory.

I would love suggestions for something better to do here @travis79 @jhugman @badboy. Is it worth trying to put these more explicitly into some sort of shared utility module that both projects can depend on?

travis79: I wouldn't mind having a shared utility module, but I guess I'm having a hard time understanding why Nimbus would need this copied over if it just depends on Glean.  Can the Nimbus Swift Package not depend on the Glean Swift Package and just import these functions directly for some reason? Even if we are megazording, could the iOS megazord itself just pull in the Glean dependence from the Swift Package instead of having Glean as a submodule, or do we lose the advantage of compiling all the Rusty bits together by doing that?

travis79: I'll try and make some time to check out your branch to get a better understanding of all of this today, it's still early and I may not have had enough ☕ yet to grep this...

rfk: > Can the Nimbus Swift Package not depend on the Glean Swift Package and just import these functions directly for some reason?

My high-level impression is that these utilities are not exported as part of the public API (e.g. they're not marked `public`) so one cannot import them from the Glean Swift Package. But I'll admit I haven't tried particularly hard to see if they can be imported, I'll take another look.

rfk: So, concrete example, if we wanted to be able to import the `Sysctl` namespace from Glean then we'd need Glean to mark this struct as `public` and commit to it being part of the public API of the module:

* https://github.com/mozilla/glean/blob/d1f57a850be8987f9d59bfeea79fb2cf3a38ea6f/glean-core/ios/Glean/Utils/Sysctl.swift#L32

/cc @badboy for visibility on the discussion as well, because I could easily be missing something.

badboy: They are private intentionally. It's not something I would want to make part of the public API really.

Now for the logger I think it might be worth standardizing on one approach, and then yes, we can probably unify this somehow.
For Sysctl, why does Nimbus need that? For now this should only be needed in Glean-internal code, so no need to expose it further.

components/nimbus/ios/Nimbus/Nimbus.swift
badboy: I think using the "unstable" API here is fine. I see that used in the wild quite often and if it ever comes to it a stabilization proposal will need to argue real good why behavior should change significantly (at which point we can also still change what we do)

megazords/ios/MozillaAppServices.xcodeproj/project.pbxproj
badboy: I _think_ this means older Xcode might refuse to load that project, but IMO that's fine 🤷

components/nimbus/ios/Nimbus/Utils/Logger.swift
travis79: > They are private intentionally. It's not something I would want to make part of the public API really.
> 
> Now for the logger I think it might be worth standardizing on one approach, and then yes, we can probably unify this somehow.
> For Sysctl, why does Nimbus need that? For now this should only be needed in Glean-internal code, so no need to expose it further.

Nimbus uses sysctl just to ensure that it grabs the device arch, mfg, and model in the same format as Glean, and it borrows the `getLocaleTag` func to ensure that targeting matches the Glean client_info.

travis79: I recall discussing in the past of creating some sort of "values provider" that both Glean and Nimbus (and anything else that needed consistent values for device info) could use, but at the time, I think it was just simpler to borrow them.

badboy: I see. Hm, for now we _could_ expose them, but still declare them not part of the actual Glean API. That way we Nimbus could use them, but apart from Nimbus we don't need to be cautious of brekaing consumers.

rfk: I'm happy to take on some amount of legwork to get to a good outcome here, even if it's a PR to Glean :-)

But, @travis79 @badboy I'd like to lean on your collective wisdom to suggest the right path forward.

travis79: I'd say let's just keep the copied code for now, and we can remove it if Glean opens these functions up publicly. I don't see either of these functions changing on the Glean side anytime soon, so I think the risk of them getting out of sync is very low.

travis79 (approved):

PR #4371, opened 2021-07-29
Fixes #4370 - History metadata enums and struct access

### LINKED ISSUE #4370: History metadata enums and structs are inaccessible for swift
Enums and Structs inside a library don't have accessible initializer and require public inits.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/SYNC-2712)

nbhasin2:
Enums and Structs inside a library don't have accessible initializer and require public inits

### DIFF
--- a/components/places/ios/Places/HistoryMetadata.swift
+++ b/components/places/ios/Places/HistoryMetadata.swift
@@ -11,7 +11,7 @@ public enum DocumentType: Int32 {
     case regular = 0
     case media = 1
     case unknown = -1
-    
+
     public init(rawValue: Int32) {
         switch rawValue {
         case 0: self = .regular
@@ -28,7 +28,7 @@ public struct HistoryMetadataKey {
     public let url: String
     public let searchTerm: String?
     public let referrerUrl: String?
-    
+
     public init(url: String, searchTerm: String?, referrerUrl: String?) {
         self.url = url
         self.searchTerm = searchTerm
@@ -44,7 +44,7 @@ public enum HistoryMetadataObservation {
     case viewTimeObservation(Int32)
     case documentTypeObservation(DocumentType)
     case unknown
-    
+
     public init(value: Any) {
         switch value {
         case let title as String: self = .titleObservation(title)
--- a/components/places/ios/Places/HistoryMetadata.swift
+++ b/components/places/ios/Places/HistoryMetadata.swift
@@ -10,6 +10,15 @@ import Foundation
 public enum DocumentType: Int32 {
     case regular = 0
     case media = 1
+    case unknown = -1
+    
+    public init(rawValue: Int32) {
+        switch rawValue {
+        case 0: self = .regular
+        case 1: self = .media
+        default: self = .unknown
+        }
+    }
 }
 
 /**
@@ -19,6 +28,12 @@ public struct HistoryMetadataKey {
     public let url: String
     public let searchTerm: String?
     public let referrerUrl: String?
+    
+    public init(url: String, searchTerm: String?, referrerUrl: String?) {
+        self.url = url
+        self.searchTerm = searchTerm
+        self.referrerUrl = referrerUrl
+    }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


components/places/ios/Places/HistoryMetadata.swift
```diff
@@ -10,6 +10,15 @@ import Foundation
 public enum DocumentType: Int32 {
     case regular = 0
     case media = 1
+    case unknown = -1
+    
+    public init(rawValue: Int32) {
```
grigoryk, line 15: So what's your use case for this? Can't you just do `DocumentType.media` when creating instances? I suppose this may be helpful when deserializing, but I don't think you'd need to persist a `DocumentType` yourself manually.

Also, these numbers aren't entirely random - they need to align w/ what other platforms are using. And currently we don't have a `-1` present elsewhere. So unless you absolutely need this change, I'd remove at least the `unknown` case.

components/places/ios/Places/HistoryMetadata.swift
```diff
@@ -28,6 +43,16 @@ public enum HistoryMetadataObservation {
     case titleObservation(String)
     case viewTimeObservation(Int32)
     case documentTypeObservation(DocumentType)
+    case unknown
```
grigoryk, line 46: What is an `unknown` observation, and why do you need it? Does seem necessary to me.

components/places/ios/Places/HistoryMetadata.swift
```diff
@@ -28,6 +43,16 @@ public enum HistoryMetadataObservation {
...
+    
+    public init(value: Any) {
+        switch value {
+        case let title as String: self = .titleObservation(title)
```
grigoryk, line 50: What if we have string observations that aren't about `title`, or Int32 that aren't `time`? Also, what's the use case here for the `init`, can't you just do `HistoryMetadataObservation.titleObservation("hello")`?

components/places/ios/Places/Places.swift
```diff
@@ -910,6 +910,8 @@ public class PlacesWriteConnection: PlacesReadConnection {
                 msg.documentType = documentType.rawValue
             case let .titleObservation(title):
                 msg.title = title
+            case .unknown:
```
grigoryk, line 913: Looks like a bug that may be easy to miss - e.g. somehow you end up `init`-ing HistoryMetadataObservation wrong, then pass it along here, and never notice that nothing actually happens.

Partially why I don't like being able to represent invalid states! E.g. if `unknown` isn't a thing, you can't pass invalid data into this method at all.

grigoryk (changes requested):
This is generally fine, but I'd like to know why you need these things before landing :)

components/places/ios/Places/HistoryMetadata.swift
nbhasin2: I really don't need unknown but removing that causes swift errors for our initializers so I added additional case for default as unknonw. Happy to change / rename it as needed.

components/places/ios/Places/HistoryMetadata.swift
nbhasin2: The major issue is that we need some sort of initializer for using it on Firefox iOS side otherwise structs and enums are inaccessible. This is just something how swift hasn't gotten to or a limitation I would say as all this is extra work 😓

grigoryk: Oh wow, so that's actually a thing! https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID3 (memberwise init section). That's pretty annoying :)

Maybe throw something like `invalidArgumentException` to avoid needing to have an `unknown` case..?

grigoryk (dismissed):
If you can remove illegal states (e.g. unknown case) by adding overloading your init methods, please do so :)

components/places/ios/Places/HistoryMetadata.swift
grigoryk: What about function overloading? Separate init functions for each data type? Then you can remove unknown, i think.

grigoryk (dismissed):

rfk (approved):
r+ based on conversation in slack

PR #4375, opened 2021-08-02
Refactor logins migration

skhamis:
Accidentally delete the previous branch https://github.com/mozilla/application-services/pull/4341 trying to fixup rebasing this ontop of logins-per-field-encryption so creating a new PR as I can't change the source of the previous one...

### Current Migration Plan
Step 1: We need to iterate through the old DB and throw it all into a struct that contains all the relevant information
```rust
struct MigrationPlan {
    logins: HashMap<String, MigrationLogin>,
}
```
```rust
struct MigrationLogin {
    local_login: Option<LocalLogin>,
    mirror_login: Option<MirrorLogin>,
    status: MigrationStatus,
}
```
Step 2: We need to perform options on the struct before throwing it into the new DB
 - Step 2a: Find any mismatched data between local/mirro and either throw away or attempt to reconcile
   - Case 1: If mirror has `is_override=1` but no matching local login
 - Step 2b: Find any logins we can fixup [LOCAL ONLY] and flip the `local_modified=true`

Step 3: Insert the struct data into the new DB
 - Step 3a: Capture the right metrics

Screen Shot 2021-08-30 at 1 55 07 PM

---

  - Note:

### DIFF
--- a/components/logins/src/migrate_sqlcipher_db.rs
+++ b/components/logins/src/migrate_sqlcipher_db.rs
@@ -7,8 +7,7 @@
 use crate::db::{MigrationMetrics, MigrationPhaseMetrics};
 use crate::encryption::EncryptorDecryptor;
 use crate::error::*;
-use crate::sync::SyncStatus;
-use crate::sync::{LocalLogin, MirrorLogin};
+use crate::sync::{LocalLogin, MirrorLogin, SyncStatus};
 use crate::util;
 use crate::LoginStore;
 use crate::{EncryptedLogin, Login, LoginFields, RecordFields, SecureLoginFields};
@@ -17,7 +16,7 @@ use sql_support::ConnExt;
 use std::collections::HashMap;
 use std::ops::Add;
 use std::path::Path;
-use std::time::Instant;
+use std::time::{Instant, SystemTime};
 use sync15::ServerTimestamp;
 
 #[derive(Debug)]
@@ -60,12 +59,13 @@ struct MigrationLogin {
     status: MigrationStatus,
 }
 
+// TODO: Kept this here as part of the initial design but doesn't seem needed as we go through this
 #[derive(Debug)]
 enum MigrationStatus {
     Processing,
-    Success,
-    Fixed,
-    Failed,
+    // Success,
+    // Fixed,
+    // Failed,
 }
 
 // Simplify the code for combining migration metrics
@@ -190,9 +190,11 @@ fn generate_plan_from_db(
             Ok(login) => {
                 let l_login = LocalLogin {
                     login: login.encrypt(&encryptor)?,
-                    local_modified: util::system_time_millis_from_row(row, "local_modified")?,
-                    is_deleted: row.get("is_deleted")?,
-                    sync_status: SyncStatus::from_u8(row.get("sync_status")?)?,
+                    local_modified: util::system_time_millis_from_row(row, "local_modified")
+                        .unwrap_or(SystemTime::now()),
+                    is_deleted: row.get("is_deleted").unwrap_or_default(),
+                    sync_status: SyncStatus::from_u8(row.get("sync_status").unwrap_or_default())
+                        .unwrap_or(SyncStatus::New),
                 };
                 let key = l_login.login.record.id.clone();
                 migration_plan
@@ -220,8 +222,10 @@ fn generate_plan_from_db(
             Ok(login) => {
                 let m_login = MirrorLogin {
                     login: login.encrypt(&encryptor)?,
-                    server_modified: ServerTimestamp(row.get::<_, i64>("server_modified")?),
-                    is_overridden: row.get("is_overridden")?,
+                    server_modified: ServerTimestamp(
+                        row.get::<_, i64>("server_modified").unwrap_or_default(),
+                    ),
+                    is_overridden: row.get("is_overridden").unwrap_or_default(),
                 };
 
                 let key = m_login.login.record.id.clone();
@@ -571,7 +575,6 @@ mod tests {
     use crate::EncryptedLogin;
     use rusqlite::types::ValueRef;
     use std::path::PathBuf;
-    use std::time::SystemTime;
     use sync_guid::Guid;
 
     static TEST_SALT: &str = "01010101010101010101010101010101";
@@ -631,7 +634,9 @@ mod tests {
                 ('a', 'test', 'password', 'https://www.example.com', NULL, 'https://www.example.com',
                 'username', 'password', 1000, 1000, 1, 10, 1000, 0, 2),
                 ('b', 'test', 'password', 'https://www.example.com', 'https://www.example.com', 'https://www.example.com',
-                'username', 'password', 1000, 1000, 1, 10, 1000, 0, 2);
+                'username', 'password', 1000, 1000, 1, 10, 1000, 0, 2),
+                ('bad_sync_status', 'test', 'password', 'https://www.example2.com', 'https://www.example.com', 'https://www.example.com',
+                'username', 'password', 1000, 1000, 1, 10, 1000, 0, 'invalid_status');
         ";
 
         const INSERT_LOGINS_M: &str = "
@@ -989,7 +994,6 @@ mod tests {
 
         assert_eq!(metrics.num_processed, 2);
         assert_eq!(metrics.num_succeeded, 2);
-        assert_eq!(metrics.num_failed, 1);
         assert_eq!(metrics.total_duration > 0, true);
     }
 }
--- a/components/logins/src/migrate_sqlcipher_db.rs
+++ b/components/logins/src/migrate_sqlcipher_db.rs
@@ -115,12 +115,16 @@ mod tests {
             ALTER TABLE loginsL RENAME passwordEnc to password;
             ALTER TABLE loginsM RENAME usernameEnc to username;
             ALTER TABLE loginsM RENAME passwordEnc to password;
-            INSERT INTO loginsL(guid, username, password, origin,
-            httpRealm, formActionOrigin, usernameField, passwordField, timeCreated, timeLastUsed,
+            ALTER TABLE loginsL RENAME origin to hostname;
+            ALTER TABLE loginsL RENAME formActionOrigin to formSubmitURL;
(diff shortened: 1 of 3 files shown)


components/logins/src/store.rs
```diff
@@ -120,6 +120,19 @@ impl LoginStore {
         self.db.lock().unwrap().update(login)
     }
 
+    // TODO: STUB FOR THE NEW API THAT WILL BE USED AND WILL BE CHANGED
+    pub fn add_or_update(&self, login: Login) -> Result<()> {
```
lougeniaC64, line 124: We decided to keep the `add` and `update` functions separate in order to avoid the necessary UI updates so I think you don't need this stub.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -4,12 +4,20 @@
 
 // Code to migrate from an sqlcipher DB to a plaintext DB
 
-use crate::db::MigrationMetrics;
+use crate::db::{MigrationMetrics, MigrationPhaseMetrics};
+use crate::encryption::EncryptorDecryptor;
 use crate::error::*;
-use crate::schema;
-use rusqlite::{Connection, NO_PARAMS};
+use crate::sync::SyncStatus;
```
lougeniaC64, line 10: nit: you can include `SyncStatus` in the import statement below

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,71 +71,582 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+    cipher_conn: &mut Connection,
+    new_db_store: LoginStore,
+    encryption_key: &str,
+) -> Result<MigrationMetrics> {
+    // encrypt the username/password data
+    let encryptor = EncryptorDecryptor::new(encryption_key)?;
+
+    // Migrate tables separately due to specific columns in each needing
+    // to be ported over
+    let local_metrics = migrate_local_logins(&cipher_conn, &new_db_store, &encryptor)?;
```
lougeniaC64, line 85: I'm wondering if the `migrate_mirror_logins` and `migrate_local_logins` functions should share the same transaction. I don't think we'd want either table to be migrated if the other isn't. IIRC in logins a combination of the records in the local and mirror table represent what was last synced so having them be in incongruous states seems undesirable. If they share this transaction, I would argue that `migrate_sync_metadata` should be included and metrics (with the exception of fixup metrics) should be collected for it.

lougeniaC64:
I left a few comments in-line but I also want to raise the fact that I'm unclear on where we landed as far as what state we'll leave the database in if we can't successfully migrate a table. It may be that my thoughts on transactions aren't in line with that decision. Disregard if that's the case.

components/logins/src/migrate_sqlcipher_db.rs
skhamis: Hmmm that does raise a few more questions too, we also probably need to revisit the discussion of end-to-end flow of rollbacks (rollforwards?) and everything inbetween:

1. I initially had it as separate tx just for ease of getting this code out, but definitely agree that maybe sharing the same transaction may help with the potential rollback logic.
  - However, it will be rare for an entire DB to be corrupted and most likely it'll be a few rows that will not be able to be fixed up/migrating at that point, do we also skip if we find that guid in the other table? From my understanding, this will only happen if they have a sync account but is migrating before syncing.
  - Since we're doing row by row instead of the entire table, if we cannot fixup a row, we just skip it -- lessening the reliance on needing to do any kind of major rollback (unless something goes horribly wrong with trying to fixup that row)
2. Since this is kind of a one-way street kind of migration. We most likely need to have a discussion of how we want to proceed if we can't even get the DB to open (I think our current stance is, trash it and start fresh) but as mentioned above, since we are doing row-by-row there is less chance of really bad things happening since we can just skip that row and move on.

lougeniaC64: > rollbacks (rollforwards?) 

LOL, right. Rolling back probably isn't a viable option. We'd have to migrate the database away from SQLCipher at some point and what would change between the time the rollback occurred and the next migration attempt that would increase the likelihood of success? So, yes, rollforwards I think.

>     1. I initially had it as separate tx just for ease of getting this code out, but definitely agree that maybe sharing the same transaction may help with the potential rollback logic.

Yeah after I suggested the shared transaction approach, I started second-guessing it. If we're not rolling back, and I don't think we should, then we'd either be skipping records that can't be ported over or trashing databases in the corrupted db scenario. 

However I think we do need to ensure that the tables are in sync (🤡) with each other. If that's the case, we could remove records in the corresponding tables if they can't be migrated as you suggested. However what's to stop us from resetting sync and forcing a sync after the migration? That's a Friday EOD question but I'm not deleting it.

mhammond: Yeah, I agree entirely with this - the problem doing local then the mirror is exactly that the tables might get out of sync. But the more I think about it, that *might* actually be an easily recoverable outcome: Consider:

* failing to migrate the `local` record but successfully migrating the `mirror` record: That means "local changes" are lost. Non sync users have lost this record entirely, Sync users will still work based on the mirror, but will have lost a local change waiting to sync up. That should be rare. I think the only problem here is that `is_overridden` would be wrong - so we should "migrate" that based on whether the actual record exists rather than the old DB value.

* successful `local` record but failing mirror record: should be impossible for non-sync users (they have no mirror records), and I think this would still work fine for sync users if we set `sync_status` to `SyncStatus::New/Changed`

A final consideration is: what happens when we've done a fixup? Does a fixup mean a "local change" (ie, that we should mark the item as locally modified and sync it up?) Fixing up a mirror record for a sync user almost certainly means a corrupt DB - that should exactly reflect the server - but we probably should not discard it entirely unless we have a 'local' record (see above, a "fixed up" local mirror record is better than discarding it because sync users, but we almost certainly want to arrange for us to re-download the mirror for that record)

I think the only way to capture all this (except metadata) as a single loop - ideally with a union statement, but if 2 read loops made it much easier that'd be ok. Then a struct/enum/handwave that represents the entire set of local states a GUID can be in and some rust pattern matching to deal with it. As as @lougeniaC64 mentions, a reasonable decision in some cases might be "also decline to migrate metadata, so we do a full server reconcile")

mhammond (commented):
Thanks Sammy, this is looking great! Sorry it's been some time since I last looked in detail and while this is in great shape, I do think we should consider doing the migration as a single loop for the local/mirror records, as I tried to describe below :) We can chat more about this tomorrow.

skhamis:
@bendk @mhammond as per conversation -- this is not meant to be the end-production migration but to get us 90% there. I will have a follow up document as well as update this PR issue to denote things to improve on in the future and edge cases to think about

mhammond (approved):
Thanks Sammy, this is in fairly good shape! It looks like the "happy case" is going to be happy, so I'm OK with landing this (although the test failures should be addressed before we do). However, I do think some of the error edge cases aren't quite going to work correctly, so there's still some stuff to do in followups - we can discuss how exactly to manage that later.

FWIW, I'm about to put a PR up to your branch, which might help explain some of the edge cases I describe here - that's really more to help explain what I said here - you can either suck that in before merging (it has `[ignore]` for the test skeletons) or after merging I'll update it to target the main branch.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -67,71 +150,532 @@ fn sqlcipher_3_compat(conn: &Connection) -> Result<()> {
...
+                Ok(_) => {
+                    if let Some(mirror_login) = &login.mirror_login {
+                        // If successful, then migrate mirror
+                        migrate_mirror_login(&store, &mirror_login)?;
+                    }
+                }
+                Err(e) => {
+                    // If not successful on local, but we have a mirror
+                    // If there is an override -> delete skip both
+                    // IF no override -> attempt to import as normal
```
bendk, line 275: This makes me think that the fixup should be a separate pass and `fixup_mismatched_records` should be last, so the code to handle the `is_overridden=true` but no local record is all in 1 place.  If needed we could collect some data in `MigrationLogin` to distinguish the different cases.  Maybe `local` wouldn't be an `Option`, it would be an enum with several cases:
  - `NotFound`
  - `Invalid`
  - `Valid(Login)`

But that's only if we actually need different behavior for those cases.

bendk (approved):
I like the direction of this a lot.  Seems like a few things need to be ironed out, and we need to decide how to handle some edge cases.  The main ones for me are:
  - We have an error writing to the new DB
  - `is_overridden=true` but there's no local login or a local login with invalid data.

So there's still some things to do, but I'm for merging and iterating on this one.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -4,12 +4,99 @@
...
+        }
+    }
+
+    fn fix_mismatched_records(mut self) -> Result<Self> {
+        let mut guids_to_remove: Vec<String> = Vec::new();
+        for (guid, login) in &self.logins {
+            // Case 1: If the mirror record has is_overridden and no local records -> delete
+            if let Some(mirror_login) = &login.mirror_login {
+                if mirror_login.is_overridden && login.local_login.is_none() {
+                    guids_to_remove.push(guid.to_string());
```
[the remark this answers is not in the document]
skhamis: This method was mainly for removing/adjusting things from the plan before we attempt to migrate all the records. I might've had the assumption that we would leave it up to the server to reconcile and flip the is_overridden as needed for all other cases (is_overridden but no local copy means the server cannot effectively reconcile, where the other cases might?)

skhamis:
Thanks for the thorough review @mhammond & @bendk! I felt most of the suggested changes/tweaks were all valid and ended up removing a bunch of cruft to leave this feeling a little slicker. I ended up removing `fix_mismatched_records` as Ben was right, it doesn't really make sense to be it's own phase before we do all the fixups since we still need to do certain changes based on if we were able to import local vs mirror successfully. I think It'll be better served to rethink that approach and group that with the next set of changes (as far as I understand, we simply need a couple more checks + branches of code to complete this).

I'll probably go ahead and close this out and capture any remaining work in subsequent tasks to keep the PRs smaller.

PR #4377, opened 2021-08-10
Update UniFFI dependency to v0.13.1

### LINKED ISSUE #4379: Update UniFFI dependency to v0.14.x
This is required before we can start landing the swift-package/xcframework stuff from #4225

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2725)
┆Story Points: 2
┆Epic: Unified a-s swift packages for iOS
┆Sprint End Date: 2021-08-20

rfk:
Sam edit: Taking over this PR from Ryan -- the tests won't pass here until https://github.com/mozilla/uniffi-rs/pull/1025 lands, however I tested with a local build and this will pass once a a new cut of uniffi is released

Fixes #4379.


components/places/src/ffi.rs
```diff
@@ -280,32 +280,6 @@ implement_into_ffi_by_delegation!(
     msg_types::BookmarkNode
 );
 
-/// Implements [`IntoFfi`] for the provided types (more than one may be passed in) implementing
```
mhammond, line 283: 🚀

mhammond (dismissed):

rfk:
Hrmph, lots of build bustage in the generated bindings here that I'm going to work my way through...

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -187,20 +187,18 @@ internal inline fun <U> rustCallUniffi(syncOn: Any, callback: () -> U): U {
     synchronized(syncOn) {
         try {
             return callback()
-        } catch (e: ErrorWrapperException.Wrapped) {
+        } catch (e: ErrorWrapper.Wrapped) {
             // uniffi-generated functions currently return just a single error
             // type, which inside its message is the underlying error code
             // and message, which we can use to construct the actual error
             // from the hand-written FFI.
```
rfk, line 194: @mhammond I had to do a bit of fixing up here because "flat" UniFFI error enums no longer carry an implicit "message" property by default. We could actually use the new rich structured error enums to avoid this hacky "code|message" setup now, but leaving that for future work.

rfk:
I've converted this to draft because there's a lot of API churn that I don't fully have my head around yet. In particular, I'm worried about losing the `message` members of the various error classes, and want to see if there's a relatively straightforward way to bring them back.

build.gradle
```diff
@@ -19,7 +19,7 @@ buildscript {
         junit_version = '4.13'
         maven_publish_version = '3.6.2'
         mockito_core_version = '2.24.5'
-        roboelectric_core_version = '4.3.1'
+        robolectric_core_version = '4.5.1'
```
rfk, line 22: I needed this new version to be able to run the Kotlin tests locally, for inscrutable reasons.

CHANGES_UNRELEASED.md
```diff
@@ -19,7 +19,26 @@ Use the template below to make assigning a version number during the release cut
...
+
+### ⚠️ Breaking Changes ⚠️
+  - Many error classes have been renamed from `FooError` or `FooErrorException` to just `FooException`,
+    to be more in keeping with Java/Kotlin idioms.
+
+## Logins
+
+### ⚠️ Breaking Changes ⚠️
+  - Members of the `LoginsStoreError` enum no longer have a `message` property giving a detailed error message.
+    It is now just a plain enum with no associated data.
```
rfk, line 32: @skhamis points out that this isn't going to work, because we depend on these error messages over in [the a-c wrapper](https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt#L85). We'll need to bring them back somehow.

skhamis: Another option is to push forward with this and create the "fix" PR in a-c to take in the new version of a-s using enums instead of message but indeed would require more planning with the a-c team than we might want with just a uniffi ver bump

skhamis:
Closing pull request and will be making a new one identical to this for uniffi release v0.14.0

PR #4387, opened 2021-08-17
[nimbus] Adds async api for features on iOS

tarikeshaq:
### Background on "why now"
With the first run targeting work on mobile, we are opting to fetch and apply experiments sequentially (on the first run only, every other run will apply then fetch concurrently). This means that it will create more of a delay till the experiment information is ready **on the first run only, every other run is unaffected**.

### Consumer experience
Consumers that want to consume our API's currently will simply call our synchronous API, and if the experiments are not ready they'll get a `nil` value (and an error will be reported). 

What I'm proposing here is an asynchronous API, where the consumer has the option to pass in a `callback` that will execute **on the main thread** once the experiment data is ready. This is to support two main use cases:
- The case where a consumer is determined to get the experiment data (and a `nil` is not acceptable), without an async API, the consumer is forced to poll until the experiment data is ready
- The case where a consumer is okay "waiting" for a result (without blocking of course) to minimize the rate of failure. This is going to be especially important for experiments running on a user's first run (since there will be an expected delay). In practice, the delay isn't great in iOS, but the API is more about creating deterministic behavior if consumers want it

Another benefit to this is it creates a clear understanding of what a client receiving a `nil` means. If a client calls the async API and the callback receives a `nil` value, this can only imply that the experiment **Doesn't exist**... because if it existed, the async API will guarantee that the callback gets it. In the synchronous api, a `nil` could mean either the experiment data is not ready **or** that the experiment doesn't exist, and there is no way for the caller to determine which is it.

### How it's implemented
What I have here is **simple**, basically have everything async executing on the same **background thread**, this is not the best way to make it async (in reality there is only one dependency, the `applyExperiments`, every other call can be concurrent as long as the `apply` is done). I can explore making the asynchrony here better, but I think that's worth its own follow-up **if we decide we want to keep this**

Another note is that we might want to modify some of the Rust logic that this API calls (or define other rust APIs for this) but I haven't thought of that deeply since this seems to get the job done (I have a demo using firefox-ios that consumes this code)

(disclaimer that I'm very new to swift and I could have easily missed something here)

### TODO:
- [ ] Change log
- [ ] Tests (haven't looked at how the existing API is tested)

### DIFF
--- a/components/nimbus/ios/Nimbus/Nimbus.swift
+++ b/components/nimbus/ios/Nimbus/Nimbus.swift
@@ -189,6 +189,15 @@ internal extension Nimbus {
 }
 
 extension Nimbus: NimbusFeatureConfiguration {
+    public func getExperimentBranchAsync(experimentId: String, callback: @escaping (String?) -> Void) {
+        catchAll(dbQueue) {
+            let result = try self.nimbusClient.getExperimentBranch(id: experimentId)
+            DispatchQueue.main.async {
+                callback(result)
+            }
+        }
+    }
+
     public func getExperimentBranch(experimentId: String) -> String? {
         return catchAll {
             try nimbusClient.getExperimentBranch(id: experimentId)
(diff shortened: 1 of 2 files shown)


mhammond:
>  in a `callback` that will execute **on the main thread** once the experiment data is read

I haven't got an opinion on this PR in general, but it did cross my mind that the dispatch to the main thread is fairly trivial, so maybe this particular part of the policy could be in the app itself - eg, we just promise to callback on "this" thread, and the callback can do the `DispatchQueue.main.async` - or make some other choice about how to proceed?

Regardless, I'll leave the review of this to James or Travis as they have more context.

travis79 (commented):
After considering @mhammond's comment, it probably is assuming a bit much to dispatch back to the main thread, especially if that isn't the thread that is awaiting the async callback. In light of that, I'm sort of having second thoughts about adding this to the API, especially since we already provide a way to accomplish this through the onFetch and onApply notifications. I'm not entirely persuaded that this is a bad idea, but maybe we should consider it some more. I feel like @jhugman would have some valuable opinions regarding this, so maybe we can defer this until he returns from PTO and has a chance to weigh in?

tarikeshaq:
Eh, closing this for now, this won't be necessary and we don't have an immediate use for it

PR #4396, opened 2021-08-20
Build iOS Rust code as an XCFramework.

skhamis:
Continuation of #4225 -- but this time merging into main.

This is an attempt to make our iOS build process more compatible with
M1 Mac build environments, and easier to consume via Swift Packages.

In our existing iOS build setup, we compile together all of the Rust
code and all of the wrapping Swift code into a single module named
"MozillaAppServices". This is a bit of a violation of clean package
separation, because it means that we risk name conflicts in the Swift
code between different components.

In this new setup, we don't pre-compile any of the Swift code. Instead
we build only the Rust code, package it up along with its associated headers,
and tell Swift to treat this as a module named "MozillaRustComponents".
Each individual Swift wrapper can then `import MozillaRustComponents`
and get access to the low-level FFI functions that it needs.

An additional benefit of this setup is that we can distribute it as
a `.xcframework` bundle for easy consumption on different platforms,
and we can build it without using Xcode.

  - Note:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,6 +177,7 @@ commands:
             cargo metadata --locked > /dev/null
             python3 ./tools/dependency_summary.py --check ./DEPENDENCIES.md
             python3 ./tools/dependency_summary.py --all-ios-targets --package megazord_ios --check megazords/ios/DEPENDENCIES.md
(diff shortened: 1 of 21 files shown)


tarikeshaq (approved):
LGTM 🚀 

It builds and works! (we'll follow up with a PR on the https://github.com/mozilla/rust-components-swift side to track a version of A-S with those changes - which we had to do to get this to build successfully)

PR #4397, opened 2021-08-20
Remove duplicate DeviceType enum implementations.

mhammond:
This is really just to prove that we *can* remove the dupe enum! It doesn't actually use `[Extern]typedef` because it is still duplicated in the .udl - there's no cross-dependency with a `.udl` to put it - but we probably could remove that duplication once `sync15` or `sync_manager` get uniffi'd - for this reason I'm leaving this as a draft while I think about whether it should land anyway, or wait until later and we can remove all duplication (and I'd welcome opinions on this!)


skhamis:
I really like this PR and personally think it's worth landing already as the rust side seems much cleaner and it's probably going to be a little before we can get sync manager/sync15 fully uniffi-ed.

mhammond:
> It doesn't actually use `[Extern]typedef` because it is still duplicated in the .udl - there's no cross-dependency with a `.udl` to put it

Hey @bendk, you wanted to add such a UDL - we should push ahead with that! In the meantime, I'd like to get this rust part out of the way for tabs.

components/support/sync15-traits/src/client.rs
```diff
@@ -23,38 +24,88 @@ pub struct RemoteClient {
...
-            "vr" => Some(DeviceType::VR),
-            "tv" => Some(DeviceType::TV),
-            _ => None,
+    // Tried to `impl From<DeviceType> for Option<DeviceType>` but that conflicts with
+    // `impl<T> From<T> for Option<T>` in the core.
+    pub fn into(dt: DeviceType) -> Option<DeviceType> {
+        // Turn DeviceType::Unknown into None, everything else into Some()
+        match dt {
+            DeviceType::Unknown => None,
+            _ => Some(dt),
```
bendk, line 56: Having both `Some(DeviceType::Unknown)` and `None` as possibilities seems like a potential footgun, but I'm guessing it's needed to support some legacy code?  Could you add a note here explaining the situation?

components/fxa-client/src/internal/device.rs
```diff
@@ -7,14 +7,15 @@ use std::collections::{HashMap, HashSet};
...
-    DeviceLocation as Location, DeviceType as Type, GetDeviceResponse as Device, PushSubscription,
+    DeviceLocation as Location, GetDeviceResponse as Device, PushSubscription,
 };
 use super::{
     commands::{self, IncomingDeviceCommand},
     error::*,
     http_client::{DeviceUpdateRequest, DeviceUpdateRequestBuilder, PendingCommand},
     telemetry, util, CachedResponse, FirefoxAccount,
 };
+use crate::DeviceType;
```
bendk, line 18: Why `crate::DeviceType` over `sync15::DeviceType`?  It feels slightly non-standard to me, but I can't see any issues either.

bendk (changes requested):
Looks like a nice improvement.  I'd just like to understand more about `DeviceType::Unknown`  before approving.

components/fxa-client/src/internal/device.rs
mhammond: Because I was lazy :) I've fixed this.

components/support/sync15-traits/src/client.rs
mhammond: good call! It's actually *this* which is legacy code - while I just moved this from somewhere, this isn't used, so I killed it.

bendk (approved):
Looks great!

PR #4398, opened 2021-08-20
Rename UpdatableLogin to LoginEntry

mhammond:
Naming is hard :( I suggested something slightly different, but this renames `UpdatableLogin` to `LoginEntry` which Ben said matches Geckoview. I thought about renaming `Login` to `FetchedLogin` or `StoredLogin` but don't really like those names either - so maybe sticking with `Login` is OK? Or if not, we can rename that in a different commit.

Anyway, I'm not particularly invested in any of this, so if you think this is an improvement, feel free to merge it, otherwise feel free to close it.


bendk:
Seems like an improvement to me, let's merge.

PR #4400, opened 2021-08-23
Test to demonstrate how blank usernames are treated

mhammond:
This is a test-only change. It shows how we all allow an unlimited number of duplicate logins if the username is empty. It's not 100% clear this behaviour is *exactly* as intended, especially with identical passwords, but it is intentional - [the SQL jumps through some hoops](https://github.com/mozilla/application-services/blob/e1abe5ad61983f00a8abf1b477afa5470876e484/components/logins/src/db.rs#L690) to make this happen. It's also work noting that no tests exercise that SQL correctly - changing the `NULLIF`'s to simple comparisons doesn't cause any tests to fail (but it *does* cause this new test to fail).

Even if we decide the behaviour needs to change, we still should have a test.

(This will not apply cleanly against our per-fields-encryption branch, but will be easy to manually convert)

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -1027,6 +1027,38 @@ mod tests {
     }
 
     #[test]
+    fn test_username_dupe_semantics() {
+        let mut login = Login {
+            id: Guid::empty().to_string(),
+            hostname: "https://www.example.com".into(),
+            http_realm: Some("https://www.example.com".into()),
+            username: "test".into(),
+            password: "sekret".into(),
+            ..Login::default()
+        };
+
+        let db = LoginDb::open_in_memory(Some("testing")).unwrap();
+        db.add(login.clone())
+            .expect("should be able to add first login");
+
+        // We will reject new logins with the same username value...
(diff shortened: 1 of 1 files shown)


bendk:
Nice catch!

I wonder if that line of SQL is just buggy.  It seems like if the author wanted to ignore the blank values they could have just done an explicit check like `:username <> "" AND :username = username`.

I've read it several times and I always thought that the idea was that there could be NULL values or blank values and it wanted to treat them identically.  But of course, changing something to NULL then doing a `=` comparison is never going to work.

lougeniaC64:
> Nice catch!
> 
> I wonder if that line of SQL is just buggy. It seems like if the author wanted to ignore the blank values they could have just done an explicit check like `:username <> "" AND :username = username`.
> 
> I've read it several times and I always thought that the idea was that there could be NULL values or blank values and it wanted to treat them identically. But of course, changing something to NULL then doing a `=` comparison is never going to work.

As the author I believe we can treat this as buggy and should feel free to update this as needed. cc/ @mhammond

lougeniaC64 (approved):

Issue #4372, opened 2021-07-29
Push component refactoring following error handling changes


tarikeshaq:
@bendk had some great suggestions for refactors while going through #4347 This is an issue to track those, I'll be taking this and doing them one by one:
- [x] For the `unsubscribe` API, using `channel_id` = None as an indicator to unsubscribe from all channels isn't intuitive, it would be better if we either separate the API, or use a more descriptive enum (ref: #4402)
- [x] ~~Having `self.auth` and `self.uaid` as two separate fields makes it easy to make mistakes (they are both created on the same request). Ben had an amazing suggestion to put them as a part of a single enum that dictates the state, (ref: https://github.com/mozilla/application-services/pull/4347#discussion_r677830655). This will takes one step towards a better understanding of the possible states our push system can be in~~ will be done in #4430 
- [x]  We also have another code smell, the `unsubscribe` functions and the `update` function never return `false`. We should change them to return `Result<()>` instead. Addtionally, the `unsubscribeAll` for the PushManager on Kotlin will have no reason to return a boolean, so we will have a breaking change to change that to return nothing. Thanks @mhammond for catching that!!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2713)
┆Epic: Uniffi Push Component
┆Sprint End Date: 2021-09-03

tarikeshaq:
Closing this, the last bit will be done in #4430

PR #4402, opened 2021-08-23
[push] create a communications.rs unsubscribe_all call

tarikeshaq:
a part of #4372

Separates the logic in `communications.rs`'s `unsubscribe` call into:
- `unsubscribe` for a single channel, and this function now accepts a `&str` and not an `Option`
- `unsubscribe_all` for unsubscribing from all channels.

The separation works well for a couple of reasons:
- It aligns better with the API calls (and that the `unsubscribe_all` wipes the `uaid` and `auth` values)
- It aligns better with the `PushManager` API, which already defines two functions with the same names.
- It prevents us from using an `Option` as a flag to dictate which API call should run

Additionally, this PR makes the FFI contract on `unsubscribe` stricter. Previously, if the Kotlin caller passes a `null` value for the `channel_id`, our `PushManager.unsubscribe` would simply do nothing and return `false` which is a little misleading. That said, this change is safe because [our handwritten Kotlin wrapper for the FFI](https://github.com/mozilla/application-services/blob/main/components/push/android/src/main/java/mozilla/appservices/push/PushManager.kt#L77) accepts a `String` (and not a `String?`) therefore it was impossible in the first place to call this API with a null value

I'll have another PR addressing the rest of #4372 

  - Note:

### DIFF
--- a/components/push/src/communications.rs
+++ b/components/push/src/communications.rs
@@ -55,7 +55,6 @@ pub enum BroadcastValue {
     Value(String),
     Nested(HashMap<String, BroadcastValue>),
 }
-
 /// A new communication link to the Autopush server
 pub trait Connection {
     // get the connection UAID
@@ -69,7 +68,10 @@ pub trait Connection {
     ) -> error::Result<RegisterResponse>;
 
     /// Drop an endpoint
-    fn unsubscribe(&mut self, channel_id: Option<&str>) -> error::Result<bool>;
+    fn unsubscribe(&self, channel_id: &str) -> error::Result<bool>;
+
+    /// drop all endpoints
+    fn unsubscribe_all(&mut self) -> error::Result<bool>;
 
     /// Update the autopush server with the new native OS Messaging authorization token
     fn update(&mut self, new_token: &str) -> error::Result<bool>;
@@ -176,6 +178,17 @@ impl ConnectHttp {
         }
         Ok(())
     }
+
+    fn format_unsubscribe_url(&self) -> String {
+        format!(
+            "{}://{}/v1/{}/{}/registration/{}",
+            &self.options.http_protocol.as_ref().unwrap(),
+            &self.options.server_host,
+            &self.options.bridge_type.as_ref().unwrap(),
+            &self.options.sender_id,
+            &self.uaid.clone().unwrap(),
+        )
+    }
 }
 
 impl Connection for ConnectHttp {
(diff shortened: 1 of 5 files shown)


components/push/src/communications.rs
```diff
@@ -259,36 +261,58 @@ impl Connection for ConnectHttp {
...
+        let options = self.options.clone();
+        let url = format!(
+            "{}://{}/v1/{}/{}/registration/{}",
+            &options.http_protocol.unwrap(),
+            &options.server_host,
+            &options.bridge_type.unwrap(),
+            &options.sender_id,
+            &self.uaid.clone().unwrap(),
+        );
+        if &self.options.sender_id == "test" {
```
mhammond, line 307: Is this only used by rust tests? If so, maybe wrapping it in a `cfg` check would be good? (Ditto for the above which existed before this patch)

components/push/src/communications.rs
```diff
@@ -259,36 +261,58 @@ impl Connection for ConnectHttp {
     ///
     /// A `None` channel_id indicates that we are unsubscribing
     /// from all channels
-    fn unsubscribe(&mut self, channel_id: Option<&str>) -> error::Result<bool> {
+    fn unsubscribe(&mut self, channel_id: &str) -> error::Result<bool> {
```
mhammond, line 264: I don't want to derail this too much, but the `bool` return values smell here given `false` is never returned. Do you think this is worth cleaning up? (ditto with the new function below). Maybe the original intention was to inspect the response for a "not an error, but there was no subscription" type result?

components/push/src/subscriber.rs
```diff
@@ -100,18 +100,14 @@ impl PushManager {
     }
 
     // XXX: maybe -> Result<()> instead
```
mhammond, line 102: heh :)

mhammond (approved):
Thanks Tarik,
  This is just copying the existing semantics, so take my comments on the `bool` result for what it cost you :) Maybe a new issue?

components/push/src/communications.rs
tarikeshaq: No, I think this is a great cleanup!! Thank you - I'll add it to the list on #4372 , I messed around with it, it'll be a breaking change (at least for the `unsubscribeAll` but I think it'll be worth the follow up

components/push/src/communications.rs
tarikeshaq: Looks like it's used by the Kotlin tests 😞 This could be a thing to keep a close eye on when `uniffing` the component though!

components/push/src/communications.rs
```diff
@@ -256,39 +258,60 @@ impl Connection for ConnectHttp {
...
         }
+        let options = self.options.clone();
+        let url = format!(
+            "{}://{}/v1/{}/{}/registration/{}",
+            &options.http_protocol.unwrap(),
+            &options.server_host,
+            &options.bridge_type.unwrap(),
+            &options.sender_id,
+            &self.uaid.clone().unwrap(),
+        );
```
bendk, line 305: There's a good deal of duplicated code between this and `unsubscribe()`, maybe it could be moved to a private helper function.

bendk (approved):
The changes look great to me, better than my suggestion in the other issue.  I left one comment, but it's not that important.

You're plan on the bool return value seems good to me.

components/push/src/communications.rs
tarikeshaq: Moves the formating bits out, I'm keeping the error checking at the top, since that will be refactored with the rest of #4372 😄

tarikeshaq:
The CI failure here is unrelated, fixed by: https://github.com/mozilla/application-services/pull/4407

PR #4403, opened 2021-08-23
Add scrubbing logic to logins

### LINKED ISSUE #4066: Add logins support for scrubbing encrypted data
This will be very similar to what was done in #4029 for Logins with some of the work having already been implemented in #4064. This work should be based on the logins encryption branch and merged into it once this work has been approved.

Note: This ticket is blocked by #4065.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2633)
┆Epic: Logins update and sqlcipher migration
┆Fix Versions: Release 94
┆Sprint End Date: 2021-09-17

lougeniaC64:
Fixes #4066.

- I added tests here but what's clearly missing is e2e sync tests. I'm omitting those here so that I can get feedback on the bulk of this logic while finishing the iOS work. 
- At the moment iOS doesn't use sync manager and android does. I tried to keep that top of mind but if I've muddled things _please_ let me know.

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -752,6 +752,13 @@ impl LoginDb {
         Ok(exists)
     }
 
+    pub fn delete_encrypted_login_data(&self) -> Result<()> {
+        let tx = self.unchecked_transaction_imm()?;
+        tx.execute("DELETE FROM loginsL", NO_PARAMS)?;
+        tx.commit()?;
+        Ok(())
+    }
+
     fn mark_mirror_overridden(&self, guid: &str) -> Result<()> {
         self.execute_named_cached(
             "UPDATE loginsM SET is_overridden = 1 WHERE guid = :guid",
@@ -1768,5 +1775,49 @@ mod tests {
         let logins = db.get_by_base_domain("www.example.com").unwrap();
         assert_eq!(logins.len(), 1);
         assert_ne!(logins[0].record.id, bad_guid, "guid was fixed");
+    }
+
+    #[test]
+    fn test_delete_encrypted_login_data() {
+        let db = LoginDb::open_in_memory().unwrap();
+
+        db.add(
+            LoginEntry {
+                fields: LoginFields {
+                    form_action_origin: Some("https://www.example.com".into()),
+                    origin: "https://www.example.com".into(),
+                    ..Default::default()
+                },
+                sec_fields: SecureLoginFields {
+                    username: "test username".to_string(),
+                    password: "test password".to_string(),
+                },
+            },
+            &TEST_ENCRYPTOR,
+        )
+        .unwrap();
+
+        db.add(
+            LoginEntry {
+                fields: LoginFields {
+                    form_action_origin: Some("https://www.example2.com".into()),
+                    origin: "https://www.example2.com".into(),
+                    ..Default::default()
+                },
+                sec_fields: SecureLoginFields {
+                    username: "test username2".to_string(),
+                    password: "test password2".to_string(),
+                },
+            },
+            &TEST_ENCRYPTOR,
+        )
+        .unwrap();
+
+        db.delete_encrypted_login_data().unwrap();
+
+        assert_eq!(
+            db.query_one::<i32>("SELECT COUNT(*) FROM loginsL").unwrap(),
+            0
+        );
     }
 }
--- a/components/logins/src/sync/engine.rs
+++ b/components/logins/src/sync/engine.rs
@@ -364,6 +364,14 @@ impl LoginsSyncEngine {
         Ok(())
     }
 
+    pub fn delete_local_sync_data(&self) -> Result<()> {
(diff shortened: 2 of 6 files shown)


mhammond (changes requested):
Thanks Lougenia, this is coming along nicely. Feel free to push back on my comments.

I agree that there should be some more tests - I'm not sure how "end to end" they need to be, but ideally they would include executing the plan rather than just checking it. That might be a bit painful though.

I think we also need some hackery around outgoing items which don't have an item on the server - eg, a login just created, then the `scrub` function called before we next synced. I think that would actually cause the sync to fail (as we tried to decrypt the fields so we can put them on the server) - I'm open to what we do in that case (although just logging and skipping probably makes sense; uploading the record with empty username/password fields almost certainly does not).

Alternatively, I think we discussed the possibility of just deleting instead of scrubbing, otherwise we probably need to change almost all API functions to become aware that records might be scrubbed. Imagine non-sync users - they will never get any of their records "repaired" by Sync. Therefore, trying to add a new login for a domain must not fail just because there's an existing locally scrubbed record (which I think would happen now). There are other cases too, and I think the simplest thing here might be to just have `decrypt_fields` handling the empty string and returning a `SecureLoginFields` with empty fields - IOW, the scrubbing then just looks like we've set all usernames and passwords to empty strings rather than failing? This probably has implications to deduping (ie, if you had 2 logins for a domain that had different usernames, after scrubbing you would have 2 which are suddenly duplicates)

(The nuclear option would be to just nuke the world, which *sounds* bad but in practice might end up being the least surprising - sync users have a short window while they seem to have lost all their logins, non-sync users lose all their data, but they have already lost all their username and passwords, and pretending otherwise and keeping the old logins around is likely to confuse them more and hit other edge-cases in our impl that we didn't really think through. I'm stopping short of actually advocating for that at the moment, but could be convinced.)

bendk (changes requested):
This code looks correct based on the plan to scrub data like autofill, but @mhammond's comment kind of convinced me that the nuclear option might be better than scrubbing.  Here's how I see it:

- Scrubbed login data is not useful to the user.  We shouldn't return it from functions like `get_by_base_domain()`.  This is a key difference from autofill, where the unencrypted data could still be useful.
- Scrubbed data is going to be overwritten by the next sync
- The last 2 bullets kind of suggest that we should just delete the login data rather than scrubbing it.  I can only think of one use for it:
  - User has sync setup
  - We scrub the login data, but can't sync yet.
  - User logs in to a site and saves the login.  Because we have the login record around, we can update it rather than creating a new login with a new GUID
  - When we get around to syncing, we don't create a new/duplicate login because we were able to reuse the GUID.
- However, it's not going to be easy to implement that:
  - The `find_existing()`/`find_login_to_update()` code that I've been messing around with gets more complicated, because it needs to search scrubbed logins, even though `get_by_base_domain()` is not going to return them.
  - What do you do if there are 2 logins for a site?  If we update the wrong one, then sync the change, this could lead to data loss for the user.

It seems to me that the best solution to all of this is just deleting all the data and starting from scratch.  If we really wanted to save the GUIDs, then I think we need a separate system for that.  Like a new table that stores which GUIDs were connected to which sites before the deletion, then when we sync we could use that to try reconcile the GUIDs between the remote logins and the newly added local logins.  That seems pretty complicated to me and probably not worth the effort though.

components/logins/src/sync/engine.rs
```diff
@@ -364,6 +364,17 @@ impl LoginsSyncEngine {
         Ok(())
     }
 
+
+    pub fn delete_local_sync_data(&self) -> Result<()> {
```
lougeniaC64, line 368: I kept this separate function because I'm not cloning the mirror, resetting the sync status of `LoginL` records, which occurs in the `do_reset` function. However, I have no problems removing this function and calling `do_reset(EngineSyncAssociation::Disconnected)` if that makes more sense.

components/logins/src/db.rs
```diff
@@ -752,6 +752,13 @@ impl LoginDb {
         Ok(exists)
     }
 
+    pub fn delete_encrypted_login_data(&self) -> Result<()> {
```
mhammond, line 755: I was going to say I don't like the name of this function, and suggested we called it something like "wipe" - and then recalled we already have something named that. Then I discovered `wipe_local` - and ISTM that this is just reimplementing `wipe_local()`? Sorry I didn't realize this before, but unless I'm missing something, it seems `wipe_local()` already does exactly what we need?

lougeniaC64: Ugh, yeah, you're right. The only unique thing this logic does is reset the last sync time which still isn't meaningful enough to warrant these changes. I'm going to close this PR and archive the scrubbing logic ticket. Thanks!

lougeniaC64:
Because the `delete_local_sync_data` function created in this PR would just duplicate the exposed `wipe_local` function, I'm closing this PR.

PR #4405, opened 2021-08-24
FxAClient as a Swift package

skhamis:
This is a first (rather successful!) attempt to turn the FxA Client crate into a swift package. I was able to import FxAClient and crate a basic client in Focus iOS with my M1 without any tweaks! Pushing this PR up to test the CI resulting artifacts.

  - Note:

### DIFF
--- a/components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
+++ b/components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
@@ -3,6 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+#if canImport(MozillaRustComponents)
+    import MozillaRustComponents
+#endif
 
 /// This class inherits from the Rust `FirefoxAccount` and adds:
 /// - Automatic state persistence through `PersistCallback`.
--- a/megazords/ios-rust/DEPENDENCIES.md
+++ b/megazords/ios-rust/DEPENDENCIES.md
@@ -18,6 +18,7 @@ the details of which are reproduced below.
 * [MIT License: h2](#mit-license-h2)
 * [MIT License: http-body](#mit-license-http-body)
 * [MIT License: hyper](#mit-license-hyper)
+* [MIT License: libsqlite3-sys](#mit-license-libsqlite3-sys)
 * [MIT License: matches](#mit-license-matches)
 * [MIT License: mio](#mit-license-mio)
 * [MIT License: nom](#mit-license-nom)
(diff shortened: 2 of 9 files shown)


megazords/ios-rust/README.md
```diff
@@ -55,7 +55,7 @@ To add a new crate to the distribution:
     ffi_module_filename = "<crate_name>FFI"
     ```
 1. Add it as a dependency in `Cargo.toml`.
-1. Add a `pub use` declaration for it in `./src/lib.rs`.
+1. Add a `pub use` declaration for it in this `./src/lib.rs`.
```
mhammond, line 58: nit: this change doesn't read correctly to me

megazords/ios-rust/README.md
```diff
@@ -64,6 +64,8 @@ To add a new crate to the distribution:
 For testing out local changes in a consuming app, you can:
 
 * Take a local checkout of https://github.com/mozilla/rust-components-swift.
+> Note: `rust-components-swift` contains a submodule of `application-services`, if you're testing any code changes you'll need
```
mhammond, line 67: I don't quite understand this sentence:

* "if you're testing any code change" - isn't that implied here? ie, what else would I be doing if I'm reading the section that starts with "For testing out local changes"? (That's a genuine question - maybe testing something that's not a "code change"?) Or maybe you mean code changes in `rust-components-swift` vs code changes in `application-services`?

* Maybe clarify what you mean by "push/publish your branch" - presumably the person making a change is reading this because they want to test their changes *before* pushing anything, so maybe clarify how they would do that?

components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
```diff
@@ -3,6 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+#if canImport(MozillaRustComponents)
```
mhammond, line 6: for my info, what will happen if we can't import it? If it will just break, but somewhere else, I wonder if we should remove the `canImport` so that it breaks here?

(I've no idea how this hangs together, so apologies if this makes absolutely zero sense :)

megazords/ios-rust/DEPENDENCIES.md
```diff
@@ -18,6 +18,7 @@ the details of which are reproduced below.
 * [MIT License: h2](#mit-license-h2)
 * [MIT License: http-body](#mit-license-http-body)
 * [MIT License: hyper](#mit-license-hyper)
+* [MIT License: libsqlite3-sys](#mit-license-libsqlite3-sys)
```
mhammond, line 21: I'm mildly surprised to see this appear, because `fxa-client` doesn't use sqlite.

mhammond (approved):
LGTM - just a couple of nits and casual questions. Not too bothered by the "why sqlite appeared", so don't bother spending time trying to figure it out (but if you happen to know I'd be interested :)

megazords/ios-rust/DEPENDENCIES.md
skhamis: I'm also surprised but I simply ran the ./tools/regenerate_dependency_summaries.sh since the CI test was failing -- I might dive a little just to make sure something is not totally off.

components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift
skhamis: totally understand the confusion! It exists simply because we're still in a hybrid state of consuming via external swift packages (Individual components: Nimbus, Viaduct, etc) vs having all the components together (current day). Ryan explains it well in this issue: https://github.com/mozilla/application-services/issues/4308

Relevant excerpt:
> The Nimbus swift bindings currently use various functions exposed by Glean, and broadly assume that they will be built together with Glean in a single combined Swift module. When we move to Swift Packages that’s no longer going to work, since consumers will want to depend on Glean as its own separate Swift package.

megazords/ios-rust/README.md
skhamis: Yeah I'll definitely clean this up -- the issue is that it's actually not possible to fully test locally, swift packages need to be consumed via a repository. What tripped me up on this was that I had some changes on application-services side and built the `.xcframework` correctly but when running `./make-tag` in `rust-components-swift` it still tries to use the submodule rather than my local checkout. 

Thanks for flagging this, I think adjust the docs to make this read easier where the specific problem lays

skhamis:
This got bundled and merged in with https://github.com/mozilla/application-services/pull/4461, closing this as a result.

PR #4394, opened 2021-08-19
Run taskgraph generation with Python 3

ahal:
This fixes issue #4393.

We'd like to drop support for Python 2 in the taskgraph repo. To ensure the ability to address issues quickly, we'd like to first update consumers of it to Python 3.

PR #4407, opened 2021-08-24
Runs cargo update [ci full]

tarikeshaq:
Our CI for dependencies was failing on an `openssl` security vulnerability. We do not use openssl except for tests (it's pulled in by `reqwest`, which is pulled in by `viaduct_reqwest` that is only ever used in tests)

Ran cargo update to get the versions not affected, and get CI to pass.

ref on the vulnerabilities: 
-  https://rustsec.org/advisories/RUSTSEC-2021-0098
-  https://rustsec.org/advisories/RUSTSEC-2021-0097

Here's ci failing: https://app.circleci.com/pipelines/github/mozilla/application-services/28022/workflows/0561703c-b69e-4985-ab91-d2d0f7d42f95/jobs/112849

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1204,9 +1204,9 @@ checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
 
 [[package]]
 name = "h2"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "825343c4eef0b63f541f8903f395dc5beb362a979b5799a84062527ef1e37726"
+checksum = "d7f3675cfef6a30c8031cf9e6493ebdc3bb3272a3fea3923c4210d1830e6a472"
 dependencies = [
  "bytes",
  "fnv",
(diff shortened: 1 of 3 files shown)


tarikeshaq:
Looks like taskcluster builds were failing https://firefox-ci-tc.services.mozilla.com/tasks/HrllhdAzSvqRA3Wr1DyIRw/runs/0/logs/public/logs/live.log

I _think_ it's because they were running in python2 instead of python3, the python interpreter couldn't recognize the `f"string {value}"`, so that was my quick guess. In response, I changed the shebang to be `#!/usr/bin/env python3`

The reason the builds didn't fail with https://github.com/mozilla/application-services/pull/4394 (that introduced the python3 support) is because that PR didn't run `[ci full]` - and thus didn't run the taskcluster builds

skhamis (approved):
Makes sense and LGTM! Nice catch 🚀 

Edit: FTR, I did try look at any other places we still might be using python2 and could only find https://github.com/mozilla/application-services/blob/e1abe5ad61983f00a8abf1b477afa5470876e484/.circleci/config.yml#L106 which seems find as it's just installing stuffs

Issue #4373, opened 2021-07-29
Create state diagrams for push component


tarikeshaq:
After chatting with @rfk  (also ref: https://github.com/mozilla/application-services/pull/4347#discussion_r674406962) , it's important for us to create a better understanding of how the different parts of the push system interact. We can get closer to that with state diagrams and how they interact:

* [ ] A state diagram for the states the consumer is in (e.g Fenix), those could be things like "connected", "subscribed to channels", "reconnecting", etc
* [ ] A state diagram for the states the server is in, those could be things like: "UAID Created", "No UAID", "Subscriptions Created", etc
* [ ] (possibly if it's different enough from the consumer) a state diagram for our component!

The examples I put above are in no way complete or completely accurate (which signals that we(well, me!) need a better complete understanding here!)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2714)
┆Story Points: 8
┆Epic: Important backlog
┆Sprint End Date: 2021-09-17

data-sync-user:
➤ Janet Dragojevic commented:

Mark Hammond I know you are doing a but of send-tab / push expiration investigation for Fenix.  Question for both you & Tarik Eshaq should we consider this as part of the uniffi push epic - so that the uniffi work comes with better docs/understanding.  (don’t want to overload the epic, so push back if I am wrong here)

data-sync-user:
➤ Tarik Eshaq commented:

Closing as this has gone stale and we’ve done a decent bit of documentation of push, including client state machines in https://firefox-source-docs.mozilla.org/dom/push/index.html ( https://firefox-source-docs.mozilla.org/dom/push/index.html|smart-link )

PR #4411, opened 2021-08-26
[push] refactors uaid and auth into same enum variant

tarikeshaq:
a part of #4372 

Refactors the `uaid` and `auth` `Option`s into an enum that guides the state of the component. This is a bit more involved as a refactor, and before I take this out of draft, I would very much like to build fenix using this and see if it works (well "works" as well as what we already have :'( )... with all the `uniffi` 0.14.0 breaking changes this might have to wait a little, but wanted to open it up for visibility

(Note: this is a breaking change since it adds errors,  something I completely missed with #4347 :( ) - I'll add the change log

  - Note:


tarikeshaq:
I'll close this PR - I don't think this is the correct refactor to do, I created #4430 which should follow #4373, which I'll add to my plate for next sprint. After we have a clean image of all the states the client can be in, we can do this properly and let the PushManager handle it

PR #4414, opened 2021-08-30
Cut release v82.3.0

tarikeshaq:
Runs the cut release script for 82.3.0

~~For some reason the changelogs were not copied over correctly, I'll do that manually~~

Had to copy over the changelogs manually, but otherwise this should be ready to land - we can cut the release right after

(note that we'll get failures on check-dependencies CI, but that's okay, to clarify, they're security vulnerability errors in **test only** code that's fixed anyways in main)


skhamis (approved):
Thanks for getting this fixed up. LGTM!

PR #4415, opened 2021-08-30
Merge back v82.3.0 into main

tarikeshaq:
Merge back v82.3.0 into main to wrap up the 82.3.0 release

### DIFF
--- a/.buildconfig-android.yml
+++ b/.buildconfig-android.yml
@@ -1,4 +1,4 @@
-libraryVersion: 82.2.0
+libraryVersion: 82.3.0
 groupId: org.mozilla.appservices
 projects:
   autofill:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# v82.3.0 (_2021-08-30_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v82.2.0...v82.3.0)
+
+### What's New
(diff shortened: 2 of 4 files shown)


tarikeshaq:
I won't update this PR until the artifacts for the release are built (then I'll deal with the merge conflict)

tarikeshaq:
This should be good to merge, the `queued` task is actually done on TaskCluster.

AND lucky for us there are no merge conflicts on this one!

I imagine if I `Squash and merge` it's okay? lightly concerned Since the release tag is based on a commit on the `release-v82` branch, but I imagine git is smart enough to preserve the tag? If uncertain we can just do a merge commit and keep the branch open

mhammond (approved):
I'm mildly surprised check-dependencies is failing as it seems resolved on main - but I can't see how this could possibly cause it.

I think we should just merge - IIRC, we previously had the repo configured to not allow merging, but that caused problems in this release process. We should probably use a consistent process each time (so I'd also welcome an update to the docs to clarify this :)

tarikeshaq:
> I'm mildly surprised check-dependencies is failing as it seems resolved on main - but I can't see how this could possibly cause it.

I _think_ it's since the branch doesn't have the fix (as it still is based off of the 82 release)

mhammond:
> I  _think_ it's since the branch doesn't have the fix (as it still is based off of the 82 release)

doh - yeah, that's correct.

PR #4417, opened 2021-08-31
Fixes dependency summaries

tarikeshaq:
The check dependencies CI is failing again (for an unrelated reason to before, a repo changed their default branch name to `main`


tarikeshaq:
Yeah... I think that's be a good idea, at the very minimum run `./tools/regenerate_dependency_summaries.sh` and push that up - we can always review it to make sure it's OK... 

We can either do that as a part of that task, or even a separate github action that runs periodically (I setup something similar last summer, but I don't remember what happened to it)

PR #4426, opened 2021-09-02
Removes unused dependencies

tarikeshaq:
while taking a look at some uniffied components to see what needs to be done with Push, realized fxa-client, logins and tabs still imported `ffi-support`, this removes that 🎉


tarikeshaq:
_wonders if our `check-dependencies` ci can make sure there are no un-used dependencies..._

components/support/error/Cargo.toml
```diff
@@ -6,7 +6,6 @@ edition = "2018"
 license = "MPL-2.0"
 
 [dependencies]
-thiserror = "1.0"
```
tarikeshaq, line 9: This is okay to remove because the macro assumes the **consumer** crate has `thiserror`, it never used its own `thiserror` dependency

skhamis (approved):
This makes sense to me, nice catch! Surprised they were in there this long -- it'll also be good to add this to the "converting a component" document https://mozilla.github.io/application-services/book/howtos/converting-a-component-to-uniffi.html but I won't block this PR for that.

mhammond (approved):
🚀

PR #4428, opened 2021-09-03
Cleanups before uniffication

### LINKED ISSUE #4423: Implement Push API changes to prepare for Uniffi
There are a few changes we can merge independently before we uniffi the push component, here's a list under "Task 1": https://github.com/mozilla/application-services/issues/4395#issuecomment-911913499

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2762)
┆Story Points: 3
┆Epic: Uniffi Push Component
┆Sprint End Date: 2021-09-17

tarikeshaq:
fixes #4423 

Point of the PR is to do as many incremental cleanups as possible without chucking everything into a big "uniffi" pr, or introducing too much temporary code.

The main goal I tried to do here is move as much logic as possible away from the ffi crate / and into the library crate..

There is still some logic I couldn't move, mainly whenever the return type I wanted to return didn't implement `ffi_support::IntoFfi` and adding a temporary Impl is not worth it (For example the `decrypt` still returns a string)

### DIFF
--- a/components/push/src/subscriber.rs
+++ b/components/push/src/subscriber.rs
@@ -13,9 +13,55 @@ use crate::config::PushConfiguration;
 use crate::crypto::{Crypto, Cryptography, KeyV1 as Key};
 use crate::error::{self, ErrorKind, Result};
 use crate::storage::{PushRecord, Storage, Store};
+// TODO(teshaq): those will be replaced with rust structs in the next
+// uniffication step
+use crate::msg_types::{
+    DispatchInfo, KeyInfo, PushSubscriptionChanged, SubscriptionInfo, SubscriptionResponse,
+};
 
 const UPDATE_RATE_LIMITER_INTERVAL: u64 = 24 * 60 * 60; // 500 calls per 24 hours.
 const UPDATE_RATE_LIMITER_MAX_CALLS: u16 = 500;
+
+impl From<(RegisterResponse, Key)> for SubscriptionResponse {
+    fn from(val: (RegisterResponse, Key)) -> Self {
+        SubscriptionResponse {
+            channel_id: val.0.channel_id,
+            subscription_info: SubscriptionInfo {
+                endpoint: val.0.endpoint,
+                keys: val.1.into(),
+            },
+        }
+    }
+}
+
+impl From<Key> for KeyInfo {
+    fn from(key: Key) -> Self {
+        KeyInfo {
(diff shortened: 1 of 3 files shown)


components/push/src/subscriber.rs
```diff
@@ -13,10 +13,56 @@ use crate::config::PushConfiguration;
...
+// TODO(teshaq): those will be replaced with rust structs in the next
+// uniffication step
+use crate::msg_types::{
+    DispatchInfo, KeyInfo, PushSubscriptionChanged, SubscriptionInfo, SubscriptionResponse,
+};
 
 const UPDATE_RATE_LIMITER_INTERVAL: u64 = 24 * 60 * 60; // 500 calls per 24 hours.
 const UPDATE_RATE_LIMITER_MAX_CALLS: u16 = 500;
 
+impl From<(RegisterResponse, Key)> for SubscriptionResponse {
```
tarikeshaq, line 25: I'm not too happy with this... 

while the `From` trait is really cool, we have too many types here, like do we really need a `KeyInfo` and a `Key` type when they basically return the same thing. 

Will need to think about this a bit and I might create a follow-up to clean this up after we are done uniffing - In the mean time, in the next PR I'll be doing a direct translation, simply replacing the `protobuf` types with rust types

components/push/src/subscriber.rs
```diff
@@ -96,10 +150,15 @@ impl PushManager {
                 self.store.set_meta("auth", &secret)?;
             }
         }
-        Ok((info, subscription_key))
+        Ok((info, subscription_key).into())
     }
 
     pub fn unsubscribe(&mut self, channel_id: &str) -> Result<bool> {
+        // TODO(teshaq): This should throw an error instead of return false
```
tarikeshaq, line 157: For context this was moved from the android code to here

components/push/src/subscriber.rs
```diff
@@ -162,13 +221,16 @@ impl PushManager {
...
         &self,
-        uaid: &str,
         chid: &str,
         body: &str,
         encoding: &str,
         salt: Option<&str>,
         dh: Option<&str>,
     ) -> Result<String> {
+        if self.conn.uaid.is_none() {
+            return Err(ErrorKind::GeneralError("No subscriptions created yet.".into()).into());
```
mhammond, line 231: This was a panic before? This is much better!

mhammond (approved):
:shipit:

PR #4431, opened 2021-09-03
Uniffi push component

### LINKED ISSUE #4424: Implement uniffi'ing the push component
This includes following the instructions here https://mozilla.github.io/application-services/book/howtos/converting-a-component-to-uniffi.html 

For this to be done, we should completely eliminate the protocol buffers and the handwritten FFI code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2763)
┆Story Points: 5
┆Epic: Uniffi Push Component
┆Fix Versions: Release 94
┆Sprint End Date: 2021-09-17

tarikeshaq:
Fixes #4424 

The commits in order:
- [x] adds a valid UDL
- [x] Rust compiles with scaffolding included
- [x] Removes ffi crate
- [x] Removes protobuf types and replaces them with rust types
- [x] Adds kotlin bindings and cleans up the API to allow android to consume this 

This is ready for review 🚀 Before this merges though, a couple of TODOS:
- [x] Create changelog entry describing the breaking changes (added: https://github.com/mozilla/application-services/pull/4431/commits/d6bf2e4186a789a9255ea70678bbafccd921b304)
- [x] Create Android components PR to fix breaking changes, get review there and make appropriate changes here (PR: https://github.com/mozilla-mobile/android-components/pull/10953)


components/push/src/push.udl
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+namespace push {};
+
+
+interface PushManager {
+    [Throws=PushError]
+    constructor(string sender_id, optional string server_host = "updates.push.services.mozilla.com", optional string http_protocol = "https", string bridge_type, string registration_id, optional string database_path = "push.sqlite");
```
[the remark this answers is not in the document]
tarikeshaq: also a reminder to myself about the breaking change here, the `bridgeType` used to be a `BridgeType` object, and our bindings used to `toString` it, now A-C will have to do that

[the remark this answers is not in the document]
tarikeshaq: ended up getting rid of the breaking change 😄

components/push/android/src/test/java/mozilla/appservices/push/PushTest.kt
```diff
@@ -141,12 +141,12 @@ class PushTest {
```
tarikeshaq, line 141: The changes in this file are the extent of the breaking changes...

- `A-C` will have to turn the `List<Byte>` into a `ByteArray` (what `decrypt` used to return) if they'd like using `toByteArray`. It was either this, or have a small wrapper to do the conversion, I opted for the "less android code in a-s" approach
- `channelID` -> `channelId`: this is trivial but interesting, I tried setting it to `channel_ID`, but the rust compiler gave me warnings to turn it into `channel_id` to make it snake case. So either we make the breaking change, or we can silence the rust compiler... I almost opted for the latter, but wanted to get a second opinion

Cargo.toml
```diff
@@ -9,7 +9,6 @@ members = [
     "components/places",
     "components/places/ffi",
     "components/push",
-    "components/push/ffi",
```
mhammond, line 12: 🚀

components/push/src/lib.rs
```diff
@@ -380,17 +380,18 @@ impl PushManager {
...
             body,
             encoding,
             Some(salt),
             Some(dh),
         )?;
-        Ok(ret.as_bytes().to_vec())
+
+        // NOTE: this returns a `Vec<i8>` since the kotlin consumer is expecting
+        // signed bytes.
+        Ok(decrypted.into_iter().map(|ub| ub as i8).collect())
```
mhammond, line 394: Is this related to unsigned types being experimental in Kotlin? If so, I think that's no longer the case.  (IOW, if it's possible to now use u8 here I think we should do so, but I've no idea how practical that is, so feel free to ignore!)

mhammond (approved):
Great work Tarik - very thorough and well thought out! Please do complete those other todos before merging though - I'm worried about overloading the android-components people, so think we should have a PR up before merging.

But great work, thanks!

components/push/src/lib.rs
tarikeshaq: Uhh yes, I used this because Android components expects a `ByteArray`, which is just one `.toByteArray()` away from a `List<Byte>`, but `List<UByte>` can't convert directly. So A-C will have to convert the `List<UByte>` to `ByteArray` through a `UByteArray`... which isn't _that_ bad, but it would be the first time they interact with unsigned types over there 😅 I guess it's just a matter of do we want to do the conversion for them, or have them do the conversion themselves

tarikeshaq:
Thanks Mark!! I created a pr on A-C to get feedback https://github.com/mozilla-mobile/android-components/pull/10953

Correct me if I'm wrong, but from yesterday's discussion, we should cut a release before this merges, resolve the _other_ breaking changes, then cut **another** release for this right?

tarikeshaq:
This should be in a mergeable state! I'm planning to merge it sometime early next week and cut another major release. Once that's done I'll update https://github.com/mozilla-mobile/android-components/pull/10953 to have the latest release. (cc: @mhammond and @jonalmeida, no action items, just letting you know my plan here in case you have any concerns or thoughts 😄)

PR #4434, opened 2021-09-03
Moves push implementation to internal module

tarikeshaq:
A part of #4424

I was able to make this a nonbreaking change 😄 The `ffi` crate basically interacts directly with the `internal` `PushManager` so from an outside view, this is just a refactor

## What's in this change
- Moves the logic into a module called `internal` similar to how the `fxa_client` component is structured.
- Defines a new `lib.rs` that defines all the APIs that `uniffi` will consume (look at #4431 for what that looks like) This makes the next steps incredibly simple, since the `PushManager` in `lib.rs` is now `Sync + Send` and the API is defined is exactly what the UDL expects (minus some error work)
- Documents each API call with Rust docs similar to what was done on the fxa client
- Makes other crates (the `ffi` crate and an example crate) reference the internal implementation directly

I slightly prefer that this merges to `main` directly as opposed to onto #4431 - well since it can, to minimize how big that PR gets - but wouldn't mind merging it onto that branch if others prefer that!


mhammond (approved):
TBH and FWIW, I'm not a huge fan of this `internal` pattern Ryan trailblazed with fxa_client, and I mentioned that to him at the time, but I concede this is completely subjective and I can see why others might like it - so I'm fine with this, especially now that you are the defacto owner of this component 🎉 :)

tarikeshaq:
Thanks for the approval @mhammond! I'll merge this for now, but I'll go find the PR Ryan introduced this pattern and read through the comments for tradeoffs... Do you think it's worth it to have a standardized approach? Like should we create future issues to figure out what our suggested way is, or do we keep it up to the developers working on the component - mostly thinking in terms of what we can add/remove from https://mozilla.github.io/application-services/book/howtos/converting-a-component-to-uniffi.html

mhammond:
> Do you think it's worth it to have a standardized approach?

I think it's worth trying to converge on common patterns, but I don't like being too prescriptive. I think a tl;dr of the fxa-client comments was Ryan saying he was trying out the idea and it seems OK, but maybe not as he initially pictured - which is completely fine.

I do agree with updating documents to have current best-practices though.

While I'm here I guess I should try to articulate what I find unsatisfying: our rust components have almost no rust consumers, so in effect, anything that's not exposed via uniffi is already "internal" - so just moving everything into an "internal" directory just adds a new directory but doesn't have any practical impact on anything else. fxa-client/src has one .rs file - I don't think much would be lost if internal/*.rs was in src. To be clear though, I don't think there's any harm - so it's not at all "I hate it" and more "I don't see the value". In some other contexts - particularly if we did have external rust consumers, it might make more sense - in which case I'd kinda expect a more even spread between files in the "internal" and "external" parts.

PR #4436, opened 2021-09-06
database_opener supports read-only connections.

mhammond:
I started to work on moving places to the database_opener. This spawned
the following changes:

* For places in particular, for various reasons including the fact that
  places needs readonly connections (which by definition can not migrate),
  the name `MigrationLogic` doesn't really work that well. I ended up
  renaming that trait to `ConnectionInitializer` (which is why this patch
  is as big as it is)

* To support read-only connections, so I slightly changed what methods are
  called and when; I won't repeat what I wrote in the comments, but I ended
  up being able to remove `setup_pragmas()` as that can now be done in
  `prepare`, and functions only called for writable connections now
  explicitly take `&Transaction` to help reinforce they are only called
  for writable.

* The tests in `open_database.rs` confused me a little in their
  intent, so the changes I made there might not make complete
  sense, but they do all pass when uncommented.

### DIFF
--- a/components/support/sql/src/open_database.rs
+++ b/components/support/sql/src/open_database.rs
@@ -5,23 +5,30 @@
 /// Use this module to open a new SQLite database connection.
 ///
 /// Usage:
-///    - Define a struct that implements MigrationLogic.  This handles:
+///    - Define a struct that implements ConnectionInitializer.  This handles:
 ///      - Initializing the schema for a new database
 ///      - Upgrading the schema for an existing database
 ///      - Extra preparation/finishing steps, for example setting up SQLite functions
 ///
 ///    - Call open_database() in your database constructor:
-///      - The first method called is setup_pragmas().  This is executed outside of a transaction
-///        to support `PRAGMA journal_mode=wal`
-///      - If the database file is not present, open_database() will create a new DB and call prepare(),
-///        init(), then finish()
-///      - If the database file exists, open_database() will open it and call prepare(),
-///        upgrade_from() for each upgrade that needs to be applied, then finish().
+///      - The first method called is `prepare()`.  This is executed outside of a transaction
+///        and is suitable for executing pragmas (eg, `PRAGMA journal_mode=wal`), defining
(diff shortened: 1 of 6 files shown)


components/support/sql/src/open_database.rs
```diff
@@ -35,89 +42,106 @@ pub enum Error {
...
     // The version that the last upgrade function upgrades to.
     const END_VERSION: u32;
 
-    // Runs first, outside of a transaction
-    fn setup_pragmas(&self, _conn: &Connection) -> Result<()> {
-        Ok(())
-    }
+    // Functions called only for writable connections all take a Transaction
+    // Initialize a newly created database to END_VERSION
+    fn init(&self, tx: &Transaction<'_>) -> Result<()>;
```
bendk, line 54: +1 to explicitly stating which methods are inside a transaction vs a connection.

components/support/sql/src/open_database.rs
```diff
@@ -35,89 +42,106 @@ pub enum Error {
...
-    // Initialize a newly created database to END_VERSION
-    fn init(&self, conn: &Connection) -> Result<()>;
-
-    // Upgrade schema from version -> version + 1
-    fn upgrade_from(&self, conn: &Connection, version: u32) -> Result<()>;
-
-    // Runs after the init/upgrade functions
+    // Runs for all types of connections. If a writable connection is being
+    // initialized, this will be called after all initialization functions,
+    // but inside their transaction.
```
bendk, line 67: Combining `prepare` and `setup_pragmas` makes a lot of sense.  `setup_pragmas` always felt overly specific me.  Also +1 to explicitly stating which methods are inside a transaction vs a connection.

bendk (approved):
Looks good to me, BTW this reminded me that I wrote some [initial code to move places to `open_database`](https://github.com/mozilla/application-services/pull/4179#issuecomment-861799072).  The first comment was about the read-only connection issue.  This change seems like the best way to handle it.

PR #4437, opened 2021-09-06
Use open_database for places.

mhammond:
This still needs love, but I'm putting it up as more context for #4436 (and note that it obvs has the changes from that PR too - just look at the changes inside places)


mhammond:
I think this is ready to roll. I think it makes sense for both @lougeniaC64 and @bendk to take a look if you don't mind!

bendk (approved):
LGTM. 

How much work is it to add a test for migrating a V2 database to the current version?  If it's easy, then maybe that could be in this one too.  If it's hard then we can work on it later.

lougeniaC64 (approved):
After reviewing #4436, LGTM--although I +1 Ben's comment about migration tests.

mhammond:
> How much work is it to add a test for migrating a V2 database to the current version? If it's easy, then maybe that could be in this one too. If it's hard then we can work on it later.

I think it's very difficult to do in a way that provides value - many of the migrations simply re-execute `CREATE_SHARED_SCHEMA_SQL` - eg, when a new table or index is added we just add a new `CREATE TABLE IF NOT EXISTS` statement, and re-execute the entire schema. So unless I'm mistaken, your patch which tried to start at version 5 would have actually leapt directly to 15 on the very first upgrade. That would make many of the later upgrades noops (eg, `CREATE INDEX IF NOT EXISTS` for version 9 would be a noop, but would not be in an upgrade from version 8)

That *is* what would actually happen if we came across a version 5 database in the wild - so there would be *some* value to that, but it seems very unlikely we would actually hit that in practice - the most value in practice is probably to check each version bump individually (ie, 5->6, then 6->7 etc), but that would mean we'd need every version checked in, and then we could test all the (to_version, from_version) permutations that exist - but that does sound very tricky, and defining success is also tricky - "doesn't fail" has some value, but not as much value as "actually ended up with the correct schema" - eg, failing to drop a table in a migration so it's re-created would not be picked up by the "didn't fail" metric, but the table would not have the correct schema at the end

And the final nail in the coffin here is that this patch isn't actually doing nor changing any migrations - the mechanism which by upgrades happen has changed, but the logic to move from version `n` to version `m` should be identical for all `n` and `m`. So I do think we should do a better job at testing upgrades, but I don't think this patch is necessarily the best place to do that.

PR #4438, opened 2021-09-06
Suppress DatabaseNotReady errors when reporting to senty

jhugman:
Fixes SDK-368.

  - Note:

### DIFF
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
@@ -393,11 +393,13 @@ open class Nimbus(
         try {
             thunk()
         } catch (e: Throwable) {
-            try {
-                errorReporter("Error in Nimbus Rust", e)
-            } catch (e1: Throwable) {
-                logger("Exception calling rust: $e")
-                logger("Exception reporting the exception: $e1")
+            if (e !is NimbusException.DatabaseNotReady) {
+                try {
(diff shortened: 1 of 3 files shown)


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -393,11 +393,13 @@ open class Nimbus(
...
-            try {
-                errorReporter("Error in Nimbus Rust", e)
-            } catch (e1: Throwable) {
-                logger("Exception calling rust: $e")
-                logger("Exception reporting the exception: $e1")
+            if (e !is NimbusException.DatabaseNotReady) {
+                try {
+                    errorReporter("Error in Nimbus Rust", e)
+                } catch (e1: Throwable) {
+                    logger("Exception calling rust: $e")
```
tarikeshaq, line 400: I'm not sure (and this is not related to your change specifically, since you just moved it into the if statement) but should this be in the "try" block? since it's logging that the issue was in the rust?

tarikeshaq (approved):
Makes sense to me! Is this worth a changelog entry?

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
jhugman: That's a great question. 

The inner try/catch is guarding ourselves from the `errorReporter` crashing. If it does we log the error with the logger, which we're betting on not failing. We also log the error in the `errorReporter`.

PR #4444, opened 2021-09-08
WIP to detect corruption and delete databases as they are opened

mhammond:
This seems to work OK - I'm not completely happy with it, but also struggling to work out how to improve it - so I thought I'd throw it up for feedback. In particular, the "try to recover" cruft needs to be in a few more places, and I don't think we are closing the connection correctly once we've started the transaction (and it's difficult to test that without getting our hands on an actual sqlite database that only fails after we create the transaction)

Flagging Ben as he probably has the most context, but welcome reviews from anyone.


components/support/sql/src/open_database.rs
```diff
@@ -87,21 +89,70 @@ pub fn open_database_with_flags<CI: ConnectionInitializer, P: AsRef<Path>>(
...
+    if open_info
+        .open_flags
+        .contains(OpenFlags::SQLITE_OPEN_READ_WRITE)
+    {
+        let tx = match conn.transaction() {
+            Ok(t) => t,
+            Err(e) => {
+                return handle_db_failure(
+                    open_info,
+                    // Hrm - can't pass 'conn' here, nor tx?!?
```
mhammond, line 143: It sucks that here we've lost the ability to pass `conn`, but don't yet have the ability to pass `tx`

components/support/sql/src/open_database.rs
```diff
@@ -87,21 +89,70 @@ pub fn open_database_with_flags<CI: ConnectionInitializer, P: AsRef<Path>>(
...
+fn do_open_database_with_flags<CI: ConnectionInitializer>(
+    open_info: OpenInfo<'_, CI>,
 ) -> Result<Connection> {
     // Try running the migration logic with an existing file
     log::debug!("{}: opening database", CI::NAME);
-    let mut conn = Connection::open_with_flags(path, open_flags)?;
-    let run_init = should_init(&conn)?;
+    // It turns out Connection::open_with_flags appears to always return without checking
+    // the validity - but that might not be true in all cases, so be safe.
+    let mut conn = match Connection::open_with_flags(open_info.path, open_info.open_flags) {
```
mhammond, line 117: everywhere that can fail needs to do this kind of dance, which is a bit ugly, but I can't see a better way.

components/support/sql/src/open_database.rs
```diff
@@ -142,6 +195,72 @@ pub fn open_memory_database_with_flags<CI: ConnectionInitializer>(
...
+    if delete {
+        log::info!(
+            "{}: the database is fatally damaged; deleting and starting fresh",
+            CI::NAME
+        );
+        match conn {
+            DroppableConnection::None => {}
+            DroppableConnection::Connection(c) => drop(c),
+            DroppableConnection::Transaction(tx) => {
+                let _ = tx.rollback(); // ignore errors here.
```
mhammond, line 243: This annoys me - I can't work out how to clean up correctly after a tx.

mhammond (commented):
There might be a different shape that can address some of these ugly bits, but it's escaping me at the moment, so though some fresh eyes might help.

bendk:
What about moving the handling code outside of `do_open_database_with_flags()`?  That way if there's an error the connection and/or transaction gets dropped automatically.  Plus you only need to do the dance once.

Updated to your commit: https://github.com/bendk/application-services/commit/bb0ff8f723a2e674d1917fae44fb85ff4a97cda2
Here's the branch: [branch](https://github.com/bendk/application-services/commits/delete-corrupt-db)

bendk:
What should we do if `remove_file()` fails?  In that branch, we just return the `std::io::Error`.  Maybe it would be better to return the original error or an error that contains both?

components/support/sql/src/open_database.rs
```diff
@@ -142,6 +195,72 @@ pub fn open_memory_database_with_flags<CI: ConnectionInitializer>(
...
+                                       // ideally we'd also call close the connection behind the transaction but
+                                       // that seems impossible - Defef into a `&Connection` fails here.
+            }
+        }
+        // XXX - should we move the path to, say ".corrupt"? Not clear there's
```
skhamis, line 248: yeah I can't think of a reason of keeping it unless we plan on maybe doing some diagnostics in a future time (then a log dump might also suffice)

components/support/sql/src/open_database.rs
skhamis: FWIW I don't think it looks too bad, the only thing I can think of is maybe a macro `open_with_flags!` that abstracts but probably won't make it that much cleaner

mhammond:
> What about moving the handling code outside of `do_open_database_with_flags()`?

Yeah, that's much better, thanks!

mhammond:
> What should we do if `remove_file()` fails? In that branch, we just return the `std::io::Error`. Maybe it would be better to return the original error or an error that contains both?

Yeah, I was torn here - if I had to choose between the 2, I'd probably prefer the IOError - returning the original error would tell us what we already knew (ie, that it must be one of exactly 2 error codes) and not tell us anything about what we don't know (what IOError). I didn't think about both though :) So that's what I've done - although it's just a stringified version of the original error, otherwise `Error` could become cyclic.

bendk (approved):
Looks great.  It'll be very interesting to see what effect this has in the wild.

PR #4445, opened 2021-09-08
The places-utils example needs to initialize viaduct

mhammond:
Some refactoring of our networking code left our places-utils example behind.


skhamis (approved):
LGTM! :shipit:

PR #4448, opened 2021-09-08
Closes #4433: Allow observing previewImageUrl for visits

### LINKED ISSUE #4433: Expand history API to allow observing & reading `preview_image_url`
We'll use this field to store "hero" images. Desktop already does this, and our mobile clients will soon need this API as well.

The task here is to allow recording `preview_image_url` via `noteObservation` API, and return this field as part of `HistoryVisitInfo` and `HistoryMetadata`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2768)
┆Epic: History Metadata Storage & Sync

grigoryk:
We already have this field in moz_places; this PR adds just enough
piping to allow observing preview image urls and reading them back via
VisitInfo and HistoryMetadata APIs.

  - Note:

### DIFF
--- a/components/places/src/observation.rs
+++ b/components/places/src/observation.rs
@@ -59,6 +59,11 @@ pub struct VisitObservation {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
     pub is_remote: Option<bool>,
+
+    /// Semantically also a url::Url, See the comment about the `url` property.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub preview_image_url: Option<String>,
 }
 
 impl VisitObservation {
@@ -73,6 +78,7 @@ impl VisitObservation {
             at: None,
             referrer: None,
             is_remote: None,
+            preview_image_url: None,
         }
     }
 
@@ -116,6 +122,12 @@ impl VisitObservation {
 
     pub fn with_referrer(mut self, v: impl Into<Option<Url>>) -> Self {
         self.referrer = v.into().map(Url::into);
+        self
+    }
+
+    // v is a String instead of a Url to allow testing invalid urls as input.
+    pub fn with_preview_image_url(mut self, v: impl Into<Option<String>>) -> Self {
+        self.preview_image_url = v.into();
         self
     }
 
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -53,6 +53,19 @@ pub fn apply_observation_direct(
     if url.as_str().len() > super::URL_LENGTH_MAX {
         return Ok(None);
     }
+    // Make sure we have a valid preview URL - it should parse, and not exceed max size.
+    // In case the URL is too long, ignore it and proceed with the rest of the observation.
+    // In case the URL is entirely invalid, let the caller know by failing.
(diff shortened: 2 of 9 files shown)


grigoryk:
Still need to add the `historymetadata` bits here (e.g. expose preview url as part of metadata, and allow observing it as part of metadata observations, for the sake of iOS).

components/places/src/observation.rs
```diff
@@ -119,6 +125,11 @@ impl VisitObservation {
         self
     }
 
+    pub fn with_preview_image_url(mut self, v: impl Into<Option<String>>) -> Self {
```
mhammond, line 128: I think this should explicitly be `Option<Url>` - I haven't checked, but I suspect the code you have would *allow* you pass an `Option<Url>` but would also let you pass an `Option<String>` with an invalid Url, which would be bad. IOW, this should be more like `new` which takes a `Url` rather than in `Into<String>`

components/places/src/storage/history.rs
```diff
@@ -53,6 +53,11 @@ pub fn apply_observation_direct(
     if url.as_str().len() > super::URL_LENGTH_MAX {
         return Ok(None);
     }
+    if let Some(ref preview_image_url) = visit_ob.preview_image_url {
```
mhammond, line 56: Like a few lines above, we want to parse the URL here to ensure it's valid and canonical, before checking the length. Also, for the URL we just drop the observation entirely, which you also do here - I'd think we'd be better off just dropping the preview image URL in this case (ie, still record the observation, just with a null preview image?)

I think we should fail on an invalid URL rather than dropping it, just like we do for the url itself - ie, "invalid URL" == error, "too long url" == "silent ignore"

components/places/src/storage/history.rs
```diff
@@ -2739,6 +2760,20 @@ mod tests {
...
+            &conn,
+            VisitObservation::new(Url::parse("https://www.example.com/").unwrap())
+                .with_preview_image_url(Some(url))
+                .with_visit_type(VisitTransition::Link)
+                .with_at(Timestamp::now()),
+        )
+        .unwrap();
+        assert!(
+            maybe_row_preview.is_none(),
+            "Shouldn't insert overlong preview URL"
```
mhammond, line 2774: so yeah, I think this test would be that the visit was recorded but the url is None. And ideally another similar test that an invalid url throws.

components/places/src/storage/mod.rs
```diff
@@ -68,6 +68,7 @@ pub struct PageInfo {
     pub row_id: RowId,
     pub title: String,
     pub hidden: bool,
+    pub preview_image_url: String,
```
mhammond, line 71: It should be a Url here.

mhammond (changes requested):
This looks fine in general, but I think it should be a `Url` rather than a `String` in some cases, and I think we should ignore just that Url when it's too long rather than the entire observation.

components/places/src/observation.rs
grigoryk: I couldn't figure out how to pass in garbage inputs with this being Into<Option<Url>>, so kept it as-is... Afaik we only use this builder in tests anyway..?

mhammond (approved):

PR #4452, opened 2021-09-09
Adds support for multi-feature DTO in the nimbus SDK

tarikeshaq:
fixes https://mozilla-hub.atlassian.net/browse/SDK-365

Changes our schema ⚠️ ⚠️  to support multiple features per branch, I tried my best to make sure backward compatibility is OK, but I would definitely appreciate the extra sets of eyes in case I missed something.

On another note, based on https://github.com/mozilla/nimbus-shared/pull/172/files#diff-1b55c0b516a39b90d8243c91647cc296d59866125186b1c9b181fbdb37834f41R140, the `enabled` field is now nullable - I added a second commit for that too, but wanted to spit it out, because I'm not 100% sure how it's related to multi-features.

One more TODO from SDK-365:
- [ ] Test the requirement that each feature is only available in one experiment. Not sure how do this yet (or if it should be done in the SDK)

  - Note:


tarikeshaq:
This should be ready! I added a test with conflicting multi-features, and verified that each feature can only be involved in one experiment

I also added default serialization for the `probeSets` field, [since it's also removed in the proposal.](https://github.com/mozilla/nimbus-shared/pull/172/files#diff-1b55c0b516a39b90d8243c91647cc296d59866125186b1c9b181fbdb37834f41L62). We are keeping the field for now for backward compatibility, since we used to serialize our structs with it, and stored that in our persisted storage. For consistency, we are keeping it (I'm a little iffy on the value of the consistency here, in practice, it should be OKAY to remove the field, but I hear we've had issues with backward compatibility with `serde` so being a little extra safe)

At some point, we should do a migration and get rid of the field (among other fields that have default serialization)

travis79:
Bravo! I'll try and review this this-afternoon!

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,9 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+# Nimbus
+## What's changed
+- Nimbus now supports multiple features on each branch. This was added with backward compatibility to ensure support for both schemas. ([#4452](https://github.com/mozilla/application-services/pull/4452)), additional changes related to multi-features:
+    - [The `enabled` field became optional](https://github.com/mozilla/nimbus-shared/pull/172/files#diff-1b55c0b516a39b90d8243c91647cc296d59866125186b1c9b181fbdb37834f41R140)
+    - [The `probSets` field was removed](https://github.com/mozilla/nimbus-shared/pull/172/files#diff-1b55c0b516a39b90d8243c91647cc296d59866125186b1c9b181fbdb37834f41L62)
```
travis79, line 26:
```suggestion
    - [The `probeSets` field was removed](https://github.com/mozilla/nimbus-shared/pull/172/files#diff-1b55c0b516a39b90d8243c91647cc296d59866125186b1c9b181fbdb37834f41L62)
```

components/nimbus/src/enrollment.rs
```diff
@@ -1445,6 +1445,88 @@ mod tests {
...
+            "slug": "another-monkey",
+            "endDate": null,
+            "branches":[
+                {"slug": "control", "ratio": 1, "features": [
+                    { "featureId": "non-conflicting-feature", "enabled": true },
+                    { "featureId": "conflicting-feature", "enabled": false}
+                ]},
+                {"slug": "treatment","ratio": 1, "features": [
+                    { "featureId": "some-non-conflicting-treatment", "enabled": true },
+                    { "featureId": "some-non-conflicting-treatment", "enabled": false },
```
travis79, line 1466: nit: indentation, and trailing comma after last item in list

components/nimbus/src/enrollment.rs
```diff
@@ -1445,6 +1445,88 @@ mod tests {
...
+                    ]
+                },
```
travis79, line 1468: These are represented on the same line in a similar list above e.g.: `]},`

components/nimbus/src/enrollment.rs
```diff
@@ -1445,6 +1445,88 @@ mod tests {
...
+            "branches":[
+                {"slug": "control", "ratio": 1, "features": [
+                    { "featureId": "some-other-non-conflicting-feature", "enabled": true },
+                    { "featureId": "conflicting-feature", "enabled": false}
+                ]},
+                {"slug": "treatment","ratio": 1, "features": [
+                    { "featureId": "some_other_treatment", "enabled": true },
+                    { "featureId": "some_other_treatment", "enabled": false },
+                    ]
+                },
```
travis79, line 1505: Let's match the formatting in the branch above this and combine these to `]},` please

components/nimbus/src/lib.rs
```diff
@@ -339,8 +339,17 @@ pub struct Experiment {
...
+    // TODO: With multi-features,
+    // the `probeSets` field was removed, however for
+    // backward compatability, we will serialize an empty
+    // array if it's missing. this should be removed
+    // once multi-feature is implemented
+    // on the server side, and
+    // we do not expect any more `probe_sets`
+    // in any experiments
+    #[serde(default)]
     pub probe_sets: Vec<String>,
```
travis79, line 351: Serde is pretty smart, doesn't it ignore the field if it's not defined here, regardless of whether or not it's in the actual payload we receive?

components/nimbus/src/lib.rs
```diff
@@ -402,14 +415,24 @@ pub struct Branch {
     pub slug: String,
     pub ratio: i32,
     pub feature: Option<FeatureConfig>,
+    // We skip serializing if the `features` is none
+    // to keep backward compatibility. if we do not skip it,
+    // new experiments that only have one feature will end up
+    // serializing to a `features: Null` which does not match what
+    // old experiments were serialized to when they were persisted
+    // since they do not include this field at all.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
travis79, line 424: Nice!

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+            "channel": "nightly",
+            "endDate": null,
+            "branches":[
+                {
+                    "slug": "control",
+                    "ratio": 1,
+                    "features": [{
+                        "featureId": "feature1",
+                        "value": {
+                            "key": "value1",
```
travis79, line 656: nit: trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+                        }
+                    }]
+                },
+                {
+                    "slug": "treatment",
+                    "ratio":1,
+                    "features": [{
+                        "featureId": "feature3",
+                        "value": {
+                            "key": "value3",
```
travis79, line 672: nit: trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+            "channel": "nightly",
+            "endDate": null,
+            "branches":[
+                {
+                    "slug": "control",
+                    "ratio": 1,
+                    "features": [{
+                        "featureId": "feature1",
+                        "value": {
+                            "key": "value1",
```
travis79, line 728: nit: trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+                        }
+                    }]
+                },
+                {
+                    "slug": "treatment",
+                    "ratio":1,
+                    "features": [{
+                        "featureId": "feature3",
+                        "value": {
+                            "key": "value3",
```
travis79, line 744: nit: trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+            "endDate": null,
+            "branches":[
+                {
+                    "slug": "control",
+                    "ratio": 1,
+                    "features": [{
+                        "featureId": "feature1",
+                        "enabled": true,
+                        "value": {
+                            "key": "value1",
```
travis79, line 802: nit: copy-pasta trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+                    }]
+                },
+                {
+                    "slug": "treatment",
+                    "ratio":1,
+                    "features": [{
+                        "featureId": "feature3",
+                        "enabled": true,
+                        "value": {
+                            "key": "value3",
```
travis79, line 820: nit: copy-pasta trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+            "endDate": null,
+            "branches":[
+                {
+                    "slug": "control",
+                    "ratio": 1,
+                    "feature": {
+                        "featureId": "feature1",
+                        "enabled": true,
+                        "value": {
+                            "key": "value",
```
travis79, line 913: nit: copy-pasta trailing comma at end of list

components/nimbus/src/lib.rs
```diff
@@ -611,6 +634,341 @@ mod test_schema_bw_compat {
...
+                    }
+                },
+                {
+                    "slug": "treatment",
+                    "ratio":1,
+                    "feature": {
+                        "featureId": "feature2",
+                        "enabled": true,
+                        "value": {
+                            "key": "value2",
```
travis79, line 924: nit: copy-pasta trailing comma at end of list

travis79 (approved):
Just a handful of nits and a question about how Serde works with non-defined values (like removing probeSet). No need to block on further review from me

components/nimbus/src/lib.rs
tarikeshaq: It does! I kept it here only for when we serialize the experiment down to our persisted storage, where it'll end up with `"probeSets": []` on experiments serialized by an old version of the SDK but no `probeSets` at all on newer experiments if we remove the field.

Honestly, I think that's OK, I can't think of how that could go wrong... when we deserialize from persisted storage, if we deserialize an "old" experiment, we will just ignore the "probeSets" and that should be OK.. It'll also save us the tech debt. I'll give it a shot and add a couple of tests to our `persistence` tests

components/nimbus/src/enrollment.rs
jhugman: Calling this `some-non-conflicting-treatment` confused me:

1. `treatment` normally refers to a branch, not a feature.
2. once we get to about 3 of test/example anything, a simple naming scheme really helps understanding: if you don't like `feature-A`, `feature-B`, `feature-C` , there's always `alice-feature`, `bob-feature` and `charlie-feature` or `antelope`, `baboon`, `chimpanzee`.

components/nimbus/src/enrollment.rs
```diff
@@ -2198,6 +2283,47 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_multi_feature_per_branch_conflict() -> Result<()> {
+        let mut test_experiments = get_test_experiments();
+        // conflicting_experiments conflict with themselves, but not with
+        // test_experiments
+        let mut conflicting_experiments = get_conflicting_multi_feature_experiments();
+        test_experiments.append(&mut conflicting_experiments);
```
jhugman, line 2292: I think this is likely a smell where we're composing `Vec`s of `Experiment`s in the tests. Unless there's a good reason, I'd prefer not to be doing this.

components/nimbus/src/lib.rs
jhugman: Can we add this to `feature` as well?

components/nimbus/src/enrollment.rs
```diff
@@ -2198,6 +2283,47 @@ mod tests {
...
+            .iter()
+            .filter(|&e| matches!(e.change, EnrollmentChangeEventType::Enrollment))
+            .count();
+        assert_eq!(
+            3, enrolled_events,
+            "exactly two events should have Enrolled event types"
+        );
+        Ok(())
+    }
+
```
jhugman, line 2326: Aside: I hate how big our testing modules are, and how disorganized they are.

Also: do you think [`test_evolver_multi_feature_experiments`][1] needs adding to/refining/refactoring?

[1]: https://github.com/mozilla/application-services/blob/5774102b78c0a96b9acd80bf4c34701c85c725b3/components/nimbus/src/enrollment.rs#L2255

jhugman (commented):
I've added comments, mostly around the code clean up of removing `probeSets` and `enabled`.

I think there's ample scope to break those out in to different issues/PRs if you want to land this quicker.

I think it would also be worth filing a `good-first-issue` around moving the `mod tests` in each file to their own files.

components/nimbus/src/lib.rs
```diff
@@ -387,9 +396,13 @@ impl Experiment {
 #[serde(rename_all = "camelCase")]
 pub struct FeatureConfig {
     pub feature_id: String,
-    pub enabled: bool,
+    // Backward Compatibility note: Originally, this used to be
+    // required, so converting it into an option should not break
+    // backward compatibility since old experiments are guaranteed
+    // to have this field
+    pub enabled: Option<bool>,
```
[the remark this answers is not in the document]
tarikeshaq: `enabled` became nullable and not completely removed... we can opt to remove it since we aren't using it, but I wasn't sure how close to the `nimbus-shared` schema we want to stick


As for removing the fields in general, I'm getting hit by a test `test_migrate_db_v1_to_db_v2_round_tripping` that expects a persisted `v1` `json` to stay the same after getting deserialized and serialized... Is that a type of backward compatibility we want to keep? Or is the ability to deserialize the only thing we care about?

Because if we remove the fields, when we deserialize and serialize the `v1` experiments, they would no longer be the **exact** same `json`. I'm happy to modify that test, to only verify that `v1` experiments can be deserialized, and I can't see any bad implications from that, but was being extra careful.

Either way, I agree that the scope here is big, so I'll split that into its own PR

components/nimbus/src/lib.rs
tarikeshaq: That would bring us to the same question I have in https://github.com/mozilla/application-services/pull/4452/files#r709390143

Do we care that after deserializing and serializing an old json, it looks different? If not, we can do that, if we do, we might have to keep serializing the `feature` until we do a migration 😞

tarikeshaq:
OK updated this to include some of the recommendations, and made it **only include adding multi-features, the other changes to the DTO will happen in a second PR I'll push up**

components/nimbus/src/lib.rs
jhugman: I think the current roundtripping tests get this backwards. 

If we're doing a test comparing the serializations:

i.e. `in json_string` --> `Experiment` --> `out json_string`; then test `in json_string` == `out json_string`.

We should be doing a test comparing the live objects themselves:

i.e. `in Experiment` --> `json_string` --> `out Experiment`; then test `in Experiment` == `out Experiment`.

components/nimbus/src/persistence.rs
```diff
@@ -1121,49 +1120,12 @@ mod tests {
         // force an upgrade & read in the upgraded database
         let db = Database::new(&tmp_dir).unwrap();
 
-        let db_experiments = db.collect_all::<Experiment>(StoreId::Experiments)?;
```
tarikeshaq, line 1124: @jhugman I took out the code in the roundtripping test that was doing the `json` -> `Experiment` -> `json` comparisons and instead made it validate that "we can still deserialize an old `json`" it doesn't validate that the Experiment is still correct though... I tried going `Experiment` -> `json` -> `Experiment` but I realized that the `Experiment` struct we have is a `v2` struct, so that flow doesn't really make sense for this test.

jhugman (approved):
Thanks for making the changes you did. Sorry, I didn't get to review this before sprint end.

PR #4453, opened 2021-09-10
Add more branch paths for logins migrations

skhamis:
Mainly building off of the previous implementation of `MigrationPlan`. Added the rest of the branches that migrations could go with accompanying tests. 

Things now go through three phases:
1. Loop through both tables and fill `MigrationPlan` accordingly
2. Loop through `MigrationPlan` and apply any fixups and/or determine if we need to fix any "out of sync" issue (i.e if we fix up a mirror without a local, we move it to the local table)
3. Insert logins into their respective DBs and track proper metrics

It might increase how many loops but we only need to read from the DB once so I think the perf hit is minimized and gives us the best chance to reduce data loss

  - Note:

### DIFF
--- a/components/logins/src/migrate_sqlcipher_db.rs
+++ b/components/logins/src/migrate_sqlcipher_db.rs
@@ -9,9 +9,8 @@ use crate::encryption::EncryptorDecryptor;
 use crate::error::*;
 use crate::sync::{LocalLogin, MirrorLogin, SyncStatus};
 use crate::util;
-use crate::{
-    EncryptedLogin, Login, LoginDb, LoginFields, LoginStore, RecordFields, SecureLoginFields,
-};
+use crate::ValidateAndFixup;
+use crate::{EncryptedLogin, Login, LoginFields, LoginStore, RecordFields, SecureLoginFields};
 use rusqlite::{named_params, Connection, Row, NO_PARAMS};
 use sql_support::ConnExt;
 use std::collections::HashMap;
@@ -38,6 +37,15 @@ impl MigrationPlan {
 struct MigrationLogin {
     local_login: Option<LocalLogin>,
     mirror_login: Option<MirrorLogin>,
+    migration_op: MigrationOp,
+}
+
+#[derive(Debug, PartialEq)]
+enum MigrationOp {
+    Normal,        // Migrate as normal
+    FixedLocal,    // Local was fixed up, any mirror should be overridden
+    MirrorToLocal, // Local is irreparable, mirror goes to local
+    Skip(String),  // Fatal issue, don't migrate
 }
 
 // Simplify the code for combining migration metrics
@@ -132,8 +140,7 @@ pub fn migrate_from_sqlcipher_db(
     let encdec = EncryptorDecryptor::new(encryption_key)?;
 
     let migration_plan: MigrationPlan = generate_plan_from_db(&cipher_conn, &encdec)?;
-
-    let migration_metrics = migrate_logins(&migration_plan, &new_db_store, &encdec)?;
+    let migration_metrics = migrate_logins(&migration_plan, &new_db_store)?;
     let metadata_metrics = migrate_sync_metadata(&cipher_conn, &new_db_store)?;
 
     Ok(migration_metrics + metadata_metrics)
@@ -141,7 +148,7 @@ pub fn migrate_from_sqlcipher_db(
 
 fn generate_plan_from_db(
     cipher_conn: &Connection,
-    encryptor: &EncryptorDecryptor,
+    encdec: &EncryptorDecryptor,
 ) -> Result<MigrationPlan> {
     let mut migration_plan = MigrationPlan::new();
 
@@ -152,7 +159,7 @@ fn generate_plan_from_db(
         match get_login_from_row(row) {
             Ok(login) => {
                 let l_login = LocalLogin {
-                    login: login.encrypt(&encryptor)?,
+                    login: login.encrypt(&encdec)?,
                     local_modified: util::system_time_millis_from_row(row, "local_modified")
                         .unwrap_or_else(|_| SystemTime::now()),
                     is_deleted: row.get("is_deleted").unwrap_or_default(),
@@ -168,6 +175,7 @@ fn generate_plan_from_db(
                         //guid: key,
                         local_login: Some(l_login),
                         mirror_login: None,
+                        migration_op: MigrationOp::Normal,
                     });
             }
             Err(e) => {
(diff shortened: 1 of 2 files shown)


components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -207,14 +216,195 @@ fn generate_plan_from_db(
...
+                        record: RecordFields {
+                            id: dec_login.record.id,
+                            ..local_login.login.record
+                        },
+                        fields: new_entry.fields,
+                        sec_fields: new_entry.sec_fields.encrypt(encdec)?,
+                    };
+                    logins_to_override.insert(
+                        guid.to_string(),
+                        MigrationLogin {
```
mhammond, line 254: This pattern repeats a bit, but it's not clear how to avoid it - but maybe see what you can come up with - anything we can do to reduce the lines in this large function would help (although there is a balance to strike - eg, splitting it up into discrete functions would likely make everything more complex rather than less)

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -225,12 +415,22 @@ fn migrate_logins(
     let tx = conn.unchecked_transaction()?;
 
     for login in migration_plan.logins.values() {
-        // Migrate local login first
+        // Could not easily use an equality here due to the inner value
+        // But neccesary to ensure we log proper metrics
+        match &login.migration_op {
```
mhammond, line 420: This should be `if let` (I expect clippy will complain about that too, and will probably tell you exactly what to replace it with too :) I suspect that's what you were getting at in the comment, but I think it will work fine (and if so, the comment should also be removed)

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -225,12 +415,22 @@ fn migrate_logins(
...
+            MigrationOp::Skip(err) => {
+                num_failed_insert += 1;
+                insert_errors.push(err.to_string());
+                continue;
```
mhammond, line 424: That enum variant is declared as `Skip(String)`, so I think `err` is already a string?

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -246,7 +446,13 @@ fn migrate_logins(
                     // Weren't successful with local login, if we have a mirror we should
                     // attempt to migrate it and flip the `is_overridden` to false
                     if let Some(mirror_login) = &login.mirror_login {
-                        match migrate_mirror_login(&conn, &mirror_login) {
+                        match migrate_mirror_login(
```
mhammond, line 449: I would have though `if let Err(err) =...` would work here too (any match with 2 arms and where 1 is empty is suspect)

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -299,13 +505,7 @@ fn migrate_logins(
...
 }
 
-// This was copied from import_multiple in db.rs with a focus on LocalLogin
-fn migrate_local_login(
-    db: &LoginDb,
-    conn: &Connection,
-    local_login: &LocalLogin,
-    encdec: &EncryptorDecryptor,
-) -> Result<()> {
+fn migrate_local_login(conn: &Connection, local_login: &LocalLogin) -> Result<()> {
```
mhammond, line 508: a bit of a nit, but I don't think `migrate` is great here as part of the name - maybe replace `migrate` with `insert` here?

mhammond (approved):
This is looking good to me, thanks! Please make the changes I suggested and wait for at least Lougenia or Ben to approve it too before landing it (and also make whatever changes clippy suggests)

components/logins/src/migrate_sqlcipher_db.rs
bendk: For this one and the other repitition, what about switching to `migration_plan.logins.iter_mut()`?  That way I think  you could just change the `MigrationLogin` data directly.  I think it also makes it more clear which data is changing and which is staying the same.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -207,14 +216,195 @@ fn generate_plan_from_db(
...
+                }
+            };
+        } else if let Some(local_login) = &migration_login.local_login {
+            // Only local
+            let dec_login = local_login.login.clone().decrypt(&encdec)?;
+            // Skip fixup if the record is fine
+            if dec_login.entry().check_valid().is_ok() {
+                continue;
+            };
+            match dec_login.entry().fixup() {
```
bendk, line 291: Could you combine this and the last method call into `maybe_fixup()`?

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -246,7 +446,13 @@ fn migrate_logins(
                     // Weren't successful with local login, if we have a mirror we should
                     // attempt to migrate it and flip the `is_overridden` to false
                     if let Some(mirror_login) = &login.mirror_login {
```
bendk, line 448: This is a weird error case.  We had valid data but failed to insert it.  It seems like something really wrong is happening, like the disk is full or failing.  I don't have any suggestions on how to change the code, but I'll just note that it seems likely that inserts are going to fail in general.

bendk (approved):
I like how you split up the fixup/validation code and the insert code.  I agree with taking care of the things  @mhammond flagged, but I couldn't see anything else important so I'm marking approve.

components/logins/src/migrate_sqlcipher_db.rs
skhamis: I'll give `iter_mut` a shot, I was wrestling with the borrow checker quite a few times to try modify the data before I gave up and went for the overriding way.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -207,14 +216,195 @@ fn generate_plan_from_db(
...
-    store: &LoginStore,
+fn apply_migration_fixups(
+    migration_plan: MigrationPlan,
     encdec: &EncryptorDecryptor,
-) -> Result<MigrationMetrics> {
+) -> Result<MigrationPlan> {
+    let mut logins_to_override: HashMap<String, MigrationLogin> = HashMap::new();
+
+    for (guid, migration_login) in &migration_plan.logins {
+        if let (Some(local_login), Some(mirror_login)) =
```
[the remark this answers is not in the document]
skhamis: Thanks for this! I really like the matching as it feels easier to read (even with an additional indentation) and more rust-like -- will definitely go this route

components/logins/src/migrate_sqlcipher_db.rs
skhamis: I agree, this is really shouldn't happen in the wild without some much bigger issue present but felt like making sure we can at least attempt to prevent data loss (however unlikely) might save us in some small way.

skhamis:
Thanks for the review all, I tried various ways to reduce the LOC but it either increased complexity or decreased readability.  Right now it's easy-to-follow, easier to remove and I believe it's in a pretty good state right now and we can certainly revisit some of the steps in each of the phases as the dust settles with the entire body of work. Gonna go ahead and merge!

PR #4455, opened 2021-09-10
Add maps, lists and enums to Variables.

jhugman:
Fixes [SDK-367][1].

[1]: https://mozilla-hub.atlassian.net/browse/SDK-367

  - Note:


jhugman:
Still in draft because it needs swiftlint and swiftdocs. Otherwise, have at the review.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
```diff
@@ -106,20 +181,126 @@ interface Variables {
...
 }
 
+inline fun <reified T : Enum<T>> String.asEnum(): T? = try {
+    enumValueOf<T>(this)
+} catch (e: IllegalArgumentException) {
+    null
+}
+
+/**
+ * Uses `getString(key: String)` to find a string value for the given key, and coerce it in to
```
tarikeshaq, line 237: nit typo:
```suggestion
 * Uses `getString(key: String)` to find a string value for the given key, and coerce it into
```

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
```diff
@@ -15,53 +17,114 @@ import org.json.JSONObject
...
      *
      * N.B. the `key` and type `String` should be listed in the experiment manifest.
      */
     fun getString(key: String): String? = null
 
+    /**
+     * Find an array for this key, and returns all the strings in that array. If none exists, `null`
+     * is returned.
+     */
+    fun getStringList(key: String): List<String>? = null
```
tarikeshaq, line 82: Hmm is it possible to have this (and the other `get*List`) as one `getList<T: SomeInterface>(key: String): List<T>? = null` generic function?

I took a look at the implementations, and each `get*List` seems different enough to warrant the different functions, so feel free to completely ignore this 😅

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
```diff
@@ -15,53 +17,114 @@ import org.json.JSONObject
...
+
+    /**
+     * Find a map for this key, and returns a map containing all the entries that have strings
+     * as their values. If none exists, then `null` is returned.
+     */
+    fun getStringMap(key: String): Map<String, String>? = null
```
tarikeshaq, line 88: ditto for the generic question I had about lists here for maps

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
```diff
@@ -106,20 +181,126 @@ interface Variables {
...
+ */
+inline fun <reified T : Enum<T>> Variables.getEnum(key: String): T? =
+    getString(key)?.asEnum<T>()
+
+/**
+ * Uses `getStringList(key: String)` to find a value that is a list of strings for the given key,
+ * and coerce each item into an `Enum<T>`.
+ * If the value doesn't correspond to a variant of the list, then `null` is
+ * returned.
+ * Items of the list that are not underlying strings, or cannot be coerced into variants,
```
tarikeshaq, line 249: I'm curious in terms of minimizing damage when things go wrong and the variables include strings that cannot be coerced into variants... is it better if we return an error if we encounter a variable like that? or at least log or report it somewhere to flag that an experiment is faulty. It might not be worth the effort if we think our experiment review process will prevent this and hopefully with the manifest work that will never happen

components/nimbus/ios/Nimbus/FeatureVariables.swift
```diff
@@ -29,23 +29,39 @@ public protocol Variables {
     /// N.B. the `key` and type `String` should be listed in the experiment manifest.
     func getString(_ key: String) -> String?
 
+    func getStringList(_ key: String) -> [String]?
```
tarikeshaq, line 32: Do we want to add the docs for each of those like we did in kotlin?

tarikeshaq (approved):
This makes sense to me! I had a couple of questions and nits, but no need to block on reviews from me

I do have a question/observation we are starting to get tons of logic in our swift/kotlin code 😞 Apart of this is `uniffi` limitations I guess, but it's starting to grow our duplication across swift/kotlin. Moving forward, what do we need from `uniffi` to move this logic into the rust code (and I guess before that, do we **want** this logic to be in Rust?)

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
```diff
@@ -106,20 +181,126 @@ interface Variables {
...
+     * Gets a map value for this key, and transforms all `JSONObject`s that are values into `Variables`.
+     *
+     * If the value isn't a `JSONObject`, then returns `null`. Values in the map that are not `JSONObject`s
+     * are omitted from the final map.
+     */
+    fun getVariablesMap(key: String): Map<String, Variables>? = null
+
     // This may be important when transforming in to a code generated object.
-    fun <T> getVariables(key: String, transform: (Variables) -> T) = getVariables(key)?.let(transform)
+    fun <T> getVariables(key: String, transform: (Variables) -> T?) = getVariables(key)?.let(transform)
```
tarikeshaq, line 221: A quick follow-up thought, is this a breaking change? I don't know enough about kotlin, but is it making a type nullable OK for backward compatibility, or do we need to mark this as a breaking change?

mhammond: > is it making a type nullable OK for backward compatibility

I'd expect that it is a breaking change - IIUC it means callers of this will need to change?

components/nimbus/ios/Nimbus/FeatureVariables.swift
jhugman: Rats, I thought I'd done that!

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
jhugman: Yes, I think strictly this would be a breaking change.

Philosophically: If an API, that no-one uses, changes: is it really a breaking change?

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
jhugman: Agreed. I think I might be able to get the `getBoolList`, `getIntList` and `getStringList` to accept a type, however, I didn't because: 

 * there are enough restrictions from the kotlin compiler (see `reified T`) to make it non-straightforward.
 * many of the fundamental types are being coerced from Strings (e.g. Enums, Text, Drawable).
 * `getStringList` and `getTextList` have the same return type.

For this reason, I left it as `get*`, `get*List` and `get*Map`.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeatureVariables.kt
jhugman: Use of Nimbus involves lots of fallbacks: the app must be able to handle when the client is not involved in an experiment, or in a control branch.

Additionally, Nimbus should never be the cause of a crash: faulty configuration can be fed to the client (via experimenter) at any time, and the app should never crash because of it (that a particular branch causes a crash is another matter, however).

So, Nimbus is one of the few places, I think, where it's okay to failing fast is the wrong thing to do.

PR #4461, opened 2021-09-14
Migrating Components to Swift Packages Pt.1

skhamis:
Adding more components to be converted to allow consumption as swift packages. These cover all the ones I was able to easily get building with minimal work

### __Current Progress:__

- Already done:
  - crashtest
  - nimbus
  - rc_log
  - viaduct

- Done in this PR:
  - **autofill** 
  - **fxa-client**
  - **logins** 
  - **tabs**
  - **push**

- Needs work to convert:
  - places

- Not planned:
  - sync_manager/sync15
  - webext-storage

  - Note:


megazords/ios-rust/Cargo.toml
```diff
@@ -14,11 +14,8 @@ viaduct = { path = "../../components/viaduct" }
...
-# ref https://github.com/mozilla/application-services/issues/4352.
-#fxa-client = { path = "../../components/fxa-client" }
-# TODO: can't include logins until we get SQLCipher working on M1 simulator,
-# ref https://github.com/mozilla/application-services/issues/4352.
-# (or until we entirely get rid of SQLCipher)
-#logins = { path = "../../components/logins" }
+fxa-client = { path = "../../components/fxa-client" }
+logins = { path = "../../components/logins" }
+autofill = { path = "../../components/autofill" }
+push = { path = "../../components/push" }
```
tarikeshaq, line 20: I'm curious what should be our approach for components that aren't being consumed in iOS yet **at all**, like `autofill` and `push` if we add them here, then we force the scripts to compile the rust code and we generate the headers for them.. not that it's a problem, the scripts shouldn't run often (unless local dev will be impacted?)

The alternative would be to omit them, and add them when/if we get an iOS consumer that wants them

Thoughts?

tarikeshaq (approved):
Makes sense to me 🕺 Had one question about adding components that don't have an iOS consumer, but regardless of what we decide I'm happy with this 😄

megazords/ios-rust/Cargo.toml
skhamis: Yeah I went back and forth on this a little trying to decide the best way but the pros kinda outweigh the cons at this point:

Pros - component completeness, ensuring all components can build with the new system, able to be consumed with no need of a PR/Cut/release, since they select the components on consumption -- no extra overhead

Cons - little longer CI build times (which we'll eventually change to only generate artifacts on release builds)

If for some reason we start seeing really long build times, issues, I think it's okay to remove until ready for consumption.

tarikeshaq: I'm convinced 💯

PR #4463, opened 2021-09-14
Logins per field encryption

bendk:
Pretty simple change. Right now it doesn't actually change any behavior, because the only username validation throws rather than tries to fixup the data. But maybe in the future we will have a different one.

This also makes it so we fixup the data twice in this path since we do it again in add() or update(), whichever is called. That seems fine to me, but we could consider splitting up those functions to avoid this.

I also added a commit to add encryption functions for secure_login_data.  It seems like people were also good with this change.


mhammond (approved):

mhammond:
The CI failures here are odd - `main` seems to be passing, but this PR doesn't touch the files that are failing linting etc.

PR #4464, opened 2021-09-14
Use correct name for espresso-core dependency

badboy:
So ... uhm ... I just ran `./gradlew nimbus:build` in a checkout and it failed.
Then it took me 10 minutes to figure out the typo and now it builds again.

I don't know how this worked before? Is that code not exercised in CI?


tarikeshaq (approved):
😆 😆 I actually laughed out loud at your branch name

💯

PR #4465, opened 2021-09-15
Exposing login migrations for clients

skhamis:
This is just a baseline to get other work unblocked. Basically just exposes a new `migration_logins` method which does some light checking of paths and throws. This way is gives more control to the clients for dealing with migration (determining when to and what to do in pass/fail scenarios).

Questions still outstanding:
1. Do we handle the deletion/removal of db files?
2. Should we own creating the new DB? or just receive the paths from clients

  - Note:


components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -78,12 +78,60 @@ impl Add for MigrationPhaseMetrics {
...
+    }
+}
+
+fn migrate_sqlcipher_db_to_plaintext(
     old_db_path: impl AsRef<Path>,
     new_db_path: impl AsRef<Path>,
     old_encryption_key: &str,
     new_encryption_key: &str,
-    salt: Option<&str>,
+    salt: Option<&String>,
```
bendk, line 134: Nit: I think `&str` is usually preferable to `&String`.  It's more general since `&String` derefs to it.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -78,12 +78,60 @@ impl Add for MigrationPhaseMetrics {
...
+        (true, false) => match path.to_str() {
+            None => throw!(ErrorKind::InvalidPath(
+                sqlcipher_path.as_os_str().to_owned(),
+            )),
+            Some(s) => throw!(ErrorKind::InvalidDatabaseFile(s.to_string())),
+        },
+        (false, false) => throw!(ErrorKind::InvalidDatabaseFile(
+            "Both paths failed".to_string(),
+        )),
+    };
```
bendk, line 115: Do we want to require that the consumer to figure out if a migration should run or not?  It seems simpler for them to always call the function, then we decide if a migration needs to run.  IOW, I think `(_, false)` should be a no-op rather than an error.

I kind of think that we should create the new DB file, rather than requiring the consumer to.  That makes it easier to identify the case where we tried to run a migration and crashed in the middle of it.  In other words:
  - `(true, false)` should be the normal path
  - On `(true, true)`, we should run some code to handle the crashed-in-the-middle case.  Maybe that's deleting the new DB before trying to migrate or maybe it's just deleting both and starting from scratch.

components/logins/src/migrate_sqlcipher_db.rs
```diff
@@ -78,12 +78,60 @@ impl Add for MigrationPhaseMetrics {
     }
 }
 
-pub fn migrate_sqlcipher_db_to_plaintext(
+pub fn migrate_logins(
+    path: impl AsRef<Path>,
+    new_encryption_key: &str,
+    sqlcipher_path: impl AsRef<Path>,
+    sqlcipher_key: &str,
+    salt: Option<String>,
```
bendk, line 86: Could you keep a note about the `salt` argument -- even just the small one from before?  I don't feel like I'm going to remember how/when that was used once this project is in the past.

bendk (commented):
My feeling on the outstanding questions is yes to both, and in general we should be driving the migration process.  The main reason is that there are some tricky edge cases that we can handle better if we're the ones that create/delete the files.  But feel free to push back on this one if I'm missing the reasons to do it the other way.

I'm in the middle of approve/request changes.  If you agree with the above logic, then maybe make the changes before merging.  If not, then I like merging to get the API closer to finalized while we discuss it.

components/logins/src/migrate_sqlcipher_db.rs
lougeniaC64: 🤔 I'm not opposed to this approach but if we decide to handle error cases internally, how will that be communicated to the consumer?

In iOS when a sqlcipher database can't be opened the db file is moved and the storage layer `open` function is called again to create a new db file. I'm not sure this is the worthwhile because of course the data contained in the moved db file is useless, but at least the consumer is aware of what has occurred.

components/logins/src/migrate_sqlcipher_db.rs
skhamis: yeah the &str and String change is quite unfortunate but there is actually no way to define `Option<&str>` in uniffi actually. Since `[ByRef] string?` makes it `&Option<String>` instead of `Option<&str>` and I couldn’t figure how to do an option with referenced string and then once I get rid of the & internally in the option — compiler complains due to unknown size at compile time for str...

components/logins/src/migrate_sqlcipher_db.rs
bendk: Good question.  It seems like this depends on at least a couple things: how are consumers going to handle the error cases and do they need to log/report the result.  Maybe we should figure that out first.  This makes me lean towards merging the current code while we discuss how things ultimately should work.

skhamis: Agreed since there is a bit more nuanced to figuring that part out due to two consumers with potentially different ways of handling. I'll add test cases then we'll get this merged.

lougeniaC64: Sounds good. FWIW on the iOS side sentry logging occurs when the old db file is moved in the scenario I described above so I'm assuming (regardless of whether the old file is deleted or moved) they'd want to do the same if the migration fails.

bendk (approved):
Looking good.

PR #4480, opened 2021-09-17
Logins per field encryption

bendk:

  - Note:

### DIFF
--- a/components/logins/src/encryption.rs
+++ b/components/logins/src/encryption.rs
@@ -29,7 +29,6 @@
 // To make life a little easier, we do that via a struct.
 
 use crate::error::*;
-use jwcrypto::{self};
 use serde::{de::DeserializeOwned, Serialize};
 
 // Rather than passing keys around everywhere we abstract the encryption
@@ -40,9 +39,10 @@ pub struct EncryptorDecryptor {
 
 impl EncryptorDecryptor {
     pub fn new(key: &str) -> Result<Self> {
-        Ok(EncryptorDecryptor {
-            jwk: serde_json::from_str(key)?,
-        })
+        match serde_json::from_str(key) {
+            Ok(jwk) => Ok(EncryptorDecryptor { jwk }),
+            Err(_) => Err(ErrorKind::InvalidKey.into()),
+        }
     }
 
     pub fn encrypt(&self, cleartext: &str) -> Result<String> {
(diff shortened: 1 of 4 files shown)


components/logins/src/encryption.rs
```diff
@@ -75,14 +75,21 @@ impl EncryptorDecryptor {
...
+//     - If it returns false, then the key is no longer valid.  It should be regenerated and the DB
+//       data should be wiped since we can no longer read it properly
+pub fn create_canary(text: &str, key: &str) -> Result<String> {
+    EncryptorDecryptor::new(key)?.encrypt(text)
 }
 
-pub fn decrypt_string(key: String, ciphertext: String) -> Result<String> {
-    EncryptorDecryptor::new(&key)?.decrypt(&ciphertext)
+pub fn check_canary(canary: &str, text: &str, key: &str) -> Result<bool> {
+    Ok(EncryptorDecryptor::new(&key)?.decrypt(&canary)? == text)
```
skhamis, line 92: I actually like these changes, they feel more natural as an API. However I wonder for the sake of consistency, we follow how autofill does so that it's exactly mirrored? We actually have the same `encrypt_string` `decrypt_string` in our encryption.rs 

Logins encryption.rs: https://github.com/mozilla/application-services/blob/logins-per-field-encryption/components/logins/src/encryption.rs#L78-L91

Autofill encryption.rs: https://github.com/mozilla/application-services/blob/c12b286312299d66a2404fa807d2a13f035095f6/components/autofill/src/encryption.rs#L74-L87

If we expose that similar to autofill's udl, it might end up being easier on the consumer kotlin/ios side?  

If not let me know, more of a thought than any kind of an opinion.

components/logins/src/logins.udl
```diff
@@ -25,6 +25,14 @@ namespace logins {
     [Throws=LoginsStorageError]
     string encrypt_fields(SecureLoginFields sec_fields, [ByRef]string encryption_key);
 
+    // Create a "canary" string, which can be used to test if the encryption key is still valid for the logins data
+    [Throws=LoginsStorageError]
+    string create_canary([ByRef]string text, [ByRef]string encryption_key);
+
+    // Check that key is still valid using the output of `create_canary`.  `text` much match the text you initially passed to `create_canary()`
+    [Throws=LoginsStorageError]
+    boolean check_canary([ByRef]string canary, [ByRef]string text, [ByRef]string encryption_key);
```
skhamis, line 34: see above comment!

components/logins/src/encryption.rs
bendk: That's a really good question.  From the Kotlin side, I think either would work and the coding effort is similar.  I don't have an opinion either way.

Long-term, I think that maybe we could do a quick refactor of both logins and autofill to make them more consistent.  I also wonder if at some point we should expose the `EncryptorDecryptor` directly rather than recreating it for each call.

skhamis (approved):
Thanks for getting this out. LGTM!

PR #4484, opened 2021-09-20
Logins per field encryption

bendk:
This changeset moves the `find_login_to_update()` into the `LoginsDb`.  It feels pretty late in the game to make this change, so please push back on this one if you think it's a bad idea, or if it's going to require too many changes to your existing code.  The reason for the change is that I found some issues with the intended flow around `find_login_to_update()`.

My initial idea was that you would call `get_by_base_domain()`, then feed those results into `find_login_to_update()`.  However, `get_by_base_domain()` is much different that the old `potential_dupes_ignoring_username()` function.  The biggest difference is that it only uses the top 2 components of `origin` when matching which changes the dupe-checking logic substantially.

On friday I proposed adding a `get_by_origin()` function and using that, but that also would be a change from the old `potential_dupes_ignoring_username()` logic, since the old logic also took into account the `http_realm` and `form_action_origin` fields.

What we really want would be something like `get_by_origin_and_either_http_realm_or_form_action_url()` but I had a really hard time finding a nice name, signature, and description for that function.  I think the reason is that the function only makes sense when coupled with `find_login_to_update()` and they should be bundled together, so that's what I did.

One issue is we lose is the ability to cache the query results and avoid extra queries as the user is editing the username field.  However this seems okay to me for a few reasons:
  - I don't think there's hard evidence that this was an actual performance issue.
  - We also have an add dialog in fenix nightly where the user can edit the origin field.  That form has the exact same performance issue, and there's no caching there.
  - If we decide later that we need the caching, there are ways to add it.  Maybe `LoginsDb` could just cache the last query by `find_login_to_update()`.

### DIFF
--- a/components/logins/src/login.rs
+++ b/components/logins/src/login.rs
@@ -22,7 +22,8 @@
 //! * [`LoginFields`], [`SecureLoginFields`], [`RecordFields`] -- These group the common fields in the
 //!   structs above.
 //!
-//! Why so many structs for similar data?  Consider some common use cases in a hypothetical browser:
+//! Why so many structs for similar data?  Consider some common use cases in a hypothetical browser
+//! (currently no browsers act exactly like this, although Fenix/android-components comes close):
 //!
 //! - User visits a page with a login form.
 //!   - We inform the user if there are saved logins that can be autofilled.  We use the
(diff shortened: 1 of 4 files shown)


components/logins/src/login.rs
```diff
@@ -32,15 +32,16 @@
 //!     implementations follow this flow.  Still, we want application-services to support it.
 //!   - If the user chooses to autofill, we decrypt the logins into a `Vec<Login>`.  We need to
 //!     decrypt at this point to display the username and autofill the password if they select one.
+//!   - When the user selects a login, we can use the already decrypted data from `Login` to fill
```
lougeniaC64, line 35: nit: This is a note about this logins autofill explanation in general and not your addition to it, it might be worth noting that this is an explanation of the flow in a-c since we don't have ultimate control over how the api will be used. And even at present this doesn't apply to iOS as they don't even use `get_by_base_domain()`.

lougeniaC64 (approved):
Left a small nit but otherwise this seems fine to me.

PR #4489, opened 2021-09-21
We rely on url version 2.2 and just got lucky that we had it

mhammond:
We rely on `Into<String>` which was added in 2.2.? When I tried to use these crates in mozilla-central things got upset because it had 2.1.


mhammond:
(for more context, we already *do* use 2.2 as [seen here](https://github.com/mozilla/application-services/blob/main/Cargo.lock#L3521) - it's just lucky that we do because someone ran `cargo update` or similar)

tarikeshaq (approved):
:shipit:

PR #4490, opened 2021-09-21
Adds a targeting specific struct for targeting

tarikeshaq:
A part of https://mozilla-hub.atlassian.net/browse/SDK-384

### What's in this
- Uses a different struct `TargetingAttributes` to do the actual targeting, the struct is not persisted, and only created before targeting is done. This creates flexibility for the SDK to define targeting attributes without worrying about the app having to add them
- Adds a new targeting attribute: `is_already_enrolled` which is only true if the user was already enrolled in the experiment
- A bunch of tests for both the targeting change, and the general "make sure the `is_already_enrolled` works correctly"

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,8 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+## Nimbus
+
+### What's new
+
+- Nimbus can now target on `is_already_enrolled`. Which is true only if the user is already enrolled in experiment. ([#4490](https://github.com/mozilla/application-services/pull/4490))
--- a/components/nimbus/examples/experiment.rs
+++ b/components/nimbus/examples/experiment.rs
@@ -4,6 +4,7 @@
 
 use clap::{App, Arg, SubCommand};
 use env_logger::Env;
+use nimbus::TargetingAttributes;
 use nimbus::{
     error::Result, AppContext, AvailableRandomizationUnits, EnrollmentStatus, NimbusClient,
     RemoteSettingsConfig,
@@ -282,8 +283,12 @@ fn main() -> Result<()> {
             'outer: loop {
                 let uuid = uuid::Uuid::new_v4();
                 let mut num_of_experiments_enrolled = 0;
+                let targeting_attributes = TargetingAttributes {
+                    app_context: context.clone(),
+                    ..Default::default()
+                };
                 for exp in &all_experiments {
-                    let enr = nimbus::evaluate_enrollment(&uuid, &aru, &context.clone(), exp)?;
+                    let enr = nimbus::evaluate_enrollment(&uuid, &aru, &targeting_attributes, exp)?;
                     if enr.status.is_enrolled() {
                         num_of_experiments_enrolled += 1;
                         if num_of_experiments_enrolled >= num {
@@ -337,7 +342,12 @@ fn main() -> Result<()> {
                 // options.
                 let uuid = uuid::Uuid::new_v4();
                 let aru = AvailableRandomizationUnits::with_client_id(&client_id);
-                let enrollment = nimbus::evaluate_enrollment(&uuid, &aru, &context.clone(), &exp)?;
+                let targeting_attributes = TargetingAttributes {
+                    app_context: context.clone(),
+                    ..Default::default()
+                };
+                let enrollment =
+                    nimbus::evaluate_enrollment(&uuid, &aru, &targeting_attributes, &exp)?;
                 let key = match enrollment.status.clone() {
                     EnrollmentStatus::Enrolled { .. } => "Enrolled",
                     EnrollmentStatus::NotEnrolled { .. } => "NotEnrolled",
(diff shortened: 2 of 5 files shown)


components/nimbus/src/evaluator.rs
```diff
@@ -48,6 +55,7 @@ pub fn evaluate_enrollment(
     available_randomization_units: &AvailableRandomizationUnits,
     app_context: &AppContext,
     exp: &Experiment,
+    previous_enrollment: Option<&EnrollmentStatus>,
```
jhugman, line 58: I think if we construct the `TargetingAttributes` in this method, we'll be forever adding new arguments to this method.

If we change the `app_context: &AppContext` to `TargetingAttributes`, and create the `TargetingAttibutes` one level up, say in `evolve_enrollments` we should be able to add things to `TargetingAttributes` more easily (e.g. `update_age`, `install_age`, `previous_experiments`).

components/nimbus/src/evaluator.rs
```diff
@@ -332,47 +365,109 @@ mod tests {
...
+                architecture: Some("x86_64".to_string()),
+                device_manufacturer: Some("Samsung".to_string()),
+                device_model: Some("Galaxy S10".to_string()),
+                locale: Some("en-US".to_string()),
+                os: Some("Android".to_string()),
+                os_version: Some("10".to_string()),
+                android_sdk_version: Some("29".to_string()),
+                debug_tag: None,
+                custom_targeting_attributes: Some(custom_targeting_attributes),
+            },
```
jhugman, line 384: Nit: there's lots of these blocks around. Perhaps we could put them in a single method?

CHANGES_UNRELEASED.md
```diff
@@ -34,6 +34,7 @@ Use the template below to make assigning a version number during the release cut
   - `get*List`, `get*Map` to get lists and maps of all types.
   - Dictionary/Map extensions to map string keys to enum keys, and string values to enum values.
 - Nimbus now supports multiple features on each branch. This was added with backward compatibility to ensure support for both schemas. ([#4452](https://github.com/mozilla/application-services/pull/4452))
+- Nimbus can now target on `is_already_enrolled`. Which is true only if the user is already enrolled in experiment.
```
jhugman, line 37: Nice.

jhugman (changes requested):
Looking good so far.

I've marked this as `Request changes` only to let you know I've reviewed this PR. It's not meant to be rude.

components/nimbus/examples/experiment.rs
```diff
@@ -282,8 +283,12 @@ fn main() -> Result<()> {
             'outer: loop {
                 let uuid = uuid::Uuid::new_v4();
                 let mut num_of_experiments_enrolled = 0;
+                let targeting_attributes = TargetingAttributes {
+                    app_context: context.clone(),
+                    ..Default::default()
+                };
```
travis79, line 289: Would this be less expensive if it was done outside of `loop`?

components/nimbus/src/evaluator.rs
tarikeshaq: Good call! Moved up to `evaluate_enrollment`!

components/nimbus/src/evaluator.rs
tarikeshaq: I made it a lot simpler using `From` for now, I'll file a follow up to clean up the tests so the `AppContext` is constructed in a helper

components/nimbus/src/evaluator.rs
```diff
@@ -250,11 +268,12 @@ mod tests {
             android_sdk_version: Some("29".to_string()),
             debug_tag: None,
             custom_targeting_attributes: None,
-        };
-        assert_eq!(targeting(expression_statement, &ctx), None);
+        }
+        .into();
```
travis79, line 272: 🤩

components/nimbus/src/evaluator.rs
```diff
@@ -58,10 +66,20 @@ pub fn evaluate_enrollment(
...
+    } else {
+        false
+    };
+
     // Get targeting out of the way - "if let chains" are experimental,
     // otherwise we could improve this.
     if let Some(expr) = &exp.targeting {
-        if let Some(status) = targeting(expr, app_context) {
+        let targeting_attributes = TargetingAttributes {
+            app_context: app_context.clone(),
```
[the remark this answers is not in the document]
tarikeshaq: Hmmm I would have to have the TargetingAttributes keep a reference to the `AppContext` which should work since the lifetime of the TargetingAttributes is pretty small compared to the `AppContext`... 

I'll keep it like this for now, since I mutate the inner `AppContext` in the tests but it could be a good clean-up for the future.. I'll file a small backlog ticket

tarikeshaq (commented):
Thanks @jhugman!! Addressed the comments, should be good for a second look

travis79 (approved):
Just a single nit in the example from me, this is looking very good!

components/nimbus/examples/experiment.rs
tarikeshaq: Oop! good catch 😄

tarikeshaq:
Talked to @jhugman  - we'll merge this and I'll create a follow up to clean up the `clone` we have in our `evolve_enrollment` function, and figure out a good way to structure the `TargetingAttributes`

PR #4491, opened 2021-09-21
Adds days since install calculation to the SDK

tarikeshaq:
Part of https://mozilla-hub.atlassian.net/browse/SDK-384

Some important TODOs before this can land:
- [x] Add tests, for the changed behaviour and for targeting.. Will probably need to write an integration test to validate this
- [x] Add changelog
- [x] modify swift/kotlin code to pass in stub value for `installation_date`, then address the actual values on another PR


components/nimbus/src/lib.rs
```diff
@@ -36,22 +37,24 @@ use once_cell::sync::OnceCell;
...
+pub const DB_KEY_INSTALLATION_DATE: &str = "installation-date";
 // The main `NimbusClient` struct must not expose any methods that make an `&mut self`,
 // in order to be compatible with the uniffi's requirements on objects. This is a helper
 // struct to contain the bits that do actually need to be mutable, so they can be
 // protected by a Mutex.
 #[derive(Default)]
 struct InternalMutableState {
     available_randomization_units: AvailableRandomizationUnits,
+    installation_date: Option<DateTime<Utc>>,
+    app_context: AppContext,
```
tarikeshaq, line 57: This is probably my biggest annoyance with this change. The app context becomes mutable..

The purpose of this is to pass down the `days_since_install` all the way down to the targeting evaluation, and adding it to the context was the *fastest* way to do that. Alternatively, we can expose the concept of "first install" to the `enrollment` and `evaluator` modules, but that seemed like it would be a ton messier, and those modules have nothing to do with installation dates.

components/nimbus/src/lib.rs
```diff
@@ -36,22 +37,24 @@ use once_cell::sync::OnceCell;
...
 use std::path::PathBuf;
 use std::sync::Mutex;
+use std::{collections::HashSet, str::FromStr};
 use updating::{read_and_remove_pending_experiments, write_pending_experiments};
 use uuid::Uuid;
 
 const DEFAULT_TOTAL_BUCKETS: u32 = 10000;
 const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
-
+pub const DB_KEY_INSTALLATION_DATE: &str = "installation-date";
```
tarikeshaq, line 48: I persist the installation date in the `Meta` store, so we don't have to go looking at the directory every single run. Happy to take alternatives/ideas!

components/nimbus/Cargo.toml
```diff
@@ -34,6 +34,7 @@ sha2 = "0.9"
 hex = "0.4"
 once_cell = "1"
 uniffi = { version = "^0.14", optional = true }
+chrono = { version = "0.4", features = ["serde"]}
```
tarikeshaq, line 37: Added `chrono` as a dependency for convenience, checked its license, it's MIT/Apatche 2.0 licensed so we should be good. I'll make sure to regen the dependency summaries.

Its types already implement `serde::Serialize` and `serde::Deserialize` and its types can be built from strings, and `SystemTime` which allows me to keep the change here small.

components/nimbus/src/lib.rs
```diff
@@ -161,10 +173,11 @@ impl NimbusClient {
     }
 
     pub fn get_available_experiments(&self) -> Result<Vec<AvailableExperiment>> {
+        let state = self.mutable_state.lock().unwrap();
```
tarikeshaq, line 176: **IMPORTANT NOTE**: Will this API ever be called on the main thread? Note for myself to check, we shouldn't lock the main thread.

components/nimbus/src/lib.rs
```diff
@@ -189,9 +202,9 @@ impl NimbusClient {
         Ok(result)
     }
 
+    // TODO: Remove this, it's not used by consumers
     pub fn update_experiments(&self) -> Result<Vec<EnrollmentChangeEvent>> {
```
tarikeshaq, line 206: I'll bring this back, totally unrelated to the change, removed it because I was trying something out:

components/nimbus/src/lib.rs
```diff
@@ -207,17 +220,43 @@ impl NimbusClient {
 
     pub fn apply_pending_experiments(&self) -> Result<Vec<EnrollmentChangeEvent>> {
         log::info!("updating experiment list");
+        // If the application did not pass in an installation date,
```
tarikeshaq, line 223: Added this logic here because it happens off the main thread... Not sure if it's the best place for it

components/nimbus/src/lib.rs
```diff
@@ -36,22 +37,24 @@ use once_cell::sync::OnceCell;
...
-
+pub const DB_KEY_INSTALLATION_DATE: &str = "installation-date";
 // The main `NimbusClient` struct must not expose any methods that make an `&mut self`,
 // in order to be compatible with the uniffi's requirements on objects. This is a helper
 // struct to contain the bits that do actually need to be mutable, so they can be
 // protected by a Mutex.
 #[derive(Default)]
 struct InternalMutableState {
     available_randomization_units: AvailableRandomizationUnits,
+    installation_date: Option<DateTime<Utc>>,
```
tarikeshaq, line 56: Note to self, we don't need this here, the caller can pass it as a part of the app context

components/nimbus/src/lib.rs
travis79: Maybe a silly question, but: Could we just add the "installation_date" inside the `AppContext`? That way it provides a place for the app to provide this info, and if it's missing on init, then we can try to derive it ourselves. I guess the AppContext would still need to be mutable or replaced if we had to derive it, but then it's already there for when we pass it into the jexl-rs `evaluate_in_context` in `fn targeting`?

components/nimbus/src/lib.rs
travis79: Ahh, nevermind my other comment, it seems you may be thinking along the same lines.

tarikeshaq:
Huh https://app.circleci.com/pipelines/github/mozilla/application-services/28408/workflows/235fedb9-adfe-4dc6-92a8-2c1980dde3e7/jobs/116169 is very weird... It's a test failure in `places`.....

jhugman (changes requested):
Main changes needed:

 * make `AppContext` immutable again.
 * make a new struct with `app_directory_path: string` and `days_since_install: i32?` and pass it in as part of a `Nimbus` constructor.
 * calculate `install_date` and `last_update_date` and store them in the database.
 * calculate `days_since_update` and `days_since_install` from the stored dates and `now()`.
 * Add a `TimeBasedAttributes` to `TargetingAttributes` with the `days_since_update` and `days_since_install`.
 *

components/nimbus/src/lib.rs
```diff
@@ -207,17 +210,40 @@ impl NimbusClient {
...
+                // if we never persisted an installation date before, we will get the installation
+                // date by inspecting the database path
+                if let Some(installation_date) = persisted_installation_date {
+                    installation_date
+                } else {
+                    self.get_installation_date_from_database_path()?
+                }
+            };
+        let diff = Utc::now() - installation_date;
+        state.app_context.days_since_install = Some(diff.num_days());
```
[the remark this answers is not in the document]
tarikeshaq: I ended up avoiding changing the Nimbus constructor altogether. The consuming app will not be involved in setting the installation date, and we do it all in the SDK (by checking persisted data, or falling back to the filesystem creation date)

The tests manually set the targeting parameters in order to test that the targeting works.

components/nimbus/src/lib.rs
tarikeshaq: Not relevant anymore, the app context is no longer mutable

tarikeshaq:
Added tests, not 100% happy with how they look, but they test both the targeting and the fallback (it doesn't test that the actual fallback works, but it tests that we end up doing the fallback when we need to) I'll do a second go through the tests to see if I can make them better or cover more logic, but opening this up for more reviews!

components/nimbus/src/lib.rs
```diff
@@ -210,25 +238,109 @@ impl NimbusClient {
...
+        // If the application did not pass in an installation date,
+        // we check if we already persisted one on a previous run:
         let db = self.db()?;
         let mut writer = db.write()?;
+
+        let installation_date = self.get_installation_date(db, &mut writer)?;
+        let update_date = self.get_update_date(db, &mut writer)?;
+        let now = Utc::now();
+        let days_since_install = now - installation_date;
+        let days_since_update = now - update_date;
```
jhugman, line 250: Not sure of the API, but does `Utc::now() - DateTime<Utc>` gives a duration back in `days`?

components/nimbus/src/lib.rs
```diff
@@ -210,25 +238,109 @@ impl NimbusClient {
...
+                let update_date: Option<DateTime<Utc>> = store.get(writer, DB_KEY_UPDATE_DATE)?;
+                if let Some(update_date) = update_date {
+                    update_date
+                } else {
+                    let update_date = Utc::now();
+                    store.put(writer, DB_KEY_UPDATE_DATE, &update_date)?;
+                    update_date
+                }
+            }
+        };
```
jhugman, line 321: The logic in this method looks like it could be cleaned up. I don't think we should touch the database if there is no version in the `app_context`, so our match statement ends up as:

```rust
(Some(persisted), Some(update_date)) if persisted == current => update_date,
(_, _) => {
     // write the current and now to the datase
     // return now.
} 
```

components/nimbus/src/lib.rs
```diff
@@ -600,6 +716,518 @@ mod tests {
...
+                "appName": "fenix",
+                "appId": "org.mozilla.fenix",
+                "bucketConfig": {
+                    "count": 10000,
+                    "start": 0,
+                    "total": 10000,
+                    "namespace": "secure-gold",
+                    "randomizationUnit": "nimbus_id"
+                },
+                "targeting": "days_since_update == 10",
```
jhugman, line 1127: Zoinks!

jhugman (approved):
This is looking lovely. There are a few bits I was confused about which you should address/answer/comment, but I don't think I should hold you up in another review cycle.

Good job!

components/nimbus/src/lib.rs
tarikeshaq: The name was a bad name 😄 , changed the name to `duration_since_install` which includes the day information I later extract

components/nimbus/src/lib.rs
```diff
@@ -210,25 +238,140 @@ impl NimbusClient {
...
+                    update_date
+                }
+            }
+            _ => {
+                let update_date = Utc::now();
+                store.put(writer, DB_KEY_UPDATE_DATE, &update_date)?;
+                update_date
+            }
+        };
+        Ok(res)
```
jhugman, line 349: I think this could benefit from making the `match` a bit more explicit:

```suggestion
        let persisted_app_version: Option<String> = store.get(writer, DB_KEY_APP_VERSION)?;
        let update_date: Option<DateTime<Utc>> = store.get(writer, DB_KEY_UPDATE_DATE)?;

        Ok(match (persisted_app_version, &self.app_context.app_version, update_date) {
            // The app been run before, but has not just been updated.
            (Some(persisted), Some(current), Some(date)) if persisted == *current => {
                date
            }
            // The app has been run before, and just been updated.
            (Some(persisted), Some(current), _) if persisted != *current => {
                let now = Utc::now();
                store.put(writer, DB_KEY_APP_VERSION, current)?;
                store.put(writer, DB_KEY_UPDATE_DATE, &now)?;
                now
            }
            // The app has just been installed
            (None, Some(current), _) => {
                let now = Utc::now();
                store.put(writer, DB_KEY_APP_VERSION, current)?;
                store.put(writer, DB_KEY_UPDATE_DATE, &now)?;
                now
            }
            // The current version is not available, or the persisted date is not available.
            (_, _, Some(date)) => {
                date
            }
            // Either way, this doesn't appear to be a good production environment.
            _ => {
                Utc::now()
            }
        })
```

components/nimbus/src/lib.rs
```diff
@@ -210,25 +238,140 @@ impl NimbusClient {
...
+            log::info!("[Nimbus] Retrieved date from Context: {}", res);
+            return Ok(res);
+        }
+        let store = db.get_store(StoreId::Meta);
+        let persisted_installation_date: Option<DateTime<Utc>> =
+            store.get(writer, DB_KEY_INSTALLATION_DATE)?;
+        let res = if let Some(installation_date) = persisted_installation_date {
+            installation_date
+        } else {
+            let installation_date = match self.get_creation_date_from_path(&self.db_path) {
```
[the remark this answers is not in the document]
tarikeshaq: Hmmmm good catch, I wrongly assumed that the db_path is something that was created even before nimbus came by

This is mainly a problem with old installations of iOS apps, or new iOS apps... There are two paths forward for me, I'll timebox a little and pick one
- Have the apps pass the home directory
- Have iOS pass an installation time like android, might need to figure out a good API for that - if I find one, we can completely eliminate the `get_creation_date_from_path`!

components/nimbus/src/matcher.rs
```diff
@@ -69,6 +70,7 @@ pub struct AppContext {
     pub os_version: Option<String>,
     pub android_sdk_version: Option<String>,
     pub debug_tag: Option<String>,
+    pub installation_date: Option<i64>,
```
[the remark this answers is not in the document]
tarikeshaq: I'll file a follow up to clean up our AppContext (and possibly rename it to something else)

components/nimbus/src/lib.rs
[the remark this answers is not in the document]
tarikeshaq: Used `NSHomeDirectory`! (also added a home directory for android using `context.applicationInfo?.dataDir`, but for the purposes of this PR, it won't be used since android can pass in the installation date directly)

components/nimbus/src/matcher.rs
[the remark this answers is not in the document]
tarikeshaq: Filed https://mozilla-hub.atlassian.net/browse/SDK-392

Issue #4502, opened 2021-09-23
Investigate intermittent places test failure


tarikeshaq:
https://app.circleci.com/pipelines/github/mozilla/application-services/28408/workflows/235fedb9-adfe-4dc6-92a8-2c1980dde3e7/jobs/116169

Came across this in a completely not-places related PR, and re-running the job passed the test.

Not sure if it's a very rare CI environment issue that we can't do anything about and we'll never see again thou

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2801)

tarikeshaq:
We haven't seen this in a while so closing

PR #4503, opened 2021-09-23
Enable rust logging in CI and in more places tests

mhammond:
#4502 is strange, and it's almost impossible to see what might have happened for 2 reasons:

* `RUST_LOG` isn't set in CI - so even if tests try and enable logging it would not be seen.
* Places enabled logging in a very ad-hoc way, so only some tests would log even with `RUST_LOG` set - and this test was one that did not. So I just moved the logging initialization to a test-only function that most tests use right at the start of the test.

The end result being that if we ever see that failure again, we should end up with a log.


mhammond:
(oh - it's worth saying it wasn't strictly necessary to remove all the now-redundant calls to initialize the logging, but I felt it was a worthwhile cleanup)

tarikeshaq (approved):
Makes sense to me!

PR #4504, opened 2021-09-26
Ignore bad records in clients engine

JN3141:
Makes the clients engine check for tombstones and any deserialisation errors when receiving a client record, and ignores them if either are present

This PR addresses issue #1801

  - Note:
  - Added a test that covers two scenarios: one where there is a tombstone, and another where the JSON only has garbage fields that do not fit the schema

### DIFF
--- a/components/sync15/src/clients/engine.rs
+++ b/components/sync15/src/clients/engine.rs
@@ -69,11 +69,20 @@ impl<'a> Driver<'a> {
         for (payload, _) in inbound.changes {
             self.interruptee.err_if_interrupted()?;
 
-            // Unpack the client record. We should never have tombstones in the
-            // clients collection, so we don't check for `is_tombstone`.
-            // https://github.com/mozilla/application-services/issues/1801
-            // tracks deleting these from the server.
-            let client: ClientRecord = payload.into_record()?;
+            // Check for a payload's tombstone
+            if payload.is_tombstone() {
+                log::debug!("Record has been deleted; skipping...");
+                continue;
+            }
+
+            // Unpack the client record
+            let client: ClientRecord = match payload.into_record() {
+                Ok(client) => client,
+                Err(e) => {
+                    log::debug!("Error in unpacking record: {:?}", e);
+                    continue;
+                }
+            };
 
             if client.id == self.command_processor.settings().fxa_device_id {
                 log::debug!("Found my record on the server");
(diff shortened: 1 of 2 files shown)


components/sync15/src/clients/engine.rs
```diff
@@ -73,7 +73,16 @@ impl<'a> Driver<'a> {
             // clients collection, so we don't check for `is_tombstone`.
             // https://github.com/mozilla/application-services/issues/1801
             // tracks deleting these from the server.
-            let client: ClientRecord = payload.into_record()?;
+            if payload.is_tombstone() {
+                continue;
+            }
+
+            let client_res: Result<ClientRecord, serde_json::Error> = payload.into_record();
```
mhammond, line 80: You can probably just use a `match` here - something like:
```
let client = match payload.into_record() {
    Some(client) => client,
    Err(_) => continue,
```

(although logging that we did skip probably makes sense)

components/sync15/src/clients/engine.rs
```diff
@@ -73,7 +73,16 @@ impl<'a> Driver<'a> {
             // clients collection, so we don't check for `is_tombstone`.
             // https://github.com/mozilla/application-services/issues/1801
             // tracks deleting these from the server.
-            let client: ClientRecord = payload.into_record()?;
+            if payload.is_tombstone() {
+                continue;
```
mhammond, line 77: it probably makes sense to log that we ignored a payload here?

mhammond (changes requested):
Thanks for jumping in here! This looks great, although I think we can clean up the error handling, and unfortunately the CHANGES note went stale already.

JN3141:
> 
> 
> Thanks for jumping in here! This looks great, although I think we can clean up the error handling, and unfortunately the CHANGES note went stale already.

Can see that the changelog has jumped quite a bit since my commit already ahaha. Have cleaned that one up, and will include your suggestions on error handling 👍

components/sync15/src/clients/engine.rs
JN3141: Yup, that sounds good; I'll add some `debug` log lines for when we drop them

components/sync15/src/clients/engine.rs
JN3141: Ah, yup, using the `match` is quite a bit cleaner for the error handling; will give that a shot.

mhammond (approved):
This looks great, thanks!

PR #4507, opened 2021-09-27
Guard against invalid view time observations

grigoryk:
Certain client bugs lead to us making view_time observations that lead
to an i32 overflow (our return type for total_view_time is i32, but due
to in-sql incrementing, it's possible to encounter a value that's larger
than i32 in the database).

This patch does two things:
- on the way in, it guards against abnormally high view time
  observations, reporting an error to the client
- on the way out, it guards against overflowing total_view_time values
  in the database, capping at them i32::MAX

Since metadata naturally expires and is deleted, it's assumed that any
incorrect values won't survive for too long on actual devices, and so we
don't bother dealing with bad data in other places where it is used (for
ranking/sorting, for example).

Reporting errors to clients allows them to detect that something's wrong
with their logic.

  - Note:

### DIFF
--- a/components/places/src/error.rs
+++ b/components/places/src/error.rs
@@ -77,6 +77,9 @@ pub enum ErrorKind {
 
     #[error("Error opening database: {0}")]
     OpenDatabaseError(#[from] sql_support::open_database::Error),
+
+    #[error("Invalid metadata observation: {0}")]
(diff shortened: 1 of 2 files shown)


mhammond (approved):
👍

PR #4526, opened 2021-10-05
Fix sqlite-based logins syncing

mhammond:
This fixes the "no such column" error, but can't see how it might have fixed the Unwrap of None we saw.


bendk (approved):
The code looks good and it worked for me when I manually tested it.

lougeniaC64:
This resolved the local sync issue in iOS for me.

Issue #4467, opened 2021-09-15
Data review for logins component migration


mhammond:
In #4064, #4102 and others, we want to record metrics to track the reliability of the migration process.  [The glean yaml can be found here](https://github.com/mozilla/application-services/blob/logins-per-field-encryption/components/logins/android/metrics.yaml#L118)

1) What questions will you answer with this data?

**How successful (or otherwise) the migration from sqlcipher to sqlite is for logins on mobile.**

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:

**Establish baselines or measure changes in product or platform quality or performance.**

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?

**No alternatives were identified**

4) Can current instrumentation answer these questions?

**no**

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Data_Collection) found on the Mozilla wiki.   

<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>The number of login records we processed, including counts for success and failure of that processing</td>
    <td>interaction and technical </td>
    <td>https://github.com/mozilla/application-services/issues/4064</td>
  </tr>
  <tr>
    <td>The first 20 errors seen during the migration. These will not have PII</td>
    <td>technical </td>
    <td>https://github.com/mozilla/application-services/issues/4064 </td>
  </tr>
</table>

6) Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.

**This collection is documented in the Glean Dictionary at https://dictionary.telemetry.mozilla.org/"**

7) How long will this data be collected?  Choose one of the following:

**This is scoped to a time-limited experiment/project until date 2022-06-30**

8) What populations will you measure?

**All channels in all countries in all locales, for Android and iOS**

9) If this data collection is default on, what is the opt-out mechanism for users?

**Standard Firefox telemetry controls**

10) Please provide a general description of how you will analyze this data.

**Look at success and error rates and determine if any corrective action needs to be taken**

11) Where do you intend to share the results of your analysis?

**With the Firefox for Android and iOS teams**

12) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If 

**No**

[...]

PR #4527, opened 2021-10-05
Add logins migration metrics to main.

mhammond:
This helps bootstrap migration metrics. data-review in #4467

This should let the new fields appear in a new dashboard even though they wont have values yet.

We'll want this in a Nightly before logins lands, which I suspect will happen naturally, but we will need to remember to make sure that's true.


travis79 (approved):
Looks good to me, still covered by the initial data-review!

PR #4529, opened 2021-10-05
nimbus attempt re-enroll if errored out

tarikeshaq:
If enrollment errors out on one run, we try again on the next. This is needed because if a user is on an old version of nimbus, and they see new targeting attributes, they would error. Then, when they update the app, we don't try to re-enroll, thus they would never be enrolled.

Note the change in the swift file is to pass swiftformat CI, which started failing on a new upgrade of swiftformat

I'll cut a release for this and we'll need to uplift it


jhugman (approved):
This looks good ✅  and changing the test 👨‍🍳👌.

Please accept this Big Green Button with my thanks.

PR #4533, opened 2021-10-05
Logins per field encryption

bendk:
Two fixes for bugs that I noticed when trying to run through the migrations.


skhamis (approved):
Thanks for making these changes and adding the appropriate tests. These look good to me and I believe solve some of the issues we were seeing

skhamis:
Please fix the clippy suggestions -- however the swiftlint one is def suspect (seems unrelated to your changes)

PR #4609, opened 2021-10-26
Adds validation for the IR

tarikeshaq:
Adds some validation logic, before this is ready I would like to:
- [x] Finish the tests, still need to add a few more to cover all the cases
- [x] ~~Adds the validation logic for defaults, I had some logic that I scrapped, will get back to it tomorrow!~~ This is becoming more work than I thought it would be, I'll do it in a separate PR

### DIFF
--- a/components/support/nimbus-fml/src/intermediate_representation.rs
+++ b/components/support/nimbus-fml/src/intermediate_representation.rs
@@ -1,10 +1,10 @@
-use std::collections::HashMap;
-
 /* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+use crate::error::{FMLError, Result};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::collections::{HashMap, HashSet};
 
 /// The `TypeRef` enum defines a reference to a type.
 ///
@@ -35,7 +35,7 @@ pub enum TypeRef {
     // JSON objects can also represent a `Map<String, V>` or a `Map` with
     // keys that can be derived from a string.
     StringMap(Box<TypeRef>),
-    // We can coerce the String keys into Enums, so this repesents that.
+    // We can coerce the String keys into Enums, so this represents that.
     EnumMap(Box<TypeRef>, Box<TypeRef>),
 
     List(Box<TypeRef>),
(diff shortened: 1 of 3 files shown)


components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -54,6 +54,140 @@ pub struct FeatureManifest {
...
+                }
+                Ok(())
+            }
+            _ => Ok(()),
+        }
+    }
+
+    fn validate_enum_defs(&self, enum_names: &mut HashSet<String>) -> Result<()> {
+        for enum_def in &self.enum_defs {
+            if !enum_names.insert(enum_def.name.clone()) {
```
jhugman, line 137: Why try the insert? Why not a simpler contains?

components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -54,6 +54,140 @@ pub struct FeatureManifest {
...
+        Ok(())
+    }
+
+    fn validate_defaults(&self) -> Result<()> {
+        // TODO: Validate that the defaults of each feature
+        // in the feature manifest agree with the props
+        // There must be a default for each property
+        // either that's defined in the property itself
+        // or in the feature definition
+        Ok(())
```
jhugman, line 187: Ok, so there are two levels of defaulting emerging. I've broken these out into two separate bugs so you can land this more quickly:

* the [property level defaults][1].
* the [feature level defaults][2].

[1]: https://mozilla-hub.atlassian.net/browse/SDK-441
[2]: https://mozilla-hub.atlassian.net/browse/SDK-424

jhugman (approved):
This looks good. I've filed issues for [property defaults][1] and [feature defaults][2], which should make this good to go.

[1]: https://mozilla-hub.atlassian.net/browse/SDK-441
[2]: https://mozilla-hub.atlassian.net/browse/SDK-424

components/support/nimbus-fml/src/intermediate_representation.rs
tarikeshaq: uh, it's so that I insert into the set and check if something exists in one function call 😄

tarikeshaq:
Coverage is failing due to a [yanked `tera` version](https://github.com/Keats/tera/issues/689) - from my understanding, it should fix itself soon `grcov` already has an issue and PR for it https://github.com/mozilla/grcov/issues/704

PR #4611, opened 2021-10-26
Ignore clap yaml-rust advisory

tarikeshaq:
Our dependency advisory CI is failing again... I'm starting to question how useful of a check this is in practice, as opposed to a different advisory checking strategy..

An alternative I'm thinking of:
- When an advisory is detected (we can run a chron job for this, github actions maybe?) that we are impacted by, it creates an issue for it. If one is already created, it's ignored
- When anyone adds a new dependency, the normal "check-dependencies" job runs, we can probably add filters or something to circle CI to only run that job if `Cargo.toml`s are changed

Anyhoo, I'll create an issue for that, in the meantime:
https://github.com/mozilla/application-services/pull/4604 added a `yaml` feature for clap, clap `2.33` uses a version of `yaml-rust` that has an advisory out. Based on https://github.com/clap-rs/clap/issues/1569 it's a false positive for clap and can be safely ignored. Once clap v3 is out, we can upgrade and we'll get the new version of `yaml-rust` for free, and thus we can remove the ignore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,7 +177,11 @@ commands:
             #                       fix this: https://github.com/chronotope/chrono/pull/578
             #                       note that both the Nimbus-SDK and glean use chrono, so if we would like to move away from it, both projects
             #                       need to do that before we can remove the ignores (assuming `chrono` doesn't release a fixed version)
-            cargo audit --ignore RUSTSEC-2021-0019 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071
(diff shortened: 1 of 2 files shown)


travis79 (approved):
I was planning on using the [`serde_yaml`](https://crates.io/crates/serde_yaml) crate which also relies on `yaml-rust`, currently at version `0.4.5`. I don't think this is a problem, but I wanted to let you know that `clap` may not be the only crate with this dependency soon.

skhamis:
It seems the check-dependencies is still failing but for a different reason -- looks like a switch from `master` to `main` might as well regen the deps so get this fully green!

PR #4619, opened 2021-10-28
Remove all Carthage and switch to xcframeworks

skhamis:
Steps needed before this can get merged:
- [x] This PR merged: https://github.com/mozilla-mobile/firefox-ios/pull/9263 
- [X] Compiles/test pass locally
- [X] Compiles/test pass in CI
- [X] Can build & package Glean

This will essentially halt all usage of Carthage and instead use XCFrameworks for all ios-specific rust bindings. There is definitely some follow ups tasks to reduce the amount of times the xcframework gets build (xcframework needs to be built before testing, and another for releasing the framework) but the build times are roughly the same.

## Testing this PR
- Pull down this PR
- Open up Xcode
- Press play to run! 

  - Note:


skhamis:
Anyone is welcome to review this code but I have added specific people simply because this might affect their works flows either via Glean, Nimbus, iOS work, etc. Let me know if there are any issues!

automation/run_ios_tests.sh
```diff
@@ -2,8 +2,7 @@
 
 set -euvx
 
-carthage bootstrap --platform iOS --cache-builds
-
+./megazords/ios-rust/build-xcframework.sh --build-profile release
```
skhamis, line 5: This is a bit unfortunate because we actually try to run `build-xcframework.sh` as one of the steps in the XCode, however if the xcframework is not built before we run the tests -- we run into a plethora of linker/cc errors and other related issues. This works for now and I'll create a follow up task to revisit when we call these scripts

automation/smoke-test-firefox-ios.py
```diff
@@ -47,8 +47,6 @@
 if not Path(repo_path, "Carthage").exists():
     step_msg("Carthage folder not present. Running the firefox-ios bootstrap script")
     run_cmd_checked(["./bootstrap.sh"], cwd=repo_path)
-step_msg("Running carthage substitution script")
-run_cmd_checked(["./appservices_local_dev.sh", "enable", find_app_services_root()], cwd=repo_path)
```
tarikeshaq, line 51: I was thinking about this, this script right now won't run against a local version of this, so it's a little deceiving - until we have a similar `appservices_local_dev` script for the swift packages, should we print an error message here, or disable the script or something?

docs/build-and-publish-pipeline.md
```diff
@@ -76,17 +76,12 @@ For iOS consumers the corresponding steps are:
     * TODO: can we prevent these steps from being able to see the tokens used
       for publishing in subsequent steps?
 4. CircleCI builds two binary artifacts:
-    * A Carthage framework containing both Rust and Swift code compiled together, as a zipfile.
```
tarikeshaq, line 79: nit the line above should say `a binary artifact` as opposed to `two`

docs/howtos/locally-published-components-in-ios.md
```diff
@@ -1,24 +0,0 @@
-# Using locally-published components in Firefox for iOS
```
tarikeshaq, line 1: 👍 It's on my plate to figure out how to have an easy locally-published flow for iOS given this new setup, but I think this will be very important for us to get our components tested before they are consumed by the app

docs/howtos/consuming-rust-components-on-ios.md
```diff
@@ -1,47 +0,0 @@
...
-### Protobuf
-
-- *SwiftProtoBuf.framework* should be automatically downloaded by carthage while pulling in the application-services dependency.
-- [Add that framework to Xcode.](#adding-a-carthage-provided-framework-to-xcode)
-
-## Third-party licenses
-
-This project incorporates code from a number of third-party dependencies,
-under a variety of open-source licenses. You should review the license info
-in the file `DEPENDENCIES.md` and decide on an appropriate way to include
```
tarikeshaq, line 42: I want to triple check here, does firefox-ios include license summaries, and given that they are migrating to the xcframework is there work we should do to make sure those are still there? I know very little about the details here so I'm just curious if you know or we need to ask

CHANGES_UNRELEASED.md
```diff
@@ -26,3 +26,9 @@ Use the template below to make assigning a version number during the release cut
     reported in [#4575](https://github.com/mozilla/application-services/issues/4575). This is
     technically a breaking change as a dictionary described in the UDL changed, but in practice,
     none of our consumers used it, so we are not declaring it as breaking in this context.
+## General
+### ⚠️ Breaking Changes ⚠️
+
+  - Building application-services via Carthage and its related dependencies is now completely replaced by XCFrameworks and Swift Packages
```
tarikeshaq, line 32: 💯 💯

docs/howtos/consuming-rust-components-on-ios.md
```diff
@@ -1,47 +0,0 @@
-# Guide to Consuming Rust Components on iOS
```
tarikeshaq, line 1: Should we also remove the reference to this doc from `docs/SUMMARY.md`?

libs/verify-ios-ci-environment.sh
```diff
@@ -37,14 +32,6 @@ if ! [[ -x "$(command -v xcpretty)" ]]; then
   exit 1
 fi
 
-# For Xcode 12 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
```
tarikeshaq, line 40: I'm assuming this is no longer needed with Xcode 13 or was this a "carthage made us do it, and now we don't need it"?

megazords/ios-rust/Cargo.toml
```diff
@@ -20,4 +20,5 @@ autofill = { path = "../../components/autofill" }
 push = { path = "../../components/push" }
 tabs = { path = "../../components/tabs" }
 places = {path = "../../components/places" }
-places-ffi = { path = "../../components/places/ffi" }
\ No newline at end of file
+places-ffi = { path = "../../components/places/ffi" }
+glean-ffi = { path = "../../components/external/glean/glean-core/ffi" }
```
tarikeshaq, line 24: I'm not familiar with how the whole glean ios situation works, I need to go get myself familiarized with it - but in here, we ship the version of glean that we have as the submodule with our AS, then in `rust-components-swift` we import `glean-swift` I'm getting the feeling that we should only do one or the other, but again I'm not familiar at all with how all of that works, but wanted to make sure that this sounds OK?

As far as I can tell from Slack, the requirement for iOS was to bundle them together into one binary to reduce the size, should we then make sure the `rust-components-swift` doesn't import `glean-swift`? and instead, uses this one binary? @travis79 @badboy

megazords/ios-rust/build-xcframework.sh
```diff
@@ -179,4 +180,3 @@ rm -rf "$XCFRAMEWORK_ROOT/common"
 # Zip it all up into a bundle for distribution.
 
 (cd "$THIS_DIR" && zip -9 -r "$FRAMEWORK_NAME.xcframework.zip" "$FRAMEWORK_NAME.xcframework")
-rm -rf "$XCFRAMEWORK_ROOT"
```
tarikeshaq, line 182: I'm curious why we delete this here, is there a reason we keep the xcframework directory after zipping it up? Is this for local debugging?

megazords/ios-rust/Cargo.toml
tarikeshaq: That would actually alleviate my concerns in https://github.com/mozilla/rust-components-swift/pull/9, if we're shipping the glean rust code with the AS binary, we don't need to treat it differently than any other component (i.e it would be tied to our releases, but that's OK, because it used to be tied to our releases anyways with iOS). I don't know how close to the truth this is, so happy to be corrected!

megazords/ios-rust/MozillaRustComponents.h
```diff
@@ -18,3 +18,4 @@
 #import "placesFFI.h"
 // This is the ffi header and will be deleted after uniffi
 #import "RustPlacesAPI.h"
+#import "GleanFfi.h"
```
tarikeshaq, line 21: Should this be `glean.h`? From https://github.com/mozilla/application-services/pull/4619/files#diff-3729d0e227487c58350ead0b2d76baa018fcca7439fb97e5ec1c2b4620fd1ee1R136, or is this different?

tarikeshaq (commented):
BYE BYE Carthage 🎉 🚀 

A couple of concerns, but nothing unsolvable!
- mostly making sure the glean situation is good
- the license summaries, and making sure iOS still gets them
- the local dev flow, but that's on my own plate

megazords/ios-rust/Cargo.toml
badboy: Original bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1661539
The one argument left is probably still "increased binary size and startup time reduction due additional load on dyld".

Given that focus-ios is fine with consuming it through rust-components-swift, maybe that isn't as big of a problem anymore? Or this was not specifically looked at at all.
So at the very least we would need to check how this affects firefox-ios. I don't know how to check that though (nor do I have the time to look into it)

automation/smoke-test-firefox-ios.py
skhamis: Hmmm that's a good point that it's essentially not doing anything to help smoke test against a local copy of a-s. Will have to think about this script more to see how to make it clearer what to do (or point them to the local testing of SPM for a-s for now)

docs/howtos/consuming-rust-components-on-ios.md
skhamis: I'll definitely need to double check on this before we land anything, however that will be more on our actual packaged product on `rust-components-swift` rather than removal of anything locally. We also still generate our `DEPENDENCIES.md` and package it with the product, so no different from what our Carthage binary was doing at this point.

libs/verify-ios-ci-environment.sh
skhamis: Yeah and this makes it impossible to build natively on M1 due to arm64 being excluded as a simulator option

megazords/ios-rust/Cargo.toml
skhamis: Yeah so this is definitely where I expected most of the discussion to take place. However we essentially are now doing what our Carthage build was doing (in megazords/ios) where we actually build glean and package it in the same shared libs. This does indeed mean we do not need to add it as a dependency anymore in `rust-components-swift` and can actually make it a target. I also do like this option more as it gives people the option of consuming glean as a separate target rather as a specific dependency of Nimbus (ex. only Glean + Logins or something)

megazords/ios-rust/build-xcframework.sh
skhamis: Yeah this is one of the other areas up for discussion. Basically, to build all our tests, we need to somehow link our built rust libs to the tests so Xcode can compile it. This means two options:

1. Use `rust-components-swift`:
   + (plus) Easier to understand/use
   - (negative) To test your code, you'll need to make a local a-s xcframework, put it into `rust-components-swift` then change the package to point the locally
Ultimately this feels like a weird workaround and probably is worse than carthage for testing new code quickly.

2. Build Xcframework locally as part of the Xcode build process, and link to that to run the tests:
  - (plus) Don't have to think about it, part of Xcode build process
  - (plus) We already have this built out, just need to run the script
  - (negative) Extra artifacts hanging around locally

I think 2 is definitely the way to go here but open to alternatives. I eventually want to make a task that deals with having these artifacts and .zip files only made when necessary as well as associated scripts to clean these when people run the respective clean scripts

megazords/ios-rust/MozillaRustComponents.h
skhamis: I believe I am following similar to how we were bundling our builds with Carthage on the `megazords/ios` side of things https://github.com/mozilla/application-services/blob/main/megazords/ios/MozillaAppServices.h#L18

docs/howtos/consuming-rust-components-on-ios.md
tarikeshaq: This makes sense! looks like our `build-xcframework` script already copies over the `Dependencies.md`!

megazords/ios-rust/build-xcframework.sh
lougeniaC64: I think the second option makes sense. Consolidating the rough parts of this conversion to the a-s local development process where possible will make it easier for us to make improvements in the future.

lougeniaC64 (approved):
I can't speak for the impact to the Nimbus or Glean development process, but that aside LGTM. I imagine some of the future work is going to reveal itself once these changes are in use. Bye, Carthage!! 🎉 🥳

nbhasin2:
@skhamis not sure if this should be a part of this pr but we can bump up XCode to v13.2.1

nbhasin2 (approved):
This is great 🔥 🔥

skhamis:
> @skhamis not sure if this should be a part of this pr but we can bump up XCode to v13.2.1

I can try update the CI and see and what happens here!

.circleci/config.yml
```diff
@@ -237,8 +220,8 @@ executors:
       - image: cimg/rust:1.53.0
   macos:
     macos:
-      xcode: 13.1.0
-    resource_class: m2.medium
+      xcode: 13.2.1
+    resource_class: medium
```
tarikeshaq, line 224: hmm if we're changing this now, I think it should be `macos.x86.medium.gen2`, ref #4803

.circleci/config.yml
```diff
@@ -330,56 +313,7 @@ jobs:
...
-      - run:
-          name: "Carthage binary snapshot URL"
-          command: |
-            JSON_URL=https://circleci.com/api/v1.1/project/github/mozilla/application-services/$CIRCLE_BUILD_NUM/artifacts/0/dist/mozilla.app-services.json
-            echo "Add the following line to your Cartfile:"
-            echo "binary \"$JSON_URL\" ~> 0.0.1-snapshot # mozilla/application-services@$CIRCLE_SHA1"
-      - persist_to_workspace:
-          root: .
-          paths:
-            - MozillaAppServices.framework.zip
```
tarikeshaq, line 381: 😍 😍

automation/smoke-test-firefox-ios.py
tarikeshaq: Makes sense! (look at me replying to a message 3 months ago 😆) should we also remove smoke testing from our [steps in circle ci](https://github.com/mozilla/application-services/blob/main/.circleci/config.yml#L427-L433)?

docs/building.md
```diff
@@ -105,7 +105,6 @@ Configure maven to use the native windows maven repository - then, when doing `.
 
 ## Building for Firefox iOS
```
tarikeshaq, line 106: Yeah I think our docs in general on iOS can use a full re-haul (there  are a few sections referencing that "Carthage will *soon* be gone"), but we can do that as a followup

libs/verify-ios-ci-environment.sh
```diff
@@ -22,11 +22,6 @@ rustup target add "${RUST_TARGETS[@]}"
 
 # If you add a dependency below, mention it in building.md in the iOS section!
 
-if ! [[ -x "$(command -v carthage)" ]]; then
-  echo 'Error: Carthage needs to be installed. See https://github.com/Carthage/Carthage#installing-carthage for install instructions.' >&2
-  exit 1
-fi
-
 if ! [[ -x "$(command -v protoc-gen-swift)" ]]; then
```
tarikeshaq, line 25: nit: we don't have to do this now, but we no longer depend on anything protobuf swift related so we can probably get rid of this too 😄

tarikeshaq (approved):
LGTM! BYE BYE CARTHAGE 🎉 

Tested it locally, works like a charm, was able to also run the tests and all is good without carthage.

One thing we might wanna do (probably follow up): we can remove the `megazords/ios` directory and move all the `xcode` related stuff into the `megazords/ios-rust` - that way we'd wipe out the old megazord - but It's not urgent

libs/verify-ios-ci-environment.sh
skhamis: good idea! I'll probably keep that separate since idk how viaduct interacts with protobuf or what that story is (I _think_ it's fine but just to be safe).

travis79:
🎉

PR #4624, opened 2021-11-02
[SDK-426] Implement a minimally viable frontend for FML

travis79:
rebase cruft

### DIFF
--- a/components/support/nimbus-fml/src/parser.rs
+++ b/components/support/nimbus-fml/src/parser.rs
@@ -2,12 +2,607 @@
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::fs::File;
+use std::{collections::HashMap, convert::TryFrom, path::Path};
 
-pub struct Parser {}
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+
+use crate::{
+    error::FMLError,
+    intermediate_representation::{
+        EnumDef, FeatureDef, FeatureManifest, ObjectDef, PropDef, TypeRef, VariantDef,
+    },
+};
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct EnumVariantBody {
+    description: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct EnumBody {
+    description: String,
+    variants: HashMap<String, EnumVariantBody>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct ObjectFieldBody {
+    description: String,
+    #[serde(default)]
+    required: bool,
+    #[serde(rename = "type")]
+    variable_type: String,
+    default: Option<serde_json::Value>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct ObjectBody {
+    description: String,
+    failable: Option<bool>,
+    fields: HashMap<String, ObjectFieldBody>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct Types {
+    enums: HashMap<String, EnumBody>,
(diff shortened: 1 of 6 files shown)


components/support/nimbus-fml/src/fixtures/intermediate_representation.rs
```diff
@@ -95,14 +95,38 @@ mod dump_to_file {
         Ok(())
     }
 
+    fn write_yaml(fm: &FeatureManifest, nm: &str) -> Result<()> {
+        let root = std::env::var("CARGO_MANIFEST_DIR")
+            .expect("Missing $CARGO_MANIFEST_DIR, cannot write fixtures files");
+        let fixtures_dir = "fixtures";
+        let path: PathBuf = [&root, fixtures_dir, nm].iter().collect();
```
tarikeshaq, line 102: super nit: we can isolate a `get_fixture_path` function and use it in both `write_*` functions (or use generics, but with only two use cases, defs not worth it)

tarikeshaq (approved):
Looks good to me!

components/support/nimbus-fml/src/parser.rs
```diff
@@ -2,12 +2,237 @@
...
+            "Enum" => Ok(TypeRef::Enum(type_name.unwrap())),
+            "Object" => Ok(TypeRef::Object(type_name.unwrap())),
+            "List" => Ok(TypeRef::List(Box::new(TypeRef::try_from(
+                type_name.unwrap(),
+            )?))),
+            "Option" => Ok(TypeRef::Option(Box::new(TypeRef::try_from(
+                type_name.unwrap(),
+            )?))),
+            "Map" => {
+                // Maps take a little extra massaging to get the key and value types
```
jhugman, line 49: Yeah, this looks narlier than it needs to be :(

Let's get rid of `StringMap` and rename `EnumMap` to `Map`. What was I thinking.

components/support/nimbus-fml/fixtures/nimbus_features.yaml
```diff
@@ -0,0 +1,34 @@
+---
```
jhugman, line 1: Could we translate the IR into YAML and put it in a directory in `fixtures`, and then generate the code again and run the same tests that are running off of the IR?

I think `workflows/mod.rs` should be the place to do the running of the tests.

jhugman (approved):
Nice work.

I've added feedback below, but nothing that should hold you up another review cycle.

components/support/nimbus-fml/fixtures/nimbus_features.yaml
```diff
@@ -0,0 +1,34 @@
...
+            "jump-back-in": "false"
+            "recently-saved": "false"
+          ]
+    default: |
+      (
+        "sections-enabled": [
+          "top-sites": "true"
+          "jump-back-in": "false"
+          "recently-saved": "false"
+        ]
```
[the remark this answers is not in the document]
travis79: We can, I just hadn't determined exactly what I wanted the defaults to look like yet, so I opted for a JSON blob. You are right, the square brackets probably should be curly braces...

components/support/nimbus-fml/src/parser.rs
```diff
@@ -2,12 +2,209 @@
...
+#[derive(Debug, Deserialize, Serialize)]
+pub(crate) struct FeatureBody {
+    description: String,
+    variables: HashMap<String, FeatureVariableBody>,
+    default: String
+}
+#[derive(Debug, Deserialize, Serialize)]
+pub(crate) struct ManifestFrontEnd {
+    enums: HashMap<String, EnumBody>,
+    features: HashMap<String, FeatureBody>
```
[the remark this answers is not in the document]
travis79: Coming in the next commit!

travis79:
I need to file an issue to track adding the "Generate IR Manifest for channel" functionality, but this is ready for another look and hopefully landing.

jhugman (approved):
I have left some comments which should be addressed, but not enough to hold up for another review cycle. Nice work.

PR #4627, opened 2021-11-03
Gh 4617

bendk:
This implements `register_with_sync_manager()` for places.

This also takes the first step in a plan for migrating away from the old sync engine code.  Here's how I think it could work:
  - Implement new sync engines that work with `register_with_sync_manager()` and `get_registered_sync_engine()` (this PR)
  - Make all syncing go through `SyncManager`.  Delete the old syncing code.
  - Refactor the importer code to use the new `BookmarksSyncEngine` and `HistorySyncEngine` structs (or maybe delete it if we don't need it anymore)
  - Remove the old `BookmarksEngine` and `HistoryEngine` structs, move the code into `BookmarksSyncEngine` and `HistorySyncEngine`.
 - Remove `open_sync_connection()` and the fields in `PlacesApi` needed to support it

  - Note:

### DIFF
--- a/components/places/src/api/places_api.rs
+++ b/components/places/src/api/places_api.rs
@@ -271,9 +271,6 @@ impl PlacesApi {
         }
 
         let sync_state = guard.as_ref().unwrap();
-        // Note that this *must* be called before either history or bookmarks are
-        // synced, to ensure the shared global state is correct.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
 
         let mut mem_cached_state = sync_state.mem_cached_state.take();
         let mut disk_cached_state = sync_state.disk_cached_state.take();
@@ -317,9 +314,6 @@ impl PlacesApi {
         }
 
         let sync_state = guard.as_ref().unwrap();
-        // Note that counter-intuitively, this must be called before we do a
-        // bookmark sync too, to ensure the shared global state is correct.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
 
         let interruptee = conn.begin_interrupt_scope();
         let bm_engine = BookmarksEngine::new(&conn, &interruptee);
@@ -353,11 +347,6 @@ impl PlacesApi {
         let _guard = self.sync_state.lock().unwrap();
         let conn = self.open_sync_connection()?;
 
-        // Somewhat ironically, we start by migrating from the legacy storage
-        // format. We *are* just going to delete it anyway, but the code is
-        // simpler if we can just reuse the existing path.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
-
         storage::bookmarks::delete_everything(&conn)?;
         Ok(())
     }
@@ -366,11 +355,6 @@ impl PlacesApi {
         // Take the lock to prevent syncing while we're doing this.
         let _guard = self.sync_state.lock().unwrap();
         let conn = self.open_sync_connection()?;
-
-        // Somewhat ironically, we start by migrating from the legacy storage
-        // format. We *are* just going to delete it anyway, but the code is
-        // simpler if we can just reuse the existing path.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
 
         bookmark_sync::reset(&conn, &sync15::EngineSyncAssociation::Disconnected)?;
         Ok(())
@@ -381,11 +365,6 @@ impl PlacesApi {
         let _guard = self.sync_state.lock().unwrap();
         let conn = self.open_sync_connection()?;
 
-        // Somewhat ironically, we start by migrating from the legacy storage
-        // format. We *are* just going to delete it anyway, but the code is
-        // simpler if we can just reuse the existing path.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
-
         storage::history::delete_everything(&conn)?;
         Ok(())
     }
@@ -394,11 +373,6 @@ impl PlacesApi {
         // Take the lock to prevent syncing while we're doing this.
         let _guard = self.sync_state.lock().unwrap();
         let conn = self.open_sync_connection()?;
-
-        // Somewhat ironically, we start by migrating from the legacy storage
-        // format. We *are* just going to delete it anyway, but the code is
-        // simpler if we can just reuse the existing path.
-        HistoryEngine::migrate_v1_global_state(&conn)?;
 
         history_sync::reset(&conn, &sync15::EngineSyncAssociation::Disconnected)?;
         Ok(())
--- a/components/places/src/api/places_api.rs
+++ b/components/places/src/api/places_api.rs
@@ -2,10 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::bookmark_sync::engine::BookmarksEngine;
+use crate::bookmark_sync::engine::{BookmarksEngine, BookmarksSyncEngine};
 use crate::db::db::PlacesDb;
 use crate::error::*;
-use crate::history_sync::engine::HistoryEngine;
+use crate::history_sync::engine::{HistoryEngine, HistorySyncEngine};
 use crate::storage::{
     self, bookmarks::bookmark_sync, delete_meta, get_meta, history::history_sync, put_meta,
 };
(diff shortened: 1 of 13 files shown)


components/places/src/api/places_api.rs
```diff
@@ -30,6 +30,34 @@ use sync15::{sync_multiple, telemetry, MemoryCachedState, SyncResult};
...
+}
+
+// Called by the sync manager to get a sync engine via the PlacesApi previously
+// registered with the sync manager.
+pub fn get_registered_sync_engine(name: &str) -> Option<Box<dyn SyncEngine>> {
+    match PLACES_API_FOR_SYNC_MANAGER.lock().unwrap().upgrade() {
+        None => None,
+        Some(places_api) => match places_api.get_sync_connection() {
+            Ok(db) => match name {
+                "bookmarks" => Some(Box::new(BookmarksSyncEngine::new(db))),
```
tarikeshaq, line 49: The reason we don't need to `Arc::clone` here is because `Weak::upgrade` returns a **new** strong reference right? This makes sense, so every time we call `get_sync_connection` and it returns the same connection, it'll add a reference count so we're good!

(I'm just speaking out loud because I had to go look at docs once I saw that we don't clone the Arc here)

components/places/src/api/places_api.rs
```diff
@@ -82,9 +110,13 @@ pub struct PlacesApi {
     write_connection: Mutex<Option<PlacesDb>>,
     sync_state: Mutex<Option<SyncState>>,
     coop_tx_lock: Arc<Mutex<()>>,
+    // Used for get_sync_connection()
+    sync_connection: Mutex<Weak<Mutex<PlacesDb>>>,
```
tarikeshaq, line 114: Doesn't look pretty to have a mutex in a mutex, but makes sense I'm assuming we need this to have interior mutability and keep it the `PlacesApi` `Sync`?

tarikeshaq (commented):
I am not familiar with how our sync engines work in-depth, (thought it's about time to start poking around) but this looks good to me! I'll leave it to someone with more background here to approve 😄

components/places/src/api/places_api.rs
bendk: Yes I'm pretty sure that's how it works.  I used to have a clone there, but I deleted it and it still compiled so I think we're good :)

components/places/src/api/places_api.rs
lougeniaC64: Definitely want to second adding comments or refactoring here. I was initially confused but the explanation makes sense.

components/places/src/api/places_api.rs
```diff
@@ -162,6 +195,37 @@ impl PlacesApi {
         }
     }
 
+    // Get a database connection to sync with
+    //
+    // This function provides a couple features to facilitate sharing the connection between
+    // different sync engines:
+    //   - Each connection is wrapped in a `Mutex<>` to synchronize access.
+    //   - The mutex is then wrapped in an Arc<>.  If last the Arc<> returned is still alive, then
```
lougeniaC64, line 203: nit: "the last Arc<>"?

components/places/src/api/places_api.rs
bendk: BTW, one reason for the design is that I wanted to keep in place the current connection sharing behavior: We create a connection when syncing starts, and share it between the sync engines.  Once they are all done, we drop the connection.

I think we could lose the outer Mutex if we went for a different model, like always having one connection open, or opening the connection on the first sync then keeping it open.

components/places/src/bookmark_sync/engine.rs
```diff
@@ -61,6 +62,93 @@ where
     }
 }
 
+// `SyncEngine` implementation for bookmarks
+pub struct BookmarksSyncEngine {
```
mhammond, line 66: Can we just change the existing engine to not use lifetimes? Probably `Arc<PlacesApi>` rather than `Arc<PlacesDb>`, but that's a minor detail (and possibly wrong :) That could probably be done independently of uniffi'ing the sync manager, but it makes sense to me to do that first rather than leaving this massive "todo" which we might never do - there's no good reason to not take this refactor on now given there's no urgency on the uniffi of the sync manager. FWIW, I'd be very surprised if this refactor ends up with more than trivial conflicts from the unification of places.

mhammond (commented):
This looks great, although I think it should go further :)

In the first comment, you say "Make all syncing go through SyncManager. Delete the old syncing code." - it's not clear exactly what you mean here, but it's worth noting that iOS does not use the sync manager but does sync - hence logins.udl etc expose a `sync` method and the `wipe` function needed by sync. We've a plan to have iOS use the sync manager, but that's trickier than it sounds while iOS has sync engines not implemented in rust.

components/places/src/api/places_api.rs
mhammond: > BTW, one reason for the design is that I wanted to keep in place the current connection sharing behavior: We create a connection when syncing starts and share it between the sync engines. Once they are all done, we drop the connection.

I think it would be perfectly fine for 2 x Connections to be opened per sync, and also think it would be fine there was 1 lazy immortal one. We should avoid being wasteful, but I think it's small beer here?

bendk: I think lazy immortal would be great, I can make that change.

Opening a new connection each time seems fine from a resource point of view, but I like having the Mutex to avoid any possibility of having 2 bookmarks/history engines updating the DB at the same time.

components/places/src/bookmark_sync/engine.rs
bendk: > Can we just change the existing engine to not use lifetimes? Probably `Arc<PlacesApi>` rather than `Arc<PlacesDb>`, but that's a minor detail (and possibly wrong :) That could probably be done independently of uniffi'ing the sync manager, but it makes sense to me to do that first rather than leaving this massive "todo" which we might never do - there's no good reason to not take this refactor on now given there's no urgency on the uniffi of the sync manager. FWIW, I'd be very surprised if this refactor ends up with more than trivial conflicts from the unification of places.

Totally on board with this one now, based on:
  - Moving iOS to using the sync manager isn't trivia, therefore we need to continue to support the `PlacesApi` sync methods for a while
  - We can delete the import functionality

components/places/src/api/places_api.rs
bendk: Hmm, maybe it's not so easy.  I wanted to use `SyncLazy` to do this, but that's only available in nightly.  It wouldn't be hard to write a struct that implemented a similar API, but that doesn't reduce the complexity that much.  I think I'm just going to keep the current implementation, with the doc strings.

bendk:
Just pushed a couple changes to this branch:
  - Fixed the nits and added some documentation around the `Mutex<Weak<Mutex>>>`
  - Removed the `migrate_v1_global_state` code

I think I could push this one farther like mhammond suggests, but I have a few questions first:
  - Is it okay that I removed support for `migrate_v1_global_state`?
  - Are we sure we can drop the importing code?  I think this means:
     - The Android team has definitely decided to stop supporting fennec migrations for 96
     - The iOS team is okay with dropping support for their bookmarks migrations
  - Would it be okay to simply open the sync connection when we create `PlacesApi`?  That's how things like `LoginStore` work and it would be the simplest way to handle the shared connection.
  - How does `global_sync_state_v2` interact with the sync manager?  The non-sync manager sync functions like `PlacesApi.sync()` modify it, but the sync engines methods don't.  I ask because the simplest way to implement this be to get/release the mutex 3 times inside one of those methods (`get_disk_persisted_state()`, the sync engine method, then ``set_disk_persisted_state()`.`  This seems potentially dangerous, but then there's the `sync_state` mutex which we hold for the entire method.  So this all seems okay as long as those methods are the sync manager methods don't also need to access that data.

mhammond:
>     * Would it be okay to simply open the sync connection when we create `PlacesApi`?  That's how things like `LoginStore` work and it would be the simplest way to handle the shared connection.

I'd suck that up, but it would be a bit of a shame for non-sync users. I'd be happy with a ticket to fix this later though.

bendk:
Thanks for all the feedback on this one.  I have a new version that takes this further, fixes the nits, and has better documentation.  I'm going to close this one and open a new PR, since the code is pretty different.

PR #4629, opened 2021-11-05
Actually use --db-path option to set NimbusClient db_path in SDK CLI example

lmorchard:


lmorchard:
Was trying to tinker with bucketing and nimbus IDs and noticed that `--db-path` had no effect in the CLI example. This seems to fix it.

tarikeshaq (approved):
Sweet! And honestly, I feel like using the `tempdir` is probably a better default anyways 🤷

lmorchard:
Hmm, I don't have write access to this repo and can't self-merge. Guess I should ask for that if I'm going to have more PRs coming :)

PR #4637, opened 2021-11-09
Merging of objects, maps, maps of objects and maps of maps

jhugman:
Fixes [SDK-435][1].

[1]: https://mozilla-hub.atlassian.net/browse/SDK-435 

  - Note:


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/Collection+.kt
```diff
@@ -4,25 +4,39 @@
...
+        defaults.entries.forEach { entry ->
+            target[entry.key] = entry.value
+        }
+        entries.forEach { entry ->
+            val override = defaults[entry.key]
+                ?.let { d -> valueTransform(entry.value, d) }
+            target[entry.key] = override ?: entry.value
+        }
+        target
+    } ?: defaults + this
```
jhugman, line 42: This method simplifies the generated code massively.

components/support/nimbus-fml/src/backends/kotlin/gen_structs/enum_.rs
```diff
@@ -31,45 +30,20 @@ impl CodeType for EnumCodeType {
...
-
     /// The name of the type as it's represented in the `Variables` object.
     /// The string return may be used to combine with an indentifier, e.g. a `Variables` method name.
     fn variables_type(&self, _oracle: &dyn CodeOracle) -> VariablesType {
         VariablesType::String
     }
 
     /// A function handle that is capable of turning the variables type to the TypeRef type.
-    fn transform(&self, oracle: &dyn CodeOracle) -> Option<String> {
+    fn create_transform(&self, oracle: &dyn CodeOracle) -> Option<String> {
```
jhugman, line 40: * `variables_type` codes for how the `Variables` object in Kotlin thinks about this type. It's used to construct method calls, e.g. `getStringList` and `getBoolMap` or `getInt`.
 * `create_transform` is a method handles or closures for creating a value as it comes out of the `Variables` object. This can be None, since sometimes you don't need to transform it.
 * `merge_transform` is a method handle of closure for merging a value with a default. If this is `None`, then the object is just overwritten.
 * `value_getter` is the method that gets the variable out of the `Variables` object.
 * `value_mapper` is the expression to make a new value. It uses the `create_transform` to make a new value. This might be used recursively for lists or maps.
 * `value_merger` is the expression to merge a value with a default.
 * `property_getter` assembles it all together in an optional chain:

```
variables
  ?.{{ value_getter }}
  ?.{{ value_mapper }}
  ?.{{ value_merger }}
  ?: _default.value
```

Note to self: this should be documented someplace!

components/support/nimbus-fml/src/backends/kotlin/gen_structs/structural.rs
```diff
@@ -216,44 +254,39 @@ impl CodeType for ListCodeType {
...
+    fn value_merger(&self, _oracle: &dyn CodeOracle, _default: &dyn Display) -> Option<String> {
+        // We never merge lists.
+        None
+    }
+
     /// The name of the type as it's represented in the `Variables` object.
     /// The string return may be used to combine with an indentifier, e.g. a `Variables` method name.
     fn variables_type(&self, _oracle: &dyn CodeOracle) -> VariablesType {
-        unimplemented!("Nesting lists in to lists and maps are not supported")
+        unimplemented!("Because lists do not merge very well, maps of lists or lists of lists are not supported")
```
jhugman, line 289: Lists don't merge well (we just replace them), but I don't think this is the reason why we won't support `Map<String, List<String>`.

I think it's because we don't yet have a `variables?.getListList()` or `variables.asStringList` yet. If we have a need for it, then we should support it, but what we're doing is already likely more than we'll ever need.

components/support/nimbus-fml/src/backends/kotlin/templates/ObjectTemplate.kt
```diff
@@ -36,16 +36,19 @@ specify all values needed for the  feature #}
...
+
     companion object {
-        fun create(variables: Variables): {{class_name}}? {
-            return {{class_name}}(_variables = variables)
-        }
+        internal fun create(variables: Variables): {{class_name}}? =
+            {{class_name}}(_variables = variables)
+
+        internal fun mergeWith(overrides: {{class_name}}, defaults: {{class_name}}): {{class_name}} =
+            overrides.mergeWith(defaults)
```
jhugman, line 52: Generating these methods makes such a difference to the recursive mapping and merging functions.

components/support/nimbus-fml/src/backends/mod.rs
```diff
@@ -55,32 +55,71 @@ pub trait CodeType {
     fn type_label(&self, oracle: &dyn CodeOracle) -> String;
 
     /// The language specific expression that gets a value of the `prop` from the `vars` object.
-    fn get_value(&self, oracle: &dyn CodeOracle, vars: &dyn Display, prop: &dyn Display) -> String {
+    fn property_getter(
```
jhugman, line 58: These defaults are all Kotlin specific. This won't do in the long run. How about changing the `identifiers` module in to `common`?

jhugman (commented):
Self review

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/Collection+.kt
travis79: Nice!

components/support/nimbus-fml/src/backends/kotlin/templates/ObjectTemplate.kt
travis79: 🤩

travis79 (approved):
"Merge" it!

PR #4642, opened 2021-11-10
Conversion of errors for places uniffication

skhamis:
Still need to convert from HandleError, but throwing this up early to get feedback and see if I might've missed the target on some of these.

  - Note:


mhammond:
This is fantastic - it's better than I was thinking because you've managed to stick with just one set of errors.

mhammond:
And if we can do the same for iOS it will give us a huge amount of confidence as we move forward as we should be able to keep CI green for almost the entire effort.

skhamis:
> This is fantastic - it's better than I was thinking because you've managed to stick with just one set of errors.

Yes! I was quite surprised at how well it worked out once I started removing the kt-specific errors for the uniffi-generated ones. We'll need to keep RustError.kt around though for the lower-level number conversion but everything else can _hopefully_ go away!

components/places/src/error.rs
```diff
@@ -163,3 +163,158 @@ pub enum InvalidMetadataObservation {
     #[error("Observed view time is invalid (too long)")]
     ViewTimeTooLong,
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum PlacesError {
```
tarikeshaq, line 168: I was thinking about this, should we abandon the `error_support` approach before we do this? This will allow us to only have exactly one high-level enum and not one `internal` and one `external`. 

The biggest benefit `error_support` **offers** is backtrace support, but that's disabled and it looks like it's been broken for a while. (I tried experimenting with it a few weeks back with no success). There is a PR I'll open today with removing `error_support` for the `push` component, maybe we can see what that looks like and if its worth it to do that here - or maybe this will be one of the cleanups after uniffication, if getting it done this way is easier!

components/places/src/error.rs
```diff
@@ -163,3 +163,158 @@ pub enum InvalidMetadataObservation {
...
+    UrlParseFailed(String),
+
+    #[error("PlacesConnectionBusy error: {0}")]
+    PlacesConnectionBusy(String),
+
+    #[error("Operation Interrupted: {0}")]
+    OperationInterrupted(String),
+
+    /**
+     * Error indicating bookmarks corruption. If this occurs, we
```
tarikeshaq, line 182: nit: should we use `///` rust docs here if we want those to show up on the docs (honestly I don't think anybody looks at those though 😢)

components/places/src/error.rs
```diff
@@ -163,3 +163,158 @@ pub enum InvalidMetadataObservation {
...
+
+// A port of the error conversion stuff that was in ffi.rs - it turns our
+// "internal" errors into "public" ones.
+impl From<Error> for PlacesError {
+    fn from(error: Error) -> PlacesError {
+        let label = error.to_string();
+        let kind = error.kind();
+        match kind {
+            ErrorKind::InvalidPlaceInfo(info) => {
+                log::error!("Invalid place info: {}", info);
```
tarikeshaq, line 246: I'm not sure at all how log here works, but will this hit sentry? (and do we want it to?)

tarikeshaq (commented):
Took an initial look and noted a couple of questions! Looks great so far 🕺

components/places/src/error.rs
tarikeshaq: (for ref, here's the PR I have for push https://github.com/mozilla/application-services/pull/4650)

skhamis: I really like the changes you did in Push! I think places should move in that direction but I'd probably err on the side of caution and maybe add this as a task near the end-tail of the uniffication. Mainly just to reduce amount of actual changes while the core APIs are being uniffied -- however I don't see why we wouldn't be able to do this relatively quickly once the dust is settled!

components/places/src/error.rs
skhamis: Not entirely sure, would have to go a little deeper to figure out where this eventually logs to. This was mainly a port from ffi.rs so I assume it's doing whatever it was previously.

mhammond: Yeah, that does seem a little heavy-handed - all error matches [here](https://github.com/mozilla/application-services/blob/574450d9f3c1f5c9c6c2a71345cf0678440e985f/components/places/src/ffi.rs#L199) log an error, and we can see them in sentry - eg, loads of `places::ffi - URL parse error` and surprisingly only 1 `Invalid place info:`.

FWIW, these errors end up in sentry via [this helper](https://github.com/mozilla-mobile/android-components/blob/6ba529f9538b8ea9ffba1846821a76cc80c3de17/components/support/rustlog/src/main/java/mozilla/components/support/rustlog/RustLog.kt#L106)

I think we might as well keep things as they are for now just to reduce the moving parts in this patch, but do agree with the implication we should think this through a bit better (eg, all the "URL parse error" entries are probably fairly useless and probably just reflect a user typing something in the awesomebar)

mhammond: (and just for fun<sup>1</sup>, check out our ["unexpected error"s](https://sentry.prod.mozaws.net/operations/firefox/?query=is%3Aunresolved+%22unexpected+error%22) 🙄 )

<sup>1</sup> May not contain fun

skhamis:
This is ready to merged in when we think it's good. The failing tests are from the `places-uniffication` commit that adds a couple of methods to the udl but the error stuff should be good to go!

components/places/src/error.rs
mhammond: See also [where a-c submits these to sentry](https://github.com/mozilla-mobile/android-components/blob/0b5b051a6751077902784a235404a577c158b9e3/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt#L92) - so I *think* that many of these exceptions end up being submitted twice.

skhamis: > See also [where a-c submits these to sentry](https://github.com/mozilla-mobile/android-components/blob/0b5b051a6751077902784a235404a577c158b9e3/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt#L92) - so I _think_ that many of these exceptions end up being submitted twice.

Hmm interesting, so this would technically be an issue that is existing in today's world? Since the ffi.rs currently has that `log:error` for the old code while this would handle errors for any of the uniffi-ed errors? This may be something I'll need to also dive into to wrap my head around.

mhammond: Sorry I wasn't clear, but yes, this is a comment on the existing state of the world and nothing to do with your patch. It's possible that sentry is smart enough to combine these (eg, there will be an exact partial stack match when both are submitted, which might be enough for sentry to know they are the same)

Just yesterday I created https://github.com/mozilla-mobile/android-components/pull/11295 to try and avoid some of these errors ending up in sentry, but it turns out that might not actually work given this code *will* still report them. We probably do need to step back and have a think about this in more general terms.

skhamis: ah totally makes sense now -- I think your patch will definitely lessen the noise of things we won't/shouldn't fix and definitely rethink just converting these errors and continuing sending up to sentry (like in the case of this patch)

mhammond:
I'm killing this PR as I believe it's all in the `places-uniffication` branch (but the branch itself will remain if I screwed something up!)

PR #4645, opened 2021-11-11
Uniffi sync manager

bendk:
Based on #4634 

I made most of these changes mechanically.  I think we need someone who knows what these fields actually mean to double check that the changes make sense.

a-c PR: https://github.com/mozilla-mobile/android-components/pull/11256

  - Note:


lougeniaC64:
> I made most of these changes mechanically. I think we need someone who knows what these fields actually mean to double check that the changes make sense.
> 
@bendk Which fields are you referring to specifically? SyncParams??

bendk:
> > I made most of these changes mechanically. I think we need someone who knows what these fields actually mean to double check that the changes make sense.
> 
> @bendk Which fields are you referring to specifically? SyncParams??

SyncParams and SyncResult,.  And both of them have several nested classes.

megazords/full/Cargo.toml
```diff
@@ -11,11 +11,12 @@ crate-type = ["cdylib"]
 [dependencies]
 fxa-client = { path = "../../components/fxa-client" }
 logins = { path = "../../components/logins" }
+tabs = { path = "../../components/tabs/" }
```
mhammond, line 14: huh - why is this necessary and/or how did it work without it?

mhammond (approved):
This is awesome, great work Ben!

mhammond:
> SyncParams and SyncResult,. And both of them have several nested types.

They look correct to me.  There are some minor semantic changes (eg, splitting success and error) which all make sense, but are obviously changes that need to carry through to a-c.

megazords/full/Cargo.toml
bendk: I think because `sync_manager_ffi` depended on tabs so it was pulled in transitively.  But then `sync_manager` also depends on tabs, so I'm not exactly sure what's going on.  In any case, it seems good to explicitly list all the components that are in the megazord.

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,92 @@
...
+};
+
+dictionary SyncResult {
+    ServiceStatus status;
+    record<DOMString, string> failures;
+    sequence<string> successful;
+    string persisted_state;
+    sequence<string>? declined;
+    timestamp? next_sync_allowed_at;
+    string? telemetry_json;
```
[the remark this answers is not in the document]
bendk: I know, so close.  I think having external types would also work, since then this could just be a `SyncTelemetryPing` record.  I also thought about wrapping this class in a kt file that we define and giving it a `telemetry()` method, but it just didn't seem worth it yet.

bendk:
Thanks for the review.  I removed the protobuf related stuff.  One last thing I want to do is add some documentation to the `SyncParams` and `SyncResults` types.  I started writing it, but then realized that there's several fields that I don't really get.  Can you tell me what I should replace the `TODO`s with in the [UDL file](https://github.com/mozilla/application-services/blob/220bbdbaa873f754f74a499c384680e76d06c218/components/sync_manager/src/syncmanager.udl)?

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,106 @@
...
+    "PlacesError",
+    "AnyhowError",
+};
+
+dictionary SyncParams {
+    // Why are we performing this sync?
+    SyncReason reason;
+    // Which engines should we sync?
+    SyncEngineSelection engines;
+    // TODO
```
mhammond, line 25: The changes that this sync should make to the "account global" list of enabled/disabled engines (for example, if the UI was used to change an engine's state since the last sync.

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,106 @@
...
+    record<DOMString, boolean> enabled_changes;
+    // Keys to encrypt/decrypt data from local database files.  These are
+    // separate from the key we use to encrypt the sync payload as a whole.
+    record<DOMString, string> local_encryption_keys;
+    // Authorization for the sync server
+    SyncAuthInfo auth_info;
+    // TODO
```
mhammond, line 32: An opaque string, as returned in the previous sync's SyncResult and persisted to disk, or `null` if no such state is available. This includes information such as the list of engines previously enabled, certain server timestamps and GUIDs etc. If this value isn't correctly persisted and round-tripped, each sync may look like a "first sync".

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,106 @@
...
+    string? persisted_state;
+    // TODO
+    DeviceSettings device_settings;
```
mhammond, line 35: Information about the current device, such as its name, formfactor and FxA device ID.

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,106 @@
...
+};
+
+dictionary SyncResult {
+    // Result from the sync server
+    ServiceStatus status;
+    // Engines that synced successfully
+    sequence<string> successful;
+    // Maps the names of engines that failed to sync to the reason why
+    record<DOMString, string> failures;
+    // TODO
```
mhammond, line 80: State that should be persisted to disk and supplied to the `sync` method on the next sync.

components/sync_manager/src/syncmanager.udl
```diff
@@ -0,0 +1,106 @@
...
+    string persisted_state;
+    // TODO
+    sequence<string>? declined;
```
mhammond, line 83: The list of engines which are marked as "declined" (ie, disabled) on the sync server. The list of declined engines is global to the account rather than to the device. Apps should use this after every sync to update the local state (ie, to ensure that their Sync UI correctly reflects what engines are enabled and disabled), because these could change after every sync.

PR #4650, opened 2021-11-15
removes error-support from push

### LINKED ISSUE #4592: Remove `support::error` from push component
We are using a `Error { kind: ErrorKind }` pattern in many of our components, which is a pattern widely used in the wild. `support::error` also promises to keep backtraces, but as far as I can tell, it's a long standing issue that it doesn't quite work in android.

The big problem we're facing now, (look at https://github.com/mozilla/application-services/pull/4587#discussion_r731493588) is that we end up defining multiple `Error` types. One internal one that follows the pattern above, and an external one that uniffi ends up converting into foreign language errors/exception... When we convert from the internal one to the external one, we always end up discarding any useful information `support::error` could have helped us preserve.

Moral of the story is that we should probably do one of the two options:
- Fix the backtrace issue in android, and have uniffi pass that information back somehow
- Clean up our code and remove this `support::error` type, and instead have exactly on error type that we use both internally and eventually return to `uniffi`. [`nimbus`](https://github.com/mozilla/application-services/blob/main/components/nimbus/src/error.rs#L11) and [`autofill`](https://github.com/mozilla/application-services/blob/main/components/autofill/src/error.rs#L9) do that, and as far as we can tell, they're not doing any worse than the other components in the error reporting department

This ticket is to only have one error type with the push component and clean it up

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2851)
┆Story Points: 3
┆Sprint End Date: 2021-11-24

tarikeshaq:
fixes #4592 

As noted in the issue, `error_support` is not providing the value that it promises for us, and removing it removes a bunch of redundant code. It was a very mechanical change and I faced no issues (and didn't have to change the UDL at all!)

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,7 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Push
+### What's changed
+  - Push internally no longer uses the `error_support` dependency to simplify the code. It now directly defines exactly one error enum and exposes that to `uniffi`. This should have no implication to the consumer code ([#4650](https://github.com/mozilla/application-services/pull/4650))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2406,7 +2406,6 @@ dependencies = [
  "base64 0.12.3",
  "bincode",
  "env_logger 0.8.4",
- "error-support",
  "hex",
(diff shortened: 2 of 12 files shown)


bendk (approved):
The changes look good to me and I've been thinking the same thing for many components.  AFAICT, once a component is using UniFFI, then `thiserror` handles everything that `error_support` does in a nicer way.

Should `error.rs` be moved out of `internal` now?

Issue #4648, opened 2021-11-15
`LoginStore.add_or_update()` can return DuplicateLoginError


bendk:
Fenix beta users are seeing this occasionally (several crashers a day): https://github.com/mozilla-mobile/android-components/issues/11251

It seems to me that add_or_update should never be able to return DuplicateLoginError.  I can think of a couple ways that it might and we should fix those:
  - `update()` can return `DuplicateLoginError` if you're trying to update a login that's already a dupe of another login ([code](https://github.com/mozilla/application-services/blob/main/components/logins/src/db.rs#L516-L519)).  I think we should change that to log a warning and update the login with the GUID that was passed in.
  - `add()` can return `DuplicateLoginError` by design.  Maybe we should change `add_or_update()` to open a transaction and pass that to `add()` to avoid any chance of a login being inserted by another thread between the check and the insert.  I doubt this is happening in practice though, since `LoginStore` has a lock around the entire DB.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2890)
┆Sprint End Date: 2021-11-24

mhammond:
>     * `update()` can return `DuplicateLoginError` if you're trying to update a login that's already a dupe of another login

Yeah, I suspect this is the most likely problem - the user is manually editing a login unaware the future-duplicate exists.

mhammond:
> Yeah, I suspect this is the most likely problem - the user is manually editing a login unaware the future-duplicate exists.

FTR, in #4655 I see that I misunderstood this. I was assuming there was an issue in that there was no duplicate before editing, but the edit made one a dupe - but this case appears to be handled correctly. I think Ben's hypothesis is that there was a dupe before for reasons we don't understand but try very hard to avoid.

PR #4655, opened 2021-11-16
Don't return errors when updated duplicate logins (#4648)

bendk:
I don't think this is the ideal behavior.  It seems better to delete the
dupe like the comment mentions.  However, that would require more
discussion and in the meantime this should fix
mozilla-mobile/android-components#11251.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,7 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Logins
+### What's changed
+  - The `update()` and `add_or_update()` methods will log rather than return an error when trying to update a duplicate login (#4648)
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -513,10 +513,19 @@ impl LoginDb {
         let now_ms = util::system_time_ms_i64(SystemTime::now());
         let tx = self.unchecked_transaction()?;
 
-        // XXX - it's not clear that throwing here on a dupe is the correct thing to do - eg, a
-        // user updated the username to one that already exists - the better thing to do is
-        // probably just remove the dupe.
-        let entry = self.fixup_and_check_for_dupes(&guid, entry, encdec)?;
+        let entry = entry.fixup()?;
+
+        // Check if there's an existing login that's the dupe of this login.  That indicates that
+        // something has gone wrong with our underlying logic.  However, if we do see a dupe login,
+        // just log an error and continue.  This avoids a crash on android-components
+        // (mozilla-mobile/android-components#11251).
+
+        if self.check_for_dupes(&guid, &entry, encdec).is_err() {
+            // Try to detect if sync is enabled by checking if there are any mirror logins
+            let has_mirror_row: bool =
+                self.db.query_one("SELECT EXISTS (SELECT 1 FROM loginsM)")?;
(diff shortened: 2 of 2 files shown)


bendk:
This one fixes the main problem identified in the issue: trying to update a login when there's another duplicate login in the DB.  It doesn't address the theoretical race condition in `add()`.  I don't think that's going to be an issue in practice though, since we lock the database around each method in `LoginStore`.  I tried updating the code to hold a transaction for the entire operation, but it added a good amount of complexity and didn't seem worth it.

mhammond (commented):
I don't think this is an improvement - it will quieten the sentry event, but overall leaves us in a slightly worse place - we've still got the identical underlying problem but no clear symptoms. ie, if the hypothesis here is true, then we've a bad bug which will also be causing Sync to fail *on other devices* for users who hit it (they should fail to sync down the dupe and record failure due to that). The status quo at least tells us how many people have this problem (which thankfully isn't large). It's not even clear it will actually solve the problem for impacted users (eg, will the UI "do the right thing" in the face of these dupes? How will the user know which one was updated?)

It might even be directly related to sync - eg, I wonder what happens if (say) a non-normalized origin comes in? I can't see where sync would fix that up (but didn't look too hard, nor did I try and work out of something like that would actually cause the problem).

So unless there's something compelling I'm missing, I think, in the short term, we should do nothing other than try out a few other theories about what might cause this state in the first place and/or work out if there are additional diagnostics (see below) - then we can revisit some band-aid like this if the mobile teams start making noise.

Maybe something that is actionable here might be to leave "sync" breadcrumbs behind in sentry - eg, if we knew with certainty that everyone hitting this had sync enabled, we'd have more clues then we have now. I think that would be generally useful in some other reports too.

bendk:
The one issue with waiting is that it's currently flagged as a crasher.  That's the one thing that I think should be fixed in the short-term.  Maybe it's better to catch the exception on the A-C side -- although that's not completely trivial because of the abstract interface I mentioned in that issue.

If we're talking long-term, then the diagnostics seems great to me and also starting the conversation with some product managers about the best behavior for this would be if we get in this state.  I basically agree with the comment, we should pick 1 login to update, delete any dupes, and record some telemetry about it.

mhammond:
>  although that's not completely trivial because of the abstract interface I mentioned in that issue.

Sorry, but I can't find that discussion here or in  #4648 - what "that issue" are you referring to?

bendk:
The android-components one: mozilla-mobile/android-components#11251.  Most of these seem to be coming from `GeckoLoginStorageDelegate`, which doesn't use our API directly, it uses the one from `mozilla.components.concept.storage`.  I think this just means we would need to wrap/convert the exceptions somehom.

bendk:
@mhammond Updated this one to log an error like we talked about on the call.

mhammond (approved):

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,11 @@ Use the template below to make assigning a version number during the release cut
...
 
 -->
+
+## Logins
+### What's changed
+  - The `update()` and `add_or_update()` methods will log rather than return an error when trying to update a duplicate login (#4648)
+
+## Push
+### What's changed
+  - Push internally no longer uses the `error_support` dependency to simplify the code. It now directly defines exactly one error enum and exposes that to `uniffi`. This should have no implication to the consumer code ([#4650](https://github.com/mozilla/application-services/pull/4650))
```
mhammond, line 28: Actually, this is wrong - this change was released.

bendk: Good catch.  I'll fix that merge error.

PR #4656, opened 2021-11-16
SDK-417 generate json for experimenter from ir

tarikeshaq:
fixes https://mozilla-hub.atlassian.net/browse/SDK-417
Early draft for generating the json for experimenter from the IR, what I have right now generates the following json:
```json
{
 "homescreen": {
    "description": "Represents the homescreen feature",
    "hasExposure": true,
    "exposureDescription": "",
    "variables": {
      "section-ordering": {
        "description": "The ordering of the sections on the homescreen",
        "type": "json"
      },
      "sections-enabled": {
        "description": "A map of booleans",
        "type": "json"
      }
    }
  }
}
```
from the `full_homescreen.json` IR. Putting this up for early feedback to make sure I'm on the right direction

A few things before this is ready to merge:
- [x] Add schema validation as a part of the generation to make sure we abide by https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/schemas/ExperimentFeatureManifest.schema.json
- [x] Adds tests that run this against all the existing fixtures and validate against the schema
- [x] Address a few TODOs I have in the code


components/support/nimbus-fml/ExperimentFeatureManifest.schema.json
```diff
@@ -0,0 +1,78 @@
+{
```
tarikeshaq, line 1: reminder to myself to think about this a little, I **THINK** this is okay to keep here, but we have to be conscious of what happens if the schema changes in moz-central, I don't think it should fail the tests, especially considering others who aren't working on this will be hit by a failure.

That said I might setup a CI job that validates the current schema here vs the schema on moz-central and if they are different, it opens an issue on github for us to go get it fixed

This defs won't be a part of the work here, but I'll file a follow-up if we think that's a good approach

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -149,4 +177,44 @@ mod test {
...
+            for err in e {
+                errs.push_str(&format!("{}\n", err));
+            }
+            panic!("{}", errs);
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn test_schema_validation() -> Result<()> {
```
tarikeshaq, line 200: This test takes a noticeably long time to run because of all the file system I/O, I think I'll make the tests avoid writing the generated json to disk (instead validate on an in-memory json), and read the schema exactly once

components/support/nimbus-fml/ExperimentFeatureManifest.schema.json
jhugman: Is this a file grabbed from nimbus-shared, or have you grabbed it then modified it? (ISTR we discussed modifying it then asking experimenter for a schema file).

components/support/nimbus-fml/src/backends/experimenter_manifest.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+impl From<TypeRef> for ExperimentManifestPropType {
+    fn from(typ: TypeRef) -> Self {
+        match typ {
+            TypeRef::Object(_)
+            | TypeRef::EnumMap(_, _)
+            | TypeRef::StringMap(_)
+            | TypeRef::List(_)
```
jhugman, line 82: :(

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -149,4 +177,44 @@ mod test {
...
+                output: curr_out.clone(),
+                load_from_ir: true,
+            };
+            generate_experimenter_manifest(None, cmd)?;
+            let generated = fs::read_to_string(curr_out)?;
+            let generated_json = serde_json::from_str(&generated)?;
+            validate_against_experimenter_schema(
+                join(pkg_dir(), "ExperimentFeatureManifest.schema.json"),
+                &generated_json,
+            )?;
```
jhugman, line 216: Oh I see, this looks like a reasonably straightforward optimization. (Good second bug perhaps)

components/support/nimbus-fml/src/backends/experimenter_manifest.rs
```diff
@@ -0,0 +1,124 @@
...
+            | TypeRef::Enum(_) => Self::Json,
```
jhugman, line 83: I think that enums come out as a string, with the variants listed in an `enum` entry.

```json
"type": "string",
"enum": ["child", "adult"]
```
https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/FeatureManifest.js#77

components/support/nimbus-fml/src/backends/experimenter_manifest.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#[derive(Debug, Default, Clone, Serialize)]
+pub struct Variables(serde_json::Value);
+
+impl From<FeatureManifest> for ExperimenterFeatureManifest {
+    fn from(fm: FeatureManifest) -> Self {
+        // TODO: Figure out how experimenter consumes features
+        // i.e, does it expect each feature individually, or all together
+        // for the sake of a draft, I just generate the first one
+        let feature = fm.feature_defs[0].clone();
```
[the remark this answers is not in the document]
tarikeshaq: Yup! That's what I ended up doing on the latest patch

components/support/nimbus-fml/ExperimentFeatureManifest.schema.json
tarikeshaq: Ya in the latest patch I modified the schema to allow for all the features to be in one json (by making them an arbitrary number of key-value pairs, where the value abides by the schema that I got from moz-central)

jhugman (approved):
This looks great! Landing imminent! Thank you.

Issue #4662, opened 2021-11-18
Sync manager panic: index out of bounds


mhammond:
[this report](https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/12007622) says `ffi_support::error - Caught a panic calling rust code: "index out of bounds: the len is 2 but the index is 2"` - calling `SyncManager::sync` but no further info.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2902)

bendk:
It's really hard to tell how often this is happening because it's getting grouped with other issues in Sentry.  It seems like switching sync manager to uniffi might fix this error or at least make it easier to debug.  Once we do that, I don't think there will be any `ffi_support` dependency.

mhammond:
We haven't seen this recently, sync manager has since moved to uniffi and parking_lot, so I think we can close it.

Issue #4661, opened 2021-11-18
Logins poison errors since per-field encryption landed.


mhammond:
as reported by @grigoryk eg, https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/14927128 and https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/14011601

Sadly, both of these are just a symptom of some other, as yet unknown error, which will be a panic while our mutex is held. Once that happens, all future attempts to use our APIs will result in a PoisonError like the links above - but it's not the code causing the PoisonError that we need to solve, it's the underlying panic.

Only 16 users so far, but we should try and nail this down.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/SYNC-2901)
┆sprintEndDate: 2022-02-04

mhammond:
I wonder if the panics are actually making it to sentry - eg, https://sentry.prod.mozaws.net/operations/firefox-nightly/?query=device.family%3A%22IN2013%22+logins

This is all events for a fairly obscure device - 8 events in total, 3 of which are these poison errors, and none of the others look like they should poison the mutex.

mhammond:
ah ha - [this report](https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/12635627/) is pointing at [this sync code](https://github.com/mozilla/application-services/blob/249e718decb6c5db82fd792f5c9f5968d4235612/components/logins/src/sync/merge.rs#L164-L169)

bendk:
I made #4667 for that particular piece of code.

Amejia481:
@mhammond and @bendk 👋🏽 
Just to make you aware that we have some new reports of this issue:
https://github.com/mozilla-mobile/fenix/issues/22910

data-sync-user:
➤ Janet Dragojevic commented:

We believe this is fixed with a-s 86.2.2 - the main source of panics which cause poison errors.

a-s 87.1.0  will have the “parking-lot” change which will fix the remainder of the errors - is in nightly.

PR #4663, opened 2021-11-18
Move to `parking_lot::Mutex` in logins, places and sync_manager

mhammond:
Avoiding panics in all cases, like #4661 and #4662, is going to be whack-a-mole, and when one happens, we pollute sentry with many many ```Caught a panic calling rust code: "called `Result::unwrap()` on an `Err` value: PoisonError { .. }"``` errors.

I don't think we rely on the poisoning strategy for correctness, so we should consider moving to parking_lot.

(Not requesting review yet - I think we should discuss a little - so thoughts welcome!)

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -22,3 +22,8 @@ Use the template below to make assigning a version number during the release cut
 ## Logins
 ### What's changed
   - The `update()` and `add_or_update()` methods will log rather than return an error when trying to update a duplicate login (#4648)
+
+## Logins, Places, SyncManager
+### What's Changed
+  - These packages all use `parking_lot::Mutex` instead of `std::Mutex`, meaning we should no
+    longer see errors about mutexes being poisoned.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1606,6 +1606,7 @@ dependencies = [
  "lazy_static",
  "log",
  "more-asserts",
+ "parking_lot",
  "rusqlite",
  "serde",
  "serde_derive",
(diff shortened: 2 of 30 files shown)


tarikeshaq:
I think I agree with this change, but before landing this, I wonder if we can audit our `panic`s? Just in case there is a situation where the poisoning of the mutex is better than the alternative - I can't think of a situation right away (which I guess is a good sign for this change?) I think the only way the poisoning would be better is if there is some type of data inconsistency or vulnerability that can be exposed if something is half done in one call, then another call goes ahead

That all said, I'm thinking if we **CAN** audit our panics extensively then we wouldn't have ended up in the situation we are in (where poison errors keep popping up..)

This is a long way for rubberstamping on my end if we can't think of any situations where the poison is better 🤷

lougeniaC64:
Agree with @tarikeshaq's sentiments. Worse case scenario the poison errors are reported as elusive panics in Sentry, but at least that gets us closer to identifying what caused the invalid state and resolving the crashes. Best case the actual issues are easy to identify and resolve. Either way, nervous but r+.

mhammond:
> I wonder if we can audit our panics?

I don't think we can in any reasonable way. Eg, the 2 issues linked above are clearly logic errors that should be impossible. We could replace those 2 instances with errors, but I don't think that solves the general whack-a-mole problem. One is an "index out of range" error - I don't think it makes sense to try and have a coding style that forbids `some_vec[i]` because it might panic. We also have a number of `assert` statements in our real code - we could remove them, but I think they have value.

IOW, the point of this patch is about making panics less damaging, making it less necessary to try and avoid panics at all costs.

> that gets us closer to identifying what caused the invalid state and resolving the crashes.

Actually, I think the opposite it true - for every panic, you are likely to see that cause *many* poison errors - eg, once there is a single panic in logins, every single other attempt to use the logins API will cause the poison panic, which might be dozens of times before the app restarts.

The poison errors carry no information about the panic, so these errors are just noise which drown out the actual panic which set off the cascading poison errors.

bendk (approved):
Seems good to me based on the discussion we had on the call.  There's a part of me that's still a bit paranoid about panics, but I can't really think of a solid case for it.

PR #4666, opened 2021-11-19
Use with_conn helper to keep exposed functions smaller

mhammond:
Thought this helper might be useful as we wrap more functions. You could argue it's not much gain and a little harder to read, so feel free to push back, but I think it's a small improvement - it avoid every function doing the `lock().unwrap()` dance.

### DIFF
--- a/components/places/src/ffi.rs
+++ b/components/places/src/ffi.rs
@@ -6,6 +6,7 @@
 
 use crate::error::{Error, ErrorKind, InvalidPlaceInfo, PlacesError};
 use crate::msg_types;
+use crate::storage::history_metadata;
 use crate::storage::history_metadata::{
     DocumentType, HistoryHighlight, HistoryHighlightWeights, HistoryMetadata,
     HistoryMetadataObservation,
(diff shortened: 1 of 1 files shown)


bendk (approved):
I personally like the style.  I was actually thinking of proposing something similar to handle type dispatching in uniffi.

My one suggestion is maybe name it `with_conn_lock()` or something like that.  Otherwise callers might make 2 `with_conn()` calls in their function without thinking about the fact the mutex is unlocked in-between.

lougeniaC64 (approved):
Really hated all the `lock().unwrap()`s cropping up so I heartily approve. Syntactic sugar for the win! 🎉

mhammond:
> Otherwise callers might make 2 `with_conn()` calls in their function without thinking about the fact the mutex is unlocked in-between.

I'd almost see that as a feature :) How strongly do you feel about that - I'm inclined towards "short and sweet" and don't think there's that much scope for confusion here given non-trivial things will use transactions, but I'm up for strong opinions here.

Issue #4667, opened 2021-11-19
Wrong guid on login


bendk:
We're seeing errors like `Caught a panic calling rust code: "Wrong guid on login in set_local`.  See the sentry report:

https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/11982194/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2905)
┆fixVersions: Release 95
┆sprintEndDate: 2021-11-24

PR #4669, opened 2021-11-19
Fix for fetch_login_data() (#4667)

bendk:
Added a test that caused the wrong guid issue.  I'm pretty sure this is
what's causing the sentry issues -- at least some of them.

The issue was that we were calling `each_chunk_mapped()` with the
expectation that the `records` and `sync_data` vecs matched up
(`records[i].0.id == sync_data[i].guid`).  However, we only push items
to `sync_data` if `from_payload()` returns `Ok`.  If it returns an error
for any item then this fails.

Some more details:
  - [Here we're assuming that the correct SyncData item for a DB row is `sync_data[guid_idx]` 
](https://github.com/mozilla/application-services/blob/main/components/logins/src/sync/engine.rs#L206-L209)
  - [`guid_idx` comes from `records` ](https://github.com/mozilla/application-services/blob/main/components/logins/src/sync/engine.rs#L154-L161)
  - [However, if `from_payload` fails, then sync_data and records won't align correctly](https://github.com/mozilla/application-services/blob/main/components/logins/src/sync/engine.rs#L141-L148)

I think this system could use a rework, but this should fix the error
for now.

  - Note:

### DIFF
--- a/components/logins/src/sync/engine.rs
+++ b/components/logins/src/sync/engine.rs
@@ -151,9 +151,11 @@ impl LoginsSyncEngine {
         }
         scope.err_if_interrupted()?;
 
-        sql_support::each_chunk_mapped(
-            records,
-            |r| r.0.id.as_str(),
+        sql_support::each_chunk(
+            &sync_data
+                .iter()
+                .map(|s| s.guid.as_str().to_string())
+                .collect::<Vec<String>>(),
             |chunk, offset| -> Result<()> {
                 // pairs the bound parameter for the guid with an integer index.
                 let values_with_idx = sql_support::repeat_display(chunk.len(), ",", |i, f| {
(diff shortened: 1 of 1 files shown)


lougeniaC64:
> I think this system could use a rework, but this should fix the error for now.

Yes! If refactoring logins syncing so it aligns with other components isn't on our tech debt list, it should be.

mhammond:
Thanks heaps Ben - so I take it that we can expect to see 1 or more "failed to deserialize" errors every time we see one of these.

Note that I landed the parking_lot change, so your patch will need a chance (drop the `.unwrap()` after the lock)

bendk:
> Thanks heaps Ben - so I take it that we can expect to see 1 or more "failed to deserialize" errors every time we see one of these.

Yes, that's what I'd expect.

> Note that I landed the parking_lot change, so your patch will need a chance (drop the `.unwrap()` after the lock)

Just updated the code to fix this.

mhammond (approved):

PR #4677, opened 2021-11-23
Update to uniffi 0.15.2

mhammond:


tarikeshaq (approved):
LGTM!

mhammond:
There's a bug in uniffi that sync-manager is exercising - it must be a regression, but whatever :) 

https://github.com/mozilla/uniffi-rs/pull/1114

After that we will need a uniffi 0.15.2 release and this PR updated.

Issue #4678, opened 2021-11-23
Track missing scoped keys in fxa-client


bendk:
Let's try to detect one theory for mozilla-mobile/android-components#8527: The server is returning a response with missing scoped keys.  In `fxa-client`.  If we request the `https://identity.mozilla.com/apps/oldsync` scope, we should get a corresponding scoped key.  If not, then let's call `log:error!` which will result in a sentry report.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2914)
┆sprintEndDate: 2021-11-24

mhammond:
> The server is returning a response with missing scoped keys

I'm quite skeptical of that, as IIUC, desktop uses the same mechanism. (I don't object to the check though as it can't hurt)

bendk:
Yeah that's probably true, but we can hope at least and eliminate one of the possibilities. 

BTW, @tarikeshaq pointed out that the code isn't actually checking the scoped keys against the requested scopes.  It's checking them against the scopes in the response.  So if this doesn't result in any sentry hits, there's still one more possibility: we're requesting the sync scope, but the scope is missing from the response along with the scoped key.

mhammond:
well, that patch did have sentry hits!

https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/16167956 and a few others have "Sync scope granted, but keys_jwe is None". I can't see hits for the log of "Sync scope granted, but no sync scoped key ..." though. Not sure what that actually means yet though!

PR #4679, opened 2021-11-23
Log fxa responses with missing scoped keys (#4678)

bendk:

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/oauth.rs
+++ b/components/fxa-client/src/internal/oauth.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 pub mod attached_clients;
+use super::scopes;
 use super::{
     error::*,
     http_client::{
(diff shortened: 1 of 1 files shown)


components/fxa-client/src/internal/oauth.rs
```diff
@@ -327,17 +328,27 @@ impl FirefoxAccount {
         resp: OAuthTokenResponse,
         scoped_keys_flow: Option<ScopedKeysFlow>,
     ) -> Result<()> {
+        let sync_scope_requested = resp.scope.split(' ').any(|s| s == scopes::OLD_SYNC);
```
tarikeshaq, line 331: Would this be `sync_scope_granted`, I [**think** oauth allows the server to not grant a scope that was requested](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) (not sure how true this is for fxa and the sync scope thou), so I think what we would be logging is a server bug, where it tells us that it is granting us a sync scoped_key, then it is missing from the payload

Not sure how to validate the case where we request a scope, and the oauth server simply rejects and decides not to grant us the scoped key (by omitting it from the `scopes` in the response 🤷 Either way I think logging is the way to go

tarikeshaq (approved):
Makes sense to me 🚀

components/fxa-client/src/internal/oauth.rs
bendk: Good point.  It would be really nice to track which scopes we requested, but I couldn't find a simple way to do that.  If this doesn't flag issues, then that could be the next step.

I'll update the variable name too.

PR #4682, opened 2021-11-26
Add delete_bookmark_node and the Guid type

mhammond:
This is really just to expose the `Guid` type, and this is the simplest function that takes a Guid as a param.


mhammond:
superseded by @lougeniaC64's branch

PR #4684, opened 2021-11-29
Merges defaults that are under the same release

tarikeshaq:
fixes https://mozilla-hub.atlassian.net/browse/SDK-436
Adds:
- Cli to generate intermediate representation, which was really useful for testing
- Merging of the defaults at the feature level **if** they are under the same channel

Before this is ready to land:
- [ ] Add tests with a variety of defaults
- [ ] filter out the defaults with `targeting` at this time

This is still a draft to get initial feedback, it needs some more testing

### DIFF
--- a/components/support/nimbus-fml/src/cli.yaml
+++ b/components/support/nimbus-fml/src/cli.yaml
@@ -49,3 +49,5 @@ subcommands:
                 global: true
     - experimenter:
         about: Generate code for Experimenter
+    - intermediate-repr:
+        about: Generate the intermediate representation json
--- a/components/support/nimbus-fml/src/main.rs
+++ b/components/support/nimbus-fml/src/main.rs
@@ -65,6 +65,14 @@ fn main() -> Result<()> {
                 load_from_ir: matches.is_present("ir"),
             },
         )?,
+        ("intermediate-repr", _) => workflows::generate_ir(
+            config,
+            GenerateIRCmd {
+                manifest: file_path("INPUT", &matches, &cwd)?,
+                output: file_path("output", &matches, &cwd)?,
+                load_from_ir: matches.is_present("ir"),
+            },
+        )?,
         (word, _) => unimplemented!("Command {} not implemented", word),
     };
 
@@ -97,6 +105,12 @@ pub struct GenerateStructCmd {
 }
 
 pub struct GenerateExperimenterManifestCmd {
+    manifest: PathBuf,
+    output: PathBuf,
+    load_from_ir: bool,
+}
+
+pub struct GenerateIRCmd {
     manifest: PathBuf,
     output: PathBuf,
     load_from_ir: bool,
--- a/components/support/nimbus-fml/src/parser.rs
+++ b/components/support/nimbus-fml/src/parser.rs
@@ -69,6 +69,69 @@ pub(crate) struct ManifestFrontEnd {
     channels: Vec<String>,
 }
 
+trait MergableDefault: Sized {
+    fn merge(self, new: Self) -> Self;
+}
+
+impl MergableDefault for serde_json::Map<String, serde_json::Value> {
+    fn merge(self, new: Self) -> Self {
+        let mut merged = serde_json::Map::new();
+        for (key, val) in self {
+            merged.insert(key, val);
+        }
+        for (key, val) in new {
+            if merged.contains_key(&key) {
+                let old_val = merged.get(&key).unwrap().clone();
+                merged.insert(key, old_val.merge(val));
+            } else {
+                merged.insert(key, val);
+            }
+        }
+        merged
+    }
+}
+
+impl MergableDefault for serde_json::Value {
+    fn merge(self, new: Self) -> Self {
+        use serde_json::Value::{Array, Number, Object, String};
+        match (self, new) {
+            (Number(old), Number(new)) => Number(old.merge(new)),
+            (String(old), String(new)) => String(old.merge(new)),
+            (Array(old), Array(new)) => Array(old.merge(new)),
+            (Object(old), Object(new)) => Object(old.merge(new)),
+            (_, new) => new,
+        }
+    }
+}
+
+impl MergableDefault for serde_json::Number {
+    fn merge(self, new: Self) -> Self {
+        new
+    }
+}
+
+impl MergableDefault for Vec<serde_json::Value> {
+    fn merge(self, new: Self) -> Self {
+        new
+    }
+}
+
+impl<V: MergableDefault> MergableDefault for Option<V> {
+    fn merge(self, new: Self) -> Self {
+        match (self, new) {
+            (None, new) => new,
+            (old, None) => old,
+            (Some(old), Some(new)) => Some(old.merge(new)),
+        }
+    }
+}
+
+impl MergableDefault for String {
+    fn merge(self, new: Self) -> Self {
+        new
+    }
+}
+
 fn parse_typeref_string(input: String) -> Result<(String, Option<String>), FMLError> {
     // Split the string into the TypeRef and the name
     let mut object_type_iter = input.split(&['<', '>'][..]);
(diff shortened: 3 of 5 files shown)


components/support/nimbus-fml/src/parser.rs
```diff
@@ -69,6 +69,69 @@ pub(crate) struct ManifestFrontEnd {
...
+
+impl MergableDefault for serde_json::Value {
+    fn merge(self, new: Self) -> Self {
+        use serde_json::Value::{Array, Number, Object, String};
+        match (self, new) {
+            (Number(old), Number(new)) => Number(old.merge(new)),
+            (String(old), String(new)) => String(old.merge(new)),
+            (Array(old), Array(new)) => Array(old.merge(new)),
+            (Object(old), Object(new)) => Object(old.merge(new)),
+            (_, new) => new,
```
tarikeshaq, line 102: note to myself that I need to think more about this if there are any more interesting cases

components/support/nimbus-fml/src/parser.rs
```diff
@@ -69,6 +69,69 @@ pub(crate) struct ManifestFrontEnd {
...
+    }
+}
+
+impl MergableDefault for serde_json::Number {
+    fn merge(self, new: Self) -> Self {
+        new
+    }
+}
+
+impl MergableDefault for Vec<serde_json::Value> {
```
tarikeshaq, line 113: this is unfortunate... I would like to be able to merge vectors a little better, but there I could think of an easy way to detect that items in the vector should be merged, so I just return the new vector

components/support/nimbus-fml/src/parser.rs
```diff
@@ -69,6 +69,69 @@ pub(crate) struct ManifestFrontEnd {
...
+    fn merge(self, new: Self) -> Self {
+        new
+    }
+}
+
+impl<V: MergableDefault> MergableDefault for Option<V> {
```
tarikeshaq, line 119: I can probably get rid of this, it's not used anywhere

components/support/nimbus-fml/src/parser.rs
```diff
@@ -278,6 +343,34 @@ impl Parser {
             feature_defs: self.features.clone(),
         })
     }
+
+    fn merge_defaults(defaults: serde_json::Value) -> Result<serde_json::Value, FMLError> {
+        let mut channel_map = HashMap::new();
+        let defaults = serde_json::from_value::<Vec<Defaults>>(defaults)?;
+        for default in defaults {
+            if channel_map.contains_key(&default.channel) {
```
tarikeshaq, line 351: should add a field `targeting` to the `Defaults` struct and add a `&& default.targeting.is_none()` here for now

tarikeshaq (commented):
quick self review

jhugman (commented):
🚨 Drive by review 🚨 

I've added some comments, but here's a bigger one not attached to any particular line of code.

I think the algorithm we're coming to is usefully described with an example.

For a given feature defined in the YAML:

```yaml
features:
    my-feature:
        variables:
            my-int:
                default: 42
            my-string:
                default: Field default
            my-button:
                default:
                    label: OK
                    color: blue
            my-map:
                default:
                    variant-one: 1
                    variant-two: 2
        defaults:
        - channel: release
          value: {
                "my-button": {
                    "color": "green"
                }
          }
        - channel: release
          value: {
                "my-string": "Release default"
          }
        - channel: release
          value: {
                "my-map": {
                    "variant-one": 42
                }
          }
```

### Step 1: extract the "field" defaults

…into a JSON representation. These are the default derived from each individual variable fields.

This _should_ be a complete set of default values for every variable of feature, but may not be.

```json
{
    "my-int": 42,
    "my-string": "Field default",
    "my-button": {
        "label": "OK",
        "color": "blue"
    },
    "my-map": {
        "variant-one": 1,
        "variant-two": 2,
    }
}
```

### Step 2: extract the "feature" defaults for the given channel

```json
{
    "my-int": 42,
    "my-string": "Release default",
    "my-button": {
        "color": "green"
    },
    "my-map": {
        "variant-one": 42
    }
}
```

Each feature has zero or more default blocks,

Each default block can have a channel and/or targeting string, and a value for a object.

For each channel, the default blocks' `value` JSON objects are merged. Blocks at the end of the list should merge over the blocks before it. If a channel is not specified, then it should be used for all channels.

In this example, we're interested in the `release` channel.

### Step 3: merge the JSON objects

…for the field defaults and the feature defaults.

The feature defaults should merge over the top of the field defaults to make a new complete set of defaults.

```json
{
    "my-int": 42,
    "my-string": "Release default",
    "my-button": {
        "label": "OK",
        "color": "green"
    },
    "my-map": {
        "variant-one": 42,
        "variant-two": 2,
    }
}
```

### Step 4

Use the new set of defaults for the entire feature to overwrite the "field" defaults.

```yaml
features:
    my-feature:
        variables:
            my-int:
                default: 42
            my-string:
                default: Release default
            my-button:
                default:
                    label: OK
                    color: green
            my-map:
                default:
                    variant-one: 42
                    variant-two: 2
```

Once this point is reached, we can start running lots of validations over the defaults, e.g. types matching, null checking, enum variant exhaustion in map keys, etc.

components/support/nimbus-fml/src/parser.rs
```diff
@@ -90,6 +214,184 @@ fn parse_typeref_string(input: String) -> Result<(String, Option<String>), FMLEr
...
+                .map(|(channel, old_default)| {
+                    (channel, merge_two_defaults(&old_default, &default.value))
+                })
+                .collect();
+        }
+    }
+    Ok(channel_map)
+}
+
+/// Transforms a feature definition with unmerged defaults into a feature
```
tarikeshaq, line 261: Not sure how we feel about doc comments in this project.. happy to get rid of it if we feel it clutters the code and doesn't add much value! (wanted to be explicit that I am totally happy to throw the comments away and I'm a bit conflicted about the "when" of doc comments and how we want to use them)

components/support/nimbus-fml/src/cli.yaml
```diff
@@ -27,15 +27,18 @@ args:
         long: output
         value_name: FILE
         global: true
+    - channel:
```
tarikeshaq, line 30: Happy to take suggestions on the best place to put this argument! I put it here because it seem like it's not super specific to android (i.e ios has beta and release) and I needed the argument when I was running the `intermediate-repr` command to validate that all is good

components/support/nimbus-fml/src/main.rs
```diff
@@ -47,6 +47,10 @@ fn main() -> Result<()> {
                         .value_of("package")
                         .map(str::to_string)
                         .or(config.package_name),
+                    channel: matches
```
tarikeshaq, line 50: Hmm should the channel be in the `config` or in the `Generate**Cmd`... not so sure

components/support/nimbus-fml/src/parser.rs
```diff
@@ -69,6 +70,130 @@ pub(crate) struct ManifestFrontEnd {
     channels: Vec<String>,
 }
 
+impl ManifestFrontEnd {
```
tarikeshaq, line 73: I refactored this file a bit to unclutter the `Parser::new` function, happy to take suggestions/feedback!

jhugman (approved):
I've left a couple of comments, which you should address, but nothing that should hold up for another review cycle.

components/support/nimbus-fml/fixtures/fe/default_merging.yaml
```diff
@@ -0,0 +1,89 @@
...
+      background-color:
+        description: This is the background color
+        type: String
+        default: white
+      player-mapping:
+        description: This is the map of the player type to a button
+        type: Map<PlayerProfile, Button>
+        default: {
+          "child": {},
+          "adult": {}
```
jhugman, line 31: A good test would test that

 * A scalar feature variable could be replaced.
 * An object feature variable with a scalar field being replace
 * A map feature variable with a scalar value being replace.
 * An object feature variable with a map with a scalar value being replaced.
 
etc. etc. 

Think of test cases to break your code.

components/support/nimbus-fml/src/parser.rs
```diff
@@ -385,6 +614,91 @@ mod unit_tests {
...
+            positive_button
+                .default
+                .get("alt-text")
+                .unwrap()
+                .as_str()
+                .unwrap(),
+            "Go Ahead!"
+        );
+        Ok(())
+    }
```
jhugman, line 700: The alogrithm you've sketched out has a number of parts/functions

For each of those parts you should write a test in which you try to explore the limits of what the function is meant to do.

When you come to put the well-tested parts together, you'll have more confidence in the pieces of the algorithm, and you'll be happy with a smoke test to test they fit together.

jhugman (commented):
I've added comments about testing in this subsequent review.

PR #4689, opened 2021-11-30
Uniffi VisitObservation and apply_observation

skhamis:
Converting VisitObservation/VisitTransition types and associated functions to be uniffi-ed. 

One interesting thing to note is that `PlacesConnection.kt` defines a type `VisitType` that parallels the rust `VisitTransition` with the exception of  `UPDATE_PLACE(-1)`. I would like to fully remove the enum so we only use the uniffi-ed version but will need to take the reasoning/use of that into consideration.

  - Note:


mhammond:
PlacesConnection.kt also needs updating to call the uniffi version of the function (and history functions don't have swift bindings currently, so best I can tell, no changes are needed there)

components/places/android/src/test/java/mozilla/appservices/places/PlacesConnectionTest.kt
```diff
@@ -382,15 +384,15 @@ class PlacesConnectionTest {
         assert(!PlacesManagerMetrics.writeQueryCount.testHasValue())
         assert(!PlacesManagerMetrics.writeQueryErrorCount["url_parse_failed"].testHasValue())
 
-        db.noteObservation(VisitObservation(url = "https://www.example.com/2a", visitType = VisitType.REDIRECT_TEMPORARY, at = 130000))
-        db.noteObservation(VisitObservation(url = "https://www.example.com/2b", visitType = VisitType.LINK, at = 150000))
-        db.noteObservation(VisitObservation(url = "https://www.example.com/3", visitType = VisitType.LINK, at = 200000))
+        db.noteObservation(VisitObservation(url = "https://www.example.com/2a", visitType = VisitTransition.REDIRECT_TEMPORARY, at = 130000u))
```
mhammond, line 387: while the move to unsigned looks correct, are we sure we want to take that on in this PR? It seems like this might have broad consequences in the a-c and fenix code? It might be possible to declare it in the udl as a signed int, and have `UniffiCustomTypeWrapper` do the conversion, and then maybe another ticket to fix it post-landing?

mhammond (commented):
This looks good to me although I'm worried by the timestamp signed/unsigned change. This probably makes a great case for the a-c PR, so we can see how badly that change leaks into the consumers? I'm r+ though if you can convince me there's no problem there.

skhamis:
Even with the panic on bad URL test fail I think overall this PR is in a good shape with the companion PR up on my a-c fork https://github.com/skhamis/android-components/pull/2 the only "yucky" thing is the whole VisitType/VisitTransition but I think we should just kick that further than the pipeline and deal with that post-conversion of the APIs. Looking for full reviews to get this in! Note: This is still going to the `places-uniffication` branch so there is lower risk here.

components/places/ffi/src/lib.rs
```diff
@@ -216,131 +199,6 @@ pub extern "C" fn places_match_url(
...
-        let output = std::slice::from_raw_parts_mut(byte_buffer, byte_buffer_len as usize);
-        let urls = url_ptrs
-            .iter()
-            .enumerate()
-            .filter_map(|(idx, &p)| {
-                let s = FfiStr::from_raw(p).as_str();
-                url::Url::parse(s).ok().map(|url| (idx, url))
-            })
-            .collect::<Vec<_>>();
-        storage::history::get_visited_into(conn, &urls, output)?;
```
mhammond, line 251: Note that get_visited_into has some magic to allow invalid URLs to be "skipped"

components/places/src/places.udl
```diff
@@ -93,6 +143,38 @@ dictionary HistoryHighlight {
     string? preview_image_url;
 };
 
+dictionary HistoryVisitInfo {
+    string url;
```
mhammond, line 147: Can we use `Url`s here and below? APIs like `getVisited()` which intentionally ignore invalid URLs can't, but I don't think most of these are in this situation.

components/places/src/types.rs
```diff
@@ -102,6 +101,22 @@ impl<'de> serde::Deserialize<'de> for VisitTransition {
     }
 }
 
+#[derive(Clone, PartialEq)]
```
mhammond, line 104: Apart from using a `Url` where possible, I think I'd rather see these new types in ffi.rs - IIUC, they exist only for the ffi and aren't used by the actual implementation, but putting them here makes them look like "core" structs.

mhammond (commented):
Sorry I didn't get very far here yet.

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+                conn,
+                types::Timestamp(start.max(0) as u64),
+                types::Timestamp(end.max(0) as u64),
+            )
+        })
+    }
+
+    fn delete_visit(&self, url: Url, timestamp: i64) -> Result<()> {
+        self.with_conn(|conn| {
+            match Url::parse(url.as_str()) {
```
mhammond, line 256: This is parsing the url twice - but given the semantics, the signature of this actually needs to be `String` (the failure to parse is consumed)

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
             )
         })
     }
+
+    /// Add an observation to the database.
+    fn apply_observation(&self, visit: VisitObservation) -> Result<()> {
+        match self.with_conn(|conn| history::apply_observation(conn, visit)) {
```
mhammond, line 150: not too bothered, but I think this could be written as:
```
self.with_conn(|conn| history::apply_observation(conn, visit))?;
Ok(())
```

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+            Err(e) => Err(e),
+        }
+    }
+
+    fn get_visited_urls_in_range(
+        &self,
+        start: i64,
+        end: i64,
+        include_remote: bool,
+    ) -> Result<Vec<String>> {
```
mhammond, line 161: I think you could make this result a vec of Urls? (in the IDL too)

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+                types::Timestamp(start.max(0) as u64),
+                types::Timestamp(end.max(0) as u64),
+                include_remote,
+            )
+        })
+    }
+
+    fn get_visit_infos(
+        &self,
+        start_date: i64,
```
mhammond, line 175: should these be your `Timestamp` type? (Same in the function above too) It abstracts away some of the signed/unsigned ugliness into one nice spot!

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+            .enumerate()
+            .filter_map(|(idx, s)| Url::parse(&s).ok().map(|url| (idx, url)))
+            .collect::<Vec<_>>();
+        self.with_conn(|conn| history::get_visited_into(conn, &url_idxs, &mut result))?;
+        Ok(result)
+    }
+
+    fn delete_visits_for(&self, url: Url) -> Result<()> {
+        self.with_conn(|conn| {
+            let guid = match Url::parse(url.as_str()) {
```
mhammond, line 230: ditto here re the double parse of the URL, and sadly this probably needs to be a string too, because the failure to parse is consumed.

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+                }
+            };
+            if let Some(guid) = guid {
+                history::delete_visits_for(conn, &guid)?;
+            }
+            Ok(())
+        })
+    }
+
+    fn delete_visits_between(&self, start: i64, end: i64) -> Result<()> {
```
mhammond, line 244: should be `Timestamp`s?

components/places/src/ffi.rs
```diff
@@ -113,6 +144,151 @@ impl PlacesConnection {
...
+            Ok(())
+        })
+    }
+}
+
+#[derive(Clone, PartialEq)]
+pub struct HistoryVisitInfo {
+    pub url: String,
+    pub title: Option<String>,
+    pub timestamp: i64,
```
mhammond, line 282: and a `Timestamp` here?

components/places/src/places.udl
```diff
@@ -4,6 +4,11 @@
 
 [Wrapped]
 typedef string Url;
+[Wrapped]
+typedef i64 Timestamp;
```
mhammond, line 8: seeing we are adding this type, we should try and use it for all timestamps!

mhammond (changes requested):
Sorry, but a few changes are needed - feel free to ping me if you want to chat about anything.

components/places/src/places.udl
```diff
@@ -93,6 +143,38 @@ dictionary HistoryHighlight {
     string? preview_image_url;
 };
 
+dictionary HistoryVisitInfo {
+    string url;
+    string? title;
+    Timestamp timestamp;
+    i32 visit_type;
```
mhammond, line 150: sorry I didn't pick this up before - this should be an enum - but I think lougenia's patch is adding to the the .udl - so maybe add an `XXX` comment to change it?

components/places/src/ffi.rs
```diff
@@ -113,6 +150,127 @@ impl PlacesConnection {
...
+                }
+            };
+            Ok(())
+        })
+    }
+}
+
+#[derive(Clone, PartialEq)]
+pub struct HistoryVisitInfo {
+    pub url: String,
```
mhammond, line 262: ditto here - a couple of things that could be Urls? I wonder what happens if you just change `visit_type` here to be of type `VisitType` - maybe the UDL will magically just work?

components/places/src/places.udl
```diff
@@ -93,6 +143,38 @@ dictionary HistoryHighlight {
...
+ */
+dictionary VisitObservation {
+    string url;
+    string? title = null;
+    VisitTransition? visit_type;
+    boolean? is_error = null;
+    boolean? is_redirect_source = null;
+    boolean? is_permanent_redirect_source = null;
+    Timestamp? at = null;
+    string? referrer = null;
```
mhammond, line 173: `referrer` is a url too!

mhammond (approved):
On slack, we agreed to leave the url changes to a followup - so please open a new ticket that covers this and the timestmap issues you've identified and we can land this - nice work :)

PR #4701, opened 2021-12-02
Compile nimbus-fml for use in a gradle plugin

jhugman:

  - Note:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,21 @@ commands:
             ./rustup-init -y --no-modify-path
             rm rustup-init
             echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $BASH_ENV
+  install-ghr:
+    steps:
+      - run:
+          name: Install GHR
+          command: |
+            GHR=ghr_v0.12.0_darwin_amd64
+            GHR_SHA256=c868ef9fc5dd8c8a397b74d84051d83693c42dd59041cb17b66f90f563477249
+            curl -sfSL --retry 5 --retry-delay 10 -O "https://github.com/tcnksm/ghr/releases/download/v0.12.0/${GHR}.zip"
+            echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
+            unzip "${GHR}.zip"
+
+            echo "export PATH=$PATH:$(pwd)/$GHR" >> $BASH_ENV
+      # Run in a different shell to source $BASH_ENV
+      - run: |
+          ghr --version
   build-libs:
     parameters:
       platform:
@@ -178,7 +193,7 @@ commands:
             #                       note that both the Nimbus-SDK and glean use chrono, so if we would like to move away from it, both projects
             #                       need to do that before we can remove the ignores (assuming `chrono` doesn't release a fixed version)
             #  * RUSTSEC-2018-0006: Uncontrolled recursion in `yaml-rust`, which is included by `clap` v2. `clap` itself already updated to a safe
-            #                       version of `yaml-rust`, which will be released in `v3` and additionally, 
+            #                       version of `yaml-rust`, which will be released in `v3` and additionally,
             #                       reading https://github.com/rustsec/advisory-db/issues/288, this is a false
             #                       positive for clap and based on our dependency tree, we only use `yaml-rust` in `clap`.
             cargo audit --ignore RUSTSEC-2021-0019 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2018-0006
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -451,30 +466,47 @@ jobs:
...
-            GHR_SHA256=c868ef9fc5dd8c8a397b74d84051d83693c42dd59041cb17b66f90f563477249
-            curl -sfSL --retry 5 --retry-delay 10 -O "https://github.com/tcnksm/ghr/releases/download/v0.12.0/${GHR}.zip"
-            echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
-            unzip "${GHR}.zip"
-            ./${GHR}/ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
+            ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
+  build-nimbus-fml:
+    executor: macos
+    steps:
+      - full-checkout
```
tarikeshaq, line 488: Definitely for a follow up... Probably not be worth it unless the fml takes a long time to compile, which it doesn't now - 5 mins is not bad at all:

We might be able to cut down some CI time by using `sccache`:
```yaml
- restore_cache:
    name: Restore sccache cache
    key: fml-sccache-cache-macos-{{ arch }}
- full-checkout
- install rust
- setup-rust-target-version
- setup-sccache

.. rest of commands
- save_cache:
    name: Save sccache cache
    key: fml-sccache-cache-macos-{{ arch }}-{{ epoch }}
    paths:
      - "~/Library/Caches/Mozilla.sccache"
- sccache-stats-and-logs
```

.circleci/config.yml
```diff
@@ -451,30 +466,47 @@ jobs:
...
+      - install-rust
```
tarikeshaq, line 489: we should probablly add a 
```yaml
- setup-rust-target-version
```
after the `install-rust` to make sure we are running the latest rust we support in app services for this

.circleci/config.yml
```diff
@@ -451,30 +466,47 @@ jobs:
...
+          path: nimbus-fml.zip
+          destination: nimbus-fml.zip
+  release-nimbus-fml:
+    executor: macos
+    steps:
+      - install-ghr
+      - run:
+          name: Release nimbus-fml archive on Github
+          command: |
+            ghr -replace "${CIRCLE_TAG}" nimbus-fml.zip
```
tarikeshaq, line 509: should we release a sha256 hash of the zip as a checksum something similar as well? so the consumer can verify against it? (I am guessing the android build process will pull this directly from over the network?)

.circleci/config.yml
```diff
@@ -451,30 +466,47 @@ jobs:
...
+          command: |
+            if [ -z "${CIRCLE_TAG}" ]; then
+              # On non-release builds do not update brew (takes ages).
+              echo 'export HOMEBREW_NO_AUTO_UPDATE=1' >> $BASH_ENV
+            fi
+            ./components/support/nimbus-fml/scripts/build-dist.sh
+      - store_artifacts:
+          name: Store nimbus-fml build archive in workspace
+          path: nimbus-fml.zip
+          destination: nimbus-fml.zip
```
tarikeshaq, line 501: if we end up releasing checksum, we can also store probably it as an artifact (or just print it, as long as someone debugging can easily get the checksum to run android against a debug build of the fml)

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,13 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+### ✨✨ What's New ✨✨
+
+#### ⛅️🔭🔬 Nimbus
+
+- Initial release of the Nimbus Feature Manifest Language tool (`nimbus-fml`).
```
tarikeshaq, line 26: 🎉 🎉 🚀

components/support/nimbus-fml/.cargo/config
```diff
@@ -0,0 +1,2 @@
+[target.x86_64-unknown-linux-musl]
```
tarikeshaq, line 1: can we add a comment here whey we need this 😅

components/support/nimbus-fml/scripts/build-dist.sh
```diff
@@ -0,0 +1,68 @@
...
+fml_dir=$root_dir/components/support/nimbus-fml
+target_dir=$root_dir/target
+filename=$(basename "$fml_dir")
+
+# But we'll dump the zip file wherever we run the script from.
+dist_file=${filename}.zip
+
+prompt='$'
+echo "## Installing tools for cross compiling"
+install_musl_cross="brew install filosottile/musl-cross/musl-cross"
```
tarikeshaq, line 29: Can you write a comment on why we need this?

I see `x86_64-unknown-linux-musl` in our targets, so I imagine the target requires some native libraries that this provides? And for my own curiosity, why do we need to target `x86_64-unknown-linux-musl`, is it for devs running Linux machines?

.circleci/config.yml
```diff
@@ -451,30 +466,47 @@ jobs:
...
           command: |
-            GHR=ghr_v0.12.0_darwin_amd64
-            GHR_SHA256=c868ef9fc5dd8c8a397b74d84051d83693c42dd59041cb17b66f90f563477249
-            curl -sfSL --retry 5 --retry-delay 10 -O "https://github.com/tcnksm/ghr/releases/download/v0.12.0/${GHR}.zip"
-            echo "${GHR_SHA256} *${GHR}.zip" | shasum -a 256 -c -
-            unzip "${GHR}.zip"
-            ./${GHR}/ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
+            ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
+  build-nimbus-fml:
+    executor: macos
```
tarikeshaq, line 486: definitely not urgent nit:

I see that we use homebrew, which makes sense why we use a macos runner here! But if we can get rid of that, we can run this on a docker runner, which I think is cheaper 😄 That said, I know we'll start shipping this to iOS devs as well... and I'm not sure if once we do that we'll be forced to use macos anyways 🤷

.circleci/config.yml
```diff
@@ -81,6 +81,21 @@ commands:
             ./rustup-init -y --no-modify-path
             rm rustup-init
             echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $BASH_ENV
+  install-ghr:
```
tarikeshaq, line 84: Love the refactoring here 😍

tarikeshaq (approved):
Tried it locally and it works! 🎉 

Dropped a few thoughts, nothing blocking - if anything just a few follow-ups that would be nice

components/support/nimbus-fml/scripts/build-dist.sh
jhugman: We have devs running linux, and Android CI builds are (IIRC) all done on linux.

.circleci/config.yml
jhugman: That was my thought, yes. Given we're already using homebrew and GHR from a macos runner, and I'm running on macos; I don't know what the M1 cross compilation story is like from linux, etc I went with the simplest option.

components/support/nimbus-fml/scripts/build-dist.sh
tarikeshaq: Aaah the CI builds of course 💯

.circleci/config.yml
jhugman: Ok, I think I'll leave this for a followup.

components/support/nimbus-fml/scripts/build-dist.sh
jhugman: I've added a comment to explain all.

Issue #4620, opened 2021-11-01
Bookmarks - support getting an arbitrary number of nested children information


Mugurell:
Saw in a recent PR for Fenix - https://github.com/mozilla-mobile/fenix/pull/21212 the need to know how many children are in current node's childrens.
Based on this we would know whether to offer a menu option to open all bookmarks from a child folder in new tabs or don't show such an option when a child folder does is empty.

`bookmarks_get_tree` allows getting info about all nested children but for our usecase this is a bit wasteful.

Tried adding a method which takes a byte argument allowing to query up to 127 levels of nested children (Byte.MAX_VALUE) but then saw that that number [gets mapped to a boolean in `lib.rs`](https://github.com/mozilla/application-services/blob/6d33f26787cccea111fd2d019f4a0ab5ea36977d/components/places/ffi/src/lib.rs#L584) so there isn't a straightforward way of getting the needed info.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2869)

mhammond:
To make sure I understand, you'd like that param to actually be treated as an integer "depth" value rather than a boolean, right? (FWIW, it's not a boolean due to some strange JNI issue that forces us to use ints in place of a bool). If so, I think that will be quite easy - [the implementation already supports that](https://github.com/mozilla/application-services/blob/a7bcd2c09eecc60469c35b6332e5bde9834af30e/components/places/src/storage/bookmarks/public_node.rs#L96) - so I think we'd just change the param to be a real int, and ensure the existing callers all pass 0 or 1 instead of False/True, and then you could specify 2, which I think would allow that Fenix issue to be implemented. Does that sound correct?

Mugurell:
Yes, that's exactly what I think we need.

mhammond:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1865194

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #4709, opened 2021-12-06
Uniffied public node type and functions

lougeniaC64:

  - Note:


components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -456,50 +457,23 @@ open class PlacesReaderConnection internal constructor(connHandle: Long, conn: U
...
-            try {
-                return rustBuf.asCodedInputStream()?.let { stream ->
-                    unpackProtobuf(MsgTypes.BookmarkNode.parseFrom(stream))
-                }
-            } finally {
-                LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuf)
-            }
+    override fun getBookmark(guid: String): PublicNode? {
+        return readQueryCounters.measure {
+            this.conn.bookmarksGetByGuid(guid, 0.toByte().toUByte())
```
mhammond, line 462: I think we can pass `false` here (uniffi itself hides the strangeness with bools).

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -456,50 +457,23 @@ open class PlacesReaderConnection internal constructor(connHandle: Long, conn: U
...
-            return rustBuf.asCodedInputStream()?.let { stream ->
-                unpackProtobuf(MsgTypes.BookmarkNode.parseFrom(stream))
-            }
-        } finally {
-            LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuf)
+    override fun getBookmarksTree(rootGUID: String, recursive: Boolean): PublicNode? {
+        if (recursive) {
+            return this.conn.bookmarksGetTree(rootGUID)
+        } else {
+            return this.conn.bookmarksGetByGuid(rootGUID, 1.toByte().toUByte())
```
mhammond, line 470: ditto here for `true`

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+#[derive(Clone, PartialEq)]
+pub struct BookmarkNode {
+    ///*
+    /// The type of this bookmark, a `BookmarkType` (from `types.rs`).
+    ///
+    /// This impacts which fields may be present.
+    ///
+    /// It's illegal to attempt to change this when updating a bookmark.
+    ///
+    /// Note: this probably should be an `enum`, but prost seems to get upset
```
mhammond, line 96: This comment is now stale - and it would indeed seem ideal if we can get an enum in place for this. The existing `BookmarkType` would be ideal, but the fact it's declared with a `val` may or may not make this tricky. Either way, the comment should change in some way :)

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    /// - Required for inserts.
+    /// - Not provided for updates.
+    pub node_type: Option<i32>,
+    ///*
+    /// The bookmarks guid.
+    ///
+    /// - Always returned on reads.
+    /// - Not allowed for inserts.
+    /// - Required for updates (specifies which record is being changed)
+    pub guid: Option<String>,
```
mhammond, line 112: Can we use `Guid` here (and below for the parent)?

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    ///*
+    /// Creation time, in milliseconds since the unix epoch.
+    ///
+    /// May not be a local timestamp, and may shift if new devices are able to
+    /// provide an earlier (but still valid) timestamp.
+    ///
+    /// - Always returned on reads.
+    /// - Ignored for insertion and update.
+    pub date_added: Option<i64>,
```
mhammond, line 121: and maybe Timestamp's here and below?

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    ///
+    /// - Returned on reads if it exists.
+    /// - Required when inserting folders.
+    pub title: Option<String>,
+    ///*
+    /// Bookmark URL. Only allowed/present for type = `BookmarkType::Bookmark`.
+    ///
+    /// - Always returned on reads (for `BookmarkType::Bookmark`).
+    /// - Required when inserting a new bookmark.
+    pub url: Option<String>,
```
mhammond, line 167: and `Url` here and below in `child_guids`

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    ///
+    /// For performance reasons, this only is provided if it's requested.
+    pub child_nodes: Vec<BookmarkNode>,
+    ///*
+    /// Returned by reads, and used to distinguish between the cases of
+    /// "empty child_nodes because the API doesn't return children" and
+    /// "empty child_nodes because this folder has no children (but
+    /// we'd populate them if it had them)".
+    ///
+    /// Only required because you can't have `optional repeated`.
```
mhammond, line 189: I think in this world we *could* have child_nodes as `Option<Vec<BookmarkNode>>` although we might decide that's beyond the scope of this PR - so maybe just update the comment to indicate that "when we used protobufs..."

components/places/src/ffi.rs
```diff
@@ -113,6 +267,38 @@ impl PlacesConnection {
             )
         })
     }
+
+    fn bookmarks_get_tree(&self, item_guid: &Guid) -> Result<Option<PublicNode>> {
+        self.with_conn(|conn| {
+            let root_id = Guid::from(item_guid.as_str());
```
mhammond, line 273: This probably isn't necessary - just use `item_guid` directly (and ditto the next function)

components/places/src/ffi.rs
```diff
@@ -113,6 +267,38 @@ impl PlacesConnection {
...
+            bookmarks::public_node::fetch_public_tree(conn, &root_id)
+        })
+    }
+
+    fn bookmarks_get_by_guid(
+        &self,
+        guid: &Guid,
+        get_direct_children: u8,
```
mhammond, line 281: This should change to a bool (and in the .udl too)

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    }
+}
+
+///*
+/// A bookmark node.
+///
+/// We use a single message type for bookmarks. It covers insertion, deletion,
+/// and update, and represents all three bookmark types.
+///
+/// This simplifies the FFI by reducing the number of types that must go across
```
mhammond, line 81: I think it's worth updating this comment to note something like "when we had hand-written FFIs"

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+/// it, and retuces boilderplate, but removes some static-ish guarantees we
+/// might have otherwise.
+///
+/// Note that these docs comments are internal, and don't necessarily impact the actual
+/// API we expose to Kotlin/Swift (this is particularly true around reads).
+#[derive(Clone, PartialEq)]
+pub struct BookmarkNode {
```
mhammond, line 88: I'm a little confused by the relationship between `BookmarkNode` and `PublicNode` but I suspect that will get clearer soon!

components/places/src/ffi.rs
```diff
@@ -43,6 +48,155 @@ impl UniffiCustomTypeWrapper for Url {
...
+    /// "empty child_nodes because this folder has no children (but
+    /// we'd populate them if it had them)".
+    ///
+    /// Only required because you can't have `optional repeated`.
+    ///
+    /// Leaving this out is equivalent to false.
+    pub have_child_nodes: Option<bool>,
+}
+///* An array of bookmark nodes, since we can't represent that directly
+#[derive(Clone, PartialEq)]
```
mhammond, line 195: This is no longer true - we should see if we can kill `BookmarkNodeList` as I think that will be transparent to consumers?

components/places/src/places.udl
```diff
@@ -45,6 +51,21 @@ interface PlacesConnection {
 
     [Throws=PlacesError]
     void metadata_delete_older_than(i64 older_than);
+
+    [Throws=PlacesError]
+    PublicNode? bookmarks_get_tree([ByRef] Guid item_guid);
+
+    [Throws=PlacesError]
+    PublicNode? bookmarks_get_by_guid([ByRef] Guid guid, u8 get_direct_children);
```
mhammond, line 59: as mentioned above, make this bool. Alternatively, lean in to it being an integer on the path to #4620 - rename the param to something like "fetchDepth", and for now, just `log::warn` if the param is > 1? Up to you :)

mhammond (commented):
This is looking great - a quick drive-by!

components/places/src/ffi.rs
lougeniaC64: Actually, this is a good catch. Initially I moved `BookmarkNode` here before I decided I was going to expose `PublicNode`. `BookmarkNode` isn't being used here so I'm going to remove it.

components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
```diff
@@ -4,7 +4,8 @@
 
 package mozilla.appservices.places
 
-import java.lang.RuntimeException
+import mozilla.appservices.places.uniffi.ReadonlyBookmarkItem
```
mhammond, line 7: Not a huge fan of `ReadonlyBookmarkItem` TBH - maybe just `BookmarkItem`? The fact we have that `Updatable` variant seems OK and is enough to differentiate IMO. This is hugely subjective though, so feel free to stick with what you prefer. :)

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -352,86 +341,39 @@ open class PlacesReaderConnection internal constructor(connHandle: Long, conn: U
...
-                LibPlacesFFI.INSTANCE.bookmarks_get_by_guid(this.handle.get(), guid, 0.toByte(), err)
-            }
-            try {
-                return rustBuf.asCodedInputStream()?.let { stream ->
-                    unpackProtobuf(MsgTypes.BookmarkNode.parseFrom(stream))
-                }
-            } finally {
-                LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuf)
-            }
+    override fun getBookmark(guid: String): ReadonlyBookmarkItem? {
```
mhammond, line 344: We have the alias setup, so I think this should read `guid: Guid`

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -352,86 +341,39 @@ open class PlacesReaderConnection internal constructor(connHandle: Long, conn: U
...
-                LibPlacesFFI.INSTANCE.bookmarks_get_by_guid(this.handle.get(), rootGUID, 1.toByte(), err)
-            }
-        }
-        try {
-            return rustBuf.asCodedInputStream()?.let { stream ->
-                unpackProtobuf(MsgTypes.BookmarkNode.parseFrom(stream))
-            }
-        } finally {
-            LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuf)
+    override fun getBookmarksTree(rootGUID: String, recursive: Boolean): ReadonlyBookmarkItem? {
```
mhammond, line 350: ditto here re guid - there are a few other below too

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -352,86 +341,39 @@ open class PlacesReaderConnection internal constructor(connHandle: Long, conn: U
...
-                LibPlacesFFI.INSTANCE.bookmarks_get_all_with_url(this.handle.get(), url, err)
-            }
-
-            try {
-                val message = MsgTypes.BookmarkNodeList.parseFrom(rustBuf.asCodedInputStream()!!)
-                return unpackProtobufItemList(message)
-            } finally {
-                LibPlacesFFI.INSTANCE.places_destroy_bytebuffer(rustBuf)
-            }
+    override fun getBookmarksWithURL(url: String): List<ReadonlyBookmarkItem> {
```
mhammond, line 358: and here re `Url`

components/places/android/src/test/java/mozilla/appservices/places/PlacesConnectionTest.kt
```diff
@@ -359,25 +361,25 @@ class PlacesConnectionTest {
...
-        val sep = db.getBookmark(sepGUID)!! as BookmarkSeparator
-        val folder = db.getBookmark(folderGUID)!! as BookmarkFolder
+        val item = db.getBookmark(itemGUID)!! as ReadonlyBookmarkItem.Bookmark
+        val sep = db.getBookmark(sepGUID)!! as ReadonlyBookmarkItem.Separator
+        val folder = db.getBookmark(folderGUID)!! as ReadonlyBookmarkItem.Folder
 
-        assertEquals(item.type, BookmarkType.Bookmark)
-        assertEquals(sep.type, BookmarkType.Separator)
-        assertEquals(folder.type, BookmarkType.Folder)
+        // assertEquals(item.type, BookmarkType.BOOKMARK)
```
mhammond, line 368: I think it's fine to delete these checks - even in the old version they seem a bit pointless - we've explicitly cast just a few lines up (ie, it is was possible to cast a Separator as a Folder, I suspect the type would be Folder.)

components/places/src/storage/bookmarks.rs
```diff
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+* License, v. 2.0. If a copy of the MPL was not distributed with this
```
mhammond, line 2: I think the old version of this was correct.

components/places/src/storage/bookmarks.rs
```diff
@@ -426,6 +431,110 @@ impl Default for UpdateTreeLocation {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ReadonlyBookmark {
+    pub guid: SyncGuid,
+    pub parent_guid: Option<SyncGuid>,
+    // Always 0 if parent_guid is None
+    pub position: BookmarkPosition,
```
mhammond, line 439: I think this should probably just be an int - `BookmarkPosition::Append` wouldn't make sense here. Ditto in the structs below.

components/places/src/storage/bookmarks.rs
```diff
@@ -426,6 +431,110 @@ impl Default for UpdateTreeLocation {
...
+    pub date_added: Timestamp,
+    pub last_modified: Timestamp,
+    pub url: Option<Url>,
```
mhammond, line 442: I *think* this should be a non-option - you can't have a bookmark without a url

components/places/src/storage/bookmarks.rs
```diff
@@ -426,6 +431,110 @@ impl Default for UpdateTreeLocation {
...
+}
+
+// The type used to update the actual item.
+#[derive(Debug, Clone)]
+pub enum ReadonlyItem {
+    Bookmark { b: ReadonlyBookmark },
+    Separator { s: ReadonlySeparator },
+    Folder { f: ReadonlyFolder },
+}
+
```
mhammond, line 492: I think we should get a ticket on file to kill `PublicNode` - I don't think it's offering any value to use at the moment.

mhammond (commented):
This is looking great - just some nits!

mhammond (approved):
This is ready whenever you are :)

components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
lougeniaC64: I went back and forth myself

components/places/src/storage/bookmarks.rs
lougeniaC64: Very much agree--I created a ticket in the after places uniffication epic for this.

lougeniaC64:
@mhammond I've made a number of

> This is ready whenever you are :)

@mhammond Thanks for all the feedback. I've made some not quite superficial changes since you approved this PR so I'm not merging for that reason. Otherwise, merge away. And feel free to amend these changes as needed.

mhammond:
This is awesome, great work @lougeniaC64

PR #4710, opened 2021-12-06
Fix swiftformat

tarikeshaq:
Fixes a swiftformat ci failure from the latest swiftformat release

Also realized that our `tests.py` still uses `--swiftversion 4`, when we are now using swift version 5 after the `Xcode` upgrade to 13


lougeniaC64 (approved):
Just running into this. 🚀

PR #4712, opened 2021-12-07
Don't provide a stdc++ on Windows

Dexterp37:
In the current state of things, the build will fail on Windows, because it won't find the stdlib.

The rustc docs mention (but I'm struggling to find the docs again) that, on Windows, this should not be provided.
It fixes the compilation for me, locally.

Update: while I can't find the documentation page that @badboy sent me a while back, he was kind enough to provide [this example](https://github.com/rust-lang/rust/blob/87dce6e8dfdae605c9c2a713cf0133066a52022a/compiler/rustc_llvm/build.rs#L288-L303) of how rustc skips everything when the target is msvc.

  - Note:

### DIFF
--- a/components/support/rc_crypto/nss/nss_build_common/src/lib.rs
+++ b/components/support/rc_crypto/nss/nss_build_common/src/lib.rs
@@ -82,7 +82,7 @@ fn link_nss_libs(kind: LinkingKind) {
     let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
     if target_os == "android" || target_os == "linux" {
         println!("cargo:rustc-link-lib=stdc++");
-    } else {
+    } else if target_os != "windows" {
         println!("cargo:rustc-link-lib=c++");
     }
 }


mhammond:
How are you building locally? I seem unable to get nss and sqcipher built both with and without this patch. I've been forced into wsl for the last few years, so would love to discover the secret-sauce here.

Dexterp37:
> How are you building locally? I seem unable to get nss and sqcipher built both with and without this patch. I've been forced into wsl for the last few years, so would love to discover the secret-sauce here.

I'm using the lines from https://github.com/mozilla-services/remote-settings-client/blob/6bb0b0d1f0a9243ffe4a58d81f682ba721e226f2/.circleci/config.yml#L23-L52 :D Calling them locally in WSL

mhammond:
ha - so using WSL to cross-compile native windows binaries! For interest, what do you do with them? I just tend to stay in the WSL environment to run as well so I don't think this will actually help my workflow in practice. Either way, this LGTM, but I'll leave the r+ to @leplatrem

leplatrem (approved):
I don't have Windows at hand, but if that change works for you, I have nothing against it! :shrug:
(the `else` here was for Mac, where the stdlib is available as `c++` instead of `stdc++`)

Dexterp37:
Gah, I don't have time sadly to chase the iOS problems. I'm also happy to keep using this patch locally by changing the `Cargo.toml` as follow (for future reference to anyone stumbling upon this, since it took us some time to figure out):

```toml
# Patch NSS until fixed by a-s
[patch.'https://github.com/mozilla/application-services']
nss_build_common = { path = "patches/nss_build_common" }
```

PR #4714, opened 2021-12-08
Uniffi Places Sync

skhamis:
No identified changes needed on either consumer for these changes!

  - Note:

### DIFF
--- a/components/places/src/ffi.rs
+++ b/components/places/src/ffi.rs
@@ -289,6 +289,83 @@ pub struct HistoryVisitInfosWithBound {
     pub bound: i64,
     pub offset: i64,
 }
+// registerWithSyncManager(handle: u64, error: &mut ExternError) {
+//     log::debug!("register_with_sync_manager");
+//     APIS.call_with_output(error, handle, |api| {
+//         api.clone().register_with_sync_manager()
+//     })
+// }
+
+// #[no_mangle]
+// pub extern "C" fn places_reset(handle: u64, error: &mut ExternError) {
+//     log::debug!("places_reset");
+//     APIS.call_with_result(error, handle, |api| -> places::Result<_> {
+//         api.reset_history()?;
+//         Ok(())
+//     })
+// }
+
+fn wipe_local(&self) -> Result<()> {
+    self.with_conn(|conn| storage::history::wipe_local(conn))
+}
+
+fn run_maintenance(&self) -> Result<()> {
+    self.with_conn(|conn| storage::run_maintenance(conn))
+}
+
+fn prune_destructively(&self) -> Result<()> {
+    self.with_conn(|conn| storage::history::prune_destructively(conn))
+}
+
+fn delete_everything(&self) -> Result<()> {
+    self.with_conn(|conn| storage::history::delete_everything(conn))
+}
+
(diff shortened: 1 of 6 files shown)


components/places/src/ffi.rs
```diff
@@ -80,6 +80,52 @@ impl PlacesApi {
         let connection = PlacesConnection { db: Mutex::new(db) };
         Ok(Arc::new(connection))
     }
+
+    // NOTE: These should be deprecated as soon as possible - that will be once
+    // we have implemented the sync manager and migrated consumers to that.
```
mhammond, line 85: maybe update this comment to something like: "they should be unused on Android but will remain needed for iOS until we can move to the sync manager there (which itself depends on replacing their existing sync engines with ours" or similar?

components/places/src/ffi.rs
```diff
@@ -255,6 +301,22 @@ impl PlacesConnection {
             Ok(())
         })
     }
+
+    fn wipe_local(&self) -> Result<()> {
```
mhammond, line 305: this is only doing history and other misc stuff (ie, not bookmarks) so even though this problem kinda before, maybe rename it to, say `wipe_local_history`? In the kotlin and swift, keep the "public" names as they were, even though they remain misleading, but add a comment saying everything other than bookmarks is removed?

components/places/src/ffi.rs
```diff
@@ -255,6 +301,22 @@ impl PlacesConnection {
...
+
+    fn run_maintenance(&self) -> Result<()> {
+        self.with_conn(|conn| storage::run_maintenance(conn))
+    }
+
+    fn prune_destructively(&self) -> Result<()> {
+        self.with_conn(|conn| history::prune_destructively(conn))
+    }
+
+    fn delete_everything(&self) -> Result<()> {
```
mhammond, line 317: eek - I guess we should squeeze the word "history" in here too - but looking at the impl, the difference between `wipe_local` and `delete_everything` is tiny - I think this one will also prevent the history from coming back via sync - so for now, either here or in the IDL, move these close to each other and add an XXX comment that we should document exactly how these work.

components/places/src/ffi.rs
```diff
@@ -255,6 +301,22 @@ impl PlacesConnection {
...
+
+    fn wipe_local(&self) -> Result<()> {
+        self.with_conn(|conn| history::wipe_local(conn))
+    }
+
+    fn run_maintenance(&self) -> Result<()> {
+        self.with_conn(|conn| storage::run_maintenance(conn))
+    }
+
+    fn prune_destructively(&self) -> Result<()> {
```
mhammond, line 313: *sob* - this is an alias for `wipe_local` :(

components/places/src/places.udl
```diff
@@ -80,6 +92,19 @@ interface PlacesConnection {
 
     [Throws=PlacesError]
     void delete_visit(string url, Timestamp timestamp);
+
+    [Throws=PlacesError]
```
mhammond, line 96: (ie, move wipe_local next to delete_everything and add a suitably sad comment here about how the distinction need to be better defined)

mhammond (approved):
perfect, thanks!

PR #4717, opened 2021-12-08
Only run dependency CI on schedule

### LINKED ISSUE #4623: Move depedency CI to run on intervals and open issues when it fails
Our dependency CI is failing too often, and turning red on PRs that have nothing to do with dependencies. In practice, we either end up regenerating dependency summaries, `cargo update` or ignoring an advisory if we're not affected by it.

This all has nothing to do with the PR that its failing against, and adds some overhead. We still **need** to make sure we tackle advisories and have our dependency summaries up to date, but I don't think it should be done on every PR.

Here's what I'm proposing:
- Have a CI job that runs every day (or more often) - if it fails due to a dependency issue, it creates a github issue with a specific label. We can then fix it (which in practice is a quick PR) but it wouldn't fail CI for any other PRs. 
- Have that job run on PRs **only** if the PR modifies a `Cargo.toml` or otherwise changes any dependency information (i.e if someone runs the regeneration of dependency summaries, we need to make sure it fixes the issue)

I'll put this as `discuss` because if the value we gain from the check is worth having it run on every PR I'm happy to close this!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2874)

tarikeshaq:
Fixes #4623 

We already have dependency CI running on a schedule, and in-practice, the fact that we have it running on PRs isn't really helping because people are more often than not getting red ci when they didn't change anything related to dependencies.

The demo I had also ran CI when a PR changed any `Cargo.toml` or the `Cargo.lock`, I'll file a follow up to implement that - I could have switched us to use GitHub actions for this (which would have made it super simple to implement the followup), but wasn't sure if it was worth the switch (and the added complexity of dealing with two different types of CI runners)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -490,9 +490,6 @@ workflows:
   check-protobuf-uptodate:
     jobs:
       - Check Protobuf files are up-to-date
-  check-dependencies:
-    jobs:
-      - Check Rust dependencies
   bash-lint:
     jobs:
       - Lint Bash scripts


mhammond (approved):
You now own our CI 🤣

tarikeshaq:
😅 😅 😰

mhammond:
What do I do when I'm actually trying to change the dependencies?

tarikeshaq:
Probably just run the `./tools/regenerate_dependency_summaries.sh` script - and if anything slips in, we'll catch it when the cron job runs... admittedly we still want to run CI on changes that change dependencies (which was my original intent, but circle ci doesn't make it easy). I filed https://github.com/mozilla/application-services/issues/4722 to deal with that

mhammond:
> Probably just run the `./tools/regenerate_dependency_summaries.sh`

IIRC, that script only works on macs (as in, some deps only exist when run from a mac), so I'm almost guaranteed to generate a file that fails. But this isn't urgent and #4722 sounds perfect, thanks.

PR #4720, opened 2021-12-09
Uniffi Search result and match url

### LINKED ISSUE #4703: Uniffi Search Result and Match Reason Types
Uniffi the _SearchResult_ and _MatchReason_ types (_SearchResult_ and _SearchResultReason_ kotlin types) as well as the below functions:

* places_query_autocomplete
* places_accept_result
* places_match_url

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2933)
┆Story Points: 3
┆epic: Uniffi Places
┆sprintEndDate: 2021-12-10

tarikeshaq:
fixes #4703 

This was a little quicker than expected. Android components builds and tests pass without any changes on that side

One thing to point out is the dance I do with the `SearchResult` and the `MatchResult`... basically duping the type for:
- AC only needs a subset of the fields for `SearchResult` that we use internally
- MatchResult has a variant `Url` which `uniffi` rejects since it shadows the wrapped type `Url`
- MatchResult has a `Tag` variant with associated data, AC doesn't use that data and we used to pass it as a flat enum across the FFI

### DIFF
--- a/components/places/ios/Places/RustPlacesAPI.h
+++ b/components/places/ios/Places/RustPlacesAPI.h
@@ -60,11 +60,6 @@ void places_note_observation(PlacesConnectionHandle handle,
                              const char *_Nonnull observation_json,
                              PlacesRustError *_Nonnull out_err);
 
-char *_Nullable places_query_autocomplete(PlacesConnectionHandle handle,
-                                          const char *_Nonnull search,
-                                          int32_t limit,
-                                          PlacesRustError *_Nonnull out_err);
-
 char *_Nullable places_match_url(PlacesConnectionHandle handle,
                                  const char *_Nonnull search,
                                  PlacesRustError *_Nonnull out_err);
--- a/components/places/ios/Places/RustPlacesAPI.h
+++ b/components/places/ios/Places/RustPlacesAPI.h
@@ -60,10 +60,6 @@ void places_note_observation(PlacesConnectionHandle handle,
                              const char *_Nonnull observation_json,
                              PlacesRustError *_Nonnull out_err);
 
-char *_Nullable places_match_url(PlacesConnectionHandle handle,
-                                 const char *_Nonnull search,
-                                 PlacesRustError *_Nonnull out_err);
-
 void places_bookmarks_import_from_ios(PlacesAPIHandle handle,
                                       const char *_Nonnull db_path,
                                       PlacesRustError *_Nonnull out_err);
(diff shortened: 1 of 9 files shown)


components/places/src/ffi.rs
```diff
@@ -379,6 +401,27 @@ impl FrecencyThresholdOption {
     }
 }
 
+// We define those types to cross the FFI
```
tarikeshaq, line 404: This is the dance I mentioned in the PR comment

One alternative is to:
- Change the variant for the internal `MatchReason` from `Url` to another name
- Change the `Tags` internal variant to not have associated data (or change `MatchUrl` to use our `.udl` syntax for enums with associated data, although... AC doesn't use the data so...)
- Then expose the internal SearchResult and MatchReason as apposed defining new types here

Thoughts about the tradeoff here?

components/places/ios/Places/RustPlacesAPI.h
```diff
@@ -60,15 +60,6 @@ void places_note_observation(PlacesConnectionHandle handle,
                              const char *_Nonnull observation_json,
                              PlacesRustError *_Nonnull out_err);
 
-char *_Nullable places_query_autocomplete(PlacesConnectionHandle handle,
```
tarikeshaq, line 63: those weren't used anywhere as far as I can tell, I searched in our repo and in firefox ios

components/places/src/ffi.rs
mhammond: This seems OK, although I think it's a uniffi bug that we should open and even try to fix. The comment in the code isn't quite clear about why renaming `Url` is a solution (ie, the comment doesn't describe the problem)

The fact `tags` has associated data but other variants do not seems quite odd and as you seem to be implying it's not actually used by Kotlin - and I can't really see how it would be helpful - so dropping that tag seems an easy solution, but I'm fine either way.

components/places/ios/Places/RustPlacesAPI.h
mhammond: yeah, ios only uses our bookmarks but not history, so this makes sense to me.

mhammond (approved):
lovely!

components/places/src/ffi.rs
tarikeshaq: filed https://github.com/mozilla/uniffi-rs/issues/1137 for the uniffi bug, I'll also file one of those follow-ups for the tags to see if there's a better approach there (like maybe just drop the associated data from `Tags` in the internal type as well, or remove that variant as a whole)

PR #4721, opened 2021-12-09
Bring in all dependabot changes

mhammond:
A bit sick of seeing those PRs hang around, and they are often red due to the dependency lists changing, so might as well kill all those birds with one stone.


mhammond:
Not sure how `PyGithub` was used previously, but it appears unused now - updating it caused a load of other transient requirements to start coming in, so I just removed it.

tarikeshaq (approved):
🚀

PR #4723, opened 2021-12-09
Upgrade rust to 1.57

tarikeshaq:
Upgrades rust to 1.57

The changes here were mechanical and based on rustc or clippy's recommendations...

Two main highlights:
- We now use rust stable for the cross-compilation of the M1 simulators, not urgent but if CI passes @skhamis it would be awesome if you can give this a shot when you have the time 🙏 🙏 
- rustc now considers code that is written, but never read "dead code" I found at least one instance where it got it wrong (there was an assert_eq that definitely read that field 😅) so I opted to allow them across the board, except for the example code. Thoughts?

  - Note:


examples/places-autocomplete/src/autocomplete.rs
```diff
@@ -24,45 +24,27 @@ struct ImportPlacesOptions {
...
-    id: i64,
     date: i64,
     visit_type: u8,
-    from_visit: i64,
 }
 
 #[derive(Default, Debug, Clone)]
 struct LegacyPlace {
     id: i64,
-    guid: String,
```
mhammond, line 36: there's not alot of value here, but I think this is reading a desktop places row, which possibly has *some* value? Maybe add a leading underscore on the names (and a comment about why?) I'm assuming this is what is described in https://github.com/rust-lang/rust/issues/81658, and that's one of the suggestions there?)

mhammond:
that new dead_code check is unfortunate :( But we can't just stay on the old rust, so we have to do *something*. I'm fine with this, but I'll leave it for others to weigh in and approve and/or suggest changes.

examples/places-autocomplete/src/autocomplete.rs
bendk: I think I have a slight preference for commenting out fields like this rather than deleting it, adding the underscore, or adding the `#[allow(dead_code)]` annotation.  This seems like a nice way to indicate that the fields are present in the source data in case we want to use them some time in the future.  That said, I'm fine with any of the solutions.

bendk:
> rustc now considers code that is written, but never read "dead code" I found at least one instance where it got it wrong (there was an assert_eq that definitely read that field)

Was the `assert_eq` in a test?  Maybe we should update the test code or just delete the test.  There doesn't seem to be much use in testing something that never gets read in the actual code.

rust-toolchain
```diff
@@ -7,14 +7,15 @@
...
-channel = "1.55.0"
+channel = "1.57.0"
 targets = [
     "aarch64-linux-android",
     "armv7-linux-androideabi",
     "i686-linux-android",
     "x86_64-linux-android",
     "aarch64-apple-ios",
     "x86_64-apple-ios",
+    "aarch64-apple-ios-sim"
```
skhamis, line 18: Exciting stuff! I believe this actually landed as part of the 1.56 work but happy for us not have to rely on nightly for our builds :). I'm a little concerned on the `ios-artifacts` failure given it's failing to build sqlcipher and nss

tarikeshaq: Are you referring to this failure https://app.circleci.com/pipelines/github/mozilla/application-services/29255/workflows/9df293bb-df35-44b4-87e9-174ae44f8ee0/jobs/123114? or the `ios-artifacts` failing on main - the latter should be fixed by this but the sqlcipher build failure on the `cargo bench` that this PR is introducing thou... I'll have to look into it a little

skhamis: ah didn't realize there was also a main one, I was looking specifically at this sqlcipher one since it seems a bunch of sqlcipher artifacts go missing. Once I free up some time I might help look as well!

tarikeshaq: Yeah the one on main is a known rust nightly bug that should get fixed soon, and it will also get fixed with this PR (since we move away from using nightly 🎉) the sqlcipher one I haven't looked into, but it's weird that `cargo test` works, but `cargo bench` causes trouble... and I can't repro locally either 🙃

tarikeshaq:
> Was the assert_eq in a test? 

Yeah, it's stuff like https://github.com/mozilla/application-services/blob/021e64902f0c1f0792d7f7249fe49d8344ce48c4/components/places/src/storage/bookmarks.rs#L2113-L2127

I'm not too sure about all that's going on there, but the `sync_change_counter` **seems** to be used as some type of sanity checking? Same goes for the other `dead_code` fields I assigned for that type, think the other structs I `dead_code`ed are fine thou

mhammond:
huh - so in that particular example, it looks like the `sync_change_counter` field isn't actually read by the implementation - it's just read by sql, so apart from those tests, could actually be removed from the struct. Still don't have a strong opinion about the best way to deal with that though :)

mhammond:
> Still don't have a strong opinion about the best way to deal with that though :)

I guess for that particular case, a `#cfg` might work, which might be an interesting experiment, but I'm completely fine with a pragmatic decision that it's not worthwhile here.

tarikeshaq:
>  a #cfg might work

👀 👀 👀 Love the idea!! I'll give it a shot

mhammond (approved):
lgtm!

PR #4726, opened 2021-12-10
Uniffi Places Interrupt methods

skhamis:

  - Note:

### DIFF
--- a/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
@@ -102,7 +102,9 @@ class PlacesApi(path: String) : PlacesManager, AutoCloseable {
         if (handle != 0L) {
             if (writeHandle != 0L) {
                 try {
-                    this.api.apiReturnWriteConn()
+                    rustCall(this) { err ->
+                        LibPlacesFFI.INSTANCE.places_api_return_write_conn(handle, writeHandle, err)
+                    }
                 } catch (e: PlacesException) {
                     // Ignore it.
                 }
@@ -215,13 +217,13 @@ open class PlacesConnection internal constructor(connHandle: Long, uniffiConn: U
     // Each method here will use one or the other, depending on whether it's been uniffi'd or not.
     protected var handle: AtomicLong = AtomicLong(0)
     protected var conn: UniffiPlacesConnection
-    protected var interruptHandle: InterruptHandle
+    protected var interruptHandle: SqlInterruptHandle
 
     init {
         handle.set(connHandle)
         conn = uniffiConn
         try {
-            interruptHandle = InterruptHandle(this.conn.newInterruptHandle())!!
+            interruptHandle = this.conn.newInterruptHandle()
         } catch (e: Throwable) {
             rustCall { error ->
                 LibPlacesFFI.INSTANCE.places_connection_destroy(this.handle.getAndSet(0), error)
@@ -1084,25 +1086,6 @@ interface WritableHistoryConnection : ReadableHistoryConnection {
     fun acceptResult(searchString: String, url: String)
 }
 
-class InterruptHandle internal constructor(raw: SqlInterruptHandle) : AutoCloseable {
-    // We synchronize all accesses, so this probably doesn't need AtomicReference.
-    private val handle: AtomicReference<SqlInterruptHandle?> = AtomicReference(raw)
-
-    @Synchronized
-    override fun close() {
-        val toFree = handle.getAndSet(null)
-        if (toFree != null) {
-            LibPlacesFFI.INSTANCE.places_interrupt_handle_destroy(toFree)
-        }
-    }
-
-    @Synchronized
-    fun interrupt() {
-        handle.get()?.let {
-            this.interrupt()
-        }
-    }
-}
 
 enum class VisitType(val type: Int) {
     /** This isn't a visit, but a request to update meta data about a page */
--- a/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
@@ -22,6 +22,7 @@ import mozilla.appservices.places.uniffi.VisitObservation
 import mozilla.appservices.places.uniffi.HistoryVisitInfo
 import mozilla.appservices.places.uniffi.HistoryVisitInfosWithBound
 import mozilla.appservices.places.uniffi.SearchResult
+import mozilla.appservices.places.uniffi.SqlInterruptHandle
 import mozilla.appservices.support.native.toNioDirectBuffer
 import mozilla.appservices.sync15.SyncTelemetryPing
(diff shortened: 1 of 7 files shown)


components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -101,9 +102,7 @@ class PlacesApi(path: String) : PlacesManager, AutoCloseable {
         if (handle != 0L) {
             if (writeHandle != 0L) {
                 try {
-                    rustCall(this) { err ->
-                        LibPlacesFFI.INSTANCE.places_api_return_write_conn(handle, writeHandle, err)
-                    }
+                    this.api.apiReturnWriteConn()
```
mhammond, line 105: I don't think this is correct - it should remain using the old technique while we still have that duality of both handle and uniffi connections. We will need *something* to kill the uniffi side of the world, but that might just be able to be a `this.writeConn.close()` method.

(To be clear, this probably isn't going to hurt because we will kill the old handle stuff before merging to main, but it still makes sense to me to keep things clear)

components/places/src/ffi.rs
```diff
@@ -131,6 +132,13 @@ impl PlacesApi {
         )?;
         Ok(serde_json::to_string(&ping).unwrap())
     }
+
+    // TODO: Relook into this and fix appropiately
```
mhammond, line 136: as above, I think just removing this from this PR is fine. We probably do need a new ticket for "remove all old handle cruft and ensure we have appropriate cleanup/close functions for the new world"

components/places/src/places.udl
```diff
@@ -21,10 +21,21 @@ enum ConnectionType {
...
 
+interface SqlInterruptHandle {
+     void interrupt();
+};
+
 interface PlacesApi {
     [Throws=PlacesError]
     PlacesConnection new_connection(ConnectionType conn_type);
 
+    //NOTE: this doesn't appear to be actively used and can probably remove
```
mhammond, line 32: In theory it *should* be used by ios so they can cancel a long running sync on shutdown, but yeah, for now, this is fine (maybe update the comment to also say the sync manager will typically arrange to interrupt syncs or something)

mhammond (commented):
This looks roughly correct! It's a shame things are broken already, but I don't see how this makes anything worse and I think it is on the right path.

tarikeshaq (commented):
I think this might be our first use of having an interface reference another interface 🤩 Looks good! Dropped a couple of thoughts

Something that maybe we can follow-up on later after we are done uniffication:

We usually don't have interfaces referencing other interfaces across the ffi because I imagine it usually happens all in Rust, is it possible to move the setup for the interrupt handler into Rust? 

So the [`PlacesConnection`](https://github.com/mozilla/application-services/blob/9067babc820f04d87b6d18504ae6b816980067c5/components/places/src/ffi.rs#L136) struct would hold an `interrupt_handler` field as opposed to the kotlin class. Then the consumer can call the `interrupt` method on the `PlacesConnection` class, and it would internally deal it.

That said, I'm not sure what the reasoning was to setup the API like this in the first place..

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
skhamis: hmmm this is an interesting point. So it actually makes sense to simply not uniffi these connection-level handle problems and mainly just focus on being able to call interrupt on the new connections.

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -1104,11 +1099,7 @@ class InterruptHandle internal constructor(raw: RawPlacesInterruptHandle) : Auto
     @Synchronized
     fun interrupt() {
         handle.get()?.let {
-            val e = RustError.ByReference()
-            LibPlacesFFI.INSTANCE.places_interrupt(it, e)
-            if (e.isFailure()) {
-                throw e.intoException()
-            }
+            this.interrupt()
```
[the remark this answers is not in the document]
skhamis: 🤦 that's what I get for putting this PR up way too early in the draft process haha. Noted!

skhamis:
The swiftlint is unrelated to this PR but everything else is passing, and no identified changes needed in either consumer!

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -216,17 +216,13 @@ open class PlacesConnection internal constructor(connHandle: Long, uniffiConn: U
     // Each method here will use one or the other, depending on whether it's been uniffi'd or not.
     protected var handle: AtomicLong = AtomicLong(0)
     protected var conn: UniffiPlacesConnection
-    protected var interruptHandle: InterruptHandle
+    protected var interruptHandle: SqlInterruptHandle
```
mhammond, line 219: I have a very vague preference for just `InterruptHandle` here, because it's (a) more than just sql and (b) kinda leaks an implementation detail out (eg, ultimately "tabs" should expose an interrupt handle too, but it doesn't even use sql!)

OTOH, I'm really just saying I hate the name in the rust code, but the UDL having the same name as the rust code does make some sense.

So ultimately I'll leave that up to you.

components/places/src/ffi.rs
```diff
@@ -147,6 +148,12 @@ impl PlacesConnection {
         Ok(f(&conn)?)
     }
 
+    fn new_interrupt_handle(&self) -> Result<Arc<SqlInterruptHandle>> {
```
mhammond, line 151: The semantics here also smell wrong. Given the arcs etc, I think the right approach would be for this to be called something like `get_interrupt_handle()` and the struct itself gets a `Weak<SqlInterruptHandle>` or something (the idea being there's exactly 1 interrupt handle per connection and getting multiple "new" ones per connection makes no sense) - but it doesn't make sense to fix this now, so please just add a comment referencing 1684 too.

(And if you get bored, starting to tackle 1684 makes some sense to me - we might decide to land it after we land this effort, but there's no good reason to not get started now.)

mhammond (approved):
LGTM - it's a shame the shape of this api gets more broken than it already was, but I think it's fine to land with this initially to help keep our blast radius down 💣

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
skhamis: hmm unfortunately the `interface SqlInterruptHandle` in UDL is in reference to the rust side. So I believe the three options we have are:
1. Rename SqlInterrupt handle in the support component (used quite a bit though)
2. Typealias in our kotlin/swift wrappers to keep the name (though still leaks a tiny bit)
3. Keep same

I have no strong preference either way but would probably lean to leaving as-is if only to keep things simple for now until the interrupt rework takes place.

components/places/src/ffi.rs
skhamis: I completely agree it feels weird to get a new handle everytime and will add proper comments,

PR #4727, opened 2021-12-10
Cleaning up Sync Manager code

bendk:
Added an enum that contains all possible sync engines and a method to
iterate over registered sync engines.  Used that, plus the existing
`get_engine()` method to remove a lot of redundant code.

I wanted to add tests for this, but I couldn't figure out a good way.  I think it's possible to create mock `dyn SyncEngine`s, then test if they had the correct methods called on them.  But this seems like a lot of work for the amount it's testing.

  - Note:

### DIFF
--- a/components/sync_manager/src/manager.rs
+++ b/components/sync_manager/src/manager.rs
@@ -7,20 +7,14 @@ use crate::types::{ServiceStatus, SyncEngineSelection, SyncParams, SyncReason, S
 use crate::{reset, reset_all, wipe};
 use parking_lot::Mutex;
 use std::collections::{HashMap, HashSet};
+use std::convert::TryFrom;
 use std::sync::{atomic::AtomicUsize, Arc};
 use std::time::SystemTime;
 use sync15::{
     self,
     clients::{Command, CommandProcessor, CommandStatus, Settings},
-    EngineSyncAssociation, MemoryCachedState, SyncEngine,
+    EngineSyncAssociation, MemoryCachedState, SyncEngine, SyncEngineId,
 };
-
-const LOGINS_ENGINE: &str = "passwords";
-const HISTORY_ENGINE: &str = "history";
-const BOOKMARKS_ENGINE: &str = "bookmarks";
-const TABS_ENGINE: &str = "tabs";
-const ADDRESSES_ENGINE: &str = "addresses";
-const CREDIT_CARDS_ENGINE: &str = "creditcards";
 
 #[derive(Default)]
 pub struct SyncManager {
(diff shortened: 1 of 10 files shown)


CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,11 @@ Use the template below to make assigning a version number during the release cut
...
 
 -->
+
+## Sync Manager
+
+### What's Changed
+  - Sync manager methods have now been standardized to support all available
+    engines.  Before each method supported its own subset of the engines
+    (`wipe()` and `reset()` didn't support tabs, `reset_all()` didn't support
+    tabs or logins, etc).
```
bendk, line 28: Maybe I should have called this a breaking change, since the behavior of `reset_all()` is different.  But it seems more like a bugfix to me and also I don't think Fenix ever calls `resetAll()`.

tarikeshaq (commented):
Only reviewing this because I don't know much about syncmanager and this was an opportunity to take a tiny look under the hood 😄 Love the clean-up!

components/sync_manager/src/manager.rs
```diff
@@ -32,180 +86,75 @@ impl SyncManager {
...
-                }
-                Ok(())
-            }
-            "addresses" | "creditcards" => {
-                if let Some(engine) = Self::autofill_engine(engine) {
-                    engine.wipe()?;
-                }
-                Ok(())
-            }
-            _ => Err(SyncManagerError::UnknownEngine(engine.into())),
```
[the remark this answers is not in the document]
bendk: Hopefully @mhammond knows more about this.  From what I can tell, Android doesn't call these sync manager methods yet and iOS doesn't have sync manager at all.  I some of the individual stores have `wipe()` and `reset()` methods, and those are called instead.

components/sync_manager/src/manager.rs
```diff
@@ -32,180 +86,75 @@ impl SyncManager {
...
+            EngineId::Addresses => Self::autofill_engine("addresses"),
+            EngineId::CreditCards => Self::autofill_engine("creditcards"),
+            EngineId::Logins => Self::logins_engine("logins"),
+            EngineId::Tabs => Self::tabs_engine("tabs"),
         }
     }
 
-    pub fn wipe(&self, engine: &str) -> Result<()> {
-        match engine {
-            "logins" => {
```
mhammond, line 68: heh - I think this is theoretically broken - I'm fairly sure engine is going be "passwords" here - although this never happens in practice (ie, nothing actually will ask for that engine to wipe itself in the current world). I think your patch actually fixes that, but should maybe go even further.

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,16 @@ Use the template below to make assigning a version number during the release cut
...
+
+## Sync Manager
+
+### What's Changed
+  - Sync manager methods have now been standardized to support all available
+    engines.  Before each method supported its own subset of the engines
+    (For example, `wipe()` and `reset()` didn't support tabs).
+
+### ⚠️ Breaking Changes ⚠️
+  - The `SyncManager.wipe()` and `SyncManager.reset()`, now expects "passwords"
```
mhammond, line 30: TBH I wouldn't call this out either - you just accidentally fixed a bug :) And it's not a breaking change to consumers here - it's kinda more like a bugfix that affects other *devices*. So I'm not sure this patch needs any README changes at all (but I'm fine with whatever you prefer)

mhammond (approved):
This is great - @skhamis actually got some of the way towards this when uniffiing logins, but then backed away due to other conflicts and never got back to it!

I think this is all correct and does include that accidental bugfix - but I think it's still worth consolidating the name "passwords" even more into the sync manager - however, I'm fine with you deciding not to do that, because there's certainly less confusion in this patch than there was before.

bendk:
Thanks for the review!  I made a few more changes based on that:

 - Removed the changelog entries.  I wasn't sure if these should be there or not, seems like it's not needed since the functionality that's actually being used isn't changing.
 - Made more use of the `EngineId` enum rather than hard-coded strings.  I moved it into `sync15-traits` and made the components use for their `get_registered_sync_engine()` functions.  I was tempted to keep going with this one and make `SyncEngineSelection` use it, but I stopped short of that.

components/sync_manager/src/manager.rs
```diff
@@ -32,180 +86,75 @@ impl SyncManager {
...
-                log::warn!("get_engine() unknown engine: {}", engine);
-                None
-            }
+    fn get_engine(engine_id: &EngineId) -> Option<Box<dyn SyncEngine>> {
+        match engine_id {
+            EngineId::History => Self::places_engine("history"),
+            EngineId::Bookmarks => Self::places_engine("bookmarks"),
+            EngineId::Addresses => Self::autofill_engine("addresses"),
+            EngineId::CreditCards => Self::autofill_engine("creditcards"),
+            EngineId::Logins => Self::logins_engine("logins"),
```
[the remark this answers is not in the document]
bendk: I decided to push the `EngineId` type into tho the `get_registered_sync_engine()` function.  This seems like the best way to avoid issues with the strings not lining up.

skhamis:
Just adding on that this is an awesome job! Definitely simplifies how we add/fix engines without needing to do so across the codebase!

mhammond:
> * Made more use of the `EngineId` enum rather than hard-coded strings.

I like this, thanks!

PR #4729, opened 2021-12-13
Run dependency check on prs and pushes that change dependencies

### LINKED ISSUE #4722: Run dependecy CI on changes to `Cargo.toml`s or the `Cargo.lock`
Follow up to #4717 

We used to run our dependency CI on every PR, which was giving us a red CI on many PRs that had nothing to do with dependencies.. but for PRs that actually update our dependencies we still want to check them

So this is to make sure CI runs for PRs that modify dependencies

tarikeshaq:
fixes #4722 

I switched us to use Github Actions, only for dependency checking because:
- mainly because CircleCi doesn't support path checking properly (it's possible, through dynamic configuration, but just adds more complex logic)
- We already use Github Actions here for the docs, and we use Github Actions in `rust-components-swift` as well

I had originally hesitated away from this, but the ci job here is small enough that i think it's ok

Note that I'm expecting this to fail first, then I'll add a commit to make it pass. I expect it to fail, because I ran `cargo update` and intentionally haven't updated our dependency summaries to see it fail

### DIFF
--- a/tools/dependency_summary.py
+++ b/tools/dependency_summary.py
@@ -552,12 +552,6 @@ PACKAGE_METADATA_FIXUPS = {
     },
     # These packages do not make it easy to infer a URL at which their license can be read,
     # so we track it down by hand and hard-code it here.
-    "ansi_term": {
-        "repository": {
-            "check": None,
-            "fixup": "https://github.com/ogham/rust-ansi-term",
-        },
-    },
     "c2-chacha": {
         "repository": {
             "check": "https://github.com/cryptocorrosion/cryptocorrosion",
@@ -742,6 +736,17 @@ PACKAGE_METADATA_FIXUPS = {
             "fixup": "https://raw.githubusercontent.com/LeopoldArkham/humansize/master/LICENSE-APACHE",
         }
     },
+
+    "encoding_rs": {
+        "license": {
+            "check": None,
+            "fixup": "(Apache-2.0 OR MIT) AND BSD-3-Clause"
+        },
+        "license_url": {
+            "check": None,
+            "fixup": "https://raw.githubusercontent.com/hsivonen/encoding_rs/master/COPYRIGHT"
+        }
+    }
 }
 
 # Sets of common licence file names, by license type.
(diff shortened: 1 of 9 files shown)


tarikeshaq:
Awesome! https://github.com/mozilla/application-services/runs/4509684783?check_suite_focus=true#step:7:389 fails as expected, I'll push the commit that fixes it and lets see if it passes 🕺

tools/dependency_summary.py
```diff
@@ -982,7 +987,7 @@ def pick_most_acceptable_license(self, id, licenseId):
         """
         # Split "A/B" and "A OR B" into individual license names.
         licenses = set(l.strip()
-                       for l in re.split(r"\s*(?:/|\sOR\s)\s*", licenseId))
+                       for l in re.split(r"\s*(?:/|\sOR\s)\s*", str(licenseId)))
```
tarikeshaq, line 990: for some reason one (or more?) of the licenseIds was not a string after the `cargo update` 🤷 I added this and all seems to work properly

tools/dependency_summary.py
```diff
@@ -742,6 +736,17 @@
             "fixup": "https://raw.githubusercontent.com/LeopoldArkham/humansize/master/LICENSE-APACHE",
         }
     },
+
+    "encoding_rs": {
```
tarikeshaq, line 740: Explanation of this fixup in https://github.com/hsivonen/encoding_rs/commit/3a4033e67b6b9d1c1e9514bcb5c20ae05bf8391d

It's an interesting case, and I'm not an expert in licensing so I would love to have someone take a look here to make sure what I'm doing is fine

what I end up doing is adding this as a special case in our licenses, and it gets listed on it own, for example: https://github.com/mozilla/application-services/pull/4729/files#diff-c0246cf339c103dac88ae8c472de424275028137e404e05ee6c54e7f9bc1f45bR1926-R1951

This seems OK to me, since we are listing the license shown on the project as-is, but  would love a second opinion on this

tools/dependency_summary.py
tarikeshaq: nvm me, it was the same license that needed the fixup 🤦

mhammond (approved):
OMG, that license change seems crazy, but I agree this is the correct fix - thanks!

PR #4730, opened 2021-12-13
Removes cargo clean from ci

tarikeshaq:
I saw that our `rust tests` ci was taking over 12 minutes and I was surprised since we use `sccache`... Then I realized it's _Probably_ because we `cargo clean`. The linked issue for why we needed to do that has been fixed (and it was for clippy, no for `cargo test` anyways)

I'll keep an eye on how fast the tests run with this.. if it's not an improvement, I'll close it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ commands:
             # Do not forget to re-export these variables in xc-universal-binary.sh!
             echo 'export "RUSTC_WRAPPER"="sccache"' >> $BASH_ENV
             echo 'export "SCCACHE_IDLE_TIMEOUT"="1200"' >> $BASH_ENV
-            echo 'export "SCCACHE_CACHE_SIZE"="2G"' >> $BASH_ENV
+            echo 'export "SCCACHE_CACHE_SIZE"="4G"' >> $BASH_ENV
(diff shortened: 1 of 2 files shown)


tarikeshaq:
I have... a theory, from https://app.circleci.com/pipelines/github/mozilla/application-services/29298/workflows/f2b8d345-9b8e-4ebe-b2da-a10e53e1cc91/jobs/123414?invite=true#step-101-2 looks like the cache is already maxed out (the max is defined in https://github.com/mozilla/application-services/blob/main/.circleci/config.yml#L42)

I'm seeing how big the cache gets if I run the test script with sccache setup locally.. and will experiment with adjusting that max slightly

tarikeshaq:
>  look like the tests are always recompiling everything on every run, even though its recovering the sccache, and that is the bottleneck

Huh, although the tests are running recompilation, from https://app.circleci.com/pipelines/github/mozilla/application-services/29298/workflows/b339c8ef-210a-4830-86dc-8ff1e588f954/jobs/123423?invite=true#step-120-7 looks like the cache stats look good, I'll still give upping the limit a shot and measure times 🤷, worst case it has no impact and I move on

tarikeshaq:
Yup, doubling the cache still ended up with the same time in tests, and more time packing/unpacking the cache... I'll close this 🙃 If I get capacity I might look at other ways to cut down testing time but I'm done here

PR #4732, opened 2021-12-15
reduce sscache usage by having jobs only update cache if dependencies change

tarikeshaq:
The hope is that this reduces our CI storage costs **significantly**

We used to use a different cache per job, and upload a new version of it each run. The totals to a 2GB `sccache` being stored each job run.. and we run multiple jobs per workflow and multiple workflows per day

I'm not an expert in `sccache`, but since all of our jobs that use it **just** run rust code, I imagine they could share the cache as long as they run the same architecture? This is what this PR does, only uploading a fresh cache if our `Cargo.lock` changes (indicating that there are new dependencies) ~~and sharing the cache across jobs running on the same architecture.~~ 

that last sentence doesn't make sense some jobs will cache more and can take advantage of more cache.. brought back the job separation, but now it only updates if the deps changes (otherwise it won't upload a new cache!) and given how rare our dependencies actually change (vs the total number of jobs) it should still be a very significant reduction

 I'll keep an eye on the performance of the cache given this scheme (I'll run it a couple of times and see the cache hit rate and the time the jobs take vs before this change)


.circleci/config.yml
```diff
@@ -209,26 +209,33 @@ commands:
...
   restore-sccache-cache:
     steps:
       - restore_cache:
           name: Restore sccache cache
-          key: sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}
+          keys: 
+          # We have multiple keys to increase the chance of a cache hit
+          # in case the Cargo.lock is updated, we still want to retrieve
+          # some cache
+            - sccache-cache-stable-{{ arch }}-{{ checksum "Cargo.lock" }}
```
tarikeshaq, line 223: I think I should bring back the `--{ .Environment.CIRCLE_JOB }}` and have it so jobs of the same **type** share caches.. because otherwise, the fastest job (which wouldn't doesn't need too much cache) would upload the cache for that checksum and every other job is stuck with it and won't update it

tarikeshaq:
Time looks exactly the same, and we have a [great hit rate on our cache](https://app.circleci.com/pipelines/github/mozilla/application-services/29345/workflows/d2e7f8c5-758a-4a3c-8b7c-cf8e49d9d776/jobs/123892?invite=true#step-120-7) and we [don't upload if our cache is up to date anymore](https://app.circleci.com/pipelines/github/mozilla/application-services/29345/workflows/d2e7f8c5-758a-4a3c-8b7c-cf8e49d9d776/jobs/123892?invite=true#step-119-1)

.circleci/config.yml
```diff
@@ -175,28 +174,31 @@ commands:
...
       - save_cache:
           name: Save sccache cache
-          # We use {{ epoch }} to always upload a fresh cache:
-          # Of course, restore_cache will not find this exact key,
-          # but it will fall back to the closest key (aka the most recent).
-          # See https://discuss.circleci.com/t/add-mechanism-to-update-existing-cache-key/9014/13
-          key: sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }}
+          # We only upload a new cache if our Cargo.lock changed
+          # otherwise, we update the exact same key
+          key: sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "Cargo.lock" }}
```
tarikeshaq, line 199: is there something better than the `Cargo.lock` as an indicator that our compilation cache should update?

.circleci/config.yml
```diff
@@ -296,27 +302,17 @@ jobs:
...
   iOS build and test:
     executor: macos
     steps:
-      # We do not use the ssccache cache helper commands as
-      # the macOS cache is in a different folder.
-      - restore_cache:
-          name: Restore sccache cache
-          key: sccache-cache-macos-{{ arch }}
       - full-checkout
+      - restore-sccache-cache
```
tarikeshaq, line 315: the `{{ arch }}` already differentiates between the caches for different machines so this can use the same `restore-sccache-cache` command the other docker runner use, so I changed it

.circleci/config.yml
```diff
@@ -381,12 +377,8 @@ jobs:
...
           command: |
             bash megazords/ios-rust/build-xcframework.sh --build-profile release
-      - save_cache:
-          name: Save sccache cache
-          key: sccache-cache-macos-{{ arch }}-{{ epoch }}
-          paths:
-            - "~/Library/Caches/Mozilla.sccache"
-      - sccache-stats-and-logs
+      - save-sccache-cache:
+          path: "~/Library/Caches/Mozilla.sccache"
```
tarikeshaq, line 381: I made the `save-sccache-cache` take a parameter to differentiate between the locations of the cache and have all executors use the same one

.circleci/config.yml
```diff
@@ -236,6 +238,7 @@ jobs:
   Lint Rust with clippy:
     executor: docker
     steps:
+      - full-checkout
```
tarikeshaq, line 241: We should checkout first, to make sure the `Cargo.lock` exists. We end up checksuming the `Cargo.lock` and use the checksum as a part of the key cache

.circleci/config.yml
mhammond: TBH, I don't even understand why Cargo.lock needs special casing - I'd expect the cache to still work out different sources are used. OTOH, I'd also expect that different architectures are transparently handled too, so 🤷 That said though, I'm all up for a low-risk approach given scache is a black-box for us, so +1 to this for now.

mhammond (approved):
Thanks Tarik - this is a bit of a rubber stamp, but the only harm it can do it to our CI, so 🚢

mhammond:
oh - the key is for *saving* - so the `arch` as part of the key does make sense, and I guess Cargo.lock does a bit too, but that's still not quite as clear. Whatever :)

Issue #4713, opened 2021-12-07
Places: Consolidate mismatched types


data-sync-user:
Based on conversation in this PR: [https://github.com/mozilla/application-services/pull/4689](https://github.com/mozilla/application-services/pull/4689|smart-link) there are a few areas that wasn’t able to be URLs due to the ffi layer. With the uniffication going on, we should convert them while in this body of work is going on as well as Timestamp and VisitType.

Areas identified:

* `VisitObservation`Struct: url, refferer, preview_image_url
* VisitType, VisitTransitionSet, and i32 and APIs that mix n match
* Timestamp, u64, i64 and APIs that may or may not use one of those three

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2940)
┆Story Points: 3
┆epic: Uniffi Places
┆sprintEndDate: 2021-12-24

data-sync-user:
➤ Tarik Eshaq commented:

This is done, there one piece that remains regarding the VisitType and VisitTransition, but we opted to leave this till after the uniffication work is done, ref: https://mozilla-hub.atlassian.net/browse/SYNC-2970 ( https://mozilla-hub.atlassian.net/browse/SYNC-2970|smart-link )

PR #4736, opened 2021-12-15
Consolidate types

tarikeshaq:
a part of #4713 

Uses URLs for the `VisitObservation` type and the `VisitTransition` for the `HistoryVisitInfo` type.

Note that our tests will fail, because we still haven't upgraded to a version of uniffi that fixed the whole wrapped types panic if they error out

this as-is also fails to compile AC, but I'll make sure to update that before this lands

I also tried to fix the `VisitTransition` and `VisitType` confusion, but since that's a bigger change I have it as its own PR on top of this PR

### DIFF
--- a/components/places/src/places.udl
+++ b/components/places/src/places.udl
@@ -228,7 +228,7 @@ dictionary HistoryVisitInfo {
     string? title;
     Timestamp timestamp;
     // XXX - This should use the VisitType enum when it gets uniffied
-    i32 visit_type;
+    VisitTransition visit_type;
     boolean is_hidden;
     Url? preview_image_url;
 };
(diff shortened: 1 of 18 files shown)


components/places/src/places.udl
```diff
@@ -228,7 +228,7 @@ dictionary HistoryVisitInfo {
     string? title;
     Timestamp timestamp;
     // XXX - This should use the VisitType enum when it gets uniffied
```
mhammond, line 230: This XXX comment refers to the wrong type, but I think can be removed now?

mhammond (approved):
Awesome, thanks!

PR #4737, opened 2021-12-15
merges the VisitType and VisitTransition types

tarikeshaq:
a part of #4713 

I'm not 100% sure we should do this, which is why it's its own PR on top of the other one

But, I don't know why we have two type for this, AC only has one `VisitType` and it seems to me like we should have the same...

That said, I'm happy to dump this and just work with #4736 if this ends up being controversial or adds any type of complexity


tarikeshaq:
From the places stand up, I'll close this and open a ticket to do this once uniffication is done, I'll link back to this so it's easy to find!

mhammond:
FTR, [here are the equivalents on desktop](https://searchfox.org/mozilla-central/source/toolkit/components/places/nsINavHistoryService.idl#947)

PR #4741, opened 2021-12-19
uniffies viaduct

tarikeshaq:
Converts viaduct to use uniffi.

It ended up being more straightforward than I thought and I had the capacity so 🤷 

The biggest thing to consider is how this impacts desktop. Desktop uses viaduct, see https://searchfox.org/mozilla-central/source/toolkit/components/viaduct

My gut is telling me that I should add back all the deleted code here (pretty easy to add back) and put it under a cargo feature, let's say `gecko` and have it disabled by default. Then use conditional compilation to decide whether we use `uniffi` or the old way of doing things.

A few things before this is ready to land:

- [ ] Need to make sure desktop is well supported as I've mentioned above
- [ ] A bit more testing on fenix, tested it and it **seems** fine sending network requests, but I've changed a few things since and I wanna try a few edge cases (how are failures handled, slow network, responses/requests with large bodies, etc)
- [ ] Need to make sure the threading model here is OK, not 100% convinced yet, but I think it is.. just need to think it through a little


components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -4,140 +4,108 @@
 
 package mozilla.appservices.httpconfig
 
-import com.google.protobuf.ByteString
-import mozilla.appservices.support.native.RustBuffer
+import mozilla.appservices.httpconfig.uniffi.FfiBackend
```
mhammond, line 7: not a fan of the `Ffi` prefix here

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -4,140 +4,108 @@
...
+        headers.append(h.key, h.value)
+    }
+    return Request(
+        url = this.url,
+        method = this.method.into(),
+        headers = headers,
+        connectTimeout = Pair(this.connectTimeoutSecs.toLong(), TimeUnit.SECONDS),
+        readTimeout = Pair(this.readTimeoutSecs.toLong(), TimeUnit.SECONDS),
+        body = if (this.body != null) {
+            Request.Body(this.body!!.toByteArray().inputStream())
```
mhammond, line 43: is the `!!` necessary here? I've seen it omitted in these scenarios where you explicitly check against null.

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -4,140 +4,108 @@
...
+            // called after the client is set
+            client!!.value.fetch(request.into()).into(request.method)
         } catch (e: Throwable) {
-            LibViaduct.INSTANCE.viaduct_log_error("doFetch failed: ${e.message}")
-            // This is our last resort. It's bad news should we fail to
-            // return something from this function.
-            return RustBuffer.ByValue()
+            // Ideally, the catching of the error should happen in `uniffi`
+            // generated code and it should be converted to a proper Rust Error
+            // but until we add support for that, this should do
```
mhammond, line 107: I don't quite understand this comment - who needs to add support, and for what exactly? :) Also wondering if the `.into()`s should be outside the exception handler?

components/viaduct/src/backend/ffi.rs
```diff
@@ -37,160 +20,115 @@ macro_rules! backend_error {
...
-    });
-    error.consume_and_log_if_error();
-}
-
-#[no_mangle]
-pub extern "C" fn viaduct_initialize(callback: FetchCallback) -> u8 {
-    ffi_support::abort_on_panic::call_with_output(|| callback_holder::set_callback(callback))
+impl Backend for FfiBackend {
+    fn send(&self, request: crate::Request) -> Result<crate::Response, Error> {
+        super::note_backend("FFI (trusted)");
```
mhammond, line 130: I'm starting to see where the `Ffi` prefix came from, but I'm a little confused. We have a backend called "FFI", and the structs we have in the udl have `Ffi` as a prefix, but *all* backends return instances of these Ffi structs?

I wonder if the default backend should be called either `default` or `internal`? I could well be missing something here though

components/viaduct/src/viaduct.udl
```diff
@@ -0,0 +1,35 @@
...
+[Enum]
+interface FfiResponse {
+    Err(string message);
+    Ok(Method req_method, Url url, i32 status, sequence<i8>? body, record<DOMString, string> headers);
+};
+
+enum Method {
+    "Get", "Head", "Post", "Put",
+    "Delete", "Connect", "Options", "Trace"
+};
```
mhammond, line 35: nit: might as well add a \n here.

components/viaduct/Cargo.toml
```diff
@@ -18,7 +18,9 @@ log = "0.4"
...
 once_cell = "1.5"
-prost = "0.8"
-prost-derive = "0.8"
-ffi-support = "0.4"
 thiserror = "1.0"
+uniffi = "^0.15"
+uniffi_macros = "^0.15"
+
+[build-dependencies]
+uniffi_build = { version = "^0.15", features = [ "builtin-bindgen" ]}
```
mhammond, line 26: nit: might as well add a \n here.

mhammond (commented):
This is awsome :)

mhammond:
> The biggest thing to consider is how this impacts desktop. Desktop uses viaduct, see https://searchfox.org/mozilla-central/source/toolkit/components/viaduct

Ouch, hrmph. I'll dig deeper, but we should be aggressive here - it would be a shame to revert alot of this code and end up with even more moving parts than now and I don't think the usage is particularly widespread yet. cc @bendk

bendk (commented):
This looks amazing to me, but I don't know enough about viaduct to properly review it.

bendk:
> Ouch, hrmph. I'll dig deeper, but we should be aggressive here - it would be a shame to revert alot of this code and end up with even more moving parts than now and I don't think the usage is particularly widespread yet. cc @bendk

Like, if we had a C++ UniFII backend?  I feel like there's a red flag waving in front of me :)

I don't think this code changes very much and there's no pressing need to merge this.  What if we held it here in a PR for a while and see if we could get something working on desktop?

components/viaduct/src/backend/ffi.rs
tarikeshaq: Yeah... I don't like the `Ffi` prefix on the UDL actually - I'll change that up!

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
tarikeshaq: Good catch, the comment isn't clear at all 😅 I'll add a reference to https://github.com/mozilla/uniffi-rs/issues/351 and make it clearer!

tarikeshaq:
> I don't think this code changes very often and there's no pressing need to merge this. What if we held it here in a PR for a while and see if we could get something working on desktop?

Yeah I'm very happy to keep this on hold, maybe early in the new year we can explore what it would take to have the C++ code to consume this new version of viaduct 🤷

> it would be a shame to revert alot of this code and end up with even more moving parts than now and I don't think the usage is particularly widespread yet.

Usage isn't widespread yet! But glean on desktop is using it (not sure if there are other consumers though)

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
tarikeshaq: Yeah I'm not sure 😅 The compiler didn't let me have it with the null check so i added it, what Ben is saying could be true - although I don't expect the conversion of the **same** response  to happen on multiple threads, maybe there is some type of guarantee I can add, I'll explore 🤔

mhammond:
> But could you use something like this?

> body = this.body?.toByteArray().inputStream()

I nearly said the same thing, but I think that would be slightly tricky without changing semantics - the trick would be to avoid calling `Request.Body` with null (which would mean an extra null check)

components/viaduct/uniffi.toml
```diff
@@ -0,0 +1,8 @@
+[bindings.kotlin]
+package_name = "mozilla.appservices.httpconfig.uniffi"
+cdylib_name = "megazord"
+
+[bindings.swift]
+ffi_module_name = "MozillaRustComponents"
+ffi_module_filename = "viaductFFI"
+generate_module_map = false
```
jhugman, line 8: I don't understand the role of the swift entry here. Is the intention to move the `viaduct-reqwest` impl into this at some point?

jhugman (commented):
A few Kotlin suggestions, but overall looks a good incremental improvement to viaduct. (though removing Protobufs is huge).

tarikeshaq:
this has grown awfully stale, I'll close it. Any effort to uniffi viaduct will likely look different given the implication on desktop

PR #4746, opened 2021-12-21
Remove the manual places FFI entirely.

mhammond:
Build fails for me on Mac - let's see what CI says


skhamis (commented):
nearly 1.5k lines removed -- AMAZING. Always fun seeing all those protos deleted. LGTM! Verbally approving but will wait until Tarik has also had a chance to look at it!

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -163,71 +128,23 @@ class PlacesApi(path: String) : PlacesManager, AutoCloseable {
...
-        handle.set(connHandle)
-        conn = uniffiConn
         try {
-            interruptHandle = this.conn.newInterruptHandle()
+            interruptHandle = uniffiConn.newInterruptHandle()
         } catch (e: Throwable) {
-            rustCall { error ->
-                LibPlacesFFI.INSTANCE.places_connection_destroy(this.handle.getAndSet(0), error)
-            }
+            uniffiConn.destroy()
```
tarikeshaq, line 140: do we need this or will this be destroyed property once it's garbage collected (I am actually not sure 😅)

components/places/src/ffi.rs
```diff
@@ -23,19 +23,13 @@ use crate::ConnectionType;
...
-use ffi_support::{ConcurrentHandleMap, ErrorCode, ExternError};
 use parking_lot::Mutex;
 use sql_support::SqlInterruptHandle;
 use std::sync::Arc;
 use sync_guid::Guid;
 use types::Timestamp;
 use url::Url;
 
-lazy_static::lazy_static! {
-    pub static ref APIS: ConcurrentHandleMap<Arc<PlacesApi>> = ConcurrentHandleMap::new();
```
tarikeshaq, line 35: LOVE TO SEE THE HANDLES GONE 🎉 🚀

megazords/full/src/lib.rs
```diff
@@ -13,7 +13,6 @@ pub use crashtest;
 pub use fxa_client;
 pub use logins;
 pub use nimbus;
-pub use places_ffi;
```
tarikeshaq, line 16: should this be replaced with `places`? (dito in the `Cargo.toml`) I'm a little concerned that our CI didn't catch this 😅 unless it's somehow imported some other way? or do our android tests not run against the real megazord?

tarikeshaq (approved):
THE SHEER AMOUNT OF DELETED CODE 🎉 I'm so excited to see this land on main 🕺 🎉 🚀 💯 

I left a comment about the import for places missing in the megazord, but otherwise, this looks great to me!

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
mhammond: that's a great question :) I believe this just winds up dropping one reference to an `Arc<>`, so it *might* mean things are cleaned up as this is called, and might not. There's no equivalent on iOS, so I'm actually inclined to remove this - if it does turn out to be necessary here, that almost certainly means we need to work out how to do the same thing for iOS. WDYT?

megazords/full/src/lib.rs
mhammond: lol - nice spot - it turns out places still gets sucked in via sync-manager - so we could kill those other uses too! However, I think it's better to be clear. (FWIW, I did strike this exact same problem for iOS, but it doesn't use the sync manager, so the fix there was to do what you suggest!)

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
tarikeshaq: Sounds like a plan! either way, it seems like a small detail - I imagine `newInterruptHandle` failing is quite unlikely 😆

PR #4748, opened 2021-12-21
folders and separators now have non-nullable parents

mhammond:
Not a huge deal, but might as well represent the data model as accurately as possible.


skhamis (approved):
LGTM! 🚀

PR #4749, opened 2021-12-21
Only test all-features if features exist

tarikeshaq:
This should give us significant savings in testing/clippy times in both CI and locally through the `automation/tests.py` script

Most of our components don't have features at all, so it doesn't make sense to re-reun their tests with `--all-features`.

### DIFF
--- a/automation/tests.py
+++ b/automation/tests.py
@@ -131,6 +131,9 @@ class RustPackage:
     def has_default_features(self):
         return bool(self.cargo_metadata.get('features').get('default'))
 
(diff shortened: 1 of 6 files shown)


tarikeshaq:
Both `rust-tests` and `clippy` ran in `11` mins on this PR, when before they used to take **at least** 16-18 minutes (assuming all cache-hits!)

bendk (approved):
33% reduction in time and CPU cost?  Sounds good to me.

PR #4750, opened 2021-12-21
Refactors ios CI into multiple jobs

tarikeshaq:
This gives us significant time savings on our ios-artifacts jobs (cutting by more than half) by allowing three different jobs to run in parallel, the jobs are:
- the ios tests
- building the carthage framework
- building the xcframework

This is also nice for two other reasons:
1. Once we get rid of carthage (hopefully soon?) we can just trash the whole job since this decouples it
2. If we want to change the frequency of how often we build the frameworks (like, we really shouldn't be building those artifacts every CI run...) this makes it really easy to change, while keeping the `ios tests` running on every CI

One reason **not** to take this: now the jobs run in parallel... meaning we use more `macos` resources so costs go up (although even then it's not by much 🤷)... UNLESS, we do what point `2` above says, then it's a win/win...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -206,6 +206,28 @@ commands:
       - store_artifacts:
           path: /tmp/sccache.log
           destination: logs/sccache.log
+  setup-ios-environment:
+    steps:
+      - run:
+          name: Toggle brew auto-updates
+          command: |
+            if [ -z "${CIRCLE_TAG}" ]; then
+              # On non-release builds do not update brew (takes ages).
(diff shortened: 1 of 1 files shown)


tarikeshaq:
the iOS Ci on this PR took 8 minutes, when it used to take ~16 minutes! This improvement + #4749 should drop all of our CI jobs < ~12 mins 🕺  (assuming a good cache hit rate, which we consistently get unless the rust version upgrades or something)

skhamis (approved):
Nice! Thanks for splitting it out (and an amazing 50% improvement). Definitely not ideal we're producing both carthage and xcframeworks but we'll get some nice improvements once we can get rid of one. However, nothing related to this PR! 🚀

PR #4755, opened 2021-12-22
The number of node types is too damn high!

mhammond:
bookmarks.rs is getting way too complex - there's a public_node, a bookmark_node and various other nodes. Many of these nodes are just to do with serializing and importing via JSON - which isn't something we actually do very often. So we end up with a new module, `json_tree`, where all such json work is done to keep a nice separation. A nice side-benefit here is that uniffi consumers could now add an entire tree in one operation (although that wasn't the motivation and isn't actually used)

The patch is large, but most is mechanical. You may find it easier to review by looking at each of the commits individually.

After this, it should be relatively easy to kill `PublicNode` entirely - it's currently just used as in intermediate type and adds a layer of confusion we are better off without.


components/places/src/storage/bookmarks/json_tree.rs
```diff
@@ -0,0 +1,808 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This supports inserting and fetching an entire bookmark tree via JSON
+// compatible data structures.
+// It's currently used only by tests, examples and our utilities for importing
+// from a desktop JSON exports.
+//
+// None of our "real" consumers currently require JSON compatibility, so try
+// and avoid using this if you can!
+// (We could possibly put this behind a feature flag?)
```
skhamis, line 12: I would vote for this path. Feels like in the future it might be all too easy to want to just via JSON it and maybe making it more explicit will deter from that, but don't feel super strongly -- these comments feel enough for any potential future consumer.

components/places/src/storage/bookmarks.rs
```diff
@@ -1271,19 +1308,19 @@ pub fn delete_everything(db: &PlacesDb) -> Result<()> {
...
-        Some(guid) => guid,
-        None => return Err(InvalidPlaceInfo::InvalidParent("<no guid>".into()).into()),
-    };
-
-    let mut insert_infos: Vec<InsertableItem> = Vec::new();
-    add_subtree_infos(parent_guid, tree, &mut insert_infos);
-    log::info!("insert_tree inserting {} records", insert_infos.len());
+pub fn insert_tree(db: &PlacesDb, tree: FolderNode) -> Result<()> {
+    // This API is strange - we don't add `tree`, but just use it for the parent.
+    // It's only used for json importing, so we can live with a strainge API :)
```
skhamis: nit: `strange`

skhamis (approved):
Awesome! This is a great cleanup and I love the consolidation of bookmark-json related functionality into it's own space to make bookmarks.rs a lot more clearer. This indeed makes it easier to just kill public node which would be a huge win! 🚀

PR #4756, opened 2021-12-24
bookmarks: Remove public_node and introduce a fetch module

mhammond:
This seems like a good cleanup to me, and I don't think it's particularly risky. I think we should consider taking it as part of landing this branch - we will be on high-alert for regressions anyway, so landing this later just means we need to be on alert for longer. Happy to get pushback on this though.


components/places/src/ffi.rs
```diff
@@ -439,39 +437,40 @@ impl PlacesConnection {
...
-            let bookmark = bookmarks::public_node::fetch_bookmark(conn, guid, get_direct_children)?;
+            let bookmark = bookmarks::fetch::fetch_bookmark(conn, guid, get_direct_children)?;
             Ok(bookmark.map(BookmarkItem::from))
         })
     }
 
     fn bookmarks_get_all_with_url(&self, url: Url) -> Result<Vec<BookmarkItem>> {
         self.with_conn(|conn| {
-            let bookmarks = bookmarks::public_node::fetch_bookmarks_by_url(conn, &url)?
+            // XXX - We should move the exact type up to the consumers!
```
mhammond, line 447: I think these changes are better done in a followup though as they will be breaking (but an improvement!) for consumers (ie, currently they would need to explicitly ignore or otherwise handle folders and separators being returned here even though they never will be)

mhammond:
(IMO, a good future followup to this would be to split the "update" parts of bookmarks.rs into its own module too, which would leave bookmarks.rs much slimmer and cleaner.)

lougeniaC64 (approved):
Sorry for the delay. I was initially concerned about the changes to the `child_guid` bookmark item logic (and the changes to the related test assertions), but populating that property where it wasn't before doesn't seem problematic. LGTM!

PR #4758, opened 2021-12-29
Fix swift warnings re non-null parentGuid

mhammond:
I think I actually caused this in 6f6c4906e1391759acb9d2b930155ceda8aad426/#4748 but CI is hiding it and it's only a warning.


skhamis (approved):
Super weird that it was passing with warning then decided to start failing without the CI catching it....in any case. LGTM!

mhammond:
> Super weird that it was passing with warning then decided to start failing without the CI catching it

FWIW, these are just warning even in the failing CI job - the failing CI job seems to not want to show the output of the test that is actually failing, but I reproduce it locally, so all good!

PR #4759, opened 2021-12-29
[ci full] Nimbus using Glean Rust Bindings

travis79:
Okay, this roughly follows what Glean does for an iOS build, by creating a virtual environment for glean_parser and then translating the metrics.yaml.  This can probably be improved, and the .venv moved to a higher place in the tree once other components migrate to using Glean-Rust bindings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,6 +119,7 @@ commands:
     steps:
       - run: sudo apt-get update
       - run: sudo apt-get install python tcl
+      - run: sudo apt-get install python3-venv
       - run:
           name: Install NSS build system dependencies
           command: sudo apt-get install ninja-build gyp zlib1g-dev
--- a/components/nimbus/build.rs
+++ b/components/nimbus/build.rs
@@ -2,7 +2,29 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::{env, process::Command};
+
 pub fn main() {
     #[cfg(feature = "uniffi-bindings")]
     uniffi_build::generate_scaffolding("./src/nimbus.udl").unwrap();
+
+    // Run Glean Parser via tools/bootstrap_glean_rust.py python script
+    let out_dir = env::var("OUT_DIR").unwrap();
+    Command::new("python")
+        .args(&[
+            "../../tools/bootstrap_glean_rust.py",
+            "online",
+            "glean_parser",
+            "5.1.1",
+            "translate",
+            "-f",
+            "rust",
+            "-o",
+            out_dir.as_str(),
+            "./metrics.yaml",
+        ])
+        .status()
+        .expect("Error generating Glean Rust bindings");
+
+    println!("cargo:rerun-if-changed=./metrics.yaml");
 }
--- a/taskcluster/docker/linux/Dockerfile
+++ b/taskcluster/docker/linux/Dockerfile
@@ -91,6 +91,8 @@ RUN apt-get update -qq \
         p7zip-full \
         # Required to rsync the `libs` folder after fetch (see taskcluster/ci/android-build/kind.yml)
         rsync \
+        # Required for creating a venv for glean_parser
+        python3-venv \
     && apt-get clean
 
 RUN pip3 install --upgrade pip
--- a/tools/bootstrap_glean_rust.py
+++ b/tools/bootstrap_glean_rust.py
@@ -1,0 +1,39 @@
+#!python3
+# Python 3
+
+import importlib
+import subprocess
+import sys
+offline = sys.argv[1] == 'offline'
+module_name = sys.argv[2]
+expected_version = sys.argv[3]
+try:
+    module = importlib.import_module(module_name)
+except ImportError:
+    found_version = None
+else:
+    found_version = getattr(module, '__version__')
+if found_version != expected_version:
+    if not offline:
+        subprocess.check_call([
+            sys.executable,
+            '-m',
+            'pip',
+            'install',
+            '--upgrade',
+            f'{module_name}=={expected_version}'
+        ])
+    else:
+        print(f'Using Python environment at {sys.executable},')
+        print(f'expected glean_parser version {expected_version}, found {found_version}.')
+        sys.exit(1)
+try:
+    subprocess.check_call([
+        sys.executable,
+        '-m',
+        module_name
+    ] + sys.argv[4:])
+except:
+    # We don't need to show a traceback in this helper script.
+    # Only the output of the subprocess is interesting.
+    sys.exit(1)


travis79:
The problem I'm running into with this right now is how to get the correct version of the glean_parser into the build system in order to run the Glean codegen step. Right now it works locally because I can install it into a Python environment and build Nimbus where the right glean_parser is accessible.

components/nimbus/src/enrollment.rs
```diff
@@ -373,40 +375,72 @@ impl ExperimentEnrollment {
...
                 branch,
                 ..
-            } => EnrollmentChangeEvent::new(
-                &self.slug,
-                enrollment_id,
-                branch,
-                None,
-                EnrollmentChangeEventType::Enrollment,
-            ),
+            } => {
```
mhammond, line 378: This function already described itself as "create a telemetry event", but this patch now has the function both call `nimbus_events::enrollment.record` *and* returns the existing `EnrollmentChangeEvent`. I'm trying to get my head around why we seem to now have 2 things, both of which seem to be broadly described as "telemetry events", and this function both emits one type and returns another (presumably for the caller to emit)?

(Maybe there's a future where the "old" style are removed?)

components/nimbus/build.rs
```diff
@@ -2,7 +2,22 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::process::Command;
+
 pub fn main() {
     #[cfg(feature = "uniffi-bindings")]
     uniffi_build::generate_scaffolding("./src/nimbus.udl").unwrap();
+
+    Command::new("glean_parser")
```
mhammond, line 11: Do you see a future where we'd just have glean_parser as a dev dependency and just call a generate function like we do for uniffi? I'm just thinking about keeping the tool up to date, people not having PATH set, etc etc.

(ah - I see you mention you are having problems doing that, but I'm happy to see that's the intent!)

components/nimbus/src/glean_metrics.rs
```diff
@@ -0,0 +1,160 @@
+// -*- mode: Rust -*-
```
mhammond, line 1: Do we really need to check this in? I already hate glean because it always changes metrics.md line endings for no good reason, and having this checked in is almost certainly going to make me hate it even more. We manage to avoid the uniffi generated rust, so I'd strongly prefer to avoid glean generated rust.

mhammond (commented):
This is very interesting - thanks for sharing! I'd love to know plans for the future of this - ie, how much of the glean stuff in .kt files do you think can be moved to Rust? (I'm also happy to just bide my time if you plan on iterating in that direction!)

I'm really excited to see this - just don't quite have in my head yet exactly how it will help us in the other components, but it certainly looks like it's headed in the right direction!

components/nimbus/src/glean_metrics.rs
travis79: No, I added it to this draft simply to make it easier to review and see. I would agree that generated code shouldn't be added and once I get the particulars of how best to get the glean_parser step integrated into the build system I'll be sure and untrack it.

components/nimbus/build.rs
travis79: Yes, what I would really dream for is to update glean_parser to be written in Rust (or maybe partially replaced by UniFFI since they both accomplish the same thing, take a human-readable file and generate some code in various languages). That way it would be easy to include it as a dependency in the build and trigger it like `generate_scaffolding`.

components/nimbus/src/enrollment.rs
travis79: Yes, I have been working on some commits locally that remove the `EnrollmentChangeEvents` that get bubbled up across the FFI, but haven't pushed them up yet. I wanted to try and keep the commits in some logical order and I was distracted tinkering with how to get the glean_parser python script into the build steps.

travis79: That was a pretty big refactor because we end up using the events in a lot of tests to ensure things operated correctly, and that `enrollment.rs` file is getting to be nearly 4k lines long :(

components/nimbus/src/enrollment.rs
```diff
@@ -373,40 +375,72 @@ impl ExperimentEnrollment {
...
+                        branch: Some(branch.clone()),
+                        enrollment_id: Some(enrollment_id.clone().to_string()),
+                        experiment: Some(self.slug.clone()),
+                        reason: match reason {
+                            DisqualifiedReason::NotTargeted => Some("targeting".to_string()),
+                            DisqualifiedReason::OptOut => Some("optout".to_string()),
+                            DisqualifiedReason::Error => Some("error".to_string()),
+                        },
+                    }
+                );
```
jhugman, line 431: I'm confused about the differences in build and dependencies for iOS and Android (and Desktop). Is the intention here allow Glean to be conditionally compiled and then falling back to the recording events via existing foreign language bindings?

jhugman (commented):
driveby question

components/nimbus/src/enrollment.rs
travis79: Sorry, this has bit-rotted a bit and doesn't reflect conversations that we have had regarding how to go about letting iOS record in Rust while Android still uses the existing Glean Kotlin interface.

This was originally done to allow me to test the Glean changes I had made in a downstream consumer. I don't think we will need to worry about conditionally compiling Glean depending on platform, Glean "should just work" regardless of what language binding is being used, even if they are mix and matched in the same product (Nimbus recording in Rust while Fenix is recording in Kotlin or iOS in Swift, etc).

travis79: What I don't know about is how we know we are working with the same version of Glean coming from GeckoView...

components/nimbus/build.rs
travis79: In the future this will probably get easier...   like once this gets more time and love: https://github.com/badboy/glean-parser

.circleci/config.yml
```diff
@@ -119,6 +119,7 @@ commands:
     steps:
       - run: sudo apt-get update
       - run: sudo apt-get install python tcl
+      - run: sudo apt-get install python3-venv
```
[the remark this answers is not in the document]
travis79: Good call, I didn't think about updating the development setup docs. In this case, if they have python installed they _should_ have venv installed. This was only necessary because the Debian images we use in CI don't set a flag (ensurepip) which means that you have to explicitly install these packages.

components/nimbus/build.rs
```diff
@@ -2,7 +2,50 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::process::Command;
+
 pub fn main() {
     #[cfg(feature = "uniffi-bindings")]
     uniffi_build::generate_scaffolding("./src/nimbus.udl").unwrap();
+
+    // Check if we have a Python virtual environment set up yet
+    let venv_exists = Command::new("[")
```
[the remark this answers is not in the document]
travis79: Yeah, it is relying on `bash` syntax.  Let me break out a Windows machine and test it. (It may just mean it needs built in the linux subsystem for windows, maybe?)

components/nimbus/Cargo.toml
```diff
@@ -35,6 +35,7 @@ once_cell = "1"
 uniffi = { version = "^0.17", optional = true }
 chrono = { version = "0.4", features = ["serde"]}
 unicode-segmentation = "1.8.0"
+glean = "42.3.2"
```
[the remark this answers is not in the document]
travis79: The Swift code and the Rust code should both be using the same Glean xcframework on iOS. I think that gets resolved when the app compiles? (looking into it to be sure)

components/nimbus/build.rs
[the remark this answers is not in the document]
tarikeshaq: Yeah I wouldn't worry too much about it working for normal windows (WSL is what the windows developers on our team use IIRC) so doing the `ci full` check to make sure that taskcluster can build the windows artifacts is probably good enough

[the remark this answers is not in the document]
mhammond: yeah, I'm fine with ignoring windows in the shortish term. This does look slightly fragile though - what "python3" exactly? I think it might be slightly better to have a .py script doing some of this, so the .rs code has less python magic - "bootstrap_glean.py" with a `#!` line that specifies python3 and let it do this cruft, and the rust code just executes it via the shell? That even has a good chance of working on Windows if the "python launcher" is installed :) Regardless though, I don't think that need block this.

mhammond (commented):
This looks great, but confusingly small :) I can't see much here that's nimbus specific - what would stop all of app-services doing the same thing? ie, what am I missing in how this works?

travis79:
@mhammond Yes, it was a lot smaller than I expected, and you are right, nothing is stopping all of A-S from using this now 😄!  

I like the idea of capturing this step in a .py script to remove some of the command line magic from the Rust build script, and make it more reusable by other components. I'll file an issue to iterate on that as an improvement to this. Once other components are using Glean-Rust, it will be useful to reuse the venv where glean_parser is installed too.

badboy:
> I like the idea of capturing this step in a .py script to remove some of the command line magic from the Rust build script, and make it more reusable by other components. I'll file an issue to iterate on that as an improvement to this. Once other components are using Glean-Rust, it will be useful to reuse the venv where glean_parser is installed too.

We already sorta _have_ that Python script, just inlined in the Groovy plugin or shell script for iOS.
I'd rather see that used here, instead of iterating on it later.

The whole Glean-through-Gecko(View) part could actually be an issue, but I haven't spend proper time thinking about it yet.

badboy:
To make this more clear: The Glean-through-GeckoView-and-how-it-all-works part is a blocking concern for this PR.
 I plan to speak with Travis next week so we can move this forward.

components/nimbus/build.rs
```diff
@@ -2,7 +2,29 @@
...
 
+use std::{env, process::Command};
+
 pub fn main() {
     #[cfg(feature = "uniffi-bindings")]
     uniffi_build::generate_scaffolding("./src/nimbus.udl").unwrap();
+
+    // Run Glean Parser via tools/bootstrap_glean_rust.py python script
+    let out_dir = env::var("OUT_DIR").unwrap();
+    Command::new("python")
```
badboy, line 13: Wait, this is not creating a virtual env now anymore anywhere, does it?
We could actually put the virtual env into the target directory. It's somewhere up from `OUT_DIR` I guess.

travis79:
As discussed in last week's Nimbus SDK meeting, I'd like to land this much of the integration that doesn't instrument anything in Rust yet. Once we have solved the problem of linking things on iOS and Android, I have a PR waiting to add the instrumentation.

cc @jhugman can I get an approving review to land this please?

components/nimbus/build.rs
```diff
@@ -2,7 +2,29 @@
...
+            "5.1.1",
+            "translate",
+            "-f",
+            "rust",
+            "-o",
+            out_dir.as_str(),
+            "./metrics.yaml",
+        ])
+        .status()
+        .expect("Error generating Glean Rust bindings");
```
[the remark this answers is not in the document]
travis79: Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1767972)

components/nimbus/Cargo.toml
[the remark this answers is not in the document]
travis79: Looks like it may require Glean to be in the iOS megazord to be able to actually record things. That's a blocker to recording metrics, but doesn't break anything with building or the integration here.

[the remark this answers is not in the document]
travis79: I'm adding that as a blocker on migration of the existing events to Rust and adding the health metrics.

components/nimbus/build.rs
travis79: Yes, once there are more than just Nimbus using the virtual environment, it will be good to have it shared among the other components.

components/nimbus/Cargo.toml
tarikeshaq, line 38: hmmm I don't feel too strongly about this, but what do you think about removing this and adding it back when we can actually consume the dependency (once we have it in iOS). That will keep the scope of this PR on the `metrics.yaml` generation from build scripts and will save us the unused dependency

Regardless I don't mind keeping it, I checked the binary sizes as well and given our megazords, it's a very small increase

tarikeshaq (approved):
Verified locally that this can be built on both firefox iOS and fenix without breaking. I recall @travis79 got a focus-ios build working as well (right? otherwise I can run through that quickly tomorrow) focus-android imports us exactly the same way as fenix so that is safe too.

This means that this is safe to land without breaking our consumers, which was my biggest concern since they already have their own glean binaries and I was afraid of linker errors 😶 

The `glean` Rust dependency isn't used anywhere and will be bloat considering our consumers have their own "glean" binaries.

In a perfect world we would only add the dependency when we actually can consume it, but I don't feel strongly about this, the difference in the binary size is small and if having it in there now helps us move faster into being able to use the rust glean api, then all the better!

components/nimbus/Cargo.toml
travis79: I'd be perfectly fine with that, given that it is very easily added back.

Issue #4762, opened 2022-01-05
database is locked errors


bendk:
We're seeing a good number of errors when trying to open the places DB for syncing: https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/16148794/

This particular version of the error is new, but I think it's been happening for a while now.  It's one of the errors listed in https://github.com/mozilla/application-services/blob/72b827c3e0f883163762857fd766df1aeb060725/docs/design/db-fatal-error-handling.md

After some investigation, this seems to be what's happening:
  - We create a sync connection and  the initialization code starts a read transaction
  - In another thread, the read/write connection performs a write (maybe just starting a write is also enough)
  - The sync connection initialization code continues and starts a write transaction

This will result in an `SQLITE_BUSY` error as described in https://sqlite.org/isolation.html:

> Another example: X starts a read transaction using BEGIN and SELECT, then Y makes a changes to the database using UPDATE. Then X tries to make a change to the database using UPDATE. The attempt by X to escalate its transaction from a read transaction to a write transaction fails with an SQLITE_BUSY_SNAPSHOT error because the snapshot of the database being viewed by X is no longer the latest version of the database. If X were allowed to write, it would fork the history of the database file, which is something SQLite does not support. In order for X to write to the database, it must first release its snapshot (using ROLLBACK for example) then start a new transaction with a subsequent BEGIN. 
>
>  If X starts a transaction that will initially only read but X knows it will eventually want to write and does not want to be troubled with possible SQLITE_BUSY_SNAPSHOT errors that arise because another connection jumped ahead of it in line, then X can issue BEGIN IMMEDIATE to start its transaction instead of just an ordinary BEGIN. The BEGIN IMMEDIATE command goes ahead and starts a write transaction, and thus blocks all other writers. If the BEGIN IMMEDIATE operation succeeds, then no subsequent operations in that transaction will ever fail with an SQLITE_BUSY error.

I think the suggestion in the second paragraph is easiest to implement.  We just need to make sure that all transactions started by the sync and read/write connections use `BEGIN IMMEDATE` to start their transactions.  The reader connections can continue to use ordinary `BEGIN` and we can keep the basic concurrency story desrcibed in https://github.com/mozilla/application-services/blob/main/components/places/doc/sql_concurrency.md

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/SYNC-2983)
┆epic: Sync Ecosystem (backlog)
[...]

PR #4763, opened 2022-01-05
Attempt at fixing the `database is locked` error (#4762)

bendk:
Added a places-utils test that opens sync connections in threads.  All
this needed to error out was:
  - The database was already created
  - 2 threads tried to open a sync connection. The Mutex in PlacesApi
    ensures that this happens sequentially, but if the second thread
    tried to immediately open the connection after the first one, then I
    would see an error.  Adding a 10ms delay in-between seemed to fix
    things.

After some investigation I narrowed it down to the `INSERT OR IGNORE`
statements inside the `create_synced_bookmark_roots()` call.  I think
the fact that that query is both a read and write causes some extra
locking to happen. I'm not sure exactly what's going on here.  Maybe the
disk needs to be flushed after the first connection closes the DB, but
`PRAGMA database.wal_checkpoint` didn't work.

In any case, it seems like we could create the roots when we are first
initializing the schema with the read/write connection so I just moved
the call there and it seemed to fix things for me.

  - Note:

### DIFF
--- a/components/places/src/db/schema.rs
+++ b/components/places/src/db/schema.rs
@@ -73,6 +73,7 @@ pub fn init(conn: &Connection) -> rusqlite::Result<()> {
     log::debug!("Initializing schema");
     conn.execute_batch(CREATE_SHARED_SCHEMA_SQL)?;
     create_bookmark_roots(conn)?;
+    create_synced_bookmark_roots(conn)?;
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


bendk:
I don't think it's correct to move the `create_synced_bookmark_roots()` call around like that, so I'm closing this one.

PR #4764, opened 2022-01-06
Attempt at fixing the `database is locked` error (#4762)

bendk:
The error was happening while opening the database so I added a
places-utils test to stress-test opening sync connections.  I was able
to reproduce the error and tracked down the source to the `INSERT OR
IGNORE` statements inside the `create_synced_bookmark_roots()` call.

The error happens when attempting to get a write lock for a transaction,
after the it already has a read lock.  This is causing SQL_BUSY errors
(which has a misleading "database is locked" error message).

This issue happened in two different ways:
  - When a separate thread held a write lock on the writer DB.  Note
    that the thread only held the lock for 2 seconds, which is short
    enough that other threads should just wait for the lock rather than
    timing out.

  - When multiple threads tried to open a sync connection at once.
    Because get_sync_connection uses a mutex this happened sequentially,
    but it still triggered the issue if open attempts happened
    immediately after the connection was dropped. Adding a 10ms sleep
    stopped the error.

Interestingly, if we used `BEGIN IMMEDIATE` to start the transaction,
which immediately tries to take the write lock, then the issue wouldn't
appear. The thread would just wait until the lock was available.  The
issue only happened when we took the read lock first, then tried to take
a write lock.

Fixed this by using the SQLITE_OPEN_SHARED_CACHE flag when opening the
connection.  This changes how sqlite handles locking and fixes the
problem.  It also seems good in general.  The sqlite site says it "can
significantly reduce the quantity of memory and IO required by the
system".

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,6 +18,10 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+## Places
+### What's Changed
+- Switched to using sqlite shared-cached mode for connections to avoid "database is locked" errors ([#4764](https://github.com/mozilla/application-services/pull/4764))
+
 ## viaduct
 ### What's New
 - Add support for PATCH methods. ([#4751](https://github.com/mozilla/application-services/pull/4751))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -893,6 +893,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "sql-support",
  "structopt",
  "sync-guid",
  "sync15",
--- a/components/places/doc/sql_concurrency.md
+++ b/components/places/doc/sql_concurrency.md
@@ -33,3 +33,7 @@ However, this *will* still result in `SQLITE_BUSY` being possible. The 2 scenari
 * One of our "small and fast" writes isn't as small and fast as we expect - the sync writer will see `SQLITE_BUSY`. This generally isn't an actual problem - it's "just" a failed sync and the next sync might work fine.
 
 * Even with our "1000ms transaction budget", some other sync transactions still end up taking more than our timeout. This will result in one of the "small and fast" writes failing. Without building a retry queue, we may need to accept this data-loss - but we should work out strategies for understanding how often it happens.
+
+## Shared cache mode
+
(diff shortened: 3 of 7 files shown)


skhamis:
Nice dive into this! Was a good read about `SQLITE_OPEN_SHARED_CACHE` and it's benefits. One thing got me thinking as you mentioned the issue is actually read/write trying to access the database and not necessarily multiple connections trying to write to the same resource (I think PlacesConnection does a lot of work to ensure there is ever only 1). Is potentially using [WAL mode for sqlite](https://www.sqlite.org/wal.html). Specifically mentions:

> - WAL is significantly faster in most scenarios.
> - **WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.**
> - Disk I/O operations tends to be more sequential using WAL.
> - WAL uses many fewer fsync() operations and is thus less vulnerable to problems on systems where the fsync() system call is broken.

However I think the current approach is fine but giving some food for thought!

bendk:
I'm pretty sure we're using WAL mode right now for all those reasons.  It definitely seems like the way to go.

mhammond (approved):
Thanks for the example. I gotta say this makes me slightly nervous - I'm quite surprised that google can't find other people noting the same thing.

Further, https://www.sqlite.org/c3ref/enable_shared_cache.html has this quote:
"Shared cache is disabled by default. It is recommended that it stay that way. In other words, do not use this routine. ... Any use of shared cache is discouraged. If shared cache must be used, [use] the SQLITE_OPEN_SHAREDCACHE flag."

But - that change to the example is compelling and I don't have any other ideas. Landing this in the same release as the places uniffication is also going to make life interesting for us, but I also don't have a good alternative to that (I don't think waiting another release serves anyone)

So, fingers crossed :)

Thanks!

mhammond:
A changelog entry might be worthwhile too?

bendk:
I think your nervousness is correct and this isn't the right way to try to fix this.  Doing some further digging on this one:

 - It's expected behavior to get an immediate `SQLITE_BUSY` when trying to upgrade a read transaction to a write transaction, when another write transaction is in-progress, or completed after the read transaction started:
   - See section 2.1 of the [transactions doc](https://sqlite.org/lang_transaction.html)
   - Here's an explanation of the general issue: https://sqlite.org/forum/forumpost/e51c0e031ceb51dc (I think this was about non-WAL mode, but the same basic issue applies to WAL mode)
  - Reading through the sqlite forum, the general consensus is that shared cached mode is bad because it limits concurrency.  Maybe that fixes this issue, but it could have a very negative impact on performance.
  - I think the best fix would be to update the code so that whenever the `ReadWrite` or `Sync` connections get a transaction, they always use `BEGIN IMMEDIATE`.  This makes them get the write lock immediately, which avoids the deadlock.

PR #4780, opened 2022-01-17
Adds swift FML code generation

tarikeshaq:
fixes [EXP-1984](https://mozilla-hub.atlassian.net/browse/EXP-1984)

- [x] Allows running swift scripts against FML generated code
- [x] Allows running swift script against a mock nimbus feature variables API 
- [x] Generate swift FML code similar to how Kotlin does it
- [x] Write tests to cover at minimum what the Kotlin tests cover
- [x] Clean up `FeatureVariables` API so UIKit is imported again
- [x] Have the "real" nimbus API use `FeatureInterface"
- [x] Clean up code and TODOs in comments
- [x] Make unit tests pass
- [x] Rebase
- [x] Test against Firefox iOS to make sure there are not breaking changes
- [x] Changelog


components/nimbus/ios/Nimbus/FeatureVariables.swift
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
-import UIKit
+//import UIKit
```
tarikeshaq, line 6: Meh I'll have to think about this harder, I want to use the Variables API in my test scripts, but those run on my Mac... `UIKit` is only for iOS. Either we mock it, or we replace `UIImage` with a protocol and have tests mock that protocol (while real code uses `UIImage`)... Not sure yet, I'll get back to this at a later point when I'm trying to clean things up

components/support/nimbus-fml/fixtures/fe/nimbus_features.yaml
```diff
@@ -27,8 +27,14 @@ features:
         description: This is the map of the player type to a button
         type: Map<PlayerProfile, Button>
         default: {
-          "child": {},
-          "adult": {}
+          "child": {
```
tarikeshaq, line 30: So my default validation that I implemented a few weeks ago works 😁 It caught this missing default values for me

components/support/nimbus-fml/fixtures/ios/tests/FeatureManifest.swift
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+public func add(a: Int, b: Int) -> Int {
```
tarikeshaq, line 5: this is to smoke test the compilation + running of scripts, this has nothing to do with the FML itself

components/support/nimbus-fml/src/backends/swift/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+            .ok_or_else(|| FMLError::InvalidPath(manifest_file.to_string_lossy().into_owned()))?;
+        let temp = TempDir::new(prefix)?;
+        let build_dir = temp.path();
+        compile_manifest_swift(manifest_file, build_dir)?;
+        run_script(build_dir, script)
+    }
+
+    #[test]
+    fn smoke_test_script() -> Result<()> {
+        run_script_with_generated_code("fixtures/ios/tests/FeatureManifest.swift".as_ref(), "fixtures/ios/tests/script.swift".as_ref())
```
tarikeshaq, line 157: This runs to make sure that the testing scripts have access to generated code + the MockNimbus Api

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -205,6 +208,17 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn smoke_test_ios_generate() -> Result<()> {
+        generate_and_assert(
```
tarikeshaq, line 213: this smoke tests that code generated from our `workflows` can be accessed from our tests. I'll replace this with multiple tests as the code generation progresses

tarikeshaq (commented):
Early notes, I think I have the testability of this setup nicely. Scripts can run and access the `MockNimbus` API. I have to think a little about the UIKit dependency, but I can do that later when this PR is in a ready state 🤷

jhugman (commented):
Small 🚲 🚗 🚜 🚌  Drive By review.

The Rust looks boring and regular; good job; otherwise, I haven't commented much.

I've added a few comments about the Swift. Keep going looks nice.

components/nimbus/ios/Nimbus/FeatureVariables.swift
```diff
@@ -332,7 +332,11 @@ internal class JSONVariables: VariablesWithBundle {
     }
     
     func asVariablesMap() -> [String: Variables]? {
-        return nil
+        let j = json.mapValues() {(value) in
+            // TODO: filter out any possible json that can't be cast to the dictionary
```
travis79, line 336: What are you going to do with the json that can't be cast to a dictionary?

components/nimbus/ios/Nimbus/FeatureVariables.swift
```diff
@@ -332,7 +332,11 @@ internal class JSONVariables: VariablesWithBundle {
...
+            return JSONVariables(with: value as! [String: Any])
```
travis79, line 337: you could use `as?` here to prevent crashing and filter out the values...  e.g.:
```
if let jsonDict = value as? [String: Any] {
  ...
} else {
  ...
}

**Review by travis79 — COMMENTED** [review]
Drive-by review (I just read James' review and I didn't have much to add beyond his suggestions to make the code a little more "swifty").

File: components/nimbus/ios/Nimbus/FeatureVariables.swift (hunk 6 above)
Comment (tarikeshaq) [reply components/nimbus/ios/Nimbus/FeatureVariables.swift]: It will be omitted! I can't think of a case where this can happen, but I went for the least destructive response if it does 😬

**Comment by tarikeshaq** [comment]
OK tested against firefox-ios locally and verified there are no breaking changes for Nimbus (had to test against the places_uniffication and the glean update patches since those are on main here, but regardless, this patch doesn't introduce a breaking change)

also added the changelog so this should be in a good place for review 😄

File: components/support/nimbus-fml/fixtures/fe/ios.yaml [hunk 8]
```diff
@@ -0,0 +1,134 @@
...
+          recently-saved:
+            description: The sites the user has bookmarked recently.
+          jump-back-in:
+            description: The tabs the user was looking immediately before being interrupted.
+          recent-explorations:
+            description: The tab groups
+          pocket:
+            description: The pocket section. This should only be available in the US.
+          library-shortcuts:
+            description: Library shortcuts section
```
travis79, line 134: nit missing blank line at end of file

components/support/nimbus-fml/fixtures/ios/runtime/UIImage.swift
```diff
@@ -0,0 +1,14 @@
...
+public class UIImage {
+    private var named: String
+    private var bundle: Bundle
+    private var compatibleWith: String?
+    public init?(named: String, in bundle: Bundle, compatibleWith: String?) {
+        self.named = named
+        self.bundle = bundle
+        self.compatibleWith = compatibleWith
+    }
+}
```
travis79, line 14: nit: missing newline at EoF

components/support/nimbus-fml/fixtures/ios/tests/FeatureManifest.swift
```diff
@@ -0,0 +1,7 @@
...
+    return a + b
+}
```
travis79, line 7: nit: missing newline at EoF

components/support/nimbus-fml/fixtures/ios/tests/script.swift
```diff
@@ -0,0 +1,11 @@
+import FeatureManifest
+
+assert(add(a: 1, b: 2) == 3)
+
+let mn = MockNimbus(("a", "{ \"a\": 12 }"))
```
travis79, line 5: Nice! This is neat!

travis79 (approved):
Just a handful of minor nits, but this looks fantastic!  Well done!

components/nimbus/ios/Nimbus/FeatureInterface.swift
```diff
@@ -0,0 +1,57 @@
...
+/// A small protocol to get the feature variables out of the Nimbus SDK.
+///
+/// This is intended to be standalone to allow for testing the Nimbus FML.
+public protocol FeaturesInterface: AnyObject {
+    /// Get the currently enrolled branch for the given experiment
+    ///
+    /// - Parameter featureId The string feature id that applies to the feature under experiment.
+    /// - Returns A String representing the branch-id or "slug"; or `nil` if not enrolled in this experiment.
+    ///
+    func getExperimentBranch(experimentId: String) -> String?
```
jhugman, line 15: I don't think this method needs to be here.

components/support/nimbus-fml/src/backends/swift/templates/EnumTemplate.swift
```diff
@@ -0,0 +1,17 @@
...
+    {% for variant in inner.variants() %}
+    {{ variant.doc()|comment("    ") }}
+    case {{ variant.name()|enum_variant_name }} = {{variant.name()|quoted}}
+    {% endfor %}
+
+    public static func enumValue(_ s: String?) -> {{class_name}}? {
+        if let s = s {
+            return {{class_name}}(rawValue: s)
+        }
+        return nil
```
jhugman, line 15: `if let` is a `nil` check, so it's redundant here.

```suggestion
        return {{class_name}}(rawValue: s)
```

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
```diff
@@ -0,0 +1,50 @@
...
+{%- for imported_module in self.imports() %}
+import {{ imported_module }}
+{%- endfor %}
+
+{% let nimbus_object = self.config.nimbus_object_name() -%}
+///
+/// An object for safely accessing feature configuration from Nimbus.
+///
+/// This is generated.
+public struct {{ nimbus_object }}Struct {
```
jhugman, line 16: Why add `Struct` here?

jhugman: Oh, I think I see. Do you think we should use statics here instead?

```swift
struct {{ nimbus_object }} {
    public static var api: FeaturesInterface? = nil
    public static let features = Features()
}
```

components/support/nimbus-fml/src/backends/swift/templates/FeatureTemplate.swift
```diff
@@ -0,0 +1,44 @@
+{%- import "macros.swift" as swift %}
+{%- let inner = self.inner() %}
+{%- let raw_name = inner.name() %}
+{% let class_name = inner.name()|class_name -%}
+
+{{ inner.doc()|comment("") }}
+public class {{class_name}} {
+    private var _variables: Variables
```
jhugman, line 8: This is smart here: this will make it easier later on when adding `Text` and `Image`.

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
```diff
@@ -0,0 +1,50 @@
+// This file was autogenerated by some hot garbage in the `nimbus-fml` crate.
+// Trust me, you don't want to mess with it!
+
+// This file was autogenerated by some hot garbage in the `nimbus-fml` crate.
+// Trust me, you don't want to mess with it!
```
jhugman, line 5: We can be _too_ self-deprecating, I think.

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
jhugman: Or, we follow the swift version of a singleton.

```swift
public struct {{ nimbus_object }} {
    public static let shared = { {{ nimbus_object }}() }()
}
```

I think I prefer this answer, but unsure if we use `shared`, `instance`, or `default` as the name.

components/nimbus/ios/Nimbus/Collections+.swift
```diff
@@ -0,0 +1,38 @@
...
+        let transformed: [(K1, V1)] = compactMap { k, v in
+            guard let k1 = keyTransform(k),
+                  let v1 = valueTransform(v)
+            else {
+                return nil
+            }
+            return (k1, v1)
+        }
+        return [K1: V1](uniqueKeysWithValues: transformed)
+    }
```
jhugman, line 29: I'm not sure about these kotlin names; in general I've tried to keep naming schemes familiar to the platform, but this is all very internal.

components/support/nimbus-fml/fixtures/ios/tests/script.swift
```diff
@@ -0,0 +1,11 @@
+import FeatureManifest
```
jhugman, line 1: This is … suprising. I'm not sure I understand why this is necessary.

jhugman (approved):
Lovely.

I've left some comments which should be addressed, but nothing to hold up another review cycle.

components/support/nimbus-fml/fixtures/ios/tests/script.swift
tarikeshaq: This is only for the tests, the reason is that the scripts are run after the FeatureManifest, and all the Nimbus helpers are compiled together into a module that's later imported by the script. In the app, the FeatureManifest would be compiled together with the app so this wouldn't be needed

components/nimbus/ios/Nimbus/FeatureInterface.swift
tarikeshaq: Moved into its own protocol to not break iOS and added a note on it that consumer should think about switching to the variables API!

components/support/nimbus-fml/src/backends/swift/templates/EnumTemplate.swift
tarikeshaq: Yeah, the nil check is intentional here because the `enumValue` in swift takes a `String?`, since the `v.getString("the-property")` returns a `String?`

PR #4782, opened 2022-01-18
Use BEGIN IMMEDIATE in `open_database`

bendk:
This implements the fix for #4762 that I suggested in the description after thinking through some of @mhammond comments:  

> We just need to make sure that all transactions started by the sync and read/write connections use BEGIN IMMEDATE to start their transactions.

After checking out the code some more, I realized that the `places::db::tx` mod was already trying to do that.  In particular, the sync connection uses [`get_tx_with_retry_on_locked()`](https://github.com/mozilla/application-services/blob/main/components/places/src/db/tx/coop_transaction.rs#L208) to create a transaction, which always uses `TransactionBehavior::Immediate`.

So the only change here is to change the `open_database` code to also use `TransactionBehavior::Immedate`.  This affects all components, not just places, but I think BEGIN IMMEDIATE makes sense in general.  Usually there should only be one connection accessing the database when it's initialized, which means it won't matter.  If we do have another
connection accessing the database as we initialize it, I think it's worth taking a small hit in concurrency in order to eliminate the possibility of SQL_BUSY errors in cases like this.

  - Note:


mhammond (approved):
Oops, sorry for the delay here!

PR #4790, opened 2022-01-20
Implementing syncing interruption for real

bendk:
Added the `interrupt_scope::InterruptScope` type which provides a
simplified API for interrupt handling.  One difference between this and
`SqlInterruptScope` is the counter is a static variable, this allows us
to create a new `InterruptScope` whenever we want, which simplied
refactoring some of the older code.

Updated the `SyncEngine` implemenations so each one stores an
`InterruptScope` to check.  Removed older/unused interruption code.

For `SyncManager` methods, where we create multiple sync engines, the
`SyncManager` creates one scope, then passes a clone to each
`SyncEngine`. This avoids any chance of some engines getting interrupted
and some not.

One thing that `InterruptScope` doesn't handle is SQL interruption.
Added places code to handle that since it's the only component that has
a dedicated sync connection.  At some point we could update the other
components to match and standardize this somehow, but for now it seems
okay to special-case places.

Made the code to use `interrupt_scope` as the field/argument name.  The existing
code was split between this one and `scope`.  I figured the extra descriptive
value was worth the extra chars.

There's still some `PlacesDb` code that uses the older `SqlInterruptScope` and
`SqlInterruptHandler` classes.  I left that alone because I think it's used for
interrupting autocomplete and I didn't want to get into that code at this
point.

  - Note:


components/support/interrupt/src/scopes.rs
```diff
@@ -0,0 +1,90 @@
...
+///      or `err_if_interrupted()` which is often more ergonomic.
+///    - Call `InterruptScope::interrupt_current_scopes()` to interrupt all previously created
+///      `InterruptScope`s. `InterruptScope`s created after this call will not be considerid
+///      interrupted unless `InterruptScope::interrupt()` is called again.
+///
+/// This type requires the code to actively if it's interrupted.  Therefore:
+///   - Make sure to sprinkle in `err_if_interrupted()` calls inside the code you want
+///     interruptable.  Loops are a particularly good place to put these.
+///   - This cannot interrupt external code.  In particular, it won't interrupt a long-running SQLite
+///     query.  Use an `rusqlite::InterruptHandler` for that.
```
mhammond, line 39: So this is what I don't quite understand in this patch - how does interrupting a sync also arrange to interrupt any running sql, other than that special places support?

components/support/interrupt/src/scopes.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+impl Default for InterruptScope {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+// Needed to make the old sync_multiple code work.  Let's remove this once all syncing goes through
```
mhammond, line 84: Who actually uses `sync_multiple`? Doesn't look like iOS does? Our examples do, but we can hack something together for them (eg, I don't care if they are not interruptable TBH)

components/sync_manager/src/manager.rs
```diff
@@ -30,42 +30,71 @@ impl SyncManager {
         SyncEngineId::try_from(engine_name).map_err(SyncManagerError::UnknownEngine)
     }
 
-    fn get_engine(engine_id: &SyncEngineId) -> Option<Box<dyn SyncEngine>> {
+    // Interrupt any active sync operations
+    pub fn interrupt(&self) {
+        InterruptScope::interrupt();
+        places::interrupt_current_sync_query();
```
mhammond, line 36: eg, I can see here how the places sync connection is interrupted, but what do we do for other engines? How does their sql connection get interrupted?

mhammond (commented):
Thanks for this! This is a great patch and I agree with your decision on the names etc even though it makes the patch scarily large :) I'm out of time for today and am off tomorrow, so wanted to offer some early feedback - many of these comments are more questions for me rather than for you, so please don't read too much into them.

In particular though, I'm not sure how this interrupts the DB operations in the general case. This might actually be impossible though - as you note, the same connection is used for some engines for both syncing and the "real" API - so how do we know what we are interrupting? This API probably does need to handle that in some way though - I think the only real use-case for this "general-purpose" interrupt code is shutdown, and in that scenario we probably just want to interrupt all the things? OTOH though, we also agreed that Android/iOS just killing us seems likely too - it's really only desktop where this would matter (and indeed, that probably does matter today for webext-storage)

I wonder if another approach for this "general purpose" approach is to explicitly declare it's expected to be a 1-shot operation explicitly for shutdown - ie, you can't just interrupt a "logins" sync - you are interrupting "the world" because you want to terminate. For use-cases where this doesn't work (eg, auto-complete), then just don't use this and use a separate mechanism (and in that specific example, the interruption is internal so isn't exposed). Then, the fact that we don't know whether we might be interrupting a "real" API call versus a sync is actually a feature.

As above though, take this with a grain of salt - I might well be missing something. Let's chat a little more about this when we catch up in a couple of days.

mhammond:
(eg: a one-shot boolean operation might also help guard against clients who don't do the right thing at shutdown - eg, imagine a client which - say - calls `manager.sync();`, catches exceptions thinking they might be transient, then does something else with our components. Having all future operations fail after a single interrupt might actually be a feature because it will be largely impossible for a client to keep doing things even after the interrupt.)

components/support/interrupt/src/scopes.rs
bendk: Long-term, I'd love to make more components work like places:  Have a connection dedicated to syncing, so we can always interrupt it.  I guess the other option would be to ignore the issue.  If the other components don't have long-running queries, maybe that's an option.

components/support/interrupt/src/scopes.rs
bendk: I think iOS uses it indirectly through methods like `LoginsStore.sync()`.  Maybe we could refactor those functions somehow though.

components/places/src/db/db.rs
```diff
@@ -207,6 +207,38 @@ impl Deref for PlacesDb {
     }
 }
 
+/// PlacesDb that's behind a mutex
+///
+/// The one extra feature that this provides is interrupting the connection without having to take
+/// the mutex lock.
+pub struct MutexPlacesDb {
```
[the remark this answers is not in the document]
bendk: The name is definitely not great, I just couldn't think of anything better.  `ProtectedPlacesDb` seems like an improvement to me.

bendk:
#4805 does this in a better way.

PR #4795, opened 2022-01-25
For EXP-2012: Move tests for enrollment.rs into their own test files

eliserichards:
[EXP-2012](https://mozilla-hub.atlassian.net/browse/EXP-2012)

This moves the two `mod test`s out of `enrollment.rs` and creates new test files for them: 
* `tests/test_enrollment.rs`
* `tests/test_enrollment_bw_compat.rs`

  - Note:


components/nimbus/tests/test_enrollment.rs
```diff
@@ -0,0 +1,2821 @@
...
+
+use nimbus::defaults::Defaults;
+use nimbus::error::Result;
+
+use std::{
+    collections::{HashMap, HashSet},
+    time::Duration,
+};
+
+const PREVIOUS_ENROLLMENTS_GC_TIME: Duration = Duration::from_secs(30 * 24 * 3600);
```
eliserichards, line 27: This is a `const` from `enrollment.rs`. Is there a way for me to access that original const without copy/pasting it here?

components/nimbus/tests/test_enrollment.rs
```diff
@@ -0,0 +1,2821 @@
...
+    let (nimbus_id, app_ctx, aru) = local_ctx();
+    let targeting_attributes = app_ctx.into();
+    let evolver = enrollment_evolver(&nimbus_id, &targeting_attributes, &aru);
+    let mut events = vec![];
+    let existing_enrollment = ExperimentEnrollment {
+        slug: "secure-gold".to_owned(),
+        status: EnrollmentStatus::WasEnrolled {
+            enrollment_id: Uuid::new_v4(),
+            branch: "control".to_owned(),
+            experiment_ended_at: now_secs() - nimbus::enrollment::PREVIOUS_ENROLLMENTS_GC_TIME.as_secs() - 60,
```
eliserichards, line 1943: Is there a better way to access this than `nimbus::enrollment::`?

components/nimbus/tests/test_enrollment.rs
eliserichards: Figured it out. Change `const` ➡️ `pub const`

components/nimbus/tests/test_enrollment.rs
```diff
@@ -0,0 +1,2823 @@
...
+    let targeting_attributes = app_ctx.into();
+    let evolver = enrollment_evolver(&nimbus_id, &targeting_attributes, &aru);
+    let mut events = vec![];
+    let existing_enrollment = ExperimentEnrollment {
+        slug: "secure-gold".to_owned(),
+        status: EnrollmentStatus::WasEnrolled {
+            enrollment_id: Uuid::new_v4(),
+            branch: "control".to_owned(),
+            experiment_ended_at: now_secs()
+                - nimbus::enrollment::PREVIOUS_ENROLLMENTS_GC_TIME.as_secs()
```
eliserichards, line 1944: Is there a better way to access this than `nimbus::enrollment::`?

tarikeshaq (approved):
Looks great!! Thanks for taking this on @eliserichards 🚀 🕺 🎉 

The only two things we are losing by making those tests integration tests (i.e in the `tests` directory) vs unit tests is the conditional compilation (all the `#[cfg(test)]`s) and that we are making more APIs than we need `pub`.

One way to keep those is to move the tests into separate files, but keep the tests as unit tests! we can use Rust's module system to help us out - for example in `lib.rs` we can have a 

```rs
#[cfg(test)]
mod tests;
```

Then we can create a directory called `tests` under the `src` directory with a file `mod.rs` that includes all the other tests. 

I don't feel strongly about this **especially since nimbus is only used by Mozilla apps through uniffi, so the visibility of the Rust API's themselves isn't really an issue**

components/nimbus/src/tests/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+mod test_enrollment;
+mod test_enrollment_bw_compat;
+mod common;
+
+use nimbus::{
+    error::Result, AppContext,
+};
+use serde_json::{json, Value};
```
tarikeshaq, line 12: We should move those into the individual test files that need them 😄 and replace `nimbus` with `crate` - since those are unit tests, they are internal modules to the crate, so we reference the crate using `crate::`

components/nimbus/src/tests/test_enrollment.rs
```diff
@@ -0,0 +1,2817 @@
...
+};
+
+use tempdir::TempDir;
+use uuid::Uuid;
+
+use nimbus::defaults::Defaults;
+
+use std::collections::{HashMap, HashSet};
+
+// const PREVIOUS_ENROLLMENTS_GC_TIME: Duration = Duration::from_secs(30 * 24 * 3600);
```
tarikeshaq, line 21: nit: 😛

tarikeshaq (changes requested):
Thanks Elise for taking my suggestion 😄 

The tests aren't actually getting compiled right now 😬 

To get them to compile, at the bottom of your [`lib.rs`](https://github.com/mozilla/application-services/blob/main/components/nimbus/src/lib.rs#L708) you'll notice a
```rs
#[cfg(test)]
mod tests {
```
The naming there is unfortunate, for now, what I suggest doing is replacing that with:
```rs
#[cfg(test)]
mod tests;

#[cfg(test)]
mod lib_tests {
// <keeping the existing tests in lib.rs unchanged, but under a different module name>
```

Then run `cargo test` and you'll notice the tests are compiled! The compiler will complain for now, but those should all be small visability problems - and the suggest changes I added below should help a little!

components/nimbus/src/tests/test_enrollment.rs
eliserichards: 😱 ❌

eliserichards:
Everything is looking peachy now. I took out the `pub`s that were no longer necessary when I created the new `tests` folder, and made the remaining ones `pub(crate)` to limit it as much as possible

Looks like codecov went down though 😂

tarikeshaq (approved):
Awesome job @eliserichards!! 

:shipit:

Issue #4775, opened 2022-01-13
IPv6 loopback in viaduct


martinthomson:
https://github.com/mozilla/application-services/blob/74c9db387bce7f6cc142393fd3ad3a985b119259/components/viaduct/src/backend.rs#L47-L49

This could probably use an additional value of "::1" (and maybe variations on that, but that's the obvious one to allowlist here).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3011)

Grubba27:
It would be like this the exemple here is on C# ?
```C#
IPAddress address = IPAddress.Parse("::1");
```

in Rust is ::1 or is 0:0:0:0:0:0:0:1?
```rust
request.url.host_str() != Some("::1");
```

martinthomson:
So the code here is imprecise.  You could do a string comparison and that is probably good enough for many cases.  Or, you could parse the string as an IP address (rust has classes for that) and then compare the result of that.  Your call.

Grubba27:
I would like to make the second one but i would do something like this ? 
```rust
let ipv6_addr = "::1".parse::<IpAddr>();

```

martinthomson:
I think that you can use the IpAddr constructor more directly.  If only for performance reasons.  I think that takes `[u8; 16]`, which you can populate with all zeros, except for a 1 in the last.

Grubba27:
i think i missed some point here 'ive created the first option as this one:
```rust
pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
    if request.url.scheme() != "https"
        && request.url.host_str() != Some("localhost")
        && request.url.host_str() != Some("127.0.0.1")
        && request.url.host_str() != Some("[::1]")
    {
        return Err(crate::Error::NonTlsUrl);
    }
    Ok(())
}

// the test

  let localhost_request_shorthand_ipv6 = crate::Request::new(
            crate::Method::Get,
            url::Url::parse("http://[::1]").unwrap(),
        );
        assert!(validate_request(&localhost_request_shorthand_ipv6).is_ok());
```
the second one that is one the i'm not getting the point:, i should be parsing too the request url ? 
```rust

pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
    let localhost_v6 = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1));
    if request.url.scheme() != "https"
[...]

PR #4798, opened 2022-01-27
ipv6 lookback issue

Grubba27:

  - Note:
- [ x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)

This pull request refers to [this issue](https://github.com/mozilla/application-services/issues/4775)
This pull request is just a refactoring of the loopback, I think it does not need an entry in Changes_unreleased.md but if this is the case i would write there too

### DIFF
--- a/components/viaduct/src/backend.rs
+++ b/components/viaduct/src/backend.rs
@@ -43,12 +43,16 @@ pub fn send(request: crate::Request) -> Result<crate::Response, crate::Error> {
 }
 
 pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
-    match request.url.host() {
-        Some(url::Host::Domain(d)) => d != "localhost",
-        Some(url::Host::Ipv4(addr)) => !addr.is_loopback(),
-        Some(url::Host::Ipv6(addr)) => !addr.is_loopback(),
-        None => return Err(crate::Error::NonTlsUrl),
-    };
+    if request.url.scheme() != "https"
+        && match request.url.host() {
+            Some(url::Host::Domain(d)) => d != "localhost",
+            Some(url::Host::Ipv4(addr)) => !addr.is_loopback(),
+            Some(url::Host::Ipv6(addr)) => !addr.is_loopback(),
+            None => true,
+        }
+    {
+        return Err(crate::Error::NonTlsUrl);
+    }
     Ok(())
 }
 
--- a/components/viaduct/src/backend.rs
+++ b/components/viaduct/src/backend.rs
@@ -96,10 +96,8 @@ mod tests {
         );
         assert!(validate_request(&localhost_request).is_err());
 
-        let localhost_request_shorthand_ipv6 = crate::Request::new(
-            crate::Method::Get,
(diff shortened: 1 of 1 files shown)


components/viaduct/src/backend.rs
```diff
@@ -47,6 +47,8 @@ pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
     if request.url.scheme() != "https"
         && request.url.host_str() != Some("localhost")
         && request.url.host_str() != Some("127.0.0.1")
+        && request.url.host_str() != Some("[::1]")
```
mhammond, line 50: In the issue, I think Martin was suggesting that the `host_str` checks can be replaced with something like:

```rust
match request.url.host() {
    Some(url::Host::Domain(d)) => d != "localhost",
    Some(url::Host::Ipv4(addr)) => !addr.is_loopback(),
    Some(url::Host::Ipv6(addr)) => !addr.is_loopback(),
    None => true,
}
```

(with a `use url` at the top of the file) which is cleaner and handles every possible way to spell an ipv4 and ipv6 loopback addresses)

mhammond (dismissed):
Thanks heaps for digging in here!

components/viaduct/src/backend.rs
Grubba27: i got it but when i do this changes this validation start to fail maybe i did something wrong?
```rust
pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
    match request.url.host() {
        Some(url::Host::Domain(d)) => d != "localhost",
        Some(url::Host::Ipv4(addr)) => !addr.is_loopback(),
        Some(url::Host::Ipv6(addr)) => !addr.is_loopback(),
        None => return Err(crate::Error::NonTlsUrl),
    };
    Ok(())
}
// the only test that fails


        let _http_request = crate::Request::new(
            crate::Method::Get,
            url::Url::parse("http://www.example.com").unwrap(),
        );
        assert!(validate_request(&_http_request).is_err());
```

mhammond: You still need the check for `request.url.scheme() != "https"` - it's only the `host_str` checks that can be replaced with that block. That's why my block has `None => true` - ie, something like:

```rust
    if request.url.scheme() != "https"
        && match request.url.host() {
            Some(url::Host::Domain(d)) => d != "localhost",
            Some(url::Host::Ipv4(addr)) => !addr.is_loopback(),
            Some(url::Host::Ipv6(addr)) => !addr.is_loopback(),
            None => true,
        }
    {
        return Err(crate::Error::NonTlsUrl);
    }
```

Grubba27: OMG, I've never tought using match this way. This really simple thing that you have shown me made me love even more this language

Grubba27:
I'think that all is ok and I'm realy glad that I learned a new thing

mhammond (approved):
Thanks!

Issue #1684, opened 2019-08-29
Sync Manager: Support for sync interruption


thomcc:
In #1447, I punted on interruption, mainly as it's hard to make logins/places use the correct interrupt handle (Instead, I gave them a dummy interrupt handle). This is probably a problem long term, as sync (probably) needs to be interruptable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2384)
┆epic: Sync Ecosystem (backlog)
┆sprintEndDate: 2022-02-18

bendk:
I'm going to try to fix this one alongside #4617

bendk:
Talked with mhammond about this one some more and I think fixing it is a bigger project.  I think we can probably do it along side the sync-manager/places work we're doing, but not with #4617.

bendk:
This one has been tricky for 2 reasons:
  - Most of the components aren't set up for interrupting sqlite.  They use a shared connection for syncing and all other operation.  This means if we just try to interrupt that connection, it's possible we'll interrupt the wrong operation.
  - It's not totally clear if this is needed.  On Android it seems like we probably don't need that since their interruption model is based on simply killing a processes.  On iOS it's unclear if this is useful or not.

I think there is a way forward on this one though:
  - We mostly focusing on just getting the sync engines to use the correct `SqlInterruptScope`.  This would allow for the Rust code to be interrupted.
  - We allow sync engines to optionally pass an `rusqlite.InterruptHandle` to the `SyncManager`. Then when the sync manager's interrupt method is call, it would interrupt those handles.  In the short term, only places would do this.  Places tends to have the longest running queries, so at least we have that covered.  In the long term, if we decide we need this for other components, we could update their code to have a dedicated sync connection and then hook them up as well.

I think this makes this issue much more tractable since we can ignore the harder issue of trying to interrupt a sql query for components with a shared DB connection

mhammond:
> * We allow sync engines to optionally pass an `rusqlite.InterruptHandle` to the `SyncManager`.

Can we just add an `interrupt` or similar methods on the sync engine trait?

But in general, this SGTM.

bendk:
I didn't add `SyncEngine.interrupt()` just because places has 2 sync engines that share a connection, so we only need to interrupt 1 of them.  Instead I just made top-level function for it, like `get_registered_sync_engine()`.

[...]

PR #4805, opened 2022-01-31
[Don't merge yet] Shutdown support (alternative try at #1684)

bendk:
In the review for #4790, @mhammond suggested trying a one-shot approach at interruption.  This is my try at it.

Added a shutdown crate which handles shutting down app-services.  The idea is that when the app goes into shutdown mode, or goes to sleep, you can call `shutdown::shutdown()` and this will interrupt all in-progress DB queries, then make the `err_if_shutdown()` start returning errors which should cause long-runner operations to start failing.  If you want to resume after sleeping you can `shutdown::restart()`.  So this is like a one-shot approach, but it also supports resuming afterwards.

This required a lot refactoring and probably more is still needed:
  - The places autocomplete code needs some refactoring and also we should implement interruption support there
  - The other components would need to be refactored to use this approach (I think this should be easier than places)
  - The iOS and Android code needs to be updated to deal with the breaking changes.  I think this should be pretty trivial.

I updated places-utils to work with this code.  To make things compile, I just ripped out all the import/export code.  The iOS/fennec code is going to go a way soon.  I wasn't sure if the other import/export code was useful, we need it then I guess that's one more coding task.  You can use places-utils to test this, by doing multiple passes and hitting `ctrl-c` in the middle to initiate shutdown/resume.  For example:

 - Execute `cargo run --example places-utils sync --nsyncs 4 --wait 1000`
 - In the middle of the first sync hit ctrl-c
 - During the 1s wait period hit ctrl-c again
 - You should see that the first sync gets interrupted, but the subsequent syncs work.

@mhammond what do you think of this approach?

  - Note:

### DIFF
--- a/components/places/src/api/places_api.rs
+++ b/components/places/src/api/places_api.rs
@@ -14,7 +14,6 @@ use crate::util::normalize_path;
 use lazy_static::lazy_static;
 use parking_lot::Mutex;
 use rusqlite::OpenFlags;
-use sql_support::{SqlInterruptHandle, SqlInterruptScope};
 use std::cell::Cell;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
@@ -110,7 +109,10 @@ pub struct SyncState {
 /// https://github.com/mozilla/uniffi-rs/pull/1063 would fix this, but got some pushback
 /// meaning we are forced into this unfortunate workaround.
 pub fn places_api_new(db_name: impl AsRef<Path>) -> Result<Arc<PlacesApi>> {
-    PlacesApi::new(db_name)
+    let api = PlacesApi::new(db_name)?;
+    let api_weakref: Weak<PlacesApi> = Arc::downgrade(&api);
+    shutdown::register_interrupt(api_weakref);
+    Ok(api)
 }
 
 /// The entry-point to the places API. This object gives access to database
@@ -252,9 +254,9 @@ impl PlacesApi {
                     mem_cached_state,
                     client_init,
                     key_bundle,
-                    // The interrupt system is not currently working (#1684), but we need to pass
-                    // in something, so let's use the engine scope.
-                    &engine.scope,
+                    // sync_multiple uses the old interrupt system, but we need to pass in
+                    // something, so let's use NeverInterrupts.  See #1684 for details.
+                    &interrupt_support::NeverInterrupts,
                     None,
                 )
             },
@@ -276,9 +278,9 @@ impl PlacesApi {
                     mem_cached_state,
                     client_init,
                     key_bundle,
-                    // The interrupt system is not currently working (#1684), but we need to pass
-                    // in something, so let's use the engine scope.
-                    &engine.scope,
+                    // sync_multiple uses the old interrupt system, but we need to pass in
+                    // something, so let's use NeverInterrupts.  See #1684 for details.
+                    &interrupt_support::NeverInterrupts,
                     None,
                 )
             },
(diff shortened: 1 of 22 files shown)


components/places/src/api/places_api.rs
```diff
@@ -450,32 +413,24 @@ impl PlacesApi {
...
-        // called in startup.
-        let _guard = self.sync_state.lock();
-        let conn = self.get_sync_connection()?;
-        let db = conn.lock();
-        Ok(Arc::new(db.new_interrupt_handle()))
+impl shutdown::ShutdownInterrupt for PlacesApi {
+    fn interrupt(&self) {
+        self.read_connection.interrupt_handle.interrupt();
+        self.write_connection.interrupt_handle.interrupt();
+        // TODO interrupt the sync connection, if it's open
```
bendk, line 433: I got to a point where I realized there was already a ton of changes and I should discuss them with someone before making more.

components/support/shutdown/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+pub fn shutdown() {
+    IN_SHUTDOWN.store(true, Ordering::Relaxed);
+    for weak in REGISTERED_INTERRUPTS.lock().iter() {
+        if let Some(interrupt) = weak.upgrade() {
+            interrupt.interrupt()
+        }
+    }
+}
+
+/// Restart after a shutdown
```
mhammond, line 28: How is this used? I wonder if we can soften expectations here, saying something like "used only for testing" or similar? Or even better, only make it available in the test cfg? Specifically, I'm thinking that to make this truly reliable we'd probably need additional protections, ensuring other threads aren't currently interrupting etc?

examples/places-utils/src/places-utils.rs
```diff
@@ -43,127 +32,8 @@ struct DesktopItem {
     children: Vec<DesktopItem>,
 }
 
-fn convert_node(dm: DesktopItem) -> Option<BookmarkTreeNode> {
```
mhammond, line 46: I'd rather not lose this code if we can avoid it. It certainly doesn't need to be interruptable though.

examples/places-utils/src/places-utils.rs
```diff
@@ -366,9 +204,18 @@ fn main() -> Result<()> {
...
     let api = PlacesApi::new(&db_path)?;
-    let db = api.open_connection(ConnectionType::ReadWrite)?;
     // Needed to make the get_registered_sync_engine() calls work.
     api.clone().register_with_sync_manager();
+    ctrlc::set_handler(move || {
+        if !shutdown::in_shutdown() {
+            println!("\nCTRL-C detected, enabling shutdown mode\n");
+            shutdown::shutdown();
+        } else {
+            println!("\nCTRL-C detected disabling shutdown mode\n");
```
mhammond, line 214: this seems a little suspect - if I hit ctrl+c multiple times I might end up in a frankenstate? (Not that I really care for this utility, but I'm trying to understand things)

components/places/src/api/places_api.rs
```diff
@@ -110,15 +109,19 @@ pub struct SyncState {
...
 }
 
 /// The entry-point to the places API. This object gives access to database
 /// connections and other helpers. It enforces that only 1 write connection
 /// can exist to the database at once.
 pub struct PlacesApi {
     db_name: PathBuf,
-    write_connection: Mutex<Option<PlacesDb>>,
+    pub read_connection: Arc<PlacesConnection>,
+    pub write_connection: Arc<PlacesConnection>,
```
mhammond, line 124: places has many read connections. Any reason you can't put interrupt support directly on `PlacesDb`, and "unregister" when dropped?

components/places/src/api/places_api.rs
mhammond: Yeah, I think we should see if we can use a lighter touch here.

mhammond (commented):
In many ways I really like this. However, once I got to the semantic changes to places I became far more wary. My main idea of a one-shot shutdown was that it would allow for a lighter touch, but that seems to have failed fairly spectacularly. The fact we want to drop code from places-utils and also perform such radical surgery on the places connection semantics mean I'd probably rather just declare no interrupt/shutdown support given no one actually seems to care.

It would be great if you can work out how to stop this bleeding out and forcing us into design choices we wouldn't otherwise make.

PR #4816, opened 2022-02-01
Adding shutdown support to sql-support

bendk:
I think 3rd time's the charm on this one.  This PR is similar to #4805, but much simpler.   It implements a shutdown system that hooks into the existing `SqlInterruptScope` and `SqlInterruptHandler` functionality, which means that the changes are a lot less invasive.  We end up with each DB being able to create an `SqlInterruptScope`, but with a global shutdown mode, that interrupts all of those scopes.  So we end up with being able to both globally shutdown everything and cause all operations to be interrupted and also locally shutdown a single operation (e.g. autocomplete).

You can test this with the `places-utils` script:
    - Execute `cargo run --example places-utils sync --nsyncs 2 --wait 1000`
   -  In the middle of the first sync hit ctrl-c
   -  You should see that the first sync gets interrupted in the middle and the second sync interrupted as soon as it starts.

  - Note:


components/support/sql/src/shutdown.rs
```diff
@@ -0,0 +1,81 @@
...
+///   - Use `SqlInterruptScope::new_with_shutdown_check()` to create a new
+///     `SqlInterruptScope`
+///   - Database connections need to be wrapped in a type that:
+///      - Implements `AsRef<SqlInterruptHandle>`.
+///      - Gets wrapped in an `Arc<>`.  This is needed so the shutdown code can get a weak reference to
+///        the instance.
+///      - Calls `shutdown::register_interrupt_handle()` on creation
+///
+///  See `PlacesDb::begin_interrupt_scope()` and `PlacesApi::new_connection()` for an example of
+///  how this works.
```
bendk, line 25: It's kind of weird that this module is inside the `sql-support` crate, but it kind of needs to be because of how it integrates with `SqlInterruptScope` and `SqlInterruptHandler`.  I think we should consider consider moving both `shutdown.rs` and `interrupt.rs` to the `interrupt-support` crate.

mhammond (approved):
Awesome, thanks for working through this!

bendk:
I think this is a step forward so I'm planning on merging once the checks pass.  However, there are still several steps left to fully implement #1684.  I left a comment in that ticket outlining the remaining work.

PR #4822, opened 2022-02-04
upgrades to rust 2021 [ci full]

tarikeshaq:
Upgrades to Rust 2021!

Seems build OK and tests pass - but lets see what CI says

Biggest win for us is that now `TryFrom` and `TryInto` are available by default (and no need to add a `use` explicitly for them)


tarikeshaq:
Uhhh of course, our minimum version hasn't hit 1.56 yet 😭 😭 😭

tarikeshaq:
huh looks like in moz-central the minimum rust version is now [1.57](https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#15). I'll change ours to reflect that (which means we can probably land this!)

tarikeshaq:
Think our taskcluster jobs still use an older version of Rust. I'll take a look on Monday to see and run some ci jobs

tarikeshaq:
>  Is there still an issue with taskcluster?

I think so 😢 ref: https://github.com/mozilla/application-services/blob/main/taskcluster/scripts/toolchain/rustup-setup.sh#L22

I'll add `ci full` then close and open this PR, hopefully the taskcluster jobs run (and I assume they will fail, but 🤷)

I'll also add as a part of this upgrade, docs to remind us to update the taskcluster rust version... It's what eventually is used to ship to our android consumers..

tarikeshaq:
huh looks like the version we have on the `rust-toolchain` is overriding the 1.24, ref: https://firefox-ci-tc.services.mozilla.com/tasks/V3V4flMCROe0C8pqSpPUyw/runs/0/logs/live/public/logs/live.log#L219

We can land this OK then once CI passes, but for sanity I'll add a note in our docs that the rust-toolchain version is also the one used by taskcluster

bendk:
I think that means we install an older version of rust-up, but that should still be able to install the latest versions of the other toolchain parts.  In fact, it can probably self-update itself.  We probably should update that, but I don't think that should block merging this.

PR #4823, opened 2022-02-04
Sync shutdown

bendk:
- Move the shutdown interruption code added in #4816 to the `interrupt` crate (this was mentioned and thumbs-uped in the comments)
- Updated `SqlInterruptScope` and `SqlInterruptHandle` to work better with the shutdown code
- Update the component code to use the new stuff

This almost allows all components to be interrupted at shutdown, the last thing issue is to actually expose the `shutdown()` method, but I wasn't sure what the best way to do that is.  We could UniFFI the `interrupt` crate, but it seems a little silly to do that for just one method.  Maybe it would make more sense to start a new crate that intended for all the miscellaneous app-services functionality (shutdown, rustlog, crashtest, etc).  Lastly, we could just add the `shutdown` method to an existing UDL file (maybe sync-manager?).  I'm not sure what's best here, what do others think?

We just discussed that question on our call and decided that a global app-services crate/UDL file would be good.  So what if we leave this PR as-is and expose `shutdown()` when we add that?

  - Note:


components/logins/src/store.rs
```diff
@@ -24,15 +24,25 @@ pub fn get_registered_sync_engine(engine_id: &SyncEngineId) -> Option<Box<dyn Sy
...
 }
 
+fn create_sync_engine(store: Arc<LoginStore>, engine_id: &SyncEngineId) -> Result<Box<dyn SyncEngine>> {
+    match engine_id {
+        SyncEngineId::Passwords => Ok(Box::new(LoginsSyncEngine::new(Arc::clone(&store))?)),
+        // panicing here seems reasonable - it's a static error if this
+        // it hit, not something that runtime conditions can influence.
+        _ => unreachable!("can't provide unknown engine: {}", engine_id),
+    }
+}
```
bendk, line 44: It's a bit annoying that we have to refactor the code like this to deal with the fact that creating a new interrupt scope can now fail, therefore creating a new sync engine can also fail.  However, this does deal with the case where we initiate shutdown mode, then another thread tries to start a sync.

components/sync_manager/src/manager.rs
```diff
@@ -109,9 +108,7 @@ impl SyncManager {
     ) -> Result<SyncResult> {
         let key_bundle = sync15::KeyBundle::from_ksync_base64(&params.auth_info.sync_key)?;
         let tokenserver_url = url::Url::parse(&params.auth_info.tokenserver_url)?;
-        // TODO(issue 1684) this isn't ideal, we should have real support for interruption.
-        let p = Arc::new(AtomicUsize::new(0));
-        let interruptee = sql_support::SqlInterruptScope::new(p);
+        let interruptee = interrupt_support::ShutdownInterruptee;
```
bendk, line 111: This should handle #1684.  Now we will interrupt syncing when when we enter shutdown mode.

mhammond (approved):
Thanks (and sorry for the delay, but in my defense there wasn't a reviewer specified, so I wasn't saved by myqonly :)

There's a CI failure that looks real, so please double-check that before landing.

PR #4824, opened 2022-02-04
Misc improvements to Android verify scripts

grigoryk:

  - Note:


grigoryk:
`set -e` is needed so that we actually bail out if there are any issues with sub-scripts (otherwise, it may look like things succeeded).

NSS_DIR path printout was helpful for debugging purposes, in case something goes wrong.

`cmdline-tools` were missing for me (default Android Studio setup), so a verify script was failing. Added a helpful error message in that case.

And finally, architecture mismatch warning printout (e.g. x86 java running on m1) to help catch weird errors while we're still in a transition phase.

libs/verify-android-environment.sh
```diff
@@ -10,10 +12,22 @@ if [[ ! -f "$(pwd)/libs/build-all.sh" ]]; then
...
 
+echo "Verifying android-specific environment..."
 "$(pwd)/libs/verify-android-ci-environment.sh"
 
+# Mac-specific checks, to help out with the M1 transition.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  java_arch=$(which java | xargs file | cut -f 5 -d ' ')
+  system_arch=$(uname -m)
+  if [ "$java_arch" != "$system_arch" ]; then
+    echo "WARNING: mismatching Java (${java_arch}) and system (${system_arch}) architectures. Make sure this is intentional, or you may get strange build errors."
```
tarikeshaq, line 29: tried running this locally on my Intel machine and got a 
```sh
WARNING: mismatching Java (with
64-bit
64-bit) and system (x86_64) architectures. Make sure this is intentional, or you may get strange build errors.
```

This could be an issue with my own setup though, as I remember I installed multiple versions of java at some point (can't remember 🙃) For reference, here's what `which java | xargs file` prints: 
```sh
/usr/bin/java: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/usr/bin/java (for architecture x86_64):	Mach-O 64-bit executable x86_64
/usr/bin/java (for architecture arm64e):	Mach-O 64-bit executable arm64e
```

If this is a very edge setup (which means I should get my java stuff fixed locally 😅) then nevermind! 😄

tarikeshaq (approved):
Thank you!! I've come across sub-script issues before, and since we didn't have the `set -e` it threw me off 😅

had a small question about the java architecture check, but LGTM!

libs/verify-android-environment.sh
grigoryk: Oh, interesting. I'll make the arch parsing a bit more robust!

mhammond:
Hey @grigoryk, should this land?

grigoryk:
Yeah. I just ran into these issues again locally, and this PR would have been helpful to have, even in its current state (@tarikeshaq comment unaddressed).

tarikeshaq:
Thanks @grigoryk! I merged this and filed https://github.com/mozilla/application-services/issues/4896 for my comment earlier 😄 It's just a warning and we'll get to it

Issue #4826, opened 2022-02-05
Create a new smoke test for Firefox iOS


tarikeshaq:
With Carthage gone (🎉 🎉) our old smoke test won't work. We should consider writing a new one

I don't think the old smoke test even ran properly but creating this issue so that we can discuss if a smoke test is something we want to add

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3048)

data-sync-user:
➤ Janet Dragojevic commented:

🚩 Flag added

qweij;eklja sdf;lkajs f;alksdjf ;adlsk

Issue #4825, opened 2022-02-05
Document how to locally test against Firefox iOS


tarikeshaq:
With Carthage gone (🎉 🎉) we have a new approach of shipping our components. This means that the old way of testing local changes in application services against Firefox iOS no longer works, and we currently don't have any documentation explaining how one would do that.

An output of this issue should be a doc, under [docs/howtos](https://github.com/mozilla/application-services/tree/main/docs/howtos) with step by step guide for an application-services developer, a Firefox iOS developer, or a contributor to integrate a local change in Rust code and have it built with Firefox iOS.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3049)
┆Story Points: 3
┆epic: Rust Components on iOS developer experience
┆sprintEndDate: 2022-05-27

data-sync-user:
➤ Tarik Eshaq commented:

App services PR: https://github.com/mozilla/application-services/pull/4959 ( https://github.com/mozilla/application-services/pull/4959|smart-link )

Rust components PR: https://github.com/mozilla/rust-components-swift/pull/59 ( https://github.com/mozilla/rust-components-swift/pull/59|smart-link )

PR #4827, opened 2022-02-05
Misc updates now that we got rid of carthage [ci full]

tarikeshaq:
This includes:
- Removing `swift-protobuf` from our scripts, CI, and dependency summaries (viaduct still uses protobufs, but not in swift/ios)
- Removing the firefox ios `smoke test` and filing #4826 
- Removing outdated documentation on firefox iOS, and referencing #4825 to add new documentation for a local setup
- Removes references in `Summary.md` to documents that don't exist anymore - [the references were causing our book to not be published](https://github.com/mozilla/application-services/actions/runs/1797947113)
- ~~Updating the taskcluster CI to build for both M1s and Intel Macs, lets see if that passes~~ this a lot more complicated than I initially thought, filed #4828
- general small changes to docs


mhammond:
Do these tests now fail to run, or is this just a cleanup? If the latter, can we keep some smoke-tests - it would be a shame to break iOS in significant ways but only discover this much later via CI in a different repo.

tarikeshaq:
> Do these tests now fail to run, or is this just a cleanup? If the latter, can we keep some smoke-tests - it would be a shame to break iOS in significant ways but only discover this much later via CI in a different repo.

worse, they are deceiving!! after [we removed carthage](https://github.com/mozilla/application-services/pull/4619/files#diff-e9bc410e36e76e1e378f90e6525206b3677b8e3ab138d288198efdecb1639d93L51) the script now runs the ios tests, without applying the changes in application services (as if somebody just installed a fresh firefox-ios and ran its tests) 😞 

I filed #4826 to write some new smoke tests... I imagine the new ones will look **very** similar to the ones we have now, but the script to substitute a local appservices will look significantly different

mhammond (approved):
> without applying the changes in application services (as if somebody just installed a fresh firefox-ios and ran its tests)

Ouch :) Thanks for keeping this stuff up-to-date here, particularly keeping the docs etc current!

PR #4830, opened 2022-02-08
Merge ios and ios-rust directories and clean up Xcode project

skhamis:
Quite a few changes but nothing too scary overall. Some high-level things I wanted to resolve:

1. Our Xcode project was antiquated and made it difficult to run any tests against our new XCFrameworks
 - So I went ahead and made a branch new project (dubbed MozillaTestServices) that uses our XCframework and removed all the building-a-framework stuff that is now outdated
 - The new project name makes it also clear we're only testing in this project and nothing else
2. We had two versions of everything, one for carthage and one for SPM -- killed the carthage directory (ios)

Another potential big change is removing the reliance on glean-core as a submodule. We can now use Glean as an SPM and that'll make sure we don't use anything private (see below) for our tests and allows us to use the SPM route to update glean rather than hidden dependency.

- NOTE: Nimbus tests currently fail with this approach due to using `experimentData.branch` which is a private variable in glean-core (and thus not using glean-core surfaced this error).

  - Note:


tarikeshaq:
> NOTE: Nimbus tests currently fail with this approach due to using experimentData.branch which is a private variable in glean-core (and thus not using glean-core surfaced this error).

Sammy and I chatted about this, and I think if all else works, we should change our tests (or get a glean-swift release with those public if they were meant to be used) - I think removing the glean submodule is worth changing our tests a little

skhamis:
@travis79 I added you as a reviewer per our conversation about commenting out the failing part of the test that uses a private member of Glean in NimbusTests.swift -- I also made a bug to ensure we get that fixed after this PR. I'm not removing glean as a submodule as of yet but this will be the first step, however let me know if there are any blockers for as this move definitely affects Nimbus!

tarikeshaq:
Just wanted to flag that I'm reviewing this - the code changes sound good. I'm having trouble building this locally, but since CI passes I probably need a small local change of some sort 😄 I'll have time tomorrow to finish reviewing

tarikeshaq (approved):
The lines diff is just 🤩 Thank you for taking this on!!! LGTM 🚢

Issue #4828, opened 2022-02-05
Enable taskcluster to publish M1 artifacts


tarikeshaq:
Right now, the artifacts generated by taskcluster are all built for [Intel Macs](https://github.com/mozilla/application-services/blob/main/taskcluster/ci/module-build/kind.yml#L43) (note that the `rust.targets` there only have `darwin`)

If we would like our android release to be accessible by M1s, we should probably change our taskcluster configs a little. I'm not sure the full extent of what's needed, but at the minimum:
- Most likely need to split the taskcluster job for [`desktop-macos`](https://github.com/mozilla/application-services/blob/main/taskcluster/ci/toolchain/desktop.yml#L17) into two, one for M1s and one for Intel (including modifying the [`desktop-macos`](https://github.com/mozilla/application-services/blob/main/taskcluster/scripts/toolchain/desktop-macos.sh) script to handle M1s.
- Include `darwin-aarch64` in https://github.com/mozilla/application-services/blob/main/taskcluster/ci/module-build/kind.yml#L43
- Add a `rust add target aarch64-apple-darwin` in https://github.com/mozilla/application-services/blob/main/taskcluster/scripts/toolchain/cross-compile-setup.sh#L33
- Properly set the environment variables like is done in https://github.com/mozilla/application-services/blob/main/taskcluster/scripts/toolchain/cross-compile-setup.sh#L2-L11 (this will probably look something like:
```sh
export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_AARCH64_APPLE_DARWIN_NSS_STATIC=1
.
.
```
- There are most likely more changes that need to be done, but this is hitting the end of my familiarity with taskcluster
- Check-in with the taskcluster team to make sure all is OK, but if we can get an M1 to run a **release** megazord artifact, then we should be good

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3050)
┆Epic: Important backlog

bendk:
Couple notes on this one since I've been working on this lately:
 -  I think https://github.com/mozilla/application-services/pull/5543 and https://github.com/mozilla/application-services/pull/5563 will affect this.
[...]

PR #4792, opened 2022-01-22
Fix gradle builds for Apple Silicon running arm64 JDKs

skhamis:
fixes #4665 

### To Test (Intel or M1 Mac)
1. Delete your libs/desktop folder (should be only darwin in there)
2. If you have any `rust.targets` in your `local.properties` update (or just comment out for now):
 - Remove any `darwin` targets
 - update to one of the new darwin-specific architectures
2. Run `./libs/verify-desktop-environment.sh`
3. also run `./build-all.sh android` (shouldn't need rebuild but just incase)
3. you should now see libs/desktop/darwin-{your-arch}
4. in the top level of the project, run `./gradlew assembleDebug`
5. using local-published flow, ensure android-components can also build successfully
6. using local-published flow, ensure fenix can also build successfully 

- [x] Need to confirm it still works on Intel (will need to clear out libs/desktop and rebuild NSS/SQLCipher)
- [X] Need to be able to successfully build android-components -> fenix with no issues

### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - Note:
    - For changes that need extra cross-platform testing, consider adding `[ci full]` to the PR title.
    - If this pull request includes a breaking change, consider [cutting a new release](https://github.com/mozilla/application-services/blob/main/docs/howtos/cut-a-new-release.md) after merging.
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry in [CHANGES_UNRELEASED.md](../CHANGES_UNRELEASED.md) or an explanation of why it does not need one
  - Any breaking changes to Swift or Kotlin binding APIs are noted explicitly
- [ ] **Dependencies**: This PR follows our [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)
  - Any new dependencies are accompanied by a summary of the due dilligence applied in selecting them.

PR #4846, opened 2022-02-15
Temporarily disable downloading native libs in android [ci full]

tarikeshaq:
Disable downloading native libraries, forcing local environments to build the native libraries (nss, etc) themselves.

This is because for macs, the `useDownloadedLibs` flow is broken, since after https://github.com/mozilla/application-services/pull/4792 `NSS_DIR` is expected to be point to a `darwin-x86-64` or `darwin-aarch64` when the downloaded libraries always have them in a `darwin` directory.

The alternative is to force set the `NSS_DIR` to point to `darwin` in gradle, but that won't work for M1s, since the [downloaded libs don't have M1 builds anyways](https://github.com/mozilla/application-services/issues/4828)


skhamis:
Nice catch! I thought cross-compile-target was CI. So technically `darwin` is still there for CI? 
https://github.com/mozilla/application-services/blob/cbfa39d756f63e95bd76b8efdbb1907eb1c368e6/libs/build-nss-desktop.sh#L30

https://github.com/mozilla/application-services/blob/cbfa39d756f63e95bd76b8efdbb1907eb1c368e6/libs/build-sqlcipher-desktop.sh#L29

mhammond (approved):

tarikeshaq:
Came across this again trying to build app services on android - I'll land this for now, if I notice this breaking for anyone's local environment (it shouldn't, but 🤷) I'll revert it

PR #4847, opened 2022-02-15
[Data review done] Includes more errors in reported telemetry

tarikeshaq:
Draft because I'm still writing the issue for the data review. Also, draft because I need to link back the data review bug into the `metrics.yaml`.

Adds all the error types in the `places.udl` as recorded telemetry.

While I'm at it, I got rid of the `InternalPanic` type (which is a breaking change)


tarikeshaq:
data review in https://github.com/mozilla/application-services/issues/4848

mhammond (approved):
I wonder if everything is now going to be "unexpected_places_exception"? :)

PR #4851, opened 2022-02-16
For EXP-2012: Move tests for updating.rs into their own file

eliserichards:
[EXP-2012](https://mozilla-hub.atlassian.net/browse/EXP-2012)

This moves the `mod test` out of `updating.rs` and creates a new test file for it: 
* `tests/test_updating.rs`

  - Note:


components/nimbus/src/updating.rs
```diff
@@ -38,41 +38,3 @@ pub fn read_and_remove_pending_experiments(
     // None is "there are no pending updates".
     Ok(experiments)
 }
-
-// This test crashes lmdb for reasons that make no sense, so only run it
-// in the "safe mode" backend.
-#[cfg(feature = "rkv-safe-mode")]
```
tarikeshaq, line 44: heh, this wasn't under a `cfg[test]` before 😆 Now it won't be compiled with non-test code 🎉 🎉

tarikeshaq (approved):
Awesome!!

components/nimbus/src/tests/test_updating.rs
```diff
@@ -0,0 +1,45 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::Result;
+use crate::updating::*;
+use crate::Database;
+
+// This test crashes lmdb for reasons that make no sense, so only run it
+// in the "safe mode" backend.
+#[cfg(feature = "rkv-safe-mode")]
```
tarikeshaq, line 11: Aaaah this is why clippy complains - for context:

Clippy runs on multiple rust ["feature" ](https://doc.rust-lang.org/cargo/reference/features.html) configurations:
- It runs against the defaults features
- Against "all features" enabled
- Against "no features" enabled

When the last runs, the test isn't compiled (since `rks-safe-mode` is disabled) and since there is no test compiled, the imports are unused.

A couple of easy ways forward:
- Add a `#![allow(unused_imports)]` at the top of the file, with a short comment
- In the `mod.rs`, right above the `mod test_updating` add a `#[cfg(feature = "rkv-safe-mode")]` and remove the same `cfg` from above the test function itself since it'll be redundant

I'm happy with either 😄 The latter feels a little nicer, but the first feels a bit more explicit 🤷 Regardless whichever you choose works!

tarikeshaq (commented):
Apologies I took longer than I wanted to get back to this and look at clippy! I dropped a comment of where the issue is 😄 Feel free to land this at your own pace

components/nimbus/src/tests/test_updating.rs
eliserichards: Thank you so much! I'm going to add the `#!allow` tag so it's clear in the test file what's going on and we can keep the "safe mode" comment

PR #4857, opened 2022-02-22
Adds tests for multiple locales geo targeting

tarikeshaq:
a part of https://mozilla-hub.atlassian.net/browse/EXP-2216

That ticket is not done with this, I still need to identify in which cases, if any, would locales not be formatted properly in experimenter vs the sdk

Adds test coverage to locale targeting. Specifically covering targeting expressions in the form of "locale in ['a', 'b']", which is what experimenter will be sending us

cc: @jrbenny35 Since you asked about where tests on targeting would go 😄 Let me know if you have any thoughts/questions

### DIFF
--- a/components/nimbus/src/tests/test_evaluator.rs
+++ b/components/nimbus/src/tests/test_evaluator.rs
@@ -11,6 +11,60 @@ use crate::{
 };
 
 #[test]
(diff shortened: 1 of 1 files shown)


jrbenny35:
This is awesome!!

travis79 (approved):

PR #4858, opened 2022-02-23
Remove duplicate DeviceType enum from .udl files

mhammond:
This is the other half of #4397. I'm not sure how I feel about it - on one hand, killing the dupe from .udl files seems great! OTOH though, adding a new .udl file to sync15 that contains exactly 1 enum sounds a bit like overkill, and it doesn't look like it will be growing anything new any time soon. Another option might be to create a "common types" crate - but even then, the amount of sharing isn't that great (but stuff like Guid and JsonObject support could live there). But maybe that's even more overkill. Another option is to just wait for something to arrive organically and add it then?

Clearly not urgent though, so marking as a draft, but welcome all thoughts.

### DIFF
--- a/components/sync15/src/sync15.udl
+++ b/components/sync15/src/sync15.udl
@@ -1,0 +1,19 @@
+// This file contains common sync15-related types which are used by various components.
+namespace sync15 {
+};
+
+// Enumeration for the different types of device.
+//
+// Firefox Accounts seprates devices into broad categories for display purposes,
+// such as distinguishing a desktop PC from a mobile phone. Upon signin, the
+// application should inspect the device it is running on and select an appropriate
+// [`DeviceType`] to include in its device registration record.
+//
+enum DeviceType {
+  "Desktop",
+  "Mobile",
+  "Tablet",
+  "VR",
+  "TV",
+  "Unknown",
+};
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3414,6 +3414,9 @@ dependencies = [
  "sync-guid",
  "sync15-traits",
  "thiserror",
+ "uniffi",
+ "uniffi_build",
+ "uniffi_macros",
  "url",
  "viaduct",
 ]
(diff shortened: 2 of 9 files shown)


mhammond:
doh - `thread 'main' panicked at 'no support for external types yet'` - IIUC, @bendk added support for kotlin recently, but swift doesn't yet have support, so this can't possibly land until it does.

components/sync15/src/sync15.udl
```diff
@@ -0,0 +1,19 @@
+// This file contains common sync15-related types which are used by various components.
```
mhammond, line 1: hrmph, `DeviceType` is actually defined in sync15-traits and re-exported from sync15, so having a sync15.udl seems wrong for this type.

bendk:
What if the shared/common crate also contained some functions?  `shutdown()` and `rc_log::log()` would be good candidates.  Maybe that makes it substantial enough to for a .udl file.

mhammond:
> What if the shared/common crate also contained some functions? `shutdown()` and `rc_log::log()` would be good candidates. Maybe that makes it substantial enough to for a .udl file.

Yeah - but where would it go? A new "utils" crate? That would lead to another odd situation where we have a crate just for the UDL and without any actual rust (other than the uniffi generated stuff)? (Not necessarily saying that's a problem, just that it seems slightly odd at first glance)

Didn't you have some use-case for a shared udl?

bendk:
It would be a crate for just the UDL.  It does feel a bit odd, but I'd prefer it over having a bunch of small UDL files with 1 enum or 1 function.

mhammond:
> doh - `thread 'main' panicked at 'no support for external types yet'` - IIUC, @bendk added support for kotlin recently, but swift doesn't yet have support, so this can't possibly land until it does.

I couldn't find an issue for this, but I think it's still a limitation, so I opened https://github.com/mozilla/uniffi-rs/issues/1218

mhammond:
Replaced by #5316

PR #4859, opened 2022-02-23
Make tabs example work on Windows and upgrade to StructOpt

mhammond:
The tabs example never worked on Windows or WSL, so I needed to fix that as part of the tabs persistence work, but thought I'd split it out. I'll also be adding another cmdline option, so figured I'd update to `StructOpt` at the same time to bring us in line with our other complex examples (eg, places_utils).

CC @lougeniaC64 as she is also working with tabs at the moment.


lougeniaC64 (approved):
LGTM!

PR #4861, opened 2022-02-24
Avoid Option<DeviceType> in all structs

### LINKED ISSUE #4860: DeviceType serialization and deserialization is a bit confused.
On the back of removing the dupes of DeviceType (#4397), , it turns out there's still significant confusion around `DeviceType` - specifically:

* We [now say that `DeviceType::Unknown` exists purely to avoid `Option<DeviceType>`](https://github.com/mozilla/application-services/blob/862e30d43f3733c514b4ebb0b26046225b3488df/components/support/sync15-traits/src/client.rs#L44-L48), but `Option<DeviceType>` exists in a number of places (including the struct immediately above that comment 🤦 ).
* We have `#[serde(skip_serializing)]` on the `Unknown` variant - so we `Err()` if we attempt to serialize that value. However, we happily serialize `None` and explicitly write `"type": null` in the serialized JSON - which is not what we want (we'd prefer to omit it entirely IMO) - but regardless, having `None` and `DeviceType::Unknown` with different semantics is clearly wrong.

All of which is confusing and a footgun. The options seems to be:
* Remove the `Unknown` variant and embrace `Option<>`
* Remove the `Option`'s

I vote the latter, simply because it being an `Option` is the minority of cases (ie, embracing Option would touch far more structs)

Which still leaves the question of: how do we serialize and deserialize? I vote:

* We continue to fail when attempting to serialize `Unknown` (ie, never serialize `null`). This seems fine, because in practice we never need to serialize records other than our own, and it seems reasonable to insist that we know what type of device *we* have.
* We will continue to deserialize null and unknown values - which implies due to the above that we will fail to round-trip such values via json, but that seems OK.
* If it turned out there was a need to serialize these values, we would opt to omit the field completely (ie, as above, we should never write `null`).
* The above is only for JSON serialization - we might still want/need to write such values to a database (eg, for tabs, where we want to record the tabs for a device, but it fails to provide the type in its client record.) What we do there will be decided on a case-by-case, but valid options might be to write `DeviceType::Mobile`, or even to write the string `Unknown`)
* We needs tests for these cases, which we currently do not.

WDYT?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3085)
┆epic: FxA Ecosystem (backlog)

mhammond:
Fixes #4860 

It's unfortunate that anyone who deserializes needs to remember to add the custom default path, but we can't add `#[serde(default=...)]` to the enum itself because that's only supported for structs. A default deserialize implementation sounds a little like overkill.

Welcome all feedback, so I'm just marking this as a draft now.

### DIFF
--- a/components/sync15/src/client_types.rs
+++ b/components/sync15/src/client_types.rs
@@ -5,7 +5,8 @@
 //! This module has to be here because of some hard-to-avoid hacks done for the
 //! tabs engine... See issue #2590
 
-use serde::{Deserialize, Serialize, Serializer};
+use crate::DeviceType;
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 /// Argument to Store::prepare_for_sync. See comment there for more info. Only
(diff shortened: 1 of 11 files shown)


components/sync15/src/client_types.rs
```diff
@@ -24,60 +24,70 @@ pub struct RemoteClient {
...
+        D: Deserializer<'de>,
+    {
+        Ok(match String::deserialize(deserializer) {
+            Ok(s) => match s.as_str() {
+                "desktop" => DeviceType::Desktop,
+                "mobile" => DeviceType::Mobile,
+                "tablet" => DeviceType::Tablet,
+                "vr" => DeviceType::VR,
+                "tv" => DeviceType::TV,
+                _ => DeviceType::Unknown,
```
bendk, line 65: Should strings that we don't know about map to `DeviceType::Unknown`?  It makes logical sense, but it seems like there could be issues with syncing:

  - Client A sets a field to `DeviceType::BrainImplant`
  - Client B is running an older FF version.  It syncs, doesn't recognize that variant, so it uses `DeviceType::Unknown`
  - Client B syncs again and uploads the record to the server
  - Client A syncs, updates their value to `DeviceType::Unknown`

I'm not sure this can happen right now, but maybe something like that could happen in the future.  Maybe it would be better to throw an error in that case and make client B ignore the record.  I'm sure what's best, so whatever you think is fine with me.  It should be documented in any case though.

bendk (approved):
I like it this way.

I was trying to think of ways to avoid hand-implementing `Deserialize` and `Serialize`, but then decided that I like the impls.  It makes everything clear and explicit.

components/sync15/src/client_types.rs
skhamis: I'm slightly partial to how it is since that device type technically is "unknown" at that point, and any other way might get a little messy trying to support old formats, etc. However, I do agree documenting it would help us down the line.

skhamis (approved):
Nice refactor! Def feels cleaner without the option. Hopefully this also makes it easier to consolidate this type to one central place soon!

mhammond:
Having unknown be, say, `Unknown(String)` seems ideal, but I'm really not sure how that flies with mobile.

mhammond:
I update this in the following way:
* Moved DeviceType to a new device_type.rs and added many comments about the semantics for serializing etc.
* Deserialize "phone" as `DeviceType::Mobile` given https://searchfox.org/mozilla-central/rev/a156a65ced2dae5913ae35a68e9445b8ee7ca457/services/sync/modules/engines/clients.js#292 - I doubt we hit that in practice, but I can't see anything wrong with doing this.

I also tested iOS and Android (although I failed to get Fenix working for reasons unrelated to this. So it seems fine to land.

Issue #3373, opened 2020-07-17
[Tabs] Add a persistent storage for Synced Tabs


jonalmeida:
In Fenix, we want to pull the Synced Tabs feature up to the tabs tray to make it more visible and prominent to the user. A WIP screenshot of what this would look like:

<img width="300" alt="Screen Shot 2020-07-17 at 3 34 34 AM" src="https://user-images.githubusercontent.com/1370580/87816720-ca3c6900-c835-11ea-9538-d736acb7cd38.png">

On a cold boot, we always seem to be retrieving the tabs from the network, so there is a long pause where we have nothing showing up if you go to the tabs tray the first time. We're trying to optimize perceived performance over accuracy of correct tabs (so we're asynchronously requesting new tabs for changes as well), but we would like to show a cached version first.

The feature request/question I'd like to bring up is if it would be prudent to having a persistent storage within the native Rust layer to provide the caching mechanism or if it should live in the a-c layer.

Some thoughts:
 - Synchronizing a database within the a-c and a-s layer can be a bit odd. We would have two copies of the state in different layers that need to be kept up-to-date.
 - Would this caching benefit other Synced Tab consumers in the future?
 - What are the perf losses/gains?
 - This affects the perceived perf of the tabs tray so we can use some smoke and mirrors until we have a better solution since it happens on the cold boot only. That is to say, this might not be a release blocker.
 - What are the constraints of the team for prioritization? We could build something out in a-c if it's quicker but might end up being duplicate work. 

cc: @eoger

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2979)
┆epic: Tabs improvements
┆sprintEndDate: 2022-03-04

eoger:
What's the timeline on this feature @jonalmeida ?

jonalmeida:
@eoger we had already started the a-c and Fenix work to have this land by the end of the sprint, so I this isn't blocking it. I think having a solution landed by the end of August is the current timeline to avoid the (perceived) perf loss.

data-sync-user:
➤ Janet Dragojevic commented:

closing this as a feature request we never acted on.  Please reopen or file a new request if this becomes an active request.

jonalmeida:
@jdragojevic I think we need to re-consider this feature request again with the new H1 work for Synced Tabs in Fenix.

grigoryk:
[...]

PR #4862, opened 2022-02-25
Persist tabs. 

mhammond:
I think this is about ready, but I need to do another self-review - but putting it up for feedback, particularly from @skhamis and @lougeniaC64. Also includes #4859, but I'll rebase once this lands.


mhammond:
Fixes #3373

components/tabs/src/schema.rs
```diff
@@ -0,0 +1,70 @@
...
+// syncChangeCounter/syncStatus nor a mirror etc.
+
+use rusqlite::{Connection, Transaction};
+use sql_support::open_database::{
+    ConnectionInitializer as MigrationLogic, Error as MigrationError, Result as MigrationResult,
+};
+
+// The payload is json and this module doesn't need to deserialize, so we just
+// store each "payload" as a row.
+// We delete the entire DB each sync and re-populate it with every record on
```
lougeniaC64, line 16: nit: I read this as the db is dropped rather than the records are deleted from the db. Maybe, "We delete all the records from the DB each sync."

components/tabs/src/storage.rs
```diff
@@ -7,42 +7,122 @@ const URI_LENGTH_MAX: usize = 65536;
...
+// Tabs has unique requirements for storage:
+// * The "local_tabs" exist only so we can sync them out. There's no facility to
+//   query "local tabs", so there's no need to store these persistently - ie, they
+//   are write-only.
+// * The "remote_tabs" exist purely for incoming items via sync - there's no facility
+//   to set them locally - they are read-only.
+// Note that this means a database is only actually needed after Sync fetches remote tabs,
+// and because sync users are in the minority, the use of a database here is purely
+// optional and created on demand. The implication here is that asking for the "remote tabs"
+// when no database exists is considered a normal situation and just implies no remote tabs exist.
```
lougeniaC64, line 58: Rewording to ensure I understand, `get_remote_tabs` will retrieve the remote tabs as of the last sync (if any exists) but the consumer should call `sync` if they want to get the absolute latest remote tabs, right?

components/tabs/src/storage.rs
```diff
@@ -115,13 +240,27 @@ mod tests {
         assert!(is_url_syncable("https://bobo.com"));
         assert!(is_url_syncable("ftp://bobo.com"));
         assert!(!is_url_syncable("about:blank"));
+        // XXX - this smells wrong - we should insist on a valid complete URL?
```
lougeniaC64, line 243: Is there much difference between a valid url and a syncable url?

lougeniaC64 (commented):
Looks good! Reviewing this PR (particularly with the comments you added) made it easier for me to build a conceptual model of how this sync engine works, specifically how it differs from the other components. I actually like the storage schema as it holds only what's needed, would be easy to extend if the task continuity work calls for it, and adheres to architectural patterns we've already established.

components/tabs/src/storage.rs
mhammond: Yep - which is very similar to how things work now, except that before the first sync we previously returned no tabs whereas now we return what was previously persisted.

components/tabs/src/storage.rs
```diff
@@ -77,20 +157,65 @@ impl TabsStorage {
...
+            }
+            Ok(None) => None,
+            Ok(Some(c)) => {
+                match c.query_rows_and_then_named_cached(
+                    "SELECT payload FROM tabs",
+                    &[],
+                    |row| -> Result<_> { Ok(serde_json::from_str(&row.get::<_, String>(0)?)?) },
+                ) {
+                    Ok(crts) => {
+                        // only return clients that actually have tabs.
```
mhammond, line 174: Best I can tell, the old code doesn't filter out clients with no open tabs - presumably it relies on the consumer doing this, and there might even be a use-case for knowing the clients exist even if they currently have no tabs? So I'm removing this filtering.

skhamis:
👍

lougeniaC64 (approved):
👍🏾

PR #3485, opened 2020-08-07
Do not pass -Dwarnings to Rust on releases CI runs

eoger:
Fixes https://github.com/mozilla/application-services/issues/3480

@MihaiTabara does this looks like the right way to do that to you?

PR #4863, opened 2022-02-25
Taskcluster tweaks [ci full]

bendk:
Some updates to our taskcluster setup.  I think this is a small improvement, but the main reason I'm doing this is just to get experience working with the system.

  - Note:


taskcluster/docker/linux/Dockerfile
```diff
@@ -149,10 +156,37 @@ RUN \
...
 VOLUME /builds/worker/.cache
 
-# run-task expects to run as root
+# Switch to the worker user to install Rust
+# After startup, run-task will downgrade it's permissions to be this user
+USER worker
+
+# Rust setup
+ENV RUST_BACKTRACE=1 \
+    CARGO_INCREMENTAL=1
```
mhammond, line 170: rustup-setup.sh has CARGO_INCREMENTAL as 0 - is the change here intentional? It also sets `RUSTFLAGS='-Dwarnings'` in some cases - we probably still want that?

mhammond (commented):
Looks good in general, but there are a couple of semantic changes I'd like to better understand.

taskcluster/docker/linux/Dockerfile
bendk: The `CARGO_INCREMENTAL` change was unintentional.  I'll fix that and also combine this with the other `ENV` statement above, not sure why I made 2 separate ones.

I wasn't sure about the `-Dwarnings`.  `rustup-setup.sh` was disabling them for releases do we want to keep that?  It feels a bit risky to me to release code that doesn't normally pass CI.  Could we disable the warning from the rust code instead?

bendk:
Just pushed a change that addresses that issue, and also another big one:  once our dockerfile is too big, then the entire pipeline breaks.  So I'm going to try a different approach: have a task that does the build, then uploads the .cargo and .rustup directories as artifacts.  Then the other tasks can depend on that and download those directories before they start.

This is probably more trouble than it's worth, but I'm hoping to use it to understand taskcluster better.  I'm going to flag this PR as a WIP since I'm still working through this.

mhammond:
> It feels a bit risky to me to release code that doesn't normally pass CI.

I'm inclined to agree with that, but IIUC, the change here is that we now allow "normal" CI runs to pass where they previously would have failed, which I don't think is the right tradeoff.

I seem to recall that we ignored that because some dot-release we needed to make caused warnings that didn't exist before - but that implies something else changed that shouldn't have - quite possibly the rust version, as I suspect that predates rust-toolchain. It was done intentionally in 456afade6965faba4a7d97be196cc136e0305bd5, but I can't find a matching PR nor any other discussion, so 🤷 - but I guess the tl;dr is that I think CI should enforce warnings and I'm fine with releases also doing that until we rediscover why it should not :)

mhammond:
> but I can't find a matching PR nor any other discussion,

It's actually #3485/#3480 - in particular:

> When cherry-picking things for hotfixes, it's okay if we don't fix warnings only present in newer versions of Rust (given that it's a cherry-pick, we can be reasonably confident that when the commits landed originally there were no lint errors).

bendk:
Oops, I had the logic reversed for `-Dwarnings`.  I'm going to set that in the dockerfile.  It seems like nowadays we can include that for releases as well.  That cherry-picking issue shouldn't be a problem now that we have `rust-toolchain` to pin the versions.

bendk:
@mhammond Okay, I finally have this working.  It now installs the rust toolchain in a task rather than as part of the Dockerfile.  This keep the docker image below the soft limit where tasks start to fail.  Can you review again and tell me what you think?

mhammond (approved):
LGTM!

PR #4866, opened 2022-03-07
Bump versions of `rusqlite`/`libsqlite3-sys` in `Cargo.toml`s

### LINKED ISSUE #4342: SQLite 3.36 changes the output of `EXPLAIN QUERY PLAN`, breaking the code in `sql_support::query_plan`.
In the next version of SQLite, the format of the output of `EXPLAIN QUERY PLAN` changed: https://www.sqlite.org/releaselog/3_36_0.html

In the past, I wrote a parser for the output of this format here: https://github.com/mozilla/application-services/blob/main/components/support/sql/src/query_plan.rs, which I believe is used in test code. This code will likely panic, or return Err, under the new format, depending on how different it is.

This is just a heads up that a future version of rusqlite will change the bundled version of SQLite (which, IIRC, is also used in test/CI configurations), and so could cause some breakage when you update.

I think the best option would either be for you to remove this code, or rewrite it for the new format. It was useful to have in the past, but SQLite doesn't guarantee the format of the response is stable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2693)

thomcc:

  - Note:

Hi. I am the maintainer of rusqlite. Unrelatedly, I also used to work here, and wrote portions of the code in this repo (I'm not sure how many people I know still work here, though). Anyway, the fact that Firefox and application-services are on such an old versions of `rusqlite` makes me very sad, and I suspect nobody is going to fix it until it becomes an actual problem. So, I've done both of us a favor and updated things. 

`cargo test --all` and `cargo clippy --all` pass locally. I'll allow your CI to let me know about the rest (I don't feel like installing android tools). Note that rusqlite supports Rust `1.57`, and I'll keep firefox's version policy in mind before updating the MSRV, since I like that Firefox is using Rusqlite. (Actually, I already was considering this, but it was moot because you have apparently been stranded in the past).

### Bulk of the changes

So, yeah, this took a lot of changes, essentially all of them are mechanical, with the possible exception of the stuff in `sql-support`. (Honestly, even the ones in sql-support were pretty mechanical, although perhaps I should have thought more when doing them. Who's to say).

Mainly, the changes are updating this code past "The Great Parameter Overhaul Of v0.25.0", which I hadn't realized would break you quite as bad as it did (I didn't intend for the `&[(":foo", &bar)]` use of named params to be a problem, but it hits an inference error I describe below). That said, I did know it was large and breaking, and if it makes you feel any better, I don't think there will be breakage like that in the future (it was necessary to clean up the API a great deal[^cleanup]).

[^cleanup]: Broadly speaking, the outcome of that was the deprecation of `NO_PARAMS` (now you can just pass in `[]`), methods like `execute_named`, `query_row_named`, etc (now they take [`impl Params`](https://docs.rs/rusqlite/latest/rusqlite/trait.Params.html), which supports, among other things, `&[(&str, &dyn ToSql)]`). One note is that you can't directly pass a `impl IntoIterator<impl ToSql>` to these anymore, and have to wrap it in [`rusqlite::params_from_iter`](https://docs.rs/rusqlite/latest/rusqlite/struct.ParamsFromIter.html). These sorts of things are most of the changes in this PR.

One unfortunate thing with that change is that `conn.execute(&[(":foo", &foo), (":bar", &bar)])` will generally hit an inference failure. It makes some sense[^sense] that this happens, but it's annoying. I think people should just use `rusqlite::named_params!` for these, but may find a way to make this better someday. That said, I didn't move you all the way to `named_params!`, since it is tedious to make that change (it requires changing code in on *several lines* 🙀). Instead, in most places I just added `as &dyn rusqlite::ToSql` on the first value (which is enough for rustc to get past the inference confusion), which I can just paste in on the start. That said, if I worked here still, I'd probably go through the trouble, and IMO you should file an issue[^issue] to change the code that does `&[(":foo", &blah as &dyn rusqlite::ToSql), ...]` to use `rusqlite::named_params!` instead, which is much cleaner.

[^sense]: Basically, while `rusqlite` *does* have a `Params` impl for `&[(&str, &dyn ToSql)]` (and `&[(&str, &dyn ToSql); N]`, which is actually what `&[...]` generally needs in generic contexts), for your usage, the compiler sees`&[(":foo", &5i64), (":bar", &format!("..."))]` and thinks that you've written an array where the first item is a `(&str, &i64)`, and the second is a `(&str, &String)`, so it complains (previously this worked because it had the "hint" of `&[(&str, &dyn ToSql)]` in the destination to help it out, but now it just has `impl Params`, which could be anything).

### Other things in this patch

Two other things I changed are:

- Removal of the `log_query_plans` feature from `sql-support`. I actually wrote a pretty good commit message about it, which I won't bore you by copypasting here. Basically, this fixes #4342. 

- I added a large comment about `UncheckedTransaction`. I actually think I added this before `0.24.1`, so this was already true. I tried to remove it all together but it was a hassle. So, that's your guys problem to worry about if you care. It doesn't matter much to me.

Anyway, I'm in your chat.mozilla.org channel now (as the user "zuurr"), so you can ping me there if you have issues. Or just comment here.

P.S. Did you know that the code in `components/autofill/src/sync/tests` is all completely dead? It wasn't even syntactically valid, although I fixed that because rust-analyzer was upset. You might be missing a `#[cfg(test)] mod tests;` or something  in the `sync` module. I may have also changed some of this as part of some of the big find and replaces I did, but it's still dead code. I didn't change that.

[^issue]: Possibly mark it as "good first bug", since it would be easy for someone with Rust experience. You can even ask that it be added as a Call For Participation in "This Week In Rust" by listing it https://users.rust-lang.org/t/twir-call-for-participation/4821 -- stuff that gets added there usually gets picked up if it's not too difficult.


mhammond (dismissed):
Thanks heaps Thom, this is great. There are some CI issues which I'll fix, but this is very helpful and I agree with everything you did!

thomcc:
I bumped the regex version in Cargo.lock for you.

mhammond (approved):
Thanks heaps Thom.

PR #4868, opened 2022-03-08
oh my god why is firefox on such an old rusqlite

mhammond:
Attempting to fix CI issues from #4866


thomcc:
I've pushed an update to my branch, but I do appreciate you taking it over if that has even more issues. In particular I don't actually have an android or ios dev environment at the moment.

PR #4869, opened 2022-03-10
cargo update

mhammond:
Thom suggested we update parking-lot, and I figured rather than just updating that, why not do a full `cargo update`? Anyone see a reason not to do this?


skhamis (approved):
Looking at the Cargo.lock, nothing seems like a wildly version bump and the `cargo build` and `cargo test` seem to work fine. LGTM!

PR #4870, opened 2022-03-10
Bootstrap Taskgraph from pypi in decision task

ahal:
Releng has tweaked the recommended way of bootstrapping Taskgraph from a
clone + checkout pinned revision, to Pypi and a requirements.txt.

This provides a few benefits:
1. Since Taskgraph uses SemVer, updating will provide a hint if there
   will be backwards incompatible changes.
2. Using version numbers is nicer than revisions (e.g, Taskgraph's
   CHANGELOG will be useful).
3. Provides a mechanism to pin other non-Taskgraph dependencies.

The reason I'm submitting this now is I'd like to migrate Taskgraph to a
new repository. By moving consumers over to Pypi first, the migration
can happen seamlessly without breaking anything.

Jira: RELENG-816


ahal:
I recommend reviewing this with the "Ignore whitespace" option checked.

ahal:
I don't *think* the toolchain-rust failure is mine, but could be wrong. Does look like it was recently added.

mhammond:
Yeah, looks like some fallout from #4863 - @bendk will probably have a look tomorrow.

bendk:
It looks like the checkout directory changed from `/builds/worker/checkouts/src` to `/builds/worker/checkouts/vcs`.  I'm not sure why, but I think it was caused by these changes somehow.

Logs:
https://firefox-ci-tc.services.mozilla.com/tasks/LZRl2gzERlSrRb1GGrMM2w/runs/0/logs/public/logs/live.log

ahal:
> It looks like the checkout directory changed from `/builds/worker/checkouts/src` to `/builds/worker/checkouts/vcs`. I'm not sure why, but I think it was caused by these changes somehow.
> 
> Logs: https://firefox-ci-tc.services.mozilla.com/tasks/LZRl2gzERlSrRb1GGrMM2w/runs/0/logs/public/logs/live.log

Ah, yes! These changes upgrade Taskgraph to the latest version, and switching the path of the checkouts folder there landed sometime last summer. So guess it's due to my patch after all, I'll fix it.

I'll note that this backwards incompatible change landed before we started releasing to Pypi.. going forward we'll be following SemVer strictly.

ahal:
Alright, this should be ready for review now (I don't have permission to request it).

bendk (approved):
LGTM.

The script changes seem fine.   I don't understand the `.taskcluster.yml` change, but none of us understood that file before either :).  If CI is passing, then I think it's good.

ahal:
Thanks, I'll need help landing this as well as I don't have permission.

ahal:
Oh oops, I clicked the rebase button and it invalidated your review, sorry :/

bendk:
No problem, I'll just rebased again and merged.

PR #4891, opened 2022-03-28
Runs cargo update to fix audit run [ci full]

tarikeshaq:
Our dependency check cron job started failing, so this is a PR to fix that by running `cargo update`

The audit that failed was for openssl, which we **only use in tests**.


tarikeshaq:
Uhh looks like the cargo update on `native-tls` broke our iOS builds, no biggie we're not impacted by the advisory. I'll keep an eye on https://github.com/sfackler/rust-native-tls/issues/218 and land this once it's fixed. We can also fix the version temporarily to be =0.2.8 but there is no urgency to this 🤷

mhammond (approved):

PR #4897, opened 2022-04-02
Bump SQLcipher version to 4.5.1 (equivalent to SQLite 3.37.2)

thomcc:
Hi. I would like to (eventually) change `rusqlite` to default to disallowing use of SQLite versions with known memory safety issues, since arguably, many of the APIs should be considered `unsafe` without this check.  I have not figured out all the details here, but currently that means I'd require SQLite 3.34.1 because it fixes a use-after-free (which could be exploitable if you run untrusted SQL -- you don't, but still) according to https://sqlite.org/releaselog/3_34_1.html. (This is also the version that windows's bundled SQLite and debian stable both seem to use, and I'm guessing that's why).

This means by default `rusqlite` would complain about your version of SQLcipher, since SQLcipher 4.4.0 is equivalent to SQLite 3.31.0 (https://www.zetetic.net/blog/2020/05/12/sqlcipher-440-release/), which may have the issue. In order to avoid giving myself a headache by worrying about it, I'm just going to just update you. This is hopefully a 2 line change and was a really old version anyway. (It also kind of seems sketchy for firefox to use an old version that has known security holes, even if the library is only used for migrations, and they shouldn't apply to your usage)

This builds and seems to work locally, but I don't have an android dev environment, and may have done something wrong. If it's too much of a quagmire, please feel free to take it over. Also, note that if this is too aggressive of an update, you could also only update to SQLcipher 4.4.3, which should be equivalent to SQLite 3.34.1, which has the use-after-free fix mentioned.

(Long term, a better option *eventually* would probably be for you to use the `libsqlite3-sys/bundled-sqlcipher` feature, but I'm not 100% sure how to make that happen with your NSS usage. Eventually `libsqlite3-sys` should support build script overrides which would solve that, but for now it does not...)

  - Note:


mhammond (approved):
Thanks Thom - I really should do a better job at keeping my eye on our deps - part of the problem is that doing an update ends up changing the set of licenses in ways that are a PITA, but that's a poor excuse and doesn't even apply here)

PR #58, opened 2018-05-22
Implement Synchronization

thomcc:
This is still WIP, but it's quite close and it seems good to get feedback on it now.

Still working on: FFI changes, tests.

PR #4905, opened 2022-04-18
Update tabs component for iOS integration

lougeniaC64:
This PR includes a few updates to the tab component to allow the integration of the tabs component in iOS. It includes:

- Resolving the naming conflict of the `DeviceType` enum that is exposed in FxA and Tabs which prevented `rust-components-swift` from generating Tabs code
- Exposing the tabs `reset` function and errors
- Updating the sync function's parameters to match the `SyncUnlockInfo` parameters
- Fixing the tabs sync example

**Note:** This PR blocks [RCS PR #58](https://github.com/mozilla/rust-components-swift/pull/58) and [Firefox iOS PR #10553](https://github.com/mozilla-mobile/firefox-ios/pull/10553).

  - Note:

### DIFF
--- a/components/tabs/ios/Tabs/Tabs.swift
+++ b/components/tabs/ios/Tabs/Tabs.swift
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+open class TabsStorage {
+    private var store: TabsStore
+    private let queue = DispatchQueue(label: "com.mozilla.tabs-storage")
+
+    public init(databasePath: String) {
+        store = TabsStore(path: databasePath)
+    }
+
+    /// Get all tabs by client.
+    open func getAll() -> [ClientRemoteTabs] {
+        return queue.sync {
+            return self.store.getAll()
+        }
+    }
+
+    /// Set the local tabs.
+    open func setLocalTabs(remoteTabs: [RemoteTabRecord]) {
+        queue.sync {
+            self.store.setLocalTabs(remoteTabs: remoteTabs)
+        }
+    }
+
+    open func reset() throws {
+        try queue.sync {
+            try self.store.reset()
+        }
+    }
+
+    open func sync(unlockInfo: SyncUnlockInfo) throws -> String {
+        return try queue.sync {
+            return try self.store
+                .sync(
+                    keyId: unlockInfo.kid,
+                    accessToken: unlockInfo.fxaAccessToken,
+                    syncKey: unlockInfo.syncKey,
+                    tokenserverUrl: unlockInfo.tokenserverURL,
+                    localId: unlockInfo.tabsLocalId
+                )
+        }
+    }
+}
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -16,6 +16,9 @@ use sql_support::ConnExt;
 use std::cell::RefCell;
 use std::path::{Path, PathBuf};
 
+pub type TabsDeviceType = crate::DeviceType;
+pub type RemoteTabRecord = RemoteTab;
+
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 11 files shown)


components/tabs/src/storage.rs
```diff
@@ -32,18 +33,18 @@ pub struct ClientRemoteTabs {
         default = "devicetype_default_deser",
         skip_serializing_if = "devicetype_is_unknown"
     )]
-    pub device_type: DeviceType,
+    pub device_type: TabsDeviceType,
```
mhammond, line 36: I think we should try and keep this code using `DeviceType` where we can - the `type` alias above is probably needed to keep uniffi happy, but there should be no need to otherwise touch the .rs.  (IOW, we should see this rename as a short term hack, so should make it as easy to back away from)

examples/tabs-sync/src/tabs-sync.rs
```diff
@@ -42,6 +42,12 @@ fn main() -> Result<()> {
     cli_support::init_logging();
     let opts = Opts::from_args();
 
+    let (_, token_info) = match get_account_and_token(get_default_fxa_config(), &opts.creds_file) {
```
mhammond, line 45: I wonder if just using `?` here instead of explicitly handing the `Err` case works? Probably fine for an example. (Not that the changes here are wrong, just longer than might be necessary)

components/tabs/src/tabs.udl
```diff
@@ -12,13 +21,16 @@ interface TabsStore {
 
     [Self=ByArc]
     void register_with_sync_manager();
+
+    [Throws=TabsError, Self=ByArc]
```
mhammond, line 25: I wonder if we should also allow the contructor to return this error? That's another breaking change, but it seems crazy that a sync function can return an error about a database being corrupt but the constructor will silence it?

mhammond (commented):
I only had time for a quick look/think, and I think this will be fine if necessary, but working out a better story for that type would be ideal! Also welcome your thoughts on having more functions return those tab errors? Or maybe sync should just log and swallow them too, making the error changes unnecessary? Not sure which one is best, but this new "mixed-messages" model seems less than ideal!

components/tabs/src/tabs.udl
lougeniaC64: So I assumed because the `TabsStore::new` returns `Self` instead of `Result` that it doesn't return an error, which means it shouldn't (or wouldn't) throw a `TabsError`, let me know if there are otther considerations here. And while I do think it's off for the sync function to throw while the constructor doesn't, we do have a precedent for throwing on the sync functions we expose.

mhammond: oops, no, I thought we caught and ignored errors there, but apparently not!

components/tabs/src/sync/store.rs
```diff
@@ -103,7 +116,7 @@ impl TabsStore {
             return Err(e.into());
         }
         match result.engine_results.remove("tabs") {
-            None | Some(Ok(())) => Ok(result.telemetry),
+            None | Some(Ok(())) => Ok(serde_json::to_string(&result.telemetry).unwrap()),
```
mhammond, line 119: We expose an error for serde errors, so I think this `.unwrap()` can be replaced with `?`

mhammond (approved):
LGTM! 🚢

components/tabs/ios/Tabs/Tabs.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    /// Set the local tabs.
+    open func setLocalTabs(remoteTabs: [RemoteTabRecord]) {
+        queue.sync {
+            self.store.setLocalTabs(remoteTabs: remoteTabs)
+        }
+    }
+
+    /// Register with the sync manager
+    open func registerWithSyncManager() {
```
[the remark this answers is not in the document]
lougeniaC64: Yeah, I was trying to be forward thinking here but this is a very fair point. I'll remove it.

lougeniaC64:
@nbhasin2 I'd appreciate your review of the [Tabs.swift](https://github.com/mozilla/application-services/pull/4905/files#diff-47356eeec8faf0937bb7c8305d3d2ab3c864af121f5fcb193a83bed53fb62c13) file in this PR.

Issue #4856, opened 2022-02-22
Error during PlacesHistoryStorage$deleteEverything: Query returned no rows


grigoryk:
Low-ish volume on Nightly. https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/17840159/?query=is%3Aunresolved%20level%3Afatal%20InternalPanic 
```
mozilla.appservices.places.uniffi.InternalPanic: Unexpected error: Error executing SQL: Query returned no rows
    at mozilla.appservices.places.uniffi.PlacesException$ErrorHandler.read(places.kt:3)
    at mozilla.appservices.places.uniffi.PlacesException$ErrorHandler$lift$1.invoke(places.kt:2)
    at mozilla.appservices.places.uniffi.PlacesException$ErrorHandler$lift$1.invoke(places.kt:1)
    at mozilla.appservices.places.uniffi.PlacesKt.liftFromRustBuffer(places.kt:2)
    at mozilla.appservices.places.uniffi.PlacesException$ErrorHandler.lift(places.kt:2)
    at mozilla.appservices.places.uniffi.PlacesException$ErrorHandler.lift(places.kt:1)
    at mozilla.appservices.places.uniffi.PlacesConnection.deleteEverythingHistory(places.kt:16)
    at mozilla.appservices.places.PlacesWriterConnection.deleteEverything(PlacesConnection.kt:3)
    at mozilla.components.browser.storage.sync.PlacesHistoryStorage$deleteEverything$2.invokeSuspend(PlacesHistoryStorage.kt:2)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:18)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
    at java.lang.Thread.run(Thread.java:764)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3082)
┆Sprint End Date: 2022-03-04

grigoryk:
Currently top 14 crash for Nightly.

[...]

PR #4907, opened 2022-04-19
Downgrade places api log from error to warn

skhamis:
The reason this is getting downgraded is due to the current state of what's happening https://github.com/mozilla/application-services/issues/4856

Summary steps:
1. Android calls tabs engine sync only (for MR1 homescreen sync work)
2. sync manager eventually calls `iter_registered_engines` which calls every engine
  - This causes each engine to run `get_registered_sync_engine`
    - most engines will return None if the engine lost it's connection
    - However, places runs a `log:error` which what is hitting sentry
3. Sync manager filters based on SyncEngineSelection
4. tabs successfully sync 

While there is a bigger discussion to be had about what `iter_registered_engines` should be doing but for now, we should lower how noisy sentry is becoming with this patch.

  - Note:

### DIFF
--- a/components/places/src/api/places_api.rs
+++ b/components/places/src/api/places_api.rs
@@ -51,7 +51,7 @@ pub fn get_registered_sync_engine(engine_id: &SyncEngineId) -> Option<Box<dyn Sy
         Some(places_api) => match create_sync_engine(&places_api, engine_id) {
             Ok(engine) => Some(engine),
             Err(e) => {
-                log::error!("places: get_registered_sync_engine: {}", e);
+                log::warn!("places: get_registered_sync_engine: {}", e);
                 None
             }
         },


mhammond:
To capture some of the slack convos between Sammy and I:

* Best we can tell, this is ending up in sentry only due to the `log::error!` call - the syncs actually happen correctly. All other engines can be in this state, but only places logs an error (most, if not all, other engines are completely silent about it)
* This scenario is not one we really expect, particularly when tracking engine changes. For example, if the most recent sync from a different device did a restore of sync bookmarks, then this device syncs just tabs while in this state, I'm not sure the right thing will happen in terms of bookmarks on this device. The tl;dr of this concern is "sync manager has a `reset_all()` method that will operate differently in these scenarios and it's not expecting to"

Meaning ultimately we need to lean into 1 or 2 scenarios:

* We do allow syncing in this scenario (ie, then engines which should be available are not) then we should embrace it, and make sure that everything works as we expect. The tl;dr of this is "sync manager should not have a `reset_all()` method at all, given what engines are available to reset isn't deterministic"

* We decide things like `reset_all()` really aren't going to work correctly, so therefore we need to ensure a-c never kills these engines so long as the app is alive. A problem with this scenario is that there might be temporary reasons engines don't exist - eg, let's say we can't create places because we are out of space, but on the next run we can - bookmarks missing a `reset_all()` in that scenario would be as bad as it is here.

So for this reason, I think this is the simplest patch we can make that will put out the current fire, then dig into which of the 2 scenarios above we should support.

mhammond (approved):

PR #4912, opened 2022-04-20
Fix fml optional enums and objects

tarikeshaq:
@bendk caught a bug where optional enums generated code that didn't compile which prevented us from upgrading AS in Fenix

I'm concerned that we didn't catch this earlier, especially with https://github.com/mozilla-mobile/fenix/blob/main/nimbus.fml.yaml#L71 in Fenix having an optional enum - @jhugman is it possible that this broke recently (I don't think so?)

Regardless this PR attempts to fix the problem and adds test coverage for optionals (including enums and objects)

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,6 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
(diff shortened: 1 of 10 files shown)


badboy:
fwiw without reviewing this: I had to run a local fenix build with updated a-s and this PR was fixing the last build issue I had.

tarikeshaq:
https://github.com/mozilla/application-services/pull/4915 fixes this too, so I'll close this PR and get that one landed

PR #4916, opened 2022-04-20
Cut release v93.0.1

tarikeshaq:
Cuts v93.0.1 with the Nimbus FML bug fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 ### What's Fixed
   - Handling of optional types which require a mapping to a usable type. ([#4915](https://github.com/mozilla/application-services/pull/4915))
 
+## Places
+
+- Downgraded places `get_registered_sync_engine` `log:error` to `log:warn` to fix an issue where places was unnecessarily creating sentry noise. This change was also cherry-picked to [v91.1.2](#v9112-2022-04-19)
+
 # v93.0.0 (_2022-04-13_)
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v92.0.1...v93.0.0)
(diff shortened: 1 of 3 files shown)


skhamis:
> This change looks good, but can you also merge the changes from #4913? I think this is just the logging change from [5a2e3c5](https://github.com/mozilla/application-services/commit/5a2e3c5536ee7e3f6134e8050f83dc0f30113322), but it would be good to get that into this release as well.

Yeah just the logging change would be good to prevent regression since that's already out in a-c. This is a better PR if we don't want to cherry-pick the commit: https://github.com/mozilla/application-services/pull/4917

tarikeshaq:
alrighty no worries, I'll rebase this on top of #4917 when it lands

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# v93.0.1 (_2022-04-20_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v93.0.0...v93.0.1)
+
+## Nimbus FML ⛅️🔬🔭🔧
+
+### What's Fixed
+  - Handling of optional types which require a mapping to a usable type. ([#4915](https://github.com/mozilla/application-services/pull/4915))
+
```
bendk, line 9: What should we do about the changelog entry from `v91.1.2`?

```
## Places
- Downgraded places `get_registered_sync_engine` `log:error` to `log:warn` to fix an issue where places was unnecessarily creating sentry noise
```

I think it's slightly confusing now, because users might assume that that fix was in `v93.0.0` as well, which is incorrect.  I'm not sure what the best way to do this, but I think I favor removing the `v91.1.2` section and moving that note to the `v93.0.1` section.   That way the changelog reflects the fact that `v91.1.2` was from a branch that diverged from `main`.

bendk (approved):
Looks good, thanks for working through this one with me.  I had one comment about the changelog, but I'm really unsure what the best format is.  Feel free to address it however you want.

CHANGELOG.md
tarikeshaq: I would opt to keep it under v91.1.2 and duplicate it under v93.0.1, and note down explicitly in both places that v91.1.2 was a release not off the main branch

Just in the unlikely case that someone is looking through our changelog in `main` for information about v91.1.2 🤷

Issue #4908, opened 2022-04-19
Expose deleteVisitsFor in our swift wrappers


electricRGB:
AS gives us `deleteHistoryMetadata(key...` which is great, but has an issue: It requires matching a full key (ie, the url, the search term and the referrer url) to delete something. So if we want to delete something from the history (where on Firefox-iOS we might not have the search term and the referrer url) we have to do what Android is currently doing.
 
<img width="913" alt="Screen Shot 2022-03-30 at 12 04 49 PM" src="https://user-images.githubusercontent.com/11182210/164046124-b36a2207-27db-4e9c-9193-9f05bd28e4f5.png">

The problem is that this code ends up being very messy, and, especially with how many operations we're doing on the homepage, we want to really lessen the performance hit as much as possible.

The request: given that URLs are unique when it comes to metadata, we'd like a API that allows us to pass in a URL (or a set of URLs), without the need for a search term or a referral url to easily delete metadata items.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3142)
┆Story Points: 1
┆sprintEndDate: 2022-05-13

mhammond:
Can you please explain a little more about how you are using this? Not that I want to push back too hard here, but the history-metadata stuff started out as an experiment, has diverged from desktop's similar API, and I fear it is a future footgun, and I'm wondering if it is still fit for purpose?

In particular, are you looking to delete all history about that URL, or just stuff in history_metadata but leave all visits alone? If the latter, I'm trying to work out what the use-case is specifically?

electricRGB:
Sure. 

So, essentially, we have two flows for deleting a history item from a user's perspective:
1. deleting from our Recently Visited section (which is a list of history highlights that we get from AS)
2. deleting from the history panel

Because on iOS we have our own local history, regardless of flow, the url the user is deleting needs to be deleted simultaneously from both data sources so that the url doesn't get deleted from history, but still shows up in our history highlights. Deleting an item from Recently Visited and having it be deleted from History is easy as we already have URLs to do that.

However, deleting an item from History means we only have a url. The current API to delete a metadata item requires a `HistoryMetadataKey` to be passed in and it needs to match fully before an item can be deleted, according to the unit tests I wrote.

[...]

PR #4918, opened 2022-04-21
places: delete_visits_for() now also deletes history metadata when the item is bookmarked

mhammond:
While looking into #4908, I noticed that `delete_visits_for()` also deleted history_metadata as a side-effect when the `moz_places` row is deleted due to an [`ON DELETE CASCADE`](https://github.com/mozilla/application-services/blob/1f01918d43ba2eba4bac921724f2be26e09ab249/components/places/sql/create_shared_schema.sql#L261) clause. The "problem" is that if the page is bookmarked, the `moz_places` row isn't deleted. This patch fixes that, and adds fairly extensive tests for this case.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -6,6 +6,11 @@
 
 <!-- WARNING: New entries should be added below this comment to ensure the `./automation/prepare-release.py` script works as expected.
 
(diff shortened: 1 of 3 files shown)


lougeniaC64 (approved):
LGTM!

PR #4920, opened 2022-04-21
Fixes map on String and Variables being internal

tarikeshaq:
Trying to upgrade ios to 93.0.1 in https://github.com/mozilla-mobile/firefox-ios/pull/10297 the build failed with this error
it's only an issue with 93.0.1

I'll cut 93.0.2 after this lands and get it bumped over there


tarikeshaq:
I tested locally and verified this fixes the build

jhugman (approved):
This looks perfect.

PR #4924, opened 2022-04-26
Make FeatureHolder threadsafe

jhugman:
This PR ensures that FeatureHolder on both Android and iOS are stateless, and thus threadsafe.

Fixes [EXP-2384][1].

[1]:  https://mozilla-hub.atlassian.net/browse/EXP-2384

  - Note:


tarikeshaq (approved):
LGTM

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
```diff
@@ -27,24 +25,16 @@ class FeatureHolder<T>(
...
-        val variables = api?.getVariables(featureId, false) ?: NullVariables.instance
-        val feature = create(variables)
-        api?.also { apiRef ->
-            val weakRef = WeakReference(apiRef)
-            exposureRecorder = {
-                weakRef.get()?.recordExposureEvent(featureId)
-            }
-        }
-        return feature
+        val variables = getSdk()?.getVariables(featureId, false) ?: NullVariables.instance
```
csadilek, line 28: I am being a bit pedantic here, but since `getSdk` is a lambda provided by the consumer of this API, the consumer can still cause thread-safety issues of this method e.g., the lambda provided by the caller could have side-effects.

We could document this here on the API level and call out, or alternatively, I am wondering: Does this need to be a lambda or could we provide an instance of `FeaturesInterface` directly? Looking at it, I assume the motivation is lazy initialization, is that right? 

In practice this is probably not a concern, as the caller of this method is in generated code i.e., the lambda is not provided by application code directly.

csadilek (commented):
Nice refactor! I have one topic/question below. I'll test this out locally as well.

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -339,6 +339,28 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn regresion_test_concurrent_access_of_feature_holder_swift() -> Result<()> {
```
csadilek, line 343: nit: regres**s**ion_

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -339,6 +339,28 @@ mod test {
...
+            "test/threadsafe_feature_holder.swift",
+            "fixtures/fe/fenix.yaml",
+            "release",
+            false,
+        )?;
+        Ok(())
+    }
+
+    #[test]
+    fn regresion_test_concurrent_access_of_feature_holder_kts() -> Result<()> {
```
csadilek, line 354: nit: regres**s**ion_

csadilek (approved):
Tested before/after locally to confirm the fix. 🚢 ✨

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
jhugman: Ah yes, I'd forgotten you'd asked this question.

As you assumed: the lambda is for lazy initialization.

We're not in control of where the singleton is, or when it gets instantiated, but the generated code needs to be able to talk to the singleton.

This design lets the app tell us where the SDK singleton is. Unfortunately, this means that there is a small window of time where the SDK is not available, but app code requires configuration OR a startup performance cost for the SDK which may not be needed until much later on.

The followup to this is to make the app pass the singleton as a lambda or a `Lazy<FeaturesInterface>`, so the generated code will always have access to the SDK.

csadilek: OK, having the app pass a `Lazy<FeaturesInterface>` sounds good to me. That's a pattern we already use to pass storage instances around that are costly to initialize. Happy to talk this through more.

PR #4933, opened 2022-05-02
Update rand and parking lot

emilio:

  - Note:
- [x] **Tests**: No behavior change, keeps passing existing tests.
- [x] **Changelog**: I don't _think_ dependency updates are worth putting in the changelog but happy to do so.


tarikeshaq (approved):
Awesome!! thank you!! I'll land this once CI passes 🚀 🎉

emilio:
@tarikeshaq Some dependency check failed, but I don't think it's my fault? I didn't change any crate or version related to windows-rs or anything like that.

emilio:
Ok I can reproduce the failure. I think there is a pre-existing issue with how that script looks for licenses but for now I just made it happy. @tarikeshaq mind taking another look? Thank you

tools/dependency_summary.py
```diff
@@ -345,6 +345,33 @@
...
+    },
+    "windows_x86_64_gnu": {
+        "repository": {
+            "check": "https://github.com/microsoft/windows-rs",
+        },
+        "license_file": {
+            "check": None,
+            "fixup": "https://raw.githubusercontent.com/microsoft/windows-rs/master/.github/license-mit",
+        }
+    },
```
tarikeshaq, line 374: 💯

tarikeshaq (dismissed):
Looks great! It happens every once in a while when we bump dependencies that the tree ends up somewhere without a clear license file - thanks a ton @emilio for taking the time to clean it up

tools/dependency_summary.py
tarikeshaq: filed #4936, we'll take a look if there is a problem with the script 😬

tarikeshaq:
@emilio if you run the `./tools/regenerate_dependency_summaries.sh` script, and check-in the changes to the dependency summaries the dependency check should pass

emilio:
Thanks, and sorry for the lag, I was digging into #4934 :)

Should be done.

emilio:
@tarikeshaq it seems the script generates different output on macOS and Linux, any chance you could run it and land this for me?

mhammond:
> it seems the script generates different output on macOS and Linux

I strike this all the time (but I'm Windows!) - I tend to just take the patch generated by CI, apply it manually, re-push and let it go green :)

tarikeshaq:
Yeah... unfortunate it's like that now to catch the iOS dependencies (I think? I haven't really looked into it, @mhammond may have the historical context 😛)

Regardless, I created #4937 with the same commits and one on top for running the generation on a mac, and I can land it there - thanks a ton for taking the time on this @emilio 🙏

mhammond:
> Yeah... unfortunate it's like that now to catch the iOS dependencies (I think? I haven't really looked into it, @mhammond may have the historical context 😛)

I don't, but I suspect you are correct. Another option is that it was initially developed on a mac and we just ran with it (although that seems less likely tbh). I'd love to see this die, but it's done for some vague legal requirement I have no interest in understanding at this point, so 🙈

tarikeshaq:
Alrighty, this was landed in https://github.com/mozilla/application-services/pull/4937 so closing here 😄 Thanks again!!

Issue #3287, opened 2020-06-28
Turn rusqlite's `unlock_notify` feature on in crates that use rusqlite.


thomcc:
This will bring us into par with how [mozStorage](https://searchfox.org/mozilla-central/source/storage/mozStorageConnection.cpp#1084-1099) performs sqlite3_step, and has a chance to reduce the frequent SQLITE_BUSY errors (https://github.com/mozilla/application-services/issues/2873) we see. However, I would not count on this, but even if it does not cause this, and a more convincing reason to do it is that it should improve latency when one connection holds a lock that another connection is waiting on.

Note that given our mildly heavy use of execute_batch we may want to wait for the next rusqlite release (a week or two) to do this, as execute_batch will not participate in unlock notifications as prior to recently they went through sqlite3_exec and not sqlite3_step. That said, there's no real harm in turning it on before then, it just won't do as much good. (I'm filing this now mostly because the code just landed in rusqlite for this, which reminded me, although I've thought we should turn this on for a while now)

This is a relatively trivial change aside from that, and just requires adding a cargo feature to the crates that use rusqlite.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1429)
┆Epic: important not urgent
┆Sprint: SYNC - end 2020-07-17

thomcc:
Thinking more, I think it actually could reduce SQLITE_BUSY errors, since this will cause connections to give up locks sooner than they otherwise would.

PR #4934, opened 2022-05-02
[ci full] Remove unused unlock_notify feature from rustqlite.

emilio:
This was causing my build to fail locally with an undefined reference to
sqlite3_unlock_notify. It seems nothing makes use of this feature tho.

  - Note:
- [x] **Tests**: No behavior change.
- [x] **Changelog**: Probably not worth putting this in the changelog.

### DIFF
--- a/components/autofill/Cargo.toml
+++ b/components/autofill/Cargo.toml
@@ -28,7 +28,7 @@ url = { version = "2.2", features = ["serde"] }
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["functions", "bundled", "serde_json", "unlock_notify"]
+features = ["functions", "bundled", "serde_json"]
 
 [dev-dependencies]
 env_logger = { version = "0.7", default-features = false }
--- a/components/logins/Cargo.toml
+++ b/components/logins/Cargo.toml
@@ -30,7 +30,7 @@ uniffi_macros = "^0.17"
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["sqlcipher", "limits", "unlock_notify"]
+features = ["sqlcipher", "limits"]
 
 [build-dependencies]
 uniffi_build = { version = "^0.17", features=["builtin-bindgen"] }
--- a/components/places/Cargo.toml
+++ b/components/places/Cargo.toml
@@ -36,7 +36,7 @@ uniffi_macros = "^0.17"
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["functions", "window", "bundled", "unlock_notify"]
+features = ["functions", "window", "bundled"]
 
 [dev-dependencies]
 pretty_assertions = "0.6"
--- a/components/push/Cargo.toml
+++ b/components/push/Cargo.toml
@@ -14,7 +14,7 @@ bincode = "1.2"
 lazy_static = "1.4"
 base64 = "0.12"
 log = "0.4"
-rusqlite = { version = "0.27.0", features = ["bundled", "unlock_notify"] }
+rusqlite = { version = "0.27.0", features = ["bundled"] }
 url = "2.2"
 viaduct = { path = "../viaduct" }
 sql-support = { path = "../support/sql" }
--- a/components/support/interrupt/Cargo.toml
+++ b/components/support/interrupt/Cargo.toml
@@ -11,4 +11,4 @@ parking_lot = "0.5"
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["functions", "limits", "bundled", "unlock_notify"]
+features = ["functions", "limits", "bundled"]
--- a/components/support/sql/Cargo.toml
+++ b/components/support/sql/Cargo.toml
@@ -18,7 +18,7 @@ tempfile = "3.1.0"
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["functions", "limits", "bundled", "unlock_notify"]
+features = ["functions", "limits", "bundled"]
 
 [dev-dependencies]
 env_logger = {version = "0.7", default-features = false}
--- a/components/tabs/Cargo.toml
+++ b/components/tabs/Cargo.toml
@@ -12,7 +12,7 @@ error-support = { path = "../support/error" }
 interrupt-support = { path = "../support/interrupt" }
 lazy_static = "1.4"
 log = "0.4"
-rusqlite = { version = "0.27.0", features = ["bundled", "unlock_notify"] }
+rusqlite = { version = "0.27.0", features = ["bundled"] }
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
--- a/components/webext-storage/Cargo.toml
+++ b/components/webext-storage/Cargo.toml
@@ -25,7 +25,7 @@ url = { version = "2.1", features = ["serde"] }
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["functions", "bundled", "serde_json", "unlock_notify", "column_decltype"]
+features = ["functions", "bundled", "serde_json", "column_decltype"]
 
 [dev-dependencies]
 env_logger = { version = "0.7", default-features = false }
--- a/examples/autofill-utils/Cargo.toml
+++ b/examples/autofill-utils/Cargo.toml
@@ -17,7 +17,7 @@ cli-support = { path = "../cli-support" }
 fxa-client = { path = "../../components/fxa-client" }
 interrupt-support = { path = "../../components/support/interrupt" } # XXX - should be removed once we do interrupts correctly!
 log = "0.4"
-rusqlite = {version = "0.27.0", features = ["functions", "bundled", "serde_json", "unlock_notify"]}
+rusqlite = {version = "0.27.0", features = ["functions", "bundled", "serde_json"]}
 serde_json = "1"
 sql-support = { path = "../../components/support/sql" }
 structopt = "0.3"


emilio:
Seems tests are unhappy. I'm not familiar with rustqlite or so, so happy to close this... But if somebody has any idea why my build fails locally otherwise I'm all ears :)

emilio:
Ok, so this is really messy. I did a bit of digging with @bendk and I have various potential fixes, but not sure what the best path forward. Ben suggested asking @mhammond for advice, so that's what I'm doing.

The issue is that, in the context of the cargo workspace, some crates enable the `sqlcipher` feature, which causes the `bundled` feature not to work:

> For backwards compatibility, feature 'sqlcipher' overrides features 'bundled' and 'bundled-windows'. If you want a bundled build of SQLCipher (available for the moment only on Unix), use feature 'bundled-sqlcipher' or 'bundled-sqlcipher-vendored-openssl' to also bundle OpenSSL crypto.

So we end up using the bundled sqlite or not depending on how we build, which is... bad?

Additionally, the `sqlcipher` feature causes us to link into `libsqlcipher.so` instead of `libsqlite3.so`. The issue is that, on my system, `libsqlcipher.so` doesn't have `sqlite3_unlock_notify`, but `libsqlite3.so` does:

```
$ nm -gD /usr/lib/libsqlcipher.so | rg sqlite3_unlock
$ nm -gD /usr/lib/libsqlite3.so | rg sqlite3_unlock                                                                                                                                                                                      
000000000009c140 T sqlite3_unlock_notify
```

First potential fix is to make sure that we consistently use the bundled libsqlite, with something like this:

```diff
diff --git a/components/logins/Cargo.toml b/components/logins/Cargo.toml
index 557ae875..7a5f0dce 100644
--- a/components/logins/Cargo.toml
+++ b/components/logins/Cargo.toml
@@ -30,7 +30,7 @@ uniffi_macros = "^0.17"
 
 [dependencies.rusqlite]
 version = "0.27.0"
-features = ["sqlcipher", "limits", "unlock_notify"]
+features = ["sqlcipher", "limits", "unlock_notify", "bundled", "bundled-sqlcipher"]
 
 [build-dependencies]
 uniffi_build = { version = "^0.17", features=["builtin-bindgen"] }
```

Second potential fix, more in the vein of the current `build.rs` hack, is to make sure we actually link in `libsqlite3.so`:

```diff
diff --git a/components/support/sql/build.rs b/components/support/sql/build.rs
index feeb5ca5..4611c62b 100644
--- a/components/support/sql/build.rs
+++ b/components/support/sql/build.rs
@@ -11,6 +11,7 @@
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rustc-link-lib=sqlite3");
 
     // If NSS_DIR isn't set, we don't really care, ignore the Err case.
     let _ = nss_build_common::link_nss();
```

Both fixes probably need tweaks for non-UNIX platforms... But anyways, the issue is that `cargo test` from the root of the workspace ends up using system libraries, and my system `libsqlcipher` doesn't have the `unlock_notify` functionality.

@mhammond, do you have any thoughts on what's preferred here?

emilio:
I reported https://bugs.archlinux.org/task/74622 for my distro to enable the relevant feature in the sqlcipher package as well...

emilio:
Ben additionally mentioned that we have an env variable to link `sqlcipher` to `libs/desktop/linux-x86-64/sqlcipher/lib`. However that doesn't seem to be working quite right because this is the output of the build script:

<details>

```
$ cat libsqlite3-sys-f430f57b96952fed/output                                                                                                                                                                                             
cargo:rerun-if-env-changed=SQLCIPHER_INCLUDE_DIR
cargo:rerun-if-env-changed=SQLCIPHER_LIB_DIR
cargo:rerun-if-env-changed=SQLCIPHER_STATIC
cargo:link-target=sqlcipher
cargo:rerun-if-env-changed=SQLCIPHER_NO_PKG_CONFIG
cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG
cargo:rerun-if-env-changed=PKG_CONFIG
cargo:rerun-if-env-changed=SQLCIPHER_STATIC
cargo:rerun-if-env-changed=SQLCIPHER_DYNAMIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
cargo:rerun-if-env-changed=PKG_CONFIG_PATH
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
cargo:rerun-if-env-changed=SYSROOT
cargo:rerun-if-env-changed=SQLCIPHER_STATIC
cargo:rerun-if-env-changed=SQLCIPHER_DYNAMIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
cargo:rustc-link-search=native=/usr/lib
cargo:rustc-link-lib=sqlcipher
cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG
cargo:rerun-if-env-changed=PKG_CONFIG
cargo:rerun-if-env-changed=SQLCIPHER_STATIC
cargo:rerun-if-env-changed=SQLCIPHER_DYNAMIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
cargo:rerun-if-env-changed=PKG_CONFIG_PATH
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
```

</details>

So looking into the sys crate build script:

https://github.com/rusqlite/rusqlite/blob/b411d27b54fd72320992bbb8bf9ff863b505fef8/libsqlite3-sys/build.rs#L415-L425

It uses that env var to do pkg-config shenanigans. There's no pkgconfig file in `libs/desktop/linux-x86-64/sqlcipher/lib`, so maybe it succeeds and returns my system's `libsqlcipher`?

emilio:
That seems to be what's going on, if I remove my local `sqlcipher` binary, then it works...

mhammond:
Sorry, I missed this conversation - sadly though I'm fairly clueless here. But I think we do not want to use a bundled sqlcipher, because we have used nss for the crypto rather than whatever it uses by default. It's for this reason we build it ourselves. I also don't expect *either* libsqlite.so or libsqlcipher.so to be used, and just `libs/desktop/linux-x86-64/sqlcipher/lib/libsqlcipher.a` statically linked (and, I assume, the "bundled" sqlite from rusqlite when used from other crates that don't use sqlcipher?) . It's for this reason we have so much pain on non-linux/mac environments - building this .a (and nss etc) is tricky - hence the only way I can use app-services on Windows is via WSL.

So I suspect the correct direction here would be to use the rusqlite `sqlcipher` feature - which is described as "looks for the SQLCipher library to link against instead of SQLite. This feature overrides bundled." - then all of our crates would be using sqlcipher (and IIRC, if you don't initialize the crypto in a db, then a sqlcipher db is just a plain old sqlite db) - which *sounds like* it would be a win, but sadly I'm really not sure about any of this.

I'm not sure exactly what context you are using this in, but sqlcipher is currently only used by our logins code to migrate old sqlcipher databases to a "plain" sqlite database and we manually encrypt certain columns manually. So we could probably add a feature-gate around that import capability, meaning consuming a-s with certain flags could avoid trying to use sqlcipher entirely (but would also fail to migrate old databases, so couldn't be enabled for our mobile clients). I'm not sure if this would help you or not.

mhammond:
> It uses that env var to do pkg-config shenanigans. There's no pkgconfig file in `libs/desktop/linux-x86-64/sqlcipher/lib`, so maybe it succeeds and returns my system's `libsqlcipher`?

Right, yeah, it sounds like we should remove that capability entirely?

mhammond:
We added `unlock_notify` in https://github.com/mozilla/application-services/issues/3287 - tl;dr of that is that it seems to be a feature not designed for direct consumption, but instead enables internal locking to perform better. IOW, removing `unlock_notify` will seem to work, but performance and actual functionality (ie, SQLITE_BUSY errors might happen more often) might take a hit.

Re pkg-config, rusqlite already has an issue for that at https://github.com/rusqlite/rusqlite/issues/510 - we could probably put a patch up to help with this (once we work out a good way to spell it), but I don't think we can do much about that in this specific issue - so I'm closing it.

PR #4935, opened 2022-05-02
Follow-up to #4933: Remove tempdir, use tempfile instead.

emilio:
tempdir is unmaintained and uses an outdated version of rand. We already
depend on tempfile.

  - Note:
- [x] **Tests**: No behavior change.
- [x] **Changelog**: Don't think it's worth an entry, doesn't change APIs or anything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -859,7 +859,7 @@ dependencies = [
  "interrupt-support",
  "log",
  "places",
- "rand 0.8.5",
+ "rand",
  "rusqlite",
  "sql-support",
  "tempfile",
@@ -887,7 +887,7 @@ dependencies = [
  "structopt",
  "sync-guid",
  "sync15",
- "tempdir",
+ "tempfile",
  "types",
  "url",
  "viaduct-reqwest",
@@ -920,7 +920,7 @@ dependencies = [
  "sql-support",
  "sync-guid",
  "sync15",
- "tempdir",
+ "tempfile",
  "viaduct-reqwest",
 ]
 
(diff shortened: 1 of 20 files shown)


emilio:
@tarikeshaq if you want to remove another related cleanup removing unused dependencies, this one should be straight-forward :-)

Btw, what is the cadence on these changes to be able to use them in Firefox? (The underlying reason I'm doing this is because I'm updating some dependencies in the CSS parser, and not updating the application-services dependencies would cause duplicate deps).

tarikeshaq (approved):
THANKS AGAIN @emilio, I'll 🚢 this!! 

Assuming you meant Firefox for android, There is no specific cadence to how long changes hit Firefox, the process is:
- We cut a release here on GitHub
- Android components would pick up the updated version (they have some automation that creates PRs for non-breaking changes) but mostly we create the PRs ourselves to bump the version
- Fenix would pick up the latest android components as they normally would (not too familiar there)

That's a long way of saying if we need a change in Firefox android/ios, I can totally cut a release for it asap (I can do so as early as Tomorrow) then it should hit nightly relatively quickly once the PRs land in the appropriate repos to bump the versions

tarikeshaq:
If you meant desktop (which we have a couple of our crates in), we usually vendor the crates using a commit hash of app-services, so we can do that anytime it is required 😄

mhammond:
> The underlying reason I'm doing this is because I'm updating some dependencies in the CSS parser, and not updating the application-services dependencies would cause duplicate deps).

I'm possibly missing something, but I'm not quite sure how this would solve that. app-services ends up in Android via a .so file which has all rust deps linked in. The CSS parser is presumably in a different .so - so IIUC, even if the deps were identical, they'd appear in both .so files?

If you mean desktop, then it *would* make sense, except for the fact that none of the components updated are in desktop. But FWIW, that process doesn't use a-s releases, but instead uses `./mach vendor` - so you could indeed take the new a-s whenever you wanted just by updating what's currently vendored. There are a few PRs though, so maybe this isn't the one that's actually going to touch the components which are in m-c, but one of the others will.

Ping us here or in slack if you'd like help to vendor a new version.

emilio:
> If you meant desktop (which we have a couple of our crates in), we usually vendor the crates using a commit hash of app-services, so we can do that anytime it is required

I meant desktop, so this answers my question, thanks!

mhammond:
(oops, missed Tarik's other comment, so apologies for saying the same thing in far more words :)

PR #4939, opened 2022-05-03
Add extra locking and caching to FeatureHolder

jhugman:
This PR fixes [EXP-2393][0]

This is part of the work to make FeatureHolder more reliable and functional on iOS.

[0]: https://mozilla-hub.atlassian.net/browse/EXP-2393

  - Note:

### DIFF
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
@@ -1,5 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 package org.mozilla.experiments.nimbus.internal
 
+import android.content.Context
 import org.mozilla.experiments.nimbus.FeaturesInterface
 import org.mozilla.experiments.nimbus.Variables
 import org.mozilla.experiments.nimbus.NullVariables
@@ -27,11 +32,15 @@ class FeatureHolder<T>(
      * in the feature manifest. This is done each call of the method, so the method should be called once, and the
      * result used for the configuration of the feature.
      *
+     * The unused parameter `Context` was relevant for `Text` and `Image` typed properties, but is no longer necessary.
+     * It will be removed in later releases.
+     *
      * @returns T
      * @throws NimbusFeatureException thrown before the Nimbus object has been constructed or `FxNimbus.initialize` has not been set.
      * This can be resolved by setting `FxNimbus.initialize`, and after that by passing in a `Context` object.
      */
-    fun value(): T =
+    @Suppress("UNUSED_PARAMETER")
+    fun value(context: Context? = null): T =
         lock.runBlock {
             if (cachedValue != null) {
                 cachedValue!!
@@ -51,12 +60,22 @@ class FeatureHolder<T>(
         getSdk()?.recordExposureEvent(featureId)
     }
 
+    /**
+     * This overwrites the cached value with the passed one.
+     *
+     * This is most likely useful during testing only.
+     */
     fun withCachedValue(value: T?) {
         lock.runBlock {
             cachedValue = value
         }
     }
 
+    /**
+     * This changes the mapping between a `Variables` and the feature configuration object.
+     *
+     * This is most likely useful during testing and other generated code.
+     */
     fun withInitializer(create: (Variables) -> T) {
         lock.runBlock {
             this.create = create
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
@@ -1,33 +1,47 @@
 package org.mozilla.experiments.nimbus.internal
 
-import android.content.Context
 import org.mozilla.experiments.nimbus.FeaturesInterface
 import org.mozilla.experiments.nimbus.Variables
 import org.mozilla.experiments.nimbus.NullVariables
+import java.util.concurrent.locks.ReentrantLock
 
+/**
+ * `FeatureHolder` is a class that unpacks a JSON object from the Nimbus SDK and transforms it into a useful
+ * type safe object, generated from a feature manifest (a `.fml.yaml` file).
+ *
+ * The two routinely useful methods are the `value()` and `recordExposure()` events.
+ *
+ * There are methods useful for testing, and more advanced uses: these all start with `with`.
+ */
 class FeatureHolder<T>(
     private val getSdk: () -> FeaturesInterface?,
     private val featureId: String,
-    private val create: (Variables) -> T
+    private var create: (Variables) -> T
 ) {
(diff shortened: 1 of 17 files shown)


electricRGB:
This look good from my side. Excited to get these additions! :D

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
```diff
@@ -36,6 +50,28 @@ class FeatureHolder<T>(
     fun recordExposure() {
         getSdk()?.recordExposureEvent(featureId)
     }
+
+    fun withCachedValue(value: T?) {
```
tarikeshaq, line 54: I'm assuming this is for testing purposes for the apps correct?

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
```diff
@@ -1,33 +1,47 @@
...
-     * This can be resolved by setting `FxNimbus.api`, and after that by passing in a `Context` object.
+     * @throws NimbusFeatureException thrown before the Nimbus object has been constructed or `FxNimbus.initialize` has not been set.
+     * This can be resolved by setting `FxNimbus.initialize`, and after that by passing in a `Context` object.
      */
-    @Suppress("UNUSED_PARAMETER")
-    fun value(context: Context? = null): T {
-        val variables = getSdk()?.getVariables(featureId, false) ?: NullVariables.instance
-        return create(variables)
-    }
+    fun value(): T =
```
tarikeshaq, line 34: This is a breaking change correct? do you know if Fenix already calls this without the argument? Otherwise we might want to mark this as a breaking change (or keep the parameter)

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
```diff
@@ -59,18 +59,14 @@ object {{ nimbus_object }} {
...
 
-    /**
-     * This should be populated at app launch.
-     *
-     * This property has to be set after the Nimbus SDK, which causes race-conditions. Please use
-     * `initialize()` much earlier in the application startup sequence.file
-     *
-     * This will be removed in future releases.
-     */
-    var api: FeaturesInterface? = null
```
tarikeshaq, line 73: This will also be a breaking change right? (assuming Fenix still uses the old way of initialization, which I think it does since they didn't take our a-s release just yet)

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
```diff
@@ -36,6 +36,14 @@ public class {{ nimbus_object }} {
     ///
     public let features = {{ nimbus_object }}Features()
 
+    /// Clear all cached values for all features.
+    /// This should be called after `applyPendingExperiments` is finished.
+    public func invalidateCachedValues() {
```
tarikeshaq, line 41: Just for my own clarity, I assume this is needed to be called after `applyPendingExperiments` is finished because it's possible a `value` was called before initialization, returning a default and we want to wipe that from the cache?

This makes me think a little about the "Nimbus is Idompotent" contract we have with mobile and whether or not this violates that

I may be missing a part of this, so maybe we can add some more context somewhere (not sure if this fits here in the doc comment, but may be the most convenient?

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
tarikeshaq: Also if we opt to remove this, should we also remove it from the swift manifest template?

tarikeshaq (changes requested):
The thread-safety/locking/caching of this makes sense 💯 🚀 

Had a question about invalidating the caches, and the `requesting changes` mostly for the breaking changes piece

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
jhugman: Now that the AS 93.0.1 has been backed out, I should put the unused parameter back.

I'd prefer to do this with the three PR doctrine we talked about.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
jhugman: Yeah, I'll put this back.

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
jhugman: Yes, this is correct. If we don't call it after `applyPendingExperiments` finishes then a call before the end of initialization (which gets less and less likely now) will poison the well for every subsequent call.

The idempotency illusion really is and always has been "idempotent between calls to `applyPendingExperiments`). This cache invalidation is needed to maintain that illusion.

components/nimbus/ios/Nimbus/FeatureHolder.swift
```diff
@@ -18,12 +30,42 @@ public class FeatureHolder<T> {
...
 
+    /// Get the JSON configuration from the Nimbus SDK and transform it into a configuration object as specified
+    /// in the feature manifest. This is done each call of the method, so the method should be called once, and the
+    /// result used for the configuration of the feature.
+    ///
+    /// Some care is taken to cache the value, this is for performance critical uses of the API.
+    /// It is possible to invalidate the cache with `FxNimbus.invalidateCachedValues()` or `with(cachedValue: nil)`.
     public func value() -> T {
+        lock.lock()
+        defer { self.lock.unlock() }
```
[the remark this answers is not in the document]
jhugman: `defer` is the way of doing the `try…finally` dance that the helper method in Kotlin does.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureHolder.kt
jhugman: Added docs.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
jhugman: Put this back, for removal after the apps upgrade.

jhugman:
I think this is timing out on CI, but ready for another review.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
tarikeshaq: did we put this back 😅 I'm not sure if it's a GitHub thing

tarikeshaq: oops ya github didn't diplay the diff properly for me 😬

tarikeshaq (approved):
Looks good! ~~I can re-run the iOS CI~~ beat me to it 🙃 , the taskcluster ones we can either for push or close/reopen the PR

PR #4940, opened 2022-05-04
Ensure that recording metadata with a view time of zero doesn't fail …

### LINKED ISSUE #4771: Unexpected error: SqlError(InvalidColumnType(0, "score", Null))
~https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/16154192/~

New Sentry instance link: https://sentry.io/organizations/mozilla/issues/3150733599/

via placesGetHistoryHighlights

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/SYNC-3000)

mhammond:
…when requesting highlights

Fixes #4771

Another option is to drop these observations, and I'm torn about the best thing to do here. Does a view time of zero represent an "accidental" visit? @jonalmeida, any thoughts?

### DIFF
--- a/components/places/src/storage/history_metadata.rs
+++ b/components/places/src/storage/history_metadata.rs
@@ -365,7 +365,7 @@ LEFT JOIN moz_places o ON o.id = m.referrer_place_id";
 // For example, given two entries we'll know that one is larger than another, but not by how much.
 const HIGHLIGHTS_QUERY: &str = "
 SELECT
-    ranked.score AS score, p.id AS place_id, p.url AS url, p.title AS title, p.preview_image_url AS preview_image_url
+    IFNULL(ranked.score, 0.0) AS score, p.id AS place_id, p.url AS url, p.title AS title, p.preview_image_url AS preview_image_url
 FROM moz_places p
 INNER JOIN
     (
@@ -1243,6 +1243,30 @@ mod tests {
             .len()
         );
 
+        // Database with "normal" history but no metadata observations is fine.
+        apply_observation(
+            &conn,
+            VisitObservation::new(
+                Url::parse("https://www.reddit.com/r/climbing").expect("Should parse URL"),
+            )
+            .with_visit_type(VisitTransition::Link)
+            .with_at(Timestamp::now()),
+        )
+        .expect("Should apply observation");
+        assert_eq!(
+            0,
+            get_highlights(
+                &conn,
+                HistoryHighlightWeights {
+                    view_time: 1.0,
+                    frequency: 1.0
+                },
+                10
+            )
+            .unwrap()
+            .len()
+        );
+
         // three observation to url1, each recording a second of view time.
         note_observation!(&conn,
             url "http://mozilla.com/1",
@@ -1330,6 +1354,32 @@ mod tests {
         let highlights4 = get_highlights(&conn, view_time_heavy_weights, 10).unwrap();
         assert_eq!(2, highlights4.len());
         assert_eq!("http://mozilla.com/2", highlights4[0].url);
+    }
+
+    #[test]
+    fn test_get_highlights_no_viewtime() {
+        let conn = PlacesDb::open_in_memory(ConnectionType::ReadWrite).expect("memory db");
+
+        // Make sure we work if the only observations for a URL have a view time of zero.
+        note_observation!(&conn,
+            url "http://mozilla.com/1",
+            view_time Some(0),
+            search_term None,
+            document_type Some(DocumentType::Regular),
+            referrer_url Some("https://news.website/tech"),
+            title None
+        );
+        let highlights = get_highlights(
+            &conn,
+            HistoryHighlightWeights {
+                view_time: 1.0,
+                frequency: 1.0,
+            },
+            2,
+        )
+        .unwrap();
+        assert_eq!(highlights.len(), 1);
+        assert_eq!(highlights[0].score, 0.0);
     }
 
     #[test]
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -27,3 +27,4 @@ Use the template below to make assigning a version number during the release cut
 ## Places
 ### What's Fixed:
 - Fixed a bug in Android where non-fatal errors were crashing. ([#4941](https://github.com/mozilla/application-services/pull/4941))
+- Fixed a bug where querying history metadata would return a sql error instead of the result ([4940](https://github.com/mozilla/application-services/pull/4940))


jonalmeida:
> Does a view time of zero represent an "accidental" visit?

It does read as an incomplete visit, i.e. open a tab and come back when it's done loading. If you return to the page later wouldn't that mean the visit time wouldn't be counted at all?

I hope @csadilek might have stronger opinions since I haven't worked on this feature at all on the Fenix/AC side.

jonalmeida (approved):
For this crash fix however, the change looks good to me.

csadilek:
> Another option is to drop these observations, and I'm torn about the best thing to do here. Does a view time of zero represent an "accidental" visit?

> I hope @csadilek might have stronger opinions since I haven't worked on this feature at all on the Fenix/AC side.

@mhammond @jonalmeida The `HistoryMetadataObservation` is actually split up in two [distinct types](https://github.com/mozilla-mobile/android-components/blob/main/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryMetadataStorage.kt#L21-L34) in A-C. A `DocumentTypeObservation` which never has a view time, and a `ViewTimeObservation` which always has a view time. The types are translated here: https://github.com/mozilla-mobile/android-components/blob/fa8e0a4cdfb464740a25b9feac260d820a579875/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Types.kt#L250

The `DocumentTypeObservation` is what we record right away when a page is opened and it will have the default (null) view time. It then may be followed `ViewTimeObservation`s. So, observations without view time are normal or as originally designed.

However, Fenix shouldn't ever crash when querying highlights, so that's new too. We should be catching all exceptions there and turning them into INFOs.

components/places/src/storage/history_metadata.rs
```diff
@@ -1332,6 +1356,32 @@ mod tests {
         assert_eq!("http://mozilla.com/2", highlights4[0].url);
     }
 
+    #[test]
+    fn test_get_highlights_no_viewtime() {
+        let conn = PlacesDb::open_in_memory(ConnectionType::ReadWrite).expect("memory db");
+
+        // three observation to url1, each recording a second of view time.
```
csadilek, line 1363: nit: leftover comment. ^^

tarikeshaq:
> However, Fenix shouldn't ever crash when querying highlights, so that's new too. We should be catching all exceptions there and turning them into INFOs.

It is very possible this was a side effect of https://github.com/mozilla/application-services/pull/4847/files#diff-f7ca17fcbe876a164e94b047ef39b32f62d4f4119bb67ec1b18d8e0c244913afR294 I'm guessing A-C catches all exceptions except for "UnexpectedPlacesException"

Hmmmm we can either catch the exception in A-C, or if we want to treat that exception as a fatal one, we should have a different exception type that captures any unexpected SQL errors. And have that get caught in A-C

jonalmeida:
RE: why are we crashing now.

In AC, the code to catch exceptions hasn't changed. We report PlacesExceptions to [the crash reporter and fail silently](https://github.com/mozilla-mobile/android-components/blob/0559e5a47e708b2f450d50621590c3e605486012/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt#L85-L94=). We can see this error (wrapped as an info) in the Sentry logs [on release](https://sentry.prod.mozaws.net/operations/firefox/issues/19455975/) that has been happening for 3 months.

Looking at the code that has changed recently, it seems that as part of [this telemetry patch](https://github.com/mozilla/application-services/commit/ab89c506addd278b4a6a0baa71f1a716fa2bdceb#diff-f7ca17fcbe876a164e94b047ef39b32f62d4f4119bb67ec1b18d8e0c244913afR294=), we are also now reporting `InternalPanic` as `UnexpectedPlacesException` which is a special exception we will always crash with. It doesn't seem right to crash the app for this particular case, so we should either change the AC code to not crash on `UnexpectedPlacesException` or we change this exception back to being an `InternalPanic`. I'm leaning towards the latter, so that we can keep using the unexpected exception for truly weird exceptions in the future when we do want a crash.

Side note: we are migrating to a new Sentry instance for release in an upcoming release so the sentry link above will stop reporting new errors over time.

mhammond:
> @mhammond @jonalmeida The `HistoryMetadataObservation` is actually split up in two [distinct types](https://github.com/mozilla-mobile/android-components/blob/main/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryMetadataStorage.kt#L21-L34) in A-C. A `DocumentTypeObservation` which never has a view time, and a `ViewTimeObservation` which always has a view time.

That makes some sense so far, but:

> And the null view time gets translated to 0 here: https://github.com/mozilla/application-services/blob/main/components/places/src/storage/history_metadata.rs#L720

Seems very odd. If I'm reading that correctly, a `ViewTimeObservation` with a time of zero can't be distinguished from a `DocumentTypeObservation` with `DocumentType::Regular` - and vice-versa.

Given the only other variant of `DocumentType` is `DocumentType::Media`, I guess the vast majority of `DocumentTypeObservation`s will be `Regular` - so that a-s code above will mostly be recording `{document_type: DocumentType::Regular, view_time: 0}` followed directly by `{document_type: DocumentType::Regular, view_time: >0}` - which seems odd.

So - the problem isn't as I described - it's not "a view time of zero was recorded", it's actually "only view times of zero are recorded for a given url" - ie, we expect many with a view time of zero, just not all - and looking at the a-s code, it will seems that we must be recording a `ViewTimeObservation` with a view time of zero. But, given what's above, we *can't* discard these, because by the time we get to the rust code, we can't distinguish it from a `DocumentTypeObservation`.

So I'll clean up my PR to reflect that, but I think the fix itself is correct - we never want that sql to return a score of null - which will avoid this specific sql error - but I agree we also want @tarikeshaq's PR so that other "unexpected" sql errors don't crash.

mhammond:
> Looking at the fix here though, the problem seems to be that you get a `null` _score_ from the query, which I don't yet understand. Why is the `0` view time causing a `null` score?

Took me longer to work this out than I hoped :)

`all_view_time` is zero [here](https://github.com/mozilla/application-services/blob/ee6674534c4659a47369a38a3c457614518d2c12/components/places/src/storage/history_metadata.rs#L376), so we divide by zero, which sqlite converts to null.

csadilek:
> all_view_time is zero [here](https://github.com/mozilla/application-services/blob/ee6674534c4659a47369a38a3c457614518d2c12/components/places/src/storage/history_metadata.rs#L376), so we divide by zero, which sqlite converts to null.

Ah, great catch! Thanks.

csadilek (approved):

mhammond:
All merged - but it's worth noting that afaict, this wasn't a regression in app-services, but presumably was a change in the behaviour of Fenix that caused it to be hit more - so reverting the app-services version seems unlikely to have the desired effect - so we should, ASAP, either cut a new major release, or a dot release with this fix. Please let us know if that sounds correct and what you'd like us to do.

jonalmeida:
Yes, we always seemed to have this error that was being caught as an INFO in Sentry (see links above). What changed is the exception type which turned it into a crash.

A dot release with this change and https://github.com/mozilla/application-services/pull/4941 would be good, so we can monitor the info rates decreasing in nightly as verification.

tarikeshaq:
@jonalmeida I'm cutting #4947 😄 I'll get the AC PR ready as well and we can probably merge it next week (it's Friday and all)

PR #4941, opened 2022-05-04
Brings back InternalError to catch non-fatal unexpected errors

tarikeshaq:
related to https://github.com/mozilla-mobile/android-components/issues/12099

This should fix the crashes in android in newer AS versions

The **real** fix here would be to expose a variant for that specific error and return it

We should in the future audit all the cases that get transformed into an `InternalError` to see if there are ones that really should be fatal - but this is the least disruptive way of making sure our a-s releases don't have any other hidden side-effects from this

(PS: I didn't name it `InternalPanic` like it used to be named because Rust panics were not involved at all in this error 😬)

### DIFF
--- a/components/places/src/error.rs
+++ b/components/places/src/error.rs
@@ -221,10 +221,8 @@ pub enum PlacesError {
     #[error("CannotUpdateRoot error: {0}")]
     CannotUpdateRoot(String),
 
-    // XX - We should kill this and replace
-    // it with other specific errors
-    // It is only alive because `UnexpectedPlacesException`
-    // is a fatal exception on android
+    // XX - Having `InternalError` is a smell and ideally it wouldn't exist
+    // it exists to catch non-fatal unexpected errors
     /// Thrown when we catch an unexpected error
     /// that shouldn't be fatal
     #[error("Unexpected error: {0}")]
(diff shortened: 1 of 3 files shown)


tarikeshaq:
Hmmm not sure why the circle CI jobs are not running... could be because I used my fork here? I'll open another PR to test it out

components/places/src/error.rs
```diff
@@ -220,6 +220,15 @@ pub enum PlacesError {
     /// - Attempting to delete any of the bookmark roots.
     #[error("CannotUpdateRoot error: {0}")]
     CannotUpdateRoot(String),
+
+    // XX - We should kill this and replace
+    // it with other specific errors
```
jonalmeida, line 225: I'm not sure if your comment implies the desire for a group of observation specific errors or a unexpected specific errors.

If you'd like to make this a specific error, that can work as part of this patch now as well. Using `UnexpectedPlacesException` is the only special exception as you can see here: https://github.com/mozilla-mobile/android-components/blob/0559e5a47e708b2f450d50621590c3e605486012/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt#L85-L94=

mhammond: I think Tarik is saying that having `InternalError` is a smell and ideally would not exist, because all the errors which fall into this category should have their own errors - but that's probably to intrusive to be actionable TBH (and chatting to Tarik, we are likely to go the other way - kill all those other internal errors, which will have the same effect)

Regardless, I agree this comment should be clarified.

(I don't think we've identified a new error specific to history metadata though)

mhammond (approved):

PR #4949, opened 2022-05-06
Bumps uniffi to 0.18 [ci full]

tarikeshaq:
Updates uniffi 0.18
changlog in https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md


tarikeshaq:
Still need to add a changelog and verify nothing is broken from the uniffi upgrade

tarikeshaq:
Aaah the dependency failure is a licence thing which is easy to patch, the iOS failure seems related to custom types being redeclared, I'll take a look - probably expected with the first breaking change in https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md

tools/dependency_summary.py
```diff
@@ -572,6 +572,51 @@
...
+            "fixup": "https://gitlab.com/CreepySkeleton/proc-macro-error/-/raw/master/LICENSE-APACHE"
+        }
+    },
+    # Based on https://github.com/Alexhuszagh/minimal-lexical/blob/main/LICENSE.md
+    # the libraries is licensed as dual licensed, however a portion of the
+    # source code is licensed under BSD-3-Clause. The portion is only
+    # used if the code uses the `compact` feature. However, tracking down
+    # where it's used in the tree, it's used by
+    # https://github.com/Geal/nom/blob/294ffb3d9e0ade2c3b7ddfff52484b6d643dcce1/Cargo.toml#L38
+    # which does **not** use the `compact` feature so we use to the Apatche license.
```
tarikeshaq, line 607: I'm gonna need some more license experienced eyes on this 😅

tarikeshaq:
Want to flag that this will be a breaking change, at least for iOS since we expose a `Timestamp` type in [the udl for places](https://github.com/mozilla/application-services/blob/42e0c7a0935fa4be69284c8949156bc05c37694e/components/places/src/places.udl#L7-L8) and [iOS have their own `Timestamp` type that conflicts with it](https://github.com/mozilla-mobile/firefox-ios/blob/c74a89eb5ad9886176588111def48f639ec88fed/Shared/TimeConstants.swift#L7)

Haven't checked in android, but whenever we decide we want to take this, it'll be breaking

components/places/src/places.udl
```diff
@@ -5,7 +5,7 @@
 [Custom]
 typedef string Url;
 [Custom]
-typedef i64 Timestamp;
+typedef i64 PlacesTimestamp;
```
tarikeshaq, line 8: This seemed like the least destructive way, this makes the change non-breaking

Normally I wouldn't mind a breaking change but the rename here is cheap and the breaking change in iOS is large (there is at least 30+ uses of `Timestamp` that would be ambigous) - if anyone feels strongly about it we can go the breaking route

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,14 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## General
+### What's new
+- Uniffi was upgraded to 0.18.0. For our consumers, this means there now exported types that used to be internal to `uniffi`. ([#4949](https://github.com/mozilla/application-services/pull/4949)).
+  - The types are:
```
tarikeshaq, line 25: Those are all of our `[Custom]` types so far (mostly in `places` and one in `nimbus`)

tools/dependency_summary.py
mhammond: typo: s/to the Apatche/the Apache/

I can't claim to have license experienced eyes, but I think your conclusion looks perfectly reasonable. It does sound like you are saying that license file isn't quite correct, so maybe opening an issue on their repo would be nice, but I don't think it's required!

mhammond (approved):
This looks fine to me, thanks for thinking it through!

Issue #4921, opened 2022-04-22
XCFramework not building Sqlite correctly for M1 simulators


skhamis:
Right now there is currently a problem where our XCFramework is not building arm64 simulator (M1) SQLite. However, everything else is being built correctly.

Summary of what I think is going on:
We use a slightly hacky approach to create xcframeworks -- instead of using the `xcodebuild` scenario we essentially custom-build the xcframework with this script https://github.com/mozilla/application-services/blob/main/megazords/ios-rust/build-xcframework.sh

Each target (x86_64 simulator, aarch64 ios hardware, and now aarch ios simulator) gets built and passed in the variables to build the binaries for those respective targets https://github.com/mozilla/application-services/blob/main/megazords/ios-rust/build-xcframework.sh#L88-L96

This has been working very well for us on firefox for iOS and M1's are able to build fx ios quite successfully. Unfortunately, there is a hidden bug where sqlite.o is not being built for aarch64-ios-sim correctly.

checking the binaries for the simulator portion of the build:
`/application-services/megazords/ios-rust/MozillaRustComponents.xcframework/ios-arm64_x86_64-simulator/MozillaRustComponents.framework`

running `otool -fahl MozillaRustComponents | grep -E 'cmd |platform |\.o'` and parsing some of the values shows:

sqlite3.o - Not being built correctly
```
../ios-arm64_x86_64-simulator/MozillaRustComponents.framework/MozillaRustComponents(sqlite3.o) (architecture x86_64):
      cmd LC_VERSION_MIN_IPHONEOS

```
```
../ios-arm64_x86_64-simulator/MozillaRustComponents.framework/MozillaRustComponents(sqlite3.o) (architecture arm64):
      cmd LC_VERSION_MIN_IPHONEOS
```
LC_VERSION_MIN_IPHONEOS is the older way to build the binaries for xcframeworks. A correct binary build looks like:
```
../ios-arm64_x86_64-simulator/MozillaRustComponents.framework/MozillaRustComponents(megazord_ios-fb888d3d9510ccdf.libsqlite3_sys-8d6c76a729212ba1.libsqlite3_sys.1af7b0bf-cgu.0.rcgu.o.rcgu.o)
 (architecture arm64):
      cmd LC_BUILD_VERSION
 platform 7
```
[...]

PR #4953, opened 2022-05-09
Ships a seperate megazord for focus-ios

### LINKED ISSUE #4821: Create a seperate "focus-ios" megazord for focus
After the [91.0.1](https://github.com/mozilla/rust-components-swift/releases/tag/91.0.1) release of `rust-components-swift` all the components are released as one target. Meaning that Focus iOS now will start getting the swift code for components it's not interested in (i.e fxa, sync, etc).  For now that shouldn't be a problem, the code is not consumed by Focus anyways - and they already load the full Rust binary anyways.

However, a clear optimization for this is to create a separate megazord for focus, one that only includes the libraries they need, then import that in `rust-components-swift` as a different target than the one we ship for Firefox iOS.

This would include:
- creating a new crate inside `megazords`, and have that crate similarly structured to [`megazords/ios-rust`](https://github.com/mozilla/application-services/tree/main/megazords/ios-rust), except in it's `Cargo.toml` it only includes the components focus uses
- Generalize the [`build-xcframework`](https://github.com/mozilla/application-services/blob/main/megazords/ios-rust/build-xcframework.sh) script to build the two separate frameworks (probably move it outside the `ios-rust` and rename `ios-rust` to `firefox-ios`
- [Change our CircleCI release job](https://github.com/mozilla/application-services/blob/main/.circleci/config.yml#L447-L457) to release both frameworks
- In `rust-components-swift`:
   - Import the new framework as a **new** binary target
   - update our `generate` script to pull in all the relevant code into a different directory than the one firefox ios consumes
   - Create a new swift target, that consumes that binary target
   - Create a new library representing that target that we ship
   
 Proably two mid-sized tickets - one for appservices and one for rust-components-swift. But let's hold off of this until the dust settles and both consumers are consuming the library with no errors

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3045)

tarikeshaq:
fixes #4821 

## Why ship a seperate megazord
There are two main reasons to ship a separate Megazord for focus:
- Builds for recent rust-components-swift fail on M1s due to an sqlite linker error, ref #4921 
- We currently ship the binaries for **all** of our components to focus, bloating the binary size. I checked manually, the difference in the  size of the compiled binary goes from ~130MB to ~50MB. The direct impact on the app will be different based on how they consume the binary

## What is in this PR
In the PR we create a new megazord, under `megazords/ios-rust/focus` that only imports the components that focus actually needs. We then build and release an xcframework of that megazord using circleci. A few notes:
- We modify the `build-xcframework` script to allow a `--focus` flag, that builds the focus xcframework instead
- Although the name of the built directory and file are `FocusRustComponents`, we keep the name of the actual module `MozillaRustComponents` because uniffi depends on it. Uniffi depends on it [because we can only define one module name in the `uniffi.toml`](https://github.com/mozilla/application-services/blob/main/components/nimbus/uniffi.toml#L15)
- Modified the license summary generation script to work with the new megazord. The **majority** of the line diff in the PR is from the new license file
- CI will only run the focus build and release on github releases. Temporarily I have it running on every push just so that I can see it work, I'll change that once this PR is about to land

Keeping this as a WIP draft until we verify that focus can build properly using it. I have https://github.com/mozilla-mobile/focus-ios/pull/3232 updated to use a focus-only megazord. It builds OK for me, but we need to verify it builds OK on M1s

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -499,18 +499,13 @@ workflows:
             tags:
               only: /^v.*/
       - Focus build XCFramework:
-          # filters:
-          #   branches:
-          #     ignore: /.*/
-          #   tags:
-          #     only: /^v.*/ # We only build Focus on release since Focus
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/ # We only build Focus on release since Focus
               # only uses a subset of the components the full iOS build
               # covers.
-          # TODO: Remove this before landing, temporary to test that CI
-          # can build the artifact properly
-          filters: # required since `XCFramework release` has tag filters AND requires this job
-            tags:
-              only: /.*/
       - Focus XCFramework release:
           requires:
             - Focus build XCFramework
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,6 +334,58 @@ jobs:
           root: .
           paths:
             - megazords/ios-rust/MozillaRustComponents.xcframework.zip
+  Focus build XCFramework:
+    executor: macos
+    steps:
+      - full-checkout
+      - restore-sccache-cache
+      - install-rust
+      - setup-rust-target-version
+      - setup-sccache
+      - setup-ios-environment
+      - run:
+          name: Build XCFramework archive
+          command: |
+            bash megazords/ios-rust/build-xcframework.sh --build-profile release --focus
+      - save-sccache-cache:
+          path: "~/Library/Caches/Mozilla.sccache"
+      - store_artifacts:
+          name: Store XCFramework bundle in workspace
+          path: megazords/ios-rust/focus/FocusRustComponents.xcframework.zip
+          destination: dist/FocusRustComponents.xcframework.zip
+      - run:
+          name: "XCFramework bundle checksum"
+          command: |
+            shasum -a 256 ./megazords/ios-rust/focus/FocusRustComponents.xcframework.zip
+            echo "Use the above checksum to depend on FocusRustComponents.xcframework.zip as a Swift Package binary target"
+      - persist_to_workspace:
+          root: .
+          paths:
+            - megazords/ios-rust/focus/FocusRustComponents.xcframework.zip
+
+  XCFramework release:
+    executor: macos
+    steps:
+      - full-checkout
+      - attach_workspace:
+          at: .
+      - install-ghr
+      - run:
+          name: Release XCFramework archive on GitHub
+          command: |
+            ghr -replace "${CIRCLE_TAG}" megazords/ios-rust/MozillaRustComponents.xcframework.zip
+
+  Focus XCFramework release:
+    executor: macos
+    steps:
+      - full-checkout
+      - attach_workspace:
+          at: .
+      - install-ghr
+      - run:
+          name: Release XCFramework archive on GitHub
+          command: |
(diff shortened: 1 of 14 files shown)


tarikeshaq:
Needs a changelog, but opening for review since we verified it builds OK on M1s!

.gitignore
```diff
@@ -28,6 +28,8 @@ components/**/ios/Generated
 
 # XCFramework artifact
 megazords/ios-rust/MozillaRustComponents.xcframework*
+megazords/ios-rust/focus/FocusRustComponents.xcframework*
```
skhamis, line 31: hmmm, I wonder if it'll be worth it eventually to move the firefox into `fx-ios` or something, but more of a musing and definitely not in scope for this PR

.circleci/config.yml
```diff
@@ -455,6 +498,27 @@ workflows:
...
               only: /^v.*/
+      - Focus build XCFramework:
+          # filters:
+          #   branches:
+          #     ignore: /.*/
+          #   tags:
+          #     only: /^v.*/ # We only build Focus on release since Focus
+              # only uses a subset of the components the full iOS build
+              # covers.
+          # TODO: Remove this before landing, temporary to test that CI
```
skhamis, line 509: Nit: Is this comment still accurate? bump here just to make sure we don't forget

megazords/ios-rust/build-xcframework.sh
```diff
@@ -118,30 +134,40 @@ rm -rf "$XCFRAMEWORK_ROOT"
...
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/logins/src/logins.udl" -l swift -o "$COMMON/Headers"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/autofill/src/autofill.udl" -l swift -o "$COMMON/Headers"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/push/src/push.udl" -l swift -o "$COMMON/Headers"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/tabs/src/tabs.udl" -l swift -o "$COMMON/Headers"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/places/src/places.udl" -l swift -o "$COMMON/Headers"
+
+# We now only move/generate the rest of the headers if we are generating a full
+# iOS megazord
+if [ -z $IS_FOCUS ]; then
+  cp "$REPO_ROOT/components/external/glean/glean-core/ffi/glean.h" "$COMMON/Headers"
```
skhamis, line 153: Could've sworn Focus uses glean, but I think the swift consumers now get glean from it's own swift packages....but I think nimbus still uses the internal glean? We should probably follow up on this.

skhamis (approved):
Awesome job on this! This is about what I'd expect if we were going down this route. It's pretty much fx-ios route but for focus. I don't think this would be too hard to revert if we end up fixing the lower level sqlite issue and if we can move back to component-based loading so i'm comfortable with moving forward with this.

LGTM! 🚀

megazords/ios-rust/build-xcframework.sh
tarikeshaq: Yeah they use glean through its own package and the internal usages will also end up using that package - the other concern would be local development, but the focus megazord shouldn't touch that, since that still uses the full xcframework

I was a little concerned something would be off, so I ran focus locally with this xcframework and verified that glean pings through nimbus get processed properly

PR #4954, opened 2022-05-10
Don't warn if we fail to open a database that doesn't exist and we di…

mhammond:
…dn't try to create

Playing with tabs I saw a scary warning:

```
[2022-05-10T00:00:18Z DEBUG sql_support::open_database] tabs storage db: opening database
[2022-05-10T00:00:18Z WARN  sql_support::open_database] tabs storage db: database operation failed: Error executing SQL: unable to open database file: ./tab-sync.db
```

digging deeper, the tabs component sometimes asks to open the database only if it exists - which is what was happening here. This patch means we now log:

```
[2022-05-10T01:03:48Z DEBUG sql_support::open_database] tabs storage db: opening database
[2022-05-10T01:03:48Z INFO  sql_support::open_database] tabs storage db: database doesn't exist, but we weren't requested to create it
```

### DIFF
--- a/components/support/sql/src/open_database.rs
+++ b/components/support/sql/src/open_database.rs
@@ -171,6 +171,15 @@ fn try_handle_db_failure<CI: ConnectionInitializer, P: AsRef<Path>>(
     _connection_initializer: &CI,
     err: Error,
 ) -> Result<()> {
+    if !open_flags.contains(OpenFlags::SQLITE_OPEN_CREATE)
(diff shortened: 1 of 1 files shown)


tarikeshaq (approved):
LGTM - also I think the right group to request reviews from is `mozilla/a-s-review` 😬 I didn't get a notification for this PR

mhammond:
> also I think the right group to request reviews from is `mozilla/a-s-review`

oops - I knew I was going to make that mistake one day :)

PR #4957, opened 2022-05-11
Cut release v93.2.0

tarikeshaq:
Cuts 93.2.0

Only for the focus-ios megazord - so that we can use it in `rust-components-swift` and update focus-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# v93.2.0 (_2022-05-11_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v93.1.0...v93.2.0)
+
+## General
+### What's new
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v93.2.0 (_2022-05-11_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v93.1.0...v93.2.0)
+
+## General
+### What's new
+- Application services now releases an xcframework with only the components needed by focus-ios (namely Nimbus, Viaduct and Rustlog). ([#4953](https://github.com/mozilla/application-services/pull/4953))
```
skhamis, line 7: This might be a super nit: but we should be explicit that we're releasing an `additional` or `separate` xcframework from the existing firefox ios one. Just incase there's any confusion

skhamis (approved):

PR #4959, opened 2022-05-16
Update ios docs

### LINKED ISSUE #4825: Document how to locally test against Firefox iOS

tarikeshaq:
This updates a number of docs that we have related to using our components in iOS.

- I'm trying to keep the docs in the `docs` directory as the source of truth, so I referenced them in the `README` in `ios-rust`
- Wrote a `howto` for testing locally published components on `firefox-ios`, this fixes #4825
- Updates the docs now that we don't have a library per component in `rust-components-swift`. If we ever revert back, should be easy to re-update the docs 😄
- Updates the `focus` docs now that they get their own xcframework/megazord
- Made sure to reference the new scripts in the `firefox-ios` docs


docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+1. Run the `build-xcframework.sh` script: 
+```bash
+$ ./build-xcframework.sh
+```
+This will produce a file name `MozillaRustComponents.xcframework.zip` that contains the following, built for all our target iOS platforms.
+- The compiled Rust code for all the crates listed in `Cargo.toml` as a static library
+- The C header files and [Swift module maps](https://clang.llvm.org/docs/Modules.html) for the components
+
+## Include the xcframework in a local checkout of `rust-components-swift`
+After you generated the `MozillaRustComponents.xcframework.zip` in the previous step, do the following to include it in a local checkout of `rust-components-swift`:
```
mhammond, line 67: I followed the "automated" instructions above, but that didn't tell me about this .zip file - probably worth noting there where it lives?

docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+1. Unzip the `MozillaRustComponents.xcframework.zip` into the `rust-components-swift` repository: (Assuming you are in the root of the `rust-components-swift` directory and `application-services` is a neighbor directory)
```
mhammond, line 68: The automated process knows the location of `rust-components-swift` - could this unzipping be automated?

docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+    ```sh
+     unzip  ../application-services/megazords/ios-rust/MozillaRustComponents.xcframework.zip -d .
+    ```
```
mhammond, line 71: This prompted me about overwriting files - I assume that's expected, so maybe add `-o` to the command-line?

docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+1. Change the `Package.swift`'s reference to the xcframework to point to unzipped folder. You can do this by uncommenting the following line:
```
mhammond, line 72: This confuses me - "point to unzipped folder" means "point to the root of the `rust-components-swift`"?

docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+    ```swift
+        path: "./MozillaRustComponents.xcframework"
+    ```
+    and commenting out the following lines:
+    ```swift
+        url: url,
+        checksum: checksum,
+    ```
```
mhammond, line 80: I'm seeing this, which (a) seems wrong compared to the instructions, but (b) seems to be where we want to end up? 
image

docs/howtos/locally-published-spm-in-firefox.md
```diff
@@ -0,0 +1,105 @@
...
+```
+Once that is done, your local checkout will now have a git tag `X.Y.Z` that can be pointed to by Xcode
+
+## Include the local checkout of `rust-components-swift` in `firefox-ios`
+This is the final step to include your local changes into `firefox-ios`. Do the following steps:
+1. Open `Client.xcodeproj` in Xcode
+1. Navigate to the Swift Packages in Xcode:
+![Screenshot of where to find the setting for Client](./img/xcode-client-package-settings.png)
+1. Remove the dependency on `rust-components-swift` as listed on Xcode, you can click the dependency then click the `-`
+1. Add a new swift package by clicking the `+`:
```
mhammond, line 95: I was surprised to find my xcode already had a file:// reference to my local checkout, but then noticed your screenshot does too - but it does point to main. Is there a reason we can't just change that to the new version?

Also, in "Add to Project" I don't have a "Blockzilla", just "Client"

This all seemed to work - but the build failed with this - so I guess I didn't clobber first?

image

mhammond (commented):
Thanks Tarik, I learned some things and nearly got a build - but I'm out of time to play any more today - I hope this is helpful

docs/howtos/locally-published-spm-in-firefox.md
tarikeshaq: > I was surprised to find my xcode already had a file:// reference to my local checkout, but then noticed your screenshot does too

I'm guessing it's because you used the script before 😬 The script does everything the manual steps do

> This all seemed to work - but the build failed with this - so I guess I didn't clobber first?

Yeah, probably re-running the `bootstrap.sh` they have on firefox-ios would fix that

docs/howtos/locally-published-spm-in-firefox.md
tarikeshaq: Yeah, what you see in the result of the script, which already does the work of replacing the lines needed

The manual steps assume you're starting from scratch without running the script

docs/howtos/locally-published-spm-in-firefox.md
tarikeshaq: Welp the wording doesn't make too much sense here, I'll update it - it should point to the unzipped `MozillaRustComponents.xcframework`

docs/howtos/locally-published-spm-in-firefox.md
tarikeshaq: Good catch - I'll update it! It's expected if you've gone through the process before (or have run the script which does the whole process)

docs/howtos/locally-published-spm-in-firefox.md
tarikeshaq: Hmmm The automated process actually does **everything** the manual process does 😅 I'll clear up the wording and put a notice at the top, but realistically you shouldn't run the manual process unless for some reason the automated process was broken...

I think I need to make it clearer that this is a "pick one" rather than "do both" process 😬

tarikeshaq:
Thanks a ton for going through this Mark!! this gave me a few things to update, most notably to make it clear to readers that what I call the "manual" process does exactly what the "automated" process does, and is only there in case something breaks (or if a user is trying to change the process itself)

docs/howtos/locally-published-spm-in-firefox.md
mhammond: > (or have run the script which does the whole process)

Right, yeah, that needs to be clearer - I paid special attention to the heading levels, so that once we were back to a top-level heading, we were back to "must be done in both cases"

lougeniaC64 (approved):
LGTM! 🚀

PR #4960, opened 2022-05-16
Adds firefox iOS smoke test

### LINKED ISSUE #4826: Create a new smoke test for Firefox iOS

tarikeshaq:
Fixes #4826 

Recreates the smoke test for firefox iOS using the new local development scripts

Important to note that I haven't added this to CI yet, I'm not sure when to run it yet... seems like an overkill to run on every PR

For Fenix, we don't even run the smoke tests on CI at all, and they are there for convenience more than anything


skhamis (approved):
Awesome job on this! I agree we shouldn't run this on every PR. I'm okay with keeping same consistency as Fenix, we also have a step when cutting a release to smoke test anyways. LGTM!

PR #4962, opened 2022-05-18
Depend on parking_lot 0.11 or 0.12 so app-services and mozilla-c…

mhammond:
…entral can use different versions

Glandium is pushing back on using parking_lot 0.12.0 in https://phabricator.services.mozilla.com/D145300 because it adds a dependency on `windows-sys`. In #build, he said

> windows-sys is very large, and it is yet to be seen whether it's going to go through more backwards incompatible iterations (which would make us require multiple versions in the end, which is even worse), or a lot of revisions (which mercurial won't like very much because some of the largest files are binary and, well, mercurial doesn't handle those very efficiently). Do you have use cases that wouldn't be satisfied by winapi? (parking_lot doesn't really count, we can stick to 0.11 which doesn't have a dependency on windows-sys, or, worse case, we can have a fake windows-sys that re-exports enough of winapi to satisfy parking_lot)

Sadly though, specifying exactly `0.11` means app-services ends up with 2 versions, because of a transitive dependency on `jsonschema` via nimbus.

Specifying `parking_lot = "0"` doesn't seem ideal, but we already do it for serde etc. 

cc @emilio and @skhamis as this seems like it might unblock vendoring a-s into m-c.

Any good reason (apart from [the obvious](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277)) not to do this?


emilio:
I'd probably go with `>=0.11,<=0.12` to allow both for now but not _any_, but I don't have a strong opinion if you prefer to do this.

mhammond:
> I'd probably go with `>=0.11,<=0.12`

Yeah, fair enough, and that is probably better - but what about just `>=0.11`?  I don't see a need to restrict the upper bound

emilio:
> Yeah, fair enough, and that is probably better - but what about just `>=0.11`? I don't see a need to restrict the upper bound

How not? Otherwise your build could break if parking-lot makes breaking changes (which they can totally do from 0.12 to 0.13 for example). I also _think_ you need the `<=something` to be able to pick 0.12 but I don't recall the rules from memory so icbw.

mhammond:
Yeah, I think I misunderstood the rules - it's late for me, so I'll double-check tomorrow, but maybe `~0.11` is what I assumed `>= 0.11` meant.

mhammond:
> I'll double-check tomorrow

I also might be misunderstanding the implication of a leading `0` - definitely a tomorrow thing :)

tarikeshaq:
> Sadly though, specifying exactly 0.11 means app-services ends up with 2 versions, because of a transitive dependency on jsonschema via nimbus.

One thing to note is that the `jsonschema` is a dev dependency so we wouldn't be creating duplicates for our consumers if we use 0.11 (and I think we can live with the duplicates in dev environments?)

bendk (approved):
That reasoning makes a lot of sense to me.

PR #4974, opened 2022-05-27
Adjust allowed version of url to avoid issues vendoring into m-c

mhammond:


mhammond:
FTR, on top of bug 1770894 there are no new dupes

mhammond:
This appears to be all we need on top of bug 1770894, so requesting review now, but I'll probably hold off merging until that lands and I'm in a position to put up the m-c patch.

components/support/sync15-traits/Cargo.toml
```diff
@@ -14,5 +14,5 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 log = "0.4"
 ffi-support = "0.4"
-url = "2.2"
+url = "2.1" # pinned to 2.1 for m-c, and 2.2 is incompatible, so don't allow that.
```
glandium, line 17: Technically, this allows 2.2. And it's fine.

mhammond: I'm not sure what you mean here - maybe the comment is misleading?

The problem IIRC is that from 2.1 -> 2.2 they added `Into<String>` (or something like that, I last struck it mid last year) but also dropped the "old" way of getting a string, which I think was a mistake on their part as it broke semver. I can dig up the details if you are interested.

mhammond: doh, sorry, I do see what you mean, yeah

mhammond: So I don't think there's anything I can really do to stop these crates from using features in 2.2, so I guess we'll just discover if they do as we vendor new versions.

glandium: You can do >=2.1,<2.2

mhammond: But then we will end up with a dupe of that crate in our mobile builds (some crates that aren't in m-c *do* need 2.2)

tarikeshaq (approved):

PR #4979, opened 2022-06-01
New error reporting system [ac: bendk/app-services-error-reporting] [fenix bendk/app-services-error-reporting]

bendk:
This PR contains code that I hope will be the basis of a new error reporting system.  Eventually I hope to replace the current system, which is based on `log::error!()` with this system.  However, before crossing that bridge let's try to test out the new code with a specific issue: #4856.

The first commit contains the general system.  The second commit uses it to generate some reports with breadcrumbs that I hope can help fix that one.  At this point it's all additive, I didn't remove any of the current reporting.

  - Note:

### DIFF
--- a/components/places/src/ffi.rs
+++ b/components/places/src/ffi.rs
@@ -23,6 +23,7 @@ use crate::ConnectionType;
 use crate::VisitObservation;
 use crate::VisitTransition;
 use crate::{PlacesApi, PlacesDb};
+use error_support::report_error;
 use interrupt_support::{register_interrupt, SqlInterruptHandle};
 use parking_lot::Mutex;
 use std::sync::Arc;
@@ -398,7 +399,18 @@ impl PlacesConnection {
     // sync metadata to only sync after most recent visit to prevent
     // further syncing of older data
     fn delete_everything_history(&self) -> Result<()> {
-        self.with_conn(history::delete_everything)
+        // Do some extra work to track down #4856
+        let conn = self.db.lock();
+        let result = history::delete_everything(&conn);
+        if let Err(e) = &result {
+            if matches!(
+                e.kind(),
+                crate::error::ErrorKind::SqlError(rusqlite::Error::QueryReturnedNoRows)
+            ) {
+                report_error!("SqlErrorQueryReturnedNoRows", "{}", e);
+            }
+        }
+        Ok(result?)
     }
 
     // XXX - This just calls wipe_local under the hood...
--- a/components/places/src/frecency.rs
+++ b/components/places/src/frecency.rs
@@ -4,6 +4,7 @@
 
 use crate::error::*;
 use crate::types::VisitTransition;
+use error_support::trace_error;
 use rusqlite::Connection;
 use types::Timestamp;
 
@@ -150,7 +151,8 @@ impl<'db, 's> FrecencyComputation<'db, 's> {
         ",
         )?;
         let mut rows = stmt.query(&[(":page_id", &page_id)])?;
-        let row = rows.next()?.ok_or(rusqlite::Error::QueryReturnedNoRows)?;
+        // trace_error to track down #4856
+        let row = trace_error!(rows.next()?.ok_or(rusqlite::Error::QueryReturnedNoRows))?;
         let typed: i32 = row.get("typed")?;
         let visit_count: i32 = row.get("visit_count")?;
         let foreign_count: i32 = row.get("foreign_count")?;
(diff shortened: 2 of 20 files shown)


components/places/src/ffi.rs
```diff
@@ -398,7 +399,17 @@ impl PlacesConnection {
     // sync metadata to only sync after most recent visit to prevent
     // further syncing of older data
     fn delete_everything_history(&self) -> Result<()> {
-        self.with_conn(history::delete_everything)
+        let conn = self.db.lock();
+        let result = history::delete_everything(&conn);
+        if let Err(e) = &result {
+            if matches!(
```
mhammond, line 405: Maybe add a comment here about why that specific error gets that specific treatment - a link to the issue maybe?

components/places/src/frecency.rs
```diff
@@ -150,7 +151,7 @@ impl<'db, 's> FrecencyComputation<'db, 's> {
         ",
         )?;
         let mut rows = stmt.query(&[(":page_id", &page_id)])?;
-        let row = rows.next()?.ok_or(rusqlite::Error::QueryReturnedNoRows)?;
+        let row = trace_error!(rows.next()?.ok_or(rusqlite::Error::QueryReturnedNoRows))?;
```
mhammond, line 154: and ditto here with the same link? I'm really just thinking about how someone in the future can undo the various parts used to track the specific problem.

(Not too sure about the breadcrumbs below - I guess probably not? It could be argued they are useful in the general case, and it seems reasonable that we could just remove breadcrumbs later if we feel they haven't added value recently?)

components/support/error/src/macros.rs
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// Tell the application to report an error
+///
+/// This will cause the error to be reporting to a Sentry-like system.  We will send extra context
```
mhammond, line 7: Really nitpicking here, but the comment implies this code will do the reporting. Maybe something like "If configured by the consuming application, this will be reported to that application, which may submit etc"?

components/support/error/src/macros.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// Log a breadcrumb if we see an `Result::Err` value
+///
+/// Use this macro to wrap a function call that returns a `Result<>`.  If that call returns an
+/// error, then we will log a breadcrumb for it.  This can be used to track down the codepath where
+/// an error happened.
+#[macro_export]
+macro_rules! trace_error {
+    ($result:expr) => {{
```
mhammond, line 35: ISTM that in some complex scenarios, we may not be 100% sure which `trace_error!` call caused this? ie, is it worth adding additional context so we know for sure which function caused it?

components/support/error/Cargo.toml
```diff
@@ -6,7 +6,15 @@ edition = "2021"
 license = "MPL-2.0"
 
 [dependencies]
+log = "0.4"
+parking_lot = "0.12"
```
mhammond, line 10: Do you think it's worth having this be `">=0.11,<=0.12"` to prepare for a future where this ends up in m-c? (OTOH, I don't really case as we don't have concrete plans there yet, and that will likely cause other yaks)

components/support/rc_crypto/nss/src/pk11/types.rs
```diff
@@ -146,7 +146,7 @@ impl PrivateKey {
         let mut id_attr: &mut nss_sys::CK_ATTRIBUTE = template
             .iter_mut()
             .find(|&&mut attr| {
-                attr.type_ == nss_sys::CKA_ID.into()
+                attr.type_ == (nss_sys::CKA_ID as nss_sys::CK_ATTRIBUTE_TYPE)
```
mhammond, line 149: Is this necessary for this patch?

mhammond (approved):
This is great and I'm looking forward to seeing what this tells us! Happy to discuss any of the changes I requested, but I don't see a good reason why I need to re-review.

components/support/error/src/macros.rs
bendk: `breadcrumb!` includes the module, line, and column for this very reason.

components/places/src/frecency.rs
bendk: That makes a lot of sense.  I added comments to the frecency breadcrumbs too, I kind of doubt they are going to be useful in general.

PR #4987, opened 2022-06-06
Put the new error-reporting capability behind a feature

mhammond:
I noticed that the new error reporting capability meant that webext-storage ended up taking a dependency on uniffi etc, which is doesn't need in practice and makes vendoring into m-c a bit painful at this instant. Even though m-c is going to end up with uniffi, there's no need for webext-storage to take the small hit, so I put the error reporting behind a feature.

This seems to work for me locally, and if I missed something I hope CI will point it out! Ben, wdyt?

### DIFF
--- a/components/places/Cargo.toml
+++ b/components/places/Cargo.toml
@@ -27,7 +27,7 @@ idna = "0.2"
 memchr = "2.3"
 dogear = "0.4"
 interrupt-support = { path = "../support/interrupt" }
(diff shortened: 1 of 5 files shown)


mhammond:
the cargo docs on features are fairly inscrutable - I *think* this is correct and does work as expected on m-c - I just hope it works as expected here!

bendk (approved):
Make sense to me

PR #4989, opened 2022-06-06
Adds function definitions for history functions to swift

tarikeshaq:
To understand our Places history APIs better, I decided a good first step is to walk through them and add their Swift equivalents with docs, and an excellent second step is to write some Swift tests for those functions to make sure they do what we expect them to do.

So far this was purely mechanical, and most my time was spent trying to write the docstrings

TODOs before this is ready:
- [x] Add tests for some of the history functions, may not need to add tests for all (we got lots of coverage in Rust 😄)

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -21,4 +21,12 @@ Use the template below to make assigning a version number during the release cut
 ## Nimbus FML ⛅️🔬🔭
 ### What's Changed
   - Validate the configuration passed from a top-level FML file to imported files. ([#5055](https://github.com/mozilla/application-services/pull/5055))
+
+## Places
+### What's new
+  - We now expose all of the Places history APIs to Swift consumers. ([#4989](https://github.com/mozilla/application-services/pull/4989))
+
+## Sync15
+### What's changed
   - `CLIENTS_TTL` has been updated to be 180 days instead of 21 ([#5054](https://github.com/mozilla/application-services/pull/5054))
+
--- a/components/places/ios/Places/Places.swift
+++ b/components/places/ios/Places/Places.swift
@@ -129,6 +129,33 @@ public class PlacesAPI {
     }
 
     /**
+     * Sync the History collection.
+     *
+     * - Returns: A JSON string representing a telemetry ping for this sync. The
+     *            string contains the ping payload, and should be sent to the
+     *            telemetry submission endpoint.
+     *
+     * - Throws:
+     *     - `PlacesError.databaseInterrupted`: If a call is made to `interrupt()` on this
+     *                                          object from another thread.
+     *     - `PlacesError.unexpected`: When an error that has not specifically been exposed
+     *                                 to Swift is encountered (for example IO errors from
+     *                                 the database code, etc).
+     *     - `PlacesError.panic`: If the rust code panics while completing this
+     *                            operation. (If this occurs, please let us know).
+     */
+    open func syncHistory(unlockInfo: SyncUnlockInfo) throws -> String {
+        return try queue.sync {
+            return try self.api.historySync(
+                keyId: unlockInfo.kid,
+                accessToken: unlockInfo.fxaAccessToken,
+                syncKey: unlockInfo.syncKey,
+                tokenserverUrl: unlockInfo.tokenserverURL
+            )
+        }
+    }
+
+    /**
      * Resets all sync metadata for history, including change flags,
      * sync statuses, and last sync time. The next sync after reset
      * will behave the same way as a first sync when connecting a new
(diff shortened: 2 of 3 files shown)


components/places/ios/History.swift
```diff
@@ -0,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import Foundation
```
tarikeshaq, line 4: this file doesn't need to exist 😬

megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/PlacesTests.swift
```diff
@@ -434,4 +434,135 @@ class PlacesTests: XCTestCase {
...
+        XCTAssertEqual("https://www.mozilla.com/foo/bar/", try! db.matchUrl(query: "mozilla.com/foo/bar"))
+        XCTAssertEqual("https://www.mozilla.com/foo/bar/baz", try! db.matchUrl(query: "mozilla.com/foo/bar/"))
+        XCTAssertEqual("https://www.mozilla.com/foo/bar/baz", try! db.matchUrl(query: "mozilla.com/foo/bar/baz"))
+        // Make sure the www/non-www doesn't confuse it
+        XCTAssertEqual("https://mozilla.com/a1/b2/", try! db.matchUrl(query: "mozilla.com/a1/"))
+
+        // Actual visit had no www
+        XCTAssertEqual(nil, try! db.matchUrl(query: "www.mozilla.com/a1"))
+        XCTAssertEqual("https://news.ycombinator.com/", try! db.matchUrl(query: "news"))
+    }
```
tarikeshaq, line 567: heh not sure we need those tests... they are very redundant and we don't really need to test the FFI layer (since it's all UniFFI'ed now!) I copied those over from the android code, but stopped halfway because I realized we actually can probably just get rid of them

mhammond: Yeah, this is a good question - I'm inclined to agree that these tests aren't valuable, and might even be harmful as we refactor (eg, it's just more redundant work if we intentionally change something.) Some "smoke" tests might make sense (eg, to test that we can load the library and open a connection or similar). I guess I feel the same about the docstrings/comments too - there's a real risk they will get out of date and even inconsistent between the kotlin and swift when the canonical source of truth remains rust. The problem here is that now there's not a good place for these docs, other than the UDL, and I'm not sure it makes sense to ask for the docs to be moved there. tl;dr, I'll be fine with whatever you think is best :)

tarikeshaq: That's a really good point with the docs, I think it's best to get rid of them here and make a point of keeping them up to date in the UDL

lougeniaC64:
> To understand our Places history APIs better, I decided a good first step is to walk through them and add their Swift equivalents with docs, and an excellent second step is to write some Swift tests for those functions to make sure they do what we expect them to do.
> 
> So far this was purely mechanical, and most my time was spent trying to write the docstrings
> 
> TODOs before this is ready:
> 
>     * [x]  Add tests for some of the history functions, may not need to add tests for all (we got lots of coverage in Rust 😄)

This looks good to me! If you were intending to add swift tests, I don't think that's necessary looking at the functions you've exposed. IMO as long as the underlying rust functions are covered, the swift tests shouldn't be required since there isn't much added logic in the swift API functions.

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,16 +1,14 @@
 {
-  "object": {
```
lougeniaC64, line 2: QQ: Why are these changes necessary?

mhammond (approved):
Assuming there's a good reason for the Package.resolved change, this looks great!

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
tarikeshaq: heh snuck in here, I imagine something to do with Xcode running and me switching branches 🤷 got rid of it!

PR #4992, opened 2022-06-09
Update to ktlint 0.45.1

badboy:
On a local checkout with `components/external/glean` checked out the older ktlint fails, because of trailing commas in Glean files (really these should be ignored regardless, so I'm adding that too).
Despite what ktlint says that's valid kotlin code and newer ktlint can handle that.

Now ktlint did move location. Glean is using that for a while now, and so is android-components, so it seems safe to switch.

Files were reformatted using `./gradlew ktlintFormat`.
Not super critical and it is a bit of churn for a whole lot of files, so feel free to drop this or defer it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -312,21 +312,25 @@ configurations {
 }
 
 dependencies {
-    ktlint "com.github.shyiko:ktlint:0.31.0"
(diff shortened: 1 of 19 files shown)


mhammond (approved):
rs=me

PR #4995, opened 2022-06-09
Do not try to lint external components

badboy:
That's only the "ignore Glean" part of #4992, no ktlint update, no reformatting, because [right now `main` is failing](https://firefox-ci-tc.services.mozilla.com/tasks/Uq56uiogRmeBHKYrG5e3ng/runs/1/logs/public/logs/live.log), so this might be quicker to apply then the lots-of-churn-because-reformatting PR.


bendk:
Looks good to me.  I'm just going to merge now, since the CI error was on main anyways.

PR #5004, opened 2022-06-17
Reworking the logins error handling (#3190)

bendk:
- Use `thiserror` only to define our errors rather than a weird
  combination of `thiserror` and `error_support`
- `LoginsError` is now our main error enum.  The only other struct left
  is `InvalidLoginsReason` which is acts as an error subtype.
- Dropped the `throw!` macro.  The error types are simple enough now
  that we can just use a return statement.
- Added note about PII
- Renamed `LoginsError::label()` to `LoginsError::error_type()` and
  added a note about how I think it should be used going forward.
- Updated the rest of the code to use the new API
- Removed the backtrace functionality from `sync-pass`.  I'm kind of
  doubt it was working anyways.

  - Note:


components/logins/src/error.rs
```diff
@@ -97,192 +86,71 @@ pub enum InvalidLogin {
...
-                }
-            }
-            // We can't destructure `err` without bringing in the libsqlite3_sys crate
-            // (and I'd really rather not) so we can't put this in the match.
-            ErrorKind::SqlError(rusqlite::Error::SqliteFailure(err, _))
+    /// Report this error to our tracking system if appropriate
+    pub fn track_error(&self) {
+        // TODO: replace this with new error reporting code:
+        //
+        //   - `log::error!` should change to `error_support::report_error!` once #4979 is merged.
```
mhammond, line 126: #4979 is merged :) So I think we can probably do that now and it would be replaced with a single line calling `report_error(self.error_type(), self.to_string())` or similar?

(hrm - I can't see where this is called though, so maybe it should just be removed?)

mhammond (changes requested):
This is a nice cleanup but I think there are a couple of things that might make it even better :)

components/logins/src/error.rs
bendk: I didn't add the `report_error()` call because I wanted to wait and see how the new system works for the `QueryReturnedNoRows` error before committing to the switch.

I completely forgot adding the call to `track_error()` and that's one of the bigger changes needed.  Before we were relying on the `Into<LoginsStorageError>` impl, which [UniFFI implicitly calls](https://github.com/mozilla/uniffi-rs/commit/775a5bf8ef9256eac8cbd42da6c78097de816f80).  I just pushed some code to implement this, but there were a few possible options and I'm not sure if I picked the best one.  Here are the options I could think of:
  - Wrap each `LoginStore` method in a macro -- this is what I ended up going with
  - Do the same thing with a proc-macro.  This would be more ergonomic, since we could wrap the entire `impl` block.  But it would be more work to implement.
  - Keep the `Error` -> `LoginsStorageError` conversion, but only for error tracking.  `LoginsStorageError` would have the exact same shape as `Error`.  The main reason I didn't go with this one, is because I'm not sure UniFFI should be implicitly calling `into()`

mhammond (changes requested):
Thanks - this is great, but I'd still like to continue the discussion about the naming, and the other error cases and lack of a sql-specific track_error string.

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
```diff
@@ -19,6 +19,27 @@ import org.mozilla.appservices.logins.GleanMetrics.LoginsStore as LoginsStoreMet
 import mozilla.telemetry.glean.private.CounterMetricType
 import mozilla.telemetry.glean.private.LabeledMetricType
 
+/**
+ * Typealias for backwards compatibility
+ */
+typealias LoginsStorageException = LoginsException
```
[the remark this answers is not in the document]
lougeniaC64: You know I would have assumed line 25 implies all of the type aliases in lines 26 - 42 but then looking at [all of the logins type alias declarations in a-c](https://github.com/mozilla-mobile/android-components/blob/b827bd3c9c90a315001be415380e8d748894017d/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L58) I'm guessing I'd be wrong. If the goal is to reduce breaking changes, then I suppose this is warranted. But since this PR already includes breaking changes (at least based on the change log), I'm on the fence. If we keep these nested type aliases we should at least use the naming in a-c's `SyncableLoginsStorage.kt` and remove those lines so we're only converting errors once.

CHANGES_UNRELEASED.md
```diff
@@ -23,3 +23,12 @@ Use the template below to make assigning a version number during the release cut
 
 ### What's New
   - Added initial work to support component FML files into an app FML. ([#4999](https://github.com/mozilla/application-services/pull/4999)). Please upgrade your nimbus-fml tooling at the same time as upgrading the megazord.
+
+## Logins
+### ⚠️ Breaking Changes ⚠️
```
lougeniaC64, line 28: Are there breaking changes in this PR? I might be missing some, but didn't you address all the breaking changes/

components/logins/src/error.rs
```diff
@@ -97,193 +83,93 @@ pub enum InvalidLogin {
...
-            ErrorKind::UrlParseError(_) => "UrlParseError",
-            ErrorKind::InvalidPath(_) => "InvalidPath",
-            ErrorKind::InvalidDatabaseFile(_) => "InvalidDatabaseFile",
-            ErrorKind::SqlError(_) => "SqlError",
-            ErrorKind::CryptoError(_) => "CryptoError",
-            ErrorKind::Interrupted(_) => "Interrupted",
-            ErrorKind::IOError(_) => "IOError",
-            ErrorKind::InvalidLogin(desc) => match desc {
+impl LoginsError {
+    // Get a short textual label identifying the type of error that occurred, but without specific
```
lougeniaC64, line 87: Great comment!

lougeniaC64:
Nice job, Ben! The [error.rs](https://github.com/mozilla/application-services/blob/5f87d242d508d6e4101ea23daa920ac67fd31c34/components/logins/src/error.rs) file is a lot easier to understand now. 🎉 I left a comment about the aliases but otherwise 👍🏾.

components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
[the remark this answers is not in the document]
bendk: I tried some more to get this to work, but couldn't.  I think the issue is related the Kotlin -> Java translation.

I think the solution @lougeniaC64 proposed would work, but would also be a breaking change, so I we should just accept the rename.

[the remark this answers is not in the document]
lougeniaC64: Also, @mhammond on another reading of your comment I'm wondering if you were asking if we need to expose all of these error types. Was that the case? If so I'm wondering what we'd expect consumers to do when errors like  `SyncAdapterError` or `SqlError` are thrown. Neither error shows up searches of a-c and iOS which seems to indicate that they're not explicitly caught. So maybe consolidating some of the non-actionable and non-fatal errors into a generic error type would be more effective? Just spitballing, mainly just wanted to make sure I didn't misinterpret and potentially bury Mark's sentiment.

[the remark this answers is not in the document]
bendk: I'm pretty sure that's what @mhammond meant, and I think it's an important point regardless.  I reworked this PR to take this into account more.

bendk:
Thanks for all the review, you both brought up some great points.  I reworked this PR to try to address two major points:

### Errors sent to the consumer should be actionable

It's better to expose a small number of variants that the application actually do something about.   I added `LoginsStorageException` back, but tried to only define [variants](https://github.com/mozilla/application-services/blob/ad94e996f0f7dc212b3f361b8705dca48b1888be/components/logins/src/error.rs#L98-L133) that might be relevant to the app.  This was largely based on the [android-components README](https://github.com/mozilla-mobile/android-components/blob/main/components/service/sync-logins/README.md#which-exceptions-do-i-need-to-handle).  This was also based on reading the `error-stack` [docs](https://docs.rs/error-stack/latest/error_stack/) and [announcment](https://hash.dev/blog/announcing-error-stack).  I don't think we want to copy they're error handling system, but one thing they discuss is having errors change when you cross an API boundary and I think that makes a lot of sense.  Lastly, this means that we can call `report_error()` in the `from()` method rather than wrapping everything in a macro.

### How are we going to track these errors?

I had some thoughts in my head about this, but it wasn't reflected in the code.  I updated [the error reporting code](https://github.com/mozilla/application-services/blob/ad94e996f0f7dc212b3f361b8705dca48b1888be/components/logins/src/error.rs#L178-L265) to reflect how I see error reporting happening.  Couple notes on this:

  - I decided to replace the old code that calls `log::error` with code that calls `error_support::report_error!`.  I'm pretty confident that this should work and a new nightly cycle is starting next week so we can fix things if this causes issues.
  - Right now the `classify()` does a [bit of digging into the error details](https://github.com/mozilla/application-services/blob/ad94e996f0f7dc212b3f361b8705dca48b1888be/components/logins/src/error.rs#L223-L233) to figure out how to group them, but I think we can more.  For example, maybe unexpected SQL errors should be grouped by their error code (something like `SqlError{sqlite_error_code}`).  I'm thinking we can iterate on this more down the road.
  - Mark brought up a good point that this is all being done at the top of the stack, but sometimes we want information that's only available near the source of the error.  `error-stack` has a pretty interesting way to do this, but I'm thinking we can leverage breadcrumbs.  For example, [this breadcrumb](https://github.com/mozilla/application-services/blob/main/components/places/src/storage/history.rs#L414) is reported right before we make the call that we think results in the NoRowsReturned error.

components/logins/src/error.rs
```diff
@@ -97,192 +85,181 @@ pub enum InvalidLogin {
...
+    fn from(error: LoginsError) -> LoginsStorageError {
+        // We convert errors before sending them across the API boundary to the consuming
+        // application, so this is a good time to report them.
+        error.report();
+
+        match error {
+            LoginsError::InvalidLogin(inner) => Self::InvalidRecord(inner),
+            LoginsError::NoSuchRecord(guid) => Self::NoSuchRecord(guid),
+            LoginsError::CryptoError(inner) => Self::CryptoError(inner.to_string()),
+            LoginsError::InvalidKey => Self::CryptoError("InvalidKey".to_string()),
```
bendk, line 145: This one is slightly risky, since the recommendation for `CryptoError` is to delete your database and start over.  I'm pretty sure that the same thing is true for `InvalidKey` since:
  - An invalid local encryption key is effectively the same as a valid key that can't decrypt database data
  - We basically don't return `InvalidKey` for sqlcipher databases anymore.
  - We could maybe return `InvalidKey` is during the sqlcipher migration, but that also means indicates the app should delete the database and start fresh.

That's my logic at least, does that make sense to you?  @mhammond @lougeniaC64

lougeniaC64: @bendk This makes sense to me. At present we recreate the database and/or report the error with both `CryptoError` and `InvalidKey`. We can always update this code if the handling needs to diverge in the future (although I can't imagine why that would happen).

bendk:
#5033 is an update version of this and I think we should go with that branch.

PR #5009, opened 2022-06-22
Adding Glean back to the iOS megazord

travis79:
This is so we can use it to record metrics in Rust on iOS


travis79:
Seems I forgot to move the exposure event to Rust completely for iOS...   not quite ready yet.

components/nimbus/src/lib.rs
```diff
@@ -488,6 +495,25 @@ impl NimbusClient {
...
     }
+
+    pub fn record_exposure_event(&self, feature_id: String) -> Result<()> {
+        // Next, we search for any experiment that has a matching featureId. This depends on the
+        // fact that we can only be enrolled in a single experiment per feature, so there should
+        // only ever be zero or one experiments for a given featureId.
+        if let Some(exp) = self
+            .get_active_experiments()?
+            .iter()
+            .find(|exp| exp.feature_ids.contains(&feature_id))
```
jhugman, line 506: I feel sure we can do something a bit more efficient than this, but I can't think of a better way that's as non-invasive as this.

components/nimbus/src/nimbus.udl
```diff
@@ -197,6 +197,13 @@ interface NimbusClient {
     // It's first use is in GleanPlumb message helper, to add extra parameters to URLs.
     [Throws=NimbusError]
     NimbusStringHelper create_string_helper(optional JsonObject? additional_context = null);
+
+    // If an active experiment exists with the provided feature_id, then an exposure event will be
+    // recorded. If there is no active experiment, then nothing happens. This makes this function
+    // safe to call at the point of exposure for the feature regardless of whether an experiment is
+    // active or not.
+    [Throws=NimbusError]
+    void record_exposure_event(string feature_id);
```
jhugman, line 206: 😻

jhugman (approved):
This is looking lovely. 😻 

I have added comments which are mostly filing bugs and questions for my own understanding. 

I haven't reviewed anything in xcodeproj files, or .h files, and so we should probably ask one of the iOS teams to look at those and give a r+.

From my point of view, this is brilliant, it's been a long time coming, so thank you.

components/nimbus/src/enrollment.rs
```diff
@@ -373,40 +375,75 @@ impl ExperimentEnrollment {
...
+                    &self.slug,
+                    enrollment_id,
+                    branch,
+                    match reason {
+                        DisqualifiedReason::NotTargeted => Some("targeting"),
+                        DisqualifiedReason::OptOut => Some("optout"),
+                        DisqualifiedReason::Error => Some("error"),
+                    },
+                    EnrollmentChangeEventType::Disqualification,
+                )
```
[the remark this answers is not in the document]
travis79: Yes, we should no longer need this. Unfortunately we rely upon this structure in our tests so removing it completely is a bit messy.

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -3,7 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
-import Glean
+#if canImport(MozillaRustComponents)
+    import MozillaRustComponents
+#endif
```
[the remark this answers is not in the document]
travis79: Same reason as before, Nimbus can't find Glean in the megazord without this. I don't know if MozillaRustComponents is always available or not, just following the pattern seen in other components that have dependencies inside of AppServices

components/nimbus/src/lib.rs
travis79: Yes, hoping @tarikeshaq might have some ideas on this. It was as close to the logic of what we were doing in Swift as I could arrive at.

components/nimbus/tests/test_telemetry.rs
```diff
@@ -0,0 +1,219 @@
...
+            ClientInfoMetrics::unknown(),
+            false,
+        );
+
+        client.set_experiments_locally(experiment_json)?;
+        let change_events = client.apply_pending_experiments()?;
+
+        // Check for Glean enrollment event
+        let events: Option<Vec<RecordedEvent>> =
+            glean_metrics::nimbus_events::enrollment.test_get_value("events");
```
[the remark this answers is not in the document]
travis79: There should already be an issue I think, around removing the Kotlin and Swift Glean instrumentations. Let me see if I can find and link it.

.circleci/config.yml
```diff
@@ -120,6 +120,7 @@ commands:
       - run: sudo apt-get update
       - run: sudo apt-get install python tcl
       - run: sudo apt-get install python3-venv
+      - run: sudo apt-get install python3-pip
```
tarikeshaq, line 123: Do you know if we need to include this in our `building` docs: https://github.com/mozilla/application-services/blob/main/docs/building.md?

components/nimbus/Cargo.toml
```diff
@@ -45,3 +46,4 @@ mockito = "0.31"
 env_logger = "0.7"
 clap = "2.33.3"
 tempfile = "3"
+glean = { path = "../external/glean/glean-core/rlb" }
```
tarikeshaq, line 49: I don't think we need to include it in the dev-dependencies if it's already a dependency

components/nimbus/Cargo.toml
```diff
@@ -35,6 +35,7 @@ once_cell = "1"
 uniffi = { version = "^0.18", optional = true }
 chrono = { version = "0.4", features = ["serde"]}
 unicode-segmentation = "1.8.0"
+glean = { path = "../external/glean/glean-core/rlb" }
```
tarikeshaq, line 38: Hmmm I'll need to think about this a little, from an "impact on android" perspective.

Two questions:
1) is always guaranteed that if the glean binary isn't initialized, all glean API's are no-ops? Because that will be the case here in android, where they use libxul
2) How OK are we with duplicating the glean binary in android?

I'm not sure about the first question, and the second question we can maybe measure?

Alternatively, for now, we can gate including glean on a feature https://doc.rust-lang.org/cargo/reference/features.html

we can even have it enabled by default, then in the [megazord for android](https://github.com/mozilla/application-services/blob/main/megazords/full/Cargo.toml) we can turn the feature off

components/nimbus/build.rs
```diff
@@ -8,14 +8,16 @@ pub fn main() {
     #[cfg(feature = "uniffi-bindings")]
     uniffi_build::generate_scaffolding("./src/nimbus.udl").unwrap();
 
+    let python = if cfg!(windows) { "python" } else { "python3" };
```
tarikeshaq, line 11: Hmmm will this always be true? Like, if a mac user follows our instructions in the building docs, will they end up with a `python3` binary?

components/nimbus/build.rs
```diff
@@ -26,5 +28,9 @@ pub fn main() {
         .status()
         .expect("Error generating Glean Rust bindings");
 
+    if !status.success() {
+        panic!("Failed generating Glean Rust bindings!")
```
tarikeshaq, line 32: Is there a way to print some more information here in case a build fails for a developer? Something that can point them in some direction

components/nimbus/ios/Nimbus/GleanPlumbHelpers.swift
```diff
@@ -3,7 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
-import Glean
+#if canImport(MozillaRustComponents)
+    import MozillaRustComponents
+#endif
```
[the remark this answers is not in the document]
tarikeshaq: The `#if canImport`... was added when we had our code used in both a Carthage and a Swift Package Manager environment. The `MozillaRustComponents` wasn't used in Firefox iOS at the time, but was used in Focus.

Now that Carthage is out of the way, we can probably get rid of the `#if canImport` and just import `MozillaRustComponents` - we can file a follow-up to clean it up from everywhere.


All that said, we probably still want to keep this in any Glean files that are added, assuming glean's local development flow and local testing doesn't build our megazord

components/nimbus/src/enrollment.rs
```diff
@@ -373,40 +375,75 @@ impl ExperimentEnrollment {
...
-                &self.slug,
-                enrollment_id,
-                branch,
-                None,
-                EnrollmentChangeEventType::Enrollment,
-            ),
+            } => {
+                nimbus_events::enrollment.record(nimbus_events::EnrollmentExtra {
+                    branch: Some(branch.clone()),
+                    enrollment_id: Some(enrollment_id.clone().to_string()),
```
tarikeshaq, line 381: nit: do we need the `clone` here? I would of thought `to_string()` is enough as that probably allocates a new string anyways

(and ditto for below)

tarikeshaq: As a tiny follow-up we file, I notice we have to clone a bunch of times, when the `branch` and `enrollment_id` values don't really change

but that makes sense since the generated glean metrics code expects `String`s... can we file a ticket/issue against glean/glean-metrics to allow for strings to be reference, or just use a [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html)

components/nimbus/src/lib.rs
tarikeshaq: Yeah I don't have any brilliant ideas here 😞 We can modify our state to include a map from feature_id -> experiment maybe? but that's a little intrusive and we'll need to be careful keeping it up to date as experiments change

Regardless I'm happy with this, and we can always improve it if we recognize that the time spent iterating over experiments is too expensive

megazords/ios-rust/Cargo.toml
```diff
@@ -20,4 +20,6 @@ autofill = { path = "../../components/autofill" }
 push = { path = "../../components/push" }
 tabs = { path = "../../components/tabs" }
 places = {path = "../../components/places" }
+glean-core = { path = "../../components/external/glean/glean-core" }
+glean = { path = "../../components/external/glean/glean-core/rlb" }
```
tarikeshaq, line 24: out of curiosity, why do we need to include both? do we need to expose the `rlb` at this layer?

megazords/ios-rust/build-xcframework.sh
```diff
@@ -146,6 +146,7 @@ cp "$WORKING_DIR/$FRAMEWORK_NAME.h" "$COMMON/Headers"
 cp "$REPO_ROOT/components/rc_log/ios/RustLogFFI.h" "$COMMON/Headers"
 cp "$REPO_ROOT/components/viaduct/ios/RustViaductFFI.h" "$COMMON/Headers"
 $CARGO uniffi-bindgen generate "$REPO_ROOT/components/nimbus/src/nimbus.udl" -l swift -o "$COMMON/Headers"
+$CARGO uniffi-bindgen generate "$REPO_ROOT/components/external/glean/glean-core/src/glean.udl" -l swift -o "$COMMON/Headers"
```
tarikeshaq, line 149: hmmm we'll need to be extra careful here to make sure the `uniffi_bindgen` version glean uses to generate its scaffolding is compatible with the version we use here. This also means that we are essentially tying our `uniffi` version in `application-services` with the `uniffi` version used in glean (at least in the short term, until we get the `uniffi` versioning work done.

Versions being tied doesn't sit well with me, but I can't see a way around it for now/

So the consequences of this are:
-  Whenever we bump the glean submodule (and it includes an updated `uniffi`) the iOS builds will fail, and CI will catch them forcing us to either align the version here in application-services, or go back and align the version in glean and cutting another release there
- We will **not** be able to bump uniffi in this repository on its own. Whenever we want to bump uniffi here, we'll have to first bump it in glean, cut a release there, then update it here

I'll try to spend a bit of time thinking about this so we can move without blocking this work
cc @mhammond and @bendk

tarikeshaq (changes requested):
Thanks a ton @travis79 🚀 🎉 !! Once we get this through we should have a good story for iOS

I had a couple of blocking thoughts/comments that we might need to work out before we land this:

1) The comment about impact on android and possibly feature gating
2) The uniffi version compatibility comment

The rest are nits and questions. I'll timebox sometime to think about the `uniffi` comment and see if we can comeup with a good way to move forward!

components/nimbus/src/lib.rs
jhugman: > We can modify our state to include a map from feature_id -> experiment maybe? but that's a little intrusive and we'll need to be careful keeping it up to date as experiments change
Yes, we do keep a map of `feature_id` to `EnrolledFeatureConfig`, which could keep an optional experiment id.

It's kept in memory and updated everytime `apply_pending_experiments` is called.

> Regardless I'm happy with this, and we can always improve it if we recognize that the time spent iterating over experiments is too expensive

Agreed.

megazords/ios-rust/build-xcframework.sh
tarikeshaq: I wanted to point out that I chatted briefly with @bendk about this, and with the upcoming changes in https://github.com/mozilla/uniffi-rs/pull/1271 I'm less concerned about the coupling between the versions. That PR changes compatibility to be less about the uniffi version and more about the low-level contract between the bindings and the scaffolding (which we don't change often at all)

And since the UDL we use is the same one from the submodule, there shouldn't be any incompatibility there.

I won't block on this anymore, and even if this PR lands first - the uniffi naming changes are already approved/agreed upon so it'll catch up quickly 🚀

.circleci/config.yml
travis79: We shouldn't, this just isn't installed by default with python in the CI environment. Typical local building you would get pip along with python3 installation (or a venv)

megazords/ios-rust/Cargo.toml
travis79: You may be correct here, I just added both, but I can try without the rlb dependency here.

components/nimbus/Cargo.toml
travis79: The answer to the first question is yes, uninitialized Glean is a no-op except for the queue for metrics recorded prior to init, but once it is full it too becomes a no-op. I don't expect much overhead from Glean in this regard and I am confident in Glean's logic for this.

Adding the RLB is just adding the binding layer on top of glean-core, so it's a fairly thin shim. There should still only be one glean-core coming from libxul and the rlb will interact with it on Android, and interact with the megazorded glean-core on iOS.  I didn't notice any big size regressions while testing in Focus, but happy to do a more detailed measurement if product has any issues with it.

components/nimbus/build.rs
travis79: Jan-Erik has created a glean-build crate as a wrapper around the glean_parser step, so this script should no longer be needed. I'll be pushing a commit to migrate to this build crate to address your concern here, this was simply a hack to get it to work with Windows development.

components/nimbus/build.rs
travis79: The glean_parser output would be the output that contains the information needed. Again, with my plan to switch to the glean-build crate rather than the script, the output should be available. If it isn't, I'll file a bug to see that the important info makes it into stdout.

components/nimbus/src/enrollment.rs
travis79: I am probably a little over-zealous with `clone()` as a way to get along with the borrow checker, I'll try your suggestions to see if I can make this better, but at least these aren't large objects so I wasn't too worried about cloning them.

megazords/ios-rust/build-xcframework.sh
travis79: I think A-S and Glean are likely going to both be using a fairly recent version of uniffi most of the time, so hopefully the compatibility problems will be minimal. Since there are already changes proposed that will make this even easier, I'm okay with keeping Glean on the same version of A-S, and I'm sure we can work it out if we need to differ for some reason.

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -166,8 +124,8 @@ private extension Nimbus {
  */
 internal extension Nimbus {
     func setGlobalUserParticipationOnThisThread(_ value: Bool) throws {
-        let changes = try nimbusClient.setGlobalUserParticipation(optIn: value)
-        postEnrollmentCalculation(changes)
+        _ = try nimbusClient.setGlobalUserParticipation(optIn: value)
+        postEnrollmentCalculation()
```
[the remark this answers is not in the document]
travis79: [Issue Filed](https://mozilla-hub.atlassian.net/browse/EXP-2630)

components/nimbus/Cargo.toml
tarikeshaq: Ideally, I still would like to see either a [`cfg` on android](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os) or some type of cargo feature that protects non-ios consumers from including a duplicate glean (and helps if anyone is confused why calls are no-ops) - if that does seem like a lot of work for what we would remove when glean-in-rust hits android I'm willing to concede, I'll leave it as a nice to have as long as android still builds and works OK

components/nimbus/build.rs
```diff
@@ -2,28 +2,15 @@
...
-            "5.1.1",
-            "translate",
-            "-f",
-            "rust",
-            "-o",
-            out_dir.as_str(),
-            "./metrics.yaml",
-        ])
-        .status()
+    Builder::default()
```
tarikeshaq, line 11: I love this!!! A lot better than shelling out 😍 🎉

components/nimbus/tests/test_telemetry.rs
```diff
@@ -0,0 +1,219 @@
...
+mod common;
+#[cfg(feature = "rkv-safe-mode")]
+#[cfg(test)]
+mod test {
+
+    use super::common::new_test_client_with_db;
+    use glean::{test_reset_glean, ClientInfoMetrics, Configuration, RecordedEvent};
+    use nimbus::glean_metrics;
+
+    #[cfg(feature = "rkv-safe-mode")]
```
tarikeshaq, line 14: nit: do we need this? would the `cfg` on the `test` module protect against this already?

megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/LoginsTests.swift
```diff
@@ -4,15 +4,13 @@
...
 
-import Glean
 import XCTest
 
 class LoginsTests: XCTestCase {
     var storage: LoginsStorage!
 
     override func setUp() {
         super.setUp()
-        Glean.shared.resetGlean(clearStores: true)
```
tarikeshaq, line 15: Just out of curiosity, what's going in here and why can we get rid of it (instead of calling some other function more directly)

megazords/ios-rust/MozillaTestServices/glean.udl
```diff
@@ -0,0 +1,549 @@
+namespace glean {
```
tarikeshaq, line 1: Curious why we needed to add a second `udl` here - is this the same `udl` in glean? Is it possible to point to the submodule in all of our scripts instead of duplicating the UDL?

megazords/ios-rust/ios/Generated/glean.swift
```diff
@@ -0,0 +1,4159 @@
+// This file was autogenerated by some hot garbage in the `uniffi` crate.
```
tarikeshaq, line 1: Why do we need this file? Shouldn't this be generated by `uniffi` on `rust-components-swift`? I don't mind it generated here, but we'll need some consistent script that runs to generated updates.


(if we need it here, we can start a discussion around generally checking-in generated uniffi files, which TBH I'm in support of, but it seems beyond the scope of this PR)

megazords/ios-rust/ios/Generated/gleanFFI.modulemap
```diff
@@ -0,0 +1,6 @@
+// This file was autogenerated by some hot garbage in the `uniffi` crate.
```
tarikeshaq, line 1: ditto on generated file, (and I guess the `.h` file too)

tarikeshaq (changes requested):
Almost there! Thanks @travis79!!!

Once you've addressed the comments I've dropped we should be in a good place to land this.

The last thing I want to confirm before we land this, can we try a local build of Fenix with this branch to confirm that we aren't causing any breakages. It would also be ideal if we can confirm that Fenix is still recording telemetry as expected and that nimbus is working properly

megazords/ios-rust/MozillaTestServices/glean.udl
travis79: Yes, that was my intention, Glean now has two udl files and does some swapping between them when building the iOS megazord (via somewhat hacky use of file renaming, but it works for now). I think I just forgot to remove this one once I moved the udl needed to build the iOS megazord compatible Glean into the Glean repo.

megazords/ios-rust/ios/Generated/glean.swift
travis79: This shouldn't be here, I think it's a side effect of me playing around with the TestServices project trying to get it to work.  Let me clean that up...

megazords/ios-rust/ios/Generated/gleanFFI.modulemap
travis79: Ditto on this shouldn't be here and may just be leftover from playing with getting the TestServices to build

components/nimbus/tests/test_telemetry.rs
travis79: I think it might, this is most likely just some copy-pasta from Glean test examples.

megazords/ios-rust/build-xcframework.sh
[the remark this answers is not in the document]
travis79: No, that udl in TestServices isn't supposed to be there so we should rely on the one in the submodule

megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/LoginsTests.swift
travis79: Which probably means I can remove the import statement too :D

components/nimbus/src/enrollment.rs
```diff
@@ -373,40 +376,88 @@ impl ExperimentEnrollment {
...
+
+                let mut extra = HashMap::new();
+                extra.insert("enrollment_id".to_string(), enrollment_id.to_string());
+
+                #[cfg(feature = "builtin-glean")]
+                glean::set_experiment_active(
+                    enrollment_id.to_string(),
+                    branch.to_string(),
+                    Some(extra),
+                );
```
tarikeshaq, line 395: super nit: we can `cfg(feature=` a block here instead of multiple `cfg`s, i.e adding curly brackets before and after the part that is for glean-in-rust

megazords/ios-rust/Cargo.toml
```diff
@@ -8,6 +8,9 @@ license = "MPL-2.0"
 [lib]
 crate-type = ["staticlib"]
 
+[features]
+builtin-glean = ["nimbus-sdk/builtin-glean"]
```
tarikeshaq, line 12: At first I was going to suggest replacing this with a `features = ["builtin-glean"]` entry in the dependencies section for nimbus, but I realized that I actually like this - in case we ever want to test the not-glean-in-rust in iOS! Awesome! 🚀

tarikeshaq (approved):
🚀 🚀 AWESOME!!! Thanks a ton for driving this @travis79


LETS 🚢 :shipit:

components/nimbus/src/enrollment.rs
tarikeshaq: eeeh I relalized there is a line in between that is not cfg'ed nvm this comment!

travis79: Yeah, I should have done that here, the extra isn't really used without Glean and I missed conditionally compiling it away anyways.  I'll update this to use the `cfg!() {}` macro and wrap the whole block of Glean related code.

travis79:
Adding the `do-not-merge` label until we get ready to land this for v107 nightlies (mid-September)

tarikeshaq:
@travis79 since this has gotten a little stale, do you know what the priority of this work is now? (i.e do we expect we'll get to it anytime soon?)

I don't object to keeping the PR open until we get to it, just curious what a timeline might look like if we have one?

travis79:
> 

@tarikeshaq Yes, the priorities on this shifted a bit. This is still desirable to do but I don't have time to get back to it immediately. Ultimately I'd love to be able to solve both iOS and Android when we get back to this. Let's close this in favor of starting fresh without any bitrot to deal with.

PR #5021, opened 2022-06-28
Update protobuf to latest releases.

rvandermeulen:
Updates protobuf to version 3.21.2 and the gradle plugin to version 0.8.18.
Upstream protobuf also has native M1 Mac support now, so remove a hack which
isn't needed anymore.

### DIFF
--- a/build-scripts/protobuf-common.gradle
+++ b/build-scripts/protobuf-common.gradle
@@ -20,13 +20,7 @@ android {
 
 protobuf {
     protoc {
-        // This is a hack for Apple M1 since protobuf does not compile for M1s yet
-        // However, M1s can just use the x86_64 just fine
-        if (osdetector.os == "osx") {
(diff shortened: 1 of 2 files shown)


rvandermeulen:
Gecko recently picked up the 21.2 release in [bug 1773604](https://bugzilla.mozilla.org/show_bug.cgi?id=1773604). Not sure how widely-used protobuf is in A-S these days, but submitting this in case we're interested in keeping it in sync. I don't have a good way of testing this besides whatever checks run in CI, however.

mhammond (approved):
Awesome, thanks for picking this up Tarik!

tarikeshaq:
Hmm looks like circleci didn't run, I imagine because of the whole github saml stuff, I'll close and re-open the PR to see if I can trigger it

tarikeshaq (approved):

Issue #5013, opened 2022-06-23
Remove Swift DispatchQueue usage from tabs and places


lougeniaC64:
Provided the work to complete issue #5012 is done and no adverse impact to the performance of logins component (or any adverse impact has been addressed), remove `DispatchQueue` instantiation and usage from [Tabs.swift](https://github.com/mozilla/application-services/blob/546520b3ad560509cde3241549c4af232c60e5be/components/tabs/ios/Tabs/Tabs.swift) and [Places.swift](https://github.com/mozilla/application-services/blob/a6af58fac64ab937a63716396443534b8ae90bfa/components/places/ios/Places/Places.swift).

Note: `DispatchQueue`s are also used in the rc_log and fxa-client components, though differently and more practically and purposefully than in the logins, tabs, and places components. Also it's worth noting that the database layer logic in iOS for the three latter components creates additional `DispatchQueue`s that our API calls are wrapped in which could cause thread locking in iOS. All that to say removing `DispatchQueue` usage from logins, tabs, and places is the primary focus of this effort.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3220)
┆Epic: Refactor iOS DispatchQueues

PR #5024, opened 2022-06-28
Remove DispatchQueue usage from logins component

### LINKED ISSUE #5012: Remove Swift DispatchQueue usage from logins
As a part of our effort to clean up the `DispatchQueue` usage in the Swift layer of our components, use the logins component (which is stable and not under active development) to test the removal of the `DispatchQueue` usage there. This involves removing the instantiation of the  DispatchQueue` in [LoginsStorage.swift](https://github.com/mozilla/application-services/blob/029c44de2b6fd800abe408da3c760727b3e285c2/components/logins/ios/Logins/LoginsStorage.swift#L18) and all the calls to submit work items to the queue within the API functions in that file.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3218)
┆Epic: Refactor iOS DispatchQueues
┆Sprint End Date: 2022-08-05

lougeniaC64:
This fixes #5012.

In the Swift layer of some of our components we are unnecessarily executing our API calls with `DispatchQueue`s. This is most likely a remnant of the manual FFI layer that existed prior to our UniFFI usage. What remains is an anti-pattern in which a call in our Swift layer executes with a `DispatchQueue` as below

```
private let queue = DispatchQueue(label: "com.mozilla.logins-storage")
...
open func reset() throws {
    try queue.sync {
        try self.store.reset()
    }
}
```

and is then called in the iOS repo with another `DispatchQueue`

```
queue = DispatchQueue(label: "RustLogins queue: \(databasePath)", attributes: [])
...
public func resetSync() -> Success {
    let deferred = Success()

    queue.async {
        ...
        try self.storage?.reset()
        ...
    }
    ...
}
```

In these cases our usage of `DispatchQueue`s is needless and potentially problematic. To address this issue, we're removing our `DispatchQueue` usage from the logins component because it is currently stable. If that removal is successful, we will address issue #5013 which addresses the remaining calls.

  - Note:

### DIFF
--- a/components/logins/ios/Logins/LoginsStorage.swift
+++ b/components/logins/ios/Logins/LoginsStorage.swift
@@ -8,14 +8,8 @@ import UIKit
 
 typealias LoginsStoreError = LoginsStorageError
 
-/*
- ** We probably should have this class go away eventually as it's really only a thin wrapper
- * similar to its kotlin equiavlent, however the only thing preventing this from being removed is
- * the queue.sync which we should be moved over to the consumer side of things
- */
 open class LoginsStorage {
     private var store: LoginStore
-    private let queue = DispatchQueue(label: "com.mozilla.logins-storage")
 
     public init(databasePath: String) throws {
         store = try LoginStore(path: databasePath)
@@ -24,38 +18,28 @@ open class LoginsStorage {
     /// Delete all locally stored login sync metadata. It's unclear if
     /// there's ever a reason for users to call this
     open func reset() throws {
-        try queue.sync {
-            try self.store.reset()
-        }
+        try store.reset()
     }
 
     /// Delete all locally stored login data.
     open func wipe() throws {
-        try queue.sync {
-            try self.store.wipe()
-        }
+        try store.wipe()
     }
 
     open func wipeLocal() throws {
-        try queue.sync {
-            try self.store.wipeLocal()
-        }
+        try store.wipeLocal()
     }
 
     /// Delete the record with the given ID. Returns false if no such record existed.
     open func delete(id: String) throws -> Bool {
-        return try queue.sync {
-            return try self.store.delete(id: id)
-        }
+        return try store.delete(id: id)
     }
 
     /// Bump the usage count for the record with the given id.
     ///
     /// Throws `LoginStoreError.NoSuchRecord` if there was no such record.
     open func touch(id: String) throws {
-        try queue.sync {
-            try self.store.touch(id: id)
-        }
+        try store.touch(id: id)
     }
 
     /// Insert `login` into the database. If `login.id` is not empty,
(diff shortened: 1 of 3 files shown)


components/logins/ios/Logins/LoginsStorage.swift
```diff
@@ -8,14 +8,8 @@ import UIKit
 
 typealias LoginsStoreError = LoginsStorageError
 
-/*
- ** We probably should have this class go away eventually as it's really only a thin wrapper
- * similar to its kotlin equiavlent, however the only thing preventing this from being removed is
- * the queue.sync which we should be moved over to the consumer side of things
- */
```
mhammond, line 15: Do you think we should eventually try and remove this file? It looks like now it's really only the `sync` method which does some arg massaging. If you think that's still a reasonable idea, maybe keep a comment here saying that's still something we should consider?

docs/adr/0005-swift-dispatchqueue-usage.md
```diff
@@ -0,0 +1,73 @@
...
+
+Lastly, tasks can be submitted to a `DispatchQueue` either synchronously (which blocks the queue until the submitted task is complete) or asynchronously (which returns immediately after the task is submitted to the queue). Best practice advises that synchronous calls be used only when needed and should be avoided entirely when scheduleding tasks concurrently or on the main queue.
+
+### Our `DispatchQueue` Usage
+
+There are places in application services code where `DispatchQueues` are appropriately used. In the FxA component where we have functions that update the UI (mostly via [closure function parameters](https://github.com/mozilla/application-services/blob/451bcc2fe8fe6675ca3de962a4b38b3fa181a806/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L108)) or [dispatch notifications via `NotificationCenter`](https://github.com/mozilla/application-services/blob/72b827c3e0f883163762857fd766df1aeb060725/components/fxa-client/ios/FxAClient/FxAccountDeviceConstellation.swift#L49), our use of non-custom queues makes sense.
```
mhammond, line 23: I'm mildly surprised that is the right thing to do for something like app-services, and the appropriate async policy can't still be owned by the main app (eg, it's not obvious from reading that FxA code that the completion handler has anything to do with UI at all). However, this is way above my pay-grade, so if this is the opinion of the iOS team it's fine with me.

mhammond (approved):
I'm slightly sad that our new approach isn't as simple as "kill all the dispatch queues" (and if it was, we probably wouldn't even need the adr), but if this reflects the consensus of the iOS team, then it LGTM!

components/logins/ios/Logins/LoginsStorage.swift
nbhasin2: Wondering if this or something similar can live in Firefox iOS? 

Does A~S team uses this class for testing any Rust component? 

cc @lougeniaC64

nbhasin2 (approved):

docs/adr/0005-swift-dispatchqueue-usage.md
lougeniaC64: @mhammond I was squinting when I made this call so I'd be very open to being wrong because this is _way_ above my pay-grade as well. I also read that sending Notification Center events doesn't require queues but I can't find that blog post for the life of me.

The long and short of it is I'm way more confident in my assessment of the storage layer of the logins, bookmarks, and tabs component than I am about our usage of queues in FxA or Rust_Log.

lougeniaC64: @nbhasin2 @OrlaM Would you agree with my assessment of the `DispatchQueue` usage in FxA and RustLog?

nbhasin2:
One small concern, what will happen if two multiple operations happen at the same time 

store -> delete and store -> add

my assumption is that sync was added to cope with threading issue and make sure all operations happen in sequential order, if that is not the case and store itself is thread-safe then we can remove dispatch queue `sync`

docs/adr/0005-swift-dispatchqueue-usage.md
OrlaM: It looks like that callback goes to a file that updates a web view within FxA (FxAWebViewController) which is UI so it probably does need to be on the main thread, but in this scenario it would be more clear if FxAWebViewController ensured it was on the main thread when doing updates instead of the business logic needing to be aware of how it's to be used.

For the changes to the DispatchQueue usage I'm having a hard time telling whats actually happening when a rust call is being made. Is it making a HTTP request, updating a data base or a combination? Is thread safety handled at a lower level and therefore it's redundant at this level?

If it's making a database update then the DispatchQueue calls made sense. This is currently the best way in Swift of making the database calls thread safe. But if this is already being handled at a lower level then these calls would be redundant. 

If it's making a HTTP request then it doesn't necessarily need to be handled by a custom queue because serialization doesn't matter but it does need to be moved off the main thread in some fashion so it doesn't block the UI.

If the concern is the duplication of this logic between LoginsStorage.swift and RustLogins.swift then it's better to remove it from RustLogins.swift instead. It's better for these kinds of thread safety checks to happen at the lowest possible level in my opinion.

lougeniaC64:
After discussing this with the iOS engineers during the a-s work week, we've decided that we're going to table these changes and allow the iOS team to prioritize their work and holistically analyze our current `DispatchQueue` usage across both the a-s and iOS repos and make any necessary improvements. @nbhasin2 @OrlaM

PR #5026, opened 2022-06-30
Reinstate older orphaned autofill tests.

mhammond:
These tests were accidentally orphaned a long time ago, and
went quite stale. This re-introduces them.


skhamis (approved):
Thanks for making these changes, definitely will help us catch some of these issues should they ever pop up again -- LGTM!

Issue #5020, opened 2022-06-27
addresses/creditcards: Store error: JSON Error: missing field `entry`


bendk:
We currently have about 2000 of these failures a day for both addresses and credit cards.  I think it's a 100% failure rate.  We should investigate this and fix it.  AFAICT, this is only happening on nightly, we don't try to sync either on beta or release (although we just switched to a new nightly cycle, so maybe we'll start seeing them on beta.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3225)
┆sprintEndDate: 2022-07-08

mhammond:
Hey @bendk, in what dashboard did you see these errors? I've worked out the problem isn't incoming tombstones, but can see an error when tombstones are stored in the mirror - but this makes it very unlikely that many users are impacted, so I'm interested to try and dig into the actual failure rate.

mhammond:
> Hey @bendk, in what dashboard did you see these errors?

Please ignore this, Sammy put me on to the new dashboard!

PR #5030, opened 2022-07-05
Fix autofill sync failures when tombstones are in the mirror.

mhammond:
The new tests, without the corresponding change in incoming.rs, fail with
`Error: JsonError(Error("missing field `entry`", line: 0, column: 0))`, which

This is the only way I can see how #5020 can come about - but the problem
solved in this PR should only come up very rarely - eg, typically only by
people disconnecting and reconnecting their sync account on any device.

So while I'm yet to be confident this is exactly #5020, I am confident
it is fixing a real issue.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -19,6 +19,11 @@ Use the template below to make assigning a version number during the release cut
 
 -->
 
+## Autofill
+
+### What's Fixed
+  - Fixed syncing of autofill when tombstones exist in the local mirror (#5030)
+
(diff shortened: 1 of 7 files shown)


lougeniaC64:
> The new tests, without the corresponding change in incoming.rs, fail with `Error: JsonError(Error("missing field `entry`", line: 0, column: 0))`, which
> 
nit: Wondering if there were further details you wanted to include here?

lougeniaC64 (approved):
LGTM - assuming we'll need to monitor the board where Ben saw these errors to be sure this is the fix for #5020 but otherwise 🚢

PR #5033, opened 2022-07-08
Error converting and reporting

mhammond:
This is built on top of @bendk's #5004, taking the concepts in that PR much further. Ben and I discussed this a little in-person, so this shouldn't be a surprise to him.

Like that branch, it splits the "internal" errors from the "external" errors returned to the consumers.

It extends that patch in the following ways:
* Instead of `Into()` to covert between the 2 error types and log/report as a side-effect, it uses an explicit conversion mechanism.
* That explicit conversion also specifies how each error should be reported and logged.
* Due to the above, there's a new `handle_error` macro, shamelessly stolen from an earlier iteration of Ben's patch, which all public functions must use. This converts and does the logging/reporting.
* It's still possible to convert between errors and/or get logging/reporting information without using that macro - so the error reporting and logging facilities are still "explicit" - it's just they are usually hidden behind that `handle_error` macro.

Open questions:
* Does this actually make sense?
* Is it an improvement and something we should take on for all components? I've kept a few commits, so it would be easy to keep this as a logins-only change if necessary, but I'm inclined to think it works OK.
* Does it work? :) Tests all pass etc, including Android tests (so I don't think this is actually a breaking change) but I'm yet to actually test it in a product.
* Assuming a positive reception, how should we land this? I think we should fold the 3 patches of mine into 1, leaving one commit from Ben and one from me we can land.

Feedback welcome. Bikesheds welcome (I'm relatively happy with the names, but welcome improvements). Happy to consider alternatives or other ideas.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -21,3 +21,8 @@ Use the template below to make assigning a version number during the release cut
 
 ### What's Changed
   - Added `MOZ_APPSERVICES_MODULE` environment variable to specify the megazord module for iOS ([#5042](https://github.com/mozilla/application-services/pull/5042)). If it is missing, no module is imported.
+
+## Logins
+
+### ⚠️ Breaking Changes ⚠️
+  - Updated the `LoginsStorageError` hierarchy.
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -24,5 +24,6 @@ Use the template below to make assigning a version number during the release cut
 
 ## Logins
 
-### ⚠️ Breaking Changes ⚠️
-  - Updated the `LoginsStorageError` hierarchy.
+  - Updated the `LoginsStorageError` implementation and introduce error reporting for unexpected errors.
+    Note that some errors were removed, which is technically a breaking change, but none of our
+    consumers use those errors so it's not a breaking change in practice.
--- a/components/logins/src/error.rs
+++ b/components/logins/src/error.rs
@@ -2,20 +2,23 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use error_support::report_error;
 use std::ffi::OsString;
 
-// TODO: this is (IMO) useful and was dropped from `failure`, consider moving it
-// into `error_support`.
-macro_rules! throw {
-    ($e:expr) => {
-        return Err(Into::into($e))
-    };
-}
+pub type Result<T> = std::result::Result<T, LoginsError>;
+pub type APIResult<T> = std::result::Result<T, LoginsStorageError>;
 
+/// Internal logins error type, this is what we use for inside this crate
 #[derive(Debug, thiserror::Error)]
-pub enum ErrorKind {
+pub enum LoginsError {
+    // WARNING: The #[error] attributes define the string representation of the error (see
+    // thiserror for details).  These strings should not contain any personally identifying
+    // information.  We operate on a best-effort basis, since we can't be completely sure that
+    // our dependencies don't leak PII in their error strings.  For example, `rusqlite::Error`
+    // could include data from a user's database in their errors, but we've never seen that in
+    // practice so we are comfortable forwading that error message in ours.
     #[error("Invalid login: {0}")]
-    InvalidLogin(InvalidLogin),
+    InvalidLogin(#[from] InvalidLogin),
 
     #[error("The `sync_status` column in DB has an illegal value: {0}")]
     BadSyncStatus(u8),
@@ -26,9 +29,6 @@ pub enum ErrorKind {
     // Fennec import only works on empty logins tables.
     #[error("The logins tables are not empty")]
     NonEmptyTable,
-
-    #[error("The provided salt is invalid")]
-    InvalidSalt,
 
     #[error("local encryption key not set")]
     EncryptionKeyMissing,
@@ -67,19 +67,7 @@ pub enum ErrorKind {
     MigrationError(String),
 }
 
-error_support::define_error! {
-    ErrorKind {
-        (SyncAdapterError, sync15::Error),
-        (JsonError, serde_json::Error),
-        (UrlParseError, url::ParseError),
-        (SqlError, rusqlite::Error),
-        (CryptoError, jwcrypto::JwCryptoError),
-        (InvalidLogin, InvalidLogin),
-        (Interrupted, interrupt_support::Interrupted),
-        (IOError, std::io::Error),
-    }
-}
-
+/// Error::InvalidLogin subtypes
 #[derive(Debug, thiserror::Error)]
 pub enum InvalidLogin {
     // EmptyOrigin error occurs when the login's origin field is empty.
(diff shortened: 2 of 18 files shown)


components/support/error/src/macros.rs
```diff
@@ -60,3 +60,19 @@ macro_rules! breadcrumb {
...
+macro_rules! handle_error {
+    { $($tt:tt)* } => {
+        let body = || {
+            $($tt)*
+        };
+        let result: Result<_> = body();
+        match result {
+            Ok(r) => Ok(r),
+            Err(e) => Err($crate::convert_log_report_error(e))
+        }
```
tarikeshaq, line 76: nit: we can use `map_err` here
```suggestion
        result.map_err($crate::convert_log_report_error)
```

components/support/error/src/handling.rs
```diff
@@ -0,0 +1,115 @@
...
+
+pub trait GetErrorHandling {
+    type ExternalError;
+
+    // Return how to handle our internal errors
+    fn get_error_handling(&self) -> ErrorHandling<Self::ExternalError>;
+
+    // Some helpers to cut the verbosity down.
+    // Just convert the error without any special logging or error reporting.
+    fn passthrough(err: Self::ExternalError) -> ErrorHandling<Self::ExternalError> {
```
tarikeshaq, line 39: Would implementers of the trait ever override this? _maybe_ can get rid of this whole trait and replace it with a bunch of `fn passthrough<E>(err: E) -> ErrorHandling<E>` etc functions and get rid of the trait?

i.e tell consumers to implement their own `impl<IE, EE> From<IE> for ErrorHandling<EE>` etc

Not sure how good of an idea that is, and this setup maybe better 🤷 but the thought crossed my mind so sharing it

tarikeshaq (commented):
Did a runthrough and dropped a few comments! My only concrete piece of feedback is that we probably should have an ADR or some type of document as we shift other components to follow the same pattern

I had a few high level thoughts that I'll try to expand on, but nothing blocking and I'd be happy to see this land as-is - I'll leave the r+ to Ben and others!

components/logins/src/error.rs
```diff
@@ -3,19 +3,49 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use std::ffi::OsString;
+pub type Result<T> = std::result::Result<T, LoginsError>;
+// Functions which are part of the public API should use this Result.
+pub type ApiResult<T> = std::result::Result<T, LoginsStorageError>;
```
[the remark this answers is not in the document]
mhammond: I'm not sure how that would work out in practice - IIUC, that would mean you'd need to convert from a sqlite error to this internal error at the place where the sqlite error might happen (ie, whereever you'd want to write `?` on a function returning a sqlite error). This would then make it tricky to handle specific errors in the crate in a sane way

eg, consider:

```
// Assume `Result` has our thin "public" error enum, so no explicit serde or sqlite errors
fn outer() -> Result<()> {
  middle()?;
}

fn middle() -> Result<()> {
  inner()?;
}

fn inner() -> Result<()> {
 do_sql()?;
 do_serde()?;
}
```

We need an `Into` our Err for the sql error and another for the serde error, and that conversion would happen at `inner()`. But what if `middle` actually *did* want to handle some sql or serde errors thrown by the inner functions? It now only has the error after transformation into the "Internal" representatation.

(Or maybe I minunderstand your suggestion or how it might be implemented?)

CHANGES_UNRELEASED.md
```diff
@@ -36,4 +36,6 @@ Use the template below to make assigning a version number during the release cut
 ## Logins
 
 ### What's Changed
+  - Updated the `LoginsStorageError` implementation and introduce error reporting for unexpected errors.
```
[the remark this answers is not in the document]
mhammond: Yes, I meant to clarify this - that note should say something like "technically a breaking change but in practice it's not"

components/support/error/src/handling.rs
```diff
@@ -0,0 +1,115 @@
...
+            report_class,
+            level,
+        } => {
+            log::log!(level, "{}", e.to_string());
+            // notify the error reporter if the feature is enabled.
+            // XXX - should we arrange for the `report_class` to have the
+            // original crate calling this as a prefix, or will we still be
+            // able to identify that?
+            #[cfg(feature = "reporting")]
+            crate::report_error(report_class, format!("{:?}", e));
```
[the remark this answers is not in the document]
mhammond: no, I agree it should be consistent.

components/support/error/src/handling.rs
mhammond: I was thinking some crates might want to override some of them maybe, but haven't really thought it through - eg, I could imagine one crate overriding `passthrough` to also log at a trace level or something like that - but then again, maybe that's not the right way to approach this. Maybe these "helpers" don't actually help and are just a potentially confusing extra level of indirection that should just be removed so things are more explicit and less magic?

It's a good thing for us to consider and I don't really have a strong sense of what's better here. We could also consider converting one or 2 other crates to reuse this mechanism too - fxaccounts would be a good candidate - to help get a better feel for it?

components/logins/src/error.rs
```diff
@@ -3,19 +3,49 @@
...
+    Interrupted(String),
 
-// TODO: this is (IMO) useful and was dropped from `failure`, consider moving it
-// into `error_support`.
-macro_rules! throw {
-    ($e:expr) => {
-        return Err(Into::into($e))
-    };
+    #[error("Unexpected Error: {0}")]
+    UnexpectedLoginsStorageError(String),
```
bendk, line 39: There's a couple errors that might be useful here in the short term: `SyncAuthError` and `RequestFailed`, which I think can be returned by the `sync()` method.  iOS is still using that method, and might want to do treat those error specially.  `SyncAuthError` means the app should tell the user they're not logged in.  `RequestFailed` is an expected error, so maybe that affects how iOS wants to log/report/track things.  I'm not really sure though, what do others think?

components/logins/src/error.rs
```diff
@@ -3,19 +3,49 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum LoginsStorageError {
+    #[error("Invalid login: {0}")]
+    InvalidRecord(String),
+
+    #[error("No record with guid exists (when one was required): {0:?}")]
+    NoSuchRecord(String),
+
+    #[error("Encryption key is in the correct format, but is not the correct key.")]
+    IncorrectKey,
```
bendk, line 33: Should keys with invalid formatting also be included with this one?  I think from the app's point of view both things require the same response.  If needed, there could be an inner enum to differentiate between the two cases.

components/logins/src/error.rs
```diff
@@ -97,193 +112,52 @@ pub enum InvalidLogin {
...
             }
+            // This list is partial - not clear if a best-practice should be to ask that every
+            // internal error is listed here (and remove this default branch) to ensure every error
+            // is considered, or whether this default is fine for obscure errors?
+            // But it's fine for now because errors were always converted with a default
+            // branch to "unexpected"
+            _ => Self::unexpected(
+                LoginsStorageError::UnexpectedLoginsStorageError(self.to_string()),
+                None,
+            ),
```
bendk, line 160: I like the mechanics of how this works, but it seems like it could be simplified some.  What do you think of these suggestions:

 - Splitting up this method into `get_error_reporting()` and `into_external_error()`.  I think that would clarify what's getting reported and what's getting passed on to the consumer app.
  - Defining `report()`, `unexpected()`, etc. as methods of `ErrorHandling` (or `ErrorReporting` if we make the previous change).  This feels more natural to me, since those methods are just creating instances of those types.

components/support/error/src/handling.rs
bendk: I personally want to go the other way and shave off a level of indirection.  If all these functions are doing is providing a convenient way to construct an `ErrorHandling` instance, then they should be in `ErrorHandling` rather than the trait.

components/support/error/src/handling.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// No logging or error reporting.
+    Nothing,
+    /// We write a log message but don't report it.
+    Log { level: log::Level },
+    /// We log a message and report via our error reporter.
+    Report {
+        level: log::Level,
+        report_class: String,
+    },
+}
```
bendk, line 18: I like how flexible this is.

You could go a step further and have a struct of options:
```
pub struct ErrorReporting {
  log: Option<log::Level>,
  report: Option<String>,
  count_in_telemetry: Option<String>, # .. eventually
}
```

I don't have a preference between the two options.  I mostly just wanted to through this out as a possibility.

bendk (commented):
Looks very nice to me.  I have a couple bikeshedding suggestions plus some questions about the error hierarchy, but I think the overall direction is great.

components/logins/src/error.rs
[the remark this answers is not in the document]
tarikeshaq: Yeah looks like the lost context is somewhat a killer to what I was suggesting :( 

> There does seem to be a fundamental split at the API boundary. The consumer code and the code in the crate care about very different things when it comes to errors.

+1! I can see that and I concede here, I think it'll be worth it to have both (or at the very least worth seeing how Logins shapes up) - it would also be nice to have a small document going over the reasoning for record-keeping!

components/logins/src/error.rs
mhammond: I did go around a few circles here and ultimately decided that InvalidFormat isn't really something the apps will need to handle tbh - it really just means "invalid json", and we expose methods to create the key. So the only time this could happen is if they decided to try and create the key themselves (which is just misuse) or if somehow the key was partially corrupted in their store (eg, trailing `}` lost or similar).

So this doesn't really seem like something apps would ever catch, and I don't think they do now. Or to put it another way, using error reporting to know this *does not* happen has value in its own right.

components/logins/src/error.rs
mhammond: > Splitting up this method into get_error_reporting() and into_external_error(). I think that would clarify what's getting reported and what's getting passed on to the consumer app.

My concern with this is that it I fear it might be error prone - logins would then have 2 different functions which match every "internal" error - so it seems like it would be easier to accidentally have some errors (say) returning an InternalError, but not logging it, because you made an assumption about what the "other" match is going some 50 lines away. The way things stand, now there's a single place to get the whole context for a single error.

IOW, I thought the single match was quite desirable. I don't feel too strongly about this though, so welcome other thoughts. 

> Defining report(), unexpected(), etc. as methods of ErrorHandling

Yeah, this does make more sense, thanks.

components/support/error/src/handling.rs
mhammond: I actually like this better, thanks, especially given your `count_in_telemetry` example - the enum approach would have meant a new variant for every *combination* of actions to take.

mhammond:
Thanks for the comments - I'm pushing a new version with most of the suggestions as individual commits, so it's easy to back it out if we think it's not an improvement (but I think they all are). I haven't tried to split the reporting and error conversions out, but would love more discussion on that (or if you are super keen, try it out!)

components/logins/src/error.rs
bendk: > My concern with this is that it I fear it might be error prone - logins would then have 2 different functions which match every "internal" error - so it seems like it would be easier to accidentally have some errors (say) returning an InternalError, but not logging it, because you made an assumption about what the "other" match is going some 50 lines away. The way things stand, now there's a single place to get the whole context for a single error.

That makes sense to me.  Maybe there's a way to keep everything in 1 match and also clarify what's happening.  Now that the `ErrorReporting` struct is using options, [what about a builder-style API to create `ErrorHandler` instances?](https://github.com/mozilla/application-services/commit/96767f5d543e87687bae55f9a09191e6b65cdf24)

components/logins/src/error.rs
```diff
@@ -97,193 +121,70 @@ pub enum InvalidLogin {
...
+                        log::Level::Warn,
+                    )
+                }
+                Sync15ErrorKind::RequestError(_) => ErrorHandling::log(
+                    LoginsStorageError::RequestFailed(e.to_string()),
+                    log::Level::Warn,
+                ),
+                _ => ErrorHandling::unexpected(
+                    LoginsStorageError::UnexpectedLoginsStorageError(self.to_string()),
+                    Some("sync"),
```
bendk, line 176: Should this and other error classification strings be prefixed with `logins-`?  That way when we add more components, each gets its own sync error classification?

mhammond:
Thanks for all the feedback here - I'm going to land it and we can keep iterating.

PR #5034, opened 2022-07-08
Fix the dep checker: the windows-rs repository moved its license files

mhammond:
My recent PR is failing because just an hour or so ago, the license files in the windows-rs repo moved.


tarikeshaq (approved):
Thank you!

PR #5046, opened 2022-07-15
delete_visits_between now deletes history metadata too

tarikeshaq:
Part of https://mozilla-hub.atlassian.net/browse/SYNC-3229, second part being exposing this API to iOS (which will happen in https://github.com/mozilla/application-services/pull/4989

Changes `delete_visits_between` to now also delete history metadata.

The deletion of the history metadata is *not* dependent on the deletion of the visits - want feedback on that, the idea is that 

- it's OK if we want to delete history metadata without there being any visits (the iOS case)
- If the whole places entry is deleted with the visit, we are OK since deletes cascade
- I thought it should be OK if we delete a history metadata entry, while the associated visit is still alive (given that the history metadata was recorded within the requested deletion frame)

I'm still getting familiar with this p

  - Note:


mhammond (approved):
Looks great! Should we get a followup so all the `delete` functions also delete metadata? I can't see a good reason for some to do so and others not doing it.

Issue #5011, opened 2022-06-23
Add pruning to places `run_maintenance()`


bendk:
Places has a `run_maintenance()` function, but it doesn't currently do all [the things that desktop does](https://searchfox.org/mozilla-central/rev/a55ac39a6e01c342d3befacc875a6e2959f6111f/toolkit/components/places/nsPlacesExpiration.js#171).  We should consider adding more functionality to it.
 
I think the starting point should be deleting older visits when over the page limit (see  `QUERY_FIND_EXOTIC_VISITS_TO_EXPIRE` and `QUERY_FIND_VISITS_TO_EXPIRE`).  I'm not sure how to set a page limit, maybe the app can pass in a size limit and we can calculate a page limit based on that.
 
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3216)
┆sprintEndDate: 2022-07-22

mhammond:
There's also https://searchfox.org/mozilla-central/source/toolkit/components/places/PlacesExpiration.jsm, which is probably what we ultimately want to do to stop the db growing without bounds. I haven't really thought too much about how much of that we should do or when though, but maybe getting the core infrastructure in place and one of the expiration strategies would be a good start? Eg, "Finds visits to be expired when history is over the unique pages limit" seems useful, and/or the `DATABASE_DISK_CONSTRAINED_THRESHOLD` handling?

bendk:
(Updated the description now that I know more about what's going on in desktop).

Another thing that desktop does is delete orphaned pages.  But I think we have [some code that already does this](https://github.com/mozilla/application-services/blob/main/components/places/src/storage/history.rs#L670-L696), that executes right when the pages are deleted from the app.  Is my reading correct?  Is there any value in still trying to delete orphans in case that code missed them somehow?

PR #5057, opened 2022-07-27
Places expiration (#5011)

bendk:
Added code to expire older visits if the places DB is over some size limit.  The easy part was calculating the size limit and figuring out which visits should be deleted.  The hard part was figuring out what steps to take to delete those visits and testing that I did the correct steps.  To help with that I did some refactoring that I think improves things, but I'm interested to hear what you think about it.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -25,6 +25,7 @@ Use the template below to make assigning a version number during the release cut
 ## Places
 ### What's new
   - We now expose all of the Places history APIs to Swift consumers. ([#4989](https://github.com/mozilla/application-services/pull/4989))
+  - Added an optional db_size_limit parameter to `run_maintenance`.  This can be used to set a target size for the places DB.  If the DB is over that size, we'll prune a few older visits. The number of visits is very small (6) to ensure that the operation only blocks the database for a short time. The intention is that `run_maintenance()` is called frequently compared to how often visits are added to the places DB.
 
 ## Sync15
 ### What's changed
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -24,6 +24,8 @@ use rusqlite::types::ToSql;
 use rusqlite::Result as RusqliteResult;
 use rusqlite::Row;
 use sql_support::{self, ConnExt};
+use std::collections::HashSet;
+use std::time::Duration;
 use sync15::EngineSyncAssociation;
 use sync_guid::Guid as SyncGuid;
 use types::Timestamp;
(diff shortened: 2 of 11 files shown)


components/places/src/storage/history/mutations.rs
```diff
@@ -0,0 +1,491 @@
...
+    let mut page_ids = HashSet::<RowId>::new();
+    for visit_to_delete in visits_to_delete.into_iter() {
+        visit_ids.insert(visit_to_delete.visit_id);
+        page_ids.insert(visit_to_delete.page_id);
+    }
+    vec![
+        DbMutation::DeleteVisitRows { visit_ids },
+        DbMutation::RecalcPages { page_ids },
+        DbMutation::DeleteFromPendingTempTables,
+    ]
```
bendk, line 154: I'm pretty sure these are the correct steps when deleting visits from the DB, but please double check that I'm not missing anything.

components/places/src/storage/history.rs
```diff
@@ -483,65 +578,24 @@ pub fn delete_everything(db: &PlacesDb) -> Result<()> {
...
-
-    db.conn().execute_cached(
-        "DELETE FROM moz_historyvisits
-         WHERE visit_date = :visit_date
-           AND place_id = :place_id",
-        &[
-            (":place_id", &place_id as &dyn rusqlite::ToSql),
-            (":visit_date", &visit_date),
-        ],
-    )?;
```
bendk, line 527: Note that there's a subtle difference between the current code and the new code if 2 visits have the same `place_id` and `visit_date`.

The current code deletes all visits with the same `visit_date`, but the new code only deletes 1 timestamp.  This seems more correct to me, since we're only inserting a single tombstone, but I wasn't really sure how that case should be handled.

Another option that makes sense to me would be to drop the `LIMIT 1` part from the query.  That way if there's multiple visits with the same `visit_date`, we'll delete all rows and insert that many rows into the tombstones table.

It's hard to say what's correct, especially since this corner case seems almost impossible in practice.

components/places/src/storage/history/mutations.rs
```diff
@@ -0,0 +1,491 @@
...
+            page_id: row.get("place_id")?,
+        })
+    }
+}
+
+/// Create a Vec of mutations to delete visits returned by a query
+///
+/// The query must include the id and place_id columns from moz_historyvisits.
+pub(super) fn delete_visit_mutations(
+    visits_to_delete: impl IntoIterator<Item = VisitToDelete>,
```
mhammond, line 142: This is only ever called with a `Vec`, right?

components/places/src/storage/history/mutations.rs
mhammond: That looks correct to me!

components/places/src/storage/mod.rs
```diff
@@ -218,12 +220,22 @@ impl TopFrecentSiteInfo {
     }
 }
 
-pub fn run_maintenance(conn: &PlacesDb) -> Result<()> {
+/// Run various maintanance on the places DB
```
mhammond, line 223: typo: maintenance

components/places/src/storage/mod.rs
```diff
@@ -218,12 +220,22 @@ impl TopFrecentSiteInfo {
...
+///
+/// This function is intended to be run during idle time and will take steps to clean up / shrink
+/// the databas.
```
mhammond, line 226: typo: database

components/support/sql/src/conn_ext.rs
```diff
@@ -223,6 +223,15 @@ pub trait ConnExt {
...
     }
+
+    /// Get the approximate DB size in bytes
+    fn get_db_size(&self) -> Result<u32, rusqlite::Error> {
+        let page_count: u32 = self.query_one("SELECT * from pragma_page_count()")?;
+        let page_size: u32 = self.query_one("SELECT * from pragma_page_size()")?;
+        let freelist_count: u32 = self.query_one("SELECT * from pragma_freelist_count()")?;
+
+        Ok((page_count - freelist_count) * page_size)
+    }
```
mhammond, line 234: for interest, have you sanity checked how closely this matches the actual size on disk? I guess I'm really wondering if we should add more caveats to the docs.

Also wondering about the freelist - that still takes space on the disk, right? If this process just ends up creating a larger freelist without actually freeing up disk space, I wonder how useful it turns out being? Or do we assume the vacuum cleans that up?

components/places/src/storage/mod.rs
```diff
@@ -218,12 +220,22 @@ impl TopFrecentSiteInfo {
...
+///
+/// db_size_limit is the approximate storage limit in bytes.  If the database is using more space
+/// than this, some older visits will be deleted to free up space.  Pass in a 0 to skip this.
+pub fn run_maintenance(conn: &PlacesDb, db_size_limit: u32) -> Result<()> {
     conn.execute_all(&[
         "VACUUM",
         "PRAGMA optimize",
         "PRAGMA wal_checkpoint(PASSIVE)",
     ])?;
+    if db_size_limit > 0 && conn.get_db_size()? > db_size_limit {
```
mhammond, line 236: shouldn't we vacuum after pruning?

mhammond (commented):
This looks great - feel free to push back or open follow-ups for my suggestions, but I wouldn't mind another look.

CHANGES_UNRELEASED.md
```diff
@@ -20,5 +20,12 @@ Use the template below to make assigning a version number during the release cut
 -->
 
 ## Places
+### ⚠️ Breaking Changes ⚠️
+  - Added a size_limit parameter to `run_maintanance`.  This can be used to set a target size for
```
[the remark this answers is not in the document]
bendk: I just looked around and desktop uses 75 MiB, then reduces that to 32.5 MiB if it determines there's low disk or memory space.  As a first pass I added comments explaining that.  Maybe we should just copy that behavior and let consumers pass in an enum with the following values: 
  - `Prune::Normal` (75 MiB)
  - `Prune:Aggresive` (32.5 MiB)
  - `Prune::Disable` (the default, to make this a non-breaking change)

components/places/src/storage/history/mutations.rs
```diff
@@ -0,0 +1,491 @@
...
+            visit_id: row.get("id")?,
+            page_id: row.get("place_id")?,
+        })
+    }
+}
+
+/// Create a Vec of mutations to delete visits returned by a query
+///
+/// The query must include the id and place_id columns from moz_historyvisits.
+pub(super) fn delete_visit_mutations(
```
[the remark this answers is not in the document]
bendk: Thanks for the naming suggestions.  I made both changes, renaming "mutations" to "actions" and also updating the function name.

components/places/src/storage/history/mutations.rs
bendk: Good point.  I had hopes of creating one from an `rusqlite::MappedRows`, but that didn't pan out.

components/support/sql/src/conn_ext.rs
bendk: Astonishingly well.  I tested the moz-central test data in `toolkit/components/passwordmgr/test/unit/data/` and `toolkit/components/places/tests/migration/`.  The result of that calculation exactly matched the output of `du -b`, for all files except the corrupt one.

I don't think those files had any pages in the free list, so I ran this test to check that:
  - Created a database file, inserted a bunch of junk data, then deleted half of it
  - Checked the output of `du -b` against the query calculation.  The query calculation was almost exactly half of `du -b`
  - Ran vaccum and checked again.  This time the query calculation and `du -b` matched exactly.

components/places/src/storage/history/mutations.rs
```diff
@@ -0,0 +1,491 @@
...
+impl DbMutation {
+    pub(super) fn execute(self, db: &PlacesDb) -> Result<()> {
+        match self {
+            Self::DeleteVisitRows { visit_ids } => Self::delete_visit_rows(db, visit_ids),
+            Self::RecalcPages { page_ids } => Self::recalc_pages(db, page_ids),
+            Self::DeleteFromPendingTempTables => Self::delete_from_pending_temp_tables(db),
+        }
+    }
+
+    pub(super) fn execute_all(db: &PlacesDb, mutations: Vec<Self>) -> Result<()> {
```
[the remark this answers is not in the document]
bendk: Initially I was thinking that consumers would construct them directly.  But then I realized that for deleting visits, it was simpler create a `Vec<VisitToDelete>` as an intermediate step.  The main reason was that it simplified combining the exotic visits and the regular visits.  I'm not sure we always want to do this though.

How about creating a follow-up bug to refactor some other operations to use this system and figure out what the right API based on that?

bendk:
> This looks great - feel free to push back or open follow-ups for my suggestions, but I wouldn't mind another look.

Your suggestions were great.  I pushed some changes based on those, but I would love another look and left a couple questions for you there.

CHANGES_UNRELEASED.md
```diff
@@ -25,6 +25,7 @@ Use the template below to make assigning a version number during the release cut
 ## Places
 ### What's new
   - We now expose all of the Places history APIs to Swift consumers. ([#4989](https://github.com/mozilla/application-services/pull/4989))
+  - Added an optional db_size_limit parameter to `run_maintanance`.  This can be used to set a target size for the places DB.  If the DB is over that size, we'll prune a few older visits. The number of visits is very small (6) to ensure that the operation only blocks the database for a short time. The intention is that `run_maintanaince()` is called frequently compared to how often visits are added to the places DB.
```
mhammond, line 28: lol - an even more creative spelling of "maintenance" here now :) It's spelt incorrectly twice here now :) Ditto below, so maybe a search-replace for "maintanance"?

components/support/sql/src/conn_ext.rs
mhammond: Fantastic!

components/places/src/storage/history/mutations.rs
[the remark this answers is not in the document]
mhammond: A followup sounds fine!

mhammond (approved):
This is awesome, thanks!

Issue #5061, opened 2022-07-28
Follow up on logins error work


bendk:
We landed the changes to logins error handling a few weeks ago and those changes hit beta at the beginning of this week.  It seems like this is a good time to follow up on those changes based on the sentry data.  Here's what I've found based on a quick look at the Fenix Beta data for the last 24 hours:

- I only found [1 report](https://sentry.io/organizations/mozilla/discover/fenix-beta:0d71a77a67ef4adbb95733d1ca16a978/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=6295551&query=%22logins%3Aunexpected%22&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29) that was generated with our new error handling code.
  - It's kind of surprising to me that we have so few errors, but I think that's accurate.  When I search for `UnexpectedLoginsStorageError` which should match the older errors I see a very small amount, rarely more than 1 per day.
  - That `target database already exists` error might be worth taking a look at if we see more reports, but since there's only 1I don't think it's worth it to address at this point.
- We should consider changing the separator char from `:` to something else like `-`.  Right now you always need to quote the string when searching, or else Sentry will interperate `:` as trying to match against a tag.  Also, there's no great way to search for all the newer-style matches.  I tried searching for `logins:`, but that matches errors with a module path in it.
- When I search for "logins", I see many reports for 2 kinds of errors and made issues for them: #5059, #5060.  I think we should try to address those issues and also update the code to call `report_error!` rather than `log::error!`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3244)
┆sprintEndDate: 2022-08-05

PR #5062, opened 2022-07-29
Logins errors phase2 (#5061)

bendk:
Various small improvements to the logins error tracking after seeing the results of the initial round of changes.

  - Note:

### DIFF
--- a/components/logins/src/error.rs
+++ b/components/logins/src/error.rs
@@ -7,7 +7,7 @@ pub type Result<T> = std::result::Result<T, LoginsError>;
 // Functions which are part of the public API should use this Result.
 pub type ApiResult<T> = std::result::Result<T, LoginsStorageError>;
 
-pub use error_support::handle_error;
+pub use error_support::{breadcrumb, handle_error, report_error};
 use error_support::{ErrorHandling, GetErrorHandling};
 use sync15::ErrorKind as Sync15ErrorKind;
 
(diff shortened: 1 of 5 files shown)


components/logins/src/login.rs
```diff
@@ -336,7 +336,8 @@ impl LoginFields {
                 }
                 Ok(None)
             }
-            Err(_) => {
+            Err(e) => {
+                breadcrumb!("Error parsing login origin: {e:?}");
```
bendk, line 340: I'm pretty sure that logging an [url::ParseError](https://docs.rs/url/latest/url/enum.ParseError.html) shouldn't leak any PII.

bendk:
I'm not sure what's going on with the Circle CI checks, but I don't think it could be related to these changes.

components/logins/src/error.rs
```diff
@@ -152,7 +152,7 @@ impl GetErrorHandling for LoginsError {
                 ErrorHandling::convert(LoginsStorageError::UnexpectedLoginsStorageError(
                     "must be an empty DB to migrate".to_string(),
                 ))
-                .report_error("logins:migration")
```
lougeniaC64, line 155: This doesn't prevent me from approving, but why were these changes from `:` to `-` were made?

components/logins/src/login.rs
lougeniaC64: Yeah IIUC this will only pass in the `ParseError` enum variant as a string so no other details will be passed. I'm going to approve with that assumption. If we find otherwise it'll be easy enough to revert this change.

lougeniaC64 (approved):

components/logins/src/error.rs
bendk: Sentry ergonomics: If you type ":" into the Sentry search box, it interprets that as filtering based on a field which means you need to quote the strings in order to search for them.

lougeniaC64: Makes sense--thanks!

PR #5073, opened 2022-08-02
Update various gradle libraries to newer releases.

rvandermeulen:
In general, the goal was to maintain parity with the versions in use by Android-Components. I also updated to the latest version of a few that aren't used by AC.


rvandermeulen:
OK, I'm going to pare this one back a bit. Bumping gradle and some of its dependencies also brings along the requirement to bump the target API version from 29 to 31. While that's probably a good idea to do, it would be better to tackle that in its own PR rather than as a ride-along change in this one.

lougeniaC64 (dismissed):

skhamis (approved):
Thanks for making these changes, CI looks good and the changes make sense (parity with a-c) LGTM! 🚀

PR #5076, opened 2022-08-03
Update gradle to version 7.3.3 and set target SDK version to 32.

rvandermeulen:
Maintaining the same versions used with Android Components.


skhamis (approved):
Awesome changes! LGTM!

PR #5077, opened 2022-08-03
Adds ios history migration and example

tarikeshaq:
While running some performance tests for places on iOS, I realized that the best way to keep the data consistent between the databases is to move the same data from one to the other... that is pretty much what we need for the migration work so I decided I'll push it up.

TODOs:
- [x] Add tests in places-tests to test that the migration logic works beyond the testing against the simulator db I've been doing

I'll be using this + a swift script to populate the BrowserDB for performance testing. (bonus points it'll also be performance testing of the migration)


components/places/src/ffi.rs
```diff
@@ -168,20 +170,25 @@ impl PlacesApi {
...
-        let metrics = import_bookmarks(self, db_path.as_str())?;
+        let metrics = import_fennec_bookmarks(self, db_path.as_str())?;
         Ok(serde_json::to_string(&metrics)?)
     }
 
     fn places_bookmarks_import_from_ios(&self, db_path: String) -> Result<()> {
-        import_bookmarks(self, db_path.as_str())?;
+        import_ios_bookmarks(self, db_path.as_str())?;
         Ok(())
     }
```
tarikeshaq, line 185: tbh I think I'll kill the other migration functions and logic, didn't want to do it in the same PR, but maybe now is as good time as ever

components/places/src/ffi.rs
```diff
@@ -168,20 +170,25 @@ impl PlacesApi {
...
 
     fn places_bookmarks_import_from_fennec(&self, db_path: String) -> Result<String> {
-        let metrics = import_bookmarks(self, db_path.as_str())?;
+        let metrics = import_fennec_bookmarks(self, db_path.as_str())?;
         Ok(serde_json::to_string(&metrics)?)
     }
 
     fn places_bookmarks_import_from_ios(&self, db_path: String) -> Result<()> {
-        import_bookmarks(self, db_path.as_str())?;
```
tarikeshaq, line 181: heh this function was wildly wrong. it called the `import_fennec_bookmarks` function to migration iOS bookmarks 😓 I fixed it here, but might just kill it

components/places/src/import/ios/history.rs
```diff
@@ -0,0 +1,137 @@
...
+            url TEXT,
+            url_hash INTEGER NOT NULL,
+            title TEXT
+        ) WITHOUT ROWID;";
+
+   static ref FILL_STAGING: &'static str = "
+    INSERT OR IGNORE INTO temp.iOSHistoryStaging(id, guid, url, url_hash, title)
+        SELECT
+            h.id,
+            sanitize_utf8(guid), -- The places record in our DB may be different, but we
```
tarikeshaq, line 94: note to self, this is copy-pasta artifiact, we don't need the guid here - iOS uses an integer primary key (id) for its history records and visits point to that

components/places/src/import/ios/history.rs
```diff
@@ -0,0 +1,137 @@
...
+            -1,
+            1
+        FROM temp.iOSHistoryStaging t"
+   ;
+
+   // Insert history visits
+   static ref INSERT_HISTORY_VISITS: &'static str =
+   "INSERT OR IGNORE INTO main.moz_historyvisits(from_visit, place_id, visit_date, visit_type, is_local)
+        SELECT
+            NULL, -- Fenec does not store enough information to rebuild redirect chains.
```
tarikeshaq, line 123:
```suggestion
            NULL, -- iOS does not store enough information to rebuild redirect chains.
```

components/places/src/ffi.rs
lougeniaC64: 🥴 Great. Well . . . an attempt was made to remove this logic in iOS two or more years ago and was only restored to make tests pass according to [this PR](https://github.com/mozilla-mobile/firefox-ios/pull/6393). So seems like this can die.

components/places/src/import/ios/bookmarks.rs
```diff
@@ -176,12 +176,12 @@ fn populate_mirror_tags(db: &crate::PlacesDb) -> Result<()> {
     {
         let mut stmt = db.prepare(
             "SELECT mirror.id, stage.tags
-             FROM main.moz_bookmarks_synced mirror
-             JOIN temp.iosBookmarksStaging stage USING(guid)
-             -- iOS tags are JSON arrays of strings (or null).
-             -- Both [] and null are allowed for 'no tags'
-             WHERE stage.tags IS NOT NULL
-               AND stage.tags != '[]'",
+              FROM main.moz_bookmarks_synced mirror
```
mhammond, line 179: The whitespace changes here and below also look accidental? (If they aren't, I don't think they are adding enough value tbh, as it's not immediately clear if you are actually changing anything - I think not, but didn't read carefully enough to know for sure)

components/places/src/import/ios/history.rs
```diff
@@ -0,0 +1,151 @@
...
+pub fn import(
+    places_api: &PlacesApi,
+    path: impl AsRef<std::path::Path>,
+) -> Result<HistoryMigrationResult> {
+    let url = crate::util::ensure_url_path(path)?;
+    do_import(places_api, url)
+}
+
+fn do_import(places_api: &PlacesApi, ios_db_file_url: Url) -> Result<HistoryMigrationResult> {
+    let conn_mutex = places_api.get_sync_connection()?;
```
mhammond, line 41: If we end up trying to migrate the "live" database this way, we might end up with any other "write" operations failing with SQLITE_BUSY - attempts to create bookmarks, for example. That *might* mean we would be better off using the normal writer connection, so we just block the attempt to create the bookmark until the import is complete. We probably need think more about this as we refine the strategy we end up with.

components/places/src/import/ios/bookmarks.rs
```diff
@@ -228,9 +228,9 @@ fn populate_mirror_tags(db: &crate::PlacesDb) -> Result<()> {
         for item_id in tagged_items {
             log::trace!("tagging {} with {}", item_id, tag);
             db.execute_cached(
-                "INSERT INTO main.moz_bookmarks_synced_tag_relation(itemId, tagId) VALUES(:item_id, :tag_id)",
-                named_params! { ":tag_id": tag_id, ":item_id": item_id },
-            )?;
+                 "INSERT INTO main.moz_bookmarks_synced_tag_relation(itemId, tagId) VALUES(:item_id, :tag_id)",
```
mhammond, line 231: (aside: I'm actually quite surprised `cargo fmt` didn't get upset at this specific change, but it seems to be OK with either)

mhammond (commented):
This is looking great!

components/places/src/ffi.rs
```diff
@@ -217,20 +220,25 @@ impl PlacesApi {
     }
 
     fn places_history_import_from_fennec(&self, db_path: String) -> Result<String> {
```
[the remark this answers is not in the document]
tarikeshaq: Yeah looks like they haven't removed the function calls to our component in a-c yet :( https://github.com/mozilla-mobile/android-components/blob/2026625c38d414fc38997ebb13932d0cf5f2bdc2/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt#L265-L274

I have another branch up to remove the migration code, but I'll make a ticket first to kill those functions in a-c

components/places/src/import/ios/bookmarks.rs
tarikeshaq: Heh I didn't touch those queries, and in fact I could swear that `cargo fmt` introduced those changes once I changed something else in the file 👀

lougeniaC64 (commented):
Other than the removal of the sql formatting changes in `ios/bookmarks.rs`, this looks good to me. A lot of the concerns I have are around the duration of the migration and how often it gets interrupted by app crashes and closures, or the app being placed in the background. But all of these concerns will be clarified by dry run data.

components/places/src/ffi.rs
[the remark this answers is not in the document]
mhammond: thank you!

mhammond (approved):
This looks great to me, but what I don't quite understand is how this is actually a "dry run"? Is it just that we aren't actually switching iOS over to this data? What's the plan for doing the "real" run later?

Another question re the metrics: I guess the iOS PR for this, which is also going to record the metrics, will want to make sure it successfully recorded the migration starting before calling this? My vague concern is that we don't want miss any migrations which started but which got killed before we completed.

But r=me for this migration code and we can discuss the final process for doing it in our meeting!

tarikeshaq:
> but what I don't quite understand is how this is actually a "dry run"?

It'll be the app that runs this as a "dry run", the app will construct a separate instance of `RustPlaces` with a path to a fake database and run this function on that.

The idea is that what we are shipping here can be both the dry run or the real based on remote configuration set by nimbus

I'll demonstrate more in the meeting, but here's the iOS side of the story: https://github.com/mozilla-mobile/firefox-ios/pull/11896

Issue #4877, opened 2022-03-14
Analyze our sqlite pragma usage for potential improvements/best practices


lougeniaC64:
Can we create a set of standard [sqlite pragmas](https://www.sqlite.org/pragma.html) for our components that could be set in `sql-support` by default based on our current pragma usage (see [here](https://github.com/mozilla/application-services/issues/4872#issuecomment-1066813381) for context)? Or can we document best practices around pragmas? 

This ticket is a spike to answer those questions which could result in additional tickets to perform any resulting optimizations.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3108)

mhammond:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1866354

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #5081, opened 2022-08-05
Update comment about the places sqlite page_size

mhammond:
Came across [this desktop places bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1782283) which shows a measurable perf win with a large page_size. I wonder if we should roll this into open_database so all our databases get it? Places is the most important, and while the others are smaller, it might still be a win? Is that worth an issue? :)


mhammond:
Related to #4877

bendk (approved):
The comment change seems great.

I'm torn on changing the page size for all components.  My guess is that it's probably an improvement, but it's very hard to know.  Maybe we should wait until we get better Glean and/or Nimbus integration and we can try to do some testing/measuring.

PR #5083, opened 2022-08-08
Move the bso_record and key_bundle into the sync15_traits crate.

### LINKED ISSUE #2712: Make ttl and sortindex handling sane
ttl and sortindex are handled in a non-obvious and hacky way. History sync does this, and we are copying that for the clients record. We should fix it - but it's not obvious how - it probably needs a bit of a rethink about a payload vs a bso_record etc.

https://github.com/mozilla/application-services/blob/04f2e0ee8ba94fc27509078381102a7b0f793ab6/components/places/src/history_sync/record.rs#L32

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2452)
┆Epic: Sync Ecosystem (backlog)

mhammond:
(Which further reinforces that the name of that crate should be something like
`sync15-types`)

This is preparatory work for a much larger patch that will get us further down
the path to #2841 and also fix #2712. It should not change any behaviour.

As part of this, the sync15_traits errors needs to be addressed. It's not
reasonable to just create another enum that we expect clients to add to
their error types because many of the errors overlap which makes handling
them very difficult (eg, if we want to catch a serde error, you don't want
to know which crate it originated in) - so sync15 duplicates the sync15_traits
errors and can convert between them - so consumers can continue to catch
sync15::Error and still get errors which originated in sync15-traits.

sync15-traits now has a `crypto` feature - sync15 enables that feature, but
crates which just use a `bridged_engine` do not. This means that when
vendoring into Desktop Firefox, this feature isn't enabled, so we don't
pull in any of the crypto libs there.

@tarikeshaq and @bendk, I'd welcome all feedback here, particularly around the error changes. They make sense to me, but I might be missing something! Note that I'll probably hold off merging this until that larger patch is up for initial review.

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/error.rs
+++ b/components/fxa-client/src/internal/error.rs
@@ -148,6 +148,19 @@ error_support::define_error_conversions! {
     }
 }
 
+// Adapts errors from the sync15_traits crate into sync15 errors.
+impl From<sync15::SyncTraitsError> for ErrorKind {
+    fn from(e: sync15::SyncTraitsError) -> Self {
+        ErrorKind::SyncError(e.into())
+    }
+}
+
+impl From<sync15::SyncTraitsError> for Error {
+    fn from(e: sync15::SyncTraitsError) -> Self {
+        Error::from(ErrorKind::from(e))
+    }
+}
+
 // The public FFI puts the errors into three buckets, this helps us
 // convert between them. Maybe in future we can use uniffi to expose
 // more error info to the caller?
--- a/components/support/sync15-traits/src/error.rs
+++ b/components/support/sync15-traits/src/error.rs
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// `Result` is unused with default features, but used with `feature=crypto`
+#[cfg(feature = "crypto")]
+pub type Result<T> = std::result::Result<T, SyncTraitsError>;
+
+// Concrete errors returned by this module. Note that the sync15 crate has
+// the ability to turn these errors into a `sync15::Error`, so most consumers
+// should leverage that rather than create an error variant specific to this
+// crate.
+#[derive(Debug, thiserror::Error)]
+pub enum SyncTraitsError {
+    #[cfg(feature = "crypto")]
+    #[error("Key {0} had wrong length, got {1}, expected {2}")]
+    BadKeyLength(&'static str, usize, usize),
+
+    #[cfg(feature = "crypto")]
+    #[error("SHA256 HMAC Mismatch error")]
+    HmacMismatch,
+
+    #[cfg(feature = "crypto")]
+    #[error("Crypto/NSS error: {0}")]
+    CryptoError(#[from] rc_crypto::Error),
+
+    #[cfg(feature = "crypto")]
+    #[error("Base64 decode error: {0}")]
+    Base64Decode(#[from] base64::DecodeError),
+
+    #[error("JSON error: {0}")]
+    JsonError(#[from] serde_json::Error),
+
+    #[error("Bad cleartext UTF8: {0}")]
+    BadCleartextUtf8(#[from] std::string::FromUtf8Error),
+
+    #[cfg(feature = "crypto")]
+    #[error("HAWK error: {0}")]
+    HawkError(#[from] rc_crypto::hawk::Error),
+}
--- a/components/sync15/src/error.rs
+++ b/components/sync15/src/error.rs
@@ -3,10 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use interrupt_support::Interrupted;
-use rc_crypto::hawk;
-use std::string;
 use std::time::SystemTime;
-use sync15_traits::request::UnacceptableBaseUrl;
+use sync15_traits::{request::UnacceptableBaseUrl, SyncTraitsError};
 /// This enum is to discriminate `StorageHttpError`, and not used as an error.
 #[derive(Debug, Clone)]
 pub enum ErrorResponse {
@@ -23,12 +21,33 @@ pub enum ErrorResponse {
 
 #[derive(Debug, thiserror::Error)]
 pub enum ErrorKind {
+    // These are errors duplicated from SyncTraitsError, so that consumers can
+    // deal with errors from just 1 of the crates and not care which one of
+    // then actually caused the error.
     #[error("Key {0} had wrong length, got {1}, expected {2}")]
     BadKeyLength(&'static str, usize, usize),
 
     #[error("SHA256 HMAC Mismatch error")]
     HmacMismatch,
 
+    #[error("Crypto/NSS error: {0}")]
+    CryptoError(#[from] rc_crypto::Error),
+
+    #[error("Base64 decode error: {0}")]
+    Base64Decode(#[from] base64::DecodeError),
+
+    #[error("JSON error: {0}")]
+    JsonError(#[from] serde_json::Error),
+
+    #[error("Bad cleartext UTF8: {0}")]
+    BadCleartextUtf8(#[from] std::string::FromUtf8Error),
+
+    #[error("HAWK error: {0}")]
+    HawkError(#[from] rc_crypto::hawk::Error),
+
+    //
+    // Errors specific to this module.
+    //
     #[error("HTTP status {0} when requesting a token from the tokenserver")]
     TokenserverHttpError(u16),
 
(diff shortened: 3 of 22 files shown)


components/sync15/src/error.rs
```diff
@@ -104,20 +108,40 @@ pub enum ErrorKind {
...
+    }
+}
+
+impl From<SyncTraitsError> for Error {
+    fn from(e: SyncTraitsError) -> Self {
+        Error::from(ErrorKind::from(e))
     }
 }
 
+// XXX - we should just move this `UnacceptableBaseUrl` into the `SyncTraitsError` enum.
```
mhammond, line 144: happy to do this, was just trying to keep the patch small :)

bendk:
I need to understand the sync15 / sync15-traits split better to comment on this one.  Couple questions here:

 - What was the motivation behind the split?  I thought there was a document about this, but I couldn't find anything.  Also, is the split still needed nowaday?  I see a note mentioning that maybe the sync15 functionality could be moved into either sync-manager or sync15-traits.
 - Should a component depend on both sync15 and sync15-traits?  It seems like it would be simpler if components only depended on one or the other.

mhammond:
> I need to understand the sync15 / sync15-traits split better to comment on this one. Couple questions here:
> 
>     * What was the motivation behind the split?  I thought there was a document about this, but I couldn't find anything.  Also, is the split still needed nowaday? 

The split was made to avoid pulling in code where it wasn't needed and to avoid circular dependencies. eg:

* None of our components depend on sync-manager - instead, sync-manager depends on them (and also on sync15)
* All our components depend on sync15 (and indirectly sync15-traits) - but after we drop stand-alone syncing (possible once iOS uses the sync-manager), they can probably just depend on sync15-traits.
* Desktop syncing doesn't depend on either sync-manager or sync15.

So in today's world I don't think we can easily merge them.
* In a future world (no stand-alone syncing), it might be possible to merge sync-manager and sync15.
* It might also be possible to merge sync15 and sync15-traits with more aggressive features to avoid pulling unused code into desktop (currently webext-storage does not depend on sync15, so combining sync15 and sync15-traits would need to be done in a way that didn't introduce unused code to desktop)

But those 2 options are mutually-exclusive - we can't merge all 3 together or there will be circular dependencies between the sync engine implementations and the newly combined crate.

> I see a note mentioning that maybe the sync15 functionality could be moved into either sync-manager or sync15-traits.

I can see a future where we can do some consolidation, but I'm not sure that now is the time. Hence, my patch doesn't really change the status-quo that much - that status-quo does need to change, but it's not 100% clear how and that's a bigger job than was trying to take on here :)

> 
>     * Should a component depend on both sync15 and sync15-traits?  It seems like it would be simpler if components only depended on one or the other.

Yeah, but that's already somewhat true today - eg, neither the tabs component nor the sync-manager components, both touched by this PR, have a direct dependency on sync15-traits. Only webext-storage needs to take an explicit dependency on sync15-traits because it doesn't depend on sync15. (autofill also takes an explicit dep on sync15-traits, but I bet it doesn't need to but was instead a copy-paste error)

bendk:
> It might also be possible to merge sync15 and sync15-traits with more aggressive features to avoid pulling unused code into desktop (currently webext-storage does not depend on sync15, so combining sync15 and sync15-traits would need to be done in a way that didn't introduce unused code to desktop)

This seems like it would simplify the error handling since you would only need 1 enum at that point.  But I'm not sure how high of a priority this is.

components/fxa-client/src/internal/error.rs
```diff
@@ -148,6 +148,19 @@ error_support::define_error_conversions! {
     }
 }
 
+// Adapts errors from the sync15_traits crate into sync15 errors.
```
bendk, line 151: It's kind of sad that consumers of `sync15` need to handle both `sync15::Error` and `sync15::SyncTraitsError`, but I can't think of a better way to handle it without changing a lot of code.  I think the docstring could be improved though, what about adding someting like "This is needed because `sync15` re-exports functions from `sync15_traits` and those functions return `Result<T, SyncTraitsError>`"

components/sync15/src/error.rs
```diff
@@ -23,12 +21,33 @@ pub enum ErrorResponse {
...
+
+    #[error("Bad cleartext UTF8: {0}")]
+    BadCleartextUtf8(#[from] std::string::FromUtf8Error),
+
+    #[error("HAWK error: {0}")]
+    HawkError(#[from] rc_crypto::hawk::Error),
+
+    //
+    // Errors specific to this module.
+    //
```
tarikeshaq, line 50: how do we feel about having only one variant like:
```rs
#[error("{0}")]
Sync15TraitsError(#[from] SyncTraitsError)
```

And just get rid of the `From` implementation.

The downside is that consumers would need to go one level deeper to get the "Actual" error though so feel free to disregard this

It would also be a breaking change though, so we'd have to make sure none of the variants deleted are consumed anywhere external

components/sync15/src/lib.rs
```diff
@@ -43,6 +39,9 @@ pub use crate::sync_multiple::{
     sync_multiple, sync_multiple_with_command_processor, MemoryCachedState, SyncRequestInfo,
 };
 pub use sync15_traits::client::DeviceType;
+pub use sync15_traits::SyncTraitsError;
```
tarikeshaq, line 42: Curious why `sync15` needs to expose the error?

I think I'm missing something, aren't all the `sync15_traits::SyncTraitsError`s converted to `sync15::Error` before they are returned from any of the APIs?

If a consumer is depending on `sync15_traits` directly, it makes sense to me that it's up to the consumer to import the error through `sync15_traits` as opposed to importing it through `sync15`

tarikeshaq: I guess I'm confused on "Why can't a consumer 'just' consume the types from sync15_traits through sync15_traits" (and subsequently, 'just' handle the error resulting from those types, as opposed to getting the types and errors from `sync15`, then dealing with two errors that `sync15` returns

components/support/sync15-traits/src/error.rs
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// `Result` is unused with default features, but used with `feature=crypto`
+#[allow(dead_code)]
```
tarikeshaq, line 6: nit: can we make this a `cfg` instead of a `dead_code`?

tarikeshaq (approved):
Had a question and a thought, but otherwise no need to block on my review (I feel like I was missing some more context here, and depth-first traversing through the linked tickets with depth of 2 wasn't enough to get it 🤪)

components/sync15/src/error.rs
mhammond: This would still leave 2 different ways to spell the same error. In this patch as it stands, the tabs component can say `TabsError::SyncAdapterError::HmacMismatch` and it will match the error whether it originated in sync15 or sync15-traits. 

(I also agree with Ben's broader point that the split is a bad thing, so this would ideally help us re-uniffy things in the future)

components/sync15/src/lib.rs
mhammond: That's a good question :) I think it's because that `Into()` implementation isn't known about in tabs. I was indeed hoping it would be that simple though, so feel free to have a poke and see if you can improve that!

mhammond: (and I'll have another poke at that too, but later when I'm closer to landing it)

mhammond: > I think I'm missing something, aren't all the sync15_traits::SyncTraitsErrors converted to sync15::Error before they are returned from any of the APIs?

For a little more detail here - even though the functions exposes by sync15-traits are re-exported by sync15, their signature is still returning the `SyncTraitsError` - so I'm not surprised they aren't magically converted by the re-exported function. So we'd either need thin-wrappers in sync15 that just converted the error, *or* arrange for the conversion in sync15 to magically end up being used when (say) tabs sees the `SyncTraitsError` - it's that second option I was hoping we'd kinda get for free and I suspect the reason why we don't is that the converter in sync15 isn't seen or isn't able to be magically used.

bendk: One source of magic would be a blanket impl, but I don't think it's possible.  I tried it 2 ways and both were rejected:
  - `sync15` could have  `impl<T> From<SyncTraitsError> for T where T: From<Error>`, but this violates the orphan rule.  Both `T` and `SyncTraitsError` are from other crates.
  - `sync15-traits` could have `impl<T, U> From<SyncTraitsError> for T where T: From<U>, U: From<SyncTraitsError>`,  but then you get the error `U` is an "unconstrained type parameter".  I think the issue here is essentially the same, it's possible for multiple types that satisfy the conditions for `U`.

I think a macro would work but that seems like overkill.

PR #5084, opened 2022-08-08
Adding more breadcrumbs to diagnose SqlErrorQueryReturnedNoRows

bendk:
Based on the current logging, it seems like multiple write operations
are happening at the same time.  Added breadcrumbs around the
transaction begin/commit to test this theory further.

Also removed commented out SQL from `delete_visits_between_in_tx()`.

  - Note:

### DIFF
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -41,10 +41,12 @@ static DELETION_HIGH_WATER_MARK_META_KEY: &str = "history_deleted_hwm";
 
 /// Returns the RowId of a new visit in moz_historyvisits, or None if no new visit was added.
 pub fn apply_observation(db: &PlacesDb, visit_ob: VisitObservation) -> Result<Option<RowId>> {
+    breadcrumb!("apply_observation: begin_transaction");
     let tx = db.begin_transaction()?;
     let result = apply_observation_direct(db, visit_ob)?;
     delete_pending_temp_tables(db)?;
     tx.commit()?;
+    breadcrumb!("apply_observation: commit");
     Ok(result)
 }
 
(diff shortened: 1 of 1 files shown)


bendk:
When Mark and I discussed this last week, we also discussed removing the `delete places page` breadcrumb, since we were seeing lots of those in the DB log.  However, I didn't do that for 2 reasons:
  - When I went through more of the Sentry events, many of them didn't have that breadcrumb
  - The original theory was that these calls came from some bulk delete method and it would be better to just log the number of pages deleted.  However, after looking through the code more, I believe the top-level call is `bookmarks_delete()`, which inputs a single GUID.  It seems like the user is actually clicking on the trash icon for individual bookmarks.

components/places/src/storage/history.rs
```diff
@@ -405,14 +413,18 @@ pub fn prune_destructively(db: &PlacesDb) -> Result<()> {
 }
 
 pub fn prune_older_visits(db: &PlacesDb) -> Result<()> {
-    let _tx = db.begin_transaction()?;
```
mhammond, line 408: oops - this was a bug that the tx wasn't committed, right?

mhammond (approved):
👍

components/places/src/storage/history.rs
bendk: I believe it gets committed when `_tx` gets dropped.

mhammond: https://docs.rs/rusqlite/0.12.0/rusqlite/struct.Transaction.html#method.drop_behavior says "Transactions will roll back by default. Use commit method to explicitly commit the transaction, or use set_drop_behavior to change what happens when the transaction is dropped." and I can't see where we call `set_drop_behavior`.

PR #5095, opened 2022-08-18
Add applyFirstRunExperiment to NimbusInterface

Amejia481:
Related ticket https://mozilla-hub.atlassian.net/browse/EXP-2697

As the actual alternative for loading first run experiments ([setExperimentsLocally](https://github.com/mozilla/application-services/blob/d5a639d758a8ca7189c5fd2d40e7f6435c35b343/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt#L467)), does it async this causes nondeterministic behaviour, as in some cases we will load the experiments before the UI is displayed and in some we won't. This proposal is exposing a new functionality `tryLoadAndApplyFirstRunExperiment` to load experiments in an sync and taking a timeout in case it takes longer than expected. 

This is a prototype for loading first run experiments 

  - Note:

### DIFF
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
@@ -18,6 +18,9 @@ import androidx.annotation.WorkerThread
 import androidx.core.content.pm.PackageInfoCompat
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.withTimeout
 import mozilla.telemetry.glean.Glean
 import org.json.JSONObject
 import org.mozilla.experiments.nimbus.GleanMetrics.NimbusEvents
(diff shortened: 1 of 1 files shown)


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -466,12 +477,20 @@ open class Nimbus(
...
+            setExperimentsFrom(file)
+        }
+    }
+
+    override fun tryLoadAndApplyFirstRunExperiments(@RawRes file: Int, timeout: Long) {
+        runBlocking {
+            try {
+                withTimeout(timeout) {
+                    setExperimentsFrom(file)
+                    applyPendingExperimentsOnThisThread()
```
Amejia481, line 489: To avoid this consuming resource on the main thread, we could lunch on `dbScope` and join on it until it finish and cancel this job if we timeout, something like


```kotlin
 runBlocking {
        try {
            withTimeout(500) {
                job.join()
            }
        } catch (ex: TimeoutCancellationException) {
            job.cancel()
        }
    }
```

tarikeshaq:
@Amejia481 do we want this in 105?

Also, I'm a little hesitant about blocking the main thread - the timeout would help but I'm nervous that we wouldn't get the timeout right and possibly discriminate against older devices... that said if it's better than the alternative I'm willing to concede and get this landed :)

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
tarikeshaq: woah beat me to it by a few seconds 😬

Amejia481:
> @Amejia481 do we want this in 105?

I think 106 is good too, we just want to have time to test it in the while.

tarikeshaq:
@Amejia481 is this safe to close? It's been stale a while

Amejia481:
Yes, sorry I completely forgot about this

PR #5097, opened 2022-08-20
Add webext-storage test showing how the outgoing payload looks.

mhammond:
This is a test only change to ensure no future regressions.


tarikeshaq (approved):
🚀

PR #5098, opened 2022-08-20
Update dependences: fs-err's branch is now named 'main'

mhammond:
dependency checks are failing on current main.


tarikeshaq (approved):
🚀

PR #5100, opened 2022-08-22
Update the bridged_engine to not rely on sync15_traits::Payload.

mhammond:
Sorry for the large PR - there's no hurry here and I'm happy to discuss the motivations behind this some more. I'm open to bikeshedding here, because some of the names and concepts will later reach further into the sync crates. This will impact the branch with Sammy's tabs work, but the fixes should be easy and I'll do them once this lands.

---
This shows how I envisage sync15_traits should work for both the bridged
engine and the bso_record based (ie, sync15) engines. This patch only
touched the bridged_engine to keep the patch small with a future
patch using this shape for the rest of the crate soon.

It leans further into the "envelope" concept introduced by the
bridged_engine, but the envelope is now split from the payload
so it can be used by bso_record.

Now, instead of going to and from the sync15_traits::Payload, we just
move to and from <T> directly:

* When converting from an incoming record to a <T>, the result is
  captured in a new enum, `IncomingDeserPayload`, which allows us
  to capture incoming tombstones and invalid record. No errors are
  returned deserializing <T>, the enum variants cover all cases.

* When converting from T to an outgoing record, we will panic if
  various constraints are not met - eg, if you have specified an
  invalid GUID. All such errors are considered "static" errors
  and aren't influenced by runtime conditions, so a panic means
  we don't have to introduce new error handling for these "impossible"
  states.

It also kills the concept of "auto" fields - it's no longer necessary
to have a field called, say, 'ttl' in your record to arrange for a TTL
on the record - these are now handled by the envelope. Note that the
'id' field on the payload does remain somewhat magic.

### DIFF
--- a/components/support/sync15-traits/src/bridged_engine.rs
+++ b/components/support/sync15-traits/src/bridged_engine.rs
@@ -2,11 +2,89 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::{error::Error, fmt};
-
+use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
-use super::{Guid, Payload, ServerTimestamp};
+use super::Guid;
+
+use crate::error;
+
+// This module hold structs which will eventually live outside of this
+// module and will be re-used by bso_record. It lives here temporarily
+// as we move the bridged_engine away from Payload so we can see the shape
+// that this and bso_record will share.
+
+// It is a separate module here to help split it out later.
+pub mod public {
+    use crate::{Guid, ServerTimestamp};
+    use serde::{Deserialize, Serialize};
+
+    /// An envelope for an incoming item. Envelopes carry all the metadata for
+    /// a Sync BSO record (`id`, `modified`, `sortindex`), *but not* the payload
+    /// itself.
+    #[derive(Debug, Clone, Deserialize)]
+    pub struct IncomingEnvelope {
+        pub id: Guid,
+
+        #[serde(skip_serializing)]
+        // If we don't give it a default, we fail to deserialize
+        // items we wrote out during tests and such.
+        // XXX - we should probably fix the tests and  kill this.
+        #[serde(default = "ServerTimestamp::default")]
+        pub modified: ServerTimestamp,
+
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub sortindex: Option<i32>,
+
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub ttl: Option<u32>,
+    }
+
+    /// An envelope for an outgoing item. This is conceptually identical to
+    /// [IncomingEnvelope], but omits fields that are only set by the server,
+    /// like `modified`.
+    #[derive(Debug, Default, Clone, Serialize, PartialEq)]
+    pub struct OutgoingEnvelope {
+        pub id: Guid,
+        pub sortindex: Option<i32>,
+        pub ttl: Option<u32>,
+    }
+
+    // Allow an envelope to be constructed with nothing but a guid.
+    impl From<Guid> for OutgoingEnvelope {
+        fn from(id: Guid) -> Self {
+            OutgoingEnvelope {
+                id,
+                ..Default::default()
+            }
+        }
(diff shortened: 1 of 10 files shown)


components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -87,134 +168,334 @@ pub trait BridgedEngine {
...
-    #[serde(default)]
-    pub sortindex: Option<i32>,
-    #[serde(default)]
-    pub ttl: Option<u32>,
+pub struct IncomingBridgeRecord {
+    #[serde(flatten)]
+    pub envelope: IncomingEnvelope,
     // Don't provide access to the cleartext directly. We want all callers to
-    // use `IncomingEnvelope::payload`, so that we can validate the cleartext.
+    // use `IncomingEnvelope::payload`, so that we can validate the cleartext
```
bendk, line 208: Nit: `IncomingBridgeRecord::payload`

components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -87,134 +168,334 @@ pub trait BridgedEngine {
...
+                        );
+                        return IncomingDeserPayload::Malformed;
+                    }
+                    if !self.envelope.id.is_valid_for_sync_server() {
+                        log::warn!("malformed incoming record: id is not valid: {}", id);
+                        return IncomingDeserPayload::Malformed;
+                    }
+                    log::warn!("incoming record has 'id' in the payload - it does match, but is still unexpected");
+                }
+                None => {
```
bendk, line 256: Note: This would capture the case where `id` exists as a non-string value, is it worth checking for that?

components/support/sync15-traits/src/bridged_engine.rs
```diff
@@ -2,11 +2,89 @@
...
+pub mod public {
+    use crate::{Guid, ServerTimestamp};
+    use serde::{Deserialize, Serialize};
+
+    /// An envelope for an incoming item. Envelopes carry all the metadata for
+    /// a Sync BSO record (`id`, `modified`, `sortindex`), *but not* the payload
+    /// itself.
+    #[derive(Debug, Clone, Deserialize)]
+    pub struct IncomingEnvelope {
+        pub id: Guid,
```
bendk, line 27: How about adding a note here about the how `id` gets special handling.

bendk:
Added a couple specific suggestions, but I think the main most important thing is to get the concepts correct.  I'm in the middle of trying to wrap my head around this, hopefully my comments make sense....

I found it hard to follow how all the structs relate to each other.  For example `IncomingBridgedRecord` contain the field `cleartext` which can then be turned into an `IncomingDeserPayload`, which may contain a `T`, which is named "record" when used as a variable.  Following that chain was quite tricky, especially since both the object on the bottom on top are called "records".  I'm not exactly sure how to best clarify this, but here's some suggestions:

- Add an overview describing the general shape of the data in the module docstring.  I think this should describe things from the 1000ft view, using words like "payload" rather than `IncomingDeserPayload`
- Figure out words to distinguish the server records from the local records.  I'm not totally sure, but maybe  "server record" vs "local record" or "server item" vs "record".

Maybe the structs could be reorganized slightly.  Here's a sketch of a potential improvement:
```
pub struct IncomingServerRecord<T: Payload> {
    pub envelop: IncomingEnvelope,
    pub payload: T,
}

pub struct OutgoingServerRecord<T: Payload> {
    pub envelop: OutgoingEnvelope,
    pub payload: T,
}

pub trait Payload: Serialize + Deserialize {
    type LocalRecord;

    // deserialize and/or decrypt this payload
    fn contents() -> PayloadContents<LocalRecord>;
}

pub enum PayloadContents<T> {
    LocalRecord(T),
    Tombstone,
    Malformed,
}

pub struct BridgePayload {
    cleartext: String
}

pub struct BsoPayload {
    encrypted_text: String
}
```

I think this allows the BSO and bridged payloads to share the same API, although I definitely don't know enough about the details to be sure.  I also like that it makes it very clear that a server record is an envelope + a payload.

components/support/sync15-traits/src/bridged_engine.rs
mhammond: yeah, I guess so, thanks.

PR #5102, opened 2022-08-23
Add a 'full-sync' feature to tabs.

mhammond:
If the new feature is enabled, tabs can sync - otherwise it can not.
The intention is that later there will be a "bridged-sync" feature,
and that in-practice, one or the other will be enabled - however, the
way features work, it must be possible that no features are enabled.

The new feature is not a default feature because I expect we will
want the default features to be what desktop consumes - ie, that
the default feature will be bridged-sync once that exists.

This isn't strictly required yet, but is being done now to make
future reviews of the bridged engine easier to reason about.


bendk (approved):
LGTM.

lougeniaC64 (approved):
🚀

PR #5111, opened 2022-08-29
Rename cleartext to payload in the bridged engine envelopes

mhammond:
Things like #5100 and the broader refactor to close the gap between "bridged engines" and normal sync engines would be a whole lot easier if the bridged-engine called its payload field "payload". I can't see a good reason to not make that change, and now seems a good time because only the webext-storage engine actually cares.

So I made this PR, and the matching [bug 1787688](https://bugzilla.mozilla.org/show_bug.cgi?id=1787688) - that bug also helps some of the tabs work.

### DIFF
--- a/components/support/sync15-traits/src/bridged_engine.rs
+++ b/components/support/sync15-traits/src/bridged_engine.rs
@@ -57,7 +57,7 @@ pub trait BridgedEngine {
     /// times per sync, once for each batch. Implementations can use the
     /// signal to check if the operation was aborted, and cancel any
     /// pending work.
-    fn store_incoming(&self, incoming_cleartexts: &[IncomingEnvelope]) -> Result<(), Self::Error>;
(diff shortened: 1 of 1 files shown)


skhamis (approved):
This is a great change with hardly any obvious drawbacks. LGTM 🚀

skhamis:
(btw, the swiftlint is a circleci issue regarding an outdated mac runner) -- this should be fine to ignore, especially since this doesn't actually touch swift files

mhammond:
> would be a whole lot easier if the bridged-engine called its payload field "payload".

I'm actually no longer sure that's true. It's more consistent, but I still need to spell out the payload explicitly for serde-flatten reasons.

bendk (approved):
Agree with this one as well, anything that brings the bridged engine and sync15-traits code closer together is good.

PR #5113, opened 2022-08-29
Upgrades the mac runners on CI to run on 13.4.1

tarikeshaq:
Our format CI jobs have been failing because the runners are on an older version of macos

Since Firefox iOS is using SPM now, they are no longer blocked by us for Xcode upgrade, and they are [**already** supporting Xcode 13.4.1](https://github.com/mozilla-mobile/firefox-ios#this-branch-main)

  - Note:
- [x] **Changelog**: CI change, doesn't need a changelog


skhamis:
Hmm for some reason `gyp` can't be found on the new runners: https://app.circleci.com/pipelines/github/mozilla/application-services/31269/workflows/f48a5814-49b6-4272-b0ee-9a5db4517570/jobs/140154

tarikeshaq:
heh looks like adding the python 3.9 bin directory to `PATH` for the runner worked - but I'm not sure if that's the "best" way to get `gyp` visible to the shell... I'm okay keeping this as is, but would take any suggestions here

tarikeshaq:
For context, the reason we need this is because the runner that run xcode 13.4.1 in Circle CI use MacOS 12.X.X when the older version of Xcode runners used MacOS 11.X.X

MacOS 12 ships with python 3.9, and it looks like the `PATH` isn't properly set by default

components/nimbus/build.rs
```diff
@@ -2,29 +2,16 @@
...
-            "-o",
-            out_dir.as_str(),
-            "./metrics.yaml",
-        ])
-        .status()
-        .expect("Error generating Glean Rust bindings");
+    Builder::default()
+        .file("./metrics.yaml")
+        .generate()
+        .unwrap();
```
tarikeshaq, line 14: @travis79 upgrading the runners caused problems here because MacOS 12 doesn't have `python` installed (it's python). I decided to just used `glean_build` here because that's what we'll do soon anyways

mhammond (approved):
It's mildly frustrating that our deps change for build tools, but meh. Using glean-build makes sense to me.

PR #5114, opened 2022-09-01
Update rust to 1.63/1.61

mhammond:
WIP as we will need a new uniffi release with https://github.com/mozilla/uniffi-rs/issues/1330 fixed.

I bumped the uniffi version but didn't update the semver specification so we remain compatible with what's in m-c (although I think @badboy is going to bump that soon anyway)

Note that the `Eq` clippy errors were trivial, but tedious and numerous. I ended up fixing most of them, but ignored them in crates I know are actively being worked on. Let me know if you think I should have just ignored clippy in more places?


mhammond:
Superseded by #5128

PR #5125, opened 2022-09-10
Merge sync15-traits back into sync15 via features.

mhammond:
~This is where I think we want to end up, but I intend splitting this into multiple PRs first.~

Now that I look at this patch, I think splitting it into individual PRs will just make the review burden higher - lots of stuff being moved around in a somewhat temporary way, just to move it again later. The PR is large, but quite mechanical, and seeing the end result makes the intent clear.


mhammond:
FWIW, I also test a test-vendor of this into m-c, and it works as expected - the sync15-traits deps there all need to change to sync15, but there are no other dependencies introduced.

Cargo.toml
```diff
@@ -23,7 +23,6 @@ members = [
     "components/support/rc_crypto/nss/nss_sys",
     "components/support/rc_crypto/nss/systest",
     "components/support/sql",
-    "components/support/sync15-traits",
```
tarikeshaq, line 26: 🤩

components/sync15/Cargo.toml
```diff
@@ -6,24 +6,48 @@ authors = ["application-services@mozilla.com"]
 license = "MPL-2.0"
 exclude = ["/android", "/ios"]
 
+[features]
+# The default feature is what mozilla-central wants to make vendoring/patching easier.
```
tarikeshaq, line 10: Could be useful to add more context here on how it's easier - like because it pulls in less dependencies etc

components/autofill/Cargo.toml
```diff
@@ -19,8 +19,7 @@ serde_derive = "1"
 serde_json = "1"
 sql-support = { path = "../support/sql" }
 sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random", "serde_support"] }
-sync15 = { path = "../sync15" }
-sync15-traits = {path = "../support/sync15-traits"}
+sync15 = { path = "../sync15", features = ["sync-engine"] }
```
tarikeshaq, line 22: since the default is `sync-engine` do we need the `features` here? it's nice to be explicit so I don't mind it!

components/sync15/Cargo.toml
```diff
@@ -6,24 +6,48 @@ authors = ["application-services@mozilla.com"]
...
+default = ["sync-engine"]
+
+random-guid = ["sync-guid/random"]
+
+# Some consumers of this just need our encrypted payloads and no other sync functionality.
+crypto = ["rc_crypto", "base16", "base64"]
+
+# Some crates need to implement a "sync engine", but aren't a "sync client" (ie, their
```
tarikeshaq, line 18: can we link to our docs that define what `engine` is here, https://mozilla.github.io/application-services/book/design/sync-manager.html?highlight=engine#clients-engine maybe?

That said, maybe it's better this goes in the `README.md` of this crate

components/places/Cargo.toml
```diff
@@ -8,9 +8,11 @@ exclude = ["/android", "/ios"]
 
 [features]
 default = []
+# TODO: we should have a standalone-sync feature which is enabled only for iOS.
```
tarikeshaq, line 11: is this TODO resolved?

components/tabs/Cargo.toml
```diff
@@ -21,8 +21,7 @@ serde_derive = "1"
 serde_json = "1"
 sql-support = { path = "../support/sql" }
 sync-guid = { path = "../support/guid", features = ["random"] }
-sync15 = { path = "../sync15", optional = true }
-sync15-traits = { path = "../support/sync15-traits" }
+sync15 = { path = "../sync15", features = ["sync-engine"] }
```
tarikeshaq, line 24: should this be `standalone-sync` or is the flag on line 10 enough for us to realize later?

tarikeshaq (approved):
Awesome! I'm a big fan of this refactor 🚀 

Had a few very small questions, I also didn't read all the code changes since they all looked like simply moving dependencies from `sync15-traits` to `sync15` and no real functional changes

components/autofill/Cargo.toml
mhammond: Good question :) You are correct that we don't need it, but I too thought that explicit was better.

mhammond:
Thanks Tarik, I ended up addressing all of your comments in various ways - mainly via extra rustdoc comments and extra comments in Cargo.toml. Have a look at the last commit in this set to see them and lmk if you think there are further improvements I can make.

tarikeshaq (approved):
The new changes make sense! Thank you for adding the extra context 🙏

bendk (approved):
This makes a lot of sense to me.  I'm assuming there aren't any functional changes and this is just moving the code around.

PR #5128, opened 2022-09-12
Updating the Rust versions

bendk:
This updates it to the min version used by Desktop nightly.  I'm not sure if we should track nightly or release in our rust version, but I'm making this PR because I'm hoping it the coverage CI failures.  Those are failing with:
```
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
```

I wanted to update the max version to 1.63, but that leads to clippy errors because https://github.com/mozilla/uniffi-rs/pull/1331/ has not been released yet.

  - Note:


rust-toolchain.toml
```diff
@@ -7,7 +7,7 @@
 #   supported and used in CI.
 
 [toolchain]
-channel = "1.60.0"
```
tarikeshaq, line 10: with https://github.com/mozilla/uniffi-rs/pull/1314 landed, we'll be able to pick up 1.63 soon too!

tarikeshaq: aaah I realize that's exactly what you said on the PR message

mhammond:
See also https://github.com/mozilla/application-services/pull/5114 :)

bendk:
@mhammond Thanks for pointing that out, I completely missed it.

I just released a new UniFFI version today so I think we can bump both the major and minor Rust versions.  I also copied some code from Mark's branch.

I think this one might be ready to merge the main [differences between this and update-rust](https://github.com/bendk/application-services/compare/11c0aed..f27bac0a9) are:
  - Updated UniFFI to `0.20.0`
  - I kept the signature of the `logins::import_multiple()` function the same.  I think Kotlin is still expecting a string return and didn't want to make a change there.
  - I chose to add more Eq derives rather than adding clippy allows.  I don't have a strong feeling about this either way though.
  - There was some extra changes to nimbus in the other branch that I didn't bring in.  Tell me if I should though.

components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -719,8 +719,9 @@ impl TypeFinder for PropDef {
 
 pub type Literal = Value;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub(crate) struct ImportedModule<'a> {
+    #[allow(dead_code)]
     pub(crate) id: ModuleId,
```
tarikeshaq, line 725: hmmm @jhugman would it be okay to get rid of this?

components/logins/src/store.rs
```diff
@@ -166,8 +166,8 @@ impl LoginStore {
     pub fn import_multiple(&self, logins: Vec<Login>, enc_key: &str) -> ApiResult<String> {
         handle_error! {
             let encdec = EncryptorDecryptor::new(enc_key)?;
-            let metrics = self.db.lock().import_multiple(logins, &encdec)?;
-            Ok(serde_json::to_string(&metrics)?)
+            self.db.lock().import_multiple(logins, &encdec)?;
+            Ok(serde_json::to_string(&())?)
```
tarikeshaq, line 170: should we add a comment here, it looks a little unconventional

tarikeshaq (approved):
Looks good to me, but i'd leave it to @mhammond to verify this makes sense given his other PR!

mhammond (approved):
> * I kept the signature of the `logins::import_multiple()` function the same.  I think Kotlin is still expecting a string return and didn't want to make a change there.

Yeah - I misread the kotlin code here - I'm sure that result is still being dropped on the floor though, so we should consider getting an issue on file to kill this - maybe a good-first.

> * I chose to add more Eq derives rather than adding clippy allows.  I don't have a strong feeling about this either way though.

Yeah, I just got sick of it :)

> * I added an `[allow(dead_code)]` to `components/support/nimbus-fml/src/intermediate_representation.rs` rather than removing the module_id field.  I'm not sure what's best here, I just didn't feel comfortable changing the nimbus code.

Like the Kotlin issue above, I was just trying to be aggressive because I think it's safe to assume that field was a mistake - if it was intentional I'd expect a comment. But yeah, totally fine with this and hopefully Nimbus treats that dead_code annotation as a smell.

lougeniaC64 (approved):
LGTM!

rvandermeulen:
@bendk Should this be added to the changelog?

bendk:
Good catch, it probably should.  #5149

PR #5139, opened 2022-09-15
Update sync15 BSO concept to use envelopes and kill sync15::Payload.

mhammond:
This is a refactor of a "BSO" - a "basic storage object - and how engines works with them - it kills `sync15::Payload`. The new implementation uses an "envelope" plus a mechanism for converting the "payload" to and from the "content" structs used by engines. This content no longer has `ttl` or `sortindex` because they are in the "envelope". 

`SyncEngine` `BridgedEngine` now consistently use BSOs, which gets us closer to being able to merge those 2 traits.

It's a very large patch because all syncable components are touched :( There's a fair bit of boiler-plate noise in the test changes which can be glossed over.

To share the review burden we could "sub-review" components? Maybe put your name next to a component, then "select" it once you are happy with it?

Obviously though, feel free to look at as much as you like!

- [x] sync15 (@bendk has already had a look at this)
- [x] autofill (lougenia)
- [x] webext-storage (skhamis)
- [x] places (bendk)
- [x] tabs (skhamis)
- [x] fxa-client (teshaq)
- [x] logins (lougenia)


mhammond:
This is a WIP for a revamp of how records are moved between SyncEngine and the servers. It gets bridged_engine and normal engines much closer to using the same types, and kills the somewhat confusing sync15::Payload entirely.

It leans into envelopes from the bridged engine, so engines no longer need to have a "ttl" or "sortindex" field, but instead get to deal with explicit envelopes.

This shows that places and tabs might look like, but *has not* touched places tests - so they are expected to fail.

I'd love early feedback here, particularly on:
* Does the new "bso" concept make sense?
* Do the names in there make sense?
* Should it just be flat single-file module? Or maybe the envelopes should move out into their own file
* All over "XXX" comments I've added.

There's no urgency here - I intent putting this down for a couple of weeks, but would love early feedback on the general shape and context.

components/places/src/storage/history.rs
```diff
@@ -1157,17 +1163,20 @@ pub mod history_sync {
...
-                    id: page.guid,
-                    title: page.title,
-                    hist_uri: page.url.to_string(),
-                    sortindex: page.frecency,
-                    ttl: HISTORY_TTL,
-                    visits,
-                }),
-            );
+            let content = HistoryRecord {
+                id: page.guid.clone(),
```
mhammond, line 1167: don't need id here

components/sync15/src/bso/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+/// # encrypted bsos:
+/// Encrypted BSOs are still a "bso" (ie, a record with a field names `payload` which is a string)
+/// but the payload is in the form of an EncryptedPayload.
+/// For example, crypto/keys:
+/// ```json
+/// {
+///   "id":"keys",
+///   "modified":1661564513.74,
+///   "payload":"{\"IV\":\"snip-base-64==\",\"hmac\":\"snip-hex\",\"ciphertext\":\"snip-base64==\"}"
+/// }```
```
mhammond, line 44: these docstrings need a bit more love

components/sync15/src/bso/content.rs
```diff
@@ -0,0 +1,385 @@
...
+        Ok(Self {
+            envelope,
+            payload: serde_json::to_string(&json)?,
+        })
+    }
+}
+
+// Helpers for packing and unpacking serde objects to and from a <T>. In particular:
+// * Helping deal complications around raw json payload not having 'id' (the envelope is
+//   canonical) but needing it to exist when dealing with serde locally.
```
bendk, line 126: Could you describe this case some more?  I think an example would be good.

components/sync15/src/bso/content.rs
```diff
@@ -0,0 +1,385 @@
...
+// Deserializing json into a T
+fn json_to_kind<T>(mut json: serde_json::Value, id: &Guid) -> IncomingKind<T>
+where
+    T: for<'de> serde::Deserialize<'de>,
+{
+    if let serde_json::Value::Object(ref map) = json {
+        if map.contains_key("deleted") {
+            return IncomingKind::Tombstone;
+        }
+    };
```
bendk, line 138: Maybe merge this into the `if let` below?

components/sync15/src/bso/content.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    // In general, the payload does not carry 'id', but <T> does - so put
+    // it into the json before deserializing the record.
```
bendk, line 141: Can you add where we get `id` from (the envelope I think)?

components/sync15/src/bso/crypto.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::error;
+use crate::key_bundle::KeyBundle;
+use crate::EncryptedPayload;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+
+// The BSO implementation we use for encrypted payloads.
+// Note that this is almost identical to the CleartextBso implementations, except
+// instead of a String payload we use an EncryptedPayload. Obviously we *could*
+// just use a String payload and transform it into an EncryptedPayload - any maybe we
+// should - but this is marginally optimal in terms of deserialization.
```
bendk, line 18: This design makes sense to me.

components/sync15/src/bso/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+/// (Note that as described above, most code working with bsos *do not* use that `payload`
+/// directly, but instead a decrypted cleartext bso.
+///
+/// Note all collection responses are the same shape as `crypto/keys` - a `payload` field with a
+/// JSON serialized EncryptedPayload, it's just that the final <T> content differs for each
+/// collection (eg, tabs and bookmarks have quite different <T>s JSON-encoded in the
+/// String payload.)
```
bendk, line 49: These examples and detail is great.  I think the last piece of the pie would be to show the flow for BSOs.  IIUC this would be something like

- Incoming encrypted data: `EncryptedIncomingBso` -> `IncomingBso` -> `IncomingContent<T>`
- Incoming cleartext data: `IncomingBso` -> `IncomingContent<T>`
- Outgoing encrypted data: `EncryptedOutgoingBso` -> `OutgoingBso`
- Outgoing cleartext data is just an `OutgoingBso`

What do you think about an `OutgoingContent<T>` struct?   It's kind of too trivial to deserve a struct, but the symmetry could be nice.

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -33,30 +34,42 @@ impl<'a> IncomingApplicator<'a> {
...
+                match value["type"].as_str() {
+                    Some("bookmark") => self.store_incoming_bookmark(timestamp, &value)?,
+                    Some("query") => self.store_incoming_query(timestamp, &value)?,
+                    Some("folder") => self.store_incoming_folder(timestamp, &value)?,
+                    Some("livemark") => self.store_incoming_livemark(timestamp, &value)?,
+                    Some("separator") => self.store_incoming_sep(timestamp, &value)?,
+                    _ => {
+                        // XXX - we probably should treat this like malformed so the entire
+                        // sync doesn't fail?
+                        // (Either that, or have ::Malformed below fail too?)
```
bendk, line 59: I personally like using `support::error::report_error!` for both of them, then continuing on.  That way we escalate it to a sentry report, but still try to continue syncing for the user.

components/sync15/src/bso/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+mod crypto;
+#[cfg(feature = "crypto")]
+pub use crypto::{IncomingEncryptedBso, OutgoingEncryptedBso};
+
+// We re-export everything in the content module.
+mod content;
+pub use content::{IncomingContent, IncomingKind};
+
+// XXX - on one hand, it seems like an "envelope" submodule makes sense.
+// OTOH, that module would be ~40 lines long, which seems overkill.
```
bendk, line 87: I'm for moving more struct definitions into this module.  If `IncomingEnvelope` just needs a struct definition, then I think it should stay here.

bendk (commented):
Looks great to me.  I didn't go over all the changes to the places/tabs engines, but it seemed fine based on a quick read.

>    Does the new "bso" concept make sense?
>   Do the names in there make sense?

IMO, yes and yes.

>  Should it just be flat single-file module? Or maybe the envelopes should move out into their own file

I left a comment about this, but my preference would be a single module that contains the high-level picture of things and submodules with the details.  I think a nice way to split it up would be to put the struct definitions in the top-level module and the `impl` blocks in submodules.

mhammond:
Thanks Ben - I did most of your comments in 341ffa923e09e947f341922d6559df8d915c27e8, so if you don't mind looking over that to make sure I didn't misunderstand your ideas it would be great. I think I'll open a new issue for the error reporting and do it as a followup just to stop this getting bigger than Ben Hur!

bendk (approved):
One minor comment on a doc string, other than that everything looks great.  Let's get this merged.

components/sync15/src/enc_payload.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[serde(rename = "IV")]
+    pub iv: String,
+    pub hmac: String,
+    pub ciphertext: String,
+}
+
+impl EncryptedPayload {
+    #[inline]
+    pub fn serialized_len(&self) -> usize {
+        (*EMPTY_ENCRYPTED_PAYLOAD_SIZE) + self.ciphertext.len() + self.hmac.len() + self.iv.len()
```
bendk, line 23: Is this always accurate?  I think that assumes there's zero overhead when JSON-encoding the strings.  What if one of them contain a `"` char that needs to be escaped to `\"`?

I'm guessing this code isn't new, but I figured it was worth flagging.

components/sync15/src/client/coll_update.rs
```diff
@@ -43,19 +46,17 @@ pub fn fetch_incoming(
         // That should cause us to re-read crypto/keys and things should
         // work (although if for some reason crypto/keys was updated but
         // not all storage was wiped we are probably screwed.)
-        let decrypted = record.decrypt(&state.key)?;
-        result.changes.push(decrypted.into_timestamped_payload());
+        result.changes.push(record.into_decrypted(&state.key)?);
```
bendk, line 49: I'm loving all the small changes like where the new code is cleaner and more readable.

bendk (changes requested):
Reviewed sync15 and places.  The changes look good overall, but there were a couple of places I had questions.

One general question I had was if we should try to hook up the new error support code to the areas that are changing. This PR already has a lot of changes, but think it would be pretty simple.  Mostly it would just mean replacing some of the `log::error()` calls with `error_support::report_error()` and the other logging calls with `error_support::report_breadcrumb()`.

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -91,136 +89,33 @@ pub trait BridgedEngine {
...
+impl From<Vec<OutgoingBso>> for ApplyResults {
+    fn from(records: Vec<OutgoingBso>) -> Self {
         Self {
-            envelopes,
+            records,
             num_reconciled: None,
         }
     }
 }
-
```
skhamis, line 122: The simplification is simply wonderfullll

components/sync15/src/bso/crypto.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Decrypt a BSO, consuming it into a clear-text version.
+    pub fn into_decrypted(self, key: &KeyBundle) -> error::Result<IncomingBso> {
+        Ok(IncomingBso::new(self.envelope, self.payload.decrypt(key)?))
+    }
+    /// Reference the payload.
+    pub fn payload(&self) -> &EncryptedPayload {
+        &self.payload
+    }
+    /// Consume the payload
+    pub fn consume_payload(self) -> EncryptedPayload {
```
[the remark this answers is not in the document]
skhamis: We don't necessarily need to follow this but https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv (also a generally great read)

would imply that `into_payload()` makes sense since it consumes the data. However `consume_payload` is also pretty clear hah.

components/sync15/src/bso/crypto.rs
```diff
@@ -0,0 +1,195 @@
...
+// The BSO implementation we use for encrypted payloads.
+// Note that this is almost identical to the IncomingBso implementations, except
+// instead of a String payload we use an EncryptedPayload. Obviously we *could*
+// just use a String payload and transform it into an EncryptedPayload - any maybe we
+// should - but this is marginally optimal in terms of deserialization.
+#[derive(Deserialize, Debug)]
+pub struct IncomingEncryptedBso {
+    #[serde(flatten)]
+    pub envelope: IncomingEnvelope,
+    // pub is slightly unfortuate, but we need it so we can get hold of our global keys.
```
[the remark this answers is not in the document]
mhammond: actually it was stale - the `payload()` and `consume_payload()` methods existed to work around the fact it wasn't pub, but I think that's just extra complication for no value - so I reworked this a little (made payload pub(crate) and deleted the 2 methods)

components/sync15/src/bso/crypto.rs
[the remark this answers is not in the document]
mhammond: As above, I killed these methods and just made the payload `pub(crate)`

components/sync15/src/enc_payload.rs
mhammond: I don't think the encrypted payload can actually have those chars? But maybe worth digging in to this to be sure, but let's not derail this patch with that because as you note, this code has existed for years

mhammond (commented):
Thanks for the reviews - Made the suggested changes and added some `report_error` calls where I thought it made sense.

components/sync15/src/bso/content.rs
```diff
@@ -0,0 +1,369 @@
...
+//! It can:
+//! * Parse JSON into some <T> while handling tombstones and invalid json.
+//! * Turn arbitrary <T> objects with an `id` field into an OutgoingBso.
+
+use super::{IncomingBso, IncomingContent, IncomingKind, OutgoingBso, OutgoingEnvelope};
+use crate::Guid;
+use serde::Serialize;
+
+// The only errors we return here are serde errors.
+type Result<T> = std::result::Result<T, serde_json::Error>;
```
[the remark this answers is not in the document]
bendk: That makes sense to me, let's keep it this way.

components/sync15/src/enc_payload.rs
bendk: I did a quick check and it looks like it's base64 encoded, so I don't think there's any chance of an issue.

bendk (approved):
The changes look great, especially that `as_json` documentation.

components/autofill/src/sync/address/incoming.rs
```diff
@@ -19,16 +20,33 @@ use sync_guid::Guid as SyncGuid;
...
+    let bso = IncomingBso {
+        envelope: IncomingEnvelope {
+            id,
+            modified: ServerTimestamp::default(),
+            sortindex: None,
+            ttl: None,
+        },
+        payload: raw,
+    };
+    // For hysterical raisins, we use an IncomingContent<AddressPayload> to convert
```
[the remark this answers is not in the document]
lougeniaC64: Wait, is this an inside joke or a typo?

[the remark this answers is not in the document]
mhammond: lol - not really an "inside" joke (google shows it relatively well known) but certainly a joke :) I think rnewman was the first person I heard use it when describing obscure parts of desktop sync, so I'm keeping the flame burning!

components/autofill/src/sync/common.rs
```diff
@@ -164,47 +164,43 @@ macro_rules! sync_merge_field_check {
...
+) -> anyhow::Result<Vec<(OutgoingBso, i64)>> {
     let mut payload = get_outgoing_records(conn, data_sql, record_from_data_row)?;
 
     payload.append(&mut get_outgoing_records(conn, tombstone_sql, &|row| {
-        Ok(OutgoingRecord::Tombstone {
-            guid: Guid::from_string(row.get("guid")?),
-        })
+        Ok((
+            OutgoingBso::new_tombstone(Guid::from_string(row.get("guid")?).into()),
+            0,
```
lougeniaC64, line 202: nit: magic number

components/autofill/src/sync/credit_card/incoming.rs
```diff
@@ -22,19 +23,37 @@ use sync_guid::Guid as SyncGuid;
...
+    let bso = IncomingBso {
+        envelope: IncomingEnvelope {
+            id,
+            modified: ServerTimestamp::default(),
+            sortindex: None,
+            ttl: None,
+        },
+        payload,
+    };
+    // For hysterical raisins, we use an IncomingContent<CCPayload> to convert
```
lougeniaC64, line 41: If hysterical raisins isn't a thing, now it is.

lougeniaC64 (approved):
Assuming we're still splitting things up by component, I reviewed logins and autofill. Nice refactor! 🚀

skhamis (approved):
Sorry forgot to add my name to the components, while I tried to generally look at all of them -- I'll put my name on tabs/webext-storage since I've either directly/indirectly worked with those two components.

tarikeshaq (approved):
heh took a look at the fxa_client, was easy when it was only a couple lines 😬 Didn't take a deeper look into the change but everything else seems covered by others reviews!

PR #5142, opened 2022-09-16
Upgrade NDK to r25b.

rvandermeulen:


rvandermeulen:
This seems to work. I know m-c is blocked on updating for reasons, but Glean already updated to r25. Any reason we shouldn't take this, @mhammond?

mhammond (dismissed):
LGTM.

mhammond (approved):

PR #5145, opened 2022-09-16
Update SQLCipher to version 4.5.2.

rvandermeulen:
Includes SQLite 3.39.2.


bendk (approved):
Seems good to me.  Note that we're only using SQLCipher to migrate old databases at this point.  Still, an upgrade doesn't seem bad to me.

rvandermeulen:
Pretty aggressive for Mergify to call a review stale because a PR gets rebased to main.

bendk (approved):

PR #5151, opened 2022-09-20
Don't dismiss stale reviews for the release-management team

mhammond:
We don't allow mergify to merge stale reviews for reasons noted in 4185342a01a68925722ed815b56e57337fc7dbcd. We might be able to fix that, but we don't want to regress what that commit fixed, so it might not be trivial.

Combining this with the fact that we reset approvals on rebase for many people, it leave @rvandermeulen in a state where he almost always needs to get multiple approvals before landing anything.

This patch means stale reviews for the release-management team should not be automatically reset. It's a copy-paste and I'm not sure how to test it other than to land it! Any thoughts/concerns?

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -15,6 +15,7 @@ pull_request_rules:
       - base=main
       - author!=@mozilla/application-services
       - author!=@mozilla/application-services-collaborators
+      - author!=@mozilla/release-management
     actions:
       dismiss_reviews:
         message: The pull request has been modified, dismissing previous reviews.


rvandermeulen:
I don't have objections to this on a higher level, but I'm not nearly comfortable enough with mergify configs to discern whether this is the right/optimal approach. It's possible that @jcristau or @JohanLorenzo may have some thoughts as they're more familiar with the mergify configs we use in the mobile repos which seem to work well enough.

jcristau:
This looks fine to me, FWIW.  The mobile repos I know best (fenix/focus-android) don't dismiss reviews, but they also don't merge PRs from non-collaborators so it seems reasonable that the tradeoffs are different.

JohanLorenzo (approved):
The rationale and the implementation both look good to me.

> It's a copy-paste and I'm not sure how to test it other than to land it! Any thoughts/concerns?

I faced the same testing issue a couple of months ago. The way I found was to:
 1. Fork the repo
 2. Install the mergify app on my fork https://github.com/apps/mergify/installations/new
 3. Create pull requests on my fork and see what mergify does.

If Mergify doesn't behave the way you expect, it's exposes a Github check like this one ⬇️ 

image

You can click on it and it will show the state of each condition. Thus, in your particular case, you can define a Github team on your fork, make your account join one or another team and see how it goes.

That said, your current change is fairly straightforward and has low impact. So, landing it and synchronizing with @rvandermeulen seems a sane and simpler solution.

rvandermeulen (approved):
Thanks for the consults! If the RelEng folks are happy about it, so am I!

PR #5152, opened 2022-09-20
Downgrade androidx.core to 1.3.2

rvandermeulen:
As noted on Slack, updating to version 1.8.0 is causing issues with Fenix which still uses 1.3.2. Downgrade back to 1.3.2 for now until the 1.8.0 bump lands downstream. Also, add a note about the dependency to hopefully avoid this happening in the future.


bendk:
Forced-pushed the same change with a different commit id so that I can test if this fixes the branch builds.

bendk (approved):
Nice, looks like that fixes things.

PR #5155, opened 2022-09-20
Change NDK version to 25 where missing [ci full]

tarikeshaq:
Testing to see if this would fix the full megazord builds


tarikeshaq:
eeeh looks like some other failures are here and #5154 is working, so closing this 😄

rvandermeulen:
Yeah, those look like the Android target API level, not related specifically to the NDK version (just a coincidence that both were 21)

PR #5156, opened 2022-09-21
Allow tabs to use uniffi 0.19 or 0.20

mhammond:
This seems to allow us to vendor into m-c using 0.19


skhamis (approved):
Glean is still on v0.19 which is why we can't just simply bump m-c to v0.20 --> This is necessary for tabs for now. LGTM!

Issue #5092, opened 2022-08-17
Migrate autofill to using the new error system


bendk:
We recently started using the new error handling system from `support::error` for logins and it seems to be working okay.  The next step is to switch the other components to also use this system.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3269)
┆epic: Error handling
┆sprintEndDate: 2022-10-28

PR #5162, opened 2022-09-22
Refactored places to use the new error handling system [#5092]

bendk:
Replaced the error_support method with the new system that consists of:
  - `PlacesInternalError`: used internally in the crate.  This replaces the old `ErrorKind` enum and is the error for `Result<>`.
  - `PlacesError`: used for external errors that we return to UniFFI consumers.  This replaces the old `Error` enum and is the error for `ApiResult<>`
  - A `GetErrorHandling` impl to convert internal errors to external errors and report to sentry.  For now, I tried to make the sentry error reporting more-or-less match the old error reporting, which means reporting lots of internal errors.  The plan is going to be to check sentry and most likely turn some of those into logs rather than errors.

Reworked all top-level API methods to return `PlacesApiResults` and use the `handle_error!` macro to do the error conversion/reporting.

Use `thiserror` to handle converting errors from other crates into `PlacesInternalError`.

`places.udl` didn't have to change any and AFAICT, this does not change the external API at all.

  - Note:


bendk:
This PR doesn't have any breaking changes (I think), but there are a couple that could be useful:
  - Change the variants of `PlacesError` so that they are more likely to be useful for the consumer apps.  I think this would mean:
     - Merging `UrlParseFailed`/`UrlTooLong` and `UnexpectedPlacesException`/`InternalError`, since they mean basically the same thing to a a consumer.
     - Remove the `JsonParseFailed` and move those errors into another variant (maybe `UnexpectedPlacesException`?).  I don't think consumers should have to care about low-level JSON errors.
  - Maybe rename `PlacesError` to `PlacesApiError` and `PlacesInternalError` to `PlacesError`.  Those names seem more descriptive to me, and it makes sense to have `PlacesError` be the internal error since that's usually what we're dealing with in our code.

Does it make sense to extend this PR to also introduce those changes?  If we did that, should we also consider doing breaking changes to the tabs/logins errors?

bendk:
Just updated this one and pushed a couple more commits.  The new commits rename some errors and combine some variants in a way that I think makes more sense.  Is this actually an improvement?  Is it worth the breaking change?  I put the changes in separate commits, so they should be easy to back out if we decide to do that.

components/places/ios/Places/Places.swift
```diff
@@ -339,17 +339,17 @@ public class PlacesReadConnection {
...
-     *     - `PlacesError.unexpected`: When an error that has not specifically been exposed
-     *                                 to Swift is encountered (for example IO errors from
-     *                                 the database code, etc).
-     *     - `PlacesError.panic`: If the rust code panics while completing this
-     *                            operation. (If this occurs, please let us know).
+     *     - `PlacesApiError.databaseInterrupted`: If a call is made to `interrupt()` on this
+     *                                             object from another thread.
+     *     - `PlacesConnectionError.connUseAfterAPIClosed`: If the PlacesAPI that returned this connection
+     *                                                      object has been closed. This indicates API
+     *                                                      misuse.
```
bendk, line 346: I'm not sure how these doc strings should be formatted.  Now that the errors use a couple extra chars, if you align everything after the colon, there's very little space to use.

mhammond:
I'm actually mildly in favour of splitting this back out into multiple PRs. The first one isn't breaking, lays a good foundation and can be landed without interacting with the mobile teams - that looks completely uncontroversial although I do think the copy/paste of the `XXX` comments re `thiserror` should instead be "resolved" (ie, either remove the use of thiserror in both this crate and from logins or remove the comment - ideally the former :) and the comment about `into` now looks stale in both places (and is my fault for leaving it in the logins PR)?

Then we can consider the breaking changes a little more holistically - eg, you mentioned combining the url errors into a single error, which probably makes sense, but isn't done here, and I suspect a little more digging might show that other errors, like the JsonError, can also be removed or made internal. That breaking change can then just linger until the mobile teams are ready to take it (eg, maybe iOS isn't going to be ready to take it until much later?) and it seems a shame to have the core PR wait until then. A PR that talks about refactoring places but also does refactoring  touches logins seems odd.

OTOH though, I'm happy to be convinced that taking it now is fine - it is less work and of the mobile PRs are trivial, then I've no objection to landing now and making incremental improvements over time.

bendk:
> I'm actually mildly in favour of splitting this back out into multiple PRs.

Makes sense to me.  I moved the breaking parts into #5166.

I guess the last thing to discuss here is the `thiserror` use for the public logins/places errors.  I personally like it, since it's a way to concisely implement `Error` and define display strings for the errors.  If that was the only error enum then I wouldn't think it was worth cost of an the extra dependency and requiring readers to know the `thiserror` macro syntax.  But given that we're also using them for our internal errors, it doesn't seem so bad.  That's a very weakly held preference though.

lougeniaC64:
> I personally like it, since it's a way to concisely implement `Error` and define display strings for the errors.

Same. I've reviewed this PR and I'm prepared to approve but don't want to do so until the `thisError` convo is resolved.

bendk:
@mhammond I think all that's left is the `thiserror` question, what do you think about it?

lougeniaC64 (approved):

PR #5166, opened 2022-10-04
Error handling breaking changes

bendk:
Breaking changes to the error handling for several components in order to make the API cleaner.  This was split off from #5162.

There are a bunch of changes here and I think it should make things simpler for our consumers, but I don't know enough about how our components are used to be at all certain about that.  I'd love feedback on any of the changes and I'll try to flag some specific questions with comments.

  - Note:

### DIFF
--- a/components/autofill/src/autofill.udl
+++ b/components/autofill/src/autofill.udl
@@ -2,16 +2,16 @@ namespace autofill {
     // We expose the crypto primitives on the namespace
 
     // Create a new, random, encryption key.
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     string create_key();
 
     // Encrypt an arbitrary string - `key` must have come from `create_key()`
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     string encrypt_string(string key, string cleartext);
 
     // Decrypt an arbitrary string - `key` must have come from `create_key()`
     // and `ciphertext` must have come from `encrypt_string()`
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     string decrypt_string(string key, string ciphertext);
 };
 
@@ -80,53 +80,55 @@ dictionary Address {
 };
 
 [Error]
-enum AutofillError {
-   "OpenDatabaseError", "SqlError", "IoError", "InterruptedError",
-   "IllegalDatabasePath", "Utf8Error", "JsonError", "InvalidSyncPayload",
-   "MissingEncryptionKey", "CryptoError", "NoSuchRecord",
+interface AutofillApiError {
+    SqlError(string reason);
+    InterruptedError();
+    CryptoError(string reason);
+    NoSuchRecord(string guid);
+    UnexpectedAutofillApiError(string reason);
 };
 
 interface Store {
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     constructor(string dbpath);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     CreditCard add_credit_card(UpdatableCreditCardFields cc);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     CreditCard get_credit_card(string guid);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     sequence<CreditCard> get_all_credit_cards();
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     void update_credit_card(string guid, UpdatableCreditCardFields cc);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     boolean delete_credit_card(string guid);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     void touch_credit_card(string guid);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     Address add_address(UpdatableAddressFields a);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     Address get_address(string guid);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     sequence<Address> get_all_addresses();
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     void update_address(string guid, UpdatableAddressFields a);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     boolean delete_address(string guid);
 
-    [Throws=AutofillError]
+    [Throws=AutofillApiError]
     void touch_address(string guid);
 
-    [Throws=AutofillError, Self=ByArc]
+    [Throws=AutofillApiError, Self=ByArc]
     void scrub_encrypted_data();
 
     [Self=ByArc]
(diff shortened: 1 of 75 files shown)


components/autofill/src/error.rs
```diff
@@ -44,4 +70,82 @@ pub enum Error {
...
+        // WARNING: The details inside the `AutofillApiError` we return should not
+        // contain any personally identifying information.
+        // However, because many of the string details come from the underlying
+        // internal error, we operate on a best-effort basis, since we can't be
+        // completely sure that our dependencies don't leak PII in their error
+        // strings.  For example, `rusqlite::Error` could include data from a
+        // user's database in their errors, which would then cause it to appear
+        // in our `AutofillApiError::SqlError` structs, log messages, etc.
+        // But because we've never seen that in practice we are comfortable
+        // forwarding that error message into ours without attempting to sanitize.
```
bendk, line 88: I copied the PII note from logins to each component since it seems useful and I'm afraid if I put it in some centralized place like the error-support README it would get overlooked.  But maybe it's overkill, WDTY?

components/autofill/src/error.rs
```diff
@@ -44,4 +70,82 @@ pub enum Error {
...
+        match self {
+            Self::OpenDatabaseError(e) => ErrorHandling::convert(AutofillApiError::SqlError {
+                reason: e.to_string(),
+            })
+            .report_error("autofill-open-database-error"),
```
bendk, line 93: I erred on the side of making error reports for everything, just so that we can get some insight into how many errors we're getting.  After this is merged and in use on mobile, we should definitely check out the errors on sentry and consider turning some of this into `log_warning` or `log_info`.

components/autofill/src/autofill.udl
```diff
@@ -80,53 +80,55 @@ dictionary Address {
...
 [Error]
-enum AutofillError {
-   "OpenDatabaseError", "SqlError", "IoError", "InterruptedError",
-   "IllegalDatabasePath", "Utf8Error", "JsonError", "InvalidSyncPayload",
-   "MissingEncryptionKey", "CryptoError", "NoSuchRecord",
+interface AutofillApiError {
+    SqlError(string reason);
+    InterruptedError();
+    CryptoError(string reason);
+    NoSuchRecord(string guid);
```
bendk, line 87: Should all error variants end in "Error"?  It seems maybe more idomatic, but I'm not sure.

components/tabs/src/sync/bridge.rs
```diff
@@ -192,65 +218,86 @@ impl TabsBridgedEngine {
...
+        &self,
+        incoming: Vec<String>,
+    ) -> ApiResult<Vec<IncomingEnvelope>> {
+        handle_error! {
+            let mut envelopes = Vec::with_capacity(incoming.len());
+            for inc in incoming {
+                envelopes.push(serde_json::from_str::<IncomingEnvelope>(&inc)?);
+            }
+            Ok(envelopes)
         }
```
bendk, line 260: This is a papercut from using the `handle_error!` macro.  It would be simpler if this code was in `store_incoming()` since that's the only place it's used.  However, I couldn't get the `handle_error!` macro to work there because it needs to wrap the body of the entire function, but that didn't work because `self.bridge_impl.store_incoming()` is already returning an `ApiResult`.

It's a bit annoying, but the fact that this is the biggest issue that I've seen with that macro seems to indicate it's working out well.

components/support/error/README.md
```diff
@@ -0,0 +1,73 @@
...
+hope to allow expected errors to be counted in telemetry (think things like
+network errors, shutdown errors, etc.).
+
+To assist this conversion, the `handle_error!` macro can be used to
+automatically convert between `Result` and `ApiResult` using
+`GetErrorHandling`.  Note that this depends on having the `Result` type
+imported in your module with a `use` statement.
+
+See the `logins::errors` and `logins::store` modules for an example of how this
+all fits together.
```
mhammond, line 73: love these docs, thanks!

components/autofill/src/error.rs
mhammond: I'm torn on this too :) Also torn whether it's better here or near the actual public enum?

An alternative might be to move some of the text into the new error doc you wrote and in each component make the note much shorter and link to that doc? I'm fine either way though.

mhammond (approved):
This is great, thanks! I think more breaking changes do need noting, and I don't think we should land this until we have PRs available for ios and Android, but this seems ready otherwise.

components/autofill/src/autofill.udl
bendk: Let's just leave this alone then, it seems like it's not important and there's no clear right answer.

components/autofill/src/error.rs
bendk: This makes a lot of sense to me, since there's also other information that could be useful there.

bendk:
I also removed the `reporting` feature from error-support and made the feature always enabled.  It was added there to [make some moz-central vendoring easier](https://github.com/mozilla/application-services/pull/4987), but I think/hope that it's not needed anymore.

bendk:
@mhammond I added an extra commit here, does it look okay?  https://github.com/mozilla/application-services/pull/5166/commits/903e0cec6523770ab812c0b1459e5b2b4fdefdbe

components/autofill/src/db/store.rs
```diff
@@ -48,10 +49,12 @@ pub struct Store {
 }
 
 impl Store {
-    pub fn new(db_path: impl AsRef<Path>) -> Result<Self> {
-        Ok(Self {
-            db: Mutex::new(AutofillDb::new(db_path)?),
-        })
+    pub fn new(db_path: impl AsRef<Path>) -> ApiResult<Self> {
+        handle_error! {
```
tarikeshaq, line 53: I like this pattern, and the more I see it I wonder if we could have a proc macro attribute like:

```rs
#[handle_err(AutofillApiError)]
pub fn new(db_path: impl AsRef<Path>) -> Result<Self> {
    Ok(Self {
        db: Mutex::new(AutofillDb::new(db_path)?),
    })
}
```

That has the same semantics as the `handle_error!` - it's orthogonal to this PR but thought it would be interesting to think out loud

tarikeshaq (commented):
Looked through a very small subset of this so far, but it looks good! Partially didn't look through the rest because it's mechanical and partially because it's a lot to look at 😅 

Had a tiny marginally related thought

components/autofill/src/db/store.rs
bendk: Yeah, I think that could a nice improvement.

One question is how should the return type read should this be `-> Result<>` or `-> ApiResult<>`?  The code itself is creating a `Result<>`, but then the proc-macro is turning it into an `ApiResult<>`.

However we answered that, I think it would be better than the current macro which I think assumes there's a `Result<>` type imported in the macro and will error out in a hard to understand way if there's not.

tarikeshaq: I'm indifferent - If we had it as a `Result` then _maybe_ we don't need an `ApiResult`? maybe the macro can replace whatever the function returns with a `Result<T, AutofillApiError>` since proc macro attributes can modify function tokens 🤷 

Regardless it'll might also nice because we can also add some error handling to print nicer messages too

tarikeshaq: Created https://github.com/mozilla/application-services/issues/5183 in case we want to do this!

mhammond:
> [903e0ce](https://github.com/mozilla/application-services/commit/903e0cec6523770ab812c0b1459e5b2b4fdefdbe)

Looks great!

components/autofill/src/db/store.rs
tarikeshaq: #5184 😬

PR #5167, opened 2022-10-05
 Switch from deprecated digital.wup.android-maven-publish to maven-publish plugin

rvandermeulen:
This was done in AC a long time ago. I don't think we need to do anything more other than changing the name given the limited usage in this case, but some local testing would be appreciated by someone with a working environment.


bendk (approved):
Rubber stamping this one...

PR #5169, opened 2022-10-05
Update some Gradle dependencies to newer releases

rvandermeulen:


mhammond (approved):
I can't find where the kotlin version is specified in a-c or fenix, but I assume this is just bringing us into line with them, so rubberstamp=me

rvandermeulen:
Did the full AC+Fenix CI run on the PR just to be safe and nothing broke.

PR #5172, opened 2022-10-10
Sync Manager syncs engines in a deterministic order, fixes #5171

### LINKED ISSUE #5171: Syncing should use a deterministic order for the engines
Our sync manager API allows for a list of engines to be specified, or if no engines are specified, we sync all registered engines. However, the order we sync them in is not deterministic and it should be - in particular, we should use a similar order to what's done on desktop (basically, tabs first, history last)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3351)

mhammond:
Before this patch, Android logs show:
```
I  Syncing 5 engines
I  Synchronizing clients engine
I  Synchronizing engines
I  Syncing passwords engine!
I  Syncing bookmarks engine!
I  The creditcards engine is declined. Skipping
I  Syncing history engine!
I  Syncing tabs engine!
I  Finished syncing engines.
```
and with it:
```
I  Syncing 5 engines
I  Synchronizing clients engine
I  Synchronizing engines
I  Syncing passwords engine!
I  Syncing tabs engine!
I  Syncing bookmarks engine!
I  The creditcards engine is declined. Skipping
I  Syncing history engine!
I  Finished syncing engines.
```

@lougeniaC64, do you happen to know the order Firefox for iOS uses?

### DIFF
--- a/components/sync15/src/engine/sync_engine.rs
+++ b/components/sync15/src/engine/sync_engine.rs
@@ -31,24 +31,30 @@ pub enum EngineSyncAssociation {
 /// The concrete `SyncEngine` implementations
 #[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SyncEngineId {
+    // Note that we've derived PartialOrd etc, which uses lexicographic ordering
+    // of the variants. We leverage that such that the higher priority engines
+    // are listed first.
+    // This order matches desktop.
     Passwords,
-    History,
-    Bookmarks,
     Tabs,
+    Bookmarks,
     Addresses,
     CreditCards,
+    History,
 }
 
 impl SyncEngineId {
-    // Iterate over all possible engines
+    // Iterate over all possible engines. Note that we've made a policy decision
+    // that this should enumerate in "order" as defined by PartialCmp, and tests
+    // enforce this.
     pub fn iter() -> impl Iterator<Item = SyncEngineId> {
         [
             Self::Passwords,
-            Self::History,
-            Self::Bookmarks,
             Self::Tabs,
+            Self::Bookmarks,
             Self::Addresses,
             Self::CreditCards,
+            Self::History,
         ]
         .into_iter()
     }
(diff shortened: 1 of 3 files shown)


lougeniaC64:
> @lougeniaC64, do you happen to know the order Firefox for iOS uses?

This is an abbreviated version of the iOS logs when you sync all engines: 

```
2022-10-10 11:26:58.715 [Debug] [Profile.swift:1264] syncWith(synchronizers:statsSession:why:) > Syncing clients…
2022-10-10 11:26:58.852 [Debug] [Profile.swift:1264] syncWith(synchronizers:statsSession:why:) > Syncing tabs…
2022-10-10 11:26:59.590 [Debug] [Profile.swift:1264] syncWith(synchronizers:statsSession:why:) > Syncing bookmarks…
2022-10-10 11:27:00.036 [Debug] [Profile.swift:1264] syncWith(synchronizers:statsSession:why:) > Syncing history…
2022-10-10 11:27:00.289 [Debug] [Profile.swift:1264] syncWith(synchronizers:statsSession:why:) > Syncing logins…
```
which is set in [syncEverything](https://github.com/mozilla-mobile/firefox-ios/blob/main/Providers/Profile.swift#L1294).

mhammond:
Thanks! That looks perfect for tabs - we should possibly consider making logins earlier than history and bookmarks though, but I don't think it's a huge deal.

components/sync15/src/engine/sync_engine.rs
```diff
@@ -31,24 +31,28 @@ pub enum EngineSyncAssociation {
...
     pub fn iter() -> impl Iterator<Item = SyncEngineId> {
         [
             Self::Passwords,
-            Self::History,
-            Self::Bookmarks,
             Self::Tabs,
+            Self::Bookmarks,
             Self::Addresses,
             Self::CreditCards,
+            Self::History,
```
bendk, line 55: Maybe add a note here that this should be kept in-sync with the enum ordering.

bendk (approved):
Looks good to me, just had a small documentation suggestion.

lougeniaC64:
> Thanks! That looks perfect for tabs - we should possibly consider making logins earlier than history and bookmarks though, but I don't think it's a huge deal.

I'm not sure of the severity if this issue so is it worth it to implement this in iOS before the sync manager integration?

lougeniaC64 (approved):

mhammond:
> I'm not sure of the severity if this issue so is it worth it to implement this in iOS before the sync manager integration?

Nope, it can wait.

components/sync15/src/engine/sync_engine.rs
mhammond: I just did that for consistency, but I don't know if it matters - should we guarantee this order is sorted, or should anyone who relies on that actually sort?

bendk: I maybe slightly prefer guaranteeing the order is sorted, but I can't see a strong reason one way or the other.

mhammond: yeah, fair enough - I added a comment that it's a policy decision and a test to ensure it remains true.

Issue #5165, opened 2022-09-27
94.3.0 release has a linkage problem in some devices


tarikeshaq:
https://sentry.io/organizations/mozilla/issues/3527488899/?query=is%3Aunresolved+load+library&statsPeriod=14d

> Unable to load library 'megazord': dlopen failed: cannot locate symbol "__extenddftf2" referenced by 
> Native library (android-x86-64/libmegazord.so) not found in resource path (.)

We are suspicious it was the NDK upgrade in https://github.com/mozilla/application-services/pull/5142 but we are not certain since we can't reproduce locally yet

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3340)
┆Epic: Prioritized from backlog
┆Sprint End Date: 2023-03-10

cpeterso:
Here is the Bugzilla bug tracking these crash reports in Socorro: https://bugzilla.mozilla.org/show_bug.cgi?id=1792799

rvandermeulen:
Thanks for that, Chris. Socorro definitely confirms that the revert to 94.2.1 resolved these crashes for now at least.

rvandermeulen:
Maybe related:
https://github.com/android/ndk/issues/1614

rvandermeulen:
Looking at https://developer.android.com/ndk/guides/common-problems, I also wonder if Gecko's still being on r21 is possibly relevant.

jonalmeida:
> Looking at https://developer.android.com/ndk/guides/common-problems, I also wonder if Gecko's still being on r21 is possibly relevant.

It shouldn't. The crash is not happening when there is any interaction between Gecko and app services code.

bendk:
I was able to reproduce this one on an emulator and confirmed that backing out the NDK upgrade fixed my issue there ([details](https://github.com/mozilla/application-services/pull/5176)).

I'm going to make a `94.3.2` release today, what should the next steps be after that?  I'm thinking we should maybe wait until the next release cycle then introduce it back into nightly?

bendk:
I think we're running into something like https://github.com/mozilla/rust-android-gradle/issues/105.

I see this error in logcat: `AndroidRuntime: Native library (android-x86-64/libmegazord.so) not found in resource path (.)`  However, `libmegazord.so` is definitely in the APK.  I see it in the `lib/x86_64/`  directory when I unzip it.

@ncalexan any ideas on how we can fix this

[...]

PR #5177, opened 2022-10-12
Reverting the Android NDK upgrade (#5165) [ci full]

bendk:
The CI in main is currently not working, so let's test this out in the branch instead.

I tested v94.3.0 using the x86_64-33_r06 image and it crashed with the Unable to load library 'megazord': error. Reverting the NDK commit stopped the crash, so I think the hypothesis is correct.

I also reverted my objcopy commit, which I think depended on the NDK upgrade. I'm building the full CI to confirm this.

This patch is v94 if we merge it, I'll make a PR to merge it back to main.

What's a good way to test this?  Is fixing things in my emulator enough, or should we try to reproduce on actual hardware? 

  - Note:


tarikeshaq (dismissed):
In case you needed a review here too 😄 - maybe worth it to just land the main PR so we don't clutter this branch unless we have to (meaning once we are on 94.4 or higher and want to cut a backport)

We can just cut 94.3.2 on main with the fix and that should be good enough?

bendk:
> We can just cut 94.3.2 on main with the fix and that should be good enough?

Yeah that makes sense to me.  Let's close this one.

PR #5182, opened 2022-10-13
Fixes missing titles in iOS history migration

tarikeshaq:
Fixes the bug where sometimes the titles would be missing.

The issue comes from history metadata. When an item is inserted there, sometimes it's without a title (this is really a bug in the iOS implementation)

However, since we want to re-use `mozplaces` rows, we ended up not updating those rows with new titles. This PR makes sure that we go through any older mozplaces rows that don't have titles and sets them to be the appropriate values

### DIFF
--- a/components/places/src/import/ios/history.rs
+++ b/components/places/src/import/ios/history.rs
@@ -9,9 +9,9 @@ use crate::history_sync::engine::LAST_SYNC_META_KEY;
 use crate::import::common::{
     attached_database, define_history_migration_functions, select_count, HistoryMigrationResult,
 };
-use crate::storage::history::update_frecencies;
-use crate::storage::put_meta;
+use crate::storage::{put_meta, update_all_frecencies_at_once};
 use crate::PlacesDb;
+use types::Timestamp;
 use url::Url;
 
 /// This import is used for iOS users migrating from `browser.db`-based
@@ -77,8 +77,9 @@ fn do_import(
     conn.execute_batch(&INSERT_HISTORY_VISITS)?;
     scope.err_if_interrupted()?;
 
-    log::info!("Updating frecencies");
-    update_frecencies(conn)?;
+    log::info!("Insert all new entries into stale frecencies");
+    let now = Timestamp::now().as_millis();
+    conn.execute(&ADD_TO_STALE_FRECENCIES, &[(":now", &now)])?;
     scope.err_if_interrupted()?;
 
     // Once the migration is done, we also migrate the sync timestamp if we have one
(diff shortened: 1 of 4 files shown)


components/places/src/import/ios/history.rs
```diff
@@ -132,6 +136,15 @@ lazy_static::lazy_static! {
         WHERE url IS NOT NULL"
    ;
 
+    // Unfortunately UPDATE FROM is not available until sqlite 3.33
+   // however, iOS does not ship with 3.33 yet as of the time of writing.
+   static ref UPDATE_PLACES_TITLES: &'static str =
```
tarikeshaq, line 141: I'm looking for more efficient ways to do this... ideally it would be in the same `INSERT OR IGNORE` below it, but that we want to ignore any issues there.

`INSERT OR REPLACE` would be cool, but we don't want to be deleting old entried and inserting new ones since lots of tables depend on the values of the places table..

This now isn't too bad, and is very far from being the bottleneck (I intentionally do it before the `FILL_MOZ_PLACES` because if we do it after, the joins will be painful and it will become the bottleneck)

components/places/src/storage/history.rs
```diff
@@ -155,6 +155,18 @@ pub fn apply_observation_direct(
     Ok(visit_row_id)
 }
 
+pub fn update_frecencies(db: &PlacesDb) -> Result<()> {
```
tarikeshaq, line 158: This also isn't ideal.. and doesn't **need** to happen during the migration, also looking for more efficient ways. I'm thinking a possibility is to even have it run async on its own, or instead, set an arbitrary non `-1` value 🤷 

I added it because with `-1` frecency the queryAutcomplete is acting weird and not even displaying values correctly (it's not searching the titles and only searching the URLs)

mhammond (approved):
This looks fine to me assuming my questions don't really make sense or wouldn't be an improvement in practice :)

components/places/src/import/ios/history.rs
```diff
@@ -132,6 +136,15 @@ lazy_static::lazy_static! {
...
+   "UPDATE main.moz_places
+        SET title = ( SELECT t.title
+                            FROM temp.iOSHistoryStaging t
+                            WHERE t.url = main.moz_places.url AND t.url_hash = main.moz_places.url_hash )"
```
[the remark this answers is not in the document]
tarikeshaq: From my testing the titles in the browserdb will be a safer call. I've seen cases where twitter would have a `Twitter` title in our places database, but have the correct `Twitter. It's what's happening / Twitter` in the browserdb. I'm yet to see a case where browserdb have a less fresh title

components/places/src/storage/history.rs
tarikeshaq: Yeah, that logic seems precisely what I want.

One thing to consider though is that iOS only allows there to exist one writer connection, so writers will actually be blocked as the whole `moz_places_stale_frecencies` process runs. This is unfortunate because the bookmark syncing logic was chunked so that other writer connections could sneak in. 

That said, as long as it's in a different transaction, the reader connection can go ahead and read the history (although with stale frecencies which is OK since they'll be updated anyways)

I'll change this so it still runs as a part of the migration process, but it runs in a separate transaction properly

lougeniaC64 (approved):
Sorry for the delay. I conflated #5181 with this one somehow.

PR #5184, opened 2022-10-14
Make handle_error as a proc macro

tarikeshaq:
~~A bit of spaghetti to show what `handle_error` could look like as a proc_macro, happy to drop this if we don't like it!~~

~~The proc macro itself could use some work, but I'm happy with what the consuming code is looking like, and the errors we get from the compiler! (which we can make even smarter)~~

PR is updated so it's less spaghetti and has tests to cover some errors


components/logins/src/encryption.rs
```diff
@@ -83,23 +83,20 @@ impl EncryptorDecryptor {
...
 
-pub fn create_key() -> ApiResult<String> {
-    handle_error! {
-        let key = jwcrypto::Jwk::new_direct_key(None)?;
-        Ok(serde_json::to_string(&key)?)
-    }
+#[handle_error(LoginsStorageError)]
+pub fn create_key() -> Result<String> {
+    let key = jwcrypto::Jwk::new_direct_key(None)?;
+    Ok(serde_json::to_string(&key)?)
```
bendk, line 99: I think this is overall a lot cleaner, but my one concern is that it's silently changing the result type.  After `handle_error` is done doing it's wrapping, this will return a `ApiResult<String>`, but you wouldn't think that based on the signature.

I'm not sure this is a great solution, but what if we reversed things?  Then this code would be:
```
#[handle_error(LoginsError)]
pub fn create_key() -> ApiResult<String> {
    let key = jwcrypto::Jwk::new_direct_key(None)?;
    Ok(serde_json::to_string(&key)?)
}
```

That seems a bit more clear to me, although still slightly magical.

tarikeshaq: Yeah there is a balance here of where to put the magic
1. In the one case, the function you write returns `ApiResult` but you wrote it to return `Result` and the signature says `Result`
2. in the other case, the function returns `ApiResult` (and so does the signature) but you have to return `Result` in its body

neither is perfect and I'm happy to change it to be the reverse!

What do you think about changing the name to something like `#[expose_error_as(LoginsStorageError)]`

That way we keep option 1 (i.e the way it is in this PR) and the magic is slightly alleviated because the macro tells us that we are mapping the error

```rs
#[expose_error_as(LoginsStorageError)]
pub fn create_key() -> Result<String> {
    let key = jwcrypto::Jwk::new_direct_key(None)?;
    Ok(serde_json::to_string(&key)?)
}
```

tarikeshaq:
If the approach in #5191 makes sense, then I can trash this! The PR there seems like a more powerful refactor of the `handle_error` pattern

components/logins/src/encryption.rs
bendk: I think 1 could work with a nice name, but I'm still not sure what that is.  One issue for me is that the macro is doing several things: coverting, logging, and reporting errors.  The other thing is that by converting the errors, you're changing the `Result<>` type you're returning.  "expose error" doesn't communicate that to me, but I'm struggling to suggest something better. 

 What if we went more verbose?  Something like `#[handle_error_and_convert_to_api_result(LoginsError)]`.

tarikeshaq:
For the sake of moving this forward and closing this PR, I switched `handle_error` to be the **exact same semantics** before it was a proc macro. 

This means that the wins for this PR are:
- Better error messages when used incorrectly, see the `.err` files
- Added documentation for when developers use the proc macro, they can understand the side effects
- It's now a proc macro that's placed top of the function as opposed to within it so it becomes more or less a "decoration" on the function (which seems to me like a win 🤷)

Otherwise kept the naming, and the semantics the same, including the assumption the original macro makes about the environment it's used in

components/support/error/macros/src/signature.rs
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+use syn::{spanned::Spanned, GenericArgument, PathArguments};
+
+const WRONG_RETURN_TYPE_MSG: &str = "Expected Result<T> or Result<T, E>";
+
+pub(crate) fn validate(sig: &syn::Signature) -> syn::Result<()> {
```
bendk, line 8: I think this is checking if the return type has 1 generic argument -- is this right?  I think this should be documented.

Or maybe it should be removed.  I'm not sure what it's guarding against and I'm slightly worried that type aliases could cause errors.

components/support/error/macros/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+///
+/// #[handle_error]
+/// fn do_something() -> std::result::Result<String, ExternalError> {
+///    // The `handle_error` macro maps from `Error` to `ExternalError`
+///    Err(Error{})
+/// }
+///
+/// // The error here is an `ExternalError`
+/// let _: ExternalError = do_something().unwrap_err();
+/// ```
```
bendk, line 61: Love these docs.

components/support/error/tests/returns_result_but_incorrect_error.stderr
```diff
@@ -0,0 +1,11 @@
+error[E0308]: mismatched types
+  --> tests/returns_result_but_incorrect_error.rs:50:2
+   |
+50 |  #[handle_error]
+   |  ^^^^^^^^^^^^^^^ expected struct `ExternalError`, found struct `OtherExternalError`
+51 |  fn func() -> Result<String, ExternalError> {
+   |               ----------------------------- expected `std::result::Result<String, ExternalError>` because of return type
```
bendk, line 7: This seems like the most important check on the return type.  How is it implemented?  Is there a way to get a bit more text, like `expected struct ExternalError, because body returns a type that implements GetErrorHandling<E = ExternalError>, found struct OtherExternalError instead`

bendk (changes requested):
Overall this looks great, and the trybuild tests are awesome.  I had some questions and maybe a suggestion about the return type validation though.

components/support/error/macros/src/signature.rs
tarikeshaq: That's a good point, I think we can do without this check and leave it to the Rust compiler to error if the return types are incorrect!

components/support/error/tests/returns_result_but_incorrect_error.stderr
tarikeshaq: It's delegated to the Rust compiler (i.e It's not implemented by the macro). We can try to be smart about it and implement it, but we'd need the macro to "know" the body's return type (as an argument for example). But even then it'll be complicated for us at the proc macro level to understand if the type the body returns implements `GetErrorHandling`. 

For now, leaving it to the compiler and having those tests verifying the errors are more or less understandable is a good start

bendk (approved):
Looking great.

PR #5191, opened 2022-10-15
Use `Into` trait instead of a new trait for error conversions

tarikeshaq:
A possible small improvement for the error handling work, using `Into<ErrorHandling<ExternalError>>` instead of a new `GetErrorHandling` trait

Thoughts?

(This could be useful because with this, and removing the `reporting` feature, we can probably completely get rid of  the `convert_log_report_error` function and inline it in the macros!)

### DIFF
--- a/components/logins/src/error.rs
+++ b/components/logins/src/error.rs
@@ -7,8 +7,8 @@ pub type Result<T> = std::result::Result<T, LoginsError>;
 // Functions which are part of the public API should use this Result.
 pub type ApiResult<T> = std::result::Result<T, LoginsStorageError>;
 
+use error_support::ErrorHandling;
 pub use error_support::{breadcrumb, handle_error, report_error};
-use error_support::{ErrorHandling, GetErrorHandling};
 use sync15::Error as Sync15Error;
 
 // Errors we return via the public interface.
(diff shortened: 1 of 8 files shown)


components/logins/src/error.rs
```diff
@@ -180,9 +178,11 @@ impl GetErrorHandling for LoginsError {
...
             _ => ErrorHandling::convert(LoginsStorageError::UnexpectedLoginsStorageError(
-                self.to_string(),
+                e.to_string(),
             ))
             .report_error("logins-unexpected"),
         }
     }
 }
+
+expose_error!(LoginsError as LoginsStorageError);
```
tarikeshaq, line 188: The idea is instead of doing `handle_error!` on every function, consuming crates would just do this macro once, then it automatically defines a `From<InternalError> for ExternalError` 

it even defines a blanket statement so any types that can be converted into an InternalError can also be converted into an External error by first being converted to the InternalError!

This is the least "magic" way I could think of, and if this makes sense I would trash the #5184

tarikeshaq:
The second commit goes beyond a
> A possible small improvement

And offers a nice refactor to get rid of `handle_error` without having to define a different kind of macro!

components/logins/src/error.rs
bendk: I think we discussed something like this for the initial work and our feeling was that it actually more "magic" than the others, since the `Into` impl was also doing things like logging and reporting errors.  My feeling is that it was a very close call though, so we could revisit that.

tarikeshaq: Heh yeah I kinda remember that 😬 I think once we close out https://mozilla-hub.atlassian.net/browse/SYNC-3318 it'll be easy for me not to reopen closed-out discussions 🤦 So to summarize:
1. We'd rather have an explicit trait than use `Into` for conversion from an `InternalError` to an `ErrorHandling<ExternalError>`
1. We'd rather have an explicit macro than use `Into` for conversion from an `InternalError` to an `ExternalError` (using the explicit trait in 1)

If so maybe I'll take this opportunity to write out some of this on the trait and macro themselves!

bendk: I think that's basically correct, but I would reword it slightly.  I think using a trait for an implicit conversion is fine and normal Rust error handling.  The issue is when that same conversion function also does error reporting/logging, since that could be very surprising.

tarikeshaq:
I'll close this, might put some time into the #5184 but regardless we won't take this one!

PR #5192, opened 2022-10-17
Refactor autofill so it doesn't lean as heavily on sync15::Payload

mhammond:
We used to convert from sync15::Payload into (eg) AddressPayload, then finally into InternalAddress as late as possible. We now convert into the AddressPayload much earlier, so sync15::Payload doesn't bleed as far into the implementations.

Also deletes `PersistablePayload` as it was just an extra layer of confusion for no good reason.

Helps on the path to #5139. No hurry on this at all.


lougeniaC64 (approved):
LGTM!

skhamis (commented):
Great cleanup!

PR #5195, opened 2022-10-18
Removes redundent xcframework job

tarikeshaq:
The `iOS test` job already builds the `xcframework`. We don't need a whole job to do that.

This PR removes the `build-xcframework` job and instead persists the artifact in the `iOS test` job

our iOS jobs are the most expensive computationally (see the [insights on circle ci](https://app.circleci.com/insights/gh/mozilla/application-services)) and removing this job so slice that cost by half

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -314,34 +314,6 @@ jobs:
       - run-tests
       - save-sccache-cache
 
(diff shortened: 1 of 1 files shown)


skhamis (approved):
This is a great cleanup! LGTM! 🚀

PR #5203, opened 2022-10-25
Update AC and Gradle Protobuf plugin to latest versions

rvandermeulen:
I left Glean alone for now, but do we want to pick up a newer release at some point?


lougeniaC64:
> I left Glean alone for now, but do we want to pick up a newer release at some point?

I _think_ Travis may have some upcoming work that requires a Glean bump. cc @travis79 to ensure I'm not telling tales.

lougeniaC64 (approved):

PR #5207, opened 2022-10-28
Bump Android SDK target to 33

rvandermeulen:


rvandermeulen:
This seems to be mostly painless. The main issue is the getPackageInfo deprecation. The change in Nimbus.kt mirrors what we're doing in AC (and seems to be recommended elsewhere as well for how to address this deprecation). I guess the question I have is whether it makes more sense to implement this same change in Nimbus.kt or if we should wait for an AC108 release to be cut with https://github.com/mozilla-mobile/android-components/pull/12396/files#diff-85d4aae3bee0e872705cabd6b01aaca7e996558b331076bb8ee570e9434c2fa7 included and use the getPackageInfoCompat method it's adding instead?

jonalmeida:
> This seems to be mostly painless. The main issue is the getPackageInfo deprecation. The change in Nimbus.kt mirrors what we're doing in AC (and seems to be recommended elsewhere as well for how to address this deprecation). I guess the question I have is whether it makes more sense to implement this same change in Nimbus.kt or if we should wait for an AC108 release to be cut with https://github.com/mozilla-mobile/android-components/pull/12396/files#diff-85d4aae3bee0e872705cabd6b01aaca7e996558b331076bb8ee570e9434c2fa7 included and use the getPackageInfoCompat method it's adding instead?

Android components depends on app services and not the other way around (with the exception of concept-fetch), so those extensions won't be usable here unfortunately.

jonalmeida (approved):
LGTM

rvandermeulen:
Sounds like this is ready for A-S review then! FWIW, it doesn't appear that we've historically called out these bumps in the changelog, but I'm happy to do so if desired.

bendk (approved):
Looks good to me as well.

I personally don't think build code like this needs a CHANGELOG entry.

PR #5215, opened 2022-11-01
Added method to unset the error reporter

bendk:
As part of the UniFFI Desktop callback interface review, we were asked to provide a way to drop callback interfaces on shutdown to break up the reference cycles.

I didn't add any tests for this since it seemed fairly trivial.

  - Note:

### DIFF
--- a/components/support/error/src/reporting.rs
+++ b/components/support/error/src/reporting.rs
@@ -53,6 +53,10 @@ pub fn set_application_error_reporter(reporter: Box<dyn ApplicationErrorReporter
     *APPLICATION_ERROR_REPORTER.write() = reporter;
 }
 
+pub fn unset_application_error_reporter() {
+    *APPLICATION_ERROR_REPORTER.write() = Box::new(DefaultApplicationErrorReporter)
+}
+
 pub fn report_error_to_app(type_name: String, message: String) {
     APPLICATION_ERROR_REPORTER
         .read()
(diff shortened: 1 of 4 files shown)


components/support/error/src/reporting.rs
```diff
@@ -53,6 +53,10 @@ pub fn set_application_error_reporter(reporter: Box<dyn ApplicationErrorReporter
     *APPLICATION_ERROR_REPORTER.write() = reporter;
 }
 
+pub fn unset_application_error_reporter() {
```
skhamis, line 56: Since for a-s rust devs, this feels a tad bit unconventional from the other ways we handle breaking up reference cycles. Maybe adding a comment here referencing "this should be called on desktop to handle ref cleanups/due to callback interfaces" or something to that tune.

skhamis (approved):
Keeping up with that desktop discussion, this both looks good and makes sense to me!

mhammond:
As a style question, in the general case, would we suggest "set" and "unset" methods, or a single "set" that takes a nullable reference? For some reason I kind of prefer the latter, but this is so subjective...

bendk:
> As a style question, in the general case, would we suggest "set" and "unset" methods, or a single "set" that takes a nullable reference? For some reason I kind of prefer the latter, but this is so subjective...

I wondered about this myself and was basically 50/50.  I guess it felt better to have different functions for startup/shutdown, but then again passing in a value at startup and `null` at shutdown also feels very natural.  I could be swayed either way.

PR #5219, opened 2022-11-02
Bug 1798754 - Migrate application-services from AWS to GCP

MasterWayZ:
This PR migrates this repo over to the GCP pools.
!!Can only land once https://phabricator.services.mozilla.com/D161061 has!!


bendk (dismissed):
This seems fine to my naive eyes, but I think @ahal should definitely approve as well.

MasterWayZ:
Migrates decision tasks as well.

MasterWayZ:
Oh whoops, I can't put ahal back myself as a reviewer.

taskcluster/ci/config.yml
```diff
@@ -20,11 +20,21 @@ workers:
       implementation: docker-worker
       os: linux
       worker-type: 'b-linux'
+    b-linux-gcp:
+      provisioner: 'app-services-{level}'
+      implementation: docker-worker
+      os: linux
+      worker-type: 'b-linux-gcp'
```
ahal, line 27: What if instead of adding a second section here, we just modify the `worker-type` key of the existing `b-linux` section? So e.g:
```
b-linux:
    ...
    worker-type: b-linux-gcp
```

This way:
A) We won't have to clean up the non-GCP pools, and
B) It'll mean we don't have to update any of the task definitions (also we won't accidentally miss updating some of them)

I think duplicating these for GCP made sense in Gecko where there were a _ton_ of tasks, but for most projects on Github it's probably not necessary as we can just do all tasks at once.

ahal (changes requested):
Thanks! I had an idea for how we can approach migrating these Github projects which might work out a bit nicer.

ahal (dismissed):
Thanks, looks great!

tarikeshaq (dismissed):

tarikeshaq:
closed and re-openned to get the Circle CI going, apologies 🙏 I'll rebase then land this once ci passes

tarikeshaq:
Actually won't land it until I get confirmation that the dependant PR in phab has landed properly 😄

ahal:
> Actually won't land it until I get confirmation that the dependant PR in phab has landed properly 😄

Yes, this should be good to merge. Thanks!

MasterWayZ:
Yes it's ready to merge @tarikeshaq

tarikeshaq (approved):

PR #5225, opened 2022-11-03
Update Glean to 51.8.0

rvandermeulen:
Notably bumping the Android target API to 33


rvandermeulen:
@badboy Feel free to r+ this whenever you're comfortable with it merging

badboy:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1799129 to get this into m-c, when that's on its way in we can land this, then both things should land in a-c without issues.

badboy (approved):

PR #5226, opened 2022-11-03
Behavioral targeting event store queries

jeddai:
This PR adds a method to query into the event store. This will be used to find core active users, as well as users meeting other behavior criteria.

#5206 must be merged first.

[EXP-2643](https://mozilla-hub.atlassian.net/browse/EXP-2643)

  - Note:

### DIFF
--- a/components/nimbus/src/behavior.rs
+++ b/components/nimbus/src/behavior.rs
@@ -7,6 +7,7 @@ use crate::error::{BehaviorError, NimbusError, Result};
 use crate::persistence::{Database, StoreId};
 use chrono::{DateTime, Timelike, Utc};
 use serde::{Deserialize, Serialize};
+use std::collections::vec_deque::Iter;
 use std::collections::{HashMap, VecDeque};
 use std::fmt;
 use std::hash::{Hash, Hasher};
@@ -219,6 +220,29 @@ pub enum EventQueryType {
     AveragePerNonZeroInterval,
 }
 
+impl EventQueryType {
+    fn perform_query(&self, buckets: Iter<u64>, num_buckets: usize) -> Result<u64> {
+        Ok(match self {
+            Self::Sum => buckets.sum::<u64>(),
+            Self::CountNonZero => buckets.filter(|v| v > &&0u64).count() as u64,
+            Self::AveragePerInterval => buckets.sum::<u64>() / num_buckets as u64,
+            Self::AveragePerNonZeroInterval => {
+                let values = buckets.fold((0, 0), |accum, item| {
+                    (
+                        accum.0 + item,
+                        if item > &0 { accum.1 + 1 } else { accum.1 },
+                    )
+                });
+                if values.1 == 0 {
+                    0
+                } else {
+                    values.0 / values.1
+                }
+            }
+        })
+    }
+}
+
 #[derive(Default, Serialize, Deserialize, Debug)]
 pub struct EventStore {
     pub(crate) events: HashMap<String, MultiIntervalCounter>,
@@ -305,24 +329,7 @@ impl EventStore {
                 let buckets = single_counter.data.buckets.range(
                     starting_bucket..usize::min(num_buckets, single_counter.data.buckets.len()),
                 );
-                return Ok(match query_type {
-                    EventQueryType::Sum => buckets.sum::<u64>(),
-                    EventQueryType::CountNonZero => buckets.filter(|v| v > &&0u64).count() as u64,
-                    EventQueryType::AveragePerInterval => buckets.sum::<u64>() / num_buckets as u64,
-                    EventQueryType::AveragePerNonZeroInterval => {
-                        let values = buckets.fold((0, 0), |accum, item| {
-                            (
-                                accum.0 + item,
-                                if item > &0 { accum.1 + 1 } else { accum.1 },
-                            )
-                        });
-                        if values.1 == 0 {
-                            0
-                        } else {
-                            values.0 / values.1
-                        }
-                    }
-                });
+                return query_type.perform_query(buckets, num_buckets);
             }
         }
         Ok(0)
--- a/components/nimbus/src/behavior.rs
+++ b/components/nimbus/src/behavior.rs
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 #![allow(dead_code)]
 
 use crate::error::{BehaviorError, NimbusError, Result};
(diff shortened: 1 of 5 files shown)


components/nimbus/src/behavior.rs
```diff
@@ -14,13 +13,30 @@ use std::hash::{Hash, Hasher};
...
 pub enum Interval {
+    Minutes,
     Hours,
     Days,
     Weeks,
     Months,
     Years,
 }
 
+impl Interval {
```
travis79, line 24: This makes sense to move here, good call!

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -0,0 +1,657 @@
...
+        let mut store = EventStore::from(vec![
+            ("event-1".to_string(), counter1),
+            ("event-2".to_string(), counter2),
+        ]);
+
+        store.record_event("event-1".to_string(), Some(Utc::now() + Duration::days(2)))?;
+        store.record_event("event-1".to_string(), Some(Utc::now() + Duration::days(3)))?;
+        store.record_event("event-1".to_string(), Some(Utc::now() + Duration::days(3)))?;
+
+        assert!(matches!(
```
travis79, line 489: You can also use `assert_eq!()` here and avoid the `matches!()`. This is absolutely fine if you prefer this, it's perfectly understandable as is.

travis79 (approved):
Looking good!

components/nimbus/src/tests/test_behavior.rs
jeddai: Ah I'll make those adjustments in my next PR -- I was reviewing something of James' earlier today and noticed it there as well!

components/nimbus/src/behavior.rs
jeddai: Yep James suggested it on Wednesday so I figured I'd do it in this PR

components/nimbus/src/behavior.rs
```diff
@@ -273,338 +287,44 @@ impl EventStore {
...
+    pub fn query(
+        &self,
+        event_id: String,
+        interval: Interval,
+        num_buckets: usize,
+        starting_bucket: usize,
+        query_type: EventQueryType,
+    ) -> Result<u64> {
+        if let Some(counter) = self.events.get(&event_id) {
+            if let Some(single_counter) = counter.intervals.get(&interval) {
```
jhugman, line 300: I'd consider having these in their respective classes:

```rs
impl MultiCounter {
    fn query(interval: Interval, num_buckets: usize, starting_bucket: usize, query_type: EventQueryType) -> Result<u64> {}
}

impl SingleCounter {
    fn query(num_buckets: usize, starting_bucket: usize, query_type: EventQueryType) -> Result<u64> {}
}
```

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -0,0 +1,657 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// cargo test --package nimbus-sdk --lib --all-features -- tests::test_behavior --nocapture
```
jhugman, line 5: 😻

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -0,0 +1,657 @@
...
+    #[test]
+    fn rotate_handles_large_rotation() -> Result<()> {
+        let mut interval = IntervalData::new(3);
+        interval.rotate(10).ok();
+
+        assert!(matches!(interval.buckets.len(), 3));
+        assert!(matches!(interval.buckets[0], 0));
+        assert!(matches!(interval.buckets[1], 0));
+        assert!(matches!(interval.buckets[2], 0));
+        Ok(())
```
jhugman, line 196: Nice!

components/nimbus/src/behavior.rs
```diff
@@ -273,338 +287,44 @@ impl EventStore {
...
+                        if values.1 == 0 {
+                            0
+                        } else {
+                            values.0 / values.1
+                        }
+                    }
+                });
+            }
+        }
+        Ok(0)
```
jhugman, line 328: Consider putting this into the `EventQueryType` enum.

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -0,0 +1,657 @@
...
+
+        let counter2 = MultiIntervalCounter::new(vec![
+            SingleIntervalCounter::new(IntervalConfig::new(12, Interval::Months)),
+            SingleIntervalCounter::new(IntervalConfig::new(28, Interval::Days)),
+        ]);
+
+        let mut store = EventStore::from(vec![
+            ("event-1".to_string(), counter1),
+            ("event-2".to_string(), counter2),
+        ]);
```
jhugman, line 352: Side question: how do new events get multi-counters, apart from at construction?

jhugman (commented):
Drive by review: this is looking good. I have a few questions, but I won't stand in the way with another review cycle.

Great job!

components/nimbus/src/behavior.rs
```diff
@@ -63,9 +79,9 @@ impl IntervalConfig {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct IntervalData {
-    buckets: VecDeque<u64>,
-    bucket_count: usize,
-    starting_instant: DateTime<Utc>,
+    pub(crate) buckets: VecDeque<u64>,
```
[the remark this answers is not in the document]
jeddai: I don't explicitly remember why I left it as a u64 but I do believe there's a piece of math or a method that returns a u64 and I felt like casting it to a u32 was somewhat dangerous so I just kinda left it. It does make sense within the confines of the type that there's virtually no way the math/method could've overflowed outside of a u32, so if we want I can change it to a u32 and put the casting where necessary.

components/nimbus/src/behavior.rs
```diff
@@ -273,338 +287,44 @@ impl EventStore {
...
-
-        Ok(())
+    pub fn query(
+        &self,
+        event_id: String,
+        interval: Interval,
+        num_buckets: usize,
+        starting_bucket: usize,
+        query_type: EventQueryType,
+    ) -> Result<u64> {
```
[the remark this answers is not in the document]
jeddai: So Travis and I chatted about this, I had initially written the function to return a float. We landed on using integers and letting the default truncation behavior do its thing -- the intent being that it might be easier to document.

components/nimbus/src/behavior.rs
jeddai: So I did do that initially, I ended up pulling them all here because the event store/multi counter methods were sort of just wrappers for the actual logic in the single counter. If it makes more sense for the method to be there on each class I'm happy to move it back!

components/nimbus/src/tests/test_behavior.rs
jeddai: I don't know that we've thought about that yet, but you're right we will need something in place to handle that as we perform releases and start to add new events.

travis79: Ideally, I see this happening automatically in `record_event`, if it can't find an existing counter it would create one.

jeddai: Do we want to include that as a part of this PR or shall I include it in the next PR I'm opening?

components/nimbus/src/behavior.rs
jhugman: I think the `EventQueryType` is the important one to grow a query function, so I'd be happy to lay off this refactor.

components/nimbus/src/behavior.rs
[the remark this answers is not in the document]
jhugman: FWIW: JEXL casts numbers to `f64`.

>  it might be easier to document.

I think there is a danger of being easier for the engineer to document, but harder for the end user to reason about.

e.g. the JEXL expression:

```jexl
'event'|my_query_function() > 2.5
```

would be incorrectly`false` if:

```
2.5 <= 'event'|my_query_function() < 3.0
```

components/nimbus/src/behavior.rs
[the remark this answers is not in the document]
jhugman: My thinking was that a `u64` was twice the disk access as `u32` etc.

Let's leave this like this until we're optimizing the startup.

components/nimbus/src/behavior.rs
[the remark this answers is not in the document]
travis79: > So I've already done the jexl work for the next PR and I did indeed notice that. I ended up casting the f64->u64s but I do agree it might make more sense to have the query result be a float. @travis79 do you have any thoughts?

No objections to changing it from me.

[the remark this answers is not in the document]
jeddai: Okay I'll get that changed in the next PR if that's okay -- the next one is where it'll really come into play with the jexl work.

PR #5228, opened 2022-11-07
add jexl transforms for behavioral targeting

jeddai:
This PR adds jexl transforms for behavioral targeting!

Includes changes from #5226. 

[EXP-2644](https://mozilla-hub.atlassian.net/browse/EXP-2644)

  - Note:


jeddai:
I've got one change to make re: swapping the types for the transforms and query from u64->f64 but this is ready to be reviewed otherwise.

components/nimbus/src/lib.rs
```diff
@@ -145,7 +145,12 @@ impl NimbusClient {
         state: &mut MutexGuard<InternalMutableState>,
     ) -> Result<()> {
         self.update_ta_install_dates(db, writer, state)?;
-        self.event_store.lock().unwrap().read_from_db(db)?;
+        let mut event_store = self.event_store.lock().unwrap();
+        event_store.read_from_db(db)?;
+
+        // Now that the EventStore has been created, place it in its current state on the targeting_attributes.
+        let mut state = self.mutable_state.lock().unwrap();
+        state.targeting_attributes.event_store = Some(event_store.clone());
```
jeddai, line 153: Going to really quickly comment here to share one thing I'm still not sure of -- I saw two options when it comes to putting the event store on the targeting attributes. 
1. As soon as it exists and every time it is updated, clone it and put it on the targeting attributes
2. When we go to apply experiments or other methods that would use the targeting attributes, set the event store there, then after it has been used set it to `None`. 

This PR is currently using option 1, but I'm not sure whether that's the best choice. Thoughts would be appreciated!

travis79: I think we really want to grab a snapshot of the event store when applying experiments, so I would say we just need to refresh the targeting_attributes copy right before it gets consumed as part of applying the updates. If If you do it here, there may be events that occur between now and the app calling `apply_updates` that would have influenced the targeting evaluation. I would argue that this should happen in `apply_updates` rather than here, for that reason.  

My thinking is that any time there is a query, we would capture a "snapshot" (.clone()) of the event_store and release the in-memory one so that events could continue to be recorded. This is exactly what you have done, but I think `initialize` is too far from where the query is being executed.

components/nimbus/src/lib.rs
```diff
@@ -575,6 +580,10 @@ impl NimbusClient {
         let mut event_store = self.event_store.lock().unwrap();
         event_store.record_event(event_id, None)?;
         event_store.persist_data(self.db()?)?;
+
+        // Once we have updated the EventStore, update its value on the targeting_attributes.
+        let mut state = self.mutable_state.lock().unwrap();
+        state.targeting_attributes.event_store = Some(event_store.clone());
```
travis79, line 586: Not sure I like keeping two concurrent in memory copies. We don't need to keep the targeting attributes copy up to date, we only need to give it a current copy whenever it is consumed. At this point the event_store that is in-memory only needs to live in the targeting attributes and we don't exactly need a second one, do we?

components/nimbus/src/tests/test_evaluator.rs
```diff
@@ -919,3 +926,231 @@ fn test_enrollment_bucketing() {
...
+
+    let event_store = EventStore::from(vec![("app.foregrounded".to_string(), counter)]);
+    let mut targeting_attributes: TargetingAttributes = AppContext {
+        ..Default::default()
+    }
+    .into();
+    targeting_attributes.event_store = Some(event_store);
+    assert_eq!(
+        targeting(
+            "'app.foregrounded'|eventsAveragePerInterval('Days', 7, 0) > 2",
```
travis79, line 1019: Do we need a floating point test of this too?

components/nimbus/src/tests/test_lib.rs
```diff
@@ -679,6 +684,219 @@ fn test_days_since_update_failed_targeting() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn event_store_exists_for_apply_pending_experiments() -> Result<()> {
```
travis79, line 688: Great thing to test for!

travis79 (commented):
A few questions and comments for my first pass through this. Very much like what I am seeing so far!

components/nimbus/src/tests/test_evaluator.rs
jeddai: Yep!

components/nimbus/src/lib.rs
```diff
@@ -860,7 +869,10 @@ impl NimbusTargetingHelper {
     }
 
     pub fn eval_jexl(&self, expr: String) -> Result<bool> {
-        evaluator::jexl_eval(&expr, &self.context)
+        evaluator::jexl_eval(
+            &expr,
+            &serde_json::from_value(self.context.clone()).unwrap(),
```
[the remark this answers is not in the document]
jeddai: I did a search of the repo and from what I could tell this struct/method are only used in tests -- is that an incorrect understanding?

components/nimbus/src/lib.rs
jeddai: Ah that's a good call -- yep I'll make that adjustment

components/nimbus/src/evaluator.rs
```diff
@@ -283,12 +287,33 @@ pub(crate) fn targeting(
...
-    expression_statement: &str,
-    context: &Context,
-) -> Result<bool> {
-    let evaluator =
-        Evaluator::new().with_transform("versionCompare", |args| Ok(version_compare(args)?));
+pub fn jexl_eval(expression_statement: &str, context: &TargetingAttributes) -> Result<bool> {
+    let evaluator = Evaluator::new()
+        .with_transform("versionCompare", |args| Ok(version_compare(args)?))
+        .with_transform("eventsSum", |args| {
+            Ok(query_event_store(context, EventQueryType::Sum, args)?)
```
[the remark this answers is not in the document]
jeddai: > What are the reasoning behind (and possible consequences of) removing the `<Context: serde::Serialize>`?

The `<Context: serde::Serialize>` was seemingly only in place for the `NimbusTargetingHelper`, and as I mentioned in a different comment as far as I could tell it was only used in tests. All other usages of `jexl_eval` had TargetingAttributes being passed in, so I figured it was a reasonable change.

> Looking at `query_event_store` it looks like it returns a `Result<Value>` and not a `Result<bool>`. I see that this is the existing way of doing things but, whe does that not conflict with the return type here?

With this adding the transform to the evaluator, what we ultimately see is the `Result<Value>` is an f64 or an error. If it's an error evaluation will stop, but if it's an f64 we can then do the comparisons we see in the tests. Effectively `'app.foregrounded'|eventsSum('Days', 28, 0) > 21` becomes `24 > 21`.

> And isn't the `Result` from the transform function call returned instead of wrapping everything in `Ok(...)?)`. Are we seeing some implicit `into()` happening here?

I think it is some implicit `into`, but I'm really not sure. I actually tried removing the `Ok`s but it was expecting the error type of the `Result` to be a uniffi typed error and not a Nimbus typed error.

components/nimbus/src/tests/test_evaluator.rs
jeddai: Just double checking -- are you suggesting a test where the right side of the operation is a floating point value?

components/nimbus/src/error.rs
```diff
@@ -59,12 +59,26 @@ pub enum NimbusError {
     BehaviorError(#[from] BehaviorError),
     #[error("TryFromIntError: {0}")]
     TryFromIntError(#[from] TryFromIntError),
+    #[error("ParseIntError: {0}")]
+    ParseIntError(#[from] ParseIntError),
+    #[error("JexlError: {0}")]
+    JexlError(#[from] JexlError),
```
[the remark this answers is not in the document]
jeddai: I suppose it's not exactly behavior-specific, it's much more jexl related. Yeah I'll move it into the larger error enum.

components/nimbus/src/lib.rs
jeddai: I've adjusted it to only set the targeting attributes value whenever it will be needed -- is there anything else I need to address for this comment?

travis79: Nope!

components/nimbus/src/tests/test_evaluator.rs
travis79: Yes, since we have changed this to a float, we might as well expose it here, right?

jeddai: Yep! Just wasn't sure if you were suggesting I test it in the transform (it'll just convert it to an int)

components/nimbus/src/lib.rs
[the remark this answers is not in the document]
travis79: No, just me commenting on an unguarded `unwrap` is all. If it can never fail because we ensure this elsewhere, it's fine.

components/nimbus/src/evaluator.rs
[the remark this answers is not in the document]
travis79: Okay, this was mostly to satisfy my curiosity than anything blocking.

components/nimbus/src/lib.rs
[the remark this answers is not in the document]
jeddai: Maybe it would be worth it to mark the struct/method as `pub(crate)` to make it more clear it's only for tests? Or I can add a comment?

components/nimbus/src/evaluator.rs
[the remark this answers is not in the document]
jeddai: I figured! Just wanted to address all your questions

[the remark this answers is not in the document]
jeddai: For real though I have no idea why doing `Ok` around those changes anything -- the `?` is still in the lambda so afaik it will end up erroring with a Nimbus error if it doesn't unwrap, which should be the same issue as just passing the function by itself 🤷

components/nimbus/src/lib.rs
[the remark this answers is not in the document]
travis79: Either sounds good to document this

travis79 (approved):
Looking good to me

components/nimbus/src/lib.rs
jeddai: I've made that adjustment -- would you mind giving your thoughts on the new code?

components/nimbus/src/lib.rs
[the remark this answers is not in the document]
jeddai: Okay so after poking around in the UDL I found that this method/struct are in fact being used by Fenix/iOS. So I might need to be making an adjustment here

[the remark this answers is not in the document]
travis79: Nice catch!

components/nimbus/src/evaluator.rs
```diff
@@ -37,6 +40,7 @@ pub struct TargetingAttributes {
     pub days_since_install: Option<i32>,
     pub days_since_update: Option<i32>,
     pub active_experiments: HashSet<String>,
+    pub event_store: Option<EventStore>,
```
jhugman, line 43: This doesn't look right here. `TargetingAttributes` are the thing that is serialized into JSON for direct access from JEXL; but `event_store` is accessed through a transform.

Would you consider a different way of moving the event store, or not serializing it on `TargetingAttibutes`?

jhugman (changes requested):
Midway through review before our 1-1.

components/nimbus/src/evaluator.rs
travis79: This was something @jeddai  and I discussed but we never settled on a good way of getting the event store into the JEXL evalutor. If this can be improved, I'm all for it.

jeddai: It is now a parameter that is passed in all the way down the chain. Seems to work well!

components/nimbus/src/lib.rs
```diff
@@ -850,17 +863,37 @@ impl NimbusStringHelper {
...
+            context,
+            event_store,
+        }
     }
 
     pub fn eval_jexl(&self, expr: String) -> Result<bool> {
-        evaluator::jexl_eval(&expr, &self.context)
+        let mut context: NimbusTargetingHelperContext =
+            serde_json::from_value(self.context.clone())?;
+        context.event_store = Some(self.event_store.clone());
```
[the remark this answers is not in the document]
jeddai: So I think the only places I make copies now are the `get_event_store` method (used in `experiments.rs`) and when creating the NimbusTargetingHelper.

components/nimbus/src/lib.rs
travis79: I'm planning on looking at this again this morning, sorry for any delay

travis79 (approved):
👍  This is looking super!

components/nimbus/src/evaluator.rs
```diff
@@ -286,9 +290,34 @@ pub(crate) fn targeting(
 pub fn jexl_eval<Context: serde::Serialize>(
     expression_statement: &str,
     context: &Context,
+    event_store: &EventStore,
 ) -> Result<bool> {
-    let evaluator =
-        Evaluator::new().with_transform("versionCompare", |args| Ok(version_compare(args)?));
+    let evaluator = Evaluator::new()
+        .with_transform("versionCompare", |args| Ok(version_compare(args)?))
+        .with_transform("eventsSum", |args| {
```
jhugman, line 297: Nit: from experience, beginning an identifier with a plural confuses things later on. Suggest replacing `events` with `event`, e.g. `eventSum`, `eventAveragePerInterval`, `eventCountNonZeroIntervals`, `eventAveragePerNonZeroInterval`

components/nimbus/src/behavior.rs
```diff
@@ -57,6 +58,24 @@ impl Hash for Interval {
...
+    type Err = NimbusError;
+
+    fn from_str(input: &str) -> Result<Self> {
+        match input {
+            "Minutes" => Ok(Self::Minutes),
+            "Hours" => Ok(Self::Hours),
+            "Days" => Ok(Self::Days),
+            "Weeks" => Ok(Self::Weeks),
+            "Months" => Ok(Self::Months),
+            "Years" => Ok(Self::Years),
```
jhugman, line 71: Ok.

components/nimbus/src/behavior.rs
```diff
@@ -213,6 +232,37 @@ impl MultiIntervalCounter {
...
+                bucket_count: 60,
+                interval: Interval::Minutes,
+            }),
+            SingleIntervalCounter::new(IntervalConfig {
+                bucket_count: 24,
+                interval: Interval::Hours,
+            }),
+            SingleIntervalCounter::new(IntervalConfig {
+                bucket_count: 56,
+                interval: Interval::Days,
```
jhugman, line 248: 😻

jhugman (approved):
This is looking good now. I still have a couple of concerns about cloning of the event store— I _think_ we need to protect the instance with an `Arc` or an `ArcMut` but I think we should land this, get a good plan of action on what to do and file other issues to fix them. 

Good job, @jeddai .

components/nimbus/src/behavior.rs
```diff
@@ -297,7 +347,16 @@ impl EventStore {
 
     pub fn record_event(&mut self, event_id: String, now: Option<DateTime<Utc>>) -> Result<()> {
         let now = now.unwrap_or_else(Utc::now);
-        let counter = self.events.get_mut(&event_id).unwrap();
+        let counter = match self.events.get_mut(&event_id) {
+            Some(v) => v,
+            None => {
+                let new_counter = MultiIntervalCounter {
+                    ..Default::default()
+                };
```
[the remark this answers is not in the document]
jeddai: Does it perform type inference based off where `new_counter` is used farther down in the method or will I need to specify the type there?

components/nimbus/src/behavior.rs
jeddai: Ok

components/nimbus/src/behavior.rs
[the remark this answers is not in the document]
jeddai: Seems as though the answer is yes and that is my new favorite thing of the day

PR #5229, opened 2022-11-08
Change custom targeting attributes into JSONObject

jhugman:
Part of fixing [EXP-2892](https://mozilla-hub.atlassian.net/browse/EXP-2892)

  - Note:


components/nimbus/ios/Nimbus/JSONUtil.swift
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+internal func stringify(jsonObject: [String: Any]) throws -> String {
```
travis79, line 7: nit: I don't know if you can create an extension for a dictionary type, but could this and `toJson` below be `internal extension`s for their respective types? That would make calling it instead of like this:
`let string = try stringify(jsonObject: additionalContext)`

To something like this:
`let string = try additionalContext.stringify()`

This seems to be a little more idomatic of Swift and doesn't change how this works, so feel free to disregard.

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -140,9 +140,9 @@ extension Nimbus: FeaturesInterface {
     internal func getFeatureConfigVariablesJson(featureId: String) -> [String: Any]? {
         do {
             if let string = try nimbusClient.getFeatureConfigVariables(featureId: featureId),
```
travis79, line 142: nit: Since the `else` case of the `if let` here only returns `nil`, I think `guard let` could be appropriate here

travis79 (approved):
Just a couple of minor nits, no need to block or re-review

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
```diff
@@ -58,6 +58,7 @@
 		1BF50F1927B1E19500A9C8A5 /* FxAccountManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1BBAC54127AE065300DAFEF2 /* FxAccountManagerTests.swift */; };
 		1BF50F1A27B1E19800A9C8A5 /* FxAccountMocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1BBAC53C27AE065300DAFEF2 /* FxAccountMocks.swift */; };
 		1BFC469827C99F250034E0A5 /* Metrics.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1BFC469627C99F250034E0A5 /* Metrics.swift */; };
+		3963A5862919A541001ED4C3 /* JSONUtil.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3963A5852919A541001ED4C3 /* JSONUtil.swift */; };
```
jeddai, line 61: So I've not worked on iOS projects within git repos before -- what kind of changes should we expect to be made to the files like this that we should be committing back to the repo?

jeddai (commented):
A small suggestion and a general question, no concerns with merging.

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
travis79: This is just what adding a new swift file to the Xcode project looks like. Xcode project files are just a bit gross to read by a huma, but, unfortunately we need to commit it to the repo. We don't edit this manually so I don't think it is as important to review manually.

PR #5236, opened 2022-11-15
Polls all commands and not just one in iOS

tarikeshaq:
Creating as a draft until I get the policy side of this understood.

There is already handling in iOS in https://github.com/mozilla-mobile/firefox-ios/blob/3f5031721b6b6e5151c41ecb2ec4dc9d14a7e347/Account/FxAPushMessageHandler.swift#L65 that handles cases where we poll multiple commands in one poll

We'll want to modify that too so we display all the tabs we got from FxA


tarikeshaq:
This should be ready to land, I'll have a PR up in iOS that makes the handling a bit more sane, but even as-is without any iOS changes it should be OK based on https://github.com/mozilla-mobile/firefox-ios/blob/3f5031721b6b6e5151c41ecb2ec4dc9d14a7e347/Account/FxAPushMessageHandler.swift#L65

components/fxa-client/src/internal/push.rs
```diff
@@ -13,12 +13,19 @@ impl FirefoxAccount {
     /// Handle any incoming push message payload coming from the Firefox Accounts
     /// servers that has been decrypted and authenticated by the Push crate.
     ///
-    /// Due to iOS platform restrictions, a push notification must always show UI.
-    /// Since FxA sends one push notification per command received,
-    /// we must only retrieve 1 command per push message,
-    /// otherwise we risk receiving push messages for which the UI has already been shown.
-    /// However, note that this means iOS currently risks losing messages for
-    /// which a push notification doesn't arrive.
+    /// ** ⚠️ Due to iOS platform restrictions, a push notification must always show UI. **
```
lougeniaC64, line 16: Thanks for this very helpful empty list explanation!

lougeniaC64 (approved):
LGTM!  🚢

tarikeshaq:
Swiftlint unrelated, this change only changes Rust

PR #5239, opened 2022-11-16
Move Nimbus startup into Application Services

jhugman:
This builds on the work of https://github.com/mozilla-mobile/fenix/pull/27650 and https://github.com/mozilla-mobile/focus-android/pull/8012 , pulling them together into a common codebase.

Part of fixing [EXP-2893](https://mozilla-hub.atlassian.net/browse/EXP-2893).

  - Note:


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/NimbusBuilder.kt
```diff
@@ -0,0 +1,144 @@
...
+     * A flag to indicate if this is being run on the first run of the app. This is used to control
+     * whether the `initial_experiments` file is used to populate Nimbus.
+     */
+    var isFirstRun: Boolean = true
+
+    /**
+     * A optional raw resource of a file downloaded at or near build time from Remote Settings.
+     */
+    @RawRes
+    var initialExperiments: Int? = null
```
jeddai, line 48: I wonder if converting this to be a String type would be helpful as we plan to integrate Nimbus into other apps. We could have other Kotlin apps that aren't Android, and as far as I'm aware the raw resource int thing is an Android thing.

If it's the string value of the initial experiments, I think that also lends itself a bit better to testing and extensibility.

jeddai (commented):
Looks solid, I've got one question.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/NimbusBuilder.kt
jhugman: I think at this point, building out a non-Android Kotlin version would be a quite a large undertaking (for example, we depend on `android.content.Context`), for currently not much demand.

I think I'm inclined to say "Let's cross that bridge when we get to it".

jeddai: Sounds good to me!

jeddai (approved):

PR #5244, opened 2022-11-17
add nimbus behavioral event recording to swift and kotlin apis

jeddai:

  - Note:


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/FeaturesInterface.kt
```diff
@@ -59,4 +59,15 @@ interface FeaturesInterface {
...
+    /**
+     * Records an event to the Nimbus event store.
+     *
+     * The method obtains the event counters for the `eventId` that is passed in, advances them if
+     * needed, then increments the counts by 1. If an event counter does not exist for the `eventId`,
+     * one will be created.
+     *
+     * @param eventId string representing the id of the event which should be recorded.
+     */
+    fun recordEvent(eventId: String)
```
travis79, line 72: This shouldn't be exposed on a Feature, but rather in the general API.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -357,6 +357,10 @@ open class Nimbus(
         recordExposure(featureId)
     }
 
+    override fun recordEvent(eventId: String) {
+        nimbusClient.recordEvent(eventId)
```
travis79, line 361: This does potentially have some IO writing going on, I think that launching this on the dbScope may be the right thing to do here so we aren't blocking the calling thread.

travis79: The only reason `recordExposure` doesn't do this above, is because Glean already manages the async bit for us.

components/nimbus/ios/Nimbus/Nimbus.swift
```diff
@@ -75,6 +75,10 @@ extension Nimbus: FeaturesInterface {
         }
     }
 
+    public func recordEvent(eventId: String) {
```
travis79, line 78: Same thing as above, this should be part of the general Nimbus API and not the FeatureInterface. Either move this to the `public class Nimbus` definition or into an appropriate extension.

travis79 (commented):
Just a few comments  here to list what we talked about in slack

travis79 (approved):
Looking good to me!

PR #5245, opened 2022-11-17
Be more careful with URL redaction

bendk:
I thought that the initial scheme was what Desktop used, but that was just wrong.  If we're going to be using a new method, then let's be extremely careful about any potential PII leaks.

The initial scheme potentially leaked PII, since URLs with many path segments would have a relatively unique redacted version.  Therefore if you had a known URL in mind like that, you could compare the redacted string to the redacted versions of the URLs in the Sentry breadcrumbs and potentially determine that it was likely the user visited the URL.

  - Note:

### DIFF
--- a/components/support/error/src/redact.rs
+++ b/components/support/error/src/redact.rs
@@ -4,9 +4,37 @@
 
 //! Functions to redact strings to remove PII before logging them
 
-/// Redact a URL, replacing all characters other than [`:`, `/`] with `x`
+/// Redact a URL.
+///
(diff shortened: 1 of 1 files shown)


skhamis (approved):
Thanks for taking the time to go back and make sure we remove PII from any logged stuff. We did something similar on desktop with: just replacing the whole thing with `<Guid>` so this aligns pretty well https://phabricator.services.mozilla.com/D150060

In any case LGTM!

mhammond (approved):
Fantastic, thanks

PR #5254, opened 2022-11-18
Update where mac cross compiler fetches precompiled NSS [ci full][ff-android: main][fenix: main]

skhamis:
- Updating NSS using the [NSS CI](https://treeherder.mozilla.org/jobs?repo=nss), which now has static builds https://phabricator.services.mozilla.com/D162252
- Updated NSS to 3.85, since that is the first build that has my patch from above (and also probably good to do)
- Remove pulling from the S3 bucket (simplifies our NSS process)

One "unfortunate" thing is that NSS CI builds by default with `PKIX` lib enabled, which is used when NSS calls it's `init` method. We do not, so I had to add an extra pull in of those libs for macos users that are running android tests on desktop. This _shouldn't_ affect anything as we don't use the APIs from the `PKIX` lib, I ran the full branch builds to confirm and looks like everything passed fine! 

  - Note:

### DIFF
--- a/components/support/rc_crypto/nss/nss_build_common/src/lib.rs
+++ b/components/support/rc_crypto/nss/nss_build_common/src/lib.rs
@@ -23,6 +23,12 @@ pub enum LinkingKind {
 pub struct NoNssDir;
 
 pub fn link_nss() -> Result<(), NoNssDir> {
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    println!(
+        "(link_nss) target_arch: {}, target_os: {}",
+        target_arch, target_os
+    );
     let is_gecko = env::var_os("MOZ_TOPOBJDIR").is_some();
     if !is_gecko {
         let (lib_dir, include_dir) = get_nss()?;
@@ -114,12 +120,13 @@ fn get_nss_libs(kind: LinkingKind) -> Vec<&'static str> {
             // Hardware specific libs.
             let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
             let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
-            // https://searchfox.org/nss/rev/08c4d05078d00089f8d7540651b0717a9d66f87e/lib/freebl/freebl.gyp#278-296
+            // https://searchfox.org/nss/rev/0d5696b3edce5124353f03159d2aa15549db8306/lib/freebl/freebl.gyp#508-542
             if target_arch == "arm" || target_arch == "aarch64" {
                 static_libs.push("armv8_c_lib");
             }
             if target_arch == "x86_64" || target_arch == "x86" {
                 static_libs.push("gcm-aes-x86_c_lib");
+                static_libs.push("sha-x86_c_lib");
             }
             if target_arch == "arm" {
                 static_libs.push("gcm-aes-arm32-neon_c_lib")
@@ -141,6 +148,9 @@ fn get_nss_libs(kind: LinkingKind) -> Vec<&'static str> {
                     static_libs.push("intel-gcm-s_lib");
                 }
             }
+            if (target_os == "macos") && (target_arch == "x86_64") {
+                static_libs.push("pkixtop");
+            }
             static_libs
         }
         LinkingKind::Dynamic { folded_libs } => {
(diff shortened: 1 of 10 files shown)


components/support/rc_crypto/nss/nss_build_common/src/lib.rs
```diff
@@ -141,6 +142,22 @@ fn get_nss_libs(kind: LinkingKind) -> Vec<&'static str> {
                     static_libs.push("intel-gcm-s_lib");
                 }
             }
+            // We need this due to using the NSS CI, which enables PKIX lib by default
+            // However this is only used for Macs testing android on desktop
+            // https://searchfox.org/mozilla-central/source/security/nss/lib/libpkix/libpkix.gyp#15-25
```
mhammond, line 147: I wonder if this should be a permalink? Or maybe just a link to the file itself without line numbers and a comment saying it should match the "dependencies" section would be ok?

libs/build-nss-desktop.sh
```diff
@@ -58,10 +58,10 @@ fi
 # https://github.com/mozilla/application-services/issues/962
 if [[ "${CROSS_COMPILE_TARGET}" =~ "darwin" ]]; then
   # Generated from nss-try@111b54aaa644978464bec98848ecba6f69d3f42e.
```
mhammond, line 60: Is this "Generated from" comment still relevant? And this is still pinned to a specific NSS version - do you think it's worth commenting here on what the process is for updating it? (I guess it's just "find the later version you want, update the task ID and get the SHA256 value (from where)?"

mhammond (commented):
This looks fantastic! Do you think we should get someone with a little more context to check it out? eg, someone from releng, even just to do a driveby? Is there any testing you think we should do locally (ie, is there anything that you think might possibly break for someone locally which CI would not pick up? I can't think of anything TBH, but that might just be due to my lack of knowledge about this)

components/support/rc_crypto/nss/nss_build_common/src/lib.rs
skhamis: Ah good point! I chatted with the NSS team and I think we can actually just avoid having libpkix all together (disable it in the CI builds) but if it turns out we have to keep this around i'll definitely make those changes

skhamis:
> This looks fantastic! Do you think we should get someone with a little more context to check it out? eg, someone from releng, even just to do a driveby? Is there any testing you think we should do locally (ie, is there anything that you think might possibly break for someone locally which CI would not pick up? I can't think of anything TBH, but that might just be due to my lack of knowledge about this)

Yes I definitely want to get releng and possibly someone from NSS have a look at this, once our team has decided this is the path we want forward. However @tarikeshaq identified that there are some issues with missing libs when trying to build locally (intel mac). I think it's because we disable libpkix locally but link to it in `nss_build`

So the two options are:

1. Disable `libpkix` in the static CI builds (chatted with the NSS team, and they mentioned they'd take accept a patch for that)
2. Enable libpkix in app-services for desktop builds -- however even if we don't use the APIs this would separate desktop & mobile which I'd rather not do

so unfortunately this might be held until I can get a patch for NSS, but the turnaround for that is fairly quick

tarikeshaq (commented):
Dropping a comment here to remove this from my review queue since there is still work that needs to be done, please feel free to re-add me whenever!

tarikeshaq (changes requested):
Just want to quickly flag this - I tried rebuilding with the newest change and `cargo test` worked after I deleted my `libs/desktop` directory and re-ran `libs/verify-desktop-environment.sh`

But to further smoke test it, I tried it against firefox-ios. At first, I got a `sha_x86_c_lib` error which was gone after I deleted the `libs/ios` directory and ran `libs/verify-ios-environment.sh`.  The `build-xcframwork` worked OK and then building firefox-ios on my **physical** iPhone also worked OK!

But once I tried building firefox-ios on a **simulator** I got:
image

I haven't debugged beyond that, and **could very well be my environment** - but happy to help test next week! Since it only shows up on the simulator, could be something about the NSS upgrade missing a library or something like that for `aarch64-apple-ios-sim`

tarikeshaq:
Now that I think about it, this is probably an issue with the `rustqlite` update we landed earlier and not this PR 🤔

tarikeshaq (approved):
> Now that I think about it, this is probably an issue with the rustqlite update we landed earlier and not this PR 🤔

Built the latest main - the latest main builds OK with firefox-ios.

Then, I cleaned, and rechecked out this branch and now it works 🙃 🙃

I'm officially confused, but it works now so I'll approve.

Also probably worth it to make announcements that developers will need to deleted their `libs/*` directories, and run their `libs/verify-*-environment.sh` scripts. Slack is as good as any other place. Maybe also worth a changelog? but this is probably irrelevant for customers so 😅

PR #5260, opened 2022-11-27
Exposes reset function swift

tarikeshaq:
## Why
I came across this when adding polling to firefox-ios. I noticed that **my polls were launching the same tabs I just got via push notification**. I tracked it down to be that the Notification Service that handles the push notifications runs as a separate process and has its own instance of our FxA client, but uses the same backing persisted state.

### What happens when a notification arrives:
- Notification service creates a new instance of the account manager
- notification service polls for the tab it just got
-  it would increment the index in-memory
- It would then persist a new state with the new index

**However, the account manager held by firefox-ios doesn't have the updated index in memory. Unless the user relaunches the app in between the push notification and the next scheduled poll (once a day) the poll would get back the same tab.**

and even if the user relaunches the app, a lot of the FxA APIs persist the state, and since our state is a simple JSON, the state persisted by the notification service could have been overwritten by then

### How it'll be used:
This PR creates a method to allow the account manager held by firefox-ios to refresh its in-memory values by re-loading the persisted state from disk.

Firefox-iOS would call this API when it's first foregrounded and notices it received a tab (there is an existing mechanism for that)This way firefox-ios would get the up-to-date index before any poll (or any other API) is run.

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -20,5 +20,13 @@ Use the template below to make assigning a version number during the release cut
 -->
 
 ## FxA Client
+
+### What's new
+- Exposed a new function for swift consumers `resetPersistedState`
+   - `resetPersistedState` can be used to refresh the account manager to reflect the latest persisted state.
+   - `resetPersistedState` should be called in between a different account manager instance persisting the state, and the current account manager persisting state
+     - For example, the Notification Service in iOS creates its own instance of the account manager, changes its state (by changing the index of the last retrieved send tab)
+     - The main account manager held by the application should call` resetPersistedState` before calling any other method that might change its state. This way it can retrieve the most up to date index that the Notification Services persisted.
 ### What's changed
-- The `processRawIncomingAccountEvent` function will now process all commands, not just one. This moves the responsibilty of ensuring each push gets a UI element to the caller.\
+- The `processRawIncomingAccountEvent` function will now process all commands, not just one. This moves the responsibilty of ensuring each push gets a UI element to the caller.
+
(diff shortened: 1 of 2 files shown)


lougeniaC64:
> I tracked it down to be that the Notification Service that handles the push notifications runs as a separate process and has its own instance of our FxA client

Does it need to have its own instance of the client?

tarikeshaq:
> Does it need to have its own instance of the client?

Great question!! Unfortunately yes, because iOS runs notification services as their own process :(

Currently, the flow of a push message is as follows:
- The notification service creates its own instance of fxa_client
- it uses it to read the push message (including polls for any missing send tabs)
- If the message is more than a notification (in the case of a send tab, or a device disconnection)
   - it persists a pref for device disconnections
   - or it persists a tab in an SQL table that the application can access later for send tab
- It displays the push notification to the user

Then, the app next time it's foregrounded:
- The app checks for the device disconnection pref, if so, disconnect itself from FxA
- The app checks for the SQL tab queue to see if the notification service passed any tabs

So the problem is that the notification service is using a separate instance of the fxa client to **read the push message**. 

One option is to change the Notification Service so it doesn't use any "state-altering" methods of the fxa_client. This probably means changing what `handle_message` does so it doesn't poll FxA and instead just acts as a way to tell the app to poll once it wakes up. We'd still need `handle_message` to decrypt the send tab in the payload to display the notification - but the actual tab that will be launched is whatever the app got from fxa by polling once it woke up

That seems like a nice cleanup, so the Notification Service becomes stateless! but for the meantime, this is a quicker solution to get us to a working polling in firefox-ios!

lougeniaC64:
> That seems like a nice cleanup, so the Notification Service becomes stateless! but for the meantime, this is a quicker solution to get us to a working polling in firefox-ios!

Yeah I like the idea of removing state from the Notification Service but I can imagine that's going to be a lot more work than what you have here. Thanks for the explanation, @tarikeshaq!

lougeniaC64 (approved):

PR #5264, opened 2022-11-29
Use incremental vacuum for places maintenance

bendk:
Metrics from GLAM indicate the VACUUM command is taking a long time and doesn't seem to be dropping.  After reading the docs a bit, it seems to me that VACUUM always copies the entire database which could indeed be slow for large DBs.

Switched to using an incremental vacuum.  This only tries to truncate N pages (set to 2 here) at time rather than all free pages.  But the real speed improvement comes from only trimming completely free pages, not trying to defragment pages, which is what VACUUM does and why it needs to copy the DB.

Incremental vacuum requires a PRAGMA setting.  If that setting isn't set before there is data in the DB, then we still need to do a single full VACUUM before it can be used.

I think the result of all of this should be that we will do one slow vacuum, but future vacuums should be quick.

Also added a places-utils command to run maintenance.  I used this to test that the incremental vacuum is ineed faster.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,8 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## places
+
+### What's Changed
+  - Switch to using incremental vacuums for maintenance, which should speed up the process.
--- a/components/autofill/src/db/schema.rs
+++ b/components/autofill/src/db/schema.rs
@@ -80,7 +80,7 @@ impl ConnectionInitializer for AutofillConnectionInitializer {
     const NAME: &'static str = "autofill db";
     const END_VERSION: u32 = 2;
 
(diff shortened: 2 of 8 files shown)


bendk:
Tested this by running maintenance on my DB using places-utils 10 times in a row.  The full vacuum version used about 650ms total, the new version was around 160ms.

mhammond (approved):
:+1

bendk:
@mhammond I needed to update this one a bit because I realized some issues with the PRAGMA statement:
  - It modifies the DB, so it can't run in a read-only connection
  - Setting the value, but not running VACUUM leaves the DB in a weird state where the old value is still active until the next VACUUM (even for fresh DBs).  I don't think this would currently lead to a bug, but it seems risky to me.  So I updated the code in `prepare()` to run VACUUM immediately after the PRAGMA.  This lead to some more code to only run that on newly-created DBs, since VACUUM might be slow for existing ones.  BTW, I couldn't move this code to `init()` because that runs in a transaction, which prevents VACUUM.

tarikeshaq (approved):
LGTM!

PR #5265, opened 2022-11-29
Adjust nimbus event store date comparison logic and add additional testing

jeddai:
I chatted with Travis and where we’ve landed on this is that all time calculations are going to be based off of relative time in whole intervals, not absolute time. For example, Dec 31 23:59:30 → Jan 01 00:01:20 would be a rotation of 1 minute, and no rotations for other intervals. As another example, Jan 01 10:30:00 → Jan 15 10:00:00 would result in rotations of 20,130 minutes, 13 days, and 1 week.

This PR does add tests to make sure all interval types are being calculated in a relative way.

Also, all the usages of EventStore in method params are updated to `Arc<Mutex<EventStore` so they can be cloned without issue.

https://mozilla-hub.atlassian.net/browse/EXP-2942

  - Note:


components/nimbus/src/enrollment.rs
```diff
@@ -648,7 +649,7 @@ impl<'a> EnrollmentsEvolver<'a> {
             &prev_rollouts,
             &next_rollouts,
             &ro_enrollments,
-            event_store,
+            event_store.clone(),
```
travis79, line 652: The function that called this also `clone()` the event store so now it's been cloned twice before being used at all

components/nimbus/src/enrollment.rs
```diff
@@ -726,7 +727,7 @@ impl<'a> EnrollmentsEvolver<'a> {
                 prev_experiments.get(slug).copied(),
                 next_experiments.get(slug).copied(),
                 Some(prev_enrollment),
-                event_store,
+                event_store.clone(),
```
travis79, line 730: Here's another clone of the event store before it is used, since this is 3 functions deep now we have made 3 copies and haven't accessed any of them.

components/nimbus/src/evaluator.rs
```diff
@@ -382,7 +387,7 @@ fn query_event_store(
         }
     } as usize;
 
-    Ok(json!(event_store.query(
+    Ok(json!(event_store.lock().unwrap().query(
```
travis79, line 390: This is the only time that the EventStore is actually pulled out of the Arc<Mutex<>> and used, can we can pass a reference to the Arc<Mutex<>> all the way to this point to reduce the number of times that it is cloned?

travis79 (changes requested):
I'm a little unsure of all the cloning that this change has produced, see my inline comments.

travis79 (approved):
My mistake, cloning is happening on the `Arc` so there should still be only the single underlying `Mutex<EventStore>`. Disregard my cloning concerns about this. 😅

PR #5270, opened 2022-12-01
Allow viaduct to make requests to the android emulator loopback address.

mhammond:
Viaduct itself gets the ability to allow exactly 1 arbitrary address to be connected to which isn't a https:// URL, but the FFI adds an android specific function which hard-codes the accepted address.

@csadilek said this seems to work in his testing. I don't think it's a breaking change but will require an a-c patch to expose the new method.

@badboy already pointed out my check should check just the host, but I also included the scheme. LMK if you would prefer different semantics.

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -30,3 +30,9 @@ Use the template below to make assigning a version number during the release cut
    - `importBookmarksFromFennec`
    - `importPinnedSitesFromFennec`
    - `importVisitsFromFennec`
+
+## Viaduct
+### What's New
+  - Allow viaduct to make requests to the android emulator's host address via
+    a new viaduct_allow_android_emulator_loopback() (in Rust)/allowAndroidEmulatorLoopback() (in Kotlin)
+    ([#5270](https://github.com/mozilla/application-services/pull/5270))
--- a/components/viaduct/Cargo.toml
+++ b/components/viaduct/Cargo.toml
@@ -15,6 +15,7 @@ log = "0.4"
 serde = "1"
 serde_json = "1"
 once_cell = "1.5"
+parking_lot = { version = ">=0.11,<=0.12" }
 prost = "0.8"
 prost-derive = "0.8"
 ffi-support = "0.4"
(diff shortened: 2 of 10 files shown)


mhammond:
I can't formally ask @badboy for review, so consider this an informal request :)

CHANGES_UNRELEASED.md
```diff
@@ -6,6 +6,13 @@
 
 <!-- WARNING: New entries should be added below this comment to ensure the `./automation/prepare-release.py` script works as expected.
 
+## Viaduct
+
+### What's New
+  - Allow viaduct to make requests to the android emulator loopback address via
+    a new viaduct_allow_android_emulator_loopback() (in Rust)/allowAndroidEmulatorLoopback() (in Kotlin)
+    ([#5270](https://github.com/mozilla/application-services/pull/5270))
+
```
badboy, line 15: This should be moved below line 20 (outside of the `<!-- ... -->` comment block)

components/viaduct/Cargo.toml
```diff
@@ -11,10 +11,12 @@ crate-type = ["lib"]
 
 [dependencies]
 url = "2.1" # mozilla-central can't yet take 2.2 (see bug 1734538)
+lazy_static = "1.4"
```
badboy, line 14: `viaduct` now uses both `lazy_static` and `once_cell`.
As a followup this should be trimmed down to only one (probably `once_cell` as the more "modern" variant and the API that's gonna get into libstd eventually)

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -54,6 +54,17 @@ object RustHttpConfig {
         }
     }
 
+    /** Allows connections to the hard-coded address the Android Emulator uses for
+      * localhost (ie, http://10.0.2.2) - if you don't call this, viaduct will fail to
```
badboy, line 58:
```suggestion
      * the host's localhost (ie, http://10.0.2.2) - if you don't call this, viaduct will fail to
```

or something like that? Or leave out `localhost` and call it `the host machine`?

components/viaduct/Cargo.toml
badboy: Ah, this is of course the first usage of `lazy_static` here. Maybe just use `once_cell` then? We can migrate `lazy_static` in `viaduct-reqwest` later.

badboy (approved):
Small nits, but I like how simple and clean this is.

mhammond:
Thanks! Applied all suggestions and even make viaduct-request use once_cell instead of lazy_static!

Issue #5230, opened 2022-11-10
[Tabs] Add support for lastModified in ClientRemoteTabs


skhamis:
We had to [manually add a lastModified](https://bugzilla.mozilla.org/show_bug.cgi?id=1800026) field manually on desktop due to [SyncedTabs needing it to display the "Last Synced" ](https://searchfox.org/mozilla-central/source/services/sync/modules/SyncedTabs.jsm#81). We should implement this in rust, which would help simplify the parts of the code on desktop that juggle that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3476)
┆Epic: Tabs on desktop follow-up work

mhammond:
(that branch is on top of #5139, so I'll put a PR up one that lands)

PR #5273, opened 2022-12-02
Add support for lastModified in ClientRemoteTabs (#5230)

mhammond:
Hey Sammy, I don't think this conflicts with the work you are doing on tabs, right? If not, I'll land it after I make sure it works ok on desktop.


skhamis (approved):
These changes look great. This def won't affect any of the work i'm doing as mine is more replacing the in-memory `last_sync`, etc items where this is more record-based.

LGTM! 🚢

Issue #5199, opened 2022-10-21
Ignore incoming sync records for devices we don't know about.


mhammond:
This bit us in desktop - it would have bitten us much less if this was fixed.

If we don't know the device we should ignore the record - so [this branch](https://github.com/mozilla/application-services/blob/1c4093bb3db502bbf151370ed56ac2ea6f9826c7/components/tabs/src/sync/engine.rs#L144) changes to, say a `log::warn` about _not_ storing, then a `continue` to ignore the record.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3441)

mhammond:
Turns out we can't actually do this because stand alone syncing (ie, store.sync()) doesn't allow us to pass the device list in, so we'd get no rows! :( I'm mildly surprised iOS tab syncing doesn't seem to be bitten by this, which I suspect is going to come back and bite us moving iOS to the sync manager.

(I'm keeping it open because I think we should still do this once we kill stand-alone syncing)

mhammond:
I think we already did this.

PR #5277, opened 2022-12-05
Dont reset the tabs engine every sync, use the db for storing last sync

skhamis:
Based on bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1800186

- We stop resetting on every sync
- Instead of storing the entire payload in one row, we do it per client_id (one row per client)
- [x] Tested on Desktop
- [x] Tested on iOS
- [x] Tested on Android

  - Note:

### DIFF
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -23,8 +23,8 @@ Use the template below to make assigning a version number during the release cut
 
 ### What's Changed
 
-The Tabs engine no longer resets after every sync:
+The Tabs engine is now more efficient in how it fetches its records:
 
-- Tabs will be stored in a DB per client.
-- Tabs now tracks the last time it synced and only fetches tabs that have changed since last sync.
-- Tabs will keep records for up to 180 days, the same as the clients engine.
+- The Tabs engine no longer clears the DB on every sync.
+- Tabs now tracks the last time it synced and only fetches tabs that have changed since the last sync.
+- Tabs will keep records for up to 180 days, in parity with the clients engine. To prevent the DB from getting too large.
--- a/CHANGES_UNRELEASED.md
+++ b/CHANGES_UNRELEASED.md
@@ -18,3 +18,13 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Tabs
+
+### What's Changed
+
+The Tabs engine no longer resets after every sync:
+
+- Tabs will be stored in a DB per client.
+- Tabs now tracks the last time it synced and only fetches tabs that have changed since last sync.
+- Tabs will keep records for up to 180 days, the same as the clients engine.
--- a/components/tabs/src/schema.rs
+++ b/components/tabs/src/schema.rs
@@ -34,9 +34,9 @@ const CREATE_META_TABLE_SQL: &str = "
 pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
 pub(crate) static GLOBAL_SYNCID_META_KEY: &str = "global_sync_id";
 pub(crate) static COLLECTION_SYNCID_META_KEY: &str = "tabs_sync_id";
-// Tabs stores this in the meta table due to a unique requirement that we only know
-// the list of connected clients when syncing, however getting list of tabs could be
-// called at anytime -- thus we need to store this in a way we can retrieve the clients list on the fly
+// Tabs stores this in the meta table due to a unique requirement that we only know the list
+// of connected clients when syncing, however getting the list of tabs could be called at anytime
+// so we store it so we can translate from the tabs sync record ID to the FxA device id for the client
 pub(crate) static REMOTE_CLIENTS_KEY: &str = "remote_clients";
 
 pub struct TabsMigrationLogic;
--- a/components/tabs/src/schema.rs
+++ b/components/tabs/src/schema.rs
@@ -18,9 +18,9 @@ use sql_support::{
 // store each client as its own row.
 const CREATE_SCHEMA_SQL: &str = "
     CREATE TABLE IF NOT EXISTS tabs (
-        guid             TEXT NOT NULL PRIMARY KEY,
-        record              TEXT NOT NULL,
-        last_modified    INTEGER NOT NULL
+        guid            TEXT NOT NULL PRIMARY KEY,
+        record          TEXT NOT NULL,
+        last_modified   INTEGER NOT NULL
     );
 ";
 
@@ -67,7 +67,6 @@ impl MigrationLogic for TabsMigrationLogic {
     }
 
     fn upgrade_from(&self, db: &Transaction<'_>, version: u32) -> MigrationResult<()> {
-        log::debug!("upgrading from {}", version);
         match version {
             1 => upgrade_from_v1(db),
             _ => Err(MigrationError::IncompatibleVersion(version)),
--- a/components/tabs/src/schema.rs
+++ b/components/tabs/src/schema.rs
@@ -7,27 +7,43 @@
 // syncChangeCounter/syncStatus nor a mirror etc.
 
 use rusqlite::{Connection, Transaction};
-use sql_support::open_database::{
-    ConnectionInitializer as MigrationLogic, Error as MigrationError, Result as MigrationResult,
+use sql_support::{
+    open_database::{
+        ConnectionInitializer as MigrationLogic, Error as MigrationError, Result as MigrationResult,
+    },
+    ConnExt,
 };
 
-// The payload is json and this module doesn't need to deserialize, so we just
-// store each "payload" as a row.
-// On each Sync we delete all local rows re-populate them with every record on
-// the server. When we read the DB, we also read every single record.
-// So we have no primary keys, no foreign keys, and really completely waste the
-// fact we are using sql.
+// The record is the TabsRecord struct in json and this module doesn't need to deserialize, so we just
+// store each client as its own row.
 const CREATE_SCHEMA_SQL: &str = "
     CREATE TABLE IF NOT EXISTS tabs (
-        payload TEXT NOT NULL
+        guid             TEXT NOT NULL PRIMARY KEY,
+        record              TEXT NOT NULL,
+        last_modified    INTEGER NOT NULL
     );
 ";
 
-pub struct TabsMigrationLogin;
+const CREATE_META_TABLE_SQL: &str = "
+    CREATE TABLE IF NOT EXISTS moz_meta (
+        key    TEXT PRIMARY KEY,
+        value  NOT NULL
+    )
+";
 
-impl MigrationLogic for TabsMigrationLogin {
+pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
+pub(crate) static GLOBAL_SYNCID_META_KEY: &str = "global_sync_id";
+pub(crate) static COLLECTION_SYNCID_META_KEY: &str = "tabs_sync_id";
+// Tabs stores this in the meta table due to a unique requirement that we only know
+// the list of connected clients when syncing, however getting list of tabs could be
+// called at anytime -- thus we need to store this in a way we can retrieve the clients list on the fly
+pub(crate) static REMOTE_CLIENTS_KEY: &str = "remote_clients";
+
+pub struct TabsMigrationLogic;
+
+impl MigrationLogic for TabsMigrationLogic {
     const NAME: &'static str = "tabs storage db";
-    const END_VERSION: u32 = 1;
+    const END_VERSION: u32 = 2;
 
     fn prepare(&self, conn: &Connection, _db_empty: bool) -> MigrationResult<()> {
         let initial_pragmas = "
@@ -45,20 +61,43 @@ impl MigrationLogic for TabsMigrationLogin {
     }
 
     fn init(&self, db: &Transaction<'_>) -> MigrationResult<()> {
-        log::debug!("Creating schema");
-        db.execute_batch(CREATE_SCHEMA_SQL)?;
+        log::debug!("Creating schemas");
+        db.execute_all(&[CREATE_SCHEMA_SQL, CREATE_META_TABLE_SQL])?;
         Ok(())
     }
 
-    fn upgrade_from(&self, _db: &Transaction<'_>, version: u32) -> MigrationResult<()> {
-        Err(MigrationError::IncompatibleVersion(version))
+    fn upgrade_from(&self, db: &Transaction<'_>, version: u32) -> MigrationResult<()> {
+        log::debug!("upgrading from {}", version);
+        match version {
+            1 => upgrade_from_v1(db),
+            _ => Err(MigrationError::IncompatibleVersion(version)),
+        }
     }
+}
+
+fn upgrade_from_v1(db: &Connection) -> MigrationResult<()> {
+    // The previous version stored the entire payload in one row
+    // and cleared on each sync -- it's fine to just drop it
+    db.execute_batch("DROP TABLE tabs;")?;
+    // Recreate the world
+    db.execute_all(&[CREATE_SCHEMA_SQL, CREATE_META_TABLE_SQL])?;
+    Ok(())
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::storage::TabsStorage;
+    use rusqlite::OptionalExtension;
+    use serde_json::json;
+    use sql_support::open_database::test_utils::MigratedDatabaseFile;
+
+    const CREATE_V1_SCHEMA_SQL: &str = "
+        CREATE TABLE IF NOT EXISTS tabs (
+            payload TEXT NOT NULL
+        );
+        PRAGMA user_version=1;
+    ";
 
     #[test]
     fn test_create_schema_twice() {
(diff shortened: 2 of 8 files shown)


components/tabs/src/schema.rs
```diff
@@ -19,13 +22,25 @@ use sql_support::open_database::{
...
+pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
+pub(crate) static GLOBAL_SYNCID_META_KEY: &str = "global_sync_id";
+pub(crate) static COLLECTION_SYNCID_META_KEY: &str = "tabs_sync_id";
+
+pub struct TabsMigrationLogic;
 
-impl MigrationLogic for TabsMigrationLogin {
+impl MigrationLogic for TabsMigrationLogic {
     const NAME: &'static str = "tabs storage db";
     const END_VERSION: u32 = 1;
```
mhammond, line 45: It seems like this should be bumped and we migrate from v1 to v2?

components/tabs/src/schema.rs
```diff
@@ -45,20 +60,40 @@ impl MigrationLogic for TabsMigrationLogin {
...
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::storage::TabsStorage;
+    use sql_support::open_database::test_utils::MigratedDatabaseFile;
+
+    const CREATE_V1_SCHEMA_SQL: &str = "
```
mhammond, line 92: and I guess this is v2? Maybe @bendk can help here if he remembers the semantics

components/tabs/src/schema.rs
```diff
@@ -67,4 +102,47 @@ mod tests {
         conn.execute_batch(CREATE_SCHEMA_SQL)
             .expect("should allow running twice");
     }
+
+    #[test]
+    fn test_tabs_meta() {
+        let mut db = TabsStorage::new_with_mem_path("test");
+        //let conn = db.open_or_create().unwrap();
```
mhammond, line 109: nit - kill this line.

Also, consider moving these tests to db.rs (ie, where the meta functions being tested are actually defined). Ideally tests here would limit themselves to the schema - eg, manually selecting/inserting into the tab makes sense here.

components/tabs/src/schema.rs
```diff
@@ -67,4 +102,47 @@ mod tests {
...
+        assert!(retrieved_value4.is_none());
+    }
+
+    #[test]
+    fn test_tabs_db_upgrade_from_v1() {
+        let db_file = MigratedDatabaseFile::new(TabsMigrationLogic, CREATE_V1_SCHEMA_SQL);
+        db_file.run_all_upgrades();
+        // Verify we can open the DB just fine, since migration is essentially a drop
+        // we don't need to check any data integrity
+        let mut storage = TabsStorage::new(db_file.path);
```
mhammond, line 144: this should check the migration actually worked (eg, selecting from the GUID column would fail if it didn't, even with no rows)

components/tabs/src/storage.rs
```diff
@@ -198,13 +201,14 @@ impl TabsStorage {
...
 
+        // For tabs it's fine if we override the existing tabs for a remote
+        // there can only ever be one record for each client
         for crt in new_remote_tabs {
             tx.execute_cached(
-                "INSERT INTO tabs (payload) VALUES (:payload);",
+                "INSERT OR REPLACE INTO tabs (guid, payload) VALUES (:guid, :payload);",
                 rusqlite::named_params! {
+                    ":guid": &crt.client_id,
                     ":payload": serde_json::to_string(&crt).expect("tabs don't fail to serialize"),
```
mhammond, line 212: nano-nit - I think avoiding "payload" here makes sense - it's not a "payload" in sync terms and I got a little confused by how the "modified" timestamp was going to work here as the "payload" doesn't have that field) - just renaming it to "json" or similar would be fine.  Possibly in the schema too maybe? (Sorry I didn't mention this before)

components/tabs/src/sync/bridge.rs
```diff
@@ -80,29 +79,29 @@ impl BridgedEngine for BridgedEngineImpl {
         handle_error! {
             let new_id = SyncGuid::random().to_string();
             let new_coll_ids = CollSyncIds {
-                global: SyncGuid::empty(),
-                coll: new_id.clone().into(),
+                global: new_id.clone().into(),
```
mhammond, line 82: I think this change is wrong - it's `coll` that we want to set, not `global`. This implies a test is probably missing - we should probably have a test that after calling `reset_sync_id`, `sync_id` returns the value we "reset" it to?

mhammond (commented):
This is looking pretty good. A few more comments! I also wonder how you feel about still doing a reset periodically? Or maybe periodically deleting records more than 3 weeks old?

It kinda scares me that we now will *never* do a reset whereas in the past we used to do it regularly. eg, something causes the database to grow without bound, and some clients now starting to show very stale devices? OTOH, we do assume that clients are going to filter records they don't know about (ie, where the fxa-id isn't known) so maybe this isn't a problem, but I'm not 100% sure mobile will always do the right thing here?

skhamis:
> This is looking pretty good. A few more comments! I also wonder how you feel about still doing a reset periodically? Or maybe periodically deleting records more than 3 weeks old?
> 
> It kinda scares me that we now will _never_ do a reset whereas in the past we used to do it regularly. eg, something causes the database to grow without bound, and some clients now starting to show very stale devices? OTOH, we do assume that clients are going to filter records they don't know about (ie, where the fxa-id isn't known) so maybe this isn't a problem, but I'm not 100% sure mobile will always do the right thing here?

Thanks for the comments, yeah I wanted to get this up to get discussions going early. I do feel like having the client filter what they don't know about is probably fine as long as the DB doesn't get ridiculously large. However I like the idea of having an optional param to send in a list of client ids (null or empty = send everything) but I guess that doesn't necessarily need to be decided/done in this patch.

 We definitely should have some periodic cleanup/maintenance, deleting records more than three weeks is consistent with what push is doing (with targets) so feels like a good starting point for now!

components/tabs/src/schema.rs
skhamis: We're actually on version 0 right now, since I _think_ this never ran the first time (no DB) the version number didn't get bumped.

bendk: Does that mean we're never saving the DB to disk right now?
  - If we don't save the DB, then I don't think you need the `v1` migration.  That's only needed for DBs that were created before the `sql_support::open_database` code.
  - If we do save the DB, then I think Mark is right and we need a v1 -> v2 migration.

bendk:
Oops, I think I mistakenly deleted one of @mhammond's comments related to the migration while trying to delete some of my own.  I didn't even realize I could do that in github.

components/tabs/src/schema.rs
mhammond: > Does that mean we're never saving the DB to disk right now?

tabs are odd - the db is only created when we need it - which roughly means, when sync has been connected.

Is it possible you are seeing it as version 0 when it doesn't actually exist?

skhamis: I think I didn't have the `PRAGMA version=1` set for the old schema in the test code, which was why I was getting weird results. I think you're both correct and it'll need to be V1 -> V2

components/tabs/src/storage.rs
```diff
@@ -226,8 +302,41 @@ impl TabsStorage {
...
+            Ok(None) => Ok(None),
+        }
+    }
+
+    pub(crate) fn delete_meta(&mut self, key: &str) -> Result<()> {
+        if let Some(db) = self.open_if_exists()? {
+            db.execute_cached("DELETE FROM moz_meta WHERE key = :key", &[(":key", &key)])?;
+        }
+        Ok(())
+    }
```
bendk, line 336: We've written these so many times, maybe it would make sense to put the code in `sql-support`.  Probably better as a follow-up though.

bendk (changes requested):
I focused on the db-related changes and just briefly skimmed the sync engine changes.

Overall it looks good, I'm marking this "request changes" because of the deletion issue.  However, I'm not sure if changes really need to be made.  If there's a reason to keep around the old tabs, then I can just change this to approve.

components/tabs/src/storage.rs
```diff
@@ -188,27 +211,80 @@ impl TabsStorage {
...
+            let record = remote_tab.0;
+            let last_modified = remote_tab.1;
             tx.execute_cached(
-                "INSERT INTO tabs (payload) VALUES (:payload);",
+                "INSERT OR REPLACE INTO tabs (guid, record, last_modified) VALUES (:guid, :record, :last_modified);",
                 rusqlite::named_params! {
-                    ":payload": serde_json::to_string(&crt).expect("tabs don't fail to serialize"),
+                    ":guid": &record.id,
+                    ":record": serde_json::to_string(&record).expect("tabs don't fail to serialize"),
+                    ":last_modified": last_modified.as_millis()
```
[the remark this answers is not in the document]
skhamis: So each client has exactly one tab record (which contains all the tabs) which is stored as a json (in `record`, though maybe I should change the name to avoid this confusion). So if a client has an updated tab list -- it would only ever overwrite it's own row. Thus preventing any dangling records.

Also I have a `remove_stale_clients()` that removes any stale clients > 3 weeks during sync (in `apply_incoming`) if a client Id changes and/or never connects again to prevent the DB to be filled with any stragglers

components/tabs/src/storage.rs
```diff
@@ -188,27 +211,80 @@ impl TabsStorage {
...
+                            record,
+                        )
+                    } else {
+                        // A record with a device that's not in our remote clients seems unlikely, but
+                        // could happen - in most cases though, it will be due to a disconnected client -
+                        // so we really should consider just dropping it? (Sadly though, it does seem
+                        // possible it's actually a very recently connected client, so we keep it)
+                        log::info!(
+                        "Storing tabs from a client that doesn't appear in the devices list: {}",
+                        id,
```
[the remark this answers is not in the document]
skhamis: That's a good idea! totally will do that.

components/tabs/src/storage.rs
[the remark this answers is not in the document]
bendk: Ahh, I should have said delete the client records that are no longer present.  Looks like that's covered too though.

bendk (approved):

components/tabs/src/storage.rs
```diff
@@ -188,27 +211,80 @@ impl TabsStorage {
                         );
                         None
                     }
+                };
+                let mut crts: Vec<ClientRemoteTabs> = Vec::new();
+                for (record, last_modified) in records.unwrap_or_default() {
+                    let id = record.id.clone();
```
mhammond, line 217: Let's add a comment somewhere here explaining that the `id` from the database is the id from the clients collection, but we need to obtain the fxa_device_id for that client to return to the consumers, and include a link to #5199 in that `else` block (becaise we want that else block to die entirely at some point as it will be doing the wrong thing)

mhammond (changes requested):
This is looking in good shape, thanks!

components/tabs/src/sync/engine.rs
```diff
@@ -39,8 +38,9 @@ pub fn get_registered_sync_engine(engine_id: &SyncEngineId) -> Option<Box<dyn Sy
     }
 }
 
+/// XXX - move to storage.rs? The struct is defined there
```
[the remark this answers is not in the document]
skhamis: TabsRecord is actually defined in sync:record -- however I think any moves unrelated to this actual PR is probably best served in a separate PR

components/tabs/src/storage.rs
```diff
@@ -29,7 +39,9 @@ pub struct RemoteTab {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ClientRemoteTabs {
-    pub client_id: String, // Corresponds to the `clients` collection ID of the client.
+    // The client id is essentially the FxA ID that is passed in from the clients engine
+    // on most clients this is the full fxa id however on desktop it is only the shortform version
```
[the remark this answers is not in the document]
skhamis: This is a great find! It's definitely hard to explain without a whole heap of context but I feel #5303 captures what's going on quite succinctly

components/tabs/src/sync/bridge.rs
```diff
@@ -372,6 +371,19 @@ mod tests {
                     "lastUsed": 1643764207
                 }]
             }),
+            // test an updated payload will replace the previous record
+            json!({
```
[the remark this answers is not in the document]
skhamis: Good catch -- I ended up moving this "test" to the engine since it was mainly making sure the apply_incoming works as intended and that's a better fit in the engine.rs

components/tabs/src/schema.rs
```diff
@@ -45,20 +61,43 @@ impl MigrationLogic for TabsMigrationLogin {
...
+            1 => upgrade_from_v1(db),
+            _ => Err(MigrationError::IncompatibleVersion(version)),
+        }
     }
 }
 
+fn upgrade_from_v1(db: &Connection) -> MigrationResult<()> {
+    // The previous version stored the entire payload in one row
+    // and cleared on each sync -- it's fine to just drop it
+    db.execute_batch("DROP TABLE tabs;")?;
```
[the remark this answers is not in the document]
skhamis: That's a good idea, I'll make sure the mobile teams are made aware and allow them to flag if it's potentially a no-go. However I do think that most users _probably_ won't notice this as they'd be having to go directly to synced tabs after updating + not first synced yet.

skhamis:
> This is looking in good shape, thanks!

Thanks for the review, I think it's looking even better with the requested changes! I have also pinged the mobile teams to make them aware of the potential no-synced-tabs-on-first-sync issue. Everything else I've been able to resolve/accept -- putting this back into review queue!

components/tabs/src/storage.rs
```diff
@@ -166,52 +176,125 @@ impl TabsStorage {
...
+                    last_modified,
+                    remote_client,
+                    record,
+                )
+            } else {
+                // A record with a device that's not in our remote clients seems unlikely, but
+                // could happen - in most cases though, it will be due to a disconnected client -
+                // so we really should consider just dropping it? (Sadly though, it does seem
+                // possible it's actually a very recently connected client, so we keep it)
+                error_support::breadcrumb!(
```
mhammond, line 242: I'm not sure this is valuable - I guess this is to make it easier to remove this branch entirely? https://github.com/mozilla/application-services/issues/5199 notes that iOS is likely to hit this, so I doubt we want every sync there to report this breadcrumb?  And wouldn't a `report_error` call be closer to what we want than a breadcrumb?

mhammond (commented):
This looks great, thanks for the changes. Let's hear from the mobile teams before landing and make sure we do some serious testing!

components/tabs/src/storage.rs
skhamis: ah I guess I was thinking this was equivalent to to `log::warn` here, I think you're right we don't necessarily need the codepath this happened and more we just need to get rid of this eventually

components/tabs/src/schema.rs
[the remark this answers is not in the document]
MatthewTighe: Paraphrasing a slack comment here:

Android at least currently does not take advantage of cached tabs, and instead always wait for a sync to complete before displaying the results. This should mean that there is no behavioral regression during the migration path.

CHANGES_UNRELEASED.md
```diff
@@ -18,3 +18,13 @@ Use the template below to make assigning a version number during the release cut
   - Description of the change with a link to the pull request ([#0000](https://github.com/mozilla/application-services/pull/0000))
 
 -->
+
+## Tabs
+
+### What's Changed
+
+The Tabs engine no longer resets after every sync:
```
mhammond, line 26: nano-nit - "reset" probably doesn't mean much to consumers - maybe change this to describe it as making the tabs engine more efficient or something?

components/tabs/src/schema.rs
```diff
@@ -7,27 +7,43 @@
...
+        key    TEXT PRIMARY KEY,
+        value  NOT NULL
+    )
+";
+
+pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
+pub(crate) static GLOBAL_SYNCID_META_KEY: &str = "global_sync_id";
+pub(crate) static COLLECTION_SYNCID_META_KEY: &str = "tabs_sync_id";
+// Tabs stores this in the meta table due to a unique requirement that we only know
+// the list of connected clients when syncing, however getting list of tabs could be
```
mhammond, line 38: For clarity, to "the list of connected clients when syncing," maybe append something like "so we can translate from the tabs sync record ID to the FxA device id for the client" or something, as that's really the core reason we need to do this.

components/tabs/src/storage.rs
```diff
@@ -166,52 +175,140 @@ impl TabsStorage {
...
+    // Keep DB from growing infinitely since we only ask for records since our last sync
+    // and may or may not know about the client it's associated with -- but we could at some point
+    // and should start returning those tabs immediately. If that client hasn't been seen in 3 weeks,
+    // we remove it until it reconnects
+    pub fn remove_stale_clients(&mut self) -> Result<()> {
+        let last_sync = self.get_meta::<i64>(schema::LAST_SYNC_META_KEY)?;
+        if let Some(conn) = self.open_if_exists()? {
+            if let Some(last_sync) = last_sync {
+                let client_ttl_ms = (TABS_CLIENT_TTL as i64) * 1000;
+                // On desktop, a quick write sets the last_sync to FAR_FUTURE
```
mhammond, line 261: maybe update this comment to better reflect it "temporarily" sets this, but if it crashes or somehow doesn't reset it, then we take this action

components/tabs/src/sync/bridge.rs
```diff
@@ -279,7 +278,7 @@ mod tests {
     use std::collections::HashMap;
     use sync15::{ClientData, DeviceType, RemoteClient};
 
-    const TTL_1_YEAR: u32 = 31_622_400;
+    const TABS_CLIENT_TTL: u32 = 15_552_000; // same as in storage.rs
```
mhammond, line 281: The one in storage.rs is `pub` so just `use` it here? (Not a big deal as it's just tests, but might as well anyway!)

components/tabs/src/storage.rs
```diff
@@ -166,52 +175,140 @@ impl TabsStorage {
...
+                    last_modified,
+                    remote_client,
+                    record,
+                )
+            } else {
+                // A record with a device that's not in our remote clients seems unlikely, but
+                // could happen - in most cases though, it will be due to a disconnected client -
+                // so we really should consider just dropping it? (Sadly though, it does seem
+                // possible it's actually a very recently connected client, so we keep it)
+                log::info!(
```
mhammond, line 241: Please add a link to https://github.com/mozilla/application-services/issues/5199 here as this branch should die.

mhammond (approved):
Looks great, thanks! Excited to see this land in desktop!

Issue #5284, opened 2022-12-08
Error parsing JSON data: missing field `last_modified`


bendk:
We're seeing a good deal of [Sentry events](https://sentry.io/organizations/mozilla/issues/3745938472/?project=6295546&referrer=issue-stream&sort=freq&statsPeriod=14d) for this, starting today after `96.1.2` was taken by a-c.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3519)

mhammond:
For more context: after upgrading to a version with this regression, the apps will fail to get any remote tabs until the next tabs sync happens. So for most users, the breakage will be relatively short and self-repair, but there's a v96.1.3 a-s release with the fix which consumers should probably take.

mhammond:
Done!

PR #5286, opened 2022-12-08
Fix breakage reading old version of tabs from database (#5284)

mhammond:


skhamis (approved):
Thanks for adding the test. Looks like the right thing for anyone that still has the old payload -- 🚢

PR #4416, opened 2021-08-31
Adds to release docs about merging back to main

tarikeshaq:
Adds a couple of points to document my experience with #4415

PR #5288, opened 2022-12-09
Merge v96.1.3

mhammond:
This merges release v96.1.3 back to main.

### DIFF
--- a/.buildconfig-android.yml
+++ b/.buildconfig-android.yml
@@ -1,4 +1,4 @@
-libraryVersion: 96.1.2
+libraryVersion: 96.1.3
 groupId: org.mozilla.appservices
 projects:
   autofill:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# v96.1.3 (_2022-12-08_)
+
+[Full Changelog](https://github.com/mozilla/application-services/compare/v96.1.2...v96.1.3)
+
+## Tabs
+
+### What's Changed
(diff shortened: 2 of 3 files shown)


tarikeshaq (changes requested):
heh this PR is empty 😝 te merge back should include the changelog so our changelog on main includes the 96.1.3 release

mhammond:
oops :)

.buildconfig-android.yml
```diff
@@ -1,4 +1,4 @@
-libraryVersion: 96.1.2
+libraryVersion: 96.1.3
```
tarikeshaq, line 1: Hmm this is tricky but I like it - usually we only merge back the changelog since technically the latest release from `main` is still the 96.1.2... It's a tricky place to be in, and I'm not really sure what our "policy" is for this 😅 

But I do like this in this case, because there haven't been any other releases since 96.1.2 and this way the next release is 96.1.4 - so I saw we go for it!

tarikeshaq (approved):

mhammond:
Yeah, I really wasn't sure about this. The release documentation just says "Merge the new release back to main" (and does mention a merge commit should be used) but you raise a good point (and it turns out you write that note in #4416 :) The last time we did this it looks like it was #4415 and we did the same. I agree it's odd though as it's going to fall down if we end up with a v97 from main and then want another 96 dot release - although I guess at that time it will become a merge conflict. So 🤷 :)

mhammond:
(stepping back, the fact we ended up a new branch to avoid taking the new rusqlite is a great demonstration of why we need to better align with the external release cycles and processes...)

PR #5291, opened 2022-12-12
Don't delete all existing tab records if we don't have any to apply.

mhammond:
This will fix bug 1801295.


skhamis (approved):
Makes sense and thank you for adding a test for it -- LGTM!

PR #5299, opened 2022-12-15
Update rust version to 1.65.0, MSR to 1.61

mhammond:
This matches m-c. Most of this was done with `clippy --fix` although the change to `components/support/nimbus-fml` surprised me and was done by hand (we could have ignore the lint to make the code look consistent, but the fix was easy and does make some sense)


tarikeshaq (approved):
Looks like clippy is still complaining 😝  but otherwise looks good to me

PR #5301, opened 2022-12-16
Sync AndroidX dependencies with AC

rvandermeulen:
Also updates AC to the latest available upstream release.


skhamis (approved):
Thanks for bumping these, LGTM!

PR #5302, opened 2022-12-16
Add an ADR for a remote settings client

mhammond:
[Click here for the rendered version](https://github.com/mhammond/application-services/blob/adr-0005-remote-settings-client/docs/adr/0005-remote-settings-client.md)

This is an ADR for a remote settings client for our mobile browsers. It was authored by @mhammond @csadilek and @MatthewTighe, but all additional feedback is appreciated.

### DIFF
--- a/docs/adr/0005-remote-settings-client.md
+++ b/docs/adr/0005-remote-settings-client.md
@@ -1,0 +1,188 @@
+# A remote-settings client for our mobile browsers.
+
+* Status: proposed
+* Discussion: https://github.com/mozilla/application-services/pull/5302
+* Deciders:
+    * csadilek for the mobile teams ✔️
+    * leplatrem for the remote-settings team ✔️
+    * mhammond for the application-services team ✔️
+
+* Date: 2022-12-16
+
+## Context and Problem Statement
+
+Mozilla’s mobile browsers have a requirement to access the remote settings service, but currently lack any libraries or tools which are suitable without some work.
+A concrete use case is the management of search engine configurations, which are stored in Remote Settings for Firefox Desktop, but shipped as individual files on our mobile browsers, requiring application releases for all changes.
+
+A constraint on any proposed solutions is that this work will be performed by Mozilla's mobile team, who have limited experience with Rust, and that it is required to be completed in Q1 2023.
+
+This document identifies the requirements, then identifies tools which already exist and are close to being suitable, then identifies all available options we can take, and outlines our decision.
+
+## Requirements
+The requirements are for a library which is able to access Mozilla’s Remote Settings service and return the results to our mobile browsers.
+This list of requirements is not exhaustive, but instead focuses on the requirements which will drive our decision making process.
+As such, it identifies the non-requirements first.
+
+### Non-requirements
+The following items all may have some degree of desirability, but they are not hard requirements for the initial version
+* While the https connection to the server must be validated, there is no requirement to verify the content received by the server - ie, there’s no requirement to check the signature of the body itself.
+* There’s no requirement to validate the result of the server conforms to a pre-defined schema - we trust the server data.
+* There’s no requirement to return strongly-typed data to the applications - returning a JSON string/object is suitable.
+* There’s no requirement to cache server responses to the file-system - if the app requests content, it’s fine for the library to always hit the server.
+* There’s no requirement for any kind of scheduling or awareness of network state - when we are requested for content, we do it immediately and return an appropriate error if it can not be fetched.
+* There’s no requirement to support publishing records, requesting reviews or providing approvals via this new library.
+* There’s no requirement that push be used to communicate changes to the application (eg, to enable rapid-enrolment type features)
+* There’s no requirement to manage buckets, groups and collections via this new library.
+
+### Initial Requirements
+The requirements we do have for the initial version are:
+* The library should allow fetching records from Mozilla’s Remote Settings servers. This includes support for attachments, and fetching [incremental changes](https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html#polling-for-changes).
+* The library should not create threads or run any event loops - the mobile apps themselves are responsible for all threading requirements. While this might change in the future, considering this kind of change to our mobile applications is out of scope for this project.
+* We must use Necko for all networking on Android, must enforce all connections are via valid https hosts (although some test-only exceptions might be helpful for QA, such as allowing localhost connections to be http)
+* The library should be the only remote-settings library used in the browser. Specifically, this means that Nimbus must also be capable of using the library, and the work to move Nimbus to the library must be considered as part of the project.
+
+## Existing Libraries
+We have identified the following libraries which may be suitable for this project.
+### Remote-settings on desktop
+There is a version of the [remote settings client in desktop](https://searchfox.org/mozilla-central/source/services/settings/RemoteSettingsClient.jsm), written in Javascript.
+It has been used and been relatively stable since at least 2018, so can be considered very capable, but the roadblock to it being suitable for use by our mobile browsers is that it is written in Javascript, so while it might be possible to expose it to Android via geckoview, there’s no reasonable path to have it made available to iOS.
+
+### Rust Remote Settings Client
+There is an existing [remote settings client on github](https://github.com/mozilla-services/remote-settings-client).
+This client is written in Rust and has evolved over a number of years.
+The most recent changes were made to support being used in [Merino](https://github.com/mozilla-services/merino), which was re-written in Python, so there are no known consumers of this library left.
+
+The main attributes of this library relevant to this discussion are:
+* It’s written in Rust, but has no FFI - ie, it’s currently only consumable by other Rust code.
+* It has recently been updated to use async rust, so requires an internal event loop.
+* It includes the capability to verify the signatures of the content.
+
+### The Nimbus-sdk Client
+The [nimbus-sdk is a component in the application-services repository](https://github.com/mozilla/application-services/tree/main/components/nimbus) written in Rust. It has client code which talks to the remote-settings server and while this has only actually been used with the "Nimbus" collection there's no reason to believe it can't be used in the more general case.
+The main attributes of this library relevant to this discussion are:
+* It’s consumed by a component which is already consumed by our mobile browsers via UniFFI.
+* It does not verify the signatures of the content - while this could be done, there hasn’t been sufficient justification made for this (ie, there are no realistic threat models which would be solved by this capability.)
+* The client itself does not persist a local cache of remote resources, but instead delegates this responsibility to the consuming application (in this case, nimbus itself, which does persist them via the [rkv library](https://github.com/mozilla/rkv))
+* It does not use async Rust, but instead everything is blocking and run on threads exclusively created by the app itself.
+* It has good test support, which run against a docker image.
+
+## Considered Options
+### Option 1: Writing a new library
+The requirements of this client are such that writing new libraries in Kotlin and Swift is currently a realistic option.
+However, we are rejecting this option because we don’t want to duplicate the effort required to write and maintain two libraries - inevitably, the features and capabilities will diverge.
+Future requirements such as supporting content signature verification would lead to significant duplication.
+
+Writing a new library from scratch in Rust and exposing it via UniFFI so it can be used by both platforms is also a possibility.
+However, we are rejecting this option because existing Rust libraries already exist, so we would be better served by modifying or forking one of the existing libraries.
+
+### Option 2: Use the existing remote settings client
+Modifying or forking the existing client is an attractive option.
+It would require a number of changes - the async capabilities would probably need to be removed (using a Rust event loop in our mobile browsers is something we are trying to avoid until we better understand the implications given these browsers already have an event loop and their own threading model).
+
+The persistence model used by this library is something that is not a requirement for the new library, which isn’t itself a problem, but it probably would preclude being able to use this library by Nimbus - so the end result is that we would effectively have two remote-settings clients written in Rust and used by our browsers.
+
+Some API changes would probably be required to make it suitable for use by UniFFI would also be necessary, but these would be tractable.
+
+We would need to update nimbus to use this client, which would almost certainly require moving this client into the application-services repository to avoid the following issues:
+* Marrying the persistence model of this client with the existing rkv-based persistence used by nimbus would be required.
+* Ensuring the upstream version changes continued to work for us.
+* Managing the circular dependency which exists due to this library needing to use viaduct.
+* Complication of our build process because the library needs to end up in our “megazord”.
+These are the exact reasons why Nimbus itself is in the application-services repo.
+
+### Option 3: Use the existing nimbus client
+Splitting the [existing client](https://github.com/mozilla/application-services/blob/6069bd26d742f9597ab7e62d8df23ad634c7bd98/components/nimbus/src/client/http_client.rs#L9) out from Nimbus in a way that allows Nimbus to continue to use it, while also making it available for stand-alone use is also an attractive option.
+
+In particular, the feature set of that client overlaps with the requirements of the new library - no local persistence is necessary and no signature verification is required. It is already used by a component which is exposed via UniFFI.
+
+Note that this option does not preclude both Nimbus and this new crate from moving to the existing remote settings client at some point in the future.
+A key benefit of this decision is that it keeps nimbus and the new crate using the same client, so updating both to use a different client in the future will always remain an option.
+
+## Chosen Option
+
+We have chosen Option 3 because it allows us to reuse the new client in Nimbus, as well as on iOS and on Android with minimal initial development effort.
+If the new library ends up growing requirements that are already in the existing remote settings client, we remain able to copy that functionality from that library into this.
+
+## Specific Plans
+
+This section is non-normative - ie, is not strictly part of the ADR, but exists
+for context.
+
+This is a very high-level view of the tasks required here.
+
+* Create a new top-level component in the application-services repository, identify the exact API we wish to expose for this new library, describe this API using UniFFI, then implement the API with “stubs” (eg, using rust `todo!()`or similar). This is depicted as `RemoteSettings` in the diagram.
+
+* Identify which parts of Nimbus should be factored out into a shared component (depicted as `rs-client` in the diagram below) and move that functionality to the new shared component. Of note:
+    * This component probably will not have a UniFFI .udl file, but is just for consumption by the new component above and the existing nimbus component.
+    * There is still some uncertaintly here - if it is a requirement that nimbus and the new component share some configuration or initialization code, we might need to do something more complex here. This seems unlikely, but possible, so is included here for completeness.
+* Identify which of the nimbus tests should move to the new client and move them.
+* Update Nimbus to take a dependency on the new component and use it, including tests.
+* Flesh out the API of the new top-level component using the new shared component (ie, replace the `todo!()` macros with real code.)
+* Identify any impact on the Nimbus android/swift code - in particular, any shared configuration and initialization code identified above in the application-services repo.
+* Implement the Android and iOS code in the application-services repo desired to make this an ergonomic library for the mobile platforms.
+* Update the mobile code for the UniFFI changes made to Nimbus, if any.
+* Implement the mobile code which consumes the new library, including tests.
+* Profit?
+
+This diagram attempts to depict this final layout. Note:
+* `rs-client` and `RemoteSettings` are both new components, everything else already exists. **Please do not consider these names as suggestions!** Names are hard, I'm sure we can do better.
+* Dashed lines are normal Rust dependencies (ie, dependencies listed in `Cargo.toml`)
+* Solid lines are where the component uses UniFFI
+* Viaduct is a little odd in that it is consumed by the mobile applications indirectly (eg, via Glean), hence it's not in `support`, but please ignore that anomaly.
+
+```mermaid
+flowchart RL
+    subgraph app-services-support[Shared components in application-services/components/support]
+    rs-client
+    other-support-components
+    end
+    subgraph app-services-components[Top-level application-services components, in application-services/components]
+    Nimbus
+    RemoteSettings
+    Viaduct
+    end
+    subgraph mobile [Code in the mobile repositories]
+        Fenix
+        Firefox-iOS
+    end
+
+    Nimbus -- nimbus.udl --> mobile
+    RemoteSettings -- remotesettings.udl --> mobile
+
+    rs-client -.-> Nimbus
+    other-support-components -.-> Nimbus
+    rs-client -.-> RemoteSettings
+    other-support-components -.-> RemoteSettings
+    Viaduct -.-> rs-client
+    other-support-components -.-> rs-client
+```
+
+## Content Signatures
+
+This section is non-normative - ie, is not strictly part of the ADR, but exists
+for context.
+
+Content Signatures have been explicitly called out as a non-requirement. Because this capability was a sticking point in the desktop version of the remote settings client, and because significant effort was spent on it, it's worth expanding on this here.
+
+Because https will be enforced for all network requests, the consumers of this library can have a high degree of confidence that:
+
+* The servers hit by this client are the servers we expect to hit (ie, no man-in-the-middle attacks will be able to redirect to a different server).
+* The response from the server is exactly what was sent by the Mozilla controlled server (ie, no man-in-the-middle attacks will be able to change the content in-flight)
+* Therefore, the content received must be exactly as sent by the Mozilla controlled servers.
+
+Content signatures offer an additional capability of checking the *content* of a remote settings response matches the signature generated with a secret key owned by Mozilla, independenty of the https certificates used for the request itself.
+
+This capability was added to the desktop version primarily to protect the integrity of the data at rest.
+Because the Desktop client cached the responses on disk, there was a risk that this data could be tampered with - so it was effectively impossible to guarantee that the data finally presented to the application is what was initially sent.
+
+The main threat-model that required this capability was 3rd party applications installed on the same system where Firefox was installed.
+Because of the security model enforced by Desktop operating systems (most notably Windows), there was evidence that these 3rd-party applications  would locate and modify the cache of remote-settings responses and modify them in a way that benefited them and caused revenue harm to Mozilla - the most obvious example is changing the search provider settings.
+
+The reason we are declaring this capability a non-requirement in the initial version is two-fold:
+
+* We have also declared caching of responses a non-requirement, meaning there's no data at rest managed by this library which is vulnerable to this kind of attack.
+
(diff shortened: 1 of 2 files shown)


travis79 (commented):
I think separating the networking and remote-settings particulars from Nimbus into a new component will make both easier to maintain and add features to. I would like to see us reuse as much of the existing Rust remote-settings-client work as possible. Happy to help untangle Nimbus from the new component when the time comes.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+## Existing Libraries
+We have identified the following libraries which may be suitable for this project.
+### Remote-settings on desktop
+There is a version of the [remote settings client in desktop](https://searchfox.org/mozilla-central/source/services/settings/RemoteSettingsClient.jsm), written in Javascript.
+It is obviously very capable, but the roadblock to it being suitable for use by our mobile browsers is that it is written in Javascript, so there’s no reasonable path to have it made available to Swift and Kotlin.
+
+### Rust Remote Settings Client
+There is an existing [remote settings client on github](https://github.com/mozilla-services/remote-settings-client).
+This client is written in Rust and has evolved over a number of years.
+The most recent changes were made to support being used in [Merino](https://github.com/mozilla-services/merino), which is currently being re-written in Python - and once that rewrite is complete, there will be no known consumers of this library left.
```
ddurst, line 48: nit: that rewrite is complete.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+
+The main attributes of this library relevant to this discussion are:
+* It’s written in Rust, but has no FFI - ie, it’s currently only consumable by other Rust code.
+* It has recently been updated to use async rust, so requires an internal event loop.
+* It includes the capability to verify the signatures of the content.
+* It maintains a cache of remote resources as JSON files on disk.
+* It performs all network operations via standard Rust libraries.
+
+### The Nimbus-sdk Client
+The [nimbus-sdk is a component in the application-services repository](https://github.com/mozilla/application-services/tree/main/components/nimbus) written in Rust. It has client code which expects to talk to servers using the same format as the remote-settings client, but at time or writing, has only actually been used to talk to the experiments server.
```
ddurst, line 58: nit: s/or/of/

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+* Update Nimbus to take a dependency on the new component and use it, including tests.
+* Flesh out the API of the new top-level component using the new shared component (ie, replace the todo!() macros with real code.)
+* Identify any changes necessary to the Nimbus android/swift code necessary for this - in particular, any shared configuration and initialization code identified above in the application-services repo.
+* Implement the Android and iOS code in the application-services repo desired to make this an ergonomic library for the mobile platforms.
+* Update the mobile code for the UniFFI changes make to Nimbus, if any.
+* Implement the mobile code which consumes the new library, including tests.
+* Profit?
```
ddurst, line 112: (not sure where to put a more general question, so putting it here) Are there concerns or mitigations for when the library has to pull a large set of changes -- like the initial pull for data for Suggest offline users? There were concerns about shipping this amount of data with the applications as well as downloading it on first run, and I'm curious if this addresses that or if that's up to the clients to solve.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+Mozilla’s mobile browsers have a requirement to access the remote settings service, but currently lack any libraries or tools which are suitable without some work.
+A concrete use case is the management of search engine configurations, which are stored in Remote Settings for Firefox Desktop, but shipped as individual files on our mobile browsers, requiring application releases for all changes.
+This document identifies the requirements, then identifies tools which already exist and are close to being suitable, then identifies all available options we can take, and outlines our decision.
+
+### Requirements
+The requirements are for a library which is able to access Mozilla’s Remote Settings service and return the results to our mobile browsers.
+This list of requirements is not exhaustive, but instead focuses on the requirements which will drive our decision making process.
+As such, it identifies the non-requirements first.
+
+### Not a requirement
```
tarikeshaq, line 22: nit: Should the title here be indented i.e a `####` and not a `###`

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+* We must use Necko for all networking on Android, must enforce all connections are via valid https hosts (although some test-only exceptions might be helpful for QA, such as allowing localhost connections to be http)
+* The library should be the only remote-settings library used in the browser. Specifically, this means that Nimbus must also be capable of using the library, and the work to move Nimbus to the library must be considered as part of the project.
+
+## Existing Libraries
+We have identified the following libraries which may be suitable for this project.
+### Remote-settings on desktop
+There is a version of the [remote settings client in desktop](https://searchfox.org/mozilla-central/source/services/settings/RemoteSettingsClient.jsm), written in Javascript.
+It is obviously very capable, but the roadblock to it being suitable for use by our mobile browsers is that it is written in Javascript, so there’s no reasonable path to have it made available to Swift and Kotlin.
```
tarikeshaq, line 43: Can we describe or link to why it's "obviously" very capable 😅 Even something along the lines of "it's been stable in desktop for X long thus it's capable"

tarikeshaq: Hmmm there's maybe a non-option here of "Move the remote settings client on desktop to xul and expose it to android using gecko-view" that doesn't solve the iOS problem and is a lot more work and high risk.. but maybe noting it down for completion?

docs/adr/0005-remote-settings-client.md
tarikeshaq, line 48: Maybe for completion we can link to the decision to re-write merino in python (a footnote even)

docs/adr/0005-remote-settings-client.md
tarikeshaq, line 58: I'm not sure "experiments server" is the right term here - maybe "only been used to access experiments in remote-settings"

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+
+The persistence model used by this library is something that is not a requirement for the new library, which isn’t itself a problem, but it probably would preclude being able to use this library by Nimbus - so the end result is that we would effectively have two remote-settings clients written in Rust and used by our browsers.
+The networking code would need to be replaced with viaduct, which would not be a significant problem assuming we have dropped all async code.
+
+Some API changes would probably be required to make it suitable for use by UniFFI would also be necessary, but these would be tractable.
+
+We would need to update nimbus to use this client, and would need to move the client into the application-services repository, otherwise we would have issues with dependency management (ie, ensuring the upstream version changes continued to work for us, and managing the circular dependency which exists due to this library needing to use viaduct) and would complicate our build process (the library would need to be in our “megazord”, so pulling it remotely would complicate CI and our release process).
```
tarikeshaq, line 84: nit: slightly hard to follow with the brackets, can we have dotted list of the issues we'll come against

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+These are the exact reasons why Nimbus itself is in the application-services repo.
```
tarikeshaq, line 85: Was there a document covering the proposal to move to application-service, linking to that here will provide good context!

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Specific Plans
+
+This section is non-normative - ie, is not strictly part of the ADR, but exists
+for context.
+
+This is a very high-level view of the tasks required here.
+
+* Create a new top-level component in the application-services repository (ie, under application-services/components), identify the exact API we wish to expose for this new library, describe this API using UniFFI, then implement the API with “stubs” (eg, having each method just be a rust todo!() macro or similar.)
+* Identify which parts of Nimbus should be factored out into a shared component (which would live under application-services/components/shared) and move that functionality to the new shared component. This component probably will not have a UniFFI .udl file, but is just for consumption by the new component above and the existing nimbus component (although this is up in the air slightly - if it is a requirement that nimbus and the new component share some configuration or initialization code, we might need to do something more complex here. This seems unlikely, but possible, so is included here for completeness). It is expected that a number of tests will also move here.
```
tarikeshaq, line 105: nit: the brackets make this harder to read a little

tarikeshaq: Can we simplify this with a diagram, so it's easier to follow what new components exists, which ones are exposed to the apps and which are library/shared ones

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+* Update Nimbus to take a dependency on the new component and use it, including tests.
+* Flesh out the API of the new top-level component using the new shared component (ie, replace the todo!() macros with real code.)
+* Identify any changes necessary to the Nimbus android/swift code necessary for this - in particular, any shared configuration and initialization code identified above in the application-services repo.
```
tarikeshaq, line 108: nit: double necessary
```suggestion
* Identify any changes necessary to the Nimbus android/swift code for this - in particular, any shared configuration and initialization code identified above in the application-services repo.
```

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Not a requirement
+The following items all may have some degree of desirability, but they are not hard requirements for the initial version
+* While the https connection to the server must be validated, there is no requirement to verify the content received by the server - ie, there’s no requirement to check the signature of the body itself.
+* There’s no requirement to validate the result of the server conforms to a pre-defined schema - we trust the server data.
+* There’s no requirement to return strongly-typed data to the applications - returning a JSON string/object is suitable.
+* There’s no requirement to cache server responses to the file-system - if the app requests content, it’s fine for the library to always hit the server.
+* There’s no requirement for any kind of scheduling or awareness of network state - when we are requested for content, we do it immediately and return an appropriate error if it can not be fetched.
+* There’s no requirement to support publishing records, requesting reviews or providing approvals via this new library.
+* There’s no requirement to manage buckets, groups and collections via this new library.
```
jaredlockhart, line 30: The desktop remote settings client also uses [webpush](https://github.com/mozilla-services/remote-settings-lambdas#sync_megaphone) to receive changes from the remote settings server, which the experimentation platform can take advantage for rapid enrollment.  Might be worth mentioning that capability as a non-requirement here?

ncalexan:
I was involved moons ago with discussions about Remote Settings clients for use outside of our browser applications, and in particular I implemented a trivial client for the Windows Default Browser Agent -- see https://searchfox.org/mozilla-central/source/toolkit/mozapps/defaultagent/rust/src/lib.rs.  It would be pleasant if we used this new client in that context as well; is that technically feasible with this proposed splitting?  (I will add that I'm not interested in actually doing this replacement: eventually, the WDBA should transition to a [Gecko background task](https://firefox-source-docs.mozilla.org/toolkit/components/backgroundtasks/index.html), removing this use case entirely.)

I'm aware that this ADR explicitly calls out that we don't need support for content signatures, but I would like to see a non-normative section addressing them, since signature validation was an early sticking point and has attracted considerable subsequent effort.  Is there a `viaduct`-like path to supporting NSS for signature validation possible with this library?

jhugman (approved):
Formally, I agree with this decision.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,112 @@
...
+* Deciders:
+    * csadilek for the mobile teams ✔️
+    * bwong for the remote-settings team
+    * mhammond for the application-services team ✔️
+    * [others?]
+* Date: 2022-12-16
+
+## Context and Problem Statement
+
+Mozilla’s mobile browsers have a requirement to access the remote settings service, but currently lack any libraries or tools which are suitable without some work.
```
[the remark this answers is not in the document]
mhammond: Good question - @csadilek do you have a good link for this?

docs/adr/0005-remote-settings-client.md
mhammond: I think the heading should have been `##`! Hopefully addressed in 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: Hopefully addressed both of these in3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: Hopefully fixed in 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: Updated in 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: That's a great question - IIUC, Nimbus has the same general requirement too. I suspect that addressing this will force us to rethink the "doesn't need local persistence" requirement in the future. I'd welcome any suggested changes I should make here.

docs/adr/0005-remote-settings-client.md
mhammond: Good idea - does anyone have a link I can use here?

docs/adr/0005-remote-settings-client.md
mhammond: 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: Good idea, thanks - 3c356141e9177b963fbaf31aa07e9196b343cc27

docs/adr/0005-remote-settings-client.md
mhammond: 3c356141e9177b963fbaf31aa07e9196b343cc27

mhammond:
> I was involved moons ago with discussions about Remote Settings clients for use outside of our browser applications, and in particular I implemented a trivial client for the Windows Default Browser Agent -- see https://searchfox.org/mozilla-central/source/toolkit/mozapps/defaultagent/rust/src/lib.rs. It would be pleasant if we used this new client in that context as well; is that technically feasible with this proposed splitting?

Huh - thanks for that link. I'm sure there will be some devil in the detail, but I can't see any reason this could not be used in that context.

> I'm aware that this ADR explicitly calls out that we don't need support for content signatures, but I would like to see a non-normative section addressing them, since signature validation was an early sticking point and has attracted considerable subsequent effort. Is there a `viaduct`-like path to supporting NSS for signature validation possible with this library?

Great idea, thanks - I added this in b8eb1b9044d685aa022efcce7bd2fb9352b81f7f and would welcome feedback on whether I accurately captured the considerations there.

docs/adr/0005-remote-settings-client.md
mhammond: I tried to simplify this and spent some time trying to find a reasonable way to represent this using Mermaid. I'm only partially happy with the result, so welcome feedback on whether it adds value or makes things worse! Pushed in 3be39fe19bee3b5faa714d7afe26e3dc097c0713

tarikeshaq: I like it! It definitely adds value to me!

docs/adr/0005-remote-settings-client.md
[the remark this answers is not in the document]
csadilek: The concrete use cases is called out right below; to consolidate our search engine configuration across Desktop and Mobile. Here's the [ticket](https://mozilla-hub.atlassian.net/browse/SNT-203), but still in exploration phase. This RS client is foundational for this work though. Various other use cases e.g., storage for wallpaper metadata have been discussed too, but are not as concrete.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,183 @@
...
+
+The main attributes of this library relevant to this discussion are:
+* It’s written in Rust, but has no FFI - ie, it’s currently only consumable by other Rust code.
+* It has recently been updated to use async rust, so requires an internal event loop.
+* It includes the capability to verify the signatures of the content.
+* It maintains a cache of remote resources as JSON files on disk.
+* It performs all network operations via standard Rust libraries.
+
+### The Nimbus-sdk Client
+The [nimbus-sdk is a component in the application-services repository](https://github.com/mozilla/application-services/tree/main/components/nimbus) written in Rust. It has client code which talks to the remote-settings client and while this has only actually been used with the "Nimbus" collection there's no reason to believe it can't be used in the more general case.
```
csadilek, line 59: nit: has client code which talks to the remote-settings **server**?

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,183 @@
...
+* Implement the Android and iOS code in the application-services repo desired to make this an ergonomic library for the mobile platforms.
+* Update the mobile code for the UniFFI changes made to Nimbus, if any.
+* Implement the mobile code which consumes the new library, including tests.
+* Profit?
+
+This diagram attempts to depict this final layout. Note:
+* `rs-client` and `RemoteSettings` are both new components, everything else already exists. **Please do not consider these names as suggestions!** Names are hard, I'm sure we can do better.
+* Dashed lines are normal Rust dependencies (ie, dependencies listed in `Cargo.toml`)
+* Solid lines are where the component uses UniFFI
+* Viaduct is a little odd in that it is consumed by the mobile applications indirectly (eg, via Glean), hence it's not in `support`, but please ignore that anomoly.
```
csadilek, line 126: nit: anomaly.

csadilek (approved):
Thanks, Mark! Responded inline.

docs/adr/0005-remote-settings-client.md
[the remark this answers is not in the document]
mhammond: Yeah, thanks - I think the existing phrasing of "A concrete use case is the management of search engine configurations..." is probably all we need here.

docs/adr/0005-remote-settings-client.md
mhammond: I kinda hate putting google doc links which aren't world viewable in a document like this, which is world viewable. Do you think that doc adds enough value to this ADR that we should include it?

tarikeshaq: That's a good point, maybe just linking to marino-py is good enough!

leplatrem (commented):
I am bit disappointed by this decision. Especially because some criterias were not evaluated precisely (persistence, viaduct, async/await removal...).

I still believe that *Option 2* (moving the existing client into the a-s monorepo) makes more sense on the long term. It supports most Remote Settings features, behind flags or traits, all pluggable. You're right that the initial dev efforts are higher than reusing the current Nimbus client. But on the long term, we are more likely to obtain an official full-featured Remote Settings Rust client that can run in all our products. Especially if, as you mention,  you end up adding caching, signature verification, attachments download, etc. I reiterate my offer: I would be delighted to participate in making the existing code base compatible with a move to the a-s monorepo!

Now, if your decision is already made, there is not much I can do. In that case, it would make more sense for us to archive the remote-settings-client repo. As a maintainer of the Remote Settings service, I would really like to avoid clients diversity as much as possible. Especially if we have to carry evolution on the server, having consistent clients behavior and few code bases to modify is a real benefit. We still have some scars from the numerous forks of the Focus repo :hurtrealbad: 

Would be happy to answer your questions or comments if you have any

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,183 @@
...
+There is a version of the [remote settings client in desktop](https://searchfox.org/mozilla-central/source/services/settings/RemoteSettingsClient.jsm), written in Javascript.
+It has been used and been relatively stable since at least 2018, so can be considered very capable, but the roadblock to it being suitable for use by our mobile browsers is that it is written in Javascript, so while it might be possible to expose it to Android via geckoview, there’s no reasonable path to have it made available to iOS.
+
+### Rust Remote Settings Client
+There is an existing [remote settings client on github](https://github.com/mozilla-services/remote-settings-client).
+This client is written in Rust and has evolved over a number of years.
+The most recent changes were made to support being used in [Merino](https://github.com/mozilla-services/merino), which was re-written in Python, so there are no known consumers of this library left.
+
+The main attributes of this library relevant to this discussion are:
+* It’s written in Rust, but has no FFI - ie, it’s currently only consumable by other Rust code.
```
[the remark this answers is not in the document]
mhammond: Thanks for commenting and I'm sorry to hear you are disappointed with the outcome. Something very relevant that the document doesn't mention is our timelines and resourcing. This work will be carried out by the mobile team, who need this in Q1, and have limited experience with Rust. To be frank, we really were looking for the simplest strategy that would meet our requirements, and offer us a path forward should those requirements change.

The point about signature verification being added for Nimbus is illuminating - Nimbus is well over 2 years old. I can say with absolute certainty that if the existing client was already being used by Nimbus, that is what we would have used - the decision is really "let's use what Nimbus uses" rather than "let's not use the remote settings client". As far as I'm aware, there's no plan to move Nimbus to this client, and we don't think it's reasonable to make that a pre-requisite for this effort when it's clear Nimbus doesn't actually require it to happen and we really don't have the resources to make it happen. The fact no resources have been found in the last 2 years to make that happen was a key factor in sticking with the status quo.

If we did end up requiring signature validation etc, I'm quite confident that we would learn from and directly use as much as possible from that client. Our decision isn't a reflection on the quality of the client, but rather a pragmatic decision based on actual requirements and resourcing. Further, it's not clear that this work would make it significantly more difficult to move to a different client implementation in the future than it is now - the fact we have 2 consumers of the same client instead of one shouldn't make it twice as difficult.

I hope that helps understand where we are coming from, and hope that you will participate with reviews and planning when requirements get more sophisticated - as mentioned above, I don't see any reason this effort would make it much more effort to take the new client in the future than it would now. Please let us know if we can answer any other questions.

docs/adr/0005-remote-settings-client.md
```diff
@@ -0,0 +1,183 @@
...
+* There’s no requirement to return strongly-typed data to the applications - returning a JSON string/object is suitable.
+* There’s no requirement to cache server responses to the file-system - if the app requests content, it’s fine for the library to always hit the server.
+* There’s no requirement for any kind of scheduling or awareness of network state - when we are requested for content, we do it immediately and return an appropriate error if it can not be fetched.
+* There’s no requirement to support publishing records, requesting reviews or providing approvals via this new library.
+* There’s no requirement that push be used to communicate changes to the application (eg, to enable rapid-enrolment type features)
+* There’s no requirement to manage buckets, groups and collections via this new library.
+
+### Initial Requirements
+The requirements we do have for the initial version are:
+* The library should allow fetching records from Mozilla’s Remote Settings servers. This includes support for attachments, and fetching [incremental changes](https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html#polling-for-changes).
```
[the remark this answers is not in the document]
mhammond: I believe the intent here is that the mobile clients might store a representation of the response (ie, the parts it actually cares about) and the timestamp, but @csadilek might be able to talk more about this.

mhammond:
I pushed 1ff4ab53337f3b6f1fb39e0b9ef702641b19860a based on comments from @leplatrem - please let me know if I missed something or if you would like additional changes made.

leplatrem:
> Something very relevant that the document doesn't mention is our timelines and resourcing. This work will be carried out by the mobile team, who need this in Q1, and have limited experience with Rust. 

From what you describe, these aspects drove the decision more than what is/was outlined about the existing client.

My initial frustration mostly came from the fact that you were justifying your final choice using invalid postulates about the existing client. It seemed like it was written backwards: fill up paperwork to sustain the decision made 🙃 Plus, being listed as decider without really having the opportunity to decide can be unpleasant 😉 

Nonetheless, if your driving criterias are resources and skills, it's totally legitimate and valid. But they should be mentioned in the ADR in my opinion. In the end, the role of this document is about providing context for readers of the future.

> [...] I don't see any reason this effort would make it much more effort to take the new client in the future than it would now. 

You are right, and it could be done in parallel. As I said, my fear is their multiplication (desktop, nimbus, ours, and now mobile),  I truly wish there was a shared desire to unify clients. Unfortunately our team spreads thin and we end up missing opportunities like this one.

> [...] and hope that you will participate with reviews and planning when requirements get more sophisticated 

Of course! Always happy to help.  

And in order to understand traffic and activity on our servers, we'd rather keep a close eye on clients implementations.

mhammond:
> > Something very relevant that the document doesn't mention is our timelines and resourcing. This work will be carried out by the mobile team, who need this in Q1, and have limited experience with Rust.
> 
> From what you describe, these aspects drove the decision more than what is/was outlined about the existing client.

That's fair - in 8eba97d2dc18beb411a040ce31210303fbe82d1c I tried to make this a little more explicit.
 
> My initial frustration mostly came from the fact that you were justifying your final choice using invalid postulates about the existing client. It seemed like it was written backwards: fill up paperwork to sustain the decision made 🙃 Plus, being listed as decider without really having the opportunity to decide can be unpleasant 😉

I see your point, but I don't think that's entirely accurate but is instead the nature of these kinds of docs. It is indeed true that @csadilek and I had discussions about this and agreed on what we thought was the most expedient approach. But in the interests of getting feedback we decided to write it as an ADR and send it to the services RFC list - so this PR *is* that review process and your opportunity to object and/or have more input into the decision. While the resourcing constraints did play a big part, this process exists so that if we were making a huge mistake or missed key points which meant we should ask for additional time/resources, they would be identified.

> > [...] and hope that you will participate with reviews and planning when requirements get more sophisticated
> 
> Of course! Always happy to help.
> 
> And in order to understand traffic and activity on our servers, we'd rather keep a close eye on clients implementations.

That's great, thank you! If you don't mind, please see if there's anything in 8eba97d2dc18beb411a040ce31210303fbe82d1c which should be expanded on, and if you don't mind, please let me know how you or your team would like to be represented in the head of that document - either as an explicit "agree/disagree/abstain" or removal of that line entirely are all fine with me. Once I hear back I'll merge it.

Thanks!

leplatrem (approved):
> please let me know how you or your team would like to be represented in the head of that document 

You can add a check mark along my name in the header 👌

mhammond:
Thanks everyone for your feedback!

PR #5303, opened 2022-12-21
Update incorrect comments and add new comments about record IDs

mhammond:
Sammy and I chatted a little about this in the context of #5277. This will end up causing a (hopefully trivial) conflict in that PR, but  felt it was easier than trying to ask for the comments to be added in that PR.

### DIFF
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -29,7 +29,10 @@ pub struct RemoteTab {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ClientRemoteTabs {
-    pub client_id: String, // Corresponds to the `clients` collection ID of the client.
+    // The fxa_device_id of the client. *Should not* come from the id in the `clients` collection,
+    // because that may or may not be the fxa_device_id (currently, it will not be for desktop
+    // records.)
+    pub client_id: String,
     pub client_name: String,
     #[serde(
         default = "devicetype_default_deser",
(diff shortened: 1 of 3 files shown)


mhammond:
(please merge this once approved, then rebase your PR on top of it keeping these new comments)

components/tabs/src/sync/engine.rs
```diff
@@ -55,6 +55,10 @@ impl ClientRemoteTabs {
         }
     }
 
+    // Note that this should die as part of https://github.com/mozilla/application-services/issues/5199
+    // If we don't have a `RemoteClient` record, then we don't know whether the ID passed here is
+    // the fxa_device_id (which is must be) or the client_id (which it will be if this ends up being
```
skhamis, line 60: nit: `which it must be`

components/tabs/src/storage.rs
```diff
@@ -29,7 +29,10 @@ pub struct RemoteTab {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ClientRemoteTabs {
-    pub client_id: String, // Corresponds to the `clients` collection ID of the client.
+    // The fxa_device_id of the client. *Should not* come from the id in the `clients` collection,
+    // because that may or may not be the fxa_device_id (currently, it will not be for desktop
+    // records.)
+    pub client_id: String,
```
skhamis, line 35: I was half tempted to just rename this to prevent confusion, but it's used quite a bit in the clients -- so the comment will be perfect for now

skhamis (approved):
Small nit but I'll just fix it in my PR -- thanks for adding this! Definitely clears up some details about how the client_id is actually being used.

Issue #5235, opened 2022-11-15
places-url-parse-error


bendk:
This is our number 2 issue by number of events.    A quick scan of the events shows that they are happening in `noteHistoryMetadataObservation()` and the message reads `relative URL without a base` (similar to #5233)`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3490)
┆Epic: Error handling
┆Sprint End Date: 2023-01-06

mhammond:
See also #4909

bendk:
Sentry errors from `96.0.1` indicate the issue is an "invalid scheme" -- based on this definition:

```
alpha *( alpha | digit | "+" | "-" | "." )
```

No clue why it doesn't, maybe we could check in with the android team and see if they have any guesses?  We could also log the invalid char, but I would want data review to be sure that doesn't constitute a PII leak.

bendk:
Looks like we've figured this out as well as we can, the next step would probably be add code in android-components to figure out where the invalid schemes are coming from (maybe [here](https://github.com/mozilla-mobile/firefox-android/blob/9a41f9e49ccab08075a0dd067550804f6bf0f4c7/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt#L403-L429)?)

However, it doesn't seem that this issue is worth investing more effort in at this point, so the app-services team is going to put it on hold.

bendk:
Checked the sentry logs after this and the volume is down, but there are still many reports.   A quick scan showed `invalid IPv4 address`, ` invalid domain character`, and  `invalid international domain name` as the top errors.

I sent a message to the android team asking what to do.  IMO we should either investigate more, or stop reporting all invalid URL errors from `note_history_metadata_observation`

bendk:
I asked on Sentry and didn't get any takers to fix this one, so let's just accept that we're going to get invalid URLs sometimes and stop reporting Sentry errors for them.  Made #5304 to do that.

bendk:
#5304 got merged a couple days ago, closing this one.

PR #5304, opened 2022-12-21
Don't report URL parsing errors

bendk:
There's a known issue where Android sends invalid URLs to us.  We've gotten about as much as we can from the Sentry reports and at this point they're more noise then helpful.

See https://github.com/mozilla/application-services/issues/5235 for details.

  - Note:

### DIFF
--- a/components/places/src/error.rs
+++ b/components/places/src/error.rs
@@ -209,16 +209,12 @@ impl GetErrorHandling for Error {
                 .report_error("places-invalid-place-info")
             }
             Error::UrlParseError(e) => {
-                let handling = ErrorHandling::convert(PlacesApiError::UrlParseFailed {
+                // This is a known issue with invalid URLs coming from Fenix. Let's just log a
+                // warning for this one. See #5235 for more details.
+                ErrorHandling::convert(PlacesApiError::UrlParseFailed {
                     reason: e.to_string(),
-                });
-                match e {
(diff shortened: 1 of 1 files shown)


mhammond (approved):
This looks fine to me, but I don't really understand why we don't just never report `UrlParseError` whereever it happens? In what cases do we now care enough to see them in sentry? It seems highly unlikely any such errors will tell us about a bug in our component?

(Approving just incase there is a good reason I don't understand, but less is more IMO :)

bendk:
> This looks fine to me, but I don't really understand why we don't just never report `UrlParseError` whereever it happens? In what cases do we now care enough to see them in sentry? It seems highly unlikely any such errors will tell us about a bug in our component?
> 
> (Approving just incase there is a good reason I don't understand, but less is more IMO :)

I was worried that we might start seeing them from other code paths in the future and we would want those to show up on Sentry.  It does seem unlikely that that would reveal an error in our code though.  Maybe it would reveal an error from the a-c code, but most likely it would just reveal issue we already know about.

You've convinced me, let's just stop reporting `UrlParseError`.  I don't think the small chance that it would catch an error is worth the extra complexity.

bendk:
Technically the new changes haven't been approved, but I'm going to listen to my inner @mhammond and merge this anyway since you approved the previous changes, the new changes follow your suggestion and they are very simple.

PR #5306, opened 2022-12-22
[Don't Merge] Expose the `logins::EncryptorDecryptor` as an UniFFI interface

bendk:
Change the logins API to use `EncryptorDecryptor` directly rather than inputting keys.  The main benefit of this is that we only construct the `EncryptorDecryptor` once, rather than for each API call.

Also, the autofill API is very similar to the logins one.  We could consider moving `EncryptorDecryptor` to some shared place, like a support crate, and using that for both.  That would avoid some code duplication.  It also is one way to sidestep an issue we have with Swift namespaces for a bit longer.  Currently, if we started using `autofill.udl` on Swift we would have a name collision with `create_key() (see uniffi-rs#1426 for details on this).

Don't merge this one yet for a couple reasons:
  - We'd need to resolve the breaking changes for Swift and Kotlin first
  - We're not sure how we want to handle Objects like this on Kotlin yet.  Right now we'd need to manually call the `destroy()` function for each `EncryptorDecryptor`.  Part of the motivation for this PR is to provide an example for the general discussion.

  - Note:

### DIFF
--- a/components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
+++ b/components/logins/android/src/main/java/mozilla/appservices/logins/DatabaseLoginsStorage.kt
@@ -80,30 +80,30 @@ class DatabaseLoginsStorage(dbPath: String) : AutoCloseable {
     }
 
     @Throws(LoginsApiException::class)
-    fun findLoginToUpdate(look: LoginEntry, encryptionKey: String): Login? {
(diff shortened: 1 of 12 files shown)


tarikeshaq:
@bendk I think we landed some of this elsewhere right? is this safe to close?

bendk:
I don't think we merged this, but it's not worth keeping around.  It would be nice to revisit this sometime, but I don't have the bandwidth right now to do that.

PR #5315, opened 2023-01-06
Fix serialization and deserialization of `Option<DeviceType>` in sync15.

mhammond:
We previously could not roundtrip a JSON string which has an unknown device_type value - we'd deserialize it as `Some(DeviceType::Unknown)`, which we'd then fail to serialize as we explicitly forbid serializing a `DeviceType::Unknown`.

We fixed this in 2 ways:
* Serializing `Some(DeviceType::Unknown)` now serializes it as though it was None (ie, as null) - just in-case our Rust code manages to explicitly set it to `Some(DeviceType::Unknown)`.
* Deserializing an unknown enum value now stores it as `None` instead of `Some(DeviceType::Unknown)` (ie, so the point above should never actually happen in practice)

Sammy and I collaborated on this patch, so also asking @bendk for review as a sanity check with a fresh pair of eyes.

  - Note:

### DIFF
--- a/components/sync15/src/client_types.rs
+++ b/components/sync15/src/client_types.rs
@@ -5,12 +5,12 @@
 //! This module has to be here because of some hard-to-avoid hacks done for the
 //! tabs engine... See issue #2590
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use std::collections::HashMap;
 
 /// Argument to Store::prepare_for_sync. See comment there for more info. Only
 /// really intended to be used by tabs engine.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ClientData {
     pub local_client_id: String,
     /// A hashmap of records in the `clients` collection. Key is the id of the record in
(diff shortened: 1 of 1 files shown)


bendk (approved):
The code seems correct to me.

It seems like we should be moving to eliminate `DeviceType::Unknown`, is this change part of that plan?  It seems like a step forward to me, since now we that variant will never be deserialized, but I don't know what the next step would be.

skhamis (approved):
I agree with @bendk, we should work towards removing that "Unknown" variant as it creates these complexities. I think with the landing of external types for swift though we can at least consolidate this DeviceType (right now it's duplicated for client, tabs, fxa)

mhammond:
> It seems like we should be moving to eliminate `DeviceType::Unknown`,

That's a great question: we should either remove `DeviceType::Unknown` or remove all use of `Option<DeviceType>`. I'm torn which one to do:

* Removing `DeviceType::Unknown` (thus forcing everyone to use `Option<DeviceType>`) has a bit of a footgun with serde - it seems impossible to add a `derive Deserialize for Option<DeviceType>` (as Option isn't in the crate), meaning that adding a simple `Option<DeviceType>` without specifying a custom deserializer compiles and seems to work, but will still fail in practice for unknown variants. It also feels mildly odd to be using 2 enums here.

* Removing `Option<DeviceType>` works fairly well and does still require a custom deserializer, but once implemented, there's no footgun left - you can just use `DeviceType` in any struct without special serde annotations, and a missing or invalid value always gives you `DeviceType::Unknown`. It seems slightly cleaner to me - there's only one Enum and there's no need to guess what `None/null/etc`  means.

I've had a bit of a poke at both of these, and think I prefer the 2nd option, only because of the footgun, but I'd welcome thoughts.

mhammond:
See #4861 for a PR which keeps `DeviceType::Unknown` but removes all `Option<DeviceType>`

PR #5324, opened 2023-01-11
Don't try and link with sha-x86_c_lib for Android x86 targets

mhammond:
Fixes gradle on Windows for me.


mhammond:
hrm, I obviously need to investigate this more.

badboy:
fwiw I had to remove this just now to get iOS building.

PR #5326, opened 2023-01-11
Adding HOWTO guide on UniFFI objects destruction

bendk:
Maybe this says more about me than anything, but I find the [prevailing Java/Kotlin wisdom on object destruction](https://www.informit.com/articles/article.aspx?p=1216151&seqNum=7) confusing when I start to think about it in relation to how we're using UniFFI objects on Android.  At one point I started to wonder if we're currently breaking the rules and/or we would be in the future after we introduced more Objects, like in #5306.  I started to write up ADRs about how we should handle them.

However, after discussing this with @mhammond and @csadilek  and thinking this through a bit on my own, my conclusion is that how we're handling them is fine and not that restrictive.  Instead of an ADR, what we need is just a howto that describes what's allowed based on the current way we're currently using them.

Hopefully we can agree that these guidelines are safe and good to use, but please push back if any that seem dangerous.

  - Note:

### DIFF
--- a/docs/howtos/uniffi-object-destruction-on-kotlin.md
+++ b/docs/howtos/uniffi-object-destruction-on-kotlin.md
@@ -1,0 +1,35 @@
+# UniFFI object destruction on Kotlin
+
+UniFFI supports [interface objects](https://mozilla.github.io/uniffi-rs/udl/interfaces.htm), which are implemented by Boxing a Rust object and sending the raw pointer to the foreign code.  Once the objects are no longer in use, the foreign code needs to destroy the object and free the underlying resources.
+
+This is slightly tricky on Kotlin.  [The prevailing Java wisdom is to use explicit destructors and avoid using finalizers for destruction](https://www.informit.com/articles/article.aspx?p=1216151&seqNum=7), which means we can't simply rely on the garbage collector to free the pointer.  The wisdom seems simple to follow, but in practice it can be difficult to know how to apply it to specific situations.  This document examines provides guidelines for handling UniFFI objects.
+
+## You can create objects in a function if you also destroy them there
+
+The simplest way to get destruction right is to create an object and destroy it in the same function.  The [use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html) function makes this really easy:
+
+```
+SomeUniFFIObject()
+  .use { obj ->
+      obj.doSomething()
+      obj.doSomethingElse()
+  }
+```
+
+## You can create and store objects in singletons
+
+If we are okay with UniFFI objects living for the entire application lifetime, then they can be stored in singletons.  This is how we handle our database connections, for example [SyncableLoginsStorage](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L98-L114) and [PlacesReaderConnection](https://github.com/mozilla-mobile/firefox-android/blob/9de5ab8cc098674aad5190ce8b00ae6be65e9bd0/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt#L75-L96).
+
+## You can create and store objects in an class, then destroy them in a corresponding lifecycle method
+
+UniFFI objects can stored in classes like the Android [Fragment](https://developer.android.com/guide/fragments/lifecycle) class that have a defined lifecycle, with methods called at different stages.  Classes can construct `UniFFI` objects in one of the lifecycle methods, then destroy it in the corresponding one.  For example, creating an object in `Fragment.onCreate` and destroying it in `Fragment.onDestroy()`.
+
+## You can share objects
+
+Several classes can hold references to an object, as long as (exactly) one class is responsible for managing it and destroying it when it's not used. A good example is the [GeckoLoginStorageDelegate](https://github.com/mozilla-mobile/firefox-android/blob/009fb6350072b8b4cf2b87bd1a8f497843f67843/android-components/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt#L46-L49).  The `LoginStorage` is initialized and managed by another object, and `GeckoLoginStorageDelegate` is passed a (lazy) reference to it.
+
+Care should be taken to ensure that once the managing class destroys the object, no other class attempts to use it.  If they do, then the generate code will raise an `IllegalStateException`.  This clearly should be avoided, although it won't result in memory corruption.
+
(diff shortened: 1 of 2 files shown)


docs/howtos/uniffi-object-destruction-on-kotlin.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## You can create and store objects in singletons
+
+If we are okay with UniFFI objects living for the entire application lifetime, then they can be stored in singletons.  This is how we handle our database connections, for example [SyncableLoginsStorage](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L98-L114) and [PlacesReaderConnection](https://github.com/mozilla-mobile/firefox-android/blob/9de5ab8cc098674aad5190ce8b00ae6be65e9bd0/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt#L75-L96).
+
+## You can create and store objects in an class, then destroy them in a lifecycle method
+
+UniFFI objects can stored in classes like the Android [Fragment](https://developer.android.com/guide/fragments/lifecycle) class that have a defined lifecycle, with methods called at different stages.  The class just needs to ensure that the UniFFI object is destroyed in one of the lifecycle methods, for example `Fragement.onDestroy()`.
+
```
bendk, line 26: Maybe this is slightly controversial, since unlike the other guidelines, there's no code that's doing something like this currently.  However, it seems pretty safe to me, what do others think?

docs/howtos/uniffi-object-destruction-on-kotlin.md
```diff
@@ -0,0 +1,37 @@
+# UniFFI object destruction on Kotlin
+
+UniFFI supports [interface objects](https://mozilla.github.io/uniffi-rs/udl/interfaces.htm), which are implemented by Boxing a Rust object and sending the raw pointer to the foreign code.  Once the objects are no longer in use, the foreign code needs to destroy the object and free the underlying resources.
+
+This is slightly tricky on Kotlin.  [The prevailing Java wisdom is to use explicit destructors and avoid using finalizers for destruction](https://www.informit.com/articles/article.aspx?p=1216151&seqNum=7), which means we can't simply rely on the garbage collector to free the pointer.  The wisdom seems simple to follow, but in practice it can be difficult to know how to apply it to specific situations.  This document examines provides guidelines for handling UniFFI objects.
+
```
mhammond, line 6: I wonder if it's worth noting somewhere that kotlin classes generated by Uniffi implement `AutoClosable` - I was going to suggest linking to uniffi, but it appears we don't actually document that - I wonder if we should also add something to https://github.com/mozilla/uniffi-rs/tree/main/docs/manual/src/kotlin about this?

docs/howtos/uniffi-object-destruction-on-kotlin.md
```diff
@@ -0,0 +1,37 @@
...
+Unlike in Rust, there's no borrow checker to ensure that code follows these rules.  If we make a mistake, then the borrowing class may try to use the object after it's already been destroyed.  In this case the generated code will raise an `IllegalStateException`.  This should obviously be avoided, but at least it won't result in memory corruption.
+
+## Avoid classes that take ownership of objects.
+
+The alternative to borrowing is taking ownership of objects: Classes that take ownership of the objects they input.  This is how the UniFFI generated code works, for example UniFFI Records/Enums that contain UniFFI objects define their own destroy() method that calls destroy on the child object.
```
mhammond, line 35: I'm not quite sure what this section is trying to say - and isn't it true that taking ownership would actually be fine, so long as you manage that ownership correctly?

IOW, it seems this section and the above "borrow" section are really just saying "you must manage lifetimes correctly. You can "borrow" or you can pass ownership around, but it's on you if you get it wrong - and if you do, you'll get `IllegalStateException` which is never going end well, so don't do that" :)

mhammond:
Thanks Ben, having this guidance makes alot of sense for people (like me!) with relatively little Kotlin experience. I'd love feedback from some of the Android team - cc @csadilek @jonalmeida @MatthewTighe

docs/howtos/uniffi-object-destruction-on-kotlin.md
```diff
@@ -0,0 +1,37 @@
...
+  }
+```
+
+## You can create and store objects in singletons
+
+If we are okay with UniFFI objects living for the entire application lifetime, then they can be stored in singletons.  This is how we handle our database connections, for example [SyncableLoginsStorage](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L98-L114) and [PlacesReaderConnection](https://github.com/mozilla-mobile/firefox-android/blob/9de5ab8cc098674aad5190ce8b00ae6be65e9bd0/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt#L75-L96).
+
+## You can create and store objects in an class, then destroy them in a lifecycle method
+
+UniFFI objects can stored in classes like the Android [Fragment](https://developer.android.com/guide/fragments/lifecycle) class that have a defined lifecycle, with methods called at different stages.  The class just needs to ensure that the UniFFI object is destroyed in one of the lifecycle methods, for example `Fragement.onDestroy()`.
```
MatthewTighe, line 25: Destructive lifecycle methods notoriously are not guaranteed to finish during events like process death (user-instantiated or not). I'm curious whether there are any additional considerations for memory leaks when using UniFFI objects in these cases, but don't know enough about how the OS cleans up to say.

docs/howtos/uniffi-object-destruction-on-kotlin.md
MatthewTighe, line 35: It seems like this section is geared towards object relationships solely in the UniFFI world, as there isn't really a concept of 'ownership' in the JVM. Could it be worth mentioning as much, in case of future Android readers?

docs/howtos/uniffi-object-destruction-on-kotlin.md
bendk: Good point, I'll add a note about this. In general, I think the OS can clean up the resources just fine.  I think custom code is the main issue, like @csadilek mentions.

bendk:
Thanks for all the help with this one!  I just pushed an update of this.  The main change is that I reworked the sections on "ownership" and "borrowing", removing the Rust-centric terminology and trying to frame things in a way that's more familiar to Kotlin coders.

bendk:
Make sure to close https://bugzilla.mozilla.org/show_bug.cgi?id=1799133 once this is merged.

mhammond (approved):

PR #5327, opened 2023-01-11
Adds adr 0005 and 0007 to the mdbook

tarikeshaq:
Those adrs were landed this week, but we forgot to link them in the mdbook


bendk (approved):
Looks good.  I'll make sure to add 0006 to the book as well.

PR #5332, opened 2023-01-13
Update uniffi to 0.22/0.23

mhammond:
Note sure when we should land this - we are hoping to do another vendor into desktop after the next merge - glean is already here but m-c isn't. Should we wait for @bendk 's  version-revamp? @badboy?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,7 +209,6 @@ dependencies = [
  "thiserror",
  "types",
  "uniffi",
- "uniffi_build",
  "url",
 ]
 
@@ -595,8 +594,6 @@ dependencies = [
  "log",
  "thiserror",
  "uniffi",
- "uniffi_build",
- "uniffi_macros",
 ]
 
 [[package]]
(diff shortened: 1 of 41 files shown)


bendk:
My plan is to get uniffi v`0.23` out early next week, with the `combined-frontend` changes, then update app-services to use that after coordinating with @badboy.  So maybe this should wait as long as https://github.com/mozilla/uniffi-rs/pull/1445 isn't controversial?

mhammond:
Totally fine with waiting - I really just put it up because of all the dependency changes to hopefully avoid someone else going through it (and most of that should be relevant in the next version too)

badboy:
We're about to land a Glean update in m-c (tomorrow to hit the 111 nightly cycle). That has uniffi 0.21 still.
After that I have a fix pending to unbreak Fenix tests. Probably best to have that on the same UniFFI in case we also uplift that.

After that I'm fine with either. Glean main is on 0.22 already.

mhammond:
I'm fine waiting. In the meantime I pushed another commit which does most of the work required for updating uniffi to the next, as yet unreleased version. It's pointing at a branch of mine and will obviously require a number of tweaks once the new version is released, but it might save someone else some work.

Issue #5345, opened 2023-01-24
Create a Rust component for a Merino client


linabutler:
As part of bringing Desktop's Firefox Suggest Online experience to Firefox for Android and iOS, we'd like to add a new Rust component that fetches search suggestions from the Merino server. The API surface of this component will be pretty small—make an HTTP request, parse the response with Serde, and expose it via UniFFI to Kotlin and Swift—and feels like a great way for our team to learn how we ship cross-platform components on mobile!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2219)
┆Epic: Merino on Mobile Experiment
┆Sprint End Date: 2023-02-07

tarikeshaq:
WOHOO I'm super excited to see more cross-platform components, so please feel free to let me know if there is any way the sync team can support here (doubt you'll need much of it 😛) 

Happy to provide reviews, an extra pair of eyes wherever needed, or any type of admin work on the repo

linabutler:
Yep, it is!

PR #5346, opened 2023-01-25
Create a Rust component for a Merino client (#5345)

linabutler:

  - Note:

### DIFF
--- a/.buildconfig-android.yml
+++ b/.buildconfig-android.yml
@@ -99,6 +99,13 @@ projects:
     - name: nimbus
       type: aar
     description: The Nimbus SDK, compiled to work with the Application Services full-megazord.
+  merinoclient:
+    path: components/merino_client/android
+    artifactId: merinoclient
+    publications:
+    - name: merinoclient
+      type: aar
+    description: Firefox Suggest Service client.
   full-megazord:
     uploadSymbols: true
     path: megazords/full/android
--- a/components/merino_client/Cargo.toml
+++ b/components/merino_client/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "merino_client"
+edition = "2021"
+version = "0.1.0"
+authors = ["contextual-services-team@mozilla.com"]
+license = "MPL-2.0"
+exclude = ["/android", "/ios"]
+
+[dependencies]
+anyhow = "1.0"
+error-support = { path = "../support/error" }
+log = "0.4"
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
+thiserror = "1.0"
+uniffi = "^0.21"
+uniffi_macros = "^0.21"
+url = "2.1" # mozilla-central can't yet take 2.2 (see bug 1734538)
+viaduct = { path = "../viaduct" }
+
+[build-dependencies]
+uniffi_build = { version = "^0.21", features = [ "builtin-bindgen" ]}
+
+[dev-dependencies]
+viaduct-reqwest = { path = "../support/viaduct-reqwest" }
--- a/components/merino_client/Cargo.toml
+++ b/components/merino_client/Cargo.toml
@@ -26,3 +26,4 @@ uniffi_build = { version = "^0.21", features = [ "builtin-bindgen" ]}
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
+mockito = "0.31"
--- a/components/merino_client/Cargo.toml
+++ b/components/merino_client/Cargo.toml
@@ -9,7 +9,6 @@ exclude = ["/android", "/ios"]
 [dependencies]
 anyhow = "1.0"
 error-support = { path = "../support/error" }
-hex = "0.4"
 log = "0.4"
 rc_crypto = { path = "../support/rc_crypto" }
 serde = "1"
--- a/components/merino_client/Cargo.toml
+++ b/components/merino_client/Cargo.toml
@@ -9,7 +9,9 @@ exclude = ["/android", "/ios"]
 [dependencies]
 anyhow = "1.0"
 error-support = { path = "../support/error" }
+hex = "0.4"
 log = "0.4"
+rc_crypto = { path = "../support/rc_crypto" }
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
--- a/components/merino_client/src/client.rs
+++ b/components/merino_client/src/client.rs
@@ -257,3 +257,125 @@ impl SessionState {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use mockito::{mock, Matcher};
+
+    #[test]
+    fn fetch_adm_suggestion() -> MerinoClientResult<()> {
+        viaduct_reqwest::use_reqwest_backend();
+        let m = mock("GET", "/api/v1/suggest")
+            .match_query(Matcher::UrlEncoded("q".into(), "test".into()))
+            .with_status(200)
+            .with_header("Content-Type", "application/json")
+            .with_body(
+                r#"{
+                    "suggestions": [{
+                        "title": "Test adM suggestion",
+                        "url": "https://example.com",
+                        "is_sponsored": false,
+                        "score": 0.1,
+                        "provider": "adm",
+                        "block_id": 1,
+                        "full_keyword": "testing",
+                        "advertiser": "Test"
+                    }]
+                }"#,
+            )
+            .create();
+
+        let client = MerinoClient::new(MerinoClientSettings {
+            server: MerinoServer::Custom {
+                url: mockito::server_url(),
+            },
+            session_duration_ms: 5000,
+            client_variants: vec![],
+            default_providers: vec![],
+        })?;
+        let suggestions = client.fetch("test", None)?;
+        m.expect(1).assert();
+
+        assert_eq!(suggestions.len(), 1);
+        match &suggestions[0] {
+            MerinoSuggestion::Adm {
+                details,
+                block_id,
+                full_keyword,
+                advertiser,
+                ..
+            } => {
+                assert_eq!(
+                    details,
+                    &MerinoSuggestionDetails {
+                        title: "Test adM suggestion".into(),
+                        url: "https://example.com".into(),
+                        is_sponsored: false,
+                        score: 0.1,
+                        icon: None,
+                    }
+                );
+                assert_eq!(*block_id, 1);
+                assert_eq!(full_keyword, "testing");
+                assert_eq!(advertiser, "Test");
+            }
+            _ => assert!(false, "Wanted adM suggestion; got {:?}", suggestions[0]),
+        };
+
+        Ok(())
+    }
+
+    #[test]
+    fn fetch_unknown_provider_suggestion() -> MerinoClientResult<()> {
+        viaduct_reqwest::use_reqwest_backend();
+        let m = mock("GET", "/api/v1/suggest")
+            .match_query(Matcher::UrlEncoded("q".into(), "test".into()))
+            .with_status(200)
+            .with_header("Content-Type", "application/json")
+            .with_body(
+                r#"{
+                    "suggestions": [{
+                        "title": "Test suggestion",
+                        "url": "https://example.com",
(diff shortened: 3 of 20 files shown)


components/merino_client/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+anyhow = "1.0"
+error-support = { path = "../support/error" }
+log = "0.4"
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
+thiserror = "1.0"
+uniffi = "^0.21"
+uniffi_macros = "^0.21"
+url = "2.1" # mozilla-central can't yet take 2.2 (see bug 1734538)
```
linabutler, line 21: It'll be a while before we can integrate Merino client into Desktop—or we might decide not to—but I copied this line from another component's `Cargo.toml` and kept it, because it's a good reminder just in case.

components/merino_client/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+viaduct = { path = "../viaduct" }
```
linabutler, line 22: Viaduct is the library that we use to make HTTP requests.

* On Android, it uses "Necko", the Gecko networking stack—Android Components sets up the Viaduct backend at launch, then Viaduct calls into Android Components, a-c makes the request via GeckoView, and passes the response back through the different layers to Viaduct.
* On iOS (and in tests), it uses the same system network stack as WebKit, via `viaduct-reqwest`.
* On Desktop, it makes requests using Necko directly.

megazords/full/Cargo.toml
```diff
@@ -21,5 +21,6 @@ nimbus-sdk = { path = "../../components/nimbus" }
 autofill = { path = "../../components/autofill" }
 crashtest = { path = "../../components/crashtest" }
 error-support = { path = "../../components/support/error" }
+merino_client = { path = "../../components/merino_client" }
```
linabutler, line 24: "Megazording" means "bundling all our Rust components in a single library" 🤖 Here's a [quick overview](https://mozilla.github.io/application-services/book/design/megazords.html) about it, and why we do it!

.buildconfig-android.yml
```diff
@@ -99,6 +99,13 @@ projects:
     - name: nimbus
       type: aar
     description: The Nimbus SDK, compiled to work with the Application Services full-megazord.
+  merinoclient:
```
linabutler, line 102: Most of the first commit is boilerplate from [this wonderful guide](https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html) for adding a new component.

components/merino_client/src/client.rs
```diff
@@ -0,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use serde_derive::*;
+
+use crate::error::{MerinoClientError, Result};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MerinoClientOptions {
```
linabutler: This is the minimal implementation of `merino_client.udl`! I found it helpful to have the UDL file opened side-by-side with this Rust code as I worked. 😊

components/merino_client/src/merino_client.udl
```diff
@@ -0,0 +1,38 @@
+namespace merino_client {
```
linabutler, line 1: The UniFFI book has a [section](https://mozilla.github.io/uniffi-rs/udl_file_spec.html) explaining the UDL file and its syntax.

components/merino_client/Cargo.toml
mhammond: We should be able to get [workspace dependencies](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) soon - we *almost* had it for UniFFI, but there was a problem vendoring into desktop so was backed out in 6ecde0b247d9c9b8129c680b454f13605c4349dc and I really can't remember whether that problem was uniffi specific or not, but @bendk said he thinks we should be able to get it sorted soon.

bendk: Actually, I'm not so sure this can be sorted.  When our crates get vendored into moz-central, we don't have our workspace `Cargo.toml` anymore so this doesn't work.  I think this is fixable, but I've decided to just duplicate the dependencies for a while more and not try to fix it.

components/merino_client/src/client.rs
```diff
@@ -79,19 +76,12 @@ impl MerinoClient {
...
-            })?
-            .require_success()
-            .map_err(|err| MerinoClientError::FetchFailed {
-                reason: err.to_string(),
-            })?
-            .json()
-            .map_err(|err| MerinoClientError::FetchFailed {
-                reason: err.to_string(),
-            })?;
+        let response = (|| -> Result<SuggestResponse, InternalError> {
```
linabutler, line 79: This is a pattern I learned from [kvstore](https://searchfox.org/mozilla-central/rev/81fbf5597044b6811e6aa4a10f3d78e8c3080f8c/toolkit/components/kvstore/src/task.rs#192-193,195,211)!

linabutler (commented):
@tiftran I left some comments on each commit, with some pointers that might be interesting for you! 😊

components/merino_client/src/merino_client.udl
```diff
@@ -0,0 +1,50 @@
...
+    boolean is_sponsored;
+    f64 score;
+    string? icon;
+};
+
+[Enum]
+interface MerinoSuggestion {
+    Adm(MerinoSuggestionDetails details, i64 block_id, string full_keyword, string advertiser, string? impression_url, string? click_url);
+    TopPicks(MerinoSuggestionDetails details, i64 block_id, boolean is_top_pick);
+    Other(MerinoSuggestionDetails details, string provider);
```
linabutler, line 41: suggestion (blocking): Let's handle `AccuWeather` suggestions, too. I think adding them will be a fun exercise that covers updating the UniFFI definition, Serde deserialization, and tests! 🎉 

/cc @tiftran

components/merino_client/src/client.rs
```diff
@@ -0,0 +1,381 @@
...
+        Self {
+            started_at: Instant::now(),
+            session_id: hex::encode(&session_id_bytes),
+            sequence_number: 0,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
linabutler, line 262: suggestion (blocking): Let's add some more tests! Ideas for test cases:

* Other kinds of suggestions (Top Picks, AccuWeather).
* Specifying additional settings, like the default providers, client variants.
* Explicitly specifying `providers` when calling `fetch`.
* Session ID and sequence number rotation.
* Error responses (non-200 response code, bad JSON in body).

/cc @tiftran

linabutler (commented):
Some things for us to work on!

I'm trying out ["conventional comments"](https://conventionalcomments.org/) for review feedback—what do you think? I'm not sure how I feel about them yet... 😄

components/merino_client/src/merino_client.udl
```diff
@@ -26,14 +26,19 @@ dictionary MerinoClientFetchOptions {
...
     string url;
-    string provider;
     boolean is_sponsored;
     f64 score;
     string? icon;
-    string request_id;
+};
+
+[Enum]
+interface MerinoSuggestion {
```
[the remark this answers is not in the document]
linabutler: thought: Even though we can't have inheritance, I've found it super helpful to have accessors for the details in both Kotlin and Swift:

```kt
val MerinoSuggestion.details: MerinoSuggestionDetails
    get() = when (this) {
        is MerinoSuggestion.Adm -> details
        is MerinoSuggestion.TopPicks -> details
        is MerinoSuggestion.Other -> details
    }
```

```swift
extension MerinoSuggestion {
    var details: MerinoSuggestionDetails {
        switch self {
        case .adm(let details, _, _, _, _, _):
            return details
        case .topPicks(let details, _, _):
            return details
        case .other(let details, _):
            return details
        }
    }
}
```

That lets us write code like `suggestion.details.url` and `suggestion.details.icon`, instead of having to pattern-match on the suggestion with `when`/`switch`, and extract the details every time we want a base suggestion field.

UniFFI enum interfaces can't have methods or getters, so we can't put that accessor in the UDL—but we can still add [hand-written Kotlin](https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html#the-kotlin-bindings) and [Swift](https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html#creating-the-directory-structure) source files with those extensions to our component! The FxA client does something like that in its [Kotlin](https://github.com/mozilla/application-services/blob/f0cf5491ce6f3ccdd80913b9a9a4c3568c6cb2da/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/Config.kt) and [Swift](https://github.com/mozilla/application-services/blob/f0cf5491ce6f3ccdd80913b9a9a4c3568c6cb2da/components/fxa-client/ios/FxAClient/FxAccountConfig.swift) bindings.

This would be a great exercise for getting more familiar with the component directory structure, and updating the Xcode project to include the hand-written Swift extension!

/cc @tiftran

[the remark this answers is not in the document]
mhammond: > thought: Even though we can't have inheritance, I've found it super helpful to have accessors for the details in both Kotlin and Swift:

That's really nice! I wonder if we can teach UniFFI about that?

> UniFFI enum interfaces can't have methods or getters, so we can't put that accessor in the UDL—but we can still add [hand-written Kotlin](https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html#the-kotlin-bindings) and [Swift](https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html#creating-the-directory-structure) source files with those extensions to our component!

This raises an interesting point. The hand-written foreign code next to the component was initially hugely valuable - our hand-written FFI was so painful that wrappers made a lot of sense. However, as UniFFI means the bindings are quite ergonomic, the app-services team has had an informal decision to keep these hand-wrappers as small as possible - if at all. Part of the reason was that we found understanding old code became quite difficult, particularly for Fenix - by the time our component ended up being consumed we were often many levels of indirection away from the component itself - we have our hand-wrappers, a couple of wrappers in android-components (often a "concept" and a "service"), then the final consumer. Nimbus has gone the other way (the "canonical" wrappers are next to their component). I don't think there is a single correct answer here, but it is a tension that we still feel.

(One example of what I'm referring to is our [autofill component](https://github.com/mozilla/application-services/tree/main/components/autofill) - the only [hand-written kotlin there is working around a UniFFI limitation](https://github.com/mozilla/application-services/blob/main/components/autofill/android/src/main/java/mozilla/appservices/autofill/CreateKey.kt) (or probably more accurately, a limitation in our megazord layout) - ie, so similar to your example above. But any additional ergonomic improvements tends to be in our consumer code (eg, in [android-components](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/sync-autofill/src/main/java/mozilla/components/service/sync/autofill/Types.kt))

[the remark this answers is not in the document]
linabutler: > That's really nice! I wonder if we can teach UniFFI about that?

I filed https://github.com/mozilla/uniffi-rs/issues/1470. 😄

> However, as UniFFI means the bindings are quite ergonomic, the app-services team has had an informal decision to keep these hand-wrappers as small as possible - if at all.

_nod_ That makes sense, for all the reasons you mentioned—having a substantial amount of hand-written code in Application Services adds another layer of wrapping that can hide important implementation details. I totally agree with your approach of keeping that layer as small as possible (if at all), and saving the ergonomics for our consumers.

linabutler:
We decided to go in a different direction, and focus on shipping the "basic" Firefox Suggest experience, instead of the "improved" experience powered by Merino.

The basic experience downloads all suggestions from a Remote Settings collection, and matches suggestions on-device. Adding a Merino backend for the improved Suggest experience to our new `suggest` component (#5723) is out of scope for now, and we have no concrete plans to ship the work in this PR—so let's close it out.

Thanks, folks!

PR #94, opened 2018-07-03
WIP on logins API based on Mentat

ncalexan:

PR #5359, opened 2023-02-02
Expose sync manager component to iOS

lougeniaC64:
This PR, in addition to (PR #94)[https://github.com/mozilla/rust-components-swift/pull/94] in `rust-components-swift`, expose the sync manager component to iOS for the sync manager integration. In also adds the following:

- `RustSyncTelemetryPing.Swift` and tests, similar to `SyncTelemetryPing.kt` for deserializing the json telemetry data in `SyncResult`
- Renaming sync manager's `DeviceType` to avoid a naming collision with other a-s components and code to avoid the breaking change for android components

This PR should not be merged until a PR to add namespacing to iOS's `SyncReason`, `SyncResult`, and `SyncManager` types is merged.

  - Note:

### DIFF
--- a/components/sync15/ios/RustSyncTelemetryPing.swift
+++ b/components/sync15/ios/RustSyncTelemetryPing.swift
@@ -1,0 +1,435 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+public class RustSyncTelemetryPing {
+    public let version: Int
+    public let uid: String
+    public let events: [EventInfo]
+    public let syncs: [SyncInfo]
+
+    private static let EMPTY_UID = String(repeating: "0", count: 32)
+
+    init(version: Int, uid: String, events: [EventInfo], syncs: [SyncInfo]) {
+        self.version = version
+        self.uid = uid
+        self.events = events
+        self.syncs = syncs
+    }
+
+    static func empty() -> RustSyncTelemetryPing {
+        return RustSyncTelemetryPing(version: 1,
+                                     uid: EMPTY_UID,
+                                     events: [EventInfo](),
+                                     syncs: [SyncInfo]())
+    }
+
+    static func fromJSON(jsonObject: [String: Any]) throws -> RustSyncTelemetryPing {
+        guard let version = jsonObject["version"] as? Int else {
+            throw TelemetryJSONError.intValueNotFound(
+                message: "RustSyncTelemetryPing `version` property not found")
+        }
+
+        let events = unwrapFromJSON(jsonObject: jsonObject) { obj in
+            try EventInfo.fromJSONArray(
+                jsonArray: obj["events"] as? [[String: Any]] ?? [[String: Any]]())
+        } ?? [EventInfo]()
+
+        let syncs = unwrapFromJSON(jsonObject: jsonObject) { obj in
+            try SyncInfo.fromJSONArray(
+                jsonArray: obj["syncs"] as? [[String: Any]] ?? [[String: Any]]())
+        } ?? [SyncInfo]()
+
+        return RustSyncTelemetryPing(version: version,
+                                     uid: try stringOrNull(jsonObject: jsonObject,
+                                                           key: "uid") ?? EMPTY_UID,
+                                     events: events, syncs: syncs)
+    }
+
+    public static func fromJSONString(jsonObjectText: String) throws -> RustSyncTelemetryPing {
+        guard let data = jsonObjectText.data(using: .utf8) else {
+            throw TelemetryJSONError.invalidJSONString
+        }
+
+        let jsonObject = try JSONSerialization.jsonObject(with: data) as? [String: Any] ?? [String: Any]()
+        return try fromJSON(jsonObject: jsonObject)
+    }
+}
+
+public class SyncInfo {
+    public let at: Int64
+    public let took: Int64
+    public let engines: [EngineInfo]
+    public let failureReason: FailureReason?
+
+    init(at: Int64, took: Int64, engines: [EngineInfo], failureReason: FailureReason?) {
+        self.at = at
+        self.took = took
+        self.engines = engines
+        self.failureReason = failureReason
+    }
+
+    static func fromJSON(jsonObject: [String: Any]) throws -> SyncInfo {
+        guard let at = jsonObject["when"] as? Int64 else {
+            throw TelemetryJSONError.intValueNotFound(
+                message: "SyncInfo `when` property not found")
+        }
+
+        let engines = unwrapFromJSON(jsonObject: jsonObject) { obj in
+            try EngineInfo.fromJSONArray(
+                jsonArray: obj["engines"] as? [[String: Any]] ?? [[String: Any]]())
+        } ?? [EngineInfo]()
+
+        let failureReason = unwrapFromJSON(jsonObject: jsonObject) { obj in
+            FailureReason.fromJSON(
+                jsonObject: obj["failureReason"] as? [String: Any] ?? [String: Any]())
+        } as? FailureReason
+
+        return SyncInfo(at: at,
+                        took: int64OrZero(jsonObject: jsonObject, key: "took"),
+                        engines: engines,
+                        failureReason: failureReason)
+    }
+
+    static func fromJSONArray(jsonArray: [[String: Any]]) throws -> [SyncInfo] {
+        var result = [SyncInfo]()
+
+        for item in jsonArray {
+            result.append(try fromJSON(jsonObject: item))
+        }
+
+        return result
+    }
+}
+
+public class EngineInfo {
+    public let name: String
+    public let at: Int64
+    public let took: Int64
+    public let incoming: IncomingInfo?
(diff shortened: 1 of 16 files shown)


components/sync_manager/src/syncmanager.udl
```diff
@@ -48,12 +48,15 @@ interface SyncEngineSelection {
...
-enum SyncReason {
+enum RustSyncReason {
     "Scheduled",
     "User",
     "PreSleep",
     "Startup",
     "EnabledChange",
+    "Backgrounded",
+    "DidLogin",
+    "ClientNameChanged",
```
mhammond, line 59: we might be able to kill this - it looks like iOS already doesn't do the right thing here so we prefer the fxa record for the name. IOW, there's probably no need to actually sync when the name changes.

components/sync_manager/src/syncmanager.udl
```diff
@@ -48,12 +48,15 @@ interface SyncEngineSelection {
...
 
-enum SyncReason {
+enum RustSyncReason {
     "Scheduled",
     "User",
     "PreSleep",
     "Startup",
     "EnabledChange",
+    "Backgrounded",
+    "DidLogin",
```
mhammond, line 58: Android treats `DidLogin` and `EnabledChange` as the same - https://github.com/mozilla-mobile/firefox-android/blob/5a961111addca08dcaa402d2a79b05920c04e731/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/Types.kt#L42. I wonder if we should just do the same here? Otherwise, we should probably get a PR/issue up for Android to use this new one (and given the above link, that seems like it would be a trivial change for them)

components/sync15/ios/RustSyncTelemetryPing.swift
```diff
@@ -0,0 +1,622 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
```
mhammond, line 2: sheesh, it really sucks that we still need this hackery (but yeah, I agree we do - it just sucks :)

mhammond (commented):
looking good!

components/sync15/ios/RustSyncTelemetryPing.swift
lougeniaC64: Yeah, sadly--particularly if we want to create more/better graphs with this telemetry data. I also think I might need to add some unit tests here (although there aren't any for the android logic this file was based on).

lougeniaC64:
> looking good!

Thanks! One thing I'll note is that I am holding off on addressing the breaking changes created for android by the UDL renaming of `DeviceType`, `SyncReason`, and `SyncResult` until the local testing of syncing with the component is finished. Adding those hacks will probably be one of the last things I do before officially opening this PR up for review so that I can reduce the amount of churn here.

components/sync_manager/src/syncmanager.udl
lougeniaC64: Yeah actually these enum variants are a relic from my first pass at the sync manager work. `ClientNameChanged` is a variant in the native `SyncReason` enum but it's only used by the swift client synchronizer code, which will ultimately be removed. I think I'm going to remove it here and just map it to `EnabledChange` where needed.

components/sync_manager/src/syncmanager.udl
lougeniaC64: Also `DidLogin` is also mainly used by the old sync code so I think I'm going to remove it to and map `SyncReason.DidLogin` to `RustSyncReason.EnabledChange`.

components/sync15/ios/RustSyncTelemetryPing.swift
```diff
@@ -0,0 +1,622 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+public class RustSyncTelemetryPing {
```
tarikeshaq, line 7: Swift might have nice support for deserializable types, what do you think about using something like [`Codable`](https://developer.apple.com/documentation/swift/codable) and [`JSONDecoder`](https://developer.apple.com/documentation/foundation/jsondecoder)?

tarikeshaq: The more I look at it the more I see custom logic that `Codable` won't give us by default... We can implement the `init(from decoder: Decoder)` but I'm not sure that's worth it..

components/sync15/ios/RustSyncTelemetryPing.swift
```diff
@@ -0,0 +1,622 @@
...
+    public var version: Int
```
tarikeshaq, line 8: nit should those be `let` and not `var`?

components/sync15/ios/RustSyncTelemetryPing.swift
```diff
@@ -0,0 +1,622 @@
...
+    public static func fromJSONString(jsonObjectText: String) throws -> RustSyncTelemetryPing {
+        guard let data = jsonObjectText.data(using: .utf8) else {
+            throw TelemetryJSONError.invalidJSONString
+        }
+
+        let jsonObject = try JSONSerialization.jsonObject(with: data) as? [String: Any] ?? [String: Any]()
+        return try fromJSON(jsonObject: jsonObject)
+    }
+
+    func toJSON() throws -> [String: Any] {
```
tarikeshaq, line 60: hmm do we need all the `toJSON` functions? do we expect this type to also be serializable?

components/sync_manager/src/syncmanager.udl
```diff
@@ -54,6 +54,7 @@ enum SyncReason {
     "PreSleep",
     "Startup",
     "EnabledChange",
+    "Backgrounded",
```
tarikeshaq, line 57: Could this be a breaking change in android? in the case where android depends on an exhaustive `when` expression that checks all the variants(I'm not sure if does or doesn't, just making sure we mark it as such if it is 😬)

megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/RustSyncTelemetryPingTests.swift
```diff
@@ -0,0 +1,51 @@
...
+@testable import MozillaTestServices
+
+import XCTest
+
+class RustSyncTelemetryPingTests: XCTestCase {
+    let json = """
+    {"version":1,"uid":"b01a6f3d11cbb0f2ae2c940ab458b30a","syncs":[ {"when":1676416271.0,"took":837,"engines":[{"name":"passwords","when":167641627.0,"took":116,"incoming":{"applied":1},"outgoing":[{}]}]}]}
+    """
+
+    func test() {
```
tarikeshaq, line 14: do we need more tests for this, or do we think it's good enough to test the happy path?

tarikeshaq (commented):
This is looking great!! 🚀 


Dropped a couple of thoughts as I was skimming through but looks good!

components/sync_manager/src/syncmanager.udl
lougeniaC64: When I built android components locally against these changes while creating the `SyncManagerDeviceType` change, no `SyncReason` errors arose. It's worth noting that android defines it's own sync reasons and converts them to our sync reasons as needed (see [here](https://github.com/mozilla-mobile/firefox-android/blob/9de5ab8cc098674aad5190ce8b00ae6be65e9bd0/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/Types.kt#L44).

lougeniaC64: Also please feel free to build a-c locally and check my math!

megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/RustSyncTelemetryPingTests.swift
lougeniaC64: This absolutely does not cover all cases, and because we ultimately plan to remove this logic I didn't want to do too much here. I was most worried about the cases where a missing property is converted to default values (e.g. `EngineInfo.validation` or `OutgoingInfo`) but I can add cases for when errors are thrown, or if there are other cases people feel strongly about.

components/sync15/ios/RustSyncTelemetryPing.swift
lougeniaC64: Yeah this and the `let` comment are fair points. I duplicated this code from `SyncTelemetryPing.swift` knowing that I needed most of it and adding the rest while I was there "just in case". But since we ultimately want to get rid of this serialization/deserialization logic and handle telemetry within a-s, I think stripping this file down to the bare essentials and treating it as a carrier of deserialized sync telemetry data makes a lot of sense.

components/sync15/ios/RustSyncTelemetryPing.swift
lougeniaC64: See my comment below 😬

components/sync_manager/src/syncmanager.udl
tarikeshaq: No makes sense! Thank you for checking 🙏

tarikeshaq (approved):
Sorry took so long to get back, LGTM! 🚀

mhammond (approved):
🚀

Issue #5106, opened 2022-08-25
Remove logging related to #5040


bendk:
To track down #5040, we've added a bunch of logging to places.  Once that error is figured out, we should remove the logging.
- #5084 
- #5105
- Some of #4979

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3294)
┆Sprint End Date: 2023-02-24

PR #5364, opened 2023-02-06
Remove SqlErrorNoRowsReturned related logging (gh-5106)

bendk:
We've investigated this, gotten info from the logging, and decided not to pursue it further.  Let's remove the logging now before it gets more stale.

  - Note:

### DIFF
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -19,7 +19,6 @@ use crate::storage::{
 };
 use crate::types::{SyncStatus, VisitTransition, VisitTransitionSet};
 use actions::*;
-use error_support::breadcrumb;
 use rusqlite::types::ToSql;
 use rusqlite::Result as RusqliteResult;
 use rusqlite::Row;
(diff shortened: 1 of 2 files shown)


components/places/src/storage/history.rs
```diff
@@ -42,12 +41,10 @@ static DELETION_HIGH_WATER_MARK_META_KEY: &str = "history_deleted_hwm";
 
 /// Returns the RowId of a new visit in moz_historyvisits, or None if no new visit was added.
 pub fn apply_observation(db: &PlacesDb, visit_ob: VisitObservation) -> Result<Option<RowId>> {
-    breadcrumb!("apply_observation: begin_transaction");
```
tarikeshaq, line 45: Could this be useful in the future? I'm not sure, but would it be possible breadcrumbing here can help us track other issue?

tarikeshaq (approved):
Rubber stamp and a small question, but makes sense to me

components/places/src/storage/history.rs
bendk: Maybe useful, but also maybe noise.  I'm not sure what the best approach is to the breadcrumbs, hopefully we can figure it out as we go along.

PR #5369, opened 2023-02-09
Create shared remote settings client and port Nimbus to it.

MatthewTighe:
This PR creates a new shared support component called rs_client that can perform simple fetch requests for collections in Remote Settings. It also ports the current Nimbus implementation to use it. This is detailed in [ADR-0005](https://github.com/mozilla/application-services/blob/main/docs/adr/0005-remote-settings-client.md).

I decided that this would make for a good first PR as it felt like a smallish, atomic piece of work and would give the Nimbus team the opportunity for early feedback on any core changes they require before the new client is used by the upcoming Remote Settings client.

For any reviewer that's unaware - I am brand new to Rust and to contributing to app-services, so would gratefully welcome any and all feedback in terms of normal conventions, process, or anything else. Thanks!

- [x] **Quality**:/application-services/blob/main/docs/howtos/cut-a-new-release.md) after merging.
- [x] **Tests**:
- [x] **Changelog**: This PR does not include and consumer-facing changes.
- [x] **Dependencies**:


components/support/rs_client/src/tests/test_http_client.rs
```diff
@@ -0,0 +1,119 @@
...
+fn test_defaults() {
+    let config = ClientConfig {
+        server_url: None,
+        bucket_name: None,
+        collection_name: String::from("the-collection"),
+    };
+    let client = Client::new(config).unwrap();
+    assert_eq!(Url::parse("https://firefox.settings.services.mozilla.com").unwrap(), client.base_url);
+    assert_eq!(String::from("main"), client.bucket_name);
+}
```
MatthewTighe, line 21: It doesn't seem like these tests are run using `cargo test` from repo root. Would this component need to included in the top-level `Cargo.toml` for that to work?

components/nimbus/src/error.rs
```diff
@@ -63,6 +64,8 @@ pub enum NimbusError {
     ParseIntError(#[from] ParseIntError),
     #[error("Transform parameter error: {0}")]
     TransformParameterError(String),
+    #[error("Error with HTTP client: {0}")]
+    ClientError(#[from] rs_client::ClientError),
```
MatthewTighe, line 68: I'm curious whether there's any desire for this mapping to be more granular, exposing more detailed errors from the shared component, and what the best way to do that would be.

MatthewTighe:
It doesn't look like I'm able to request reviewers on the PR, so I'll ping @mhammond and @jhugman directly and beg forgiveness if there's a better way for me to be doing that 🙂

MatthewTighe:
Running `cargo run --bin protobuf-gen tools/protobuf_files.toml` doesn't produce or change any artifacts so I would appreciate any advice on what's going wrong there, as well.

components/nimbus/src/client/mod.rs
```diff
@@ -33,7 +34,7 @@ pub(crate) fn create_client(
                 };
                 Box::new(FileSystemClient::new(path)?)
             } else {
-                Box::new(Client::new(config)?)
+                Box::new(Client::new(config.to_client_config())?)
```
mhammond, line 37: Because `config` is passed by value into this function, it would be better to "consume" it - so `to_client_config()` could take `self` instead of `&self`, and then you wouldn't need to clone in its impl.

Even more idiomatic rust would be to use Into/From - so you could replace `to_client_config()` with

```rust
impl From<RemoteSettingsConfig> for ClientConfig {
    fn from(c: RemoteSettingsConfig) -> ClientConfig {
        ClientConfig {
            server_url: Some(c.server_url),
            bucket_name: None,
            collection_name: c.collection_name,
        }
    }
````
which then describes an automatic way of going from one struct to another. Then this line then changes to `Box::new(Client::new(config.into())?)`

File: components/nimbus/src/error.rs (hunk 2 above)
Comment (mhammond) [reply components/nimbus/src/error.rs:68]: It looks like this already does that - this enum variant could be unpacked as `NimbusError::ClientError(internal)` where `internal` will itself be a `rs_client::error::ClientError` enum - so code that cares could do something like:

```rust
match err {
  NimbusError::ClientError(rs_client::error::ClientError::RequestError(_)) -> println!(...),
  _ -> println!(...),
}
```
if it really wanted to know about just that one error, or could do:
```rust
match err {
  NimbusError::ClientError(_) -> println!(...),
  _ -> println!(...),
}
```
if it wanted to handle all `ClientError`s together.

However, that means that some of the other errors here are redundant - eg, it seems likely that `BackoffError` etc should be removed from this enum as they are now sub-variants of `ClientError`.

components/support/rs_client/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum ClientError {
+    #[error("Error sending request: {0}")]
+    RequestError(#[from] viaduct::Error),
+    #[error("Error parsing URL: {0}")]
+    UrlParsingError(#[from] url::ParseError),
+    #[error("Server asked the client to back off ({0} seconds remaining)")]
+    BackoffError(u64),
+    #[error("Error in network response: {0}")]
+    ResponseError(String),
```
mhammond, line 14: Is there a reason you don't have this be a `viaduct::Response` instead of a string? That way consumers of the error can inspect it for more details if they care - eg, for the status code.

components/support/rs_client/src/tests/test_http_client.rs
```diff
@@ -0,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mhammond, line 1: this is purely stylistic and a personal preference, but this is file and http_client.rs are both small enough that I'd be inclined to include these tests directly in http_client.rs. But feel free to keep it this way if you really do prefer it this way!

components/support/rs_client/src/http_client.rs
```diff
@@ -0,0 +1,110 @@
...
+pub(crate) enum RemoteState {
+    Ok,
+    Backoff {
+        observed_at: Instant,
+        duration: Duration,
+    },
+}
+
+impl Client {
+    #[allow(unused)]
```
mhammond, line 35: is this annotation still needed? It looks like nimbus now calls it.

components/support/rs_client/src/tests/test_http_client.rs
```diff
@@ -0,0 +1,122 @@
...
+    http_client.remote_state = Cell::new(RemoteState::Backoff {
+        observed_at: Instant::now() - Duration::from_secs(31),
+        duration: Duration::from_secs(30),
+    });
+    assert!(http_client.get().is_ok());
+    m.expect(1).assert();
+}
+
+fn response_body() -> String {
+    format!(
```
mhammond, line 111: clippy is complaining here - it just wants you to drop the `format!()` and instead just use `r#"..."#.to_string()` or similar.

mhammond (commented):
This is coming along nicely!

mhammond:
(and FWIW, the dependency failures are unrelated to your PR!)

components/support/rs_client/src/tests/test_http_client.rs
mhammond: ah yes, this should be in the "workspace" (ie, referenced in the top-level Cargo.toml)

components/support/rs_client/src/error.rs
MatthewTighe: I just copied it over from Nimbus tbh. I would guess they stringified it so that a consuming application could get details of the error which otherwise get lost since the `viaduct::Response` does not implement `Display`. 

Is there a good way to use `viaduct::Response` here and have the error use full display text from Nimbus and the top-level Remote Settings client? Otherwise I think it may be best to keep as is so the application layer can benefit fully.

components/support/rs_client/src/tests/test_http_client.rs
MatthewTighe: Putting tests in the same file is definitely new to me, but I'm here for it! This explains why I wasn't seeing test files in the support components.

components/support/rs_client/src/tests/test_http_client.rs
MatthewTighe: Ah I bet I actually did have it there in my other branch where I created the top-level Remote Settings stubs and then scrubbed it while I was separating this commit. Thanks!

components/nimbus/src/client/mod.rs
MatthewTighe: Oh, this is great stuff to know. Thanks!

MatthewTighe (commented):
Addressed most of Mark's feedback! I pushed changes as a separate commit, but in Fenix reviews we will often just overwrite the most recent commit. Let me know if there's a preference around these parts 🙂

components/support/rs_client/src/error.rs
mhammond: yeah, I later realized that this will be because it's exposed via uniffi, which a viaduct::Response can't do, so yeah, this is fine.

components/nimbus/src/tests/client/test_http_client.rs
```diff
@@ -2,14 +2,10 @@
...
-};
+use crate::client::SettingsClient;
+use crate::{Branch, BucketConfig, Experiment, FeatureConfig, RandomizationUnit, SCHEMA_VERSION};
 use mockito::mock;
-use std::cell::Cell;
-use std::time::{Duration, Instant};
+use rs_client::{Client, ClientConfig};
 
 #[test]
 fn test_fetch_experiments_from_schema() {
```
tarikeshaq, line 11: a possible followup: 

I paused here and I'm not sure I have a good answer to my own following question 😬

What **should** this test actually test? I feel a little weird that we are actually sending an HTTP request here now that the client itself is down in `rs-client`.

This should be a nice-to-have follow up but there might be an opportunity here to use trait mocking to mock the `SettingsClient`, see how [`FxAClient`](https://github.com/mozilla/application-services/blob/bd5ff1a29747702cbc95be998903be206c24506a/components/fxa-client/src/internal/http_client.rs#L55-L56) is mocked, and then [the tests depend on a generated type that mocks it](https://github.com/mozilla/application-services/blob/bd5ff1a29747702cbc95be998903be206c24506a/components/fxa-client/src/internal/profile.rs#L124-L141). 

That way we remove both the `viaduct_reqwest` and the mockito dependencies for nimbus! but all this is test code, so later follow up 😄

components/support/rs_client/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "rs_client"
+version = "0.1.0"
+edition = "2021"
+
```
tarikeshaq, line 5: we should add at least the `license`, `description` and `author` fields here: https://github.com/mozilla/application-services/blob/bd5ff1a29747702cbc95be998903be206c24506a/components/nimbus/Cargo.toml#L4-L9

components/support/rs_client/src/http_client.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        Ok(Self {
+            base_url: Url::parse(&server_url)?,
+            bucket_name,
+            collection_name: config.collection_name,
+            remote_state: Cell::new(RemoteState::Ok),
+        })
+    }
+
+    pub fn get(&self) -> Result<Response> {
```
tarikeshaq, line 49: how do you feel about the name `get`, maybe `get_records` is clearer? I'm not sure if it's easy to imply by consumers that `get` refers to the records but I don't have too much context on remote_settings as a whole 😅

components/nimbus/src/nimbus.udl
```diff
@@ -76,10 +76,10 @@ enum NimbusError {
     "InvalidPersistedData", "RkvError", "IOError",
     "JSONError", "EvaluationError", "InvalidExpression", "InvalidFraction",
     "TryFromSliceError", "EmptyRatiosError", "OutOfBoundsError","UrlParsingError",
-    "RequestError", "ResponseError", "UuidError", "InvalidExperimentFormat",
```
tarikeshaq, line 79: This **may** be a breaking change depending on if the consuming apps depend on those errors. Which is totally okay, but lets note that in our changelog! (or if we can easily verify this definitely isn't a breaking change, just noting tha apps can no longer depend on them is good enough!)

tarikeshaq (commented):
Thanks a ton @MatthewTighe (and sorry for the unsolicited review 😆) this looks great and I'm excited to see a remote-settings client starting off 🎉 🚀 

I'm adding @jeddai for review since this changes nimbus code and will leave final approval to @mhammond

components/support/rs_client/src/http_client.rs
tarikeshaq: Maybe I could add something here, how do you feel about adding doc comments to the `http_client` functions here? Since it's its own library now, it could be useful for other Rust consumers to see the docs for each method! ref: https://doc.rust-lang.org/rust-by-example/meta/doc.html

MatthewTighe: I'm not at all married to the name, and I agree that `get_records` is clearer. This was intended to match with the [existing platform client](https://remote-settings.readthedocs.io/en/latest/introduction.html#what-does-the-client-api-look-like), but if we even deem that to be important it could be done in the top-level Remote Settings client instead of here.

components/nimbus/src/nimbus.udl
MatthewTighe: This is a good point, I see that Fenix treats `RequestError` and `ResponseError` as non-reportable, so although a breaking change I don't think they are necessary there at least. Is the iOS app the only other place I should be checking?

components/nimbus/src/client/mod.rs
```diff
@@ -47,3 +48,13 @@ pub(crate) trait SettingsClient {
     fn get_experiments_metadata(&self) -> Result<String>;
     fn fetch_experiments(&self) -> Result<Vec<Experiment>>;
 }
+
+impl From<RemoteSettingsConfig> for ClientConfig {
```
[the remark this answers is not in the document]
MatthewTighe: You can [thank mark for that one](https://github.com/mozilla/application-services/pull/5369#discussion_r1100911153) 😄

components/nimbus/src/tests/client/test_http_client.rs
MatthewTighe: I do like this idea, but I would love some input from the Nimbus team on whether they find a more integration-y test useful to keep around before I make changes. Maybe @jeddai can weigh in?

tarikeshaq: Yeah for the sake of this PR being a "move the rs client down to its own crate" I'm sure we can land this as is, good follow up though to play around with trait mocking

components/nimbus/src/nimbus.udl
tarikeshaq: Yup! iOS and Fenix are the only consumers (right now!) - given Fenix depends on it, let's flag this change as a breaking change in the https://github.com/mozilla/application-services/blob/main/CHANGES_UNRELEASED.md, so when we do the next a-s upgrade we replace them with new variants

components/support/rs_client/Cargo.toml
MatthewTighe: The question of authorship is an interesting one. I don't want to volunteer the Nimbus team, but I'm also not sure signing the Android mobile team up for it makes a ton of sense either in case of lost Rust context. I can check in with my team, but I'd be interested in any suggestions here as well

tarikeshaq: We can add multiple so  Nimbus team  + Android team seems sensible - Nimbus wrote the pre-existing bits and you are moving down/evolving it!

We don't use the `Cargo.toml` as the absolute source of truth for ownership, but it's a good reference for who wrote the code!

jeddai:
Coming up to speed now, aiming to have a review in by eod

components/nimbus/src/tests/client/test_http_client.rs
jeddai: So my thought on this is that the structure of this test – calling `get` on the client and validating the type and value of the response – should move to the `rs_client` module. For this new test, the library making the http call and the function's response value are under test. 

As for this test, while it _can_ remain here, what I think this is actually testing now is the `fetch_experiments` and `parse_experiments` methods in `http_client`. I think this should become two different sets of tests:
1. A test on `fetch_experiments` to validate that it a) calls `self.get()`, and b) passes the `text` from that into `parse_experiments`. 
    * This could very likely be done with mocking, though I'll admit I've not had to do much mock testing in Rust yet so I'm not sure off the top of my head how to do that.
2. Tests covering the potential branches of `parse_experiments`. Admittedly this seems like tech debt, and I don't think it should be required for your PR, but the only cases I've found of `parse_experiments` being tested cover the `Ok` case.

jeddai (commented):
Overall I'm really liking these changes! I've got a couple comments, mostly on top of existing comments. Once stuff has been addressed feel free to re-request my review and I'll approve!

components/nimbus/src/tests/client/test_http_client.rs
MatthewTighe: @tarikeshaq After looking into this, I have some more questions about what we'd like to do here. I'm happy to address this in a follow-up or now in case it has any impact on the API of `rs_client`. Also happy to include an additional Nimbus test to capture a failure branch for `parse_experiments`. The more Rust practice the better 🙂 

I'm not sure that using trait mocking similar to `FxAClient` as mentioned will accomplish the goals stated in @jeddai's post directly above this. That kind of mocking seems to be useful when you have a trait that is being used by a separate type under test, but it seems in both of the cases we really want to test a specific implementation of the trait itself.  

For example:
```
#[cfg_attr(test, mockiato::mockable)]
pub(crate) trait SettingsClient {
    fn get_experiments_metadata(&self) -> Result<String>;
    fn fetch_experiments(&self) -> Result<Vec<Experiment>>;
}

#[test]
fn test_fetch_experiments_from_schema() {
    let client = SettingsClientMock::new()
    client
        .expect_fetch_experiments(...)
        .returns_once(...);
    let resp = client.fetch_experiments().unwrap();
    assert(resp...);
}
```
seems to only really be testing the mocking framework itself.  

I'm curious if, instead, there is a way to mock the type the trait is implemented _for_, in this case `rs_client::Client`. Is it possible to expose that base type as mockable from the support component? Or does this indicate that we might want to change the underlying design, like changing `Client` to be a trait?

For example, I think we want something closer to:
```
fn test_parse_fetched_experiments() {
    let client = ClientMock::new()
    client
        .expect_get_records()
        .returns_once(<response>);

    let resp = client.fetch_experiments().unwrap();
    assert(<parsed response>);
}

fn test_fail_parse_fetched_experiments() {
    ...
}
```

The other option would be to move forward with mocking `SettingsClient` in a higher-level test, but then I think we're missing out on covering the cases above.

MatthewTighe:
Addressed most of the rest of the feedback, just want to make sure we have a clear resolution to our [conversation about testing](https://github.com/mozilla/application-services/pull/5369#discussion_r1103084587).

Thanks all for the patience and thoroughness during this review cycle!

components/nimbus/src/tests/client/test_http_client.rs
tarikeshaq: Thanks a ton for looking deeply into this @MatthewTighe 🚀 

> That kind of mocking seems to be useful when you have a trait that is being used by a separate type under test, but it seems in both of the cases we really want to test a specific implementation of the trait itself.

That is exactly right - if we are testing behaviour specific to the HTTP client, which from @jeddai 's responses is what we want here, then mocking the trait won't help us!


> I'm curious if, instead, there is a way to mock the type the trait is implemented for, in this case rs_client::Client

Unfortunately Rust doesn't let us consume test-only types in other crates :( So even if we had defined `rs_client::Client` as a trait first, its mock won't be consumable here in Nimbus.


**Let's open a follow-up ticket for this case**. I think it's a small enough detail that we can land with the test as-is

(the only win in making this better is getting rid of a test-only dependency!)


FWIW I believe a way forward here could include:
- Exposing a `ClientTrait` trait from `rs_client`, and having `rs_client::Client` implement it
- The Nimbus Http Client own either a `Box<dyn ClientTrait>` or be generic on a type that implements `ClientTrait`
- In the test code, we can create our own TestClient type that implements `ClientTrait` and returns whatever we want it to return, then we can validate the behavior of the nimbus client.


That all adds complexity though and I'm not exactly sure is the best way forward, so a follow-up is probably best!

tarikeshaq (approved):
This looks amazing 🚀 🚀 +1 from me

components/nimbus/src/tests/client/test_http_client.rs
jeddai: I agree on the follow-up ticket. I'm going to do a quick scan through the rest of the changes and will approve after that!

CHANGES_UNRELEASED.md
```diff
@@ -20,8 +20,9 @@ Use the template below to make assigning a version number during the release cut
...
 ## Nimbus ⛅️🔬🔭
-
 ### 🦊 What's Changed 🦊
 - Updated the Nimbus Gradle Plugin to fix a number of issues after migrating it to this repository ([#5348](https://github.com/mozilla/application-services/pull/5348))
 - Good fences: protected calls out to the error reporter with a `try`/`catch` ([#5366](https://github.com/mozilla/application-services/pull/5366))
-- Updated the Nimbus FML CLI to only import the R class if it will be used by a feature property ([#5361](https://github.com/mozilla/application-services/pull/5361))
\ No newline at end of file
+- Updated the Nimbus FML CLI to only import the R class if it will be used by a feature property ([#5361](https://github.com/mozilla/application-services/pull/5361))
+### ⚠️ Breaking Changes ⚠️
+  - Android and iOS: Several errors have been moved to an internal support library and will no longer be reported as top-level Nimbus errors. They should still be accessible through `NimbusError.ClientError`. They are: `RequestError`, `ResponseError`, and `BackoffError`.
```
jeddai, line 28: Can you add a link to this PR here like the changelogs above?

jeddai (approved):
r+wc -- this looks great!

mhammond (approved):
Thanks @tarikeshaq and @jeddai for the detailed reviews, adding my rubber stamp

components/nimbus/src/tests/client/test_http_client.rs
MatthewTighe: Just to close the loop, filed https://github.com/mozilla/application-services/issues/5394 as a follow-up and will be working on it shortly.

PR #5374, opened 2023-02-09
Move the EventStore into the TargetingHelper

jhugman:
This started off as a fix for [EXP-2943](https://mozilla-hub.atlassian.net/browse/EXP-2943).

This is not ready for that, it is a pre-requisite for that work: however, it is also very useful for the Cirrus work of separating out the event store from the enrollment evolution path.

This work is to unify all calls into the `EnrollmentEvolver` so that we're not passing an `EventStore` around, but a `NimbusTargetingHelper`.

Thus: bringing this out of Draft, so it may be landed and become useful to the Cirrus work.

  - Note:

### DIFF
--- a/components/nimbus/src/enrollment.rs
+++ b/components/nimbus/src/enrollment.rs
@@ -1,17 +1,17 @@
-use crate::behavior::EventStore;
-use crate::defaults::Defaults;
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
+use crate::defaults::Defaults;
 use crate::error::{NimbusError, Result};
-use crate::evaluator::TargetingAttributes;
 use crate::persistence::{Database, StoreId, Writer};
 use crate::{evaluator::evaluate_enrollment, persistence::Readable};
-use crate::{AvailableRandomizationUnits, EnrolledExperiment, Experiment, FeatureConfig};
+use crate::{
+    AvailableRandomizationUnits, EnrolledExperiment, Experiment, FeatureConfig,
+    NimbusTargetingHelper,
+};
 
 use ::uuid::Uuid;
 use serde_derive::*;
-use std::sync::{Arc, Mutex};
 use std::{
     collections::{HashMap, HashSet},
     time::{Duration, SystemTime, UNIX_EPOCH},
@@ -81,9 +81,8 @@ impl ExperimentEnrollment {
         is_user_participating: bool,
         nimbus_id: &Uuid,
         available_randomization_units: &AvailableRandomizationUnits,
-        targeting_attributes: &TargetingAttributes,
         experiment: &Experiment,
-        event_store: Arc<Mutex<EventStore>>,
+        targeting_helper: &NimbusTargetingHelper,
         out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
     ) -> Result<Self> {
         Ok(if !is_user_participating {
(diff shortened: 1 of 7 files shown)


components/nimbus/src/enrollment.rs
```diff
@@ -900,18 +886,18 @@ impl<'a> EnrollmentsEvolver<'a> {
             false
         };
 
-        let mut targeting_attributes = self.targeting_attributes.clone();
-        targeting_attributes.is_already_enrolled = is_already_enrolled;
+        let th = self
+            .targeting_helper
+            .put("is_already_enrolled", is_already_enrolled);
```
jhugman, line 891: This step is a bit icky, and should be removed, but we need to change `experimenter` sticky logic to use the `active_experiments` form of sticky targeting.

components/nimbus/src/lib.rs
```diff
@@ -888,16 +892,32 @@ pub struct NimbusTargetingHelper {
...
+        } else {
+            self.context.clone()
+        };
+
+        let event_store = self.event_store.clone();
+        Self {
+            context,
+            event_store,
+        }
+    }
```
jhugman, line 920: This goes away when we `experimenter` writes sticky targeting in terms of `active_experiments`.

components/nimbus/src/tests/test_evaluator.rs
```diff
@@ -23,12 +23,8 @@ fn test_locale_substring() -> Result<()> {
...
         ..Default::default()
     };
-    let targeting_attributes = ctx.into();
-    let event_store = Arc::new(Mutex::new(EventStore::new()));
-    assert_eq!(
-        targeting(expression_statement, &targeting_attributes, event_store),
-        None
-    );
+    let th = ctx.into();
+    assert_eq!(targeting(expression_statement, &th), None);
```
jeddai, line 27: Okay so if I'm understanding this correctly, the `AppContext` type is able to `.into()` on `TargetingHelper` types because `new` on the `TargetingHelper` has the generic `<C: Serialize>`, and `AppContext` is serializable.

One thing I'm not tracking here is how it knows how to instantiate the EventStore when used in this way? Does it effectively use `Default::default()` for `Arc<Mutex<EventStore>>`?

jeddai (approved):
r+wc but the comment I'd like a response on is just a question. Overall this looks great and I think it ultimately is a great solution to this problem!

Only downside I see is working with the AppContext on the targeting helper is a little less safe than before (unless I've misunderstood something).

components/nimbus/src/enrollment.rs
jhugman: Yes, agreed with caveats: we lose the type-safety within the evolver: so I'm leaning towards getting rid of mutability of the targeting helper (mutability/clone with changes that we have here) within the evolver. In practice, I think this means changing just this sticky flag.

components/nimbus/src/lib.rs
jhugman: Created issue https://github.com/mozilla/experimenter/issues/8661.

components/nimbus/src/enrollment.rs
jhugman: Created https://github.com/mozilla/experimenter/issues/8661

components/nimbus/src/tests/test_evaluator.rs
jhugman: There are [several `impl`s of `From` and `Default` used in testing in `test_enrollment.rs`, here](https://github.com/mozilla/application-services/blob/32de3d621b7f0662d0b5da5bbb700e1c251c35e7/components/nimbus/src/tests/test_enrollment.rs#L1199-L1226).

PR #5375, opened 2023-02-10
#[handle_error] macro now requires the internal error type be specified

mhammond:
The fact that `#[handle_error]` assumed a `Result<>` was in scope and defined the internal error type just bit me as I was working on something else, and I saw it was described as a but of a smell, so I thought I'd look at fixing it.

This now requires every use of that macro to specify the internal error like `#[handle_error(E="path::to::Error")`. I've only converted tabs and places so far, but thought I'd put it up now for general comment and bikeshedding on the `E="..."` arg I came up with before converting the others.

### DIFF
--- a/components/support/error/macros/src/argument.rs
+++ b/components/support/error/macros/src/argument.rs
@@ -2,16 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-pub(crate) fn validate(arguments: &syn::AttributeArgs) -> syn::Result<()> {
-    // For now we validate that there are no arguments.
-    // if we want to improve the macro to support attributes, for example,
-    // if we'd like to pass in the return type of the body we can modify this
-    // logic to parse and return the type.
-    if !arguments.is_empty() {
-        return Err(syn::Error::new(
-            proc_macro2::Span::call_site(),
-            "Expected #[handle_error] with no arguments",
-        ));
+use syn::spanned::Spanned;
+
+const ERR_MSG: &str = "Expected #[handle_error(path::to::Error)]";
+
+/// Returns the path to the type of the "internal" error.
+pub(crate) fn validate(arguments: &syn::AttributeArgs) -> syn::Result<&syn::Path> {
+    if arguments.len() != 1 {
+        return Err(syn::Error::new(proc_macro2::Span::call_site(), ERR_MSG));
     }
-    Ok(())
+
+    let nested_meta = arguments.iter().next().unwrap();
+    if let syn::NestedMeta::Meta(syn::Meta::Path(path)) = nested_meta {
+        Ok(path)
+    } else {
+        Err(syn::Error::new(nested_meta.span(), ERR_MSG))
+    }
 }
--- a/components/support/error/macros/src/lib.rs
+++ b/components/support/error/macros/src/lib.rs
@@ -48,9 +48,11 @@ mod argument;
 ///    }
 /// }
 ///
+/// // The `handle_error` macro maps from the error supplied in the mandatory argument
+/// // (ie, `Error` in this example) to the error returned by the function (`ExternalError`
+/// // in this example)
 /// #[handle_error(Error)]
 /// fn do_something() -> std::result::Result<String, ExternalError> {
-///    // The `handle_error` macro maps from `Error` to `ExternalError`
 ///    Err(Error{})
 /// }
 ///
--- a/components/support/error/macros/src/lib.rs
+++ b/components/support/error/macros/src/lib.rs
@@ -48,7 +48,7 @@ mod argument;
 ///    }
 /// }
 ///
-/// #[handle_error]
+/// #[handle_error(Error)]
 /// fn do_something() -> std::result::Result<String, ExternalError> {
 ///    // The `handle_error` macro maps from `Error` to `ExternalError`
 ///    Err(Error{})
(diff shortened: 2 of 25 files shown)


components/support/error/macros/src/argument.rs
```diff
@@ -2,16 +2,33 @@
...
-    // if we'd like to pass in the return type of the body we can modify this
-    // logic to parse and return the type.
-    if !arguments.is_empty() {
-        return Err(syn::Error::new(
+/// Returns the name of the type of the "internal" error.
+pub(crate) fn validate(arguments: &syn::AttributeArgs) -> syn::Result<String> {
+    // We insist on a single arg of form `E=ErrorType`. If we end up growing
+    // more args we'd probably just add a struct and return that, but this seems
+    // ok for now.
+    fn oops(inner: &str) -> syn::Error {
```
tarikeshaq, line 10: a bit of a nit: We can have this function take a [`Span`](https://docs.rs/proc-macro2/latest/proc_macro2/struct.Span.html) and pass it to the error creation, which will make the errors a bit more localized!

(if we import the trait `Spanned` from `syn` we can get a `Span` from all the `syn` types!

ref: https://docs.rs/syn/latest/syn/spanned/trait.Spanned.html

components/support/error/macros/src/argument.rs
```diff
@@ -2,16 +2,33 @@
...
+        )
+    }
+    if arguments.len() != 1 {
+        return Err(oops(".."));
+    }
+
+    let nested_meta = arguments.iter().next().unwrap();
+    match nested_meta {
+        syn::NestedMeta::Meta(meta) => match meta {
+            syn::Meta::NameValue(meta_name_value) => match meta_name_value.path.get_ident() {
```
tarikeshaq, line 23: What if we use a `Meta::Path` instead of a `NameValue`

We'd lose the `E = ` but we can have the macro take the type itself.

so `#[handler_error(Error)]` would be the way to use it. I don't feel strongly, but it's a little nicer not having to put the type in a string literal :(

tarikeshaq: if we do ^ we'd have to return the path or a `TokenStream` from the `validate` function, which seems OK?

components/support/error/macros/src/lib.rs
```diff
@@ -60,30 +60,30 @@ mod argument;
 #[proc_macro_attribute]
 pub fn handle_error(args: TokenStream, input: TokenStream) -> TokenStream {
     let args = syn::parse_macro_input!(args as syn::AttributeArgs);
-    let parsed = syn::parse_macro_input!(input as syn::Item);
-    TokenStream::from(match impl_handle_error(&parsed, &args) {
-        Ok(res) => res,
-        Err(e) => e.to_compile_error(),
-    })
+    match argument::validate(&args) {
```
tarikeshaq, line 63: if we keep the `validate` in the `impl_handle_error` function, then we can get rid of this match block!

any error would be caught by the inner `match` and converted to a compile error

tarikeshaq (commented):
Wrote a brief comment about what I think re: the argument

I have a slight preference for `#[handler_error(Type)]` vs `#[handle_error(E = "Type")]` but I see the value of the latter if we start introducing more named arguments, thoughts?

mhammond:
> I have a slight preference for `#[handler_error(Type)]` vs `#[handle_error(E = "Type")]` but I see the value of the latter if we start introducing more named arguments, thoughts?

I thought I saw a comment implying that later we might want a `T` as well, but now I can't find it! I also can't see why that would every actually be necessary. I do think it makes sense to think about the possibility of other args later though - eg, I could imagine, say, `report_error = false` or similar as a way to get finer-grained per-function control. I don't *think* your suggestion would preclude that, right?

tarikeshaq:
> I don't think your suggestion would preclude that, right?

Right! I don't know why I assumed we can't have both in the future. i.e the macro can evolve into something like `#[handler_error(Error, report_error = false)` and that should be okay! - we'd just use [`Meta::List`](https://docs.rs/syn/latest/syn/enum.Meta.html#variant.List)  and have each argument parsed separately

components/support/error/macros/src/lib.rs
mhammond: I got a bit lazy there - I moved that around in a failed attempt to reuse this in another macro, but didn't change it back.

components/support/error/macros/src/lib.rs
```diff
@@ -48,7 +48,7 @@ mod argument;
 ///    }
 /// }
 ///
-/// #[handle_error]
+/// #[handle_error(Error)]
```
tarikeshaq, line 51: nit: Should we add a note in the doc comment to what `Error` is? maybe even a small comment above the macro invocation in the example that `Error` is the internal error and that it's mandatory to specify

components/support/error/tests/macro_arguments.rs
```diff
@@ -34,10 +34,35 @@
      }
  }
 
- #[handle_error(String)]
- fn func() -> ::std::result::Result<String, ExternalError> {
-     Err(Error{})
- }
+#[handle_error(Error)] // Works.
```
tarikeshaq, line 37: nit: Can we add another failing case of `#[handle_error(String)]` or some other type that isn't the correct one?

components/support/error/tests/parse.rs
```diff
@@ -35,32 +32,36 @@ impl GetErrorHandling for Error {
...
 
-#[handle_error]
+#[handle_error(Error)]
 fn func4() -> FullyAliasedResult {
     Err(Error{})
 }
 
+mod submodule {
+    #[super::handle_error(super::Error)]
+    fn func() -> ::std::result::Result<String, super::ExternalError> {
```
tarikeshaq, line 61: nit: can we call this function in the `main` to test that the macro still returns what we expect it to, similar to to other functions?

bendk (approved):
Looks good to me and I like making the error type more explicit.

tarikeshaq (approved):
Awesome!!! I added a couple of nits as I was going through, but I love this

PR #5376, opened 2023-02-10
Trim the title of tabs and length of list before trying to upload

### LINKED ISSUE #4986: Rust Tabs engine should trim tabs list before upload to keep payload under the max size
Desktop tabs does a couple of things we should:

* [Checks each URL is under the max size for a URL](https://searchfox.org/mozilla-central/rev/9dc5ffe42635b602d4ddfc9a4b8ea0befc94975a/services/sync/modules/engines/tabs.js#208-211)
* [Checks that the total payload size is smaller than the max, and drops tabs until it is](https://searchfox.org/mozilla-central/source/services/sync/modules/util.js#431)

The second in particular will stop *any* tabs for a device being listed, and is what I suspect is happening in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1772521)

We might as well do this in the rust component before we vendor it into m-c. cc @skhamis

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3197)
┆Epic: Prioritized from backlog
┆Sprint End Date: 2023-02-24

skhamis:
Fixes #4986 

1. Sets the maximum upload amount to 1MB
2. Trims any tab titles above 512 characters (and adds ellipsis)
3. Sorts the tab list by last_used

  - Note:

### DIFF
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -183,9 +183,8 @@ impl TabsStorage {
             // Sort the tabs so when we trim tabs it's the oldest tabs
             sanitized_tabs.sort_by(|a, b| b.last_used.cmp(&a.last_used));
             // If trimming the tab length failed for some reason, just return the untrimmed tabs
-            return Some(
-                trim_tabs_length(&mut sanitized_tabs, MAX_PAYLOAD_SIZE).unwrap_or(sanitized_tabs),
-            );
+            trim_tabs_length(&mut sanitized_tabs, MAX_PAYLOAD_SIZE);
+            return Some(sanitized_tabs);
         }
         None
     }
@@ -379,29 +378,25 @@ impl TabsStorage {
 }
 
 // Trim the amount of tabs in a list to fit the specified memory size
-fn trim_tabs_length(
-    tabs: &mut Vec<RemoteTab>,
-    payload_size_max_bytes: usize,
-) -> Result<Vec<RemoteTab>> {
+fn trim_tabs_length(tabs: &mut Vec<RemoteTab>, payload_size_max_bytes: usize) {
     // Ported from https://searchfox.org/mozilla-central/rev/84fb1c4511312a0b9187f647d90059e3a6dd27f8/services/sync/modules/util.sys.mjs#422
     // See bug 535326 comment 8 for an explanation of the estimation
     let max_serialized_size = (payload_size_max_bytes / 4) * 3 - 1500;
-    let size = compute_serialized_size(tabs)?;
+    let size = compute_serialized_size(tabs);
     if size > max_serialized_size {
         // Estimate a little more than the direct fraction to maximize packing
         let cutoff = (tabs.len() * max_serialized_size) / size;
         tabs.truncate(cutoff);
 
         // Keep dropping off the last entry until the data fits.
-        while compute_serialized_size(tabs)? > max_serialized_size {
+        while compute_serialized_size(tabs) > max_serialized_size {
             tabs.pop();
         }
     }
-    Ok(tabs.to_vec())
 }
 
-fn compute_serialized_size(v: &Vec<RemoteTab>) -> Result<usize> {
-    Ok(serde_json::to_string(v)?.len())
+fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
+    serde_json::to_string(v).unwrap_or_default().len()
 }
 
 // Try to keep in sync with https://searchfox.org/mozilla-central/rev/2ad13433da20a0749e1e9a10ec0ab49b987c2c8e/modules/libpref/init/all.js#3927
@@ -586,14 +581,14 @@ mod tests {
                 last_used: 0,
             });
         }
-        let tabs_mem_size = compute_serialized_size(&too_many_tabs).unwrap();
+        let tabs_mem_size = compute_serialized_size(&too_many_tabs);
         // ensure we are definitely over the payload limit
         assert!(tabs_mem_size > MAX_PAYLOAD_SIZE);
         // Add our over-the-limit tabs to the local state
         storage.update_local_state(too_many_tabs.clone());
         // prepare_local_tabs_for_upload did the trimming we needed to get under payload size
         let tabs_to_upload = &storage.prepare_local_tabs_for_upload().unwrap();
-        assert!(compute_serialized_size(tabs_to_upload).unwrap() <= MAX_PAYLOAD_SIZE);
+        assert!(compute_serialized_size(tabs_to_upload) <= MAX_PAYLOAD_SIZE);
     }
     // Helper struct to model what's stored in the DB
     struct TabsSQLRecord {
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -20,6 +20,7 @@ use sql_support::open_database::{self, open_database_with_flags};
 use sql_support::ConnExt;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::mem;
 use std::path::{Path, PathBuf};
 use sync15::{RemoteClient, ServerTimestamp};
 pub type TabsDeviceType = crate::DeviceType;
@@ -27,6 +28,8 @@ pub type RemoteTabRecord = RemoteTab;
 
 pub(crate) const TABS_CLIENT_TTL: u32 = 15_552_000; // 180 days, same as CLIENTS_TTL
 const FAR_FUTURE: i64 = 4_102_405_200_000; // 2100/01/01
+const MAX_PAYLOAD_SIZE: usize = 256 * 1024; // Same as Desktop, see service.js or sync15/request.rs
+const MAX_TITLE_CHAR_LENGTH: usize = 512; // A high limit, but also sane
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct RemoteTab {
@@ -146,30 +149,34 @@ impl TabsStorage {
         self.local_tabs.borrow_mut().replace(local_state);
     }
 
+    // We try our best to fit as many tabs in a payload as possible, this includes
+    // limiting the url history entries, title character count and finally drop enough tabs
+    // until we have small enough payload that the server will accept
     pub fn prepare_local_tabs_for_upload(&self) -> Option<Vec<RemoteTab>> {
         if let Some(local_tabs) = self.local_tabs.borrow().as_ref() {
-            return Some(
-                local_tabs
-                    .iter()
-                    .cloned()
-                    .filter_map(|mut tab| {
-                        if tab.url_history.is_empty() || !is_url_syncable(&tab.url_history[0]) {
-                            return None;
+            let mut sanitized_tabs: Vec<RemoteTab> = local_tabs
+                .iter()
+                .cloned()
+                .filter_map(|mut tab| {
+                    if tab.url_history.is_empty() || !is_url_syncable(&tab.url_history[0]) {
+                        return None;
+                    }
+                    let mut sanitized_history = Vec::with_capacity(TAB_ENTRIES_LIMIT);
+                    for url in tab.url_history {
+                        if sanitized_history.len() == TAB_ENTRIES_LIMIT {
+                            break;
                         }
-                        let mut sanitized_history = Vec::with_capacity(TAB_ENTRIES_LIMIT);
-                        for url in tab.url_history {
-                            if sanitized_history.len() == TAB_ENTRIES_LIMIT {
-                                break;
-                            }
-                            if is_url_syncable(&url) {
-                                sanitized_history.push(url);
-                            }
+                        if is_url_syncable(&url) {
+                            sanitized_history.push(url);
                         }
-                        tab.url_history = sanitized_history;
-                        Some(tab)
-                    })
-                    .collect(),
-            );
+                    }
+
+                    tab.url_history = sanitized_history;
+                    tab.title.truncate(MAX_TITLE_CHAR_LENGTH);
+                    Some(tab)
+                })
+                .collect();
+            return trim_tabs_length(&mut sanitized_tabs, MAX_PAYLOAD_SIZE);
         }
         None
     }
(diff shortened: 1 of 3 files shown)


mhammond:
> * What should the character limit of tab titles we sync be? It ideally just a little bit higher than the max we display anywhere

I don't think any such max exists TBH. The idea is really just to trim insanely long, so I was thinking 1024 might be ok (but the case could be made for quite a bit less I'd expect). I guess we should probably append a unicode ellipsis character too?

> * I didn't put any offset on the upper limit of the tabs payload, I expect we want a little lower to fit the meta stuffs but haven't found a good number for that yet, @mhammond  thoughts?

I don't quite understand this question :)

tarikeshaq (commented):
Apologies for the unsolicited review, especially that you marked this as a draft 🙏 I thought this was cool and dropped a couple of thoughts. Please feel free to ignore them as this evolves

Excited to see the impact here (across all three platforms!!!)

skhamis:
> > * What should the character limit of tab titles we sync be? It ideally just a little bit higher than the max we display anywhere
> 
> I don't think any such max exists TBH. The idea is really just to trim insanely long, so I was thinking 1024 might be ok (but the case could be made for quite a bit less I'd expect). I guess we should probably append a unicode ellipsis character too?

Yeah no strong preference here, though I'd lean towards 512 or even 256 since that still offers quite a bit of reduction while keeping the amount of characters still in a high limit. Adding an ellipsis is interesting, because on one hand, are we ever really displaying it? Should that be up to the clients whether an ellipsis is shown? (though we are trimming it ourselves, so we're responsible for communicating that).


> > * I didn't put any offset on the upper limit of the tabs payload, I expect we want a little lower to fit the meta stuffs but haven't found a good number for that yet, @mhammond  thoughts?
> 
> I don't quite understand this question :)

Yeah I realized that was probably unclear (and also slightly wrong based on what I was doing anyways).

My original question was mainly about what overhead in addition to the tabs themselves we'd need to make sure we leave space so our trimming doesn't still also end up failing in the end. I briefly saw the comment on what desktop was doing: https://searchfox.org/mozilla-central/source/services/sync/modules/util.js#436. However I'm trimming based on the `max_payload_bytes` and not the `max_total_bytes` so I _think_ this is not really an issue.

components/tabs/src/storage.rs
```diff
@@ -362,6 +369,25 @@ impl TabsStorage {
...
+    // Calculate the difference needed to hit the memory payload
+    // then drop any over
+    let difference = tabs_mem_size - memory_limit;
+    let elements_to_remove = difference / mem::size_of::<RemoteTab>();
+    tabs.truncate(tabs.len() - elements_to_remove);
+    Some(tabs.to_vec())
+}
+
+fn calc_memory_size<T>(v: &Vec<T>) -> usize {
+    v.capacity() * std::mem::size_of::<T>() + std::mem::size_of::<Vec<T>>()
```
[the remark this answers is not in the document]
skhamis: > 



> Hmm what does the server reject records on? I imagine it's the size of the encrypted payload json? Are we trying to do best-effort trimming at this layer (i.e we don't need to be accurate?)
> 
> I say that because [`std::mem::size_of`](https://doc.rust-lang.org/std/mem/fn.size_of.html) is the compile time size and:
> 
>     * adding the size of a `Vec` (which in practice is a raw pointer and two `usize`s)
> 
>     * the `std::mem::sizeof::<T>()` is probably not an accurate size, since that is the compile time size. i.e `RemoteTab` is
>       ```rust
>       pub struct RemoteTab {
>         pub title: String,
>         pub url_history: Vec<String>,
>         pub icon: Option<String>,
>         pub last_used: i64, // In ms.
>       }
>       ```
>         
>       since most values there are heap-allocated we could underestimate the size of the actual tab once it's serialized
> 
>     * Using `capacity` instead of `len` may overestimate the size of the vector
> 
> 
> if that is OK, and we aren't aiming for full accuracy (which I think is very reasonable TBH) then it's alright, but we'd want some documentation to convince our future selves what the trimming logic is doing

Yes! As a first pass at this, I definitely wanted to implement something that was (1) easy to maintain, (2) relatively easy to read/understand (though needs more comments I agree), (3) comparatively fast.

re telemetry: I don't think we _know_ exactly how many records are failing to upload, since we don't keep track of that. However, that'll be something I'll also want to implement so we can see (1) implementing telemetry success/failure for individual records (2) then see tabs records failing as a result of "too large" dropping

mhammond:
FTR, desktop currently trims the payload at https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a1d6bf0b8fc0c7481d513f7cda91a885 and, in particular, https://searchfox.org/mozilla-central/rev/5ccb73c0217d1710b10d6e6e297cf3396d10ec23/services/sync/modules/util.js#427 which makes some interesting assumptions about the size of the encrypted version! That code has been in desktop for a few years and seems to work fine.

Note that code does not currently trim titles though, but that does seem orthogonal.

All encryption in the rust code happens at https://github.com/mozilla/application-services/blob/main/components/sync15/src/client/coll_update.rs#L14, called by

components/tabs/src/storage.rs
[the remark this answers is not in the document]
mhammond: Sorry for the quick comment before, but I think we really just need to reimplement that `tryFitItems` function (ie, it's all about how big the encrypted json representation is, and nothing about the raw memory size (even though they are obviously related)

And off the top of my head, we might as well trim *every* title, even if they would fit? ie, a user with 1 tab with a short URL but a 1.5MB title will fit, but we aren't doing anyone a favor by including all that)

components/tabs/src/storage.rs
```diff
@@ -146,29 +149,36 @@ impl TabsStorage {
...
+                            sanitized_history.push(url);
                         }
-                        tab.url_history = sanitized_history;
-                        Some(tab)
-                    })
-                    .collect(),
+                    }
+
+                    tab.url_history = sanitized_history;
+                    tab.title.truncate(MAX_TITLE_CHAR_LENGTH);
```
mhammond, line 175: I do think we should add an ellipses - the client can't know whether we trimmed it or not so it can't add one. I believe it's likely to be seen in the tooltip for tabs at 512 chars on desktop devices. Another option would be to add a new bool to indicate if it was truncated so the client can add it, but that doesn't seem necessary.

components/tabs/src/storage.rs
```diff
@@ -362,6 +372,34 @@ impl TabsStorage {
...
+        while compute_serialized_size(&tabs) > max_serialized_size {
+            tabs.pop();
+        }
+    }
+    Ok(tabs.to_vec())
+}
+
+fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
+    let json = serde_json::to_string(v).unwrap();
+    mem::size_of_val(&*json)
```
mhammond, line 400: isn't this just `json.len()`?

components/tabs/src/storage.rs
```diff
@@ -362,6 +372,34 @@ impl TabsStorage {
...
+        // Keep dropping off the last entry until the data fits.
+        while compute_serialized_size(&tabs) > max_serialized_size {
+            tabs.pop();
+        }
+    }
+    Ok(tabs.to_vec())
+}
+
+fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
+    let json = serde_json::to_string(v).unwrap();
```
mhammond, line 399: This unwrap is wrong (even though I don't think it will ever fail tbh, but everywhere else handles serde failures so we should too). It looks like this is the only reason the calling functions need a `Result<>`

components/tabs/src/storage.rs
```diff
@@ -362,6 +372,34 @@ impl TabsStorage {
     }
 }
 
+// Trim the amount of tabs in a list to fit the specified memory size
+fn trim_tabs_length(
+    tabs: &mut Vec<RemoteTab>,
+    payload_size_max_bytes: usize,
+) -> Result<Vec<RemoteTab>> {
+    // Same as what desktop does in util.js
```
mhammond, line 380: nit: maybe the full path (ie, services/sync/modules/util.js)?

components/tabs/src/storage.rs
```diff
@@ -146,29 +149,36 @@ impl TabsStorage {
         self.local_tabs.borrow_mut().replace(local_state);
     }
 
+    // We try our best to fit as many tabs in a payload as possible, this includes
+    // limiting the url history entries, title character count and finally drop enough tabs
+    // until we have small enough payload that the server will accept
```
mhammond, line 154: I wonder if it makes sense to call this sanitization code in `update_local_state`, especially given we hard-code the max payload size. Let's say that someone gives us 10MB worth of tab data, we will hold on to that 10MB in memory forever and every sync we will do the same trimming. OTOH, doing it in `update_local_state` means we have less in-memory and each sync doesn't need to recompute.

OTOH though, if `update_local_state` ends up being called many times before we actually sync, then it means we'd do this trimming even though we never upload it. Even on desktop, we have a 5 second time before upload, so there is a good chance this would be called a number of time before actually syncing.

So I *think* I've convinced myself the way you have done it is fine and that the extra memory overhead is probably OK when compared against the extra CPU overhead, but I'm leaving this comment in-place just incase someone else wants to argue otherwise :)

components/tabs/src/storage.rs
```diff
@@ -20,13 +20,16 @@ use sql_support::open_database::{self, open_database_with_flags};
...
 use std::collections::HashMap;
+use std::mem;
 use std::path::{Path, PathBuf};
 use sync15::{RemoteClient, ServerTimestamp};
 pub type TabsDeviceType = crate::DeviceType;
 pub type RemoteTabRecord = RemoteTab;
 
 pub(crate) const TABS_CLIENT_TTL: u32 = 15_552_000; // 180 days, same as CLIENTS_TTL
 const FAR_FUTURE: i64 = 4_102_405_200_000; // 2100/01/01
+const MAX_PAYLOAD_SIZE: usize = 256 * 1024; // Same as Desktop, see service.js or sync15/request.rs
```
mhammond, line 31: I believe desktop is stale in this regard, and that tabs have the same max-payload as anything else, which inspecting info/configuration tells me is `max_record_payload_bytes: 2097152`, so much bigger than this. It's actually super unfortunate that the sync engine doesn't have this available to it so it can't easily be passed as a param. I'm not sure if we need to go all the way to 2MB though, but it does seem unfortunately to be this small if we don't need to and users are actually not seeing all the tabs they want. I'll ask on #sync

mhammond (changes requested):

mhammond:
oh, I think we also need to ensure the tabs are sorted before trimming so the least recently used are removed?

components/tabs/src/storage.rs
skhamis: yeah from a quick test it looks like they're equivalent. I'll just use len here since it's simpler (might not even need to be a separate fn anymore, but i'll leave it as is)

```
[2023-02-15T21:36:30Z DEBUG tabs::storage] mem::size_of_val: 628768
[2023-02-15T21:36:30Z DEBUG tabs::storage] json.len(): 628768
```

components/tabs/src/storage.rs
skhamis: Per the conversation, we don't seem to have an agreed amount vs just going to the max -- I decided to put it at 1MB since it's (1) > 256k and (2) < 2MB, but obviously it's super easy to change.

skhamis:
Thanks for the feedback! This is looking in a much better state, took it out of draft and re-requesting reviews

components/tabs/src/storage.rs
```diff
@@ -362,6 +378,32 @@ impl TabsStorage {
...
+    let max_serialized_size = (payload_size_max_bytes / 4) * 3 - 1500;
+    let size = compute_serialized_size(tabs)?;
+    if size > max_serialized_size {
+        // Estimate a little more than the direct fraction to maximize packing
+        let cutoff = (tabs.len() * max_serialized_size) / size;
+        tabs.truncate(cutoff);
+
+        // Keep dropping off the last entry until the data fits.
+        while compute_serialized_size(tabs)? > max_serialized_size {
+            tabs.pop();
```
tarikeshaq, line 397: nit: This while loop may get expensive if a user has **A LOT** of tabs, since we clear them one at at time, then re-serialize them. but I imagine that's what desktop does too so I see no real practical harm... and any logic that tries to be too smart might get it wrong and over trim

components/tabs/src/storage.rs
```diff
@@ -362,6 +378,32 @@ impl TabsStorage {
     }
 }
 
+// Trim the amount of tabs in a list to fit the specified memory size
+fn trim_tabs_length(
+    tabs: &mut Vec<RemoteTab>,
+    payload_size_max_bytes: usize,
+) -> Result<Vec<RemoteTab>> {
```
tarikeshaq, line 385: Do we need to return the Vec if we are modifying it in places by passing a mutable reference? We might be able to get away by either:

- pass in an owned mut Vec, and return the same one back
- keep the mutable reference, but don't clone at the end and have the caller consume what was borrowed

components/tabs/src/storage.rs
```diff
@@ -362,6 +378,32 @@ impl TabsStorage {
     }
 }
 
+// Trim the amount of tabs in a list to fit the specified memory size
+fn trim_tabs_length(
```
tarikeshaq, line 382: I think this only ever practical for tabs, but if we happen to have a limit for history records/etc we might be able to make it generic over any `Vec<T>`, such that `T: Serialize` and have it a part of `sync15`

But since Tabs is the only one that we might hit limit with I like this here not to overengineer!

tarikeshaq (commented):
Looks great to me 🚀 Dropped a couple of thoughts and will leave approval to Mark!

components/tabs/src/storage.rs
mhammond: That's an interesting idea! One slight issue with this would be that the `Vec<>` in question might not be the entire payload - eg, for history, it would be the `visits` element which gets trimmed. I *think* this could still be expressed though, but agree we should wait for an actual other use-case.

components/tabs/src/storage.rs
mhammond: yeah, thanks - I agree, let's do this (ie, drop the result vec and just re-use the (not trimmed) arg we passed in

components/tabs/src/storage.rs
mhammond: I nearly wrote something similar before - where we could do something even funkier by assuming the JSON payload is just an array, and serialize it somewhat manually - eg, serialize each element, and after doing it, see if we can fit it in, and finally put the array payload together manually with something like `format!("[{}]", serialized_tabs.join(",")` or something - then we only serialize each payload once and the ones that don't fit (bar 1) don't get serialized at all! But that's also a bigger change that I think we can optimize later if we care.

mhammond (changes requested):
I think that trivial change to avoid the new vec makes sense!

components/tabs/src/storage.rs
skhamis: Great suggestion @tarikeshaq, done!

components/tabs/src/storage.rs
```diff
@@ -146,30 +148,43 @@ impl TabsStorage {
...
-                    .collect(),
-            );
+                    }
+
+                    tab.url_history = sanitized_history;
+                    // Truncate the title to some limit and append ellipsis
+                    // to incate that we've truncated
+                    if tab.title.len() > MAX_TITLE_CHAR_LENGTH {
+                        tab.title.truncate(MAX_TITLE_CHAR_LENGTH - 3);
+                        tab.title.push_str("...");
```
mhammond, line 178: damn - sorry Sammy, I missed this before, but I think we need to use a unicode ellipsis character - `\x2026`

mhammond (approved):
Thanks Sammy!

PR #5384, opened 2023-02-16
Update Android Components and related Gradle dependencies to v110

rvandermeulen:
Nothing earth-shattering here, just updating AC to the recently-released 110.0.1 release and updating common Gradle dependencies to the same versions it's using.


skhamis (approved):
LGTM! Thanks for keeping on this!

PR #5386, opened 2023-02-17
Ensure syncing history sees the most recent 5000 records

mhammond:
The [docs](https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.1.html#url-semantics) don't specify what the default value is. I asked in slack whether the default is `newest`, but regardless, it makes sense for us to specify it - [desktop does](https://searchfox.org/mozilla-central/rev/66aa740e65a343659a7446b890145781f1b6a344/services/sync/modules/engines.sys.mjs#1221)

### DIFF
--- a/components/places/src/history_sync/engine.rs
+++ b/components/places/src/history_sync/engine.rs
@@ -10,7 +10,7 @@ use interrupt_support::SqlInterruptScope;
 use std::sync::Arc;
 use sync15::engine::{
     CollSyncIds, CollectionRequest, EngineSyncAssociation, IncomingChangeset, OutgoingChangeset,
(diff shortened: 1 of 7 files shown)


mhammond:
Desktop avoids this foot-gun, so I pushed a larger version. Also avoid confusion between "get" and "post" params - makes no sense to have an `?newer=timestamp` param on a POST.

skhamis (approved):
Sorry this took awhile, thanks for making this patch! -- makes sense to make what desktop does (and obv ensure we always see most recent first). LGTM!

bendk (approved):
Looks good to me to.  I like the cleanup.

PR #5389, opened 2023-02-20
Refactor of push component so PushManager handles uaid [firefox-android: tarikeshaq:1829982]

tarikeshaq:
It was about time we give the push component a bit more love.

Opening as a draft for now - might break it down to different PRs... but it tackles some confusion we've had for a while

The main driver for all of this, is making sure our handling of the `uaid` is reasonable. Now the story is:
- it gets loaded from persisted storage (if it exists) on creation of PushManager
- When we get a new one, we update both the persisted storage and the in-memory value
- every other operation that doesn't update the `uaid` will consume the in-memory value
- PushManager owns the uaid and passes it around. No other object holds on to it

## List of changes
- Most importantly, it removes the caching happening at the lightweight "Connection" layer.
- Makes the `Connection` responsible for mostly straightforward HTTP requests
- Moves the in-memory caching of the `uaid` and `auth` to the `PushManager`
- Makes the `connection` long-lasting, as opposed to being created each request, it's an HTTP client that's attached to the push manager, but it doesn't host any in-memory storage beyond it's own configuration
- The push manager now loads up the uaid and secret on the creation and holds on to them. The push manager will appropriately update it's in-memory value when it updates the underlying persisted store
- Most of the `Connection` functions now take the `uaid` and `auth`. Error handling is done at the push manager layer to ensure we already have all we need if we're triggering the HTTP client.
- Removes all the testing code that lived within the actual functions themselves.
- This PR has the `PushManager` now consume a generic type that implements `Connection`, which allows us to mock `Connection` instead of implementing if statements to check if we are in test mode.
- Various improvements for the structure of the crate


components/push/src/internal/subscriber.rs
```diff
@@ -276,21 +300,89 @@ impl PushManager {
...
+            let components = EcKeyComponents::new(
+                base64::decode_config(PRIV_KEY_D, base64::URL_SAFE_NO_PAD).unwrap(),
+                base64::decode_config(PUB_KEY_RAW, base64::URL_SAFE_NO_PAD).unwrap(),
+            );
+            let auth = base64::decode_config(AUTH_RAW, base64::URL_SAFE_NO_PAD).unwrap();
+            Ok(Key {
+                p256key: components,
+                auth,
+            })
+        });
```
mhammond, line 385: love these new tests 😍

mhammond (commented):
Just a drive-by with a few random thoughts, but I like this!

components/push/src/internal/storage/db.rs
```diff
@@ -257,6 +240,14 @@ impl Storage for PushDb {
         self.execute_cached(query, &[(":k", &key), (":v", &value)])?;
         Ok(())
     }
+
+    fn from_path(path: &Option<String>) -> Result<Self> {
```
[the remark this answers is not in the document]
tarikeshaq: Good idea!! split this into an `open` function, that has two implementations, a `cfg(not(test))` one that opens the db and a `cfg(test)` one that runs in memory.. made the `PushDb`'s `open_in_memory` function also `#[cfg(test)]`

components/push/src/internal/communications.rs
```diff
@@ -211,7 +195,7 @@ impl Connection for ConnectHttp {
             &options.sender_id
         );
         // Add the Authorization header if we have a prior subscription.
```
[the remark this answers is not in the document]
tarikeshaq: Yeah I think so.

`auth` is an auth token that gets sent as an http header, but all requests we send with the auth token, have the `uaid` in their path. The moment we lose our `uaid`, we purge both `uaid` and `auth` and get new ones next time we subscribe.

given that we store both along each other, and end up sending both, and resetting both makes me wonder why we **need** both... It also makes me wonder if there is an edge case here we're missing 🤔 Like what if the `auth` gets revoked on us, I wonder if that ever happens and what type of response we get from autopush

components/push/src/internal/communications.rs
```diff
@@ -189,11 +154,30 @@ impl ConnectHttp {
...
+    // Connect to the Autopush server
+    fn connect(options: PushConfiguration) -> error::Result<ConnectHttp> {
+        // find connection via options
+
+        if options.socket_protocol.is_some() && options.http_protocol.is_some() {
+            return Err(CommunicationError(
+                "Both socket and HTTP protocols cannot be set.".to_owned(),
+            ));
+        };
+        if options.socket_protocol.is_some() {
```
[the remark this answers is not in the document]
tarikeshaq: Good call! Took it as an opportunity to also clean up the `bridge_type` and make it an enum in the configuration

components/push/android/src/test/java/mozilla/appservices/push/PushTest.kt
```diff
@@ -1,277 +0,0 @@
...
-        // On a new native OS registration ID change:
-        manager.update(newSubscriptionID)
-            // sets the new registration ID (sender ID) on the server. Returns a false if this
-            // operation fails. A failure may prevent future messages from being received.
-
-      ```
-    */
-
-    @Test
-    fun testAesgcmDecryption() {
```
tarikeshaq, line 138: All those tests were moved into the Rust layer, they mostly were testing that UniFFI works - but I don't think we need such tests anymore!

We can re-add more tests at this layer if we have logic at this layer (for example, a new telemetry layer here, etc)

components/push/src/internal/subscriber.rs
```diff
@@ -186,39 +216,44 @@ impl PushManager {
...
         Ok(())
     }
 
     pub fn update(&mut self, new_token: &str) -> error::Result<bool> {
-        self.store.set_registration_id(new_token)?;
-        let mut conn = self.make_connection()?;
-        // It's OK if we don't have a uaid yet - that means we can't have any subscriptions,
-        // and we've saved our registration_id, so will use it on our first subscription.
-        if conn.uaid.is_none() {
+        if self.registration_id.as_deref() == Some(new_token) {
```
tarikeshaq, line 234: We shouldn't send a network request if we're already up to date! If the server loses our registration everything collapses anyways and we can recover through the existing recovery mechanism (verify_connection)

components/push/src/lib.rs
```diff
@@ -183,20 +183,17 @@
 uniffi::include_scaffolding!("push");
 // All implementation detail lives in the `internal` module
 mod internal;
+use std::fmt::Display;
 use std::sync::Mutex;
 mod error;
 use error::*;
 
-// The following are only exposed for use by the examples
-pub use error::Result as InternalResult;
```
tarikeshaq, line 191: The example now uses the high-level `PushManager` itself! and no longer needs access to internal values 😄

components/push/src/internal/config.rs
```diff
@@ -2,44 +2,112 @@
...
+/// Please contact services back-end for any additional bridge protocols.
+///
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BridgeType {
+    Fcm,
+    Adm,
+    Apns,
+}
+
+#[cfg(test)]
```
mhammond, line 26: nice

mhammond (approved):
This is fantastic, thanks Tarik!

components/push/src/internal/communications.rs
```diff
@@ -5,138 +5,192 @@
...
 pub use rate_limiter::PersistedRateLimiter;
 
 const UAID_NOT_FOUND_ERRNO: u32 = 103;
-#[derive(Debug)]
+#[derive(Deserialize, Debug)]
+/// The response from the `/registration` endpoint
 pub struct RegisterResponse {
     /// The UAID associated with the request
-    pub uaid: String,
+    pub uaid: Option<String>,
```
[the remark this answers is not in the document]
tarikeshaq: I think your confusion is rooted from a confusion in the code - the register endpoint (for the very first subscription) is different from the subscribe endpoint (for subsequent subscriptions)

Before the PR they were fully coupled, before my latest commit, I separated them only at the HTTP layer, but tried to force their responses to look the same (thus the UAID being an `Option`)

Now, I fully separate them, even in how they are handled. The [register endpoint returns the uaid and secret](https://github.com/mozilla-services/autopush-rs/blob/7641d18d9bfe53311099b2130302668dd7d2e113/autoendpoint/src/routes/registration.rs#L71-L76), the [subscribe endpoint does not](https://github.com/mozilla-services/autopush-rs/blob/7641d18d9bfe53311099b2130302668dd7d2e113/autoendpoint/src/routes/registration.rs#L148-L152)

I fixed the return types accordingly and removed all the code that was confused.. we had code that assumed that autopush could return a UAID on subscription requests, but that was false

tarikeshaq:
TLDR: I'll probably land this next week

I think this is in a good place to land, but I'll hold off until the start of 113 nightly cycle and I'll also do some testing on the simulator

I have an iOS branch that **almost** works with this too, so I'll see if I can get that working. If so, the push on iOS epic should be trivial 😄

tarikeshaq:
This is pretty much ready to land but waiting for approval on https://github.com/mozilla-mobile/firefox-android/pull/1780

tarikeshaq:
The android PR is approved, and ran branch builds in https://github.com/mozilla/application-services/runs/13875375673 and and the all branch build tests passed. Going to land this, manually trigger a nightly release and land the android change!

PR #5390, opened 2023-02-21
BridgedEngine now uses no generic types and is `Send+Sync`.

mhammond:
* A key, fundamental difference between `SyncEngine` and `BridgedEngine`
  is the latter's use of a generic `Error` type. This isn't necessary
  in practice and makes the uniffication of this trait more difficult,
  so this patch uses `anyhow::Error` instead.

* While adjusting `golden_gate` for this change, I noticed alot of
  additional complexity around the lifetime and the fact the trait
  is not `Send+Sync` - which will be a requirement if we want to
  expose this via UniFFI, so thought I'd do this at the same time.
  Tabs was already Send+Sync, but for webext-storage we now use
  a similar pattern used in all other engines - although here,
  inspired by `golden_gate`'s teardown code, we use a `Weak<Store>`
  instead of an `Arc<>` - which is probably what all engines should
  also do!

I'll put up a phabricator patch for the m-c side of this.

### DIFF
--- a/components/sync15/src/engine/bridged_engine.rs
+++ b/components/sync15/src/engine/bridged_engine.rs
@@ -2,8 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::bso::{IncomingBso, OutgoingBso};
+use anyhow::Result;
 
+use crate::bso::{IncomingBso, OutgoingBso};
 use crate::Guid;
 
 /// A BridgedEngine acts as a bridge between application-services, rust
@@ -16,27 +17,24 @@ use crate::Guid;
 /// [SyncEngine](crate::SyncEngine) trait), so this BridgedEngine trait adapts
 /// between the 2.
 pub trait BridgedEngine {
-    /// The type returned for errors.
-    type Error;
-
     /// Returns the last sync time, in milliseconds, for this engine's
     /// collection. This is called before each sync, to determine the lower
     /// bound for new records to fetch from the server.
-    fn last_sync(&self) -> Result<i64, Self::Error>;
+    fn last_sync(&self) -> Result<i64>;
 
     /// Sets the last sync time, in milliseconds. This is called throughout
     /// the sync, to fast-forward the stored last sync time to match the
     /// timestamp on the uploaded records.
-    fn set_last_sync(&self, last_sync_millis: i64) -> Result<(), Self::Error>;
+    fn set_last_sync(&self, last_sync_millis: i64) -> Result<()>;
 
     /// Returns the sync ID for this engine's collection. This is only used in
     /// tests.
-    fn sync_id(&self) -> Result<Option<String>, Self::Error>;
+    fn sync_id(&self) -> Result<Option<String>>;
 
     /// Resets the sync ID for this engine's collection, returning the new ID.
     /// As a side effect, implementations should reset all local Sync state,
     /// as in `reset`.
-    fn reset_sync_id(&self) -> Result<String, Self::Error>;
+    fn reset_sync_id(&self) -> Result<String>;
 
     /// Ensures that the locally stored sync ID for this engine's collection
     /// matches the `new_sync_id` from the server. If the two don't match,
@@ -45,48 +43,48 @@ pub trait BridgedEngine {
     /// `new_sync_id`, or a different one if the engine wants to force other
     /// devices to reset their Sync state for this collection the next time they
     /// sync.
-    fn ensure_current_sync_id(&self, new_sync_id: &str) -> Result<String, Self::Error>;
+    fn ensure_current_sync_id(&self, new_sync_id: &str) -> Result<String>;
 
     /// Tells the tabs engine about recent FxA devices. A bit of a leaky abstration as it only
     /// makes sense for tabs.
     /// The arg is a json serialized `ClientData` struct.
-    fn prepare_for_sync(&self, _client_data: &str) -> Result<(), Self::Error> {
+    fn prepare_for_sync(&self, _client_data: &str) -> Result<()> {
         Ok(())
     }
 
     /// Indicates that the engine is about to start syncing. This is called
     /// once per sync, and always before `store_incoming`.
-    fn sync_started(&self) -> Result<(), Self::Error>;
+    fn sync_started(&self) -> Result<()>;
 
     /// Stages a batch of incoming Sync records. This is called multiple
     /// times per sync, once for each batch. Implementations can use the
     /// signal to check if the operation was aborted, and cancel any
     /// pending work.
-    fn store_incoming(&self, incoming_records: Vec<IncomingBso>) -> Result<(), Self::Error>;
+    fn store_incoming(&self, incoming_records: Vec<IncomingBso>) -> Result<()>;
 
     /// Applies all staged records, reconciling changes on both sides and
     /// resolving conflicts. Returns a list of records to upload.
-    fn apply(&self) -> Result<ApplyResults, Self::Error>;
+    fn apply(&self) -> Result<ApplyResults>;
 
     /// Indicates that the given record IDs were uploaded successfully to the
     /// server. This is called multiple times per sync, once for each batch
     /// upload.
-    fn set_uploaded(&self, server_modified_millis: i64, ids: &[Guid]) -> Result<(), Self::Error>;
+    fn set_uploaded(&self, server_modified_millis: i64, ids: &[Guid]) -> Result<()>;
 
     /// Indicates that all records have been uploaded. At this point, any record
     /// IDs marked for upload that haven't been passed to `set_uploaded`, can be
     /// assumed to have failed: for example, because the server rejected a record
     /// with an invalid TTL or sort index.
-    fn sync_finished(&self) -> Result<(), Self::Error>;
+    fn sync_finished(&self) -> Result<()>;
 
     /// Resets all local Sync state, including any change flags, mirrors, and
     /// the last sync time, such that the next sync is treated as a first sync
     /// with all new local data. Does not erase any local user data.
-    fn reset(&self) -> Result<(), Self::Error>;
+    fn reset(&self) -> Result<()>;
 
     /// Erases all local user data for this collection, and any Sync metadata.
     /// This method is destructive, and unused for most collections.
-    fn wipe(&self) -> Result<(), Self::Error>;
+    fn wipe(&self) -> Result<()>;
 }
 
 // TODO: We should replace this with OutgoingChangeset to reduce the number
--- a/components/sync15/src/engine/bridged_engine.rs
+++ b/components/sync15/src/engine/bridged_engine.rs
@@ -16,7 +16,7 @@ use crate::Guid;
 /// implemented in Rust use a different shape (specifically, the
 /// [SyncEngine](crate::SyncEngine) trait), so this BridgedEngine trait adapts
 /// between the 2.
-pub trait BridgedEngine {
+pub trait BridgedEngine: Send + Sync {
     /// Returns the last sync time, in milliseconds, for this engine's
     /// collection. This is called before each sync, to determine the lower
     /// bound for new records to fetch from the server.
--- a/components/tabs/src/sync/bridge.rs
+++ b/components/tabs/src/sync/bridge.rs
@@ -94,7 +94,7 @@ impl BridgedEngine for BridgedEngineImpl {
             };
             sync_impl.reset(&EngineSyncAssociation::Connected(new_coll_ids))?;
         }
-        Ok(sync_id.to_string()) // this is a bit odd, why the result?
+        Ok(sync_id.to_string())
     }
 
     fn prepare_for_sync(&self, client_data: &str) -> Result<()> {
--- a/components/tabs/src/sync/bridge.rs
+++ b/components/tabs/src/sync/bridge.rs
@@ -2,12 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::sync::{Arc, Mutex};
-
-use crate::error::{ApiResult, TabsApiError};
+use crate::error::TabsApiError;
 use crate::sync::engine::TabsSyncImpl;
 use crate::TabsStore;
-use error_support::handle_error;
+use anyhow::Result;
+use std::sync::{Arc, Mutex};
 use sync15::bso::{IncomingBso, OutgoingBso};
 use sync15::engine::{ApplyResults, BridgedEngine, CollSyncIds, EngineSyncAssociation};
 use sync15::{ClientData, ServerTimestamp};
@@ -45,10 +44,7 @@ impl BridgedEngineImpl {
 }
 
 impl BridgedEngine for BridgedEngineImpl {
-    type Error = TabsApiError;
-
-    #[handle_error(crate::Error)]
-    fn last_sync(&self) -> ApiResult<i64> {
+    fn last_sync(&self) -> Result<i64> {
         Ok(self
             .sync_impl
             .lock()
@@ -58,8 +54,7 @@ impl BridgedEngine for BridgedEngineImpl {
             .as_millis())
     }
 
-    #[handle_error(crate::Error)]
-    fn set_last_sync(&self, last_sync_millis: i64) -> ApiResult<()> {
+    fn set_last_sync(&self, last_sync_millis: i64) -> Result<()> {
         self.sync_impl
             .lock()
             .unwrap()
@@ -67,16 +62,14 @@ impl BridgedEngine for BridgedEngineImpl {
         Ok(())
     }
 
-    #[handle_error(crate::Error)]
-    fn sync_id(&self) -> ApiResult<Option<String>> {
+    fn sync_id(&self) -> Result<Option<String>> {
         Ok(match self.sync_impl.lock().unwrap().get_sync_assoc()? {
             EngineSyncAssociation::Connected(id) => Some(id.coll.to_string()),
             EngineSyncAssociation::Disconnected => None,
         })
     }
 
-    #[handle_error(crate::Error)]
-    fn reset_sync_id(&self) -> ApiResult<String> {
+    fn reset_sync_id(&self) -> Result<String> {
         let new_id = SyncGuid::random().to_string();
         let new_coll_ids = CollSyncIds {
             global: SyncGuid::empty(),
@@ -89,8 +82,7 @@ impl BridgedEngine for BridgedEngineImpl {
         Ok(new_id)
     }
 
-    #[handle_error(crate::Error)]
-    fn ensure_current_sync_id(&self, sync_id: &str) -> ApiResult<String> {
+    fn ensure_current_sync_id(&self, sync_id: &str) -> Result<String> {
         let mut sync_impl = self.sync_impl.lock().unwrap();
         let assoc = sync_impl.get_sync_assoc()?;
         if matches!(assoc, EngineSyncAssociation::Connected(c) if c.coll == sync_id) {
@@ -105,26 +97,23 @@ impl BridgedEngine for BridgedEngineImpl {
         Ok(sync_id.to_string()) // this is a bit odd, why the result?
     }
 
-    #[handle_error(crate::Error)]
-    fn prepare_for_sync(&self, client_data: &str) -> ApiResult<()> {
+    fn prepare_for_sync(&self, client_data: &str) -> Result<()> {
         let data: ClientData = serde_json::from_str(client_data)?;
-        self.sync_impl.lock().unwrap().prepare_for_sync(data)
+        Ok(self.sync_impl.lock().unwrap().prepare_for_sync(data)?)
     }
 
-    fn sync_started(&self) -> ApiResult<()> {
+    fn sync_started(&self) -> Result<()> {
         // This is a no-op for the Tabs Engine
         Ok(())
     }
 
-    #[handle_error(crate::Error)]
-    fn store_incoming(&self, incoming: Vec<IncomingBso>) -> ApiResult<()> {
+    fn store_incoming(&self, incoming: Vec<IncomingBso>) -> Result<()> {
         // Store the incoming payload in memory so we can use it in apply
         *(self.incoming.lock().unwrap()) = incoming;
         Ok(())
     }
 
-    #[handle_error(crate::Error)]
-    fn apply(&self) -> ApiResult<ApplyResults> {
+    fn apply(&self) -> Result<ApplyResults> {
         let mut incoming = self.incoming.lock().unwrap();
         // We've a reference to a Vec<> but it's owned by the mutex - swap the mutex owned
         // value for an empty vec so we can consume the original.
@@ -141,22 +130,20 @@ impl BridgedEngine for BridgedEngineImpl {
         })
     }
 
-    #[handle_error(crate::Error)]
-    fn set_uploaded(&self, server_modified_millis: i64, ids: &[SyncGuid]) -> ApiResult<()> {
-        self.sync_impl
+    fn set_uploaded(&self, server_modified_millis: i64, ids: &[SyncGuid]) -> Result<()> {
+        Ok(self
+            .sync_impl
             .lock()
             .unwrap()
-            .sync_finished(ServerTimestamp::from_millis(server_modified_millis), ids)
+            .sync_finished(ServerTimestamp::from_millis(server_modified_millis), ids)?)
     }
 
-    #[handle_error(crate::Error)]
-    fn sync_finished(&self) -> ApiResult<()> {
+    fn sync_finished(&self) -> Result<()> {
         *(self.incoming.lock().unwrap()) = Vec::default();
         Ok(())
     }
 
-    #[handle_error(crate::Error)]
-    fn reset(&self) -> ApiResult<()> {
+    fn reset(&self) -> Result<()> {
         self.sync_impl
             .lock()
             .unwrap()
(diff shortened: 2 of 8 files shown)


mhammond:
https://phabricator.services.mozilla.com/D170467 is the m-c changes. I only pinged @linabutler there as she has quite a bit of context around that code, but welcome all feedback.

components/webext-storage/src/db.rs
```diff
@@ -69,24 +69,16 @@ impl StorageDb {
...
-                err.into(),
-            )
+    /// remain open and the connection will be leaked - we used to return the
+    /// underlying connection so the caller can retry but (a) that's very tricky
+    /// in an Arc<Mutex<>> world and (b) we never actually took advantage of
+    /// that retry capability.
+    pub fn close(self) -> Result<()> {
+        self.writer.close().map_err(|(writer, err)| {
+            // If we just let `writer` drop, the close would panic on failure.
+            // (Actually, not sure that's true? Regardless, this seems fine)
```
linabutler, line 79: I think [this is where the panic happens](https://github.com/rusqlite/rusqlite/blob/9880cdef12c1b624e351ed077dd71a2f77978f77/src/inner_connection.rs#L380), but maybe that's wrong?

components/webext-storage/src/sync/bridge.rs
```diff
@@ -38,56 +46,75 @@ impl<'a> BridgedEngine<'a> {
         delete_meta(tx, LAST_SYNC_META_KEY)?;
         Ok(())
     }
-}
 
-impl<'a> sync15::engine::BridgedEngine for BridgedEngine<'a> {
-    type Error = Error;
+    fn get_shared_db(&self) -> Result<Arc<SharedStorageDb>> {
+        match self.db.upgrade() {
```
tarikeshaq, line 51: nit: we can use `ok_or_else` here on the option!

components/webext-storage/src/sync/bridge.rs
```diff
@@ -20,14 +21,21 @@ const SYNC_ID_META_KEY: &str = "sync_id";
 /// `storage.sync` store work with Desktop's Sync implementation.
 /// Conceptually, it's similar to `sync15::Store`, which we
 /// should eventually rename and unify with this trait (#2841).
-pub struct BridgedEngine<'a> {
-    db: &'a StorageDb,
+///
+/// Unlike most of our other implementation which hold a strong reference
+/// to the store, this engine keeps a weak reference in an attempt to keep
+/// the desktop semantics as close as possible to what they were when the
+/// engines all took lifetime params to ensure they don't outlive the store.
```
tarikeshaq, line 28: In other words: What used to be enforced by the compiler, will now become an error when we upgrade the reference. 

I think it's totally fine, and worth it to make the engine Sync + Send to be UniFFIed!

components/webext-storage/src/store.rs
```diff
@@ -110,18 +116,32 @@ impl Store {
...
-        self.db.close().map_err(|(db, err)| (Store { db }, err))
+    pub fn close(self) -> Result<()> {
+        // Even though this consumes `self`, the fact we use an Arc<> means
+        // we can't guarantee we can actually consume the inner DB - so do
+        // the best we can.
+        let shared: SharedStorageDb = match Arc::try_unwrap(self.db) {
+            Ok(shared) => shared,
+            _ => {
+                // Probably means the sync engine remains alive?
+                log::warn!("Attempting to close a store while other DB references exist.");
```
tarikeshaq, line 138: Hmmm what do you think about implementing `Drop` on the `SharedStorageDb`? That way whenever the underlying memory is dropped, we close the actual connection, regardless of when that happens... which should be **soon** after this `close` function runs... thoughts?


But maybe this `close` function could also be a `Drop` impl on the `Store`, but I'm not sure if we want to close a connection without dropping the whole store so that's another problem

tarikeshaq (commented):
Awesome!! Dropped a few comments, happy to approve if I'm the last on the queue, but this will benefit from a @linabutler review  😝

components/webext-storage/src/db.rs
```diff
@@ -105,6 +97,53 @@ impl DerefMut for StorageDb {
     }
 }
 
+// We almost exclusively use this SharedStorageDb
+pub struct SharedStorageDb {
+    db: Mutex<StorageDb>,
```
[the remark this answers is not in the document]
linabutler: Yep! `Store` always had some kind of external synchronization—that was `LazyStore` in Desktop, which kept it inside a `OnceCell<Mutex<Store>>`, or a `ConcurrentHandleMap` in the (hand-written 😩) FFI bindings—so it didn't need to be `Send + Sync`.

Pushing the synchronization down into this crate lets `Store` be `Send + Sync` for the transition to UniFFI later.

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -2,8 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::bso::{IncomingBso, OutgoingBso};
+use anyhow::Result;
```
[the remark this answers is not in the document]
linabutler: I think that could work! The tricky part before was the `sync15` / `sync15_traits` split, so we didn't have a common error type to use—I think that's why we used the associated `Error` type. As a bonus, the associated `Error` type also made it "easier" to turn a bridged engine error into something we could bubble up through Golden Gate, before turning it into an `nsresult`.

But now that @mhammond has merged `sync15` and `sync15_traits`, and the "bonus" of converting to `nsresult` turned out to not be helpful, I think it should be easier to use `sync15::Error` down the road!

components/webext-storage/src/db.rs
mhammond: I think that is completely correct, thanks! I'll update the comment

components/tabs/src/sync/bridge.rs
```diff
@@ -105,26 +97,23 @@ impl BridgedEngine for BridgedEngineImpl {
         Ok(sync_id.to_string()) // this is a bit odd, why the result?
```
linabutler, line 97: This isn't at all related to your change, but I saw this comment and have an answer! 😆

* `ensure_current_sync_id` returns an `ApiResult` in case the bridged engine persists the sync ID in SQLite or another store that can return errors.
* It returns the sync ID to use (instead of `()`), because the engine can decide it doesn't want to adopt the `sync_id` it got from the server, and wants the server to use the device's sync ID instead. I think the only engine to take advantage of that is bookmarks, on the first sync after restoring from a backup.

In just about every case, the device handles sync ID mismatches by going "my sync ID is X but you're saying it should be Y, let me take X, reset all my local sync state, and merge all your records with mine."

On a first sync after a bookmarks restore, though, the device goes "my sync ID is X, and even though you're saying it should be Y, I want you to forget Y, write X to your `meta/global`, erase everything in your `bookmarks` collection, and take all my new records."

components/sync15/src/engine/bridged_engine.rs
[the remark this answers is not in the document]
mhammond: I'm not sure I fully understand how this would be useful. Consider, eg, places:
* The places sync code works almost exclusively with `places::Error` (ie, errors returned by the impl will always be `places::Error` and converted to anyhow right at the trait boundary)
* Places already has an [Into for `sync15::Error`](https://github.com/mozilla/application-services/blob/main/components/places/src/error.rs#L59)

Because `sync15` doesn't depend on places (and indeed, doesn't know how many engines might exist), I think the only sane way of doing this would be to have a new sync15::Error variant with, say `EngineError(String)` which all engines would need to use to convert their specific errors into a `sync15:Error` - which is just `anyhow:Error` but with more steps :)

Further, this would mean you can go from a `sync15::Error` into a `places::Error` *and* a way to go from a `places::Error` into a `sync15::Error` - which seems a little odd and would be impossible to round-trip sanely - ie, I don't think you could go from `places::Error` -> `sync15::Error` -> `places::Error`.

Although there's a good chance I'm misunderstanding something here :)

components/webext-storage/src/db.rs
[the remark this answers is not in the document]
mhammond: Further, note that for uniffi'd components, the stores *already* are `Send + Sync` - eg, tabs didn't require any changes in this area because the fact that it's already exposed via uniffi meant the requirement already existed. To put it another way, if we were ever asked to expose webext-storage for use in Android via UniFFI, we'd be somewhat forced to make the store `Send + Sync` even without considering the bridged engine.

components/webext-storage/src/db.rs
```diff
@@ -105,6 +97,53 @@ impl DerefMut for StorageDb {
     }
 }
 
+// We almost exclusively use this SharedStorageDb
+pub struct SharedStorageDb {
```
linabutler, line 101: thought: `SharedStorageDb` doesn't feel like quite the right name ("What makes it shared? With whom? 'Shared' as opposed to what?" are a few questions that went through my mind reading this the first time), but I'm struggling to think of a more descriptive one! 😅

Maybe `InterruptibleStorageDb`? It looks like we have a similar concept of an `InterruptibleConnection` in the hand-written Kotlin bindings that's a thread-safe database connection + interrupt handle.

WDYT?

components/sync15/src/engine/bridged_engine.rs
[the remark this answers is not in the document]
tarikeshaq: Ooof I didn't realize the engines already had `From<sync15::Error>` implementations :(

> , I think the only sane way of doing this would be to have a new sync15::Error variant with, say EngineError(String) which all engines would need to use to convert their specific errors into a sync15:Error - which is just anyhow:Error but with more steps :)

Yeah that doesn't sound great :( It might be possible to define **YET ANOTHER** error type in `sync15` that is meant to be a target for the SyncEngine traits.. but it's starting to get messy...

 I concede, I don't see anything wrong with `anyhow` **unless consumers want to handle specific errors without deref'ing**, which I don't think any consumers do


As a slightly related question: Why does `places::Error` implement `From<sync15::Error>`, I imagine it's because of the "stand-alone" sync we support for iOS's sake?

components/webext-storage/src/error.rs
```diff
@@ -49,8 +49,16 @@ pub enum ErrorKind {
...
 
-    #[error("Sync Payload Error: {0}")]
-    IncomingPayloadError(#[from] sync15::Error),
+    // When trying to close a connection but we aren't the exclusive owner of the containing Arc<>
+    #[error("Other shared references to this connection are alive")]
+    OtherConnectionReferencesExist,
+
+    // When `Weak::upgrade()` returns None.
+    #[error("The storage database has been closed")]
+    WeakReferenceDropped,
```
linabutler, line 58: suggestion:

```suggestion
    DatabaseClosed,
```

"Weak reference" is a bit of an implementation detail—and pedantically, I think it's the strong ref that got dropped, not the weak ref 🫢 I think your `#[error(...)]` message is super clear about what happened, though—let's rename the variant to match?

components/webext-storage/src/store.rs
```diff
@@ -110,18 +116,32 @@ impl Store {
...
 
     /// Closes the store and its database connection. See the docs for
     /// `StorageDb::close` for more details on when this can fail.
-    pub fn close(self) -> result::Result<(), (Store, Error)> {
-        self.db.close().map_err(|(db, err)| (Store { db }, err))
+    pub fn close(self) -> Result<()> {
+        // Even though this consumes `self`, the fact we use an Arc<> means
+        // we can't guarantee we can actually consume the inner DB - so do
+        // the best we can.
+        let shared: SharedStorageDb = match Arc::try_unwrap(self.db) {
```
linabutler, line 134: thought: Could we unwrap this and panic if we didn't consume the inner DB? Cloning the `Arc` with `self.db.clone()` feels like a programmer bug that we'd want to catch early—I don't think an application could / should be expected to recover from `OtherConnectionReferencesExist`...

```suggestion
        let shared = Arc::try_unwrap(self.db).expect("Shouldn't have any other DB references");
```

components/sync15/src/engine/bridged_engine.rs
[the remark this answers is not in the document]
mhammond: > I imagine it's because of the "stand-alone" sync we support for iOS's sake?

Yep, that's exactly why, so I guess eventually what I wrote above need not be true.

But as you imply above, nothing in `sync15` can possibly handle a specific `places` error because (a) it doesn't depend places and (b) given it doesn't even know which engines might exist it doesn't really make sense. If places ever did need to see a `sync15::Error` and know what the underlying error was we'd be in a bit of trouble, but `anyhow::Error` probably makes that easier anyway.

Where this *does* fall apart is a possible future where the sync15 engines needed to call a sync15 implemented function as part of doing the sync, and that error needs to be inspected by the sync15 caller of the engine trait (phew!) - eg, if the design of the engine was such that an engine would explicitly request a http request be made. But the design of the engine avoids that - eg, the http requests are all initiated by the sync process and the engine is provided with the result.

components/webext-storage/src/store.rs
tarikeshaq: I was wondering how we would end up in this state too!

One possible (but unlikely?) case is that while we were closing the store, there was a sync operation still running that upgraded it's `Weak` reference into a strong one before the dropping :( I'm not sure however if that is prevented at a higher level in m-c though 🤔

mhammond: huh - that's a good point. I don't think anything will prevent this. I was thinking we could lock the mutex, but you can't into_inner() a locked mutex. hrmph

tarikeshaq: FWIW this was the bases for my comment on the `Drop` implementation 😬  as long as the `Store` is dropped, eventually the `SharedStorageDb` will be dropped too! We can then close the underlying connection?

components/webext-storage/src/store.rs
linabutler: > whenever the underlying memory is dropped, we close the actual connection, regardless of when that happens...

I think the tricky part is that `close()` can fail. Rusqlite has something similar, and panics if it can't close a dropped connection, which unfortunately turns into a shutdown crash in Desktop because of `panic = "abort"`. (For comparison, `std::fs::File` is a little kinder, and ignores errors in its `Drop` implementation).

Surfacing `close()` errors on Desktop lets `AsyncShutdown.jsm` make some logging noise while still unblocking shutdown, and leaking the `StorageDb` avoids the shutdown crash.

components/sync15/src/engine/bridged_engine.rs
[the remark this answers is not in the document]
tarikeshaq: Aaah yeah that makes since - thanks @mhammond !!

components/webext-storage/src/store.rs
mhammond: but - to back up a bit, if `Arc::try_unwrap()` works, then we must have the only reference to the mutex, so it can't possibly be locked, right? eg, changing this to use `Mutex::get_mut()` also compiles

components/webext-storage/src/store.rs
```diff
@@ -26,23 +25,25 @@ use serde_json::Value as JsonValue;
 /// and causing write contention. For this reason, you should only call
 /// `Store::new()` (or `webext_store_new()`, from the FFI) once.
 pub struct Store {
-    db: StorageDb,
+    db: Arc<SharedStorageDb>,
```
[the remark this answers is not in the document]
linabutler: Yep, you got it! The mutex protects the `StorageDb` from concurrent mutation, and the `Arc` is what lets us give each `BridgedEngine` instance a weak reference to the `Mutex<StorageDb>`. The `Mutex<StorageDb>` could be accessed from any thread at any time.

components/webext-storage/src/store.rs
tarikeshaq: Yup! If we can get the owned SharedStorageDb, we are guaranteed to have the only reference! I think it's the case where `try_unwrap` fails that's the problem :(

mhammond: But that doesn't change the underlying problem - a sync *might* be running as we hit this code, so it's that `Arc::try_unwrap()` which fails. We could take Lina's advice to panic in that situation, but I'm not sure how we can actually guarantee that never happens? ie, it's not what will guarantee an in-flight sync operation completes before we hit this code if we are syncing at shutdown?

components/webext-storage/src/sync/bridge.rs
linabutler: Yep! This "one strong ref, many weak refs" pattern also helps when `close()` is racing against other operations. In that case, we don't want those other operations to prevent the connection from closing.

components/webext-storage/src/store.rs
tarikeshaq: > ie, what will guarantee an in-flight sync operation completes before we hit this code if we are syncing at shutdown?

Unless there is synchronization at a higher level I don't think we can unless we add **additional synchronization** :(

We do know though, that unless the process is killed before the sync operation actually finishes, we'll **eventually** drop the inner `SharedStorageDb` because the `Store` is consumed (thus the only long-lasting strong reference). The problem becomes that the underlying SQL connection is not closed (i think?)

linabutler: > One possible (but unlikely?) case is that while we were closing the store, there was a sync operation still running that upgraded it's Weak reference into a strong one before the dropping :( I'm not sure however if that is prevented at a higher level in m-c though 🤔
>
> [...]
>
> what will guarantee an in-flight sync operation completes before we hit this code if we are syncing at shutdown?

Oooh, no, that's a great point! Desktop currently uses a single [task queue](https://searchfox.org/mozilla-central/rev/234ae43d3793289167926ef91bfb5145da91d4aa/xpcom/threads/TaskQueue.h#27-31) shared between the bridged engine and the store, which guarantees that all operations scheduled on it will execute serially, and which is why I thought we could panic here—but I don't think we can guarantee that for the other backends? 🤯

If you access the same `Store` from two different Swift dispatch queues (or a single concurrent dispatch queue), or two different Kotlin coroutine contexts, or even two different concurrently executing Rust threads...I think you're both totally right; it's very possible to end up with an in-flight sync operation that upgraded its `Weak` ref to a strong one running on one queue / thread, and a close operation running on a different queue, at the same time.

mhammond: I also think the [mutex that the bridge holds around the store](https://searchfox.org/mozilla-central/rev/234ae43d3793289167926ef91bfb5145da91d4aa/toolkit/components/extensions/storage/webext_storage_bridge/src/store.rs#45) helped here too. For something like places, I don't think the writer is ever explicitly closed 😬 (IIUC, Fenix treats the places api as a singleton and happily leaks it at shutdown)

components/webext-storage/src/store.rs
tarikeshaq: eeek this is the comment I missed 😵‍💫 yeah this is getting tricky - thanks @linabutler 🙏 🙇‍♂️

components/webext-storage/src/store.rs
linabutler: Yeah, definitely! I think you and @tarikeshaq are on the money—without some kind of additional synchronization to ensure that all `Store` calls execute serially, `Arc::try_unwrap` can fail.

But, it's okay for callers to ignore that, _as long as_ they're treating the `Store` as a singleton that's only closed at shutdown. We might leak the connection then, but the leak won't last for long, because the kernel will close all open file descriptors when the process goes away...whereas using `.expect()`/`.unwrap()` will cause a crash.

Does that sound right?

mhammond: That does sound right. The other part of the picture is avoiding the panic, which I think we can do by, as Tarik suggests, implementing `Drop` on `StorageDb` and explicitly doing the closing/forget on `writer`?

The main reason I think "this is fine" is because this problem already exists for all our other engines - including tabs, which is in desktop, and because desktop has `panic=abort`, this wouldn't go unnoticed - so I'm also interested to know when the connection close will panic in practice given I don't think we've ever seen evidence of it actually failing (which isn't to suggest we shouldn't pretend it can't though)

linabutler: > The main reason I think "this is fine" is because this problem already exists for all our other engines - including tabs, which is in desktop, and because desktop has `panic=abort`, this wouldn't go unnoticed...

For sure, and I'm wondering if Desktop avoids running into this today—and why we haven't seen evidence of it actually failing—because it's always using a single task queue. The store is making the assumption that it won't be used concurrently, and, because the task queue guarantees that there's only one in-flight call at a time, that assumption holds.

components/tabs/src/sync/bridge.rs
mhammond: Thanks! I meant to delete that comment as I later realized you had already documented by the result value exists in bridged_engine.rs :)

components/webext-storage/src/db.rs
mhammond: The "shared" is really just to reflect the `Mutex<>` - what I don't particularly like about `InterruptibleConnection` is that the inner `StorageDb` protected by the mutex is *also* interruptible. Maybe `ConcurrentConnection`? Or the even more explicit `ThreadSafeConnection`?

components/webext-storage/src/store.rs
mhammond: So I think the resolution to this and the other massive thread is that:
* We hope/pray that the consumers somehow manage to avoid contention here, and we believe desktop does via its uses of the task queue.
* But just in-case, this code remains the same (ie, we never panic if we aren't the exclusive arc holder.)
* And just in-double-case, we implement `Drop` on the inner `StorageDb` as a fail-safe so we never implicitly panic as the connection drops.

I'll update the patch accordingly.

components/webext-storage/src/db.rs
linabutler: > Or the even more explicit `ThreadSafeConnection`?

Oh, I like that! `ThreadSafeStorageDb`?

components/webext-storage/src/store.rs
linabutler: Sounds perfect, thanks!

components/webext-storage/src/db.rs
```diff
@@ -105,6 +96,67 @@ impl DerefMut for StorageDb {
     }
 }
 
+/* ugh - can't call writer.close() as it consumes self
```
mhammond, line 99: I can't work out how to do this without a lot of pain. We can't use a cell, and Option seems quite painful to just handle this case. Does anyone have any ideas?

I wonder if we should consider opening an issue on rusqlite for this? I had a quick look and can't find any open or closed issues asking for this - for the sake of a single bool they might be open to a flag which just leaks the connection on failure. (Or maybe the flag arranging to call [sqlite_close_v2](https://www.sqlite.org/c3ref/close.html) would also be fine and we'd get the new behavior largely for "free" in terms of ruqslite implementation - but I think calling `sqlite_close()` probably remains best because at least you get an error telling you something went wrong but can just avoid the panic.

I made it a draft in the meantime while we discuss next steps.

mhammond:
https://github.com/rusqlite/rusqlite/issues/1292

components/webext-storage/src/db.rs
mhammond: I agree we can move ahead here. It looks like the rusqlite team are going to be fine with just not doing the panic on drop, so I might wait a week for so (there's no real hurry to get this in) and if it looks like we can upgrade to a new rusqlite soon I'll just update all the comments here, and probably even the `mem::forget` hacks etc so we can better document "we try hard to close it, but things are ok if we can't"

mhammond:
rusqlite is fairly slow making updates and I don't really want to ask them for a new release just for this given we are convinced these panics can't actually happen in practice.

I just added d4ae58c60ad293dc92bceb76302a04d154eae26d which tries to explain this situation, and I'm inclined to agree with @tarikeshaq that it's landable even with this theoretical panic. @linabutler @tarikeshaq do you mind having a quick look at that commit to check it's both accurate and complete and whether you agree it's fine to land this?

components/webext-storage/src/sync/bridge.rs
```diff
@@ -38,56 +46,74 @@ impl<'a> BridgedEngine<'a> {
         delete_meta(tx, LAST_SYNC_META_KEY)?;
         Ok(())
     }
-}
 
-impl<'a> sync15::engine::BridgedEngine for BridgedEngine<'a> {
-    type Error = Error;
+    fn get_shared_db(&self) -> Result<Arc<ThreadSafeStorageDb>> {
```
linabutler, line 50: nitpick: Let's name this method `thread_safe_storage_db()`, to match the new name you gave the type.

components/webext-storage/src/db.rs
```diff
@@ -227,6 +268,10 @@ pub mod test {
         let counter = ATOMIC_COUNTER.fetch_add(1, Ordering::Relaxed);
         StorageDb::new_memory(&format!("test-api-{}", counter)).expect("should get an API")
     }
+
+    pub fn new_shared_mem_db() -> Arc<ThreadSafeStorageDb> {
```
linabutler, line 272: nitpick: `new_thread_safe_storage_mem_db()`? `shared_db` is shorter, but feels a little out of place with `ThreadSafeStorageDb`.

components/webext-storage/src/sync/bridge.rs
```diff
@@ -160,94 +203,115 @@ mod tests {
...
-        assert!(get_meta::<i64>(engine.db, LAST_SYNC_META_KEY)?.is_some());
+        assert!(get_meta::<i64>(&db, LAST_SYNC_META_KEY)?.is_some());
         Ok(())
     }
 
     #[test]
     fn test_wipe() -> Result<()> {
-        let db = new_mem_db();
-        let engine = super::BridgedEngine::new(&db);
+        let strong = new_shared_mem_db();
```
linabutler, line 272: nitpick: Ditto here.

linabutler (approved):
Looks great, thanks! Let's get this landed.

PR #5391, opened 2023-02-22
Upgrades glean to 52.2.0

tarikeshaq:
Upgrades glean to 52.2.0

This is to unblock iOS from upgrading, see https://github.com/mozilla-mobile/firefox-ios/pull/13323#issuecomment-1440415294

Jan-Erik is on PTO and Travis is out sick, but this should be a safe upgrade to land.

I already tested locally that Firefox-iOS builds OK once they take a release with this patch

cc @lmarceau


lougeniaC64 (approved):
I was just trying to decide what to do with these changes in my a-s sync manager PR. Can definitely confirm that these changes build locally with iOS.

Issue #5394, opened 2023-02-22
Refactor rs_client::Client to be a trait and use it to improve Nimbus test cases.


MatthewTighe:
See [this comment thread](https://github.com/mozilla/application-services/pull/5369#discussion_r1103084587) for details.

This may not be the exact outcome, but suggestion is to:
1. Add client trait to rs_client and implement it using the exposed Client API
2. In Nimbus tests overwrite client with fake implementation to better capture all use cases

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3614)

PR #5405, opened 2023-02-24
Clean up Nimbus test dependencies and cover all parsing paths.

MatthewTighe:
This was intended to address https://github.com/mozilla/application-services/issues/5394 and the conversation that was started in [this comment thread](https://github.com/mozilla/application-services/pull/5369#discussion_r1103084587) but during implementation I discovered a couple things:

1. Turning `rs_client::Client` into a trait seemed non-trivial and I ran into difficulties updating the design accordingly. I think the best way to achieve it would probably be to add a trait, like `rs_client::ioHandler`, as a struct member of `Client` which could then be replaced under test. However, I felt it would be prudent to move forward with implementing the top-level Remote Settings client _first_ and see if that helps inform a design change like that. Otherwise, I am not sure complicating the design for test-only purposes is necessarily worth it given that `parse_experiments` is directly testable and `fetch_experiments` is 2 lines. 

2. In @jeddai's [comment here](https://github.com/mozilla/application-services/pull/5369#discussion_r1103251594) it was mentioned that the existing Nimbus test only covered the `Ok` path. I took this to mean that the test didn't cover any "unhappy paths", but when I read closer I saw that the test actually hides quite a bit of it. The malformed experiment case and the non-matching schema version case are both covered. Since I had to read closely to discover that, I added a couple comments. I also added a test to cover the `Error` case.

All that said, I thought it would be best for now to remove the unnecessary test dependencies, add the new test, and move on to the Remote Settings client. If any refactoring of the design of `rs_client` happens I will of course revisit that in Nimbus as well.  

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2007,7 +2007,6 @@ dependencies = [
  "hex",
  "jexl-eval",
  "log",
- "mockito",
  "once_cell",
  "rkv",
  "rs_client",
--- a/components/nimbus/Cargo.toml
+++ b/components/nimbus/Cargo.toml
@@ -43,7 +43,6 @@ glean-build = { path = "../external/glean/glean-core/build" }
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
-mockito = "0.31"
 env_logger = "0.7"
 clap = "2.33.3"
 tempfile = "3"
--- a/components/nimbus/src/tests/client/test_http_client.rs
+++ b/components/nimbus/src/tests/client/test_http_client.rs
@@ -2,35 +2,20 @@
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::client::SettingsClient;
-use crate::{Branch, BucketConfig, Experiment, FeatureConfig, RandomizationUnit, SCHEMA_VERSION};
-use mockito::mock;
-use rs_client::{Client, ClientConfig};
+use crate::parse_experiments;
+use crate::{
+    Branch, BucketConfig, Experiment, FeatureConfig, NimbusError, RandomizationUnit, SCHEMA_VERSION,
+};
 
 #[test]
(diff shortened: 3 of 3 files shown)


tarikeshaq (commented):
Thank you for taking the time to dig into this @MatthewTighe 🙏 

LGTM and +1 to thinking about the `rs_client` holistically before we invest too much in fixing tests

I'll leave approval to @jeddai 


side note, CI is complaining because of the Cargo.lock isn't up to date.. it's most likely because you need to update your local glean git submodule - because we landed an upgrade recently - so running:
```
$ git submodule update --init --recursive
```
Then any cargo operation should update it!

tarikeshaq:
Aaaah looks like `viaduct_reqwest` is also being used in the examples - probably worth keeping it around

MatthewTighe:
Trying to run `cargo run --package nimbus-sdk --example experiment --config components/nimbus/examples/config/config.json` hits an error:

```
error: failed to load config from `/Users/matt-tighe/dev/application-services/components/nimbus/examples/config/config.json`

Caused by:
  could not parse TOML configuration in `/Users/matt-tighe/dev/application-services/components/nimbus/examples/config/config.json`

Caused by:
  could not parse input as TOML

Caused by:
  TOML parse error at line 1, column 1
    |
  1 | {
    | ^
  Unexpected `{`
  Expected key or end of input
```

that I don't really understand since the config file is JSON. Am I misunderstanding how to use the command? I'm happy to look into removing viaduct from the example if that's the best move. Just let me know what you think @jeddai 🙂

tarikeshaq:
@MatthewTighe 

I think you'd want to run 
```
cargo run --package nimbus-sdk --example experiment -- --config components/nimbus/examples/config/config.json
```

To pass the config to the example (note the extra `--`)

Without the `--` the config gets passed as an argument to `cargo` and not our example, and [`cargo` expects configs to be tomls!](https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides)

MatthewTighe:
That did it, thanks!

jeddai (approved):
This looks great! Thank you! 🚢

PR #5407, opened 2023-02-24
Tabs storage should create DB when adding meta rows

### LINKED ISSUE #5404: Tabs engine drops client data if DB isn't there
Since https://github.com/mozilla/application-services/pull/5277, the Tabs engine fully relies on a DB for processing tabs. There are certain cases (new user first sync, TPS tests) Where a DB doesn't exist and needs to get created (Note: it doesn't on initialization https://github.com/mozilla/application-services/blob/main/components/tabs/src/storage.rs#L88 unlike other engines). 

While most functions create the DB if it doesn't exist, one scenario  [prepare_for_sync](https://github.com/mozilla/application-services/blob/main/components/tabs/src/sync/engine.rs#L121) doesn't and instead tries to put it in storage and returns `None` if it fails. We store the client data during sync so anytime a consumer tries to fetch the remote tabs for other clients, we have an up-to-date list with the most recently connected clients.

We should fix this so it creates the DB if it doesn't exist.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3628)
┆Epic: Desktop prioritized P2s
┆Sprint End Date: 2023-02-24

skhamis:
Fixes #5404 

1. `put_meta` should create the DB if doesn't exists (see issue, new user first sync or TPS tests)
2. get/delete shouldn't create the DBs, those are unchanged
3. Updated tests to ensure this scenario works

  - Note:

### DIFF
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -442,8 +442,9 @@ mod tests {
 
     #[test]
     fn test_tabs_meta() {
-        // Does not create the db on new
-        let mut db = TabsStorage::new("test.db");
+        let dir = tempfile::tempdir().unwrap();
+        let db_name = dir.path().join("test_tabs_meta.db");
+        let mut db = TabsStorage::new(db_name);
         let test_key = "TEST KEY A";
         let test_value = "TEST VALUE A";
         let test_key2 = "TEST KEY B";
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -346,12 +346,11 @@ impl TabsStorage {
     }
 
     pub(crate) fn put_meta(&mut self, key: &str, value: &dyn ToSql) -> Result<()> {
-        if let Some(db) = self.open_if_exists()? {
-            db.execute_cached(
-                "REPLACE INTO moz_meta (key, value) VALUES (:key, :value)",
-                &[(":key", &key as &dyn ToSql), (":value", value)],
-            )?;
-        }
+        let db = self.open_or_create()?;
+        db.execute_cached(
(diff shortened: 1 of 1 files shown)


components/tabs/src/storage.rs
```diff
@@ -443,12 +442,14 @@ mod tests {
 
     #[test]
     fn test_tabs_meta() {
-        let mut db = TabsStorage::new_with_mem_path("test");
+        // Does not create the db on new
```
mhammond, line 445: I don't understand this test change?

skhamis: The test before didn't catch that when we call `put_meta` on an empty DB -- the DB is not created. The test would've failed on `get_meta` due to no value (since no DB).

When I try to use `new` vs `new_with_mem_path` it expectantly fails before the new patch in `put_meta`. Even though both of them should've NOT created the db, i believe `new_with_mem_path` still might be creating the DB in mem.

I'll see if I can demonstrate this a little better in one of the tests

skhamis: Running the test:
```
let mut db = TabsStorage::new_with_mem_path("test");
assert!(db.open_if_exists().unwrap().is_none());
```
Fails with:
```
thread 'storage::tests::test_tabs_meta' panicked at 'assertion failed: db.open_if_exists().unwrap().is_none()', components/tabs/src/storage.rs:452:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

so I wanted to adjust the test to ensure that the any writes to the tabs DB, also creates it

mhammond (approved):
Thanks!

PR #5411, opened 2023-03-01
Ensure we only truncate at char bounds

skhamis:
Fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1819723

Even though this change was backed out from m-c, we will still need to fix this before it lands on any of the platforms. Per conversation with #l10n, using graphenes or code-points is probably fine since we're most likely truncating outside of the users view. The main thing is we're ONLY truncating at char bounds

  - Note:

### DIFF
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -175,7 +175,7 @@ impl TabsStorage {
                     // Truncate the title to some limit and append ellipsis
                     // to incate that we've truncated
                     if tab.title.len() > MAX_TITLE_CHAR_LENGTH {
-                        tab.title = safe_truncate(&tab.title, MAX_TITLE_CHAR_LENGTH);
+                        tab.title = safe_truncate(&tab.title, MAX_TITLE_CHAR_LENGTH - 1);
                         // Append an ellipsis '...' so clients know it's been truncated
                         tab.title.push('\u{2026}');
                     }
@@ -566,7 +566,7 @@ mod tests {
             last_used: 0,
         }]);
         let ellipsis_char = '\u{2026}';
-        let mut truncated_title = "a".repeat(MAX_TITLE_CHAR_LENGTH);
+        let mut truncated_title = "a".repeat(MAX_TITLE_CHAR_LENGTH - 1);
         truncated_title.push(ellipsis_char);
         assert_eq!(
             storage.prepare_local_tabs_for_upload(),
@@ -600,8 +600,8 @@ mod tests {
             },
         ]);
         let ellipsis_char = '\u{2026}';
-        let mut truncated_title = "😍".repeat(MAX_TITLE_CHAR_LENGTH);
-        let mut truncated_jp_title = "を".repeat(MAX_TITLE_CHAR_LENGTH);
+        let mut truncated_title = "😍".repeat(MAX_TITLE_CHAR_LENGTH - 1);
+        let mut truncated_jp_title = "を".repeat(MAX_TITLE_CHAR_LENGTH - 1);
         truncated_title.push(ellipsis_char);
         truncated_jp_title.push(ellipsis_char);
         let remote_tabs = storage.prepare_local_tabs_for_upload().unwrap();
@@ -623,15 +623,9 @@ mod tests {
                 },
             ]
         );
-        // We should be at max + our ellipsis
-        assert_eq!(
-            remote_tabs[0].title.chars().count(),
-            MAX_TITLE_CHAR_LENGTH + 1
-        );
-        assert_eq!(
-            remote_tabs[1].title.chars().count(),
-            MAX_TITLE_CHAR_LENGTH + 1
-        );
+        // We should be less than max
+        assert!(remote_tabs[0].title.chars().count() <= MAX_TITLE_CHAR_LENGTH);
+        assert!(remote_tabs[1].title.chars().count() <= MAX_TITLE_CHAR_LENGTH);
     }
     #[test]
     fn test_trim_tabs_length() {
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -172,13 +172,8 @@ impl TabsStorage {
                     }
 
                     tab.url_history = sanitized_history;
-                    // Truncate the title to some limit and append ellipsis
-                    // to incate that we've truncated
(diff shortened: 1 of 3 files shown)


components/tabs/src/storage.rs
```diff
@@ -174,7 +174,7 @@ impl TabsStorage {
                     // Truncate the title to some limit and append ellipsis
                     // to incate that we've truncated
                     if tab.title.len() > MAX_TITLE_CHAR_LENGTH {
-                        tab.title.truncate(MAX_TITLE_CHAR_LENGTH - 3);
+                        tab.title = safe_truncate(&tab.title, MAX_TITLE_CHAR_LENGTH - 1).to_string();
```
bendk, line 177: Is `- 1`  correct?  That depends on how many bytes that unicode character below uses.

I think this calculation should be use  `'\u{2026}'.len()` and maybe it should be inside `safe_truncate`

components/tabs/src/storage.rs
```diff
@@ -399,6 +399,13 @@ fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
     serde_json::to_string(v).unwrap_or_default().len()
 }
 
+fn safe_truncate(s: &str, max_chars: usize) -> &str {
+    match s.char_indices().nth(max_chars) {
+        None => s,
+        Some((idx, _)) => &s[..idx],
+    }
```
bendk, line 406: I think this truncates things based on unicode chars, but you want to be truncating based on the `idx` values.

components/tabs/src/storage.rs
```diff
@@ -571,6 +578,44 @@ mod tests {
...
+        }, RemoteTab {
+            title: "を".repeat(MAX_TITLE_CHAR_LENGTH + 5), // Fill a string more than max
+            url_history: vec!["https://foo_jp.bar".to_owned()],
+            icon: None,
+            last_used: 0,
+        }]);
+        let mut truncated_title = "😍".repeat(MAX_TITLE_CHAR_LENGTH - 1);
+        let mut truncated_jp_title = "を".repeat(MAX_TITLE_CHAR_LENGTH - 1);
+        truncated_title.push('\u{2026}');
+        truncated_jp_title.push('\u{2026}');
```
bendk, line 598: I think you should check that the truncated titles are exactly `MAX_TITLE_CHAR_LENGTH` bytes here to makes sure you're testing the right thing.

bendk (changes requested):
I think this still doesn't do the right thing for some unicode values, but I'm not totally sure.  I think the main thing is to update the unit test to make sure that  `truncated_title` and `truncated_jp_title` have exactly the right number of bytes.

components/tabs/src/storage.rs
skhamis: you're correct on this! I updated to use the length of the ellipsis (and also moved all of that to the truncating method itself

components/tabs/src/storage.rs
skhamis: hmm I'm not sure if this is quite right, Rust treats all strings as UTF8 https://doc.rust-lang.org/book/ch08-02-strings.html#what-is-a-string but i'll have to double check on this

components/tabs/src/storage.rs
skhamis: Given that we now truncate "correctly" we might not be able to exactly get the `MAX_TITLE_CHAR_LENGTH` since if something was `513` but 2 bytes then we'd get `511` if i'm not mistaken. Though I should add some extra tests to validate we are within the proper bytes

mhammond:
IIUC, we are currently truncating at byte boundaries. We probably want to truncate at a grapheme boundary though - I don't think just truncating at a character boundary is going to be ok in all locales

skhamis:
Places might need to eventually get updated: https://github.com/mozilla/application-services/blob/ca14a8c332e4e0f8d57593d42c1e9f673466c7a7/components/places/src/util.rs#L11 if we switch to a grapheme-style truncation

mhammond (approved):
if #l10n is happy with this, I'm happy too :)

mhammond (changes requested):
Outcome is that we should just do code-points, so don't need the new dep - sorry about that!

components/tabs/src/storage.rs
```diff
@@ -399,6 +394,23 @@ fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
...
 
+// Similar to places/utils.js
+// handles the case where `s.is_char_boundary(max_len)`
+// is false (which would otherwise panic).
+pub fn slice_up_to(s: String, max_len: usize) -> String {
+    if max_len >= s.len() {
+        return s;
+    }
+    let mut idx = max_len;
+    while !s.is_char_boundary(idx) {
```
bendk, line 405: Note: I'm pretty sure this check is O(1), and therefore this is a better method than iterating with `char_indices()` (
https://doc.rust-lang.org/src/core/num/mod.rs.html#882).

components/tabs/src/storage.rs
```diff
@@ -399,6 +394,23 @@ fn compute_serialized_size(v: &Vec<RemoteTab>) -> usize {
...
 }
 
+// Similar to places/utils.js
+// handles the case where `s.is_char_boundary(max_len)`
+// is false (which would otherwise panic).
+pub fn slice_up_to(s: String, max_len: usize) -> String {
+    if max_len >= s.len() {
+        return s;
+    }
+    let mut idx = max_len;
```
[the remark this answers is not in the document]
skhamis: You're right on this, however since our "limit" is quite arbitrary (just to ensure we don't have insanely long titles in the payload) and these are only for local tabs to be uploaded, there wouldn't be a scenario where their constantly serialized. Though I believe to keep the code consistent I'll go with this.

bendk (approved):

components/tabs/src/storage.rs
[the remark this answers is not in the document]
bendk: That's fine with me, just make sure to document how it works.

components/tabs/src/storage.rs
skhamis: Interestingly enough, `is_utf8_char_boundary()` gets called by `is_char_boundary` https://doc.rust-lang.org/src/core/str/mod.rs.html#211 so I think there's something there

mhammond (approved):

PR #5413, opened 2023-03-03
logins syncing roundtrips data it doesn't know about.

mhammond:
Needs many more tests and a db migration, but I think this will work.

### DIFF
--- a/components/logins/src/schema.rs
+++ b/components/logins/src/schema.rs
@@ -93,8 +93,9 @@ use lazy_static::lazy_static;
 use rusqlite::Connection;
 use sql_support::ConnExt;
 
-/// The current schema version is 1.  We reset it after the SQLCipher -> plaintext migration.
-const VERSION: i64 = 1;
+/// Version 1: SQLCipher -> plaintext migration.
+/// Version 2: addition of `loginsM.enc_unknown_fields`.
+pub(super) const VERSION: i64 = 2;
 
 /// Every column shared by both tables except for `id`
 ///
@@ -161,7 +162,10 @@ lazy_static! {
             -- Milliseconds (a sync15::ServerTimestamp multiplied by
             -- 1000 and truncated)
             server_modified INTEGER NOT NULL,
-            is_overridden   TINYINT NOT NULL DEFAULT 0
+            is_overridden   TINYINT NOT NULL DEFAULT 0,
+            -- fields on incoming records we don't know about and roundtrip.
+            -- a serde_json::Value::Object as an encrypted string.
+            enc_unknown_fields   TEXT
         )",
         common_sql = COMMON_SQL
     );
@@ -214,7 +218,7 @@ pub(crate) fn init(db: &Connection) -> Result<()> {
 
 // Allow the redundant Ok() here.  It will make more sense once we have an actual upgrade function.
 #[allow(clippy::unnecessary_wraps)]
-fn upgrade(_db: &Connection, from: i64) -> Result<()> {
+fn upgrade(db: &Connection, from: i64) -> Result<()> {
     log::debug!("Upgrading schema from {} to {}", from, VERSION);
     if from == VERSION {
         return Ok(());
@@ -224,7 +228,15 @@ fn upgrade(_db: &Connection, from: i64) -> Result<()> {
         "Upgrading from user_version = 0 should already be handled (in `init`)"
     );
 
-    // Schema upgrades that should happen after the sqlcipher -> plaintext migration go here
+    // Schema upgrades.
+    if from == 1 {
+        // Just one new nullable column makes this fairly easy
+        db.execute_batch("ALTER TABLE loginsM ADD enc_unknown_fields TEXT;")?;
+    }
+    // XXX - next migration, be sure to:
+    // from = 2;
+    // if from == 2 ...
+    db.execute_batch(&SET_VERSION_SQL)?;
     Ok(())
 }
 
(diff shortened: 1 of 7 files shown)


components/logins/src/sync/payload.rs
```diff
@@ -61,44 +113,19 @@ pub struct LoginPayload {
 
     #[serde(default)]
     pub times_used: i64,
+
+    // Additional "unknown" round-tripped fields
+    #[serde(flatten, skip_serializing_if = "serde_json::Map::is_empty")]
```
mhammond, line 118: This is quite magic 😮

components/logins/src/sync/engine.rs
```diff
@@ -504,6 +506,22 @@ impl SyncEngine for LoginsSyncEngine {
...
+// fields because they are, um, unknown :) So we never merge or inspect them, just send
+// whatever we have in outgoing.
+fn unknown_fields_from_row(
+    row: &rusqlite::Row<'_>,
+) -> Result<Option<super::payload::UnknownFields>> {
+    Ok(match row.get::<_, Option<String>>("unknown_fields")? {
+        None => None,
+        Some(s) => match serde_json::from_str(&s)? {
+            serde_json::Value::Object(m) => Some(m),
+            // ???? Should maybe log/report an error?
```
galich, line 519: Definitely should, otherwise how would we notice/troubleshoot data loss?

components/logins/src/sync/engine.rs
```diff
@@ -504,6 +506,22 @@ impl SyncEngine for LoginsSyncEngine {
...
+            _ => None,
+        },
+    })
+}
```
galich, line 523: nit: perhaps something like this can help to avoid nested matches?

```suggestion
match row.get::<_, Option<String>>("unknown_fields")?.map(serde_json::from_str) {
    serde_json::Value::Object(m) => Ok(m),
    _ => Ok(None),
},

```

components/logins/src/sync/engine.rs
```diff
@@ -504,6 +506,22 @@ impl SyncEngine for LoginsSyncEngine {
...
 }
 
+// This seems misplaced - but everywhere else a mirror row is read we don't need the unknown
+// fields because they are, um, unknown :) So we never merge or inspect them, just send
+// whatever we have in outgoing.
+fn unknown_fields_from_row(
+    row: &rusqlite::Row<'_>,
+) -> Result<Option<super::payload::UnknownFields>> {
+    Ok(match row.get::<_, Option<String>>("unknown_fields")? {
+        None => None,
```
bendk, line 516: Maybe this should be an error?  I think it indicates that you want to get the `unknown_fields`, but forgot to include it in your SELECT list.

components/logins/src/sync/engine.rs
bendk, line 519: I like reporting an error.  It seems like the data was corrupted in some way and now we're going to drop it.

components/logins/src/sync/engine.rs
```diff
@@ -504,6 +506,22 @@ impl SyncEngine for LoginsSyncEngine {
     }
 }
 
+// This seems misplaced - but everywhere else a mirror row is read we don't need the unknown
+// fields because they are, um, unknown :) So we never merge or inspect them, just send
+// whatever we have in outgoing.
+fn unknown_fields_from_row(
+    row: &rusqlite::Row<'_>,
+) -> Result<Option<super::payload::UnknownFields>> {
```
bendk, line 514: I think I like `Result<UnknownFields>` better -- both for this return type and the argument types for functions like `into_bso()`.   That way there's no ambiguity between `None` and an empty map.  To me, `UnknownFields::new()` communicates "there's no unknown fields" clearer than `None` (maybe we could define `UnknownFields::empty()` to be even more clear)

components/logins/src/sync/payload.rs
```diff
@@ -16,8 +16,60 @@ use serde_derive::*;
...
 
+pub(super) type UnknownFields = serde_json::Map<String, serde_json::Value>;
+
+/// What we get from the server after parsing the payload. We need to round-trip "unknown"
+/// fields, but don't want to carry them around in `EncryptedLogin`.
+#[derive(Debug)]
+pub(super) struct IncomingLogin {
+    pub login: EncryptedLogin,
+    pub unknown: UnknownFields,
+}
```
bendk, line 27: Maybe down the line this could be made generic/shared.  Like having a `sync15::IncomingRecord<T>` type that wraps a concrete sync record adds an `unknown_fields`.

components/logins/src/sync/payload.rs
bendk: This is definitely cool.  Does `flatten` apply to serializing as well as deserializing?  If so, is the `skip_serializing_if` needed?

galich (approved):
LGTM!

We need the same for credit cards & addresses, right?

bendk (commented):
Looks pretty great to me.  I'm glad that this wasn't too painful.  I left some minor comments, but overall it looks very nice.

components/logins/src/sync/engine.rs
skhamis: I think this is a little interesting -- should we always assume there is an `unknown_fields` but could be empty? I _slightly_ believe the *absence* of `unknown_fields` (None/null) is more clearer that if that field is null or undefined then we don't need to worry about it. This also keeps the "status quo" of today's world.

However I do understand the argument for an empty map. We'll probably just need to see a potential case where `None` doesn't work and see what ambiguity lies there.

skhamis (commented):
Awesome to have this to actually see the proposal in action!

components/logins/src/sync/payload.rs
mhammond: Good question :) I expect the difference might be the key missing vs existing as `{}`; the former seems preferable even if the same in practice? Needs tests to clarify though.

components/logins/src/sync/engine.rs
mhammond: Yeah - in this specific case, it's a reflection of a nullable sql column - it's flattened out later, and there's literally zero difference between an empty map and `None` - so I agree that option should die

components/logins/src/sync/engine.rs
mhammond: If the data was corrupted, it's more likely the `from_str()` would fail. I'm inclined to think that it would actually be OK to panic if we ended up with valid json that wasn't an object - we have a number of other places we panic in clearly impossible cases and no evidence of actually hitting them.

(On the fence still about the `from_str()` failing though - I mean, we don't really try and ignore other corruption scenarios. autofill stores json in a string in the db and I'm fairly sure it just has a normal error return if that fails)

components/logins/src/sync/engine.rs
mhammond: As discussed above though, in one scenario the None is expected, but the other needs to make lots of noise (and as mentioned, actually think a panic is OK)

components/logins/src/sync/engine.rs
galich: If one record is broken, we can not panic and stop processing all others. I'm sure you meant to return an error for broken record.

mhammond: This is almost a philosophical question :) We are talking about our local DB and sqlite is quite a loosely typed DB. To my mind, this problem would be analogous to, say, finding that a column our schema asserts is a number is actually a string, or a non-nullable column has a null - all things that could happen with corrupt databases and all things that would bring us down today. I'd agree we can't panic if this was about incoming data, but it's just us trusting both our implementation and our database. Some sync engines have explicit `assert()` statements in similar "impossible" cases, and I'm generally fine with that, and indeed they have saved us in the past (ie, the one case I can think of when an assert failed, we agreed that outcome was better than had the assert not existed and we charged on regardless in the face of a real bug)

mhammond: eg, https://www.sqlite.org/faq.html#q3

mhammond:
I think this is ready to roll. The fact the fields are now stored encrypted has changed a few things, in particular the equation around using an `Option<>` when no extra fields exist - it seems pointless to me to encrypt an empty map when (a) that going to be the vast majority of cases and (b) our DB needs to handle null columns anyway for a migration. Still open to everything though!

components/logins/src/schema.rs
```diff
@@ -224,7 +228,15 @@ fn upgrade(_db: &Connection, from: i64) -> Result<()> {
...
-    // Schema upgrades that should happen after the sqlcipher -> plaintext migration go here
+    // Schema upgrades.
+    if from == 1 {
+        // Just one new nullable column makes this fairly easy
+        db.execute_batch("ALTER TABLE loginsM ADD enc_unknown_fields TEXT;")?;
+    }
+    // XXX - next migration, be sure to:
+    // from = 2;
+    // if from == 2 ...
+    db.execute_batch(&SET_VERSION_SQL)?;
```
mhammond, line 239: we really should use our sql_support code for migrating etc, but I didn't want to block this on that especially while we still support sqlcipher migrations.

components/logins/src/sync/engine.rs
```diff
@@ -855,4 +862,114 @@ mod tests {
...
+            }),
+        );
+
+        let payload = get_outgoing_payload(&engine);
+
+        // The outgoing payload for our item should have the unknown fields.
+        assert_eq!(payload.get("unknown1").unwrap().as_str().unwrap(), "?");
+        assert_eq!(
+            payload
+                .get("unknown2")
```
mhammond, line 933: can probably simplify this with `!json({})`

bendk (approved):
This looks great to me.  The code feels very clean and the tests are there.

PR #5415, opened 2023-03-03
Update Kotlin to version 1.8.10 [ci full] [firefox-android: main]

rvandermeulen:
Also remove some unnecessary kotlin-stdlib usage as it's already implicit.


mhammond (approved):
Fenix recently updated to 1.8, so rs=me.

PR #5416, opened 2023-03-04
Various Android build updates

rvandermeulen:
Various commits to clean up the Dockerfile a bit and update some things.


bendk (approved):
LGTM.  I made https://github.com/mozilla/application-services/issues/5418 for the switftlint failure.  It also failed on the latest merge to `main`, so I don't think it has anything to do with this change.

PR #5423, opened 2023-03-07
Add Remote Settings client component

MatthewTighe:
I wanted to make sure the API was settled before adding in crate docs and a README. I'm also open to writing some more test cases, though I don't personally see a ton of value in doing so at the Rust layer with completely mocked responses for things like `get_records_since` or `download_attachment_to_path`. IMO if we want more complete (but mocked) paths tests there it is more interesting to do them at the application layer so that we are at least verifying the integration of the layers of the local stack. 

I'll seek input from the Android team on our part of the API once the Rust folks have had an opportunity for a once-over in case there's anything that immediately stands out as needing changes.

  - Note:


components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        };
+        let remote_settings = RemoteSettings::new(config).unwrap();
+
+        remote_settings
+            .download_attachment_to_path(
+                "d3a5eccc-f0ca-42c3-b0bb-c0d4408c21c9.jpg".to_string(),
+                "test.jpg".to_string(),
+            )
+            .unwrap();
+    }
```
MatthewTighe, line 134: My intent was to leave this in as demonstrative context until the API for this component is solidified

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+                {}
+            ]
+          }}"#,
+            jpg_attachment(),
+            pdf_attachment(),
+            no_attachment()
+        )
+    }
+
+    fn jpg_attachment() -> String {
```
tarikeshaq, line 157: does this and the following functions need to be functions? can we define `const` string slices? for example:

```rs
const JSG_ATTACHMENT: &str = r#"
          {
            "title": "jpg-attachment",
            "content": "content",
            "attachment": {
            "filename": "jgp-attachment.jpg",
            "location": "the-bucket/the-collection/d3a5eccc-f0ca-42c3-b0bb-c0d4408c21c9.jpg",
            "hash": "2cbd593f3fd5f1585f92265433a6696a863bc98726f03e7222135ff0d8e83543",
            "mimetype": "image/jpeg",
            "size": 1374325
            },
            "id": "c5dcd1da-7126-4abb-846b-ec85b0d4d0d7",
            "schema": 1677694447771,
            "last_modified": 1677694949407
          }
        "#

```

megazords/full/src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub use push;
 // TODO: Drop this dependency once android-components switches to using `rust_log_forwarder` for
 // log forwarding.
 pub use rc_log_ffi;
+pub use remote_settings;
```
tarikeshaq, line 22: Do we want to expose this to firefox-ios as well? I think that is best for a follow up PR, but the steps are in https://mozilla.github.io/application-services/book/howtos/adding-a-new-component.html#the-swift-bindings

components/remote_settings/src/config.rs
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! This module defines the custom configurations that consumers can set.
+//! Those configurations override default values and can be used to set a custom server url,
+//! collection name, and bucket name.
+//! The purpose of the configuration parameters are to allow consumers an easy debugging option,
+//! and the ability to be explicit about the server.
+
+/// Custom configuration for the client.
```
tarikeshaq, line 11: Love the docs 😍 !!!

tarikeshaq (changes requested):
This looks great @MatthewTighe 🚀 🎉 🚀  it's coming together!! I dropped comments as I was going through it - it's mostly questions and discussion so please do push back!

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    pub fn download_attachment_to_path(
+        &self,
+        attachment_location: String,
+        path: String,
+    ) -> Result<()> {
+        let client = self.client.lock().unwrap();
+        let resp = client.get_attachment(&attachment_location)?;
+        let mut file = File::create(path)?;
```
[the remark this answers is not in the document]
MatthewTighe: Is there something that is known to work? @mhammond and I talked about this a bit, but it seemed riskier to me to pass something like `Vec<u8>` across the FFI layer, since attachments can technically be of arbitrary size.

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+impl RemoteSettings {
+    pub fn new(config: RemoteSettingsConfig) -> Result<Self> {
+        Ok(RemoteSettings {
+            config: config.clone(),
+            client: Mutex::new(Client::new(config.into())?),
+        })
+    }
+
+    pub fn get_records(&self) -> Result<Vec<JsonObject>> {
```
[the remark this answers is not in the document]
MatthewTighe: Each collection in Remote Settings can define an arbitrary schema for their records, so if I'm understanding this correctly I don't think it would work. Each call to the records endpoint will return on array of (record) objects that matches the arbitrary schema. Here's another example:

```
{
  "data": [
    {
      "field1": "a",
      "another-field": true,
      "string-type-2": "a",
      "int-type": 100,
      "schema": 1678231127074,
      "id": "c92e34b3-40e1-4989-86bb-c7f48eacab31",
      "last_modified": 1678231137883
    }
  ]
}
```

components/remote_settings/src/lib.rs
MatthewTighe: Definitely, I had just been following the convention in my last PR. Thanks for the suggestion!

megazords/full/src/lib.rs
MatthewTighe: Ideally, yes. I am happy to do it here or as a follow-up. Would it be worth seeking input on the API from the iOS team as well?

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+pub use error::{RemoteSettingsError, Result};
+mod config;
+use rs_client::{Client, ClientConfig};
+use std::{fs::File, io::prelude::Write, sync::Mutex};
+
+uniffi_macros::include_scaffolding!("remotesettings");
+
+pub struct RemoteSettings {
+    pub config: RemoteSettingsConfig,
+    client: Mutex<Client>,
```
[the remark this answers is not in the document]
MatthewTighe: A comment probably would've been a good thing to include about it 😅 good sleuthing! 

I like shifting it down to the layer where it is most directly, but I am wondering if there's any case where other that's used `rs_client` would not want that. I don't really know enough to speculate though

components/remote_settings/src/error.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum RemoteSettingsError {
+    #[error("Network failure")]
+    NetworkFailure,
+    #[error("Error with HTTP client: {0}")]
+    ClientError(#[from] rs_client::ClientError),
+    #[error("JSON Error: {0}")]
+    JSONError(#[from] serde_json::Error),
+    #[error("Unexpected JSON format")]
+    InvalidJSONError,
```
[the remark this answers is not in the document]
MatthewTighe: This is a side-effect of the potential error from deserializing with`serde_json::from_str`, but different errors when actually parsing as shown below. 

```
        let value: serde_json::Value = serde_json::from_str(&json_text)?;
        let data = value
            .get("data")
            .ok_or(RemoteSettingsError::InvalidJSONError)?;
        let vec = data
            .as_array()
            .ok_or(RemoteSettingsError::InvalidJSONError)?;
```

TBH, I would love to know a better way to do this kind of error comprehension in Rust, but it seems like each `Result` to be assigned separately when using the ? operator. For example, I would have written this in Kotlin more along the lines of:

```
fun parse_records(text: String): Result<List<JsonObject>> {
    Result.runCatching {
        // one liner to deserialize and parse because intermediate assignments aren't required
    }.getOrDefault(<custom error>)
}
```

components/remote_settings/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+/Cargo.lock
```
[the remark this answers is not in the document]
MatthewTighe: I'll double-check and remove if they aren't required! I spliced some of the old project infrastructure from an old commit and things didn't go as cleanly as I'd have liked

components/remote_settings/src/error.rs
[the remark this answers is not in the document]
tarikeshaq: Aaah I see!

> seems like each Result to be assigned separately when using the ? operator.

Since it's all about the `From` implementation, we can have multiple errors map to our same variant! each type implements `From` itself, so if we are already returning a `RemoteSettingsError::InvalidJSONError`, it'll just map to the same variant, but if we have another error we need an explicit `From` implementation. 

`thiserror` gives us that `From` implementation for free when we do something like `JSONError(#[from] serde_json::Error)` (note the `#[from]`)

 if we do the suggestion in https://github.com/mozilla/application-services/pull/5423#discussion_r1128699873 we might get away with one error (i.e we'd remove the need to define our own `InvalidJSONError`


But if we don't, as long as `From<serde_json::Error>` returns the same variant as what our manual parsing returns, we should be in the clear!

we can do that by either:
- running `serde_json::from_str(&json_text).map_err(|_| RemoteSettingsError::InvalidJSONError)?` and removing the variant `JSONError` - that however means we lost context on the error string
- Changing `JSONError` to `JSONError(String)` and implementing our own, non-thiserror  `impl From<serde_json::Error> for RemoteSettingsError` that saves just the error string and not the error itself, then removing `InvalidJSONError` and replacing it with `JSONError(String)`

components/remote_settings/src/lib.rs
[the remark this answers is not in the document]
tarikeshaq: This is probably fine, we definitely know we can open already existing files (we do that for our DBs for example) - so if we can ensure that the consumer has already created the file, writing to it should work!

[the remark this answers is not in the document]
mhammond: We create our DBs too. I thought the important thing was that they tell us the location?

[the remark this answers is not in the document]
tarikeshaq: Eeeh nvm mind me, I'm confusing the opening/creating apis with some more specific file metadata apis that I've seen not work properly in some platforms

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(RemoteSettings {
+            config: config.clone(),
+            client: Mutex::new(Client::new(config.into())?),
+        })
+    }
+
+    pub fn get_records(&self) -> Result<Vec<JsonObject>> {
+        let client = self.client.lock().unwrap();
+        let resp = client.get_records()?;
+        self.parse_records(resp.text().to_string())
```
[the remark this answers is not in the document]
MatthewTighe: Wow yeah, that is very slick. To double-check my understanding, basically, the compiler is able to propagate the `serde_json::Value` type through the generic chain in `RemoteSettingsResponse` and the placeholder in `Ok(resp.json::<RemoteSettingsResponse<_>>()?.data)` will force the compiler to reify that generic during compile time?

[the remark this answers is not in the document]
tarikeshaq: Yup exactly!!

the `resp.json` function takes a type parameter in 
https://github.com/mozilla/application-services/blob/9657aebb7450c5b58e8b9a88bec12bd5e9e0f700/components/viaduct/src/lib.rs#L246-L252

We specify that type parameter when we do `::<...>` (this is called a t[urbofish](https://github.com/jplatte/turbo.fish)!)

The type we provide is `RemoteSettingsResponse<T>`, which itself takes a type parameter. We do `RemoteSettingsResponse<_>` to let the compiler infer what `T` is, it can do that because the return type of `get_records` includes the `JsonOnbject`, so the compiler infers that we wanted `RemoteSettingsResponse<JsonObject>`.


I wrote that not knowing whether we'd expect `data` to have concrete types, buf if we expect `data` to always be an arbitrary schema, it might be worth removing the generic in `RemoteSettingsResponse` so it's always holding on to a `data: Vec<JsonObject>` which should function the same way!

components/support/rs_client/src/http_client.rs
```diff
@@ -76,14 +92,15 @@ impl Client {
...
         if let RemoteState::Backoff {
             observed_at,
             duration,
-        } = self.remote_state.get()
+        } = current_state
         {
             let elapsed_time = observed_at.elapsed();
             if elapsed_time >= duration {
-                self.remote_state.replace(RemoteState::Ok);
+                *self.remote_state.lock().unwrap() = RemoteState::Ok;
```
tarikeshaq, line 103: If we do the above comment, we should just assign the value to `*current_state` instead of relocking the mutex

components/support/rs_client/src/http_client.rs
```diff
@@ -4,11 +4,11 @@
 
 //! This is a simple HTTP client that uses viaduct to retrieve Remote Settings data from the server.
 
+use std::sync::Mutex;
```
tarikeshaq, line 7: Aaah should we use `parking_lot::Mutex` here instead of `Mutex`?

At some point in the past we switched to using `parking_lot` because the `Mutex`s there don't get poisoned, I imagine we should probably do the same here

components/support/rs_client/src/http_client.rs
```diff
@@ -76,14 +92,15 @@ impl Client {
     }
 
     fn ensure_no_backoff(&self) -> Result<()> {
+        let current_state = self.remote_state.lock().unwrap().clone();
```
tarikeshaq, line 95: Can we get away without the clone here? if we remove the clone then simply deref `current_state` (by using `*current_state`) we might be able to avoid the clone!

components/support/rs_client/src/http_client.rs
tarikeshaq: Since we're already going through the overhead of the lock in `ensure_no_backoff`, it might be worth it to only lock once at the beginning of the function!

tarikeshaq (changes requested):
This looks great!!! My main `requesting changes` is to use `parking_lot` for the Mutex

re: the API design, I don't see anything wrong, but of course, I have a limited understanding of how this will be used so I'd defer to consumers of the API (i.e mobile!)

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+}
+
+#[derive(Deserialize)]
+struct RemoteSettingsResponse<T> {
+    data: Vec<T>,
+}
+
+#[derive(Deserialize)]
```
mhammond, line 65: Why the custom `JsonObject` instead of using `serde_json::Value` directly? IOW, something like `use serde_json::Value as JsonObject;` at the top of the file will work fine with a few other adjustments I've I've noted:

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+pub struct JsonObject(pub serde_json::Value);
+
+impl UniffiCustomTypeConverter for JsonObject {
+    type Builtin = String;
+
+    fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+        Ok(JsonObject(serde_json::from_str(&val)?))
```
mhammond, line 72: This just becomes `Ok(serde_json::from_str(&val)?)`

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    fn from_custom(obj: Self) -> Self::Builtin {
+        serde_json::to_string_pretty(&obj.0).unwrap()
```
mhammond, line 76: seems mildly wasteful for this to be pretty, especially given the Kotlin code never actually sees the string (ie, it doesn't even seem able to be inspected anywhere for debugging)

Re the type stuff above, the `.0` here (and below in the test get removed)

components/remote_settings/src/lib.rs
mhammond: Depending on how often you might actually want to run it against a test server, you could put this behind a "feature", so running it would just mean a command-line like `cargo test --features=[with-local-test-server]` or something - not at all bothered by that though - could be done as a followup or never done at all :)

components/remote_settings/src/remotesettings.udl
```diff
@@ -0,0 +1,33 @@
...
+
+dictionary RemoteSettingsConfig {
+    string? server_url;
+    string? bucket_name;
+    string collection_name;
+};
+
+[Error]
+enum RemoteSettingsError {
+    "NetworkFailure",
```
mhammond, line 18: nano-nit - one being `Failure` and the rest being `Error` seems odd?

components/remote_settings/src/remotesettings.udl
```diff
@@ -0,0 +1,33 @@
...
+    "JSONError",
+    "FileError",
+};
+
+interface RemoteSettings {
+    [Throws=RemoteSettingsError]
+    constructor(RemoteSettingsConfig remote_settings_config);
+
+    [Throws=RemoteSettingsError]
+    sequence<JsonObject> get_records();
```
mhammond, line 29: This maybe relates to what Tarik mentioned above - while the records themselves are schemaless, it seems possible there will be metadata you care about - eg, a better return type might be a struct with, say, `last_modified: u64, records: Vec<JsonObject>` or similar. Then you are one step closer to being able to trivially add `get_records_since(u64)` and also have a place to add extra metadata (eg, etag or something else from the server communicated via headers)

components/support/rs_client/src/http_client.rs
mhammond: Yeah, I think so - they roughly work the same, just with less explicit `unwrap`s!

components/support/rs_client/src/http_client.rs
```diff
@@ -59,14 +59,30 @@ impl Client {
...
+    pub fn get_attachment(&self, attachment_location: &str) -> Result<Response> {
+        let path = format!(
+            "attachments/{}/{}/{}",
+            &self.bucket_name, &self.collection_name, attachment_location
+        );
+        self.make_request(path)
     }
 
-    fn make_request(&self, request: Request) -> Result<Response> {
+    pub fn get_records_since(&self, last_modified: i64) -> Result<Response> {
```
mhammond, line 73: ah, I see you have it here but not in the UDL - my comment above is about how you might communicate the timestamp on the first call.

components/remote_settings/src/remotesettings.udl
```diff
@@ -0,0 +1,33 @@
...
+dictionary RemoteSettingsConfig {
+    string? server_url;
+    string? bucket_name;
+    string collection_name;
+};
+
+[Error]
+enum RemoteSettingsError {
+    "NetworkFailure",
+    "ClientError",
```
mhammond, line 19: I *think* that all of these end up as enums with strings? If so, communicating the sub-variants might be tricky - I'm thinking specifically of `ClientError::Backoff(duration)` and whether the Kotlin caller will be able to sanely handle that and extract the duration. I don't think this needs to block though because I suspect a first version will not care, but might be worth opening an issue?

(This certainly would not be the first crate with this kind of issue though!)

components/support/rs_client/src/http_client.rs
mhammond: yeah - I agree! Alternatively, can the caller lock it and pass it as a mutable ref? I imagine the caller is going to lock it for itself after this succeeds anyway.

mhammond (commented):
This is looking in great shape, nice work!

components/remote_settings/src/lib.rs
MatthewTighe: That's a good point, I don't think my brain had caught up from thinking about the result being a list of strings

components/remote_settings/src/remotesettings.udl
MatthewTighe: And it wasn't even used! Think it was leftover from when I was stubbing everything out

components/remote_settings/src/remotesettings.udl
MatthewTighe: AFAICT Fenix was not doing anything with the `NimbusError::Backoff` before it was moved into `rs_client::ClientError::Backoff`, so I don't suspect that this will be required for the first version either

components/remote_settings/src/lib.rs
tarikeshaq: heh our ci does run tests for all features - if we'd like to have an example of how to use this remote-settings client against a test server, we can probably make an example like the [nimbus one](https://github.com/mozilla/application-services/tree/main/components/nimbus/examples) or the [push one](https://github.com/mozilla/application-services/blob/main/examples/push-livetest/src/livetest.rs)

But I agree with Mark, we can probably do that as a follow-up ticket!

components/support/rs_client/src/http_client.rs
MatthewTighe: Oh thanks for the reminder, this should be added now!

components/remote_settings/src/remotesettings.udl
MatthewTighe: Latest push has some more significant changes to the data model - much happier with where it is now so thanks for the suggestions. I _think_ I handled the fact that arbitrary JSON can be included with some of these fields like `last_modified` and and `id`, but let me know if you think there are further improvements that can be made

MatthewTighe:
Addressed the second round of feedback. Thanks again all for the patience and thoroughness on these, I'm learning a lot!

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+pub use error::{RemoteSettingsError, Result};
+mod config;
+use rs_client::{Client, ClientConfig};
+use serde::Deserialize;
+use std::{collections::HashMap, fs::File, io::prelude::Write};
+
+uniffi_macros::include_scaffolding!("remotesettings");
+
+pub struct RemoteSettings {
+    pub config: RemoteSettingsConfig,
```
tarikeshaq, line 16: nit: does this need to be `pub`?

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    pub fn get_records(&self) -> Result<Vec<Record>> {
+        let resp = self.client.get_records()?;
+        Ok(resp.json::<RemoteSettingsResponse>()?.data)
+    }
+
+    pub fn get_records_since(&self, last_modified: u64) -> Result<Vec<Record>> {
+        let resp = self.client.get_records_since(last_modified)?;
+        Ok(resp.json::<RemoteSettingsResponse>()?.data)
```
tarikeshaq, line 35: ❤️ ❤️

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+struct RemoteSettingsResponse {
+    data: Vec<Record>,
+}
+
+#[derive(Deserialize, PartialEq)]
+pub struct Record {
+    id: String,
+    last_modified: u64,
+    attachment: Option<Attachment>,
+    #[serde(flatten)]
```
tarikeshaq, line 70: We love `flatten`!!!

tarikeshaq (approved):
Coming around to remove the "Request for changes" - this looks great to me!! I'm happy with this Rust-wise, but I imagine you're still waiting on android approvals

components/remote_settings/src/lib.rs
MatthewTighe: I had imagined that it might be useful to be able to inspect the configuration at the application layer after construction, but I'm open to making it private if that isn't typical

MatthewTighe:
Thanks for revisiting!

> but I imagine you're still waiting on android approvals

I am actually going to wait on @mhammond's next cycle, since his last feedback resulted in some fairly significant (and good!) updates to the API surface. I want to make sure we have that nailed down before I get more Android folks involved 🙂

components/remote_settings/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        Ok(RemoteSettings {
+            config: config.clone(),
+            client: Client::new(config.into())?,
+        })
+    }
+
+    pub fn get_records(&self) -> Result<RemoteSettingsResponse> {
+        let resp = self.client.get_records()?;
+        let records = resp.json::<RemoteSettingsRecordsResponse>()?.data;
+        let last_modified = resp.headers.get_as::<u64, _>("etag").unwrap()?;
```
mhammond, line 31: You can drop the explicit types (ie, the `::<u64, _>`) here and below as the compiler can work it out given it's being put into a struct member with the same type.

It's probably worth a comment here too about the unwrap so casual readers know it is intentional - something like "If the request succeeded, kinto documents the etag header exists and is the timestamp - we assume that's true via the unwrap" or something?

mhammond (approved):
Fantastic!

components/remote_settings/src/lib.rs
MatthewTighe: Cool, will do! I am intending to do a more thorough docs pass and clean up things like the unused test now that the API is looking good from the Rust perspectiv

jonalmeida (approved):
I only reviewed the RemoteSettingsTest since you were looking for an android engineer's review - lgtm!

MatthewTighe:
Drafting this for a bit while I work through UniFFI errors related to moving the `RemoteSettingsConfig` into the shared component

components/remote_settings/src/remotesettings.udl
```diff
@@ -0,0 +1,53 @@
...
+
+[Error]
+enum RemoteSettingsError {
+    "ClientError",
+    "JSONError",
+    "FileError",
+    "ParseIntError",
+};
+
+interface RemoteSettings {
```
linabutler, line 41: thought: Should other Rust crates depend on this `remote_settings` crate, or use `support/rs_client`?

In my [Suggest Offline prototyping branch](https://github.com/mozilla/application-services/commit/8902fa73227b744a63ade0848adf8c3e710ea708), I added a skeleton for a new Suggest component that—for now—just pulls down the `quicksuggest` collection and collects some stats on it.

At first, I tried depending on this `remote_settings` crate, but couldn't construct a `RemoteSettings` object, because the `RemoteSettingsConfig` struct (I see you've removed that and made it an extern in your latest push 👀) wasn't public. Changing it to `pub use config::RemoteSettingsConfig` fixed that, but then—after a quick fix to resolve attachment URLs; I left a separate comment on `get_attachment` about that—I was curious how to get at the attachment without downloading it to disk.

So I switched to `support/rs_client` instead. That gave me the interface I wanted, to access the attachment's contents in memory, but then I needed to duplicate the `RemoteSettingsResponse`, `Record`, and `Attachment` definitions in my crate, because they're defined in this crate, not in `rs_client`. If you're 👀 at my branch, the dupes are called `SuggestRecordsResponse`, `SuggestRecord`, and `SuggestAttachment` there 😄

I guess I really have three related questions, about how you're envisioning the split between `remote_settings` and `rs_client`: 😅 

* Which crate should my crate—and future crates, if we add more a-s components that pull data from Remote Settings—depend on?
* If `rs_client` is better, how would you feel about lifting the definitions of `RemoteSettingsResponse`, `Record`, and `Attachment` into that crate, and making them externs in the UDL, too?
* If `remote_settings` is better, how would you feel about adding a new method to download an attachment without writing it to a file?

(My gut feel is that adding the types to the lower-level `support/rs_client` feels like a better option, especially since it'll take a dependency on Serde to parse the capabilities response, anyway, and it looks like Nimbus also uses it—but I don't have the full picture that you do about the split between the two crates!)

linabutler (commented):
I used your new client to do some prototyping with the `quicksuggest` collection, and it worked wonderfully! 🎊  Just a couple of things I noticed while trying it out.

components/remote_settings/src/remotesettings.udl
MatthewTighe: 👋 Hi @linabutler, thank you for the feedback! Wanted to respond to some of this now so you didn't think it had been left in complete limbo. I am still slowly working through some iOS build issues with the `extern` UDL type, and unfortunately this has taken a slight backseat to other work, especially while mhammond is on PTO for the next bit. If this slowness is impacting you, please let me know. I can work with my team to try and reprioritize a bit. 

Your intuition about `rs_client` being the crate that other crates ingest is correct, and so I agree that it would be best to move some more of those foundational types into it. In retrospect I probably should've started with that, but I had aimed for the smallest lift when moving the original client out of Nimbus, which expected pure JSON results. 

I do think those updates can be quick, once I get a working example of one extern type. These build issues have been plaguing me for a bit since I don't have a strong understanding of megazording and UniFFI generation steps 😅

tarikeshaq: >These build issues have been plaguing me for a bit since I don't have a strong understanding of megazording and UniFFI generation steps

Hey @MatthewTighe 👋 !! I just wanted to volunteer to help with any build issues you're facing if you're looking to rubber duck with someone from application-services in case there's context we could help provide once this work gets prioritized again

MatthewTighe: 🙏 thank you @tarikeshaq, that would be phenomenal! I'm never sure exactly the right place to go for app services questions/help so I appreciate the volunteerism. I'm hoping to get back to this starting next week if you have time then

components/support/rs_client/src/rs_client.udl
```diff
@@ -0,0 +1,34 @@
...
+    RsJsonObject fields;
+};
+
+dictionary Attachment {
+    string filename;
+    string mimetype;
+    string location;
+    string hash;
+    u64 size;
+};
```
MatthewTighe, line 34: Should I write comments here for these types? Will they propagate through the `extern` references in the top level components?

MatthewTighe (commented):
Alright, I think this is ready for another round of reviews. I incorporated @linabutler's suggestion of moving the RS primitives to rs_client. 

This also includes changes to dynamically look up the attachments base URL for the server. I switched that to be done on client initialization, but if we think it was more appropriate to do it the first time an attachment was downloaded I am happy to switch it. I just wanted to avoid dealing with mutexes and pointers 😅

components/support/rs_client/src/http_client.rs
```diff
@@ -37,53 +95,111 @@ pub(crate) enum RemoteState {
...
+    /// During construction, the [Client] will fetch the base attachments
+    /// URL from the server's metadata, assuming attachments are enabled for
+    /// the server.
+    pub fn new(config: RemoteSettingsConfig) -> Result<Self> {
         let server_url = config
             .server_url
             .unwrap_or_else(|| String::from("https://firefox.settings.services.mozilla.com"));
         let bucket_name = config.bucket_name.unwrap_or_else(|| String::from("main"));
+        let base_url = Url::parse(&server_url)?;
+        let lookup = Self::lookup_attachments_base_url(base_url.clone());
```
linabutler, line 107: I definitely hear you on sidestepping the complexity of mutexes and optionals to do this the first time an attachment is downloaded!

There are a couple of reasons why I'm nervous about looking up the base URL at client creation time, though:

* `Client::new` now hides a blocking network call that can take an unknown amount of time to complete. In general, I try to avoid constructors doing I/O, or having side effects beyond throwing / returning errors for invalid parameters. Effectful constructors can make testing harder, and require consumers to remember to call the constructor off the main thread. The complexity is still there, but it's pushed to consumers.
* The network request can take an unknown amount of time, blocking whatever thread creates the `Client` until it returns. Consumers will need to remember to call the constructor off the main thread
* What happens if the request fails—maybe the server had a hiccup, or the user is on a spotty network? If we look up the base URL lazily, when an attachment is downloaded, we can bubble that error up, and try again on the next `get_attachment` call.
* Every consumer of `Client` has to wait for the network call, even if they never call `get_attachment`.

linabutler, line 107: I definitely hear you on sidestepping the complexity of mutexes and optionals to do this the first time an attachment is downloaded! 😅

There are a few reasons why I'm nervous about looking up the base URL at client creation time, though:

* `Client::new` now hides a side effect. It's definitely a personal preference, but I usually try to avoid I/O in constructors. From past experience, effectful constructors can make testing harder, and, even when called out (like you've done here!), can be surprising for callers.
* The network request can take an unknown amount of time, blocking the thread that creates the `Client` until it completes. Consumers of the client will need to remember to construct it lazily, and off the main thread. The complexity with mutexes / optionals / `OnceCell`s will still be there, but pushed to the consumers.
* If the request fails—maybe the server had a hiccup, or the user is on a spotty network—we'll fail to create the client.
* Every consumer of `Client` will need to wait for the network call, even if they're not interested in attachments.

It's more work for us to look up the base URL lazily, the first time `get_attachment` is called, but it would mean that:

* Consumers can call `Client::new` on the main / any thread, and not worry about it blocking.
* `Client::new` errors are much more deterministic: the only reason it'll fail is if the caller gives it a bad server URL, not a potentially transient network failure.
* Consumers that never call `get_attachment` won't make the extra network call at all!

WDYT? Would you be open to discussing more, or changing the behavior to be lazy?

components/remote_settings/src/remotesettings.udl
linabutler: Thanks so much, @MatthewTighe! I've been meaning to reply to your comment for a while (sorry!), and saw you just pushed your latest changes, so now's a perfect time! 😁

Definitely no immediate rush on getting this in. We're very keen to take advantage of your wonderful work for Cross-Platform Suggest, and I know the Search Consolidation project is interested also—but they're still in the planning and discussion phases, and I think it'll be at least a week or two more before we're ready for engineering beyond prototyping.

We'd be totally open to helping you get it over the line if other work takes priority, and I also volunteer to help with build issues, reviews, pairing, and anything you need for this! 😊

> I had aimed for the smallest lift when moving the original client out of Nimbus, which expected pure JSON results.

That makes perfect sense, thanks for the additional context! I'll do a proper review pass now, but I wanted to make sure not to forget to reply to your note.

.buildconfig-android.yml
```diff
@@ -43,6 +43,20 @@ projects:
...
+    - name: rsclient
+      type: aar
+    description: A Remote Settings client
+  remotesettings:
+    path: components/remote_settings/android
+    artifactId: remotesettings
+    publications:
+    - name: remotesettings
+      type: aar
+    description: A Remote Settings client
```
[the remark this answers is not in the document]
linabutler: Another idea: keep `remotesettings` UniFFIed, but don't UniFFI `rs_client`.

It'll be more boilerplate-y for sure, but I think you could still make that work, by having the UniFFIed `remotesettings` types implement `From<rs_client::{Type}>`. So instead of using externs, `remotesettings::Attachment` (for example) could `impl From<rs_client::Attachment> for remotesettings::Attachment`, and so on.

Then we can explain the split as: "`rs_client` is the low-level Remote Settings HTTP client for other Rust code only, and `remotesettings` is the high-level API for other consumers."

But I'm not sure how I feel about doing it this way 👀 WDYT?

components/support/rs_client/src/http_client.rs
```diff
@@ -22,7 +23,64 @@ pub struct Client {
...
+// This shouldn't really impact Android, since the type is converted into the platform
+// JsonObject thanks to the UniFFI binding.
+pub type RsJsonObject = serde_json::Map<String, serde_json::Value>;
+impl UniffiCustomTypeConverter for RsJsonObject {
+    type Builtin = String;
+    fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+        let json: serde_json::Value = serde_json::from_str(&val)?;
+
+        match json.as_object() {
+            Some(obj) => Ok(obj.clone()),
```
linabutler, line 74:
```suggestion
        match json {
            serde_json::Value::Object(obj) => Ok(obj),
```

components/support/rs_client/src/http_client.rs
```diff
@@ -22,7 +23,64 @@ pub struct Client {
...
+    pub last_modified: u64,
+}
+
+#[derive(Deserialize)]
+struct RemoteSettingsRecordsResponse {
+    data: Vec<Record>,
+}
+
+/// A parsed Remote Settings record. Records can contain arbitrary fields, so clients
+/// are required to further extract expected values from the [fields] member.
```
linabutler, line 43: The way you're exposing these additional fields to Kotlin with `RsJsonObject` is really slick! 🥇

components/support/rs_client/src/http_client.rs
```diff
@@ -107,13 +227,37 @@ impl Client {
...
+        let server_info = req.send()?.json::<RemoteSettingsServerInfo>()?;
+        match server_info.capabilities.attachments {
+            Some(capability) => Ok(Url::parse(&capability.base_url)?),
+            None => Err(ClientError::AttachmentsUnsupportedError),
+        }
+    }
+}
+
+#[derive(Deserialize)]
+struct RemoteSettingsServerInfo {
```
linabutler, line 249:
```suggestion
struct ServerInfo {
```

(Since we're already in the `rs_client` crate, let's drop the `RemoteSettings` qualifier! 😊)

components/support/rs_client/src/http_client.rs
```diff
@@ -107,13 +227,37 @@ impl Client {
...
+    capabilities: RemoteSettingsCapabilities,
+}
+
+#[derive(Deserialize)]
+struct RemoteSettingsCapabilities {
```
linabutler, line 254:
```suggestion
struct Capabilities {
```

components/support/rs_client/src/http_client.rs
```diff
@@ -107,13 +227,37 @@ impl Client {
...
+    attachments: Option<RemoteSettingsAttachmentsCapability>,
+}
+
+#[derive(Deserialize)]
+struct RemoteSettingsAttachmentsCapability {
```
linabutler, line 259:
```suggestion
struct AttachmentsCapability {
```

components/support/rs_client/src/http_client.rs
```diff
@@ -194,41 +417,122 @@ mod test {
...
-        });
+        };
         assert!(http_client.get_records().is_ok());
         m.expect(1).assert();
     }
 
-    fn response_body() -> String {
-        r#"
-        {{ "data": [
+    fn attachment_metadata(base_url: String) -> String {
```
linabutler, line 446: Great tests, as always!

linabutler (changes requested):
Thank you so much @MatthewTighe, this is tremendous work! I did call out a few important things, so I wouldn't mind seeing this again, and discussing some of those points with you, but it looks like it's in great shape overall!

components/support/rs_client/src/http_client.rs
MatthewTighe: You're totally right, I (ironically) was just being lazy. I agree that the the lazy implementation is the more correct way to handle this. I'll update accordingly. I do really appreciate the detailed write-up!

components/nimbus/src/client/mod.rs
```diff
@@ -23,18 +22,19 @@ pub(crate) fn create_client(
         Some(config) => {
             // XXX - double-parsing the URL here if it's not a file:// URL - ideally
             // config would already be holding a Url and we wouldn't parse here at all.
-            let url = Url::parse(&config.server_url)?;
+            let server_url = config.server_url.clone().unwrap();
+            let url = Url::parse(&server_url)?;
```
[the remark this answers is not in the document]
MatthewTighe: Oh neat! Didn't realize we could early return from match expressions

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
```diff
@@ -0,0 +1,149 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.appservices.remotesettings
```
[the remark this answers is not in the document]
MatthewTighe: It was frustrating to keep track of while dealing with tooling, so I think this nit has merit anyway. I only did it because I had been under the impression that neither underscores or hyphens were valid characters in package names. This comment made me look into it though, it turns out that [Oracle suggests using underscores and that hyphens are illegal characters](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html). Without more research, I would guess the hyphens in your example are legal because that is a reference to a plugin and not a package.

components/support/rs_client/src/http_client.rs
```diff
@@ -22,7 +23,64 @@ pub struct Client {
...
+
+/// A parsed Remote Settings record. Records can contain arbitrary fields, so clients
+/// are required to further extract expected values from the [fields] member.
+#[derive(Deserialize, PartialEq)]
+pub struct Record {
+    pub id: String,
+    pub last_modified: u64,
+    pub attachment: Option<Attachment>,
+    #[serde(flatten)]
+    pub fields: serde_json::Map<String, serde_json::Value>,
```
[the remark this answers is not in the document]
MatthewTighe: Do you mean something like `type Fields: serde_json::Map<RsJsonObject, serde_json::Value>`?

components/support/rs_client/src/http_client.rs
MatthewTighe: Removing this from the private types, but leaving them in the public API for now, assuming that was the intent of this comment.

components/remote_settings/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[derive(Debug, thiserror::Error)]
+pub enum RemoteSettingsError {
+    #[error("Error with HTTP client: {0}")]
+    ClientError(#[from] rs_client::ClientError),
+    #[error("JSON Error: {0}")]
+    JSONError(#[from] serde_json::Error),
+    #[error("Error writing downloaded attachment: {0}")]
```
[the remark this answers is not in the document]
MatthewTighe: I dig the principle here, but am having trouble conceptualizing it as code. Could you point me to an example?

components/support/rs_client/src/http_client.rs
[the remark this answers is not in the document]
mhammond: I just meant `pub fields: RsJsonObject`, which better matches the udl of `RsJsonObject fields`

components/remote_settings/src/error.rs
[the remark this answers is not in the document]
mhammond: I really don't think you should go this far in this PR (unless you are a masochist! ;) but an example is logins:

Here's the error returned to consumers: https://github.com/mozilla/application-services/blob/main/components/logins/src/error.rs#L17
The errors used internally (which in many ways is just a unions of errors from all the other crates we depend on): https://github.com/mozilla/application-services/blob/main/components/logins/src/error.rs#L41
and the thing that ties it together, including policy of sentry, logging etc: https://github.com/mozilla/application-services/blob/main/components/logins/src/error.rs#L116

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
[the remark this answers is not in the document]
MatthewTighe: I did try this, but am now running into Android build issues. I'm not sure they're related yet, because reverting didn't seem to fix them though I may have missed a step in the revert. FWIW, the issue is:

`A problem occurred configuring project ':nimbus-gradle-plugin'.`

Stacktrace top is:
`Cannot get property 'build' on extra properties extension as it does not exist`

tarikeshaq (commented):
I think you've gotten two quality reviews with Mark and Lina so I'll take myself off the review queue - excited to see this land!

components/support/rs_client/src/http_client.rs
linabutler: Ahhh, sorry, nope, I meant literally renaming `RemoteSettingsServerInfo` to `ServerInfo` (and doing the same for the other `RemoteSettings`-prefixed types)!

Since we're already in a crate that deals with "Remote Settings things", I wasn't sure that the `RemoteSettings` prefix was adding a whole lot of value—and naming it `ServerInfo`, etc. would be in line with `Attachment` and `Record` 😅

.buildconfig-android.yml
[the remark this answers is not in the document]
linabutler: >> Another idea: keep `remotesettings` UniFFIed, but don't UniFFI `rs_client`.
>
> I'm not sure I agree with that exactly - in other places we are ending with with crates being uniffi'd just for their types - eg, I just landed a sync15.udl with just types because ending up with duplicate struct definitions across different crates quickly became annoying.

Oooh, thanks, this is helpful to know! To be honest, I was already feeling pretty lukewarm about suggesting this idea already, but now I think I'm also against it—for all the reasons you mentioned, for the increase in boilerplate, and also because consolidating into a single crate / component is a better way forward in every way! 😄

> That said, I'm wondering if we would see more benefit from consolidating into a single crate/component. It seems like it could continue to be confusing to future consumers at the Rust layer whether they should use `rs_client` or `remote_settings`, and I don't immediately see any reason that methods exposed to the application layer couldn't also be exposed to the Rust layer.

Yeah, I agree, too! We can have one crate and one UDL file that exposes the types (that are currently in the `rs_client` crate) and the `RemoteSettings` interface (that's currently in the `remotesettings` crate).

As for `rs_client::Client`, I think we can either have it in the single crate but not UniFFIed (so only other Rust code can use it), or expose it in UDL as a lower-level `Client` or `HttpClient` interface.

WDYT @MatthewTighe @mhammond?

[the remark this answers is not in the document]
linabutler: > WDYT @MatthewTighe @mhammond?

To make the "what" a little more concrete, I'm suggesting that we:

1. Choose which crate we want to use: `rs_client` or `remotesettings`.
2. Move everything from the crate that we didn't choose to the one we chose. If we chose `remotesettings`, we'd move `RemoteSettingsConfig`, `RemoteSettingsResponse`, `Record`, `Attachment`, `Client`, etc. out of `rs_client` and into `remotesettings`. If we chose `rsclient`, we'd move `RemoteSettings`, `RemoteSettingsError`, etc. out of `remotesettings` and into `rs_client`.
3. Consolidate the contents of `remotesettings.udl` and `rs_client.udl` in the crate we chose.
4. (Optional: decide if we want to expose `Client` in UDL. If not, we can still keep it a `pub` type that's exposed to Rust code, but not foreign code).
5. Delete the crate that we didn't choose.

How does that sound? Are there reasons we wouldn't want to do this, and instead keep the two-crate `rs_client` / `remotesettings` setup we have now?

components/remote_settings/src/error.rs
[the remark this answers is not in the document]
linabutler: I also think this can wait until a follow-up! 😄

If you'd like to see another example of that pattern in the works, though, you might also like https://github.com/mozilla/application-services/commit/cafe35d9bc22943e51aea41c494f17363119e37e#diff-56f5ce6ace9bcf94d4ef9b46cc333bf21a39e94c7e1a201872015291df47a4b8.

In that commit:

`Error` is the error type that's used internally in the crate. The magic `#[from]` attribute on each variant is what lets it "unify" (😅) all the different kinds of errors into a single type. It makes working with the `?` operator much easier.

For example, a function might do something that can fail with a `serde_json::Error`, and then further down, something that can fail with a `rusqlite::Error`. The function can return a `Result<T, Error>`, and both the `serde_json::Error` and `rusqlite::Error` will be converted into variants of `Error` when those calls are made with the `?` operator.

`SuggestError` is the error that's returned to consumers. (I think I'll actually rename it to `SuggestApiError`, to match the convention we use in tabs and logins!) It has a UniFFI binding, and a `From<Error>` implementation that creates it from an internal `Error`. For now, it just has a catch-all `Other` variant, which is kind of like the `InternalError` and `UnexpectedError` that @mhammond mentioned above. As we build out the component, though, we might add other kinds of variants to `Suggest{Api}Error`.

For example, we might add `Interrupted` to signal that a database read or write was interrupted. This would be an "expected" error—maybe the work manager told our ingestion worker to stop, or maybe the user typed new input into the address bar, and we want to abort the query for their old input—in that we'd want Kotlin / Swift code to catch it.

But Remote Settings returning a weird response, or an attachment failing to parse, would be an "unexpected" `Other` error. For Kotlin / Swift consumers, it's not interesting that ingestion failed specifically with an `rs_client::Error` vs. a `serde_json::Error`—it failed, and the worker should retry (and maybe report the error to Sentry).

Hopefully that helps!

.buildconfig-android.yml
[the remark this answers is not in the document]
mhammond: That sounds perfect to me. However, I thought the idea was that nimbus would reference rs_client. So you are suggesting that `remote_settings` could be *both* a "top level" component (ie, exposed directly to the mobile clients) *and* a dependency of nimbus? If that works I'd be fine with it - less is more!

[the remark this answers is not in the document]
linabutler: > So you are suggesting that `remote_settings` could be *both* a "top level" component (ie, exposed directly to the mobile clients) *and* a dependency of nimbus?

Yes, exactly!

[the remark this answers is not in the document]
MatthewTighe: Awesome, glad this was resolved while I was out last week. I'm in agreement with this plan as well!

components/support/rs_client/src/http_client.rs
MatthewTighe: Yeah, I think we were on the same page! `RemoteSettingsResponse` might be worth keeping to avoid future naming conflicts in the iOS megazord since there's no real namespacing strategy available there AFAIK.

linabutler: Oop, yes, I'm sorry I forgot about `RemoteSettingsResponse`! :joy: Yep, we have to worry about naming conflicts for anything in UDL, but non-`pub` types (and `pub` types that aren't in UDL) in Rust are okay!

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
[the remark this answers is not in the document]
MatthewTighe: Still working through these, but it seems like the linter won't allow package names with hyphens anyway. I'm of the opinion that we should still keep snake-cased crate names for legibility, and combine the words in Android package names to avoid these issues.

components/remote_settings/src/error.rs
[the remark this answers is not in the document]
MatthewTighe: Decided to punt on this for now and filed https://github.com/mozilla/application-services/issues/5566 as a follow-up

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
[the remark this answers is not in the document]
MatthewTighe: Deleted a couple old comments but TL;DR: the complexity analysis tool we use `detekt`, won't allow hyphenated names. I'd prefer to just switch the naming convention here for Android's sake instead of switching all the Rust code to use concatenated names, but let me know if you'd prefer that solution.

linabutler (approved):
I think this is ready to go! Thanks _so_ much for sticking with this, @MatthewTighe, I'm super excited to see this land! 🚀

LGTM, pending a green CI run. It looks like there's a `cargo clippy` nit, and an iOS issue?

For iOS, I think you'll need to:

* Add a line to [this section](https://github.com/mozilla/application-services/blob/b79e5e0d8b5bd0a6eb3e0ee0b5f57c8887fb414d/megazords/ios-rust/build-xcframework.sh#L160-L168) to generate the headers.
* Add the dependency and `extern crate` to `megazords/ios-rust`.

And then you should be good to go!

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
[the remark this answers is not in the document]
mhammond: I agree 100% - but...

> fxaclient project => fxa-client crate, errorsupport => error-support, syncmanager => sync_manager

We'd probably still want some consistency there, right? ie, IIUC, in all the above cases, the crate names would *always* have underscores, but the package names would be inconsistent between dashes and underscores - which is already true today. Further, there's also the directory name to consider :)

Personally, I think I'd vote for "use dashes everywhere they are allowed" - so we'd have a package and directory with name "sync-manager" while the crate name would be "sync_manager". But this is getting off-topic for this PR - I think what's here is great (and it is all that much cleaner now we've killed that extra crate!)

components/remote_settings/src/client.rs
```diff
@@ -0,0 +1,551 @@
...
+
+    /// Fetches all records for a collection that can be found in the server,
+    /// bucket, and collection defined by the [ClientConfig] used to generate
+    /// this [Client].
+    pub fn get_records(&self) -> Result<RemoteSettingsResponse> {
+        let path = format!(
+            "v1/buckets/{}/collections/{}/records",
+            &self.bucket_name, &self.collection_name
+        );
+        let resp = self.make_request(path)?;
```
mhammond, line 55: Can these few lines be replaced with `let resp = self.get_records_raw()?;`?

components/remote_settings/src/client.rs
```diff
@@ -0,0 +1,551 @@
...
+    /// for a collection that can be found in the server, bucket, and
+    /// collection defined by the [ClientConfig] used to generate this [Client].
+    pub fn get_records_since(&self, timestamp: u64) -> Result<RemoteSettingsResponse> {
+        let path = format!(
+            "v1/buckets/{}/collections/{}/records?_since={}",
+            &self.bucket_name, &self.collection_name, timestamp
+        );
+        let resp = self.make_request(path)?;
+        let records = resp.json::<RecordsResponse>()?.data;
+        let last_modified = resp.headers.get_as("etag").unwrap()?;
```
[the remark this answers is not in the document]
mhammond: There is a `ResponseError(String)` that seems suitable, so the cleaner story is probably just `.ok_or_else(|| RemoteSettingsError::ResponseError("no etag in response".into())?;` or something?  (OTOH though, IIRC this unwrap already existed and has forever - no etag does mean something very unexpected, and panics in our mobile clients aren't as catastrophic as they are on desktop)

components/remote_settings/src/client.rs
```diff
@@ -0,0 +1,551 @@
...
+    /// Fetches all records for a collection that can be found in the server,
+    /// bucket, and collection defined by the [ClientConfig] used to generate
+    /// this [Client].
+    pub fn get_records(&self) -> Result<RemoteSettingsResponse> {
+        let path = format!(
+            "v1/buckets/{}/collections/{}/records",
+            &self.bucket_name, &self.collection_name
+        );
+        let resp = self.make_request(path)?;
+        let records = resp.json::<RecordsResponse>()?.data;
```
mhammond, line 56: Will this work correctly for a 304? If the body was empty (which I think is typical with that response?) then I suspect this json parsing might fail.

components/remote_settings/src/client.rs
```diff
@@ -0,0 +1,551 @@
...
+
+        let url = self.base_url.join(&path)?;
+        let req = Request::get(url);
+        let resp = req.send()?;
+
+        let mut current_remote_state = self.remote_state.lock();
+        self.handle_backoff_hint(&resp, &mut current_remote_state.backoff)?;
+
+        if resp.is_success() || resp.status == status_codes::NOT_MODIFIED {
+            Ok(resp)
```
mhammond, line 128: If the body really is empty here, a possibility might be a new enum, or even `Result<Option<Response>>`?

mhammond (commented):
This looks fantastic. I think it's worth making sure I'm incorrect about the 304 handling though, or if I'm not, then it probably needs to be addressed in some way?

components/remote_settings/android/src/test/java/org/mozilla/appservices/remotesettings/RemoteSettingsTest.kt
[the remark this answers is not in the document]
linabutler: > IIUC, in all the above cases, the crate names would always have underscores, but the package names would be inconsistent between dashes and underscores - which is already true today.

Hmmm, am I looking in the wrong place? 😅 I see mostly concatenated (`oneword`) names, a couple of hyphens, and no underscores in the package or project names in `.buildconfig-android.yml`—the only underscores there are in the directory names—and most of the package names in `AndroidManifest.xml` are concatenated, with a couple of underscores.

> Further, there's also the directory name to consider :)

Ohh, I was using "directory name" and "crate name" interchangeably, and didn't even think about them being different! Except for the `support/` components, I don't see other examples under `components/` where the directory name is different from the crate name (in `Cargo.toml`).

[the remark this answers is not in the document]
mhammond: >  package or project names in .buildconfig-android.yml

Right! I was talking just about the rust package names - so many places to get it wrong :)

components/remote_settings/src/client.rs
linabutler: Oh, nice catch! I wonder if we can remove the 304 check entirely, since we aren't (AFAICT?) ever sending an `If-None-Match` or an `If-Modified-Since` request header, and Kinto's docs (and RFC 9110) say that's the only time we'll receive a 304 response.

mhammond: ah - I went looking for them too, but then noticed the "since" param in the url, which I thought might trigger it - but yeah, you are probably correct that in that case we'll get a 200 with no records - so yeah, if that's the case, I agree we should kill the 304 handling.

components/remote_settings/src/client.rs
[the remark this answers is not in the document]
MatthewTighe: This is a little further complicated because `get_as` returns an `Option<Result<....>>` and not just `Result<...>`. I matched on that pretty explicitly, but please let me know if there's a way to simplify it down some

components/remote_settings/src/client.rs
MatthewTighe: Just to confirm, the above all means that we don't want to add explicit 304 handling?

components/remote_settings/src/client.rs
MatthewTighe: This relates to the 304 handling above, correct?

components/remote_settings/src/client.rs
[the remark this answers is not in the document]
linabutler: Nice, thanks! You can tighten it up a bit with:

```rs
let last_modified = resp.headers.get_as("etag").ok_or_else(|| RemoteSettingsError::ResponseError("no etag header".into()))??;
```

`ok_or_else` turns your `Option<Result<...>>` into a `Result<Result<...>>`, then the first `?` unwraps the outer one, and the second `?` unwraps the inner one.

components/remote_settings/src/client.rs
linabutler: > the above all means that we don't want to add explicit 304 handling?

Correct.

I double-checked, and `curl "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/quicksuggest/records?_since=1683313881574"` (where `1683313881574` is an `ETag` that I got from a previous request without the `_since`), indeed, returns a 200 with no records:

```
< HTTP/2 200
< ...
{"data":[]}
```

...So we don't need the explicit 304 handling! 🚀

components/remote_settings/src/client.rs
linabutler: Correct, let's just drop the 304 handling, and then we don't have to worry about empty bodies.

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
```diff
@@ -70,13 +70,15 @@
 		39F5D7642956161E004E2384 /* Operation+.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39F5D7632956161E004E2384 /* Operation+.swift */; };
 		39F5D766295616E3004E2384 /* NimbusBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39F5D765295616E3004E2384 /* NimbusBuilder.swift */; };
 		45CC574A28AD9C86006D55AA /* errorsupport.udl in Sources */ = {isa = PBXBuildFile; fileRef = 45CC574828AD9C31006D55AA /* errorsupport.udl */; };
+		AB65933C2A0C527900DBF059 /* remote_settings.udl in Sources */ = {isa = PBXBuildFile; fileRef = AB65933B2A0C524E00DBF059 /* remote_settings.udl */; };
+		AB6593402A0C52A000DBF059 /* remote_settings.swift in Sources */ = {isa = PBXBuildFile; fileRef = AB65933F2A0C52A000DBF059 /* remote_settings.swift */; };
```
linabutler, line 74:
```suggestion
```



Oh, sorry I didn't catch this earlier—please remove this `remote_settings.swift` line.

(I think the equivalent operation in the Xcode UI is to remove `remote_settings.swift` from the "Compile Sources" section of the "Build Phases" tab of the target?)

There's a rule in Xcode that automatically generates Swift sources from `*.udl` files (in the "Build Rules" tab), and adding the generated `.swift` file as a source actually causes all sorts of problems 😬

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
```diff
@@ -774,13 +800,15 @@
 				1B3BC94027B1D62800229CF6 /* Places.swift in Sources */,
 				F81C7B9829DE305C00FAF8F9 /* SyncManagerTelemetry.swift in Sources */,
 				1BF50F1527B1E17B00A9C8A5 /* FxAccountState.swift in Sources */,
+				AB6593402A0C52A000DBF059 /* remote_settings.swift in Sources */,
```
linabutler, line 803:
```suggestion
```

...And nix this line, too, please. 😊

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
MatthewTighe: Thanks for the keen eye! I had hard reset the iOS setup to try and get it working again since I had xcode open when I rebased and things had become kinda funky. Guess I missed a few steps 😅

components/remote_settings/src/client.rs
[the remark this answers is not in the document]
MatthewTighe: TY, that's much better 🙏

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
MatthewTighe: Hmm, it looks like these lines are automatically added when adding these files to the newly created group when following [this documentation](https://github.com/mozilla/application-services/blob/main/docs/howtos/adding-a-new-component.md#the-swift-bindings):
```
Finally, in the Project navigator, add a sub-group named "Generated", pointing to the ./Generated/ subdirectory, and containing entries for the files generated by UniFFI: * <your_crate_name>.swift * <your_crate_name>FFI.h Make sure that "Copy items if needed" is unchecked, and that nothing is checked in "Add to targets".
```

Though I'm also seeing that my local copy does not have a bunch of `.udl` files in the `Compile Sources` section any more either, so I'm wondering if perhaps the documentation has recently become out of date.

linabutler: Oh, how odd! It automatically added them to the target even with nothing checked in "Add to targets"? I wonder if there's another setting that controls that... 🤔 Sorry you had to untangle all that, but it looks like you fixed the build and we're good to go!

I think this is the note about double-checking that the generated `.swift` file doesn't appear in "Compile Sources", though:

> Double-check that `<your_crate_name>.swift` does not appear in the "Compile Sources" section.

(That one bit me in #5346, along with the "Copy Bundle Resources" callout above it that caused a dependency cycle error during the build. Xcode's default behavior unfortunately makes things worse here 😔)

But it's very subtle, and I wonder if we can call it out more explicitly that `<your_crate_name>.swift` is referring to the `Generated/<your_crate_name>.swift`.

components/nimbus/src/lib.rs
```diff
@@ -19,22 +19,21 @@ pub use error::{NimbusError, Result};
 #[cfg(debug_assertions)]
 pub use evaluator::evaluate_enrollment;
 pub use matcher::AppContext;
+pub use remote_settings::RemoteSettingsConfig;
```
jeddai, line 22: This should move to the "stateful" section of the `cfg_if` below

jeddai (changes requested):
This looks really great! I haven't tested functionality yet but I do have some small change requests re: conditional compilation in the nimbus library!

MatthewTighe:
> re: conditional compilation in the nimbus library!

I knew I had gotten this wrong. Somehow got it in my head that the rs_client was conditionally compiled for when the `stateful` was disabled, ran into compilation errors, and sorted it out the easiest way. Thanks for stopping by and helping with my error!

jeddai (approved):
No indication the one branch build test is failing because of app-services so I'm comfortable approving it!

mhammond:
woohoo 🎉

jonalmeida:
Congrats! 🙌

linabutler:
Awesome! 💯🏁🏆

PR #5432, opened 2023-03-13
Allow autofill to roundtrip data it doesn't know about

skhamis:
This includes both addresses and credit cards.

1. We essentially do nothing on incoming (added a test to validate that we can de-serialize though)
2. Outgoing is where most of the heavy lifting is done, but we essentially just pull whatever `unknown_fields` were in the corresponding payload in the mirror table and attach it to our outgoing payload.


components/autofill/sql/create_shared_schema.sql
```diff
@@ -22,7 +22,11 @@ CREATE TABLE IF NOT EXISTS addresses_data (
     time_last_modified  INTEGER NOT NULL,
     times_used          INTEGER NOT NULL,
 
-    sync_change_counter INTEGER NOT NULL
+    sync_change_counter INTEGER NOT NULL,
+
+    -- fields on incoming records we don't know about and roundtrip.
+    -- a serde_json::Value::Object as an encrypted string.
```
mhammond, line 28: I don't think it will be encrypted for addresses, right? But TBH I don't think we need this column at all - we can just extract it from the mirror table's payload.

components/autofill/src/sync/address/mod.rs
```diff
@@ -96,10 +97,16 @@ struct PayloadEntry {
...
+    // Fields that the current schema did not expect, we store them only internally
+    // to round-trip them back to sync without processing them in any way
+    // Additional "unknown" round-tripped fields
+    #[serde(flatten)]
+    unknown_fields: UnknownFields,
 }
 
 impl InternalAddress {
     fn from_payload(p: AddressPayload) -> Result<Self> {
+        // TODO - should it always be version 1?
```
mhammond, line 109: yeah, it should be. I think this field is now effectively "frozen".

mhammond (commented):
I think this isn't quite the correct direction here. The advantage this engine has is that it already has the complete payload as seen on the server, so I don't think we need any new DB columns at all and we get all the encryption for free, and don't even need anything on InternalAddress - but probably *do* need the unknown_fields struct on `PayloadEntry`

So I'm pretty sure incoming needs almost zero work (other than maybe tests - but maybe not) The complete payload is in the "staging" table which is copied to the mirror (ie, the mirror doesn't store something that went via an InternalAddress, it stores the raw payload)

Outgoing is where most of the work is going to need to be done, in `fetch_outgoing_records()`, where you select records with a change counter > 0, you will need an outer join to read the `payload` record from the mirror table, and records where there is no mirror obviously have a `null` for that value.

So in `fetch_outgoing_records` you have:
```
        let record_from_data_row: &dyn Fn(&Row<'_>) -> Result<(OutgoingBso, i64)> = &|row| {
            Ok((
                OutgoingBso::from_content_with_id(InternalAddress::from_row(row)?.into_payload()?)?,
```

All the magic can be done in there. Something like *deserialize* the mirror payload column if not null into one of the `AddressPayload` struct - its `PayloadEntry` has `unknown_fields` - which are the unknown fields as last seen on the server. The `unknown_fields` from that structure can be copied into the result of `InternalAddress::from_row(row)?.into_payload()?` - which is the `AddressPayload` struct we are about to upload.

To put it another way - that function now deals with exactly 1 `AddressPayload` struct - the one to upload. With these changes it will deal with 2 - the new one will be one deserialized from the mirror row just so we can get the `unknown_fields` out of it and move them to the one we are about to upload.

skhamis:
> I think this isn't quite the correct direction here. The advantage this engine has is that it already has the complete payload as seen on the server, so I don't think we need any new DB columns at all and we get all the encryption for free, and don't even need anything on InternalAddress - but probably do need the unknown_fields struct on PayloadEntry

Thanks for the initial review/comments! I agree that we should keep it with the mirror to prevent having to do any DB stuff at all. I've updated the PR to go down this route (almost everything was already in place for the most part). I do want to try get a couple more tests -- I also will keep credit cards in a separate PR (but stacked) since it'll be easier to review

components/autofill/src/sync/credit_card/incoming.rs
```diff
@@ -558,4 +581,22 @@ mod tests {
         assert!(get_credit_card(&db.writer, &local_guid).is_err());
         assert!(get_credit_card(&db.writer, &incoming_guid).is_ok());
     }
+
+    #[test]
+    fn test_get_incoming_unknown_fields() {
+        let json = test_json_record('D');
+        let address_payload = serde_json::from_value::<CreditCardPayload>(json).unwrap();
```
mhammond, line 588: nit - 'address' should be 'credit_card' here and below?

components/autofill/src/sync/mod.rs
```diff
@@ -18,6 +18,10 @@ use sync15::{CollectionName, ServerTimestamp};
 use sync_guid::Guid;
 use types::Timestamp;
 
+// This type is used as a snazzy way to capture all unknown fields from the payload
+// upon deserialization without having to work with a concrete type
+pub(super) type UnknownFields = serde_json::Map<String, serde_json::Value>;
```
mhammond, line 23: I think `super` here is the crate. `pub(self)` is probably all you need here as nothing outside the sync module knows about this. But, in that case and because this is in the root of the crate, I'm not sure you need *any* pub here?

components/autofill/src/sync/credit_card/mod.rs
```diff
@@ -78,7 +79,7 @@ pub(crate) struct CreditCardPayload {
 
     // For some historical reason and unlike most other sync records, creditcards
     // are serialized with this explicit 'entry' object.
-    entry: PayloadEntry,
+    pub(crate) entry: PayloadEntry,
```
mhammond, line 82: I think these pub changes are only needed because of the visibility of `UnknownEntry` - it's probably not necessary if you change that as I note below.

components/autofill/src/sync/credit_card/mod.rs
```diff
@@ -105,6 +106,10 @@ struct PayloadEntry {
     #[serde(rename = "timesUsed")]
     pub times_used: i64,
     pub version: u32, // always 3 for credit-cards
+    // Fields that the current schema did not expect, we store them only internally
+    // to round-trip them back to sync without processing them in any way
+    #[serde(flatten)]
+    pub unknown_fields: UnknownFields,
```
mhammond, line 112: but even then, you probably don't need this `pub` at all? At least, addresses doesn't have it.

mhammond (approved):
LGTM with the request changes. Let me know if you can remove all the `pub` stuff entirely - I think you can.

components/autofill/src/sync/address/outgoing.rs
```diff
@@ -256,4 +293,46 @@ mod tests {
...
+        do_test_outgoing_synced_with_local_change(
+            &tx,
+            &ao,
+            &test_record.guid,
+            DATA_TABLE_NAME,
+            MIRROR_TABLE_NAME,
+            STAGING_TABLE_NAME,
+            COLLECTION_NAME.into(),
+        );
+    }
```
bendk, line 337: How about some more tests?

- **3 way merge:** there's a record in the mirror table with `unknown_fields` set, an incoming record with new `unknown_fields`, and local changes.  I think we should take the new incoming `unknown_fields` in this case.
- **Fork:** Like before there's a mirror record with `unknown_fields`, an incoming record with changes, and a local record with changes.  In this case, the incoming and local record have conflicts, so we "fork" the record.  I'm not totally sure how this should work, but I think:
  - The forked local record should have the `unknown_fields` from the existing mirror record.  There should be a new mirror record with corresponding values.
  - We shouldn't include the incoming record in the outgoing changes -- this means it will keep it's `unknown_fields` values.  We should update the mirror record to match the incoming record.

bendk (changes requested):
I think the addresses code is correct, but there are a couple corner cases that I think we should test. 

I didn't look at the credit card code too deeply.  It looks pretty much the same to me.  I think we should have the same corner case tests for credit cards.

components/autofill/src/sync/address/outgoing.rs
skhamis: Thanks for the review! Hmmm, definitely something to think about more (and maybe add appropriate tests to reflect what's actually happening). However, if I'm thinking of this right...

1. The three-way merge is happening on the `InternalAddress` struct, which `unknown_fields` is not a part of -- due to `unknown_fields` not being in the local db and strictly on the mirror table
2. Upon outgoing (so I presume all incoming issues have been merged/planned) we take the the `unknown_fields` from the `mirror` table, which means it should be whatever was latest fetched from the server

So I'm thinking that `unknown_fields` should *never* be in a 3-way merge situation, at least with the current code. 

Forking is a tad bit more complicated, because according to `apply_incoming_action` https://github.com/mozilla/application-services/blob/b14fd3566879f927901691407c0341061bbdeccb/components/autofill/src/sync/mod.rs#L370-L374

we change the local guid, so when we try to get the corresponding `unknown_fields` from the mirror, it'll of course not be there. There is also not an easy way to keep track it's previous guid

mhammond: Ben's note that we need tests there is excellent.

Re 3-way merge, I agree the mirror should end up with whatever unknown fields were incoming, and I think the current impl will correctly do that, but should have a test which proves it.

Re forking: ugh. I think it should actually be possible to just change the guid in the mirror table too, so long as we make sure the record has an appropriate change counter (which I think it will)? Again, test demonstrating that is important.

bendk: Changing the GUID on the mirror table seems scary to me.  How will other clients see that change when they sync next?

Could `change_local_guid()` also update the GUID in the mirror table so that the forked record would see that when it came to upload the changes, or is it too late by then since the new mirror record was already added.

mhammond: > Changing the GUID on the mirror table seems scary to me. How will other clients see that change when they sync next?

> Could `change_local_guid()` also update the GUID in the mirror table so that the forked record would see that when it came to upload the changes, or is it too late by then since the new mirror record was already added.

I'm slightly confused here - that second sentence is exactly what I meant, but the first sentence says you think that is scary?

skhamis: I wonder if I'm thinking about the forking right -- from what I understand looking at the code:

1. We have a mirror record that has `given-name` = "mir"
2. We have a record that changed locally `given-name` = "loc"
3. We have a incoming record that also changed it `given-name` = "inc"

Given that it's impossible to reconcile this -- we implement a forking strategy, which means
1. The mirror record now has `inc`
2. The local record now has `inc`
3. A NEW local record has `loc` <--> This does not exist on the server yet and will be written as a new record on next sync


So the current suggestion is to change the `guid` of `1` to match the guid of `3` --> Which I think may be a bad idea as the data doesn't match given that the "server wins" for the existing record. A good visualization of this is in the `test_reconcile.rs` json (which is where I'm adding my additional testing).

https://github.com/mozilla/application-services/blob/d8cdb7a15ff27a25336265e49a1936147789cd7e/components/autofill/src/sync/tests/test_reconcile.rs#L279-L310

mhammond: Ah yeah, I think you are correct - at the time we detect we need to fork, we've already overwritten the mirror record with any new incoming fields - we've then lost what the unknown fields were.

So I'm on-board with that - the forked record gets no unknown fields - in effect, it looks like a brand-new record created from scratch on this device, which seems perfectly fine to me.

mhammond: > Ah yeah, I think you are correct - at the time we detect we need to fork, we've already overwritten the mirror record with any new incoming fields

Actually, I don't think that's true - we've *staged* the incoming record in `addresses_sync_staging` - `addresses_mirror` should have the old version until `finish_incoming()` is called?

skhamis: > > Ah yeah, I think you are correct - at the time we detect we need to fork, we've already overwritten the mirror record with any new incoming fields
> 
> Actually, I don't think that's true - we've _staged_ the incoming record in `addresses_sync_staging` - `addresses_mirror` should have the old version until `finish_incoming()` is called?

That's correct, this turned out to be somewhat easy and I added it as a separate commit to see the changes for this. I still think there is a theoretical question here of _do we want the forked record to have an accompanying local mirror_ -- it'll get uploaded all the same and treated as a "changed" record rather than "new".

components/autofill/src/sync/tests/test_reconcile.rs
```diff
@@ -307,6 +327,11 @@ lazy_static::lazy_static! {
             "given-name": "Kip",
             "family-name": "Jones",
         },
+        // Because our record has been "forked" the local change we send out
+        // should have the ORIGINAL unknown fields
+        "outgoing": {
```
mhammond, line 332: Let's add `"given-name": "Skip",` to this object, just as an extra test that the record which is outgoing is what we think it is.

The other thing this test should do is somehow check that `reconciled` now has `new-unknown-field` in its mirror, which isn't as simple as adding it to that json. So it's not immediately clear how to get this test to check that :( Maybe something like a new `reconciled_unknown` object or something, and the test has to actually hit the db for the mirror record or something?

mhammond (changes requested):

components/autofill/src/sync/mod.rs
```diff
@@ -370,6 +381,10 @@ fn apply_incoming_action<T: std::fmt::Debug + SyncRecord>(
         IncomingAction::Fork { forked, incoming } => {
             // `forked` exists in the DB with the same guid as `incoming`, so fix that.
             rec_impl.change_local_guid(tx, incoming.id(), forked.id())?;
+            // The mirror will be overridden by what's newer on the server, causing us
+            // to potentially lose any unknown_fields here: so we change the guid to match the new
+            // forked record so we can extract any unknown fields
+            rec_impl.change_mirror_guid(tx, incoming.id(), forked.id())?;
```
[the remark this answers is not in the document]
mhammond: > change_local_guid() is also used by UpdateLocalGuid 

doh! FWIW, in that case I'm fairly sure the mirror must not exist for that guid (get_local_dupe's queries explicitly rule out items in the mirror). That probably doesn't mean my suggestion is OK though, it does seem wrong to carry all these assumptions through all these levels.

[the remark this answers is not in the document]
mhammond: (OTOH, maybe it *is* OK if well commented and maybe with a new name. Changing the local guid already implicitly assumed there was no mirror record, but now we have a case where we might, and the mirror must change too. I can't think we'd ever have a case for changing a local guid but wanting to leave the old guid in the mirror "orphaned" with the old guid)

bendk (approved):
I've only been following from afar, but this looks great to me.  I agree with Mark's testing suggestion and I had a small style suggestion as well, but this seems great overall to me.

components/autofill/src/sync/common.rs
```diff
@@ -75,9 +75,12 @@ pub(super) fn common_remove_record(conn: &Connection, table_name: &str, guid: &G
     Ok(())
 }
 
+// This optionally takes a mirror table name to update if
+// there is a corresponding record with the same guid we'd like to update
 pub(super) fn common_change_guid(
     conn: &Connection,
     table_name: &str,
+    mirror_table_name: Option<&str>,
```
mhammond, line 83: Why is this an option? It never seems to be called with `None`

mhammond (approved):
This LGTM after changing that `Option` - thanks Sammy!

PR #5434, opened 2023-03-15
Add an ADR for sync payload evolution

mhammond:
This is a summary of the google doc and our discussions. Please feel free to add commits to this branch if you have ideas for significant changes. CC @galich as I can't explicitly request review.

### DIFF
--- a/docs/adr/0008-sync-payload-evolution.md
+++ b/docs/adr/0008-sync-payload-evolution.md
@@ -14,11 +14,11 @@ Sync exists on all platforms (Desktop, Android, iOS), all channels (Nightly, Bet
 Whenever there are feature changes or requests that potentially involve schema changes, there are not a lot of good options to ensure sync doesn’t break for any specific client.
 Since sync data is synced from all channels, we need to make sure each client can handle the new data and that all channels can support the new schema.
 Issues like [credit card failing on android and desktop release channels due to schema change on desktop Nightly](https://bugzilla.mozilla.org/show_bug.cgi?id=1812235)
-is an example of such cases we can run into.
+are examples of such cases we can run into.
 This document describes our decision on how we will support payload evolution over time.
 
-Note that even though this document existing in the application-services repoository, it should
-be considered to apply too all sync implementation, whether in this repository, in mozilla-central,
+Note that even though this document exists in the application-services repository, it should
+be considered to apply to all sync implementations, whether in this repository, in mozilla-central,
 or anywhere else.
 
 ## Definitions
@@ -47,6 +47,8 @@ or anywhere else.
 * Because such evolution should be rare, we do not want to set an up-front policy about locking out
   "old" versions just because they might have a problem in the future. That is, we want to avoid
   a policy that dictates versions more than (say) 2 years old will break when syncing "just in case"
+* Any solution to this must be achievable in a relatively short timeframe as we know of product
+  asks coming down the line which require this capability.
 
 ## Considered Options
 
@@ -55,6 +57,7 @@ or anywhere else.
 * A policy which prevents "recent" clients from syncing, or editing data, or other restrictions.
 * A formal schema-driven process.
 * Consider the sync payloads frozen and never change them.
+* Use separate collections for new data
 
 ## Decision Outcome
 
@@ -81,7 +84,7 @@ A summary of this option is a policy by which:
   that "new" clients must support both new fields *and* fields which are considered deprecated
   by these "new" clients because they are still used by "recent" versions.
 
-The pros and conts:
+The pros and cons:
 
 * Good, because it meets the requirements.
 
@@ -121,7 +124,7 @@ This was rejected because:
 
 Ideally we could formally describe schemas, but we can't come up with anything here which
 works with the constraints of supporting older clients - we simply can't update older released
-Firefoxes so they know how to work with the new schemas. We also couldn't some up with a solution
+Firefoxes so they know how to work with the new schemas. We also couldn't come up with a solution
 where a schema is downloaded dynamically which also allowed the *semantics* (as opposed to simply
 validity) of new fields to be described.
 
@@ -137,6 +140,25 @@ A process where payloads are frozen was rejected because:
   particularly around data-loss for older clients. For example, adding a credit-card
   on a Nightly version but having it be completely unavailable on a release firefox
   isn't acceptable.
+
+### Use separate collections for new data
+
+We could store the new data in a separate collection. For example define a
+bookmarks2 collection where each record has the same guid as one in bookmarks alongside any new fields.
+Newer clients use both collections to sync.
+
+The pros and cons:
+
+* Good, because it allows newer clients to sync new data without affecting recent or older clients
+* Bad, because sync writes would lose atomicity without server changes.
+  We can currently write to a single collection in an atomic way, but don't have a way to write to multiple collections.
+* Bad because this number of collections grows each time we want to add fields.
+* Bad because it potentially leaks extra information to an attacker that gets access to the encrypted server records.
+  For example if we added a new collection for a single field, then the attacker could guess if that
+  field was set or not based on the size of the encrypted record.
+* Bad because it's difficult to handle nested data with this approach,
+  for example adding a field to a history record visit.
+* Bad because it has the same issue with dependent data as the chosen solution.
 
 ## Links <!-- optional -->
 
--- a/docs/adr/0008-sync-payload-evolution.md
+++ b/docs/adr/0008-sync-payload-evolution.md
@@ -5,7 +5,7 @@
 * Date: 2023-03-15
 
 Technical Story:
-* https://github.com/mozilla/application-services/pull/5???
+* https://github.com/mozilla/application-services/pull/5434
 * https://docs.google.com/document/d/1ToLOERA5HKzEzRVZNv6Ohv_2wZaujW69pVb1Kef2jNY
 
 ## Context and Problem Statement
--- a/docs/adr/0008-sync-payload-evolution.md
+++ b/docs/adr/0008-sync-payload-evolution.md
@@ -1,0 +1,144 @@
+# Handling the evolution of Sync payloads
+
+* Status: Accepted
+* Deciders: sync team, credentials management team
+* Date: 2023-03-15
+
+Technical Story:
+* https://github.com/mozilla/application-services/pull/5???
+* https://docs.google.com/document/d/1ToLOERA5HKzEzRVZNv6Ohv_2wZaujW69pVb1Kef2jNY
+
+## Context and Problem Statement
+
+Sync exists on all platforms (Desktop, Android, iOS), all channels (Nightly, Beta, Release, ESR) and is heavily used across all Firefox features.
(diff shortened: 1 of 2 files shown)


docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+
+* The most naive approach here would not meet the needs of Firefox in the future.
+
+* A complicated system where we started creating new payload and new collections
+  (ie, freezing "old" schemas but then creating "new" schemas only understood by
+  newer clients) could not be conceived in a way that still met the requirements,
+  particularly around data-loss for older clients. For example, adding a credit-card
+  on a Nightly version but having it be completely unavailable on a release firefox
+  isn't acceptable.
+
```
bendk, line 140: Maybe one more?

### Use separate collections for new data

We could store the new data in a separate collection.  For example define a `bookmarks2` collection where each record has the same `guid` as one in `bookmarks` alongside any new fields.  Newer clients use both collections to sync.

Pros and cons:

* Good, because it allows newer clients to sync new data without affecting recent or older clients
* Bad, because sync writes would lose atomicity without server changes.  We can currently write to a single collection in an atomic way, but don't have a way to write to multiple collections.
* Bad because this number of collections grows each time we want to add fields.
* Bad because it potentially leaks extra information to an attacker that gets access to the encrypted server records.  For example if we added a new collection for a single field, then the attacker could guess if that field was set or not based on the size of the encrypted record.
* Bad because it's difficult to handle nested data with this approach, for example adding a field to a history record visit.
* Bad because it has the same issue with dependent data as the chosen solution.

bendk (approved):
Couple small comments, but I think this distills the google doc discussion very well.

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+* The user would certainly perceive data-loss if we ignored the incoming data entirely.
+* If we still wanted older versions to "partially" see the record (eg, but disallow editing) we'd
+  still need most of the chosen option anyway - specifically, we could still never
+  deprecate fields etc.
+* The UI/UX of trying to explain to the user why they can't edit a record was deemed impossible
+  to do in a satisfactory way.
+* This would effectively penalize users who chose to use Nightly Firefoxes in any way. Simply
+  allowing a Nightly to sync would effectively break Release/Mobile Firefox versions.
+
+### A formal schema-driven process.
```
tarikeshaq, line 120: This is what protobuf does, but of course that's a land very far away from where we are :) Just reminded me of some reading I did

(https://earthly.dev/blog/backward-and-forward-compatibility/)

tarikeshaq (approved):
Rubber stamp as I wasn't involved in the conversations and have little to no context here!

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+Sync exists on all platforms (Desktop, Android, iOS), all channels (Nightly, Beta, Release, ESR) and is heavily used across all Firefox features.
+Whenever there are feature changes or requests that potentially involve schema changes, there are not a lot of good options to ensure sync doesn’t break for any specific client.
+Since sync data is synced from all channels, we need to make sure each client can handle the new data and that all channels can support the new schema.
+Issues like [credit card failing on android and desktop release channels due to schema change on desktop Nightly](https://bugzilla.mozilla.org/show_bug.cgi?id=1812235)
+is an example of such cases we can run into.
+This document describes our decision on how we will support payload evolution over time.
+
+Note that even though this document existing in the application-services repoository, it should
+be considered to apply too all sync implementation, whether in this repository, in mozilla-central,
+or anywhere else.
```
skhamis, line 22:
```suggestion
Note that even though this document exists in the application-services repository, it should
be considered to apply to all sync implementations, whether in this repository, in mozilla-central,
or anywhere else.
```

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+* Different engines must identify different locations where this might happen. For example, the
+ `passwords` engine would identify the "root" of the payload, `addresses` and `creditcards` would
+ identify the `entry` sub-object in the payload, while the history engine would probably identify
+ *both* the root of the payload and the `visits` array.
+
+* Fields can not change type, nor be removed for a significant amount of time. This might mean
+  that "new" clients must support both new fields *and* fields which are considered deprecated
+  by these "new" clients because they are still used by "recent" versions.
+
+The pros and conts:
```
skhamis, line 84:
```suggestion
The pros and cons:
```

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+* The UI/UX of trying to explain to the user why they can't edit a record was deemed impossible
+  to do in a satisfactory way.
+* This would effectively penalize users who chose to use Nightly Firefoxes in any way. Simply
+  allowing a Nightly to sync would effectively break Release/Mobile Firefox versions.
+
+### A formal schema-driven process.
+
+Ideally we could formally describe schemas, but we can't come up with anything here which
+works with the constraints of supporting older clients - we simply can't update older released
+Firefoxes so they know how to work with the new schemas. We also couldn't some up with a solution
```
skhamis, line 124:
```suggestion
Firefoxes so they know how to work with the new schemas. We also couldn't come up with a solution
```

skhamis (approved):
Awesome ADR! I think this nails the problem we are facing and how we hope to solve it. Some minor nits and the other reviewers got to most of the other stuff, but I think it's good to go!

docs/adr/0008-sync-payload-evolution.md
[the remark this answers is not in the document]
mhammond: huh - so yeah, maybe I should actually just land this under https://firefox-source-docs.mozilla.org/services/sync/index.html, as I guess that's really where the canonical public sync docs should live - WDYT?

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,144 @@
...
+* It must be possible to change what data is carried by Sync to meet future product requirements.
+* Both desktop and mobile platforms must be considered.
+* We must not break "recent" Firefox installations, nor cause what the user might perceive as data-loss,
+  when a "new" Firefox installation syncs, and vice-versa.
+* Some degree of breakage for "old" Firefox installations when "new" or "recent" firefoxes sync
+  might be considered acceptable if absolutely necessary.
+* However, breakage of "new" or "recent" Firefoxes should "old" versions sync is *not* acceptable.
+* Because such evolution should be rare, we do not want to set an up-front policy about locking out
+  "old" versions just because they might have a problem in the future. That is, we want to avoid
+  a policy that dictates versions more than (say) 2 years old will break when syncing "just in case"
```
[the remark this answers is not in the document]
mhammond: I actually don't like "decision drivers" because I don't quite understand what it means :) If I renamed this to simply "requirements" (which I do :) then I think this para still belongs there, right? Or are you suggesting that it's misplaced because it's really just stating a non-requirement?

docs/adr/0008-sync-payload-evolution.md
mhammond: That's what I was trying to get at in the bullet point "A complicated system where we started creating new payload and new collections", but yeah, yours goes into more detail, so I added it, thanks!

docs/adr/0008-sync-payload-evolution.md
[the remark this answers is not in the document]
bendk: That seems like a good place to me.

docs/adr/0008-sync-payload-evolution.md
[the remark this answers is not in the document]
bendk: The way I see it "Some degree of breakage for "old" Firefox installations when "new" or "recent" firefoxes sync might be considered acceptable if absolutely necessary." is the requirement, which is why we don't want to have a blanket policy of locking out older versions.  I would either remove the bullet or discuss locking out older versions as in the options section.  But I'm definitely open to keeping it here if you like it.

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,166 @@
...
+where a schema is downloaded dynamically which also allowed the *semantics* (as opposed to simply
+validity) of new fields to be described.
+
+### Consider the sync payloads frozen and never change them.
+
+A process where payloads are frozen was rejected because:
+
+* The most naive approach here would not meet the needs of Firefox in the future.
+
+* A complicated system where we started creating new payload and new collections
```
linabutler, line 137: nitpick (editorial):

```suggestion
* A complicated system where we started creating new payloads and new collections
```

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Use separate collections for new data
+
+We could store the new data in a separate collection. For example define a
+bookmarks2 collection where each record has the same guid as one in bookmarks alongside any new fields.
+Newer clients use both collections to sync.
+
+The pros and cons:
+
+* Good, because it allows newer clients to sync new data without affecting recent or older clients
```
linabutler, line 152: nitpick (editorial):

```suggestion
* Good, because it allows "new" clients to sync new data without affecting "recent" or "old" clients
```

linabutler (approved):
Looks wonderful, thanks for capturing this @mhammond! 🎉

docs/adr/0008-sync-payload-evolution.md
```diff
@@ -0,0 +1,166 @@
...
+### A backwards compatible schema policy
+
+A summary of this option is a policy by which:
+
+* All sync engines should identify where in their payload new fields are likely to be added in the
+  future, and when syncing these records, arrange to persist any fields from the payload which it
+  does not understand. The next time that engine needs to upload that record to the storage server,
+  it must arrange to add all such "unknown" fields back into the payload.
+
+* Different engines must identify different locations where this might happen. For example, the
```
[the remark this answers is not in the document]
mhammond: [this is replying to both comments above]

I'd like to see some room for best judgement here though - eg, as written, it sounds like we should also persist unknown fields in the top-level of an "address" payload, even though I believe the design is explicitly that new fields go in "entry" so we should never find them in the top-level? Ditto for tabs - as written, we should (a) persist them for tabs too, even though theory are only written by the owner, but must also persist them in the top of the payload, the `tabs` array, and `urlHistory` for each tab.

Another example in our future might be passkeys. IIUC, they are likely to be stored as a sub-object, and each site can have any number of passkeys associated with it. It's not clear to me now that we must support unknown fields in that sub-object, but if we make this section too prescriptive then we will be forcing it even if the use-case ends up not really existing. IOW, I think it's something that the implementers of passkeys need to *consider* when implementing the change, but not something this document should *insist* they do against their judgement at the time.

WDYT?

mhammond:
Folded all the comments into [bug 1831630](https://bugzilla.mozilla.org/show_bug.cgi?id=1831630) as we decided a better place to land this is m-c.

PR #5438, opened 2023-03-21
Bookmarks sync evolution

bendk:

  - Note:

### DIFF
--- a/components/places/sql/create_shared_schema.sql
+++ b/components/places/sql/create_shared_schema.sql
@@ -196,7 +196,9 @@ CREATE TABLE IF NOT EXISTS moz_bookmarks_synced(
     loadInSidebar BOOLEAN,
     smartBookmarkName TEXT,
     feedURL TEXT,
-    siteURL TEXT
+    siteURL TEXT,
+    -- All unknown fields from the server record, encoded as a JSON object.
+    unknownFields TEXT
 );
 
 CREATE INDEX IF NOT EXISTS moz_bookmarks_synced_urls ON moz_bookmarks_synced(placeId);
--- a/components/places/src/bookmark_sync/incoming.rs
+++ b/components/places/src/bookmark_sync/incoming.rs
@@ -2,7 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use super::record::BookmarkRecordId;
+use super::record::{
+    BookmarkItemRecord, BookmarkRecord, BookmarkRecordId, FolderRecord, LivemarkRecord,
+    QueryRecord, SeparatorRecord,
+};
 use super::{SyncedBookmarkKind, SyncedBookmarkValidity};
 use crate::error::*;
 use crate::storage::{
@@ -35,10 +38,11 @@ impl<'a> IncomingApplicator<'a> {
     }
 
     pub fn apply_bso(&self, record: IncomingBso) -> Result<()> {
-        // This is a little odd - we need to go via serde_json::Value so we can
-        // work out what we want to deserialize it as.
         let timestamp = record.envelope.modified;
-        let json_content = record.into_content::<serde_json::Value>();
+        let mut validity = SyncedBookmarkValidity::Valid;
+        let json_content = record.into_content_with_fixup::<BookmarkItemRecord>(|json| {
+            validity = fixup_bookmark_json(json)
+        });
         match json_content.kind {
             IncomingKind::Tombstone => {
                 self.store_incoming_tombstone(
@@ -46,23 +50,23 @@ impl<'a> IncomingApplicator<'a> {
                     BookmarkRecordId::from_payload_id(json_content.envelope.id.clone()).as_guid(),
                 )?;
             }
-            IncomingKind::Content(value) => {
-                match value["type"].as_str() {
-                    Some("bookmark") => self.store_incoming_bookmark(timestamp, &value)?,
-                    Some("query") => self.store_incoming_query(timestamp, &value)?,
-                    Some("folder") => self.store_incoming_folder(timestamp, &value)?,
-                    Some("livemark") => self.store_incoming_livemark(timestamp, &value)?,
-                    Some("separator") => self.store_incoming_sep(timestamp, &value)?,
-                    _ => {
-                        // We support all known bookmark types, so this is probably just malformed.
-                        error_support::report_error!(
-                            "unknown-bookmark-type",
-                            "Unknown bookmark type: {}",
-                            value["type"]
-                        );
-                    }
-                };
-            }
+            IncomingKind::Content(item) => match item {
+                BookmarkItemRecord::Bookmark(b) => {
+                    self.store_incoming_bookmark(timestamp, &b, validity)?
+                }
+                BookmarkItemRecord::Query(q) => {
+                    self.store_incoming_query(timestamp, &q, validity)?
+                }
+                BookmarkItemRecord::Folder(f) => {
+                    self.store_incoming_folder(timestamp, &f, validity)?
+                }
+                BookmarkItemRecord::Livemark(l) => {
+                    self.store_incoming_livemark(timestamp, &l, validity)?
+                }
+                BookmarkItemRecord::Separator(s) => {
+                    self.store_incoming_sep(timestamp, &s, validity)?
+                }
+            },
             IncomingKind::Malformed => {
                 log::trace!(
                     "skipping malformed bookmark record: {}",
@@ -77,51 +81,13 @@ impl<'a> IncomingApplicator<'a> {
         Ok(())
     }
 
-    fn store_incoming_bookmark(&self, modified: ServerTimestamp, b: &JsonValue) -> Result<()> {
-        let mut validity = SyncedBookmarkValidity::Valid;
-
-        let record_id = unpack_id("id", b)?;
-        let parent_record_id = unpack_optional_id("parentid", b);
-        let date_added = unpack_optional_i64("dateAdded", b, &mut validity);
-        let title = unpack_optional_str("title", b, &mut validity);
-        let keyword = unpack_optional_keyword("keyword", b, &mut validity);
-
-        let raw_tags = &b["tags"];
-        let tags = if let Some(array) = raw_tags.as_array() {
-            let mut seen = HashSet::with_capacity(array.len());
-            for v in array {
-                if let JsonValue::String(s) = v {
-                    let tag = match validate_tag(s) {
-                        ValidatedTag::Invalid(t) => {
-                            log::trace!("Incoming bookmark has invalid tag: {:?}", t);
-                            set_reupload(&mut validity);
-                            continue;
-                        }
-                        ValidatedTag::Normalized(t) => {
-                            set_reupload(&mut validity);
-                            t
-                        }
-                        ValidatedTag::Original(t) => t,
-                    };
-                    if !seen.insert(tag) {
-                        log::trace!("Incoming bookmark has duplicate tag: {:?}", tag);
-                        set_reupload(&mut validity);
-                    }
-                } else {
-                    log::trace!("Incoming bookmark has unexpected tag: {:?}", v);
-                    set_reupload(&mut validity);
-                }
-            }
-            seen
-        } else {
-            if !raw_tags.is_array() {
-                log::trace!("Incoming bookmark has unexpected tags list: {:?}", raw_tags);
-            }
-            HashSet::new()
-        };
-
-        let url = unpack_optional_str("bmkUri", b, &mut validity);
-        let url = match self.maybe_store_href(url) {
+    fn store_incoming_bookmark(
+        &self,
+        modified: ServerTimestamp,
+        b: &BookmarkRecord,
+        mut validity: SyncedBookmarkValidity,
+    ) -> Result<()> {
+        let url = match self.maybe_store_href(b.url.as_deref()) {
             Ok(u) => Some(String::from(u)),
             Err(e) => {
                 log::warn!("Incoming bookmark has an invalid URL: {:?}", e);
@@ -130,12 +96,13 @@ impl<'a> IncomingApplicator<'a> {
                 None
             }
         };
+        let unknown_fields = serde_json::to_string(&b.unknown_fields)?;
 
         self.db.execute_cached(
             r#"REPLACE INTO moz_bookmarks_synced(guid, parentGuid, serverModified, needsMerge, kind,
-                                                 dateAdded, title, keyword, validity, placeId)
+                                                 dateAdded, title, keyword, validity, unknownFields, placeId)
                VALUES(:guid, :parentGuid, :serverModified, 1, :kind,
-                      :dateAdded, NULLIF(:title, ""), :keyword, :validity,
+                      :dateAdded, NULLIF(:title, ""), :keyword, :validity, :unknownFields,
                       CASE WHEN :url ISNULL
                       THEN NULL
                       ELSE (SELECT id FROM moz_places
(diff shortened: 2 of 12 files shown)


components/places/src/bookmark_sync/tests/payload_evolution.rs
```diff
@@ -0,0 +1,358 @@
...
+                "Animals",
+                vec!["A", "B"],
+                vec![("device_type", "phone"), ("color", "blue")],
+            ),
+            remote_separator(guid('c'), 2, vec![("device_type", "desktop")]),
+            // We re-upload query and livemark records after a reset, so test them here
+            // ... except for some reason apply_incoming() flags these as deleted and wants to sync
+            // outgoing tombstones for them
+            // remote_query(guid('d'), "dog search", "place:folder=123&excludeItems=1", vec![("device_type", "desktop")]),
+            // remote_livemark(guid('e'), vec![("device_type", "desktop")]),
```
bendk, line 100: I really wanted to test this case, but I couldn't figure out this issue.  I'm hoping @mhammond knows what's going on here.

linabutler: You actually uncovered an intentional behavior! Livemarks were [removed in Firefox Desktop 64](https://support.mozilla.org/en-US/kb/feed-reader-replacements-firefox), and none of our other platforms ever supported them, so Sync deletes them.

There's a longer saga behind livemarks if you're curious! 😄 Forever ago, Desktop's Places stored them as folders with special annotations—the feed and site URLs—and real bookmark children for each feed item. Sync had a special record type for them, too, but otherwise treated them as regular folders. That scheme caused write and sync churn, though, so livemarks became empty folders with annotations, and the items were populated dynamically in the UI.

Some of our power users loved livemarks, but they never became a super popular feature, and couldn't justify their maintenance burden—so the Desktop team decided to remove them in 64. That release had a migration step that exported livemarks to OPML, then deleted them from Places.

We didn't want Sync to resurrect them—or turn them into folders—so we also added logic in [the merger](https://github.com/mozilla/dogear/pull/37) and [Desktop's mirror](https://bugzilla.mozilla.org/show_bug.cgi?id=1546035) to delete them.

TL;DR: Rust Places inherited the logic to delete livemarks from Desktop. Even though Fenix and iOS never supported livemarks, they'll still upload tombstones if they ever see one.

Since they're not a thing anymore, I think it's also totally OK not to support unknown fields for them—it doesn't hurt to, but it is extra code that none of our platforms should ever invoke.

I hope that helps!

bendk: That's definitely helpful, thanks for the context here.

Do you know if queries are in the same category or are they still supported?

components/places/src/bookmark_sync/tests/payload_evolution.rs
```diff
@@ -0,0 +1,358 @@
...
+                guid('b'),
+                "Animals",
+                vec!["A", "B"],
+                vec![("device_type", "phone"), ("color", "blue")],
+            ),
+            remote_separator(guid('c'), 2, vec![("device_type", "desktop")]),
+            // We re-upload query and livemark records after a reset, so test them here
+            // ... except for some reason apply_incoming() flags these as deleted and wants to sync
+            // outgoing tombstones for them
+            // remote_query(guid('d'), "dog search", "place:folder=123&excludeItems=1", vec![("device_type", "desktop")]),
```
linabutler, line 99: Here's the test fix for queries (plus plumbing `store_incoming_query` / `BookmarkItemRecord::unknown_fields`):

```suggestion
            remote_item(
                json!({
                    "id": "unfiled",
                    "title": "Unfiled Bookmarks",
                    "type": "folder",
                    "parentid": "places",
                    "parentName": "places",
                    "dateAdded": 0,
                    "children": vec![guid('d')],
                }),
                vec![],
            ),
            remote_query(guid('d'), "dog search", "place:folder=123&excludeItems=1", vec![("device_type", "desktop")]),
```

I'll post the explanation in a bit!

linabutler: Queries are still supported in Desktop Places. Fenix and iOS don't show them in the UI, but Rust Places still stores and round-trips them. There are three special cases in Sync's handling of queries:

1. Rewriting tag folder queries to reference tag names instead of folders.
2. Rewriting non-tag `folder` queries to add `excludeItems`.
3. Removing suspected "orphaned left pane queries".

I think your test unintentionally triggered (3), which is why it was uploading a tombstone.

What's a left pane query? If you open Bookmarks > Manage Bookmarks in Firefox Desktop, you'll see a two-pane window, with "All Bookmarks", "Bookmarks Toolbar", "Bookmarks Menu", "Other Bookmarks", and "Mobile Bookmarks" in the left pane. These left-pane queries show the contents of those roots in Places. Since 2017-ish, these queries are implemented entirely in the UI.

Before 2017, these left-pane queries were stored in Places as real query items, as children and grandchildren of the "left pane root". This root was hidden from the UI, and not synced. Before our Desktop bookmarks tracker and mirror rewrites, though, it was possible in some cases for Sync to observe changes to those left pane queries–like localized titles, or query URLs, changing between releases—and upload records for them. Crucially, it never uploaded the left pane root, so receiving devices would see those left pane queries, decide they were orphaned, and move them into `unfiled`.

This caused all sorts of issues in the UI. In the worst case, expanding one of those orphaned left pane queries at the wrong time would cause an infinite loop. We didn't have a good way of identifying these queries—their URLs and titles could change between releases, and the left pane root had a random GUID per device—so we used a [heuristic](https://github.com/mozilla/dogear/blob/f83e6db9640332bb4d8b9689740a7db68ba087f3/src/tree.rs#L1583-L1584). **If we saw any orphaned query—where it had a `parentid` that either didn't exist, or didn't mention the query in its `children` array—we'd delete it.**

TL;DR: Your test was accidentally creating one of these orphaned queries: `remote_query()` returns a record with `"parentid: "unfiled"`, but `initial_remote_records` didn't have a record for `unfiled`, so Dogear assumed it was an orphaned left pane query, and deleted it.

Your test didn't trigger the other two special cases, but here's the explanation for completeness.

(1) and (2) have the same underlying cause: the `folder` param references the SQLite row ID, which isn't portable across devices. When a query references a nonexistent `folder` ID, it'll return _all_ bookmarks in the database, which breaks the UI in the same way as (3).

To fix (1), Sync rewrites tag queries (with `type=7`) from `folder=<tag folder ID>` to `tag=<tag name>`. Desktop Places added support for `tag=` params in Firefox 61, and Sync query records store the tag name in the `folderName` in their payloads, so it has enough information to fix these up.

Sync doesn't have enough information to fix up (2)—the receiving device doesn't know what folder ID the query wants—so it compromises by adding `excludeItems=1` to the query. The device that originally uploaded the query has to rewrite it from `folder=<folder ID> to `parent=<folder GUID>`, but, in the meantime, adding `excludeItems=1` will make sure the query doesn't return anything, to keep it from breaking the UI.

bendk: Wow, that history is totally wild.  I copied your code and was able to get the remote_query test to fail properly and then was able to fix it.  I think payload evolution is working for all bookmark types other than livemarks, thanks for all the help!

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -38,31 +40,21 @@ impl<'a> IncomingApplicator<'a> {
         // This is a little odd - we need to go via serde_json::Value so we can
         // work out what we want to deserialize it as.
         let timestamp = record.envelope.modified;
-        let json_content = record.into_content::<serde_json::Value>();
+        let json_content = record.into_content::<BookmarkItemRecord>();
```
linabutler, line 43: question: Will this still parse and flag records with invalid (wrong type or missing) properties, or will those now fall through to the `IncomingKind::Malformed` case? And, is that a case we should still be worried about?

If memory serves right, we switched to treating records as dynamically-checked `JsonValue` bags in #1661, which is also how Desktop behaves. We don't fail the entire sync anymore, but I'm wondering if we could end up in a situation where Desktop would fix up (replace / reupload) an invalid record, but Rust Places would treat that record as malformed and skip it.

Maybe keeping an eye on the `malformed-incoming-bookmark` probe is enough...

components/places/src/storage/bookmarks.rs
```diff
@@ -907,13 +907,9 @@ fn get_raw_bookmarks_for_url(db: &PlacesDb, url: &Url) -> Result<Vec<RawBookmark
 }
 
 fn reset_in_tx(db: &PlacesDb, assoc: &EngineSyncAssociation) -> Result<()> {
-    // Remove all synced bookmarks and pending tombstones, and mark all
-    // local bookmarks as new.
+    // Mark all local bookmarks as new so that we re-upload them on the next sync
     db.execute_batch(&format!(
-        "DELETE FROM moz_bookmarks_synced;
-
-        DELETE FROM moz_bookmarks_deleted;
```
linabutler, line 915: question: How come we want to carry over unknown fields between resets? Wouldn't re-downloading all records on the first sync after a reset also download those unknown fields?

mhammond: One argument for keeping them is just that we might be the first device to show up after a reset, right?

But although I haven't looked much at this patch, it does seem a red-flag that a reset isn't killing the mirror.

mhammond: (I haven't put as much thought into resets as I should. Eg, I don't think autofill will keep them because it relies on what's stored in the mirror and deletes it on reset). It would seem odd for some engines to carry them and not others though.)

components/places/src/bookmark_sync/incoming.rs
bendk: I mentioned this in [the commit](https://github.com/mozilla/application-services/pull/5438/commits/8e1c752d08e182df52585c063b47f042414f1e68), but I forgot to call it out in the PR.  I think you're analysis is correct and it means we will skip over records with incorrect types / missing required fields.  I think that's okay, given the payload evolution discussions, but I wonder what other's think.

BTW, I also forgot to call out Mark's excellent new sync15 code that makes this all work.  It was a real joy doing that refactor.

components/places/src/storage/bookmarks.rs
bendk: Maybe we need to define what a reset represents.  If reset means "forget the server data", then I think it should also mean that we forget the unknown fields.  If it means "reset your timestamps/counters so you do a full sync next time", then I think it's okay to keep the mirror data around.

components/places/src/bookmark_sync/incoming.rs
mhammond: > it means we will skip over records with incorrect types / missing required fields.

I'm a little torn here - bookmarks is somewhat special in that it has a fuzzy concept of requiring a kind of atomicity. Eg, [allow_failed](https://github.com/mozilla/application-services/blob/main/components/sync15/src/client/request.rs#L188) when posting exists only for bookmarks. The main concern is the possibility of creating an invalid tree - eg, if the POST creating a parent node failed, we don't want to allow the children. The same concern applies for incoming records - if the parent is malformed, we really can't skip it. So the decision to just give up was kinda explicit, but mainly in a "let's not think about this now" kind of way.

dogear *should* prevent us actually having a bad tree locally though - but if it really is just this device that considers the record malformed and dogear reparents items because of it, then that contagion ends up spreading to other devices too.

All that said though, one malformed record preventing bookmarks syncing *at all* on this device certainly doesn't seem an ideal outcome either. So there's no correct answer here, it's all a tradeoff of what we think might go wrong.

I'd love to hear what others think here.

components/places/src/storage/bookmarks.rs
mhammond: It means "we have disconnected from the server", either explicitly due to user action, or because the server state changed (eg, password reset). In the former case, we may never reconnect, so keeping old server data around forever just smells wrong from a privacy POV - their DB still having bookmarks deleted years sounds quite bad. In a server-state change, then we are going to reconnect soon, but that differentiation doesn't currently exist. I guess we could consider looking into adding it.

components/places/src/bookmark_sync/incoming.rs
mhammond: *sigh* - I really should not try and comment without more context. I just deleted the above comment which totally missed the point - we do already skip malformed records as Lina noted. I'm actually not sure that was the right call for reasons noted by Lina and in my deleted comment. I wonder if we can change the "unpack" functions to take the json as `mut &` and them call [take](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.take), then any non-null values left are unknown?

(Or we assert that the last 5 years *has* cleaned up things such that we can trust data better, however issues like https://mozilla.sentry.io/issues/3796658061 worry me - sentry hates me, but we see some reports with text `Invalid incoming T: invalid type: string "1676479234.51", expected i64` as well as null confusion in `Invalid incoming T: invalid type: null, expected a string`. We can be reasonably sure that the `i64` problem is *not* bookmarks, because `unpack_optional_i64` handles that case - so those kinds of records would regress here)

bendk: Actually, I don't think that will work for this purpose since the functions were doing much more complicated things like trying to convert strings to ints and also marking when a record needs to be reuploaded.

bendk:
Pushed a new version of this based on the conversation here.  There's two main changes:

- Based on the slack convo, I removed the commit to avoid deleting data from the mirror table and removed the test for round-tripping data after a reset.
- I went back to the older code that has workarounds for mistyped data.  I'd love to know what others think about this, I'm not totally sure about it myself.

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -521,83 +471,122 @@ impl<'a> IncomingApplicator<'a> {
...
+///
+/// This is extra important since bookmarks form a tree.  If a parent node is invalid, then we will
+/// have issues tring to merge its children.
+fn fixup_bookmark_json(json: &mut JsonValue) -> SyncedBookmarkValidity {
+    let mut validity = SyncedBookmarkValidity::Valid;
+    // the json value should always be on object, if not don't try to do any fixups.  The result will
+    // be that into_content_with_fixup() returns an IncomingContent with IncomingKind::Malformed.
+    if let JsonValue::Object(ref mut obj) = json {
+        obj.entry("parentid")
+            .and_modify(|v| fixup_optional_str(v, &mut validity));
```
bendk, line 485: Before there was a an `unpack_optional_id` that worked like `unpack_optional_str`, but didn't modify the validity.  I don't see why we wouldn't want to re-upload when the `parentid` field is wrong, so I made them use the same codepath.

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -521,83 +471,122 @@ impl<'a> IncomingApplicator<'a> {
...
+                        log::trace!("Incoming bookmark has duplicate tag: {:?}", tag);
+                        set_reupload(validity);
+                    }
+                } else {
+                    log::trace!("Incoming bookmark has unexpected tag: {:?}", v);
+                    set_reupload(validity);
+                }
+            }
+            if valid_tags.len() != tags.len() {
+                *val = JsonValue::Array(valid_tags.into_iter().map(JsonValue::from).collect());
```
linabutler, line 583: nitpick: Let's do this unconditionally; I think `valid_tags` might still have normalized tags even when `valid_tags.len() == tags.len()`?

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -521,83 +471,122 @@ impl<'a> IncomingApplicator<'a> {
...
-    match &data[key] {
-        JsonValue::String(ref s) => {
-            // Like Desktop, we don't reupload if a keyword has leading or
-            // trailing whitespace, or isn't lowercase.
-            let k = s.trim();
-            if k.is_empty() {
-                None
-            } else {
-                Some(k.to_lowercase())
+fn fixup_optional_tags(val: &mut JsonValue, validity: &mut SyncedBookmarkValidity) {
```
linabutler, line 555: Thanks for untangling this logic from `store_incoming_bookmark`! 🏆

components/places/src/bookmark_sync/tests/payload_evolution.rs
```diff
@@ -0,0 +1,318 @@
...
+        )],
+        outgoing_unknown_fields: vec![
+            (guid('a'), vec![("device_type", "mini-phone")]),
+            (guid('b'), vec![("device_type", "desktop")]),
+        ],
+    }
+    .run()
+}
+
+struct RoundtripTest {
```
linabutler, line 103: ❤️ table-driven tests!

linabutler (approved):
LGTM!

components/places/src/bookmark_sync/tests/payload_evolution.rs
bendk: Yeah, they're pretty great.  I saw a few in the app-services code when I first came on and copied the style.  The more I write the more I like them.

components/places/sql/create_shared_schema.sql
```diff
@@ -196,7 +196,9 @@ CREATE TABLE IF NOT EXISTS moz_bookmarks_synced(
     loadInSidebar BOOLEAN,
     smartBookmarkName TEXT,
     feedURL TEXT,
-    siteURL TEXT
+    siteURL TEXT,
+    -- All unknown fields from the server record, encoded as a JSON object.
+    unknownFields TEXT
```
mhammond, line 201: This needs a db migration, right?

components/places/src/bookmark_sync/incoming.rs
```diff
@@ -130,12 +96,13 @@ impl<'a> IncomingApplicator<'a> {
                 None
             }
         };
+        let unknown_fields = serde_json::to_string(&b.unknown_fields)?;
```
mhammond, line 99: How do you feel about writing a null if the map is empty? (And ditto for other items below)

mhammond (commented):
This is a great refactor, thanks. I think storing `NULL` seems optimal, but not critical, however the DB migration is. Probably worth running the `places_utils` example too to ensure the upgrade works and that syncing still works after.

PR #5441, opened 2023-03-24
Narrow the gap between a SyncEngine and a BridgedEngine.

mhammond:
A SyncEngine was implemented when syncing was first implemented in Rust. A
few years later, a BridgedEngine was created when we needed to interact
with desktop's sync.

SyncEngine didn't really capture the semantics very well though, whereas
BridgedEngine was a much better fit. It used a "staging" model and better
supported having incoming and outgoing records being "batched".

This patch:
* Re-defines SyncEngine in a way that's much closer to BridgedEngine - it
  doesn't quite get all the way, as that would require changes to Desktop,
  but it's very close!
* Keeps the old SyncEngine as a LegacySyncEngine, and defines a simple
  trait so old engines can support the new world order with very little
  effort. Only 2 engines actually leverage this though, so it's not clear
  it's particularly worthwhile.
* Defines a BridgedEngineAdaptor so that code which implements a SyncEngine
  can trivially also implement BridgedEngine. The tabs crate leverages this.
* Places moves to the new SyncEngine, primary because that engine already
  used a staging model, and to check the new SyncEngine trait actually
  captured the semantics it needed to.

At this stage, I'm primarily after feedback on the sync15 changes - does the new trait make sense? Do the adapters make sense? Is this approach worthwhile? Once there's general agreement there, looking at the individual crates makes more sense. Note also that I'll be on PTO for most of April and have no intention of trying get this landed before I go, so there's no hurry.


components/places/src/history_sync/engine.rs
```diff
@@ -83,40 +83,60 @@ impl SyncEngine for HistorySyncEngine {
...
     }
 
-    fn sync_finished(
-        &self,
-        new_timestamp: ServerTimestamp,
-        records_synced: Vec<Guid>,
-    ) -> anyhow::Result<()> {
-        do_sync_finished(&self.db.lock(), new_timestamp, records_synced)?;
+    fn set_uploaded(&self, new_timestamp: ServerTimestamp, ids: Vec<Guid>) -> anyhow::Result<()> {
+        // XXX - this is WRONG!
```
mhammond, line 116: I can't remember why I wrote that 😅

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -87,8 +90,118 @@ pub trait BridgedEngine: Send + Sync {
...
 
-// TODO: We should replace this with OutgoingChangeset to reduce the number
-// of types engines need to deal with.
+// This is an adaptor trait - the idea is that engines can implement this
+// trait along with SyncEngine and get a BridgedEngine for free. It's temporary
+// so we can land this trait without needing to update desktop.
+// Longer term, we should remove both this trait and BridgedEngine entirely, sucking up
+// the breaking change for desktop. The main blocker to this is moving desktop away
+// from the explicit timestamp handling and moving closer to the `get_collection_request`
+// model.
```
bendk, line 99: My initial impression is that I like this trait and the underlying plan.  It's also nice that it shows the difference between the two traits.

components/sync15/src/engine/legacy_engine.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn get_collection_requests(
+        &self,
+        server_timestamp: ServerTimestamp,
+    ) -> Result<Vec<CollectionRequest>>;
+    fn get_sync_assoc(&self) -> Result<EngineSyncAssociation>;
+    fn reset(&self, assoc: &EngineSyncAssociation) -> Result<()>;
+    fn wipe(&self) -> Result<()>;
+
+    // the magic
+    fn get_legacy_engine_state(&self) -> &LegacySyncEngineState;
```
bendk, line 97: What if this returns a `&mut`, since we're going to be storing the staged records in the state?  Would that remove the need for the `RefCell<>` wrappers?

bendk (approved):
I like this change and the underlying plan for it.  The adapters make sense to me, they help identify what the differences are between all the traits and should make the rest of the changes relatively smooth.

One thing that's unfortunate is that so much other code needs to change, even with the adapters.  But I've gotten comfortable with trusting Rust + our tests to ensure that there's no behavior change.  BTW, is there any behavior changes?  I can't see any, but I'm not sure.

components/sync15/src/engine/legacy_engine.rs
mhammond: that turns into this function needing a `&mut self` which explodes into everything needing it. We could juggle where the `RefCell`s are, but I wonder if we should lean into the engine traits taking `&mut self`? None of this stuff needs shared references. I might play with that in a different PR :)

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -87,8 +90,118 @@ pub trait BridgedEngine: Send + Sync {
...
+        let mut telem = telemetry::Engine::new(engine.collection_name());
+        engine.stage_incoming(incoming_records, &mut telem)
+    }
+
+    fn apply(&self) -> Result<ApplyResults> {
+        let engine = self.engine();
+        let mut telem = telemetry::Engine::new(engine.collection_name());
+        // we tell it to apply with an invalid timestamp then update it.
+        // this isn't atomic and we probably shouldn't do it, but to get a feel...
+        let records = engine.apply(ServerTimestamp::from_millis(0), &mut telem)?;
```
linabutler, line 175: Can you remind me why this isn't atomic, please? I paged this out of my brain 😅

It is wrong, but the only engines we have to worry about are tabs (and Extension Storage, maybe? 🤔), right? And it looks like its `apply` doesn't use the `timestamp`, so I think it's okay to leave in until we migrate!

components/sync15/src/client/sync.rs
```diff
@@ -40,54 +40,72 @@ pub fn synchronize_with_clients_engine(
...
+            let incoming = super::fetch_incoming(client, &coll_state, collection_request)?;
+            log::info!("Downloaded {} remote changes", incoming.len());
+            engine.stage_incoming(incoming, telem_engine)?;
+            interruptee.err_if_interrupted()?;
+        }
     };
 
-    let outgoing = engine.apply_incoming(incoming, telem_engine)?;
+    // Should probably add a new `fetch_outgoing()` and only call `apply()` when something was
+    // staged?
```
linabutler, line 74: Hmmm, what's this comment referring to? Do you mean skip calling `apply()` if we haven't staged any incoming changes?

components/sync15/src/client/sync.rs
```diff
@@ -40,54 +40,72 @@ pub fn synchronize_with_clients_engine(
...
-    log::info!("Uploading {} outgoing changes", outgoing.changes.len());
-    let upload_info =
-        CollectionUpdate::new_from_changeset(client, &coll_state, outgoing, fully_atomic)?
-            .upload()?;
+
+    // XXX - this upload strategy is buggy due to batching. With enough records, we will commit
+    // 2 batches on the server. If the second fails, we get an Err<> here, so can't tell the
+    // engine about the successful server batch commit.
+    // Most stuff below should be called per-batch rather than at the successful end of all
+    // batches, but that's not trivial.
```
linabutler, line 83: Ooof...yeah, I think the lack of per-batch calls (and reading everything in one request) is an artifact of us implementing batches "transparently" on Desktop way back when, and carrying that through all our implementations.

In case of the first batch failing and the second succeeding, I think we decided that not advancing the last sync time after the successful batch, and reading our own changes on the next sync, was good enough back then...but it would sure be nice to handle them properly.

linabutler (commented):
Still working through this, but wanted to get an early round of comments up at the end of my day—I'll go through the rest tomorrow! This is really exciting, thanks @mhammond! 🥇

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -87,8 +90,118 @@ pub trait BridgedEngine: Send + Sync {
...
+                coll: sync_id.into(),
+            };
+            engine.reset(&EngineSyncAssociation::Connected(new_coll_ids))?;
+        }
+        Ok(sync_id.to_string())
+    }
+
+    fn prepare_for_sync(&self, client_data: &str) -> Result<()> {
+        // unwrap here is unfortunate, but can hopefully go away if we can
+        // start using the ClientData type instead of the string.
```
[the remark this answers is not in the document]
mhammond: The `client_data` param is only used for tabs, however, if we assume the sync engines live longer than a single sync - which is sadly *not* true for bridged engines created by desktop - then this method is also the only way those engines know a new sync is starting.

I do agree we should try and communicate client_data to tabs differently though!

components/sync15/src/engine/bridged_engine.rs
```diff
@@ -87,8 +90,118 @@ pub trait BridgedEngine: Send + Sync {
...
+
+    fn sync_id(&self) -> Result<Option<String>> {
+        Ok(match self.engine().get_sync_assoc()? {
+            EngineSyncAssociation::Disconnected => None,
+            EngineSyncAssociation::Connected(c) => Some(c.coll.into()),
+        })
+    }
+
+    fn reset_sync_id(&self) -> Result<String> {
+        let global = Guid::empty();
```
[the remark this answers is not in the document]
mhammond: > Desktop's base SyncEngine really wants to be in charge of making the request; 

When I looked at this, I felt it wasn't as bad as I remember - it's really just [these few lines](https://searchfox.org/mozilla-central/rev/445a5ab684b73eb56d807d0f3b2fabcc85a7c3dd/services/sync/modules/engines.sys.mjs#1207-1232) in desktop which still delegate to the engine for many of the params.

> Maybe a comment that the global sync ID is irrelevant here
good call!

components/sync15/src/client/sync.rs
mhammond: yeah, exactly - I think the new shape of the trait is now capable of expressing this correctly, so it's just a matter of having sync15 actually do it.

linabutler (approved):
This looks amazing, thanks so much @mhammond—and sorry for the delay in reviewing! I really like the shape of the new trait, and exciting to see the merge get closer! 🏆

components/places/src/history_sync/engine.rs
```diff
@@ -64,6 +60,9 @@ fn do_sync_finished(
 // Short-lived struct that's constructed each sync
 pub struct HistorySyncEngine {
     pub db: Arc<SharedPlacesDb>,
+    // We should stage these in a temp table! For now though we just hold them
+    // in memory.
+    pub staged_incoming: RefCell<Vec<IncomingBso>>,
```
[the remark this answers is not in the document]
mhammond: That's a really good point, and I admit I forgot to come back to this - while I think the temp storage in the `RefCell<>` probably isn't actually any worse then the current behaviour (the incoming records are held in memory for just marginally longer then they are now), it's still a bit ugly, and I agree with your assessment. To really tip me over the line, we (a) only ever do 5k records and (b) don't actually batch incoming records yet, so what could possibly go wrong 😅 . So yeah, I update the PR to apply them as they come in.

components/sync15/src/client/sync.rs
mhammond: > My thinking is, if the engine stages incoming records in a temporary table, and is interrupted in the middle of apply

In that scenario I'd expect the engine should drop and create the temp tables as part of each sync.

> or if apply ends up "fixing up" any local items that weren't necessarily part of the last incoming batch

Do you mind expanding on this? You mean an engine might want to fixup a local item we've already uploaded in the past, which isn't currently marked as outgoing and wasn't incoming in this sync? I'm struggling to think where that might happen, but might be missing your point entirely.

Regardless, I updated that comment to point at this discussion :)

Issue #5414, opened 2023-03-03
Help building on Fedora


jo:
I'm trying to build the project - at least running the test suite - on Fedora. I followed [the steps](https://github.com/mozilla/application-services/blob/main/docs/building.md#linux) and I get a `NoNssDir` error:

```sh
$ RUST_BACKTRACE=1 cargo test
   Compiling this and that...
   Compiling num-traits v0.2.15
error: failed to run custom build command for `nss_sys v0.1.0 (/home/jo/git/application-services/components/support/rc_crypto/nss/nss_sys)`

Caused by:
  process didn't exit successfully: `/home/jo/git/application-services/target/debug/build/nss_sys-eddc484f54a37c61/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=NSS_DIR

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NoNssDir', components/support/rc_crypto/nss/nss_sys/build.rs:6:34
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:575:5
     1: core::panicking::panic_fmt
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:65:14
     2: core::result::unwrap_failed
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1791:5
     3: core::result::Result<T,E>::unwrap
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1113:23
     4: build_script_build::main
               at ./build.rs:6:5
     5: core::ops::function::FnOnce::call_once
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ops/function.rs:251:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I have searched through the NSS docs but was not able to find any hint.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3634)

jo:
Hi @tarikeshaq thank you very much, this helped me coming way further!

[...]

PR #5443, opened 2023-03-27
Remove dependency on the clipboard crate for the tabs example

mhammond:
As noted in #5414, the fact the tabs example relies on the `clipboard` crate is a bit of a footgun and it keeps shooting me, so I think we should just kill it. Instead of trying to read from the clipboard the user instead pastes it to a file and gives the filename to the example.


lougeniaC64 (approved):
Tested both the new dummy tab path and the file-backed local state path and they worked for me! 🚢

PR #5446, opened 2023-03-27
Fixup intermittent test failures in the event_store

jhugman:

  - Note:

### DIFF
--- a/components/nimbus/src/tests/test_behavior.rs
+++ b/components/nimbus/src/tests/test_behavior.rs
@@ -780,7 +780,7 @@ mod multi_interval_counter_tests {
 
 #[cfg(test)]
 mod event_store_tests {
-    use chrono::Duration;
+    use chrono::{Datelike, Duration};
 
     use super::*;
 
@@ -1232,7 +1232,10 @@ mod event_store_tests {
     fn record_past_event_should_reflect_in_queries() -> Result<()> {
         let mut store: EventStore = Default::default();
         let event_id = "app_launch";
-        store.record_past_event(5, event_id, None, Duration::days(1))?;
+        let delta = Duration::days(1);
+        let now = Utc::now();
+        let then = now - delta;
+        store.record_past_event(5, event_id, None, delta)?;
 
         // eventSum, for all intervals
         assert_eq!(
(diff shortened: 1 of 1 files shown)


jeddai:
Lol I just had this test fail on me locally haha. Thanks for the fix!

jeddai:
@bendk noted it's failing things on `main`, I'm going to merge it now

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -1306,11 +1317,7 @@ mod event_store_tests {
             store.query(event_id, Interval::Weeks, 52, 0, EventQueryType::LastSeen)?
         );
         assert_eq!(
-            0.0,
-            store.query(event_id, Interval::Months, 12, 0, EventQueryType::LastSeen)?
-        );
```
jhugman, line 1311: This is the test that was broken.

It was failing because [when the test was written](https://github.com/mozilla/application-services/pull/5431), 1 week ago and now were both in the same month.

But it's the 27th today and 1 week ago would be 20th of the same month. Is that right?

Yes, because the `event_store` considers 1 month == 28 days.

Yesterday, was the 84th day of the year, so from today, for a week, this test would fail.

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -1278,8 +1281,13 @@ mod event_store_tests {
...
             store.query(event_id, Interval::Days, 56, 0, EventQueryType::LastSeen)?
         );
+
         assert_eq!(
-            0.0,
+            if now.weekday().number_from_monday() < then.weekday().number_from_monday() {
+                0.0
+            } else {
+                1.0
+            },
```
jhugman, line 1290: This, on reflection, is definitely wrong, and will fail tomorrow.

jeddai: Ah dang

components/nimbus/src/tests/test_behavior.rs
jeddai: Is there a way we could stub out `Utc::now()`? We could set the test up with a static date if the rolling dates prove to be a problem

PR #5448, opened 2023-03-27
Adjust nimbus features and add feature attributes where appropriate

jeddai:
One thing worth noting is this blocks out a bunch of methods and introduces a ton of unused imports at the moment. Planning on fixing those soon. Opening this early to see if CI catches anything I didn't.

https://mozilla-hub.atlassian.net/browse/EXP-3159

  - Note:


jeddai:
Alrighty @jhugman this is officially ready for a review! Lots of commits, sorry, I'll definitely be squashing this when it's merged.

components/nimbus/src/lib.rs
```diff
@@ -658,18 +88,19 @@ pub struct Experiment {
     // but we ignore them because they're for internal use by RemoteSettings.
 }
 
+#[allow(unused)]
```
jeddai, line 91: You'll see a number of these around now, they're required because the `rust-clippy` tests run against both `--no-default-features` and with features. When the `--no-default-features` flag is used, a lot of the stuff in `lib` and other places doesn't get used, resulting in the `rust-clippy` job failing. 

As we build out the Cirrus feature we should be able to remove these (hopefully).

jeddai: One thing I'm curious about that I've not been able to find is if there's a way to group attributes, or have attributes apply to attributes. 

As an example, what I'd like to do here is something like 
```rust
#[allow(unused), cfg(not(feature = "nimbus"))]
```
to note allow unused only when the `nimbus` feature is disabled (this exact syntax is definitely not supported, and the only things related to something like it I've found were closed out and not implemented).

jeddai:
The CI is being consistently wonky wrt crates.io, I'm going to re-run the failing ones tomorrow.

components/nimbus/src/enrollment.rs
```diff
@@ -83,7 +89,7 @@ impl ExperimentEnrollment {
         available_randomization_units: &AvailableRandomizationUnits,
         targeting_attributes: &TargetingAttributes,
         experiment: &Experiment,
-        event_store: Arc<Mutex<EventStore>>,
+        #[cfg(feature = "stateful")] event_store: Arc<Mutex<EventStore>>,
```
jhugman, line 92: Should we land https://github.com/mozilla/application-services/pull/5374 before we land this?

components/nimbus/src/enrollment.rs
```diff
@@ -1015,6 +1037,7 @@ fn map_features(
     map
 }
 
+#[allow(unused)]
```
jhugman, line 1040: I'm guessing that we'll need to start using this again once we start re-organizing `dbcache.rs`: this is the function that takes current enrollments and experiments and outputs the merged feature configs.

components/nimbus/src/tests/test_persistence.rs
```diff
@@ -4,7 +4,7 @@
 
 use crate::enrollment::ExperimentEnrollment;
 use crate::persistence::*;
-use crate::{error::Result, Database, Experiment};
+use crate::{error::Result, persistence::Database, Experiment};
```
jhugman, line 7: Nit: is `Database` included in `persistence::*`?

components/nimbus/src/lib.rs
```diff
@@ -2,620 +2,50 @@
...
 mod matcher;
-pub mod persistence;
 mod sampling;
 mod strings;
+#[cfg(feature = "stateful")]
 mod updating;
+
+pub mod error;
+#[cfg(feature = "stateful")]
+pub mod nimbus;
```
jhugman, line 23: I wonder if this the only thing that should be in this file.

The rest of this current file is schema declarations for interacting with remote settings, so perhaps should be in its own `schema.rs`. WDYT?

components/nimbus/src/enrollment.rs
```diff
@@ -1182,9 +1205,11 @@ pub enum EnrollmentChangeEventType {
     EnrollFailed,
     Disqualification,
     Unenrollment,
+    #[allow(unused)]
     UnenrollFailed,
 }
 
+#[cfg(feature = "stateful")]
```
jhugman, line 1212: These make me think we should separate them out to different modules (either now, or Eventually™, e.g. `enrollment::stateful` which can be imported into `nimbus.rs`/`nimbus_client.rs`.

jhugman (commented):
This is looking very good! I don't think it's ready to land, but I've added some comments on where I think it needs to go next.

Teasing out the event store from the enrollment evolver is the key next step (e.g. https://github.com/mozilla/application-services/pull/5374).

Good job!

components/nimbus/src/tests/test_persistence.rs
jeddai: 🤦

components/nimbus/src/enrollment.rs
jeddai: Yep yep, just not being used now when the stateful feature is disabled.

components/nimbus/src/lib.rs
jeddai: That's a good idea!

components/nimbus/src/enrollment.rs
jeddai: It functions fine like this but you're right, it would definitely be better for readability if the stateful pieces are in their own module.

jeddai: Do you think that's worth me making that change in this PR?

components/nimbus/src/lib.rs
yashikakhurana: may be doing this` #[cfg_attr(not(feature = "nimbus"), allow(unused))]`
this allow(unused) attribute only when the nimbus feature is disabled. When the nimbus feature is enabled, the allow(unused) attribute will not be applied.

jeddai: AHH That's what I was trying to find, thank you! I was expecting there to simply be an option to join cfg attributes together, but `cfg_attr` seems to be specifically for this exact purpose. Great find!!

components/nimbus/src/enrollment.rs
jeddai: I decided to preemptively merge #5374 into this PR! I made some additional changes to support this, such as pulling the NimbusTargetingHelper out into its own module.

components/nimbus/src/enrollment.rs
```diff
@@ -17,7 +15,18 @@ use std::{
...
+        use crate::{
+            persistence::{Database, Readable, StoreId, Writer},
+            EnrolledExperiment,
+        };
+
+        const DB_KEY_GLOBAL_USER_PARTICIPATION: &str = "user-opt-in";
+    }
+}
+
+#[cfg_attr(not(feature = "stateful"), allow(unused))]
```
jeddai, line 29: Massive thank you @yashikakhurana for finding `cfg_attr`, this is a great adjustment 🎉

components/nimbus/src/tests/mod.rs
```diff
@@ -2,20 +2,24 @@
...
+    if #[cfg(feature = "stateful")] {
+        mod test_behavior;
+        mod test_nimbus;
+        mod test_persistence;
+        mod test_updating;
+        mod client {
+            mod test_http_client;
+            mod test_null_client;
+        }
+    }
```
jeddai, line 24: @jhugman here's what `cfg_if` allows us to do – I'm very pleased with it

CHANGES_UNRELEASED.md
```diff
@@ -28,7 +28,8 @@ Use the template below to make assigning a version number during the release cut
 ## Nimbus ⛅️🔬🔭
 
 ### 🦊 What's Changed 🦊
-
+- Added a `stateful` cargo feature and added appropriate feature flag attributes ([#5448](https://github.com/mozilla/application-services/pull/5448)).
+  - This does not functionally change build processes, as the `stateful` feature is now the default feature for the `nimbus-sdk` library.
 - Changed the ordering around for optional arguments for Python compatibility ([#5460](https://github.com/mozilla/application-services/pull/5460)).
   - This does not change Kotlin or Swift APIs, but affects code that uses the uniffi generated FFI for `record_event` and `record_past_event` directly.
 - Refactor the `EnrollmentEvolver` in preparation for a larger refactor to split out a `stateful` feature. ([#5374](https://github.com/mozilla/application-services/pull/5374)).
```
jhugman, line 35: The reference to #5374 can probably be folded in to your log entry.

At a bare minimum it should happen occur chronologically before (and next to) your log entry.

components/nimbus/tests/common/mod.rs
```diff
@@ -2,43 +2,99 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[cfg(feature = "rkv-safe-mode")]
+use rkv::StoreOptions;
+
 // utilities shared between tests
-use nimbus::{error::Result, AppContext, NimbusClient, RemoteSettingsConfig};
+cfg_if::cfg_if! {
+    if #[cfg(feature = "stateful")] {
```
jhugman, line 10: Where the whole file is behind this, would it make more sense to change the `use` site(s)— in the same way `nimbus_client.rs` doesn't have any cfg/feature attributes?

components/nimbus/src/lib.rs
```diff
@@ -2,945 +2,44 @@
...
-        log::info!("updating experiment list");
-        let db = self.db()?;
-        let mut writer = db.write()?;
+cfg_if::cfg_if! {
+    if #[cfg(feature = "stateful")] {
+        mod behavior;
+        mod client;
+        mod config;
+        mod dbcache;
+        mod updating;
```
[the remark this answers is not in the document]
jeddai: AFAIK it was there previously for `crate::` imports, but I do agree it's not strictly necessary anymore. 

I'm content with removing it if you'd prefer, but I'd prefer to handle it in a different ticket because it'll result in a lot of import changes.

components/nimbus/src/targeting.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    pub fn eval_jexl(&self, expr: String) -> Result<bool> {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "stateful")] {
+                jexl_eval(&expr, &self.context, self.event_store.clone())
+            } else {
+                jexl_eval(&expr, &self.context)
+            }
+        }
```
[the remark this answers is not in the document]
jeddai: Yep it makes it quite clean, and actually doesn't change tests like at all. The tests focus on the `targeting` method, and that uses a targeting helper (which acts as a proxy for the `jexl_eval` method)!

yashikakhurana:
@jeddai I have reviewed the code, overall looks good, just don't have enough experience with rust :D

PR #5449, opened 2023-03-27
Fixup days in a week test failure

jhugman:
https://github.com/mozilla/application-services/pull/5446#discussion_r1149856015

Without this fix, the tests will pass on mondays, but no other day of the week.

  - Note:

### DIFF
--- a/components/nimbus/src/tests/test_behavior.rs
+++ b/components/nimbus/src/tests/test_behavior.rs
@@ -1250,24 +1250,63 @@ mod event_store_tests {
             5.0,
             store.query(event_id, Interval::Days, 56, 0, EventQueryType::Sum)?
         );
+        if same_year(now, then) {
+            assert_eq!(
+                5.0,
+                store.query(
+                    event_id,
+                    Interval::Weeks,
+                    1,
+                    usize::from(!same_week(now, then)), // 0 if the same week, 1 if not
+                    EventQueryType::Sum
+                )?
+            );
+            assert_eq!(
+                5.0,
+                store.query(
+                    event_id,
+                    Interval::Months,
+                    1,
+                    usize::from(!same_month(now, then)), // 0 if the same month, 1 if not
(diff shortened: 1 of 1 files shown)


components/nimbus/src/tests/test_behavior.rs
```diff
@@ -1323,4 +1325,27 @@ mod event_store_tests {
 
         Ok(())
     }
+
+    fn same_week(now: DateTime<Utc>, then: DateTime<Utc>) -> bool {
+        now.weekday().num_days_from_sunday() >= then.weekday().num_days_from_sunday()
```
jeddai, line 1330: I don't think `from_sunday` would work either because the default event store boots up with the beginnings of everything set to the first day of the year. I think you'd need to do something like `(<now|then>.day_of_year() / 7).floor()` to get the week w/i the context of the event store

jhugman: I think you're right. I've added `same_month`, `same_week` based on day_of_year / `ordinal0` and `same_year` tests so as not to break in January.

components/nimbus/src/tests/test_behavior.rs
```diff
@@ -1250,18 +1250,32 @@ mod event_store_tests {
...
-        );
+        if same_year(now, then) {
+            assert_eq!(
+                5.0,
+                store.query(
+                    event_id,
+                    Interval::Weeks,
+                    1,
+                    usize::from(!same_week(now, then)), // 0 if the same week, 1 if not
+                    EventQueryType::Sum
```
jeddai, line 1261: Just a nit but since these are SUM queries, we don't necessarily need to use the `same_<time>` methods and could just passively check two buckets instead of one. No problem leaving it like this though

jhugman: Added some comments and extra tests around using the appropriate granularity of interval.

PR #5451, opened 2023-03-28
Update chrono to avoid RUSTSEC-2020-0159. Fixes #4590

### LINKED ISSUE #4590: Remove cargo audit ignores once chrono advisory is resolved
With https://github.com/mozilla/application-services/pull/4589, we added `cargo audit` ignores for two security advisories. This ticket is to remove the ignores

This means one of two things:

* Either remove chrono all together and use `time` directly
* If `chrono` resolves the issue and releases a fixed version, upgrade to that

We should do this once enough time has passed and/or:

* https://github.com/chronotope/chrono/pull/578 is merged
* https://github.com/chronotope/chrono/issues/602 is fixed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2849)
┆Epic: Important backlog

mhammond:
Updated via `cargo update` - we can't specify a non-impacted version in toml as m-c is still on an impacted version.

### DIFF
--- a/.github/workflows/dependency-check.yaml
+++ b/.github/workflows/dependency-check.yaml
@@ -55,18 +55,13 @@ jobs:
         run: |
             cargo install cargo-audit
             # Explanation for ignored issues:
-            #  * RUSTSEC-2020-0159:  A possible Segfault in `chrono`'s `localtime_r' invocation, at the time of this
-            #                        patch, there is no fixed versions available, but an issue is filed on chrono: https://github.com/chronotope/chrono/issues/602
-            #  * RUSTSEC-2020-0071: Related to the one above, `chrono` pulls in a version of `time` that has the same problem, where invocations of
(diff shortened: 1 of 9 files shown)


mhammond:
RUSTSEC-2020-0071 is a larger yak to shave as there are still transitive deps to `time@0.1` via things other than chrono

mhammond:
damn - nimbus uses deprecated functions, so needs changes, and the audit is failing for reasons I don't understand.

mhammond:
Fixed the deprecated functions (which were primarily about pushing the policy of panicing on invalid values to the consumer) so if CI looks good I'll mark this as ready.

jeddai:
This actually works out incredibly well for me because I ran into issues earlier today with the same deprecated function uses! Thanks for fixing those, I can review in just a bit

jeddai (approved):
This looks good to me! I'm trusting the change to functionality with the `naive_date` use works the same because that section of code is well tested.

Thanks for getting this in!

PR #5452, opened 2023-03-28
Update openssl to avoid a couple of advisories

mhammond:
Because `cargo audit` is only run in CI when certain files are changed, we miss new vulnerabilities reported since that last happened - which is what's happening here - openssl has a couple of new vulns.

Updated via `cargo update`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2284,9 +2284,9 @@ checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
 [[package]]
 name = "openssl"
-version = "0.10.40"
+version = "0.10.48"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


mhammond:
```
Crate:         openssl
Version:       0.10.40
Title:         `openssl` `SubjectAlternativeName` and `ExtendedKeyUsage::other` allow arbitrary file read
Date:          2023-03-24
ID:            RUSTSEC-2023-0023
URL:           https://rustsec.org/advisories/RUSTSEC-2023-0023
Solution:      Upgrade to >=0.10.48

Crate:         openssl
Version:       0.10.40
Title:         `openssl` `X509Extension::new` and `X509Extension::new_nid` null pointer dereference
Date:          2023-03-24
ID:            RUSTSEC-2023-0024
URL:           https://rustsec.org/advisories/RUSTSEC-2023-0024
Solution:      Upgrade to >=0.10.48
```

skhamis (approved):
LGTM! Thanks for keeping up with this. The fact that we don't catch it without certain files occasionally getting hit is a bit concerning but we can solve that in a separate issue.

PR #5466, opened 2023-04-04
gh-5465 -- Add extra checks for the places 15-16 migration

bendk:
The rollout of this migration was unfortunate, since the new code was before the DB schema version was bumped.  This means that:

- Some users installed a nightly which had the v16 schema, but had the schema version set to 15.
- dce0f34 fixed the code so the version was correctly listed as 16
- This caused those users to run the 15 -> 16 migration, which fails since the `unknownFields` column is already present.

Updated the 15 -> 16 migration to handle this case.  Refactored the places schema code a bit so follows the new pattern we're using for schema upgrades with `open_database()`.  The new pattern gives us more flexibility which simplifies the fix.

  - Note:


linabutler (approved):
Oops, good catch!

PR #5468, opened 2023-04-04
places: Added test for running all DB migrations

bendk:
The issues with the latest fenix nightlies makes me think a test like this is important to have.

  - Note:

### DIFF
--- a/components/places/src/db/schema.rs
+++ b/components/places/src/db/schema.rs
@@ -299,6 +299,7 @@ mod tests {
     use crate::db::{db::PlacesInitializer, PlacesDb};
     use crate::error::Result;
     use sql_support::open_database::test_utils::MigratedDatabaseFile;
+    use std::collections::BTreeSet;
     use sync_guid::Guid as SyncGuid;
     use url::Url;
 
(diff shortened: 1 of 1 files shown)


components/places/src/db/schema.rs
```diff
@@ -775,4 +775,75 @@ mod tests {
         drop(db);
         db_file.upgrade_to(16);
     }
+
+    #[test]
+    fn test_db_migration_sanity() {
```
linabutler, line 780: suggestion: Let's try to find a different term for this if possible, please—maybe `test_all_upgrades`, to match `autofill` and `webext-storage`?

I know it's a common phrase (and I've definitely used it before), but it can have hurtful connotations, and I wonder if there's a more precise term we can use to describe this test—and improve (rip)grep-ability while we're at it!

linabutler (commented):
This is a lovely test! (Also, I think your patches are the first time I've seen a use of a `pragma_` as a table-valued function in the wild—so cool! 🎊)

components/places/src/db/schema.rs
bendk: Thanks for raising this!  I hadn't even considered the issue.  I just pushed a new commit that fixes the naming.

linabutler (approved):
You rock, thanks @bendk!

PR #5470, opened 2023-04-06
History schema evolution

bendk:

  - Note:

### DIFF
--- a/components/places/src/history_sync/payload_evolution_tests.rs
+++ b/components/places/src/history_sync/payload_evolution_tests.rs
@@ -1,0 +1,444 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Tests for sync payload evolution.  These are in a separate file because they test the
+//! functionality from several modules.
+
+use crate::{
+    api::places_api::{test::new_mem_api, PlacesApi},
+    history_sync::{record::HistoryRecord, HistorySyncEngine},
+    observation::VisitObservation,
+    storage::history::apply_observation,
+    types::{UnknownFields, VisitTransition},
+};
+use serde_json::{json, Value};
+use std::collections::{HashMap, HashSet};
+use sync15::{
+    bso::{IncomingBso, IncomingKind, OutgoingBso},
+    engine::{IncomingChangeset, SyncEngine},
+    telemetry, ServerTimestamp,
+};
+
+// Macro to quickly make `UnknownFields` values.  It's like `json!`, but avoids the need to append
+// `as_object().unwrap()` to each call.
+macro_rules! unknown_fields(
+   {
+       {
+           $($key:literal : $value: expr),* $(,)?
+       }
+   } => {
+       serde_json::Map::from_iter([
+           $(
+               (String::from($key), serde_json::json!($value)),
+           )*
+       ])
+   }
+);
+
+#[test]
+fn test_history_records_roundtrip_fields() {
+    // Test that we roundtrip unknown fields from remote history records after making changes to them
+    RoundtripTest {
+        initial_remote_records: vec![
+            json!({
+                "id": "a___________",
+                "title": "dogs",
+                "histUri": "https://dogs.com/",
+                "visits": [],
+                // Simulate 2 new features that nobody is asking for:
+                //   - themes for specific URLs
+                //   - prank the user when they try to open a URL
+                "theme": "solarized",
+                "prank": "rick-roll",
+                "visits": [
+                    {
+                        "date": timestamp(1),
+                        "type": 1,
+                    },
+                ]
+            }),
+            json!({
+                "id": "b___________",
+                "title": "cats",
+                "histUri": "https://cats.com/",
+                "visits": [
+                    {
+                        "date": timestamp(2),
+                        "type": 1,
+                    },
+                    {
+                        "date": timestamp(3),
+                        "type": 1,
+                    },
+                ],
+            }),
+        ],
+        local_visits: vec![
+            visit("https://dogs.com/", timestamp(10)),
+            visit("https://cats.com/", timestamp(11)),
+        ],
+        incoming_remote_records: vec![],
+        outgoing_unknown_fields: vec![
+            OutgoingRecordUnknownFields {
+                guid: "a___________".into(),
+                fields: unknown_fields!({
+                    "theme": "solarized",
+                    "prank": "rick-roll",
+                }),
+                visits_fields: vec![
+                    (timestamp(1), unknown_fields!({})),
+                    (timestamp(10), unknown_fields!({})),
+                ],
+            },
+            OutgoingRecordUnknownFields {
+                guid: "b___________".into(),
+                fields: unknown_fields!({}),
+                visits_fields: vec![
+                    (timestamp(2), unknown_fields!({})),
+                    (timestamp(3), unknown_fields!({})),
+                    (timestamp(11), unknown_fields!({})),
+                ],
+            },
+        ],
+    }
+    .run()
+}
+
+#[test]
+fn test_history_records_new_unknown_field_values() {
+    // If there are incoming history records with new unknown fields, we should use those rather than our
+    // stored unknown fields
+    RoundtripTest {
+        initial_remote_records: vec![json!({
+            "id": "a___________",
+            "title": "dogs",
+            "histUri": "https://dogs.com/",
+            "visits": [],
+            "theme": "solarized",
+            "prank": "rick-roll",
+            "visits": [
+                {
+                    "date": timestamp(1),
+                    "type": 1,
+                },
+            ]
+        })],
+        local_visits: vec![visit("https://dogs.com/", timestamp(10))],
+        incoming_remote_records: vec![json!({
+            "id": "a___________",
+            "title": "dogs",
+            "histUri": "https://dogs.com/",
+            "visits": [],
+            "theme": "nord",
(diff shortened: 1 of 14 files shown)


components/places/src/history_sync/payload_evolution_tests.rs
```diff
@@ -0,0 +1,336 @@
...
+        }],
+    }
+    .run()
+}
+
+// The last case to consider is incoming visits with differerent unknown fields.  Maybe we should
+// handle that, but it would mean changing a lot of legacy code that assumes that visits are
+// immutable and can be compared based on the timestamp + transition fields (or the other code that
+// compares only the timestamps).  If we want to address this case, let's do it as part of a
+// general rework of the places sync merge code.
```
bendk, line 215: What do others think about this one?  The current code treats visits are currently considered immutable, which means if we add unknown fields, they can't change either.

Should we try to rework the code so it doesn't make that assumption?  I kind of want to rework the merging code anyways and this could be a good excuse.

Another question is if we want to allow new fields to be part of the "key".  Visits don't have a GUID, we [treat two visits the same if they have the same timestamp + transition type](https://github.com/mozilla/application-services/blob/main/components/places/src/history_sync/plan.rs#L106-L122) (except [sometimes it's just timestamp](https://github.com/mozilla/application-services/blob/main/components/places/src/storage/history.rs#L921-L928) -- one more reason that we may want to rework this code).  Would we ever want to allow new fields to be added to this key?  Right now I don't think it's worth the complexity, but I wonder what others think.

skhamis: One thing I'm trying to keep in mind of roundtripping this data, is really to prevent data loss until the clients that can "catch up" and actually understand the data in those fields. 

Which means -- `unknown_fields` _should_ eventually disappear as a client becomes "recent", having unknown_fields stick around even if every client can understand the schema -- seems bad (IMO). I do like the idea of reworking so that we avoid this pitfall but depends on our timelines.

To your second question, I agree given the added complexity and we don't have a spec we'd be trying to design the would-be new "keys" around -- might be better to de-prioritize that.

components/places/src/storage/maintenance.rs
```diff
@@ -0,0 +1,225 @@
...
+/// The `run_maintenance_*()` functions are intended to be run during idle time and will take steps
+/// to clean up / shrink the database.  They're split up so that we can time each one in the
+/// Kotlin wrapper code (This is needed because we only have access to the Glean API in Kotlin and
+/// it supports a stop-watch style API, not recording specific values).
+pub fn run_maintenance_checkpoint(conn: &PlacesDb) -> Result<()> {
+    conn.execute_one("PRAGMA wal_checkpoint(PASSIVE)")?;
+    Ok(())
+}
+
+fn prune_unknown_fields(conn: &PlacesDb, now: Timestamp) -> Result<()> {
```
skhamis, line 86: This is a nice solution! Would be nice to have some comments of why this exists and why we need to prune (I'm assuming it's related to the immutable visits discussion). 

That being said, I'm not entirely sure how I feel of separating the maintenance of pruning unknown_fields -- wouldn't they get removed with the visits their associated with (in `prune_older_visits`) or is this used for something else? Comments here might also lessen the confusion of this.

skhamis (commented):
Seems like a great start! It seems close, but the pruning of unknown_fields separate from visits gives me a little pause. Feels we might need to understand the implications of that a bit more

components/places/src/history_sync/payload_evolution_tests.rs
bendk: > Which means -- unknown_fields should eventually disappear as a client becomes "recent", having unknown_fields stick around even if every client can understand the schema -- seems bad (IMO). I do like the idea of reworking so that we avoid this pitfall but depends on our timelines.

I'm not understanding this in relation to this question.  I think we will eventually null out `unknown_fields` once we upgrade no matter what.  My question is if we should support changes to `unknown_fields` before that happens.

However, it's a really good point in general and it's making me rethink the pruning idea.

components/places/src/storage/maintenance.rs
bendk: Sorry, I meant to add a comment here and totally forgot about it.  But now that I'm writing it down, I'm having second thoughts.  Maybe we shouldn't special case the `unknown_fields` and just continue to use the existing pruning code.

The reasoning behind the `prune_unknown_fields` is that we only sync the most recent 20 visits for each URL, so storing `unknown_fields` for visits older than that seems like a waste of space.  However, that's not really true:
  - As you point out, the client will upgrade and at that point it will want to migrate the `unknown_fields` data and start using it.
  - If you delete recent visits, then that will push older URLs into the most recent 20.  (This is related to https://github.com/mozilla/application-services/issues/5472).

Part of the motivation here is there's a lot of visits in the DB and `unknown_fields` is space inefficient.  You want to add an enum column that could be stored in as an integer, you end up storing an entire JSON string.  Maybe we can address this another way though:
  - Lets say that we are on nightly version v120 and want to add a new feature that requires a new field
  - Before starting work on that feature, we add a migration to store the new field in a column and round-trip the data.
  - The migration should be pretty trivial, so we uplift it into v119 and an ESR patch ahead of the v120 release.
  - That way users who are doing a reasonable job keeping up-to-date don't have to deal with the space penalty.

skhamis (approved):
Thanks for making those changes. I think it looks good! I think we'll be able to better prune the unknown_fields when the time comes to "consume" said fields.

👍 from me! But I'd wait for Linas review before merging!

components/places/src/history_sync/payload_evolution_tests.rs
mhammond: I think a visit is considered immutable because it's recording a historical event - so changing it in the future really doesn't make sense. I can't see how a different device would somehow say "this old visit I previously recorded now should have different attributes".

Similarly, I think the key question is roughly the same - it's a record of something happening at an instant - I don't see how other fields being part of the key makes sense - you'd never want to record 2 different visits of the same type at the same timestamp with different attributes - there can only really have been one visit at that instant.

If that all makes sense, maybe just update the comment to reflect that?

mhammond (approved):
This looks correct to me!

components/places/src/history_sync/payload_evolution_tests.rs
```diff
@@ -0,0 +1,336 @@
...
+                        "emotion": "anger",
+                    }),
+                ),
+                (timestamp(10), unknown_fields!({})),
+            ],
+        }],
+    }
+    .run()
+}
+
```
linabutler, line 210: Could you please add a test with deduping involved? Maybe one where we have a local record and 2 remote records, all with the same URL but different GUIDs?

AFAICR, history deduping works a little oddly in this case. On the server, only the GUID can be unique, but locally, there's an additional requirement that URLs are unique. But because of how we rotate history, and sync only a subset of your history on the first sync, it's possible for the server to end up with multiple records—each with different GUIDs—for the same URL.

(This is another reason #5472 is tricky: the tombstone only carries the GUID, and if your device deduped the page to a different GUID than the one that uploaded the tombstone, your device won't know to delete that page).

I think our deduping logic just smushes visits for all those together, and picks the first (or the last?) page GUID that it saw as the winner. I'm not sure the right semantics for unknown fields in that case—keep the fields from the record with the winning GUID, and discard the others?—but it would be good to capture them in a test, if it's not too cumbersome to add.

linabutler (approved):
This looks great to me, too; thanks @bendk, and apologies for the delay!

My only question is how deduping will interact with this, but more about capturing the current semantics in a test, and not trying to define new semantics for it.

I also think your approach of adding unknown fields to just the page record, instead of individual visits, is great. If we do decide to look closer at visits, I think it's worth having that discussion in the context of other changes, like supporting deletions for them, instead of trying to solve all that here.

Thank you again! 🎊

components/places/src/history_sync/payload_evolution_tests.rs
```diff
@@ -0,0 +1,336 @@
...
+};
+use serde_json::{json, Value};
+use std::collections::{HashMap, HashSet};
+use sync15::{
+    bso::{IncomingBso, IncomingKind, OutgoingBso},
+    engine::{IncomingChangeset, SyncEngine},
+    telemetry, ServerTimestamp,
+};
+
+macro_rules! unknown_fields(
```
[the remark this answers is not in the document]
bendk: Yeah, it's basically `json!()`, but without having to type `as_object().unwrap()`.

components/places/src/history_sync/payload_evolution_tests.rs
bendk: > I think our deduping logic just smushes visits for all those together, and picks the first (or the last?) page GUID that it saw as the winner. I'm not sure the right semantics for unknown fields in that case—keep the fields from the record with the winning GUID, and discard the others?—but it would be good to capture them in a test, if it's not too cumbersome to add.

AFAICT, the current behavior is very weird:
 - We pick the GUID from the first record
 - We pick the title from the last record
 - We hope the records are in the same order next sync (which I don't think is guaranteed).

I added a test to verify the title comes from the last record.  Based on that, I made the `unknown_fields` also come from the last record.  I'm not really sure if this is right or not, what do you think?

linabutler: > I added[ a test to verify the title comes from the last record](https://github.com/mozilla/application-services/blob/6879a0b13b62ca23f223130871b0973d51bf8f95/components/places/src/history_sync/plan.rs#L480-L538). Based on that, and the fact that it was simpler to code, I made the `unknown_fields` also come from the last record. I'm not really sure if this is right or not, what do you think?

That sounds great to me, thank you!

PR #5471, opened 2023-04-06
Add cirrus_client to Nimbus

jeddai:

  - Note:


components/nimbus/src/enrollment.rs
```diff
@@ -1132,9 +1163,12 @@ impl Defaults for EnrolledFeatureConfig {
...
-#[cfg(test)]
-impl EnrolledFeatureConfig {
-    pub fn is_rollout(&self) -> bool {
+impl SharedExperimentMetadata for EnrolledFeatureConfig {
+    fn get_slug(&self) -> String {
+        self.slug.clone()
+    }
+
+    fn is_rollout(&self) -> bool {
         self.branch.is_none()
```
jeddai, line 1172: So this was previously used to determine if something was a rollout in tests, but I'm iffy on doing the same here.

Would it make more sense to put a field for `is_rollout` on the `EnrolledFeatureConfig` struct and make sure it's set properly? Not sure how big of a lift that would be or if that would result in a little bit of a nasty refactor with Swift/Kotlin.

FWIW making that adjustment can come later — the first Nimbus web adopters don't plan to hold onto the enrolled data in the browser or anywhere else, meaning `prev_experiments` will always be empty coming into the function for Nimbus web.

yashikakhurana (commented):
overall looks good to me, I will let James do a final pass, I just have couple of question to understand it better

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,117 @@
...
+        context: TargetingAttributes,
+        is_user_participating: bool,
+        prev_experiments: &[EnrolledFeatureConfig],
+        next_experiments: &[Experiment],
+        prev_enrollments: &[ExperimentEnrollment],
+    ) -> Result<EnrollmentResponse> {
+        let nimbus_id = Uuid::new_v4();
+        let available_randomization_units =
+            AvailableRandomizationUnits::with_client_id(client_id.as_str());
+        let th = context.into();
```
jhugman, line 97: I'd just use `TargetingHelper::new(context)` here.

components/nimbus/src/targeting.rs
```diff
@@ -61,6 +61,26 @@ impl NimbusTargetingHelper {
...
+    }
+}
+
+impl From<AppContext> for NimbusTargetingHelper {
+    fn from(context: AppContext) -> Self {
+        let ta: TargetingAttributes = context.into();
+        ta.into()
+    }
+}
+
```
jhugman, line 83: If you're just using `TargetingHelper::new()` in [`cirrus_client.rs`](https://github.com/mozilla/application-services/pull/5471/files#diff-f13145056c9a37d6aec2871fb99b16ebd1340212a3354e1dfee5a4bdb4219332R97) then this can stay where it is.

components/nimbus/src/lib.rs
```diff
@@ -43,3 +49,34 @@ pub use evaluator::TargetingAttributes;
...
+        impl UniffiCustomTypeConverter for JsonObject {
+            type Builtin = String;
+
+            fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+                let json: Value = serde_json::from_str(&val)?;
+
+                match json.as_object() {
+                    Some(obj) => Ok(obj.clone()),
+                    _ => Err(uniffi::deps::anyhow::anyhow!(
+                        "Unexpected JSON-non-object in the bagging area"
```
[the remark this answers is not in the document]
jhugman: and now it's Charlie's `git blame`. :)

components/nimbus/src/enrollment.rs
jhugman: `EnrolledFeatureConfig` doesn't currently cross the FFI. As alluded to in https://github.com/mozilla/application-services/pull/5475, I'm not sure it should now.

I wonder if we should have a slimmed down `prev_enrollments` with `next_experiments` which we can derive `prev_experiments`? But as you said, the client isn't going to be storing any enrolled data in the browser.

jhugman (commented):
I've held off a full review of this, because I've requested changes for https://github.com/mozilla/application-services/pull/5475.

jeddai:
@jhugman oh this PR is actually the first of the three that will need to be merged. I'll take relevant feedback from the other PR and apply it here though!

components/nimbus/src/enrollment.rs
```diff
@@ -527,6 +527,16 @@ impl EnrollmentStatus {
...
 
+pub(crate) trait SharedExperimentMetadata {
+    fn get_slug(&self) -> String;
+
+    fn is_rollout(&self) -> bool;
+
+    fn is_not_rollout(&self) -> bool {
+        !self.is_rollout()
+    }
+}
```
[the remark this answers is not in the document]
jeddai: The `not` method is there because of how it is being used — @jhugman commented on #5475 that I can instead make a temporary variable that's a lambda for the same functionality so that's my plan

components/nimbus/src/enrollment.rs
```diff
@@ -972,14 +994,17 @@ fn map_enrollments(enrollments: &[ExperimentEnrollment]) -> HashMap<String, &Exp
...
-    let experiments: Vec<Experiment> = filter_experiments(experiments, filter_fn);
+    filter_fn: fn(&E) -> bool,
+) -> (Vec<E>, Vec<ExperimentEnrollment>)
+where
+    E: SharedExperimentMetadata + Clone,
+{
+    let experiments: Vec<E> = filter_experiments(experiments, filter_fn);
 
-    let slugs: HashSet<String> = experiments.iter().map(|e| e.slug.clone()).collect();
+    let slugs: HashSet<String> = experiments.iter().map(|e| e.get_slug()).collect();
```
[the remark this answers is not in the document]
jeddai: OH I also forgot to mention the type change. Instead of the type being `&[Experiment]`, it is now `&[E]` — `where E: SharedExperimentMetadata + Clone`. This is akin to having a generic where the generic implements an interface.

Experiment and EnrolledFeatureConfig both have implementations for the `SharedExperimentMetadata` trait, which allows us to use the same method — `get_slug()` — across each struct that implements that trait.

`Clone` is added there as well because we need to be able to clone/move the object and its members.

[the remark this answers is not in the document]
jeddai: Happy to chat about this one on a call if you'd like

components/nimbus/src/enrollment.rs
jeddai: Hm we could probably do something similar to what I did with the `ExperimentMetadata` trait? I'd have to dig through the code to double-check everything `prev_enrollments` is used for, but I do wholly agree. 

I think if we want to make that adjustment though I can make it in the UDL PR or in a later refactor (I'm leaning towards the latter)

components/nimbus/src/enrollment.rs
```diff
@@ -634,13 +638,16 @@ impl<'a> EnrollmentsEvolver<'a> {
         Ok(enrollments_change_events)
     }
 
-    pub(crate) fn evolve_enrollments(
+    pub(crate) fn evolve_enrollments<E>(
         &self,
         is_user_participating: bool,
-        prev_experiments: &[Experiment],
+        prev_experiments: &[E],
         next_experiments: &[Experiment],
```
yashikakhurana, line 645: do you think this type should also be changed to `E`

components/nimbus/src/enrollment.rs
```diff
@@ -693,13 +699,16 @@ impl<'a> EnrollmentsEvolver<'a> {
 
     /// Evolve and calculate the new set of enrollments, using the
     /// previous and current state of experiments and current enrollments.
-    pub(crate) fn evolve_enrollment_recipes(
+    pub(crate) fn evolve_enrollment_recipes<E>(
         &self,
         is_user_participating: bool,
-        prev_experiments: &[Experiment],
+        prev_experiments: &[E],
         next_experiments: &[Experiment],
```
yashikakhurana, line 706: same in here

yashikakhurana (commented):
thank you @jeddai for explaining the previous doubts that I had

components/nimbus/src/enrollment.rs
jeddai: Not _yet_, but it's something we could do in the future. At the moment, the intent was only to add support for the `EnrolledFeatureConfig` type on `prev_experiments`. This plan is actually flawed as pointed out by @jhugman, but I'm going to address that in a [future ticket](https://mozilla-hub.atlassian.net/browse/EXP-3430).

components/nimbus/src/matcher.rs
```diff
@@ -54,3 +56,35 @@ pub struct AppContext {
...
+    pub app_id: String,
+    pub channel: String,
+    pub app_version: Option<String>,
+    pub app_build: Option<String>,
+    pub locale: Option<String>,
+    pub os: Option<String>,
+    pub os_version: Option<String>,
+    #[serde(flatten)]
+    pub custom_targeting_attributes: Option<Map<String, Value>>,
+}
```
jhugman, line 90: At some point, I prefer this re-declaration approach over the preffing on an off of properties.

I think you have it right here.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,118 @@
...
+            is_user_participating.unwrap_or(true),
+            &match prev_experiments {
+                Some(exps) => exps,
+                None => Vec::new(),
+            },
+            &next_experiments,
+            &match prev_enrollments {
+                Some(enrollments) => enrollments,
+                None => Vec::new(),
+            },
```
jhugman, line 82: Nit: this is… a little ugly. Can we use default values here, instead of an option?

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub(crate) fn enroll(
+        &self,
+        client_id: String,
+        context: TargetingAttributes,
+        is_user_participating: bool,
+        prev_experiments: &[EnrolledFeatureConfig],
+        next_experiments: &[Experiment],
+        prev_enrollments: &[ExperimentEnrollment],
+    ) -> Result<EnrollmentResponse> {
+        let nimbus_id = Uuid::new_v4();
```
jhugman, line 95: This needs a comment here, I think.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,118 @@
...
+        let available_randomization_units =
+            AvailableRandomizationUnits::with_client_id(client_id.as_str());
+        let th = NimbusTargetingHelper::new(context);
```
jhugman, line 98: 😻

jhugman (approved):
This is looking good. 

I've left some comments, which should be addressed, and also a ✅ so as not to hold you up.

Well done!

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct EnrollmentRequest {
+    pub client_id: Option<String>,
+    pub context: TargetingAttributes,
+    pub is_user_participating: Option<bool>,
+    pub prev_experiments: Option<Vec<EnrolledFeatureConfig>>,
```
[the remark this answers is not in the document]
jeddai: So I want to talk about this more to make sure I'm not missing anything important, but it does seem like that's absolutely something we can do. So yeah, I'll pull out previous experiments here. I opened a ticket to do that refactor [here](https://mozilla-hub.atlassian.net/browse/EXP-3430).

components/nimbus/src/matcher.rs
jeddai: Same, it was so many `#[cfg(feature = "")]`'s one after the other and I couldn't wrap it in a multiline because `cfg_if` doesn't seem to support wrapping around struct members.

components/nimbus/src/stateless/cirrus_client.rs
jeddai:
```rust
&match prev_enrollments {
    Some(enrollments) => enrollments,
    _ => Default::default(),
},
```

Like this?

components/nimbus/src/enrollment.rs
```diff
@@ -1183,7 +1207,7 @@ impl EnrollmentChangeEvent {
     }
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
jeddai: Yep it's required for the `EnrollmentChangeEvent`s to be sent as part of the response.

PR #5475, opened 2023-04-10
Set up Cirrus UDL

jeddai:
**Wait for #5471**

Testing:
1. Build the UniFFI bindings and cdylib
```bash
cd components/nimbus
cargo build --no-default-features
cargo uniffi-bindgen generate src/cirrus.udl --language python -o .
cargo build --manifest-path ../../megazords/cirrus/Cargo.toml --release
mv ../../target/release/libcirrus.dylib ./libuniffi_cirrus.dylib
```

2. This python script tests a number of things; importantly it tests the custom type bindings in the `uniffi.toml`
```python
from components.nimbus.cirrus import CirrusClient

# All experiments using Cirrus will need to use the client_id for bucket config
bucket_config = dict({
    "randomizationUnit": "client_id",
    "count": 100,
    "namespace": '',
    "start": 1,
    "total": 100
})

branches = list([
    dict({
        "slug": 'control',
        "ratio": 1,
        "feature": dict({
            "featureId": 'feature1',
            "value": {
                "value key": "value"
            }
        })
    }),
    dict({
        "slug": 'treatment',
        "ratio": 1,
        "feature": dict({
            "featureId": 'feature1',
            "value": {
                "value key": "treatment value"
            }
        })
    })
])

exp = dict({
    "schemaVersion": '1.0.0',
    "slug": 'experiment-slug',
    "userFacingName": '',
    "userFacingDescription": '',
    "appId": 'test app id',
    "appName": 'test app name',
    "channel": 'dev',
    # Use is_already_enrolled for sticky targeting, otherwise we check lang, region, and a custom attribute
    "targeting": '(is_already_enrolled) || (username in ["test", "jeddai"])',
    "bucketConfig": bucket_config,
    "isRollout": False,
    "isEnrollmentPaused": False,
    "proposedEnrollment": 10,
    "branches": branches,
    "featureIds": ['feature1']
})

# Set up the context to match and add the custom targeting attributes
# Consider changing app_name, channel, and other values to see it fail to enroll
ta = {
    "app_id": 'test app id',
    "app_name": 'test app name',
    "channel": 'dev',
    "username": "jeddai",
    "active_experiments": [],
    "is_already_enrolled": False,
}

# Create the request; no need to pass in prev experiments/enrollments
request = {
    "clientId": 'jeddai',
    "appContext": ta,
    "requestContext": dict(),
    "nextExperiments": [exp]
}

# Create the client object
client = CirrusClient()
# Request handling of an enrollment
response = client.handle_enrollment(request)

print(response, end='\n\n')
print(response["enrollments"][0], end='\n\n')
print(response["enrolledFeatureConfigMap"]['feature1'], end='\n\n')
print(response["events"][0], end='\n\n\n')
```

Separate ticket for Python tests: https://mozilla-hub.atlassian.net/browse/EXP-3378

  - Note:


megazords/cirrus/Cargo.toml
jeddai, line 1: So I'm not sure if there's a better way to handle this. 

I was hoping I could find a way to add `cfg` attributes like features to add the `extern "C"` stuff when the crate-type is cdylib, but I couldn't find that. I was also hoping to find a way to configure `cargo build` to build the crate as a cdylib but I also couldn't find that. 

So that's how I landed on a separate megazord crate — if anyone knows of ways to do the above please share!

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+typedef string JsonObject;
+
+[Custom]
+typedef string Value;
+
+[Custom]
+typedef string Uuid;
+
+[Custom]
+typedef string StringHashSet;
```
jeddai, line 175: Took me a while to realize the reason `JsonObject` is a custom type is because UniFFI doesn't handle types with generics very well (outside of its built-in types).

components/nimbus/uniffi.toml
```diff
@@ -11,7 +11,34 @@ imports = [ "org.json.JSONObject" ]
...
+[bindings.python.custom_types.Uuid]
+type_name = "str"
+into_custom = "{}"
+from_custom = "{}"
+
+[bindings.python.custom_types.StringHashSet]
+type_name = "set"
+imports = ["json"]
+into_custom = "set(json.loads({}))"
+from_custom = "json.dumps(list({}))"
```
jeddai, line 44: The python script in the description should be testing all of these afaict. There's a `Uuid` that comes back on enrollment (`enrollment_id`), the `StringHashSet` is on the `TargetingAttributes`, `JsonObject` is on `AppContext`, and `Value` is on a few things. 

On `Uuid` — do I actually need to include it here? Not sure if UniFFI handles it by treating it as a string in both directions by default.

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+};
+
+[Enum]
+interface EnrollmentStatus {
+    Enrolled(Uuid enrollment_id, EnrolledReason reason, string branch);
+    NotEnrolled(NotEnrolledReason reason);
+    Disqualified(Uuid enrollment_id, DisqualifiedReason reason, string branch);
+    WasEnrolled(Uuid enrollment_id, string branch, u64 experiment_ended_at);
+    EnrollError(string reason);
+};
```
jeddai, line 77: This threw me off — I didn't realize that parameterized enums a) had a separate syntax, and b) required you to include all the parameters.

I like the syntax though, seems to handle it great and makes a lot of sense.

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+
+dictionary TargetingAttributes {
+    AppContext app_context;
+    boolean is_already_enrolled;
+    StringHashSet active_experiments;
+    string? language = null;
+    string? region = null;
+    i32? days_since_install = null;
+    i32? days_since_update = null;
+};
```
jeddai, line 30: Something I noticed with setting defaults for struct members — it doesn't seem to matter in which order you include the member variables. Regardless of the order in the UDL UniFFI seems to handle the translations appropriately, which is important since right now the UniFFI Python bindings need all members/params with defaults to be after all non-default ones (Python's positional vs named parameters ordering rules).

components/nimbus/build.rs
```diff
@@ -8,6 +8,9 @@ pub fn main() {
     #[cfg(feature = "stateful-uniffi-bindings")]
     uniffi::generate_scaffolding("./src/nimbus.udl").unwrap();
 
+    #[cfg(not(feature = "stateful-uniffi-bindings"))]
```
jeddai, line 11: This is... eh. I'm not super pleased with this but unless we want to pull Cirrus out into its own crate we can't really make this any better as far as I can tell.

megazords/cirrus/Cargo.toml
jeddai, line 1: So I'm not sure if there's a better way to handle this.

I was hoping I could find a way to add cfg attributes like features to add the extern "C" stuff when the crate-type is cdylib, but I couldn't find that. I was also hoping to find a way to configure cargo build to build the crate as a cdylib but I also couldn't find that.

So that's how I landed on a separate megazord crate — if anyone knows of ways to do the above please share!

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+    string? channel = null;
+    string? targeting = null;
+    string? start_date = null;
+    string? end_date = null;
+    u32? proposed_duration = null;
+    string? reference_branch = null;
+};
+
+dictionary EnrollmentResponse {
+    record<DOMString, EnrolledFeatureConfig> enrolled_feature_config_map;
```
jhugman, line 140: Oh, ha. This is what comes out of `map_features_by_feature_id`.

By the time it goes to the FML, it just needs to be

```
record<string, JsonObject>
```

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+    sequence<ExperimentEnrollment> enrollments;
+    sequence<EnrollmentChangeEvent> events;
+};
+
+dictionary EnrollmentRequest {
+    string? client_id;
+    TargetingAttributes context;
+    sequence<Experiment> next_experiments;
+    boolean? is_user_participating = null;
```
jhugman, line 149: Should this be default to `true` rather than `null`?

components/nimbus/src/cirrus.udl
jhugman, line 175: I think these all disappear.

components/nimbus/src/enrollment.rs
```diff
@@ -521,12 +523,22 @@ impl EnrollmentStatus {
...
 }
 
+pub(crate) trait SharedExperimentMetadata {
+    fn get_slug(&self) -> String;
+
+    fn is_rollout(&self) -> bool;
+
+    fn is_not_rollout(&self) -> bool {
+        !self.is_rollout()
+    }
```
jhugman, line 539: I'm not in love with this method.

components/nimbus/src/enrollment.rs
```diff
@@ -521,12 +523,22 @@ impl EnrollmentStatus {
...
                 ..
             } => *enrollment_id = Uuid::nil(),
-            EnrollmentStatus::NotEnrolled { .. } | EnrollmentStatus::Error { .. } => (),
+            EnrollmentStatus::NotEnrolled { .. } | EnrollmentStatus::EnrollError { .. } => (),
         };
         updated
     }
 }
 
+pub(crate) trait SharedExperimentMetadata {
```
jhugman, line 532: Nit: why is this called `Shared`?

components/nimbus/src/enrollment.rs
```diff
@@ -669,9 +683,10 @@ impl<'a> EnrollmentsEvolver<'a> {
         // We need to mop up all the enrollments that aren't rollouts (not just belonging to experiments that aren't rollouts)
         // because some of them don't belong to any experiments recipes, and evolve_enrollment_recipes will handle the error
         // states for us.
-        let experiments_only = |e: &Experiment| !e.is_rollout();
-        let prev_experiments = filter_experiments(prev_experiments, experiments_only);
-        let next_experiments = filter_experiments(next_experiments, experiments_only);
+        let prev_experiments =
+            filter_experiments(prev_experiments, SharedExperimentMetadata::is_not_rollout);
+        let next_experiments =
+            filter_experiments(next_experiments, SharedExperimentMetadata::is_not_rollout);
```
jhugman, line 689: Oh, ok. I think I'd prefer seeing a closure which we use twice declared in this method.

components/nimbus/src/lib.rs
```diff
@@ -43,3 +53,85 @@ pub use evaluator::TargetingAttributes;
...
+                    _ => Err(anyhow::anyhow!(
+                        "Unexpected JSON-non-array in the bagging area"
+                    )),
+                }
+            }
+
+            fn from_custom(val: Self) -> Self::Builtin {
+                Value::Array(Vec::from_iter(val).iter().map(|x| Value::String(x.to_owned())).collect()).to_string()
+            }
+        }
```
jhugman, line 133: I don't think these converters are necessary once you cut down on the API in the UDL.

components/nimbus/src/lib.rs
```diff
@@ -43,3 +53,85 @@ pub use evaluator::TargetingAttributes;
...
+    }
+
+    fn from_custom(obj: Self) -> Self::Builtin {
+        Value::Object(obj).to_string()
+    }
+}
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "stateful-uniffi-bindings")] {
+        include!(concat!(env!("OUT_DIR"), "/nimbus.uniffi.rs"));
```
jhugman, line 78: I think let's move these `include!` calls to their respective `nimbus_client.rs` and `cirrus_client.rs` files, and move the custom  converters to their own files.

components/nimbus/src/cirrus.udl
```diff
@@ -0,0 +1,183 @@
...
+    sequence<EnrolledFeatureConfig>? prev_experiments = null;
+    sequence<ExperimentEnrollment>? prev_enrollments = null;
```
jhugman, line 151: So is the intention here for the client to store this, but it's ok if not?

Would this be better to default to an empty list, rather than a `null`?

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub(crate) fn enroll(
+        &self,
+        client_id: String,
+        context: TargetingAttributes,
+        is_user_participating: bool,
+        prev_experiments: &[EnrolledFeatureConfig],
+        next_experiments: &[Experiment],
+        prev_enrollments: &[ExperimentEnrollment],
+    ) -> Result<EnrollmentResponse> {
+        let nimbus_id = Uuid::new_v4();
```
jhugman, line 94: This looks… strange. This would make `enroll` non-deterministic, I think? Is this transitory, or is this intentional?

jhugman (changes requested):
I think there are enough comments to be going on here, but I'm going to go have a look at the other PRs you have!

components/nimbus/src/enrollment.rs
jeddai: I called it `Shared` because the information the trait's methods returns is shared across a number of different types — happy to just call it `ExperimentMetadata` if that makes more sense. Or something else entirely

components/nimbus/src/enrollment.rs
jeddai: Using a single closure twice proved difficult because of the typing — it's not a fan of dynamic typing a lambda that supports both `&E` and `&Experiment`. Or at least I've not found a way to get it to support that.

I can totally replace it with a `|exp| !exp.is_rollout()` though

components/nimbus/src/cirrus.udl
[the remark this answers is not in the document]
jeddai: Active experiments comes from the database when the initialization finishes

components/nimbus/src/lib.rs
jeddai: It seems if I move these into their respective `client` files, then I have to write the definition for the `JsonObject` type converter in each of those files as well. The `UniffiCustomTypeConverter` is private so it's only available to the file the scaffolding rust is included in.

jeddai:
@jhugman I'm now generally torn on how to handle this. Part of me thinks we should just have a bare-minimum UDL like this:

```
namespace cirrus {};

[Error]
enum NimbusError {
    "InvalidPersistedData", "IOError",
    "JSONError", "EvaluationError", "InvalidExpression", "InvalidFraction",
    "TryFromSliceError", "EmptyRatiosError", "OutOfBoundsError","UrlParsingError",
    "UuidError", "InvalidExperimentFormat",
    "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch",
    "DatabaseNotReady", "VersionParsingError", "TryFromIntError",
    "ParseIntError", "TransformParameterError", "CirrusError"
};

[Custom]
typedef string JsonObject;

interface CirrusClient {
    constructor();

    // Handles an enrollment request and returns an enrollment response.
    [Throws=NimbusError]
    JsonObject handle_enrollment(JsonObject request);
};
```

Then either a) work with everything as `dict`s in python, or b) allow Python code to re-implement the dataclasses necessary for this since they don't work well with JSON and DEFINITELY won't work well for typed FastAPI definitions.

jhugman:
> @jhugman I'm now generally torn on how to handle this. Part of me thinks we should just have a bare-minimum UDL like this:
> 
> ```
> namespace cirrus {};
> 
> [Error]
> enum NimbusError {
>     "InvalidPersistedData", "IOError",
>     "JSONError", "EvaluationError", "InvalidExpression", "InvalidFraction",
>     "TryFromSliceError", "EmptyRatiosError", "OutOfBoundsError","UrlParsingError",
>     "UuidError", "InvalidExperimentFormat",
>     "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch",
>     "DatabaseNotReady", "VersionParsingError", "TryFromIntError",
>     "ParseIntError", "TransformParameterError", "CirrusError"
> };
> 
> [Custom]
> typedef string JsonObject;
> 
> interface CirrusClient {
>     constructor();
> 
>     // Handles an enrollment request and returns an enrollment response.
>     [Throws=NimbusError]
>     JsonObject handle_enrollment(JsonObject request);
> };
> ```
> 
> Then either a) work with everything as `dict`s in python, or b) allow Python code to re-implement the dataclasses necessary for this since they don't work well with JSON and DEFINITELY won't work well for typed FastAPI request/response definitions.

Just to close the circle: we talked in a 1-1 about having a _slightly_ more structured request/response, but I like this less structure in python, more passing through of things straight into the Cirrus client.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -22,10 +24,11 @@ use uuid::Uuid;
...
 /// - `enrollments`: This is the list of ExperimentEnrollments — it should be returned to the client.
 /// - `events`: This is the list of EnrollmentChangeEvents. These events should be recorded to Glean.
-#[derive(Debug, Default, Clone)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[serde(rename_all = "camelCase")]
 pub struct EnrollmentResponse {
-    pub enrolled_feature_config_map: JsonObject,
-    pub enrollments: Vec<ExperimentEnrollment>,
+    pub enrolled_feature_config_map: HashMap<String, JsonObject>,
+    pub enrollments: JsonArray,
```
jhugman, line 31: Note to self: this is because we're expecting the enrollments to be stored as an opaque blob, to be passed back in to the SDK, stored, or discarded. 

Perhaps we should make this more clear, by naming them `OpaqueEnrollments` or `OpaqueString`— and not even convert it from a string into a `dict` on the python side.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -86,35 +97,45 @@ impl CirrusClient {
...
         let client_id = if let Some(client_id) = client_id {
             client_id
         } else {
             return Err(NimbusError::CirrusError(
                 CirrusClientError::RequestMissingParameter("client_id".to_string()),
             ));
         };
+        let next_experiments: Vec<Experiment> = from_value(Value::Array(next_experiments)).unwrap();
+        let prev_enrollments: Vec<ExperimentEnrollment> =
+            from_value(Value::Array(prev_enrollments)).unwrap();
```
jhugman, line 120: Nit: a comment here re: why we're doing this in Rust, rather than in Uniffi-generated code.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -142,12 +162,60 @@ impl CirrusClient {
             map_features_by_feature_id(&enrollments, next_experiments);
 
         Ok(EnrollmentResponse {
-            enrolled_feature_config_map: enrolled_feature_config_map
+            enrolled_feature_config_map: HashMap::from_iter(
+                enrolled_feature_config_map
+                    .into_iter()
+                    .map(|(k, v)| (k, to_value(v).unwrap().as_object().unwrap().clone())),
```
jhugman, line 168: 👨‍🍳

components/nimbus/src/evaluator.rs
```diff
@@ -29,21 +32,52 @@ impl Bucket {
...
+#[cfg(not(feature = "stateful"))]
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct TargetingAttributes {
+    #[serde(flatten)]
+    pub app_context: AppContext,
+    #[serde(flatten)]
+    pub request_context: RequestContext,
+    pub language: Option<String>,
+    pub region: Option<String>,
+    pub is_already_enrolled: bool,
```
jhugman, line 57: Nit: I don't think makes sense in this context.

jhugman (approved):
I'm sorry, I left this on Thursday, and didn't come back to it until today. 

I've r+, but with comments. It's shaping up nicely!

components/nimbus/src/stateless/cirrus_client.rs
jeddai: As discussed today I'm going to switch the typings in the request/response types back to the original Rust types, and swap out `JsonObject` for `String` in all the Python-callable code.

components/nimbus/src/evaluator.rs
jeddai: No feature-gate required! It works without it fine

PR #5478, opened 2023-04-10
Cirrus python tests and Cirrus SDK changes

jeddai:
**Wait for #5475**

I don't feel like a changelog is necessary but I'm happy to add one if it's preferred.

This covers four small tickets:
* https://mozilla-hub.atlassian.net/browse/EXP-3378
* https://mozilla-hub.atlassian.net/browse/EXP-3391
* https://mozilla-hub.atlassian.net/browse/EXP-3435
* https://mozilla-hub.atlassian.net/browse/EXP-3446

  - Note:

### DIFF
--- a/automation/python-tests/conftest.py
+++ b/automation/python-tests/conftest.py
@@ -2,6 +2,7 @@ import pytest
 import json
 from components.nimbus.cirrus import CirrusClient
 
+
 @pytest.fixture
 def bucket_config():
     return {
@@ -41,7 +42,6 @@ def branches():
 
 @pytest.fixture
 def experiment(bucket_config, branches):
-
     return {
         "schemaVersion": '1.0.0',
         "slug": 'experiment-slug',
--- a/automation/python-tests/conftest.py
+++ b/automation/python-tests/conftest.py
@@ -1,0 +1,98 @@
+import pytest
+import json
+from components.nimbus.cirrus import CirrusClient
+
+@pytest.fixture
+def bucket_config():
+    return {
+        "randomizationUnit": "client_id",
+        "count": 100,
+        "namespace": '',
+        "start": 1,
+        "total": 100
+    }
+
+
+@pytest.fixture
+def branches():
+    return [
+        {
+            "slug": 'control',
+            "ratio": 1,
+            "feature": {
+                "featureId": 'feature1',
+                "value": {
+                    "value key": "value"
+                }
+            }
+        },
+        {
+            "slug": 'treatment',
+            "ratio": 1,
+            "feature": {
+                "featureId": 'feature1',
+                "value": {
+                    "value key": "treatment value"
+                }
+            }
+        }
+    ]
+
+
+@pytest.fixture
+def experiment(bucket_config, branches):
+
+    return {
+        "schemaVersion": '1.0.0',
+        "slug": 'experiment-slug',
+        "userFacingName": '',
+        "userFacingDescription": '',
+        "appId": 'test app id',
+        "appName": 'test app name',
+        "channel": 'dev',
+        # Use is_already_enrolled for sticky targeting, otherwise we check lang, region, and a custom attribute
+        "targeting": '(is_already_enrolled) || (username in ["test", "jeddai"])',
+        "bucketConfig": bucket_config,
+        "isRollout": False,
+        "isEnrollmentPaused": False,
+        "proposedEnrollment": 10,
+        "branches": branches,
+        "featureIds": ['feature1']
+    }
+
+
+@pytest.fixture
+def app_context():
+    return json.dumps({
+        "app_id": 'test app id',
+        "app_name": 'test app name',
+        "channel": 'dev',
+    })
+
+
+@pytest.fixture
+def request_context():
+    return {
+        "username": "jeddai",
+    }
+
+
+@pytest.fixture
+def req(request_context):
+    def _req(*, client_id='jeddai', request_context=request_context):
+        return json.dumps({
+            "clientId": client_id,
+            "requestContext": request_context,
+        })
+
+    return _req
+
+
+@pytest.fixture
+def client(app_context, experiment):
+    c = CirrusClient(app_context)
+    data = json.dumps({
+        "data": [experiment]
+    })
+    c.set_experiments(data)
+    return c
--- a/automation/python-tests/test_cirrus.py
+++ b/automation/python-tests/test_cirrus.py
@@ -1,10 +1,9 @@
 from components.nimbus.cirrus import NimbusError
 import json
 
+
 def test_one_experiment_should_enroll(client, req):
-    response = None
-    for i in range(0, 1000):
-        response = json.loads(client.handle_enrollment(req()))
+    response = json.loads(client.handle_enrollment(req()))
 
     assert len(response["enrollments"]) == 1
     assert "Enrolled" in response["enrollments"][0]["status"].keys()
@@ -27,15 +26,23 @@ def test_one_experiment_should_not_enroll(client, req, request_context):
     assert len(response["events"]) == 0
 
 
-def test_failure_case_no_client_id(client, req):
+def test_one_experiment_should_enroll_1000_times(client, req):
+    response = None
+    for i in range(0, 1000):
+        response = json.loads(client.handle_enrollment(req()))
 
+    assert len(response["enrollments"]) == 1
+    assert "Enrolled" in response["enrollments"][0]["status"].keys()
+    assert len(list(response["enrolledFeatureConfigMap"].values())) == 1
+    assert len(response["events"]) == 1
+
+
+def test_failure_case_no_client_id(client, req):
     request = req(client_id = None)
 
-    err = None
     try:
         client.handle_enrollment(request)
-    except NimbusError as e:
-        err = e
-
-    if err is None:
+    except NimbusError:
+        pass
+    else:
         assert False, 'client.handle_enrollment did not throw an error'
--- a/automation/python-tests/test_cirrus.py
+++ b/automation/python-tests/test_cirrus.py
@@ -1,0 +1,41 @@
+from components.nimbus.cirrus import NimbusError
+import json
+
+def test_one_experiment_should_enroll(client, req):
+    response = None
+    for i in range(0, 1000):
+        response = json.loads(client.handle_enrollment(req()))
+
+    assert len(response["enrollments"]) == 1
+    assert "Enrolled" in response["enrollments"][0]["status"].keys()
+    assert len(list(response["enrolledFeatureConfigMap"].values())) == 1
+    assert len(response["events"]) == 1
+
+
+def test_one_experiment_should_not_enroll(client, req, request_context):
+    request_context["username"] = "not test"
+
+    request = req(
+        request_context=request_context
+    )
+
+    response = json.loads(client.handle_enrollment(request))
+
+    assert len(response["enrollments"]) == 1
+    assert "NotEnrolled" in response["enrollments"][0]["status"].keys()
+    assert len(list(response["enrolledFeatureConfigMap"].values())) == 0
+    assert len(response["events"]) == 0
+
+
+def test_failure_case_no_client_id(client, req):
+
+    request = req(client_id = None)
+
+    err = None
+    try:
+        client.handle_enrollment(request)
+    except NimbusError as e:
+        err = e
+
+    if err is None:
+        assert False, 'client.handle_enrollment did not throw an error'
--- a/automation/run_python_tests.sh
+++ b/automation/run_python_tests.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+cd components/nimbus
+cargo build --no-default-features
+cargo uniffi-bindgen generate src/cirrus.udl --language python -o .
+cargo build --manifest-path ../../megazords/cirrus/Cargo.toml --release
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  mv ../../target/release/libcirrus.dylib ./libuniffi_cirrus.dylib
+else
+  mv ../../target/release/libcirrus.so ./libuniffi_cirrus.so
+fi
+
+cd ../..
+PYTHONPATH=$PYTHONPATH:$(pwd) pytest -s automation/python-tests
--- a/components/nimbus/src/cirrus.udl
+++ b/components/nimbus/src/cirrus.udl
@@ -12,12 +12,14 @@ enum NimbusError {
 };
 
 interface CirrusClient {
+    [Throws=NimbusError]
     constructor(string app_context);
 
     // Handles an enrollment request string and returns an enrollment response string.
     [Throws=NimbusError]
     string handle_enrollment(string request);
 
+    // Sets the experiments list in the CirrusClient's internal state.
     [Throws=NimbusError]
     void set_experiments(string experiments);
 };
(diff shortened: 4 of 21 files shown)


automation/run_python_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
jrbenny35, line 1: We could add ```set -e``` to exit if a command fails, instead of using ```|| exit``` for each command.

jrbenny35 (commented):
Just some quick thoughts I wanted to add, we can take a look more later.

components/nimbus/src/schema.rs
```diff
@@ -88,6 +88,37 @@ impl ExperimentMetadata for Experiment {
...
+        .get("data")
+        .ok_or(NimbusError::InvalidExperimentFormat)?;
+    let mut res = Vec::new();
+    for exp in data
+        .as_array()
+        .ok_or(NimbusError::InvalidExperimentFormat)?
+    {
+        // XXX: In the future it would be nice if this lived in its own versioned crate so that
+        // the schema could be decoupled from the sdk so that it can be iterated on while the
+        // sdk depends on a particular version of the schema through the Cargo.toml.
```
jaredlockhart, line 106: 💯 Yes I've been hoping we could add rust bindings to the Nimbus Experiment schema that's defined in nimbus-shared so we could do exactly this.  Maybe this is something we can do next half.

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -133,27 +135,39 @@ impl CirrusClient {
         let nimbus_id = Uuid::new_v4();
         let available_randomization_units =
             AvailableRandomizationUnits::with_client_id(client_id.as_str());
-        let th = NimbusTargetingHelper::new(targeting_attributes);
+        let ta = TargetingAttributes::new(self.app_context.clone(), request_context);
+        let th = NimbusTargetingHelper::new(ta);
         let enrollments_evolver =
             EnrollmentsEvolver::new(&nimbus_id, &available_randomization_units, &th);
+        let state = self.state.lock().unwrap();
```
jaredlockhart, line 142: Does this panic at runtime if it fails?

jeddai: Yep, it's worth noting a few things about that though.

1. It's our general paradigm used throughout the Nimbus client, as well as generally what is expected when using Mutexes according to the docs.
    * nimbus use: https://github.com/mozilla/application-services/blob/main/components/nimbus/src/nimbus_client.rs#L93
2. As long as I'm understanding the docs correctly, `lock` is going to block the current thread until it's able to acquire the Mutex. So if it's not available immediately it should be okay but I don't think we'll run into that very often in Python. 
3. FWIW I keep seeing the Python tests run in .5s, and it's running enrollment 1000 times plus some other stuff so I personally am not concerned. Happy to write some threaded tests if that'd be preferred

Mutex docs: https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock

The docs _seem_ to imply that panics only happen if you try to lock it a second time on the same thread, but I'm not entirely certain that's what the "panic" block in the docs means. I'll do some research and get back to you on that

components/nimbus/src/evaluator.rs
```diff
@@ -51,27 +51,28 @@ pub struct TargetingAttributes {
...
 impl TargetingAttributes {
-    pub fn new(app_context: AppContext, request_context: RequestContext) -> Self {
-        let (language, region) = request_context
-            .locale
-            .clone()
-            .map(split_locale)
-            .unwrap_or_else(|| (None, None));
+    pub fn new(app_context: AppContext, request_context: Map<String, Value>) -> Self {
+        let (language, region) = match request_context
+            .get("locale")
```
jaredlockhart, line 63: So `locale` is required in the request context?  Is there any way to remove it as a special case and not put any hard requirements on the structure of the targeting context?

jaredlockhart (commented):
Awesome, interface looks very nice, setting the experiment list looks nice, initialization and enrollment interfaces look correct.  I'm curious about all the naked unwraps, will those bubble back as Python exceptions?

jeddai:
> I'm curious about all the naked unwraps, will those bubble back as Python exceptions?

I'm actually not sure — I can run some tests and find out. Definitely worth having an answer to that question even if it's an unlikely case.

jaredlockhart:
> > I'm curious about all the naked unwraps, will those bubble back as Python exceptions?
> 
> I'm actually not sure — I can run some tests and find out. Definitely worth having an answer to that question even if it's an unlikely case.

Yeah exactly, as long as a panic in Rust space is wrapped in an Exception in Python space we should be fine.  What I wouldn't want is Rust to panic and then Python to `exit()`.  Also I'm curious if the SDK panics, does it lose its state?  This is one possible concern with making the SDK in memory stateful.

jeddai:
> What I wouldn't want is Rust to panic and then Python to exit(). Also I'm curious if the SDK panics, does it lose its state?

For sure on the first part. As for the panic/state piece I'll check on that as well. It's going to be difficult to cause that particular error on the mutex unwraps in Python tests, so I'm going to write a function that will throw that specific panic (unwrapping an errored result) and test with it. It won't be committed but I'll log the results in a comment for posterity.

components/nimbus/src/evaluator.rs
jeddai: Nope it's not required! I left it as optional so that if they choose to include it we parse it like we do for iOS/Android.

Just to quickly walk through the control flow here:
```rust
let (language, region) = match request_context // perform a `match` on the result of this execution
            .get("locale")            // `get` returns an Option and does not error on missing key
            .unwrap_or(&Value::Null)  // `unwrap_or` unwraps if it's there, and uses a different value if it's not. This changes the value in the execution chain from an `Option<Value>` to `Value` — but if it _wasn't_ present the `Value` will be JSON-equiv null
            .as_str()                 // `as_str` returns `Value` instance as an optional Rust `str` type if the `Value` is a string, otherwise it returns the option as `None`
        {
            Some(locale) => split_locale(locale.to_string()), // Here we handle the split if the string is present
            _ => (None, None),          // Here we handle the case where it's not, setting lang/region to `None`
        };
```

Though if we're not a fan of having any amount of extra logic around the targeting for Cirrus I can just remove this and the lang/region members entirely.

jrbenny35:
Is there a python formatter within the AS repo we could use?

jeddai:
> Is there a python formatter within the AS repo we could use?

Uh not that I'm aware of

jaredlockhart (commented):
@jhugman pointed out the [question mark operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator) when used with unwrap will return an `Error` fast with the return type being `Result<T>`.  Maybe that pattern can be applied here for cases where we unwrap to get the exception back from the Rust side?

jeddai:
> @jhugman pointed out the [question mark operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator) when used with unwrap will return an Error fast with the return type being Result<T>. Maybe that pattern can be applied here for cases where we unwrap to get the exception back from the Rust side?

I suppose we could, I'll also note I just tested it and it turns out rust panics are returned as an `InternalError` — an error type generated by UniFFI and exposed in the generated Python file. So we could also set up very explicit try/except blocks around it instead.

jeddai:
Also just tested the panic+state concern and a rust panic does not cause us to lose the experiments state in the client object.

components/nimbus/src/schema.rs
```diff
@@ -4,7 +4,7 @@
 
 use crate::{defaults::Defaults, enrollment::ExperimentMetadata, NimbusError, Result};
 use serde_derive::*;
-use serde_json::{Map, Value};
+use serde_json::{from_str, from_value, Map, Value};
```
jhugman, line 7: Nit: I'd prefer to see well known functions like this be used explicitly/qualified.

i.e. use

```rust
let foo: Value = serde_json::from_str(payload)?;
```

rather than

```rust
let foo: Value = from_str(payload)?;
```

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -67,21 +65,31 @@ impl Default for EnrollmentRequest {
...
+pub struct CirrusClient {
+    app_context: AppContext,
+    state: Mutex<InternalMutableState>,
+}
 
 impl CirrusClient {
-    pub fn new() -> Self {
-        Self {}
+    pub fn new(app_context: String) -> Self {
+        let app_context: AppContext = from_str(&app_context).unwrap();
```
jhugman, line 88:
```suggestion
    pub fn new(app_context: String) -> Result<Self> {
        let app_context: AppContext = from_str(&app_context)?;
```

A panic should really only be used for unusual errors that are neither caller nor callees fault (e.g. out of memory, or malformed installation, corrupted memory), rather than "you used me wrong".

components/nimbus/src/cirrus.udl
```diff
@@ -12,9 +12,12 @@ enum NimbusError {
 };
 
 interface CirrusClient {
-    constructor();
+    constructor(string app_context);
```
jhugman, line 15: To move from panic to `NimbusError`:
```suggestion
    [Throws=NimbusError]
    constructor(string app_context);
```

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -133,27 +135,39 @@ impl CirrusClient {
...
                 is_user_participating,
                 Default::default(),
-                next_experiments,
+                &state.experiments,
                 prev_enrollments,
             )?;
 
         let enrolled_feature_config_map =
-            map_features_by_feature_id(&enrollments, next_experiments);
+            map_features_by_feature_id(&enrollments, &state.experiments);
```
jhugman, line 153: IIRC, you had something to unpack the feature id --> raw feature JSON, rather than passing the `EnrolledFeatureConfig` back. What happened to that? Did I imagine that? 😊

components/nimbus/src/schema.rs
jhugman: Also: `serde_json::to_string`, `serde_json::to_value` etc.

jhugman (approved):
This is looking good.

I have some comments, which should be addressed, but not enough to hold you up for another review cycle.

Good job!

jrbenny35:
I am wondering what is the best way to add a python linter to this, as it should have one. @bendk any thoughts? It doesn't seem like there is any python in this repo, outside of taskcluster stuff

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -67,21 +65,31 @@ impl Default for EnrollmentRequest {
...
             is_user_participating: true,
-            next_experiments: Default::default(),
             prev_enrollments: Default::default(),
         }
     }
 }
 
 #[derive(Default)]
-pub struct CirrusClient {}
+pub struct InternalMutableState {
```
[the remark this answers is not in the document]
jeddai: Yep good call

CHANGES_UNRELEASED.md
```diff
@@ -25,7 +25,7 @@ Use the template below to make assigning a version number during the release cut
 - Added processing of command line arguments (or intent extras) to be driven by a command line tool. ([#5482](https://github.com/mozilla/application-services/pull/5482), [#5497](https://github.com/mozilla/application-services/pull/5497))
   - Requires passing `CommandLine.arguments` to `NimbusBuilder` in iOS.
   - Requires passing `intent` to `NimbusInterface` in Android.
-- Added Cirrus client object for working with Nimbus in a static, stateless manner ([#5471](https://github.com/mozilla/application-services/pull/5471)).
+- Added Cirrus client object for working with Nimbus in a static, stateless manner ([#5471](https://github.com/mozilla/application-services/pull/5471), [#5478](https://github.com/mozilla/application-services/pull/5478)).
```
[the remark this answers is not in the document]
jeddai: Yeah I hadn't rebased from main yet since that adjustment

components/nimbus/src/schema.rs
jeddai: Good to know — totally fine with that change. Out of curiosity, is there a particular reason for explicit qualification vs `use`ing?

jeddai:
> I am wondering what is the best way to add a python linter to this, as it should have one. @bendk any thoughts? It doesn't seem like there is any python in this repo, outside of taskcluster stuff

I'm on-board with adding one but I do want to note that if we decide to I will handle it in a separate PR.

components/nimbus/src/stateless/cirrus_client.rs
jeddai: So I don't _think_ I had ever had something like that, but I know we talked about something that could take a `FeatureConfig`, validate it against a FeatureManifest, and merge it into the feature default JSON object.

That has been done over in the FML library, as a part of this PR: https://github.com/mozilla/application-services/pull/5557/files#diff-c664e1568e3d1b6098e9fbc6422dc27647782af7afa92b0e957e3e191d82f2e4R139-R174 (it's still a draft, just wanted to get it up so Jared/Yashika could review the bindings)

automation/python-tests/test_cirrus.py
```diff
@@ -0,0 +1,41 @@
...
+    response = json.loads(client.handle_enrollment(request))
+
+    assert len(response["enrollments"]) == 1
+    assert "NotEnrolled" in response["enrollments"][0]["status"].keys()
+    assert len(list(response["enrolledFeatureConfigMap"].values())) == 0
+    assert len(response["events"]) == 0
+
+
+def test_failure_case_no_client_id(client, req):
+
```
jrbenny35, line 31: extra space could be removed

automation/python-tests/test_cirrus.py
```diff
@@ -0,0 +1,41 @@
...
+    request = req(client_id = None)
+
+    err = None
+    try:
+        client.handle_enrollment(request)
+    except NimbusError as e:
+        err = e
+
+    if err is None:
+        assert False, 'client.handle_enrollment did not throw an error'
```
jrbenny35, line 41: You could use the ```else``` clause here, at the end of the try catch, since it seems you are expecting only 1 type of error.

automation/python-tests/conftest.py
```diff
@@ -0,0 +1,98 @@
+import pytest
+import json
+from components.nimbus.cirrus import CirrusClient
+
+@pytest.fixture
```
jrbenny35, line 5: Can you add a space here, it should be 2 lines after imports.

jrbenny35 (changes requested):
Some thoughts

automation/python-tests/test_cirrus.py
```diff
@@ -0,0 +1,41 @@
+from components.nimbus.cirrus import NimbusError
+import json
+
+def test_one_experiment_should_enroll(client, req):
+    response = None
+    for i in range(0, 1000):
+        response = json.loads(client.handle_enrollment(req()))
```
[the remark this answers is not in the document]
jeddai: While it doesn't _do_ anything in particular, I personally like seeing the amount of time the tests take for 1000 enrollments. Mostly just for visibility into general speed than anything else.

I can remove it if it's a concern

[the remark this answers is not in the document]
jrbenny35: Could you add a separate test for it?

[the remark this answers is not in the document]
jeddai: Sure

jrbenny35 (approved):
Looks great, I'll find a solution for a python formatter

components/nimbus/src/evaluator.rs
yashikakhurana: I think eventually they will be using locale, they use it in unleash too

PR #5485, opened 2023-04-17
SYNC-3590: Track logins crypto deserialization errors with Sentry

bendk:
Report errors to Sentry and include some info on what we were trying to encrypt/decrypt when it happened

  - Note:


components/logins/src/encryption.rs
```diff
@@ -36,6 +36,19 @@ pub struct EncryptorDecryptor {
...
+
+impl<T> CryptoResultExt<T> for std::result::Result<T, jwcrypto::JwCryptoError> {
+    fn map_crypto_err(self, action: &str) -> Result<T> {
+        self.map_err(|e| Error::CryptoError {
+            action: action.into(),
+            error: e,
+        })
+    }
+}
+
```
bendk, line 51: I like the [extend](https://crates.io/crates/extend) crate for things like this.  We use it in UniFFI, so I think it's already moz-central approved.  Maybe we should add it to app-services, what do you think?

components/logins/src/error.rs
```diff
@@ -141,9 +145,8 @@ impl GetErrorHandling for Error {
                 })
                 .report_error("logins-migration")
             }
-            Self::CryptoError(_) => {
-                ErrorHandling::convert(LoginsApiError::IncorrectKey).log_warning()
-            }
+            Self::CryptoError { .. } => ErrorHandling::convert(LoginsApiError::IncorrectKey)
+                .report_error("logins-crypto-error"),
```
bendk, line 149: `IncorrectKey` may not be the the right error for the API.  It's possible that we're getting invalid ciphertext.  Once we get error reports back, we could consider changing it.

tarikeshaq: I might be misunderstanding what happens here, is the `action` we add to the `CryptoError` reported?

components/logins/src/login.rs
```diff
@@ -367,7 +367,11 @@ pub struct SecureLoginFields {
 
 impl SecureLoginFields {
     pub fn encrypt(&self, encdec: &EncryptorDecryptor) -> Result<String> {
-        encdec.encrypt_struct(&self)
+        encdec.encrypt_struct(&self, "encrypt SecureLoginFields")
```
tarikeshaq, line 370: nit suggestion: Maybe another way of ensuring that the `action` is set properly is to introduce a trait `Encryptable`  

```rs
trait Encryptable: Serializable {
  fn action_name() -> &'static str;

  fn encrypt(&self, encdec: &EncryptorDecryptor) -> Result<String> {
      encdec.encrypt_struct(&self)
  }
}

```

And changing the signature of `encrypt_struct` to

```rs
fn encrypt_struct<E: Encryptable>(&self, fields: &E) -> Result<String> {
   let name = fields.action_name();
   ....
} 
```

Thoughts?

components/logins/src/encryption.rs
```diff
@@ -36,6 +36,19 @@ pub struct EncryptorDecryptor {
...
+impl<T> CryptoResultExt<T> for std::result::Result<T, jwcrypto::JwCryptoError> {
+    fn map_crypto_err(self, action: &str) -> Result<T> {
+        self.map_err(|e| Error::CryptoError {
+            action: action.into(),
+            error: e,
+        })
+    }
+}
```
tarikeshaq, line 50: What do you think about moving this to the `error` module?

Also, is this an abstraction credit cards can use too? if we feel like it is, we can move it to our shared `types` crate and have the `Error` be an associated type and implement it for both

But we can also just keep this here for logins if you feel like that would be pre-mature optimization!

tarikeshaq (changes requested):
I like the idea of reporting it to sentry!

Only requesting changes because I think the `action` isn't going to be sent to senty, and I wanted to confirm that is expected

components/logins/src/encryption.rs
tarikeshaq: eeh I just realized moving this trait to another crate means we can't implement it for `std::result::Result` - never mind this comment - leaving it open for the "moving this to the `error` module part

tarikeshaq (approved):
Removing my request for changes, because @bendk kindly explained how the error is reported to me 🙏 

The rest are suggestions/nits, so LGTM!

bendk:
@tarikeshaq I still like your suggestion to make this more general.  I just pushed a new version that moves more of the core functionality to the `support/jwcrypto` crate.  I made both logins and autofill use it, so both of them should get the improved reporting.

components/support/jwcrypto/src/encdec.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+impl<E: From<EncryptorDecryptorError>> EncryptorDecryptor<E> {
+    /// Create a key that can be used to construct an EncryptorDecryptor
+    pub fn create_key() -> Result<String, E> {
+        let key = crate::Jwk::new_direct_key(None).to_encdec_result("create key")?;
+        Ok(serde_json::to_string(&key).to_encdec_result("create_key (serialization)")?)
+    }
+
+    pub fn new() -> Result<Self, E> {
```
tarikeshaq, line 30: What do you think about having `new` take a required key, and the other function be `new_with_random_key` or something like that to make it explicit that a random key is being generated?

components/support/jwcrypto/src/encdec.rs
```diff
@@ -0,0 +1,142 @@
...
+/// This struct wraps the jwcrypto functionality in a convenient package.  Also, it uses
+/// `EncryptorDecryptorError`, which includes a description that can help track down where
+/// encryption errors are happening.
+///
+/// EncryptorDecryptor takes a generic Error parameter that it uses for its results.  This can be
+/// anything that implement `From<EncryptorDecryptorError>`, typically it's the internal error type
+/// for a crate.
+pub struct EncryptorDecryptor<E = EncryptorDecryptorError> {
+    jwk: Jwk,
+    phantom: PhantomData<E>,
```
tarikeshaq, line 20: nit: do we think https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check applies here?

components/support/jwcrypto/src/error.rs
```diff
@@ -22,4 +22,25 @@ pub enum JwCryptoError {
...
+/// This has a nice interaction with `error_support::ErrorHandling`. Crates will typically define
+/// an internal error with a variant that wraps `EncryptorDecryptorError`.  As long as the Display
+/// implementation for that variant includes the `EncryptorDecryptorError` string, then the
+/// description will be included in error reports to Sentry and other systems.
+#[derive(Error, Debug)]
+#[error("{from} ({description})")]
+pub struct EncryptorDecryptorError {
+    // Underlying error
+    pub from: JwCryptoError,
+    pub description: String,
```
tarikeshaq, line 45: Wow! This is neat!

How do we feel this works vs having `EncryptorDecryporError` be "just" another variant in `JwCryptoError`?

That way we can even remove the `E` type parameter from `EncryptorDecryptor` and have each function in `EncryptorDecryptor` just return `Result<T, JwCryptoError>` and it's the consumer's job to turn that into its own error (and report it)

i.e: I realize this sounds a lot like the opposite of what I was suggesting earlier so sorry for that, curious what you think works best here

components/support/jwcrypto/src/encdec.rs
```diff
@@ -0,0 +1,142 @@
...
+/// High-level struct for handling Encryption/Decryption
+///
+/// This struct wraps the jwcrypto functionality in a convenient package.  Also, it uses
+/// `EncryptorDecryptorError`, which includes a description that can help track down where
+/// encryption errors are happening.
+///
+/// EncryptorDecryptor takes a generic Error parameter that it uses for its results.  This can be
+/// anything that implement `From<EncryptorDecryptorError>`, typically it's the internal error type
+/// for a crate.
+pub struct EncryptorDecryptor<E = EncryptorDecryptorError> {
```
tarikeshaq, line 18: I love that this is here now and used by both logins and autofill!!

components/support/jwcrypto/src/encdec.rs
```diff
@@ -0,0 +1,142 @@
...
+    jwk: Jwk,
+    phantom: PhantomData<E>,
+}
+
+impl<E: From<EncryptorDecryptorError>> EncryptorDecryptor<E> {
+    /// Create a key that can be used to construct an EncryptorDecryptor
+    pub fn create_key() -> Result<String, E> {
+        let key = crate::Jwk::new_direct_key(None).to_encdec_result("create key")?;
```
tarikeshaq, line 26: nit: is there a way for us to move around the magic strings? either make them an enum `EncryptionDecryptionReason`, or maybe consts? - thoughts?

tarikeshaq (approved):
Did a quick pass end of my day, looks good so far!

Most my comments are nits and questions!

components/support/jwcrypto/src/error.rs
bendk: I like the idea and I don't even think we need a new variant.  But many of the current variants would need to store `description` strings.

The reason for a different error type is that I was trying to leave the existing API alone and define a new one on-top of it.  `fxa-client` is still using the current API and I don't think I want to change that code yet.

How about as a follow up as part of the fxa work?

components/support/jwcrypto/src/encdec.rs
bendk: TIL

components/support/jwcrypto/src/encdec.rs
bendk: The issue is that `EncryptorDecryptorError` always expects a string.  I don't think it would be too hard to clean this up, but let's do it as a follow-up (#5493).

PR #5510, opened 2023-04-25
Updating the release process [ci full]

bendk:
There's two parts to this PR: updating the a-s release process and adding taskcluster code to build artifacts for the new process.

## New release process

We don't really need to "cut releases" anymore, just push commits to release branches.

I modified `automation/prepare-release.py` to handle the new process.  It should create PRs like this:
  - [Updating the release branch for a new rversion](https://github.com/mozilla/application-services/pull/5507)
  - [Updating main to start a new major version](https://github.com/mozilla/application-services/pull/5508)

After those PRs are merged and we've built the release, then you need to tag the version with `automation/tag-release.py`

The new release process simplifies a lot of things and I think that means we can just keep a `CHANGELOG.md` and forget about `CHANGES_UNRELEASED.md`

## Taskcluster stuff

I refactored the taskcluster code more-or-less execute the nightly process, but do it for a real release.  It's not working yet, I'm using this PR to test it.

@ahal couple questions that you can hopefully answer while I'm away:
  - I'm hoping to get this code to run whenever there's a push to a `release-v*` branch.  However, when I tried to test that I didn't see any taskcluster actions.  Do we need to do something to git taskcluster to run on github pushes?  It's also very possible that this is already working and I just did something wrong when testing.
  - I was hoping to use a `kind-dependency` to get the [release-publish](https://github.com/mozilla/application-services/blob/599f7ae8bd7f8b69052239dabe14152903ab8215/taskcluster/ci/release-publish/kind.yml) task to only run after [build-summary](https://github.com/mozilla/application-services/blob/599f7ae8bd7f8b69052239dabe14152903ab8215/taskcluster/ci/build-summary/kind.yml) succeeds.  My thinking is that build-summary only succeeds if all the build task succeed, which means we would only publish the `release.json` file after we knew everything worked.  However,  the `kind-dependency` doesn't seem to translate into an actual dependency in the taskcluster tasks.  `build-publish` only waits on a few tasks then runs.
  
  - Note:


docs/howtos/releases.md
```diff
@@ -0,0 +1,99 @@
...
+
+Release builds are generated from the `release-vXXX` branches.  Whenever a commit is pushed there,
+we build all artifacts needed for the release.  Once we're ready to publish the release, we run the
+taskcluster release promotion action which signs and publishes the artifacts.
+
+TODO: explain release-promotion more.
+
+## What to do at the end of a nightly cycle
+
+1. Start this 2 workdays before the nightly cycle ends
```
bendk, line 23: I just made up this 2 workdays number, not sure when we should start it.  We should make sure we have plenty of time to finish before firefox-android or firefox-ios want to start cutting their releases.

docs/howtos/releases.md
```diff
@@ -0,0 +1,99 @@
...
+index 8cd923873..6482018e0 100644
+--- a/.buildconfig-android.yml
++++ b/.buildconfig-android.yml
+@@ -1,4 +1,4 @@
+-libraryVersion: 114.0a1
++libraryVersion: 114.0
+ groupId: org.mozilla.appservices
+```
+
+ * Create a PR on `main` that starts a new CHANGELOG haeder.
```
bendk, line 41: Eventually, maybe this could be rolled into the `promote` action.

docs/howtos/releases.md
```diff
@@ -0,0 +1,99 @@
...
+
+3. Trigger release-promotion once the PRs are approved, merged, and CI has completed
+
+ * TODO: explain how to do this
+
+4. Tag the release with `automation/tag-release.py [major-version-number]`
```
bendk, line 47: And eventually, maybe this could be rolled into the `ship` action.

ahal:
> I'm hoping to get this code to run whenever there's a push to a release-v* branch. However, when I tried to test that I didn't see any taskcluster actions. Do we need to do something to git taskcluster to run on github pushes? It's also very possible that this is already working and I just did something wrong when testing.

This is because we're ignoring pushes to all non-main branches here:
https://github.com/mozilla/application-services/blob/main/.taskcluster.yml#L100

Luckily VPN is set up the way you want, so it should just be a matter of copying them:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/.taskcluster.yml#L98

> I was hoping to use a kind-dependency to get the [release-publish](https://github.com/mozilla/application-services/blob/599f7ae8bd7f8b69052239dabe14152903ab8215/taskcluster/ci/release-publish/kind.yml) task to only run after [build-summary](https://github.com/mozilla/application-services/blob/599f7ae8bd7f8b69052239dabe14152903ab8215/taskcluster/ci/build-summary/kind.yml) succeeds. My thinking is that build-summary only succeeds if all the build task succeed, which means we would only publish the release.json file after we knew everything worked. However, the kind-dependency doesn't seem to translate into an actual dependency in the taskcluster tasks. build-publish only waits on a few tasks then runs.

Yeah, all `kind-dependency` does is make sure those kinds are loaded before the current kind. You still need to make the dependency link to whatever specific tasks are desired within the kind dependencies. You can do this by specifying the `dependencies` key manually, using a transform, or in this case, I'd recommend using the [multi_dep](https://github.com/mozilla/application-services/blob/main/taskcluster/app_services_taskgraph/loader/multi_dep.py) loader.

The `multi_dep` loader is designed for cases when you want to automatically create tasks from a set of indeterminate dependencies. Here's an example of how it's used:
https://github.com/mozilla/application-services/blob/main/taskcluster/ci/signing/kind.yml

This will create one `signing` task for every `module-build` task. Come to think of it, you may want your summary task to run after `signing` rather than `module-build`.

Aside: I'm actually working on implementing this functionality in a nicer way directly into core Taskgraph over in [this PR](https://github.com/taskcluster/taskgraph/pull/236), but it's not worth blocking on that.

ahal:
Oh, and if we start running tasks on `releases/*` don't forget to add Github branch protection rules at least as strict as the ones on `main` :).

docs/howtos/releases.md
```diff
@@ -0,0 +1,99 @@
...
+  https://firefox-ci-tc.services.mozilla.com/tasks/index/project.application-services.v2.branch.main.latest.taskgraph/decision-nightly
+- If you don't see a decision task from the day before, then contact releng.  It's likely that the cron decision task is broken.
+
+## Release builds
+
+Release builds are generated from the `release-vXXX` branches.  Whenever a commit is pushed there,
+we build all artifacts needed for the release.  Once we're ready to publish the release, we run the
+taskcluster release promotion action which signs and publishes the artifacts.
+
+TODO: explain release-promotion more.
```
ahal, line 19: Here's [a link](https://firefox-source-docs.mozilla.org/taskcluster/release-promotion.html) to the Gecko docs on the topic if you want to draw inspiration.

docs/howtos/releases.md
ahal, line 41: typo: s/haeder/header

docs/howtos/releases.md
```diff
@@ -0,0 +1,99 @@
...
+-libraryVersion: 114.0a1
++libraryVersion: 114.0
+ groupId: org.mozilla.appservices
+```
+
+ * Create a PR on `main` that starts a new CHANGELOG haeder.
+
+3. Trigger release-promotion once the PRs are approved, merged, and CI has completed
+
+ * TODO: explain how to do this
```
ahal, line 45: Ideally this will be done via shipit, but until then we may want only Releng to have permission to do this. This way it acts as a poor man's dual signoff.

docs/howtos/releases.md
ahal: Yes, we have a `treescript` which can not only tag releases, but land changes to e.g, increment a version number (if desired).

taskcluster/app_services_taskgraph/__init__.py
```diff
@@ -33,6 +32,8 @@ def register(graph_config):
         # "nightly" for the nightly builds.  Other strings indicate making a
         # preview build for a particular application-services branch.
         'preview-build': str,
+        # Release type.  We set this to "release" for non-nightly builds, it defaults to "nightly".
+        'release': Optional(str),
```
ahal, line 36: Gecko calls this parameter `release-type`, might be worth making it consistent.

bendk:
@ahal Thanks for the info on those issues, I just pushed a [commit](https://github.com/mozilla/application-services/pull/5510/commits/ac3444bfc87ab01919998d55ed92dbdd4e54421c) that I think should fix them.

For the dependencies, I went with a transform instead of a loader, since we already implemented this code for `build-summary`.  I think it should work and we can clean it up once that PR you mentioned lands.

I'm going to update the branch protection settings now.

If you think this looks good, then I think we can merge this to main.

taskcluster/app_services_taskgraph/target_tasks.py
```diff
@@ -10,18 +10,19 @@
 def target_tasks_pr_skip(full_task_graph, parameters, graph_config):
     return []
 
-@_target_task('preview')
-def target_tasks_preview(full_task_graph, parameters, graph_config):
-    # Run all tasks for preview builds
+@_target_task('release')
+def target_tasks_release(full_task_graph, parameters, graph_config):
+    # Run all tasks for release builds
```
ahal, line 15: What's the intent of this target task phase exactly? Shouldn't it be called `full` or something?

Fyi I'll be adding new methods for the `release_promotion` aspects, so I'm a bit unclear if this one will still be needed or not.

bendk: The intention is to run the tasks needed for a release (nightly or otherwise).  `full` is probably a better name, but we also have `pr-full` which is run when you add `[ci: full]` in the PR title.  It's all a bit confusing to me.

Maybe `release` and `pr-full` should both be renamed to `full` and the only difference between a release, nightly, and a PR with `[ci: full]` is the indexing.

taskcluster/app_services_taskgraph/__init__.py
bendk: Sounds good, I updated the param.

taskcluster/app_services_taskgraph/target_tasks.py
ahal: Yeah that latter suggestion sounds sensible, though we don't need to worry about it now. I was mostly curious.

For release promotion, we'll schedule only the tasks that are needed for the actual release. Things like builds and tests presumably ran on the push already (e.g, we "promote" an existing build). So fyi this target_tasks method probably won't end up being used there.

taskcluster/app_services_taskgraph/transforms/release_publish.py
```diff
@@ -0,0 +1,65 @@
...
+    if config.params['level'] == '3':
+        if config.params.get('preview-build') is None:
+            maven_channel = "maven-production"
+        else:
+            maven_channel = "maven-nightly-production"
+    else:
+        if config.params.get('preview-build') is None:
+            maven_channel = "maven-staging"
+        else:
+            maven_channel = "maven-nightly-staging"
```
ahal, line 25: suggestion: you could reduce this to:
```
instance = "production" if config.params["level"] == "3" else "staging"
nightly = "-nightly" if config.params.get("preview-build") else ""
maven_channel = f"maven{nightly}-{instance}"
```

taskcluster/app_services_taskgraph/transforms/release_publish.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Convert kind dependencies into soft-dependencies
+
+    The `release-publish` task lists the `build-summary` task as a kind dependency, but we need a
+    transform to setup the actual dependency.  When
+    https://github.com/taskcluster/taskgraph/pull/236 is merged, we could simplify this code.
+    """
+    for task in tasks:
+        task.setdefault("soft-dependencies", [])
+        task["soft-dependencies"] += [
```
ahal, line 61: You *might* want to use the `if-dependencies` feature here:
https://taskcluster-taskgraph.readthedocs.io/en/latest/concepts/task-graphs.html#if-dependencies

In practice soft-dependencies will probably work exactly the same, but the difference is that this task will still run even if all its dependencies get optimized away. But the odds of that happening are very small, so I wouldn't worry about it too much.

ahal (approved):
Thanks, this looks great! I think we'll need to make a few tweaks for release promotion, but this is a good base to build off of and we can do those in the relpro PR.

taskcluster/app_services_taskgraph/target_tasks.py
bendk: The target tasks method should be used for the push though right?  I renamed the key to `full` and added a comment on how I think it should work.  Hopefully that's enough good for now.

bendk:
Pushed what's hopefully the last commit.  I tried to clean up / document the target tasks method some.  I also made it so `swift-build` and `nimbus-fml-build` run when `[ci full]` is in the title, but they only use the index when we're doing an actual release build.

Does the last commit seem okay?  I agree that this probably will need to change once we implement release-promotion, but hopefully it's good enough as a base to get started.

taskcluster/ci/swift/kind.yml
```diff
@@ -17,9 +17,12 @@ kind-dependencies:
 tasks:
   build:
     attributes:
+      run-on-pr-type: full-ci
       resource-monitor: true
+      release-routes:
+        - index.project.application-services.v2.swift.{appservices_version}
```
ahal, line 23: This is the same route from before, so no need to block this PR. Though I notice that there's no `release-type` in the route. This means if you have both staging and prod release types, the former could overwrite the latter. Maybe it's not a problem in this case, but would be good to double check this.

ahal (approved):
Thanks, this looks good!

taskcluster/ci/swift/kind.yml
bendk: Good point.  I don't think it'll be an immediate issue, but it's good to fix.  I made this issue for it: https://github.com/mozilla/application-services/issues/5545.

I think we can do this as a follow-up, hopefully next week.

PR #5514, opened 2023-04-27
Building nss now forces the integrated assember to work with ndk 25

### LINKED ISSUE #5491: Can't build NSS for Android on Mac post NDK upgrade to 25
After the NDK upgrade, the `libs/verify-android-environment` script fails for mac users. We've established there is a problem with the build script attempting to run the system assembler and passing in a `--64` option, which the mac assembler doesn't support.

In the mean time, a few notes:

- This does not affect anyone with an already existing build of NSS (i.e you've run `verify-android-environment` before the upgrade) and still have a working `libs/android` directory
- If this affects you, one remedy is to:
   - Rever https://github.com/mozilla/application-services/pull/5442 locally
   - Install NDK `21.3.6528147` in Android Studio (or verify it's already installed)
   - delete `libs/android`
   - run `libs/verify-android-environment.sh`
   - Remove any changes you have because you reverted https://github.com/mozilla/application-services/pull/5442
   - Now android builds should work, but will break again if you delete `libs/android`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3694)
┆Sprint End Date: 2023-05-04

mhammond:
Fixes #5491.

### DIFF
--- a/libs/build-nss-android.sh
+++ b/libs/build-nss-android.sh
@@ -65,6 +65,7 @@ gyp -f ninja-android "${NSS_SRC_DIR}/nss/nss.gyp" \
   --depth "${NSS_SRC_DIR}/nss/" \
   --generator-output=. \
   -DOS=android \
+  -Dforce_integrated_as=1 \
   -Dnspr_lib_dir="${NSPR_BUILD_DIR}/dist/lib" \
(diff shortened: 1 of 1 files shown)


skhamis (approved):
Really nice find!!

I was able to reproduce the `--64` issue and then applied this patch and the build works! so LGTM!

PR #5525, opened 2023-05-01
Adds sync15.udl to nightly builds for iOS

tarikeshaq:
While testing the nightly builds for iOS, came across the fact that our other components now depend on sync15, but it wasn't included in the nightly builds


mhammond (approved):
oops :)

PR #5528, opened 2023-05-01
Bumps cross-compiled NSS artifact for mac [ci full]

tarikeshaq:
Bumps the NSS artifact for mac to a more latest one


skhamis (approved):
Thanks for this. I also created https://github.com/mozilla/application-services/issues/5529 to keep on record that we should now get this process moved over to something more automated.

PR #5531, opened 2023-05-01
Fixes sync15 not referencing the megazord

tarikeshaq:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1830793

the new sync15 DeviceType type can't be accessed from Fenix because it's attempting to be loaded through it's own shared object.

Exposing it through the megazord


mhammond (approved):
oops, thanks.

PR #5534, opened 2023-05-02
Update sqlcipher to version 4.5.4

rvandermeulen:
## 4.5.4 - April 2023
- Updates baseline to upstream SQLite 3.41.2
- Updates minimum Apple SDK versions in podspec for new Xcode compatibility
- Return runtime OpenSSL version from PRAGMA cipher_provider_version (instead of hardcoded value)
- Adds guard against zero block size and crash if cryptographic provider initialization fails
- When an ATTACH occurs creating a new encrypted database as the first operation after keying the main database, the new database will have the same salt value.


tarikeshaq (approved):
Given Ben is out on PTO, this LGTM!

PR #5543, opened 2023-05-05
Taskcluster nimbus-fml / nimbus-cli build [ci full]

bendk:

  - Note:


bendk:
Looks like this one is working, WDYT?

You can see the nimbus-fml and nimbus-cli zip files in the artifacts section of this task: https://firefox-ci-tc.services.mozilla.com/tasks/Q3xEk2tCQGu9GnNDQ_Pffw

bendk:
BTW, the task I linked above just assembles the zip files, you can see the tasks that generate the individual binaries here: https://firefox-ci-tc.services.mozilla.com/tasks/groups/T-rpicTbRKOsE-PKIOSkDw#nimbus-build

jhugman (commented):
I'm not sure I can be the one to Approve this.

However, a few questions: 

- if I'm programmatically fetching the latest from main, how would I determine the URL to fetch it from?
- if I wanted to get a version of the lib (so I could tell whether to upgrade or not), would you suggest checking `Cargo.toml` on main, or some other way (perhaps in taskcluster)?
- if I wanted to distribute the `nimbus-cli` binaries (for different architectures) as different artifacts would that be advisable/difficult/annoying?

bendk:
>     * if I'm programmatically fetching the latest from main, how would I determine the URL to fetch it from?
> 
>     * if I wanted to get a version of the lib (so I could tell whether to upgrade or not), would you suggest checking `Cargo.toml` on main, or some other way (perhaps in taskcluster)?

Right now we built and publish `nimbus-fml` with each nightly (https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/project.application-services.v2.nightly.latest/artifacts/public%2Fbuild%2Fnightly.json).  It would be trivial to add a link for `nimbus-cli` there.

But maybe it would be better to have a separate system since it seems like there are some differences:
  - Devs are going to always want the latest `nimbus-cli`, rather than pinning the version to the application-services version (which is how I think we want to handle `nimbus-fml`).
  - We want to build `nimbus-cli` on every push rather than do a nightly build.

Does that seem correct?  If so, I think we could set up a separate index and publish a `nimbus-cli.json` file with:
  - Links to the binaries
  - Git commit it was built from, to help with auto-upgrading.


> 
>     * if I wanted to distribute the `nimbus-cli` binaries (for different architectures) as different artifacts would that be advisable/difficult/annoying?

I don't think this would be too hard.

tarikeshaq (approved):
The scripts looks okay to me, but I'd defer the taskcluster parts of this to folks more familiar with taskcluster so will add releng to the review!

We might also want to wait for @jhugman 's +1 to the ongoing discussion but if he's happy with how we deliver the binaries, I'm happy!

jhugman (approved):

bendk:
> I think my ideal would be if it's the last commit hash to touch the nimbus-cli, rather than QA and devs installing new versions every day. I'm not sure if that was what you meant however.

I think this would be ideal as well, but "touching" is hard to define.  It should include changes to `rust-toolchain.toml` as well as any dependent crates.  If you wanted to setup a list of globs, I think we could follow that.

taskcluster/app_services_taskgraph/transforms/nimbus.py
```diff
@@ -0,0 +1,124 @@
...
+        'nimbus-fml',
+        'nimbus-cli',
+    ]
+    for task in tasks:
+        dependencies = {}
+        fetches = {}
+
+        for (label, build_task) in config.kind_dependencies_tasks.items():
+            # Since `nimbus-build` is our only kind-dependency, this loops through all the
+            # nimbus-build tasks.  For each task, fetch it's artifacts.
```
ahal, line 97: Might be worth replacing this comment with:
```
if build_task.kind != "nimbus-build":
    continue
```

just in case more kind-dependencies get added in the future.

ahal (approved):

PR #5546, opened 2023-05-06
Don't trigger a nightly build on every push to main

bendk:
Added `run-on-tasks-for` for the nightly publish task so that it doesn't run on every push to main.  I'm not sure if this is exactly the right configuration -- it likely will need to change again when we implement release-promotion.  But this matches how beetmover tasks work, which seems like a good place for now.

Added taskcluster test params for a push to main.

  - Note:


bendk:
Side question: has any other project implemented tests to try to prevent breakage like this?  It seems like we could setup a test suite that runs `taskgraph  target --json` against the each of the test parameters and check that the result matches the expected output.

rvandermeulen:
Makes sense to me, but this isn't something I should review.

bhearsum (approved):
> it likely will need to change again when we implement release-promotion.

Yeah, you'll probably end up with a `shipping-phase` at that point. But this certainly seems like the reasonable thing to do in the meantime (I see VPN doing it as well...).

PR #5556, opened 2023-05-08
Exposes retrieving an event for a push message [firefox-android: tarikeshaq:1832676]

tarikeshaq:

  - Note:

### DIFF
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -457,7 +457,7 @@ interface FirefoxAccount {
   //
   // **💾 This method alters the persisted account state.**
   //
-  // Applications should call this method whenever they receive a push notiication on subscription
+  // Applications should call this method whenever they receive a push notification on subscription
   // endpoint previously registered with the Firefox Accounts server. Such messages typically indicate
   // a noteworthy change of state on the user's account, such as an update to their profile information
   // or the disconnection of a client. The [`FirefoxAccount`] struct will update its internl state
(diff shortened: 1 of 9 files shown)


tarikeshaq:
Wanted to write down some context before I move on to something else and get back to this:

We want this because:
In Firefox iOS, the notification service runs in its own process. If the notification service keeps using the existing `handle_push_messages` to handle push messages, it will modify the internal `last_command_index`. The notification service being in its own process, means that it could race with the main process, causing the main process to overwrite the `last_command_index` the notification service set.


We would like to move towards a model where the notification service **does not** modify the last command index, and instead, always instructs the main process to `pollDeviceCommands` when a push notification arrives.

Much of the above logic would live in the notification service extension code (In firefox-ios), but we would need an API from the fxa client that does not modify the command index, and instead acts as a pure-ish API to return the decrypted command (send tab) associated with a single push message.

skhamis (approved):
This is slightly rubber-stampy but it makes sense to me and looks good! Awesome job!

components/fxa-client/src/internal/error.rs
```diff
@@ -123,6 +123,12 @@ pub enum ErrorKind {
 
     #[error("Integer conversion error: {0}")]
     IntegerConversionError(#[from] std::num::TryFromIntError),
+
+    #[error("Command not found by fxa")]
+    CommandNotFound,
+
+    #[error("Invalid Push Event")]
+    InvalidPushEvent,
```
mhammond, line 131: I wonder if "unknown" is a better description than "invalid"?

mhammond (changes requested):

tarikeshaq:
(regardless of review I'll wait till tomorrow to land this - I'll have a PR for Fenix fixing the breaking change then)

components/fxa-client/src/fxa_client.udl
```diff
@@ -451,22 +451,24 @@ interface FirefoxAccount {
...
-  // or the disconnection of a client. The [`FirefoxAccount`] struct will update its internl state
-  // accordingly and return a list of [`AccountEvent`] structs describing the events, which the application
+  // Applications should call this method whenever they receive a push notification from the Firefox Accounts server.
+  // Such messages typically indicate a noteworthy change of state on the user's account, such as an update to their profile information
+  // or the disconnection of a client. The [`FirefoxAccount`] struct will update its internal state
+  // accordingly and return an individual [`AccountEvent`] struct describing the event, which the application
   // may use for further processing.
   //
+  // It's important to note if the event is [`AccountEvent::CommandReceived`], the caller should immediately call
+  // [`FirefoxAccount::poll_device_commands`]
```
mhammond, line 467: nano doc nit: The "immediately" is over-emphasized here because the real point is something like "If the event is ..., the caller should not actually process the event - it should instead use the event information to notify the user, then arrange for `poll_device_commands()` to be called so that this, and any missed commands, are all correctly processed and the state of the commands correctly persisted to disk", or something like that?

mhammond (approved):
fantastic, thanks!

tarikeshaq:
This is ready to land, but want to get approvals on both: 
- iOS: https://github.com/mozilla-mobile/firefox-ios/pull/14305 
- Android: https://github.com/mozilla-mobile/firefox-android/pull/2000

tarikeshaq:
Failures are not relevant to this PR, the test suite for FxA runs OK and I verified locally against both iOS and Android

PR #5560, opened 2023-05-09
Don't build nightlies if there isn't a new commit

bendk:
Added a new `nightly` target tasks method.  That works like `full`, except it checks the `nightly.json` file to see if we're still on the same commit as the last nightly build.

  - Note:

### DIFF
--- a/taskcluster/app_services_taskgraph/__init__.py
+++ b/taskcluster/app_services_taskgraph/__init__.py
@@ -83,7 +83,6 @@ def get_decision_parameters(graph_config, parameters):
         # We don't have a great way of determining if something is a nightly or
         # not.  But for now, we can assume all cron-based builds are nightlies.
         parameters["preview-build"] = "nightly"
-        parameters["target_tasks_method"] = "full"
         parameters["release-type"] = "nightly"
 
     parameters['branch-build'] = branch_builds.calc_branch_build_param(parameters)
--- a/taskcluster/app_services_taskgraph/target_tasks.py
+++ b/taskcluster/app_services_taskgraph/target_tasks.py
@@ -2,9 +2,13 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import json
+import logging
 
 from taskgraph.target_tasks import _target_task, filter_for_tasks_for
+from taskgraph.util import taskcluster
 
+logger = logging.getLogger(__name__)
 
 @_target_task('pr-skip')
 def target_tasks_pr_skip(full_task_graph, parameters, graph_config):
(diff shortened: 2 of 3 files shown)


rvandermeulen:
I mentioned this to @JohanLorenzo already, but just posting it here for posterity as well. At a high level, I'm totally in favor of making this change. It makes perfect sense to not waste resources building & shipping multiple builds off the same revision.

However, I do want to point out one area where the current A-S nightly flow is lacking which this change will exacerbate if not addressed prior to landing. Right now, the nightly graphs for A-S are not easily discoverable - the only reliable way I know of is finding the latest build in the taskcluster index and then backtracking to a taskgraph from there. This is particularly important when trying to track down a busted update.

Especially now that we won't attempt to beetmove the nightly artifacts until all builds are finished, one of the only indications we currently have that something went wrong with the last Nightly build is the *lack* of a bump PR in firefox-android every day. When this patch lands, it will be very difficult to discern whether the lack of a nightly bump PR is due to something going wrong with the build or just lack of new changes. Ideally, we'd either be able to easily find the nightly decision task from the Github commit checks, or better yet, the application-services repo would be visible on Treeherder so we can easily see them there.

taskcluster/app_services_taskgraph/__init__.py
```diff
@@ -82,7 +82,7 @@ def get_decision_parameters(graph_config, parameters):
         # We don't have a great way of determining if something is a nightly or
         # not.  But for now, we can assume all cron-based builds are nightlies.
         parameters["preview-build"] = "nightly"
-        parameters["target_tasks_method"] = "full"
+        parameters["target_tasks_method"] = "nightly"
```
jcristau, line 85: Might be a stupid question.. why do we set target_tasks_method here?  If we're coming from cron, target-tasks-method should already be set (in .cron.yml)?

bendk:
More visibility into the taskgraph seems great to me.  However, I think we should know if a nightly build fails because of the slack alert that runs from the `build-summary` task.  I'm not totally sure that alert is working properly because we haven't had a build failure yet, but it was reporting errors properly back when it was just doing the `branch-build` test.

taskcluster/app_services_taskgraph/target_tasks.py
```diff
@@ -22,6 +25,20 @@ def target_tasks_pr_skip(full_task_graph, parameters, graph_config):
 def target_tasks_release(full_task_graph, parameters, graph_config):
     return full_task_graph.tasks
 
+@_target_task('nightly')
+def target_tasks_nightly(full_task_graph, parameters, graph_config):
+    try:
+        nightly_json = taskcluster.get_artifact_from_index(
```
jcristau, line 31: Any reason not to look for `project.application-services.v2.branch.main.revision.{head_rev}.taskgraph.decision-nightly` instead? (That would be more in line with how other projects handle this)

bendk: No reason at all, I just didn't realize it was that simple.  Updated the code to do this.

taskcluster/app_services_taskgraph/__init__.py
bendk: Not a stupid question at all, I agree that this line isn't needed.

jcristau (approved):
Thanks.

PR #5563, opened 2023-05-10
Mac taskcluster improvements [ci full]

bendk:
Move OSX Rust setup and library building to toolchain tasks.  This way we can only build them once, rather than in each task that requires it.
    
- Updated `setup-fetched-rust-toolchain.sh` to work with both the docker worker and mac worker.
- Build the NSS/SQJCipher libraries as a toolchain task that other tasks can depend on.  Right now, we're building these libraries in each task.

  - Note:

### DIFF
--- a/taskcluster/app_services_taskgraph/transforms/worker.py
+++ b/taskcluster/app_services_taskgraph/transforms/worker.py
@@ -1,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from taskgraph.transforms.base import TransformSequence
+
+transforms = TransformSequence()
+
+@transforms.add
+def setup_worker(_, tasks):
+    for task in tasks:
+        task_name = task['name']
+        try:
+            worker_type = task['worker-type']
+        except KeyError:
+            raise ValueError(f"worker-type not set for {task_name}")
+        if worker_type == 'b-linux':
+            worker = task.setdefault('worker', {})
+            worker['docker-image'] = {'in-tree': 'linux'}
+        else:
+            raise ValueError(f"Unknown worker type for {task_name} ({worker_type})")
+        yield task
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -15,7 +15,8 @@ targets = [
     "x86_64-linux-android",
     "aarch64-apple-ios",
     "x86_64-apple-ios",
-    "aarch64-apple-ios-sim"
+    "aarch64-apple-ios-sim",
+    "aarch64-apple-darwin",
 ]
 # The "rust-src" component is currently required for building for the M1 iOS simulator.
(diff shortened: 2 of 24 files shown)


tarikeshaq:
I imagine this is safe to close @bendk ?

bendk:
This one was blocked by the lack of a L3 mac worker.  I believe we're going to get one very soon, so let's keep it around.

bendk:
Looks like this one is working.  The `fetch-nss-artifact` task is failing because the source link has expired, but that's a separate [issue](https://github.com/mozilla/application-services/issues/5681).

I think the optimization is real.  In this branch, the [nimbus-build-fml-*](https://firefox-ci-tc.services.mozilla.com/tasks/dnRI5NfKSeOFoN0q46-zDA) tasks are taking about 5 minutes, on main [they take about 10](https://firefox-ci-tc.services.mozilla.com/tasks/Nh8dEFdURhGrhpOvx7a9qg).  5 minutes speedup per OSX task seems pretty nice.

taskcluster/app_services_taskgraph/transforms/toolchain.py
```diff
@@ -41,3 +41,10 @@ def resolve_keys(config, tasks):
             routes.append(TOOLCHAIN_OLD_INDEX[task['name']].format(sha=sha))
 
         yield task
+
+@transforms.add
+def fixup_mac_worker(config, tasks):
```
ahal, line 46: nit: technically `mac` isn't the commonality here, it's `generic-worker`

But couldn't we just avoid adding `docker-image` for generic-worker tasks in the first place (like you do in the `setup_worker` transforms below)?

Actually looks like the `toolchain` kind is already using those transforms, so I'm not sure this is even necessary?

taskcluster/app_services_taskgraph/transforms/worker.py
```diff
@@ -0,0 +1,25 @@
...
+        try:
+            worker_type = task['worker-type']
+        except KeyError:
+            raise ValueError(f"worker-type not set for {task_name}")
+        if worker_type == 'linux':
+            task['worker-type'] = 'b-linux'
+            worker = task.setdefault('worker', {})
+            worker['docker-image'] = {'in-tree': 'linux'}
+        elif worker_type == 'osx':
+            task['worker-type'] = 'b-osx'
```
ahal, line 22: I think I would avoid translating the `worker-type` here as it just adds another layer of indirection. If you want to get rid of the `b-` prefix, you can rename the aliases here:
https://github.com/mozilla/application-services/blob/main/taskcluster/ci/config.yml#L17

You could continue setting up the `docker-image` here if you like however.

ahal (approved):
Looks good!

I have a couple suggestions, though no need for re-review later

taskcluster/app_services_taskgraph/transforms/toolchain.py
bendk: I'm pretty sure you're right it wasn't necessary anymore after I added the worker transform.  I just deleted that code and am doing one last `[full ci]` check to make sure.

PR #5570, opened 2023-05-12
Block beetmover tasks on the build-summary task being completed

rvandermeulen:
Per Slack discussion, let's hold off on running beetmover tasks until the same phase as release-publish to avoid bump PRs running too early in downstream repos.


bhearsum (approved):
Ryan and I worked on this together and I'm pretty sure this is what we want -- it basically just reverses the relationship between build-summary & beetmover. We had to add the extra attribute to `build-summary` to make `multi_dep.py` happy (as we previously did with my `post-signing-dummy` kind.)

Since I kindof helped write this someone else might want to look at it as well.

rvandermeulen:
Maybe @ahal can do the honors

ahal (approved):

PR #5588, opened 2023-05-18
Move sync manager metrics from firefox-android into application-services [firefox-android: skhamis:move-sync-telem-appservices]

skhamis:
Paired with: https://github.com/mozilla-mobile/firefox-android/pull/2117
Mainly wrangling with build.gradle to run these tests but otherwise the porting went somewhat okay. Still need to do the android/fenix side of things.

If we decide to go down this path the next steps will roughly look like:
1. This lands
    - iOS and Android now use the same metrics/pings.yaml
2. We'll need to update the name of all metrics to `_v2` -> https://bugzilla.mozilla.org/show_bug.cgi?id=1827949#c15
3. Look into adding failure reasons for individual records, similar to desktop
4. Look into seeing if mobile platforms can move forward via syncing even if there are corrupted/non-syncable records incoming/outgoing
5. Look into consolidating metrics to be called from the rust side

  - Note:

### DIFF
--- a/components/sync_manager/android/build.gradle
+++ b/components/sync_manager/android/build.gradle
@@ -1,10 +1,27 @@
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.26"
+}
 
 apply from: "$rootDir/build-scripts/component-common.gradle"
 apply from: "$rootDir/publish.gradle"
 
+// Needs to happen before `dependencies` in order for the variables
+// exposed by the plugin to be available for this project.
+ext.gleanYamlFiles = ["${project.projectDir}/../metrics.yaml", "${project.projectDir}/../pings.yaml"]
+ext.gleanNamespace = "mozilla.telemetry.glean"
+apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
+
 dependencies {
     // Part of the public API.
     api project(':sync15')
+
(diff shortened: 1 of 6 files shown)


bendk (approved):
This seems fine to me.  I only skimmed `SyncTelemetry.kt`, but I assume it's mostly copied from other places.

My one question is where the final location of this code should be.  I was thinking the end goal was to put them in a separate, non-UniFFIed, component.  That way our UniFFIed components could eventually shed all of their wrapper code and the build code gets simplified.

I'm not sure exactly how this would work, but I guess it would be kind of similar to the logging code: 
  - Define a `metrics` crate that has functions to handle each ping
  - The `metrics` crate doesn't depend on Glean directly, and doesn't need any wrapper layer, it defines a UniFFI CallbackInterface and a way for foreign code to register a global instance.
  - Define non-UniFFIed Kotlin/Swift modules that implement the interface, register themselves, and forward the pings to Glean.

This is definitely not trivial and maybe not even a good idea.  I'm fine landing this as-is, but I'd love to discuss this and plan it out some.

lougeniaC64 (approved):
Yes, this all looks very familiar. 🫠 LGTM!

skhamis:
This is superseded by https://github.com/mozilla/application-services/pull/5624

PR #5590, opened 2023-05-19
Add opening and enrolling with deeplink

jhugman:
Fixes [EXP-3510](https://mozilla-hub.atlassian.net/browse/EXP-3510).

  - Note:

### DIFF
--- a/components/support/nimbus-cli/src/cmd.rs
+++ b/components/support/nimbus-cli/src/cmd.rs
@@ -9,7 +9,7 @@ use crate::{
     },
     AppCommand, ExperimentListSource, ExperimentSource, LaunchableApp, NimbusApp,
 };
-use anyhow::Result;
+use anyhow::{bail, Result};
 use console::Term;
 use serde_json::{json, Value};
 use std::{path::PathBuf, process::Command};
@@ -31,6 +31,7 @@ pub(crate) fn process_cmd(cmd: &AppCommand) -> Result<bool> {
             preserve_targeting,
             preserve_bucketing,
             preserve_nimbus_db,
+            deeplink,
             ..
         } => app.enroll(
             params,
@@ -40,6 +41,7 @@ pub(crate) fn process_cmd(cmd: &AppCommand) -> Result<bool> {
             preserve_targeting,
             preserve_bucketing,
             preserve_nimbus_db,
+            deeplink,
         )?,
         AppCommand::FetchList { params, list, file } => params.fetch_list(list, file)?,
         AppCommand::FetchRecipes {
@@ -50,6 +52,7 @@ pub(crate) fn process_cmd(cmd: &AppCommand) -> Result<bool> {
         AppCommand::Kill { app } => app.kill_app()?,
         AppCommand::List { params, list } => params.list(list)?,
         AppCommand::LogState { app } => app.log_state()?,
+        AppCommand::Open { app, deeplink, .. } => app.open(deeplink.as_ref())?,
         AppCommand::Reset { app } => app.reset_app()?,
         AppCommand::TailLogs { app } => app.tail_logs()?,
         AppCommand::Unenroll { app } => app.unenroll_all()?,
@@ -106,7 +109,9 @@ impl LaunchableApp {
                 .spawn()?
                 .wait()?
                 .success(),
-            Self::Ios { app_id, device_id } => {
+            Self::Ios {
+                app_id, device_id, ..
+            } => {
                 let _ = self
                     .exe()?
                     .args(["terminate", device_id, app_id])
@@ -118,7 +123,7 @@ impl LaunchableApp {
 
     fn unenroll_all(&self) -> Result<bool> {
         let payload = json! {{ "data": [] }};
-        self.start_app(false, Some(&payload), true)
+        self.start_app(false, Some(&payload), true, None)
     }
 
     fn reset_app(&self) -> Result<bool> {
@@ -130,7 +135,9 @@ impl LaunchableApp {
                 .spawn()?
                 .wait()?
                 .success(),
-            Self::Ios { app_id, device_id } => {
+            Self::Ios {
+                app_id, device_id, ..
+            } => {
                 self.exe()?
                     .args(["privacy", device_id, "reset", "all", app_id])
                     .status()?;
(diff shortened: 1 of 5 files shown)


components/support/nimbus-cli/src/cmd.rs
```diff
@@ -118,7 +123,7 @@ impl LaunchableApp {
 
     fn unenroll_all(&self) -> Result<bool> {
         let payload = json! {{ "data": [] }};
-        self.start_app(false, Some(&payload), true)
+        self.start_app(false, Some(&payload), true, None)
```
eliserichards, line 126: Do we need this `None` here since it's optional?

components/support/nimbus-cli/src/cmd.rs
```diff
@@ -318,15 +332,42 @@ impl LaunchableApp {
...
+        reset_db: bool,
+        payload: Option<&Value>,
+        log_state: bool,
+        deeplink: Option<&String>,
+    ) -> Result<bool> {
+        let deeplink = self.create_deeplink(deeplink)?;
         Ok(match self {
             Self::Android { .. } => self
-                .android_start(reset_db, payload, log_state)?
+                .android_start(reset_db, payload, log_state, deeplink.as_ref())?
```
eliserichards, line 365: What happens with `deeplink.as_ref()` if `deeplink` doesn't exist?

components/support/nimbus-cli/src/main.rs
```diff
@@ -105,24 +107,48 @@ impl TryFrom<&Cli> for LaunchableApp {
...
+                "developer" => "fenix-dev",
+                "nightly" => "fenix-nightly",
+                "beta" => "fenix-beta",
+                "release" => "fenix",
+                _ => unreachable!(),
+            }),
+            "firefox_ios" => Some(match channel {
+                "developer" => "fennec",
+                "beta" => "firefox-beta",
+                "release" => "firefox-internal",
```
eliserichards, line 122: why is it called `"firefox-internal"`?? is that really what they call their release channel? 🤔

eliserichards (commented):
This is looking lovely! I have a few questions:

components/support/nimbus-cli/src/cmd.rs
jeddai: `deeplink` as a reference will always exist, but it may not always have a value stored in it. I've kinda mentally mapped `Option` to the concept of an enum where the only values are `Some(Inner)` and `None`. Not sure if you'd find that line of thought helpful though.

If you see places where we use `deeplink.unwrap()`, `deeplink?` or `let Some(x) = deeplink`, those are all ways we unwrap an `Option` into its inner value/type.

components/support/nimbus-cli/src/cmd.rs
jeddai: Yep, we do still need to pass _something_, even if it's a `None` `Option`.

Rust doesn't do default parameter values or method overloading and as far as I've been able to determine from research there aren't plans to support either.

components/support/nimbus-cli/src/cmd.rs
```diff
@@ -318,15 +332,42 @@ impl LaunchableApp {
...
+    fn create_deeplink(&self, deeplink: Option<&String>) -> Result<Option<String>> {
+        if deeplink.is_none() {
+            return Ok(None);
+        }
+        let deeplink = deeplink.unwrap();
+        Ok(if deeplink.contains("://") {
+            Some(deeplink.clone())
+        } else if let Some(scheme) = match self {
+            Self::Android { scheme, .. } | Self::Ios { scheme, .. } => scheme,
+        } {
```
jeddai, line 348: This is a really cool way to unwrap multiple enum options' values into a single value, I haven't considered this before!

components/support/nimbus-cli/src/cmd.rs
```diff
@@ -338,49 +379,50 @@ impl LaunchableApp {
...
+                ]);
             } else {
-                None
-            };
-            let experiments_arg = if let Some(s) = json {
+                args.push(format!("-n {}/{}", package_name, activity_name));
+                args.extend(vec![
+                    "-a android.intent.action.MAIN".to_string(),
+                    "-c android.intent.category.LAUNCHER".to_string(),
+                ]);
```
jeddai, line 404: Is there a particular reason for not including the contents of the `push` in the `extend`'s vector?

jeddai (commented):
Code looks good! Just a couple comments and one question.

Is there a patch for fenix/fxios I could use to test the deeplinking before approving? I'm not super concerned if not, just figured it was worth asking.

components/support/nimbus-cli/src/main.rs
jhugman: I started typing an explanation, but then realised it should be a comment in the code. I've added a comment.

components/support/nimbus-cli/src/cmd.rs
jhugman: No, you're right! I also changed to using a slice rather than a `vec![]`.

jhugman:
> Code looks good! Just a couple comments and one question.
> 
> Is there a patch for fenix/fxios I could use to test the deeplinking before approving? I'm not super concerned if not, just figured it was worth asking.

Yes, you should be able to test with:

```sh
nimbus-cli --app fenix --channel developer open --deeplink make_default_browser
```

For the iOS simulator, we can't send command line args, so we may want to think again on how to get things into the SDK; but that would be after some demand for it.

For Firefox for iOS, I've added a "no-clobber" option, which doesn't terminate the app before sending the URL to be opened, so running the following commands should open the expected page.

```sh
nimbus-cli --app firefox_ios --channel developer open
nimbus-cli --app firefox_ios --channel developer open --deeplink deep-link?url=default-browser/tutorial --no-clobber
```

We could add a different option which forces the app to be open before sending the URL, but I wasn't sure which was better. WDYT?

jeddai:
> For Firefox for iOS, I've added a "no-clobber" option, which doesn't terminate the app before sending the URL to be opened, so running the following commands should open the expected page.
>
> ```
> nimbus-cli --app firefox_ios --channel developer open
> nimbus-cli --app firefox_ios --channel developer open --deeplink deep-link?url=default-browser/tutorial --no-clobber
> ```
>
> We could add a different option which forces the app to be open before sending the URL, but I wasn't sure which was better. WDYT?

Since we know this is a limitation of iOS, and there's not really a way for us to get around it, I think I'd suggest that when a deeplink is passed to the `open` command for an iOS app we should just innately have a check/requirement that the app is already open on the device (failing if it's not). Is that doable or do you think that's a bad approach?

PR #5591, opened 2023-05-20
remote_settings: Add `Client::get_records_raw_with_options`.

linabutler:
For the Suggest collection, we'd like to specify some extra options:

* Fetch all records newer than a given timestamp. This is currently possible using `get_records_since`.
* Sort records oldest-to-newest. Remote Settings orders records newest-to-oldest by default, but we want to invert the order to support incremental downloading.
* Specify an optional limit for the number of records to fetch.

This commit:

* Adds a `GetItemsOptions` builder that sets filtering, sorting, and pagination options.
* Implements `Client::get_records_{raw, since}` in terms of the new `get_records_raw_with_options`.

Since Suggest will use `remote_settings::Client` directly from Rust, `GetItemsOptions` and `get_records_raw_with_options` are not UniFFIed.

  - Note:


components/remote_settings/src/client.rs
```diff
@@ -61,22 +64,16 @@ impl Client {
...
     /// collection defined by the [ClientConfig] used to generate this [Client].
     pub fn get_records_since(&self, timestamp: u64) -> Result<RemoteSettingsResponse> {
-        let path = format!(
-            "v1/buckets/{}/collections/{}/records?_since={}",
-            &self.bucket_name, &self.collection_name, timestamp
-        );
-        let resp = self.make_request(path)?;
+        let resp = self.get_records_raw_with_options(
+            GetItemsOptions::new().gt("last_modified", timestamp.to_string()),
+        )?;
```
linabutler, line 76: According to https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html#polling-for-changes, `_since` is an alias for `gt_last_modified`. 😊

mhammond (approved):
Nice!

MatthewTighe (approved):
Looks great from my perspective, thank you for continuing to iterate on this!

components/remote_settings/src/client.rs
```diff
@@ -271,10 +281,217 @@ struct AttachmentsCapability {
...
+                }
+            })
+            .chain({
+                // For selecting fields (https://docs.kinto-storage.org/en/latest/api/1.x/selecting_fields.html),
+                // the query pair syntax is `_fields=field1,field2`.
+                if self.fields.is_empty() {
+                    None
+                } else {
+                    Some(("_fields".into(), self.fields.join(",").into()))
+                }
```
pjenvey, line 426: The newish `bool` method can be nice in places like this (and above)
```suggestion
                !(self.fields.is_empty())
                    .then_some(("_fields".into(), self.fields.join(",").into()))
```

linabutler: Thanks @pjenvey! 🏆 I actually tried writing it with `.then(...)` first, like:

```rs
.chain(
    (!self.fields.is_empty()).then(|| ("_fields".into(), self.fields.join(",").into())),
)
```

...But that felt a little cramped, and I thought the extra parentheses around `(!self.fields.is_empty())` made it easy to miss the `!`, so I wrote it out explicitly. Although, with the comment and block there, maybe it's not too bad?

```rs
.chain({
    // For selecting fields (https://docs.kinto-storage.org/en/latest/api/1.x/selecting_fields.html),
    // the query pair syntax is `_fields=field1,field2`.
    (!self.fields.is_empty()).then(|| ("_fields".into(), self.fields.join(",").into()))
})
```

WDYT?

components/remote_settings/src/client.rs
```diff
@@ -88,6 +85,20 @@ impl Client {
...
+        let path = format!(
+            "v1/buckets/{}/collections/{}/records",
+            &self.bucket_name, &self.collection_name
+        );
+        let mut url = self.base_url.join(&path)?;
+        for (name, value) in options.iter_query_pairs() {
+            url.query_pairs_mut().append_pair(&name, &value);
+        }
+        self.make_request(url)
+    }
```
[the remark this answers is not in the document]
linabutler: Definitely, thanks for noting that; I'll get a follow-up bug on file! We could very likely expose `GetItemsOptions` directly in UDL, and put its mutable state behind a mutex, without sacrificing ergonomics or performance: we aren't expecting this to be hot code, and `parking_lot`'s mutexes have a fast path for the uncontended case also. That would be a great first bug for learning more about Rust and UniFFI!

components/remote_settings/src/client.rs
pjenvey: It's just a suggestion so feel free to ignore. I'll just note that I've used this a few times chaining onto the same `!is_empty()` method call and it already feels like a pretty common, easily identifiable pattern to me: if there's something in this collection, use it.

linabutler: Perfect, thanks—it helps to know that you've used this a few times, and that it feels like a common pattern! I folded in your suggestion 😁

pjenvey (approved):

components/remote_settings/src/client.rs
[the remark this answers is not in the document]
linabutler: (Filed #5608).

NguyenCuong17012021 (commented):
Seting

PR #5595, opened 2023-05-22
Split up the `fxa-client/lib.rs` code and switch to the new error system.

bendk:
This is prep for the upcoming FxA client work.  I think the smaller modules are easier to understand and will make it easier to communicate changes.  Also, it's useful to know which structs go together with which methods.

  - Note:


components/fxa-client/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use thiserror::Error;
+
+/// Generic error type thrown by many [`FirefoxAccount`] operations.
+///
+/// Precise details of the error are hidden from consumers, mostly due to limitations of
+/// how we expose this API to other languages. The type of the error indicates how the
+/// calling code should respond.
```
bendk, line 11: I don't think we have the same UniFFI limitations anymore, should we be adding more data to these?  Or is it actually preferable to hide them?

components/fxa-client/src/devices.rs
```diff
@@ -0,0 +1,474 @@
...
+//! Applications that connect to a user's account may register additional information
+//! about themselves via a "device record", which allows them to:
+//!
+//!    - customize how they appear in the user's account management page
+//!    - receive push notifications about events that happen on the account
+//!    - participate in the FxA "device commands" ecosystem
+//!
+//! For more details on FxA device registration and management, consult the
+//! [Firefox Accounts Device Registration docs](
+//! https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/device_registration.md).
```
bendk, line 16: I wonder if this could be split up more, maybe `devices.rs` and `push.rs`?  WDYT?

components/fxa-client/src/lib.rs
```diff
@@ -53,78 +61,18 @@ pub mod internal;
...
-    Other,
-}
-
 /// Object representing the signed-in state of an application.
 ///
 /// The `FirefoxAccount` object is the main interface provided by this crate.
 /// It represents the signed-in state of an application that may be connected to
 /// user's Firefox Account, and provides methods for inspecting the state of the
 /// account and accessing other services on behalf of the user.
-///
```
bendk, line 106: I removed a lot of these trailing newline comments since I think they're non-standard.  Is there a reason to keep them?

components/fxa-client/src/devices.rs
tarikeshaq: That sounds reasonable, maybe we should go further and consolidate the modules inside the `internal` module with the modules at this layer?

components/fxa-client/src/error.rs
tarikeshaq: I would switch us to use the error handling model we introduced with logins and places, I like how simple the errors at this layer are (they make me think that `FxaError` sounds like a nice `ExternalError`!)

components/fxa-client/src/error.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// of the flow by an attacker. The signin attempt cannot be completed.
+    ///
+    /// **Note:** This error is currently only thrown in the Swift language bindings.
+    #[error("the requested authentication flow was not active")]
+    WrongAuthFlow,
+    /// Thrown if there is a panic in the underlying Rust code.
+    ///
+    /// **Note:** This error is currently only thrown in the Kotlin language bindings.
+    #[error("panic in native code")]
+    Panic,
```
tarikeshaq, line 44: For a followup: I would be surprised if this is actually used correctly, worth investigating

components/fxa-client/src/lib.rs
```diff
@@ -36,13 +36,21 @@
...
+mod error;
+mod migration;
+mod persistence;
+mod profile;
+mod telemetry;
+mod tokens;
+pub use auth::{AuthorizationInfo, MetricsParams};
+pub use devices::{AccountEvent, AttachedClient, Device, DeviceCapability, DevicePushSubscription, IncomingDeviceCommand, SendTabPayload, TabHistoryEntry};
+pub use error::FxaError;
+pub use migration::{FxAMigrationResult, MigrationState};
```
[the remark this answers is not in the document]
tarikeshaq: FWIW iOS should probably no longer try to migrate users but for the naming, I agree we should be conscious if it causes breakage in iOS but also take liberty in changing it if it doesn't!

components/fxa-client/src/lib.rs
tarikeshaq: Nope, good cleanup!

tarikeshaq (approved):
Overall this makes sense to me - and if this helps grow the API lets ship it!

Further, I would like us to also consider what the internal modules look like as well. There might be an opportunity here to consolidate the high level API with those (i.e I believe you mentioned offline that we could have the public and internal API of each "feature" be one module and a submodule)

bendk:
Pushed some more commits, I think this one could be ready to merge if you think it's good.

- I split up `device` and `push` into two modules.
- I didn't restructure the crate so that both the public API and internal code was in the same module (for example moving `internal::device` to `device::internal`).  Maybe this would be good as a follow-up, but it's a decent chunk of work.  There are also some things to like about the current structure for example, everything in `internal` can be private rather than `pub(crate).`.
- I didn't add the `Fxa` prefix to `MigrationState`.  I think we're going to end up dropping that struct anyways.

components/fxa-client/src/account.rs
```diff
@@ -0,0 +1,82 @@
...
+//! web.
+//!
+//! The methods in this section provide URLs at which the user can perform various
+//! account-management activities.
+
+use crate::{ApiResult, Error, FirefoxAccount};
+use error_support::handle_error;
+
+impl FirefoxAccount {
+    /// Get the URL at which to access the user's sync data.
```
mhammond, line 19: heh - this comment kinda lies as it skips a level of indirection - It "Get the URL at which the URL and access token to access the user's sync data can be obtained" :)

(I don't care if you don't update it as I'm sure you just copied it)

components/fxa-client/src/error.rs
mhammond: I agree, only errors which would ever be actually caught should be distinguished, the rest lumped together and reported via error reporting. I'm not sure this PR captures that correctly though, but not sure if it is the intention - specifically, I'd never expect consumers to catch `Panic` specifically, but could imagine it *would* want to handle `BackoffError` (or maybe more generally a "server error"?)

Looking lower, maybe this PR *is* attempting to get those errors correct - if so I think it needs changes.

components/fxa-client/src/error.rs
```diff
@@ -2,10 +2,56 @@
...
+    /// expected by the client.
+    /// This may indicate a stale OAuth flow, or potentially an attempted hijacking
+    /// of the flow by an attacker. The signin attempt cannot be completed.
+    ///
+    /// **Note:** This error is currently only thrown in the Swift language bindings.
+    #[error("the requested authentication flow was not active")]
+    WrongAuthFlow,
+    /// Thrown if there is a panic in the underlying Rust code.
+    ///
+    /// **Note:** This error is currently only thrown in the Kotlin language bindings.
```
mhammond, line 43: To follow up from Tarik's comment, something seems wrong if Kotlin/Swift are *throwing* these errors?

components/fxa-client/src/error.rs
bendk: Good point.  I think we can get rid of `NoExistingAuthFlow`, `WrongAuthFlow`, and `Panic`, then maybe rename the remaining to `AuthError`, `NetworkError`, and `UnexpectedError`.

But I'm thinking we can get this one merged now and work on the breaking changes in the next pass.

components/fxa-client/src/internal/migrator.rs
```diff
@@ -2,9 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-pub use crate::{FxAMigrationResult, MigrationState};
-
-use super::{error::*, scoped_keys::ScopedKey, scopes, FirefoxAccount};
+use super::{scopes, FirefoxAccount};
```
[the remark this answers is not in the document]
bendk: I'm pretty sure that it's not used in Android and we can get it removed from the iOS code.  However, we are going to need a way to migrate Desktop.  So, yes we can kill it, but then it's brother will rise from the grave :)

mhammond:
I'm fine with landing this and fixing the errors later if this isn't a breaking change - I assumed it was and wanted to avoid too many of them, but go for it!

bendk:
> I'm fine with landing this and fixing the errors later if this isn't a breaking change - I assumed it was and wanted to avoid too many of them, but go for it!

I believe this shouldn't have breaking changes, @tarikeshaq can you see any?

tarikeshaq (approved):
Nope, I don't see any breaking changes. No UDL changes and only real practical change here is the error handling, but the errors the consumer sees are still the same - :shipit:

bendk:
Okay, let's try this out :crossed_fingers:

PR #5609, opened 2023-05-25
[Discuss] FxA new API

bendk:
This is what I'm thinking as a new FxA API for the upcoming work.  It's based on #5595, I think we can merge that one and that should make these changes more clear.  For now, just know that the real changes begin with the `Refactoring FirefoxAccount constructor` commit.

All of this is in a very early stage.  I'm sure many things will have to change, I'm mostly wondering if the general shape seems good to others.

The "big idea" here is to start using async more, especially async callback interfaces supplied by the foreign consumer.  This allows the API to be expressed more naturally and to cover more functionality.  The main goal is for this to replace almost all of the code in `services-firefox-accounts` in android-components and the Swift wrapper code in app-services.  The secondary goal is to replace the Desktop FxA code, although that's probably more of a long-term project that will require several iterations.

The consumer will now supply callback interfaces to provide capabilities to the FxA client: an HTTP client, an OAuth handler,  storage layer, etc.  It also provides an event handler which the client can use to communicate when things change.  This leads to some nice improvements in the API, for example instead of having a `begin_oauth_flow()` and `complete_oauth_flow()` method, there's now a single `begin_oauth_flow()` method that will make async HTTP requests and start an oauth flow, and if everything is successful send a `StateChanged` event with the logged in state.

This allows the client to track more about the general state of the application.  I'm hoping that the `get_state()` method can return basically all the information the consumer needs to know in order to update it's menus, buttons, etc.


bendk:
1) The other PR was merged, so this one should be much cleaner to look at now.
2) Mark pointed out that my initial proposal was probably trying to do too much at once.  It's reworking most of the API and also switching it to async.  This leads to the possibility of the async switch failing for some reason and then we don't get the nice API changes.  I'm thinking we can break this up into two parts:
  - The code in this PR which focuses on updating the API.
  - A subsequent effort to switch the API to async.

This means that there will be a transition period when we've done the first but not the second.  During that time:
  - We'll need to keep the wrapper code around to transform this API into some sort of async API, which is what the consumers expect (eg: [firefox-android](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt#L153-L157) and [our swift wrapper](https://github.com/mozilla/application-services/blob/main/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L215-L229))
  - We'll be "wasting" some threads because we don't support async suspend.  However, this won't be a big difference from the current situation, where we need to use a thread for any HTTP operation over viaduct.  I think the one potential difference is that we might need to use a thread when going through the oauth process.
  - I think we still want some limited async support to handle [loading the saved state](https://github.com/mozilla/application-services/blob/4e4af2e8d13601a72927951d8228c927bd8d9dbf/components/fxa-client/src/lib.rs#L95-L97), but if this isn't possible then it should be possible to add a hack involving the wrapper code to do it.

The other way I could see splitting this up would be to:
  - Implement the async API we want in the wrapper code first
  - Then work on getting UniFFI to generate the async API

I think this one requires more overall work through, since that first step is going to require writing a lot of code that we will then throw away.

tarikeshaq (commented):
Thank you for this @bendk and for sharing it early as well!

I had a few feedback comments throughout. Generally, I see the earlier `async` version of this as the "perfect" world to be in, and this is one step away from that.

This may be a bit much to ask, but I wonder if we could have a document that describes the changes we made here and the motivations behind them. What do we think about maybe an ADR? I wonder if we can communicate the new Swift/Kotlin APIs and threading requirements - do an internal review first, then once it's ready for broader review expose it to the consumer teams for sign-off.

That way we'd have a clear contract of what the accepted API looks like starting from the consumer, and that could help inform our API at this layer (i.e the use of async callbacks, the threading requirements, the new functions and interfaces to implement, etc)

A bit unrelated to the above, I think this PR successfully isolates some pain points in the existing API and I really like that! eg: demonstrates that:
- The existing auth flow is not ideal to work with
- The way storage is done makes it hard to reason about it
- Event handling can be done better with callbacks!

The above pain points can be improved without changing the public API too much, and maybe there are small steps that we can take so that we get to the `perfect` world sooner rather than later

components/fxa-client/src/account.rs
```diff
@@ -19,8 +19,6 @@ impl FirefoxAccount {
     /// Get the token server URL
     ///
     /// The token server URL can be used to get the URL and access token for the user's sync data.
-    ///
-    /// **💾 This method alters the persisted account state.**
```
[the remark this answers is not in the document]
bendk: They still will, but the plan is to have them call the `FxaStorage.save_state()` method which will handling saving the state.

That said, it's not like the docstring is wrong so we could keep it around.  I removed them because I thought the only reason for these messages was to inform the consumer that they needed to call `to_json()` and save the results after the method call.

components/fxa-client/src/lib.rs
```diff
@@ -85,36 +89,67 @@ pub struct FirefoxAccount {
...
+
+/// Current state of the FxA client
+pub enum FxaState {
+    /// User is logged out
+    Disconnected,
+    /// User is logged in
+    Connected,
+    /// User has been logged out by some external event and needs to re-authenticate, for example a
+    /// password change from another device
+    ReauthenticationNeeded,
```
bendk, line 144: Note: The android-components code makes a formal distinction between "account states" and "progress states".  Account states are the ones above this comment where there's no in-progress operation that could change the state.  Progress states are like the ones below this comment.  I don't love their terminology, but maybe we should split up this enum somehow to reflect that.  The fact that both `Authenticating` and `Loading` have the same warning makes me think this might be a good idea.

components/fxa-client/src/auth.rs
```diff
@@ -23,21 +23,20 @@
...
 //! documentation hub](https://mozilla.github.io/ecosystem-platform/docs/features/firefox-accounts/pairing).
 
-use crate::{ApiResult, Error, FirefoxAccount};
+use crate::{ApiResult, Error, FirefoxAccount, FxaError};
 use error_support::handle_error;
 use std::collections::HashMap;
 
 impl FirefoxAccount {
-    /// Initiate a web-based OAuth sign-in flow.
+    /// Initiate an web-based OAuth sign-in flow.
```
mhammond, line 31: I think the change of `a` to `an` here is incorrect.

components/fxa-client/src/auth.rs
```diff
@@ -23,21 +23,20 @@
...
     ///
-    /// This method initializes some internal state and then returns a URL at which the
-    /// user may perform a web-based authorization flow to connect the application to
-    /// their account. The application should direct the user to the provided URL.
+    /// - This method calls `perform_flow` on the `OAuthHandler` instance passed to the constructor.
+    /// - While `perform_flow()` is running, the `FirefoxAccount` instance will be in the
+    ///   [FxaState::Authenticated] state (see the FxaState documentation for details)
```
mhammond, line 35: I think this should be `Authenticating`?

components/fxa-client/src/auth.rs
```diff
@@ -23,21 +23,20 @@
...
+    /// - When `perform_flow()` returns the state will be changed to `Authenticated` on success and
```
mhammond, line 36: I think this is `Connected` in the source?

components/fxa-client/src/device.rs
```diff
@@ -250,3 +203,13 @@ pub struct AttachedClient {
     pub last_access_time: Option<i64>,
     pub scope: Option<Vec<String>>,
 }
+
+/// Device record information for the client.  This is passed in at startup.
```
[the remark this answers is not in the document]
mhammond: Unless I'm missing something, isn't the dilemma here going to be that the mobile apps fetching the device record from local storage might be non-trivial - you might have managed to avoid IO in the FxA constructor, but that's not much help if it's mandatory that consumers of FxA must do IO before using it.

components/fxa-client/src/events.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// The account's profile was updated
+    ProfileUpdated,
+    /// The list of devices changed
+    DevicesChanged {
+        /// Device record for this client
+        client_devices: Device,
+        /// Device record for all other known clients
+        other_devices: Vec<Device>,
+    },
+    // I don't think we need these:
```
mhammond, line 35: These are delivered by push, so I think we probably do want them - eg, desktop still displays UI for `DeviceConnected` and IIRC even checks if itself was disconnected in the `DeviceDisconnected` state to leave the browser as though it never signed in.

components/fxa-client/src/events.rs
```diff
@@ -0,0 +1,50 @@
...
+    // DeviceConnected
+    // DeviceDisconnected
+    // DeviceCommandIncoming
+    //
+    // I also don't think we should implement the `SyncStatusObserver` functionality.  If needed,
+    // we should add that to SyncManager
+}
+
+/// Received FxA command
+pub enum FxaCommand {
```
mhammond, line 45: It's difficult to know where this is heading, but I wonder if we should support multiple event receivers? I don't see a great reason to insist on exactly 1 receiver but then forcing the consuming app to handle the multi-receiver case. Eg, desktop wants many receivers (but it's probably true that desktop would make the adapting to multiple receivers trivial via its observer models)

Why isn't a command just part of the `FxaEvent` enum, meaning you just need an `on_event()` handler which also handles `on_command()`? Although I guess one reason might be if you did handle multiple event handlers - you probably *would* want to insist on a single command handler?

components/fxa-client/src/lib.rs
mhammond: hrm, that's interesting. Before I saw this comment I was going to suggest that `FxaState::Loading` seemed a bit smelly and that maybe `get_state()` should just return `Option<FxaState>`, with none representing that state. If we did that, then all we have left is "authenticating" - which if we think about it, probably applies to the other states too - eg, `ReauthenticationNeeded` will also need to hit the network and losing that state while we are reauthenticating seems unfortunate. Ditto for disconnection really - we need to hit the network to disconnect - how do we get back to the initial state if those flows fail?

IOW, I think I'm saying that `Authenticating` almost sounds a little like a boolean `is_state_transitioning` or something - the state is still in the *old* state, we are just trying to change it, and probably should reject further attempts to change it while set.

mhammond (commented):
I didn't get very far here, but I think this is looking reasonable. You deleted the udl, which I assume is temporary, but I wonder if you might be able to put it back - eg, some of the traits I had to assume would be supplied by the app, but really wasn't sure. IOW, I found it tricky to form the "bigger picture" - but as mentioned, ran out of time today and I'll get back to it ASAP.

components/fxa-client/src/storage.rs
```diff
@@ -2,54 +2,43 @@
...
-    #[handle_error(Error)]
-    pub fn from_json(data: &str) -> ApiResult<FirefoxAccount> {
-        Ok(FirefoxAccount {
-            internal: std::sync::Mutex::new(internal::FirefoxAccount::from_json(data)?),
-        })
-    }
+    /// This is called at startup to check if we should try to load the saved state.  If this
+    /// returns true, the `FirefoxAccount` instance will invoke `load_state()` and into a loading
+    /// state while that executes.
+    fn has_saved_state(&self) -> bool;
```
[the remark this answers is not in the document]
mhammond: I see `load_state` now returns an option - should this die? (IMO it should - doesn't seem valuable to ask them to do IO to check a file exists then immediately do IO in it)

components/fxa-client/src/account.rs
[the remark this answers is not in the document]
mhammond: Maybe these comments should be updated to say "will callback into the app-supplied storage handler to persist a new state"?

(TBH I'm not that bothered by the comments existing or not, just trying to get my head around the model)

components/fxa-client/src/device.rs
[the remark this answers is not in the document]
bendk: I thought that they could just construct a local device record without IO, but I could definitely be wrong about that.   If I am, then maybe this should be part of the `SavedState` data that the `FxaStorage` callback interface loads.

components/fxa-client/src/events.rs
bendk: My hope was that `DevicesChanged` could be used a a single event type for all of these, but if that's not going to work out we can definitely add these back in.

components/fxa-client/src/lib.rs
bendk: I like this -- more information for the consumer seems great in this case.  I updated the code to track 2 things: the current state and if we are in the middle of a transition.  I like the shape of this a lot, although I'm not sure about the type names.

components/fxa-client/src/device.rs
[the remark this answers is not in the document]
mhammond: Yeah, most of the info comes from the server so IIUC is persisted locally. I agree it's probably best as part of SavedState.

components/fxa-client/src/events.rs
mhammond: Oh right - yeah, that might work just fine.

bendk:
This one is getting stale and the discussion is moving to the newer/smaller PRs.  I'm going to close this in favor of those.

PR #5612, opened 2023-05-26
Remove LegacySyncEngine

mhammond:
The LegacySyncEngine from #5441 was always going to be temporary, this kills it.


skhamis (approved):
Great cleanup! Reads much easier imo. LGTM!

PR #5615, opened 2023-05-27
Bump XCode to version 14.3.1

rvandermeulen:
Firefox and Focus iOS are on XCode 14.3 now, so being aligned with them seems sensible to me.


rvandermeulen:
@tarikeshaq Feel free to add any other reviewers who may want to be in the loop on this!

tarikeshaq (approved):
Looks good! No need for additional review - this isn’t controversial 🚀

PR #5618, opened 2023-05-30
Add installer for nimbus-cli

jhugman:
Fixes [EXP-3475](https://mozilla-hub.atlassian.net/browse/EXP-3475).

  - Note:

### DIFF
--- a/install-nimbus-cli.sh
+++ b/install-nimbus-cli.sh
@@ -1,0 +1,284 @@
+#!/usr/bin/env bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+: "${BINARY_NAME:=nimbus-cli}"
+: "${TASKCLUSTER_HOST:=https://firefox-ci-tc.services.mozilla.com}"
+: "${USE_SUDO:=false}"
+: "${SHELL:=bash}"
+: "${DEBUG:=false}"
+: "${NIMBUS_INSTALL_DIR:=""}"
+
+# This is largely inspired by the MIT-licensed get-helm-3 install script
+# https://github.com/helm/helm/blob/main/scripts/get-helm-3
+
+HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
+HAS_WGET="$(type "wget" &> /dev/null && echo true || echo false)"
+
+# echoProgress displays a message to the user.
+echoProgress() {
+    echo "✅ ${*}"
+}
+
+# echoError displays an error message to the user.
+echoError() {
+    echo "❎ ${*}"
+}
+
+echoInfo() {
+    # Double space because the `i` seems to eat the first space.
+    echo "ℹ️  ${*}"
+}
+
+# initArch discovers the architecture for this system.
+initArch() {
+  ARCH=$(uname -m)
+  case $ARCH in
+    aarch64) ARCH="aarch64" ;;
+    arm64) ARCH="aarch64" ;;
+    x86_64) ARCH="x86_64" ;;
+    amd64) ARCH="x86_64" ;;
+  esac
+}
+
+# initOS discovers the operating system for this system.
+initOS() {
+  OS=$(uname|tr '[:upper:]' '[:lower:]')
+
+  case "$OS" in
+    # Minimalist GNU for Windows
+    mingw*|cygwin*) OS='windows' ;;
+    darwin*) OS='darwin' ;;
+    linux*) OS='linux' ;;
+  esac
+}
+
+# verifySupported checks that the os/arch combination is supported for
+# binary builds, as well whether or not necessary tools are present.
+verifySupported() {
+  if [ "${HAS_CURL}" != "true" ] && [ "${HAS_WGET}" != "true" ]; then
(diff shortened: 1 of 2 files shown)


install-nimbus-cli.sh
```diff
@@ -0,0 +1,284 @@
...
+}
+
+success() {
+  echoProgress "Success!"
+}
+
+# runs the given command as root (detects if we are root already)
+runAsRoot() {
+  if [ $EUID -ne 0 ] && [ "$USE_SUDO" = "true" ]; then
+    sudo "${@}"
```
tarikeshaq, line 169: How do you feel about instructing the user of the script explicitly run it again with `sudo` if it is necessary to do so? Just to be extra careful of any damage this script can do?

Maybe it's possible to avoid sudo altogether and leave any operations that require root access to the user to do manually?

who reasonable is it to make this script run so that if we need to write to a protected directory/file we can instruct the user to either explicitly set the directory to be another one that is not protected or run the script with sudo if they know what they're doing

I don't feel strongly especially that it's meant to be used by QA and Android/iOS devs so feel free to dismiss!

install-nimbus-cli.sh
```diff
@@ -0,0 +1,284 @@
...
+}
+
+# installFile installs the nimbus-cli binary.
+installFile() {
+  NIMBUS_TMP="$NIMBUS_TMP_ROOT/unzipped"
+  mkdir -p "$NIMBUS_TMP"
+  unzip -qj "$NIMBUS_TMP_FILE" -d "$NIMBUS_TMP"
+  local binary
+  if [[ "$OS" == "windows" ]] ; then
+    binary="${BINARY_NAME}.exe"
```
tarikeshaq, line 182: Does this installer support windows ;) If not we might want to return early with instructions?

install-nimbus-cli.sh
```diff
@@ -0,0 +1,284 @@
...
+fi
+
+# Parsing input arguments (if any)
+export INPUT_ARGUMENTS="${*}"
+set -u
+
+while (( "$#" )); do
+  case "$1" in
+    '--no-sudo')
+       USE_SUDO="false"
```
tarikeshaq, line 251: Hmm does this get overwritten once `initInstallDirectory` runs? i.e even if the user sets `--no-sudo` we might run `sudo` anyways if the directory is protected (I also could be misfollowing the logic as well so feel free to correct me!)

tarikeshaq (commented):
Thank you @jhugman! I had a couple of comments about the `sudo` usage but otherwise, this looks good - (I didn't actually run the script or follow the details of the script, so I'll leave approval to Charlie!)

jeddai:
Still reviewing the code but wanted to go ahead and note I validated it functions as expected when running `curl <url> | sh`! It installed into `/usr/local/bin` and I was able to run it no problem.

install-nimbus-cli.sh
```diff
@@ -0,0 +1,284 @@
...
+
+: "${BINARY_NAME:=nimbus-cli}"
+: "${TASKCLUSTER_HOST:=https://firefox-ci-tc.services.mozilla.com}"
+: "${USE_SUDO:=false}"
+: "${SHELL:=bash}"
+: "${DEBUG:=false}"
+: "${NIMBUS_INSTALL_DIR:=""}"
+
+# This is largely inspired by the MIT-licensed get-helm-3 install script
+# https://github.com/helm/helm/blob/main/scripts/get-helm-3
```
jeddai, line 15: Great choice, also I 10/10 recommend Helm if you do any k8s work 😅 

My comments will be based on a diff between the helm script and this one, I'm trusting that Helm's install script works well (from experience).

install-nimbus-cli.sh
```diff
@@ -0,0 +1,284 @@
...
+
+HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
+HAS_WGET="$(type "wget" &> /dev/null && echo true || echo false)"
+
+# echoProgress displays a message to the user.
+echoProgress() {
+    echo "✅ ${*}"
+}
```
jeddai, line 23: I _really_ like these, great addition

jeddai (commented):
I've got one request re: checking for `unzip`, but otherwise it's looking good so far! Waiting to approve bc @tarikeshaq has some good questions

install-nimbus-cli.sh
jhugman: On reflection, I think I'm going to remove the `--no-sudo` option: for those with a more advanced set up (e.g. having `$HOME/bin` in their `PATH`, sudo isn't asked for.

> it's meant to be used by QA and Android/iOS dev so feel free to dismiss!

I think I'm leaning this way.

install-nimbus-cli.sh
jhugman: I've written it as though it does, but only via WSL: in which case, I think this works. If not, then the installer just won't work— it's a bash script!

jeddai: Oh I forgot to mention in a top-level comment but I did test the install on WSL and it worked!

install-nimbus-cli.sh
jhugman: Yes, you're right.

I'm going to remove the `--no-sudo` option; I don't think it'll get any usage: it's a hangover from the get_helm script.

jeddai (approved):
Super excited for this! 🚢

PR #5631, opened 2023-06-02
Don't write duplicate entries to dependency-licenses.xml

mhammond:
This no longer tries to add the dupe `syn` but there are a few other dupes this removes


tarikeshaq (approved):
Wohoo 🚢 🚢 🚢

PR #5645, opened 2023-06-07
Add fetch task for NSS mac opt-static binary

skhamis:
Part 1 of 2: We need to land this first to create/store the index, then update the scripts, which will be in: https://github.com/mozilla/application-services/pull/5648


skhamis:
>     * Should we wait to delete the patch until the second PR?
That patch is not actually used in any flow at this point, so it's really just a cleanup (it was the old way of updating the phab patch, etc)

>     * Why do we need to split these into 2 PRs?  L1 fetch tasks run on every PR and the `nss-artifact` index already seems to be populated.

ah maybe I'm misunderstanding, I thought those fetch tasks _shouldn't_ run on every PR, only when the link changes. Otherwise this unfortunately will fail once the NSS CI link expires.

skhamis:
Thanks for the review! I'm going to merge this as I think we can the L1 fetch and see if I can reach out to the CI folks to get their input (once this is landed, i'm sure they can tweak it to how we'd like it to be)

PR #5651, opened 2023-06-08
Target JVM 17 bytecode [firefox-android: main][ci full]

rvandermeulen:


rvandermeulen:
This should be landable once https://github.com/mozilla-mobile/firefox-android/pull/1936 is merged upstream.

rvandermeulen:
@tarikeshaq This should be ready for review. I can't do a [ci full] run on the PR right now because of issues with the macOS worker, but I was able to successfully build the megazord locally. I'll do another full [firefox-android: main] run on this after that upstream PR lands as well before attempting to merge this PR.

taskcluster/docker/linux/Dockerfile
```diff
@@ -2,9 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-FROM ubuntu:18.04
-
-MAINTAINER Edouard Oger "eoger@mozilla.com"
```
tarikeshaq, line 7: 😬

docs/building.md
```diff
@@ -95,8 +95,8 @@ The following instructions assume that you are building `application-services` f
 
 1. Install Android SDK, JAVA, NDK and set required env vars
    1. Clone the [firefox-android](https://github.com/mozilla-mobile/firefox-android) repository (**not** inside the Application Service repository).
-   1. Install [Java **11**](https://www.oracle.com/java/technologies/downloads/#java11) for your system
-   1. Set `JAVA_HOME` to point to the JDK 11 installation directory.
+   1. Install [Java **17**](https://www.oracle.com/java/technologies/downloads/#java17) for your system
```
tarikeshaq, line 98: Do app services developers also need to update? we can make a mini announcement in slack if so

taskcluster/docker/linux/Dockerfile
```diff
@@ -159,7 +159,7 @@ ENV CCACHE=sccache \
 RUN \
     curl -sfSL --retry 5 --retry-delay 10 \
          -o /usr/local/bin/tooltool.py \
-         https://raw.githubusercontent.com/mozilla/build-tooltool/36511dae0ead6848017e2d569b1f6f1b36984d40/tooltool.py && \
+         https://raw.githubusercontent.com/mozilla-releng/tooltool/master/client/tooltool.py && \
```
tarikeshaq, line 162: I see `tooltool` has since changed homes - I imagine the `master` branch is stable enough for us not to fix a tag/revision? +1 for this because it allows us to catch and report any oddities with the script quickly - just curious

tarikeshaq (approved):
I don't know much gradle so a bit of a number stamp on the changes there, otherwise LGTM to the upgrade and alignment with AC assuming CI passes when we run the branch build

docs/building.md
rvandermeulen: If they still have JDK11 installed, they'll need to update to 17, yeah. That said, they really *should* be on at least 17 at this point anyway! 😄

taskcluster/docker/linux/Dockerfile
rvandermeulen: Yeah, my reasoning was that it changes infrequently enough to not be an issue. Especially since we're only going to be attempting to pick up a newer version if/when the docker image rebuilds, which should be a pretty infrequent occurrence.

PR #5659, opened 2023-06-12
Reworked FirefoxAccount construction

bendk:
Made `FirefoxAccount` input a single `FxaConfig` instance for all of it's configuration.  I want to add more constructor arguments but don't want to end up with a huge argument count.

Made the example code use the public API, not the internal module. `FxaConfig` helps ease this transition.  The motivation for this is that I want to move FxaClent and some other fields to the public `FirefoxAccount` instance, which means the example code needs to start using it. Deleted the `fxa-client/migration` example, we're not really using the migration code anymore.

Adding some handling for dead code now that we're not exposing the `internal` module as pub.

Created an `FxaServer` enum.  This matches how the android wrapper handles it and I don't think we need to input arbitrary URLs here.

  - Note:


components/fxa-client/src/internal/config.rs
```diff
@@ -204,24 +148,104 @@ impl Config {
...
+    }
+
+    pub fn userinfo_endpoint(&self) -> Result<Url> {
+        Url::parse(&self.remote_config()?.userinfo_endpoint).map_err(Into::into)
+    }
+}
+
+#[allow(dead_code)]
+// Methods that aren't actually used.  Should we delete these?
+impl Config {
```
bendk, line 166: I feel like we should delete, but I don't know about the history of these.  What do others think?

components/fxa-client/src/fxa_client.udl
```diff
@@ -723,6 +709,27 @@ interface FirefoxAccount {
...
+dictionary FxaConfig {
+    // FxaServer to connect with
+    FxaServer server;
+    // Registered OAuth client id of the application.
+    string client_id;
+    // `redirect_uri` - the registered OAuth redirect URI of the application.
+    string redirect_uri;
+    //  URL for the user's Sync Tokenserver. This can be used to support users who self-host their
+    //  sync data. If `None` then it will default to the Mozilla-hosted Sync server.
+    string? token_server_url_override;
```
mhammond, line 721: I think the fact that this exists at all is suspect tbh - I don't see why this module needs to know about the override - but that existed well before this :)

mhammond (commented):
This makes sense to me, but it's a breaking change, right?

components/fxa-client/src/internal/config.rs
mhammond: I'm always up for killing unused code 🔥

components/fxa-client/src/lib.rs
```diff
@@ -89,30 +83,78 @@ impl FirefoxAccount {
...
-    ) -> FirefoxAccount {
+    pub fn new(config: FxaConfig) -> FirefoxAccount {
         FirefoxAccount {
-            internal: std::sync::Mutex::new(internal::FirefoxAccount::new(
-                content_url,
-                client_id,
-                redirect_uri,
-                token_server_url_override.as_deref(),
-            )),
+            internal: std::sync::Mutex::new(internal::FirefoxAccount::new(config)),
```
bendk, line 88: I'm thinking that things that block like the oauth handler, API client, storage handler, etc. should live here, next to the `Mutex<internal::FirefoxAccount>`.  That would facilitate things like releasing the lock while the application performs the oauth flow.  This is why I made the CLI examples construct the public FirefoxAccount rather than the internal one.

Once those changes are made, then there's going to be a bigger distinction between `FirefoxAccount` and `internal::FirefoxAccount`.  What do others think about changing the name to reflect that?  Maybe calling it `FxaInternal` or `FxaInternalState`?

There's an alternative would be to keep `FirefoxAccount` as a thin wrapper around `internal::FirefoxAccount`, and moving a lot of functionality into the `State` struct.  I don't think I like that as much though.

bendk:
> This makes sense to me, but it's a breaking change, right?

Yes it is.  I added the `do-not-merge` label, let's not remove it until we have approved PRs for iOS and Android.

components/fxa-client/src/lib.rs
tarikeshaq: I think it will all be clearer once we start seeing some of those changes, but here are my initial non-blocking thoughts

- By switching `internal::FirefoxAccount` into `internal::FxaInternalState`, are we going to consolidate the states into one place (putting the caches, the `State`, and the HTTP backoff state into one place and wrapping that up in the mutex?). If so I like it, in a way, we are collapsing the two `FirefoxAccount`s and making the locking more granular and +1 for that!
- How will the other members (HTTP client for example) interact with the state that requires locking the mutex? Will that be the responsibility of the high-level `FirefoxAccount`? For example after an http request if we want to update the back off for an endpoint?

components/fxa-client/src/fxa_client.udl
tarikeshaq: Probably for pre-sync manager implementations, we might have an API that retrieves the token server URL for consumers

tarikeshaq (approved):
A couple of non-blocking comments, but this LGTM - +1 to making sure consumer PRs are up if needed (might not need one for iOS since a lot of that logic is handled here in application-services)

components/fxa-client/src/internal/config.rs
```diff
@@ -204,24 +148,104 @@ impl Config {
...
+            FxaServer::China => "https://accounts.firefox.com.cn",
+            FxaServer::LocalDev => "http://127.0.0.1:3030",
+        }
+        .to_string();
+
+        Self {
+            content_url,
+            client_id: fxa_config.client_id,
+            redirect_uri: fxa_config.redirect_uri,
+            token_server_url_override: fxa_config.token_server_url_override,
```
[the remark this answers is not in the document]
bendk: Good catch, I pushed an update for this.

components/fxa-client/src/lib.rs
bendk: My goal is to move some of the state out of `internal::FirefoxAccount`.  I'm thinking the public `FirefoxAccount` will have fields that look something like this:
  - `FxaInternalState`: holds the current `State` field, plus some new internal state, like the current state transition.
  - `ApiClient`: contains the HTTP functionality + the caches and rate limiter/backoff
  - `OAuthHandler`: The application-provided oauth handler
  - `StorageHandler`: application-provider storage handler
  - `EventHandler`, `...`: I think all other application-provided callback interfaces will probably also live here, but haven't thought all of it through.

Each of these will have its own locking, if they need it.  Let's take the `get_devices()` call as an example:
  - Call `internal_state.get_device_request()`.  That locks the state to figure out which URL, refresh token, etc. to use.  Once we know the request, we release the lock.
  - Call `api_client.execute_request(device_request)` to perform the request.
     - Part of performing the request means checking/saving to the cache which will lock a mutex internal to `ApiClient`
     - For auth-related requests it would also check the rate-limiter, which would currently require a mutex, but I think we can avoid that with a refactor
   - Finally return the result of the request, maybe after some conversion.

By having multiple top-level components, each with their own locking, we allow more concurrency.  In particular, we don't want to block on the user-interaction in OAuthhandler.  `FxaInternalState`, or whatever we end up calling it, will be able to access the internal state, but should not block on IO, network, user-interaction, etc.

Each of the components will be independent of each other so there's no chances of deadlocks.  Maybe they will need some lock-free communication channels though, for example `FxaInternalState` may need a way to signal to `ApiClient` that it should clear it's cache/reset it's rate limiter.  Maybe not though, I could see the top-level `FirefoxAccount` handling that.

components/fxa-client/src/fxa_client.udl
mhammond: yeah, that might even be true today, but it still seems wrong for this to know about the override - this crate should probably just return what fxa told us, and the choice about whether to use the override made by the consumer of this

lougeniaC64:
> > This makes sense to me, but it's a breaking change, right?
> 
> Yes it is. I added the `do-not-merge` label, let's not remove it until we have approved PRs for iOS and Android.

These changes make sense to me but I have reservations around merging this PR prior to us vetting the overall approach for this effort. I wouldn't have an issue if this were an internal refactor without breaking changes but is right on the line between refactor and the actual work to complete the FxA refactor.

I'm not opposed to these changes, but I'd feel better about introducing the breaking changes once we're more certain about the scope and strategy for this effort.

mhammond:
> I'm not opposed to these changes, but I'd feel better about introducing the breaking changes once we're more certain about the scope and strategy for this effort.

@tarikeshaq seems to think we might be able to make it a non-breaking change by adjusting our wrappers, but the point is a really good one in general. We seem to have a tension between:

* We don't want this to be a long-running project that makes a massive PR with many breaking changes.
* We don't want to land smaller breaking changes until we have a clear picture of what we are trying to do.
The problem is I agree with *both* of them, but they seem to contradict 😅 .

components/fxa-client/src/fxa_client.udl
```diff
@@ -723,6 +709,29 @@ interface FirefoxAccount {
...
+
+// FxA server to connect to
+[Enum]
+interface FxaServer {
+    Release();
+    Stable();
+    Stage();
+    China();
+    LocalDev();
+    Custom(string content_url);
```
mhammond, line 732: How do you feel about using `Url` here? As a nit, I'd say just `url` is better than `content_url`, particularly for external people who might misunderstand what `content` means in this context.

bendk:
> These changes make sense to me but I have reservations around merging this PR prior to us vetting the overall approach for this effort. I wouldn't have an issue if this were an internal refactor without breaking changes but it's right on the line between refactor and the actual work to complete the FxA refactor.

> @tarikeshaq seems to think we might be able to make it a non-breaking change by adjusting our wrappers, but the point is a really good one in general. We seem to have a tension between:
> 
> * We don't want this to be a long-running project that makes a massive PR with many breaking changes.
> * We don't want to land smaller breaking changes until we have a clear picture of what we are trying to do.
The problem is I agree with *both* of them, but they seem to contradict 😅 .

I agree that there's a tension in general, but I think this PR is on the "okay to introduce now" side.  This actually isn't a breaking change for our consumers, I updated the wrappers to ensure that.  It's also not that much of a change from the current status quo.   The [Swift](https://github.com/mozilla/application-services/blob/main/components/fxa-client/ios/FxAClient/FxAccountConfig.swift) and [Kotlin](https://github.com/mozilla/application-services/blob/main/components/fxa-client/ios/FxAClient/FxAccountConfig.swift) already have a version of the `Config` and `Server` types that get passed to the fxa client constructor.  This is just changing the Rust code to be more inline with that.

mhammond:
I agree - the fact it's *not* a breaking change for consumers makes this a no-brainer. I was talking more in the abstract :)

components/fxa-client/src/fxa_client.udl
bendk: `url` makes more sense to me than `content_url`.  I'll make that change.

I tested out the using `Url` directly and the main issue is that parsing URLs is failing.  I think that means we need to either make the constructor failing  or panic in the face of invalid URLs.  I'm thinking we can hold off on this one for now.

bendk:
Updated the wrapper code to remove the breaking change.  I think I applied all the suggestions except switching to `Url`, which seems better as a follow-up.  The CI is green.  Should we merge this one?

components/fxa-client/src/lib.rs
```diff
@@ -40,6 +40,11 @@ mod account;
 mod auth;
 mod device;
 mod error;
+// We need to make the `internal` mod public for `sync-test` to work
```
mhammond, line 43: I think we fixed that so it's no longer true. I put a PR up for your branch to kill this!

mhammond (approved):
This looks great, thanks! I think you might as well fold my PR into this too assuming it's green. Or as a followup, I don't really mind, but it's logically a part of this.

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/Config.kt
```diff
@@ -9,23 +9,28 @@ package mozilla.appservices.fxaclient
...
     val redirectUri: String,
     val tokenServerUrlOverride: String? = null,
 ) {
-    enum class Server(val contentUrl: String) {
-        RELEASE("https://accounts.firefox.com"),
-        STABLE("https://stable.dev.lcip.org"),
-        STAGE("https://accounts.stage.mozaws.net"),
-        CHINA("https://accounts.firefox.com.cn"),
-        LOCALDEV("http://127.0.0.1:3030"),
+    enum class Server(val rustServer: FxaServer) {
```
mhammond, line 17: One interesting issue with this refactor is that it seems impossible for consuming apps to discover what the actual URL being used it? ie, they might construct with `FxaServer.Stage`, but later when they want to open, say, the "manage account" page, they seem unable to discover what the base URL is. Before this change they are able to obtain it via the enum.

Not sure what the answer is here - ie, whether we should expose the base URL, or expose a way to build a URL - eg, how desktop does it is [here](https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccountsConfig.sys.mjs#79) - the code is responsible for knowing whether the email address should be included in the URL etc.

components/fxa-client/src/lib.rs
bendk: Sounds great to me.  I just merged the branch in.

PR #5664, opened 2023-06-13
Fxa client oauth handling

bendk:
Rework FxA Oauth handling by defining a callback interface that preforms the oauth flow and changing `FirefoxAccount` to have a single method that performs the entire flow.

This is based off of `fxa-client-construction` to make the changes clearer.  Let's merge that branch first, then update the PR to use `main` as the base branch.  I'm pretty sure all of the example clients and the Swift/Kotlin code are broken with this change, let's fix that in `fxa-client-construction` first then fix it here.

  - Note:


components/fxa-client/src/auth.rs
```diff
@@ -168,6 +157,28 @@ pub struct AuthorizationInfo {
...
+    /// sessions.
+    fn perform_flow(&self, url: String) -> CallbackResult<OAuthResult>;
+}
+
+// Result of an Oauth flow
+//
+// Normally, the field values are extracted from the URL query parameters when the browser reaches
+// the redirect_uri.
+pub struct OAuthResult {
+    pub code: String,
```
bendk, line 179: I didn't realize this before, but this change means the consumer only needs to pass us the `code` query param, not the `state` anymore.  That makes sense though, since once we have a single method that handles the oauth process, we no longer need to round-trip data through the FxA server to get the oauth state setup at the begining.

tarikeshaq: The `state` is used to verify that the authorization code returned to us by the server is the one associated with the same request we sent (see https://datatracker.ietf.org/doc/html/rfc6749#section-10.12)

We should still get the `state` back from the consumer and match it against the one we generated. If they don't match we should error out before sending the authorization request to the user

components/fxa-client/src/auth.rs
```diff
@@ -51,41 +42,47 @@ impl FirefoxAccount {
...
-        self.internal
-            .lock()
-            .begin_oauth_flow(&scopes, entrypoint, metrics)
+
+        // Lock self.internal to determine the oauth URL
+        let mut internal = self.internal.lock();
+        let (url, flow) = internal.begin_oauth_flow(&scopes, entrypoint, metrics)?;
+
+        // Release the lock while blocking on the OAuth handler
+        drop(internal);
```
tarikeshaq, line 59: nit: What do we think about introducing another scope (i.e a `{ ` block) that way we avoid the explicit drop and make easier for the reader that the lock only lives within that scope

components/fxa-client/src/internal/oauth.rs
```diff
@@ -277,7 +282,6 @@ impl FirefoxAccount {
         url.query_pairs_mut()
             .append_pair("client_id", &self.state.config.client_id)
             .append_pair("scope", &scopes.join(" "))
-            .append_pair("state", &state)
```
tarikeshaq, line 280: we would want to bring this back as well (re: my comment on `state`)

components/fxa-client/src/auth.rs
```diff
@@ -51,41 +42,47 @@ impl FirefoxAccount {
...
+        let result = self.oauth_handler.perform_flow(url)?;
```
tarikeshaq, line 60: I'm both excited and nervous to see this, and I'm really curious what the implementation on the foreign language would look like.

It would help us if we have some tests here as well, maybe simple unit tests here that mock the handler and more importantly some Android and iOS tests that try as close as possible to mimic what the consumer might try to do:
- A simple case were the consumer returns right away in the same thread
- A more complex case where a consumer blocks the thread, then resumes it later
- A case covering an error
- A case covering cancellation

components/fxa-client/src/auth.rs
```diff
@@ -12,32 +12,23 @@
...
-    /// Initiate a web-based OAuth sign-in flow.
-    ///
-    /// This method initializes some internal state and then returns a URL at which the
-    /// user may perform a web-based authorization flow to connect the application to
-    /// their account. The application should direct the user to the provided URL.
-    ///
-    /// When the resulting OAuth flow redirects back to the configured `redirect_uri`,
-    /// the query parameters should be extracting from the URL and passed to the
-    /// [`complete_oauth_flow`](FirefoxAccount::complete_oauth_flow) method to finalize
-    /// the signin.
```
tarikeshaq, line 40: We're losing some context by deleting those comments, we should document at least the expectation that we are enforcing in our oauth handler. 

I'm not sure if documenting them here as doc comments is best (and I know we have some notes on the callback interface itself) - but we should make sure the context isn't fully lost

components/fxa-client/src/fxa_client.udl
```diff
@@ -101,7 +108,7 @@ interface FirefoxAccount {
   //
   // This method constructs as new [`FirefoxAccount`] instance configured to connect
   // the application to a user's account.
-  constructor(FxaConfig config);
+  constructor(FxaConfig config, OAuthHandler oauth_handler);
```
tarikeshaq, line 111: Do you think it's worth it for us to also comment on the limitation of callback interfaces when it comes to releasing memory?

Particularly whatever callback interface is passed here will be held on to and **must not be destroyed** while the object is alive?

This makes me wonder how the JVM handles that 🤔

tarikeshaq (changes requested):
Thanks @bendk! I had a few comments as I was going through, mostly around how we are handling the OAuth `state` and adding more tests

components/fxa-client/src/fxa_client.udl
bendk: I'm pretty sure UniFFI ensures that the callback interface stays alive as long as Rust has a reference to it.  For example [CallbackInterfaceRuntime.kt](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceRuntime.kt) keeps an entry in a map to keep it alive.   At some point I want to refactor it to use [UniFfiHandleMap](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt#L130-L154) which does something similar.

components/fxa-client/src/auth.rs
bendk: Is this an issue for us though?  AIUI, the attack is against a stateless webserver where you redirect a user to the oauth complete URL, but with your code rather than theirs.  This doesn't seem possible because we control the entire OAuth flow inside our method.  When `OauthHandler::perform_oauth()` returns, I think we can be trust the code.  Also, if an attacker was somehow redirecting things inside the flow, they could probably see the state.

I'm not really sure about this though, so I'm very happy to put the state back in with a note that explains it's just in case.

tarikeshaq: I definitely don't have the right knowledge in security or the OAuth 2.0 RFC itself so take everything I say with a grain of salt 😅 

I think you may be right about our case within the browsers, where the `code` and `state` are passed through a web channel configured **only** for the firefox.com origin. So the risk comes in:
- if the content server is compromised but there are worse things that will happen in that case (attackers gets user's credentials etc).
- Somehow the consuming apps try to complete the auth process by getting a `code` and `state` from a different source.. this implies a very bad bug in Firefox

So for Firefox it'll be precautionary and just in case (but I still think it's good to have regardless)

For hypothetical non-Firefox consumers, the issue becomes more apparent, we can imagine VPN or relay starting an oauth flow, where the OAuth handler blocks a thread, and resumes the thread with a `code` and `state` once a web redirect is done. If a third-party web application redirects to VPN or relay and passes their own `code` then it might not be possible for the relying party to confirm that it was indeed the oauth code it requested

Regardless, I think we should error on the side of caution here (I'm also not sure if the FxA server enforces a `state` query parameter or treats it as optional so we might be forced to keep it even if we didn't want to)

mhammond: This kinda ties into my comment above - if it's possible for multiple oauth requests to be happening concurrently, you can imagine a confused client causing something like this

components/fxa-client/src/error.rs
```diff
@@ -193,7 +197,28 @@ impl GetErrorHandling for Error {
...
+
+/// Errors thrown by by callback interfaces
+#[derive(Debug, thiserror::Error)]
+pub enum CallbackError {
+    #[error("Cancelled")]
+    Cancelled,
+
+    #[error("CallbackError: {reason}")]
+    Other { reason: String },
+}
```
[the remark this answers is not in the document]
mhammond: Seems fine to me to have an error enum specifically for things we might expect callbacks to do - what's your concern?

To my mind, calling `Cancelled` an error is a bit of a smell - it's almost like we want an enum much like `Result<>` but with a 3rd option for cancelled to reinforce it's an expected non-error state - but I don't think we should actually do that :)

components/fxa-client/src/fxa_client.udl
mhammond: I think we want to refactor it to use traits (and deprecate callbacks while we are at it), then it's just an `Arc<>` like everything else :)

mhammond (commented):
This is interesting as it really does start to experiment with long running callbacks.

components/fxa-client/src/auth.rs
bendk: I'm still not sure this is needed, even in that VPN situation.  ISTM that if some attacker got in the middle of the flow, then they would be able to see the `state` and send it back.  The param is there to protect against a complete oauth endpoint from receiving the result of an an oauth flow initiated somewhere else.

But I'm not a security expert, so let's put it back.  I do think we can kill the `flow_store` field though.

tarikeshaq: Just for completeness: My concern is that the delivery mechanism some other consumers might use to deliver the `code` and `state` may not have the same properties Firefox does. I could imagine a world where VPN uses [web redirect to redirect to its native app](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/), and a third-party can take advantage of that to deliver its own oauth code.

I.e the concern is less that an attacker gets in the middle of the flow, it's more that an attacker inserts itself to the end of the flow and the relying party resolves the callback with that OAuth code.

But of course, I'm not a security expert either so +1 to putting it back if only for the reason of keeping the status quo here

bendk: I updated the commit to still pass the `state` query, but removes the `flow_store` field.  I think think this should be very safe and avoids the issues of a confused client (if they mix up the requests, then they'll get 2 errors).

components/fxa-client/src/error.rs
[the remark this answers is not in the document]
bendk: James actually mentioned the same thing.  Maybe a `Result<Option<T>>` is better.

That gets a bit weird, since then `CallbackError` only has one variant.  I guess that's just more motivation for the error work that you're doing.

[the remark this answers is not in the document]
bendk: I think the biggest issue with `Result<Option<T>>` is that it means that consumers have to return `null`  to signal cancelled, which doesn't seem right.  Also, they have several nested functions, than it might be more idiomatic to let a `Cancelled` error bubble up than for each of them to handle `null`.

components/fxa-client/src/auth.rs
bendk: I'm actually very glad you're bringing this up.  I want to open another PR to address locking using this example, but I was worried it was a bit too obscure.  I'll try to open that one up tomorrow, let's discuss more.

components/fxa-client/src/auth.rs
bendk: I think the documentation in `internal/auth.rs` is decent.  It describes both halves of the operation in `begin_oauth_flow()/begin_pairing_flow()` and `complete_oauth_flow()` .  Is there something I should add there?

components/fxa-client/src/error.rs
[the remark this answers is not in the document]
mhammond: I don't like `Result<Option<T>` for the same reason. `enum CallbackResult ...` is what I had in mind, but I suspect the cure will be worse than the disease...

components/fxa-client/src/auth.rs
tarikeshaq: Not sure if it belongs here, but possibly belongs at whatever layer the consumer will be blocked. The message we should send is "This API will block until the handler given in the constructor is completed", I understand that we have that information in the `OAuthHandler` itself - but it might be beneficial to have it on the API itself as a warning

components/fxa-client/src/auth.rs
mhammond: Not clear we actually want "locking" here - indeed, not clear what semantics we want in general (ie, I think I know what we don't want, just not sure what we do :) - looking forward to chatting about it though.

PR #5668, opened 2023-06-14
Move sync-test to the fxa cli helper.

mhammond:
I initially did this on @bendk's FxA branch, but given it's so large I thought I'd put it in main first, then @bendk can rebase and make alot of this go away (although I do still have a version for that branch)

Not that happy with the new `with_internal` for FxA, but that's obviously not needed on Ben's branch, but am I missing something?

### DIFF
--- a/testing/sync-test/src/auth.rs
+++ b/testing/sync-test/src/auth.rs
@@ -1,233 +1,30 @@
 /* Any copyright is dedicated to the Public Domain.
 http://creativecommons.org/publicdomain/zero/1.0/ */
 
-use crate::Opts;
 use anyhow::Result;
 use autofill::db::store::Store as AutofillStore;
-use fxa_client::internal::{auth, config::Config as FxaConfig, FirefoxAccount};
+use cli_support::fxa_creds::CliFxa;
+use fxa_client::internal::{config::Config as FxaConfig};
+use fxa_client::Device;
 use logins::LoginStore;
-use serde_json::json;
 use std::collections::HashMap;
 use std::sync::Arc;
-use sync15::DeviceType;
+use sync15::{client::{SetupStorageClient, Sync15StorageClient}, DeviceType};
 use sync_manager::{
-    manager::SyncManager, DeviceSettings, SyncAuthInfo, SyncEngineSelection, SyncParams, SyncReason,
+    manager::SyncManager, SyncEngineSelection, SyncParams, SyncReason, DeviceSettings
 };
 use tabs::TabsStore;
-use url::Url;
-use viaduct::Request;
 
-pub const CLIENT_ID: &str = "3c49430b43dfba77"; // Hrm...
-pub const SYNC_SCOPE: &str = "https://identity.mozilla.com/apps/oldsync";
+// Note that in revision ba9cc53710243c357e5bf125636bb94a7c6a4a48 or so, this had support for
+// username/password authentication and for "temporary" accounts using restmail etc.
+// It has since been refactored to use only oauth and to use the "cli-support" crate to help
+// manage the account, but you can use that revision if you ever want to revive username/password
+// support.
 
-// TODO: This is wrong for dev?
-pub const REDIRECT_URI: &str = "https://stable.dev.lcip.org/oauth/success/3c49430b43dfba77";
-
-// It's important that this doesn't implement Clone! (It destroys it's temporary fxaccount on drop)
-#[derive(Debug)]
-pub struct TestAccount {
-    pub email: String,
-    pub pass: String,
-    pub cfg: FxaConfig,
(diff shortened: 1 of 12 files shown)


components/external/glean
tarikeshaq, line 1: oops looks like this slid in, might need to re-sync the glean versions since an upgraded landed in main

testing/sync-test/src/auth.rs
```diff
@@ -1,233 +1,46 @@
...
-use url::Url;
-use viaduct::Request;
 
-pub const CLIENT_ID: &str = "3c49430b43dfba77"; // Hrm...
-pub const SYNC_SCOPE: &str = "https://identity.mozilla.com/apps/oldsync";
+// Note that in revision ba9cc53710243c357e5bf125636bb94a7c6a4a48 or so, this had support for
+// username/password authentication and for "temporary" accounts using restmail etc.
+// It has since been refactored to use only oauth and to use the "cli-support" crate to help
+// manage the account, but you can use that revision if you ever want to revive username/password
+// support.
```
tarikeshaq, line 22: Another thing we lose here (that we didn't use much in the first place) is the ability to fully automate the sync-tests where someone can run the sync-tests and it would go through the whole flow of creating the account, and syncing

Now the runner of the test would have to manually input the account information in a web browser window... given we no longer run the sync tests in CI I'm totally okay with it, it makes me a little sad that we couldn't get to a place where trust our integration tests enough to put them CI :(

testing/sync-test/src/auth.rs
```diff
@@ -1,233 +1,46 @@
...
+    fn fxa_account(&self) -> &FirefoxAccount {
+        &self.cli.account
     }
 }
 
+// Many of these tests simulate multiple devices (aka clients).
 pub struct TestClient {
-    pub fxa: fxa_client::internal::FirefoxAccount,
-    pub test_acct: Arc<TestAccount>,
+    pub acct: Arc<Account>,
```
tarikeshaq, line 42: What do you think about killing the `Account` struct all together and attaching its members here, looks like it only has one getter and it retrieved the inner `FirefoxAccount` so maybe we can just attach that?

testing/sync-test/src/main.rs
```diff
@@ -68,7 +71,19 @@ pub fn run_test_groups(opts: &Opts, groups: Vec<TestGroup>) {
...
 
 pub fn run_test_group(opts: &Opts, group: TestGroup) {
-    let mut user = TestUser::new(opts, 2).expect("Failed to get test user.");
+    if opts.helper_debug {
+        // What are these used for?
+        std::env::set_var("DEBUG", "nightmare");
+        std::env::set_var("HELPER_SHOW_BROWSER", "1");
+    }
+
+//    let cli_fxa = get_cli_fxa(opts.fxa_stack.to_config(), &opts.credential_file).expect("can't initialize cli");
```
tarikeshaq, line 80: nit: comment

testing/sync-test/src/auth.rs
```diff
@@ -237,40 +50,20 @@ pub struct TestClient {
...
-            let oauth_uri = fxa.begin_oauth_flow(&[SYNC_SCOPE], "integration_test", None)?;
-            let redirect_uri = acct.execute_oauth_flow(&oauth_uri)?;
-            let redirect_uri = Url::parse(&redirect_uri)?;
-            let query_params = redirect_uri
-                .query_pairs()
-                .into_owned()
-                .collect::<HashMap<String, String>>();
-            (query_params["code"].clone(), query_params["state"].clone())
+    pub fn new(acct: Arc<Account>, device_name: &str) -> Result<Self> {
+        // XXX - not clear if/how this device gets cleaned up!
```
tarikeshaq, line 54: The device is only destroyed when we disconnect the account (calling `disconnect`)

tarikeshaq (approved):

bendk (approved):
Looks good to me, once Tarik's suggestions are implemented (mainly the Glean one).  I don't know much about this client, but I hope to start using it soon to test some FxA changes.  We can continue to iterate as we do that.

testing/sync-test/src/auth.rs
bendk: A bit sad, but I'm not sure that our repo is the place to test some of this stuff.  Account creation should probably be tested alongside the FxA server code.

If we want to prioritize it, we could re-add some sort of integration test as part of the FxA work.  I always wonder if it's worth it or not though.

mhammond: I think the point is more that now the tests can't run from a clean environment without user interaction - to complete an oauth flow. Before this change all you needed was an email and password and it was fully automated.

PR #5671, opened 2023-06-16
Update detekt to version 1.23.0 [ci full]

rvandermeulen:
I wonder if at some point it'd be worth trying to align the config with AC. But that's a lot of errors I don't want to look at, so it won't be today! 😁


tarikeshaq (approved):
LGTM 🚀

PR #5683, opened 2023-06-23
Add some more commands to places-utils

mhammond:
* create-fake-visits will create a (possibly large) number of visits.
* delete-history will remove all history
* show-stats will show some stats for the places DB.

This is primarily so we can see the performance of delete-history


skhamis (approved):
Nice! Will definitely be handy in the near future! LGTM!

PR #5684, opened 2023-06-26
Add coenrolling features to enrollments evolver

jhugman:
Fixes [EXP-3622](https://mozilla-hub.atlassian.net/browse/EXP-3622).

  - Note:

### DIFF
--- a/components/nimbus/src/enrollment.rs
+++ b/components/nimbus/src/enrollment.rs
@@ -581,6 +581,7 @@ pub(crate) struct EnrollmentsEvolver<'a> {
     nimbus_id: &'a Uuid,
     available_randomization_units: &'a AvailableRandomizationUnits,
     targeting_helper: &'a NimbusTargetingHelper,
+    coenrolling_feature_ids: &'a HashSet<&'a str>,
 }
 
 impl<'a> EnrollmentsEvolver<'a> {
@@ -588,11 +589,13 @@ impl<'a> EnrollmentsEvolver<'a> {
         nimbus_id: &'a Uuid,
         available_randomization_units: &'a AvailableRandomizationUnits,
         targeting_helper: &'a NimbusTargetingHelper,
+        coenrolling_feature_ids: &'a HashSet<&str>,
     ) -> Self {
         Self {
             nimbus_id,
             available_randomization_units,
             targeting_helper,
+            coenrolling_feature_ids,
         }
     }
 
@@ -709,12 +712,13 @@ impl<'a> EnrollmentsEvolver<'a> {
     {
         let mut enrollment_events = vec![];
         let prev_experiments = map_experiments(prev_experiments);
-        let next_experiments = map_experiments(next_experiments);
-        let prev_enrollments = map_enrollments(prev_enrollments);
+        let next_experiments_map = map_experiments(next_experiments);
+        let prev_enrollments_map = map_enrollments(prev_enrollments);
 
         // Step 1. Build an initial active_features to keep track of
         // the features that are being experimented upon.
         let mut enrolled_features = HashMap::with_capacity(next_experiments.len());
+        let mut coenrolling_features = HashMap::with_capacity(next_experiments.len());
 
         let mut next_enrollments = Vec::with_capacity(next_experiments.len());
 
@@ -724,7 +728,7 @@ impl<'a> EnrollmentsEvolver<'a> {
         // start building up active_features, the map of feature_ids under
         // experiment to EnrolledFeatureConfigs, and next_enrollments.
 
-        for prev_enrollment in prev_enrollments.values() {
+        for prev_enrollment in prev_enrollments {
             if matches!(
                 prev_enrollment.status,
                 EnrollmentStatus::NotEnrolled {
@@ -738,7 +742,7 @@ impl<'a> EnrollmentsEvolver<'a> {
             let next_enrollment = match self.evolve_enrollment(
                 is_user_participating,
                 prev_experiments.get(slug).copied(),
-                next_experiments.get(slug).copied(),
+                next_experiments_map.get(slug).copied(),
                 Some(prev_enrollment),
                 &mut enrollment_events,
             ) {
@@ -756,15 +760,16 @@ impl<'a> EnrollmentsEvolver<'a> {
 
             self.reserve_enrolled_features(
                 next_enrollment,
-                &next_experiments,
+                &next_experiments_map,
                 &mut enrolled_features,
+                &mut coenrolling_features,
                 &mut next_enrollments,
             );
         }
 
         // Step 3. Evolve the remaining enrollments with the previous and
         // next data.
-        for next_experiment in next_experiments.values() {
+        for next_experiment in next_experiments {
             let slug = &next_experiment.slug;
 
             // Check that the feature ids that this experiment needs are available.  If not, then declare
@@ -816,7 +821,7 @@ impl<'a> EnrollmentsEvolver<'a> {
             // But we evolved all the existing enrollments in step 2,
             // (except the feature conflicted ones)
             // so we should be mindful that we don't evolve them a second time.
-            let prev_enrollment = prev_enrollments.get(slug).copied();
+            let prev_enrollment = prev_enrollments_map.get(slug).copied();
 
             if prev_enrollment.is_none()
                 || matches!(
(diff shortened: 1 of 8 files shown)


components/nimbus/src/enrollment.rs
```diff
@@ -1049,14 +1073,56 @@ pub fn map_features_by_feature_id(
...
+        // If we're not allowing co-enrollment for this feature, then add it to enrolled_features.
+        // We'll use this map to prevent collisions.
+        enrolled_features.insert(feature_id.clone(), enrolled_feature);
+    } else if let Some(existing) = coenrolling_features.get(feature_id) {
+        // Otherwise, we'll add to the coenrolling_features map.
+        // In this branch, we've enrolled in one experiment already before this one.
+        // We take care to merge this one with the existing one.
+        let merged = enrolled_feature
+            .defaults(existing)
+            .expect("A feature config hasn't been able to merge; this is a bug in Nimbus");
```
jeddai, line 1109: Do we have it set up such that the order in which enrolled feature configs will enter this method will be the same across all clients/enrollment cases? Thinking about the case where two experiments set different values for the same feature variable — wondering if each client could end up seeing different things as a result of that

components/nimbus/src/enrollment.rs
```diff
@@ -1318,3 +1384,94 @@ pub(crate) fn now_secs() -> u64 {
         .expect("Current date before Unix Epoch.")
         .as_secs()
 }
+
+#[cfg(test)]
+mod unit_tests {
```
jeddai, line 1389: Is this unit test module here because of access specification of the `populate_feature_maps` method? Curious if it could be moved into the base `test_enrollment` file.

components/nimbus/src/nimbus_client.rs
```diff
@@ -85,6 +86,10 @@ impl NimbusClient {
             app_context,
             database_cache: Default::default(),
             db_path: db_path.into(),
+            // With this being default, i.e. empty, Nimbus doesn't support coenrolling ids.
+            // Once the API is connected with Swift/Kotlin it will.
+            // This is the subject of https://mozilla-hub.atlassian.net/browse/EXP-3623
+            coenrolling_feature_ids: Default::default(),
```
jeddai, line 92: Okay so the idea here is the Swift/Kotlin generated code will send in the coenrolling feature ids when instantiating Nimbus?

components/nimbus/src/stateless/cirrus_client.rs
```diff
@@ -88,6 +89,10 @@ impl CirrusClient {
         let app_context: AppContext = serde_json::from_str(&app_context)?;
         Ok(Self {
             app_context,
+            // As it currently stands, with no coenrolling feature ids, cirrus cannot
+            // support coenrolling features.
+            // This is the subject of https://mozilla-hub.atlassian.net/browse/EXP-3623
+            coenrolling_feature_ids: Default::default(),
```
jeddai, line 95: Hmmmm this is where I'm now kinda wishing the FML library was available here because I'm not sure how we're gonna handle this from a deployment perspective

components/nimbus/src/tests/test_enrollment.rs
```diff
@@ -1595,6 +1633,367 @@ fn test_evolver_experiment_update_was_enrolled() -> Result<()> {
...
+    );
+    let exp4 = get_multi_feature_experiment(
+        "exp4",
+        "another",
+        json!({"p": 1 }),
+        "coenrolling",
+        json!({ "d": 7, "e": 8 }),
+    );
+
+    let all_experiments = [exp1, exp2, exp3.clone(), exp4.clone()];
```
jeddai, line 1927: > Thinking about the case where two experiments set different values for the same feature variable — wondering if each client could end up seeing different things as a result of that

As an example of this, what happens if we swap the order of these experiments around? Is there something somewhere that re-orders them based on something deterministic? 

Additionally, if you apply say experiments 1 and 2, then later apply experiments 1, 2, 3, and 4 do we end up with the same result as just applying all four of them at once? That might be more for a `nimbus_client` integration-style test, but I feel like it could be worthwhile to have.

jeddai (approved):
r+wc 

I've got some in-depth questions but nothing worth holding the PR up over (I think).

components/nimbus/src/tests/test_enrollment.rs
jeddai: I do want to say I recognize that in practice this scenario _should_ never happen, as coenrolling features should be highly configurable and key-value based (so key collisions are unlikely at best), but considering we don't have fine-tuned control over how users/devs are able to use coenrolling I think it'd be wise of us to at least know what we can expect if an app ends up in that scenario.

Do we have any plans to update Experimenter to identify targeting/key collisions across experiments for coenrolling features?

components/nimbus/src/enrollment.rs
```diff
@@ -847,17 +852,24 @@ impl<'a> EnrollmentsEvolver<'a> {
...
+        enrolled_features.extend(coenrolling_features);
+
         // Check that we generate the enrolled feature map from the new
         // enrollments and new experiments.  Perhaps this should just be an
         // assert.
-        let updated_enrolled_features = map_features(&next_enrollments, &next_experiments);
+        let updated_enrolled_features = map_features(
+            &next_enrollments,
+            &next_experiments_map,
+            self.coenrolling_feature_ids,
```
eliserichards, line 871: should this be `coenrolling_features`?

components/nimbus/src/enrollment.rs
```diff
@@ -1022,24 +1040,30 @@ where
 fn map_features(
     enrollments: &[ExperimentEnrollment],
     experiments: &HashMap<String, &Experiment>,
+    coenrolling_ids: &HashSet<&str>,
 ) -> HashMap<String, EnrolledFeatureConfig> {
-    let mut map = HashMap::with_capacity(enrollments.len());
+    let mut map1 = HashMap::with_capacity(enrollments.len());
+    let mut map2 = HashMap::with_capacity(enrollments.len());
```
eliserichards, line 1046: Nit: can we find better names than map1 and map2? why do we need two of them?

components/nimbus/src/stateless/cirrus_client.rs
eliserichards: What do you mean? 🤔

eliserichards (commented):
So 👏 many 👏 tests 👏 !

components/nimbus/src/tests/test_enrollment.rs
jeddai: Also also do we need to consider salting bucket sampling for experiments targeting only coenrolling features?

components/nimbus/src/stateless/cirrus_client.rs
jeddai: TL;DR I was wrong, it works fine we just need to add a way to pull the coenrolling feature ids from the FML client

So what I was referring to is that Nimbus will get these values because the generated code has them effectively hard-coded in it and will pass them into the NimbusClient object during instantiation. 

Cirrus does not have the luxury of generated code, so we'd need an environment variable with a comma-separated list or a config file or something of the sort to contain the list of ids. That lends to more chance for human error, so I was noting that it'd be nice if we could import the FML so that we had the FML client object here.

But in writing that out, I realized we don't need that. We just need to update the FML Client to have a method that returns the coenrolling feature ids — then the Python code would be adjusted to look something like this
1. have the FML file mounted somewhere (current plan)
2. instantiate the FML client(s) BEFORE the SDK client(s)
3. get the coenrolling feature ids from the FML client(s) via an FFI'd method
4. instantiate the SDK client(s) passing in the feature ids obtained in the previous step

jhugman: Please see [EXP-3624](https://mozilla-hub.atlassian.net/browse/EXP-3624) and [EXP-3626](https://mozilla-hub.atlassian.net/browse/EXP-3626) 😉 .

components/nimbus/src/nimbus_client.rs
jhugman: The `NimbusBuilder` will mediate: grabbing the generated list of ids from the generated code, and poke it into the `NimbusClient` at construction time.

components/nimbus/src/enrollment.rs
```diff
@@ -709,12 +712,13 @@ impl<'a> EnrollmentsEvolver<'a> {
     {
         let mut enrollment_events = vec![];
         let prev_experiments = map_experiments(prev_experiments);
-        let next_experiments = map_experiments(next_experiments);
-        let prev_enrollments = map_enrollments(prev_enrollments);
+        let next_experiments_map = map_experiments(next_experiments);
+        let prev_enrollments_map = map_enrollments(prev_enrollments);
```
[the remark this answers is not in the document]
jhugman: FWIW: renaming this to `map` was so that we could iterate through the`prev_enrollments` and `next_experiments` deterministically.

jeddai:
Yeah I’m coming on board with this feature a bit late I really need to read
up on all the docs/tickets

components/nimbus/src/enrollment.rs
jhugman: I made a change to `evolve_enrollment_recipes` to make it more deterministic, though I'm not yet ready to declare the SDK completely deterministic. However, in practice, I think this is for the app features to define FML schema that will merge well.

components/nimbus/src/tests/test_enrollment.rs
jhugman: > Also also do we need to consider salting bucket sampling for experiments targeting only coenrolling features?

Could you elaborate? perhaps in a document or meeting.

jeddai: Never mind, a quick look through the available experiments on Experimenter indicates to me we already do what I was referring to with namespaces right now

PR #5686, opened 2023-06-27
Proposed new Kotlin FxA API

bendk:
This is my proposal for what a Kotlin FxA API should look like, based on reviewing the [android-components service-fxa component](https://github.com/mozilla-mobile/firefox-android/tree/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa).  It's definitely a WIP, any feedback/criticism is welcome.

When I reviewed the android-components code, I felt like it was doing too much.  In addition to the FxA code, it was also doing a lot of work to handle sync.  This is attempt to isolated and factor out the functionality directly related to FxA from there.

  - Note:

### DIFF
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/Devices.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/Devices.kt
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fxa
+
+import mozilla.appservices.fxaclient.Device
+import mozilla.appservices.fxaclient.DeviceCapability
+import mozilla.appservices.fxaclient.DevicePushSubscription
+import mozilla.appservices.fxaclient.IncomingDeviceCommand
+import mozilla.appservices.sync15.DeviceType
+
+/**
+ * Manages the list of connected devices for a [FirefoxAccount]
+ */
+abstract class DeviceManager {
+    /**
+     * Set name of the current device.
+     * @param name New device name.
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun setDeviceName(name: String)
+
+    /**
+     * Get the current device list. May be incomplete if state was never queried.
+     * @return [DeviceList] describes current and other known devices
+     */
+    abstract fun getKnownDevices(): DeviceList
+
+    /**
+     * Set a [DevicePushSubscription] for the current device.
+     * @param subscription A new [DevicePushSubscription].
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun setDevicePushSubscription(subscription: DevicePushSubscription)
+
+    /**
+     * Send a command to a specified device.
+     * @param targetDeviceId A device ID of the recipient.
+     * @param title tab title
+     * @param url tab URL
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun sendSingleTab(targetDeviceId: String, title: String, url: String)
+
+    /**
+     * Refreshes [DeviceList]. Registered [AccountEventsObserver] observers will be notified.
+     *
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun refreshDevices()
+
+    /**
+     * Polls for any pending [DeviceCommandIncoming] commands.
+     * In case of new commands, registered [AccountEventsObserver] observers will be notified.
+     *
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun pollForDeviceCommands()
+}
+
+/**
+ * Configuration for the current device.
+ *
+ * @property name An initial name to use for the device record which will be created during authentication.
+ * This can be changed later via [setDeviceName].
+ * @property type Type of a device - mobile, desktop - used for displaying identifying icons on other devices.
+ * This cannot be changed once device record is created.
+ * @property capabilities A set of device capabilities, such as SEND_TAB.
+ * @property secureStateAtRest A flag indicating whether or not to use encrypted storage for the persisted account
+ * state.
+ */
+data class DeviceConfig(
+    val name: String,
+    val type: DeviceType,
+    val capabilities: Set<DeviceCapability>,
+    val secureStateAtRest: Boolean = false,
+)
+
+/**
+ * Describes current device and other known devices.
+ */
+data class DeviceList(val currentDevice: Device, val otherDevices: List<Device>)
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
@@ -1,0 +1,153 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fxa
+
+import mozilla.appservices.fxaclient.AccessTokenInfo
+import mozilla.appservices.fxaclient.Device
+import mozilla.appservices.fxaclient.DeviceCapability
+import mozilla.appservices.fxaclient.DevicePushSubscription
+import mozilla.appservices.fxaclient.FxaConfig
+import mozilla.appservices.fxaclient.FirefoxAccount as RustFirefoxAccount
+import mozilla.appservices.fxaclient.IncomingDeviceCommand
+import mozilla.appservices.fxaclient.Profile
+import mozilla.appservices.sync15.DeviceType
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Top-level class to manage a FxA account
+ *
+ * This class contains the top-level functionality like fetching tokens and listening for account
+ * events.  See also:
+ *   - StateManager: manages the account state machine.  Defined in `State.kt` and available from
+ *     the `stateManager()` method.
+ *   - DeviceManager: manages the list of connected devices to an FxA account.  Defined in
+ *     `Devices.kt` and available from the `deviceManager()` method.
+ */
+abstract class FirefoxAccount(
+    private val config: FxaConfig,
+    private val deviceConfig: DeviceConfig,
+    private val storageHandler: StorageHandler,
+    private val applicationScopes: List<Scope> = listOf(),
+) {
+    /**
+     * Tries to fetch an access token for the given scope.
+     *
+     * @param singleScope Single OAuth scope (no spaces) for which the client wants access
+     * @return [AccessTokenInfo] that stores the token, along with its scope, key and
+     *                           expiration timestamp (in seconds) since epoch when complete
+     */
+    abstract suspend fun getAccessToken(singleScope: String): AccessTokenInfo?
+
+    /**
+     * Fetches the profile object for the current client either from the existing cached state
+     * or from the server (requires the client to have access to the profile scope).
+     *
+     * @param ignoreCache Fetch the profile information directly from the server
+     * @return Profile (optional, if successfully retrieved) representing the user's basic profile info
+     */
+    abstract suspend fun getProfile(ignoreCache: Boolean = false): Profile?
+
+    /**
+     * Fetches the token server endpoint, for authentication using the SAML bearer flow.
+     *
+     * @return Token server endpoint URL string, `null` if it couldn't be obtained.
+     */
+    abstract suspend fun getTokenServerEndpointURL(): String?
+
+    /**
+     * Get the pairing URL to navigate to on the Authority side (typically a computer).
+     *
+     * @return The URL to show the pairing user
+     */
+    abstract fun getPairingAuthorityURL(): String
+
+    /**
+     * Returns current FxA Device ID for an authenticated account.
+     *
+     * @return Current device's FxA ID, if available. `null` otherwise.
+     */
+    abstract fun getCurrentDeviceId(): String?
+
+    /**
+     * Returns session token for an authenticated account.
+     *
+     * @return Current account's session token, if available. `null` otherwise.
+     */
+    abstract fun getSessionToken(): String?
+
+    /**
+     * Indicates if account needs to be re-authenticated via [beginAuthentication].
+     * Most common reason for an account to need re-authentication is a password change.
+     *
+     * TODO this may return a false-positive, if we're currently going through a recovery flow.
+     * Prefer to be notified of auth problems via [AccountObserver], which is reliable.
+     *
+     * @return A boolean flag indicating if account needs to be re-authenticated.
+     */
+    abstract fun accountNeedsReauth(): Boolean
+
+    /**
+     * This method should be called when a request made with an OAuth token failed with an
+     * authentication error. It will re-build cached state and perform a connectivity check.
+     *
+     * In time, fxalib will grow a similar method, at which point we'll just relay to it.
+     * See https://github.com/mozilla/application-services/issues/1263
+     *
+     * @param singleScope An oauth scope for which to check authorization state.
+     * @return An optional [Boolean] flag indicating if we're connected, or need to go through
+     * re-authentication. A null result means we were not able to determine state at this time.
+     */
+    abstract suspend fun checkAuthorizationStatus(singleScope: String): Boolean?
+
+    /**
+     * Register for account events
+     */
+    abstract suspend fun registerAccountEventHandler(handler: AccountEventHandler)
+
+    abstract fun stateManager(): StateManager
+
+    abstract fun deviceManager(): DeviceManager
+}
+
+interface AccountEventHandler {
+    fun onAccountEvent(event: AccountEvent)
+}
+
+sealed class AccountEvent  {
+    /** The account's profile was updated */
+    object ProfileUpdated : AccountEvent()
+
+    /** The authentication state of the account changed - eg, the password changed */
+    data class AccountStateChanged(val state: AccountState) : AccountEvent()
+
+    /** The account itself was destroyed */
+    object AccountDestroyed : AccountEvent()
+
+    /** An incoming command from another device */
+    data class DeviceCommandIncoming(val command: IncomingDeviceCommand) : AccountEvent()
+
+    /** Another device connected to the account */
+    data class devicesChanged(val deviceList: DeviceList) : AccountEvent()
+}
+
+interface StorageHandler {
+    /**
+     * Called when the FirefoxAccount state has changed and should be saved
+     *
+     * The `state` that was last passed to this method is what should be passed back in
+     * `StateManager.initialize()`
+     */
+    fun saveState(state: String)
+}
+
+/** Scopes that can be requested from the FxA Server */
+enum class Scope {
+    // Fetch the user's profile.
+    Profile,
+    // Obtain sync keys for the user
+    Sync,
+    // Obtain a sessionToken, which gives full access to the account.
+    Session,
+}
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/State.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxa/State.kt
@@ -1,0 +1,126 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fxa
+
+/**
+ * Manages state transitions for a FirefoxAccount
+ *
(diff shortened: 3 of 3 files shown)


components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+    abstract suspend fun checkAuthorizationStatus(singleScope: String): Boolean?
+
+    /**
+     * Register for account events
+     */
+    abstract suspend fun registerAccountEventHandler(handler: AccountEventHandler)
+
+    abstract fun stateManager(): StateManager
+
+    abstract fun deviceManager(): DeviceManager
```
bendk, line 111: One thing I liked about the android components code was that it split things up into different components, which helped keep things understandable.  What do you think of doing this for our component?  What is the correct split?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/Devices.kt
```diff
@@ -0,0 +1,83 @@
...
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun sendSingleTab(targetDeviceId: String, title: String, url: String)
+
+    /**
+     * Refreshes [DeviceList]. Registered [AccountEventsObserver] observers will be notified.
+     *
+     * @throwns FxaException.Authentication
+     */
+    abstract suspend fun refreshDevices()
```
bendk, line 51: The abstract functions don't make this clear, but I want most of these methods should have some common functionality, probably achieved by some sort of wrapper:

 - If they alter the internal state, then `StorageHandler.saveState()` should be called (similar to [this android-components code](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt#L44-L65))
 - If we see an error with one of our auth tokens, then the internal state should be changed to `AccountState.Disconnected`, and `AccountEvent.AccountStateChanged()` should be sent out.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/State.kt
```diff
@@ -0,0 +1,126 @@
...
+     * devices will no longer see the current device in their device lists.
+     *
+     * @throws FxaException.InvalidState: the account was not in the `Connected` state
+     */
+    abstract suspend fun disconnect()
+}
+
+sealed class AccountState  {
+    /** The account is waiting for the `StateManager.initialize()` call */
+    object Uninitialized : AccountState()
```
bendk, line 107: I'm not sure if this is needed, but it seems like a lot of consumers are creating a nullable firefox account instance to defer the disk IO until after startup.  It might be better to formally encode that as a state of the account.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+     * @return The URL to show the pairing user
+     */
+    abstract fun getPairingAuthorityURL(): String
+
+    /**
+     * Returns current FxA Device ID for an authenticated account.
+     *
+     * @return Current device's FxA ID, if available. `null` otherwise.
+     */
+    abstract fun getCurrentDeviceId(): String?
```
bendk, line 71: A general question is how to handle disconnected accounts.  I made it so that getter functions like this return an optional result, while setter methods throw `FxaException.Authentication`.  WDYT?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    /** The account itself was destroyed */
+    object AccountDestroyed : AccountEvent()
+
+    /** An incoming command from another device */
+    data class DeviceCommandIncoming(val command: IncomingDeviceCommand) : AccountEvent()
+
+    /** Another device connected to the account */
+    data class devicesChanged(val deviceList: DeviceList) : AccountEvent()
+}
```
bendk, line 133: android-components has too many different event handlers, there's [AccountObserver](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt#L277-L319), [AccountEventsObserver](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/AccountEvent.kt#L10-L41) and [DeviceConstellationObserver](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt#L113-L115).

I think we should just one event type, but make it an enum / sealed class.  Observers match the event type and ignore the ones they don't care about.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+     * authentication error. It will re-build cached state and perform a connectivity check.
+     *
+     * In time, fxalib will grow a similar method, at which point we'll just relay to it.
+     * See https://github.com/mozilla/application-services/issues/1263
+     *
+     * @param singleScope An oauth scope for which to check authorization state.
+     * @return An optional [Boolean] flag indicating if we're connected, or need to go through
+     * re-authentication. A null result means we were not able to determine state at this time.
+     */
+    abstract suspend fun checkAuthorizationStatus(singleScope: String): Boolean?
```
mhammond, line 102: It looks like we should drop the param from this function - a-c only ever passes the profile scope and the equiv function in fxa-client takes no param.

Also wondering if we need the bool return here? The implication is that if this found an auth problem, `accountsNeedsReauth()` would start returning `true` (or in my suggestion above, `getAccountState()` would start returning `AuthenticationProblem`) - which seems a bit clearer and easier to document (ie, after this function you should check the account state via the usual means" seems mildly better?)

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Indicates if account needs to be re-authenticated via [beginAuthentication].
+     * Most common reason for an account to need re-authentication is a password change.
+     *
+     * TODO this may return a false-positive, if we're currently going through a recovery flow.
+     * Prefer to be notified of auth problems via [AccountObserver], which is reliable.
+     *
+     * @return A boolean flag indicating if account needs to be re-authenticated.
+     */
+    abstract fun accountNeedsReauth(): Boolean
```
mhammond, line 89: This seems a little wrong to me - the [existing function in a-c](https://github.com/mozilla-mobile/firefox-android/blob/d6b6f3c63cae8203c989775d9d24833a6c56dcb7/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L312) checks some other internal state enum, which in that case has a `AuthenticationProblem` - but our state enum has no equivalent state - which IMO it probably should.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+    data class AccountStateChanged(val state: AccountState) : AccountEvent()
+
+    /** The account itself was destroyed */
+    object AccountDestroyed : AccountEvent()
+
+    /** An incoming command from another device */
+    data class DeviceCommandIncoming(val command: IncomingDeviceCommand) : AccountEvent()
+
+    /** Another device connected to the account */
+    data class devicesChanged(val deviceList: DeviceList) : AccountEvent()
```
mhammond, line 132: Probably want upper-case `D` at the start of this one? And I guess similar question to above - why isn't this an enum in udl?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/Devices.kt
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @property name An initial name to use for the device record which will be created during authentication.
+ * This can be changed later via [setDeviceName].
+ * @property type Type of a device - mobile, desktop - used for displaying identifying icons on other devices.
+ * This cannot be changed once device record is created.
+ * @property capabilities A set of device capabilities, such as SEND_TAB.
+ * @property secureStateAtRest A flag indicating whether or not to use encrypted storage for the persisted account
+ * state.
+ */
+data class DeviceConfig(
```
mhammond, line 73: Not quite sure of the scope of this PR - is it to describe a shape but not the impl, or is is something that would be checked in?

Specifically, I'm wondering why this `data class`, and the `enum` below, wouldn't be in UDL? And some of the functions below are duplicated - is that mainly because of the `suspend` attribute, and we'd ideally just remove those wrappers once uniffi could help, or are these wrappers a "feature"?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+ * Top-level class to manage a FxA account
+ *
+ * This class contains the top-level functionality like fetching tokens and listening for account
+ * events.  See also:
+ *   - StateManager: manages the account state machine.  Defined in `State.kt` and available from
+ *     the `stateManager()` method.
+ *   - DeviceManager: manages the list of connected devices to an FxA account.  Defined in
+ *     `Devices.kt` and available from the `deviceManager()` method.
+ */
+abstract class FirefoxAccount(
```
mhammond, line 28: I don't quite understand why we'd want an abstract class like this in our repo and why these private vals exist? eg, is `applicationScopes` how we'd obtain the list of scopes we care about and how does it relate to the `singleScope` params used below?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
mhammond: Which part of a-c are you referring to? I still don't really understand some of the splits in a-c - I can't tell you how a `GlobalAccountManager` relates to an `FxaAccountManager`, and `SyncEnginesStorage`, `AccountStorage` and `SyncAuthInfoCache` all seems vaguely related to storage but aren't in a directory names `store`?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+     * @return Profile (optional, if successfully retrieved) representing the user's basic profile info
+     */
+    abstract suspend fun getProfile(ignoreCache: Boolean = false): Profile?
+
+    /**
+     * Fetches the token server endpoint, for authentication using the SAML bearer flow.
+     *
+     * @return Token server endpoint URL string, `null` if it couldn't be obtained.
+     */
+    abstract suspend fun getTokenServerEndpointURL(): String?
```
mhammond, line 57: should we be using urls in the new api?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+     * @return Current device's FxA ID, if available. `null` otherwise.
+     */
+    abstract fun getCurrentDeviceId(): String?
+
+    /**
+     * Returns session token for an authenticated account.
+     *
+     * @return Current account's session token, if available. `null` otherwise.
+     */
+    abstract fun getSessionToken(): String?
```
mhammond, line 78: when would this return null? If only when the account is in a bad state I wonder if an error return is more appropriate?

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,153 @@
...
+     * Called when the FirefoxAccount state has changed and should be saved
+     *
+     * The `state` that was last passed to this method is what should be passed back in
+     * `StateManager.initialize()`
+     */
+    fun saveState(state: String)
+}
+
+/** Scopes that can be requested from the FxA Server */
+enum class Scope {
```
mhammond, line 146: same question re this being in the UDL, and also why it's not used as the param to, say, `getAccessToken()`.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
mhammond: I think I'd prefer an error return - most code is going to assume the account remains in a good state and probably *want* to throw an exception when this turns out to not be true, so somewhere further up the call stack and notice the transition.

To put it another way - the code that is calling `getCurrentDeviceId()` is almost certainly never going to be called if the account was known to be in a disconnected state, so I don't think having it handle null makes sense - what would be better is somewhere up the callstack handles the error, checks the new state, and updates things in such a way that nothing is going to call `getCurrentDeviceId()`

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
bendk: It's confusing to me as well and I think there's way too many files overall.  But I do like the general idea of splitting up the API surface.  The current logic of this PR goes like this:
  - `FirefoxAccount` is kind of like [FxaAccountManager](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt).  It's the top-level class and handles general functionality
  - `DeviceManager` is kind of like [FxaDeviceConstellation](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt).  It handles device-related functionality.
  - `StateManager` is kind of like [FirefoxAccount](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt) it has methods to initiate state transitions.  This one has a lot of changes compared to the android-components class. I made it focus only on state transitions, which meant removing methods like `getAccessToken()`, and I also added the `Uninitialized` state.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
bendk: This method feels weird to me, can we kill it?  The docs say "This method should be called when a request made with an OAuth token failed with an authentication error.".  Why should this be the responsibility of the consumer?  It seems better for the component to make this check and emit a `AccountStateChanged` event if it notices there's an issue.

mhammond: It makes sense to me that this becomes internal.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
bendk: `AccountState.Disconnected` has the `fromAuthIssues` flag, which can represent that.  I tried to lower the number of top-level states, since I think a lot of code isn't only going to care if the account is connected vs disconnected.  I'm not sure if that's a good idea or not though, what do you think?

However we decide to handle the state enum, maybe this method should be removed since `getState()` can tell you the same information.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/Devices.kt
bendk: It's more to describe the shape.  I don't want to write any actual code until Lougenia does her analysis with the Swift API and we can compare and contrast the two.

I consider the wrappers a necessary evil that's hopefully temporary.  My hope is that once we decide on an API, we implement almost all of the API in Rust and the Swift/Kotlin wrappers would just deal with the async parts.  And then those wrappers can fall off once UniFFI has a better async story.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
bendk: It's abstract just because I didn't want to write the code.  I don't think it should be in the final version.  And the `private` keyword was a mistake, I think all of those should be constructor arguments.

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
mhammond: > If the UniFFI custom converter fails, then I think we would get a panic,

Good point, but I think this is probably solvable in the same way you solved it for callback functions, right? If so, I'll open a uniffi issue

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,158 @@
...
+    val config: FxaConfig,
+    val deviceConfig: DeviceConfig,
+    val storageHandler: StorageHandler,
+    val applicationScopes: List<Scope> = listOf(),
+) {
+    /**
+     * Tries to fetch an access token for the given scope.
+     *
+     * @param singleScope Single OAuth scope (no spaces) for which the client wants access
+     * @return [AccessTokenInfo] that stores the token, along with its scope, key and
```
tarikeshaq, line 38: Noting for future: We might have an opportunity here to improve the return type here, `AccessTokenInfo` carrying the encryption key feels like a footgun...I could see a future consumer making a mistake and sending over the key to a server along with the access token

I'd prefer if the operation to retrieve the scoped key was separate.. [i.e a `getKeyForScope` function like desktop does](https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccountsKeys.sys.mjs#124)

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
tarikeshaq: +1 to killing the method and having it internally delt with.

In the future, I can see us re-adding it if more non-firefox scopes get introduced to our component - for example, if Fenix wants to check if it already has Monitor scoped access token, but it doesn't, it wouldn't be an error state. Fenix would just need to make sure the user approves the new scope or force-issue a new access token using the session token.

But in the meantime, it doesn't add much value being an exposed API to Fenix or iOS

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,158 @@
...
+     * @param singleScope An oauth scope for which to check authorization state.
+     * @return An optional [Boolean] flag indicating if we're connected, or need to go through
+     * re-authentication. A null result means we were not able to determine state at this time.
+     */
+    abstract suspend fun checkAuthorizationStatus(singleScope: String): Boolean?
+
+    /**
+     * Register for account events
+     */
+    abstract suspend fun registerAccountEventHandler(handler: AccountEventHandler)
```
tarikeshaq, line 112: What do you think about moving this to the constructor? forcing the consumer to make decision about what to do with the events... I could see the argument for allowing the consumer not to set the handlers so I could go for either 🤔

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
```diff
@@ -0,0 +1,158 @@
...
+    data class AccountStateChanged(val state: AccountState) : AccountEvent()
+
+    /** The account itself was destroyed */
+    object AccountDestroyed : AccountEvent()
+
+    /** An incoming command from another device */
+    data class DeviceCommandIncoming(val command: IncomingDeviceCommand) : AccountEvent()
+
+    /** Another device connected to the account */
+    data class DevicesChanged(val deviceList: DeviceList) : AccountEvent()
```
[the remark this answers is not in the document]
bendk: That might make sense for this one, how does the event get emitted anyways?

In general, I think we may want to support two ways to get certain information.  For example, the `getState()` method and `AccountStateChanged` event.  Listening to events seems better for GUI consumers.  But `example-fxa-client` would probably rather call `getState()`

components/fxa-client/android/src/main/java/mozilla/appservices/fxa/FirefoxAccount.kt
bendk: I'm in the middle too.  Originally I thought it should be in the constructor, but it's possible that consumers might not be ready to construct their event handler yet.

PR #5688, opened 2023-06-27
Fixes EXP-3624: Allow coenrollment in Nimbus FML

eliserichards:

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # v117.0 (In progress)
 
+## Nimbus FML ⛅️🔬🔭🔧
+
+### ✨ What's New ✨
+
+- Add `allow-coenrollment` property for features in the Feature Manifest Language. This relaxes the feature exclusion rules for features marked with `allow-coenrollment: true`. ([#5688](https://github.com/mozilla/application-services/pull/5688)).
+  - This adds a non-user-facing method to the `FeatureManifestInterface`, `getCoenrollingFeatureIds`, in both Kotlin and Swift.
+
 [Full Changelog](In progress)
 
 # v116.0 (_2023-07-03_)
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureManifestInterface.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureManifestInterface.kt
@@ -41,4 +41,6 @@ interface FeatureManifestInterface<T> {
      * is quite useful for introspection.
      */
     fun getFeature(featureId: String): FeatureHolder<*>?
+
+    fun getCoenrollingFeatureIds(): List<String>
 }
--- a/components/nimbus/ios/Nimbus/FeatureManifestInterface.swift
+++ b/components/nimbus/ios/Nimbus/FeatureManifestInterface.swift
@@ -31,4 +31,6 @@ public protocol FeatureManifestInterface {
     ///
     /// This happens automatically if you use the `NimbusBuilder` pattern of initialization.
     func invalidateCachedValues()
+
+    func getCoenrollingFeatureIds() -> [String]
 }
--- a/components/support/nimbus-fml/fixtures/fe/including/ios_homescreen.yaml
+++ b/components/support/nimbus-fml/fixtures/fe/including/ios_homescreen.yaml
@@ -28,6 +28,7 @@ features:
             "library-shortcuts": true
           }
         }
+    allow_coenrollment: false
 enums:
   HomeScreenSection:
     description: The identifiers for the sections of the homescreen.
--- a/components/support/nimbus-fml/fixtures/fe/including/ios_homescreen.yaml
+++ b/components/support/nimbus-fml/fixtures/fe/including/ios_homescreen.yaml
@@ -28,7 +28,6 @@ features:
             "library-shortcuts": true
           }
         }
-    allow_coenrollment: false
 enums:
   HomeScreenSection:
     description: The identifiers for the sections of the homescreen.
--- a/components/support/nimbus-fml/fixtures/fe/including/ios_search.yaml
+++ b/components/support/nimbus-fml/fixtures/fe/including/ios_search.yaml
@@ -22,6 +22,7 @@ features:
             "keep-for-days": 100
           }
     defaults: null
+    allow_coenrollment: false
 objects:
   AwesomeBar:
     description: "Represents the awesome bar object"
--- a/components/support/nimbus-fml/fixtures/fe/including/ios_search.yaml
+++ b/components/support/nimbus-fml/fixtures/fe/including/ios_search.yaml
@@ -22,7 +22,6 @@ features:
             "keep-for-days": 100
           }
     defaults: null
-    allow_coenrollment: false
 objects:
   AwesomeBar:
     description: "Represents the awesome bar object"
--- a/components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
+++ b/components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
@@ -96,6 +96,15 @@ object {{ nimbus_object }} : FeatureManifestInterface<{{ nimbus_object }}.Featur
     }
 
     /**
+     * Get a list of feature ids where the feature allows co-enrollment.
+     */
+    override fun getCoenrollingFeatureIds(): List<String> =
+        listOf(
+            {%- for f in self.fm.get_coenrolling_feature_ids() %}
+            {{- f|quoted }}
+            {%- if !loop.last %}, {% endif %}
+            {%- endfor %})
+    /**
      * Accessor object for generated configuration classes extracted from Nimbus, with built-in
      * default values.
      */
--- a/components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
+++ b/components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
@@ -47,6 +47,14 @@ public class {{ nimbus_object }} : FeatureManifestInterface {
     public let features = Features()
 
     {% let blocks = self.initialization_code() -%}
+
+    public func getCoenrollingFeatureIds() -> [String] {
+        [{%- for f in self.fm.get_coenrolling_feature_ids() %}
+            {{- f|quoted }}
+            {%- if !loop.last %}, {% endif %}
+            {%- endfor %}]
+    }
+
     ///
     /// All generated initialization code. Clients shouldn't need to override or call
     /// this.
@@ -104,4 +112,4 @@ public class {{ features_object }} {
 {{- code }}
 {%- endfor %}
 
-{% import "macros.kt" as kt %}
+{% import "macros.swift" as swift %}
--- a/components/support/nimbus-fml/src/frontend.rs
+++ b/components/support/nimbus-fml/src/frontend.rs
@@ -131,6 +131,8 @@ pub(crate) struct FeatureBody {
     #[serde(alias = "defaults")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) default: Option<Vec<DefaultBlock>>,
+    #[serde(default)]
+    pub(crate) allow_coenrollment: bool,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, Default)]
@@ -251,6 +253,7 @@ impl ManifestFrontEnd {
                         .iter()
                         .map(|v| self.get_prop_def_from_field(v))
                         .collect(),
+                    allow_coenrollment: body.allow_coenrollment.clone(),
                 };
 
                 merger.merge_feature_defaults(&mut def, &body.default)?;
--- a/components/support/nimbus-fml/src/frontend.rs
+++ b/components/support/nimbus-fml/src/frontend.rs
@@ -120,6 +120,7 @@ pub(crate) struct ImportBlock {
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct FeatureBody {
     pub(crate) description: String,
     // We need these in a deterministic order, so they are stable across multiple
@@ -132,6 +133,7 @@ pub(crate) struct FeatureBody {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) default: Option<Vec<DefaultBlock>>,
     #[serde(default)]
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
     pub(crate) allow_coenrollment: bool,
 }
 
@@ -253,7 +255,7 @@ impl ManifestFrontEnd {
                         .iter()
                         .map(|v| self.get_prop_def_from_field(v))
                         .collect(),
-                    allow_coenrollment: body.allow_coenrollment.clone(),
+                    allow_coenrollment: body.allow_coenrollment,
                 };
 
                 merger.merge_feature_defaults(&mut def, &body.default)?;
--- a/components/support/nimbus-fml/src/intermediate_representation.rs
+++ b/components/support/nimbus-fml/src/intermediate_representation.rs
@@ -640,9 +640,10 @@ impl FeatureManifest {
         self.iter_feature_defs().find(|f| f.name() == nm)
     }
 
-    pub fn get_coenrolling_feature_ids(&self) -> Vec<&FeatureDef> {
-        self.iter_feature_defs()
-            .filter(|f| f.allow_coenrollment())
+    pub fn get_coenrolling_feature_ids(&self) -> Vec<String> {
+        self.iter_all_feature_defs()
+            .filter(|(_, f)| f.allow_coenrollment())
+            .map(|(_, f)| f.name())
             .collect()
     }
 
@@ -725,7 +726,7 @@ impl FeatureDef {
         self.props.clone()
     }
     pub fn allow_coenrollment(&self) -> bool {
-        self.allow_coenrollment.clone()
+        self.allow_coenrollment
     }
 
     pub fn default_json(&self) -> Value {
@@ -963,6 +964,27 @@ pub mod unit_tests {
         ];
         fm.validate_manifest()
             .expect_err("Should fail on duplicate obj_defs");
+        Ok(())
+    }
+
+    #[test]
+    fn validate_allow_coenrollment() -> Result<()> {
+        let mut fm = get_simple_homescreen_feature();
+        fm.feature_defs.push(FeatureDef::new(
+            "some_def",
+            "my lovely qtest doc",
+            vec![PropDef {
+                name: "some prop".into(),
+                doc: "some prop doc".into(),
+                typ: TypeRef::String,
+                default: json!("default"),
+            }],
+            true,
+        ));
+        fm.validate_manifest()?;
+        let coenrolling_ids = fm.get_coenrolling_feature_ids();
+        assert_eq!(coenrolling_ids, vec!["some_def".to_string()]);
+
         Ok(())
     }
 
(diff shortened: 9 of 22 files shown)


components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
```diff
@@ -95,6 +96,21 @@ object {{ nimbus_object }} : FeatureManifestInterface<{{ nimbus_object }}.Featur
...
+        {%- let features = self.iter_feature_defs() %}
+        {%- if features.is_empty() %}
+        return emptyList()
+        {%- let coenrolledFeatures = ArrayList<String> %}
+        {%- for f in features %}
+        {%- if f.allow_coenrollment() %}
+        coenrolledFeatures += features.{{- f.name() -}}
+        {%- endfor %}
+        return coenrolledFeatures
+    }
```
jhugman, line 112: A few comments here:
- I'd prefer to standardize on one of `conenrolling` or `coenrolled`; #5684 at the moment uses `coenrolling`; let's pick one and stick to it.
- The return type doesn't need to be optional.
- I think you're doing the hard work in `get_coenrolling_feature_ids`; no need to duplicate it in the templates.

```suggestion
    override fun getCoenrollingFeatureIds() = listOf(
            {%- for f in self.fm.get_coenrolling_feature_ids() %}{{ f|quoted }}
            {%- if !loop.last %}, {% endif %}
            {%- endfor %})
```

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureManifestInterface.kt
```diff
@@ -41,4 +41,9 @@ interface FeatureManifestInterface<T> {
      * is quite useful for introspection.
      */
     fun getFeature(featureId: String): FeatureHolder<*>?
+
+    /**
+     * Get a list of feature ids where the feature allows co-enrollment.
+     */
+    fun getCoenrolledFeatureIds(): List<String>? = null
```
jhugman, line 48: I don't think this should be optional. You can always use `listOf()` if you want an empty list.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
```diff
@@ -149,4 +165,6 @@ object {{ nimbus_object }} : FeatureManifestInterface<{{ nimbus_object }}.Featur
 {{- code }}
 {%- endfor %}
 
-{% import "macros.kt" as kt %}
\ No newline at end of file
+{% import "macros.kt" as kt %}
+{% import "java.util.Collections.emptyList"%}
+{% import "java.util.ArrayList"%}
```
jhugman, line 170: [Askama's `import` directive](https://djc.github.io/askama/template_syntax.html?highlight=import#macros) is a little different to Kotlin; I don't _think_ you need to add anymore Kotlin imports, but you can do it with [`FeatureCodeDeclaration`](https://github.com/mozilla/application-services/blob/f2e7ee575bdd35be3c61111a767aa83484c25b8c/components/support/nimbus-fml/src/backends/kotlin/gen_structs/feature.rs#L40).

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
```diff
@@ -80,6 +80,22 @@ public class {{ nimbus_object }} : FeatureManifestInterface {
...
+        {%- let features = self.iter_feature_defs() %}
+        {%- if features.is_empty() %}
+        return []
+        {%- let coenrolledFeatures = ArrayList<String> %}
+        {%- for f in features %}
+        {%- if f.allow_coenrollment() %}
+        coenrolledFeatures += features.{{- f.name() -}}
+        {%- endfor %}
+        return coenrolledFeatures
+    }
```
jhugman, line 97: I think get one language going before doing the second one. That way you don't need to deal with the same failing tests twice, you get to deal with two different sets of failing tests!

components/support/nimbus-fml/src/backends/swift/templates/FeatureManifestTemplate.swift
```diff
@@ -104,4 +121,4 @@ public class {{ features_object }} {
 {{- code }}
 {%- endfor %}
 
-{% import "macros.kt" as kt %}
\ No newline at end of file
+{% import "macros.swift" as s %}
```
jhugman, line 124: Well this is weird. It seems like no macros were being used in this file. It should read:

```suggestion
{% import "macros.swift" as swift %}
```

or deleted.

components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -954,6 +967,26 @@ pub mod unit_tests {
...
+            "some_def",
+            "test doc",
+            vec![PropDef {
+                name: "prop name".into(),
+                doc: "prop doc".into(),
+                typ: TypeRef::EnumMap(
+                    Box::new(TypeRef::Enum("SectionId".into())),
+                    Box::new(TypeRef::String),
+                ),
+                default: json!(null),
```
jhugman, line 983: You likely want a simpler test here:

```suggestion
                name: "my-variable".into(),
                doc: "my documentation".into(),
                typ: TypeRef::String,
                default: json!("arbitrary string"),
```

components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -954,6 +967,26 @@ pub mod unit_tests {
...
+            }],
+            true,
+        ));
+        fm.validate_manifest()
```
jhugman, line 987: Then check if `fm.get_coenrolling_feature_ids()` works, and returns `vec!["some_def".to_string()]`.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
```diff
@@ -38,6 +38,7 @@ object {{ nimbus_object }} : FeatureManifestInterface<{{ nimbus_object }}.Featur
         {%- let raw_name = f.name() %}
         {%- let class_name = raw_name|class_name %}
         {{ f.doc()|comment("        ") }}
+        {%- let allow_coenrollment = f.allow_coenrollment() %}
```
jhugman, line 41: I don't think this is doing anything. :(

components/support/nimbus-fml/src/frontend.rs
```diff
@@ -131,6 +131,8 @@ pub(crate) struct FeatureBody {
     #[serde(alias = "defaults")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) default: Option<Vec<DefaultBlock>>,
+    #[serde(default)]
+    pub(crate) allow_coenrollment: bool,
```
jhugman, line 135: I think you should be able to skip this if it's `false`.

Edit: you can, but it's a little clunky:
```suggestion
    #[serde(default)]
    #[serde(skip_serializing_if = "std::ops::Not::not")]
    pub(crate) allow_coenrollment: bool,
```

jhugman (commented):
Overall, this is a good first stab at this task. Good job!

I would tackle this in stages: 

- ignore the Kotlin & Swift side altogether: add `allow_coenrollment` and get the Rust to compile, and add some Rust tests.
- do the Kotlin code generation; get used to how finickety Askama can be. Compiling the templates is tricky, and the error messages are difficult to debug. Move in small steps, commiting every time you move forward until you get more confident!
- once the Kotlin generates and compiles, write some tests in Kotlin (with some FML fixtures)
- once the kotlin compiles and tests, then repeat the process but for Swift.

Keep at it! It's hard! But you're doin' it!

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
eliserichards: what is `{{ f|quoted }}` doing here?

jeddai: It's a pipe similar to what you might see in other templating frameworks — Angular/Helm both have them, as examples. 

`quoted` is a function that's declared in the `gen_structs/common` file, and the pipe passes the left-hand value into the function

https://github.com/mozilla/application-services/blob/c2823f590b26c259becd103e76050bbff1d272f2/components/support/nimbus-fml/src/backends/kotlin/gen_structs/common.rs#L22-L32

jeddai: I'm realizing now you may have been asking _why_ it's there and not what it is doing, so sorry

eliserichards: no, you were right! thank you so much for the explanation :D

components/support/nimbus-fml/src/frontend.rs
```diff
@@ -251,6 +254,7 @@ impl ManifestFrontEnd {
                         .iter()
                         .map(|v| self.get_prop_def_from_field(v))
                         .collect(),
+                    allow_coenrollment: body.allow_coenrollment.clone(),
```
eliserichards, line 257: Does this need the `.clone()` here?

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
```diff
@@ -95,6 +95,15 @@ object {{ nimbus_object }} : FeatureManifestInterface<{{ nimbus_object }}.Featur
         {%- endfor %}
     }
 
+    /**
+     * Get a list of feature ids where the feature allows co-enrollment.
+     */
+    override fun getCoenrollingFeatureIds(): List<String> = listOf(
```
eliserichards, line 101: When this is generated into `app.fml_release.kt`, `simple_nimbus_validation_release.kt`, etc, it's generating this:

```
override fun getCoenrollingFeatureIds(): List<String> = listOf()
```

Is that right? Are these files just showing the default (which should just be `listOf<string>()`), or should it be showing the actual logic (aka calling the feature manifest's `get_coenrolling_feature_ids()`) 🤔

components/support/nimbus-fml/src/frontend.rs
eliserichards: image

ahh, thank you clippy 📎

jeddai: Yep the way I think about it is if you're going to move a variable and ALSO use it afterwards, then you'll need to clone it. Otherwise you can just move it and call it a day.

components/support/nimbus-fml/src/backends/kotlin/templates/FeatureManifestTemplate.kt
jeddai: Nope if that's the default then I think that's working correctly! Unlike the FFI places in the Nimbus SDK where we do call back into the Rust, the templates won't. The FML templates may generate code that would call into _SDK_ Rust code, but they'll never generate code that would call into FML Rust code.

jeddai: Another way you can think about this is it's very similar to rendering a React/Angular component's template. The template itself has a bunch of references into component code, but you never actually see the JS that's inside the template because it all gets rendered into HTML.

eliserichards: That totally makes sense! Ty! 🙏

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -954,4 +954,15 @@ mod test {
         test_single_merged_manifest_file("fixtures/fe/importing/diamond/02-sublib.yaml", "debug")?;
         Ok(())
     }
+
+    #[test]
+    fn test_with_coenrolled_features_and_imports_kotlin() -> Result<()> {
```
eliserichards, line 959: todo: make this pass
also todo: add the same for swift 🐦

components/support/nimbus-fml/src/intermediate_representation.rs
```diff
@@ -640,6 +640,13 @@ impl FeatureManifest {
         self.iter_feature_defs().find(|f| f.name() == nm)
     }
 
+    pub fn get_coenrolling_feature_ids(&self) -> Vec<String> {
+        self.iter_all_feature_defs()
+            .filter(|(_, f)| f.allow_coenrollment())
+            .map(|(_, f)| f.name())
+            .collect()
+    }
```
jhugman, line 648: Lovely.

components/support/nimbus-fml/src/frontend.rs
```diff
@@ -120,6 +120,7 @@ pub(crate) struct ImportBlock {
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
```
jhugman, line 123: I'd probably put this further up: i.e. above `ManifestFrontend`.

components/support/nimbus-fml/src/backends/kotlin/gen_structs/mod.rs
```diff
@@ -118,6 +118,7 @@ impl<'a> FeatureManifestDeclaration<'a> {
                 "org.mozilla.experiments.nimbus.internal.FeatureManifestInterface".to_string(),
                 "org.mozilla.experiments.nimbus.FeaturesInterface".to_string(),
                 "org.json.JSONObject".to_string(),
+                "kotlin.collections.listOf".to_string(),
```
jhugman, line 121: I'm surprised this is necessary.

components/support/nimbus-fml/fixtures/fe/including/ios_search.yaml
```diff
@@ -22,6 +22,7 @@ features:
             "keep-for-days": 100
           }
     defaults: null
+    allow-coenrollment: false
```
jhugman, line 25: You can probably remove this, now you're defaulting to `false`.

components/support/nimbus-fml/fixtures/fe/including/ios_homescreen.yaml
```diff
@@ -28,6 +28,7 @@ features:
             "library-shortcuts": true
           }
         }
+    allow-coenrollment: false
```
jhugman, line 31: You can probably remove this, now you're defaulting to `false`.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/internal/FeatureManifestInterface.kt
```diff
@@ -41,4 +41,9 @@ interface FeatureManifestInterface<T> {
      * is quite useful for introspection.
      */
     fun getFeature(featureId: String): FeatureHolder<*>?
+
+    /**
+     * Get a list of feature ids where the feature allows co-enrollment.
```
jhugman, line 46: Nit: Since this is public API, I think a note saying that this isn't useful for app developers for day-to-day usage of Nimbus.

components/nimbus/ios/Nimbus/FeatureManifestInterface.swift
```diff
@@ -31,4 +31,7 @@ public protocol FeatureManifestInterface {
     ///
     /// This happens automatically if you use the `NimbusBuilder` pattern of initialization.
     func invalidateCachedValues()
+
+    /// Get a list of feature ids where the feature allows co-enrollment.
```
jhugman, line 35: Same nit as above.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+# v117.0 (In progress)
+
+## Nimbus FML ⛅️🔬🔭🔧
+
+### ✨ What's New ✨
+
+- Add `allowCoenrollment` to the FML `FeatureDef` and `FeatureBody`. This...([#5688](https://github.com/mozilla/application-services/pull/5688)).
+- Add `getCoenrollingFeatureIds` command to the FML. This command ... ([#5688](https://github.com/mozilla/application-services/pull/5688)).
```
jhugman, line 8:
```suggestion
- Add `allow-coenrollment` property for features in the Feature Manifest Language. This relaxes the feature exclusion rules for features marked with `allow-coenrollment: true`. ([#5688](https://github.com/mozilla/application-services/pull/5688)).
  - This adds a non-user-facing method to the `FeatureManifestInterface`, `getCoenrollingFeatureIds`, in both Kotlin and Swift.
```

components/support/nimbus-fml/src/workflows.rs
```diff
@@ -954,4 +954,15 @@ mod test {
...
+        generate_and_assert(
+            "test/allow_coenrolling.kts",
+            "fixtures/fe/importing/coenrolling/app.yaml",
+            "release",
+            true,
+        )?;
```
jhugman, line 965:
```suggestion
        generate_multiple_and_assert(
            "test/allow_coenrolling.kts",
            &[
                ("fixtures/fe/importing/coenrolling/app.fml.yaml", "release"),
                ("fixtures/fe/importing/coenrolling/ui.fml.yaml", "release"),
            ],
        )?;
```

components/support/nimbus-fml/fixtures/fe/importing/coenrolling/app.yaml
```diff
@@ -0,0 +1,16 @@
+version: 1.0
```
jhugman, line 1: Nit: rename this to `app.fml.yaml`.

The naming scheme I was using was that `includes` use `.yaml`, and files that have a generated file associated with them (i.e. have an `about` block) use `.fml.yaml`.

components/support/nimbus-fml/test/allow_coenrolling.kts
```diff
@@ -0,0 +1,17 @@
...
+var injected: MockNimbus? = null
+MyNimbus.initialize { injected }
+val feature = MyNimbus.features.homescreen.value()
+
+val coenrollingFeatureIds = MyNimbus.getCoenrollingFeatureIds()
+
+assert(coenrollingFeatureIds == listOf("fun-app-menu"))
+// Record exposure
+MyNimbus.features.nimbusValidation.recordExposure()
+assert(api.isExposed("nimbus-validation"))
```
jhugman, line 17: This can be as simple as:

```suggestion
import com.example.app.nimbus.AppNimbus as MyNimbus

val coenrollingFeatureIds = MyNimbus.getCoenrollingFeatureIds()

assert(coenrollingFeatureIds == listOf("fun-app-menu"))

```

Note: that the import of `AppNimbus` needs to match up with the about block in the [`app.fml.yaml` file](https://github.com/mozilla/application-services/pull/5688/files#diff-3f97979c331bb21174f195ea7cdff20170e119929aa00f23bb84075269eb0035).

megazords/cirrus/tests/python-tests/resources/test.fml.yml
```diff
@@ -28,6 +28,7 @@ features:
         value: {
            "something": "wicked"
         }
+    allow-coenrollment: false
```
jhugman, line 31: You can probably delete this line now that `allow-coenrollment: false` is the default.

jhugman (approved):
Super. There are a few nits, and you should get a swft version of the `allow_coenrolling.kts` going.

But this is really nice. Well done.

components/support/nimbus-fml/src/frontend.rs
eliserichards: It didn't change the case when it was up on `ManifestFrontend`, so it looks like it needs to stay here

components/support/nimbus-fml/src/backends/kotlin/gen_structs/mod.rs
eliserichards: it doesn't need this! I forgot to remove it 👍

PR #5692, opened 2023-06-29
Remove support for migrating from pre-app-services fxaccount implementations [firefox-android: main]

mhammond:
…tations

Over 3 years ago we implemented support for migrating from a stand-alone session token to an fxa-client account. We no longer need to support this and it will get in the way of the fxa refactor we are planning, so let's get rid of it.

(A worse-case scenario is that a 3 year old iOS device is fired up and finds itself needing to sign in to FxA)


tarikeshaq (commented):
+1 to removing the migration - this is a breaking change for iOS though so we should make sure the iOS side of the story is cleaned up as well

PR #5694, opened 2023-06-30
Update ktlint to version 0.50.0 and fix new warnings

rvandermeulen:
Changelog: https://github.com/pinterest/ktlint/releases/tag/0.50.0

Nothing super interesting in this update other than getting more strict about how suppressions are declared.


skhamis (approved):
👍 Though the only thing is we might want to wait for the cut as I think Ben wanted a somewhat code-freeze for the release

PR #5717, opened 2023-07-06
Adding Fxa functionality

bendk:
This is my attempt to add more functionality to the FxA client with the hope that this shared code can replace a lot of the current Swift/Kotlin wrapper code.  This is based on top of https://github.com/mozilla/application-services/pull/5713.

This is basically non-breaking, although we now throw exceptions in more cases than before.  This mostly happens when you try to call a method that's not valid for the current auth state (`disconnect()` when the client isn't connected, `begin_oauth_flow()` when the client is uninitialized, etc).  I think we might be able merge this now without changing the consumer code.

The next step would be to update the Swift/Kotlin wrapper code to use this functionality and delete the wrapper classes there.

The best way to understand this is probably commit-by-commit.

  - Note:

### DIFF
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -70,7 +70,30 @@ impl FirefoxAccount {
         self.internal.lock().get_current_device_id()
     }
 
+    /// Get the list of known devices registered on the user's account.
+    ///
+    /// The application might use this information to e.g. display a list of appropriate
+    /// send-tab targets.
+    ///
+    /// # Notes
+    ///
+    ///    - Device metadata is only visible to applications that have been
+    ///      granted the `https://identity.mozilla.com/apps/oldsync` scope.
+    #[handle_error(Error)]
+    pub fn get_device_list(&self) -> ApiResult<DeviceList> {
+        self.internal.lock().get_device_list(false)
+    }
+
+    /// Like `get_device_list()`, but always fetch a new device list rather than using cached data.
+    #[handle_error(Error)]
+    pub fn refresh_device_list(&self) -> ApiResult<DeviceList> {
+        self.internal.lock().get_device_list(true)
+    }
+
     /// Get the list of devices registered on the user's account.
+    ///
+    /// **Deprecated**: use [FirefoxAccount::get_device_list] or
+    /// [FirefoxAccount::refresh_device_list] instead.
     ///
     /// This method returns a list of [`Device`] structs representing all the devices
     /// currently attached to the user's account (including the current device).
@@ -184,12 +207,21 @@ impl FirefoxAccount {
     }
 }
 
+/// List of known devices registered to the user's account
+#[derive(Debug, PartialEq, Eq)]
+pub struct DeviceList {
+    /// The device using the FxA client
+    pub local_device: Device,
+    /// All other known devices
+    pub remote_devices: Vec<Device>,
+}
+
 /// A device connected to the user's account.
 ///
 /// This struct provides metadata about a device connected to the user's account.
 /// This data would typically be used to display e.g. the list of candidate devices
 /// in a "send tab" menu.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Device {
     pub id: String,
     pub display_name: String,
@@ -210,7 +242,7 @@ pub struct Device {
 /// use the variants of this enum to do so.
 ///
 /// In practice, the only currently-supported command is the ability to receive a tab.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum DeviceCapability {
     SendTab,
 }
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -22,8 +22,6 @@ use sync15::DeviceType;
 impl FirefoxAccount {
     /// Create a new device record for this application.
     ///
-    /// **💾 This method alters the persisted account state.**
-    ///
     /// This method register a device record for the application, providing basic metadata for
     /// the device along with a list of supported [Device Capabilities](DeviceCapability) for
     /// participating in the "device commands" ecosystem.
@@ -74,8 +72,6 @@ impl FirefoxAccount {
 
     /// Get the list of devices registered on the user's account.
     ///
-    /// **💾 This method alters the persisted account state.**
-    ///
     /// This method returns a list of [`Device`] structs representing all the devices
     /// currently attached to the user's account (including the current device).
     /// The application might use this information to e.g. display a list of appropriate
@@ -125,8 +121,6 @@ impl FirefoxAccount {
 
     /// Update the display name used for this application instance.
     ///
-    /// **💾 This method alters the persisted account state.**
-    ///
     /// This method modifies the name of the current application's device record, as seen by
     /// other applications and in the user's account management pages.
     ///
@@ -145,8 +139,6 @@ impl FirefoxAccount {
 
     /// Clear any custom display name used for this application instance.
     ///
-    /// **💾 This method alters the persisted account state.**
-    ///
     /// This method clears the name of the current application's device record, causing other
     /// applications or the user's account management pages to have to fill in some sort of
     /// default name when displaying this device.
@@ -161,8 +153,6 @@ impl FirefoxAccount {
     }
 
     /// Ensure that the device record has a specific set of capabilities.
-    ///
-    /// **💾 This method alters the persisted account state.**
     ///
     /// This method checks that the currently-registered device record is advertising the
     /// given set of capabilities in the FxA "device commands" ecosystem. If not, then it
--- a/components/fxa-client/src/events.rs
+++ b/components/fxa-client/src/events.rs
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::{AuthState, CallbackResult, DeviceList, FirefoxAccount, IncomingDeviceCommand};
+
+impl FirefoxAccount {
+    /// Register to receive FxaEvent notifications
+    ///
+    /// If there is EventListener already registered, it will be unregistered.  Pass in None to
+    /// remove the current listener
+    pub fn register_event_listener(&self, listener: Option<Box<dyn EventListener>>) {
+        self.internal.lock().register_event_listener(listener);
+    }
+}
+
+/// Implemented by a foreign callback interface to listen for FxA events.
+/// Use `FirefoxAccount::register_event_listener()` to receive these events.
+pub trait EventListener: Send + Sync {
+    fn on_event(&self, event: FxaEvent) -> CallbackResult<()>;
+}
+
+#[derive(Debug)]
+pub enum FxaEvent {
+    /// The account's profile was updated
+    ProfileUpdated,
+    /// Account auth state changed
+    AuthStateChanged { state: AuthState },
+    /// The account itself was destroyed
+    AccountDestroyed,
+    /// An command from another device was received
+    DeviceCommandIncoming { command: IncomingDeviceCommand },
+    /// The list of devices was changed
+    DeviceListChanged { device_list: DeviceList },
+}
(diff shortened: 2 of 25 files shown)


components/fxa-client/src/fxa_client.udl
```diff
@@ -241,13 +280,18 @@ interface FirefoxAccount {
   // user's last-seen profile information, if any. This may be useful in helping
   // the user to reconnnect to their account. If reconnecting to the same account
   // is not desired then the application should discard the persisted account state.
-  //
+  [Throws=FxaError]
   void disconnect();
```
bendk, line 284: As mentioned in the description, `disconnect()` now throws methods like `begin_oauth_flow()` have more cases where they throw because I added checks to make sure the auth state is what we expect.  Is this a good idea?  In many cases we could just log warnings and proceed.  I'm not sure what's best.

I made the opposite choice for `cancel_oauth_flow()`.  In that method, if we're not in an oath flow it's just a no-op.  I can't exactly explain why, but it seems like cancellation methods have more reason to be infallible.

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -0,0 +1,446 @@
...
+            event_listener: None,
+        }
+    }
+
+    fn initial_auth_state(persisted_state: &PersistedState, uninitialized: bool) -> AuthState {
+        if uninitialized {
+            AuthState::Uninitialized
+        } else if persisted_state.session_token.is_some() || persisted_state.refresh_token.is_some()
+        {
+            AuthState::Connected
```
bendk, line 59: As I mentioned on the call, it's not totally straightforward to know when an account is connected vs disconnected by looking at the persisted state.  But this test seems pretty reasonable to me.

components/fxa-client/src/internal/push.rs
```diff
@@ -83,6 +80,69 @@ impl FirefoxAccount {
             }
         }
     }
+
+    /// Process a push message, updating the internal state and emit FxaEvent
+    pub fn process_push_message(&mut self, payload: &str) -> Result<()> {
```
bendk, line 85: I did my best to code this correctly, but another set of eyes or two would be great.

components/fxa-client/src/fxa_client.udl
```diff
@@ -310,11 +352,27 @@ interface FirefoxAccount {
...
+  //
+  // The application might use this information to e.g. display a list of appropriate
+  // send-tab targets.
+  //
+  // # Notes
+  //
+  //    - Device metadata is only visible to applications that have been
+  //      granted the `https://identity.mozilla.com/apps/oldsync` scope.
+  [Throws=FxaError]
+  DeviceList get_device_list();
```
bendk, line 366: I decided to add this method directly to `FirefoxAccount`.  The wrapper code for both Kotlin and Swift create a new class for this.  I don't have strong feelings either way, but I think I slightly prefer having the entire API in one class.

components/fxa-client/src/events.rs
```diff
@@ -0,0 +1,35 @@
...
+    ProfileUpdated,
+    /// Account auth state changed
+    AuthStateChanged { state: AuthState },
+    /// The account itself was destroyed
+    AccountDestroyed,
+    /// An command from another device was received
+    DeviceCommandIncoming { command: IncomingDeviceCommand },
+    /// The list of devices was changed
+    DeviceListChanged { device_list: DeviceList },
+}
```
bendk, line 35: The wrappers often have several different event hierarchies: one for the main event, one for account events, one for device events, etc.  I kind of like just defining one and letting the consumer code use handle the variants they care about and ignore the ones they done.

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -28,6 +31,9 @@ pub struct StateManager {
     auth_state: AuthState,
     /// Application-supplied storage handlers
     storage_handler: Option<Box<dyn StorageHandler>>,
+    /// Application-supplied event listener.  Wrapped with an Arc since it's shared with
+    /// FirefoxAccount
+    pub event_listener: Option<Arc<Box<dyn EventListener>>>,
```
tarikeshaq, line 36: This feels a little like a leaking abstraction, what do you think about keeping the event notification at a higher level than the state manager, and keeping the state manager only responsible for writing and reading state?

components/fxa-client/src/internal/mod.rs
```diff
@@ -54,6 +54,11 @@ pub struct FirefoxAccount {
     devices_cache: Option<CachedResponse<Vec<http_client::GetDeviceResponse>>>,
     auth_circuit_breaker: AuthCircuitBreaker,
     telemetry: FxaTelemetry,
+    /// Application-supplied event listener.  Wrapped with an Arc since it's shared with StateManager.
+    ///
+    /// Arc<Box<dyn T>> is kind of silly, but I couldn't figure out how to properly move the inner
+    /// value out of a Box and into an Arc.
+    pub event_listener: Option<Arc<Box<dyn EventListener>>>,
```
tarikeshaq, line 61: related to my earlier comment: what if it's only owned by this layer and not shared with the StateManager? That might add some complexity in passing data but could simplify that event handling

tarikeshaq, line 61: nit: can we make this not `pub` and expose an API to any consumers of it?

tarikeshaq (changes requested):
Thanks a ton, Ben, there was obviously a lot of well-thought-out work here and various improvements - but because there's a lot here it was really difficult to unwrangle but I think I got to do a good first pass. Some high-level thoughts:

- The PR is fairly large and introducing quite a few new concepts - this is to say I probably missed a few things going over it and it makes it a little hard to confidently land. Going commit by commit helped though and I understand parts of this are the other PR so maybe it'll be easier to review once that one is complete!
- I really like the event handling part on a high level, consolidating the things consumers can "subscribe" to could be really valuable and I imagine cut done tons of Kotlin/Swift code
- I'm a little hesitant about the `AuthState`s part, do we have pain points in the consumers about not being able to identify what the state of the component is? (or the need to initialize a component early?)
- Moving the state persistence into a callback interface I'm curious, do we anticipate any risks? I had a few questions about what constraints should we enforce for any consumer implementing that
- I'd love to see an example of consumption of some of those changes (probably in a separate branch maybe against firefox-android? 😅) I'm curious how the changes we make here interact with the app and how much we simplify it! That would answer a lot of my questions about trying to understand the value we are creating here

Overall there are a lot of great quality changes here, I do think I'll need to do a few more passes here as we figure out how valuable some of those changes are to the consumer

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -186,6 +207,31 @@ impl StateManager {
...
+    fn save_state(&self) {
+        if let Some(storage_handler) = &self.storage_handler {
+            match state_to_json(&self.persisted_state) {
+                Ok(json) => {
+                    if let Err(e) = storage_handler.save_state(json) {
+                        error_support::report_error!(
+                            "fxaclient-persist-state",
+                            "Error saving state: {e}"
+                        );
+                    }
```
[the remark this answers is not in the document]
bendk: > How does this change error handling in persisting the state? Do we know how android and iOS react nowadays when the state persistence fails?

From what I can tell [Swift catches all errors and converts them to warnings](https://github.com/mozilla/application-services/blob/main/components/fxa-client/ios/FxAClient/PersistedFirefoxAccount.swift#L246-L258).  [Kotlin does that for JSON conversion errors, but lets other errors bubble up](https://github.com/mozilla/application-services/blob/main/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/PersistedFirefoxAccount.kt#L77-L88)

> Another side effect here we should double check is that previously, the state was only ever updated once the high-level operation was fully completed, where now it can be partially updated (possibly multiple times) before the operation is complete.

That's a very good point.  Maybe we only emit one save state call per high-level function.  I think that would mean that saving a flag if the persistent state changed that we would then check at the top-level.  I think we could probably have a standard wrapper function for all top-level methods that did that.

components/fxa-client/src/storage.rs
```diff
@@ -54,3 +58,17 @@ impl FirefoxAccount {
...
+pub trait StorageHandler: Send + Sync {
+    // This is called whenever the saved state changes.  The StorageHandler should ensure that the
+    // state is saved to disk.  The next time the FirefoxAccount is constructed, it should be
+    // through `from_json()` with this json data.
+    //
+    // **⚠️ Warning:** the serialized state may contain encryption keys and access
+    // tokens that let anyone holding them access the user's data in Firefox Sync
+    // and/or other FxA services. Applications should take care to store the resulting
+    // data in a secure fashion, as appropriate for their target platform.
+    fn save_state(&self, json: String) -> CallbackResult<()>;
```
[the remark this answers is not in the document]
bendk: +1 to documenting the constraints.

I think a FIFO queue would be essentially equivalent to last-writer wins.  Is there a case where that's not true?

components/fxa-client/src/internal/mod.rs
```diff
@@ -49,7 +49,7 @@ unsafe impl<'a> Sync for http_client::FxAClientMock<'a> {}
 // to be modified.
 pub struct FirefoxAccount {
     client: Arc<FxAClient>,
-    state: StateManager,
+    pub state: StateManager,
```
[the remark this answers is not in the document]
bendk: I considered this and am definitely open to it, but I slightly favor having this `pub`.  My end-goal is to have `internal::FirefoxAccount` be a container for `StateManager`, `FxAClient`, and maybe a couple more top-level components.  Each of those components should protect their internal state, but `internal::FirefoxAccount` can have each component as a `pub` field.

components/fxa-client/src/device.rs
```diff
@@ -70,8 +70,31 @@ impl FirefoxAccount {
...
+    #[handle_error(Error)]
+    pub fn get_device_list(&self) -> ApiResult<DeviceList> {
+        self.internal.lock().get_device_list(false)
+    }
+
+    /// Like `get_device_list()`, but always fetch a new device list rather than using cached data.
+    #[handle_error(Error)]
+    pub fn refresh_device_list(&self) -> ApiResult<DeviceList> {
+        self.internal.lock().get_device_list(true)
+    }
```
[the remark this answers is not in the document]
bendk: > keeping it as one API with the ignore_cache, or a force_fetch parameter?

My issue with the current system that it's not clear when your reading the core that `true` means to ignore the cache or use the cache.   What about separate functions, but naming the second one `get_device_list_ignore_cache()`?

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -57,6 +87,39 @@ impl StateManager {
...
+                "{operation}: client already connected"
+            ))),
+            _ => Ok(()),
+        }
+    }
+
+    pub fn check_connected(&self, operation: &str) -> Result<()> {
+        match self.auth_state {
+            AuthState::Uninitialized => Err(Error::AuthState(format!(
+                "{operation}: client not initialized"
```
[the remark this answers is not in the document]
bendk: Yup, totally agree.

components/fxa-client/src/internal/state_manager.rs
bendk: I think this goes well with your comment above...
  - `StateManager` could have methods to check if the state has changed since some time in the past -- one to check if the any persisted state has changed and one to check if the auth state has changed.
  - The top-level `FirefoxAccount` would wrap each API call with state-change checks
  - If the persisted state changed, then it would call `StorageHandler::save_state()`
  - If the auth state changed, then it would pass an `AuthStateChanged` event `EventHandler::on_event()`

bendk:
>    I'm a little hesitant about the AuthStates part, do we have pain points in the consumers about not being able to identify what the state of the component is? (or the need to initialize a component early?)

Not necessarily pain points, but both the Kotlin and Swift implement a notification system for this.  [Swift](https://github.com/mozilla/application-services/blob/main/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L7-L12) has separate notifications for each auth state.  [Kotlin](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/AccountEvent.kt#L28) has a single auth state changed notification.  I figure it's better to move the functionality into the Rust code.  They also both have a lazy-initialized FirefoxAccount which I thought could be simplified if we supported an `Uninitialized` state, although maybe that's not needed.

Note: Even if we move this into Rust code we might still decide to support the current public API, especially those Swift Notification classes.  I think the wrapper layer would be simplified because it would just be a bridge, rather than implementing the logic itself.

bendk:
This one is getting pretty stale and the plan is to approach this from a different direction, starting with the wrapper code first.  Let's close it.

PR #5723, opened 2023-07-12
Fixes DISCO-2470: Implement the foundation of a Firefox Suggest component

linabutler:
This commit adds a new Suggest component to power the [Firefox Suggest](https://support.mozilla.org/en-US/kb/firefox-suggest-faq) feature. It downloads a dataset of suggestions from Remote Settings, stores them in an SQLite database with a simple schema, and makes them available to queries.

We'll iterate on this over time: more comprehensive tests, additions to the interface, and schema changes will definitely be coming soon, and you can take a peek at some of that upcoming work in this [Jira epic](https://mozilla-hub.atlassian.net/browse/DISCO-2468)! This initial PR is meant to provide a foundation for that work, and not so much a production-quality component that we'll ship right away.

It can be reviewed commit-by-commit—that might be helpful for folks newer to Rust and Application Services, as an example of how to get a new component off the ground—or as a mega-diff. For the newer folks, some of the patterns in this new component show up in our other components, too; if you're curious, I definitely recommend checking out the `.udl` files for the others!

/cc @mozilla/disco-services

  - Note:


components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+/// suggestions to ingest at once. This is for two reasons:
+///
+/// 1. The primitives for scheduling background work vary between platforms. You
+///    might use an idle timer on Desktop, `WorkManager` on Android, or
+///    `BGTaskScheduler` on iOS.
+/// 2. Ingestion limits can change, depending on the platform and the needs of
+///    the application. A mobile device on a metered connection might want to
+///    request a small subset of the Suggest data and fetch the rest later,
+///    while a desktop on a fast link might request the entire dataset on first
+///    launch.
```
linabutler, line 43: 💭 Some other ideas for using constraints:

* Imagine a future where we have "offline" location-based suggestions. We could have a scheme that chunks these suggestion by region in Remote Settings, and constrain the ingestion to just the chunk based on the user's current (coarse) location—so that we'll only download suggestions near where the user is, instead of downloading the _entire_ dataset and using up space to store suggestions for places they might never visit.
* If the user just installed Firefox, we might want to ingest an initial set of "higher-value" suggestions that we can show right away, and ingest the rest later. This would also require changing our Remote Settings schema, as I don't think records currently know what suggestions are in their attachments—but it's a thought for making Suggest more useful to first-time users.
* If the user is on cellular, we could similarly limit the amount of work we do. Android's `WorkManager` lets you schedule a task based on the network connection type, but we might still want to fetch a smaller set of suggestions even on cellular—just not a full fetch.

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+    path: PathBuf,
+    dbs: OnceCell<SuggestStoreDbs>,
+    settings_client: remote_settings::Client,
+}
+
+/// Limits for an ingestion from Remote Settings.
+pub struct IngestLimits {
+    /// The maximum number of records to request from Remote Settings.
+    /// Each record has about 200 suggestions.
+    pub records: Option<u64>,
```
linabutler, line 54: 💭 I added this constraint mostly for smoke testing against a real Remote Settings, so that I could kick the tires on the schema without ingesting the entire 7 MB dataset every time...

But, I think "approximate number of suggestions" might be a more useful constraint than number of records, if it's valuable at all. "Number of records" (and chunks) is an implementation detail; if the consumer requests about 500 suggestions, we can do the calculation in the component that it's 3 records, without leaking that detail to consumers.

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+            dbs: OnceCell::new(),
+            settings_client,
+        })
+    }
+
+    /// Returns this provider's database connections, initializing them if
+    /// they're not already open.
+    fn dbs(&self) -> Result<&SuggestStoreDbs> {
+        self.dbs
+            .get_or_try_init(|| SuggestStoreDbs::open(&self.path))
```
linabutler, line 88: 🗒️ Putting the reader and writer behind a `OnceCell` is a little different than some of our other components, but it lets the constructor avoid doing I/O—so the store is safe to create on the main thread without blocking it.

components/suggest/src/db.rs
```diff
@@ -0,0 +1,262 @@
...
+        let interrupt_handle = Arc::new(SqlInterruptHandle::new(&conn));
+        Self {
+            conn: Mutex::new(conn),
+            interrupt_handle,
+        }
+    }
+
+    /// Fetches all suggestions that match the given keyword from the database.
+    pub fn fetch_by_keyword(&self, keyword: &str) -> Result<Vec<Suggestion>> {
+        let conn = self.conn.lock().unwrap();
```
mhammond, line 82: It's a bit of a shame each of these needs to acquire the mutex, because it makes transactions tricky - eg, the store's `ingest` method can't really use a single transaction.

One idea, which adds complexity so may not be worth it would be to have a new struct, say
```
impl <'a> SuggestDbApi<'a> {
    pub fn new(conn: &'a Connection, interrupt_handle: &Arc<SqlInterruptHandle>) -> Result<Self> {
        let interrupt_scope = interrupt_handle.begin_interrupt_scope()?;
        Ok(Self { conn, interrupt_scope })
    }

    pub fn err_if_interrupted(&self) -> Result<()> {
        Ok(self.interrupt_scope.err_if_interrupted()?)
    }
...
```

with all these methods defined on it. A "reader" might then look like:

```
        let db = self.dbs()?.reader.conn.lock().unwrap();
        let reader = SuggestDbApi::new(&db, &self.dbs()?.reader.interrupt_handle)?;
```

but a writer could do:
```
        let mut db = self.dbs()?.writer.conn.lock().unwrap();
        let tx = db.transaction()?;
        let writer = SuggestDbApi::new(&tx, &self.dbs()?.writer.interrupt_handle)?;
```

and that way, `ingest()` gets to do everything in a single transaction.

Again, complexity might not be worth it, and introducing lifetimes might make people new to rust hate it, but it's just a thought I had while reading this.

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+                    id: record_id,
+                    last_modified,
+                    attachment,
+                }) => {
+                    // Drop any suggestions that we previously ingested from
+                    // this record's attachment. Suggestions don't have a
+                    // stable identifier, and determining which suggestions in
+                    // the attachment actually changed is more complicated than
+                    // dropping and re-ingesting all of them.
+                    writer.drop_suggestions(record_id)?;
```
mhammond, line 161: to follow up on my comment above, not being able to do `.drop_suggestions()` in a transaction seems a shame from a perf POV?

Also, you hold the writer mutex while doing http, which is probably OK from a correctness POV, but depending on the foreign bindings implementations, does given the lock might be held for many seconds, does it make sense to "queue" multiple calls to ingest, or might it be worth doing `try_lock()` and failing if an operation is in progress?

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+        before_each();
+
+        // Ensure that `SuggestStore` is usable with UniFFI, which requires
+        // exposed interfaces to be `Send` and `Sync`.
+        fn is_send_sync<T: Send + Sync>() {}
+        is_send_sync::<SuggestStore>();
+    }
+
+    #[test]
+    fn ingest() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
mhammond: A followup is fine IMO, especially if we assume you'd like others on the team to help contribute to them. Your team can then make their own decision about when the coverage is "good enough"

components/suggest/src/suggest.udl
```diff
@@ -0,0 +1,55 @@
...
+namespace suggest {
+
+};
+
+[Error]
+interface SuggestApiError {
+    Other(string reason);
+};
+
+dictionary IngestLimits {
```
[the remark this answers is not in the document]
mhammond: Still mildly salty that swift will force us to namespace things which uniffi tries to already have in namespaces, but that's not this PRs problem, so +1 :)

components/suggest/src/db.rs
```diff
@@ -0,0 +1,262 @@
...
+use crate::{
+    keyword::full_keyword, schema::SuggestConnectionInitializer, RemoteRecordId, RemoteSuggestion,
+    Result, Suggestion,
+};
+
+/// The metadata key whose value is the timestamp of the last ingested Remote
+/// Settings record.
+pub const LAST_FETCH_META_KEY: &str = "last_fetch";
+
+/// A list of IAB categories that contain non-sponsored suggestions.
```
mhammond, line 28: I wonder if `IAB` is worth expanding, and/or adding a link to something which might explain this sorcery?

mhammond (approved):
This is fantastic! While some further tests would be nice, I see no reason they can't be done in a followup - indeed, there's literally nothing here which I think must be fixed before landing even if some tweaks are desired before releasing - so I think you should land this whenever you like :)

components/suggest/src/store.rs
mhammond: > Also, you hold the writer mutex while doing http,

Doh - of course you don't - sorry about that :) But the point remains. I do like the db abstractions here!

components/suggest/src/db.rs
linabutler: Definitely! I copied this constant from Desktop, but the comment is unhelpful in explaining what an "IAB category" is (Internet Advertising Bureau taxonomy for web pages), or where it comes from (Remote Settings, where it's a property of each suggestion).

components/suggest/src/db.rs
```diff
@@ -0,0 +1,262 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+use std::{
+    path::Path,
+    sync::{Arc, Mutex},
```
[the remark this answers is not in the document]
linabutler: Oh, nice!

components/suggest/src/store.rs
tarikeshaq: +1 to hiding the details of what a `record` is - I like the idea of constraints in that it helps for Mobile vs Desktop network connection, but I wonder how the consumer will make the decision of the constraint.. is the plan to run some smoke tests with real devices on throttled networks to see how long the network takes? Or do we plan to have dynamic configuration of limits based on a user's network 👀 

Regardless I feel like this is a followup but I'm interested in the methodology to pick a limit as I'm sure we can learn from that in other places (Nimbus, other consumers of Remote Settings, Sync?)

components/suggest/src/keyword.rs
```diff
@@ -0,0 +1,102 @@
...
+        .map(|phrase_words| phrase_words[..min_phrase_words_len].join(" "))
+        .unwrap_or_else(|| {
+            keywords
+                .iter()
+                .map(AsRef::as_ref)
+                .filter(|phrase| phrase.starts_with(query) && query.len() < phrase.len())
+                .max_by_key(|phrase| phrase.trim().len())
+                .unwrap_or(query)
+                .to_owned()
+        })
```
tarikeshaq, line 38: I love the iterator use here 🤩 🚀 🎉

components/suggest/src/keyword.rs
```diff
@@ -0,0 +1,102 @@
...
+        // word, so that the completed phrase can show a word after the space.
+        query_words_len + 1
+    } else {
+        query_words_len
+    };
+    keywords
+        .iter()
+        .map(AsRef::as_ref)
+        .filter(|phrase| phrase.starts_with(query))
+        .map(|phrase| phrase.split_whitespace().collect::<Vec<_>>())
```
tarikeshaq, line 27: what do you think about using `filter_map` with `then_some` here and merge those three? I'm conflicted whether that would make it easier or harder to read 🤔 Something like: (although my indentation is off)
```suggestion
        .filter_map(|phrase| phrase
                              .as_ref()
                              .starts_with(query)
                              .then_some(
                                 phrase
                                   .as_ref()
                                   .split_white_space()
                                   .collect::<Vec<_>>()
                               )
                 )
```

tarikeshaq (commented):
Thanks @linabutler 🚀 🚀 This looks great!!

Had mostly questions as I was going through - didn't dig too much into the business logic 😛 This looks great to me and I'm super excited for another cross-platform component 🤩

components/suggest/src/lib.rs
```diff
@@ -0,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+use remote_settings::RemoteSettingsConfig;
+use serde_derive::*;
```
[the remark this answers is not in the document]
linabutler: That sounds great to me! Digging into it a bit more, it looks like `serde_derive` is still around for compatibility, but all the examples in Serde's docs use `serde`'s `derive` feature, so...let's give it a spin? 🤷🏼‍♀️

components/suggest/src/store.rs
linabutler: > Or do we plan to have dynamic configuration of limits based on a user's network 👀

Yep, I was thinking something closer to this. I'd love to explore generalizing it to our other components, too, but definitely in the far future!

My half-baked idea is, a consumer could use platform APIs to query whether the device is on a metered connection (and/or has Data Saver or Low Data Mode on), or to check its battery level, and then pick a set of 👋🏼 arbitrary but sensible 👋🏼 limits based on that.

Taking that even further, we could make the limits adaptive: see how long it takes to download some data, and figure out what to prioritize and how much to download based on that.

I'm not sure it's super helpful for where we are right now, since Suggest is a ~7 MB, infrequently changing dataset (1-2x/month; when it does change, it's only a few chunks, not the entire dataset) that's chunked arbitrarily now.

But I could see us wanting to do this if we ever want to support novel suggestion types, like location-based suggestions, in the future—where mirroring the entire dataset might not work well.

How does that line up with what you're thinking? 😊

components/suggest/src/suggest.udl
```diff
@@ -0,0 +1,55 @@
...
+    string full_keyword;
+    string title;
+    string url;
+    sequence<u8>? icon;
+    string? impression_url;
+    string? click_url;
+};
+
+dictionary SuggestionQuery {
+    string keyword = "";
```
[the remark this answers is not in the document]
linabutler: That sounds good to me, I think making it optional was an oversight I carried over from my experimenting with a builder API! Not all suggestions will have keywords in the future, but that'll be a breaking change, anyway, so let's make it required until then.

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+        }
+        Ok(self
+            .dbs()?
+            .reader
+            .fetch_by_keyword(&query.keyword)?
+            .into_iter()
+            .filter(|suggestion| {
+                (suggestion.is_sponsored && query.include_sponsored)
+                    || (!suggestion.is_sponsored && query.include_non_sponsored)
+            })
```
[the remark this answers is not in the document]
linabutler: For sure!

The short answer: I didn't want to complicate `fetch_by_keyword` too much at this stage by doing the equivalent of the `!NONSPONSORED_IAB_CATEGORIES.contains(&iab_category)` check in SQL.

The longer answer: We'll almost certainly be making a big schema change that'll affect `fetch_by_keyword` and `is_sponsored`, and it's not clear that we'll actually want to filter suggestions this way.

Right now, we're storing everything in one big `suggestions` table, but I think we'll want to partition the suggestions into separate tables for each "provider": adM, Wikipedia, and soon MDN, AMO, and Pocket. adM and Wikipedia suggestions currently use the same format, even though a lot of the adM fields aren't relevant for Wikipedia, and MDN / AMO / Pocket suggestions will use a different format entirely.

Since we'll be changing the schema and the query, anyway, I went with the quick approach here 😅

components/suggest/src/db.rs
ncloudioj: Is the mutex here used to serialize all transactions on the *same* sqlite connection? If so, I wonder if we can use a `Transaction` helper and move all those read, write, drop functions to it. Something like:

```rs
// For ingestion
let mut txn = self.db.txn(); // This will acquire a lock and begin a transaction off `self.db.conn`.
txn.clear();
txn.insert_suggestions();
txn.insert_icons();
txn.commit(); // This will commit using the underlying sqlite transaction.

// For queries
let mut txn = self.db.txn(); // Readonly ops will also require a transaction.
txn.fetch_by_keyword();
// a txn could be aborted implicitly when it goes out of the scope.
``` 

Just a thought of mine, no intent to interrupt the original conversation. :)

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+            .collect())
+    }
+
+    /// Interrupts any ongoing queries. This should be called when the
+    /// user types a new keyword into the address bar, to ensure that they
+    /// see fresh suggestions as they type.
+    pub fn interrupt(&self) {
+        if let Some(dbs) = self.dbs.get() {
+            // Only interrupt if the databases are already open.
+            dbs.reader.interrupt_handle.interrupt();
```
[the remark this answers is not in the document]
linabutler: Ooo, good question! We intentionally don't want to interrupt the writer connection at the same time as the reader, because typing a new query into the address bar shouldn't interrupt an in-progress ingest 😊

I can think of a couple of cases where might want to interrupt ingestion separately, though: app shutdown, and when the scheduler tells us to stop our background work. But I think you're totally right that it would be more valuable to interrupt the network I/O in that case, instead of just the connection.

Still, I'm wondering if there's a better way to clarify that. We don't surface the reader / writer split in the `SuggestStore` API, but maybe we could? Something like:

```webidl
interface SuggestStoreReader {
    sequence<Suggestion> query();
    void interrupt();
};

interface SuggestStoreWriter {
    void ingest();
    // `interrupt()` could go here eventually, too, to interrupt the writer...
};

interface SuggestStore {
    SuggestStoreReader reader();
    SuggestStoreWriter writer();
};
```

WDYT?

components/suggest/src/db.rs
mhammond: > Is the mutex here used to serialize all transactions on the same sqlite connection?

The mutex is actually used to gain access to the connection - the mutex "wraps" it, as opposed to something you acquire "next to" the connection. Once you've acquired the mutex you have exclusive access to it, so transactions and commands you execute are serialized by the nature of your code.

So your suggestion is roughly what I was trying to convey - it's just that mine used many more words due to Rust's general verbosity, lifetime requirements and lack of duck-typing. I hope that if it ends up working out for Lina, she'll be able to find a way to trim some of the noise.

components/suggest/src/store.rs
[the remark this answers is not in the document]
tarikeshaq: This makes total sense - thanks!

components/suggest/src/store.rs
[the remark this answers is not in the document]
tarikeshaq: > Still, I'm wondering if there's a better way to clarify that. We don't surface the reader / writer split in the SuggestStore API, but maybe we could? Something like:

I like it because it simplifies how we handle writers/readers a little internally (and this is I think what we do for places too?)

But at the same time, it leaks the abstraction of the DB connections out to the consumer who may not know what to do with them and may misuse them  - We see this confusion places too! especially in iOS where the consumer ended up creating exactly one reader and exactly one writer, each synchronized using multiple layers of DispatchQueues beating the purpose of the internal SQLite Reader/writer connection invariants we enforce)

This is all to say I'm conflicted but leaning a little towards keeping the UDL interface simple like you have it currently even if it means a bit more complexity internally, and we can maybe iterate on it if we find issues with the interrupts (or otherwise misuse of the API).

components/suggest/src/store.rs
```diff
@@ -0,0 +1,421 @@
...
+    reader: SuggestDb,
+}
+
+impl SuggestStoreDbs {
+    fn open(path: &Path) -> Result<Self> {
+        // Order is important here: the writer must be opened first, so that it
+        // can set up the database and run any migrations.
+        let writer = SuggestDb::open(path, ConnectionType::ReadWrite)?;
+        let reader = SuggestDb::open(path, ConnectionType::ReadOnly)?;
+        Ok(Self { writer, reader })
```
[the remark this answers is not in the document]
linabutler: I like that idea! But also agree that the complexity is probably not worth it. I do like that opening the write connection first helps guarantee that the database is in a good state before we try to read from it, and avoids a class of errors like what happens if opening one succeeds, but the other fails later on.

components/suggest/src/db.rs
linabutler: @tiftran and I paired on this today, and I think we've found an elegant way to do this that combines Mark's and Nan's suggestions! I'm kind of excited about it 😁 https://github.com/mozilla/application-services/pull/5723/commits/7ddf41de4369802e7525518b97503cf26ee98216

We moved all the SQL for reading and writing into a new `SuggestDao` type, pretty much in line with Mark's sketch. This made the other call sites a little verbose at first, because we had to lock the `conn` (sometimes doing it in a nested scope, to avoid holding the lock for too long), optionally start a transaction, create a `SuggestDao`, do the work we wanted, and finally commit the transaction.

We reduced that boilerplate by adding `read` and `write` methods to `SuggestDb`, that do all the locking and transaction management for us. These methods take a closure that receives the `SuggestDao`, and return whatever the closure returns. Swift likes to use that pattern a lot, and I think it works well here, too 😊

As a nice bonus, `SuggestDao` methods that write to the database can take a `&mut self`, so we can have compile-time checking for cases like accidentally calling a write method from a `read` closure. Maybe that's going a little too far? But I like that Rust lets us model that...

How does that look?

components/suggest/src/store.rs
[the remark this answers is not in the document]
linabutler: Perfect, let's keep this as it is now, then!

> We see this confusion places too! especially in iOS where the consumer ended up creating exactly one reader and exactly one writer, each synchronized using multiple layers of DispatchQueues beating the purpose of the internal SQLite Reader/writer connection invariants we enforce

Oooh, thanks for that context! Places actually inspired my sketch (except `reader()` would return a reference to the one reader connection, instead of opening a new one), so it's very helpful—if a little unfortunate—to hear that surfacing those concepts to consumers can be confusing.

components/suggest/src/keyword.rs
linabutler: Oh, thanks for reminding me about `then` and `then_some`! I wrote it with `filter_map` at first, but with the `if` check written out. I think the extra level of indentation and nested closure still make it a little harder to read, though:

```rust
.filter_map(|phrase| {
    phrase
        .as_ref()
        .starts_with(query)
        .then(|| phrase.as_ref().split_whitespace().collect::<Vec<_>>())
})
```

(I used `then` instead of `then_some` to avoid eagerly evaluating `split_whitespace`).

Would you be OK with keeping it as-is? There's a good chance we'll end up dropping this function in favor of using `full_keywords` from the Remote Settings data, so this might end up going away.

I did love getting to use Rust's iterators for these, and was impressed with how well they were able to follow the [English description](https://searchfox.org/mozilla-central/rev/c615dc4db129ece5cce6c96eb8cab8c5a3e26ac3/browser/components/urlbar/private/AdmWikipedia.sys.mjs#209-231)!

components/suggest/src/store.rs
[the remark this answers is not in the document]
linabutler: Thanks, I found it really helpful to write this when sketching out the architecture, too!

Also, based on our other conversation, this might be a good place to mention `interrupt` 😁👍🏼

components/suggest/src/schema.rs
```diff
@@ -0,0 +1,76 @@
...
+        PRIMARY KEY (keyword, suggestion_id)
+    ) WITHOUT ROWID;
+
+    CREATE UNIQUE INDEX keywords_suggestion_id_rank ON keywords(suggestion_id, rank);
+
+    CREATE TABLE suggestions(
+        id INTEGER PRIMARY KEY,
+        record_id TEXT NOT NULL,
+        block_id INTEGER NOT NULL,
+        advertiser TEXT NOT NULL,
```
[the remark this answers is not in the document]
linabutler: Good idea!

Going a little further—since we're not actually querying by `block_id`, `impression_url`, `click_url`, or `iab_category` (or, for that matter, by any of the other columns in `suggestions`, except for `id`, `record_id`, and maybe `provider`/`advertiser`)...we could make `custom_details` be a regular `TEXT` column in the `suggestions` table, and just store a big ol' JSON blob with all those fields, instead of unpacking them into a separate table!

(If we really wanted to query on those fields, SQLite also supports [JSON natively now](https://www.sqlite.org/json1.html), so that would be an option. It wouldn't be as efficient as normalizing them, and it might increase the database size a bit—so we may want to compare using JSON vs. separate tables for the custom details).

@tiftran, maybe something to think about when you work on this (and thanks for taking it on! 🚀)

components/suggest/src/db.rs
mhammond: That looks great! I feel silly for not thinking of using a closure there :) Being able to take a `&mut self` could certainly come in handy in the future and the fact we are able to use transactions "correctly" and not use our [`unchecked_transaction`](https://github.com/mozilla/application-services/blob/aa3a34b6cb47f63e7db9ef76a4c7e2b8c25ffb68/components/support/sql/src/conn_ext.rs#L271-L278) hacks is already a big win.

ncloudioj: +1. Assuming the transaction will get aborted automatically when the closure returns an error, this looks pretty neat.

components/suggest/src/schema.rs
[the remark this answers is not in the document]
linabutler: I filed https://mozilla-hub.atlassian.net/browse/DISCO-2537 for this work.

components/suggest/src/db.rs
linabutler: Rad, thank you both! I feel like we've found new patterns for storage as we've learned what works and what doesn't—but we're definitely converging on a design 😁

I'm also happy that this lets us avoid the `unchecked_transaction` escape hatch!

> Assuming the transaction will get aborted automatically when the closure returns an error

It will!

components/suggest/src/db.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn with_connection(conn: Connection) -> Self {
+        let interrupt_handle = Arc::new(SqlInterruptHandle::new(&conn));
+        Self {
+            conn: Mutex::new(conn),
+            interrupt_handle,
+        }
+    }
+
+    /// Fetches all suggestions that match the given keyword from the database.
+    pub fn fetch_by_keyword(&self, keyword: &str) -> Result<Vec<Suggestion>> {
```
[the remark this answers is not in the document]
linabutler: Great point! I filed https://mozilla-hub.atlassian.net/browse/DISCO-2538; it's a "good first bug" for working with UDL, Rust, and SQL.

ncloudioj (approved):
Thanks, Lina! This is cool stuff. 🚢

components/suggest/src/schema.rs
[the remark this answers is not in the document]
ncloudioj: Yup, those are dumb fields that we'd never need a secondary index on any of them. We can pack them up into a single text blob for now.

components/suggest/src/store.rs
tarikeshaq: Didn't realize I didn't get back to this 😅
> My half-baked idea is, a consumer could use platform APIs to query whether the device is on a metered connection (and/or has Data Saver or Low Data Mode on), or to check its battery level, and then pick a set of 👋🏼 arbitrary but sensible 👋🏼 limits based on that.

 I like it! And seems like it would serve our use case here of making sure the user gets a good experience even if they are on a low data network connection - once we ship and see how well our configuration story is in practice I'd love for us to visit if there's value in the other components, but like you said, that's in the far future!

mhammond: > My half-baked idea is, a consumer could use platform APIs to query whether the device is on a metered connection (and/or has Data Saver or Low Data Mode on), or to check its battery level, and then pick a set of 👋🏼 arbitrary but sensible 👋🏼 limits based on that.

I guess one vague concern about this is that "arbitrary but sensible" is really a contradiction. Picking something sensible probably means the app needs to know more about the implementation than really makes sense. I'm wondering if it's better for the app to communicate what it does know for sure (ie, it's in a kind of data-saver mode) which the component uses to choose appropriate (but not arbitrary ;) limits.

PR #5743, opened 2023-07-26
Fixes DISCO-2493: Add tests for the Suggest component

linabutler:
Some refactoring to make testing easier, and then adding some tests on top of that! It's a little easier to review this commit-by-commit, especially after the first commit that moved the Remote Settings types into their own `rs.rs` (😅) module.

Marking this as a "draft" for now, because there are a few more things (deleting records, records with unknown types, constraining the number of records, clearing) I'd like to cover before landing—but how does the overall shape of this look?

/cc @mtbottle, I can't tag you as a reviewer, but please feel free to review if you'd like!

  - Note:

### DIFF
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -1,0 +1,126 @@
+use std::ops::Deref;
+
+use remote_settings::Attachment;
+use serde::Deserialize;
+
+/// The Suggest Remote Settings collection name.
+pub(crate) const REMOTE_SETTINGS_COLLECTION: &str = "quicksuggest";
+
+/// The maximum number of suggestions in a Suggest record's attachment.
+///
+/// This should be the same as the `BUCKET_SIZE` constant in the
+/// `mozilla-services/quicksuggest-rs` repo.
+pub(crate) const SUGGESTIONS_PER_ATTACHMENT: u64 = 200;
+
+/// The response body for a Suggest Remote Settings collection request.
+#[derive(Debug, Deserialize)]
+pub(crate) struct SuggestRemoteSettingsResponse {
+    pub data: Vec<SuggestRecord>,
+}
+
+/// A record with a known or an unknown type, or a tombstone, in the Suggest
+/// Remote Settings collection.
+///
+/// Because `#[serde(other)]` doesn't support associated data
+/// (serde-rs/serde#1973), we can't define variants for all the known types and
+/// the unknown type in the same enum. Instead, we have this "outer", untagged
+/// `SuggestRecord` with the "unknown type" variant, and an "inner", internally
+/// tagged `TypedSuggestRecord` with all the "known type" variants.
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub(crate) enum SuggestRecord {
+    /// A record with a known type.
+    Typed(TypedSuggestRecord),
+
+    /// A tombstone, or a record with an unknown type, that we don't know how
+    /// to ingest.
+    ///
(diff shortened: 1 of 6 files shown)


components/suggest/src/store.rs
```diff
@@ -71,31 +100,22 @@ pub struct SuggestIngestionConstraints {
...
-    pub fn new(
-        path: &str,
-        settings_config: Option<RemoteSettingsConfig>,
-    ) -> SuggestApiResult<Self> {
-        Ok(Self::new_inner(path, settings_config)?)
-    }
+/// The implementation of the store. This is generic over the Remote Settings
+/// client, and is split out from the concrete [`SuggestStore`] for testing
+/// with a mock client.
+pub(crate) struct GenericSuggestStore<S> {
```
ncloudioj, line 106: Do we want to bound it?
```suggestion
pub(crate) struct GenericSuggestStore<S: SuggestRemoteSettingsClient> {
```

components/suggest/src/store.rs
```diff
@@ -460,13 +409,469 @@ mod tests {
...
+        );
+        store.ingest(SuggestIngestionConstraints::default())?;
+
+        assert_eq!(
+            store.query(SuggestionQuery {
+                keyword: "la".into(),
+                include_sponsored: true,
+                ..Default::default()
+            })?,
+            &[Suggestion {
```
ncloudioj, line 509: Looks like we have to write lots of those fixture records in the assertions throughout the file, shall we consider using a snapshot test crate (e.g. [insta](https://github.com/mitsuhiko/insta)) to ease that?

ncloudioj (approved):
It looks good, nice refactors!

components/suggest/src/store.rs
linabutler: Good question about adding a bound to the type declaration! I think the Rust convention leans away from doing that (rust-lang/rust-clippy#1689), and prefers adding the bound to just the `impl` blocks that need it. I don't feel strongly about it, though, and would be happy to add the bound if you'd like—the issues with bounds on the type mentioned in that Clippy ticket aren't as relevant here.

You're right about `Generic` being redundant; I had a hard time thinking of a good name! Let's do `SuggestStoreInner` 😁

components/suggest/src/rs.rs
ncloudioj, line 1: Shall we call it `o.rs`? Then we will get the compiled obj `o.o`

linabutler: 😂😂😂 I _definitely_ didn't call it `rs.rs` intentionally... 😉

components/suggest/src/store.rs
linabutler: Yep, that Clippy issue is still open; there's no lint for it or anything now! 👀 But it did help me understand why trait bounds aren't as commonly used, and where they're needed (the `Cow` 🐮 example, where we need the bound to express the `Owned` variant).

linabutler:
I iterated a bit more on the tests, and came up with an approach that I'm a lot happier with!

There's a `Snapshot` helper with convenience constructors that take JSON for the fake records and attachments, and deserialize them into the right data structures. The `SnapshotSettingsClient` just wraps the snapshot, and a nice advantage of making `SuggestStoreInner` generic (vs. using a trait object, from my very first iteration) is that it's super easy to swap out the snapshot.

I'd like to check out https://docs.rs/expect-test/latest/expect_test/ next, and add some more coverage for the store.

components/suggest/src/store.rs
linabutler: I added `expect-test` to get us started with snapshot testing, and I'm such a fan! Thanks for the pointer!

PR #5745, opened 2023-07-26
[DISCO-2537] Suggest db schema change to accomodate 'custom details'

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -16,7 +16,7 @@ use sql_support::{open_database::open_database_with_flags, ConnExt};
 
 use crate::{
     keyword::full_keyword,
-    rs::{DownloadedSuggestion, SuggestRecordId},
+    rs::{DownloadedSuggestion, SuggestRecordId, SuggestionProvider},
     schema::SuggestConnectionInitializer,
     Result, Suggestion,
 };
@@ -122,41 +122,78 @@ impl<'a> SuggestDao<'a> {
     /// Fetches suggestions that match the given keyword from the database.
     pub fn fetch_by_keyword(&self, keyword: &str) -> Result<Vec<Suggestion>> {
         self.conn.query_rows_and_then_cached(
-            "SELECT s.id, k.rank, s.block_id, s.advertiser, s.iab_category,
-                    s.title, s.url, s.impression_url, s.click_url,
-                    (SELECT i.data FROM icons i WHERE i.id = s.icon_id) AS icon
-             FROM suggestions s
-             JOIN keywords k ON k.suggestion_id = s.id
-             WHERE k.keyword = :keyword
-             LIMIT 1",
+            "SELECT s.id, k.rank, s.title, s.url, s.provider,
+                        (SELECT i.data FROM icons i WHERE i.id = s.icon_id) AS icon
+                  FROM suggestions s
+                  JOIN keywords k ON k.suggestion_id = s.id
+                  WHERE k.keyword = :keyword
+                  LIMIT 1",
             named_params! {
                 ":keyword": keyword,
             },
-            |row| -> Result<Suggestion> {
-                let keywords: Vec<String> = self.conn.query_rows_and_then(
+            |row| -> Result<Suggestion>{
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let url = row.get("url")?;
+                let icon = row.get("icon")?;
+                let provider = row.get("provider")?;
+
+                let keywords: Vec<String> = self.conn.query_rows_and_then_cached(
                     "SELECT keyword FROM keywords
                      WHERE suggestion_id = :suggestion_id AND rank >= :rank
                      ORDER BY rank ASC",
                     named_params! {
-                        ":suggestion_id": row.get::<_, i64>("id")?,
+                        ":suggestion_id": suggestion_id,
                         ":rank": row.get::<_, i64>("rank")?,
                     },
                     |row| row.get(0),
                 )?;
-                let iab_category = row.get::<_, String>("iab_category")?;
-                let is_sponsored = !NONSPONSORED_IAB_CATEGORIES.contains(&iab_category.as_str());
-                Ok(Suggestion {
-                    block_id: row.get("block_id")?,
-                    advertiser: row.get("advertiser")?,
-                    iab_category,
-                    is_sponsored,
-                    title: row.get("title")?,
-                    url: row.get("url")?,
-                    full_keyword: full_keyword(keyword, &keywords),
-                    icon: row.get("icon")?,
-                    impression_url: row.get("impression_url")?,
-                    click_url: row.get("click_url")?,
-                })
+
+                match provider {
+                    SuggestionProvider::Amp => {
+                        self.conn.query_row_and_then(
+                            "SELECT amp.advertiser, amp.block_id, amp.iab_category, amp.impression_url, amp.click_url
+                            FROM amp_custom_details amp WHERE amp.suggestion_id = :suggestion_id", 
+                            named_params! {
+                                ":suggestion_id": suggestion_id
+                            },
+                            |row|{
+                                let iab_category = row.get::<_, String>("iab_category")?;
+                                let is_sponsored = !NONSPONSORED_IAB_CATEGORIES.contains(&iab_category.as_str());
+                                Ok(Suggestion {
+                                    block_id: row.get("block_id")?,
+                                    advertiser: row.get("advertiser")?,
+                                    iab_category,
+                                    is_sponsored,
+                                    title,
+                                    url,
+                                    full_keyword: full_keyword(keyword, &keywords),
+                                    icon,
+                                    impression_url: row.get("impression_url")?,
+                                    click_url: row.get("click_url")?,
+                                    provider
+                                })
+                            }
+                        )
+                    },
+                    SuggestionProvider::Wikipedia =>{
+                        Ok(Suggestion {
+                            block_id: 0,
+                            advertiser: "Wikipedia".to_string(),
+                            iab_category: "5 - Education".to_string(),
+                            is_sponsored: false,
+                            title,
+                            url,
+                            full_keyword: full_keyword(keyword, &keywords),
+                            icon,
+                            impression_url: None,
+                            click_url: None,
+                            provider
+                        })
+                    }
+                }
+
+
             },
         )
     }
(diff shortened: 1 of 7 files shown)


components/suggest/src/schema.rs
```diff
@@ -26,14 +26,21 @@ pub const SQL: &str = "
...
+    CREATE TABLE adm_custom_details(
+        suggestion_id INTEGER,
+        block_id INTEGER,
+        advertiser TEXT NOT NULL,
+        iab_category TEXT NOT NULL,
         icon_id TEXT NOT NULL,
         impression_url TEXT,
-        click_url TEXT
+        click_url TEXT,
+        FOREIGN KEY(suggestion_id) REFERENCES suggestions(id)
```
mhammond, line 43: I can't remember how suggestions are pruned and I'm too lazy to look, but I wonder if an "on cascade delete" should happen?

components/suggest/src/rs.rs
```diff
@@ -146,18 +147,91 @@ impl SuggestRecordId {
...
+            #[serde(rename = "Wikipedia")]
+            Wikipedia(DownloadedBaseSuggestion),
+        }
+
+        Ok(
+            match UntaggedDownloadedSuggestion::deserialize(deserializer)? {
+                UntaggedDownloadedSuggestion::KnownTag(TaggedDownloadedSuggestion::Wikipedia(
+                    base,
+                )) => Self::Wikipedia(base),
+                UntaggedDownloadedSuggestion::UnknownTag(base) => Self::Adm(base),
```
linabutler, line 233:
```suggestion
                UntaggedDownloadedSuggestion::UnknownTag(adm) => Self::Adm(adm),
```

components/suggest/src/schema.rs
```diff
@@ -26,14 +26,21 @@ pub const SQL: &str = "
...
-        iab_category TEXT NOT NULL,
+        provider TEXT NOT NULL,
         title TEXT NOT NULL,
         url TEXT NOT NULL,
-        icon_id TEXT NOT NULL,
+        icon_id TEXT NOT NULL
+    );
+
+    CREATE TABLE adm_custom_details(
+        suggestion_id INTEGER,
```
linabutler, line 36:
```suggestion
        suggestion_id INTEGER PRIMARY KEY,
```

components/suggest/src/schema.rs
```diff
@@ -26,14 +26,21 @@ pub const SQL: &str = "
...
+        advertiser TEXT NOT NULL,
+        block_id INTEGER,
```
linabutler, line 38:
```suggestion
        block_id INTEGER NOT NULL,
```

components/suggest/src/schema.rs
```diff
@@ -26,14 +26,21 @@ pub const SQL: &str = "
...
+        iab_category TEXT NOT NULL,
         impression_url TEXT,
```
linabutler, line 40:
```suggestion
        impression_url TEXT NOT NULL,
```

components/suggest/src/schema.rs
```diff
@@ -26,14 +26,21 @@ pub const SQL: &str = "
...
-        click_url TEXT
+        click_url TEXT,
```
linabutler, line 41:
```suggestion
        click_url TEXT NOT NULL,
```

components/suggest/src/schema.rs
linabutler, line 1: suggestion (blocking): Important—please bump the `VERSION` constant to `2`.

It's OK not to have a migration now, but we want to bump the version any time we change the schema.

linabutler (dismissed):
Woohooo! 🚀 This is looking great, @tiftran! I left a lot of small nits and suggestions for follow-up, but the only blocking thing is bumping the `VERSION`. I wouldn't mind seeing your patch one more time, then we can get it landed.

components/suggest/src/db.rs
```diff
@@ -132,31 +133,68 @@ impl<'a> SuggestDao<'a> {
...
+                            is_sponsored: false,
+                            title,
+                            url,
+                            full_keyword: full_keyword(keyword, &keywords),
+                            icon,
+                            impression_url: None,
+                            click_url: None,
+                        })
+                    },
+                    &_ => Err(Error::Sql(rusqlite::Error::ExecuteReturnedResults))
```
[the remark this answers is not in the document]
tiftran: ahhh yesss, i forgot mention that this was something i would like your opinion on

components/suggest/src/db.rs
```diff
@@ -132,31 +133,68 @@ impl<'a> SuggestDao<'a> {
...
-                    is_sponsored,
-                    title: row.get("title")?,
-                    url: row.get("url")?,
-                    full_keyword: full_keyword(keyword, &keywords),
-                    icon: row.get("icon")?,
-                    impression_url: row.get("impression_url")?,
-                    click_url: row.get("click_url")?,
-                })
+
+                match provider.as_str() {
```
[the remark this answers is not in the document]
tiftran: Soooo i don't think I have it as intended...

components/suggest/src/db.rs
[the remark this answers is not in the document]
tiftran: considering I wasn't able to handle the match arm...I don't think I introduced `ProviderType` correctly

components/suggest/src/db.rs
```diff
@@ -132,31 +132,70 @@ impl<'a> SuggestDao<'a> {
...
                 ":keyword": keyword,
             },
-            |row| -> Result<Suggestion> {
-                let keywords: Vec<String> = self.conn.query_rows_and_then(
+            |row| -> Result<Suggestion>{
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let url = row.get("url")?;
+                let icon = row.get("icon")?;
+                let provider = ProviderType::from_u8(row.get::<_, u8>("provider")?);
```
linabutler, line 140:
```suggestion
                let provider = row.get("provider")?);
```

Here's what was causing your problem—a one-liner that makes all the difference! 😅 Instead of writing out the `ProviderType::from_u8` conversion yourself, you can write `row.get::<_, ProviderType>("provider")`, which will call your new `FromSql::column_result` conversion!

Bonus: you don't need the explicit `::<_, ProviderType>` annotation, either! If you change the match arms to just `ProviderType::Adm` and `ProviderType::Wikipedia`, Rust's type inferencer will figure out that `provider` must be a `ProviderType`.

components/suggest/src/suggest.udl
```diff
@@ -26,6 +32,7 @@ dictionary Suggestion {
     sequence<u8>? icon;
     string? impression_url;
     string? click_url;
+    ProviderType? provider;
```
linabutler, line 35:
```suggestion
    ProviderType provider;
```

...And, if you do the `provider` trick I mentioned above, this can change to be non-optional!

linabutler (commented):
You almost got it! 😸

components/suggest/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub struct Suggestion {
     pub icon: Option<Vec<u8>>,
     pub impression_url: Option<String>,
     pub click_url: Option<String>,
+    pub provider: Option<ProviderType>,
```
mhammond, line 34: I'm really stepping out of my wheelhouse here, so please feel free to ignore this - but it's not clear this needs to be an option - it looks like you panic when reading the database and find an unknown option. Even if you eventually do not panic, it seems likely you would just ignore an unknown provider - so having this as a non-option might make life easier for consumers if they never have to deal with a `None` in practice.

If you do want to keep it: the DB has the column as non-null, so if you translate an unknown value to `None`, you've lost the actual (unknown) int value - so another option would be another enum variant `Unknown(u8)` - this means you still avoid the `Option<>` while still being able to handle "unknown" values *and* being able to know what the unknown value is (eg, for logging etc).

tiftran: it's just my newbie rust abilities shining through 😬 .

mhammond: > it's just my newbie rust abilities shining through 😬 .

Don't worry, we all remember that feeling very well :) Learning Rust is guaranteed to bring the latent imposter syndrome in everyone to the surface :) It takes longer to feel confident in Rust than learning most other languages in my experience, but it does happen :)

components/suggest/src/rs.rs
```diff
@@ -146,18 +149,126 @@ impl SuggestRecordId {
...
+        // we want to represent them separately. To distinguish between the two,
+        // we use an "untagged" outer enum and a "tagged" inner enum.
+        //
+        // Wikipedia suggestions always use the `"Wikipedia"` advertiser, so
+        // they'll deserialize successfully into the `KnownTag` variant.
+        // AMP suggestions will try the `KnownTag` variant first, fail, then
+        // try the `UnknownTag` variant and succeed.
+        //
+        // This strategy is an implementation detail, so we turn the nested
+        // enums into a friendlier `DownloadedAmpSuggestion` enum after
```
linabutler, line 248:
```suggestion
        // enums into a friendlier `DownloadedSuggestion` enum after
```

components/suggest/src/rs.rs
```diff
@@ -146,18 +147,91 @@ impl SuggestRecordId {
...
 
-/// A suggestion to ingest from a downloaded Remote Settings attachment.
-#[derive(Clone, Debug, Deserialize)]
-pub(crate) struct DownloadedSuggestion {
-    #[serde(rename = "id")]
-    pub block_id: i64,
-    pub advertiser: String,
-    pub iab_category: String,
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+pub(crate) struct DownloadedBaseSuggestion {
```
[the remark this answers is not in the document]
linabutler: (We decided on `DownloadedSuggestionCommonDetails` on Slack 😊)

components/suggest/src/db.rs
```diff
@@ -122,41 +122,78 @@ impl<'a> SuggestDao<'a> {
...
-             FROM suggestions s
-             JOIN keywords k ON k.suggestion_id = s.id
-             WHERE k.keyword = :keyword
-             LIMIT 1",
+            "SELECT s.id, k.rank, s.title, s.url, s.provider,
+                        (SELECT i.data FROM icons i WHERE i.id = s.icon_id) AS icon
+                  FROM suggestions s
+                  JOIN keywords k ON k.suggestion_id = s.id
+                  WHERE k.keyword = :keyword
+                  LIMIT 1",
```
linabutler, line 130:
```suggestion
            "SELECT s.id, k.rank, s.title, s.url, s.provider,
                    (SELECT i.data FROM icons i WHERE i.id = s.icon_id) AS icon
             FROM suggestions s
             JOIN keywords k ON k.suggestion_id = s.id
             WHERE k.keyword = :keyword
             LIMIT 1",
```

components/suggest/src/db.rs
```diff
@@ -122,41 +122,78 @@ impl<'a> SuggestDao<'a> {
...
-                    icon: row.get("icon")?,
-                    impression_url: row.get("impression_url")?,
-                    click_url: row.get("click_url")?,
-                })
+
+                match provider {
+                    SuggestionProvider::Amp => {
+                        self.conn.query_row_and_then(
+                            "SELECT amp.advertiser, amp.block_id, amp.iab_category, amp.impression_url, amp.click_url
+                            FROM amp_custom_details amp WHERE amp.suggestion_id = :suggestion_id",
```
linabutler, line 156:
```suggestion
                            "SELECT amp.advertiser, amp.block_id, amp.iab_category, amp.impression_url, amp.click_url
                             FROM amp_custom_details amp
                             WHERE amp.suggestion_id = :suggestion_id",
```

linabutler (approved):
r+ with formatting nits and the `base` => `common_details` name change! Thank you so much, @tiftran, this is lovely—let's get it landed! 🚀

components/suggest/src/db.rs
tiftran: 🤔 how did this not get in.... 🤦‍♀️

linabutler (approved):
🚀 :shipit: 👏🏼

PR #5747, opened 2023-07-27
Update Glean to v53.1.0

rvandermeulen:
#5678 already updated the submodule to the v53.1.0 commit, so just need to update the Android and iOS references.


linabutler (approved):
Thanks!

PR #5760, opened 2023-07-31
Update Cargo.lock for a previous commit which missed a syn upgrade

mhammond:
It looks like 6c90ecc6b59e823364bdffae40d61bf7c7aef116 missed a crate.


tarikeshaq (approved):
🚀

PR #5764, opened 2023-08-01
Fixes DISCO-2566: Improve Remote Settings client ergonomics for the Suggest component

linabutler:
This PR adds a few conveniences to the Remote Settings Rust client, and reworks the Suggest component to use them. These conveniences let Suggest use the `RemoteSettingsResponse` and `RemoteSettingsRecord` types directly, instead of defining its own duplicate "shadow types".

* The new `get_records_with_options` method returns a `RemoteSettingsResponse`, complementing the other `_raw` variants that return the Viaduct `Response`.
* `get_attachment` now returns a `Vec<u8>`, and the old logic moves to `get_attachment_raw`. This change makes it easier to mock the Remote Settings client in Suggest—check out the simplified `SuggestRemoteSettingsClient` trait.
* There's a new `RemoteSettingsRecord::deleted` property that surfaces whether a record is a tombstone. This makes it easier to tell tombstones apart, and handle them separately.
* The bulk of the Suggest changes are tests, to remove an extra layer of nesting for the fake Remote Settings records.

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -15,6 +15,7 @@ use url::Url;
 use viaduct::{Request, Response};
 
 const HEADER_BACKOFF: &str = "Backoff";
+const HEADER_ETAG: &str = "ETag";
 const HEADER_RETRY_AFTER: &str = "Retry-After";
 
 /// A simple HTTP client that can retrieve Remote Settings data using the properties by [ClientConfig].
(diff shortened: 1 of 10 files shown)


components/remote_settings/src/client.rs
```diff
@@ -102,22 +101,27 @@ impl Client {
...
-                    None => Err(RemoteSettingsError::AttachmentsUnsupportedError)?,
-                }
-            }
+    pub fn get_attachment(&self, attachment_location: &str) -> Result<Vec<u8>> {
+        Ok(self.get_attachment_raw(attachment_location)?.body)
+    }
+
+    /// Fetches a raw network [Response] for an attachment.
+    pub fn get_attachment_raw(&self, attachment_location: &str) -> Result<Response> {
+        if let Some(attachments_base_url) = self.remote_state.lock().attachments_base_url.clone() {
```
mhammond, line 110: This confused me a little - is it possible to match on the value in the state, so we don't early return here but still fall through to the `make_request` at the bottom?

components/remote_settings/src/client.rs
```diff
@@ -102,22 +101,27 @@ impl Client {
...
+        let attachments_base_url = match server_info.capabilities.attachments {
+            Some(capability) => Url::parse(&capability.base_url)?,
+            None => Err(RemoteSettingsError::AttachmentsUnsupportedError)?,
         };
-        current_remote_state.attachments_base_url = Some(attachments_base_url.clone());
-        drop(current_remote_state);
+        self.remote_state
+            .lock()
+            .attachments_base_url
+            .get_or_insert(attachments_base_url.clone());
```
mhammond, line 124: This also confused me and the semantics are subtly changed. On one hand this is better because it doesn't hold the lock while we make a request, but OTOH, now it seems possible multiple threads will race to update the base URL and the first will win. If the expectation is that the server will never update the URL (which are the semantics pre this patch IIUC), should we assert/log/err the existing value is the same once we've re-acquired the lock? Or if we want to handle the face that the server might update the base URL, should we use `Option::insert()` to ensure the last request wins?

(TBH this is such an edge-case I don't particularly care, but maybe some comments with the implicit assumptions might help)

mhammond (approved):

PR #5766, opened 2023-08-02
Fixes DISCO-2569: Prepare to support other types of suggestions in the Suggest Rust component

linabutler:
This commit:

* Renames some of our crate-internal types to clarify that they're specific to AMP-Wikipedia suggestions.
* Removes `icon_id` from the common `suggestions` table, because not all suggestions that we'll support in the future have icon data. For example, Pocket suggestions don't have icons at all, and AMO suggestions have URLs instead of binary data.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -16,7 +16,8 @@ use sql_support::{open_database::open_database_with_flags, ConnExt};
 
 use crate::{
     keyword::full_keyword,
-    rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId, SuggestionProvider},
+    provider::SuggestionProvider,
+    rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId},
     schema::SuggestConnectionInitializer,
     Result, Suggestion,
 };
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -16,7 +16,7 @@ use sql_support::{open_database::open_database_with_flags, ConnExt};
 
 use crate::{
     keyword::full_keyword,
-    rs::{DownloadedSuggestion, SuggestRecordId, SuggestionProvider},
+    rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId, SuggestionProvider},
     schema::SuggestConnectionInitializer,
     Result, Suggestion,
 };
@@ -122,12 +122,11 @@ impl<'a> SuggestDao<'a> {
     /// Fetches suggestions that match the given keyword from the database.
     pub fn fetch_by_keyword(&self, keyword: &str) -> Result<Vec<Suggestion>> {
         self.conn.query_rows_and_then_cached(
-            "SELECT s.id, k.rank, s.title, s.url, s.provider,
-                        (SELECT i.data FROM icons i WHERE i.id = s.icon_id) AS icon
-                  FROM suggestions s
-                  JOIN keywords k ON k.suggestion_id = s.id
-                  WHERE k.keyword = :keyword
-                  LIMIT 1",
+            "SELECT s.id, k.rank, s.title, s.url, s.provider
+             FROM suggestions s
+             JOIN keywords k ON k.suggestion_id = s.id
+             WHERE k.keyword = :keyword
+             LIMIT 1",
             named_params! {
                 ":keyword": keyword,
             },
@@ -135,7 +134,6 @@ impl<'a> SuggestDao<'a> {
                 let suggestion_id: i64 = row.get("id")?;
                 let title = row.get("title")?;
                 let url = row.get("url")?;
-                let icon = row.get("icon")?;
                 let provider = row.get("provider")?;
 
                 let keywords: Vec<String> = self.conn.query_rows_and_then_cached(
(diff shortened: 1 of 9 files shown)


components/suggest/src/rs.rs
```diff
@@ -132,109 +126,81 @@ pub(crate) struct DownloadedAmpSuggestion {
...
+    #[serde(rename = "icon")]
+    pub icon_id: String,
 }
 
-/// A suggestion to ingest from a downloaded Remote Settings attachment.
+/// A suggestion to ingest from an AMP-Wikipedia attachment downloaded from
+/// Remote Settings.
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub(crate) enum DownloadedSuggestion {
+pub(crate) enum DownloadedAmpWikipediaSuggestion {
```
ncloudioj, line 144: Are we looking to add more suggestion types (as enum variants) here in the future? If so, having `Amp` and `Wikipedia` in the type identifier seems a bit inflexible to me. `DownloadedSuggestion` sounds perfect to me. If we will introduce new enums instead, plz disregard my comment :)

components/suggest/src/rs.rs
```diff
@@ -132,109 +126,81 @@ pub(crate) struct DownloadedAmpSuggestion {
...
+    fn deserialize<D>(
+        deserializer: D,
+    ) -> std::result::Result<DownloadedAmpWikipediaSuggestion, D::Error>
     where
         D: Deserializer<'de>,
     {
-        // AMP and Wikipedia suggestions conform to the same JSON schema, but
-        // we want to represent them separately. To distinguish between the two,
-        // we use an "untagged" outer enum and a "tagged" inner enum.
+        // AMP and Wikipedia suggestions are stored in the same attachment
```
ncloudioj, line 174: To clarify, we don't mix those two suggestions within the same RS attachment, and that's enforced at the publishing time.

ncloudioj (approved):
Nice refactor! We're one step closer to support more suggestion types! 🎉

components/suggest/src/db.rs
```diff
@@ -152,58 +147,57 @@ impl<'a> SuggestDao<'a> {
...
                     },
-                    SuggestionProvider::Wikipedia =>{
-                        Ok(Suggestion {
-                            block_id: 0,
-                            advertiser: "Wikipedia".to_string(),
-                            iab_category: "5 - Education".to_string(),
-                            is_sponsored: false,
+                    SuggestionProvider::Wikipedia => {
+                        let icon = self.conn.try_query_one::<Vec<u8>, _>(
+                            "SELECT i.data AS icon
```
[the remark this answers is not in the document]
linabutler: Thanks for pointing that out! It was part of the reason I moved `icon_id` into the AMP and Wikipedia tables 😊 I noticed that AMO suggestions use `https://` URLs for icons, instead of including the icon data directly; and Pocket suggestions didn't have icons in Remote Settings at all.

Android and iOS can load icons from `https://` and `data:` URLs (yay!) but not `chrome://` URLs (boo!): on Android, the awesomebar code is "pure" Android, outside of GeckoView, and iOS doesn't have Gecko at all.

components/suggest/src/db.rs
```diff
@@ -231,43 +223,59 @@ impl<'a> SuggestDao<'a> {
...
                     ":url": common_details.url,
-                    ":icon_id": common_details.icon_id
 
                 },
                 |row| row.get(0),
                 true,
             )?;
             match suggestion {
-                DownloadedSuggestion::Amp(amp) => {
+                DownloadedAmpWikipediaSuggestion::Amp(amp) => {
```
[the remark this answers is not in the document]
linabutler: Oh, those are great links! We use the `VALUES (...), (...)` trick [quite](https://github.com/mozilla/application-services/blob/5384d4893b8eab259b6c9398363b6803c97d2eb0/components/logins/src/sync/engine.rs#L176-L180) [extensively](https://github.com/mozilla/application-services/blob/5384d4893b8eab259b6c9398363b6803c97d2eb0/components/autofill/src/sync/common.rs#L40-L59) to [chunk](https://github.com/mozilla/application-services/blob/5384d4893b8eab259b6c9398363b6803c97d2eb0/components/webext-storage/src/sync/incoming.rs#L72-L105) [binding](https://github.com/mozilla/application-services/blob/5384d4893b8eab259b6c9398363b6803c97d2eb0/components/places/src/bookmark_sync/engine.rs#L139-L152) [parameters](https://github.com/mozilla/application-services/blob/5384d4893b8eab259b6c9398363b6803c97d2eb0/components/places/src/bookmark_sync/engine.rs#L254-L269) and [reduce](https://github.com/mozilla/application-services/blob/main/components/places/src/bookmark_sync/incoming.rs#L189-L211) the number of statements we run—definitely a lower-hanging perf boost! Once our schema settles a bit, I'd love to benchmark our reads and writes, and see how we're doing.

components/suggest/src/rs.rs
linabutler: I'm glad you noticed this! 😄 I was thinking of introducing new structs for the new types, which is why I renamed it, and the `SuggestDao` method, to be specific to AMP-Wikipedia.

IIUC, AMP-Wikipedia attachments have the same Remote Settings record type (`data`), but the newer ones have different record types (`amo-suggestions` and `pocket-suggestions`). We could definitely have variants for those in `DownloadedSuggestion`, but I mostly wanted to avoid that for simplicity—we can avoid complicating the custom deserializer with cases for these other different types.

components/suggest/src/rs.rs
linabutler: Ooh, thanks for clarifying that; I actually wasn't sure if that was enforced, so I'll update the comment! It doesn't change the logic, but let's make sure the comment is accurate.

components/suggest/src/db.rs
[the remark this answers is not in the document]
ncloudioj: Gotcha! Thanks for that note! We use those TippyTop icons (`chrome://`) for Dynamic Wikipedia and Top Picks for desktop, definitely need to revisit that once we bring those providers to x-platform.

components/suggest/src/rs.rs
ncloudioj: Ah, I see. So we're modelling against those record types here. Makes sense, let's leave it as is then. Maybe add a quick comment here for future reference.

components/suggest/src/db.rs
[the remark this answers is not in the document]
linabutler: Filed https://mozilla-hub.atlassian.net/browse/DISCO-2573 for our future selves, thanks! 😊

components/suggest/src/rs.rs
linabutler: Added some docs for how to add new suggestion types in https://github.com/mozilla/application-services/pull/5766/commits/839f4990dc395ad471fe9e08199599366009acd7 😁👍🏼 (/cc @tiftran)

PR #5767, opened 2023-08-03
Fixes DISCO-2571: Fix re-ingesting icons in the Suggest Rust component

linabutler:
We used `INSERT` instead of `REPLACE` to store icons, so trying to re-ingest an updated icon would fail with a unique constraint error.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -352,7 +352,7 @@ impl<'a> SuggestDao<'a> {
     /// Sets the value for a metadata key.
     pub fn put_meta(&mut self, key: &str, value: impl ToSql) -> Result<()> {
         self.conn.execute_cached(
-            "REPLACE INTO meta(key, value) VALUES(:key, :value)",
+            "INSERT OR REPLACE INTO meta(key, value) VALUES(:key, :value)",
             named_params! { ":key": key, ":value": value },
         )?;
         Ok(())
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -292,10 +292,10 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
-    /// Inserts an icon for a suggestion into the database.
-    pub fn insert_icon(&mut self, icon_id: &str, data: &[u8]) -> Result<()> {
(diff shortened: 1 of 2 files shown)


components/suggest/src/db.rs
```diff
@@ -292,10 +292,10 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
-    /// Inserts an icon for a suggestion into the database.
-    pub fn insert_icon(&mut self, icon_id: &str, data: &[u8]) -> Result<()> {
+    /// Inserts or replaces an icon for a suggestion into the database.
+    pub fn put_icon(&mut self, icon_id: &str, data: &[u8]) -> Result<()> {
         self.conn.execute(
-            "INSERT INTO icons(
+            "REPLACE INTO icons(
```
ncloudioj, line 298: Q: per SQLite's [doc](https://www.sqlite.org/lang_replace.html), `REPLACE` is a shorthand of `INSERT OR REPLACE` borrowed from MySQL, I'd recommend using the latter for clarity.

Also, icons are append-only on RemoteSettings, we'd never delete/modify icons once published to RS. So technically, we can also do `INSERT OR IGNORE`.

components/suggest/src/store.rs
```diff
@@ -869,6 +869,209 @@ mod tests {
...
+                                97,
+                                103,
+                                110,
+                                97,
+                                45,
+                                105,
+                                99,
+                                111,
+                                110,
+                            ],
```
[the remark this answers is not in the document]
ncloudioj: nod, nod, i trust u :P

ncloudioj (approved):
:shipit:

components/suggest/src/db.rs
linabutler: Sounds good to me! I've been using `REPLACE` for a while, but let's use `INSERT OR REPLACE` in Suggest. I'll fix that here and in `put_meta`.

Good to know about icons being append-only, thanks! I think I'll keep it as `INSERT OR REPLACE` just in case, even though we don't expect updates, but it helps to know that this isn't an issue we'll expect to see in practice.

PR #5770, opened 2023-08-04
Discuss: async component APIs

bendk:
This changes the tabs component to by async rather than use a dispatch queue so that we can discuss the proper way to do this.

  - Note:

### DIFF
--- a/components/tabs/ios/Tabs/Tabs.swift
+++ b/components/tabs/ios/Tabs/Tabs.swift
@@ -6,28 +6,27 @@ import Foundation
 
 open class TabsStorage {
     private var store: TabsStore
-    private let queue = DispatchQueue(label: "com.mozilla.tabs-storage")
 
     public init(databasePath: String) {
         store = TabsStore(path: databasePath)
     }
 
     /// Get all tabs by client.
-    open func getAll() -> [ClientRemoteTabs] {
-        return queue.sync {
+    open func getAll() async -> [ClientRemoteTabs] {
+        Task {
             self.store.getAll()
         }
     }
 
     /// Set the local tabs.
-    open func setLocalTabs(remoteTabs: [RemoteTabRecord]) {
-        queue.sync {
+    open func setLocalTabs(remoteTabs: [RemoteTabRecord]) async {
+        Task {
             self.store.setLocalTabs(remoteTabs: remoteTabs)
         }
     }
 
-    open func registerWithSyncManager() {
-        queue.sync {
+    open func registerWithSyncManager() async {
+        Task {
             self.store.registerWithSyncManager()
         }
     }


components/tabs/ios/Tabs/Tabs.swift
```diff
@@ -6,28 +6,27 @@ import Foundation
...
         store = TabsStore(path: databasePath)
     }
 
     /// Get all tabs by client.
-    open func getAll() -> [ClientRemoteTabs] {
-        return queue.sync {
+    open func getAll() async -> [ClientRemoteTabs] {
+        Task {
             self.store.getAll()
         }
```
bendk, line 18: I think this is the main question: can we simply replace `DispatchQueue` with `Task` and be done with it?

In particular: `getAll()` runs a database query that could potentially fetch a lot of data.  Would using a `Task` like this block the main thread or other background tasks while the query is running or would they run in parallel?  Should we be using `Task.detached` instead?

OrlaM: Task will never run the block of code on the main thread unless you specifically request it to.
That syntax would look like this:
```
Task { @MainActor in
    // Some code
 }
```
For Task.detached it's generally recommended to avoid using these unless you have a compelling reason to do so. There is a good write up here on what the difference is if you want to know more https://www.hackingwithswift.com/quick-start/concurrency/whats-the-difference-between-a-task-and-a-detached-task

OrlaM:
Is the goal of using the Tasks in this PR to make sure the requests are never run on the main thread or is it to ensure all of the work is done serially to avoid potential data races, or both?

linabutler:
> Is the goal of using the Tasks in this PR to make sure the requests are never run on the main thread

I think just to make sure that the requests are never run on the main thread. It's not necessary for the work to be done serially.

components/tabs/ios/Tabs/Tabs.swift
linabutler: Oh, thanks for clarifying that! Just to make super sure—will making a blocking I/O call inside a task (detached or not) prevent all other tasks from running until that call is done?

I wanted to double-check because all the examples of Swift concurrency I've seen so far use `await` inside the operation passed to `Task`. These examples also take care to mention that tasks should always make forward progress.

But our case feels a bit different. In our case, we're calling into foreign Rust code that does synchronous disk (and sometimes network) I/O. The Rust code has no awareness of the Swift task runtime, so it can't suspend itself, and can't be canceled until it's done.

At the same time, we want Swift callers to be able to do other things while the Rust code is blocked on I/O. For a concrete example, let's say that `SearchViewController` calls `TabsStorage.getAll()`. We want the controller to be able to run `TabsStorage.getAll()` concurrently with the network call to query Google for search engine suggestions, and with calls to Places to search history and bookmarks. The thing we want to avoid is forcing the controller to wait for `TabsStorage.getAll()` to run to completion before it can start fetching search engine suggestions, or querying bookmarks and history.

Zooming out a bit, I think what I'm struggling to understand with Swift concurrency is how to adapt existing blocking functions to work with it.

With GCD, we gave `TabsStorage` and `RustPlaces` their own private serial queues—not because we need the sequential execution semantics, but to avoid thread explosion with concurrent queues. IIUC, that meant that all `TabsStorage` operations would run serially, and all `RustPlaces` operations would run serially, but `TabsStorage` operations could (but not necessarily guaranteed to) run concurrently with `RustPlaces` operations, and vice versa. Is that mental model accurate?

With Swift concurrency, I'm wondering if that's still the case—just without the need to create our own dispatch queues—or if the task runtime executes _all_ tasks sequentially, and so a single long-running task will block all other tasks from running until it's done.

(So sorry for that long wall of text! 😅 I haven't really found a good explanation for this, and it would be amazing to have your insights and guidance, when you have some time—and thank you!)

bendk:
> Is the goal of using the Tasks in this PR to make sure the requests are never run on the main thread or is it to ensure all of the work is done serially to avoid potential data races, or both?

Hi Orla, just getting back to this one now that I understand more about async stuff.  The goal is basically "never run on the main thread", but a many Rust libraries make a further distinction:
  - Tasks that will always run quickly then suspend.  For example execute async IO / async network requests, then quickly process the results.  These tasks get scheduled on a queue with a relatively small amount of threads (maybe 4).  The small number of threads is fine, since these tasks always suspend quickly after they're started.
  - Tasks that run cpu-intensive work, execute blocking IO/network calls, etc.  These get scheduled on a different queue with a growable amount of threads (maybe up to 512).

I think we're mostly going to be running a lot of the second kind of task -- at least in the short term.  We're going to take our current sync rust functions and wrap them with something to make them async.  I don't think we're going to need anywhere near 512 threads, but I could imagine a handful of Rust tasks running that are waiting for SQLite results or something similar.  So the question is, should we be using `Task`, `DispatchQueue`, or something else for this?

OrlaM:
Since you are going to be making calls that block threads that are not using swift concurrency under the hood I think using DispatchQueue would be the way to go and then continuations anywhere it interfaces with code that uses swift concurrency.

bendk:
> Since you are going to be making calls that block threads that are not using swift concurrency under the hood I think using DispatchQueue would be the way to go and then continuations anywhere it interfaces with code that uses swift concurrency.

That makes sense, thanks.

BTW, I also found this thread in SO over the weekend which backs up yours: https://stackoverflow.com/questions/72862670/integrate-a-blocking-function-into-swift-async

I'm going to close this one because it looks like we have a clear direction.  Feel free to re-open if there's more to discuss though.

PR #5774, opened 2023-08-10
Fixes DISCO-2576: Replace Suggest databases with incompatible schema versions.

linabutler:
We have some prototype Suggest integrations for Desktop and Android that will fail to open databases with older schema versions. We haven't been writing migration logic for these versions, since we haven't shipped the component to any users yet, and the Suggest dataset being "read-only" means we aren't super worried about user data loss.

We can work around this by deleting the database ourselves, but, since we already have machinery for replacing corrupt databases, let's extend it so that migrations can force replacement.

  - Note:

### DIFF
--- a/components/support/sql/src/open_database.rs
+++ b/components/support/sql/src/open_database.rs
@@ -22,6 +22,8 @@
 ///        it and call prepare(), upgrade_from() for each upgrade that needs to be applied, then
 ///        finish(). As above, a read-only connection will panic if upgrades are necessary, so
 ///        you should ensure the first connection opened is writable.
+///      - If the database file is corrupt, or upgrade_from() returns [`Error::Corrupt`], the
+///        database file will be removed and replaced with a new DB.
 ///      - If the connection is not writable, `finish()` will be called (ie, `finish()`, like
 ///        `prepare()`, is called for all connections)
 ///
@@ -38,11 +40,25 @@ use thiserror::Error;
 pub enum Error {
     #[error("Incompatible database version: {0}")]
     IncompatibleVersion(u32),
+    #[error("Database is corrupt")]
+    Corrupt,
     #[error("Error executing SQL: {0}")]
-    SqlError(#[from] rusqlite::Error),
-    // `.0` is the original `Error` in string form.
-    #[error("Failed to recover a corrupt database ('{0}') due to an error deleting the file: {1}")]
-    RecoveryError(String, std::io::Error),
+    SqlError(rusqlite::Error),
+    #[error("Failed to recover a corrupt database due to an error deleting the file: {0}")]
+    RecoveryError(std::io::Error),
+}
+
+impl From<rusqlite::Error> for Error {
+    fn from(value: rusqlite::Error) -> Self {
+        match value {
+            RusqliteError::SqliteFailure(e, _)
+                if matches!(e.code, ErrorCode::DatabaseCorrupt | ErrorCode::NotADatabase) =>
+            {
+                Self::Corrupt
+            }
+            _ => Self::SqlError(value),
+        }
+    }
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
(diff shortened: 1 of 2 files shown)


components/support/sql/src/open_database.rs
```diff
@@ -22,6 +22,8 @@
 ///        it and call prepare(), upgrade_from() for each upgrade that needs to be applied, then
 ///        finish(). As above, a read-only connection will panic if upgrades are necessary, so
 ///        you should ensure the first connection opened is writable.
+///      - If the database file is corrupt, or your upgrade returns [`Error::Corrupted`], the
+///        database file will be removed and replaced with a new DB.
```
linabutler, line 26: note: Desktop does this in a few...places 😅 already, and I thought it's a nice pattern to expose more generally!

* [Places](https://searchfox.org/mozilla-central/rev/dcf50463d3e1e1f3a91bc1406e3772b2a5f62774/toolkit/components/places/Database.cpp#1112-1116) will replace databases from Firefox versions [older than the last ESR before the most recent watershed release](https://bugzilla.mozilla.org/show_bug.cgi?id=1847592).
* The [bookmarks mirror](https://searchfox.org/mozilla-central/rev/dcf50463d3e1e1f3a91bc1406e3772b2a5f62774/toolkit/components/places/SyncedBookmarksMirror.sys.mjs#1536-1541) throws away any databases from when the mirror was pref'd off, instead of trying to migrate them.

linabutler:
```suggestion
///      - If the database file is corrupt, or your upgrade returns [`Error::Corrupt`], the
///        database file will be removed and replaced with a new DB.
```

mhammond: Agree "Corrupt" isn't great - I really don't see a reason you can't reuse `IncompatibleVersion` here though?

mhammond (approved):
This looks great and any error name you choose is going to be perfectly fine in practice :)

components/support/sql/src/open_database.rs
linabutler: Oooh, good question! I wasn't sure about throwing the database away in case of an incompatible version—that felt like an error that we'd want to know about, especially since there's no way to downgrade a profile on mobile, and other migration errors don't trigger automatic replacement. But I'd also be okay with having `IncompatibleVersion` trigger the replacement behavior!

mhammond:
I'm sorry, I admit to a quick rubber stamp late last night 😅 On second thought, I actually think `Corrupt` is a reasonable name for the error itself. I agree it seems like the wrong name to use for a lower schema version than we want to support, but given how it's being used I think it's actually fine - it's just signalling to the sql-support crate it should be *treated* as if it's corrupt. We also don't want to delete "future" versions, and while today it's probably reasonable to not bother deleting "old" versions (ie, just make it the developer's problem to delete manually), in the longer term we will want something like this (ie, a version which has actually been released can't simply refuse to work when it finds an old version). So I actually think the patch is exactly correct as it stands.

linabutler:
Oh, wonderful! Thank you so much for the speedy review, and for continuing to think more about this—I'm actually really happy you like `Corrupt` as a name; I didn't like `NeedsReplacement` as much!

> We also don't want to delete "future" versions

This is also a great point! I went to write a "downgrade" test yesterday, but noticed that we [bail early if we see a future version](https://github.com/mozilla/application-services/blob/65723831c566ad3405781cfa65975a93936df486/components/support/sql/src/open_database.rs#L124-L126). Whether we keep that behavior as it is, or relax it to match Desktop's (where we'll [roll back the schema version, but not the actual changes to the schema](https://searchfox.org/mozilla-central/rev/503938c13ef2dd174705dc0f6d0683ae43074ccc/toolkit/components/places/Database.cpp#1095-1098)) in the future, it feels like silently deleting the database in that case would be the wrong thing to do.

PR #5786, opened 2023-08-17
Suggest: fixes [DISCO-2564] - Store unknown records

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -28,7 +28,7 @@ use crate::{
 pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
 /// The metadata key whose value keeps track of records of suggestions
 /// that aren't parsable and which schema version it was first seen in.
-pub const UNKNOWN_RECORDS_META_KEY: &str = "unknown_records";
+pub const UNPARSABLE_RECORDS_META_KEY: &str = "unparsable_records";
 
 /// The database connection type.
 #[derive(Clone, Copy)]
@@ -383,32 +383,40 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
-    /// Insert unknown record into the unknown records meta field.
-    pub fn put_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
-        let mut unknown_records = self
-            .get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?
+    /// Adds an entry for a Suggest Remote Settings record to the list of
+    /// unparsable records.
+    ///
+    /// This is used to note records that we don't understand how to parse and
+    /// ingest yet.
+    pub fn put_unparsable_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
+        let mut unparsable_records = self
+            .get_meta::<UnparsableRecords>(UNPARSABLE_RECORDS_META_KEY)?
             .unwrap_or_default();
-        unknown_records.0.insert(
+        unparsable_records.0.insert(
             record_id.as_str().to_string(),
             UnparsableRecord {
                 schema_version: VERSION,
             },
         );
-        self.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)?;
+        self.put_meta(UNPARSABLE_RECORDS_META_KEY, unparsable_records)?;
         Ok(())
     }
 
-    /// Remove the record id from the unknown records meta field.
-    pub fn drop_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
-        let Some(mut unknown_records) =
-            self.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?
+    /// Removes an entry for a Suggest Remote Settings record from the list of
+    /// unparsable records. Does nothing if the record was not previously marked
+    /// as unparsable.
+    ///
+    /// This indicates that we now understand how to parse and ingest the
+    /// record, or that the record was deleted.
+    pub fn drop_unparsable_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
+        let Some(mut unparsable_records) =
+            self.get_meta::<UnparsableRecords>(UNPARSABLE_RECORDS_META_KEY)?
         else {
-            // We don't have any unknown record IDs, so nothing to drop.
             return Ok(());
         };
-        if unknown_records.0.remove(record_id.as_str()).is_none() {
+        if unparsable_records.0.remove(record_id.as_str()).is_none() {
             return Ok(());
         };
-        self.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)
+        self.put_meta(UNPARSABLE_RECORDS_META_KEY, unparsable_records)
     }
 }
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -18,13 +18,17 @@ use crate::{
     keyword::full_keyword,
     provider::SuggestionProvider,
     rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId},
-    schema::SuggestConnectionInitializer,
+    schema::{SuggestConnectionInitializer, VERSION},
+    store::{UnparsableRecord, UnparsableRecords},
     Result, Suggestion,
 };
 
 /// The metadata key whose value is the timestamp of the last record ingested
 /// from the Suggest Remote Settings collection.
 pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
+/// The metadata key whose value keeps track of records of suggestions
+/// that aren't parsable and which schema version it was first seen in.
+pub const UNKNOWN_RECORDS_META_KEY: &str = "unknown_records";
 
 /// The database connection type.
 #[derive(Clone, Copy)]
@@ -364,5 +368,47 @@ impl<'a> SuggestDao<'a> {
             named_params! { ":key": key, ":value": value },
         )?;
         Ok(())
+    }
+
+    /// Updates the last ingest timestamp if the given last modified time is
+    /// newer than the existing one recorded.
+    pub fn put_last_ingest_if_newer(&mut self, record_last_modified: u64) -> Result<()> {
+        let last_ingest = self
+            .get_meta::<u64>(LAST_INGEST_META_KEY)?
+            .unwrap_or_default();
+        if record_last_modified > last_ingest {
+            self.put_meta(LAST_INGEST_META_KEY, record_last_modified)?;
+        }
+
+        Ok(())
+    }
+
+    /// Insert unknown record into the unknown records meta field.
+    pub fn put_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
+        let mut unknown_records = self
+            .get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?
+            .unwrap_or_default();
+        unknown_records.0.insert(
+            record_id.as_str().to_string(),
+            UnparsableRecord {
+                schema_version: VERSION,
+            },
+        );
+        self.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)?;
+        Ok(())
+    }
+
+    /// Remove the record id from the unknown records meta field.
+    pub fn drop_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
+        let Some(mut unknown_records) =
+            self.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?
+        else {
+            // We don't have any unknown record IDs, so nothing to drop.
+            return Ok(());
+        };
+        if unknown_records.0.remove(record_id.as_str()).is_none() {
+            return Ok(());
+        };
+        self.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)
     }
 }
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -3,19 +3,33 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-use std::path::{Path, PathBuf};
+use std::{
+    collections::BTreeMap,
+    path::{Path, PathBuf},
+};
 
 use once_cell::sync::OnceCell;
-use remote_settings::{self, GetItemsOptions, RemoteSettingsConfig, SortOrder};
+use remote_settings::{
+    self, GetItemsOptions, RemoteSettingsConfig, RemoteSettingsRecord, SortOrder,
+};
+use rusqlite::{
+    types::{FromSql, ToSqlOutput},
+    ToSql,
+};
+use serde::{Deserialize, Serialize};
 
 use crate::{
-    db::{ConnectionType, SuggestDb, LAST_INGEST_META_KEY},
+    db::{ConnectionType, SuggestDb, LAST_INGEST_META_KEY, UNKNOWN_RECORDS_META_KEY},
     rs::{
         SuggestAttachment, SuggestRecord, SuggestRecordId, SuggestRemoteSettingsClient,
         REMOTE_SETTINGS_COLLECTION, SUGGESTIONS_PER_ATTACHMENT,
     },
+    schema::VERSION,
     Result, SuggestApiResult, Suggestion, SuggestionQuery,
 };
+
+/// The chunk size used to request unparsable records.
+pub const UNPARSABLE_IDS_PER_REQUEST: usize = 150;
 
 /// The store is the entry point to the Suggest component. It incrementally
 /// downloads suggestions from the Remote Settings service, stores them in a
@@ -47,6 +61,35 @@ use crate::{
 ///    on the first launch.
 pub struct SuggestStore {
     inner: SuggestStoreInner<remote_settings::Client>,
+}
+
+/// For records that aren't currently parsable,
+/// the record ID and the schema version it's first seen in
+/// is recorded in the meta table using `UNKNOWN_RECORDS_META_KEY` as its key.
+/// On the first ingest after an upgrade, re-request those records from Remote Settings,
+/// and try to ingest them again.
+#[derive(Deserialize, Serialize, Default, Debug)]
+#[serde(transparent)]
+pub(crate) struct UnparsableRecords(pub BTreeMap<String, UnparsableRecord>);
+
+impl FromSql for UnparsableRecords {
+    fn column_result(value: rusqlite::types::ValueRef<'_>) -> rusqlite::types::FromSqlResult<Self> {
+        serde_json::from_str(value.as_str()?)
+            .map_err(|err| rusqlite::types::FromSqlError::Other(Box::new(err)))
+    }
+}
+
+impl ToSql for UnparsableRecords {
+    fn to_sql(&self) -> rusqlite::Result<rusqlite::types::ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(serde_json::to_string(self).map_err(
+            |err| rusqlite::Error::ToSqlConversionFailure(Box::new(err)),
+        )?))
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub(crate) struct UnparsableRecord {
+    pub schema_version: u32,
 }
 
 impl SuggestStore {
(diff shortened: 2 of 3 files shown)


components/suggest/src/store.rs
```diff
@@ -197,15 +235,23 @@ where
...
                 // We don't recognize this record's type, so we don't know how
-                // to ingest its suggestions. Skip to the next record.
-                writer.write(|dao| dao.put_meta(LAST_INGEST_META_KEY, record.last_modified))?;
+                // to ingest its suggestions. Record this in the meta table.
+                writer.write(|dao| {
+                    // TODO: Read `unknowns`, update JSON, and write it back
+                    dao.put_meta(LAST_INGEST_META_KEY, record.last_modified)?;
+                    let mut unknown_records = dao.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?.unwrap_or_default();
+                    unknown_records.0.insert(record.id.clone(), UnparsableRecord{schema_version: VERSION});
+                    dao.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)?;
```
ncloudioj, line 252: Potential concern: this would load/store `UnparsableRecords` from the meta table for every single unknown record observed. What do you think about stashing all the unknown records and storing them in bulk at the end of the ingestion?

components/suggest/src/store.rs
```diff
@@ -234,6 +280,15 @@ where
                         // attachment.
                         dao.insert_amp_wikipedia_suggestions(&record_id, attachment.suggestions())?;
 
+                        // Remove this `record.id` from unknowns, since we
+                        // understand it now.
+                        if let Some(mut unknown_records) =
+                            dao.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)?
```
ncloudioj, line 286: Likewise, the removal can be stashed and persisted once at the end.

components/suggest/src/store.rs
linabutler: Tif actually did it that way at first, and we chatted about the different approaches over Zoom! 😁

Storing the unknown record IDs in bulk at the end of the ingestion avoids the extra loads and stores—but risks losing the IDs if we're interrupted during ingestion. Interruptions might be for reasons outside of our control, like the OS terminating the background worker.

Since we're advancing the last ingest time after each attachment, and known records can be sandwiched between unknown ones, we could end up in a scenario where we advance the last ingest time past an unknown record, but get terminated before we can save that unknown record's ID at the end. At that point, we won't see the record again, until the next time it's changed in Remote Settings.

(We saw a version of this in Firefox Sync, where it "missed" changes because it advanced the "last sync" timestamp out of lockstep with updating the database).

One way we could work around that—and still store the unknown records in bulk—is to keep an SQLite transaction open for the entire ingestion. But I think that's tricky; we'd need to hold the transaction open while we do HTTP, and we'd lose progress on the entire ingestion if we were interrupted halfway through.

We considered using SQLite's [built-in `json_patch` function](https://www.sqlite.org/json1.html#jpatch) to update the `UNKNOWN_RECORDS_META_KEY`, or a separate table. Serde has [excellent performance](https://github.com/serde-rs/json-benchmark), though, on par with SQLite's JSON implementation, and it's a bit easier to work with Serde. A separate table could be a good option to explore if we find ourselves accumulating large numbers of unknown record IDs.

Does that track for you?

components/suggest/src/store.rs
```diff
@@ -48,6 +56,25 @@ use crate::{
...
+
+impl FromSql for UnparsableRecords {
+    fn column_result(value: rusqlite::types::ValueRef<'_>) -> rusqlite::types::FromSqlResult<Self> {
+        Ok(serde_json::from_str(value.as_str()?).unwrap())
+    }
+}
+
+impl ToSql for UnparsableRecords {
+    fn to_sql(&self) -> rusqlite::Result<rusqlite::types::ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(serde_json::to_string(self).unwrap()))
```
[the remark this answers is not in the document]
linabutler: That suggestion actually looks perfect! ✨

components/suggest/src/store.rs
```diff
@@ -48,6 +56,25 @@ use crate::{
 pub struct SuggestStore {
     inner: SuggestStoreInner<remote_settings::Client>,
 }
+#[derive(Deserialize, Serialize, Default, Debug)]
+#[serde(transparent)]
+pub struct UnparsableRecords(BTreeMap<String, UnparsableRecord>);
+
+impl FromSql for UnparsableRecords {
+    fn column_result(value: rusqlite::types::ValueRef<'_>) -> rusqlite::types::FromSqlResult<Self> {
+        Ok(serde_json::from_str(value.as_str()?).unwrap())
```
linabutler, line 65: I might use a different error type here:

```suggestion
        serde_json::from_str(value.as_str()?)
            .map_err(|err| rusqlite::types::FromSqlError::Other(Box::new(err)))
```

...But yes, it's a great suggestion to avoid `unwrap` here! 😊

components/suggest/src/store.rs
ncloudioj: Of course, interruptions! I notice that `SuggestStoreInner::interrupt()` would only interrupt the reader, looks like the writer can also be interrupted by other means.

Yup, let's leave it as is since there are only a dozen of unknown records in production. We can revisit this later if needed.

components/suggest/src/store.rs
```diff
@@ -197,15 +235,23 @@ where
...
                 })?;
                 continue;
             }
             let Ok(fields) = serde_json::from_value(serde_json::Value::Object(record.fields.clone())) else {
                 // We don't recognize this record's type, so we don't know how
-                // to ingest its suggestions. Skip to the next record.
-                writer.write(|dao| dao.put_meta(LAST_INGEST_META_KEY, record.last_modified))?;
+                // to ingest its suggestions. Record this in the meta table.
+                writer.write(|dao| {
+                    // TODO: Read `unknowns`, update JSON, and write it back
```
ncloudioj, line 248:
```suggestion

```

components/suggest/src/store.rs
```diff
@@ -253,12 +308,20 @@ where
                     writer.write(|dao| {
                         dao.put_icon(icon_id, &data)?;
                         dao.put_meta(LAST_INGEST_META_KEY, record.last_modified)?;
+                        // Remove this `record.id` from unknowns, since we
+                        // understand it now.
+                        if let Some(mut unknown_records) =
```
ncloudioj, line 313: A couple of notes:
- This removal block was used by both arms of the match, looks like we can take them out and do it once
- It doesn't need to update the meta table if `record.id` is not in the map
- Optionally, I think we can only load `UNKNOWN_RECORDS_META_KEY` once out of the loop and mutate it as we go. Need to add `Clone` trait to both `UnparsableRecords` and `UnparsableRecord`, so feel free to disregard this if you mind the extra complexity

It will look like this:

```rs
        // Load unknown records from db once
        let mut unknown_records = writer.read(|dao| {
            dao.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY)
        })?.unwrap_or_default();

        for record in &records {
            let record_id = SuggestRecordId::from(&record.id);
            ...
            let Ok(fields) = serde_json::from_value(serde_json::Value::Object(record.fields.clone())) else {
                // We don't recognize this record's type, so we don't know how
                // to ingest its suggestions. Record this in the meta table.
                unknown_records.0.insert(record.id.clone(), UnparsableRecord{schema_version: VERSION});
                let clone = unknown_records.clone();
                writer.write(|dao| {
                    dao.put_meta(LAST_INGEST_META_KEY, record.last_modified)?;
                    dao.put_meta(UNKNOWN_RECORDS_META_KEY, clone)?;
                    Ok(())
                })?;
                continue;
            };
            match fields {
                // Will do the removal below.
            }
            
            // Drop from unknown records if needed. 
            if unknown_records.0.remove(&record.id).is_some() {
                let clone = unknown_records.clone();
                writer.write(|dao| {
                    dao.put_meta(UNKNOWN_RECORDS_META_KEY, clone)?;
                    Ok(())
                })?;
            }
```

linabutler: Oooh, I definitely agree about the duplication—I might suggest fixing that by moving the read-write-update logic to a pair of methods on `SuggestDao`, though—`put_unknown_record_id` and `delete_unknown_record_id`?—instead of DRYing it out here.

There's a super subtle difference in the way that we're doing it now, and your example. We're reading, updating, and writing `UNKNOWN_RECORDS_META_KEY` inside the `writer.write` block. That matches how SQLite's `json_patch` would work if we were using it, but, more importantly, it ensures that the update happens atomically in a transaction, so concurrent writes to `moz_meta` won't race.

Hoisting the load of `UNKNOWN_RECORDS_META_KEY` out of the loop is generally a little dangerous—we'd read it outside of a transaction, and then periodically write the mutated in-memory state in a transaction, so concurrent ingests could race to update the column. We don't expect concurrent calls to `ingest()`, but I like to be defensive about them 😅

components/suggest/src/store.rs
linabutler: Sounds good! Yeah, between crashes, the user quitting the app, and the OS stopping background work—especially on mobile—there are lots of ways of being interrupted that we can't control.

components/suggest/src/store.rs
ncloudioj: Yup yup, loading/mutating the map in memory definitely requires (implicitly) the single-writer semantics, which is a big assumption to make. So your proposal makes a lot sense to me.

Is it possible to drop a record id that is not in the map? I didn't fully check the code path, so I just assumed "everything is possible", meaning that we don't need to write the map back if the `map.remove()` returns `None`.

linabutler: > we don't need to write the map back if the `map.remove()` returns `None`.

Oh, nice catch! Yes, we can definitely avoid the write-back if `map.remove()` returns `None`.

components/suggest/src/db.rs
```diff
@@ -365,4 +369,29 @@ impl<'a> SuggestDao<'a> {
         )?;
         Ok(())
     }
+    pub fn put_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
```
linabutler, line 372:
```suggestion

    pub fn put_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
```

components/suggest/src/store.rs
```diff
@@ -49,6 +57,35 @@ pub struct SuggestStore {
...
+        Ok(ToSqlOutput::from(serde_json::to_string(self).map_err(
+            |err| rusqlite::Error::ToSqlConversionFailure(Box::new(err)),
+        )?))
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct UnparsableRecord {
+    pub(crate) schema_version: u32,
+}
```
linabutler, line 87:
```suggestion
pub(crate) struct UnparsableRecord {
    pub schema_version: u32,
}
```

nitpick: I like to mark crate-internal types as `pub(crate)`, so that it's a bit easier to tell what's exposed to consumers, and what's part of the crate's implementation.

For fields and methods _within_ a `pub(crate)` type, though, I usually go with `pub`—Rust won't let `pub(crate)` types leak outside of the crate, so it's OK to use a more permissive (and shorter! 😊) visibility specifier for its fields.

components/suggest/src/store.rs
```diff
@@ -176,6 +213,17 @@ where
             // was interrupted, we'll pick up where we left off.
             options.gt("last_modified", last_ingest.to_string());
         }
+
+        if let Some(unknown_records) =
+            writer.read(|dao| dao.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY))?
+        {
+            for (record_id, unparasble_record) in unknown_records.0 {
+                if unparasble_record.schema_version < VERSION {
```
linabutler, line 221:
```suggestion
            for (record_id, unparsable_record) in unknown_records.0 {
                if unparsable_record.schema_version < VERSION {
```

components/suggest/src/store.rs
```diff
@@ -176,6 +213,17 @@ where
...
+                    options.eq("id", record_id);
```
linabutler, line 222: issue (blocking): Hmmm, unfortunately, I don't think this will do what we want 😞 Sorry I didn't catch this earlier!

From peeking at the [API docs](https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html) and [code](https://github.com/Kinto/kinto/blob/6edf6453033e0106410fe1f8c70323b6fea2f2fe/kinto/core/resource/model.py#L108-L111), Kinto seems to always combine filters with `AND`.

So, if we have any unparsable records, these options will tell Kinto, "fetch all records with these unparsable IDs that also changed since the last ingest". But what we really want is "fetch all records with these unparsable IDs"—regardless of the last time they changed.

So I think we might need to make two requests for records: one to backfill the unparsable ones, and the second one for any that changed since the last ingest.

We'll process both sets of records in exactly the same way—so it might be worth factoring out the `for record in &records` loop into a separate method on the store—but I didn't see a way that we can request them all at once from Kinto.

linabutler (changes requested):
Almost there!

components/suggest/src/store.rs
linabutler: Another benefit of doing this in multiple requests: we might need to limit the number of IDs that we include in `options` for each request, so that we don't run over the URL length limit.

It's easy-ish to layer that on, though, once we have the request working! The flow of logic would change from something like this:

```rs
let all_unparsable_ids = unknown_records
    .0
    .iter()
    .filter(|(_, unparsable_record)| unparsable_record.schema_version < VERSION)
    .map(|(record_id, _)| record_id)
    .collect::<Vec<_>>();

let mut options = GetItemsOptions::new();
for record_id in all_unparsable_ids {
    options.eq("id", record_id);
}

// ... Make the request, and re-ingest all the records.
```

To this:

```rs
let all_unparsable_ids = unknown_records
    .0
    .iter()
    .filter(|(_, unparsable_record)| unparsable_record.schema_version < VERSION)
    .map(|(record_id, _)| record_id)
    .collect::<Vec<_>>();

for unparsable_ids in all_unparsable_ids.chunks(150) {
    let mut options = GetItemsOptions::new();
    for unparsable_id in unparsable_ids {
        options.eq("id", unparsable_id);
    }
    // ... Make the request, and re-ingest up to 150 of the unparsable records.
}
```

components/suggest/src/store.rs
tiftran: lol i really do suck at spelling 😓

components/suggest/src/db.rs
```diff
@@ -365,4 +369,42 @@ impl<'a> SuggestDao<'a> {
         )?;
         Ok(())
     }
+
+    /// update last ingest if record last modified is newer than the existing one recorded
```
linabutler, line 373: nitpick: Let's follow the [Rustdoc conventions](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text) for docstrings, and make this a complete sentence.

```suggestion
    /// Update the last ingest timestamp if the given last modified time is newer than the existing one recorded.
```

components/suggest/src/db.rs
```diff
@@ -365,4 +369,42 @@ impl<'a> SuggestDao<'a> {
...
+            .get_meta::<u64>(LAST_INGEST_META_KEY)?
+            .unwrap_or_default();
+        if record_last_modified > last_ingest {
+            self.put_meta(LAST_INGEST_META_KEY, record_last_modified)?;
+        }
+
+        Ok(())
+    }
+
+    /// insert unknown record into the unknown records meta field
```
linabutler, line 385: nitpick: Let's make this docstring a complete sentence, too, please.

linabutler, line 385: nitpick: Let's make this docstring a complete sentence, too, please.

components/suggest/src/db.rs
```diff
@@ -365,4 +369,42 @@ impl<'a> SuggestDao<'a> {
...
+            record_id.as_str().to_string(),
+            UnparsableRecord {
+                schema_version: VERSION,
+            },
+        );
+        self.put_meta(UNKNOWN_RECORDS_META_KEY, unknown_records)?;
+        Ok(())
+    }
+
+    pub fn drop_unknown_record_id(&mut self, record_id: &SuggestRecordId) -> Result<()> {
```
linabutler, line 400: nitpick: Docstring for this function, too, please.

components/suggest/src/store.rs
```diff
@@ -166,6 +206,29 @@ where
...
+        if let Some(unknown_records) =
+            writer.read(|dao| dao.get_meta::<UnparsableRecords>(UNKNOWN_RECORDS_META_KEY))?
+        {
+            let all_unparsable_ids = unknown_records
+                .0
+                .iter()
+                .filter(|(_, unparsable_record)| unparsable_record.schema_version < VERSION)
+                .map(|(record_id, _)| record_id)
+                .collect::<Vec<_>>();
+            for unparsable_ids in all_unparsable_ids.chunks(150) {
```
linabutler, line 218: suggestion: Let's put this `150` into a constant, please (and give it a docstring), so that our future selves aren't left wondering what it means. We could call it something like `UNPARSABLE_IDS_PER_REQUEST`—how does that sound to you?

components/suggest/src/store.rs
```diff
@@ -187,6 +251,12 @@ where
             .settings_client
             .get_records_with_options(&options)?
             .records;
+        self.ingest_records(records, writer)?;
+
+        Ok(())
+    }
+
+    fn ingest_records(&self, records: Vec<RemoteSettingsRecord>, writer: &SuggestDb) -> Result<()> {
```
linabutler, line 259: suggestion: This looks wonderful, thank you so much for factoring this out! 🚀 A couple of suggestions for the signature:

1. Let's flip the order of the arguments around, so that the writer comes first. In general, I like to order arguments in order of "specificity" (maybe there's a better word to describe that; it's the closest one I could think of! 😅)—first `self` (the store), then `writer` (that we got from the store, and that we need to ingest), then the `records` themselves. How does that sound to you?
2. I think we can take a `&[RemoteSettingsRecord]` instead of a `Vec<RemoteSettingsRecord>`. It doesn't look like this function needs an owning reference to the records—the first line does a `for record in &records`, so it immediately borrows it, anyway—and taking a slice would help clarify the ownership semantics. I think the conventions in Rust tend to be "take ownership if you need it, but borrow if you don't".

```suggestion
    fn ingest_records(&self, writer: &SuggestDb, records: &[RemoteSettingsRecord]) -> Result<()> {
```

components/suggest/src/store.rs
```diff
@@ -1567,7 +1651,198 @@ mod tests {
...
+            }]),
+        )?;
+
+        let store = unique_test_store(SnapshotSettingsClient::with_snapshot(snapshot));
+
+        store.ingest(SuggestIngestionConstraints::default())?;
+
+        store.dbs()?.reader.read(|dao| {
+            // Unknown records should still advance the last ingest time, but
+            // we don't try to store or ingest any suggestions from them.
```
linabutler, line 1722: We do now! 😁

```suggestion
```

components/suggest/src/store.rs
```diff
@@ -1567,7 +1651,198 @@ mod tests {
             // Unknown records should still advance the last ingest time, but
             // we don't try to store or ingest any suggestions from them.
```
linabutler, line 1652:
```suggestion
```

components/suggest/src/store.rs
```diff
@@ -1567,7 +1651,198 @@ mod tests {
...
+        store.dbs()?.writer.write(|dao| {
+            dao.put_meta(UNKNOWN_RECORDS_META_KEY, initial_data)?;
+            Ok(())
+        })?;
+
+        store.ingest(SuggestIngestionConstraints::default())?;
+
+        store.dbs()?.reader.read(|dao| {
+            // Unknown records should still advance the last ingest time, but
+            // we don't try to store or ingest any suggestions from them.
```
linabutler, line 1830:
```suggestion
```

components/suggest/src/store.rs
```diff
@@ -1567,7 +1651,198 @@ mod tests {
...
+                "title": "Los Pollos Hermanos - Albuquerque",
+                "url": "https://www.lph-nm.biz",
+                "icon": "5678",
+                "impression_url": "https://example.com/impression_url",
+                "click_url": "https://example.com/click_url",
+            }]),
+        )?;
+
+        let store = unique_test_store(SnapshotSettingsClient::with_snapshot(snapshot));
+        let mut initial_data = UnparsableRecords::default();
```
linabutler, line 1813: These are fantastic tests, thank you for adding them! 🏆

linabutler (dismissed):
One more quick round of suggestions and nits, and then this looks perfect for landing! Thanks @tiftran!

linabutler (approved):
Thanks so much, @tiftran, let's get this landed! 🚀

I pushed two more commits—they're small enough to where I didn't think it was worth another round of feedback, especially since you're at RustConf this week! 🦀

The first one is a find-and-replace that standardizes on the "unparsable records" terminology. We were using a mix of "unknown" and "unparsable" before, and I think "unparsable" is much more descriptive—and it avoids confusion of describing the same thing in two different ways.

The second bumps the schema version, which will make our consumers throw away their databases, and re-download all records again. Except this time, instead of silently ignoring any unparsable records, they'll store them in the meta table for re-ingestion.

PR #5804, opened 2023-09-05
Initial commit of nimbus-experimenter megazord build [ci full]

jhugman:
Fixes [EXP-3790](https://mozilla-hub.atlassian.net/browse/EXP-3790).

This PR adds a new megazord type: the server side megazord. Primarily, this is to get FML in experimenter, and cirrus (nimbus SDK + nimbus FML).

It is heavily based on the nimbus-build tasks, but builds cdylibs with python bindings.

  - Note:


jeddai (commented):
This looks good to me but I'm going to defer approval to someone more familiar with the build system

jhugman:
Each of @bdk, @ahal , @tarikeshaq and @jeddai has said "they shouldn't be the one to review this, try one of the others".

In which case, I'm going to suggest landing this only after getting two or three white ticks.

bendk:
> Each of @bdk, @ahal , @tarikeshaq and @jeddai has said "they shouldn't be the one to review this, try one of the others".
> 
> In which case, I'm going to suggest landing this only after getting two or three white ticks.

:rofl: 

The build changes look okay to me, but I think the best method is to just check the CI results.  Sammy is getting back from PTO tomorrow, how about waiting until then, with the hope of fixing NSS and getting `full ci` to pass?

jeddai (approved):
Went through the builds/artifacts and they look good to me! Of course we'll need to actually fully test its integration but afaik that kinda has to come after this.

taskcluster/ci/server-megazord-assemble/kind.yml
```diff
@@ -0,0 +1,47 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+# release-publish: publish artifacts for successful nightly/release builds
+#
+# This task publishes a JSON file for the nightly/release builds. The JSON file is used by consumer
+# apps for app-services version bumps.
+#
+# This task only runs if build-summary succeeds, which implies that all other
+# tasks suceeded.  This task should not fail, since that could leave the
+# build artifacts in an inconsistent state.
```
tarikeshaq, line 12: Just confirming that this comment is accurate? I don't see `build-summary` in `kind-dependencies` and from the changes to `build-summary/kind.yaml` it seems like it's the opposite, this job is a dependency of the build summary task

tarikeshaq (commented):
For the NSS artifact, you could try the documentation on https://mozilla.github.io/application-services/book/howtos/upgrading-nss-guide.html?highlight=NSS#updating-the-cross-compiled-nss-artifacts that might get you there

otherwise, I had a tiny comment and this looks good to me if folks more familiar with taskcluster think it's good!

taskcluster/scripts/setup-mac-worker.sh
```diff
@@ -7,6 +7,8 @@
 mkdir -p "$HOME/bin"
 export PATH="$HOME/bin:$HOME/Library/Python/3.7/bin:$PATH"
 
+pip3 install pytest
```
ahal, line 10: It's probably worth pinning this, otherwise pytest is going to make some backwards incompatible change at some point that breaks CI outside of a commit.

It might also be worth using a `requirements.txt` generated via `pip-compile` that includes hashes, but it's just for testing so up to you.

taskcluster/docker/linux/Dockerfile
```diff
@@ -103,9 +103,12 @@ RUN apt-get update -qq \
 
 RUN pip3 install --upgrade pip
 RUN pip3 install \
+    pytest \
     pyyaml \
     toml \
     taskcluster \
+    # For formatting uniffi generated python code
+    yapf \
```
ahal, line 111: Does this get used anywhere? Fwiw, we tend to use `black` for most Python projects at Mozilla.

taskcluster/ci/server-megazord-assemble/kind.yml
ahal: Good catch, yes this patch sets up `build-summary` to depend on this task, not the other way around. I'm not sure whether the comment or the implementation is incorrect, but one of them should be fixed :)

taskcluster/app_services_taskgraph/transforms/server_megazord.py
```diff
@@ -0,0 +1,139 @@
...
+        build_task_deps = [
+            MegazordBuildDep(label, build_task.attributes['target'])
+            for (label, build_task) in config.kind_dependencies_tasks.items()
+            if build_task.kind == "server-megazord-build" and build_task.attributes.get('megazord') == binary
+        ]
+
+        task['dependencies'] = {
+            dep.label: dep.label
+            for dep in build_task_deps
+        }
```
ahal, line 110: It would be possible to define these dependencies using Taskgraph's [from_deps](https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/transforms/from_deps.html) transforms, but it's probably not worth changing at this point unless you were keen to keep the logic here to a minimum.

taskcluster/app_services_taskgraph/transforms/server_megazord.py
```diff
@@ -0,0 +1,139 @@
...
+        task['fetches'] = {
+            dep.label: [
+                {
+                    'artifact': f'{binary}-{dep.target}.zip',
+                    'dest': binary,
+                    'extract': True, # if binary == 'nimbus-fml' else False
```
ahal, line 116: nit: forget to remove this commented out code?

taskcluster/app_services_taskgraph/transforms/server_megazord.py
ahal: `from_deps` could also handle the fetches section below I think:
https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/transforms/from_deps.html#adding-fetches

In fact this whole file could likely be defined entirely in the task yaml, but it's a matter of preference so not something that needs fixing.

ahal (changes requested):

taskcluster/scripts/setup-mac-worker.sh
jhugman: I'm not going to do this in this PR; we're using `pip install` and `gem`  in several places without hashes. This is a task on its own.

jhugman:
@ahal I've updated the PR with the suggested changes. I fixed up the nss artifact too, and now it clean builds. Would you mind another r?

taskcluster/docker/linux/Dockerfile
ahal: Thanks for the explanation! Let's just leave it as is then.

taskcluster/scripts/setup-mac-worker.sh
ahal: Sounds good, fwiw this can be as simple as changing this to `pip3 install pytest~=7.4`. But keeping scope down is reasonable.

ahal (approved):
Thanks, lgtm!

PR #5826, opened 2023-09-15
Fixes DISCO-2605: Replace template parameters in AMP suggestion URLs.

linabutler:
Some AMP suggestion URLs contain template parameters, like `%YYYYMMDDHH%`, which we must "cook" with real values before returning the suggestion. This means that an AMP suggestion's URL can vary depending on when it's shown.

When deduping suggestions, though, we want to treat all variations of these URLs as equivalent, and ignore the replaced parameters. Otherwise, the address bar might show an AMP suggestion twice, as on Desktop in bug 1735894: once from history, with one set of parameters; and once from Suggest, with a different set of parameters.

To address both of these needs, this commit adds:

* A `SuggestionUrl` type that holds both the "cooked" URL string, and the locations of replaced template parameters within that string.
* A `suggestion_url_matches` function that compares a `SuggestionUrl` with a URL string, to determine if they're identical except for their replaced parameters.

/cc @0c0w3

  - Note:

### DIFF
--- a/components/suggest/src/suggest.udl
+++ b/components/suggest/src/suggest.udl
@@ -8,13 +8,14 @@ typedef extern RemoteSettingsConfig;
 
 namespace suggest {
 
+boolean raw_suggestion_url_matches([ByRef] string raw_url, [ByRef] string url);
+
 };
 
 [Error]
 interface SuggestApiError {
     Other(string reason);
 };
-
 
 enum SuggestionProvider {
     "Amp",
@@ -26,13 +27,15 @@ interface Suggestion {
     Amp(
         string title,
         string url,
+        string? raw_url,
         sequence<u8>? icon,
         string full_keyword,
         i64 block_id,
         string advertiser,
         string iab_category,
         string impression_url,
-        string click_url
+        string click_url,
+        string? raw_click_url
     );
     Wikipedia(
         string title,
--- a/components/suggest/src/suggest.udl
+++ b/components/suggest/src/suggest.udl
@@ -27,7 +27,7 @@ interface Suggestion {
     Amp(
         string title,
         string url,
-        string? raw_url,
+        string raw_url,
         sequence<u8>? icon,
         string full_keyword,
         i64 block_id,
@@ -35,7 +35,7 @@ interface Suggestion {
         string iab_category,
         string impression_url,
         string click_url,
-        string? raw_click_url
+        string raw_click_url
     );
     Wikipedia(
         string title,
--- a/components/suggest/src/suggestion.rs
+++ b/components/suggest/src/suggestion.rs
@@ -68,7 +68,12 @@ pub fn raw_suggestion_url_matches(raw_url: &str, cooked_url: &str) -> bool {
     ) else {
         return false;
     };
-    raw_url_prefix == cooked_url_prefix && raw_url_suffix == cooked_url_suffix
+    if raw_url_prefix != cooked_url_prefix || raw_url_suffix != cooked_url_suffix {
+        return false;
+    }
+    let maybe_timestamp =
+        &cooked_url[raw_url_prefix.len()..raw_url_prefix.len() + TIMESTAMP_LENGTH];
+    maybe_timestamp.bytes().all(|b| b.is_ascii_digit())
 }
 
 #[cfg(test)]
@@ -146,6 +151,16 @@ mod tests {
         assert!(!raw_suggestion_url_matches(
             raw_url_with_trailing_segment,          // `a&b`.
             "https://example.com?a=5555555555&c=c"  // `a&c`.
+        ));
+
+        // Not a timestamp.
+        assert!(!raw_suggestion_url_matches(
+            raw_url_with_one_timestamp,
+            "https://example.com?a=bcdefghijk"
+        ));
+        assert!(!raw_suggestion_url_matches(
+            raw_url_with_trailing_segment,
+            "https://example.com?a=bcdefghijk&b=c"
         ));
     }
 
--- a/components/suggest/src/suggestion.rs
+++ b/components/suggest/src/suggestion.rs
@@ -3,9 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-use std::fmt::Write;
-
-use chrono::{Datelike, Local, Timelike};
+use chrono::Local;
 
 /// The template parameter for a timestamp in a "raw" sponsored suggestion URL.
 const TIMESTAMP_TEMPLATE: &str = "%YYYYMMDDHH%";
@@ -22,7 +20,7 @@ pub enum Suggestion {
     Amp {
         title: String,
         url: String,
-        raw_url: Option<String>,
+        raw_url: String,
         icon: Option<Vec<u8>>,
         full_keyword: String,
         block_id: i64,
@@ -30,7 +28,7 @@ pub enum Suggestion {
         iab_category: String,
         impression_url: String,
         click_url: String,
-        raw_click_url: Option<String>,
+        raw_click_url: String,
     },
     Wikipedia {
         title: String,
@@ -48,30 +46,11 @@ impl Suggestion {
 }
 
 /// Replaces all template parameters in a "raw" sponsored suggestion URL,
-/// producing a "cooked" URL with real values. Returns `None` if the raw
-/// URL doesn't contain any template parameters.
-pub(crate) fn cook_raw_suggestion_url(raw_url: &str) -> Option<String> {
-    let now = Local::now();
-    let mut cooked_url = String::new();
-    let mut last_index = 0;
-    for (index, _) in raw_url.match_indices(TIMESTAMP_TEMPLATE) {
-        cooked_url.push_str(&raw_url[last_index..index]);
-        write!(
-            &mut cooked_url,
-            "{:04}{:02}{:02}{:02}",
-            now.year(),
-            now.month(),
-            now.day(),
-            now.hour()
-        )
-        .expect("Failed to replace timestamp template parameter");
-        last_index = index + TIMESTAMP_TEMPLATE.len();
-    }
-    if cooked_url.is_empty() {
-        return None;
-    }
-    cooked_url.push_str(&raw_url[last_index..]);
-    Some(cooked_url)
+/// producing a "cooked" URL with real values.
+pub(crate) fn cook_raw_suggestion_url(raw_url: &str) -> String {
+    let replacement = Local::now().format("%Y%m%d%H").to_string();
+    debug_assert!(replacement.len() == TIMESTAMP_LENGTH);
+    raw_url.replace(TIMESTAMP_TEMPLATE, &replacement)
 }
 
 /// Determines whether a "raw" sponsored suggestion URL is equivalent to a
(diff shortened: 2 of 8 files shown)


0c0w3:
Thanks @linabutler. Desktop stores the original URL alongside the cooked URL (thanks for that term) in the suggestion object so that we can quickly and easily store and look up dismissed suggestions. It looks like clients can use the replacements concept in this PR to produce uncooked URLs to accomplish the same thing, but that's not quite as convenient. wdyt?

components/suggest/src/suggest.udl
```diff
@@ -8,35 +8,46 @@ typedef extern RemoteSettingsConfig;
...
-
 enum SuggestionProvider {
     "Amp",
     "Wikipedia",
 };
 
+dictionary SuggestionUrl {
+    string string;
+    sequence<SuggestionUrlReplacement> replacements;
+};
```
linabutler, line 28: This interface is a little awkward because of mozilla/uniffi-rs#1372. At first, I modeled `SuggestionUrl` as an object, like:

```webidl
interface SuggestionUrl {
    string string();
    boolean matches([ByRef] string other_url_string);
};
```

But `Suggestion` is an enum with associated data, and those can't have object fields yet. Even if they could, though, `SuggestionUrl` would need an explicit destructor in Kotlin (mozilla/uniffi-rs#1394).

Another idea I considered was turning `Suggestion` into an "enum-like" object with optional-returning getters for each variant, like this:

```webidl
interface Suggestion {
    AmpSuggestion? as_amp_suggestion();
    WikipediaSuggestion? as_wikipedia_suggestion();
};

dictionary AmpSuggestion {
    string title;
    SuggestionUrl url;
    sequence<u8>? icon;
    // ...
};

dictionary WikipediaSuggestion {
    string title;
    SuggestionUrl url;
    // ...
};
```

But that makes `Suggestion` feel clunkier, and forces explicit destructors for `Suggestion` _and_ `SuggestionUrl` in Kotlin.

Also, UniFFI dictionaries can't have methods, so that's why `suggestion_url_matches` is a free function instead of a method on `SuggestionUrl`.

/cc @bendk @mhammond Do you folks have any suggestions (😅) for other things to try, or does this interface look okay to you, given UniFFI's restrictions on it?

linabutler:
Thanks @0c0w3!

> Desktop stores the original URL alongside the cooked URL (thanks for that term) in the suggestion object so that we can quickly and easily store and look up dismissed suggestions. It looks like clients can use the replacements concept in this PR to produce uncooked URLs to accomplish the same thing, but that's not quite as convenient. wdyt?

Ooh, thanks for mentioning dismissed suggestions—I was so focused on equivalence that I didn't think about that being another case where we'd need the original "raw" URL, oops! 😬

I wonder, do you think we could get away with passing just the raw and cooked URLs back, and drop the "replacements" concept (and not worry about implementing equivalence in the Rust component?) altogether?

It would definitely simplify `SuggestionUrl`:

```webidl
interface SuggestionUrl {
    string cooked;
    string raw;
};
```

0c0w3:
Yes, that's how I would do it. Equivalence testing is still useful though, especially for deduping as you mentioned. Since it's the Rust component that will be cooking these URLs -- it's the thing that knows which substrings need replacing and how to replace them -- IMO it should provide the equivalence test too. The `suggestion_url_matches` free function looks good to me, and desktop implements a similar function by storing another property on each AMP suggestion object, `urlTimestampIndex` (the timestamp length is assumed to be constant).

That approach may or may not be scalable to more or different types of templates, I'm not sure. Actually, a related thing that came up recently is UTM params. Only a couple suggestion types use them and they never change, so we've lucked out so far and we don't have any special handling for these at all, but if they ever do change, like timestamp templates do, we'd need to handle them too.

components/suggest/src/suggest.udl
bendk: I can't really think of anything better.  This makes me want to address 1372 and let users configure if they want to treat their interface as a resource or not.

linabutler:
Great, thanks so much, @0c0w3! Here's what I changed in the latest push:

* Dropped `SuggestionUrl` and `SuggestionUrlReplacement` 🗑️
* Added `raw_url` and `raw_click_url` fields for `Amp` suggestions. I made these optional at first, and only set when the raw URL != the cooked URL, but then changed them to be set always. Please let me know if you'd prefer it the first way, and I can drop the second commit 😊
* Renamed `suggestion_url_matches` to `raw_suggestion_url_matches`. Functionally, it works the same way, and compares the non-template segments of the raw and cooked URLs.

`raw_suggestion_url_matches` makes the same assumption as Desktop that the timestamp length is constant, and that it's the only param. We've been trying to avoid taking a dependency on the `regex` crate, since it impacts our binary size by quite a bit—but I think we can still extend `raw_suggestion_url_matches` to do variable-length matching if and when we need it for other params.

components/suggest/src/suggest.udl
```diff
@@ -26,13 +27,15 @@ interface Suggestion {
     Amp(
         string title,
         string url,
+        string raw_url,
```
linabutler, line 30: note: This is a source breaking change for Swift consumers (which we don't have, yet! 🤫), but not Kotlin or JS.

0c0w3 (approved):
Thanks, lgtm! The desktop impl has a couple optimizations over this I thought I'd mention in case you're interested.

First, it compares cooked URLs to other cooked URLs, so right off the bat it can return false if the two URLs aren't the same length, short circuiting the later substring operations. (A side benefit of comparing cooked URLs is it makes the raw URL less a part of the interface and more of an internal detail, although consumers end up needing the raw URL anyway to handle dismissals.)

Second, when it initially replaces a suggestion's templates, it also stores their indexes in the suggestion object. That lets the matching function avoid another full-string search for templates, keeping the function on the order of the number of segments in the URLs instead of their length.

Are these optimizations important? Maybe not. At most N calls to the matching function happen per keystroke, N = the number of history results alongside the Suggest result, and for desktop N is always a very small number, less than 10 by default.

linabutler:
That makes perfect sense! I think my original implementation was closer to Desktop's—`SuggestionUrl` held the cooked URL, and the positions of any replacements in it.

After thinking about it, though, I wasn't sure if the benefit of avoiding the linear scan was enough to overcome the other costs. The biggest one is that exposing `SuggestionUrl` as a dictionary would add a UniFFI serialization toll, which we'd pay twice: once when returning the `Suggestion` from `query`, and a second time when calling the (since-renamed) `suggestion_url_matches` function.

In addition to serialization, there's a bit of additional complexity to validate that the replacement positions are valid. Because `SuggestionUrl` is a dictionary, all its fields become part of our component's public interface—so, even though we don't expect our consumers to ever create their own `SuggestionUrl`s, we'd still need to handle potentially invalid positions.

I think the API that I'd really love to have—and that would also solve the issue of exposing replacement positions over the FFI—is a `SuggestionUrl` object, rather than a dictionary. The object could encapsulate the "raw" URL and template replacement positions, so that consumers wouldn't need to know about them at all. IOW, something like:

```webidl
interface SuggestionUrl {
    // Returns the URL as a "cooked" string, with all template strings replaced
    // with real values.
    string string();

    // Returns `true` if this URL is equivalent to a "cooked" `other_url`.
    // Internally, this would compare this URL's cooked string to `other_url`,
    // then compare this URL's "raw" string to `other_url`, ignoring any
    // replaced templates.
    boolean matches(string other_url_string);
};

[Enum]
interface Suggestion {
    Amp(
        string title,
        SuggestionUrl url,
        SuggestionUrl click_url
        /* ...Other fields elided... */
    );
    Wikipedia(
        string title,
        SuggestionUrl url
        /* ...Other fields elided... */
    );
};

interface SuggestStore {
    /* ...Other store methods elided... */

    // Marks a suggestion URL as dismissed. Internally, this would get the
    // "raw" URL from the `url`, and mark that as dismissed.
    void dismiss(SuggestionUrl url);
};
```

...But I think we'd need to wait for mozilla/uniffi-rs#1372 (and mozilla/uniffi-rs#1394) to have that.

0c0w3:
Thanks for your reply! That interface does look nice. I hadn't considered that the suggest component would keep track of dismissed suggestions too -- I was thinking each platform would need to do it -- but that makes perfect sense and it would certainly make my job easier. :-)

components/suggest/src/suggestion.rs
```diff
@@ -31,3 +46,184 @@ impl Suggestion {
...
+        last_index = index + TIMESTAMP_TEMPLATE.len();
+    }
+    cooked_url.push_str(&raw_url[last_index..]);
+    cooked_url
+}
+
+/// Determines whether a "raw" sponsored suggestion URL is equivalent to a
+/// "cooked" URL. The two URLs are equivalent if they are identical except for
+/// their replaced template parameters, which can be different.
+pub fn raw_suggestion_url_matches(raw_url: &str, cooked_url: &str) -> bool {
```
ncloudioj, line 76: Likewise, if we leverage the invariant of at-most-1-template-for-each-url, this can be simplified as well.

components/suggest/src/suggestion.rs
```diff
@@ -3,19 +3,34 @@
...
+/// 2 bytes shorter than [`TIMESTAMP_TEMPLATE`].
+const TIMESTAMP_LENGTH: usize = 10;
+
 /// A suggestion from the database to show in the address bar.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub enum Suggestion {
     Amp {
         title: String,
         url: String,
+        raw_url: String,
```
ncloudioj, line 25: Worth noting that there can be no template in the raw urls for certain advertisers, meaning that `url` and `raw_url` will be the same for those. Though the vast majority of suggestions do have that, so I think we can live with that tiny overhead for now.

ncloudioj (approved):
Looks good, r+wc.

components/suggest/src/suggestion.rs
```diff
@@ -31,3 +46,184 @@ impl Suggestion {
...
     }
 }
+
+/// Replaces all template parameters in a "raw" sponsored suggestion URL,
+/// producing a "cooked" URL with real values.
+pub(crate) fn cook_raw_suggestion_url(raw_url: &str) -> String {
+    let now = Local::now();
+    let mut cooked_url = String::new();
+    let mut last_index = 0;
+    for (index, _) in raw_url.match_indices(TIMESTAMP_TEMPLATE) {
```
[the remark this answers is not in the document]
linabutler: Replacing just the first occurrence sounds good to me!

> If the partner breaks that contract by including more than one template, we should reject it at the ingestion phase and let them fix that. WDYT?

Agreed—just to make sure we're on the same page, though, I think it would be valuable to do this when we ingest into Remote Settings, not when we ingest from Remote Settings into our SQLite database.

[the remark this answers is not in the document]
ncloudioj: Yes, I meant enforcing that at the ingestion phase before loading data into Remote Settings. Oh, right. We also have another ingestion here from RS to SQLite, but that's not what was referring to, sorry about that. 😅

components/suggest/src/suggestion.rs
linabutler: Definitely! I made these `Optional<String>` at first, and only passed them if `url` and `raw_url` are different—but I think always passing them makes things a bit easier for consumers, at the cost of some overhead for us. Otherwise, our consumers would need to do something like `suggestion.rawUrl ?? suggestion.url` to get at the raw URL.

components/suggest/src/suggestion.rs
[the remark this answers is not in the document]
linabutler: No worries, I figured that's what you meant, and wanted to double-check!

PR #5829, opened 2023-09-20
Use the taskgraph `decision` image instead of `decision-mobile`

rvandermeulen:


rvandermeulen:
This is NPOTB for A-S, just changes the image that Taskcluster Decision Tasks run on.

jcristau (approved):
Thanks!

PR #5832, opened 2023-09-21
[DISCO-2505] Add support for ingesting AMO suggestions

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -347,6 +347,28 @@ where
                         Ok(())
                     })?;
                 }
+                SuggestRecord::Amo => {
+                    let Some(attachment) = record.attachment.as_ref() else {
+                        writer.write(|dao| dao.put_last_ingest_if_newer(record.last_modified))?;
+                        continue;
+                    };
+
+                    let attachment: SuggestAttachment<_> = serde_json::from_slice(
+                        &self.settings_client.get_attachment(&attachment.location)?,
+                    )?;
+
+                    writer.write(|dao| {
+                        dao.drop_suggestions(&record_id)?;
+
+                        dao.insert_amo_suggestions(&record_id, attachment.suggestions())?;
+
+                        dao.drop_unparsable_record_id(&record_id)?;
+
+                        dao.put_last_ingest_if_newer(record.last_modified)?;
+
+                        Ok(())
+                    })?;
+                }
             }
         }
         Ok(())
@@ -1360,6 +1382,18 @@ mod tests {
                 "hash": "",
                 "size": 0,
             },
+
+        }, {
+            "id": "data-2",
+            "type": "amo-suggestions",
+            "last_modified": 15,
+            "attachment": {
+                "filename": "data-2.json",
+                "mimetype": "application/json",
+                "location": "data-2.json",
+                "hash": "",
+                "size": 0,
+            },
         }, {
             "id": "icon-2",
             "type": "icon",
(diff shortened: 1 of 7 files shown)


components/suggest/src/store.rs
```diff
@@ -344,6 +344,42 @@ where
...
                         dao.drop_unparsable_record_id(&record_id)?;
 
+                        Ok(())
+                    })?;
+                }
+                SuggestRecord::Amo => {
+                    let Some(attachment) = record.attachment.as_ref() else {
+                        // An Amo record should always have an
+                        // attachment with suggestions. If it doesn't, it's
+                        // malformed, so skip to the next record.
```
linabutler, line 354:
```suggestion
                        // An Amo record should always have an
                        // attachment with suggestions. If it doesn't, it's
                        // malformed, so skip to the next record.
```

nitpick, style: I think we can drop the comments (and extra blank lines) in this new block, since they're the same as the block above.

This is totally a judgment call, and I'm wondering how you feel about it! My thinking is, because your new logic is so similar to what's just above it, and we already explain how that code works—the duplicated comments in this arm don't tell a reader anything new. If I were reading this code for the first time, I'd be thinking, "the code structure looks super similar to what I just read, but the spacing and comments here give me the feeling I should be paying closer attention".

But that's me. What do you think? 😊

You mentioned the duplication on Zoom earlier, too, and that got me thinking if it would be worth DRYing out this logic. But I don't think we're there just yet. While it is really similar to AMP-Wikipedia, the duplication isn't too bad now, and it's the kind of thing that's great to do when you're pretty sure you've found the generalization.

I _just_ saw [this post](https://programmingisterrible.com/post/176657481103/repeat-yourself-do-more-than-one-thing-and) earlier today, and thought you might like it, too! 😁 I feel like we're at the "repeat yourself to find abstractions" stage now—and the next time we go to add another suggestion type, we can see if we're following the same pattern, and DRY it out then. IME, it's usually been much easier to add abstractions later, than to undo a premature one. WDYT?

components/suggest/src/store.rs
```diff
@@ -1387,6 +1439,20 @@ mod tests {
...
+                "description": "amo suggestion",
+                "url": "https://addons.mozilla.org/en-US/firefox/addon/example",
+                "guid": "{b9db16a4-6edc-47ec-a1f4-b86292ed211d}",
+                "keywords": ["relay", "spam", "masking", "alias"],
+                "title": "Firefox Relay",
+                "icon": "https://addons.mozilla.org/user-media/addon_icons/2633/2633704-64.png?modified=2c11a80b",
+                "rating": "4.9",
+                "number_of_ratings": 888,
+                "score": 0.25
+            },]),
```
linabutler, line 1454: nitpick: `cargo fmt` doesn't (can't?) do a great job with macros, unfortunately—but could you please reformat this to fix the indentation in the `json!(...)`, closer to how we do `data-1.json` above?

components/suggest/src/suggest.udl
```diff
@@ -40,6 +41,16 @@ interface Suggestion {
         sequence<u8>? icon,
         string full_keyword
     );
+    Amo(
+        string title,
+        string url,
+        string icon,
```
linabutler, line 47:
```suggestion
        string icon_url,
```

suggestion: Ditto about `icon_url` here, to match our other `_url`-suffixed names for URLs.

linabutler (changes requested):
Thanks @tiftran! The core logic looks great—thank you for the test, too! 🏆—and I left mostly style and naming nits, plus the schema version bump. Almost there!

linabutler (approved):
Perfect!

linabutler:
@tiftran Oops, sorry, I think this needs a rebase! Would you mind squashing your commits while you're there, too, please? 🙏🏼

linabutler (approved):

PR #5841, opened 2023-09-25
[DISCO-2503] Suggest: Pocket suggestion ingestion

tiftran:

  - Note:


linabutler:
Thank you so much, Tif—and thanks for making the changes since yesterday. Gentle reminder to bump the schema version and migration function in `schema.rs` again 😊 I'll do another review pass later today!

I tested your patch out against the real dataset in Remote Settings, and saw some unique constraint errors on both `pocket_high_confidence_keywords` and `pocket_low_confidence_keywords`. There are only a handful of duplicates, but they are there:

```
sqlite> select keyword, group_concat(suggestion_id, ',') from pocket_high_confidence_keywords group by keyword having count(*) > 1;
african billionaires|71,71
april fools day|7,311
arthur brooks column|24,24
arthur c brooks|24,24
arthur c brooks column|24,24,24,24
budget-friendly sustainability|247,247
cheesy corn dip|26,26
dallas cowboy cheerleaders|111,111
festive vegan dishes|17,275
how to focus|119,119
how to get clearer skin|231,231
how to recognize burnout|232,232
how to write anything|168,168
making new friends as adults|251,251
skin maintenance|231,231
sustainable habits|247,247
the atlantic arthur brooks|24,24
```

(I dropped the `PRIMARY KEY` definition from both tables to get a database that I could query to build that list). 

The ones with different suggestion IDs—`april fools day`, `festive vegan dishes`—shouldn't be a problem, because the Pocket keywords tables have `PRIMARY KEY (keyword, suggestion_id)`, so it's OK for different suggestions to have the same keyword. We might need to do an `ORDER BY score` in the `SELECT` for `fetch_by_keyword` to break ties, though, or drop the `LIMIT 1`.

But the ones with multiple copies of the same suggestion ID are more interesting. It really looks like their keywords arrays have duplicates!

```text
DownloadedPocketSuggestion {
  description: "The best advice for finding your people, staying close, and getting through the hard parts.",
  url: "https://getpocket.com/collections/the-grown-ups-guide-to-making-and-keeping-friends",
  title: "The Grown-Up’s Guide to Making and Keeping Friends",
  low_confidence_keywords: ["what to do if you're lonely", "dealing with loneliness", "lonely what to do", "what to do lonely"],
  high_confidence_keywords: ["how to make friends as adults", "making new friends as adults", "how to make friends as an adult", "making new friends as adults", "making new friends as an adult", "how to make friends", "making friendships", "keeping friendships", "friendship maintenance", "adult social networks", "new friends", "adult social life", "making friends", "making new friends", "how to make new friends", "how to keep friends", "how to maintain friendships", "maintain friendships", "maintaining friendships", "adult friendships", "socializing tips", "nurturing friendships", "friendship longevity", "friendships in adulthood", "grown up friendships", "grown-up friendships"],
  score: 0.25 }
```

We can fix that by using either a `HashSet<String>` instead of a `Vec<String>` for the keywords lists in `DownloadedPocketSuggestion`, or an `INSERT OR IGNORE` when we insert them into the tables—but I wonder if we should take care of this in `quicksuggest-rs`, during ingestion instead. WDYT?

/cc @ncloudioj @0c0w3 for visibility, too!

ncloudioj:
> but I wonder if we should take care of this in quicksuggest-rs, during ingestion instead. WDYT?

Yes, we can easily dedup keywords for each suggestion at the ingestion phase over [here](https://github.com/mozilla-services/merino-py/blob/main/merino/jobs/csv_rs_uploader/base.py#L45-L50) (BTW, Pocket suggestions are ingested by [this Merino job](https://github.com/mozilla-services/merino-py/tree/main/merino/jobs/csv_rs_uploader) instead of quickssugest-rs). Do we also want to do deduping among low and high-confidence keywords for each suggestion?

@0c0w3 What do you think?

components/suggest/src/store.rs
```diff
@@ -344,6 +344,42 @@ where
...
+                        // attachment.
+                        dao.insert_pocket_suggestions(&record_id, attachment.suggestions())?;
+
+                        // Remove this record's ID from the list of unparsable
+                        // records, since we understand it now.
+                        dao.drop_unparsable_record_id(&record_id)?;
+
+                        // Advance the last fetch time, so that we can resume
+                        // fetching after this record if we're interrupted.
+                        dao.put_last_ingest_if_newer(record.last_modified)?;
```
ncloudioj, line 381: Looks like this sequence of (drop-insert-drop-put) has been used for all the suggestion types, shall we wrap them up into another helper in DAO to keep `ingest_records()` DRY?

components/suggest/src/schema.rs
```diff
@@ -43,6 +57,12 @@ pub const SQL: &str = "
         ON DELETE CASCADE
     );
 
+    CREATE TABLE pocket_custom_details(
+        suggestion_id INTEGER PRIMARY KEY REFERENCES suggestions(id) ON DELETE CASCADE,
+        description TEXT NOT NULL,
```
ncloudioj, line 62: Looks like `description` is not included in the Pocket suggestion & the udl below, also looks like Firefox doesn't use it for now. Shall we exclude it from the schema here as well? We can always add it back later if needed.

ncloudioj (commented):
Looks good, added a few comments for your consideration.

0c0w3:
Wow it's great to see so much progress being made on the Rust implementation.

Yeah the Pocket keywords were hand-edited by a group of people and it's entirely possible some keyword lists contain duplicates. I agree we should remove the duplicates at ingestion time, which we can do as you said, @ncloudioj. I can take care of that.

> Do we also want to do deduping among low and high-confidence keywords for each suggestion?

It looks like this isn't a problem for the Rust implementation (please correct me if I'm wrong), so I would vote for not worrying about it. Arguably this type of duplication isn't wrong; imagine if users could turn off high-confidence keywords and the related UI but keep the low-confidence ones, for example.

0c0w3:
> Yeah the Pocket keywords were hand-edited by a group of people and it's entirely possible some keyword lists contain duplicates. I agree we should remove the duplicates at ingestion time, which we can do as you said, @ncloudioj. I can take care of that.

Should be fixed now.

ncloudioj:
> Arguably this type of duplication isn't wrong; imagine if users could turn off high-confidence keywords and the related UI but keep the low-confidence ones, for example.

Sounds good, thanks for confirming! @0c0w3 Could you briefly outline how low/high-confidence keywords work in Firefox?

0c0w3:
Sure, Tif and I were chatting about this last week and here's the gist:

When the user triggers a suggestion by typing a high confidence keyword, the suggestion is shown as a top pick. When they trigger it using a low confidence keyword, it’s shown at the bottom in the usual Suggest position. Also, high confidence keywords must be typed in full, but Firefox will start matching on a low confidence keyword after its first word is typed.

References:

* [Product spec](https://docs.google.com/document/d/1r9f-yhlukrIVLJTG8cpEKNZ4JuYPnljOgt38WrQzeB4/edit?usp=sharing)
* [Desktop impl](https://searchfox.org/mozilla-central/source/browser/components/urlbar/private/PocketSuggestions.sys.mjs)
* [Figma](https://www.figma.com/file/3m1GVo9lejjorUZrPSn7VW/Pocket-collections-suggestions?type=design&node-id=3050%3A32721&mode=dev)
* To enable on desktop (118+): `browser.urlbar.pocket.featureGate` = true (also make sure Suggest is enabled, it is by default if you’re in the U.S.)

Example:

* High: `["wildlife stories"]`
* Low: `["esther the wonder pig"]`

```
Query                   | Match
------------------------+------
esthe                   | none
esther                  | low
esther (trailing space) | low
esther t                | low
esther th               | low
(etc.)                  | low
esther the wonder pig   | low
wildlif                 | none
wildlife                | none
(etc.)                  | none
wildlife storie         | none
wildlife stories        | high
```

ncloudioj:
Ah, I see. Thanks for sharing that here! And now I understand why the keyword lookup queries were made that way in this PR.

components/suggest/src/db.rs
```diff
@@ -123,14 +124,29 @@ impl<'a> SuggestDao<'a> {
...
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, false as is_top_pick
              FROM suggestions s
              JOIN keywords k ON k.suggestion_id = s.id
              WHERE k.keyword = :keyword
+             UNION
+             SELECT s.id, k.rank, s.title, s.url, s.provider, true as is_top_pick
+             FROM suggestions s
+             JOIN pocket_high_confidence_keywords k ON k.suggestion_id = s.id
+             WHERE k.keyword = :keyword
+             UNION
```
ncloudioj, line 141: If it already finds a record from the high confidence keywords table, doing another lookup against the low-confidence keywords table is unnecessary. Looks like it's not easy to inject that conditional query logic inside this query, guess we will live with that.

components/suggest/src/db.rs
```diff
@@ -123,14 +124,29 @@ impl<'a> SuggestDao<'a> {
...
+             SELECT s.id, k.rank, s.title, s.url, s.provider, false as is_top_pick
+             FROM suggestions s
+             JOIN pocket_low_confidence_keywords k ON k.suggestion_id = s.id
+             WHERE k.keyword BETWEEN :pocket_keyword AND :pocket_keyword || x'ffff'
              LIMIT 1",
```
ncloudioj, line 146: Now that we have multiple keyword sources, maybe it's time to return all the hits back (sorted by score) to the consumer instead of 1. Even returning the one with the highest score could lead to some surprises, e.g. if the user dismissed that suggestion, all other candidates for the same keyword will never get a chance to show.

Also cc @linabutler @0c0w3 for their thoughts.

components/suggest/src/db.rs
```diff
@@ -123,14 +124,29 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches suggestions that match the given keyword from the database.
     pub fn fetch_by_keyword(&self, keyword: &str) -> Result<Vec<Suggestion>> {
+        let pocket_keyword = match keyword.contains(' ') {
+            true => keyword.to_string(),
+            false => keyword.to_owned() + " ",
```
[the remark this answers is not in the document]
linabutler: Oooh, I really like Nan's suggestion of using `Cow` that way!

But I think that this logic won't handle single-word low-confidence keywords correctly. For example, I don't think that `workaholism` will match the Pocket suggestion in the test now—but it should. Is that right?

IIUC, another way to write our algorithm for matching low-confidence keywords is:

1. The head (first word) of the query must be the same as the low-confidence keyword's head.
2. The tail (anything after the first word) of the query—if any—should be a prefix of the keyword's tail.

So all these queries should match the Pocket suggestion: `workaholism`, `soft` (first word of `soft life`), `soft l`, `soft li` (first words match; `l` and `li` are prefixes of `life`).

And these shouldn't, because their first words don't match: `work`, `sof`, etc.

components/suggest/src/db.rs
0c0w3: I bet there's a way to do it, maybe:

```
UNION
SELECT IFNULL(
  (
    SELECT s.id, k.rank, s.title, s.url, s.provider, true as is_top_pick
    FROM suggestions s
    JOIN pocket_high_confidence_keywords k ON k.suggestion_id = s.id
    WHERE k.keyword = :keyword
  ),
  (
    SELECT s.id, k.rank, s.title, s.url, s.provider, false as is_top_pick
    FROM suggestions s
    JOIN pocket_low_confidence_keywords k ON k.suggestion_id = s.id
    WHERE k.keyword BETWEEN :pocket_keyword AND :pocket_keyword || x'ffff'
  )
)
```

The high and low keywords could also be stored in the same table, so you do one lookup to get the matching keyword and its type, whether it's high or low, and then a join to get the suggestion.

components/suggest/src/db.rs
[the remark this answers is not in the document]
linabutler: One way we could make this work is to store the head and tail in separate columns when we insert them into the table. Then, when we fetch suggestions, we could split the user's query the same way, and match the head and tail prefix.

Here's a quick sketch of how that all could work! Let's say we have two suggestions, and all their low-confidence keywords start with `breakfast`:

```text
title: "How to Build a Better Breakfast"
keywords: ["breakfast"]

title: Your New Favorite Breakfast Sandwich
keywords: ["breakfast crunch wrap", "breakfast torta"]
```

If we stored those suggestions and keywords in a schema like this:

```sql
CREATE TABLE suggestions(id INTEGER PRIMARY KEY, title TEXT NOT NULL);
CREATE TABLE keywords(keyword_head TEXT NOT NULL, keyword_tail TEXT NOT NULL, suggestion_id INTEGER NOT NULL REFERENCES suggestions(id), PRIMARY KEY (keyword_head, keyword_tail));

INSERT INTO suggestions(id, title) VALUES(1, 'How to Build a Better Breakfast');
INSERT INTO suggestions(id, title) VALUES(2, 'Your New Favorite Breakfast Sandwich');

INSERT INTO keywords(keyword_head, keyword_tail, suggestion_id) VALUES('breakfast', '', 1);
INSERT INTO keywords(keyword_head, keyword_tail, suggestion_id) VALUES('breakfast', 'crunch wrap', 2);
INSERT INTO keywords(keyword_head, keyword_tail, suggestion_id) VALUES('breakfast', 'torta', 2);
```

Here's how a query plan for `breakfast` would look:

```text
sqlite> SELECT s.id, s.title FROM suggestions s JOIN keywords k ON k.suggestion_id = s.id WHERE k.keyword_head = 'breakfast' AND k.keyword_tail BETWEEN '' AND '' || x'ffff' GROUP BY s.id, s.title;
QUERY PLAN
|--SEARCH k USING INDEX sqlite_autoindex_keywords_1 (keyword_head=? AND keyword_tail>? AND keyword_tail<?)
|--SEARCH s USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR GROUP BY
1|How to Build a Better Breakfast
2|Your New Favorite Breakfast Sandwich
```

(To play around with that schema in the `sqlite3` shell, I saved the schema into a file, then ran `sqlite3 -init file.sql` to load it).

Since `Your New Favorite Breakfast Sandwich` has two keywords that start with `breakfast`, we need a `GROUP BY` or a `DISTINCT` to filter out duplicates—or we'll get two rows for it, one for each of its keywords. But the search over `keywords` can use all columns of the primary key (that's the `sqlite_autoindex_keywords_1` automatic index), so the most expensive part of the query is still pretty efficient. By the time we build the temporary b-tree for the `GROUP BY`, we should have reduced the number of rows to a small handful.

The query plan for `breakfast cru` looks the same, and finds just the one suggestion:

```text
sqlite> SELECT s.id, s.title FROM suggestions s JOIN keywords k ON k.suggestion_id = s.id WHERE k.keyword_head = 'breakfast' AND k.keyword_tail BETWEEN 'cru' AND 'cru' || x'ffff' GROUP BY s.id, s.title;
QUERY PLAN
|--SEARCH k USING INDEX sqlite_autoindex_keywords_1 (keyword_head=? AND keyword_tail>? AND keyword_tail<?)
|--SEARCH s USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR GROUP BY
2|Your New Favorite Breakfast Sandwich
```

And so does `breakfast torta`:

```text
sqlite> SELECT s.id, s.title FROM suggestions s JOIN keywords k ON k.suggestion_id = s.id WHERE k.keyword_head = 'breakfast' AND k.keyword_tail BETWEEN 'torta' AND 'torta' || x'ffff' GROUP BY s.id, s.title;
QUERY PLAN
|--SEARCH k USING INDEX sqlite_autoindex_keywords_1 (keyword_head=? AND keyword_tail>? AND keyword_tail<?)
|--SEARCH s USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR GROUP BY
2|Your New Favorite Breakfast Sandwich
```

components/suggest/src/db.rs
ncloudioj: Good point, but unfortunately, both `IFNULL` and `COALESCE` in SQLite only take scalars as the arguments, selecting multiple columns will lead to a syntax error.

components/suggest/src/db.rs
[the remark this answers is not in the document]
ncloudioj: Wow, that's some really cool SQL magic there! 👏

I'd like to offer another approach: How about we move this qualifying logic over to the row handler closure?

We currently only fetch data from other tables and then assemble the suggestion accordingly in that closure. I think we can also use that as the __post-processor__ for each candidate. Specifically, the post-processor can transform, filter, and augment each row if needed. Two benefits I can see here:
- In certain cases, it's much easier to do stuff in native code than in SQL
- To keep those keywords table schema lean and generic. Imagine if we need to change the matching logic for the low-confidence keywords, we can still do prefix queries in SQL and apply more specific matching logic in the post-processor. No change is needed to the schema!

WDYT?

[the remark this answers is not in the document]
linabutler: > How about we move this qualifying logic over to the row handler closure?

Filtering in the row handler is definitely a great option, with a small result set. For example, if our query returns 5 rows, and we filter all of them out—that's OK, because we only had to scan and read 5 rows. But if our query returns hundreds or thousands of rows, filtering in the row handler isn't as efficient—we still end up scanning and reading in all those rows, and then filtering them out in code.

For low-confidence Pocket suggestions, this is a bit tricky to do naively, because we want the whole first word to match, and only match prefixes for subsequent words. If we wanted to keep the schema super generic, with a single `keyword` column, I think we'd need to do a prefix match unconditionally. So, if the user searches for `b` or `br`, we'd return suggestions for all keywords that start with `b` or `br`—and then filter out all of them, because `b` and `br` aren't words.

We could combine the approaches, through—a less generic schema, but also doing more work in Rust code instead of SQL:

1. Split each keyword into `keyword_head` and `keyword_tail` (I'm realizing `keyword_prefix` and `keyword_suffix` might actually be better names! 😅), like I suggested, and...
2. Always filter on `keyword_prefix` in the SQL query, then filter on `keyword_suffix` in the row handler, like you suggested!

I think this would also line up nicely with @0c0w3's suggestion to use a single table to store the keywords, and have a column for whether it's low or high-confidence.

So we'd have a schema like:

```sql
CREATE TABLE pocket_keywords(
    keyword_prefix TEXT NOT NULL,
    keyword_suffix TEXT NOT NULL DEFAULT "",
    confidence_type INTEGER NOT NULL,
    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id),
    PRIMARY KEY (keyword_prefix, keyword_suffix)
);
```

With the data from [Drew's example ](https://github.com/mozilla/application-services/pull/5841#issuecomment-1736521129)like:

```
keyword_prefix | keyword_suffix | confidence_type
---------------+----------------+----------------
esther         | the wonder pig | LOW
wildlife       | stories        | HIGH
```

(Assume `LOW` and `HIGH` are integer constants).

And then, in the row handler, we'd have logic like:

```rs
let suffixes_match = match row.get("confidence_type")? {
    LOW => row.get("keyword_suffix")?.starts_with(user_keyword_suffix),
    HIGH => row.get("keyword_suffix")? == user_keyword_suffix,
};
```

Let's say the user typed in `esther th`. Splitting that on the first space, the prefix is `esther`, and the suffix is `th`. The query for that would look like this:

```sql
SELECT s.*, k.keyword_suffix, k.confidence_type
FROM suggestions s
WHERE k.keyword_prefix = 'esther'
```

...Then, our row handler would see that it's a low-confidence keyword, and take the `LOW` branch. `the wonder pig` starts with `th`, so it's a match.

Now let's say they typed in `wildlife stories`. The prefix is `wildlife`; the suffix is `stories`. The query for that looks the same as for `esther th`:

```sql
SELECT s.*, k.keyword_suffix, k.confidence_type
FROM suggestions s
WHERE k.keyword_prefix = 'wildlife'
```

...Then our row handler sees that it's a high-confidence keyword, so it takes the `HIGH` branch. `stories` == `stories`, so it's a match.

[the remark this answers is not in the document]
ncloudioj: Nice, that'll work like a charm!

components/suggest/src/store.rs
```diff
@@ -348,31 +323,76 @@ where
...
+            // this record's attachment. Suggestions don't have a
+            // stable identifier, and determining which suggestions in
+            // the attachment actually changed is more complicated than
+            // dropping and re-ingesting all of them.
+            dao.drop_suggestions(&record_id)?;
+
+            // Ingest (or re-ingest) all suggestions in the
+            // attachment.
+            block(dao, &record_id, attachment.suggestions())?;
+            // dao.insert_amp_wikipedia_suggestions(&record_id, attachment.suggestions())?;
```
linabutler, line 381:
```suggestion
```

Nit: Let's remove this commented-out code.

components/suggest/src/store.rs
```diff
@@ -348,31 +323,76 @@ where
...
+    ) -> Result<()>
+    where
+        T: DeserializeOwned,
+    {
+        let record_id = SuggestRecordId::from(&record.id);
+
+        let Some(attachment) = record.attachment.as_ref() else {
+            // An AMP-Wikipedia record should always have an
+            // attachment with suggestions. If it doesn't, it's
+            // malformed, so skip to the next record.
```
linabutler, line 362:
```suggestion
            // A record should always have an attachment with suggestions.
            // If it doesn't, it's malformed, so we can't ingest it.
```

Nit: This isn't specific to AMP-Wikipedia anymore! 🎉

components/suggest/src/schema.rs
```diff
@@ -21,6 +21,13 @@ pub const SQL: &str = "
         PRIMARY KEY (keyword, suggestion_id)
     ) WITHOUT ROWID;
 
+    CREATE TABLE pocket_keywords( 
+        keyword_prefix TEXT NOT NULL, 
+        keyword_suffix TEXT NOT NULL DEFAULT '', 
+        confidence_type INTEGER NOT NULL, 
+        suggestion_id INTEGER NOT NULL REFERENCES suggestions(id), 
+        PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id) );
```
linabutler, line 29:
```suggestion
        PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id)
    );
```

Style nit.

components/suggest/src/db.rs
```diff
@@ -417,6 +470,120 @@ impl<'a> SuggestDao<'a> {
...
+                },
+            )?;
+            for (_, keyword) in suggestion.low_confidence_keywords.iter().enumerate() {
+                let (keyword_prefix, keyword_suffix) = match keyword.find(' ') {
+                    Some(index) => {
+                        let (prefix, suffix) = keyword.split_at(index);
+                        (prefix.to_string(), suffix[1..].to_string())
+                    }
+                    None => (keyword.clone(), String::new()),
+                };
```
linabutler, line 528: Let's lean on the Rust standard library here!

```suggestion
                let (keyword_prefix, keyword_suffix) = keyword.split_once(' ').unwrap_or_else(|| (keyword, ""));
```

`split_once` will do the checking and slicing for you, and returns references, so we can also avoid the extra copies with `to_string()` and `clone()`.

linabutler: These loops also look mostly identical! We can use some of Rust's iterator methods to cut down on the repetition:

```rust
for ((rank, keyword), confidence) in suggestion
    .high_confidence_keywords
    .iter()
    .enumerate()
    .zip(std::iter::repeat(PocketKeywordConfidence::High))
    .chain(
        suggestion
            .low_confidence_keywords
            .iter()
            .enumerate()
            .zip(std::iter::repeat(PocketKeywordConfidence::Low)),
    )
{
    // ...
}
```

`std::iter::repeat` returns an infinite iterator, and `zip`-ping another iterator with it will combine elements from both of them, and stop as soon as one of the iterators is exhausted.

components/suggest/src/store.rs
```diff
@@ -1664,6 +1706,64 @@ mod tests {
                 ]
                 "#]],
             ),
+            (
+                "keyword = `soft`; non-sponsored only",
```
linabutler, line 1710: Let's also add a test for `soft li`, so that we can make sure the prefix matching works.

components/suggest/src/db.rs
```diff
@@ -417,6 +470,120 @@ impl<'a> SuggestDao<'a> {
...
+            }
+
+            for (_, keyword) in suggestion.high_confidence_keywords.iter().enumerate() {
+                let (keyword_prefix, keyword_suffix) = match keyword.find(' ') {
+                    Some(index) => {
+                        let (prefix, suffix) = keyword.split_at(index);
+                        (prefix.to_string(), suffix[1..].to_string())
+                    }
+                    None => (keyword.clone(), String::new()),
+                };
```
linabutler, line 558: Same comment here about using `split_once`.

linabutler (commented):
Thanks @tiftran! Putting up a quick round of comments before breakfast; I'll do a more thorough pass later!

components/suggest/src/db.rs
```diff
@@ -124,16 +125,28 @@ impl<'a> SuggestDao<'a> {
...
-            "SELECT s.id, k.rank, s.title, s.url, s.provider
+        let (keyword_prefix, keyword_suffix) = match keyword.find(' ') {
+            Some(index) => {
+                let (prefix, suffix) = keyword.split_at(index);
+                (prefix.to_string(), suffix[1..].to_string())
+            }
+            None => (keyword.to_string(), String::new()),
+        };
+        Ok(self.conn.query_rows_and_then_cached(
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, -1 as confidence_type
```
ncloudioj, line 136: I'd recommend using `NULL` instead of `-1` for the missing field. Also, how about we move this field to `pocket_custom_details` as it's Pocket-specific? We will need to do more in the handler, but I think it's still better than introducing a dummy column to the suggestion table.

WDYT?

components/suggest/src/db.rs
```diff
@@ -216,12 +229,52 @@ impl<'a> SuggestDao<'a> {
...
                                         guid: row.get("guid")?,
                                         score: row.get("score")?,
-                                    })
+                                    }))
                             })
                     },
+                    SuggestionProvider::Pocket => {
+                        let confidence_type = row.get("confidence_type")?;
+                        self.conn.query_row_and_then(
+                            "SELECT keyword_suffix FROM pocket_keywords WHERE suggestion_id = :suggestion_id AND keyword_prefix = :keyword_prefix",
```
ncloudioj, line 238:
```suggestion
                            "SELECT k.keyword_suffix FROM pocket_keywords k
                             WHERE k.suggestion_id = :suggestion_id
                             AND k.keyword_prefix = :keyword_prefix", 
```

components/suggest/src/pocket_confidence.rs
```diff
@@ -0,0 +1,37 @@
+use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, ValueRef};
+use rusqlite::{Result as RusqliteResult, ToSql};
+
+/// Classification of Pocket confidence keywords.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[repr(u8)]
+pub enum PocketKeywordConfidence {
```
ncloudioj, line 7: Suggestion: the module already has Pocket and Confidence, we can simply call it `KeywordConfidence` or `Confidence`.

components/suggest/src/store.rs
```diff
@@ -348,31 +323,76 @@ where
...
             }
         }
         Ok(())
     }
+
+    fn ingest_suggestions_from_record<T>(
+        &self,
+        writer: &SuggestDb,
+        record: &RemoteSettingsRecord,
+        block: impl FnOnce(&mut SuggestDao<'_>, &SuggestRecordId, &[T]) -> Result<()>,
```
ncloudioj, line 352: Nit: let's call it `ingestion_handler` or the like as `block` has special meanings in the context of database ops.

components/suggest/src/db.rs
```diff
@@ -124,16 +125,28 @@ impl<'a> SuggestDao<'a> {
...
+             SELECT s.id, -1 as rank, s.title, s.url, s.provider, k.confidence_type
+             FROM suggestions s
+             JOIN pocket_keywords k  ON k.suggestion_id = s.id
+             WHERE k.keyword_prefix = :keyword_prefix",
             named_params! {
                 ":keyword": keyword,
+                ":keyword_prefix": keyword_prefix,
             },
-            |row| -> Result<Suggestion>{
+            |row | -> Result<Option<Suggestion>>{
```
ncloudioj, line 149:
```suggestion
            |row| -> Result<Option<Suggestion>> {
```

components/suggest/src/db.rs
tiftran: 😵 wow this is fancy

components/suggest/src/db.rs
```diff
@@ -216,12 +229,52 @@ impl<'a> SuggestDao<'a> {
...
+                                        },
+                                        |row| {
+                                            Ok(Some(Suggestion::Pocket {
+                                                title,
+                                                url: raw_url,
+                                                score: row.get("score")?,
+                                                is_top_pick: match confidence_type{
+                                                    PocketKeywordConfidence::Low => false,
+                                                    PocketKeywordConfidence::High => true,
+                                                }
```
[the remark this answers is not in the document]
tiftran: ohh this is fancy too

components/suggest/src/db.rs
tiftran: sooo confidence is sits in the pocket_keywords table, the -1 is used as a dummy value cause UNION requires the same number of columns. Is there a way to smush the two results together without having the same number of columns?

components/suggest/src/db.rs
tiftran: 🙈

components/suggest/src/db.rs
ncloudioj: Right, it's in the pocket_keywords table. If we don't include it here, we'd have to do another lookup against the keywords table. NVM, let's use NULL as the dummy value then.

components/suggest/src/db.rs
```diff
@@ -124,16 +125,22 @@ impl<'a> SuggestDao<'a> {
...
+        Ok(self.conn.query_rows_and_then_cached(
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence
              FROM suggestions s
              JOIN keywords k ON k.suggestion_id = s.id
              WHERE k.keyword = :keyword
-             LIMIT 1",
+             UNION ALL
+             SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence
+             FROM suggestions s
+             JOIN pocket_keywords k  ON k.suggestion_id = s.id
```
linabutler, line 137:
```suggestion
             JOIN pocket_keywords k ON k.suggestion_id = s.id
```

Style nit.

components/suggest/src/db.rs
```diff
@@ -216,12 +223,60 @@ impl<'a> SuggestDao<'a> {
...
+                        )
+                    }
                 }
-            },
-        )
+            }
+        )?.into_iter().flatten().collect())
+    }
+
+    fn split_keyword(keyword: &str) -> (&str, &str) {
```
linabutler, line 277: suggestion: Let's make this a `pub fn` in `pocket_confidence.rs`, instead of a method here—and maybe rename that module to `pocket.rs`, too!

components/suggest/src/db.rs
```diff
@@ -216,12 +223,60 @@ impl<'a> SuggestDao<'a> {
...
+                                        |row| {
+                                            Ok(Some(Suggestion::Pocket {
+                                                title,
+                                                url: raw_url,
+                                                score: row.get("score")?,
+                                                is_top_pick: matches!(
+                                                    confidence,
+                                                    KeywordConfidence::High
+                                                )
+
```
linabutler, line 261:
```suggestion
```

Style nit: stray blank line.

components/suggest/src/db.rs
```diff
@@ -216,12 +223,60 @@ impl<'a> SuggestDao<'a> {
...
+                                            }))
+                                        }
+                                    )
+                                }
+                                else{
```
linabutler, line 266:
```suggestion
                                } else {
```

Style nit.

(I'm not sure why `cargo fmt` doesn't catch these! I know it doesn't do well with macros, so maybe the `named_params!` are tripping it up in the rest of the file? 🤔)

components/suggest/src/db.rs
```diff
@@ -417,6 +472,96 @@ impl<'a> SuggestDao<'a> {
...
+                     :url
+                 )
+                 RETURNING id
+                 ",
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":provider": SuggestionProvider::Pocket,
+                    ":title": suggestion.title,
+                    ":url": suggestion.url,
+
```
linabutler, line 504:
```suggestion
```

Stray blank line.

components/suggest/src/schema.rs
```diff
@@ -6,7 +6,7 @@
 use rusqlite::{Connection, Transaction};
 use sql_support::open_database::{self, ConnectionInitializer};
 
-pub const VERSION: u32 = 5;
+pub const VERSION: u32 = 6;
```
linabutler, line 9: Nice, thanks!

Remember to update `upgrade_from`, too, please 😁

components/suggest/src/store.rs
```diff
@@ -1664,6 +1705,64 @@ mod tests {
                 ]
                 "#]],
             ),
+            (
+                "keyword = `soft l`; non-sponsored only",
+                SuggestionQuery {
+                    keyword: "soft l".into(),
```
linabutler, line 1711: Oh, I'm so sorry—I meant, let's keep both `soft` and `soft l`, not change it—just to make sure that we handle empty suffixes for low-confidence matches correctly.

linabutler (commented):
This is looking great, thanks for folding in all those changes!

components/suggest/src/schema.rs
tiftran: 🙈

components/suggest/src/db.rs
```diff
@@ -216,12 +223,42 @@ impl<'a> SuggestDao<'a> {
...
+                            )
+                        } else {
+                            Ok(None)
+                        }
+                    }
                 }
-            },
-        )
+            }
+        )?.into_iter().flatten().collect())
```
ncloudioj, line 261: Discussion: we could provide some extra convenience by sorting the returning suggestions by certain criteria, e.g. by score or rank. For example, if a keyword matches both the high confidence and the low confidence keywords for a Pocket suggestion, it'll be nice to put the high confidence one in the front.

ncloudioj (approved):
Looks great, thanks for cranking it out!
r=nanj with a few nits and comments for discussion.

components/suggest/src/pocket.rs
```diff
@@ -48,11 +48,8 @@ impl ToSql for KeywordConfidence {
 /// Return an empty string as the suffix if there is no whitespace.
 ///
 /// # Examples:
-///
-/// ```
 /// assert_eq!(split_keyword("foo"), ("foo", ""));
 /// assert_eq!(split_keyword("foo bar baz"), ("foo", "bar baz"));
-/// ```
```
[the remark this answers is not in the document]
linabutler: @tiftran Did you happen to get a build failure with the backticks? 😅

The triple backticks will also compile and run the code between them as a ["doctest"](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html). Under the hood, rustdoc builds each doctest as a separate program, so it has to import any functions and types that it uses, like this:

````
/// ```
/// # use suggest::pocket::split_keyword;
/// assert_eq!(split_keyword("foo"), ("foo", ""));
/// assert_eq!(split_keyword("foo bar baz"), ("foo", "bar baz"));
/// ```
````

(The `#` [hides the import](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example) from the rendered snippet).

But that runs into another issue: the `pocket` module is private, so the doctest can't use it. It's called out (though a bit buried) in [this section](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests):

> Note that they will still link against only the public items of your crate; if you need to test private items, you need to write a unit test.

We have a few options to make this work:

1. Add the [`ignore` attribute](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) to the code block.
2. Make this module public in `lib.rs`, with `pub mod pocket;`, and add the `# use` above. It does mean that our Rust component's public API is different than the UniFFIed API—but, given that the Rust component is never consumed directly by other Rust code, I think that's okay.
3. Rewrite this doctest as a unit test. This is what `keyword.rs` and `suggestion.rs` does.

File: components/suggest/src/db.rs [hunk 27]
```diff
@@ -124,16 +125,22 @@ impl<'a> SuggestDao<'a> {
...
-            "SELECT s.id, k.rank, s.title, s.url, s.provider
+        let (keyword_prefix, keyword_suffix) = split_keyword(keyword);
+        Ok(self.conn.query_rows_and_then_cached(
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
              FROM suggestions s
              JOIN keywords k ON k.suggestion_id = s.id
              WHERE k.keyword = :keyword
-             LIMIT 1",
+             UNION ALL
+             SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
```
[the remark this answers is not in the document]
linabutler: Yes, let's get a follow-up on file to let callers specify which providers they want—we can use that to optimize which query we run, and on mobile to restrict suggestions to AMP and Wikipedia.

components/suggest/src/schema.rs
```diff
@@ -96,7 +110,7 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
 
     fn upgrade_from(&self, _db: &Transaction<'_>, version: u32) -> open_database::Result<()> {
         match version {
-            1..=4 => {
+            1..=6 => {
```
linabutler, line 113:
```suggestion
            1..=5 => {
```

Let's make this one version lower than the current. `upgrade_from` shouldn't be called if the database version  is current, but we want that to fail nosily with an error if it does.

linabutler (approved):
Wonderful! r+wc from me, too! :shipit:

components/suggest/src/pocket.rs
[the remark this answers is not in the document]
tiftran: oooh

components/suggest/src/pocket.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+/// Split the keyword by the first whitespace into the prefix and the suffix.
+/// Return an empty string as the suffix if there is no whitespace.
+/// ```
+/// # use suggest::pocket::split_keyword;
+/// assert_eq!(split_keyword("foo"), ("foo", ""));
+/// assert_eq!(split_keyword("foo bar baz"), ("foo", "bar baz"));
+/// ```
```
linabutler, line 53: Oops—I think this last commit removed a bit too much! The `Examples` heading is [rustdoc's convention](https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments) for doctests.

```suggestion
/// Split the keyword by the first whitespace into the prefix and the suffix.
/// Return an empty string as the suffix if there is no whitespace.
///
/// # Examples
///
/// ```
/// # use suggest::pocket::split_keyword;
/// assert_eq!(split_keyword("foo"), ("foo", ""));
/// assert_eq!(split_keyword("foo bar baz"), ("foo", "bar baz"));
/// ```
```

components/suggest/src/schema.rs
```diff
@@ -21,6 +21,15 @@ pub const SQL: &str = "
         PRIMARY KEY (keyword, suggestion_id)
     ) WITHOUT ROWID;
 
+    CREATE TABLE pocket_keywords( 
+        keyword_prefix TEXT NOT NULL, 
+        keyword_suffix TEXT NOT NULL DEFAULT '', 
+        confidence INTEGER NOT NULL,
+        rank INTEGER NOT NULL,
+        suggestion_id INTEGER NOT NULL REFERENCES suggestions(id), 
+        PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id)
```
linabutler, line 30: Oh no, GitHub ate my comment! 😭 Posting it again...

I tested with our live Remote Settings dataset, and it failed to ingest a suggestion that had the same keyword (`"friendship expert"`) in `low_confidence_keywords` and `high_confidence_keywords`:

```
DownloadedPocketSuggestion {
    url: "https://getpocket.com/collections/pocket-best-of-2022-marisa-franco-friendship",
    title: "Best of 2022: Dr. Marisa G Franco on the Importance of Friendship",
    low_confidence_keywords: ["friendship", "friendship expert", "marisa franco", "marisa g franco", "social connections"],
    high_confidence_keywords: ["are friends important", "friendship benefits", "friendship expert", "friendship support", "how important are friends", "importance of friendship", "quality friendships", "the importance of friendship", "the power of friendship", "why you need friends"],
    score: 0.25
}
```

Per [this comment from @0c0w3](https://github.com/mozilla/application-services/pull/5841#issuecomment-1736336089), it sounds like we do want to support these. Is that right, Drew?

If so, let's make `confidence` part of the primary key:

```suggestion
        PRIMARY KEY (keyword_prefix, keyword_suffix, confidence, suggestion_id)
```

Could you also please fix up the trailing whitespace on lines 25, 26, and 29, @tiftran? Thank you! 🙏🏼 (If you haven't already set it, VS Code's `files.trimTrailingWhitespace` setting is really helpful!)

0c0w3: > Per [this comment from @0c0w3](https://github.com/mozilla/application-services/pull/5841#issuecomment-1736336089), it sounds like we do want to support these. Is that right, Drew?

It's not a requirement, and re-reading my previous comment I'm not sure I agree with it on second thought, sorry! Having been involved in the process of picking these keywords, I'm pretty sure this kind of duplication is not intentional. Unless it's trivial to support in the Rust implementation, I think we should probably remove these kinds of duplicates at ingestion time too.

We could check with Nive for each of these individual cases, but right now desktop will show them as high-confidence top pick matches since it always looks up high-confidence keywords first, and I think it would be OK for us to assume that high confidence is intended.

0c0w3: I've uploaded new suggestions to stage and prod with these dupes removed from `lowConfidenceKeywords` (awaiting review) and made https://github.com/mozilla-services/merino-py/pull/434 to fix the uploader.

ncloudioj: Duplicates are removed at ingestion time so the current schema should be fine now. Going to mark this as resolved.

components/suggest/src/store.rs
```diff
@@ -348,31 +323,75 @@ where
...
+            dao.drop_unparsable_record_id(&record_id)?;
+
+            // Advance the last fetch time, so that we can resume
+            // fetching after this record if we're interrupted.
+            dao.put_last_ingest_if_newer(record.last_modified)?;
+
+            Ok(())
+        })?;
+
+        Ok(())
```
ncloudioj, line 393:
```suggestion
        })
```

ncloudioj (approved):

linabutler (dismissed):
Thanks all—they're all ingested correctly from Remote Settings now! 🎉

Please fold in @ncloudioj's last suggestion, squash, and let's get this landed!

linabutler (approved):

PR #5849, opened 2023-09-27
Avoid `brew update-reset` each time we check Swift formatting

### LINKED ISSUE #4025: Work out how to revert #4024
To get around CI failing, https://github.com/Homebrew/discussions/discussions/691#discussioncomment-604291 suggested we execute `brew update-reset`

* which seems to have worked, but shouldn't be necessary for every build. We merged that in the hope it goes away by itself or something, so we should try and work out how to revert it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-2623)
┆Epic: Important backlog

mhammond:
This was a hack for a one-time problem which has almost certainly gone away now. If we did want to do this, it should probably be formalized and not a hack just for swift-lint, so let's just kill it if CI passes.

Fixes #4025

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -212,8 +212,6 @@ jobs:
     executor: macos
     steps:
       - full-checkout
-      # swiftlint moved? See https://github.com/Homebrew/discussions/discussions/691
-      - run: brew update-reset
       - run: brew install swiftlint swiftformat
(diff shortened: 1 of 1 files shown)


jhugman (approved):
LGTM.

PR #5853, opened 2023-10-04
Add a Places API to recursively count the number of descendants in a … [firefox-android: mhammond/count-bookmarks-in-trees] 

### LINKED ISSUE #5789: Add a Places API to recursively count the number of descendants in a folder
[Fenix's `HomeActivity.onPause` fetches the entire bookmarks tree](https://github.com/mozilla-mobile/firefox-android/blob/a415402dcdaf57c6d71889d696e7a9de496f2850/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#L560-L570), but it's only interested in the descendant count.

Fetching and building the tree does a fair bit of work behind the scenes, and creates lots of extra objects on both the Rust and Kotlin sides. I'm a little nervous about Fenix doing that in `onPause`, too; [the Android docs warn](https://developer.android.com/reference/android/app/Activity#onPause()) not to do anything lengthy in that method.

Let's make things a little easier for Fenix, and add an API to count descendants. The new API can use a more efficient SQL query, and pass just a single number over the FFI, instead of fully-formed `BookmarkNode` objects for every bookmark in the library. I think this would especially help users with large bookmark libraries.

Split out from #5773.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3887)

mhammond:
…set of folders

Fixes #5789 - that issue implies the API should just count in a single folder, but the link android-components code ends up wanting to do multiple folders, so having this API take a list of folders makes life easier for them - ie, a-c does something like:

```
bookmarksStorage.countBookmarksInTrees(listOf(BookmarkRoot.Menu.id, BookmarkRoot.Toolbar.id, BookmarkRoot.Unfiled.id)
```

(I'm shaving many many yaks trying to get a-c to build with this, which are unrelated to the patch, so thought I'd get this up while I fight that)

@linabutler, wdyt?

### DIFF
--- a/components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
+++ b/components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
@@ -119,6 +119,19 @@ interface ReadableBookmarksConnection : InterruptibleConnection {
      * has its `interrupt()` method called on another thread.
      */
     fun getRecentBookmarks(limit: Int): List<BookmarkItem>
+
+    /**
+     * Counts the number of bookmark items in the bookmark trees under the specified GUIDs.
+
+     * @param guids The guids of folders to query.
+     * @return Count of all bookmark items (ie, not folders or separators) in all specified folders recursively.
+     * Empty folders, non-existing GUIDs and non-existing items will return zero.
+     * The result is implementation dependant if the trees overlap.
+     *
+     * @throws OperationInterrupted if this database implements [InterruptibleConnection] and
+     * has its `interrupt()` method called on another thread.
+     */
+    fun countBookmarksInTrees(guids: List<Guid>): UInt
 }
 
 /**
--- a/components/places/src/storage/bookmarks.rs
+++ b/components/places/src/storage/bookmarks.rs
@@ -14,7 +14,7 @@ use crate::types::{BookmarkType, SyncStatus};
 use rusqlite::{self, Connection, Row};
 #[cfg(test)]
 use serde_json::{self, json};
-use sql_support::{self, ConnExt};
+use sql_support::{self, repeat_sql_vars, ConnExt};
 use std::cmp::{max, min};
 use sync15::engine::EngineSyncAssociation;
 use sync_guid::Guid as SyncGuid;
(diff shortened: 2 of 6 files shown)


components/places/src/storage/bookmarks.rs
```diff
@@ -786,6 +786,40 @@ pub fn bookmarks_get_url_for_keyword(db: &PlacesDb, keyword: &str) -> Result<Opt
...
+    if item_guids.is_empty() {
+        return Ok(0);
+    }
+    let vars = repeat_vars(item_guids.len());
+    let sql = format!(
+        r#"
+        WITH RECURSIVE bookmark_tree(id, parent, type)
+        AS (
+            SELECT id, parent, type FROM moz_bookmarks
+            WHERE parent IN (SELECT id from moz_bookmarks WHERE guid IN ({vars}))
```
linabutler, line 811: Could we use `sql_support::repeat_sql_vars` here? I thought for sure we had a helper like `repeat_vars` already... 🤔

components/places/src/storage/bookmarks.rs
```diff
@@ -786,6 +786,40 @@ pub fn bookmarks_get_url_for_keyword(db: &PlacesDb, keyword: &str) -> Result<Opt
...
+            UNION ALL
+            SELECT bm.id, bm.parent, bm.type FROM moz_bookmarks bm, bookmark_tree bt WHERE bm.parent = bt.id
+        )
+    SELECT COUNT(*) from bookmark_tree WHERE type = {};
+    "#,
+        BookmarkType::Bookmark as u8
+    );
+    let params = rusqlite::params_from_iter(item_guids);
+    Ok(db.query_row_and_then(&sql, params, |row| row.get(0))?)
```
linabutler, line 820: If the types work out, could we use our `db.query_one` extension from `sql_support` here?

linabutler (approved):
This is exactly what I had in mind, thanks @mhammond!

components/places/android/src/main/java/mozilla/appservices/places/Bookmarks.kt
```diff
@@ -119,6 +119,19 @@ interface ReadableBookmarksConnection : InterruptibleConnection {
...
      */
     fun getRecentBookmarks(limit: Int): List<BookmarkItem>
+
+    /**
+     * Counts the number of bookmark items in the bookmark trees under the specified GUIDs.
+
+     * @param guids The guids of folders to query.
+     * @return Count of all bookmark items (ie, not folders or separators) in all specified folders recursively.
+     * Empty folders, non-existing GUIDs and non-existing items will return zero.
+     * The result is implementation dependant if the trees overlap.
```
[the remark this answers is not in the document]
mhammond: Yeah - my first version took just one arg, but then looking at the code in a-c it was clear that it would always need to sum the values from Menu, Toolbar and Unfiled, then use a separate count for Mobile - so I figured I'd optimize for the actual use-case. FWIW, https://github.com/mhammond/firefox-android/tree/count-bookmarks-in-trees is that branch.

components/places/src/storage/bookmarks.rs
mhammond: doh - yes, thanks!

components/places/src/storage/bookmarks.rs
mhammond: It's unfortunate that `db.query_one` doesn't take params :( try_query_one does, so there's odd asymmetry there - I moved to try_query_one anyway even though I don't think we could ever return 0 rows.

PR #5867, opened 2023-10-13
Fixes DISCO-2637: Replace `SuggestionQuery` options with `providers`. [firefox-android: linabutler/DISCO-2637] [firefox-ios: linabutler/DISCO-2637]

linabutler:
This commit removes the `include_sponsored` and `include_non_sponsored` options, and replaces them with a list of `providers`. This lets consumers request specific suggestion types: AMP and Wikipedia only on mobile; and AMP, Wikipedia, and active experiments on Desktop.

This is a backward-incompatible API change.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -14,16 +14,16 @@ use rusqlite::{
 };
 use sql_support::{open_database::open_database_with_flags, ConnExt};
 
-use crate::pocket::{split_keyword, KeywordConfidence};
 use crate::rs::{DownloadedAmoSuggestion, DownloadedPocketSuggestion};
 use crate::{
     keyword::full_keyword,
+    pocket::{split_keyword, KeywordConfidence},
     provider::SuggestionProvider,
     rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId},
     schema::{SuggestConnectionInitializer, VERSION},
     store::{UnparsableRecord, UnparsableRecords},
     suggestion::{cook_raw_suggestion_url, Suggestion},
-    Result,
+    Result, SuggestionQuery,
 };
 
 /// The metadata key whose value is the timestamp of the last record ingested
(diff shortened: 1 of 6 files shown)


components/suggest/src/db.rs
```diff
@@ -271,23 +294,24 @@ impl<'a> SuggestDao<'a> {
...
+                         url
+                     )
+                     VALUES(
+                         :record_id,
+                         {},
+                         :title,
+                         :url
+                     )
+                     RETURNING id",
+                    SuggestionProvider::Amo as u8
```
linabutler, line 311: The `SuggestionProvider` variants are all integers, so, as with `providers_to_sql_list` above, we can safely interpolate them directly into the query string, without worrying about SQL injection. This is a trick we use in Places, too! ✨

components/suggest/src/db.rs
```diff
@@ -123,24 +123,45 @@ impl<'a> SuggestDao<'a> {
...
-            named_params! {
-                ":keyword": keyword,
-                ":keyword_prefix": keyword_prefix,
-            },
-            |row| -> Result<Option<Suggestion>> {
+    /// Fetches suggestions that match the given query from the database.
+    pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
+
+        let (mut statement, params) = if query.providers.contains(&SuggestionProvider::Pocket) {
```
linabutler, line 130: Pocket keywords use a separate query, so, if we aren't matching Pocket suggestions, we can skip querying `pocket_keywords` entirely.

That also means we end up with a dynamically-sized list of `params`—two for Pocket; one for non-Pocket—so we have to use a `Vec<(&str, &dyn ToSql)>` to hold them.

components/suggest/src/db.rs
```diff
@@ -258,7 +279,9 @@ impl<'a> SuggestDao<'a> {
                     }
                 }
             }
-        )?.into_iter().flatten().collect())
+        )?.flat_map(Result::transpose).collect::<Result<_>>()?;
```
linabutler, line 282: This transformation filters out the `None`s from our `query_and_then(...)` iterator, then accumulates the remaining suggestions into a vector. It leans on a few iterator helpers to do this:

* `transpose` turns a `Result<Option<Suggestion>>` into an `Option<Result<Suggestion>>`.
* An `Option` is an iterator—it yields zero elements if it's `None`, and one if it's `Some`—so we can use it with `flat_map` to unwrap optionals. This turns our iterator of `Option<Result<Suggestion>>` into an iterator of `Result<Suggestion>`.
* `collect::<Result<_>>` accumulates the iterator of `Result<Suggestion>` into a `Result<Vec<Suggestion>>`.

ncloudioj (approved):
LGTM. Thanks!

I wonder if we could use a "cheaper" construct for `SuggestionQuery.providers`. Creating a new vector (despite the providers mostly remaining the same) for each query seems a bit wasteful to me. Would a bitflag fit the bill? I am thinking of this:

```rs
use bitflags::bitflags;

bitflags! {
    pub struct Providers: u32 {
        const Amp = 1;
        const Wikipedia = 1 << 1;
        const Pocket = 1 << 2;
        const Amo = 1 << 3;
    }
}

pub struct SuggestionQuery {
    pub keyword: String,
    pub providers: Providers, 
}

// For mobile
let query = SuggestionQuery {
    keyword: "foo".to_own(),
    providers = Providers::Amp | Providers::Wikipedia,
};

if query.providers.contains(Providers::Pocket) {
} else ...
```

WDYT?

linabutler:
Thanks for the review!

I thought bitflags—or even a set—would be perfect here, too! Unfortunately, they're tricky to do with UniFFI. Neither UDL nor UniFFI's serialization format supports custom values for fieldless enums, so I don't think it would be possible today for Swift, Kotlin, or JS to pass a bitmask, and have Rust see the same value.

We could use a [custom type](https://mozilla.github.io/uniffi-rs/udl/custom_types.html) to work around that, but then we couldn't use the `enum` syntax in UDL, because we can't declare a custom value for each variant.

ncloudioj:
Bummer. Yep, the FFI limitations! If we use an opaque `u32`, then we'd sacrifice the type safety  🤷‍♂️. let's leave it as is then.

linabutler:
PRs resolving the breaking changes for our consumers:

* Android: https://github.com/mozilla-mobile/firefox-android/pull/4073
* iOS: https://github.com/mozilla-mobile/firefox-ios/pull/16818

The [process](https://mozilla.github.io/application-services/book/howtos/breaking-changes.html) for landing breaking changes like this is roughly:

* Land the Application Services PR first.
* Wait for the automated a-s version bumps in Firefox for Android and iOS to fail. (I'll drop links to those failing PRs here when that happens, so that folks can follow the process along).
* Fold in the Android and iOS PRs that fix the breakages into the version bump PRs, push, and land that as a squashed commit. This keeps the version bump, and the fix for the breakages, in a single commit.

PR #5870, opened 2023-10-13
Add an option for getting number of suggestions

mtbottle:
The current search an suggest query doesn't enforce a limit for how many suggestions to return. This allows us to define a limit at the client level. If we do not put a limit, then the default is to return all of the suggestions.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -126,6 +126,7 @@ impl<'a> SuggestDao<'a> {
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
         let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
+        let suggestions_limit = query.limit.unwrap_or(-1);
 
         let (mut statement, params) = if query.providers.contains(&SuggestionProvider::Pocket) {
             (self.conn.prepare_cached(
@@ -139,12 +140,15 @@ impl<'a> SuggestDao<'a> {
                      SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
                      FROM suggestions s
                      JOIN pocket_keywords k ON k.suggestion_id = s.id
-                     WHERE k.keyword_prefix = :keyword_prefix",
+                     WHERE k.keyword_prefix = :keyword_prefix
+                     ORDER BY s.provider
+                     LIMIT :suggestions_limit",
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
                 (":keyword", &query.keyword as &dyn ToSql),
                 (":keyword_prefix", &keyword_prefix as &dyn ToSql),
+                (":suggestions_limit", &suggestions_limit as &dyn ToSql),
             ])
         } else {
             (self.conn.prepare_cached(
@@ -153,11 +157,14 @@ impl<'a> SuggestDao<'a> {
                      FROM suggestions s
                      JOIN keywords k ON k.suggestion_id = s.id
                      WHERE s.provider IN ({}) AND
-                           k.keyword = :keyword",
+                           k.keyword = :keyword
+                     ORDER BY s.provider
+                     LIMIT :suggestions_limit",
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
                 (":keyword", &query.keyword as &dyn ToSql),
+                (":suggestions_limit", &suggestions_limit as &dyn ToSql),
             ])
         };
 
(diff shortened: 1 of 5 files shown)


linabutler:
I'm so sorry I bitrotted you 😬 But the shape of this looks great; thanks for taking this bug!

As far as interpolating the limit into the SQL query, I think we could take advantage of two things to make it a little bit easier:

1. `limit` is a `u64`, so we can include it in the `format!(...)` string directly, instead of using a binding parameter, without worrying about SQL injection. `providers_to_sql_list` does the same trick.
2. A [negative LIMIT expression](https://www.sqlite.org/lang_select.html#limitoffset) is the same as "no limit"—so we could always specify a `LIMIT` clause in the query, and default to `-1` if it's not specified.

components/suggest/src/db.rs
```diff
@@ -139,12 +140,15 @@ impl<'a> SuggestDao<'a> {
                      SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
                      FROM suggestions s
                      JOIN pocket_keywords k ON k.suggestion_id = s.id
-                     WHERE k.keyword_prefix = :keyword_prefix",
+                     WHERE k.keyword_prefix = :keyword_prefix
+                     ORDER BY s.provider
+                     LIMIT :suggestions_limit",
```
mtbottle, line 145: @linabutler You mentioned that I don't need to use a binding parameter for this, but I feel like it's a bit more readable than passing into `providers_to_sql_list`. Is there a reason to not using binding params?

components/suggest/src/db.rs
```diff
@@ -126,6 +126,7 @@ impl<'a> SuggestDao<'a> {
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
         let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
+        let suggestions_limit = if let Some(l) = query.limit { l } else { -1 };
```
linabutler, line 129:
```suggestion
        let suggestions_limit = query.limit.unwrap_or(-1);
```

`Option` has [a handful of convenience methods](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or) for unwrapping values; we can use one of those here! ✨

components/suggest/src/db.rs
```diff
@@ -139,12 +140,14 @@ impl<'a> SuggestDao<'a> {
                      SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
                      FROM suggestions s
                      JOIN pocket_keywords k ON k.suggestion_id = s.id
-                     WHERE k.keyword_prefix = :keyword_prefix",
+                     WHERE k.keyword_prefix = :keyword_prefix
+                     LIMIT :suggestions_limit",
                     providers_to_sql_list(&query.providers),
```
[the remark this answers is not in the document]
linabutler: One other note for our future selves: unless it can use an indexed column, SQLite will build an in-memory B-Tree for sorting with `ORDER BY`. The cost can add up with a lot of rows, or with large rows, but I think it's totally okay in our case here: it makes results and tests deterministic, it sets the stage for ranking suggestions later, and, if we have more than a handful of suggestions to show, the awesomebar will run out of space long before sorting the suggestions becomes a bottleneck.

components/suggest/src/store.rs
```diff
@@ -1583,11 +1634,118 @@ mod tests {
...
+                            title: "Multimatching",
+                            url: "https://getpocket.com/collections/multimatch",
+                            score: 0.25,
+                            is_top_pick: true,
+                        },
+                    ]
+                "#]],
+            ),
+            (
+                "multimatch; all providers, limit 2",
```
linabutler, line 1694: I love these tests! 🏆 Thank you!

components/suggest/src/db.rs
linabutler: Oops, I left a review comment earlier, and didn't see your reply; the binding param here is great! I think it could've made more sense to interpolate the limit directly into the string if we were conditionally adding the `ORDER BY` and `LIMIT` clauses together (so the string would be `"ORDER BY s.provider LIMIT {}"` if we had a limit, and `""` if we didn't), but we aren't doing that—or any fancy chunking—so there's no reason to prefer it over the param.

linabutler (dismissed):
Oops, I meant to approve—this looks great; please land away with the `unwrap_or` change!

components/suggest/src/db.rs
mtbottle: This is so much nicer! I'll add this change, thanks! :)

linabutler (approved):

PR #5884, opened 2023-10-24
[DISCO-2649] Suggest: Amo keyword logic to match prefix and suffix

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -247,7 +247,7 @@ impl<'a> SuggestDao<'a> {
                                 ":suggestion_id": suggestion_id
                             },
                             |row| {
-                                if full_suffix.contains(keyword_suffix) {
+                                if full_suffix.starts_with(keyword_suffix) {
                                     Ok(Some(Suggestion::Amo{
                                         title,
                                         url: raw_url,
@@ -265,9 +265,10 @@ impl<'a> SuggestDao<'a> {
                     },
                     SuggestionProvider::Pocket => {
                         let confidence = row.get("confidence")?;
+                        let full_suffix = row.get::<_, String>("keyword_suffix")?;
                         let suffixes_match = match confidence {
-                            KeywordConfidence::Low => row.get::<_, String>("keyword_suffix")?.starts_with(keyword_suffix),
-                            KeywordConfidence::High => row.get::<_, String>("keyword_suffix")? == keyword_suffix,
+                            KeywordConfidence::Low => full_suffix.starts_with(keyword_suffix),
+                            KeywordConfidence::High => full_suffix == keyword_suffix,
                         };
                         if suffixes_match {
                             self.conn.query_row_and_then(
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -128,7 +128,11 @@ impl<'a> SuggestDao<'a> {
         let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
         let suggestions_limit = query.limit.unwrap_or(-1);
 
-        let (mut statement, params) = if query.providers.contains(&SuggestionProvider::Pocket) {
+        let (mut statement, params) = if query
+            .providers
+            .iter()
+            .any(|p| matches!(p, SuggestionProvider::Pocket | SuggestionProvider::Amo))
+        {
             (self.conn.prepare_cached(
                 &format!(
                     "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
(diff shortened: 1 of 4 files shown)


components/suggest/src/db.rs
```diff
@@ -128,7 +128,9 @@ impl<'a> SuggestDao<'a> {
         let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
         let suggestions_limit = query.limit.unwrap_or(-1);
 
-        let (mut statement, params) = if query.providers.contains(&SuggestionProvider::Pocket) {
+        let (mut statement, params) = if query.providers.contains(&SuggestionProvider::Pocket)
+            || query.providers.contains(&SuggestionProvider::Amo)
```
ncloudioj, line 132: Nit: you can also do this with the benifit that it will scan the provider array once.
```suggestion
        let (mut statement, params) = if query
            .providers
            .iter()
            .any(|p| matches!(p, SuggestionProvider::Pocket | SuggestionProvider::Amo))

```

components/suggest/src/db.rs
```diff
@@ -141,6 +143,11 @@ impl<'a> SuggestDao<'a> {
                      FROM suggestions s
                      JOIN pocket_keywords k ON k.suggestion_id = s.id
                      WHERE k.keyword_prefix = :keyword_prefix
+                     UNION ALL
```
ncloudioj, line 146: This will query both amo and pocket keyword tables despite that the consumer might need one of them. I think we have two options here:
- Merge `amo_keywords` and `pocket_keywords` as they're almost identical with the same matching logic. We can split them up if they diverge in the future
- Split this SQL into multiple so that it will query suggestions per provider, which means it'd issue multiple calls against the database. The upside is that we don't need to maintain this SQL for different keyword tables

components/suggest/src/db.rs
tiftran: oooh fancy

components/suggest/src/db.rs
tiftran: makes sense, let me try to fiddle with both options

tiftran: okie doke, so i think i was able to implement the first option, currently trying the second

components/suggest/src/db.rs
```diff
@@ -242,6 +247,7 @@ impl<'a> SuggestDao<'a> {
                                 ":suggestion_id": suggestion_id
                             },
                             |row| {
+                                if full_suffix.contains(keyword_suffix){
```
linabutler, line 250:
```suggestion
                                if full_suffix.contains(keyword_suffix) {
```

components/suggest/src/schema.rs
```diff
@@ -21,10 +21,10 @@ pub const SQL: &str = "
         PRIMARY KEY (keyword, suggestion_id)
     ) WITHOUT ROWID;
 
-    CREATE TABLE pocket_keywords( 
+    CREATE TABLE prefix_keywords(
         keyword_prefix TEXT NOT NULL,
         keyword_suffix TEXT NOT NULL DEFAULT '',
-        confidence INTEGER NOT NULL,
+        confidence INTEGER NOT NULL DEFAULT '',
```
linabutler, line 27:
```suggestion
        confidence INTEGER NOT NULL DEFAULT 0,
```

`''` is technically legal—SQLite's type system is [relaxed by default](https://www.sqlite.org/datatype3.html)—but let's use a number if the column is declared as a number.

components/suggest/src/store.rs
```diff
@@ -1945,14 +1945,34 @@ mod tests {
                 "#]],
             ),
             (
-                "keyword = `soft`; AMP, Wikipedia, and AMO",
+                "keyword = `masking e`; AMO only",
```
linabutler, line 1948: Just for completeness, could you please add another test for one of the single-word keywords, like `relay` or `spam`?

linabutler (approved):
r+wc, thanks! This looks like it'll be a very easy uplift 😊

components/suggest/src/db.rs
```diff
@@ -242,6 +247,7 @@ impl<'a> SuggestDao<'a> {
                                 ":suggestion_id": suggestion_id
                             },
                             |row| {
+                                if full_suffix.contains(keyword_suffix) {
```
linabutler, line 250: Sorry, I just noticed this: `.contains()` will match anywhere in the string, so `masking mail`, `masking i`, `masking il`, `masking l`, etc. will all match `masking email`.

Is that intentional? Or do we want this to be a prefix match, so only `masking e`, `masking em`, `masking ema`, etc. should match?

(Either way, let's add another test for that case, too, please—either a positive one if it's intentional, or a negative one that verifies we don't return anything. Thanks!)

linabutler: Side note: I also really like how you factored out `full_suffix` into a variable, and wondering if we can give the logic in the Pocket match arm the same treatment, while you're here?

```rs
let confidence = row.get("confidence")?;
let full_suffix = row.get::<_, String>("keyword_suffix")?;
let suffixes_match = match confidence {
    KeywordConfidence::Low => full_suffix.starts_with(keyword_suffix),
    KeywordConfidence::High => full_suffix == keyword_suffix,
};
```

linabutler (approved):
Thank you! 🙌🏼

PR #5887, opened 2023-10-25
Fxa internal changes [firefox-android: bendk/resolve-rust-fxa-client-changes]

bendk:
This PR contains internal app-services changes to prepare for android-components changes for the fxa-client rework.

The [RFC](https://docs.google.com/document/d/16UUdfx1yvK4W-jBvczL5tPUaJ-R-TFIRiLckAS7BJmM/edit) proposes this as the first step in the plan.  After this, we can start making changes in the `firefox-android` repo.

I don't believe any of this should be breaking for Swift, but we should double check that.

  - Note:

### DIFF
--- a/components/fxa-client/src/auth.rs
+++ b/components/fxa-client/src/auth.rs
@@ -28,6 +28,11 @@ use error_support::handle_error;
 use std::collections::HashMap;
 
 impl FirefoxAccount {
+    /// Get the high-level authentication state of the client
+    pub fn get_auth_state(&self) -> FxaRustAuthState {
+        self.internal.lock().get_auth_state()
+    }
+
     /// Initiate a web-based OAuth sign-in flow.
     ///
     /// This method initializes some internal state and then returns a URL at which the
@@ -158,6 +163,26 @@ impl FirefoxAccount {
     pub fn disconnect(&self) {
         self.internal.lock().disconnect()
     }
+
+    /// Update the state based on authentication issues.
+    ///
+    /// **💾 This method alters the persisted account state.**
+    ///
+    /// Call this if you know there's an authentication / authorization issue that requires the
+    /// user to re-authenticated.  It transitions the user to the [FxaRustAuthState.AuthIssues] state.
+    pub fn on_auth_issues(&self) {
+        self.internal.lock().on_auth_issues()
+    }
+
+    /// Used by the application to test auth token issues
+    pub fn simulate_temporary_auth_token_issue(&self) {
+        self.internal.lock().simulate_temporary_auth_token_issue()
+    }
+
+    /// Used by the application to test auth token issues
+    pub fn simulate_permanent_auth_token_issue(&self) {
+        self.internal.lock().simulate_permanent_auth_token_issue()
+    }
 }
 
 /// Information about the authorization state of the application.
@@ -171,4 +196,19 @@ pub struct AuthorizationInfo {
 /// Additional metrics tracking parameters to include in an OAuth request.
 pub struct MetricsParams {
     pub parameters: HashMap<String, String>,
+}
+
+/// High-level view of the authorization state
+///
+/// This is named `FxaRustAuthState` because it doesn't track all the states we want yet and needs
+/// help from the wrapper code.  The wrapper code defines the actual `FxaAuthState` type based on
+/// this, adding the extra data.
+///
+/// In the long-term, we should track that data in Rust, remove the wrapper, and rename this to
+/// `FxaAuthState`.
+#[derive(Debug, PartialEq, Eq)]
+pub enum FxaRustAuthState {
+    Disconnected,
+    Connected,
+    AuthIssues,
 }
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -75,6 +75,9 @@ enum FxaError {
   // **Note:** This error is currently only thrown in the Swift language bindings.
   "WrongAuthFlow",
 
+  // The sync scoped key was missing in the server response
+  "SyncScopedKeyMissingInServerResponse",
+
   // Thrown if there is a panic in the underlying Rust code.
   //
   // **Note:** This error is currently only thrown in the Kotlin language bindings.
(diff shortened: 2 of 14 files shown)


components/fxa-client/src/internal/state_manager.rs
```diff
@@ -157,23 +169,64 @@ impl StateManager {
...
+        self.persisted_state.access_token_cache = HashMap::new();
+        self.persisted_state.device_capabilities = HashSet::new();
+        self.persisted_state.server_local_device_info = None;
+        self.persisted_state.session_token = None;
+        self.persisted_state.logged_out_from_auth_issues = false;
+        self.flow_store.clear();
+    }
+
+    /// Called when the account is logged out because of auth issues.  This clears some of the auth
+    /// state, but less than disconnect().
```
tarikeshaq, line 206: what do you think about being explicit what data is not deleted (device id, right?)

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -157,23 +169,64 @@ impl StateManager {
...
         self.flow_store.clear();
     }
 
     /// Called when the account is disconnected.  This clears most of the auth state, but keeps
     /// some information in order to eventually reconnect to the same user account later.
     pub fn disconnect(&mut self) {
-        self.persisted_state = self.persisted_state.start_over();
+        // Clear (almost all of) the persisted state of the account.
+        //
+        // This method keep just enough information to be able to eventually reconnect
```
tarikeshaq, line 188:
```suggestion
        // This method keeps just enough information to be able to eventually reconnect
```

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -157,23 +169,64 @@ impl StateManager {
...
+        // to the same user account later. To completely forget the previously-signed-in
+        // user, simply discard the persisted data.
```
tarikeshaq, line 190: this comment is a bit redundant, the doc comment on top of the function says the same thing, what do you think of deleting this one?

components/fxa-client/src/fxa_client.udl
```diff
@@ -243,7 +246,18 @@ interface FirefoxAccount {
...
-  
+
+  // Log out because of authorization / authentication issues
+  //
+  //
+  // **💾 This method alters the persisted account state.**
+  //
+  // Call this if you know there's an authentication / authorization issue that requires the
+  // user to re-authenticated.  It transitions the user to the [FxaRustAuthState.AuthIssues] state.
+  void logout_from_auth_issues();
```
tarikeshaq, line 257: IIUC, we want this to preserve the device id in the case of an authentication issue, so that we don't issue the user a new device id, correct?

What do you think of adding tests for this in the Rust layer to make sure that an authentication after a auth issue correctly uses the old device id?

components/fxa-client/src/auth.rs
```diff
@@ -172,3 +198,18 @@ pub struct AuthorizationInfo {
 pub struct MetricsParams {
     pub parameters: HashMap<String, String>,
 }
+
+/// High-level view of the authorization state
+///
+/// This is named `FxaRustAuthState` because it doesn't track all the states we want yet and needs
```
tarikeshaq, line 204: Hmmm maybe `InternalAuthState` or something of the sort? I'm also happy to keep it as-is since it seems its temporary

tarikeshaq (commented):
Thanks a ton @bendk! Overall this looks okay to me, I had a few questions but mostly I wonder how we could add some tests (especially for the new API changes, like the auth issue trigger)

I'm going through the document you wrote as well and will jot down some thoughts, I'm curious if we can handle the auth issues internally in a way that the consumer doesn't have to call an explicit function, but that might be a larger undertaking

components/fxa-client/src/auth.rs
bendk: Yeah, I think that name would make more sense if it was permanent.  I almost want a name that feels a bit off, to match the temporary situation we're going to be in for the next few months.

bendk:
Thanks for the review!  I think I got all the comments, let me know if I missed any.

tarikeshaq (approved):
LGTM! looks like the branch build didn't run (I imagine a missing `:` in the PR title?) so we should quickly give that a go, but otherwise looks great!

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -157,23 +169,59 @@ impl StateManager {
...
+        self.persisted_state.server_local_device_info = None;
+        self.persisted_state.session_token = None;
+        self.persisted_state.logged_out_from_auth_issues = false;
+        self.flow_store.clear();
+    }
+
+    /// Called when the account is logged out because of auth issues.  This clears the auth state,
+    /// but leaves `current_device_id` untouched.  That way, if the user re-authenticates they can
+    /// continue to use their existing device entry.
+    pub fn logout_from_auth_issues(&mut self) {
```
mhammond, line 202: I don't like this name tbh - I don't see it as the user being logged out, but instead being in a bad state. For example, do we really want to reset `last_handled_command` here? If a user on desktop changes their password then later sends a tab to this device (forgetting they are yet to re-auth it), I think we'd want this device to see the tab immediately after reauthenticating.  Something like "update_for_auth_issues()" or similar seems better - but I don't want to block landing on this if it ends up being a PITA.

(I guess you could argue that this is the only method called "log out" and can't be confused with "disconnect()", which I also can't argue with :)

Also: it's not completely clear how the app responds once the user reauths. Can we be *sure* it will re-call `set_device_capabilities()` - given that is about the *device* state and independent of the *user* state, I can see that it might not, so maybe this should not be cleared too? Ditto the device name - Tarik mentions the device ID should remain unchanged, can we be sure the app will do whatever it does to ensure the device name also doesn't change?

mhammond (approved):
Take my comments for what they cost you, feel free to open issues for anything you think is worthwhile for later.

components/fxa-client/src/internal/state_manager.rs
bendk: I don't really like the name either, it's hard to pick one.  How about `transition_to_auth_issues()`?  I don't love that one either, but it makes it clear that it's a state-transition at least.

Good point on `last_handled_command`.  I'll make it so it doesn't reset that.  I'm also thinking `device_capabilities` shouldn't be reset, especially since this PR changes the semantics on that a bit.  It's now only used to know store what the consumer app last sent to `ensure_capabilities` or `initialize_device`.  Seems like that should persist through auth issues.

mhammond: `note_auth_issue()` maybe?

mhammond: or "observe", or "on"?

bendk: Thanks for the naming help.  I went with `on_auth_issues`.

The method currently resets the same fields as `disconnect`, except `current_device_id`, `device_capabilities`, and `last_handled_command`.  Does that seem right?

bendk:
I also squashed this all together into one commit.  I think this is almost ready to merge.

components/fxa-client/src/internal/state_manager.rs
mhammond: That seemed right to me when I was looking over them.

PR #5892, opened 2023-10-27
[DISCO-2661] Suggest: fix: use lowercase keyword to find suggestion

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -125,7 +125,8 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
-        let (keyword_prefix, keyword_suffix) = split_keyword(&query.keyword);
+        let keyword_lowercased = &query.keyword.to_lowercase();
+        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
         let suggestions_limit = query.limit.unwrap_or(-1);
 
         let (mut statement, params) = if query
@@ -150,7 +151,7 @@ impl<'a> SuggestDao<'a> {
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
-                (":keyword", &query.keyword as &dyn ToSql),
+                (":keyword", keyword_lowercased as &dyn ToSql),
                 (":keyword_prefix", &keyword_prefix as &dyn ToSql),
                 (":suggestions_limit", &suggestions_limit as &dyn ToSql),
             ])
@@ -167,7 +168,7 @@ impl<'a> SuggestDao<'a> {
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
-                (":keyword", &query.keyword as &dyn ToSql),
+                (":keyword", keyword_lowercased as &dyn ToSql),
                 (":suggestions_limit", &suggestions_limit as &dyn ToSql),
             ])
         };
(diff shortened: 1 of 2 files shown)


linabutler:
This looks great, thanks!

I have a suggestion for another way to do this, that might be interesting to think about, and if you're curious 😊 Columns in SQLite can have ["collation sequences"](https://www.sqlite.org/datatype3.html#collation) that define how to compare strings in that column.

As an alternative to lowercasing the keywords at query time, we can declare the `NOCASE` collation on the columns we want, like this:

```sql
CREATE TABLE keywords(
        keyword TEXT NOT NULL COLLATE NOCASE,
        ...
);

CREATE TABLE prefix_keywords(
        keyword_prefix TEXT NOT NULL COLLATE NOCASE,
        keyword_suffix TEXT NOT NULL DEFAULT '' COLLATE NOCASE,
        ...
);
```

And then SQLite will store and query those keywords in a case-insensitive way. I like it for two reasons: it's not possible to accidentally forget to normalize the string at query or storage time, and it lets you define language-specific collations.

`COLLATE NOCASE`'s biggest drawback is that it doesn't do case-folding for non-ASCII characters—which isn't a problem for Suggest now, but good to keep in mind for when we expand. Rust's `to_lowercase()` is more Unicode-aware than `COLLATE NOCASE`, but still not suitable for general case-insensitive comparisons, because those rules can be language-specific. In German, `SS` and `ß` are equivalent ignoring case; meanwhile, in Turkish, `I` and `i` are two different letters that are not equivalent. It turns out [there are a lot of rules that go into a case-insensitive string comparison](https://unicode-org.github.io/icu/userguide/collation/)!

Luckily, though, we don't have to worry about any of those for a while!

linabutler (approved):

tiftran:
> This looks great, thanks!
> 
> I have a suggestion for another way to do this, that might be interesting to think about, and if you're curious 😊 Columns in SQLite can have ["collation sequences"](https://www.sqlite.org/datatype3.html#collation) that define how to compare strings in that column.
> 
> As an alternative to lowercasing the keywords at query time, we can declare the `NOCASE` collation on the columns we want, like this:
> 
> ```sql
> CREATE TABLE keywords(
>         keyword TEXT NOT NULL COLLATE NOCASE,
>         ...
> );
> 
> CREATE TABLE prefix_keywords(
>         keyword_prefix TEXT NOT NULL COLLATE NOCASE,
>         keyword_suffix TEXT NOT NULL DEFAULT '' COLLATE NOCASE,
>         ...
> );
> ```
> 
> And then SQLite will store and query those keywords in a case-insensitive way. I like it for two reasons: it's not possible to accidentally forget to normalize the string at query or storage time, and it lets you define language-specific collations.
> 
> `COLLATE NOCASE`'s biggest drawback is that it doesn't do case-folding for non-ASCII characters—which isn't a problem for Suggest now, but good to keep in mind for when we expand. Rust's `to_lowercase()` is more Unicode-aware than `COLLATE NOCASE`, but still not suitable for general case-insensitive comparisons, because those rules can be language-specific. In German, `SS` and `ß` are equivalent ignoring case; meanwhile, in Turkish, `I` and `i` are two different letters that are not equivalent. It turns out [there are a lot of rules that go into a case-insensitive string comparison](https://unicode-org.github.io/icu/userguide/collation/)!
> 
> Luckily, though, we don't have to worry about any of those for a while!

Would it be overkill to have both?

linabutler:
> Would it be overkill to have both?

Yeah, I feel like it would be overkill to have both—one or the other would be good 😅 The way you have it now is great to land!

tiftran:
@mergifyio backport release-v120

Issue #5902, opened 2023-11-02
Intermittent ios-test failures


skhamis:
There's been an uptick in ios-test intermittent -- Example: failures: https://app.circleci.com/pipelines/github/mozilla/application-services/35740/workflows/41ab9956-c4bc-4a02-848e-05e5e367c14c/jobs/177795

Excerpt from link above:
```
MozillaTestServicesTests.PlacesTests
  testHistoryMetadataBasics, XCTAssertEqual failed: ("1") is not equal to ("2")
  /Users/distiller/project/megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/PlacesTests.swift:312

        XCTAssertEqual(1, try! db.getHistoryMetadataSince(since: afterLastMeta1Update).count)
        XCTAssertEqual(2, try! db.getHistoryMetadataSince(since: beginning).count)

  testHistoryMetadataBasics, XCTAssertEqual failed: ("1") is not equal to ("2")
  /Users/distiller/project/megazords/ios-rust/MozillaTestServices/MozillaTestServicesTests/PlacesTests.swift:339

        XCTAssertEqual(2, try! db.getHistoryMetadataBetween(start: beginning, end: afterLastMeta2Update).count)
        XCTAssertEqual(1, try! db.getHistoryMetadataBetween(start: afterLastMeta1Update, end: afterLastMeta2Update).count)
```

It also seems slightly tangential to https://github.com/mozilla-mobile/firefox-android/pull/3703 which also disabled tests due to intermittent issues and is also using `getHistoryMetadataBetween`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3991)

mhammond:
I see 2 main possibilities here:
* it's using timestamps, so it's possible not enough care is being taken with them (eg, using `now()` immediately after writing history might *usually* work, but sometimes the value returned by `now()` will be different than what `now()` was at write time, etc.
* is it possible these tests are running in parallel?

mhammond:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1866537

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #5903, opened 2023-11-05
Switch fxa_client from using mockiato to using mockall 

### LINKED ISSUE #5368: Switch fxa_client from using mockiato to using mockall
mockiato is a trait mocking library for unit tests used throughout fxa_client. Mockiato hasn't been updated in 4 or so years, and [mockall](https://crates.io/crates/mockall) seems like what the Rust community is adopting for trait mocking.

This ticket will be a great start for someone new to fxa_client to get around the component

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3599)

nox1134:
Fixes #5368 
Please refer to the previous PR (https://github.com/mozilla/application-services/pull/5890). 
This new PR adds support for mockall for fxa_client.

  - Note:

### DIFF
--- a/components/fxa-client/Cargo.toml
+++ b/components/fxa-client/Cargo.toml
@@ -32,7 +32,7 @@ uniffi = { version = "0.24.1", features = ["build"] }
 
 [dev-dependencies]
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
-mockiato = "0.9"
+mockall = "0.11"
 mockito = "0.31"
 
 [features]
--- a/components/fxa-client/src/internal/http_client.rs
+++ b/components/fxa-client/src/internal/http_client.rs
@@ -53,7 +53,9 @@ const DEVICES_FILTER_DAYS: u64 = 21;
 /// it return something called a "refresh token"? Using unambiguous
 /// verbs to start each method helps avoid confusion here.
 ///
-#[cfg_attr(test, mockiato::mockable)]
+// Due to limitations in mockall, we have to add explicit lifetimes that the Rust compiler would have been happy to infer.
+#[allow(clippy::needless_lifetimes)]
+#[cfg_attr(test, mockall::automock)]
 pub(crate) trait FxAClient {
     fn create_refresh_token_using_authorization_code(
         &self,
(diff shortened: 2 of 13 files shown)


nox1134:
I did changes in this path only `components/fxa-client/` and all the test cases for this `fxa-client` also passed as well still while running tests for the whole repo, i am getting those errors.
`error: expect test failed
   --> components/remote_settings/src/client.rs:852:9` 
   You can check the CI to know about these errors. How do i fix these? and how are these related to changes that i did

tarikeshaq (changes requested):
Thanks a ton @nox1134 🎉 🎉 

For the failing tests, I see there are quite a few changes to the `Cargo.lock` that would have caused it, you might have run `cargo update` at some point?

For now, if you revert the changes to `Cargo.lock` and run any cargo command, it'll be updated only to include the changes we'd expect (mockiato -> mockall) and CI should start passing from there! We'll fix the failing tests afterwards when we update our dependencies the next time - thank you so much 🙏

nox1134:
Thanks for the review @tarikeshaq! Seems like it worked
How do i go about Check-formatting and other failed tests? I tried fixing the `clippy` tests but the issue was with the fact that whatever i fix in the `clippy` fails the cargo tests with `lifetime parameter` error.

components/fxa-client/src/internal/http_client.rs
```diff
@@ -53,37 +53,37 @@ const DEVICES_FILTER_DAYS: u64 = 21;
 /// it return something called a "refresh token"? Using unambiguous
 /// verbs to start each method helps avoid confusion here.
 ///
-#[cfg_attr(test, mockiato::mockable)]
+#[cfg_attr(test, mockall::automock)]
```
tarikeshaq, line 56: Lets tell clippy to ignore the issue here, looks like we're hitting a limitation with mockall!
```suggestion
// Due to limitations in mockall, we have to add explicit lifetimes that the Rust compiler would have been happy to infer.
#[allow(clippy::needless_lifetimes)]
#[cfg_attr(test, mockall::automock)]
```

tarikeshaq (changes requested):
Thanks again @nox1134! Lets ignore the clippy error in this case, and write a comment for our future selves why we did that (see my comment for what that looks like in practice)

for the formatting, try running `cargo fmt` then check-in the changes that the formatter makes

once we make those two changes we might be ready to land 👀 🎉

tarikeshaq:
We're dealing with an intermittent test for the `ios-test` job, so you can ignore that failure (see #5902)

tarikeshaq:
Awesome! This is almost ready to land, I ran the dependency check script (which attempts to make sure we include the licenses of any upstream dependencies) and there are some manual fixups that need to happen.. but not subject you to those fixups, I pushed a commit on top of your commits and will open it in a separate PR to land it (your commits are still there as-is to preserve authorship!)

components/fxa-client/Cargo.toml
```diff
@@ -26,13 +26,14 @@ thiserror = "1.0"
 anyhow = "1.0"
 sync-guid = { path = "../support/guid", features = ["random"] }
 uniffi = "0.24.1"
+mockall = "0.11"
```
tarikeshaq, line 29: Aaah! this is why the dependency summaries were failing, lets remove this from there, it should only be a test dependency!

This way we don't need any fixups and we can land your PR without having to create a new one 👀

tarikeshaq (dismissed):
Actually, I realized that the reason those fixups were needed was because we added `mockall` to the regular dependencies when it should've only lived as a test dependency!

If we fix this up we can land this PR without having to do any fixups to our scripts!

components/fxa-client/Cargo.toml
nox1134: okay! thank you

tarikeshaq (approved):
Thank you so much @nox1134 🙏 I'll go ahead and land this

tarikeshaq:
Aaah looks like one of the commits in your branch conflicts with the main branch (but not the full changset) - @nox1134 do you mind squashing your commits into a single commit with a descriptive name, then force-pushing? That way what lands is a clean single commit

nox1134:
I squashed all my commits and force-pushed!
Big thanks 💯  @tarikeshaq for guiding me through contributing to Mozilla — I learned a lot while working on this issue. 🚀

tarikeshaq (approved):

mhammond:
Hey @nox1134, thanks for your contribution and for sticking with this even after you hit a few hurdles!

PR #5910, opened 2023-11-09
Adding twin ADRS on wrapper code and async Rust

bendk:
These are independant but related decisions, so two ADRs in one pull request is a good way to handle them.

  - Note:

### DIFF
--- a/docs/adr/0008-wrapper-code.md
+++ b/docs/adr/0008-wrapper-code.md
@@ -1,0 +1,91 @@
+# Wrapper code
+
+* Status: draft
+* Deciders: 
+* Date: ???
+
+## Context and Problem Statement
+
+Application-services components currently consist of a Rust core, and UniFFI-generated bindings for Swift and Kotlin.
+Additionally, some of our components have hand-written Swift and Kotlin code, containing wrappers and extensions for the UniFFI bindings.
+In the past, these wrappers were strictly necessary because of deficiencies in our FFI strategy.
+However, UniFFI is reaching the point where it can support all our requirements and it's possible to remove the wrapper code.
+
+We should decide how much effort, if any, we want to invest into getting rid of the wrapper code now that it's technically possible.
+We should also decide if there are places where wrapper code makes sense and we don't want to replace it, even ignoring the time investment.
+
+### Possible changes
+
+This section explores some possibilities for removing wrapper code.
+Deciding on any particular possibility listed here is out of scope for the ADR, the decision is if we should invest our time investigating some of these.
+
+#### Async
+
+One of the main reasons for our current wrapper code is to wrap our sync API and present an async interface.
+ADR-0009 lays out a couple of alternatives to this.
+
+#### Feature flags for breaking changes
+
+Another main reason for our current wrapper code is to mitigate breaking API changes.
+The wrapper layer allows us to make breaking changes at the Rust level, but keep the same API at the wrapper layer.
+
+An alternative to this would be to use Rust feature flags to manage breaking changes.
+Any breaking change, or large change in general, would be behind a feature flag.
+We would wait to enable the feature flag on the megazord until consumer application was ready to take the change.
+Maybe we could have a transition period where we built two megazords, for example `megazord` and `megazord-with-logins-local-encrytion` and the consumer app could pick between the two.
+This would simplify the consumer PRs since they could run CI normally.
+
+Some potential uses of of features flags would be:
+
+- **Cosmetic changes**. If we want to rename a field/function name, we could put that rename behind a feature flag.
+- **Architectural changes**. For bigger changes, like the logins local encryption rework, we could maintain two pieces of code at once. For example, copy db.rs to db/old.rs and db/new.rs. Create db/mod.rs which runs `use old::*` or `use new::*` depending on the feature flag. Then we do our work in db/new.rs.
+
+Maintaining many feature flags at once would require significant effort, so we should aim to get our consumers to use the new feature flag soon after our work is complete.
+
+#### UniFFI-supported interfaces
+
+One last reason for wrapper code is to present idiomatic interfaces to our Rust code -- especially for callback interfaces.
+For example, it's possible to define a UniFFI callback interface for notifications, but the FxA wrapper code uses the `NotificationCenter` on Swift which is not supported by UniFFI.
+If we wanted to remove all wrapper code we would need to commit to only using interfaces that UniFFI could support.
+
+## Decision Drivers
+
+## Considered Options
+
+* **(A) Keep using our current strategy**
+Don't invest time into removing wrapper code.
+
+* **(B) Remove all wrapper code**
+Invest time into removing wrapper code with the intention of removing all of it.
+
+* **(C) Remove most wrapper code, keep an additive wrapper layer**
+Invest time into removing wrapper code, but keep some wrapper code.
+In particular, keep a wrapper layer that adds new API surfaces rather than replaces existing ones.
+For example, we could define an `FxaEventListener` interface with UniFFI, then add a `IosEventListener` class that implemented that interface by forwarding the messages to the `NotificationCenter`.
+
+## Decision Outcome
+
+## Pros and Cons of the Options
+
+### (A) Keep using our current strategy
+
+* Good, because we can spend our time on other improvements
+* Good, because there's no chance of wasting time on implementing solutions that may not work out in practice.
+
+### (B) Remove all wrapper code
+* Good, because it simplifies our documentation strategy.
+  There is active work in UniFFI to auto-generate the bindings documentation from the Rust docstrings (https://github.com/mozilla/uniffi-rs/pull/1498, https://github.com/mozilla/uniffi-rs/pull/1493).
+  If there are no wrappers, then we could potentially use this to auto-generate a high-level documentation site and/or docstrings in the generated bindings code.
+  If there are wrappers, then this probably isn't going to work.
+  In general, wrappers mean we are have multiple public APIs which makes documentation harder.
+* Bad, because it may lead to worse documentation.
+  Hand-written documents can be better than auto-generated ones and especially since they can specifically target javadoc and swiftdoc.
+* Good, because a "vanilla" API may be easier to integrate with multiple consumer apps.
+  `NotificationCenter` might be the preferred choice for firefox-ios, but another Swift app may want to use a different system.
+  By only using UniFFI-supported types we can be fairly sure that our code will work with any system.
+
+### (C) Remove most wrapper code, keep an additive wrapper layer
+* Good, because most documentation can be auto-generated and some can be hand-written.
+  The hand-written documentation would be the language-specific parts, which probably need to be written by hand.
+* Bad because it may lead to worse documentation, for the same reasons an (B).
+* Good, because consumer apps can choose to use the wrapper interfaces or not.
--- a/docs/adr/0009-async-rust.md
+++ b/docs/adr/0009-async-rust.md
@@ -1,0 +1,151 @@
+# Using Async Rust
+
+* Status: draft
+* Deciders: 
+* Date: ???
+
+## Context and Problem Statement
+
+Our Rust components are currently written using synchronous Rust.
+The components are then wrapped in Kotlin to present an async interface.
+Swift also wraps them to present an async-style interface, although it currently uses `DispatchQueue` and completion handlers rather than `async` functions.
+
+UniFFI has been adding async capabilities in the last year and it seems possible to switch to using async Rust and not having a hand-written async wrapper.
+It also seems possible to auto-generate the async wrapper with UniFFI.
+
+What should our async strategy be?
+
+### Scope
+
+This ADR discusses what our general policy on wrapper code should be.
+It does not cover how we should plan our work.
+If we decide to embrace async Rust, we do not need to commit to any particular timeline for actually switching to it.
+
+### Desktop and gecko-js
+
+On desktop, we can't write async wrappers because it's not possible in Javascript.
+Instead we use a strategy where every function is automatically wrapped as async in the C++ layer.
+Using a config file, it's possible to opt-out of this strategy for particular functions/methods.
+
+### Android-components
+
+In Kotlin, the async wrapper layer currently lives in `android-components`.
+For the purposes of this ADR, it doesn't really matter, and this ADR will not make a distinction between wrapper code in our repo and `android-components`.
+
+## How it would work
+
+### SQLite queries
+
+One of the reasons our code currently blocks is to run SQLite queries.
+https://github.com/mozilla/uniffi-rs/pull/1837 has a system to run blocking code inside an async function.
+It would basically mean replacing code like this:
+
+```kotlin
+    override suspend fun wipeLocal() = withContext(coroutineContext) {
+        conn.getStorage().wipeLocal()
+    }
+```
+
+with code like this:
+```rust
+
+    async fn wipe_local() {
+        self.queue.execute(|| self.db.wipe_local()).await
+    }
+```
+
+We would need to merge #1837, which is currently planned for the end of 2023.
+
+### Locks
+
+Another reason our code blocks is to wait on a `Mutex` or `RWLock`.
+There are a few ways we could handle this:
+
+* The simplest is to continue using regular locks, inside a `execute()` call, which would be very similar to our current system.
+* We could also consider switching to `async_lock` and reversing the order: lock first, then make a `execute()` call.
+  This may be more efficient since the async task would suspend while waiting for the lock rather than blocking a thread
+* We could also ditch locks and use [actors and channels](https://ryhl.io/blog/actors-with-tokio/) to protect our resources.
+  It's probably not worth rewriting our current components to do this, but this approach might be useful for new components.
+
+### Network requests
+
+The last reason we block is for network requests.
+To support that we would probably need some sort of "async viaduct" that would allow consumer applications to choose either:
+- Use async functions from the `reqwest` library.
+  This matches what we currently do for `firefox-ios`.
+- Use the foreign language's network stack via an async callback interface.
+  This matches what we currently do for `firefox-android`.
+  This would require implemnenting https://github.com/mozilla/uniffi-rs/issues/1729, which is currently planed for the end of 2023.
+
+## Decision Drivers
+
+## Considered Options
+
+* **(A) Experiment with async Rust**
+
+* Pick a small component like `tabs` or `push` and use it to test our async Rust.
+* Use async Rust for new components.
+* Consider slowly switching existing components to use async Rust.
+
+* **(B) Keep hand-written Async wrappers**
+
+Don't change the status quo.
+
+* **(C) Auto-generate Async wrappers**
+
+We could also make the `gecko-js` model the official model and switch other languages to use it as well.
+For example, we could support something like this in `uniffi.toml`:
+
+```toml
+[[bindings.async_wrappers]]
+# Class to wrap, methods wrapped with an async version
+wrapped = "LoginStore"
+# Name of the wrapper class.
+# UniFFI would generate async wrapper methods that worked exactly like the current hand-written code.
+# For most languages, the wrapper class constructors would input an extra parameter to handle the async wrapping (for example `CoroutineContext` or `DispatchQueue`).
+wrapper = "LoginStoreAsync"
+# methods to skip wrapping and keep sync (optional)
+sync_methods = [...]
+```
+
+We could also support async wrappers for callback interfaces.
+These would allow the foreign code to implement an sync callback interface using async code
+The Rust code would block while waiting for the result.
+
+## Decision Outcome
+
+## Pros and Cons of the Options
+
+### (A) Experiment with async Rust
+
+* Good, if we decide to avoid wrappers in `ADR-0008` because it allows us to remove the async wrappers.
+* Bad, because there's a risk that the UniFFI async code will cause issues and our current async strategy is working okay.
+  Even if we pick a small component to experiment with, it would be bad if that component crashes or stops responding because of async issues.
+* Good because it allows us to be more efficient with our thread usage.
+  When an async task is waiting on a lock or network request, it can suspend itself and release the thread for other async work.
+  Currently, we need to block a thread while we are waiting for this.
+  However, it's not clear this would meaningfully affect our consumers since we don't run that many blocking operations.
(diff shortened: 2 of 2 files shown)


docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Pros and Cons of the Options
+
+### (A) Embrace wrapper code and continue to use it
+
+* Good, because wrapper code allows us to wrap sync Rust code in async wrapper functions and this is a low-risk approach to async.
+* Good, because wrapper code allows us to mitigate breaking changes in the Rust code.
+* Bad, because there are better ways to handle breaking changes.
+  We could introduce breaking changes, and large changes in general, behind a Rust feature flag.
+  We would wait to enable the feature flag on the megazord until consumer application was ready.
```
bendk, line 39: This is my current feeling, but I'm not that confident in it.  What do others think?

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,58 @@
...
+
+* **(A) Embrace wrapper code and continue to use it**
+* **(B) Reject wrapper code and remove it**
+* **(C) Use wrapper code for async wrapping only**
+
+## Decision Outcome
+
+## Pros and Cons of the Options
+
+### (A) Embrace wrapper code and continue to use it
```
MatthewTighe, line 33: 👋 Happened to notice this and thought I'd stop in. Happy to provide a perspective on these proposals from the Android side if there is interest in that. That said, it would be helpful for my understanding to have concrete examples of current wrapper practices.

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+    }
+```
+
+with code like this:
+```rust
+
+    async fn wipe_local() {
+        self.queue.run_blocking(|| self.db.wipe_local()).await
+    }
+```
```
MatthewTighe, line 57: Would the Kotlin usage functionally change here? It seems like we would still need to run this in a coroutine to avoid blocking threads at that level

bendk: From the high level perspective, most/all of the sync team's components are currently async.  It's just that the async is being implemented by a wrapper in android-components.  The change would be to handle the wrapping inside Rust instead.

For components that currently present a sync API, I don't think there's any reason to change them to be async.

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,58 @@
...
+* Good, because wrapper code allows us to mitigate breaking changes in the Rust code.
+* Bad, because there are better ways to handle breaking changes, like Rust feature flags.
+  We could introduce breaking changes, and large changes in general, behind a feature flag.
+  We would wait to enable the feature flag on the megazord until consumer application was ready.
+* Good, because it allows us to present APIs using idiomatic, native, types.
+  For example, UniFFI callback interfaces may meet our requirements, but the Swift `NotificationCenter` may provide a more ergonomic API.
+
+### (B) Reject wrapper code and remove it
+
+* Good, because it simplifies our documentation strategy.
```
tarikeshaq, line 45: I don't fully agree with this one - I agree the active work in UniFFI will truly improve the documentation strategy, but I'm not sure that removing wrapper code makes the documentation story any better. Both javadocs and swift docs have very rich documentation systems that would be hard to fully conform to (found this great blog post about [swift's documentation](https://nshipster.com/swift-documentation/)) 

I still believe what we're doing in UniFFI is many times better than the status quo, where the generated code has no doc comments at all, but I doubt that story alone makes the documentation story better than hand-written docs in swift

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+
+We would need to merge this code, which is currently planed for the end of 2023.
```
tarikeshaq, line 59: To clarify, we would need to merge **support** for this in UniFFI, correct?

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Locks
+
+Another reason our code blocks is to wait on a `Mutex` or `RWLock`.
+To handle this, we could switch from using `parking_lot` to `async_lock`.
```
tarikeshaq, line 64: Just for discussion, I've heard opinions along the lines of "only use an async mutex if you absolutely have to", eg: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

Would the performance implication of switching to async mutexes anything to be worries about here?

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Network requests
+
+The last reason we block is for network requests.
+To support that we would probably need some sort of "async viaduct" that would allow consumer applications to choose either:
+- Use async functions from the `reqwest` library.
+  This matches what we currently do for `firefox-ios`.
+- Use the foreign language's network stack via an async callback interface.
+  This matches what we currently do for `firefox-android`.
+  This would require implemnenting https://github.com/mozilla/uniffi-rs/issues/1729, which is currently planed for the end of 2023.
```
tarikeshaq, line 75: +1 that if we embrace async, having async callback support will be critical for things like network support, I'd almost want to see how that system works in practice before making a final call here.

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+* **(B) Avoid Async Rust**
+
+## Decision Outcome
+
+## Pros and Cons of the Options
+
+### (A) Embrace Async Rust
+
+* Good, if we decide to avoid wrappers in `ADR-0008` because it allows us to remove the async wrappers.
+* Bad, because there's a risk that the UniFFI async code will cause issues and our current async strategy is working okay.
```
tarikeshaq, line 91: This is the largest `Bad` I see here, the cost of sticking to our existing, non-async Rust is cheap (no work!) but it might be expensive engineering time spent to make all the async bits work (async callback interfaces, migrating away from sync calls and the breaking changes that follow, async viaduct, etc)

There is also the cost of having to handle possible inconsistencies we never had to handle before (if we use async mutexes that drop locks on `await`s, allowing different tasks to interleave in ways we never had to deal with before)

This almost makes me want to take the position of: "It's too costly and risky to embrace because of uncertainty - however if we can quantify possible wins, let's try it on a small surface and realize those wins to give us the confidence we're doing the right thing"

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+* Good, because it makes it easier to integrate with new languages that expect async.
+  For example, WASM integration libraries usually returns `Future` objects to Rust which can only be evaluated in an async context.
+  Note: this is a separate issue from UniFFI adding WASM support.
+  If we switched our component code to using async Rust, it's possible that we could use `wasm-bindgen` instead.
+* Bad, because it makes it harder to provide bindings on new languages that don't support async, like C and C++.
+  Maybe we could bridge the gap with some sort of callback-based async system, but it's not clear how that would work.
+
+### (B) Avoid Async Rust
+
+This is basically just the inverse of the last section.
```
tarikeshaq, line 106: 'Good, this is the status quo, and doesn't require any work'

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+
+* Good, if we decide to avoid wrappers in `ADR-0008` because it allows us to remove the async wrappers.
+* Bad, because there's a risk that the UniFFI async code will cause issues and our current async strategy is working okay.
+* Good because it allows us to be more effecient with our thread usage.
+  When an async task is waiting on a lock or network request, it can suspend itself and release the thread for other async work.
+  Currently, we need to block a thread while we are waiting for this.
+  However, it's not clear this would meaningfully affect our consumers since we don't run that many blocking operations.
+  We would be saving maybe 1-2 threads at certain points.
+* Good, because it makes it easier to integrate with new languages that expect async.
+  For example, WASM integration libraries usually returns `Future` objects to Rust which can only be evaluated in an async context.
```
tarikeshaq, line 98: To provide an example for this, I've experimented with having the `fxa_client` compile to WASM and the dependency on NSS was a problem, I've opted to web apis exposed through the JS runtime the wasm is running in, and those web apis often return promises, that have to be wrapped in Futures.

The above makes it so it's difficult to expose our fxa_client to web applications without exposing an async Rust API.

docs/adr/0008-wrapper-code.md
lougeniaC64: What would this feature flag infrastructure look like? One the one hand I could see it being hugely beneficial for a lot of scenarios including releases backend logic for consumer features when the frontend isn't in place. On the other hand, this could get onerous to maintain with different consumers depending on the situation.

tarikeshaq (commented):
Thank you so much for taking this on and writing this out @bendk - I wrote down my initial thoughts going through this, but feel free to challenge my comments.

My own position on those two is: (but I'm curios to hear what others think as well!)
- For wrappers, I believe it's a case-by-case basis. And occasionally adding wrapper to make our API more idiomatic shouldn't be discouraged. That said, I do think it'll be educational to evaluate what are the existing cases that **couldn't** be satisfied with UniFFI, if the answer is "close to none", that would change my position
- For Async Rust, it's too early to commit to embracing it but I would love to see a small component running using Async Rust as support.
   -  It's an expensive commitment, so being diligent with quantifying the wins would go a long way
   - Maybe the decision could be along the lines of "Lets experiment with Async Rust, starting with the XX component, then re-evaluate". That way we commit to trying it, and commit to re-visiting the large question of embracing it.

docs/adr/0009-async-rust.md
jonalmeida: > There is also the cost of having to handle possible inconsistencies we never had to handle before (if we use async mutexes that drop locks on awaits, allowing different tasks to interleave in ways we never had to deal with before)

I think you are describing my concern here from the embedder perspective: we have our platform async system (Coroutines) and we will have the native rust async system which have to both sync (heh) together correctly because the race bugs from this would be painful to debug.

bendk: Agreed that this is the biggest concern.

FWIW, UniFFI uses the platform async system to schedule the Rust work, so races don't seem that much more likely to me.  If there's a bug, I would expect either hangs if the task gets lost somehow or just regular exceptions / crashes.

bendk:
> For Async Rust, it's too early to commit to embracing it but I would love to see a small component running using Async Rust as support. 

I wouldn't want to move faster than that either.  I think calling the strategy "embracing Async Rust" is too much, it should be something more like "timidly waving at Async Rust from 100ft away".

docs/adr/0009-async-rust.md
bendk: Agreed.  I think it's a case-by-case basis thing.  If we're just going to run a simple query, than a standard mutex is probably better.  I think the main reason we'd want to take an async lock is if we expect to hold it for a long time, for example it might be useful in the the places cooperative transaction strategy, where the sync connection wants exclusive access to the DB for a second or two.

That last sentence is pretty interesting too.  Maybe some mutexes would be replaced with [actors and channels](https://ryhl.io/blog/actors-with-tokio/).

docs/adr/0008-wrapper-code.md
bendk: I'll try to add this as a new bullet.  Maybe it's most precise to say that auto-generating the docs would simplify things, but we might end up with worse docs if we don't hand-write the wrapper doc-strings.

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,106 @@
...
+    override suspend fun wipeLocal() = withContext(coroutineContext) {
+        conn.getStorage().wipeLocal()
+    }
+```
+
+with code like this:
+```rust
+
+    async fn wipe_local() {
+        self.queue.run_blocking(|| self.db.wipe_local()).await
```
linabutler, line 55: I wonder if it would help to use a different name than `run_blocking` here, like `launch` or `execute`, to help convey the semantics?

In the Rust world—if I'm understanding your example right—`run_blocking` means "run this blocking code asynchronously".

But in Kotlin, [`runBlocking`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html) means "run this asynchronous function and block the current thread until it's done"—the very opposite of what your Rust code is doing!

It's a very small change—I know that it's just an example; and we can decide on all the names later 😊—but I wonder if it would help avoid confusion, given that `run_blocking` means two totally opposite things depending on if you're coming from Rust or Kotlin. WDYT?

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,58 @@
...
+It does not cover how we should plan our work.
+If we decide to reject wrapper code, we do not need to commit to any particular timeline for actually removing it.
+
+## Decision Drivers
+
+## Considered Options
+
+* **(A) Embrace wrapper code and continue to use it**
+* **(B) Reject wrapper code and remove it**
+* **(C) Use wrapper code for async wrapping only**
```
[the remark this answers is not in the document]
bendk: I think these options are wrong and basically everyone supports some form of `D`.  AFAICT, no one is advocating for keeping wrapper code for the sake of having wrappers and nor does anyone believe that wrapper code is so bad we should remove it regardless of the consequences.  I'm going to update the ADR to frame the question better.  I think the real question is how much effort should we put into removing wrapper code.

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,58 @@
+# Wrapper code
+
+* Status: draft
+* Deciders: 
+* Date: ???
+
+## Context and Problem Statement
+
+Application-services components currently consist of a Rust core that gets wrapped by Swift and Kotlin wrappers.
```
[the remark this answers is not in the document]
bendk: `android-components` kind of confuses things, but we don't need to mention that here.  Is there a better way to write this sentence?

bendk:
Thanks for all the feedback here.  I just pushed an update with a bunch of changes, the main ones are:

* Significantly re-framed the wrapper questions.  Instead of "do we like wrappers or not?", it's now "how much effort are we willing to put into removing wrappers?"
* Added an option (C) for the async question

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,151 @@
...
+### Desktop and gecko-js
+
+On desktop, we can't write async wrappers because it's not possible in Javascript.
+Instead we use a strategy where every function is automatically wrapped as async in the C++ layer.
+Using a config file, it's possible to opt-out of this strategy for particular functions/methods.
+
+### Android-components
+
+In Kotlin, the async wrapper layer currently lives in `android-components`.
+For the purposes of this ADR, it doesn't really matter, and this ADR will not make a distinction between wrapper code in our repo and `android-components`.
```
linabutler, line 33: At first, I was wondering if this was a bit inconsistent with your "Wrapper code" ADR, because that ADR (seems to?) explicitly exclude wrappers outside of Application Services.

But thinking about it a bit more, the way you've written this makes sense to me! All our wrappers—wherever they live—do two things:

1. Wrap the synchronous Rust component interface in an asynchronous foreign interface.
2. Bridge the Rust interface to platform-specific Android and iOS interfaces (`WorkManager`, `BGTaskScheduler`, `UIKit`, etc.)

We can lift (1) into Application Services, and still keep the wrappers in our consuming applications for (2).

docs/adr/0009-async-rust.md
linabutler: Agreed! This one might also need some support from Android Components; IIRC, it currently wraps GeckoView's async networking client in a blocking interface. It would be really compelling (and a fair bit of work! 😅) to have networking be one of our first uses of async callbacks, though!

docs/adr/0009-async-rust.md
```diff
@@ -0,0 +1,151 @@
...
+  This matches what we currently do for `firefox-android`.
+  This would require implemnenting https://github.com/mozilla/uniffi-rs/issues/1729, which is currently planed for the end of 2023.
+
+## Decision Drivers
+
+## Considered Options
+
+* **(A) Experiment with async Rust**
+
+* Pick a small component like `tabs` or `push` and use it to test our async Rust.
```
linabutler, line 86: 🙋🏼‍♀️ I volunteer `remote_settings` as ~a tribute~ another test component!

* It has a small API surface.
* It could be a good "launch customer" for async networking.
* It uses a lock to keep track of some client state, and races there shouldn't be catastrophic.
* It doesn't do any persistence (e.g., to SQLite) directly.
* It has a [small Android Components wrapper](https://github.com/mozilla-mobile/firefox-android/blob/44e0f1b7ae93850a58044743be5173a8e74d9582/android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt), that won't be too disruptive to migrate.

linabutler (commented):
These both look great; thank you so much for taking the time to write them up!

docs/adr/0009-async-rust.md
bendk: First off: Thanks for offering this!  Secondly: maybe we should be thinking about this component-by-component.  

Async makes sense for `remote_settings` so that's a natural place to start exploring.

Async seems to make the least sense for our syncable-storage components.  I wonder if we should just lean into saying the non-async API is the "real" API.  We could still have an async wrapper class, but this would be more of an add-on.  In practice, this would mean that when publish documentation, we would create detailed documentation for the sync API while the async wrapper would just have some text like "all methods work the same as LoginsStore, but are wrapped with `withContext` to present an async API"

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Considered Options
+
+* **(A) Keep using our current strategy**
+Don't invest time into removing wrapper code.
+
+* **(B) Remove all wrapper code**
+Invest time into removing wrapper code with the intention of removing all of it.
+
+* **(C) Remove most wrapper code, keep an additive wrapper layer**
```
lougeniaC64, line 61: I agree with other comments that suggest we need to analyze our current wrapper code to see how much of it we can remove and that the result will inform our future strategy. I for one am assuming that we won't be able to remove all of it so we will end up at option C. We also might want to have a one-to-one wrapper to consumer strategy so that we can put the more platform-specific, opinionated code there. But in order to reduce our API surface I think we should do what we can to reduce the wrapper code we have.

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,90 @@
...
+### (B) Remove all wrapper code
+* Good, because it simplifies our documentation strategy.
+  There is active work in UniFFI to auto-generate the bindings documentation from the Rust docstrings (https://github.com/mozilla/uniffi-rs/pull/1498, https://github.com/mozilla/uniffi-rs/pull/1493).
+  If there are no wrappers, then we could potentially use this to auto-generate a high-level documentation site and/or docstrings in the generated bindings code.
+  If there are wrappers, then this probably isn't going to work.
+  In general, wrappers mean we are have multiple public APIs which makes documentation harder.
+* Bad, because it may lead to worse documentation.
+  Hand-written documents can be better than auto-generated ones and especially since they can specifically target javadoc and swiftdoc.
+* Good, because a "vanilla" API may be easier to integrate with multiple consumer apps.
+  `NotificationCenter` might be the preferred choice for firefox-ios, but another Swift app may want to use a different system.
```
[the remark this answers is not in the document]
mhammond: Not wanting to distract from the main point here, but it would be super interesting to know how uniffi "CustomTypes" work here - eg, we could probably use something like `Image` in the UDL, which is just an alias for `Vec<u8>`, and Android and iOS each has special code that knows how to turn that vec into the local type.

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,91 @@
...
+  By only using UniFFI-supported types we can be fairly sure that our code will work with any system.
+
+### (C) Remove most wrapper code, keep an additive wrapper layer
+* Good, because most documentation can be auto-generated and some can be hand-written.
+  The hand-written documentation would be the language-specific parts, which probably need to be written by hand.
+* Bad because it may lead to worse documentation, for the same reasons an (B).
+* Good, because consumer apps can choose to use the wrapper interfaces or not.
```
mhammond, line 91: I think I'd be leaning towards an option described as something like "as few wrappers as possible, but no less" - ie, a perfect world has no wrappers, but this world is going to force *some* on us. Deciding "no wrappers" would be a fool's errand and not survive reality.

I guess this is closest to (C) Is there some feature uniffi is missing to enable this without the "bad" parts? It almost gets back to the "decorators" idea we had all that time ago - ie, something more like "augmentation" than "wrappers"?

mhammond (commented):
Sorry for the delay and sorry that I only got to the first ADR - I'll look at the second soon!

docs/adr/0008-wrapper-code.md
```diff
@@ -0,0 +1,91 @@
...
+- **Cosmetic changes**. If we want to rename a field/function name, we could put that rename behind a feature flag.
+- **Architectural changes**. For bigger changes, like the logins local encryption rework, we could maintain two pieces of code at once. For example, copy db.rs to db/old.rs and db/new.rs. Create db/mod.rs which runs `use old::*` or `use new::*` depending on the feature flag. Then we do our work in db/new.rs.
+
+Maintaining many feature flags at once would require significant effort, so we should aim to get our consumers to use the new feature flag soon after our work is complete.
+
+#### UniFFI-supported interfaces
+
+One last reason for wrapper code is to present idiomatic interfaces to our Rust code -- especially for callback interfaces.
+For example, it's possible to define a UniFFI callback interface for notifications, but the FxA wrapper code uses the `NotificationCenter` on Swift which is not supported by UniFFI.
+If we wanted to remove all wrapper code we would need to commit to only using interfaces that UniFFI could support.
```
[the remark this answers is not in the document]
bendk: I think it would be more like "our swift consumers need to do the work to hook up a callback interface to the NotificationCenter themselves".  Definitely a drawback but not totally out of the question.  I do prefer `C` over `B` at this point though.

bendk:
Thanks for all the great discussion here.  We never came to a conclusion, but I think that's largely because my ADRs weren't asking the questions in the right way.  I just opened a new version of this here: https://github.com/mozilla/application-services/pull/6549.  Let's continue the discussion there.

PR #5911, opened 2023-11-09
Added prune_limit pramater to the places maintenance

bendk:
This allows us to adjust the numbers until we reach a point where we are effectively limiting the DB size

https://bugzilla.mozilla.org/show_bug.cgi?id=1854383

  - Note:


components/places/ios/Places/Places.swift
```diff
@@ -462,7 +462,7 @@ public class PlacesWriteConnection: PlacesReadConnection {
     open func runMaintenance(dbSizeLimit: UInt32 = 0) throws {
         return try queue.sync {
             try self.checkApi()
-            _ = try self.conn.runMaintenancePrune(dbSizeLimit: dbSizeLimit)
+            _ = try self.conn.runMaintenancePrune(dbSizeLimit: dbSizeLimit, pruneLimit: 6)
```
mhammond, line 465: future us might like a comment explaining why this is not the same as Android?

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
```diff
@@ -344,7 +344,7 @@ class PlacesWriterConnection internal constructor(conn: UniffiPlacesConnection,
     override fun runMaintenance(dbSizeLimit: UInt) {
         val pruneMetrics = PlacesManagerMetrics.runMaintenanceTime.measure {
             val pruneMetrics = PlacesManagerMetrics.runMaintenancePruneTime.measure {
-                this.conn.runMaintenancePrune(dbSizeLimit)
+                this.conn.runMaintenancePrune(dbSizeLimit, 12U)
```
mhammond, line 347: I know we want to make conservative changes here, but I wonder if this specifically can be more aggressive. On. Android in particular, I'm still not sure that this runs often enough to keep up with an active user.

Specifically, maybe doubling (ie, 24) again is ok? I think I'd even be fine with more.

mhammond (approved):

components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt
bendk: I'd be fine with that, but there's also a firefox-android PR to increase the frequency by two.  So if we bumped it to 24 we would be increasing the rate 8x.  Seems fine with me, but I wanted to double check that.

bendk:
It's [merged on Android](https://github.com/mozilla-mobile/firefox-android/pull/4432).  Should we merge this one here?  I'm happy to bump again to 24 before merging, either way seems fine with me.

mhammond:
Don't let my last-minute rambling derail more important things, let's just keep it simple :)

bendk:
Simple sounds great to me, let's merge.

PR #5951, opened 2023-11-29
Add Yelp suggestions feature

dadaa:

  - [] This PR follows the breaking change policy:
  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -115,7 +115,7 @@ impl SuggestDb {
 /// reference (`&mut self`).
 pub(crate) struct SuggestDao<'a> {
     pub conn: &'a Connection,
-    scope: SqlInterruptScope,
+    pub scope: SqlInterruptScope,
 }
 
 impl<'a> SuggestDao<'a> {
@@ -125,6 +125,10 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        if let Some(suggestion) = self.fetch_yelp_suggestion(query)? {
+            return Ok(vec![suggestion]);
+        }
+
         let keyword_lowercased = &query.keyword.to_lowercase();
         let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
         let suggestions_limit = query.limit.unwrap_or(-1);
@@ -294,7 +298,8 @@ impl<'a> SuggestDao<'a> {
                         } else {
                             Ok(None)
                         }
-                    }
+                    },
+                    _ => Ok(None),
                 }
             }
         )?.flat_map(Result::transpose).collect::<Result<_>>()?;
@@ -615,6 +620,18 @@ impl<'a> SuggestDao<'a> {
     pub fn drop_suggestions(&mut self, record_id: &SuggestRecordId) -> Result<()> {
         self.conn.execute_cached(
             "DELETE FROM suggestions WHERE record_id = :record_id",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
+            "DELETE FROM yelp_subjects WHERE record_id = :record_id",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
+            "DELETE FROM yelp_modifiers WHERE record_id = :record_id",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
+            "DELETE FROM yelp_location_signs WHERE record_id = :record_id",
             named_params! { ":record_id": record_id.as_str() },
         )?;
         Ok(())
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -90,6 +90,8 @@ pub(crate) enum SuggestRecord {
     Amo,
     #[serde(rename = "pocket-suggestions")]
     Pocket,
+    #[serde(rename = "yelp-suggestions")]
+    Yelp,
 }
 
 /// Represents either a single value, or a list of values. This is used to
@@ -265,4 +267,22 @@ pub(crate) struct DownloadedPocketSuggestion {
     #[serde(rename = "highConfidenceKeywords")]
     pub high_confidence_keywords: Vec<String>,
     pub score: f64,
+}
+/// A location sign for Yelp to ingest from a Yelp Attachment
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct DownloadedYelpLocationSign {
+    pub keyword: String,
+    #[serde(rename = "needLocation")]
+    pub need_location: bool,
+}
+/// A Yelp suggestion to ingest from a Yelp Attachment
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct DownloadedYelpSuggestion {
+    pub subjects: Vec<String>,
+    #[serde(rename = "preModifiers")]
+    pub pre_modifiers: Vec<String>,
+    #[serde(rename = "postModifiers")]
+    pub post_modifiers: Vec<String>,
+    #[serde(rename = "locationSigns")]
+    pub location_signs: Vec<DownloadedYelpLocationSign>,
 }
--- a/components/suggest/src/schema.rs
+++ b/components/suggest/src/schema.rs
@@ -6,7 +6,7 @@
 use rusqlite::{Connection, Transaction};
 use sql_support::open_database::{self, ConnectionInitializer};
 
-pub const VERSION: u32 = 9;
+pub const VERSION: u32 = 10;
 
 pub const SQL: &str = "
     CREATE TABLE meta(
@@ -78,6 +78,24 @@ pub const SQL: &str = "
         id TEXT PRIMARY KEY,
         data BLOB NOT NULL
     ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_subjects(
+        keyword TEXT PRIMARY KEY,
+        record_id TEXT NOT NULL
+    ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_modifiers(
+        type INTEGER NOT NULL,
+        keyword TEXT NOT NULL,
+        record_id TEXT NOT NULL,
+        PRIMARY KEY (type, keyword)
+    ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_location_signs(
+        keyword TEXT PRIMARY KEY,
+        need_location INTEGER NOT NULL,
+        record_id TEXT NOT NULL
+    ) WITHOUT ROWID;
 ";
 
 /// Initializes an SQLite connection to the Suggest database, performing
@@ -108,7 +126,7 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
 
     fn upgrade_from(&self, _db: &Transaction<'_>, version: u32) -> open_database::Result<()> {
         match version {
-            1..=8 => {
+            1..=9 => {
                 // These schema versions were used during development, and never
                 // shipped in any applications. Treat these databases as
                 // corrupt, so that they'll be replaced.
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -282,6 +282,7 @@ where
                 })?;
                 continue;
             };
+
             match fields {
                 SuggestRecord::AmpWikipedia => {
                     self.ingest_suggestions_from_record(
@@ -328,6 +329,16 @@ where
                         record,
                         |dao, record_id, suggestions| {
                             dao.insert_pocket_suggestions(record_id, suggestions)
+                        },
+                    )?;
+                }
+                SuggestRecord::Yelp => {
+                    self.ingest_suggestions_from_record(
+                        writer,
+                        record,
+                        |dao, record_id, suggestions| match suggestions.first() {
+                            Some(suggestion) => dao.insert_yelp_suggestions(record_id, suggestion),
+                            None => Ok(()),
                         },
                     )?;
                 }
@@ -1705,6 +1716,17 @@ mod tests {
                 "size": 0,
             },
         }, {
+            "id": "data-4",
+            "type": "yelp-suggestions",
+            "last_modified": 15,
+            "attachment": {
+                "filename": "data-4.json",
+                "mimetype": "application/json",
+                "location": "data-4.json",
+                "hash": "",
+                "size": 0,
+            },
+        }, {
             "id": "icon-2",
             "type": "icon",
             "last_modified": 20,
@@ -1813,6 +1835,20 @@ mod tests {
                 },
             ]),
         )?
+        .with_data(
+            "data-4.json",
+            json!({
+                "subjects": ["ramen", "spicy ramen", "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789", "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789Z"],
+                "preModifiers": ["best", "super best", "same_modifier"],
+                "postModifiers": ["delivery", "super delivery", "same_modifier"],
+                "locationSigns": [
+                    { "keyword": "in", "needLocation": true },
+                    { "keyword": "near", "needLocation": true },
+                    { "keyword": "near by", "needLocation": false },
+                    { "keyword": "near me", "needLocation": false },
+                ]
+            }),
+        )?
         .with_icon("icon-2.png", "i-am-an-icon".as_bytes().into())
         .with_icon("icon-3.png", "also-an-icon".as_bytes().into());
 
@@ -1830,6 +1866,7 @@ mod tests {
                         SuggestionProvider::Wikipedia,
                         SuggestionProvider::Amo,
                         SuggestionProvider::Pocket,
+                        SuggestionProvider::Yelp,
                     ],
                     limit: None,
                 },
@@ -1846,6 +1883,7 @@ mod tests {
                         SuggestionProvider::Wikipedia,
                         SuggestionProvider::Amo,
                         SuggestionProvider::Pocket,
+                        SuggestionProvider::Yelp,
                     ],
                     limit: None,
                 },
@@ -2393,6 +2431,340 @@ mod tests {
                 []
                 "#]],
             ),
+            (
+                "keyword = `best spicy ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "best spicy ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=best+spicy+ramen+delivery&find_loc=tokyo",
+                        title: "best spicy ramen delivery in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `BeSt SpIcY rAmEn DeLiVeRy In ToKyO`; Yelp only",
+                SuggestionQuery {
+                    keyword: "BeSt SpIcY rAmEn DeLiVeRy In ToKyO".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=BeSt+SpIcY+rAmEn+DeLiVeRy&find_loc=ToKyO",
+                        title: "BeSt SpIcY rAmEn DeLiVeRy In ToKyO",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `best ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "best ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=best+ramen+delivery&find_loc=tokyo",
+                        title: "best ramen delivery in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `best invalid_ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "best invalid_ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `best delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "best delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `super best ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "super best ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=super+best+ramen+delivery&find_loc=tokyo",
+                        title: "super best ramen delivery in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `invalid_best ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "invalid_best ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `ramen delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen+delivery&find_loc=tokyo",
+                        title: "ramen delivery in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen super delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen super delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen+super+delivery&find_loc=tokyo",
+                        title: "ramen super delivery in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen invalid_delivery in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen invalid_delivery in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `ramen in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen&find_loc=tokyo",
+                        title: "ramen in tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen near tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen near tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen&find_loc=tokyo",
+                        title: "ramen near tokyo",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen invalid_in tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen invalid_in tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `ramen in San Francisco`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen in San Francisco".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen&find_loc=San+Francisco",
+                        title: "ramen in San Francisco",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen in`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen in".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen",
+                        title: "ramen in",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen near by`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen near by".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen+near+by",
+                        title: "ramen near by",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen near me`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen near me".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen+near+me",
+                        title: "ramen near me",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `ramen near by tokyo`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen near by tokyo".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `ramen`; Yelp only",
+                SuggestionQuery {
+                    keyword: "ramen".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=ramen",
+                        title: "ramen",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = maximum chars; Yelp only",
+                SuggestionQuery {
+                    keyword: "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                [
+                    Yelp {
+                        url: "https://www.yelp.com/search?find_desc=012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789",
+                        title: "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789",
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = over chars; Yelp only",
+                SuggestionQuery {
+                    keyword: "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789Z".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `best delivery`; Yelp only",
+                SuggestionQuery {
+                    keyword: "best delivery".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `same_modifier same_modifier`; Yelp only",
+                SuggestionQuery {
+                    keyword: "same_modifier same_modifier".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
+            (
+                "keyword = `same_modifier `; Yelp only",
+                SuggestionQuery {
+                    keyword: "same_modifier ".into(),
+                    providers: vec![SuggestionProvider::Yelp],
+                    limit: None,
+                },
+                expect![[r#"
+                []
+                "#]],
+            ),
         ];
         for (what, query, expect) in table {
             expect.assert_debug_eq(
@@ -2483,10 +2855,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                         },
                     ),
@@ -2551,10 +2923,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                         },
                     ),
@@ -2657,10 +3029,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 9,
+                                schema_version: 10,
                             },
                         },
                     ),
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -1,0 +1,362 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+use rusqlite::types::ToSqlOutput;
+use rusqlite::{named_params, Result as RusqliteResult, ToSql};
+use sql_support::ConnExt;
+use url::form_urlencoded;
+
+use crate::{
+    db::SuggestDao,
+    provider::SuggestionProvider,
+    rs::{DownloadedYelpSuggestion, SuggestRecordId},
+    suggestion::Suggestion,
+    Result, SuggestionQuery,
+};
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[repr(u8)]
+enum Modifier {
+    PreModifier = 0,
+    PostModifier = 1,
+}
+
+impl ToSql for Modifier {
+    fn to_sql(&self) -> RusqliteResult<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(*self as u8))
+    }
+}
+
+/// This module assumes like following query.
+/// "Pre-modifier? Subject Post-modifier? (Location-modifier | Location-sign Location?)?"
+/// For example, the query below is valid.
+/// "Best(Pre-modifier) Ramen(Subject) Delivery(Post-modifier) In(Location-sign) Tokyo(Location)"
+/// Also, as everything except Subject is optional, "Ramen" will be also valid query.
+/// However, "Best Best Ramen" and "Ramen Best" is out of the above appearance order rule,
+/// parsing will be failed. Also, every words except Location needs to be registered in DB.
+/// Please refer to the query test in store.rs for all of combination.
+/// Currently, the maximum query length is determined while refering to having word lengths in DB
+/// and location names.
+/// max subject: 50 + pre-modifier: 10 + post-modifier: 10 + location-sign: 7 + location: 50 = 127 = 150.
+const MAX_QUERY_LENGTH: usize = 150;
+
+/// The max number of words consisting the modifier. To improve the SQL performance by matching with
+/// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
+const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
+
+impl<'a> SuggestDao<'a> {
+    /// Inserts the suggestions for Yelp attachment into the database.
+    pub fn insert_yelp_suggestions(
+        &mut self,
+        record_id: &SuggestRecordId,
+        suggestion: &DownloadedYelpSuggestion,
+    ) -> Result<()> {
+        for keyword in &suggestion.subjects {
+            self.scope.err_if_interrupted()?;
+            self.conn.execute_cached(
+                "INSERT INTO yelp_subjects(record_id, keyword) VALUES(:record_id, :keyword)",
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":keyword": keyword,
+                },
+            )?;
+        }
+
+        for keyword in &suggestion.pre_modifiers {
+            self.scope.err_if_interrupted()?;
+            self.conn.execute_cached(
+                "INSERT INTO yelp_modifiers(record_id, type, keyword) VALUES(:record_id, :type, :keyword)",
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":type": Modifier::PreModifier,
+                    ":keyword": keyword,
+                },
+            )?;
+        }
+
+        for keyword in &suggestion.post_modifiers {
+            self.scope.err_if_interrupted()?;
+            self.conn.execute_cached(
+                "INSERT INTO yelp_modifiers(record_id, type, keyword) VALUES(:record_id, :type, :keyword)",
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":type": Modifier::PostModifier,
+                    ":keyword": keyword,
+                },
+            )?;
+        }
+
+        for sign in &suggestion.location_signs {
+            self.scope.err_if_interrupted()?;
+            self.conn.execute_cached(
+                "INSERT INTO yelp_location_signs(record_id, keyword, need_location) VALUES(:record_id, :keyword, :need_location)",
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":keyword": sign.keyword,
+                    ":need_location": sign.need_location,
+                },
+            )?;
+        }
+
+        Ok(())
+    }
+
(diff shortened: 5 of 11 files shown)


0c0w3:
Thanks @dadaa! I'll need to continue looking at this tomorrow. For now here are some initial thoughts in random order:

Move Yelp helpers into yelp.rs, similar to pocket.rs?

We should be able to use a single RS record to store all Yelp data. We chunk the other suggestions into chunks of 200, and they also contain a lot more data than Yelp suggestions need -- titles, URLs, lots of keywords. I don't believe all the Yelp data would be bigger than one AMP chunk for example? (The Yelp keywords sheet seems to have been deleted, so I can't check it.) The record schema could look like: `{ subjects, preModifiers, postModifiers, locationSigns, templates }`

`find_keyword_and_split()` is clever but it does a full table scan as it looks up each `value` in the query. It's not the end of the world because this dataset is pretty small (and I haven't looked in detail but I think other suggestion types may be doing full table scans too?), but I think there's room for improvement. Would it be possible to do something [like I suggested on your earlier JS patch](https://phabricator.services.mozilla.com/D189335#6271394)? If we did that, and we also added indexes on the pre and post modifier tables and the subject table, then lookup would be `O(number of words in the query)` instead of `O(number of subjects)`.

Just thinking out loud: It would be nice to implement a providers system like urlbar and Merino use, so `fetch_suggestions()` would loop through a `providers` list and call a `query()` method on each provider and not need to know much about the different implementations. That's outside the scope of this PR of course.

dadaa:
Thank you very much for the feedback, @0c0w3!
I will try them.

0c0w3:
> and we also added indexes on the pre and post modifier tables and the subject table

I forgot Sqlite creates indexes for primary key columns, so with the schema in this patch, creating indexes isn't necessary.

linabutler (commented):
Thanks so much for this PR, @dadaa, and @0c0w3 for the initial review pass!

I'm _really_ excited that you're contributing to the Rust component! 🎉🚀 We should see about adding both of you (and anyone else from the SNG team who's interested?) to the `mozilla/application-services-collaborators` team, so that you can run CI, tag and be tagged for PR  reviews, and so on.

Replying to @0c0w3's comment:

> I think other suggestion types may be doing full table scans too?), but I think there's room for improvement.

The other suggestion types avoid full table scans 😊 (You're totally right, though; the Yelp dataset sounds like it's small enough where one would still be totally okay).

AMP and Wikipedia use an exact match on the primary key, and AMO and Pocket have a compound primary key for exact matching on the first word, and prefix matching everything after, both to reduce the number of rows we scan.

...Which gives me some ideas about your next suggestion:

> Would it be possible to do something [like I suggested on your earlier JS patch?](https://phabricator.services.mozilla.com/D189335#6271394)

Oh, I really like your approach there—and your distillation of the matching algorithm made it very easy to follow!

Noodling on it just a bit more, I thought of a couple of things that we could try in SQL:

*1*. We could store all the modifiers (pre- and post-) and location signs in a single table, with a compound primary key over `(type, modifier)`. IIUC, we might be able to query this table in one go, and use the result of that query to figure out which words to remove from the search string.

Something like:

```sql
-- The first number is the number of words to remove from the start.
-- The second is the number of words to remove from the end.
SELECT 1, 0 FROM modifiers WHERE (modifier, type) = (PRE, :firstWord)
UNION ALL
SELECT 0, 1 FROM modifiers WHERE (modifier, type) = (POST, :lastWord)
UNION ALL
SELECT 0, 2 FROM modifiers WHERE (modifier, type) = (POST, :lastTwoWords)
```

For space efficiency, we could also use this table to store all the location signs, with a special `type`. This set looks so small that reading it all into a `Vec<String>`, and matching each one against the words in the search string, might be even more efficient than trying to craft a SQL query with the substring check.

**2.** We can store all the cities in another table, with a primary key over the city name. If the search string contains `"in"` (or another location sign?) we can do a prefix match for everything after it, like you suggested in your comment, and remove the location sign and city from the search string if we have a match:

```sql
SELECT EXISTS(
  SELECT 1 FROM cities
  WHERE name BETWEEN :everythingAfterIn AND :everythingAfterIn || x'ffff'
)
```

**3.** Finally, after cleaning up the search string, we can query the `subjects` table, with an exact match on the primary key.

**4.** Do we expect to have more than one URL template? It looks like `https://www.yelp.com/search?find_desc={description}&find_loc={location}` is the only one, so it could even make sense to have it as a constant, instead of storing it in the database.

I'm sure the devil will be in the details, but I wanted to jot down those quick first impressions!

> It would be nice to implement a providers system like urlbar and Merino use, so `fetch_suggestions()` would loop through a `providers`` list and call a `query()` method on each provider and not need to know much about the different implementations.

+1, it would be lovely to see this in a future refactoring! :trophy:

dadaa:
Thank you very much for your feedback, @linabutler!
I'm sorry, as I have fixed another bug, and could not take the time to look at this today. I will check it next week!

dadaa:
> Replying to @0c0w3's comment:
> 
> > I think other suggestion types may be doing full table scans too?), but I think there's room for improvement.
> 
> The other suggestion types avoid full table scans 😊 (You're totally right, though; the Yelp dataset sounds like it's small enough where a scan would still be okay).

The reason that I did this was I wanted to reduce the number of SQL executed, especially when we can't find any `subject`.  So, only the initial query can determine if the user query has a `subject`. On the other hand, if the subject is examined last, SQL and string manipulations for pre/post modifiers and location may be wasted.

It seems that the number of currently registered candidates is 860. I’m not sure how often this number will provide Yelp suggestions.
https://docs.google.com/spreadsheets/d/1MPW_mm0c3q_ZBJhAh-nc0V04BYrQDb-QR3dpP6bD4Fg/edit#gid=1467415472
I just wanted to think about when we can't suggest.

However, as you both said, it seems that the performance due to full table scans might be low.

> > Would it be possible to do something [like I suggested on your earlier JS patch?](https://phabricator.services.mozilla.com/D189335#6271394)
> 
> I like your approach—and your distillation of the matching algorithm made it very easy to follow!

Okay!

> Noodling on it just a bit more, I thought of a couple of things that we could try to make that same algorithm work in SQL:
> 
> __1_._ We could store all the modifiers (pre- and post-) and location signs in a single table, with a compound primary key over `(type, modifier)`. IIUC how modifiers work, we might be able to query this table in one go, and use the result of that query to figure out which words to remove from the search string.
> 
> Something like:
> 
> ```sql
> -- The first number is the number of words to remove from the start.
> -- The second is the number of words to remove from the end.
> SELECT 1, 0 FROM modifiers WHERE (type, modifier) = (PRE, :firstWord)
> UNION ALL
> SELECT 0, 1 FROM modifiers WHERE (type, modifier) = (POST, :lastWord)
> UNION ALL
> SELECT 0, 2 FROM modifiers WHERE (type, modifier) = (POST, :lastTwoWords)
> ```

Oh, cool, interesting!

> **2.** We can store all the cities in another table, with a primary key over the city name. If the search string contains `"in"` (or another location sign?) we can do a prefix match for everything after it, like you suggested in your comment, and then remove the location sign and city from the search string if we have a match:

Yes, there are other words besides `"in"`.  `"close to"`, `"near"`, and so on, but not been fixed yet.
However, I don't think we can store all city names since it will be super huge :)

> **3.** Finally, after cleaning up the search string, we can query the `subjects` table, with an exact match on the primary key.
> 
> **4.** Do we expect to have more than one URL template? It looks like `https://www.yelp.com/search?find_desc={description}&find_loc={location}` is the only one, so it could even make sense to have it as a constant in the code, instead of storing it in the database.

I'm not sure yet, but I imagine that there might be a possibility the URL will be changed to add a parameter for like telemetry etc.

So, as a result, now, I am thinking about below:

1. Find the `location-sign` and `location` from the user's query. It does full table scans, but the `location-sign` is very few, it will be better compared to full scans of the `subjects`.
2. Remove `location-sign` and `location` string from the query.
3. Find pre/post modifiers using like SQL suggested in 1 Lina suggested.
4. Remove the above words.
5. Find `subject` by the remaining query.

> I'm sure the devil will be in the details, but I wanted to jot down those quick first impressions!

Thank you so much! It was very helpful!

> > It would be nice to implement a providers system like urlbar and Merino use, so `fetch_suggestions()` would loop through a `` providers`` list and call a  ``query()` method on each provider and not need to know much about the different implementations.
> 
> +1, it would be lovely to see this in a future refactoring! 🏆

Cool!

dadaa:
Hi @0c0w3, @linabutler!
I updated the patch and addressed your comments.
When you have time, could you check? Thanks!

0c0w3 (commented):
Thanks Daisuke, overall this is looking good to me, and I'd like to hear what Lina thinks. I'm still a little doubtful about whether we should try to handle locations, and how we should do it, but the approach here is simple enough I'm not too worried about it, and we can give it a shot and always iterate on it.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Fetch Yelp suggestion from given user's query.
+pub(crate) fn fetch_suggestion(
+    conn: &Connection,
+    query: &SuggestionQuery,
+) -> Result<Option<Suggestion>> {
+    if !query.providers.contains(&SuggestionProvider::Yelp) {
+        return Ok(None);
+    }
+
+    if !&query.keyword.contains(" ") {
```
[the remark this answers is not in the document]
dadaa: Yeah, I agree!
At this time, the suggestions module can provide several suggestions including Yelp, Wikipedia.
And, the limitation might change by the suggestion. For example, Yelp can accept 100 chars, Wikipedia  255 chars etc. If so, the consumer can't restrict since does not know which suggestion will be returned by the user's query on the consumer side. And I think it would be more independent. So, I feel we can restrict here.

On a similar note, and related to the Yelp title, I feel that it might be better not to make the query lowercase in the consumer. As much as possible, why not just pass the query as it is? What do you think??
https://searchfox.org/mozilla-central/rev/16526d690cccc9c60cacf09cc08e2c3041ef884e/browser/components/urlbar/private/SuggestBackendRust.sys.mjs#100

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    // Find the modifiers.
+    let (subject_candidate, pre_modifier, post_modifier) = if query_without_location.contains(" ") {
+        let mut words: Vec<&str> = query_without_location.split_whitespace().collect();
+        let pre_modifier_candidate = words[0];
+        let post_modifier_candidate = words[words.len() - 1];
+
+        let (is_pre_modifier, is_post_modifier) =
+            is_modifiers(conn, pre_modifier_candidate, post_modifier_candidate)?;
```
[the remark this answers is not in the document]
dadaa: The modifiers' candidates are below, for now.
pre: `cheap`, `good`, `affordable`, `budget`
post: `nearby`, `near by`, `delivery`, `in [city]`
And for post-modifier, except `delivery` can be `location sign`, I think. If we can do so, can assume all modifiers are single word. I will ask this to Chris.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        if is_pre_modifier {
+            words.remove(0);
+        }
+        if is_post_modifier {
+            words.pop();
+        }
+
+        (
+            words.join(" "),
```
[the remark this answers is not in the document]
dadaa: Okay, I will try that!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    if let Some(loc) = &location {
+        url_template = url_template.replace("{location}", &loc);
+    }
+
+    let Ok(url) = Url::parse(&url_template) else {
+        return Ok(None);
+    };
+
+    // Build Title.
```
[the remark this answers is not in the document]
dadaa: Yeah yeah, super cool idea, indeed!
One concern is that the query passed here from JS is lowercase. Is not necessary if we need to capitalize the initial letter and the beginning of the place name? Or, as I wrote above though, can we pass the query as it is without changing to lowercase?

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
0c0w3: I agree with both of your comments (restricting here in the Rust component and not lowercasing on desktop). I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1868922 for the desktop part.

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
0c0w3: That's true -- if the consumer uses the query string as the title, as I suggested, then the title will use whatever case the user used, probably just lowercase. I think that's fine and I don't think it's worth the effort to try and capitalize certain substrings, which we might not do a perfect job of anyway.

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
dadaa: As discussed on Slack, we handle `near by`, `near me` and related word as `location sign`.
And, we will extend the `location_sign` table to know whether the sign needs specific location.
And, if that doesn't need the location, we don't set the location to the URL and the Title.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,319 @@
...
+    if let (Some(sign), false) = (&location_sign, need_location) {
+        // This location sign such the 'near by' needs to add as a description parameter.
+        description.append(&mut sign.split_whitespace().collect::<Vec<&str>>());
+    }
+
+    let mut url_template: String = conn
+        .query_row_and_then("SELECT url FROM yelp_templates", [], |row| {
+            row.get::<_, String>(0)
+        })?
+        .replace("{description}", &description.join("+"));
```
0c0w3, line 172: To summarize, you're splitting the modifiers and subject into vecs of words, putting them all in `description`, and then joining the words with `"+"`, right? And the reason is to URL-encode the description, essentially, right? I think we might want to do some proper encoding here, which will be safer in case the subject or modifiers contain characters that should be encoded. I did a quick search for a Rust equivalent to `encodeURIComponent()` and found [this](https://rust-lang-nursery.github.io/rust-cookbook/encoding.html), and then did a quick search in the application-services repo and found that the `places` component [already uses](https://github.com/mozilla/application-services/blob/1d259497eb43e6fb416667dbdf81723e9e04d2e2/components/places/src/match_impl.rs#L371-L374) the [percent_encoding](https://docs.rs/percent-encoding/2.3.1/percent_encoding/) crate. So we can use that here too, and then we don't have to worry about vetting it when we vendor this code into m-c. Does that sound OK?

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    if let (Some(loc), true) = (&location, need_location) {
+        url_template = url_template.replace("{location}", &loc);
```
0c0w3, line 175: And we'll want to encode the location too.

components/suggest/src/yelp.rs
dadaa: Thanks! Yes, I intended to do that.  And I did the `percent-encoding` using `Url::parse()`.

The reason why I have done this is if we search on Yelp, it seems that they replace the whitespace of keywords with `+'` not `%20`.
e.g. https://www.yelp.com/search?find_desc=Ramen+Near+Me&find_loc=San+Francisco
So, to make it exactly match with the Yelp URL, I did it. I will try to find a better way.

dadaa: https://docs.rs/form_urlencoded/latest/form_urlencoded/ ?

components/suggest/src/rs.rs
```diff
@@ -266,3 +268,22 @@ pub(crate) struct DownloadedPocketSuggestion {
     pub high_confidence_keywords: Vec<String>,
     pub score: f64,
 }
+
+/// A Yelp dataset to ingest from a Yelp Attachment
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct LocationSign {
+    pub keyword: String,
+    pub need_location: i8,
```
0c0w3, line 276: Is there a reason this can't be a `bool`? (I don't know anything about Serde yet so I'm happy to learn 🙂)

components/suggest/src/schema.rs
```diff
@@ -80,6 +80,25 @@ pub const SQL: &str = "
...
+    ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_location_signs(
+        keyword TEXT PRIMARY KEY,
+        need_location INTEGER NOT NULL
+    ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_templates(
+        url TEXT PRIMARY KEY
+    ) WITHOUT ROWID;
```
0c0w3, line 101: I agree with @linabutler that simply storing this in memory somewhere seems simpler and sufficient. In practice, there should only ever be one template I think. If there's more than one, then it's likely because we're providing two different UXes to the user, e.g., one might be a Yelp SERP URL, another might be a direct link to a business's Yelp page or something. And in that case, it's likely that each UX will need to know its specific URL template, so storing them all together in a table doesn't really seem helpful.

0c0w3 (commented):
Thank you @dadaa, this looks good to me! I don't see any tests -- didn't an earlier version have them?

components/suggest/src/yelp.rs
0c0w3: > and then did a quick search in the application-services repo and found that the `places` component [already uses](https://github.com/mozilla/application-services/blob/1d259497eb43e6fb416667dbdf81723e9e04d2e2/components/places/src/match_impl.rs#L371-L374) the [percent_encoding](https://docs.rs/percent-encoding/2.3.1/percent_encoding/) crate. So we can use that here too, and then we don't have to worry about vetting it when we vendor this code into m-c.

Correction w/r/t vetting, this probably isn't true because `places` isn't used in m-c, but I'm not 100% sure.

linabutler: It looks like `percent_encoding` and `form_urlencoded` are both already vendored and vetted (🎉), because they're dependencies of `url`.

[Bug 1866135](https://bugzilla.mozilla.org/show_bug.cgi?id=1866135) was the last one to bump the vendored and vetted versions for both  `percent_encoding` and `form_urlencoded`, so I think either one should work, without extra work from us!

linabutler:
(Sorry for the delay, I'll find some time to look at this before the end of the week!)

components/suggest/src/rs.rs
dadaa: Oh, indeed, it seems that we can use `bool` here. Thanks!

components/suggest/src/schema.rs
dadaa: Okay, I will do it!

components/suggest/src/store.rs
```diff
@@ -331,6 +334,33 @@ where
                         },
                     )?;
                 }
+                SuggestRecord::Yelp => {
+                    let Some(attachment) = record.attachment.as_ref() else {
```
ncloudioj, line 338: There is a helper function [`ingest_suggestion_from_record()`](https://github.com/mozilla/application-services/blob/main/components/suggest/src/store.rs#L339) to use to save some boilerplate code, can you try using that here? You can find examples in the `AMO` and `Pocket` branches above.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+    PostModifier = 1,
+}
+
+impl ToSql for Modifier {
+    fn to_sql(&self) -> RusqliteResult<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(*self as u8))
+    }
+}
+
+/// max subject: 50 + pre-modifier: 10 + post-modifier: 10 + location-sign: 7 + location: 50 = 127 = 150.
```
ncloudioj, line 31: Could you please briefly outline the query strategy here? A link to the doc for details also works.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+        &subject_candidate,
+        pre_modifier,
+        post_modifier,
+        location_sign,
+        location,
+        need_location,
+    )
+}
+
+fn to_suggestion(
```
ncloudioj, line 134: Can you also add a docstring for this function describing what it does? An example would be even nicer.

components/suggest/src/rs.rs
```diff
@@ -90,6 +90,8 @@ pub(crate) enum SuggestRecord {
     Amo,
     #[serde(rename = "pocket-suggestions")]
     Pocket,
+    #[serde(rename = "yelp-dataset")]
```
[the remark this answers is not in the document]
dadaa: Thank you very much for your review, @ncloudioj!
The reason why I named this was that we are not preparing complete suggestions, unlike other suggestions.
But, okay, I will rename it!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+        conn.execute(
+            "INSERT INTO yelp_location_signs(keyword, need_location) VALUES(:keyword, :need_location)",
+            named_params! {
+                ":keyword": sign.keyword,
+                ":need_location": sign.need_location,
+            },
+        )?;
+    }
+
+    *URL_TEMPLATE.lock().unwrap() = dataset.template.to_string();
```
linabutler, line 81: I left a comment on `URL_TEMPLATE` about this.

components/suggest/src/rs.rs
```diff
@@ -266,3 +268,22 @@ pub(crate) struct DownloadedPocketSuggestion {
     pub high_confidence_keywords: Vec<String>,
     pub score: f64,
 }
+
+/// A Yelp dataset to ingest from a Yelp Attachment
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct LocationSign {
+    pub keyword: String,
+    pub need_location: bool,
```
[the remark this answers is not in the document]
dadaa: What about renaming it with `DownloadedYelpLocationSign`, for example, to keep the consistency of meaning that the data was downloaded? And, keep this snake case. What do you think? I will do it, but if you have another opinion, please let me know!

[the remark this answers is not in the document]
dadaa: Ah, sorry, I might misunderstand. You meant, how to set the `need_location` data, right?  Yes, I agree, will change to camel case.

components/suggest/src/yelp.rs
linabutler: The number of arguments (and optionals) that this function takes definitely makes it a little tricky to read—and it also does a combination of I/O (via `is_subject`) and string manipulation.

It'll be a bit more code, but I'm wondering how you'd feel about:

1. Lifting the `is_subject` check into `fetch_suggestion`. That function already calls other methods that do I/O (like `find_location` and `find_modifiers`), and then this function can focus on just the string manipulation. This would also make it easier to unit test separately, without having to pull in a `Connection`.
2. Changing this function to either take a struct of options—or make it a method on that struct!

Something like:

```rs
struct SuggestionBuilder<'a> {
    query: &'a SuggestionQuery,
    subject: &'a str,
    pre_modifier: Option<String>,
    // ...
}

impl<'a> From<SuggestionBuilder<'a>> for Suggestion {
    fn from(builder: SuggestionBuilder<'a>) -> Suggestion {
        let mut description = String::from(builder.subject);
        
        if let Some(modifier) = &self.pre_modifier {
            // ...
        }

        Suggestion::Yelp {
            url,
            // Use user’s query as title as it is.
            title: builder.query.keyword.clone(),
        }
    }
}
```

Then you could use it like this to build your `Suggestion` from a `SuggestionBuilder`:

```rs
let builder = SuggestionBuilder {
    query: query,
    subject: &subject,
    // ...
};

Ok(Some(suggestion.into()))
```

How does that sound?

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+        description.push(' ');
+        description.push_str(&sign);
+    }
+
+    let mut url = URL_TEMPLATE.lock().unwrap().clone();
+    url = url.replace(
+        "{description}",
+        &form_urlencoded::byte_serialize(description.as_bytes())
+            .collect::<Vec<_>>()
+            .concat(),
```
linabutler, line 173: There's a [`Serializer` struct](https://docs.rs/form_urlencoded/latest/form_urlencoded/struct.Serializer.html#method.finish) that might be a bit more ergonomic to use than converting back and forth to bytes.

linabutler (changes requested):
Thanks so much for working on this, @dadaa, and for folding in all the feedback changes so far!

This is looking good, and I think I need a bit more time to digest the modifier and location sign processing—but I didn't want to block the other feedback I had on that.

I'll be back from holiday break on January 8th, and would be happy to keep working with you on this then, if it's still open—and @ncloudioj and @0c0w3 can also help with additional reviews and getting it landed, if they're around and you'd like to get it finished up sooner!

components/suggest/src/rs.rs
```diff
@@ -266,3 +268,22 @@ pub(crate) struct DownloadedPocketSuggestion {
     pub high_confidence_keywords: Vec<String>,
     pub score: f64,
 }
+
+/// A Yelp dataset to ingest from a Yelp Attachment
```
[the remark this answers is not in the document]
dadaa: Oh, indeed. I will add a comment for the location sign as well.

components/suggest/src/schema.rs
```diff
@@ -80,6 +80,21 @@ pub const SQL: &str = "
         id TEXT PRIMARY KEY,
         data BLOB NOT NULL
     ) WITHOUT ROWID;
+
+    CREATE TABLE yelp_subjects(
+        keyword TEXT PRIMARY KEY
```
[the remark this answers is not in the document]
dadaa: I see! Thanks!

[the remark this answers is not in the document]
linabutler: Before you land, we'll want to bump the `VERSION` constant, and bump [this version range](https://github.com/mozilla/application-services/blob/abfb70513cdce28755583e0f831bf33b82d05b70/components/suggest/src/schema.rs#L128), too.

Here's an example: https://github.com/mozilla/application-services/pull/6016

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn to_sql(&self) -> RusqliteResult<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(*self as u8))
+    }
+}
+
+/// max subject: 50 + pre-modifier: 10 + post-modifier: 10 + location-sign: 7 + location: 50 = 127 = 150.
+const MAX_QUERY_LENGTH: usize = 150;
+const MAX_MODIFIER_WORDS_LENGTH: usize = 2;
+
+static URL_TEMPLATE: Mutex<String> = Mutex::new(String::new());
```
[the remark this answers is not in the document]
dadaa: Thank you, Lina! Okay, I will write as hardcoding now. I will modify this to get from RS, if necessary.

components/suggest/src/yelp.rs
dadaa: Oh, cool! I will do that :)

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+    query: &str,
+) -> Result<(String, Option<String>, Option<String>, bool)> {
+    let results: Vec<(String, String, String, i8)> = conn.query_rows_and_then(
+        "
+        SELECT
+          TRIM(SUBSTR(q, 0, i)) AS without_location,
+          TRIM(SUBSTR(q, i, LENGTH(k))) AS location_sign,
+          TRIM(SUBSTR(q, i + LENGTH(k))) AS location,
+          need_location
+        FROM (
```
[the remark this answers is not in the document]
dadaa: Yeah, I will try it!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+        subject_candidate.trim().to_string(),
+        pre_modifier,
+        post_modifier,
+    ))
+}
+
+fn is_modifier(conn: &Connection, word: &str, modifier_type: Modifier) -> Result<bool> {
+    let count = conn.query_row_and_then(
+        "
+        SELECT COUNT(*) FROM yelp_modifiers WHERE (type, keyword) = (:type, LOWER(:word))
```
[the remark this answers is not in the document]
dadaa: I see, thanks!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,322 @@
...
+    // Build URL.
+    let mut description = String::from(subject);
+
+    if let Some(modifier) = &pre_modifier {
+        description.insert(0, ' ');
+        description.insert_str(0, &modifier);
+    }
+
+    if let Some(modifier) = &post_modifier {
+        description.push(' ');
```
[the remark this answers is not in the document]
dadaa: Wao, it seems Rust World :) I will do it, thanks!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+) -> Result<(String, Option<String>, Option<String>)> {
+    if !query.contains(" ") {
+        return Ok((query.to_string(), None, None));
+    }
+
+    let words: Vec<_> = query.split_whitespace().collect();
+
+    let mut pre_modifier = None;
+    for i in 0..MAX_MODIFIER_WORDS_LENGTH {
+        let mut candidate_chunks = words.chunks(MAX_MODIFIER_WORDS_LENGTH - i);
```
ncloudioj, line 235:
```suggestion
    for n in (1..=MAX_MODIFIER_WORDS_LENGTH).rev() {
        let mut candidate_chunks = words.chunks(n);
```

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+        let candidate = candidate_chunks.next().unwrap().join(" ");
```
ncloudioj, line 236: Since you only call `candidate_chunks.next()` once inside the foo loop, it will only check the _first_ chunk, is that expected?

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+        if is_modifier(conn, &candidate, Modifier::PreModifier)? {
+            pre_modifier = Some(candidate);
+            break;
+        }
+    }
+
+    let mut post_modifier = None;
+    for i in 0..MAX_MODIFIER_WORDS_LENGTH {
+        let mut candidate_chunks = words.rchunks(MAX_MODIFIER_WORDS_LENGTH - i);
+        let candidate = candidate_chunks.next().unwrap().join(" ");
```
ncloudioj, line 247: Likewise, this would only check the _last_ chunk for post-modifiers. Did you mean to only check the first two words for pre-modifier(s) and the last two words for post-modifier(s)? If so, I'd suggest calling `candidate_chunks.nth(0)` to make it clearer.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+            break;
+        }
+    }
+
+    let mut subject_candidate = query;
+    if let Some(ref modifier) = pre_modifier {
+        subject_candidate = &subject_candidate[modifier.len()..];
+    }
+    if let Some(ref modifier) = post_modifier {
+        subject_candidate = &subject_candidate[..subject_candidate.len() - modifier.len()];
```
ncloudioj, line 259: In the special case that there is only a pre-modifier and a post-modifier but not the subject, e.g. `best delivery`, would it only return a empty string as the subject?

components/suggest/src/db.rs
```diff
@@ -125,6 +125,10 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        if let Some(suggestion) = yelp::fetch_suggestion(self.conn, query)? {
```
ncloudioj, line 128: Just a note for future refactors. We might want to refactor this to move it to the query execution block down below. Otherwise, all the calls regardless of the interested providers will unconditionally call `yelp::fetch_suggestion()`.

components/suggest/src/schema.rs
```diff
@@ -6,7 +6,7 @@
 use rusqlite::{Connection, Transaction};
 use sql_support::open_database::{self, ConnectionInitializer};
 
-pub const VERSION: u32 = 9;
+pub const VERSION: u32 = 10;
```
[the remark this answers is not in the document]
dadaa: Oh, indeed, I will fix it!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+/// "Best(Pre-modifier) Ramen(Subject) Delivery(Post-modifier) In(Location-sign) Tokyo(Location)"
+/// Also, as everything except Subject is optional, "Ramen" will be also valid query.
+/// However, "Best Best Ramen" and "Ramen Best" is out of the above appearance order rule,
+/// parsing will be failed. Also, every words except Location needs to be registered in DB.
+/// Please refer to the query test in store.rs for all of combination.
+/// Currently, the maximum query length is determined while refering to having word lengths in DB
+/// and location names.
+/// max subject: 50 + pre-modifier: 10 + post-modifier: 10 + location-sign: 7 + location: 50 = 127 = 150.
+const MAX_QUERY_LENGTH: usize = 150;
+const MAX_MODIFIER_WORDS_LENGTH: usize = 2;
```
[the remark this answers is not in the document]
dadaa: Ah, I'm sorry, it makes you confused.
This is the number of words that make up the modifier. There are no such modifiers now, but we want to accept the modifier consisting of multiple words such as the `very best`. Now, we are trying to improve the SQL performance by matching with `keyword=:modifier` (please see `is_modifier()`).  So, we define this here how many words we should check. I will change the variable name and add the comment. Thanks!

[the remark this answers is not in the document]
dadaa: So, it does not mean multiple modifiers. We assume at most one pre/post-modifier.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,343 @@
...
+                None
+            } else {
+                Some(location.to_string())
+            },
+            *need_location == 1,
+        ))
+    }
+}
+
+fn find_modifiers(
```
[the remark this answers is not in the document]
dadaa: `find_location` also parses the original query and returns the location and the remaining query. So, I wanted to name that focuses on the modifier, what do you think?

components/suggest/src/yelp.rs
dadaa: Oh, cool, thanks!

components/suggest/src/yelp.rs
dadaa: If the `words` are `["very", "best", "ramen", "delivery"]`, I wanted to check `very best` and `best`. So, we get the chunk by `MAX_MODIFIER_WORDS_NUMBER`..1 and then get its first value. So, it is expected.

components/suggest/src/yelp.rs
ncloudioj: Another special case: if both the pre and the post modifiers are the same, and the `query` is exactly the modifier, this will crash because `subject_candidate.len() - modifier.len()` would overflow. Let's be extra cautious when manipulating those indices.

ncloudioj (commented):
Sorry for the delay. I just dropped a few more comments, but I think we're pretty close :)

ncloudioj:
@dadaa could you also do a rebase against the main branch because we've made some new changes there?

components/suggest/src/store.rs
dadaa: The reason why I did so was the data of Yelp suggestion is not an array, unlike other suggestions.
But, yes, I will use `ingest_suggestion_from_record()`, as it looks better to keep the deserialization consistency. Thanks!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,350 @@
...
+    query: &SuggestionQuery,
+) -> Result<Option<Suggestion>> {
+    if !query.providers.contains(&SuggestionProvider::Yelp) {
+        return Ok(None);
+    }
+
+    if query.keyword.len() > MAX_QUERY_LENGTH {
+        return Ok(None);
+    }
+
```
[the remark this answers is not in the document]
dadaa: Thanks!
Now, as we are keeping upper/lower case for the query of URL like `https://www.yelp.com/search?find_desc=Best+Ramen`, I wanted to pass the keyword as it is. 
However, it seems that the result of the URL is the same regardless of string cases. The only concern I have is the query form on Yelp. As it reflects the query as it is, might be better to do the same thing.
Please see the screenshot. What do you think?

Screenshot 2024-01-09 at 11 07 22

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,350 @@
...
+) -> Result<(String, Option<String>, Option<String>, bool)> {
+    let query_with_spaces = format!(" {} ", query.trim());
+    let results: Vec<(usize, usize, i8)> = conn.query_rows_and_then(
+        "
+        SELECT
+          INSTR(:query, ' ' || keyword || ' ') AS sign_index,
+          LENGTH(keyword) AS sign_length,
+          need_location
+        FROM yelp_location_signs
+        WHERE sign_index ORDER BY sign_length DESC LIMIT 1
```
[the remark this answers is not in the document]
dadaa: Thanks!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,350 @@
...
+fn find_location(
+    conn: &Connection,
+    query: &str,
+) -> Result<(String, Option<String>, Option<String>, bool)> {
+    let query_with_spaces = format!(" {} ", query.trim());
+    let results: Vec<(usize, usize, i8)> = conn.query_rows_and_then(
+        "
+        SELECT
+          INSTR(:query, ' ' || keyword || ' ') AS sign_index,
+          LENGTH(keyword) AS sign_length,
```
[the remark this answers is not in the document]
dadaa: Ah, it might be the same answer of upper/lower cases. 
https://github.com/mozilla/application-services/pull/5951/files#r1445554770
Especially, we set `location_sign` that has `need_location` = 0 such as the `near by` to the `find_desc` parameter. So, to keep the string cases, I am doing it so. If we can use lowercase for the all characters for URL, yes, your suggestion looks much better!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,350 @@
...
+                Some(location_sign.to_string())
+            },
+            if location.is_empty() {
+                None
+            } else {
+                Some(location.to_string())
+            },
+            *need_location == 1,
+        ))
+    }
```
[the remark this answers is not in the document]
dadaa: Thanks!

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
dadaa: I'd like to hear your opinion as well, @0c0w3!

components/suggest/src/yelp.rs
dadaa: > In the special case that there is only a pre-modifier and a post-modifier but not the subject, e.g. `best delivery`, would it only return a empty string as the subject?

Yes, then, as we check it by `is_subject()`, it should be no problem, I think.
But, the `is_subject()` does not check whether the query is empty, I will add the check.
Thanks!

> Another special case: if both the pre and the post modifiers are the same, and the `query` is exactly the modifier, this will crash because `subject_candidate.len() - modifier.len()` would overflow. Let's be extra cautious when manipulating those indices.

If the query was one word such as the `best`, it was possible to touch the failure.
But, we check whether the query has a whitespace at the beginning, the failure never happens, I think.
https://github.com/mozilla/application-services/pull/5951/files#diff-22307d3249667224ac6502dabf1e4b86e075333c0f279d93b126d25a8d7bb7c8R119
To be safe, I add the tests.

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
dadaa: Ah, it seems that we can do trimming here at least. I will do it.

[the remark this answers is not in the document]
ncloudioj: I'd be surprised if yelp handles queries case sensitively. If so, please disregard my suggestion :)

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,377 @@
...
+        post_modifier,
+        location_sign,
+        location,
+        need_location,
+    };
+    Ok(Some(builder.into()))
+}
+
+/// Find the location information from the given query string.
+/// The returned result is consited from tuple of below.
```
ncloudioj, line 172:
```suggestion
/// It returns the location tuple as follows:
```

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,377 @@
...
+    };
+    let pre_location = query_with_spaces
+        .get(..sign_index)
+        .map(str::trim)
+        .map(str::to_string)
+        .unwrap_or_default();
+    let location_sign = query_with_spaces
+        .get(sign_index..sign_index + sign_length)
+        .map(str::trim)
+        .unwrap_or_default();
```
ncloudioj, line 222: If `.get(sign_index..sign_index + sign_length)` returns `None`, all the subsequent `map` calls will also return `None`, so no need to call `unwrap_or_default()` at the end because we want `location_sign` to be `Option<String>`. Likewise, you don't need
```rs
        if location_sign.is_empty() {
            None
        } else {
            Some(location_sign.to_string())
        }
```
down below if you do
```rs
    let location_sign = query_with_spaces
        .get(sign_index..sign_index + sign_length)
        .map(str::trim)
        .map(str::to_string);
```
As `location_sign` is exactly of type `Option<String>`. `.unwrap_or_default()` is called for `pre_location` because it wants a `String` rather than `Option<String>` :)

ncloudioj (approved):
Looks good. Thanks! r+ with some minor comments.

components/suggest/src/yelp.rs
dadaa: Thanks, @ncloudioj!
Yeah, I had thought so. But it seems that if the `get()` gets `""`, it returns `Some(""))`.
For example, if run the following code, it shows `Some("")`.

```
let r = "test".get(0..0);
println!("{:?}", r);
```

As I wanted to handle empty string as `None`, checked it explicitly. If you have better idea, could you tell me, please?

ncloudioj: Ah, I see. You wanted to handle the empty string case. I think you can attach a `filter()` on it.

```
    let location_sign = query_with_spaces
        .get(sign_index..sign_index + sign_length)
        .filter(|s| !s.is_empty())
        .map(str::trim)
        .map(str::to_string);
```

dadaa: Cool! 
I changed the order of the processing like the below to trim first before filtering, it works fine!
Thank you so much!

```
    let location_sign = query_with_spaces
        .get(sign_index..sign_index + sign_length)
        .map(str::trim)
        .filter(|s| !s.is_empty())
        .map(str::to_string);
```

ncloudioj (approved):
r+ 🚢

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+/// Insert Yelp suggestions into db.
+pub(crate) fn insert_suggestions(
+    conn: &Connection,
+    scope: &SqlInterruptScope,
+    record_id: &SuggestRecordId,
+    suggestion: &DownloadedYelpSuggestion,
+) -> Result<()> {
+    for keyword in &suggestion.subjects {
+        scope.err_if_interrupted()?;
+        conn.execute(
```
linabutler, line 58:
```suggestion
        conn.execute_cached(
```

Let's cache the statements when we run them. I think using bulk insert like @ncloudioj suggested would be a bigger performance boost, but caching would let us reuse the underlying SQLite prepared statement.

linabutler (approved):
Thanks @dadaa! And thank you for reviewing while I was out, @ncloudioj!

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    Ok((
+        subject_candidate.trim().to_string(),
+        pre_modifier,
+        post_modifier,
+    ))
+}
+
+fn is_modifier(conn: &Connection, word: &str, modifier_type: Modifier) -> Result<bool> {
+    let result = conn.query_row_and_then(
```
linabutler, line 287:
```suggestion
    let result = conn.query_row_and_then_cachable(
```

(And `true` for the extra last argument, to cache it 😊)

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    Ok(result)
+}
+
+fn is_subject(conn: &Connection, word: &str) -> Result<bool> {
+    if word.is_empty() {
+        return Ok(false);
+    }
+
+    let result = conn.query_row_and_then(
```
linabutler, line 308:
```suggestion
    let result = conn.query_row_and_then_cachable(
```

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+/// and location names.
+/// max subject: 50 + pre-modifier: 10 + post-modifier: 10 + location-sign: 7 + location: 50 = 127 = 150.
+const MAX_QUERY_LENGTH: usize = 150;
+
+/// The max number of words consisting the modifier. To improve the SQL performance by matching with
+/// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
+const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
+
+/// Insert Yelp suggestions into db.
+pub(crate) fn insert_suggestions(
```
linabutler, line 50: Actually, I'm wondering—can you turn these functions into methods on `SuggestDao`, please?

I was thinking a bit more about it, and realized that you're passing the `Connection` (and sometimes the `SqlInterruptScope`) around as arguments—and that's actually why `SuggestDao` exists as a type! 😄

In Rust, `impl` blocks can live in different files—they don't have to live in the same file that declares the type, and you can have more than one—so you can declare this block right in `yelp.rs`:

```rust
impl<'a> SuggestDao<'a>  {
    pub fn insert_yelp_suggestions(&mut self, record_id: &SuggestRecordId, suggestion: &DownloadedYelpSuggestion) -> -> Result<()> {
        // ...
    }

    pub fn fetch_suggestion(&self, query: &SuggestionQuery) -> Result<Option<Suggestion>> {
        // ...
    }

    fn find_location(query: &str) -> -> Result<(String, Option<String>, Option<String>, bool)> {
        // ...
    }

    fn find_modifiers(query: &str)-> Result<(String, Option<String>, Option<String>)> {
        // ...
    }

    // ...
}
```

By convention, `&mut self` methods write, and `&self` methods read—it helps with separating which methods need a transaction.

WDYT?

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
0c0w3: Typically in urlbar we try to keep the user's casing and sometimes whitespace, that's true. I don't feel too strongly about it here. As Nan said, Yelp probably isn't case sensitive, so practically it shouldn't matter. If we need a tie breaker, I guess I would agree with Nan that we should normalize in one place since that would make the code a little simpler. I wouldn't expect most users to care about casing or even use uppercase when searching anyway.

0c0w3 (approved):
Thanks @dadaa, lgtm! There have been a lot of comments since the new year and I might have missed some that were addressed to me -- please let me know if so!

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
dadaa: Yeah, the Yelp results don't seem to change by the string casing.
However, the string the user inputs into the Yelp form (like the screenshot) will remain in the same case on the result page. That's part of it, I thought that reflecting the cases user inputs to the form might be natural.
What do you think? If we don't have to care about it, I will change the code! Or, shall we consider this in a follow-up?

components/suggest/src/store.rs
```diff
@@ -331,6 +332,15 @@ where
                         },
                     )?;
                 }
+                SuggestRecord::Yelp => {
+                    self.ingest_suggestions_from_record(
+                        writer,
+                        record,
+                        |dao, record_id, suggestions| {
+                            dao.insert_yelp_suggestions(record_id, &suggestions[0])
```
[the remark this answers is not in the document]
dadaa: I see, thanks!

components/suggest/src/yelp.rs
dadaa: Cool! I will try it!

dadaa:
@linabutler 
I implemented Yelp-related code as `SuggestDao`. Could you check it, just to be safe??

components/suggest/src/store.rs
```diff
@@ -331,6 +332,15 @@ where
                         },
                     )?;
                 }
+                SuggestRecord::Yelp => {
+                    self.ingest_suggestions_from_record(
+                        writer,
+                        record,
+                        |dao, record_id, suggestions| {
+                            dao.insert_yelp_suggestions(record_id, suggestions.first())
```
linabutler, line 340:
```suggestion
                            match suggestions.first() {
                                Some(suggestion) => dao.insert_yelp_suggestions(record_id, suggestion),
                                None => Ok(()),
                            }
```

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub fn insert_yelp_suggestions(
+        &self,
+        record_id: &SuggestRecordId,
+        maybe_suggestion: Option<&DownloadedYelpSuggestion>,
+    ) -> Result<()> {
+        if maybe_suggestion.is_none() {
+            return Ok(());
+        }
+
+        let suggestion = maybe_suggestion.unwrap();
```
linabutler, line 60: It's a bit more idiomatic to use a `match` or a `let-else` to unwrap optionals:

```rust
let Some(suggestion) = maybe_suggestion else {
    return Ok(());
};
```

But I'd actually suggest moving the check for the `None` into `store.rs` (I left a suggestion there 😊), and changing this method to take a non-optional `DownloadedYelpSuggestion`.

components/suggest/src/yelp.rs
```diff
@@ -0,0 +1,368 @@
...
+const MAX_QUERY_LENGTH: usize = 150;
+
+/// The max number of words consisting the modifier. To improve the SQL performance by matching with
+/// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
+const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
+
+impl<'a> SuggestDao<'a> {
+    /// Inserts the suggestions for Yelp attachment into the database.
+    pub fn insert_yelp_suggestions(
+        &self,
```
linabutler, line 52:
```suggestion
        &mut self,
```

This method writes to the database, so let's make it `mut` to match our other write methods.

linabutler (dismissed):
Still looks great to me—thank you very much for making the changes! 🚀

dadaa:
Hi @linabutler, @ncloudioj, @0c0w3, thank you very much for your review!
I have some questions. 
I'm not sure though, `mergify` bot removes r+ when the patch is updated. Is this expected?
And, always `swiftlint` failed. Is there anything I should do?

linabutler:
Hi @dadaa! I just added you to @mozilla/application-services-collaborators, so you should be able to merge PRs now, and Mergify should no longer dismiss stale reviews when you push. I think it does that for "outside contributors", but not if you're a collaborator.

I think the `swiftlint` failures should be fixed if you rebase and pick up #6045!

linabutler (approved):
Once more, with feeling!

dadaa:
> Hi @dadaa! I just added you to @mozilla/application-services-collaborators, so you should be able to merge PRs now, and Mergify should no longer dismiss stale reviews when you push. I think it does that for "outside contributors", but not if you're a collaborator.

I see, thanks!

> I think the `swiftlint` failures should be fixed if you rebase and pick up #6045!

Okay, I will try it.

dadaa:
@linabutler The bot still removed r+..

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
0c0w3: I'm finding it hard to form an opinion either way -- I just don't think it's a big deal. There's a small perf cost to unnecessarily lowercasing the same strings more than once, but I would think it's miniscule overall, especially considering we force a limit on the length of the search string. Likewise there's a small amount of code inflation due to the multiple `to_lowercase()` calls. There's also a higher likelihood of introducing bugs due to case mismatches if we don't normalize early on in one place, like at the beginning of `fetch_suggestions()`. All these drawbacks seem minor to me.

If you want to preserve the case, I'm fine with that, but I'm not sure @ncloudioj is?

0c0w3 (approved):

dadaa:
Thank you so much, @0c0w3!

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
ncloudioj: It was just a nitpick for readability, not a blocker. Feel free to leave it as is. :)

I was wondering if those query normalization had special treatments in different places When I looked at the code. Looks like they all just do trimming and casing. So I suggested normalizing the original query once at the top to slightly reduce the cognitive load when reading those code paths in future.

ncloudioj (approved):

components/suggest/src/yelp.rs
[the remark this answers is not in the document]
0c0w3: (Sorry for this long comment!)

One thing I do have a stronger opinion on, that I just thought of: The Yelp provider shouldn't bypass any query normalization that's done for every other provider.

Right now, `fetch_suggestions()` isn't really provider agnostic. i.e., it doesn't simply loop through a list of items that `impl` some generic `Provider` trait and call `fetch()` on each of them. So at the moment, we really don't have a clear implementation strategy where (1) normalization is done on the query and then (2) the normalized query is handed off to each provider. (Although https://github.com/mozilla/application-services/commit/f3cd74035a5eafcdd5925aadc2859ec59583ecdb did kind of start us down that path, which is good IMO.)

However, we've talked about rewriting `fetch_suggestions()` in the future so that it does work like that. And in that world, Yelp shouldn't be treated specially. It should get the same query as all the other providers, whether it's normalized or not.

I don't think we need to worry about this in this PR right now, but it's something to keep in mind if/when we rewrite `fetch_suggestions()`.

dadaa:
Thank you very much, @ncloudioj!
I will try to merge this patch.

PR #5961, opened 2023-11-30
Set a minimum Python version of 3.8 [ci full]

rvandermeulen:
The minimum Python version was recently bumped to version 3.8 in mozilla-central - see [bug 1843209](https://bugzilla.mozilla.org/show_bug.cgi?id=1843209). I'm not sure if we have an official way of enforcing that in this repo, but we can at least bump some dependencies which have been blocked on making 3.8 the minimum.


bendk:
Made https://bugzilla.mozilla.org/show_bug.cgi?id=1867813 for upgrading the mac worker.

rvandermeulen:
I feel like we should document this somewhere too? Even if to say that A-S stays in sync with m-c. @tarikeshaq Any suggestions as to where that should go?

tarikeshaq:
We have https://github.com/mozilla/application-services/blob/main/docs/design/rust-versions.md for our Rust versions (which ends up in https://mozilla.github.io/application-services/book/design/rust-versions.html)

We can make that document more general and include Python there or place another document next to it

rvandermeulen:
Failing dependency check job will be handled by #6230 or something along those lines. The failing fetch-nss-artifact job will be handled by #6229 which I hope to land immediately after this one.

PR #5987, opened 2023-12-13
Updating the Kotlin getAccessToken docstring

bendk:

  - Note:

### DIFF
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
@@ -250,12 +250,17 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
      * This performs network requests, and should not be used on the main thread.
      * It may modify the persisted account state.
      *
+     * On `FxaException.Unauthorized` and `FxaException.SyncScopedKeyMissingInServerResponse`, the
+     * caller should indicate to the user that there are authentication issues and allow them to
+     * re-login by starting a new OAuth flow.
+     *
      * @param scope Single OAuth scope (no spaces) for which the client wants access
      * @param ttl time in seconds for which the token will be valid
      * @return [AccessTokenInfo] that stores the token, along with its scopes and keys when complete
-     * @throws FxaException.Unauthorized We couldn't provide an access token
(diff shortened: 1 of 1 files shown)


components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
```diff
@@ -250,12 +250,15 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
      * This performs network requests, and should not be used on the main thread.
      * It may modify the persisted account state.
      *
+     * On exceptions, the caller should indicate to the user that there are authentication issues
```
mhammond, line 253: Is this strictly true? Aren't there a class of errors which do not imply this, such as network errors? Also, below you call out `SyncScopedKeyMissingInServerResponse`, but it's not clear to me that exception also implies a new flow is required (ie, aren't we still a little confused by what that *does* mean?)

(This also raises interesting meta-questions about docstrings in general - in a perfect world, this docstring would actually existing in our UDL, right? But would that end up exposed to our consumers in a sensible way? This has nothing at all to do with this patch though, just pondering!)

bendk: Good point, I updated the docs to reflect that.

I do believe that consumers should transition to the auth-problems state when they see `SyncScopedKeyMissingInServerResponse`.  I believe the only way to recover from that state is to re-login.  I'm still a bit confused on why users get in that state, but I'm pretty sure that it's a server issue since we're seeing 
[Sentry reports for fxaclient-scoped-key](https://mozilla.sentry.io/issues/3726733595/?project=6295546&query=is%3Aunresolved+fxaclient-scoped-key&referrer=issue-stream&statsPeriod=30d&stream_index=0) which comes from [here](https://github.com/mozilla/application-services/blob/5e9c3f11f09328a2436df32110b0340d02691869/components/fxa-client/src/internal/oauth.rs#L332-L339).

mhammond: (Re the meta-question, I guess I was really referring to the fact we have hand-written wrappers - would consumers of these wrappers ever actually see docstrings emitted via uniffi? But yeah, that 2024's problem :)

PR #5995, opened 2023-12-14
Uniffi webext-storage component

lougeniaC64:
Uniffi webext-storage component for mobile.

Note: Added `Do not merge` label to this PR until next release is cut. Will remove and a changelog entry thereafter.

  - Note:

### DIFF
--- a/components/webext-storage/build.rs
+++ b/components/webext-storage/build.rs
@@ -10,4 +10,6 @@ fn main() {
 
     // If NSS_DIR isn't set, we don't really care, ignore the Err case.
     let _ = nss_build_common::link_nss();
+
+    uniffi::generate_scaffolding("./src/webext-storage.udl").unwrap();
 }
--- a/components/webext-storage/ffi/src/lib.rs
+++ b/components/webext-storage/ffi/src/lib.rs
@@ -5,7 +5,7 @@
 use std::os::raw::c_char;
 
 use ffi_support::{define_handle_map_deleter, ConcurrentHandleMap, ExternError, FfiStr};
-use webext_storage::{error, store::Store};
+use webext_storage::{error, store::WebExtStorageStore as Store};
 
 lazy_static::lazy_static! {
     static ref STORES: ConcurrentHandleMap<Store> = ConcurrentHandleMap::new();
--- a/components/webext-storage/src/error.rs
+++ b/components/webext-storage/src/error.rs
@@ -2,17 +2,46 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use error_support::{ErrorHandling, GetErrorHandling};
 use interrupt_support::Interrupted;
+use std::fmt;
 
-#[derive(Debug)]
+/// Result enum used by all implementation functions in this crate.
+/// These wll be automagically turned into `WebExtStorageApiError` at the
+/// FFI layer.
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Clone, Copy)]
 pub enum QuotaReason {
     TotalBytes,
     ItemBytes,
     MaxItems,
 }
 
+impl fmt::Display for QuotaReason {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            QuotaReason::ItemBytes => write!(f, "ItemBytes"),
+            QuotaReason::MaxItems => write!(f, "MaxItems"),
+            QuotaReason::TotalBytes => write!(f, "TotalBytes"),
+        }
+    }
+}
+
 #[derive(Debug, thiserror::Error)]
-pub enum ErrorKind {
+pub enum WebExtStorageApiError {
+    #[error("Unexpected webext-storage error: {reason}")]
+    UnexpectedError { reason: String },
+
+    #[error("Error parsing JSON data: {reason}")]
+    JsonError { reason: String },
+
+    #[error("Quota exceeded: {reason}")]
+    QuotaError { reason: QuotaReason },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
     #[error("Quota exceeded: {0:?}")]
     QuotaError(QuotaReason),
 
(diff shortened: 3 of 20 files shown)


lougeniaC64:
@mhammond FYI I took out the updates to the error reporting because I would need to update the desktop FFI functions to use the `Result` with the new external error type (and I'd like to leave that code untouched and stable so it remains a control in this integration) or do something else dumb and ultimately temporary. I'll add the new reporting logic after the hand-written FFI code is removed.

components/webext-storage/src/store.rs
```diff
@@ -58,10 +58,10 @@ impl Store {
 
     /// Sets one or more JSON key-value pairs for an extension ID. Returns a
     /// list of changes, with existing and new values for each key in `val`.
-    pub fn set(&self, ext_id: &str, val: JsonValue) -> Result<StorageChanges> {
```
mhammond, line 61: This is looking great! Sorry I didn't get very far on this, but if you change the UDL to, eg, `StorageChanges set([ByRef] string ext_id, JsonValue val);` these can all remain `&str`. I'll look in more detail next week!

mhammond:
This is looking great! I think we should do a little more with the errors sooner rather than later.

Looking at the [ffi error code](https://github.com/mozilla/application-services/blob/main/components/webext-storage/src/ffi.rs#L14-L47), it seems like we already have a distinction between "internal" and "external" errors - "internal" errors are all the types in that enum, whereas there are exactly 5 "external" errors (those values in the line above). So we should be able to add a new enum with 5 entries as the "public" error - so we then end up with the same internal->external mapping done by our other crates, and that FFI code can work purely in terms of the new external enum. Later, of course, we still remove that ffi code.

Ping me if some of that doesn't make much sense!

lougeniaC64:
> This is looking great! I think we should do a little more with the errors sooner rather than later.
> 
> Looking at the [ffi error code](https://github.com/mozilla/application-services/blob/main/components/webext-storage/src/ffi.rs#L14-L47), it seems like we already have a distinction between "internal" and "external" errors - "internal" errors are all the types in that enum, whereas there are exactly 5 "external" errors (those values in the line above). So we should be able to add a new enum with 5 entries as the "public" error - so we then end up with the same internal->external mapping done by our other crates, and that FFI code can work purely in terms of the new external enum. Later, of course, we still remove that ffi code.
> 
> Ping me if some of that doesn't make much sense!

Just put up the changes I had with some updates. Let me know if this is what you had in mind.

components/webext-storage/build.rs
```diff
@@ -10,4 +10,6 @@ fn main() {
 
     // If NSS_DIR isn't set, we don't really care, ignore the Err case.
     let _ = nss_build_common::link_nss();
+
+    uniffi::generate_scaffolding("./src/webextstorage.udl").unwrap();
```
mhammond, line 14: nano-nit - shall we use a hyphen in this name for consistency with the crate?

components/webext-storage/ffi/src/lib.rs
```diff
@@ -14,7 +14,7 @@ lazy_static::lazy_static! {
 #[no_mangle]
 pub extern "C" fn webext_store_new(db_path: FfiStr<'_>, error: &mut ExternError) -> u64 {
     log::debug!("webext_store_new");
-    STORES.insert_with_result(error, || -> error::Result<Store> {
+    STORES.insert_with_result(error, || -> error::ApiResult<Store> {
```
mhammond, line 17: I think we can probably stick with using the "internal" errors here and below.

components/webext-storage/src/error.rs
```diff
@@ -2,17 +2,46 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum WebExtStorageApiError {
+    #[error("Unexpected webext-storage error: {reason}")]
+    UnexpectedError { reason: String },
+
+    #[error("Error parsing JSON data: {reason}")]
+    JsonError { reason: String },
+
+    #[error("Quota exceeded: {reason}")]
+    QuotaError { reason: String },
```
mhammond, line 40: This can probably stick with the enum rather than a string. It will mean declaring the enum in the UDL, but that seems fine.

components/webext-storage/src/error.rs
```diff
@@ -60,13 +89,154 @@ pub enum ErrorKind {
...
+        match self {
+            Error::QuotaError(r) => ErrorHandling::convert(WebExtStorageApiError::QuotaError {
+                reason: r.to_string(),
+            })
+            .report_error("webext-storage-quota-error"),
+            Error::JsonError(e) => ErrorHandling::convert(WebExtStorageApiError::JsonError {
+                reason: e.to_string(),
+            })
+            .report_error("webext-storage-json-error"),
+            Error::OpenDatabaseError(e) => {
```
mhammond, line 105: All of these cases can probably be combined into one.

components/webext-storage/src/error.rs
```diff
@@ -60,13 +89,154 @@ pub enum ErrorKind {
...
+            Error::QuotaError(QuotaReason::TotalBytes) => WebExtStorageApiError::QuotaError {
+                reason: QuotaReason::TotalBytes.to_string(),
+            },
+            Error::QuotaError(QuotaReason::ItemBytes) => WebExtStorageApiError::QuotaError {
+                reason: QuotaReason::ItemBytes.to_string(),
+            },
+            Error::QuotaError(QuotaReason::MaxItems) => WebExtStorageApiError::QuotaError {
+                reason: QuotaReason::MaxItems.to_string(),
+            },
+            Error::OpenDatabaseError(e) => WebExtStorageApiError::UnexpectedError {
```
mhammond, line 188: ditto here - we can probably collapse all these.

components/webext-storage/src/ffi.rs
```diff
@@ -32,15 +32,16 @@ mod error_codes {
...
-            ErrorKind::QuotaError(QuotaReason::TotalBytes) => {
-                error_codes::QUOTA_TOTAL_BYTES_EXCEEDED
-            }
-            ErrorKind::QuotaError(QuotaReason::ItemBytes) => error_codes::QUOTA_ITEM_BYTES_EXCEEDED,
-            ErrorKind::QuotaError(QuotaReason::MaxItems) => error_codes::QUOTA_MAX_ITEMS_EXCEEDED,
+impl From<WebExtStorageApiError> for ExternError {
+    fn from(err: WebExtStorageApiError) -> ExternError {
+        let code = ErrorCode::new(match &err {
+            WebExtStorageApiError::JsonError { reason: _ } => error_codes::INVALID_JSON,
+            WebExtStorageApiError::QuotaError { reason } => match reason.as_str() {
```
mhammond, line 39: (sticking to the enum in the error makes this a lot cleaner)

components/webext-storage/src/store.rs
```diff
@@ -28,14 +28,14 @@ use serde_json::Value as JsonValue;
...
+pub struct WebExtStorageStore {
     db: Arc<ThreadSafeStorageDb>,
 }
 
-impl Store {
+impl WebExtStorageStore {
     /// Creates a store backed by a database at `db_path`. The path can be a
     /// file path or `file:` URI.
-    pub fn new(db_path: impl AsRef<Path>) -> Result<Self> {
+    pub fn new(db_path: impl AsRef<Path>) -> ApiResult<Self> {
```
mhammond, line 38: Like above, I think these can probably stat with the internal error.

megazords/full/src/lib.rs
```diff
@@ -25,6 +25,7 @@ pub use suggest;
 pub use sync_manager;
 pub use tabs;
 pub use viaduct;
+pub use webext_storage;
```
mhammond, line 28: We probably need to work out a way to arrange for the local kotlin tests to work without adding it to the megazord at this point as it will end up shipped on Android even while unused. I'll think about this but it should be doable. It's fine to keep it for now, just not for actually landing it.

mhammond (changes requested):
This is looking great!

lougeniaC64:
@mhammond If these changes aren't quite right, I'll update this PR when I return. If things look okay, feel free to approve and merge.

mhammond (approved):
Lovely! 🚢

mhammond:
oops, had to back this out because m-c didn't build and suggest meeded a new vendor. I think the m-c changes will be quite easy, but we probably need to get them landed before re-landing this.

PR #6011, opened 2023-12-20
Bug 1866370: Recovers logins when modified remotely if deleted locally

tarikeshaq:
This PR probably requires a bit of careful review, especially the semantics of how login syncs with regard to local tombstones

I believe this PR would fix both https://bugzilla.mozilla.org/show_bug.cgi?id=1866370 and https://bugzilla.mozilla.org/show_bug.cgi?id=1870876 and help bring up our sync success rate up from 95% to closer to the other engines ~99%

The issue was **very** subtle, and is triggered in the following case:
1. Locally, we deleted a login, modifying it locally to be a tombstone
2. Another client modified they login before we synced it
3. When it's time for our next sync, we detect a conflict and initiate a three-way merge
   - In the three-way merge, we start by calculating the delta between our local record (the tombstone!) and the mirror record
   - To even start calculating the delta, we attempt to decrypt the secure data... However, that data has already been wiped in the delete in step 1 so we fail the three-way merge, and subsequently the sync.

## Important Note
 I made an assumption here that in this case, the incoming record should win and the login gets re-established. Another possible approaches:
- Make sure server deletes the login
- Based on which change is "younger" we can either delete, or keep the login

Neither approach is perfect and I'd be happy to do it the other way around

  - Note:

### DIFF
--- a/components/logins/src/sync/engine.rs
+++ b/components/logins/src/sync/engine.rs
@@ -19,6 +19,7 @@ use sql_support::ConnExt;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::sync::Arc;
+use std::time::{Duration, UNIX_EPOCH};
 use sync15::bso::{IncomingBso, OutgoingBso, OutgoingEnvelope};
 use sync15::engine::{CollSyncIds, CollectionRequest, EngineSyncAssociation, SyncEngine};
 use sync15::{telemetry, ServerTimestamp};
@@ -91,8 +92,8 @@ impl LoginsSyncEngine {
                     telem.applied(1);
                 }
                 (None, Some(local)) => {
-                    log::debug!("  Conflicting record without shared parent, using newer");
-                    plan.plan_two_way_merge(&local.login, (upstream, upstream_time));
+                    log::debug!("  Conflicting record without shared parent,  Resolving with 2WM");
+                    plan.plan_two_way_merge(local, (upstream, upstream_time));
                     telem.reconciled(1);
                 }
                 (None, None) => {
@@ -102,7 +103,13 @@ impl LoginsSyncEngine {
                             upstream.guid(),
                             dupe.guid()
                         );
-                        plan.plan_two_way_merge(&dupe, (upstream, upstream_time));
+                        let local_modified = UNIX_EPOCH
+                            + Duration::from_millis(dupe.record.time_password_changed as u64);
+                        let local = LocalLogin::Alive {
+                            login: dupe,
+                            local_modified,
+                        };
+                        plan.plan_two_way_merge(local, (upstream, upstream_time));
                     } else {
                         log::debug!("  No dupe found, inserting into mirror");
                         plan.plan_mirror_insert(upstream, upstream_time, false);
@@ -509,10 +516,9 @@ mod tests {
 
     // Wrap sync functions for easier testing
     fn run_fetch_login_data(
-        store: LoginStore,
+        engine: &mut LoginsSyncEngine,
         records: Vec<IncomingBso>,
     ) -> (Vec<SyncLoginData>, telemetry::EngineIncoming) {
-        let mut engine = LoginsSyncEngine::new(Arc::new(store)).unwrap();
         engine
             .set_local_encryption_key(&TEST_ENCRYPTION_KEY)
             .unwrap();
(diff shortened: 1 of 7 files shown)


mhammond:
Thanks Tarik,
  As we discussed, I kinda like the idea of more explicit tombstone handling via an enum. I'm fairly sure this uncovered a few problems with "2 way" merge. This is exceedingly unlikely in the wild, but still possible. There's still an issue with an unnecessary outgoing record in this scenario, but I'm not too worried about it as it only happens on resolution of a conflict, which should be even rarer.

Do you mind doing a review of the 2nd commit?

components/logins/src/sync/engine.rs
```diff
@@ -950,4 +963,111 @@ mod tests {
             "else"
         );
     }
+
+    fn count(engine: &LoginsSyncEngine, sql: &str) -> u32 {
```
tarikeshaq, line 967: nit: What do you think of having this function take the table name to count, instead of the whole SQL statement?

components/logins/src/sync/engine.rs
```diff
@@ -102,7 +103,13 @@ impl LoginsSyncEngine {
                             upstream.guid(),
                             dupe.guid()
                         );
-                        plan.plan_two_way_merge(&dupe, (upstream, upstream_time));
+                        let local_modified = UNIX_EPOCH
+                            + Duration::from_millis(dupe.record.time_password_changed as u64);
```
tarikeshaq, line 107: Out of curiosity, why don't we use the `local_modified` column instead here? I get that the struct `EncryptedLogin` doesn't have it as a field (but we could easily pull it from the db if we needed to so I'm curious if there's another reason)

components/logins/src/sync/merge.rs
```diff
@@ -33,29 +33,56 @@ impl MirrorLogin {
...
+            }
         })
     }
 
-    pub fn is_tombstone(&self) -> bool {
-        // For all sync purposes, if the secure fields are not present, this is a tombstone
-        self.login.sec_fields.is_empty()
+    // Only used by tests where we want to get the "raw" record - ie, a tombstone will still
+    // be returned here, just with many otherwise invalid empty fields
+    #[allow(unused)]
```
tarikeshaq, line 83: nit: since this is test only and only used by unit tests, can we #[cfg(test)] it?

components/logins/src/sync/engine.rs
```diff
@@ -950,4 +963,111 @@ mod tests {
...
+
+    #[test]
+    fn test_incoming_non_mirror_tombstone_local_newer() {
+        do_test_incoming_with_local_unmirrored_tombstone(true);
+    }
+
+    #[test]
+    fn test_incoming_non_mirror_tombstone_local_older() {
+        do_test_incoming_with_local_unmirrored_tombstone(false);
+    }
```
tarikeshaq, line 1072: Tests looks great!!!

components/logins/src/sync/merge.rs
```diff
@@ -33,29 +33,56 @@ impl MirrorLogin {
...
-            local_modified: util::system_time_millis_from_row(row, "local_modified")?,
+        let local_modified = util::system_time_millis_from_row(row, "local_modified")?;
+        Ok(if row.get("is_deleted")? {
+            let id = row.get("guid")?;
+            LocalLogin::Tombstone { id, local_modified }
+        } else {
+            let login = EncryptedLogin::from_row(row)?;
+            if login.sec_fields.is_empty() {
+                error_support::report_error!("logins-crypto", "empty ciphertext in the db",);
+            }
```
tarikeshaq, line 73: nice! This will most likely fail the sync if we hit it, Is it worth us returning an error as well report it?

tarikeshaq (commented):
This looks good to me!! (GitHub wouldn't let me approve since I'm the author of the PR)

components/logins/src/sync/merge.rs
mhammond: Good question! I was trying to remain conservative and not introduce new error conditions, but OTOH, the user is still going to have a bad time eventually - at decrypt time. So I'm leaning towards just doing this and let whatever existing failures might have happened continue to happen and re-evaluate once we start seeing sentry reports, but could easily be convinced to do something different instead.

components/logins/src/sync/engine.rs
mhammond: I agree we probably should, but I didn't want to change the semantics in this patch, mainly because I was slightly worried the semantics of that column might not be what we want (ie, I didn't want to dig into the semantics of that field and how sync itself updates it, for example)

mhammond (approved):

PR #6027, opened 2023-12-29
Update Rust version to 1.75, MSR to 1.70

mhammond:
m-c is now on [1.75 for workers](https://bugzilla.mozilla.org/show_bug.cgi?id=1857090), [1.70 for MSR](https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#14). As per [our policy](https://github.com/mozilla/application-services/blob/main/docs/design/rust-versions.md), we should update.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ commands:
   setup-rust-min-version:
     steps:
       # https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#26
-      - run: rustup override set 1.65.0
+      - run: rustup override set 1.70.0
 
   full-checkout:
     steps:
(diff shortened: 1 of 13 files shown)


bendk (approved):
Looks great!  :rocket:

bendk:
I just updated this branch purely to test if the CI would pass now.  I believe we had some updates to the mac workers, so maybe it will just work?

bendk:
Looks like that worked.  What do you think about rebase and merging?

mhammond:
I'm fine with that - I was waiting to see if @badboy had any concerns or wanted us to hold off, but now he is back from PTO I think we can take silence as consent :)

badboy:
oh sorry, no blocker from my side!

PR #6051, opened 2024-01-12
Support swiftlint and swiftformat on non-macOS via docker

brennie:
Fixes [Bug 1874452](https://bugzilla.mozilla.org/show_bug.cgi?id=1874452)

  - Note:

### DIFF
--- a/automation/tests.py
+++ b/automation/tests.py
@@ -490,5 +490,14 @@ def exit_with_error(code, text, print_exception=False):
         traceback.print_exc()
     sys.exit(code)
 
+def docker_installed():
+    result = subprocess.run(
+        ["docker"],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+
+    return result.returncode == 0
+
 if __name__ == '__main__':
     main()
--- a/automation/tests.py
+++ b/automation/tests.py
@@ -358,14 +358,39 @@ def cargo_fmt(package=None, fix_issues=False):
     run_command(cmdline)
 
 def swift_format():
+    swift_format_args = [
+        'megazords',
+        'components/*/ios',
+        '--exclude',
+        '**/Generated',
+        '--exclude',
+        'components/nimbus/ios/Nimbus/Utils',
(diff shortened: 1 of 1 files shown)


automation/tests.py
```diff
@@ -311,8 +311,27 @@ def run_ktlint():
...
+            [
+                "docker",
+                "run",
+                "-it",
+                "--rm",
+                "-v",
+                f"{cwd}:{cwd}",
+                "-w",
+                cwd,
+                "ghcr.io/realm/swiftlint:latest",
```
tarikeshaq, line 330: Hmm do we know if this will always match the version of swiftlint running in CI (CI's version is not fixed, although I'm slightly in the camp of fixing the version - if we do that this will be fixed to that version which will be ok!)

automation/tests.py
```diff
@@ -339,14 +358,39 @@ def cargo_fmt(package=None, fix_issues=False):
     run_command(cmdline)
 
 def swift_format():
+    swift_format_args = [
```
tarikeshaq, line 361: thank you for the refactor!!

automation/tests.py
```diff
@@ -339,14 +358,39 @@ def cargo_fmt(package=None, fix_issues=False):
...
+            [
+                "docker",
+                "run",
+                "-it",
+                "--rm",
+                "-v",
+                f"{cwd}:{cwd}",
+                "-w",
+                cwd,
+                "ghcr.io/nicklockwood/swiftformat:latest",
```
tarikeshaq, line 390: ditto on the versioning, but it's still better than the status quo so I'm not going to block on figuring the versioning out! Will see if we can fix all the version at once as a followup

tarikeshaq (approved):
Thank you! 🚀

automation/tests.py
brennie: The version in CI is from `brew install swiftlint`, which should be the latest version. I also am pro fixing the verisons, esp because they seem to add new default lints in versions which causes PRs to fail for unrelated reasons :)

PR #6055, opened 2024-01-16
[DISCO-2718] Suggest: fix: Add score to AMP suggestions

tiftran:
So I think that firefox [defaults to a score 0.2](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarProviderQuickSuggest.sys.mjs#33) when there's no score and @ncloudioj mentioned that firefox sorts by score (I think that's done [here?](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarProviderQuickSuggest.sys.mjs#193), which would explain the amo/amp keyword collision shows an amo suggestion since amo has a score of 0.25?

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -33,6 +33,9 @@ pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
 /// that aren't parsable and which schema version it was first seen in.
 pub const UNPARSABLE_RECORDS_META_KEY: &str = "unparsable_records";
 
+// Default value when Suggestion does not have a value for score
+const DEFAULT_SUGGESTION_SCORE: f64 = 0.2;
+
 /// The database connection type.
 #[derive(Clone, Copy)]
 pub(crate) enum ConnectionType {
@@ -140,40 +143,40 @@ impl<'a> SuggestDao<'a> {
         {
             (self.conn.prepare_cached(
                 &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
+                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
                      FROM suggestions s
                      JOIN keywords k ON k.suggestion_id = s.id
                      WHERE s.provider IN ({}) AND
                            k.keyword = :keyword
                      UNION ALL
-                     SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
+                     SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, k.confidence, k.keyword_suffix
                      FROM suggestions s
                      JOIN prefix_keywords k ON k.suggestion_id = s.id
                      WHERE k.keyword_prefix = :keyword_prefix
-                     ORDER BY s.provider
+                     ORDER BY s.score DESC
                      LIMIT :suggestions_limit",
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
                 (":keyword", keyword_lowercased as &dyn ToSql),
                 (":keyword_prefix", &keyword_prefix as &dyn ToSql),
-                (":suggestions_limit", &suggestions_limit as &dyn ToSql),
+                (":suggestions_limit", &suggestions_limit as &dyn ToSql)
             ])
         } else {
             (self.conn.prepare_cached(
                 &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
+                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
                      FROM suggestions s
                      JOIN keywords k ON k.suggestion_id = s.id
                      WHERE s.provider IN ({}) AND
                            k.keyword = :keyword
-                     ORDER BY s.provider
+                     ORDER BY s.score DESC
                      LIMIT :suggestions_limit",
                     providers_to_sql_list(&query.providers),
                 ),
             )?, vec![
                 (":keyword", keyword_lowercased as &dyn ToSql),
-                (":suggestions_limit", &suggestions_limit as &dyn ToSql),
+                (":suggestions_limit", &suggestions_limit as &dyn ToSql)
             ])
         };
 
@@ -182,6 +185,7 @@ impl<'a> SuggestDao<'a> {
                 let title = row.get("title")?;
                 let raw_url = row.get::<_, String>("url")?;
                 let provider = row.get("provider")?;
+                let score = row.get::<_, f64>("score")?;
 
                 let keywords: Vec<String> = self.conn.query_rows_and_then_cached(
                     "SELECT keyword FROM keywords
@@ -220,6 +224,7 @@ impl<'a> SuggestDao<'a> {
                                     impression_url: row.get("impression_url")?,
                                     click_url: cooked_click_url,
                                     raw_click_url,
+                                    score,
                                 }))
                             }
                         )
@@ -245,7 +250,7 @@ impl<'a> SuggestDao<'a> {
                     SuggestionProvider::Amo => {
                         let full_suffix = row.get::<_, String>("keyword_suffix")?;
                         self.conn.query_row_and_then(
-                            "SELECT amo.description, amo.guid, amo.rating, amo.icon_url, amo.number_of_ratings, amo.score
+                            "SELECT amo.description, amo.guid, amo.rating, amo.icon_url, amo.number_of_ratings
                              FROM amo_custom_details amo
                              WHERE amo.suggestion_id = :suggestion_id",
                             named_params! {
@@ -261,7 +266,7 @@ impl<'a> SuggestDao<'a> {
                                         rating: row.get("rating")?,
                                         number_of_ratings: row.get("number_of_ratings")?,
                                         guid: row.get("guid")?,
-                                        score: row.get("score")?,
+                                        score,
                                     }))
                                 } else {
                                     Ok(None)
@@ -276,26 +281,16 @@ impl<'a> SuggestDao<'a> {
                             KeywordConfidence::High => full_suffix == keyword_suffix,
                         };
                         if suffixes_match {
-                            self.conn.query_row_and_then(
-                                "SELECT p.score
-                                FROM pocket_custom_details p
-                                WHERE p.suggestion_id = :suggestion_id",
-                                named_params! {
-                                    ":suggestion_id": suggestion_id
-                                },
-                                |row| {
+
                                     Ok(Some(Suggestion::Pocket {
                                         title,
                                         url: raw_url,
-                                        score: row.get("score")?,
+                                        score,
                                         is_top_pick: matches!(
                                             confidence,
                                             KeywordConfidence::High
-                                        )
-                                    }))
-                                }
-                            )
-                        } else {
+                                        )}))
+                                    } else {
                             Ok(None)
                         }
                     },
@@ -322,13 +317,15 @@ impl<'a> SuggestDao<'a> {
                          record_id,
                          provider,
                          title,
-                         url
+                         url,
+                         score
                      )
                      VALUES(
                          :record_id,
                          {},
                          :title,
-                         :url
+                         :url,
+                         :score
                      )
                      RETURNING id",
                     SuggestionProvider::Amo as u8
@@ -337,6 +334,7 @@ impl<'a> SuggestDao<'a> {
                     ":record_id": record_id.as_str(),
                     ":title": suggestion.title,
                     ":url": suggestion.url,
+                    ":score": suggestion.score,
                 },
                 |row| row.get(0),
                 true,
@@ -348,8 +346,7 @@ impl<'a> SuggestDao<'a> {
                              guid,
                              icon_url,
                              rating,
-                             number_of_ratings,
-                             score
+                             number_of_ratings
                          )
                          VALUES(
                              :suggestion_id,
@@ -357,8 +354,7 @@ impl<'a> SuggestDao<'a> {
                              :guid,
                              :icon_url,
                              :rating,
-                             :number_of_ratings,
-                             :score
+                             :number_of_ratings
                          )",
                 named_params! {
                     ":suggestion_id": suggestion_id,
@@ -366,8 +362,7 @@ impl<'a> SuggestDao<'a> {
                     ":guid": suggestion.guid,
                     ":icon_url": suggestion.icon_url,
                     ":rating": suggestion.rating,
-                    ":number_of_ratings": suggestion.number_of_ratings,
-                    ":score": suggestion.score,
+                    ":number_of_ratings": suggestion.number_of_ratings
                 },
             )?;
             for (index, keyword) in suggestion.keywords.iter().enumerate() {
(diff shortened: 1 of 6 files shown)


components/suggest/src/rs.rs
```diff
@@ -168,8 +168,24 @@ pub(crate) struct DownloadedAmpSuggestion {
     pub impression_url: String,
     #[serde(rename = "icon")]
     pub icon_id: String,
+    pub score: f64,
```
ncloudioj, line 171: According to the [schema](https://github.com/mozilla-services/quicksuggest-rs/blob/main/schema/quicksuggest.1.schema.json#L70-L75), `score` is optional meaning that it's possible to see `null` in the RS attachments. So we should make it `Option<f64>`. This also requires us to handle the `null` case for `PartialEq` and we can't implement `Eq` for it for total ordering.

components/suggest/src/schema.rs
```diff
@@ -48,6 +48,7 @@ pub const SQL: &str = "
         impression_url TEXT NOT NULL,
         click_url TEXT NOT NULL,
         icon_id TEXT NOT NULL,
+        score REAL NOT NULL,
```
ncloudioj, line 51: Should be nullable.

ncloudioj (dismissed):

components/suggest/src/rs.rs
linabutler: Feel free to remove the `PartialEq` and `Eq` derives from `DownloadedAmpSuggestion`, `DownloadedAmpWikipediaSuggestion`, and the other downloaded suggestion types, too!

They're here "just in case" (and will be optimized out as dead code), but we don't expose these types outside our crate, and we don't use them in our code or in unit tests, so we can just drop them 😊

components/suggest/src/suggest.udl
```diff
@@ -38,7 +38,8 @@ interface Suggestion {
         string iab_category,
         string impression_url,
         string click_url,
-        string raw_click_url
+        string raw_click_url,
+        f64 score
```
linabutler, line 42: This will be a backward-incompatible change for iOS and Android—we'll need to add the field [here](https://github.com/mozilla-mobile/firefox-ios/blob/889120b984079af73e075b47dae918f4cba924da/firefox-ios/Storage/Rust/RustFirefoxSuggestion.swift#L42) for iOS, and the [tests](https://github.com/mozilla-mobile/firefox-android/blob/51f88bd6180d80fee4cc69d56607269455cf831c/android-components/components/feature/fxsuggest/src/test/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProviderTest.kt#L82-L101) on Android.

There's a bit of a [manual process](https://mozilla.github.io/application-services/book/howtos/breaking-changes.html#merging) for handling those:

1. Put up PRs for iOS and Android.
2. Get the iOS and Android PRs approved, but not landed yet.
3. Land this Application Services PR.
4. Wait for the automated PRs on Android and iOS that bump their Application Services versions to fail.
5. Merge the PRs from (1) into the version bump branches from (4).

tiftran: oh lol, sounds like im going on cross platform adventure 😂 , will do!

ncloudioj (dismissed):
LGTM! r=nanj

components/suggest/src/suggest.udl
ncloudioj: TIL! Gotta say that process looks quite involved to a newbie like me. Out of curiosity, what happens if we land it w/o fixing breakages in downstream? Will we get backouts like what Desktop does :D?

linabutler: Yeah, it's pretty manual!

For Android and iOS, there are bots that automatically bump the Application Services dependencies. Here's an example for iOS: mozilla-mobile/firefox-ios#18117 and Android: mozilla-mobile/firefox-android#5130. On Android, if the tests pass, the PR is merged automatically; on iOS, someone from that team needs to manually trigger CI, approve, and merge.

On Android, if the PR fails on CI, the Release Owner usually pings `#application-services-eng` on Slack, at which point the author of the breaking change merges their PR _into the branch for the version bump PR_. Here's an example of that: mozilla-mobile/firefox-android#4088. That PR folded in my fix for a breaking change in Suggest, and a fix from James for a breaking change in Nimbus. Once that combined branch was green, it was merged using "squash and merge".

On iOS, if CI fails, `@firefox-ios-alerts-bitrise` will notify `#firefox-ios`—and someone from the team might tag you or someone from the Sync team. It's the same deal there—the author of the breaking change merges their PR into the branch with the version bump, and then someone from the iOS team re-triggers CI, and squash-merges once it's all green. Here's an example: mozilla-mobile/firefox-ios#16840

There are no backouts—Android and iOS will stay on the old Application Services versions until the breakage is resolved 😅 We try to avoid backward-incompatible changes (but sometimes they can't be helped, like here), and resolve them pretty quickly. Breakages don't prevent iOS and Android from building, but they do prevent them from getting more A-S version bumps.

(On Desktop, there's no automated vendoring at all 😢)

ncloudioj: Hey Lina, thank you so much for walking us through with examples! I have a better grasp on how things interact behind the scenes!

While we might be able to find those details in the [A-S doc](zilla.github.io/application-services/book/index.html), I just feel it's much easier to get up to speed with guided walk-through like this. This should be part of the "Hack A-S: Episode II" :)

components/suggest/src/db.rs
```diff
@@ -140,17 +140,17 @@ impl<'a> SuggestDao<'a> {
         {
             (self.conn.prepare_cached(
                 &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
+                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
```
tiftran, line 143: i tried to add `IFNULL(s.score, 0.2) AS s.score` here and it wasn't happy with me

linabutler: Try `IFNULL(s.score, 0.2) AS score` 😊

`SELECT s.score` is shorthand for `SELECT s.score AS score`—the left-hand side of the `AS` is the qualified column name, and the right-hand side is the alias; a qualified column name can’t appear on the right-hand side.

components/suggest/src/db.rs
```diff
@@ -140,17 +140,17 @@ impl<'a> SuggestDao<'a> {
...
                      WHERE s.provider IN ({}) AND
                            k.keyword = :keyword
                      UNION ALL
-                     SELECT s.id, k.rank, s.title, s.url, s.provider, k.confidence, k.keyword_suffix
+                     SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, k.confidence, k.keyword_suffix
                      FROM suggestions s
                      JOIN prefix_keywords k ON k.suggestion_id = s.id
                      WHERE k.keyword_prefix = :keyword_prefix
-                     ORDER BY s.provider
+                     ORDER BY s.score DESC
```
linabutler, line 153: I think you’ll want `IFNULL(s.score, 0.2)` here, too, or any suggestions without a score will sort last.

Could you please make `0.2` a constant, too? It’ll help clarify that it’s the default score, so that we aren’t embedding magic numbers in the query. Thanks!

components/suggest/src/db.rs
```diff
@@ -162,12 +162,12 @@ impl<'a> SuggestDao<'a> {
...
             (self.conn.prepare_cached(
                 &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
+                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
                      FROM suggestions s
                      JOIN keywords k ON k.suggestion_id = s.id
                      WHERE s.provider IN ({}) AND
                            k.keyword = :keyword
-                     ORDER BY s.provider
+                     ORDER BY s.score
```
linabutler, line 170: Don’t forget `IFNULL` and `DESC` here, too, please!

components/suggest/src/db.rs
linabutler: Another tip: If you write `IFNULL(s.score, 0.2) AS score` in the `SELECT`, you can just write `ORDER BY score DESC` here! 🎉 `ORDER BY` can refer to any columns or aliased expressions from the `SELECT`.

SQLite's [documentation for SELECT](https://sqlite.org/lang_select.html) is a long read, but I promise, it's _super_ worth it for helping the syntax feel less arcane.

linabutler (dismissed):
I replied to your comments 😊 Almost there!

components/suggest/src/store.rs
```diff
@@ -1990,6 +2032,24 @@ mod tests {
                 },
                 expect![[r#"
                     [
+                        Pocket {
```
linabutler, line 2035: I think this test is getting a bit too long...and I think we might also be missing some test coverage that would've caught the missing `DESC` in the first query. We can save this for a follow-up, though, unless you'd like to tackle it now!

To make sure we're covering all the cases:

Let's add one test, like `amp_wikipedia_scores`, that ingests AMP and Wikipedia suggestions. This test can exercise sorting and limits for the query that just selects from `keywords`. We can have it ingest two AMP suggestions—one with a score of 0.3, and one with 0.1—and a Wikipedia suggestion without a score.

We can then verify that the suggestions are sorted correctly, with and without a limit: AMP with 0.3, Wikipedia with the default of 0.2, and AMP with 0.1. I know we won't see AMP suggestions like that IRL (yet?); this is more to verify that the `ORDER BY` works, and that `ORDER BY` combined with a `LIMIT` returns them in the right order.

Let's have our second test, `wikipedia_pocket_scores`, ingest Wikipedia and Pocket suggestions. This test can exercise sorting and limits for the query that selects from `prefix_keywords` and `keywords`.

We can follow the same "sandwich" approach for this second test as for the first: a Pocket suggestion with a score higher than 0.2, a Wikipedia suggestion without a score, and a second Pocket suggestion with a lower score.

components/suggest/src/suggest.udl
```diff
@@ -38,7 +38,8 @@ interface Suggestion {
         string iab_category,
         string impression_url,
         string click_url,
-        string raw_click_url
+        string raw_click_url,
+        f64? score
```
linabutler, line 42: Since we always use the default score now, I wonder if we should make this non-optional. WDYT?

components/suggest/src/db.rs
```diff
@@ -140,17 +140,17 @@ impl<'a> SuggestDao<'a> {
         {
             (self.conn.prepare_cached(
                 &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, NULL as confidence, NULL as keyword_suffix
+                    "SELECT s.id, k.rank, s.title, s.url, s.provider, IFNULL(s.score, 0.2) AS score, NULL as confidence, NULL as keyword_suffix
```
linabutler, line 143: I'd ❤️ to see this in a constant eventually—like `DEFAULT_SUGGESTION_SCORE` to match Desktop's naming—so that we aren't embedding magic numbers in our queries, but that can be in a follow-up.

linabutler (approved):
r+wc, thanks!

components/suggest/src/db.rs
linabutler: Oh, I was thinking that we leave the database field as nullable, so that if we ever wanted to experiment with changing the default score, we'd have a way of distinguishing them!

components/suggest/src/db.rs
```diff
@@ -182,6 +182,7 @@ impl<'a> SuggestDao<'a> {
                 let title = row.get("title")?;
                 let raw_url = row.get::<_, String>("url")?;
                 let provider = row.get("provider")?;
+                let score = row.get::<_,Option<f64>>("score")?;
```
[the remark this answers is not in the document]
linabutler:
```suggestion
                let score = row.get::<_, f64>("score")?;
```

Oops, sorry I didn't catch this! Since you're using `IFNULL` in the query, this doesn't need to be an `Option` anymore—it'll never be `None`.

components/suggest/src/db.rs
```diff
@@ -261,7 +263,7 @@ impl<'a> SuggestDao<'a> {
                                         rating: row.get("rating")?,
                                         number_of_ratings: row.get("number_of_ratings")?,
                                         guid: row.get("guid")?,
-                                        score: row.get("score")?,
+                                        score: score.unwrap_or_default(),
```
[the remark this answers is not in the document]
linabutler: Or if you change to `row.get::<_, f64>("score")?` above.

components/suggest/src/suggest.udl
linabutler: On further thought, let's make this non-optional—there's no need to keep it nullable if we always set its value.

components/suggest/src/db.rs
ncloudioj: > Oh, I was thinking that we leave the database field as nullable, so that if we ever wanted to experiment with changing the default score, we'd have a way of distinguishing them!

I think using a default score in the `suggestion` table should be fine. For experimentation, we can always override it via external sources (like what we do for AMP & Pocket suggestions). In the longer term, I'd foresee a path that we'd adjust the `score`s from DB for relevance ranking perhaps in the application.

linabutler: Sounds good to me, thanks! 👍🏼

components/suggest/src/db.rs
```diff
@@ -423,8 +426,7 @@ impl<'a> SuggestDao<'a> {
                     ":record_id": record_id.as_str(),
                     ":title": common_details.title,
                     ":url": common_details.url,
-                    ":score": common_details.score,
-
+                    ":score": common_details.score.unwrap_or(DEFAULT_SUGGESTION_SCORE)
```
tiftran, line 429: So I tried a couple things with the schema `score REAL DEFAULT 0.2` it was unhappy with me try to insert a NULL value. 
I landed on `score REAL NOT NULL ON CONFLICT REPLACE DEFAULT 0.2` and i was uncertain of that too

components/suggest/src/schema.rs
```diff
@@ -51,11 +52,6 @@ pub const SQL: &str = "
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
 
-    CREATE TABLE pocket_custom_details(
```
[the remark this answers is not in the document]
tiftran: how does one do that 😅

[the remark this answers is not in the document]
ncloudioj: I believe `DROP TABLE IF EXISTS pocket_custom_details` should do.

After a second thought, since we're bumping the schema version as well, we will rebuild the entire db anyway. Maybe we don't need to drop anything. @linabutler Could you please hold me true? :)

components/suggest/src/db.rs
ncloudioj: Right, `score DEFAULT 0.2` wont work because we always specify `score` in `INSERT`. This works equally fine 👍

components/suggest/src/db.rs
```diff
@@ -33,6 +33,9 @@ pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
 /// that aren't parsable and which schema version it was first seen in.
 pub const UNPARSABLE_RECORDS_META_KEY: &str = "unparsable_records";
 
+// Default value when Suggestion does not have a value for score
+pub const DEFAULT_SUGGESTION_SCORE: f64 = 0.2;
```
ncloudioj, line 37: Does this need to be `pub`?

ncloudioj (approved):
LGTM, thanks!

components/suggest/src/schema.rs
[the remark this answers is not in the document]
linabutler: Nope, no need for a `DROP TABLE`. If we were doing a durable migration, where we wanted to preserve the database, we would—but the `1..10` branch in `upgrade_from` is replacing the entire database, so we can just delete this `CREATE TABLE` statement.

linabutler (approved):

PR #6060, opened 2024-01-17
Don't report sentry errors for metadata view times too long

bendk:
We're getting around 50-75 reports of these a day in Sentry for Fenix nightly and I don't think they're useful.

- [search for the Sentry errors](https://mozilla.sentry.io/discover/results/?field=title&field=release&field=environment&field=user.display&field=timestamp&name=places-unexpected-error%3A+Invalid+metadata+observation%3A+Observed+view+time+is+invalid+%28too+long%29&project=6295546&query=Observed+view+time+is+invalid+%28too+long%29&sort=-timestamp&statsPeriod=7d&yAxis=count%28%29)
- [the android-components code that I believe causes this](https://github.com/mozilla-mobile/firefox-android/blob/205ece4bf9234d24135ee780f886367d64b11735/fenix/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataService.kt#L92-L121)

Maybe there's a bug in that code that we could fix (for example not reporting an observation if the `tab.id` key is missing), but I'm not sure it's worth it.  I also could see plenty of reasons for valid code to cause this, for example if the user's clock is adjusted.  I'm currently thinking that the best policy is to just ignore and clean up our sentry reports.

  - Note:

### DIFF
--- a/components/places/src/error.rs
+++ b/components/places/src/error.rs
@@ -268,6 +268,12 @@ impl GetErrorHandling for Error {
                     .report_error("places-unexpected-sync-error"),
                 }
             }
+            Error::InvalidMetadataObservation(InvalidMetadataObservation::ViewTimeTooLong) => {
+                ErrorHandling::convert(PlacesApiError::UnexpectedPlacesException {
+                    reason: self.to_string(),
+                })
+                .log_warning()
+            }
             _ => ErrorHandling::convert(PlacesApiError::UnexpectedPlacesException {
                 reason: self.to_string(),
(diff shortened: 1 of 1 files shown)


mhammond:
I believe that linked a-c code has a bug with the timestamps - eg, a device going to sleep and waking without having any focus changes might not trigger the "stop the counter" logic which exists. But yeah, I agree there's no value in recording this in sentry.

mhammond (approved):

PR #6066, opened 2024-01-19
Update Gradle to version 8.5 and AGP to version 8.0.2 [ci full][firefox-android: main]

rvandermeulen:


rvandermeulen:
Looks like AGP 8.1+ has the same kinds of issues here that we've also been seeing in Glean and firefox-android (i.e. overly-aggressive R8 optimizations causing problems like `UnresolvedLinkError`). Going to stick with 8.0.2 for now like we're using elsewhere as the most expedient option for unblocking this update.

rvandermeulen:
I've also asked mcarare to look over this as well given his extensive experience doing this update elsewhere 😄

bendk (approved):
Looks good, let's try this out for the 124 release.

PR #6072, opened 2024-01-22
[ci full] Update NSS to 3.97

skhamis:
(thanks @mhammond  for this thought) To successfully build NSS to 3.97, we had to bump the SDK to Mac OSX 11.0. This has the implication that the minimum version to build our android stuffs to require having Mac OSX SDK 11.0 or higher. 

While I don't think this will impact anyone whose updated their computer in the last couple of years, there is a slight chance that some older machines might run into trouble. I'll/we'll monitor this closely within the next couple of weeks.  

  - Note:

### DIFF
--- a/libs/build-nss-desktop.sh
+++ b/libs/build-nss-desktop.sh
@@ -62,8 +62,8 @@ if [[ "${CROSS_COMPILE_TARGET}" =~ "darwin" ]]; then
     # run-task has already downloaded + extracted the dependency
     NSS_DIST_DIR="${MOZ_FETCHES_DIR}/dist"
   else
-    curl -sfSL --retry 5 --retry-delay 10 -O "https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/app-services.cache.level-3.content.v1.nss-artifact.latest/artifacts/public/dist.tar.bz2"
-    SHA256="15e8cab63a4e0484564231a24d23d8be349217ba6add53b56609c339bb306829"
+    curl -sfSL --retry 5 --retry-delay 10 -O "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/yFaMDJo8TgSJeZ51Npqxbg/runs/0/artifacts/public/dist.tar.bz2"
+    SHA256="1f4ba2d130a59f3c95e316d39667fc8a0e9d5db052469091ed17bef18a563066"
     echo "${SHA256}  dist.tar.bz2" | shasum -a 256 -c - || exit 2
     tar xvjf dist.tar.bz2 && rm -rf dist.tar.bz2
     NSS_DIST_DIR=$(abspath "dist")
--- a/libs/macos-cc-tools.manifest
+++ b/libs/macos-cc-tools.manifest
@@ -1,13 +1,5 @@
 [
   {
-    "size": 31991917,
-    "visibility": "internal",
-    "digest": "c5c0be09972b56b5980dc9d06b61ff49cf58c4572913437256a79b202e19e936af3c0ab0924df72b9f648d518c257597f84800a84bb80e68af4eabdaf1df5f24",
-    "algorithm": "sha512",
-    "unpack": true,
-    "filename": "MacOSX10.12.sdk.tar.xz"
-  },
-  {
     "size": 68798037,
     "visibility": "internal",
     "digest": "5bdc3270d0b660889b66194aa8cbcdbdb9163188bf976f31cc87837e6b0b5590db8d76cc17e9ea863b50f6735779cfd4338a4121997888b13139f4af1fdbb504",
--- a/libs/macos-cc-tools.manifest
+++ b/libs/macos-cc-tools.manifest
@@ -6,5 +6,13 @@
     "algorithm": "sha512",
     "unpack": true,
     "filename": "MacOSX10.12.sdk.tar.xz"
+  },
+  {
+    "size": 68798037,
+    "visibility": "internal",
+    "digest": "5bdc3270d0b660889b66194aa8cbcdbdb9163188bf976f31cc87837e6b0b5590db8d76cc17e9ea863b50f6735779cfd4338a4121997888b13139f4af1fdbb504",
+    "algorithm": "sha512",
+    "unpack": true,
+    "filename": "MacOSX11.0.sdk.tar.zst"
   }
 ]
--- a/taskcluster/scripts/toolchain/cross-compile-setup.sh
+++ b/taskcluster/scripts/toolchain/cross-compile-setup.sh
@@ -11,9 +11,9 @@ export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_TOOLCHA
 export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_AR=/builds/worker/cctools/bin/x86_64-apple-darwin-ar
 export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_RANLIB=/builds/worker/cctools/bin/x86_64-apple-darwin-ranlib
 export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_LD_LIBRARY_PATH=/builds/worker/clang/lib
-export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS="-C linker=/builds/worker/clang/bin/clang -C link-arg=-fuse-ld=/builds/worker/cctools/bin/x86_64-apple-darwin-ld -C link-arg=-B -C link-arg=/builds/worker/cctools/bin -C link-arg=-target -C link-arg=x86_64-apple-darwin -C link-arg=-isysroot -C link-arg=/tmp/MacOSX10.12.sdk -C link-arg=-Wl,-syslibroot,/tmp/MacOSX10.12.sdk -C link-arg=-Wl,-dead_strip"
+export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS="-C linker=/builds/worker/clang/bin/clang -C link-arg=-fuse-ld=/builds/worker/cctools/bin/x86_64-apple-darwin-ld -C link-arg=-B -C link-arg=/builds/worker/cctools/bin -C link-arg=-target -C link-arg=x86_64-apple-darwin -C link-arg=-isysroot -C link-arg=/tmp/MacOSX11.0.sdk -C link-arg=-Wl,-syslibroot,/tmp/MacOSX11.0.sdk -C link-arg=-Wl,-dead_strip"
 # For ring's use of `cc`.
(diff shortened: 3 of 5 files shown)


rvandermeulen:
@skhamis https://github.com/mozilla/application-services/blob/main/libs/build-nss-desktop.sh#L66 needs updating also.

skhamis:
> @skhamis https://github.com/mozilla/application-services/blob/main/libs/build-nss-desktop.sh#L66 needs updating also.

Ah I was gonna see if the L1/L2 cache was going to get stored in TC for this but then I realized I would need to update it to T3 after it merges anyways 😶‍🌫️ -- thanks for the catch!

skhamis:
So unfortunately this does looks like we're missing some symbols:

error:
```
note: Undefined symbols for architecture x86_64:
"_syslog$DARWIN_EXTSN", referenced from:
_sftk_LogAuditMessage in libnss_sys-eb10ec44728ac76a.rlib(softokn_static.fipstokn.o)
"____chkstk_darwin", referenced from:
_CERT_NameToAsciiInvertible in libnss_sys-eb10ec44728ac76a.rlib(certdb.alg1485.o)
_sftkdb_Update in libnss_sys-eb10ec44728ac76a.rlib(softokn_static.sftkdb.o)
_point_mul_secp521r1_wrap in libnss_sys-eb10ec44728ac76a.rlib(freebl_static.ecp_secp521r1.o)
_point_mul_two_secp521r1_wrap in libnss_sys-eb10ec44728ac76a.rlib(freebl_static.ecp_secp521r1.o)
_pqcrystals_kyber768_ref_keypair_derand in libnss_sys-eb10ec44728ac76a.rlib(freebl_static.kyber-pqcrystals-ref.o)
_pqcrystals_kyber768_ref_indcpa_enc in libnss_sys-eb10ec44728ac76a.rlib(freebl_static.kyber-pqcrystals-ref.o)
 _pqcrystals_kyber768_ref_dec in libnss_sys-eb10ec44728ac76a.rlib(freebl_static.kyber-pqcrystals-ref.o)
..
"___darwin_check_fd_set_overflow", referenced from:
_poll in libnss_sys-eb10ec44728ac76a.rlib(unix.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

From what I _can tell_ it seems we're missing LogAuditMessage that's in `fipstokn`. It also seems like it's mentioning some kyber crystals (heh) lib that was added in 3.97 via https://bugzilla.mozilla.org/show_bug.cgi?id=1826451.

I'll see if I can reach out for help on this.

tarikeshaq:
@skhamis  not sure if this will help in this specific case, but have some docs on tips for missing symbol build bustages when upgrading NSS: https://mozilla.github.io/application-services/book/howtos/upgrading-nss-guide.html#tips-for-fixing-bustage

mhammond:
Sorry if this is obvious, but I think there are "only" 3 symbols missing; `_syslog$DARWIN_EXTSN, `____chkstk_darwin` and `___darwin_check_fd_set_overflow`?

skhamis:
> Sorry if this is obvious, but I think there are "only" 3 symbols missing; `_syslog$DARWIN_EXTSN`, `____chkstk_darwin` and `___darwin_check_fd_set_overflow`?

Correct, though it's not clear if we need to manually add these or if we're missing a library to include. I was able to build NSS successfully locally though I'm M1 Mac.

It's looking like it's failing from the logs `Task :full-megazord:cargoBuildDarwin` so it's Linux trying to build x86 darwin for android. So I suspect either missing some mac lib or we'll need to export these manually

skhamis:
> > Sorry if this is obvious, but I think there are "only" 3 symbols missing; `_syslog$DARWIN_EXTSN`, `____chkstk_darwin` and `___darwin_check_fd_set_overflow`?
> 
> Correct, though it's not clear if we need to manually add these or if we're missing a library to include. I was able to build NSS successfully locally though I'm M1 Mac.
> 
> It's looking like it's failing from the logs `Task :full-megazord:cargoBuildDarwin` so it's Linux trying to build x86 darwin for android. So I suspect either missing some mac lib or we'll need to export these manually

Wait that actually doesn't make any sense. I _thought_ we should not have a `cargoBuildDarwin` anymore due to the fact that we need x86-64 and aarch64. 

https://github.com/mozilla/application-services/blob/b9adb61a235af1a9919651c58da46b664d0f7bb4/publish.gradle#L265 

https://github.com/mozilla/application-services/blob/b9adb61a235af1a9919651c58da46b664d0f7bb4/build.gradle#L217-L236

tarikeshaq:
There's a good chance that's coming from https://github.com/mozilla/application-services/blob/b9adb61a235af1a9919651c58da46b664d0f7bb4/taskcluster/kinds/module-build/kind.yml#L37 although I'm not sure it's exactly related to the issue we have here

darwin being there, but not it's x86 and arm variants is probably still https://github.com/mozilla/application-services/issues/4828

skhamis:
> There's a good chance that's coming from
> 
> https://github.com/mozilla/application-services/blob/b9adb61a235af1a9919651c58da46b664d0f7bb4/taskcluster/kinds/module-build/kind.yml#L37
> although I'm not sure it's exactly related to the issue we have here
> 
> darwin being there, but not it's x86 and arm variants is probably still #4828

Ah okay that makes sense, i guess at some point we'll need to update it to include M1 but `darwin` defaults to x86-64 anyways (lol I forgot I wrote that patch 2 years ago) https://github.com/mozilla/rust-android-gradle/blame/master/plugin/src/main/kotlin/com/nishtahir/RustAndroidPlugin.kt#L28

tarikeshaq:
All those like you said, look suspiciously like we're trying to build against an older MacOS SDK when NSS updated it's minimum version 🤔

The below seems to be using the Mac OS 10.X SDK, where at least one of the above symbols seems to have only been [introduced in Mac OS 11](https://stackoverflow.com/questions/68095018/dyld-symbol-not-found-darwin-check-fd-set-overflow)
 https://github.com/mozilla/application-services/blob/b9adb61a235af1a9919651c58da46b664d0f7bb4/taskcluster/scripts/toolchain/cross-compile-setup.sh#L14

tarikeshaq:
Looks like there is magic there that you might be more familiar with in terms of bumping our Mac OS SDK, (looks like we bumped a minor version last week in https://github.com/mozilla/application-services/pull/6073/files) but trying to bump it seems like a good place to start

skhamis:
> Looks like there is magic there that you might be more familiar with in terms of bumping our Mac OS SDK, (looks like we bumped a minor version last week in https://github.com/mozilla/application-services/pull/6073/files) but trying to bump it seems like a good place to start

Great idea! I found we do have a MacOSX 11.0 in tooltool so I tweaked our settings a bit and it looks like that did the trick!

skhamis:
Thanks @tarikeshaq and @mhammond for helping brainstorm/debug this! Looks like CI is green and I've added ya'll as reviewers since ya'll have the closest context at this point.

libs/macos-cc-tools.manifest
```diff
@@ -6,5 +6,13 @@
     "algorithm": "sha512",
     "unpack": true,
     "filename": "MacOSX10.12.sdk.tar.xz"
```
tarikeshaq, line 8: Do we need to keep the 10.12 SDK for any reason, or can we kill it and replace it with the new one?

taskcluster/scripts/toolchain/cross-compile-setup.sh
```diff
@@ -44,5 +44,10 @@ tooltool.py \
   --url=http://taskcluster/tooltool.mozilla-releng.net/ \
   --manifest="/builds/worker/checkouts/vcs/libs/macos-cc-tools.manifest" \
   fetch
+# tooltool doesn't know how to unpack zstd-files,
+# so we do it manually.
+tar -I zstd -xf "MacOSX11.0.sdk.tar.zst"
 
 popd || exit
+
+set +eu
```
tarikeshaq, line 53: nit: no new line at the end of the file

taskcluster/scripts/toolchain/cross-compile-setup.sh
```diff
@@ -11,9 +11,9 @@ export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_TOOLCHA
...
+export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS="-C linker=/builds/worker/clang/bin/clang -C link-arg=-fuse-ld=/builds/worker/cctools/bin/x86_64-apple-darwin-ld -C link-arg=-B -C link-arg=/builds/worker/cctools/bin -C link-arg=-target -C link-arg=x86_64-apple-darwin -C link-arg=-isysroot -C link-arg=/tmp/MacOSX11.0.sdk -C link-arg=-Wl,-syslibroot,/tmp/MacOSX11.0.sdk -C link-arg=-Wl,-dead_strip"
 # For ring's use of `cc`.
-export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_CFLAGS_x86_64_apple_darwin="-B /builds/worker/cctools/bin -target x86_64-apple-darwin -isysroot /tmp/MacOSX10.12.sdk -Wl,-syslibroot,/tmp/MacOSX10.12.sdk -Wl,-dead_strip"
+export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_CFLAGS_x86_64_apple_darwin="-B /builds/worker/cctools/bin -target x86_64-apple-darwin -isysroot /tmp/MacOSX11.0.sdk -Wl,-syslibroot,/tmp/MacOSX10.12.sdk -Wl,-dead_strip"
```
tarikeshaq, line 16:
```suggestion
export ORG_GRADLE_PROJECT_RUST_ANDROID_GRADLE_TARGET_X86_64_APPLE_DARWIN_CFLAGS_x86_64_apple_darwin="-B /builds/worker/cctools/bin -target x86_64-apple-darwin -isysroot /tmp/MacOSX11.0.sdk -Wl,-syslibroot,/tmp/MacOSX11.0.sdk -Wl,-dead_strip"
```

libs/build-nss-desktop.sh
```diff
@@ -62,8 +62,8 @@ if [[ "${CROSS_COMPILE_TARGET}" =~ "darwin" ]]; then
     # run-task has already downloaded + extracted the dependency
     NSS_DIST_DIR="${MOZ_FETCHES_DIR}/dist"
   else
-    curl -sfSL --retry 5 --retry-delay 10 -O "https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/app-services.cache.level-3.content.v1.nss-artifact.latest/artifacts/public/dist.tar.bz2"
```
tarikeshaq, line 65: Looks like we're switching here from using our own taskcluster index to a specific run's artifacts, is this intentional? I vaguely assumed that we now pull from the CI in the `fetch` taskcluster job, then upload it here, then we later just use it from our index?

I'm out of my depth here so I could be very off 😅

tarikeshaq (changes requested):
Thanks @skhamis!! Had a few questions but this looks great!! Thank you for going down the painful path to get this to this state 🙏

libs/build-nss-desktop.sh
skhamis: Yeah this will be switched to our L3 cache, but I think it's more clear to just use NSS TC since it's the same as CI right now then update to use the L3 cache once this lands. I feel our docs are not as clear on the best approach here and will make a task to update that.

libs/macos-cc-tools.manifest
skhamis: yeah I was thinking we might want to keep it around when I was testing both versions, but I think it makes sense to just remove this entirely to reduce confusion.

tarikeshaq (approved):
Awesome thanks! I'm assuming CI will pass so once it does land away!

PR #6077, opened 2024-01-24
Adding network/auth retries to the fxa state machine

bendk:
While trying to clean up the FxA tickets, I realized I had this code that I forgot to push out.  This implements https://bugzilla.mozilla.org/show_bug.cgi?id=1871061.

  - Note:

### DIFF
--- a/components/fxa-client/src/state_machine/internal_machines/mod.rs
+++ b/components/fxa-client/src/state_machine/internal_machines/mod.rs
@@ -98,23 +98,17 @@ impl State {
         account: &mut FirefoxAccount,
         device_config: &DeviceConfig,
     ) -> Result<Event> {
-        let is_ensure_capabilities = matches!(self, State::EnsureDeviceCapabilities);
-        self.make_call_inner(account, device_config).or_else(|e| {
-            // All errors get converted to events, except StateMachineLogicError
-            if matches!(e, Error::StateMachineLogicError(_)) {
-                Err(e)
-            } else {
-                // This call is mostly to report the error, but converting `Error` to `FxaError`
-                // also simplifies the match for authentication errors since multiple `Error`
-                // variants map to `FxaError::Authentication`.
-                let fxa_error = convert_log_report_error(e);
-                if is_ensure_capabilities && matches!(fxa_error, FxaError::Authentication) {
-                    Ok(Event::EnsureCapabilitiesAuthError)
-                } else {
-                    Ok(Event::CallError)
-                }
-            }
-        })
+        let mut error_handling = CallErrorHandler::new(self);
+        loop {
+            return match self.make_call_inner(account, device_config) {
+                Ok(event) => Ok(event),
+                Err(e) => match error_handling.handle_error(e, account) {
+                    CallResult::Retry => continue,
+                    CallResult::Finished(event) => Ok(event),
+                    CallResult::InternalError(err) => Err(err),
+                },
+            };
+        }
     }
 
     fn make_call_inner(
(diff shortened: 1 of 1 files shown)


bendk:
@linabutler I was going to also add you as a reviewer, but then I figured you probably have a bunch of work on your plate for the rotation, so I decided against it.  But I'd love to hear your thoughts in any case and feel free to add yourself as a reviewer if you have the time.

components/fxa-client/src/state_machine/internal_machines/mod.rs
```diff
@@ -177,6 +171,66 @@ impl State {
     }
 }
 
+/// Number of times to retry fxa calls in the face of network errors
+const NETWORK_RETRY_LIMIT: usize = 3;
+/// Number of times to retry fxa calls in the face of auth issues
+const AUTH_RETRY_LIMIT: usize = 1;
```
mhammond, line 177: I'm interested to know why we bother retrying here at all?

tarikeshaq: Hmm yeah - logically what we would be doing is:
- get a 4xx that tells us our access token has expired,
- use our refresh token to get a new access token,
- try again

Just retrying won't do anything so +1 to make sure we check our refresh token and ask for a new access token before retrying

components/fxa-client/src/state_machine/internal_machines/mod.rs
```diff
@@ -177,6 +171,88 @@ impl State {
...
+                    //   - Clear the access token
+                    //   - Call `check_authorization_status`.  If successful we can retry the operation.
+                    account.clear_access_token_cache();
+                    match account.check_authorization_status() {
+                        Ok(status) if status.active => {
+                            self.auth_retries += 1;
+                            CallResult::Retry
+                        }
+                        _ => CallResult::Finished(self.event_for_auth_error()),
+                    }
```
bendk, line 225: This code is trying to match what the android-components code currently does.  That's tricky, because that code is not so easy to read, but here's how I understand it to work:

- [When operations fail, it transitions to the RecoveringFromAuthProblem state](https://github.com/mozilla-mobile/firefox-android/blob/3872fae14232f1f9d1d2486647a5d9f88c608064/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/State.kt#L146).
  - Note: The android-components code uses a state transition for this, while this code stays in the same state while it tries to recover/retry.  I think that's an improvement, because when you do successfully recover from an auth error, it's not clear exactly how you should transition back to the previous state to retry.  I think I found at least one [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1868418) in how the android-components code does it.
- When the android-components code reaches the [RecoveringFromAuthProblem](https://github.com/mozilla-mobile/firefox-android/blob/3872fae14232f1f9d1d2486647a5d9f88c608064/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L654) state, it clears the access token cach and calls `check_authorization_status()`.  Note: the call to [authErrorDetected](https://github.com/mozilla-mobile/firefox-android/blob/3872fae14232f1f9d1d2486647a5d9f88c608064/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt#L190-L194) is just another way to clear out the access tokens.
- There is also code to implement a "circuit-breaker" to avoid infinite loops where the component transitions to the RecoveringFromAuthProblem state, recovers and transitions back to the previous state, hits another auth error and starts the loop again.  I don't think that's an issue with this code because don't mode the retry as a state transition and `CallErrorHandler` ensures that it only does the auth retry once.
  - See also https://bugzilla.mozilla.org/show_bug.cgi?id=1864994 for my thoughts on how the current circuit breaker doesn't prevent a class of auth retry issues and how I think we should handle those.

tarikeshaq: on the retry there are a couple of cases, the code we have here properly handles one, so we just need to make sure the other one is OK
- case 1: The refresh token itself is no longer valid, it was revoked, expired, etc. Refresh tokens are long-lived so this shouldn't happen often but we should still handle it. In that case, the best we can do is put the user in an "auth error state" so they can sign back in and that seems well handled here
- case 2: The access token, which we provided to the app, is expired - this happens regularly, and in our code to "get_access_token" [we periodically would ask for a new access token instead of returning from cache](https://github.com/mozilla/application-services/blob/43b8447cb16fecfc8f935154f5e9a411154a4964/components/fxa-client/src/internal/oauth.rs#L43-L67)

The second case is tricky, in that the 4xx error isn't usually first caught here in our component, and is usually caught outside because we provide the access token to our consumers to use (so in practice the 4xx comes from sync manager). In other words, we still need the consumer to "manually" clear the access token and ask for a new one when they hit a 4xx while consuming our access token. Do you know if that will stay in-place or if this logic is meant to replace that all together?

bendk: I *think* this will be handled correctly.  In the [PR to start using the new state machine code](https://github.com/mozilla-mobile/firefox-android/pull/4912/files) here's how it works:
  - Some component notifies `GlobalAccountManager` that there was an account manager, which then notifiies `FxaAccountManager`
  - [FxaAccountManager then processes the AuthenticationError event](https://github.com/mozilla-mobile/firefox-android/blob/19876e2460ee05cbc9a81bdcda741c3bb385d447/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L404) (the changes start after this).
  - [FxaAccountManger converts that event to CheckAuthorizationStatus](https://github.com/mozilla-mobile/firefox-android/blob/19876e2460ee05cbc9a81bdcda741c3bb385d447/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L462) to send to the new state machine
  - The new state machine code calculates the new state.  [That state, plus the original `AuthenticationError` event gets sent to accountStateSideEffects](https://github.com/mozilla-mobile/firefox-android/blob/19876e2460ee05cbc9a81bdcda741c3bb385d447/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L436C65-L436C70)
  - If the recovery worked, the user will stay in the `Connected` and [this code](https://github.com/mozilla-mobile/firefox-android/blob/19876e2460ee05cbc9a81bdcda741c3bb385d447/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L524) will clear the stale access token.  (Also the call to `resetAccount()` will do the same thing if the recovery fails).

So, from the POV of app-services, I don't think we need to do anything special since consumers can figure out for themselves when they need to clear their access token.  Maybe eventually we could have a callback interface with an event for that, but that can be a future change.  From the POV of android-components, we need to make sure that that path is accounted for and handled correctly.  I'm kind of worried that we'll forget about that, so I just added a checklist of things to make sure we get right to that bug.  Hopefully our future selves will see it and double-check.

tarikeshaq: Alrighty that sounds good then - it's a shame we've setup in this way where consumers have to know what to do when they hit a 400, but any change there is a larger scope :(

bendk: > It's a shame we've setup in this way where consumers have to know what to do when they hit a 400, but any change there is a larger scope :(

I'd love to see the state machine expand to handling more operations than auth.  That way we could handle the 400 errors internally.

PR #6078, opened 2024-01-24
Re-order members of RemoteSettingsConfig UDL dict [firefox-android: main] [firefox-ios: main]

brennie:
The order of dictionary members matters to uniffi when they are used to generate arguments for Python functions. In this case, the order was generating a method signature with defaulted kwargs before undefaulted args, which is a syntax error. Re-arranging these members will result in generating a valid Python method signature. The Swift bindings also need updating, because Swift enforces argument order for named arguments, but Kotlin does not.

Fixes bug 1874030

  - Note:

### DIFF
--- a/components/remote_settings/src/remote_settings.udl
+++ b/components/remote_settings/src/remote_settings.udl
@@ -8,9 +8,9 @@ typedef string RsJsonObject;
 namespace remote_settings {};
 
 dictionary RemoteSettingsConfig {
-    string? server_url = null;
-    string? bucket_name = null;
     string collection_name;
+    string? bucket_name = null;
+    string? server_url = null;
 };
 
 dictionary RemoteSettingsResponse {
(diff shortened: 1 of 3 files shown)


components/remote_settings/src/remote_settings.udl
```diff
@@ -8,9 +8,9 @@ typedef string RsJsonObject;
 namespace remote_settings {};
 
 dictionary RemoteSettingsConfig {
-    string? server_url = null;
-    string? bucket_name = null;
     string collection_name;
+    string? bucket_name = null;
+    string? server_url = null;
```
jhugman, line 13: Nit: the URL is more significant than the `bucket_name` or `collection_name`; would you consider having the url before the bucket?

jhugman (approved):
This requires a Breaking Change in the CHANGELOG.md; there may not be any changes in Firefox for iOS, or Android, but we should still mark it as so.

components/remote_settings/src/remote_settings.udl
brennie: It is, but its also probably the most likely to be defaulted vs the other two. I think either this order or the previous order with no default value support are the most correct options. I'm going to keep this as is.

mhammond:
IIUC, this is a breaking change for suggest in desktop.

PR #6085, opened 2024-01-30
[DISCO-2727] Suggest: Refactor fetch_suggestions()

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-use std::{path::Path, sync::Arc};
+use std::{collections::HashSet, path::Path, sync::Arc};
 
 use interrupt_support::{SqlInterruptHandle, SqlInterruptScope};
 use parking_lot::Mutex;
@@ -34,7 +34,7 @@ pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
 pub const UNPARSABLE_RECORDS_META_KEY: &str = "unparsable_records";
 
 // Default value when Suggestion does not have a value for score
-const DEFAULT_SUGGESTION_SCORE: f64 = 0.2;
+pub const DEFAULT_SUGGESTION_SCORE: f64 = 0.2;
 
 /// The database connection type.
 #[derive(Clone, Copy)]
@@ -128,63 +128,46 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
-        if let Some(suggestion) = self.fetch_yelp_suggestion(query)? {
-            return Ok(vec![suggestion]);
-        }
-
-        let keyword_lowercased = &query.keyword.to_lowercase();
-        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
-        let suggestions_limit = query.limit.unwrap_or(-1);
-
-        let (mut statement, params) = if query
-            .providers
+        let unique_providers = query.providers.iter().collect::<HashSet<_>>();
+        unique_providers
             .iter()
-            .any(|p| matches!(p, SuggestionProvider::Pocket | SuggestionProvider::Amo))
-        {
-            (self.conn.prepare_cached(
-                &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
-                     FROM suggestions s
-                     JOIN keywords k ON k.suggestion_id = s.id
-                     WHERE s.provider IN ({}) AND
-                           k.keyword = :keyword
-                     UNION ALL
-                     SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, k.confidence, k.keyword_suffix
-                     FROM suggestions s
-                     JOIN prefix_keywords k ON k.suggestion_id = s.id
-                     WHERE k.keyword_prefix = :keyword_prefix
-                     ORDER BY s.score DESC
-                     LIMIT :suggestions_limit",
-                    providers_to_sql_list(&query.providers),
-                ),
-            )?, vec![
-                (":keyword", keyword_lowercased as &dyn ToSql),
-                (":keyword_prefix", &keyword_prefix as &dyn ToSql),
-                (":suggestions_limit", &suggestions_limit as &dyn ToSql)
-            ])
-        } else {
-            (self.conn.prepare_cached(
-                &format!(
-                    "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, NULL as confidence, NULL as keyword_suffix
-                     FROM suggestions s
-                     JOIN keywords k ON k.suggestion_id = s.id
-                     WHERE s.provider IN ({}) AND
-                           k.keyword = :keyword
-                     ORDER BY s.score DESC
-                     LIMIT :suggestions_limit",
-                    providers_to_sql_list(&query.providers),
-                ),
-            )?, vec![
-                (":keyword", keyword_lowercased as &dyn ToSql),
-                (":suggestions_limit", &suggestions_limit as &dyn ToSql)
-            ])
-        };
+            .try_fold(vec![], |mut acc, provider| {
+                let suggestions = match provider {
+                    SuggestionProvider::Amp => self.fetch_amp_suggestions(query),
+                    SuggestionProvider::Wikipedia => self.fetch_wikipedia_suggestions(query),
+                    SuggestionProvider::Amo => self.fetch_amo_suggestions(query),
+                    SuggestionProvider::Pocket => self.fetch_pocket_suggestions(query),
+                    SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
+                }?;
+                acc.extend(suggestions);
+                Ok(acc)
+            })
+            .map(|mut suggestions| {
+                suggestions.sort();
+                if let Some(limit) = query.limit.and_then(|limit| usize::try_from(limit).ok()) {
+                    suggestions.truncate(limit);
+                }
+                suggestions
+            })
+    }
 
-        let suggestions = statement.query_and_then(&*params, |row| -> Result<Option<Suggestion>> {
+    /// Fetches Suggestions of type Amp provider that match the given query
+    pub fn fetch_amp_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        let keyword_lowercased = &query.keyword.to_lowercase();
+        let suggestions = self.conn.query_rows_and_then_cached(
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score
+             FROM suggestions s
+             JOIN keywords k ON k.suggestion_id = s.id
+             WHERE s.provider = :provider AND
+             k.keyword = :keyword",
+             named_params! {
+                ":keyword": keyword_lowercased,
+                ":provider": SuggestionProvider::Amp
+            },
+            |row| -> Result<Suggestion>{
                 let suggestion_id: i64 = row.get("id")?;
                 let title = row.get("title")?;
                 let raw_url = row.get::<_, String>("url")?;
-                let provider = row.get("provider")?;
                 let score = row.get::<_, f64>("score")?;
 
                 let keywords: Vec<String> = self.conn.query_rows_and_then_cached(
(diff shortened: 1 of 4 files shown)


components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
+                    | Suggestion::Amo { score, .. } => score,
+                    _ => &DEFAULT_SUGGESTION_SCORE,
+                };
+                b_score
+                    .partial_cmp(a_score)
+                    .unwrap_or(std::cmp::Ordering::Equal)
+            });
+            if suggestions.len() > suggestions_limit as usize {
+                suggestions.truncate(suggestions_limit as usize);
+            }
```
tiftran, line 175: so sorting after fetching from each provider came with a snag and seems a little messy. Since Wikipedia and Yelp doesn't have a score field in their `Suggestion` structs, I used the `DEFAULT_SUGGESTION_SCORE`.

components/suggest/src/db.rs
```diff
@@ -197,58 +206,106 @@ impl<'a> SuggestDao<'a> {
...
+    }
+
+    pub fn fetch_amo_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        let keyword_lowercased = &query.keyword.to_lowercase();
+        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
+        let suggestions = self.conn.query_rows_and_then_cached(
+            "SELECT s.id, k.rank, s.title, s.url, s.provider, s.score, k.confidence, k.keyword_suffix
+                     FROM suggestions s
+                     JOIN prefix_keywords k ON k.suggestion_id = s.id
+                     WHERE k.keyword_prefix = :keyword_prefix AND s.provider = :provider",
```
tiftran, line 297: I was unsure whether it's worth using `ORDER BY` and `LIMIT` here and the other select statements to to reduce the number of suggestions we have to sort after. I don't think it's worth it for AMP since we're matching on keyword/keyword partials and that should only return 1 suggestion, but for AMO and Pocket that have prefixes...

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
+                        SuggestionProvider::Amp => self.fetch_amp_suggestions(query),
+                        SuggestionProvider::Wikipedia => self.fetch_wikipedia_suggestions(query),
+                        SuggestionProvider::Amo => self.fetch_amo_suggestions(query),
+                        SuggestionProvider::Pocket => self.fetch_pocket_suggestions(query),
+                        SuggestionProvider::Yelp => {
+                            if let Some(yelp_suggestion) = self.fetch_yelp_suggestion(query)? {
+                                Ok(vec![yelp_suggestion])
+                            } else {
+                                Ok(vec![])
+                            }
```
tiftran, line 148: this felt jank

tiftran:
just wanted to gather some feedback on my initial attempt to break things up 🙂

components/suggest/src/db.rs
ncloudioj: Now we can make `fetch_yelp_suggestion()` return `Result<Vec<Suggestion>>` instead of `Result<Option<Suggestion>>`.

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
     /// Fetches suggestions that match the given query from the database.
     pub fn fetch_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
-        if let Some(suggestion) = self.fetch_yelp_suggestion(query)? {
-            return Ok(vec![suggestion]);
-        }
-
-        let keyword_lowercased = &query.keyword.to_lowercase();
-        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
         let suggestions_limit = query.limit.unwrap_or(-1);
+        let suggestions =
```
ncloudioj, line 132: I think we can dedup providers before running them though just in case.

```rs
let unique_providers = query.providers.iter().collect::<HashSet<_>>();
unique_providers.iter().try_fold(...)
```

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
+                    | Suggestion::Pocket { score, .. }
+                    | Suggestion::Amo { score, .. } => score,
+                    _ => &DEFAULT_SUGGESTION_SCORE,
+                };
+                b_score
+                    .partial_cmp(a_score)
+                    .unwrap_or(std::cmp::Ordering::Equal)
+            });
+            if suggestions.len() > suggestions_limit as usize {
+                suggestions.truncate(suggestions_limit as usize);
```
ncloudioj, line 174: When `suggestions_limit` is greater than `suggestions.len()`, this has no effect. So you don't need the if check here.

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
+                                Ok(vec![])
+                            }
+                        }
+                    }?;
+                    acc.extend(suggestions);
+                    Ok(acc)
+                });
+
+        suggestions.map(|mut suggestions| {
+            suggestions.sort_by(|a, b| {
```
ncloudioj, line 156: I think it's time to implement `std::ops::PartialOrd` for `Suggestion` as we have a use case here. You can move everything to that implementation.

components/suggest/src/db.rs
ncloudioj: Let's add a test case for this :)

components/suggest/src/db.rs
linabutler: Thanks for highlighting this! Here are our top prefixes for AMO and Pocket, and the number of times they're repeated:

```
sqlite> SELECT keyword_prefix, count(*) FROM prefix_keywords GROUP BY keyword_prefix HAVING count(*) > 1 ORDER BY 2 DESC LIMIT 5;
how|209
best|56
the|37
vegan|37
what|35
```

...So doing `ORDER BY` and `LIMIT` in these queries sounds like a great idea to me! Using the original `query.limit` feels like a good starting point: if the caller passes a limit of `1`, even in the most impossible case of each provider returning 1 suggestion, that's still 5 suggestions at most, that we'll then truncate to 1.

If even that turns out to be inefficient, we can try dynamically adjusting the limit for each provider, based on the number of suggestions remaining, and the priority of the provider—but starting with the original limit still sounds good to me!

components/suggest/src/db.rs
```diff
@@ -197,58 +206,106 @@ impl<'a> SuggestDao<'a> {
...
+                ":keyword_prefix": keyword_prefix,
+                ":provider": SuggestionProvider::Amo
+            },
+            |row| -> Result<Option<Suggestion>>{
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let raw_url = row.get::<_, String>("url")?;
+                let score = row.get::<_, f64>("score")?;
+
+                let full_suffix = row.get::<_, String>("keyword_suffix")?;
```
linabutler, line 308: While you're here, do you think it's worth lifting the `full_suffix.starts_with(keyword_suffix)` check up here—so that we don't even query `amo_custom_details` if the suffixes don't match?

In other words:

```rs
let full_suffix = row.get::<_, String>("keyword_suffix")?;
if !full_suffix.starts_with(keyword_suffix) {
    return Ok(None);
}
self.conn.query_row_and_then(
    "SELECT amo.description, amo.guid, amo.rating, amo.icon_url, amo.number_of_ratings
```

components/suggest/src/db.rs
```diff
@@ -272,33 +329,50 @@ impl<'a> SuggestDao<'a> {
...
+                let raw_url = row.get::<_, String>("url")?;
+                let score = row.get::<_, f64>("score")?;
+                let confidence = row.get("confidence")?;
+                let full_suffix = row.get::<_, String>("keyword_suffix")?;
+                let suffixes_match = match confidence {
+                    KeywordConfidence::Low => full_suffix.starts_with(keyword_suffix),
+                    KeywordConfidence::High => full_suffix == keyword_suffix,
+                };
+                if suffixes_match {
+
```
linabutler, line 362: Don't forget to fix the indentation before landing 😊

components/suggest/src/db.rs
linabutler: Oof, yes, the `f64` sure makes sorting cumbersome. I really like @ncloudioj's suggestion to implement `Ord` for `Suggestion`, though—then we can lift that complexity, and the default score for Wikipedia, out of here, and then use `suggestions.sort()`.

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
-
-        let keyword_lowercased = &query.keyword.to_lowercase();
-        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
         let suggestions_limit = query.limit.unwrap_or(-1);
+        let suggestions =
+            query
+                .providers
+                .clone()
+                .into_iter()
+                .try_fold(vec![], |mut acc, provider| {
```
linabutler, line 137: Nice use of `try_fold`! 🏆 Those nested `Result<Vec<Suggestion>>` really make working with them awkward, but this looks great.

There's a Rust crate inspired by Python's called [`itertools`](https://docs.rs/itertools/latest/itertools/), and it has a [`flatten_ok`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.flatten_ok) helper for this exact case. It looks like `nimbus-fml` and `glean` already use `itertools`, so we could use it here if we wanted—but it doesn't seem worth it just for one method, when your approach is already clean and readable.

components/suggest/src/db.rs
linabutler: Yeah, I like the idea of renaming `fetch_yelp_suggestion` to `fetch_yelp_suggestions`, and having it return a `Vec` instead of a single `Option`.

linabutler (commented):
This is definitely on the right track, great job!

components/suggest/src/db.rs
```diff
@@ -128,63 +128,72 @@ impl<'a> SuggestDao<'a> {
...
-        }
-
-        let keyword_lowercased = &query.keyword.to_lowercase();
-        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
         let suggestions_limit = query.limit.unwrap_or(-1);
+        let suggestions =
+            query
+                .providers
+                .clone()
+                .into_iter()
```
linabutler, line 136: I know you're going to be changing this anyway, using @ncloudioj's `unique_providers` suggestion, but it doesn't look like you need `.clone()` here:

```suggestion
                .providers
                .iter()
```

components/suggest/src/db.rs
```diff
@@ -128,63 +128,49 @@ impl<'a> SuggestDao<'a> {
...
+                    SuggestionProvider::Pocket => {
+                        self.fetch_pocket_suggestions(query, suggestions_limit)
+                    }
+                    SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
+                }?;
+                acc.extend(suggestions);
+                Ok(acc)
+            });
+
+        suggestions.map(|mut suggestions| {
```
ncloudioj, line 149: `suggestions.and_then(|suggestions| {..})` feels more natural to me, what do you think? Oh, also, i think you could attach `and_then` right after `try_fold` above :P

components/suggest/src/db.rs
```diff
@@ -128,63 +128,49 @@ impl<'a> SuggestDao<'a> {
...
+            suggestions.sort();
+            suggestions.truncate(suggestions_limit as usize);
```
ncloudioj, line 151: Doh, looks like we set `suggestion_limit` to -1 by default, which would be problematic since truncate expects a usize. Maybe we should do `let suggestion_limit = query.limit.unwrap_or(i32::MAX)` 😀

ncloudioj (approved):
r+wc. Thanks!

components/suggest/src/db.rs
tiftran: fancy

components/suggest/src/db.rs
```diff
@@ -197,58 +183,119 @@ impl<'a> SuggestDao<'a> {
...
+                ":suggestions_limit": suggestions_limit,
+            },
+            |row| -> Result<Option<Suggestion>>{
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let raw_url = row.get::<_, String>("url")?;
+                let score = row.get::<_, f64>("score")?;
+
+                let full_suffix = row.get::<_, String>("keyword_suffix")?;
+                    if !full_suffix.starts_with(keyword_suffix){
```
ncloudioj, line 294: It looks like misaligned also there is a missing whitespace `suffix) {`. I opted for `bool::then()` [here](https://mozilla.github.io/application-services/book/howtos/branch-builds.html#breaking-changes-in-an-application-services-branch) to cut an if-else nesting, it's up to you though :)

components/suggest/src/db.rs
```diff
@@ -128,63 +128,48 @@ impl<'a> SuggestDao<'a> {
...
-            return Ok(vec![suggestion]);
-        }
-
-        let keyword_lowercased = &query.keyword.to_lowercase();
-        let (keyword_prefix, keyword_suffix) = split_keyword(keyword_lowercased);
-        let suggestions_limit = query.limit.unwrap_or(-1);
-
-        let (mut statement, params) = if query
-            .providers
+        let suggestions_limit = query.limit.unwrap_or(i32::MAX);
```
linabutler, line 131: Actually, I don't think you need to unwrap here—you can pass along `query` as-is to all the `fetch_*_suggestions()` functions, unwrap-or-default in those, and use an `if-let` when you truncate.

```suggestion
```

components/suggest/src/db.rs
```diff
@@ -128,63 +128,48 @@ impl<'a> SuggestDao<'a> {
...
+                        self.fetch_pocket_suggestions(query, suggestions_limit)
+                    }
+                    SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
+                }?;
+                acc.extend(suggestions);
+                Ok(acc)
+            })
+            .and_then(|mut suggestions| {
+                suggestions.sort();
+                suggestions.truncate(suggestions_limit as usize);
```
linabutler, line 150:
```suggestion
                if let Some(limit) = query.limit.and_then(|limit| usize::try_from(limit).ok()) {
                    suggestions.truncate(limit);
                }
```

components/suggest/src/db.rs
```diff
@@ -197,108 +182,185 @@ impl<'a> SuggestDao<'a> {
...
+        Ok(suggestions)
+    }
+
+    /// Fetches Suggestions of type Amo provider that match the given query
+    pub fn fetch_amo_suggestions(
+        &self,
+        query: &SuggestionQuery,
+        suggestions_limit: i32,
+    ) -> Result<Vec<Suggestion>> {
+        let keyword_lowercased = &query.keyword.to_lowercase();
```
linabutler, line 273: Let's do the `suggestions_limit = query.limit.unwrap_or(-1)` in these functions—`query` already has the limit; it's a little awkward to pass it and the unwrapped limit as arguments.

components/suggest/src/db.rs
tiftran: lol looks like cargo clippy doesn't like it though 
```
error: using `Result.and_then(|x| Ok(y))`, which is more succinctly expressed as `map(|x| y)`
   --> components/suggest/src/db.rs:133:9
    |
133 | /         unique_providers
134 | |             .iter()
135 | |             .try_fold(vec![], |mut acc, provider| {
136 | |                 let suggestions = match provider {
...   |
152 | |                 Ok(suggestions)
153 | |             })
    | |______________^
```

components/suggest/src/db.rs
```diff
@@ -197,108 +182,185 @@ impl<'a> SuggestDao<'a> {
...
+                            title,
+                            url: raw_url,
+                            icon_url: row.get("icon_url")?,
+                            description: row.get("description")?,
+                            rating: row.get("rating")?,
+                            number_of_ratings: row.get("number_of_ratings")?,
+                            guid: row.get("guid")?,
+                            score,
+                        }))
+                    })).unwrap_or(Ok(None))
```
ncloudioj, line 313:
```suggestion
                        Ok(Suggestion::Amo {
                            title,
                            url: raw_url,
                            icon_url: row.get("icon_url")?,
                            description: row.get("description")?,
                            rating: row.get("rating")?,
                            number_of_ratings: row.get("number_of_ratings")?,
                            guid: row.get("guid")?,
                            score,
                        })
                    })).transpose()
```

components/suggest/src/db.rs
ncloudioj: clippy strikes. ok, fine, i will :🤐

components/suggest/src/db.rs
tiftran: it's always the space, sorry 🤦‍♀️

components/suggest/src/db.rs
linabutler: Ooooh, I like the `.then(|| { ... }).transponse()`! ✨

components/suggest/src/suggestion.rs
```diff
@@ -59,6 +61,33 @@ pub enum Suggestion {
...
+            | Suggestion::Amo { score, .. } => score,
+            _ => &DEFAULT_SUGGESTION_SCORE,
+        };
+        b_score
+            .partial_cmp(a_score)
+            .unwrap_or(std::cmp::Ordering::Equal)
+    }
+}
+
+impl Eq for Suggestion {}
```
linabutler, line 90: Floaaaaats! 😱

linabutler (approved):
Looks great! Thank you very much! :shipit:

PR #6086, opened 2024-01-30
Bug 1876211 - Add MDN for Suggest [firefox-android: main]

ncloudioj:

  - Note:

First stab, will need to rebase against #6085. Also noticed a few duplicate code paths that require more refactors perhaps in this patch.

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -14,7 +14,7 @@ use rusqlite::{
 };
 use sql_support::{open_database::open_database_with_flags, ConnExt};
 
-use crate::rs::{DownloadedAmoSuggestion, DownloadedPocketSuggestion};
+use crate::rs::{DownloadedAmoSuggestion, DownloadedMdnSuggestion, DownloadedPocketSuggestion};
 use crate::{
     keyword::full_keyword,
     pocket::{split_keyword, KeywordConfidence},
@@ -138,6 +138,7 @@ impl<'a> SuggestDao<'a> {
                     SuggestionProvider::Amo => self.fetch_amo_suggestions(query),
                     SuggestionProvider::Pocket => self.fetch_pocket_suggestions(query),
                     SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
+                    SuggestionProvider::Mdn => self.fetch_mdn_suggestions(query),
                 }?;
                 acc.extend(suggestions);
                 Ok(acc)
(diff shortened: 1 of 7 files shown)


ncloudioj:
@0c0w3 wanted to confirm with you that the current implementation only returns the raw data from RS, do we want to process the data here? (e.g. add utm params; add the hardcoded icon URI, etc.)

0c0w3 (approved):
Thanks @ncloudioj, this looks great! Sorry for the delay, I would have r+'ed this earlier today but I wanted to do an end-to-end test in Firefox and that took some time. It works great.

Yes, the current impl only reflects the data in RS. The desktop urlbar JS handles UTM params and the icon. (Maybe they should be handled by Rust too but that's a future discussion.)

ncloudioj:
Thanks for confirming that, Drew!

components/suggest/src/db.rs
```diff
@@ -294,6 +295,33 @@ impl<'a> SuggestDao<'a> {
...
+                             FROM
+                                 mdn_custom_details
+                             WHERE
+                                 suggestion_id = :suggestion_id
+                            "#,
+                            named_params! {
+                                ":suggestion_id": suggestion_id
+                            },
+                            |row| {
+                                if full_suffix.starts_with(keyword_suffix) {
```
linabutler, line 313: I wonder if we can do this check before we do the `query_row_and_then`? (I think we could do the same for AMO, but that can be a different PR; no need to do that here 😊) No need to query `mdn_custom_details` if the suffixes don't match.

components/suggest/src/db.rs
```diff
@@ -583,6 +611,84 @@ impl<'a> SuggestDao<'a> {
...
+                named_params! {
+                    ":record_id": record_id.as_str(),
+                    ":title": suggestion.title,
+                    ":url": suggestion.url,
+                    ":score": suggestion.score,
+                },
+                |row| row.get(0),
+                true,
+            )?;
+            self.conn.execute(
```
linabutler, line 651:
```suggestion
            self.conn.execute_cached(
```

linabutler (approved):
Thanks @ncloudioj!

Please be sure to trigger a [branch build](https://mozilla.github.io/application-services/book/howtos/branch-builds.html) for iOS and Android. Adding new suggestion types should be a backward-compatible change (I'm realizing it would be super helpful to document what's backward-compatible, and what's a breaking change 😅), but let's make sure there's no unexpected breakage.

components/suggest/src/db.rs
ncloudioj: Yup, nice catch! Let's do that.

components/suggest/src/db.rs
ncloudioj: Doh, copy&pasta 🙈, will update both.

components/suggest/src/store.rs
```diff
@@ -2865,6 +2897,89 @@ mod tests {
                 []
                 "#]],
             ),
+            (
+                "keyword = prefix; MDN only",
```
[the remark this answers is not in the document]
ncloudioj: Absolutely, will do.

ncloudioj:
Hey @linabutler, are you suggesting to follow this [particular step](https://mozilla.github.io/application-services/book/howtos/branch-builds.html#breaking-changes-in-an-application-services-branch) (i.e. for breaking changes) for this patch? If so, would adding `[firefox-android: 1876211-suggest-mdn]` to the PR title be sufficient?

mhammond:
The branch-name in `[firefox-android: branch-name]` is a branch on the firefox-android repo. eg, if you knew this was a breaking change, you'd make a branch on that repo with the fixes and specify that name in this PR (I don't think it allows branches in forks, which is a shame). In this scenario where you assume there are no breaking changes, specifying `[firefox-android: main]` is probably what you want

ncloudioj:
Gotcha! Thanks for the pointers, Mark!

PR #6088, opened 2024-01-31
Suggest - switch to using `error_support`

bendk:
Start using `error_support` for internal -> external error conversion.

Added SuggestApiError variants for network errors and interrupted operations.

  - Note:

### DIFF
--- a/components/suggest/src/error.rs
+++ b/components/suggest/src/error.rs
@@ -3,6 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+use error_support::{ErrorHandling, GetErrorHandling};
+use remote_settings::RemoteSettingsError;
+
 /// A list of errors that are internal to the component. This is the error
 /// type for private and crate-internal methods, and is never returned to the
 /// application.
@@ -18,7 +21,7 @@ pub(crate) enum Error {
     Json(#[from] serde_json::Error),
 
     #[error("Error from Remote Settings: {0}")]
-    RemoteSettings(#[from] remote_settings::RemoteSettingsError),
+    RemoteSettings(#[from] RemoteSettingsError),
 
     #[error("Operation interrupted")]
     Interrupted(#[from] interrupt_support::Interrupted),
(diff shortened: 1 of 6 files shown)


components/suggest/src/error.rs
```diff
@@ -29,15 +32,35 @@ pub(crate) enum Error {
...
+// See `support/error/README.md` for how this works, especially the warning about PII.
+impl GetErrorHandling for Error {
+    type ExternalError = SuggestApiError;
+
+    fn get_error_handling(&self) -> ErrorHandling<Self::ExternalError> {
+        match self {
+            // Do nothing for interrupted errors, this is just normal operation.
+            Self::Interrupted(_) => ErrorHandling::convert(SuggestApiError::Interrupted),
+            // Network errors are fairly common so let's log, but not report them.
+            Self::RemoteSettings(RemoteSettingsError::RequestError(_)) => {
```
bendk, line 54: From what I could tell from Sentry, this was the overwhelming source of network errors.  However, are there other potential sources?

components/suggest/src/store.rs
```diff
@@ -130,13 +134,15 @@ impl SuggestStore {
...
 
     /// Ingests new suggestions from Remote Settings.
+    #[handle_error(Error)]
     pub fn ingest(&self, constraints: SuggestIngestionConstraints) -> SuggestApiResult<()> {
-        Ok(self.inner.ingest(constraints)?)
+        self.inner.ingest(constraints)
     }
 
     /// Removes all content from the database.
+    #[handle_error(Error)]
```
bendk, line 143: The main drawback to the `error_support` crate is this macro, which is slightly magical since it not only converts errors but also potentially reports them.  Overall it hasn't been that bad to work with in app-services though.  On the plus side, it removes some boilerplate in the inner code.

Another option would be manually doing running the conversion.  I think that would mean adding something like `.map_err(::error_support::convert_log_report_error)` at the end of all returns.

components/suggest/src/error.rs
```diff
@@ -29,15 +32,35 @@ pub(crate) enum Error {
 #[derive(Debug, thiserror::Error)]
 #[non_exhaustive]
 pub enum SuggestApiError {
+    #[error("Network error: {reason}")]
+    NetworkError { reason: String },
```
linabutler, line 36:
```suggestion
    Network { reason: String },
```

Question: Should we adopt the `Error` suffix for our error variants, or drop it?

I noticed some of our crates have it, and other don't, so I went with "no suffix", on the grounds that `SuggestApiError.OtherError` felt like one `Error` too many. But as long as we're consistent, either one sounds good to me!

components/suggest/src/error.rs
linabutler: I wonder if we should also treat `ResponseError` (and maybe `BackoffError`, too?) as a network error...a 503 or a bad response from Remote Settings could be interesting to note, but doesn't feel super actionable as far as there being something for us to fix in the client. WDYT?

components/suggest/src/store.rs
linabutler: I like it! (And I love the documentation you added for it ✨)

The reduced boilerplate is great, and I think it actually calls out "this call is special, its errors get reported" better than a `.map_err(...)` at each return would.

components/suggest/src/error.rs
```diff
@@ -29,15 +32,35 @@ pub(crate) enum Error {
...
+    // The application interrupted a request
+    #[error("Interrupted")]
+    Interrupted,
     #[error("Other error: {reason}")]
     Other { reason: String },
```
linabutler, line 41: If we go with `NetworkError`, let's maybe rename `Other` to `OtherError` or `UnexpectedError`, too?

linabutler (approved):
Lovely, thanks!

components/suggest/src/error.rs
bendk: I think that last part is the most important: what actions should the client take?  With that in mind I updated the handling:
- `SuggestApiError::Network` now only happens when we see `viaduct::NetworkError`.  If a client sees a network error, they should retry when the network comes back online.  I targetted this very specifically at `viaduct::NetworkError`, because AFAIK, there's no reason to believe that the other errors are transient in the same way -- but please correct me if I'm wrong about this.
- I added `SuggestApiError::Backoff` which corresponds to `remove_settings::BackoffError`, which implies a very specific action for the client.
- All other networky errors get mapped to `SuggestApiError::Other`.  My reasoning is the client should just treat these as something bad happened and try to recover as best as they can, without any expectation that sending the same request again will work.

What do you think about that?  This is mostly based on my gut feelings, so I'm happy to take another pass and update the logic again.

components/suggest/src/error.rs
bendk: I ended up dropping the trailing `Error` from all variants.  I always wonder what the best way is, but at least this way is consistent.

mhammond: my 2c is that the redundant trailing `Error` is redundant :)

linabutler: Sold! 😆

components/suggest/src/error.rs
linabutler: That sounds good to me!

> there's no reason to believe that the other errors are transient in the same way

We've seen Remote Settings return 503s occasionally, so the client could retry those, too—but let's not worry about them now; Viaduct network errors are so noisy by comparison. We can always revisit reporting these HTTP errors if they show up frequently in Sentry.

linabutler (approved):
Still LGTM ☺️

PR #6089, opened 2024-02-01
Bug 1877300 - Add weather to Suggest

0c0w3:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -14,12 +14,14 @@ use rusqlite::{
 };
 use sql_support::{open_database::open_database_with_flags, ConnExt};
 
-use crate::rs::{DownloadedAmoSuggestion, DownloadedMdnSuggestion, DownloadedPocketSuggestion};
 use crate::{
     keyword::full_keyword,
     pocket::{split_keyword, KeywordConfidence},
     provider::SuggestionProvider,
-    rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId},
+    rs::{
+        DownloadedAmoSuggestion, DownloadedAmpWikipediaSuggestion, DownloadedMdnSuggestion,
+        DownloadedPocketSuggestion, DownloadedWeatherData, SuggestRecordId,
+    },
     schema::{SuggestConnectionInitializer, VERSION},
     store::{UnparsableRecord, UnparsableRecords},
     suggestion::{cook_raw_suggestion_url, Suggestion},
@@ -139,6 +141,7 @@ impl<'a> SuggestDao<'a> {
                     SuggestionProvider::Pocket => self.fetch_pocket_suggestions(query),
                     SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
                     SuggestionProvider::Mdn => self.fetch_mdn_suggestions(query),
+                    SuggestionProvider::Weather => self.fetch_weather_suggestions(query),
                 }?;
                 acc.extend(suggestions);
                 Ok(acc)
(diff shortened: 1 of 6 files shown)


0c0w3:
This doesn't match the [current weather record](https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/quicksuggest/records/weather) in RS, which looks like:

```json
{
  "weather": {
    "keywords": [],
    "min_keyword_length": 3,
    "min_keyword_length_cap": 6
  }
}
```

Instead this patch expects:

```json
{
  "weather": {
    "keywords": [],
    "score": 0.2,
  }
}
```

I'm trying to separate suggestion-related data from config data. I'll handle config in bug 1877501. `min_keyword_length` is specific to weather suggestions, so I'm thinking we can store that using @linabutler's idea of per-provider config. `min_keyword_length_cap` is an older version of the `show_less_frequently_cap` in the [global config](https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/quicksuggest/records/firefox-suggest-configuration) and so is redundant, so I'd like to drop it altogether.

0c0w3:
> `min_keyword_length` is specific to weather suggestions, so I'm thinking we can store that using @linabutler's idea of per-provider config.

Or simpler, just include all desired prefix keywords in `keywords`. There aren't many weather keywords so that doesn't seem too bad.

0c0w3:
Hmm after thinking about it more, maybe the weather record should include `min_keyword_length`. If we had to generate all keyword prefixes and add them to `keywords`, then the code for doing that would have to live somewhere, and in that case why not just do it in the suggest component itself. Another thing: The user can change the min keyword length by clicking "Show less frequently", so the `min_keyword_length` config value is just the initial value. Maybe in the future the Rust component could store the user value too but I don't want to worry about that right now -- desktop can keep storing it in a pref and compare it to the user's search string length at query time.

Also, I was thinking of putting provider-specific config inside the existing `firefox-suggest-configuration` record, but maybe it's better for it to live inside each provider's dataset, where it will be closer to the rest of the provider data. If a provider's data is split among multiple records, it could live in the first record.

ncloudioj (approved):
Looks great, thanks!

0c0w3:
I just pushed these changes:

* While testing this I found RS can't store floats in record JSON, so use a stringified float for the score and add a `serde(deserialize_with)` attribute for it
* Do prefix matching starting at the 3rd char
* Trim the query string

components/suggest/src/rs.rs
```diff
@@ -301,3 +303,26 @@ pub(crate) struct DownloadedMdnSuggestion {
...
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct DownloadedWeatherData {
+    pub weather: DownloadedWeatherDataInner,
+}
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct DownloadedWeatherDataInner {
+    pub keywords: Vec<String>,
+    // Remote settings doesn't support floats in record JSON so we use a
+    // stringified float instead. If a float can't be parsed, this will be None.
+    #[serde(deserialize_with = "de_stringified_f64")]
```
ncloudioj, line 317: `#[serde(default)]` is required with the custom deserializer. Otherwise, it will error out when `score` is missing from the JSON payload.
```suggestion
    #[serde(default)]
    #[serde(deserialize_with = "de_stringified_f64")]
```

components/suggest/src/rs.rs
```diff
@@ -301,3 +303,26 @@ pub(crate) struct DownloadedMdnSuggestion {
...
+    pub score: Option<f64>,
+}
+
+fn de_stringified_f64<'de, D>(deserializer: D) -> std::result::Result<Option<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    use std::str::FromStr;
+    let s = String::deserialize(deserializer)?;
+    Ok(f64::from_str(&s).ok())
```
ncloudioj, line 327:
```suggestion
    String::deserialize(deserializer).map(|s| s.parse().ok())
```

ncloudioj (approved):
r+wc. Thanks!

0c0w3:
Thanks, latest push has these changes:

* Rebase on attachment-ingestion changes from #6096 that were merged to main a few hours ago
* Address @ncloudioj's comments

components/suggest/src/db.rs
```diff
@@ -14,12 +14,14 @@ use rusqlite::{
 };
 use sql_support::{open_database::open_database_with_flags, ConnExt};
 
-use crate::rs::{DownloadedAmoSuggestion, DownloadedMdnSuggestion, DownloadedPocketSuggestion};
 use crate::{
     keyword::full_keyword,
     pocket::{split_keyword, KeywordConfidence},
     provider::SuggestionProvider,
-    rs::{DownloadedAmpWikipediaSuggestion, SuggestRecordId},
+    rs::{
```
linabutler, line 21: Thanks for cleaning up these imports!

linabutler (approved):
Looks excellent! 🚢 Thank you for the nice refactoring to `ingest_suggestions_from_record`!

PR #6091, opened 2024-02-01
Applying history tombstones checks foreign count

tarikeshaq:
Possible fix to https://bugzilla.mozilla.org/show_bug.cgi?id=1866364

When applying a tombstone, we first delete all the visits, then only delete the places entry if no more bookmarks and keywords are referencing it

  - Note:

### DIFF
--- a/components/places/src/storage/history.rs
+++ b/components/places/src/storage/history.rs
@@ -1019,8 +1019,20 @@ pub mod history_sync {
     }
 
     pub fn apply_synced_deletion(db: &PlacesDb, guid: &SyncGuid) -> Result<()> {
+        // First we delete any visits for the page
+        // because it's possible the moz_places foreign_count is not 0
+        // and thus the moz_places entry won't be deleted.
         db.execute_cached(
-            "DELETE FROM moz_places WHERE guid = :guid",
+            "DELETE FROM moz_historyvisits
+              WHERE place_id IN (
+                  SELECT id
+                  FROM moz_places
+                  WHERE guid = :guid
+              )",
+            &[(":guid", guid)],
+        )?;
+        db.execute_cached(
+            "DELETE FROM moz_places WHERE guid = :guid AND foreign_count = 0",
             &[(":guid", guid)],
         )?;
         Ok(())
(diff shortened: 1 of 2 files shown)


components/places/src/storage/history.rs
```diff
@@ -1019,8 +1019,20 @@ pub mod history_sync {
...
-            "DELETE FROM moz_places WHERE guid = :guid",
+            "DELETE FROM moz_historyvisits
+              WHERE place_id IN (
+                  SELECT id
+                  FROM moz_places
+                  WHERE guid = :guid
+              )",
+            &[(":guid", guid)],
+        )?;
+        db.execute_cached(
```
tarikeshaq, line 1034: I verified that if we had bookmarks referencing the place_id and and applied a tombstone, we do get a `FOREIGN KEY constraint failed` error back from sqlite

components/places/src/storage/history.rs
```diff
@@ -1019,8 +1019,20 @@ pub mod history_sync {
...
+        // First we delete any visits for the page
+        // because it's possible the moz_places foreign_count is not 0
+        // and thus the moz_places entry won't be deleted.
         db.execute_cached(
-            "DELETE FROM moz_places WHERE guid = :guid",
+            "DELETE FROM moz_historyvisits
+              WHERE place_id IN (
+                  SELECT id
+                  FROM moz_places
+                  WHERE guid = :guid
```
tarikeshaq, line 1030: I thought of refactoring the guid scan here so we do it only once.. but we have unique indexes on the guid so that wouldn't have made too much of a difference in practice (but I could be corrected!)

linabutler: I think SQLite's optimizer will do the right thing here! `(SELECT id FROM moz_places WHERE guid = :guid)` isn't a correlated subquery, so it should be run just once to build the `IN` list—which, thanks to the unique constraint on `moz_places.guid`, will contain just one entry at most—and then use the `placedateindex` on `moz_historyvisits` to search and delete all matching rows.

components/places/src/storage/history.rs
linabutler: Nice, thanks for checking!

linabutler (approved):
Thank you so much!

PR #6092, opened 2024-02-01
Implement prefix-match for Yelp suggestion

dadaa:
This PR is for https://bugzilla.mozilla.org/show_bug.cgi?id=1877873.

  - Note:

### DIFF
--- a/components/suggest/src/suggest.udl
+++ b/components/suggest/src/suggest.udl
@@ -72,7 +72,8 @@ interface Suggestion {
     );
     Yelp(
         string url,
-        string title
+        string title,
+        boolean is_top_pick
     );
     Mdn(
         string title,
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -47,6 +47,9 @@ const MAX_QUERY_LENGTH: usize = 150;
 /// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
 const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
 
+/// The threshold that enables prefix-match.
+const PREFIX_MATCH_THRESHOLD: usize = 6;
+
 impl<'a> SuggestDao<'a> {
     /// Inserts the suggestions for Yelp attachment into the database.
     pub fn insert_yelp_suggestions(
@@ -128,24 +131,25 @@ impl<'a> SuggestDao<'a> {
 
         let query_string = &query.keyword.trim();
         if !query_string.contains(' ') {
-            if !self.is_subject(query_string)? {
+            let Some((subject, subject_exact_match)) = self.find_subject(query_string)? else {
                 return Ok(vec![]);
-            }
-
+            };
             let builder = SuggestionBuilder {
-                query,
-                subject: query_string,
+                subject: &subject,
+                subject_exact_match,
                 pre_modifier: None,
                 post_modifier: None,
                 location_sign: None,
                 location: None,
                 need_location: false,
+                pre_yelp_modifier: None,
+                post_yelp_modifier: None,
             };
             return Ok(vec![builder.into()]);
         }
 
         // Find the yelp keyword modifier and remove them from the query.
-        let (query_without_yelp_modifiers, _, _) =
+        let (query_without_yelp_modifiers, pre_yelp_modifier, post_yelp_modifier) =
             self.find_modifiers(query_string, Modifier::Yelp, Modifier::Yelp)?;
 
         // Find the location sign and the location.
@@ -166,18 +170,19 @@ impl<'a> SuggestDao<'a> {
         let (subject_candidate, pre_modifier, post_modifier) =
             self.find_modifiers(&query_without_location, Modifier::Pre, Modifier::Post)?;
 
-        if !self.is_subject(&subject_candidate)? {
+        let Some((subject, subject_exact_match)) = self.find_subject(&subject_candidate)? else {
             return Ok(vec![]);
-        }
-
+        };
         let builder = SuggestionBuilder {
-            query,
-            subject: &subject_candidate,
+            subject: &subject,
+            subject_exact_match,
             pre_modifier,
             post_modifier,
             location_sign,
             location,
             need_location,
+            pre_yelp_modifier,
+            post_yelp_modifier,
         };
         Ok(vec![builder.into()])
     }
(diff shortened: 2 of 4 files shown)


dadaa:
Hello! I implemented for https://bugzilla.mozilla.org/show_bug.cgi?id=1877873.
Could you review this when you have time?
Thanks!

components/suggest/src/yelp.rs
```diff
@@ -337,27 +376,29 @@ impl<'a> SuggestDao<'a> {
...
+    pre_yelp_modifier: Option<String>,
+    post_yelp_modifier: Option<String>,
 }
 
 impl<'a> From<SuggestionBuilder<'a>> for Suggestion {
     fn from(builder: SuggestionBuilder<'a>) -> Suggestion {
         // This location sign such the 'near by' needs to add as a description parameter.
         let location_modifier = if !builder.need_location {
-            builder.location_sign
+            builder.location_sign.clone()
```
ncloudioj, line 394: We can save a couple of clones by this:

```rs
        let location_modifier = if !builder.need_location {
            builder.location_sign.as_deref()
        } else {
            None
        };
        let description = [
            builder.pre_modifier.as_deref(),
            Some(builder.subject),
            builder.post_modifier.as_deref(),
            location_modifier,
        ]
        .iter()
        .flatten()
        .map(|&s| s)
        .collect::<Vec<_>>()
        .join(" ");
```

components/suggest/src/yelp.rs
```diff
@@ -375,10 +416,25 @@ impl<'a> From<SuggestionBuilder<'a>> for Suggestion {
         }
         url.push_str(&parameters.finish());
 
+        let title = [
```
ncloudioj, line 419: Same here:

```rs
        let title = [
            builder.pre_yelp_modifier.as_deref(),
            builder.pre_modifier.as_deref(),
            Some(builder.subject),
            builder.post_modifier.as_deref(),
            builder.location_sign.as_deref(),
            builder.location.as_deref(),
            builder.post_yelp_modifier.as_deref(),
        ]
        .iter()
        .flatten()
        .map(|&s| s)
        .collect::<Vec<_>>()
        .join(" ");
```

ncloudioj (approved):
Looks good to me, just a few nits.

components/suggest/src/yelp.rs
```diff
@@ -296,6 +301,44 @@ impl<'a> SuggestDao<'a> {
...
+                Some((candidate.to_string(), true))
+            } else {
+                None
+            });
+        }
+
+        // Otherwise, apply prefix-match.
+        Ok(
+            match self.conn.query_row_and_then_cachable(
+                "SELECT keyword FROM yelp_subjects WHERE keyword LIKE :candidate || '%' ORDER BY LENGTH(keyword) LIMIT 1",
```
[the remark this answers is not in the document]
0c0w3: Shortest makes sense to me! We don't have product direction either way. @dadaa could you add an explicit `ASC` to the `ORDER BY` just so it's clear?

0c0w3 (approved):
Thanks, lgtm! How do you think the client could use `subject_exact_match`? Would it do something differently if it's true instead of false?

components/suggest/src/yelp.rs
```diff
@@ -296,6 +301,44 @@ impl<'a> SuggestDao<'a> {
...
+                named_params! {
+                    ":candidate": candidate.to_lowercase(),
+                },
+                |row| row.get::<_, String>(0),
+                true,
+            ) {
+                Ok(keyword) if candidate.len() == keyword.len() => Some((candidate.to_string(), true)),
+                Ok(keyword) =>  Some((candidate.to_owned() + keyword.get(candidate.len()..).unwrap(), false)),
```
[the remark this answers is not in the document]
dadaa: > I assume we want to keep the casing for the original string, correct?
Yes, right!

> Let's add a debug assert for the invariant keyword.len() >= candidate.len() so we can slice keyword with peace of mind.
Ah, cool. Thank you so much!

components/suggest/src/yelp.rs
dadaa: Oh, cool.. I didn't know the `as_deref()`.

dadaa:
> Thanks, lgtm! How do you think the client could use `subject_exact_match`? Would it do something differently if it's true instead of false?

Hi @0c0w3, thanks!
Yes, I want to change how to show dependent on the flag. If `subject_exact_match` is true, the suggestion will be shown as `top_pick`, otherwise normal.

components/suggest/src/yelp.rs
dadaa: A lint error happens from `rust-clippy`.
https://app.circleci.com/pipelines/github/mozilla/application-services/36787/workflows/52aeceb3-9d98-4fe4-8037-ffccae4cd57e/jobs/186366?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary
So, as it said, I use `.flatten().copied()`.

ncloudioj: Ah, of course, thanks Clippy!

0c0w3:
> Hi @0c0w3, thanks! Yes, I want to change how to show dependent on the flag. If `subject_exact_match` is true, the suggestion will be shown as `top_pick`, otherwise normal.

I see, that makes sense. Loren changed plans today: There will be one experiment branch where Yelp is always shown as a top pick and another branch where it's always shown in the usual Firefox Suggest section, similar to the higher-placement sponsored experiment. So we won't need `subject_exact_match` immediately but it may be useful in the future and let's definitely keep it as part of your patch.

components/suggest/src/yelp.rs
```diff
@@ -296,6 +301,46 @@ impl<'a> SuggestDao<'a> {
...
+                Some((candidate.to_string(), true))
+            } else {
+                None
+            });
+        }
+
+        // Otherwise, apply prefix-match.
+        Ok(
+            match self.conn.query_row_and_then_cachable(
+                "SELECT keyword FROM yelp_subjects WHERE keyword LIKE :candidate || '%' ORDER BY LENGTH(keyword) ASC LIMIT 1",
```
linabutler, line 328: Small suggestion: Using `BETWEEN` instead of `LIKE` will let SQLite take advantage of the automatic index on the primary key, instead of doing a full table scan:

```suggestion
                "SELECT keyword FROM yelp_subjects WHERE keyword BETWEEN :candidate AND :candidate || x'FFFF' ORDER BY LENGTH(keyword) ASC LIMIT 1",
```

The optimizer [can optimize `x%` `LIKE` patterns to be prefix matches](https://www.sqlite.org/optoverview.html#the_like_optimization), but won't by default—because `TEXT` columns are case-sensitive unless they're explicitly declared `COLLATE NOCASE`, but `LIKE` is case-insensitive by default.

The `ORDER BY LENGTH(keyword)` will still need to create a temporary sorting B-Tree, but it should only have a small handful of entries.

components/suggest/src/suggest.udl
```diff
@@ -72,7 +72,8 @@ interface Suggestion {
     );
     Yelp(
         string url,
-        string title
+        string title,
+        boolean subject_exact_match
```
linabutler, line 76: Thought / question: For Pocket, we have a `boolean is_top_pick` field, and decide in the component what that means (e.g., if the keyword is a high-confidence keyword). Do you think it makes sense to do the same for Yelp?

You mentioned that we use `subject_exact_match` to decide whether to use the Top Pick treatment or not, so I'm wondering, does it make sense to call this `is_top_pick`, and set it to `true` / `false` in `yelp.rs` based on any criteria we might have now or in the future.

linabutler (approved):
r+wc! 😁

components/suggest/src/yelp.rs
dadaa: Oh, I see, thanks!

components/suggest/src/suggest.udl
dadaa: Yeah, as you said, this means the same meaning for now. The reason I named this was that I thought it would be better to return only the facts as much as possible here and let the client interpret them. But, for consistency, better to use `is_top_pick`, as you said. Okay, I will rename it with `is_top_pick`. Thanks!

components/suggest/src/yelp.rs
ncloudioj: Nice catch, @linabutler! TIL💡

dadaa:
@0c0w3 @ncloudioj @linabutler

dadaa:
> @0c0w3 @ncloudioj @linabutler Do you know this error? https://app.circleci.com/pipelines/github/mozilla/application-services/36820/workflows/84ae8144-7b51-4eaa-aaa5-bbc884d746cb/jobs/186641?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

It seems that, even if we run test `./automation/tests.py rust-min-version-tests` locally without my patch, got test failed.

dadaa:
I will try to rebase and push it again.

dadaa:
Oh, I'm not sure, but there is no issue in this time.. I will merge now. Thanks!

PR #6095, opened 2024-02-01
Add Suggest Ingestion Constaint

mtbottle:
This adds a suggestion constaint that gets passed to the Remote Settings Client. We iterate over all the suggestion record type.

Tests pass, but they aren't actually working. I will need to fix tests because the snapshots are returning the same snapshot multiple times (incorrectly).

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -309,7 +309,7 @@ struct AttachmentsCapability {
 /// Options for requests to endpoints that return multiple items.
 #[derive(Clone, Debug, Default)]
 pub struct GetItemsOptions {
-    filters: Vec<Filter>,
+    pub filters: Vec<Filter>,
     sort: Vec<Sort>,
     fields: Vec<String>,
     limit: Option<u64>,
@@ -463,7 +463,7 @@ pub enum SortOrder {
 }
 
 #[derive(Clone, Debug)]
-enum Filter {
+pub enum Filter {
     Eq(String, String),
     Not(String, String),
     Contains(String, String),
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -309,7 +309,7 @@ struct AttachmentsCapability {
 /// Options for requests to endpoints that return multiple items.
 #[derive(Clone, Debug, Default)]
 pub struct GetItemsOptions {
-    pub filters: Vec<Filter>,
+    filters: Vec<Filter>,
     sort: Vec<Sort>,
     fields: Vec<String>,
     limit: Option<u64>,
@@ -463,7 +463,7 @@ pub enum SortOrder {
 }
 
 #[derive(Clone, Debug)]
-pub enum Filter {
+enum Filter {
     Eq(String, String),
     Not(String, String),
     Contains(String, String),
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -31,9 +31,9 @@ use crate::{
     Result, SuggestionQuery,
 };
 
-/// The metadata key whose value is the timestamp of the last record ingested
-/// from the Suggest Remote Settings collection.
-pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
+/// The metadata key prefix for the last ingested unparsable record. These are
+/// records that were not parsed properly, or were not of the "approved" types.
+pub const LAST_INGEST_META_UNPARSABLE: &str = "last_quicksuggest_ingest_unparsable";
 /// The metadata key whose value keeps track of records of suggestions
 /// that aren't parsable and which schema version it was first seen in.
 pub const UNPARSABLE_RECORDS_META_KEY: &str = "unparsable_records";
@@ -148,20 +148,28 @@ impl<'a> SuggestDao<'a> {
         self.put_unparsable_record_id(&record_id)?;
         // Advance the last fetch time, so that we can resume
         // fetching after this record if we're interrupted.
-        self.put_last_ingest_if_newer(record.last_modified)
+        self.put_last_ingest_if_newer(LAST_INGEST_META_UNPARSABLE, record.last_modified)
     }
 
-    pub fn handle_ingested_record(&mut self, record: &RemoteSettingsRecord) -> Result<()> {
+    pub fn handle_ingested_record(
+        &mut self,
+        last_ingest_key: &str,
+        record: &RemoteSettingsRecord,
+    ) -> Result<()> {
         let record_id = SuggestRecordId::from(&record.id);
         // Remove this record's ID from the list of unparsable
         // records, since we understand it now.
         self.drop_unparsable_record_id(&record_id)?;
         // Advance the last fetch time, so that we can resume
         // fetching after this record if we're interrupted.
-        self.put_last_ingest_if_newer(record.last_modified)
+        self.put_last_ingest_if_newer(last_ingest_key, record.last_modified)
     }
 
-    pub fn handle_deleted_record(&mut self, record: &RemoteSettingsRecord) -> Result<()> {
+    pub fn handle_deleted_record(
+        &mut self,
+        last_ingest_key: &str,
+        record: &RemoteSettingsRecord,
+    ) -> Result<()> {
         let record_id = SuggestRecordId::from(&record.id);
         // Drop either the icon or suggestions, records only contain one or the other
         match record_id.as_icon_id() {
@@ -173,7 +181,7 @@ impl<'a> SuggestDao<'a> {
         self.drop_unparsable_record_id(&record_id)?;
         // Advance the last fetch time, so that we can resume
         // fetching after this record if we're interrupted.
-        self.put_last_ingest_if_newer(record.last_modified)
+        self.put_last_ingest_if_newer(last_ingest_key, record.last_modified)
     }
 
     // =============== Low level API ===============
@@ -1224,12 +1232,14 @@ impl<'a> SuggestDao<'a> {
 
     /// Updates the last ingest timestamp if the given last modified time is
     /// newer than the existing one recorded.
-    pub fn put_last_ingest_if_newer(&mut self, record_last_modified: u64) -> Result<()> {
-        let last_ingest = self
-            .get_meta::<u64>(LAST_INGEST_META_KEY)?
-            .unwrap_or_default();
+    pub fn put_last_ingest_if_newer(
+        &mut self,
+        last_ingest_key: &str,
+        record_last_modified: u64,
+    ) -> Result<()> {
+        let last_ingest = self.get_meta::<u64>(last_ingest_key)?.unwrap_or_default();
         if record_last_modified > last_ingest {
-            self.put_meta(LAST_INGEST_META_KEY, record_last_modified)?;
+            self.put_meta(last_ingest_key, record_last_modified)?;
         }
 
         Ok(())
--- a/components/suggest/src/provider.rs
+++ b/components/suggest/src/provider.rs
@@ -8,6 +8,8 @@ use rusqlite::{
     Result as RusqliteResult,
 };
 
+use crate::rs::SuggestRecordType;
+
 /// A provider is a source of search suggestions.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 #[repr(u8)]
@@ -44,6 +46,52 @@ impl SuggestionProvider {
             6 => Some(SuggestionProvider::Mdn),
             7 => Some(SuggestionProvider::Weather),
             _ => None,
+        }
+    }
+
+    pub(crate) fn records_for_provider(&self) -> Vec<SuggestRecordType> {
+        match self {
+            SuggestionProvider::Amp => {
+                vec![
+                    SuggestRecordType::AmpWikipedia,
+                    SuggestRecordType::Icon,
+                    SuggestRecordType::GlobalConfig,
+                ]
+            }
+            SuggestionProvider::Wikipedia => {
+                vec![
+                    SuggestRecordType::AmpWikipedia,
+                    SuggestRecordType::Icon,
+                    SuggestRecordType::GlobalConfig,
+                ]
+            }
+            SuggestionProvider::Amo => {
+                vec![SuggestRecordType::Amo, SuggestRecordType::GlobalConfig]
+            }
+            SuggestionProvider::Pocket => {
+                vec![SuggestRecordType::Pocket, SuggestRecordType::GlobalConfig]
+            }
+            SuggestionProvider::Yelp => {
+                vec![
+                    SuggestRecordType::Yelp,
+                    SuggestRecordType::Icon,
+                    SuggestRecordType::GlobalConfig,
+                ]
+            }
+            SuggestionProvider::Mdn => {
+                vec![SuggestRecordType::Mdn, SuggestRecordType::GlobalConfig]
+            }
+            SuggestionProvider::Weather => {
+                vec![SuggestRecordType::Weather, SuggestRecordType::GlobalConfig]
+            }
+            SuggestionProvider::AmpMobile => {
+                vec![
+                    SuggestRecordType::AmpMobile,
+                    SuggestRecordType::AmpWikipedia,
+                    SuggestRecordType::Icon,
+                    SuggestRecordType::GlobalConfig,
+                ]
+            }
         }
     }
 }
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -47,6 +47,20 @@ pub(crate) const REMOTE_SETTINGS_COLLECTION: &str = "quicksuggest";
 /// `mozilla-services/quicksuggest-rs` repo.
 pub(crate) const SUGGESTIONS_PER_ATTACHMENT: u64 = 200;
 
+/// A list of default record types to download if nothing is specified.
+/// This currently defaults to all of the record types.
+pub(crate) const DEFAULT_RECORDS_TYPES: [SuggestRecordType; 9] = [
+    SuggestRecordType::Icon,
+    SuggestRecordType::AmpWikipedia,
+    SuggestRecordType::Amo,
+    SuggestRecordType::Pocket,
+    SuggestRecordType::Yelp,
+    SuggestRecordType::Mdn,
+    SuggestRecordType::Weather,
+    SuggestRecordType::GlobalConfig,
+    SuggestRecordType::AmpMobile,
+];
+
 /// A trait for a client that downloads suggestions from Remote Settings.
 ///
 /// This trait lets tests use a mock client.
@@ -100,6 +114,59 @@ pub(crate) enum SuggestRecord {
     GlobalConfig(DownloadedGlobalConfig),
     #[serde(rename = "amp-mobile-suggestions")]
     AmpMobile,
+}
+
+/// Enum for the different record types that can be consumed.
+/// Extracting this from the serialization enum so that we can
+/// extend it to get type metadata.
+#[derive(Clone, PartialEq, PartialOrd, Eq, Ord)]
+pub enum SuggestRecordType {
+    Icon,
+    AmpWikipedia,
+    Amo,
+    Pocket,
+    Yelp,
+    Mdn,
+    Weather,
+    GlobalConfig,
+    AmpMobile,
+}
+
+impl From<SuggestRecord> for SuggestRecordType {
+    fn from(suggest_record: SuggestRecord) -> Self {
+        match suggest_record {
+            SuggestRecord::Amo => Self::Amo,
+            SuggestRecord::AmpWikipedia => Self::AmpWikipedia,
+            SuggestRecord::Icon => Self::Icon,
+            SuggestRecord::Mdn => Self::Mdn,
+            SuggestRecord::Pocket => Self::Pocket,
+            SuggestRecord::Weather(_) => Self::Weather,
+            SuggestRecord::Yelp => Self::Yelp,
+            SuggestRecord::GlobalConfig(_) => Self::GlobalConfig,
+            SuggestRecord::AmpMobile => Self::AmpMobile,
+        }
+    }
+}
+
+impl SuggestRecordType {
+    pub fn get_rs_type(&self) -> &str {
+        match self {
+            Self::Icon => "icon",
+            Self::AmpWikipedia => "data",
+            Self::Amo => "amo-suggestions",
+            Self::Pocket => "pocket-suggestions",
+            Self::Yelp => "yelp-suggestions",
+            Self::Mdn => "mdn-suggestions",
+            Self::Weather => "weather",
+            Self::GlobalConfig => "configuration",
+            Self::AmpMobile => "amp-mobile-suggestions",
+        }
+    }
+
+    /// Return the meta key for the last ingested record.
+    pub fn last_ingest_meta_key(&self) -> String {
+        format!("last_quicksuggest_ingest_{}", self.get_rs_type())
+    }
 }
 
 /// Represents either a single value, or a list of values. This is used to
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -31,7 +31,7 @@
 //!     the new suggestion in their results, and return `Suggestion::T` variants
 //!     as needed.
 
-use std::borrow::Cow;
+use std::{borrow::Cow, fmt};
 
 use remote_settings::{GetItemsOptions, RemoteSettingsResponse};
 use serde::{Deserialize, Deserializer};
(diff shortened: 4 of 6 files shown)


components/suggest/src/suggest.udl
```diff
@@ -69,6 +69,14 @@ interface Suggestion {
     );
 };
 
+enum SuggestRecord {
```
linabutler, line 72: Instead of surfacing `SuggestRecord` and "record types" to consumers, how would you feel about letting them specify providers instead? And then, in our component, we can do the mapping of provider to records.

So `SuggestionProvider.Amp` and `SuggestionProvider.Wikipedia` would fetch both `SuggestRecord.AmpWikipedia` and `SuggestRecord.Icon` records, `SuggestionProvider.Pocket` would fetch `SuggestRecord.Pocket` records, and so on.

That way, consumers won't need to know about our Remote Settings storage scheme, and we can switch it around—split up AMP and Wikipedia, remove icons and host them on a different server, remove attachments altogether—without them knowing.

components/suggest/src/rs.rs
```diff
@@ -81,7 +91,7 @@ impl SuggestRemoteSettingsClient for remote_settings::Client {
 /// suggestions are stored in each record's attachment.
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 #[serde(tag = "type")]
-pub(crate) enum SuggestRecord {
+pub enum SuggestRecord {
```
linabutler, line 94: One other reason I wanted to suggest (😝) using `SuggestionProvider` instead of `SuggestRecord` is that we use `SuggestRecord` to parse Remote Settings record payloads.

Right now, this is OK, because the records are empty (save for the `type` tag), and none of the variants have associated data. But this won't work once we add associated data to variants, like #6089—because the weather prefixes are stored in the record, not in the attachment.

components/suggest/src/rs.rs
```diff
@@ -94,6 +104,18 @@ pub(crate) enum SuggestRecord {
     Yelp,
 }
 
+impl SuggestRecord {
+    pub fn get_rs_type(&self) -> &str {
```
linabutler, line 108: I like the way you structured this helper! I think it can move to `SuggestionProvider`—and remember that `Amp` and `Wikipedia` need to fetch two types of records, `AmpWikipedia` and `Icon`.

components/suggest/src/store.rs
```diff
@@ -221,38 +222,48 @@ where
...
 
-                self.ingest_records(writer, &records_chunk)?;
+                self.ingest_records(writer, LAST_INGEST_META_UNPARSABLE, &records_chunk)?;
             }
         }
 
+        let ingest_record_types = if let Some(rt) = constraints.allowed_record_types {rt} else {DEFAULT_RECORDS_TYPES.to_vec()};
+        for ingest_record_type in ingest_record_types {
+            self.ingest_records_by_type(ingest_record_type, writer, constraints.max_suggestions)?;
+        }
```
mtbottle, line 232: To reduce the number of calls to Remote Settings, I can also use the `in_field`. I might change this to do that (will need to update the Remote Settings client though...). I tested that using `in_type` on multiple suggestions type returns all the types inclusive.

components/suggest/src/suggest.udl
mtbottle: > So SuggestionProvider.Amp and SuggestionProvider.Wikipedia would fetch both SuggestRecord.AmpWikipedia and SuggestRecord.Icon records,

Ah cool this would work. I wasn't sure if what we would do with Icons if I used providers, which is why I opted for more control of records. If we know what the mapping looks like, I can do the Provider -> Record mapping.

components/suggest/src/store.rs
```diff
@@ -221,38 +222,48 @@ where
...
+
+        Ok(())
+    }
+
+    fn ingest_records_by_type(&self, ingest_record_type: SuggestRecord, writer: &SuggestDb, max_suggestions:Option<u64> ) -> Result<()> {
```
linabutler, line 237: One other suggestion for making this a bit more Rusty...

* Let's give `SuggestionProvider` two methods: `get_rs_types()`, which you already added to `SuggestRecord`, and `last_ingest_meta_key()`, to build the meta key string for us.
* `ingest_records_by_type()` can take a `SuggestionProvider`, and use those methods to figure out which records it needs to fetch for that provider, and which key to use for the last ingest time.

linabutler (changes requested):
This is looking wonderful, @mtbottle; thank you for taking this on! My one big suggestion is to surface "providers" instead of "records" to consumers, and the other comments I left flow from that—but this is looking very complete 😁👍🏼

components/suggest/src/store.rs
linabutler: Oh, that would be wonderful! Yes, please feel free to change the Remote Settings client—we're the only consumer of `GetItemsOptions` now.

components/suggest/src/store.rs
mtbottle: > , and last_ingest_meta_key(), to build the meta key string for us. 

Good idea!

linabutler (changes requested):
I like the direction you're going, but I did have a question about using `get_last_ingest()`...

components/suggest/src/store.rs
```diff
@@ -311,12 +314,30 @@ where
...
+        options.eq_any("type", ingest_record_types);
+
         // Remote Settings returns records in descending modification order
         // (newest first), but we want them in ascending order (oldest first),
         // so that we can eventually resume downloading where we left off.
         options.sort("last_modified", SortOrder::Ascending);
-        if let Some(last_ingest) = writer.read(|dao| dao.get_meta::<u64>(LAST_INGEST_META_KEY))? {
+
+        // Get the last ingest value. This is the max of the last_ingest_keys
+        // that are in the database.
```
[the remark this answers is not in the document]
mtbottle: Reverted it back to doing multiple calls!

components/suggest/src/provider.rs
```diff
@@ -45,6 +47,22 @@ impl SuggestionProvider {
...
+            }
+            SuggestionProvider::Wikipedia => {
+                vec![SuggestRecordType::AmpWikipedia, SuggestRecordType::Icon]
+            }
+            SuggestionProvider::Amo => vec![SuggestRecordType::Amo],
+            SuggestionProvider::Pocket => vec![SuggestRecordType::Pocket],
+            SuggestionProvider::Yelp => vec![SuggestRecordType::Yelp],
+            SuggestionProvider::Mdn => vec![SuggestRecordType::Mdn],
+            SuggestionProvider::Weather => vec![SuggestRecordType::Weather],
+        }
```
mtbottle, line 64: I see that we have a new `SuggestRecord` type called `GlobalConfig`. Which providers should it be mapped with?

components/suggest/src/db.rs
```diff
@@ -921,6 +923,14 @@ impl<'a> SuggestDao<'a> {
         )?)
     }
 
+    pub fn get_last_ingest<T: FromSql>(&self) -> Result<Option<T>> {
```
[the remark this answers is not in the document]
mtbottle: This function is no longer relevant, so I removed it.

components/remote_settings/src/client.rs
```diff
@@ -330,6 +330,19 @@ impl GetItemsOptions {
...
     }
 
+    /// Sets an option to return items whose `field` matches any one of the `values`.
+    ///
+    /// `field` can be a simple or dotted field name, like `author` or
+    /// `author.name`. `values` is a vector of Strings.
+    pub fn eq_any(
+        &mut self,
+        field: impl Into<String>,
+        values: impl Into<Vec<String>>,
```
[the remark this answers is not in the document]
mtbottle: This function is no longer relevant, so I removed it.

components/remote_settings/src/client.rs
```diff
@@ -330,6 +330,19 @@ impl GetItemsOptions {
         self
     }
 
+    /// Sets an option to return items whose `field` matches any one of the `values`.
+    ///
+    /// `field` can be a simple or dotted field name, like `author` or
+    /// `author.name`. `values` is a vector of Strings.
+    pub fn eq_any(
```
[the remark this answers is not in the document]
mtbottle: Ah ya, I used `eq_any` because `in` is a keyword, so it's kind of confusing. Also, we already have an `eq` function which does a match to one particular field. So, this is _kind_ of an extension of that: it has to match _any_ of the fields in the list. It is definitely more descriptive than `in`!

components/suggest/src/store.rs
```diff
@@ -4034,13 +4132,67 @@ mod tests {
...
+            );
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_pocket-suggestions")?,
+                None
+            );
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_amo-suggestions")?,
+                None
+            );
             Ok(())
```
mtbottle, line 4196: I was originally not super happy with the tests because the snapshot returns the same data every time we call remote settings (in this case, twice). This means that although we're trying to get only "Pocket" data, the snapshot returns whatever we configured it to return (for all the other types as well). 

Now that I'm thinking of it, the core functionality of the test, which is to separate the last_ingest record, kind of _works_ as it does properly separate the values in storage properly. I don't know, what do you think @linabutler ?

components/suggest/src/rs.rs
```diff
@@ -100,6 +113,41 @@ pub(crate) enum SuggestRecord {
     GlobalConfig(DownloadedGlobalConfig),
 }
 
+/// Enum for the different record types that can be consumed.
+/// Extracting this from the serialization enum so that we can
+/// extend it to get type metadata.
```
bendk, line 118: What do you think about implementing `From<SuggestRecord> for SuggestRecordType`?  I say that because it explicitly shows the relationship between the two types (I think `SuggestRecordType` is just `SuggestRecord` without the field data?  If I saw the mapping I would know for sure).

components/suggest/src/suggest.udl
```diff
@@ -95,6 +95,7 @@ dictionary SuggestionQuery {
 
 dictionary SuggestIngestionConstraints {
     u64? max_suggestions = null;
+    sequence<SuggestionProvider>? providers = null;
```
bendk, line 98: Really like the API design here.  It's nice that you can just extend the query options by adding a field.

bendk (approved):
Looks great to me FWIW.  I don't have a solution for the tests and just had a couple of minor suggestions.  Feel free to take them or leave them.

components/suggest/src/db.rs
```diff
@@ -30,9 +30,11 @@ use crate::{
     Result, SuggestionQuery,
 };
 
-/// The metadata key whose value is the timestamp of the last record ingested
-/// from the Suggest Remote Settings collection.
-pub const LAST_INGEST_META_KEY: &str = "last_quicksuggest_ingest";
+/// The metadata key prefix for the key whose value is the timestamp
+/// of the last record ingested from the Suggest Remote Settings collection.
+/// This will need to be appended with the record type to get the actual key.
+pub const LAST_INGEST_META_PREFIX: &str = "last_quicksuggest_ingest_";
```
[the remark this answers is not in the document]
mtbottle: Oh you're right! I only use this in one place, so I can just declare it in that method, rather than pulling it out as a constant.

components/suggest/src/rs.rs
mtbottle: Ah, I didn't know you can do that! I think adding the mapping will definitely show the relationship better, so I'll add that. 👍

ncloudioj: FYI, [`strum`](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumDiscriminants.html) is made exactly for use cases like this: you want an enum w/o all the data fields. Doing this manually could be tedious as now we will have to add it to various places when adding a new record type. The following will generate an enum `SuggestRecordType` for you automatically with all the bells and whistles. 

```rs
#[derive(Clone, Debug, Deserialize, EnumDiscriminants]
#[strum_discriminants(name(SuggestRecordType))]
pub(crate) enum SuggestRecord {
  #[serde(rename = "icon")]
  #[strum_discriminants(strum(message = "icon"))]  // This maps to what you define in `get_rs_type()` and can be used as `SuggestRecordType::Icon.get_message()`
  Icon,
}
```

mtbottle: >   #[strum_discriminants(strum(message = "icon"))]  // This maps to what you define in `get_rs_type()` and can be used as `SuggestRecordType::Icon.get_message()`

I'm a bit confused on how to use this (Rust enums confuses me :p). So, let's say I want to access the "icon" value from `SuggestRecord::Icon`, how do I do that without the match in `get_rs_type` method?

I did manage to reduce duplication on the `SuggestRecordType` though! Thanks for the suggestion!

ncloudioj: Oh, actually, I think we can use `Strum::Display` to further simplify it:

```rs
use strum_macros::Display as StrumDisplay;

#[derive(Clone, Debug, Deserialize, EnumDiscriminants, StrumDisplay)]
#[strum_discriminants(name(SuggestRecordType))]
#[strum_discriminants(derive(StrumDisplay))]
#[serde(tag = "type")]
pub(crate) enum SuggestRecord {
    #[serde(rename = "icon")]
    #[strum(to_string = "icon")]
    Icon,
    #[serde(rename = "data")]
    #[strum(to_string = "data")]
    AmpWikipedia,
    #[serde(rename = "amo-suggestions")]
    #[strum(to_string = "amo-suggestions")]
    Amo,
    #[serde(rename = "pocket-suggestions")]
    #[strum(to_string = "pocket-suggestions")]
    Pocket,
    #[serde(rename = "yelp-suggestions")]
    #[strum(to_string = "yelp-suggestions")]
    Yelp,
    #[serde(rename = "mdn-suggestions")]
    #[strum(to_string = "mdn-suggestions")]
    Mdn,
    #[serde(rename = "weather")]
    #[strum(to_string = "weather")]
    Weather(DownloadedWeatherData),
    #[serde(rename = "configuration")]
    #[strum(to_string = "configuration")]
    GlobalConfig(DownloadedGlobalConfig),
}

impl SuggestRecordType {
    /// Return the meta key for the last ingested record.
    pub fn last_ingest_meta_key(&self) -> String {
        format!("last_quicksuggest_ingest_{}", self) // because `SuggestRecordType` now implements `Display`
    }
}

// For other places where you need `ingest_record_type.get_rs_type()`, you can just do `ingest_record_type.to_string()`  

```

mtbottle: Cool nice! I'll try this as it'll reduce duplication!

ncloudioj: There is more 😃 , if you stack `#[strum_discriminants(derive(EnumIter))]` to SuggestRecord, then we can also replace `DEFAULT_RECORDS_TYPES` with `SuggestRecordType::iter().collect::<Vec<_>>()`.

linabutler:
> FYI, [strum](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumDiscriminants.html) is made exactly for use cases like this: you want an enum w/o all the data fields. Doing this manually could be tedious as now we will have to add it to various places when adding a new record type.

Nice find, @ncloudioj!

It looks like we don't use `strum` in m-c yet, so we'll need to do a [`cargo vet` audit](https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html#third-party-crate-dependencies) the next time we vendor in m-c after this PR lands. Here are some [examples of audit notes](https://searchfox.org/mozilla-central/rev/4fe00e0322377316390da6faa2d645cae53d08f4/supply-chain/audits.toml).

We have a [similar, but less formal policy](https://mozilla.github.io/application-services/book/dependency-management.html#rust-code) specifically for Application Services, since we exist outside m-c for Android and iOS.

We don't have a direct connection to `strum`, but it looks like it doesn't have any `unsafe` code, all its dependencies have already been vetted for m-c, it's actively maintained, the code it generates looks very reasonable, and it provides non-trivial functionality that we wouldn't want to hand-write. I think the `#supply-chain-reviewers` Phabricator review group will sign off on it easily.

mtbottle:
After having a chat with a few folks about adding dependencies Application Services, we've decided that it's better to not add `strum`. So this is Ready For Review again.

components/suggest/src/store.rs
```diff
@@ -337,16 +340,43 @@ where
                     .get_records_with_options(&options)?
                     .records;
 
-                self.ingest_records(writer, &records_chunk)?;
+                self.ingest_records(LAST_INGEST_META_UNPARSABLE, writer, &records_chunk)?;
             }
         }
 
+        let ingest_record_types = if let Some(rt) = &constraints.providers {
+            rt.iter().flat_map(|x| x.records_for_provider()).collect()
```
ncloudioj, line 348: Do we want to dedup the types in the result vector? `Icon` and `GlobalConfig` are shared by many providers, would it re-ingest those record types w/o deduping?

components/suggest/src/store.rs
```diff
@@ -388,14 +422,24 @@ where
 
             match fields {
                 SuggestRecord::AmpWikipedia => {
-                    self.ingest_attachment(writer, record, |dao, record_id, suggestions| {
-                        dao.insert_amp_wikipedia_suggestions(record_id, suggestions)
-                    })?;
+                    self.ingest_attachment(
+                        &SuggestRecordType::AmpWikipedia.last_ingest_meta_key(),
```
ncloudioj, line 426: Q: we already have an argument `last_ingest_meta_key`, why do we need to generate another meta key for each `ingest_attachment()`? Are there different?

I find it quite hard to track things as now we have to pass this `last_ingest_meta_key` all the way to the bottom, e.g. `ingest_attachment() => ingest_record() => dao.handle_ingested_record()` because we call `self.put_last_ingest_if_newer(last_ingest_key)` in all the high-level APIs of `SugegstDao`. I wonder if it'd be easier to do that at the store level, such as in `Store::ingest_record()` so that `last_ingest_meta_key` won't contaminate stuff in `db.rs`.

ncloudioj: Alternatively, if updating last_ingest_meta_key at the store level is non-trivial to refactor, we can combine this meta key and `record` into a new struct as the latter is already passed around everywhere.

WDYT?

components/suggest/src/store.rs
mtbottle: Oh ya, there was a comment there originally about doing the dedupe. I'll see if I can add it easily.

components/suggest/src/store.rs
mtbottle: Oh, I think this was an artifact of refactoring the code to return multiple types. Now, it's only a singular type, so we can just use the type passed in params. I'll fix this.

> I wonder if it'd be easier to do that at the store level, such as in Store::ingest_record() so that last_ingest_meta_key won't contaminate stuff in db.rs.

This would be messier and kind of works against the refactoring to move logical pieces of logic to the Dao. 

> Alternatively, if updating last_ingest_meta_key at the store level is non-trivial to refactor, we can combine this meta key and record into a new struct as the latter is already passed around everywhere.

It's only one extra value, so I don't feel like it's a huge deal to pass it as a param. If we find that we have a lot of logical pieces of data that we continually pass with records, then I this a new struct would make sense at that point. Since it's only one piece of data, I'm not sure it's worth creating a new object.

> I find it quite hard to track things as now we have to pass this last_ingest_meta_key all the way to the bottom, e.g. ingest_attachment() => ingest_record() => dao.handle_ingested_record() because we call self.put_last_ingest_if_newer(last_ingest_key) in all the high-level APIs of SugegstDao. I

I think the confusion is because `ingest_attachment` has logic that calls `put_last_ingest_if_newer`, which isn't encapsulated in the high level dao functions. Ideally, we have it's just `ingest_record` -> `dao.handle_ingested_record`, which isn't as many levels of indirection. I don't really understand how attachments work, so not sure if there's a way we can refactor this to make it simpler...

I would like to get this PR in as soon as possible and let someone more skilled with Rust refactor it. What do you think about leaving the code as is (with the fix mentioned above), then letting someone else do the refactor?

ncloudioj: Yup, for sure. Feel free to leave it as is. I think some future refactors (e.g. merge `record` and `last_ingest_meta_key` into an ingestion context struct and pass it down) would make this easier to follow and maintain.

mtbottle: Arg, the passing of `last_ingest_key` won't work with the current tests (because the full snapshots are being passed). So, I'll need to keep the full reference for the tests to pass, or otherwise figure out how to fix the tests referenced here: https://github.com/mozilla/application-services/pull/6095#discussion_r1484415054 Thoughts on whether this is a dealbreaker? @ncloudioj I can try to fix the tests instead...

ncloudioj: Hmm, I am still not entirely sure about the problem, but I'd try fixing the tests to make the code easier to follow here.

components/suggest/src/store.rs
```diff
@@ -4426,13 +4531,67 @@ mod tests {
...
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_icon")?,
+                Some(30)
+            );
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_pocket-suggestions")?,
+                None
+            );
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_amo-suggestions")?,
```
ncloudioj, line 4592: Let's also test other suggestion types (`Yelp, AmpMobile`, `Mdn`, and `Weather`) were not ingested for completeness.

ncloudioj (dismissed):
Dropping my r+ here as I will be out for several days. I think it looks good in general, just a few minor comments. Thanks!

components/suggest/src/store.rs
```diff
@@ -337,16 +340,45 @@ where
                     .get_records_with_options(&options)?
                     .records;
 
-                self.ingest_records(writer, &records_chunk)?;
+                self.ingest_records(LAST_INGEST_META_UNPARSABLE, writer, &records_chunk)?;
```
mtbottle, line 343: @linabutler I've been thinking about "unparsable" records a bit, and I see 2 options:

1. If we decide that we don't need to reprocess the record, then we can probably just get rid of all of the unparsable logic. It's a larger change, so might be better done in a separate PR. From my limited understanding of what unparsable records are meant for, I think it seems reasonable to simplify this to not have to reprocess unparsable records (because we won't be processing records of a type we don't understand with my change). For corrupted data, we will reprocessed the updated file, so we will unlikely need to reprocess an existing version.
2. If we decide to keep the logic for unparsable records (which is probably the safer approach), then we can keep the code that's kind of "duplicating" (i.e. use the exact record type like SuggestType::Icons.get_rs_type(), rather than pass in the last_ingest_key). That's because the match for the type will be done at the "match" statement to match the type of record. It's kind of annoying to pass the type multiple times, but it will prevent another matching done on this line to get the "actual" record type. 

Not sure if you have an opinion, but writing it out here so we can discuss this tomorrow.

components/suggest/src/store.rs
```diff
@@ -337,16 +340,45 @@ where
...
             }
         }
 
+        let mut ingest_record_types = if let Some(rt) = &constraints.providers {
+            rt.iter().flat_map(|x| x.records_for_provider()).collect()
+        } else {
+            DEFAULT_RECORDS_TYPES.to_vec()
+        };
+        ingest_record_types.sort_unstable();
+        ingest_record_types.dedup();
```
ncloudioj, line 353: This does the deduping and maintains the order. Also no need to dedup the constant type array. 
```suggestion
        // use std::collections::BTreeSet;
        let ingest_record_types = if let Some(rt) = &constraints.providers {
            rt.iter()
                .flat_map(|x| x.records_for_provider())
                .collect::<BTreeSet<_>>()
                .into_iter()
                .collect()
        } else {
            DEFAULT_RECORDS_TYPES.to_vec()
        };
```

components/suggest/src/store.rs
ncloudioj: @mtbottle can you add a brief comment to note why it needs to generate the last ingest meta key again here so that we know how to refactor it in the future. Adding one to here would be sufficient, not need to do so for other match variants. Thanks!

components/suggest/src/store.rs
```diff
@@ -721,7 +804,10 @@ mod tests {
         store.ingest(SuggestIngestionConstraints::default())?;
 
         store.dbs()?.reader.read(|dao| {
-            assert_eq!(dao.get_meta::<u64>(LAST_INGEST_META_KEY)?, Some(15));
+            assert_eq!(
+                dao.get_meta::<u64>("last_quicksuggest_ingest_data")?,
```
ncloudioj, line 808: Nit: Just in case we change the key format in the future, let's not use the hardcoded keys in the tests, we can call `SuggestRecordType::last_ingest_meta_key()` instead with the associated type.

ncloudioj (dismissed):
r+wc. It's almost there.

components/remote_settings/src/client.rs
```diff
@@ -309,7 +309,7 @@ struct AttachmentsCapability {
 /// Options for requests to endpoints that return multiple items.
 #[derive(Clone, Debug, Default)]
 pub struct GetItemsOptions {
-    filters: Vec<Filter>,
+    pub filters: Vec<Filter>,
```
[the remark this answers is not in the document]
mtbottle: Arg, artifact of some changes I've made. This file shouldn't be changed at all, so I'll remove it.

components/suggest/src/rs.rs
mtbottle: Ok sure! I'll update this.

PR #6096, opened 2024-02-01
Bug 1875113 - Handle JSON errors when parsing attachments

bendk:
This handles the `Invalid JSON in a record or an attachment.` bullet point from the issue.

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -381,31 +381,37 @@ where
             return Ok(());
         };
 
-        let attachment: SuggestAttachment<T> =
-            serde_json::from_slice(&self.settings_client.get_attachment(&attachment.location)?)?;
+        let attachment_data = self.settings_client.get_attachment(&attachment.location)?;
+        match serde_json::from_slice::<SuggestAttachment<T>>(&attachment_data) {
+            Ok(attachment) => writer.write(|dao| {
+                // Drop any suggestions that we previously ingested from
+                // this record's attachment. Suggestions don't have a
+                // stable identifier, and determining which suggestions in
+                // the attachment actually changed is more complicated than
+                // dropping and re-ingesting all of them.
+                dao.drop_suggestions(&record_id)?;
 
-        writer.write(|dao| {
-            // Drop any suggestions that we previously ingested from
-            // this record's attachment. Suggestions don't have a
-            // stable identifier, and determining which suggestions in
-            // the attachment actually changed is more complicated than
-            // dropping and re-ingesting all of them.
-            dao.drop_suggestions(&record_id)?;
+                // Ingest (or re-ingest) all suggestions in the
+                // attachment.
+                ingestion_handler(dao, &record_id, attachment.suggestions())?;
 
-            // Ingest (or re-ingest) all suggestions in the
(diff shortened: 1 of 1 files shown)


components/suggest/src/store.rs
```diff
@@ -366,31 +366,37 @@ where
...
-            // records, since we understand it now.
-            dao.drop_unparsable_record_id(&record_id)?;
-
-            // Advance the last fetch time, so that we can resume
-            // fetching after this record if we're interrupted.
-            dao.put_last_ingest_if_newer(record.last_modified)?;
-
-            Ok(())
-        })
+        let attachment_data = self.settings_client.get_attachment(&attachment.location)?;
```
bendk, line 369: Should we also handle errors fetching the attachment in the same way?

components/suggest/src/store.rs
```diff
@@ -366,31 +366,37 @@ where
...
+                // Advance the last fetch time, so that we can resume
+                // fetching after this record if we're interrupted.
+                dao.put_last_ingest_if_newer(record.last_modified)?;
+
+                Ok(())
+            }),
+            Err(_) => writer.write(|dao| {
+                // The attachment was not able to be parsed, record this and continue on
+                dao.put_unparsable_record_id(&record_id)?;
+                dao.put_last_ingest_if_newer(record.last_modified)?;
```
bendk, line 396: Would it make sense to put this code in a `dao` method?  I slightly worry that somewhere down the line there will be code that only does the first operation, but not the second one.  Alternatively, maybe `SuggestDao` only have low-level CRUD methods and a higher-level component could handle compound operations like this one.

components/suggest/src/store.rs
linabutler: Ooh, good question! I was thinking no, because `get_attachment` just does the network request, and the error would most likely be a network error (or some other kind of error that we'd want to report). But an error when parsing means "we don't understand this attachment, and trying to fetch it again won't help".

WDYT?

components/suggest/src/store.rs
```diff
@@ -3452,6 +3458,122 @@ mod tests {
         Ok(())
     }
 
+    /// Tests ingesting suggestions into an empty database.
```
linabutler, line 3461:
```suggestion
    /// Tests that records with invalid attachments are ignored and marked as unparsable.
```

linabutler (approved):
Woohoo!

components/suggest/src/store.rs
linabutler: I think it would make sense, and `SuggestDao` sounds like a good place to put it! If it starts getting unwieldy having the lower-level methods and the higher-level ones in `SuggestDao`, maybe we could think about separating them down the road, but I'm not sure we've reached that point just yet.

Our other access pattern is to do `drop_unparsable_record_id + put_last_ingest_if_newer` for records that we did parse and store successfully—worth having a second `SuggestDao` helper for that, too?

components/suggest/src/store.rs
bendk: Yeah that makes sense to me as well.  I wasn't sure what work `get_attachment` was doing.

PR #6097, opened 2024-02-02
Bug 1877501 - Teach Suggest about the configuration data in RS

0c0w3:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -19,12 +19,11 @@ use crate::{
     pocket::{split_keyword, KeywordConfidence},
     provider::SuggestionProvider,
     rs::{
-        DownloadedAmoSuggestion, DownloadedAmpWikipediaSuggestion, DownloadedConfig,
-        DownloadedMdnSuggestion, DownloadedPocketSuggestion, DownloadedWeatherData,
-        SuggestRecordId,
+        DownloadedAmoSuggestion, DownloadedAmpWikipediaSuggestion, DownloadedMdnSuggestion,
+        DownloadedPocketSuggestion, DownloadedWeatherData, SuggestRecordId,
     },
     schema::{SuggestConnectionInitializer, VERSION},
-    store::{SuggestConfig, UnparsableRecord, UnparsableRecords},
+    store::{SuggestGlobalConfig, SuggestProviderConfig, UnparsableRecord, UnparsableRecords},
     suggestion::{cook_raw_suggestion_url, Suggestion},
     Result, SuggestionQuery,
 };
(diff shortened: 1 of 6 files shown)


0c0w3:
@linabutler I still like your idea of per-provider config. We don't need it at this time so I didn't implement support for it. It's not hard to imagine modifying the API and storing per-provider rows in `meta`.

There are two pieces of config we currently have that could be considered per-provider:

* The global config has sponsored impression caps, but they aren't turned on in desktop and never have been by default. If we need these in the future, that would be a great opportunity to add support for per-provider and use it for Amp suggestions.
* Weather has `min_keyword_length`. I went ahead and hardcoded a similar value in the weather PR. The problem is `min_keyword_length` is just the initial min length. The user can increase it by clicking "Show less frequently". We don't currently store user data in this component and I didn't want to get caught up in that right now. Alternatively the client could pass in the current min length as a new `query()` param, but I don't want to make that larger change now either (not sure it even makes sense), and if we're relying on the client to store the min length anyway, then it can just discard a weather suggestion if the min length isn't satisfied. I think we can revisit this in the future.

0c0w3:
Well the client will need to know the initial min length though. Let me see if I can add weather config after all.

linabutler:
> Let me see if I can add weather config after all.

Sounds good, thanks! Your weather provider PR looks great; if you'd like to land that first, that might make it easier to implement config for it in this PR, too?

> We don't currently store user data in this component and I didn't want to get caught up in that right now. 

Yeah, I think this would be totally fine to save for a follow-up. @bendk took the first step toward having the Suggest component store user data over in #6101, so maybe the first use case for that could be storing the `min_keyword_length`, instead of dismissals.

components/suggest/src/store.rs
```diff
@@ -157,6 +174,34 @@ pub struct SuggestIngestionConstraints {
     pub max_suggestions: Option<u64>,
 }
 
+/// Global Suggest configuration data.
```
linabutler, line 177: Small suggestion: Shall we put these types in a new `config.rs` module?

linabutler (approved):
Looks great! I like the new helpers, and the new `fetch_{global, provider}_config()` interfaces!

components/suggest/src/db.rs
```diff
@@ -979,4 +990,45 @@ impl<'a> SuggestDao<'a> {
...
+        &self,
+        provider: SuggestionProvider,
+    ) -> Result<Option<SuggestProviderConfig>> {
+        self.get_meta::<String>(&provider_config_meta_key(provider))?
+            .map_or_else(|| Ok(None), |json| Ok(serde_json::from_str(&json)?))
+    }
+}
+
+fn provider_config_meta_key(provider: SuggestionProvider) -> String {
+    format!("{}{}", PROVIDER_CONFIG_META_KEY_PREFIX, provider as u8)
```
ncloudioj, line 1033: Question: since we don't have a `suggestion_provider` table to store this enum in the db. We'd have to reference the source to decode provider config for each key, which might get annoying for debugging. How about we store the string presentation instead in the meta table (i.e. to implement "std::fmt::Display" for `SuggestionProvider`)? Or we could create a table `suggestion_provider` as other tables with the `provider` column can also use it to get the human readable providers.

WDYT?

ncloudioj (approved):
LGTM! Just a question for discussion.

components/suggest/src/db.rs
0c0w3: Yeah, I like both ideas. wdyt about continuing to use `PROVIDER_CONFIG_META_KEY_PREFIX` in that case? I'm hesitant to store names or u8's by themselves in a `meta` column since we'd need to be careful to avoid name conflicts with other meta keys. Actually I like your `suggestion_provider` table idea because then it could store both names and config -- a table for provider info.

0c0w3: All right, I'll go ahead and merge this now then, thanks.

components/suggest/src/store.rs
0c0w3: Sounds good!

components/suggest/src/db.rs
0c0w3: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1878941 so we don't lose track of this idea.

PR #6101, opened 2024-02-02
Suggest - add a data path field

bendk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1876208 points out that we need a permanent database to store user dismissal data -- one that won't be deleted when the cache is cleared.

`SuggestStoreInner` stores the new field, but doesn't actually use it yet.  Let's wait until the consumer code is updated before that.

I added this capability using a builder object for two reasons:
  - This allows us to keep the current constructor to avoid any breaking changes while consumers adapt to the new system.
  - It forces consumers to specify `cache_path` or `data_path`.  If the constructor inputted 2 strings it would be easy to mix them up and hard to notice.

Lastly, input paths as owned strings rather than refs.  I think this avoids an extra copy.

  - Note:

### DIFF
--- a/components/suggest/src/suggest.udl
+++ b/components/suggest/src/suggest.udl
@@ -106,3 +106,19 @@ interface SuggestStore {
     [Throws=SuggestApiError]
     void clear();
 };
+
+interface SuggestStoreBuilder {
+    constructor();
+
+    [Self=ByArc]
+    SuggestStoreBuilder data_path(string path);
+
+    [Self=ByArc]
+    SuggestStoreBuilder cache_path(string path);
+
+    [Self=ByArc]
+    SuggestStoreBuilder remote_settings_config(RemoteSettingsConfig config);
+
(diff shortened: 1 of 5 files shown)


components/suggest/src/suggest.udl
```diff
@@ -106,3 +106,19 @@ interface SuggestStore {
...
+    SuggestStoreBuilder data_path(string path);
+
+    [Self=ByArc]
+    SuggestStoreBuilder cache_path(string path);
+
+    [Self=ByArc]
+    SuggestStoreBuilder remote_settings_config(RemoteSettingsConfig config);
+
+    [Throws=SuggestApiError]
+    SuggestStore finish();
```
linabutler, line 123: Nano-nit: What do you think of calling this `build()`, to match Java / Kotlin conventions?

linabutler (approved):
Love this!

bendk:
Both of those changes make sense to me.

PR #6107, opened 2024-02-05
Bug 1876219 -- Consider attachments with newer fields unparsable

bendk:
I got this working by adding `deny_unknown_fields` to all attachment structs that implement `Deserialize`.

The tests pass, but I'm not sure I'm doing the right thing with the various serde structs.

  - Note:

### DIFF
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -34,7 +34,7 @@
 use std::borrow::Cow;
 
 use remote_settings::{GetItemsOptions, RemoteSettingsResponse};
-use serde::{Deserialize, Deserializer};
+use serde::{de::DeserializeOwned, Deserialize, Deserializer};
 
 use crate::{provider::SuggestionProvider, Result};
 
@@ -100,27 +100,42 @@ pub(crate) enum SuggestRecord {
     GlobalConfig(DownloadedGlobalConfig),
 }
 
-/// Represents either a single value, or a list of values. This is used to
-/// deserialize downloaded attachments.
+/// Implemented by the various attachment types
+pub trait SuggestAttachment: DeserializeOwned + std::fmt::Debug {
+    /// Get fields that we saw when ingesting this attachment, but don't recognize
+    fn unknown_fields(&self) -> &serde_json::Map<String, serde_json::Value>;
+
+    fn has_unknown_fields(&self) -> bool {
+        !self.unknown_fields().is_empty()
+    }
+}
+
+/// The payload of a downloaded attachment.
+///
+/// This can either be a single value or a list of values.  Values can either be suggestions or
+/// provider configuration data.
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
-enum OneOrMany<T> {
+pub enum AttachmentPayload<T> {
     One(T),
     Many(Vec<T>),
 }
 
-/// A downloaded Remote Settings attachment that contains suggestions.
-#[derive(Clone, Debug, Deserialize)]
-#[serde(transparent)]
-pub(crate) struct SuggestAttachment<T>(OneOrMany<T>);
-
-impl<T> SuggestAttachment<T> {
-    /// Returns a slice of suggestions to ingest from the downloaded attachment.
-    pub fn suggestions(&self) -> &[T] {
-        match &self.0 {
-            OneOrMany::One(value) => std::slice::from_ref(value),
-            OneOrMany::Many(values) => values,
+impl<T> AttachmentPayload<T> {
+    pub fn as_slice(&self) -> &[T] {
+        match &self {
+            AttachmentPayload::One(value) => std::slice::from_ref(value),
+            AttachmentPayload::Many(values) => values,
         }
+    }
+}
+
+impl<T: SuggestAttachment> AttachmentPayload<T> {
+    /// Did any of the payloads (i.e. JSON values) in this attachment have an unknown field?
+    pub fn has_unknown_fields(&self) -> bool {
+        self.as_slice()
+            .iter()
+            .any(SuggestAttachment::has_unknown_fields)
     }
 }
 
@@ -175,6 +190,8 @@ pub(crate) struct DownloadedAmpSuggestion {
     pub impression_url: String,
     #[serde(rename = "icon")]
     pub icon_id: String,
+    #[serde(flatten)]
+    unknown_fields: serde_json::Map<String, serde_json::Value>,
 }
 
 /// A Wikipedia suggestion to ingest from an AMP-Wikipedia attachment.
@@ -184,6 +201,8 @@ pub(crate) struct DownloadedWikipediaSuggestion {
     pub common_details: DownloadedSuggestionCommonDetails,
     #[serde(rename = "icon")]
     pub icon_id: String,
+    #[serde(flatten)]
+    unknown_fields: serde_json::Map<String, serde_json::Value>,
 }
 
 /// A suggestion to ingest from an AMP-Wikipedia attachment downloaded from
@@ -263,6 +282,8 @@ pub(crate) struct DownloadedAmoSuggestion {
     pub title: String,
     pub keywords: Vec<String>,
     pub score: f64,
+    #[serde(flatten)]
+    unknown_fields: serde_json::Map<String, serde_json::Value>,
 }
 /// A Pocket suggestion to ingest from a Pocket Suggestion Attachment
 #[derive(Clone, Debug, Deserialize)]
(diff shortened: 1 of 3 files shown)


components/suggest/src/rs.rs
```diff
@@ -162,6 +162,10 @@ pub(crate) struct DownloadedSuggestionCommonDetails {
 
 /// An AMP suggestion to ingest from an AMP-Wikipedia attachment.
 #[derive(Clone, Debug, Deserialize)]
+// We use deny_unknown_fields for all attachment types so that unrecognized fields result in
+// `UnparsableRecord` entries.  This means that if we add new attachment fields, older clients will
+// ignore the suggestions until they know how to parse them.
+#[serde(deny_unknown_fields)]
```
linabutler, line 168: Hmmm, I think this is going a bit too far the other way—it means that older clients will stop being able to display suggestions that they previously could, as soon as we add new fields to them.

I wonder if we can use `#[serde(other)]` to soak up any unknown fields, then check if the map is empty. If it is, we know we can understand the suggestion completely; if it's not, we store the fields we do know, and remember the ID to refetch after a schema version upgrade.

The high-level helpers you added in #6106 would need to be taught about this new case—a record could now be in `unparsable_records` and in `suggestions.record_id` if we understood some of its fields—which is making me wonder if we actually want to track these record IDs in a separate `meta` key, because they're "partially parsable", not "unparsable".

WDYT?

components/suggest/src/rs.rs
```diff
@@ -272,16 +280,21 @@ pub(crate) struct DownloadedPocketSuggestion {
     #[serde(rename = "highConfidenceKeywords")]
     pub high_confidence_keywords: Vec<String>,
     pub score: f64,
+    #[allow(unused)]
+    #[serde(default)]
+    pub description: String,
```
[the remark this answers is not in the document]
linabutler: That's correct!

linabutler (changes requested):
I left a comment that `deny_unknown_fields` could be too much, but let's discuss more—requesting changes so that I don't forget to check back later 😅

components/suggest/src/rs.rs
bendk: I think that makes a lot of sense.  I updated the PR to follow those semantics.

The one thing I didn't do was create a separate meta key for this.  I think it might be simpler to keep 1 key and to change the naming from "unparsable records" to "records to retry" or something like that.  Do we need to differentiate between the two cases?  If so, maybe we could still keep one key but add a field that distinguishes between the two.

Adding another meta key also seems fine to me, I just wanted to discuss it more with you before committing to anything.

components/suggest/src/rs.rs
```diff
@@ -100,30 +100,44 @@ pub(crate) enum SuggestRecord {
...
+        !self.unknown_fields().is_empty()
+    }
+}
+
 /// Represents either a single value, or a list of values. This is used to
 /// deserialize downloaded attachments.
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
-enum OneOrMany<T> {
+pub enum OneOrMany<T> {
```
ncloudioj, line 117: Suggestion: how about renaming this to something more readable as it's now public and `SuggestAttachment<T>` has been re-purposed as a trait.

`OneOrMany` was used to model the payload for each RS attachment, which can be either a list of JSON values or a single JSON value. Maybe call it `AttachmentPayload`.

WDYT?

components/suggest/src/rs.rs
```diff
@@ -100,30 +100,44 @@ pub(crate) enum SuggestRecord {
...
 }
 
-/// A downloaded Remote Settings attachment that contains suggestions.
-#[derive(Clone, Debug, Deserialize)]
-#[serde(transparent)]
-pub(crate) struct SuggestAttachment<T>(OneOrMany<T>);
-
-impl<T> SuggestAttachment<T> {
+impl<T> OneOrMany<T> {
     /// Returns a slice of suggestions to ingest from the downloaded attachment.
```
ncloudioj, line 123: Unrelated to this patch, but this comment needs some updates as the attachment payload could be either suggestions or provider configurations now.

components/suggest/src/rs.rs
```diff
@@ -100,30 +100,44 @@ pub(crate) enum SuggestRecord {
...
+    pub fn as_slice(&self) -> &[T] {
+        match &self {
             OneOrMany::One(value) => std::slice::from_ref(value),
             OneOrMany::Many(values) => values,
         }
     }
 }
 
+impl<T: SuggestAttachment> OneOrMany<T> {
+    /// Did any of the attachments have an unknown field?
```
ncloudioj, line 133: Nit: as discussed above, `OneOrMany` would only involve one attachment, so maybe `Did any of the payloads (i.e. JSON values) in this attachment have an unknown field?`.

components/suggest/src/rs.rs
```diff
@@ -339,3 +369,48 @@ where
...
+impl SuggestAttachment for DownloadedAmpWikipediaSuggestion {
+    fn unknown_fields(&self) -> &serde_json::Map<String, serde_json::Value> {
+        match self {
+            Self::Amp(DownloadedAmpSuggestion { unknown_fields, .. }) => unknown_fields,
+            Self::Wikipedia(DownloadedWikipediaSuggestion { unknown_fields, .. }) => unknown_fields,
+        }
+    }
+}
+
+impl SuggestAttachment for DownloadedAmoSuggestion {
```
ncloudioj, line 382: Looks like most `Downloaded()Suggestionn` types share the same implementation, what do you think about providing a macro for that? Please disregard if macros are not a thing for you.

```rs
macro_rules! impl_suggest_attachment {
    ( 
      $(
        $t:ident
      ),+
    ) => {
        $(
          impl SuggestAttachment for $t {
             fn unknown_fields(&self) -> &serde_json::Map<String, serde_json::Value> {
                 &self.unknown_fields
             }
          }
        )+
    }
}

impl_suggest_attachment!(
  DownloadedAmoSuggestion,
  DownloadedPocketSuggestion,
  ...
);
```

components/suggest/src/store.rs
```diff
@@ -486,10 +490,13 @@ where
         };
 
         let attachment_data = self.settings_client.get_attachment(&attachment.location)?;
-        match serde_json::from_slice::<SuggestAttachment<T>>(&attachment_data) {
-            Ok(attachment) => self.ingest_record(writer, record, |dao, record_id| {
-                ingestion_handler(dao, record_id, attachment.suggestions())
-            }),
+        match serde_json::from_slice::<OneOrMany<T>>(&attachment_data) {
+            Ok(attachments) => self.ingest_record(
```
ncloudioj, line 494: Nit: calling those `attachments` might confuse the readers, `attachment_payload` could be better IMHO.

ncloudioj (commented):
Looks good, just a few comments for your consideration.

To clarify, this will mark the record as unparsable if we see any unknown fields in its attachment payload(s), though the recognized fields would still be ingested as the current behavior. Is that correct?

bendk:
> To clarify, this will mark the record as unparsable if we see any unknown fields in its attachment payload(s), though the recognized fields would still be ingested as the current behavior. Is that correct?

Yes that's how it should work.

components/suggest/src/rs.rs
bendk: I ended up adding the note about what items an attachment payload can have top-level docstring and removing this docstring altogether.  I think it's more like to remain correct going forward that way.

ncloudioj (approved):
LGTM, r=nanj.

This would allow us to "tolerate" more schema changes that don't necessarily require a version bump and an entire db re-ingestion. Very cool!

components/suggest/src/store.rs
```diff
@@ -405,7 +405,7 @@ where
                     let data = self.settings_client.get_attachment(&attachment.location)?;
                     writer.write(|dao| {
                         dao.put_icon(icon_id, &data)?;
-                        dao.handle_ingested_record(record)
+                        dao.handle_ingested_record(record, false)
```
linabutler, line 408: I wonder if we could make this an enum instead of a Boolean, or split up `handle_ingested_record` into `handle_ingested_record` and `handle_ingested_record_with_unknown_fields`?

`handle_ingested_record(record, false)` feels a bit too close to a ["Boolean trap"](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) to me—it's not super obvious from context what `false` means.

components/suggest/src/rs.rs
```diff
@@ -100,30 +100,45 @@ pub(crate) enum SuggestRecord {
     GlobalConfig(DownloadedGlobalConfig),
 }
 
-/// Represents either a single value, or a list of values. This is used to
-/// deserialize downloaded attachments.
+/// Implemented by the various attachment types
+pub trait SuggestAttachment: DeserializeOwned + std::fmt::Debug {
```
linabutler, line 104: This is a lovely use of traits!

linabutler (approved):
LGTM! One suggestion about changing the Boolean to an enum, but the rest looks great—thanks for doing this!

bendk:
I'm going to close this one because we decided to stop tracking unparsable records.  The current system where we throw away all suggestion records and re-download them when the schema changes is working well enough and the unparsable records code adds more complexity than we want.

PR #6111, opened 2024-02-06
Recieves session token from application before oauth for user agents

tarikeshaq:
Putting it up a little early, still needs:
- [x] unit tests,
- [x] fixing up swiftlint 
- [x] ios-tests
- [x] Two patches for iOS and Android
- [x] Changelog 

I'm adding those in a separate commit but I've tested this against both iOS and Android

## Context
User Agents such as Firefox iOS and Android, need access to a session token, this is so that they can issue access tokens, but also so they can load the settings page. This is a super power that usually only the web content would have.

Historically, well at least since the inception of the FxA client in Rust, the session token was delivered through a special OAuth scope. That works OK when the application is authenticating using pairing, however it leads to an interesting case when the application is authenticating using email/password. 

When signing in using email and password, the web content holds on to a session token, representing the web session the user just signed into, and additionally, the user agent would get a second session token through the special session scope. This leads to there being two seperate session tokens associated with the one authority (your firefox application). This easily reproducible, all you have to do is sign into Firefox Android or iOS using email and password, then go to manage account - you'll see two entries for the same device!

## This patch

This patch moves us so that we do what desktop does. Where desktop shares the same session token the content got when the user first signs in using email password

### DIFF
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -138,6 +138,11 @@ interface FirefoxAccount {
   [Throws=FxaError]
   string to_json();
   
+  // Sets the users information based on the web content's login information
+  // This is intended to only be used by user agents (eg: Firefox) to set the users
+  // session token and tie it to the refresh token that will be issued at the end of the
+  // oauth flow.
+  void set_user_data(UserData user_data);
 
   // Initiate a web-based OAuth sign-in flow.
   //
@@ -1097,4 +1102,11 @@ interface FxaStateCheckerState {
   GetProfile();
   Complete(FxaState new_state);
   Cancel();
+};
+
+dictionary UserData {
+  string session_token;
+  string uid;
+  string email;
+  boolean verified;
 };
--- a/components/fxa-client/src/internal/http_client.rs
+++ b/components/fxa-client/src/internal/http_client.rs
@@ -59,9 +59,10 @@ const DEVICES_FILTER_DAYS: u64 = 21;
 #[allow(clippy::needless_lifetimes)]
 #[cfg_attr(test, mockall::automock)]
 pub(crate) trait FxAClient {
-    fn create_refresh_token_using_authorization_code(
+    fn create_refresh_token_using_authorization_code<'a>(
         &self,
         config: &Config,
+        session_token: Option<&'a str>,
         code: &str,
         code_verifier: &str,
     ) -> Result<OAuthTokenResponse>;
@@ -200,6 +201,7 @@ impl FxAClient for Client {
     fn create_refresh_token_using_authorization_code(
         &self,
         config: &Config,
+        session_token: Option<&str>,
         code: &str,
         code_verifier: &str,
     ) -> Result<OAuthTokenResponse> {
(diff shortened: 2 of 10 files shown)


tarikeshaq:
The flaky test is because of a race condition in the unit tests for push, see https://github.com/mozilla/application-services/pull/6126. re-running the test here

components/fxa-client/src/internal/state_manager.rs
```diff
@@ -175,7 +175,9 @@ impl StateManager {
             self.persisted_state.scoped_keys.insert(scope, key);
         }
         self.persisted_state.refresh_token = Some(refresh_token);
-        self.persisted_state.session_token = session_token;
+        if let Some(session_token) = session_token {
+            self.persisted_state.session_token = Some(session_token)
+        }
```
bendk, line 180: Can you add a comment describing how this interacts with the session token from `set_user_data()`?  It seems like it would override it if set, but a comment would be great.

Also, maybe it would be more clear if the code was `self.persisted_state.session_token = session_token.or(self.persisted_state.session_token)`.  I don't feel strongly about that though.

components/fxa-client/src/internal/http_client.rs
```diff
@@ -479,10 +485,20 @@ impl Client {
...
     ) -> Result<OAuthTokenResponse> {
         let url = config.token_endpoint()?;
-        Ok(self.make_request(Request::post(url).json(&body))?.json()?)
+        if let Some(session_token) = session_token {
+            let key = derive_auth_key_from_session_token(session_token)?;
+            let request = HawkRequestBuilder::new(Method::Post, url, &key)
+                .body(body)
+                .build()?;
+
+            Ok(self.make_request(request)?.json()?)
```
bendk, line 498: What will be in the `session_token` in the return value in this case?

components/fxa-client/src/fxa_client.udl
```diff
@@ -138,6 +138,11 @@ interface FirefoxAccount {
   [Throws=FxaError]
   string to_json();
   
+  // Sets the users information based on the web content's login information
+  // This is intended to only be used by user agents (eg: Firefox) to set the users
+  // session token and tie it to the refresh token that will be issued at the end of the
+  // oauth flow.
+  void set_user_data(UserData user_data);
```
bendk, line 145: I'm not sure, but I kind of think this data should be sent via a state machine event.  Could it be an optional field on `FxaEvent::CompleteOAuthFlow`?  Alternatively, could there be a `FxaEvent::StoreUserData` event?

components/fxa-client/src/internal/http_client.rs
bendk: I think that it shouldn't.  In that case, I think we should add a check that `session_token=None`.  I really want to avoid overwriting it here.

bendk (changes requested):
This looks great to me.  My only real concern is having an extra check that we don't overwrite the session token.  Integrating this with the state machine could be nice, but that can be done later.

components/fxa-client/src/fxa_client.udl
tarikeshaq: We chatted a bit about this and decided that it would be a nice followup but not necessary here - resolving!

components/fxa-client/src/internal/http_client.rs
tarikeshaq: Yeah that's a good idea, I'm pushing up a change to make it explicit that we prioritize the existing session token

components/fxa-client/src/internal/state_manager.rs
bendk: The new version of this is super clear thanks!

bendk (approved):

PR #6114, opened 2024-02-08
Bug 1878735 - Ingest Yelp icon from Remote Settings

ncloudioj:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -1071,6 +1071,10 @@ impl<'a> SuggestDao<'a> {
             "DELETE FROM yelp_location_signs WHERE record_id = :record_id",
             named_params! { ":record_id": record_id.as_str() },
         )?;
+        self.conn.execute_cached(
+            "DELETE FROM yelp_custom_details WHERE record_id = :record_id",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
         Ok(())
     }
 
--- a/components/suggest/src/schema.rs
+++ b/components/suggest/src/schema.rs
@@ -92,6 +92,11 @@ pub const SQL: &str = "
         record_id TEXT NOT NULL
     ) WITHOUT ROWID;
 
+    CREATE TABLE yelp_custom_details(
+        icon_id TEXT PRIMARY KEY,
+        record_id TEXT NOT NULL
+    ) WITHOUT ROWID;
+
     CREATE TABLE mdn_custom_details(
         suggestion_id INTEGER PRIMARY KEY,
         description TEXT NOT NULL,
--- a/components/suggest/src/suggest.udl
+++ b/components/suggest/src/suggest.udl
@@ -73,7 +73,8 @@ interface Suggestion {
     Yelp(
         string url,
         string title,
-        boolean is_top_pick
+        boolean is_top_pick,
+        sequence<u8>? icon
     );
     Mdn(
         string title,
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -294,6 +294,8 @@ pub(crate) struct DownloadedYelpSuggestion {
     pub location_signs: Vec<DownloadedYelpLocationSign>,
     #[serde(rename = "yelpModifiers")]
     pub yelp_modifiers: Vec<String>,
+    #[serde(rename = "icon")]
+    pub icon_id: String,
 }
 
 /// An MDN suggestion to ingest from an attachment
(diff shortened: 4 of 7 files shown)


components/suggest/src/schema.rs
```diff
@@ -92,6 +92,11 @@ pub const SQL: &str = "
         record_id TEXT NOT NULL
     ) WITHOUT ROWID;
 
+    CREATE TABLE yelp_custom_details(
```
ncloudioj, line 95: Unlike other suggestions, we build Yelp suggestions on the fly, so there is no records in the `suggestions` table for Yelp. Thus, we don't have the `suggestion_id` in this table. Also, there should be only one record in this table, which feels odd. As discussed in another thread, maybe we can move the `icon_id` to the `suggestion_providers` table in the future.

components/suggest/src/db.rs
```diff
@@ -1071,6 +1071,10 @@ impl<'a> SuggestDao<'a> {
             "DELETE FROM yelp_location_signs WHERE record_id = :record_id",
             named_params! { ":record_id": record_id.as_str() },
         )?;
+        self.conn.execute_cached(
+            "DELETE FROM yelp_custom_details WHERE record_id = :record_id",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
```
0c0w3, line 1077: Does the icon data in `icons` need to be deleted too? Does that happen somewhere else?

0c0w3 (approved):
Looks good, thanks again for doing this!

Repeating the icon bytes in the test is kind of too bad -- would it be reasonable to factor it out into a const or something?

components/suggest/src/db.rs
ncloudioj: Yup, icon deletion happens [here](https://github.com/mozilla/application-services/blob/a757c1ec4083259622e366ea4ce5a2c2189b2088/components/suggest/src/db.rs#L167).

ncloudioj:
> Repeating the icon bytes in the test is kind of too bad -- would it be reasonable to factor it out into a const or something?

I know, it's kinda annoying. But the entire component is stateless, so I couldn't find a better way to cache that to avoid repeated loads.

@linabutler @bendk Wonder if you have better recommendations?

bendk:
> > Repeating the icon bytes in the test is kind of too bad -- would it be reasonable to factor it out into a const or something?
> 
> I know, it's kinda annoying. But the entire component is stateless, so I couldn't find a better way to cache that to avoid repeated loads. This also happens to Wikipedia, we store 7,000 icon ids (they are identical) in `wikipedia_custom_details`! And we load the icon from db for every single hit of Wikipedia.
> 
> @linabutler @bendk Wonder if you have better recommendations?

I think you're talking about two different things.

Drew's comment seems to be about these [lines in the tests](https://github.com/mozilla/application-services/blob/0b51ee950b799f1a4c193aa0ad7d4d1eee855a8f/components/suggest/src/store.rs#L816-L830).  Could that `icon` value be factored out into a const at the top to avoid the repetition?

Nan's comment seems to be about the fact that icon data might be loaded multiple times, even though it's easily cachable.  I guess we could start storing some cache in the component somewhere, but it doesn't seem that high-priority to me.  My guess is that SQLite + the disk cache will be fast enough.

ncloudioj:
> I think you're talking about two different things.

Oops, sorry, I misread Drew's comment. Those test fixtures are automatically populated and managed by [expect_test](https://docs.rs/expect-test/latest/expect_test/). So I don't think we can swap in some constants to minimize those unfortunately.

> I guess we could start storing some cache in the component somewhere, but it doesn't seem that high-priority to me. My guess is that SQLite + the disk cache will be fast enough.

Agreed, no need to worry too much right now. Just be mindful about the icon size (~10KB across all the providers), repeatedly loading those from the db might be sub-optimal.

dadaa (approved):
Thank you very much, @ncloudioj! LGTM!

components/suggest/src/suggest.udl
```diff
@@ -73,7 +73,8 @@ interface Suggestion {
     Yelp(
         string url,
         string title,
-        boolean is_top_pick
+        boolean is_top_pick,
+        sequence<u8>? icon
```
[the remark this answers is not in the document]
ncloudioj: Yes, I think so. While the `icon` is a required field for `DownloadedYelpAttachment`, it's just an ID pointing to the icon record and we can't guarantee that icon's present on RS (e.g. someone could delete it by accident).

PR #6118, opened 2024-02-09
Use single name field for addresses [firefox-android: issammani:feat/use-single-name-field-address]

issammani:
This PR:
 - Ports [FormAutofillNameUtils.sys.mjs](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/shared/FormAutofillNameUtils.sys.mjs) with tests to make sure the implementations behave the same way.
 - Ports [this patch](https://phabricator.services.mozilla.com/D197658) from desktop so we can have consistent logic for handling missing `name` fields. This also includes tests.
 - Migrates db version to v3.
 - Exposes a single `name` field to consumers  instead of 3 separate `*-name` fields.
  
     - There are corresponding PRs for our consumer applications that resolve the breaking changes and have been approved:
     - https://github.com/mozilla-mobile/firefox-ios/pull/18690 ( approved )
     - https://github.com/mozilla-mobile/firefox-android/pull/5586 ( approved )
  - Note:


components/autofill/src/sync/address/mod.rs
```diff
@@ -75,6 +77,9 @@ pub struct AddressPayload {
 #[derive(Default, Deserialize, Serialize)]
 #[serde(default, rename_all = "kebab-case")]
 struct PayloadEntry {
+    pub name: String,
+    // given_name, additional_name and family_name
+    // should exist in PayloadEntry but not in InternalAddress
```
mhammond, line 82: maybe expand this comment to explain it exists for "old" clients?

components/autofill/src/sync/tests/test_reconcile.rs
```diff
@@ -502,16 +502,16 @@ lazy_static::lazy_static! {
         ],
         "remote": {
             "version": 1,
```
mhammond, line 504: It seems like some of these test cases should exercise this name handling - eg, we can expect that the "mirror" (ie, the parent) will already hold records without name but with the 3 split names. Similarly, some incoming records will be from an old device (no combined name) or from a new device (both combined and split names), some tests with those variations makes sense.

mhammond (commented):
I'm going to need a little more time to look at this, but this is a nice patch and I don't see anything fundamental which will block this. It's safe to assume that the changes to the UDL will land as they stand, so I think you can start to consider the mobile consumers in the meantime. Ping me in slack if you need any assistance there.

components/autofill/src/sync/address/incoming.rs
```diff
@@ -18,10 +19,50 @@ use rusqlite::{named_params, Transaction};
 use sql_support::ConnExt;
 use sync_guid::Guid as SyncGuid;
 
-// Takes a raw payload, as stored in our database, and returns an InternalAddress
-// or a tombstone. Addresses store the raw payload as cleartext json.
-fn raw_payload_to_incoming(id: SyncGuid, raw: String) -> Result<IncomingContent<InternalAddress>> {
-    // Make an IncomingBso from the payload.
+fn update_name(payload_content: &mut IncomingContent<AddressPayload>, local_name: String) {
```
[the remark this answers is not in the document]
mhammond: (actually, I think my comment above is wrong - it's not "before storing in the mirror", which is great! As you can see, I haven't fully got my head around it yet :)

lougeniaC64:
@issammani Disregard if you and @mhammond have already discussed this, but I'm curious what version you are targeting for these changes and how any necessary consumer changes will be handled. Also gonna cc @skhamis since he's been working with the iOS team on implementing address syncing.

issammani:
> @issammani Disregard if you and @mhammond have already discussed this, but I'm curious what version you are targeting for these changes and how any necessary consumer changes will be handled. Also gonna cc @skhamis since he's been working with the iOS team on implementing address syncing.

Hey @lougeniaC64 , I am targeting 125 for now. I have opened PRs for android and iOS that support this change. I linked them in the description.

components/autofill/src/autofill.udl
```diff
@@ -43,9 +43,7 @@ dictionary CreditCard {
 
 // What you pass to create or update an address.
 dictionary UpdatableAddressFields {
-    string given_name;
-    string additional_name;
-    string family_name;
+    string name;
```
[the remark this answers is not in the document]
issammani: I have added both PRs to the description. They are still in review for now. Thanks !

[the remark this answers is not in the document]
issammani: I also added  `[firefox-android: issammani:feat/use-single-name-field-address]` to the title but not sure if it triggered the workflow or not.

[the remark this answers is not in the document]
mhammond: I believe you need to do another push for this to happen, which will probably happen organically anyway.

components/autofill/src/sync/tests/test_reconcile.rs
issammani: We have these in a separate test file here. For example where the mirror is old and using split names. Let know if you feel like we need to add more cases here.
https://github.com/mozilla/application-services/blob/804f0fb2ecef3d1e02320085f148f6df8c859c93/components/autofill/src/sync/tests/test_migrate_remote_address.rs#L129-L151

components/autofill/src/sync/address/mod.rs
issammani: updated in https://github.com/mozilla/application-services/pull/6118/commits/340d170c26936bfe2be5a108ae8301ba6f7d7ce9

components/autofill/src/sync/address/incoming.rs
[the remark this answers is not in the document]
issammani: Updated in https://github.com/mozilla/application-services/pull/6118/commits/340d170c26936bfe2be5a108ae8301ba6f7d7ce9 with a more comprehensive example

components/autofill/src/sync/tests/test_reconcile.rs
mhammond: Fair enough - how about a command at the top of this huge set explaining that?

mhammond (dismissed):
Sorry for the delay, this is great work!

components/autofill/src/sync/tests/test_reconcile.rs
issammani: Done in [921c0ea](https://github.com/mozilla/application-services/pull/6118/commits/921c0ea2feecf9ee84ba39d658307e00510bd79c). I added a comment at the top with the different scenarios also added comments to each test case explaining the testcase itself 😄

issammani: ( also not sure what the preferred workflow here is but it seems after I rebased onto main the previous review from you was dismissed 🤔 )

mhammond: This happens because you are outside the org. If you get in the mozilla org I can add you as a contributor to the repo.

mhammond (approved):

PR #6120, opened 2024-02-10
Bug 1875909 - Suggest: Add support for mobile AMP suggestions

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -21,12 +21,13 @@ use crate::{
     pocket::{split_keyword, KeywordConfidence},
     provider::SuggestionProvider,
     rs::{
-        DownloadedAmoSuggestion, DownloadedAmpWikipediaSuggestion, DownloadedMdnSuggestion,
-        DownloadedPocketSuggestion, DownloadedWeatherData, SuggestRecordId,
+        DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedAmpWikipediaSuggestion,
+        DownloadedMdnSuggestion, DownloadedPocketSuggestion, DownloadedWeatherData,
+        SuggestRecordId,
     },
     schema::{SuggestConnectionInitializer, VERSION},
     store::{UnparsableRecord, UnparsableRecords},
-    suggestion::{cook_raw_suggestion_url, Suggestion},
+    suggestion::{cook_raw_suggestion_url, AmpSuggestionType, Suggestion},
     Result, SuggestionQuery,
 };
 
@@ -186,7 +187,12 @@ impl<'a> SuggestDao<'a> {
             .iter()
             .try_fold(vec![], |mut acc, provider| {
                 let suggestions = match provider {
-                    SuggestionProvider::Amp => self.fetch_amp_suggestions(query),
+                    SuggestionProvider::Amp => {
+                        self.fetch_amp_suggestions(query, AmpSuggestionType::Desktop)
+                    }
+                    SuggestionProvider::AmpMobile => {
+                        self.fetch_amp_suggestions(query, AmpSuggestionType::Mobile)
+                    }
                     SuggestionProvider::Wikipedia => self.fetch_wikipedia_suggestions(query),
                     SuggestionProvider::Amo => self.fetch_amo_suggestions(query),
                     SuggestionProvider::Pocket => self.fetch_pocket_suggestions(query),
@@ -207,8 +213,16 @@ impl<'a> SuggestDao<'a> {
     }
 
     /// Fetches Suggestions of type Amp provider that match the given query
-    pub fn fetch_amp_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+    pub fn fetch_amp_suggestions(
+        &self,
+        query: &SuggestionQuery,
+        suggestion_type: AmpSuggestionType,
+    ) -> Result<Vec<Suggestion>> {
         let keyword_lowercased = &query.keyword.to_lowercase();
+        let provider = match suggestion_type {
+            AmpSuggestionType::Mobile => SuggestionProvider::AmpMobile,
+            AmpSuggestionType::Desktop => SuggestionProvider::Amp,
+        };
         let suggestions = self.conn.query_rows_and_then_cached(
             r#"
             SELECT
@@ -229,7 +243,7 @@ impl<'a> SuggestDao<'a> {
             "#,
             named_params! {
                 ":keyword": keyword_lowercased,
-                ":provider": SuggestionProvider::Amp
+                ":provider": provider
             },
             |row| -> Result<Suggestion> {
                 let suggestion_id: i64 = row.get("id")?;
@@ -276,6 +290,7 @@ impl<'a> SuggestDao<'a> {
                         let cooked_url = cook_raw_suggestion_url(&raw_url);
                         let raw_click_url = row.get::<_, String>("click_url")?;
                         let cooked_click_url = cook_raw_suggestion_url(&raw_click_url);
+
                         Ok(Suggestion::Amp {
                             block_id: row.get("block_id")?,
                             advertiser: row.get("advertiser")?,
@@ -736,6 +751,7 @@ impl<'a> SuggestDao<'a> {
             self.scope.err_if_interrupted()?;
             let common_details = suggestion.common_details();
             let provider = suggestion.provider();
+
             let suggestion_id: i64 = self.conn.query_row_and_then_cachable(
                 &format!(
                     "INSERT INTO suggestions(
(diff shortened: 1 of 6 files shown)


components/suggest/src/db.rs
```diff
@@ -186,7 +186,9 @@ impl<'a> SuggestDao<'a> {
             .iter()
             .try_fold(vec![], |mut acc, provider| {
                 let suggestions = match provider {
-                    SuggestionProvider::Amp => self.fetch_amp_suggestions(query),
+                    SuggestionProvider::Amp | SuggestionProvider::AmpMobile => {
+                        self.fetch_amp_suggestions(query, **provider)
+                    }
```
tiftran, line 191: So I tried my best to recycle the existing bits as much as I can, but as a result, this look a little goofy since at the point of `fetch_amp_suggestions` we don't know which provider it came from.

components/suggest/src/db.rs
```diff
@@ -731,11 +754,16 @@ impl<'a> SuggestDao<'a> {
...
     ) -> Result<()> {
         for suggestion in suggestions {
             self.scope.err_if_interrupted()?;
             let common_details = suggestion.common_details();
-            let provider = suggestion.provider();
+            let provider = if let SuggestRecord::AmpMobile = fields {
+                SuggestionProvider::AmpMobile
+            } else {
+                suggestion.provider()
+            };
```
tiftran, line 766: the clunky-ness continues here, since  `suggestion.providers()` differentiates amp and wikipedia by the advertiser (I think), and AMP mobile and AMP are the exact same in terms of structure?

components/suggest/src/db.rs
```diff
@@ -731,11 +754,16 @@ impl<'a> SuggestDao<'a> {
         &mut self,
         record_id: &SuggestRecordId,
         suggestions: &[DownloadedAmpWikipediaSuggestion],
+        fields: SuggestRecord,
```
ncloudioj, line 757: Nit: since this is only used to determine if the record is of type `amp_mobile_suggestions`, we can just pass in a boolean `is_amp_mobile`.

```suggestion
        is_amp_mobile: bool,
```

```rs
dao.insert_amp_wikipedia_suggestions(
   record_id,
   suggestions,
   matches!(fields, SuggestRecrod::AmpMobile)
);
```

components/suggest/src/suggestion.rs
```diff
@@ -33,6 +33,20 @@ pub enum Suggestion {
         raw_click_url: String,
         score: f64,
     },
+    AmpMobile {
```
ncloudioj, line 36: Since AMP suggestions share the same structure on desktop and mobile, I think we can use `Amp` for both. Let's add a doc string to to the `Amp` variant to note this.

components/suggest/src/db.rs
```diff
@@ -276,20 +282,37 @@ impl<'a> SuggestDao<'a> {
...
-                            raw_url,
-                            full_keyword: full_keyword(keyword_lowercased, &keywords),
-                            icon: row.get("icon")?,
-                            impression_url: row.get("impression_url")?,
-                            click_url: cooked_click_url,
-                            raw_click_url,
-                            score,
-                        })
+                        if provider == SuggestionProvider::Amp {
+                            Ok(Suggestion::Amp {
```
ncloudioj, line 286: As noted below, we can use `Suggestion::Amp` for both mobile and desktop.

components/suggest/src/store.rs
```diff
@@ -3558,6 +3558,191 @@ mod tests {
...
+                        score: 0.3,
+                    },
+                ]
+                "#]],
+            ),
+            (
+                "keyword = `las `; amp and amp mobile",
+                SuggestionQuery {
+                    keyword: "las".into(),
+                    providers: vec![SuggestionProvider::Amp, SuggestionProvider::AmpMobile],
```
ncloudioj, line 3698: Note that, in practice, consumers would only issue queries against either of those two providers, and it's up to the consumers to enforce that. But I guess having a test here to query both is fine for completeness.

ncloudioj (changes requested):
Looks good, just a few comments for simplification.

components/suggest/src/suggestion.rs
linabutler: Hmmm, I'm not sure about this one! Right now, there's a 1:1 correspondence between the variants of `SuggestionProvider` and `Suggestion`—so if you ask for `SuggestionProvider::Amp` and `SuggestionProvider::Wikipedia`, you'll get `Suggestion::Amp` and `Suggestion::Wikipedia` back.

Even though the AMP mobile suggestions have the same structure, I think passing `SuggestionProvider::AmpMobile` and `SuggestionProvider::Wikipedia`, and getting `Suggestion::Amp` and `Suggestion::Wikipedia` back, feels a bit too special case-y for me 🤔

components/suggest/src/db.rs
linabutler: I might actually suggest having a separate `dao.insert_amp_mobile_suggestions` method, at the risk of duplicating a bit of code!

The reason I'd suggest that is that AMP Desktop and Wikipedia suggestions are commingled in the same attachment, _but_ AMP Mobile only has mobile suggestions! So maybe we can have a method on `SuggestDao` like:

```rs
fn insert_amp_mobile_suggestions(&mut self, record_id: &SuggestRecordId, suggestions: &[DownloadedAmpSuggestion]) -> Result<()>
```

...That doesn't have to worry about Wikipedia.

components/suggest/src/db.rs
linabutler: Yeah, I think that's why I'm leaning toward having a separate `insert_amp_mobile_suggestions` method. `insert_amp_wikipedia_suggestions` has to disentangle AMP Desktop and Wikipedia suggestions, and with the addition of AMP Mobile, this method now handles three different cases. A separate method would bring it down to two.

linabutler (changes requested):
Almost there! My suggestions are a little different than @ncloudioj's, but I'm happy to discuss them more, and open to having my mind changed! ☺️

* I think we can keep `Suggestion::AmpMobile` and `Suggestion::Amp` separate.
* I'd suggest adding a new `AmpSuggestionType` enum to `db.rs`, and using it in `fetch_amp_suggestions`.
* I'd also suggest adding a separate `SuggestDao::insert_amp_mobile_suggestions` method, because AMP Mobile doesn't need to untangle Wikipedia suggestions.

components/suggest/src/suggestion.rs
linabutler: I thought a bit more about this one, and now I like @ncloudioj's suggestion more :joy:

There's no reason why a `SuggestionProvider` couldn't return different types of `Suggestion`s, and I could see us doing this in the future.

components/suggest/src/db.rs
```diff
@@ -207,7 +209,11 @@ impl<'a> SuggestDao<'a> {
     }
 
     /// Fetches Suggestions of type Amp provider that match the given query
-    pub fn fetch_amp_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+    pub fn fetch_amp_suggestions(
+        &self,
+        query: &SuggestionQuery,
+        provider: SuggestionProvider,
```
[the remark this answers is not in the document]
ncloudioj: YES! +1. Let's [make-invalid-states-unrepresentable](https://geeklaunch.io/blog/make-invalid-states-unrepresentable/)!

Actually, I was gonna suggest adding this exact enum `AmpSuggestionType` to the `Provider` so that consumers have to pick either variant (but not both!) when they query. Then I realized that they could still "violate" that as they could pass both `Mobile` and `Desktop` to `SuggestionQuery::providers` 😂.

components/suggest/src/db.rs
ncloudioj: Ticket for the split: https://bugzilla.mozilla.org/show_bug.cgi?id=1879962

components/suggest/src/db.rs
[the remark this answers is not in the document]
linabutler: Oooh, thank you for that link, it's going into my bookmarks for sure! 🏆 I think [this Aphyr article](https://aphyr.com/posts/341-hexing-the-technical-interview) (it's a _read_! :joy:) was the first time I heard that expression:

> “But… those aren’t even the same type. That’s… that’s illegal.”
>
> “If it were meant to be illegal,” you remind him sagely, “Sun Microsystems would have made it unrepresentable.”

I liked your second point, too:

> Then I realized that they could still "violate" that as they could pass both `Mobile` and `Desktop` to `SuggestionQuery::providers`.

Yeah, it's tricky; I think this is the first time that we're segmenting suggestions by platform like this! I'd be curious to see if we end up with similar cases in the future—and if a more fluent way of expressing this separation emerges from those other cases! For now, it feels like the list of providers is the most sensible abstraction we have, but that might change as we add providers and new types of filters.

tiftran:
okie doke, i think we ended up converging in agreement with changes, if i read everything correctly.

- added an enum, for desktop and mobile
- removed `AmpMobile` as a suggestion type
- added `SuggestDao::insert_amp_mobile_suggestions`

components/suggest/src/suggestion.rs
```diff
@@ -16,6 +16,11 @@ const TIMESTAMP_TEMPLATE: &str = "%YYYYMMDDHH%";
 /// 2 bytes shorter than [`TIMESTAMP_TEMPLATE`].
 const TIMESTAMP_LENGTH: usize = 10;
 
+/// Suggestion Types for Amp
+pub enum AmpSuggestionType {
```
linabutler, line 20:
```suggestion
pub(crate) enum AmpSuggestionType {
```

Nit: Let's stick to `pub` for things that we expose in the UDL, and `pub(crate)` for things that we use within the crate.

components/suggest/src/store.rs
```diff
@@ -3999,6 +4004,191 @@ mod tests {
...
+            "data-2.json",
+            json!([
+              {
+                  "id": 0,
+                  "advertiser": "Good Place Eats",
+                  "iab_category": "8 - Food & Drink",
+                  "keywords": ["la", "las", "lasa", "lasagna", "lasagna come out tomorrow"],
+                  "title": "Desktop - Lasagna Come Out Tomorrow",
+                  "url": "https://www.lasagna.restaurant",
+                  "icon": "2",
```
linabutler, line 4074:
```suggestion
                  "icon": "3",
```

components/suggest/src/store.rs
```diff
@@ -3999,6 +4004,191 @@ mod tests {
...
+            "data-1.json",
+            json!([
+               {
+                   "id": 0,
+                   "advertiser": "Good Place Eats",
+                   "iab_category": "8 - Food & Drink",
+                   "keywords": ["la", "las", "lasa", "lasagna", "lasagna come out tomorrow"],
+                   "title": "Mobile - Lasagna Come Out Tomorrow",
+                   "url": "https://www.lasagna.restaurant",
+                   "icon": "2",
```
linabutler, line 4057:
```suggestion
                   "icon": "3",
```

Just to make sure that we ingest icons correctly for mobile, too ☺️

linabutler (approved):
Looks great, thanks! :shipit:

ncloudioj (approved):
LGTM, thanks!

Once AMP mobile suggestions are ready in production (pending on AMP's confirmation), we will need two tiny client changes to swap in the new AMP mobile provider.

PR #6121, opened 2024-02-10
Bug 1879640 - Suggest: Update Yelp prefix matching

0c0w3:

  - Note:

### DIFF
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -47,8 +47,8 @@ const MAX_QUERY_LENGTH: usize = 150;
 /// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
 const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
 
-/// The threshold that enables prefix-match.
-const PREFIX_MATCH_THRESHOLD: usize = 6;
+/// At least this many characters must be typed for a subject to be matched.
+const SUBJECT_PREFIX_MATCH_THRESHOLD: usize = 2;
 
 impl<'a> SuggestDao<'a> {
     /// Inserts the suggestions for Yelp attachment into the database.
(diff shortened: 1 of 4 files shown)


0c0w3:
The only substantive change is to lower the subject prefix threshold from 6 to 2. I talked with Loren and we came up with these rules for prefix matching:

* The entire search string has to be at least 5 characters long, **and**
* At least 2 characters of the subject have to be typed

The 5 char overall limit needs to be configurable. We've added a Nimbus variable for it and desktop will enforce it. Maybe we can also add some Yelp RS config and have the Rust component enforce it, but as long as we have a Nimbus variable, the client will need to do it.

I renamed `top_pick` back to @dadaa's original `subject_exact_match`. Desktop will use this to bypass the 5 char limit for subjects that are shorter than 5 chars. `top_pick` doesn't fit this use case and is too prescriptive.

components/suggest/src/yelp.rs
```diff
@@ -360,7 +360,11 @@ impl<'a> SuggestDao<'a> {
         // Otherwise, apply prefix-match.
         Ok(
             match self.conn.query_row_and_then_cachable(
-                "SELECT keyword FROM yelp_subjects WHERE keyword BETWEEN :candidate AND :candidate || x'FFFF' ORDER BY LENGTH(keyword) ASC LIMIT 1",
+                "SELECT keyword
+                 FROM yelp_subjects
+                 WHERE keyword BETWEEN :candidate AND :candidate || x'FFFF'
+                 ORDER BY LENGTH(keyword) ASC, keyword ASC
```
0c0w3, line 366: The only substantive change here is to add a backstop `keyword ASC` to the `ORDER BY` to ensure a consistent ordering for keywords that are the same length.

dadaa (approved):
Thanks!

ncloudioj (approved):
LGTM!

PR #6131, opened 2024-02-14
Proposed relevancy component work for the validation phase

bendk:
This is how I'm thinking we could handle the validation phase.  Do you think it will work?  Am I missing any important details?

  - Note:


components/relevancy/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        todo!()
+    }
+
+    /// Get the highest 3 interests by score.
+    ///
+    /// To calculate the scores:
+    ///   - Take the interest vector from `ingest()`
+    ///   - Find the maximum 3 values, by raw numbers (excluding Interest::Unknown)
+    ///   - Normalize the value by dividing by the norm of the vector
+    ///      - The norm is `sqrt(sum([n*n for n in vec]))`
```
bendk, line 70: If I understood the wikipedia article correctly, this is taking the [cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity) between the entire interest vector and a unit vector in the direction that one interest.

components/relevancy/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// An interest combined with a relevancy score
+///
+/// Scores will be [0-10].  Consumers would use this score to alter the order of their suggestions.
+///
+/// For example, consumers may assign a default score of 5 to sponsored suggestions and 0 to
+/// Wikipedia suggestions.  By default, the sponsored suggestions will show up over Wikipedia
+/// suggestions, but if a Wikipedia suggestion had a interest that was highly relevant for the
+/// user, that suggestion would show up first instead.
+#[derive(uniffi::Record)]
```
bendk, line 30: I was thinking about trying to test out the new UniFFI proc-macros, but on second thought that's going to cause a lot of issues on Desktop.  I don't actually know how we're going to be able to get those to work with the FF build system and the current JS bindings implementation.

In general I'd like to try out some newer UniFFI features with this component, since there's not that much risk.  If we run into some new UniFFI bug and that causes us to lose some experiment data, that should be fine.  

However, we should prioritize getting the experiment out over quickly over testing out UniFFI features.  So if some UniFFI feature is going to require work on the `gecko-js` bindings, let's not delay the experiments.

components/relevancy/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+///   - If possible, I'd also like to have make "about::relevancy" where users can go to see the
+///     results themselves.  This way we can pair the Glean metrics with asking people if they
+///     think the interest rankings seem correct to them.
+#[derive(uniffi::Object)]
+pub struct RelavancyStore {
+}
+
+#[uniffi::export]
+impl RelavancyStore {
+    /// Ingest relevancy data from remote settings
```
bendk, line 52: Here's my proposal for storing the data in remote settings:

- The data is stored in a binary attachment
- In that file, for each interest we store:
   - The count of URLs for that interest (32 bits)
   - For each URL, we store a SHA-256 hash of that URL
- In the Rust component, we build a `HashSet<u8[32], Topic>` to map URL hashes to topics.  (Assuming URLs, can only have 1 topic.  If not, then we can use `Vec<Topic>` as the value).
  
This is similar to the bloom filter approach, but the hashing is simpler.  If we use a bloom filter, then I think it would be somewhat tricky to get the server and the component to agree on the hashing system.  SHA-256 avoids those issues.

I don't know how to think about the size differences, but I believe it should be similar.  Lookups will be slower, but I think a `HashSet` should perform well enough for our needs.

bendk: (Maybe MD5 would be be better, since it outputs 128 bit value, which we could store in a `u128`).  The collision chance is still very low, and if we did hit a collision that wouldn't be the end of the world -- probably it's a negligible loss in accuracy).

ncloudioj: I think that's a solid proposal for now, love it!

The current dataset, each domain could have more than 1 category, but we can simplify that by only assigning one category (the good news is we also have a score for each category, so we could choose the one with the higher score).

Using MD5 is totally fine, actually we encode URLs via MD5 extensively in Desktop. This is quite similar to how we store domains/categories for the SERP search term categorization project. Expect a GH invite to that repo so that you can see more details there.

Lastly, I think we might want to add another field on RS indicating what part of URLs used (host + TLD or subdomain + host + TLD) so that when we classify interests, we can extract the right parts as well.

components/relevancy/src/lib.rs
ncloudioj: Ah, cosine similarities! Though I am a little confused that we normally use cosine similarities to compare two vectors, looks like we only one vector here, correct? At the first glance, I thought we wanted to use that to calculate the relevance score for some categorized content against user's interests (so that we have two vectors).

There are a couple of ways to normalize a category hit vector into scores, such as arithmetic means or [soft-max](https://en.wikipedia.org/wiki/Softmax_function).

ncloudioj (commented):
This looks good, I think we can incorporate those comments into the doc as well :)

components/relevancy/src/lib.rs
bendk: This comes from an assumption that consumers would only want to specify a single interest for a suggestion, so we would have a vector where one element was `1` and all others were `0` (not sure what to call this, maybe a "single-entry vector").    The question I wanted to explore was what would be the maximum possible cosine similarities when comparing those vectors against the user's interest vector.

Maybe that's not the best approach though.  For one thing, I think you're right that consumers might want to specify more complex vectors for their content -- maybe we give them an easy way to specify a single-entry vector but also a way to specify all components.

Based on that, maybe there are some other ways to measure the strength of the interest detection:
  - Measure the cosine similarity between the user's interest vector and the "non-unknown vector" (constructed by using `1` for all interests, except unknown which is `0`.
  - Measure the maximum cosine similarity that content could possibly have (I think you can get this by copying the raw counts from the user's interest vector, but setting the unknown element to `0`).

Are there other ways to measure this?  How do you think we should approach it?

> There are a couple of ways to normalize a category hit vector into scores, such as arithmetic means or [soft-max](https://en.wikipedia.org/wiki/Softmax_function).

Thanks!  I'll look into those and probably switch to using one of those.

ncloudioj: Ah, I see. You're already thinking about relevance metrics!

Yup, i think we call those "single-entry vector" or "unit vector". Likewise, [one-hot encoding](https://en.wikipedia.org/wiki/One-hot) is also popular to convert categorical data into numerical vectors.

There are also many other similarity metrics which do not need encoding, such as [Jaccard_index](https://en.wikipedia.org/wiki/Jaccard_index). All in all, we have lots of options (including the cosine one). I think we can pick a few to examine and experiment with.

> This comes from an assumption that consumers would only want to specify a single interest for a suggestion, so we would have a vector where one element was 1 and all others were 0.

It depends. For suggest, yes, it's more common to assign one category to each suggestion type. Whereas for other content such as Pocket stories on New Tab, they normally come with more than 1 category. So we might want to design for more generic cases for similarity calculation.

bendk: I'm going to need to do some more research on all of this, but I updated `metrics.yaml` with my current plan:
- Calculate 3 measures of cosine similarity, by comparing the user's interest vector with:
   - The single-entry vector that's closest to the user's interest vector
   - The multiple-entry vector that's closest to the user's interest vector (but with `0` for the inconclusive component)
   - The "non-inconclusive" vector
- Multiply by 100, to make it an integer for Glean, but don't try to normalize.   We can figure that out later.

bendk:
Thanks for all the feedback here, I used it to open https://github.com/mozilla/application-services/pull/6152/ which I'm hoping can be a starting point for our work.

PR #6146, opened 2024-02-27
Bug 1882381 - Switch to using the data path for the suggest DB

bendk:
- Use the `data_path` rather than `cache_path` for the suggest DB.  This is prep for for storing the suggestion dismissal data in the DB, which should not be reset on schema upgrades.
- Don't always drop and recreate the database when the schema upgrades.  Instead, I'm hoping we can use the code from `SuggestDao.clear` to delete the suggestion data so that we re-ingest it.
- Other than adding the `dismissed_suggestion` table, this doesn't implement any of the suggestion dismissal functionality.

  - Note:

### DIFF
--- a/components/suggest/src/schema.rs
+++ b/components/suggest/src/schema.rs
@@ -13,7 +13,9 @@ use sql_support::open_database::{self, ConnectionInitializer};
 ///  1. Bump this version.
 ///  2. Add a migration from the old version to the new version in
 ///     [`SuggestConnectionInitializer::upgrade_from`].
-pub const VERSION: u32 = 16;
+///    a. If suggestions should be re-ingested after the migration, call `clear_database()` inside
+///       the migration.
+pub const VERSION: u32 = 17;
 
 /// The current Suggest database schema.
 pub const SQL: &str = "
@@ -121,6 +123,12 @@ pub const SQL: &str = "
         description TEXT NOT NULL,
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
+
+    -- Just store the MD5 hash of the dismissed suggestion.  The collision rate is low and the
+    -- impact of a collision is not showing a suggestion, which is not that bad.
+    CREATE TABLE dismissed_suggestions (
+        url_hash INTEGER PRIMARY KEY
+    ) WITHOUT ROWID;
 ";
 
 /// Initializes an SQLite connection to the Suggest database, performing
@@ -149,7 +157,7 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
         Ok(db.execute_batch(SQL)?)
     }
 
-    fn upgrade_from(&self, _db: &Transaction<'_>, version: u32) -> open_database::Result<()> {
+    fn upgrade_from(&self, tx: &Transaction<'_>, version: u32) -> open_database::Result<()> {
         match version {
             1..=15 => {
                 // Treat databases with these older schema versions as corrupt,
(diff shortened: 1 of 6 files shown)


bendk:
Putting this up slightly early for feedback, but don't merge until Android and iOS is using the `SuggestStoreBuilder` ([iOS PR](https://github.com/mozilla-mobile/firefox-ios/pull/18732), [Android PR](https://github.com/mozilla-mobile/firefox-android/pull/5546))

bendk:
See #6147 for how this will be used.

bendk:
Pushing this one out again because:
 - The Android and iOS PRs are both merged!   Nothing is blocking this anymore
 - I realized that my scheme to automatically delete some of the tables but keep the rest was really fragile.  I don't think it would work correctly if we ever added or removed tables from the temp tables list.  Instead, I'm hoping we can use the code from `SuggestDao.clear`  to do what we want.  Does that make sense?

components/suggest/src/schema.rs
```diff
@@ -121,6 +121,10 @@ pub const SQL: &str = "
         description TEXT NOT NULL,
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
+
+    CREATE TABLE dismissed_suggestions (
+        url TEXT PRIMARY KEY
```
ncloudioj, line 126: Directly indexing URLs might lead to sub-optimal query performance as vast majority of them begin with "https://www." especially against large tables. We could either index the reversed URLs or follow Places' wisdom of indexing on URL hashes instead.

```
CREATE TABLE dismissed_urls (
    url_hash INTEGER NOT NULL,
    url TEXT NOT NULL,
) WITHOUT ROWID;

CREATE INDEX idx_dismissed_urls_url_hash ON dismissed_urls (url_hash);

SELECT
  1
FROM
  dismissed_urls
WHERE
  url_hash = MD5(target_url)  -- Much faster index lookup
AND
  url = target_url  -- Needed to avoid hash collisions
```

Alternatively, for dismissal records, we often only store the URL hashes in the browser as the collision rate should be super low, and the collision impact (i.e. not serving a suggestion) is normally acceptable.

What do you think?

components/suggest/src/schema.rs
```diff
@@ -149,15 +153,35 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
...
+    fn upgrade_from(&self, tx: &Transaction<'_>, version: u32) -> open_database::Result<()> {
         match version {
             1..=15 => {
                 // Treat databases with these older schema versions as corrupt,
                 // so that they'll be replaced by a fresh, empty database with
                 // the current schema.
                 Err(open_database::Error::Corrupt)
             }
+            16 => {
+                clear_database(tx)?;
```
ncloudioj, line 165: Why do we need to clear those three tables here? How about other ones?

bendk: That's a great question.  I was trying to copy the behavior of the old database code that dropped all tables.  I think the reasoning was that we probably want to re-ingest the suggestions after the schema upgrade.   That said, I don't think that's needed for this migration so I removed from the migration code and added a note to the "For any changes to the schema..." section.  WDYT?

PS:  I thought that deleting from those 3 tables should cause all of the data to be dropped since all other tables have a foreign key to them.   However, I notice that the `yelp_*` tables actually have a foreign key to `record_id` with an `ON DELETE CASCADE` clause.  How should those work?  Should I add `ON DELETE CASCADE` to those tables in the migration?

components/suggest/src/schema.rs
bendk: That makes sense to me.  I switched to just storing the hash.

components/suggest/src/schema.rs
```diff
@@ -149,15 +157,157 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
...
+
+    PRAGMA user_version=16;
+"#;
+
+    /// Test running all schema upgrades from V16, which was the first schema with a "real"
+    /// migration.
+    #[test]
+    fn test_all_upgrades() {
+        let db_file = MigratedDatabaseFile::new(SuggestConnectionInitializer, V16_SCHEMA);
+        db_file.run_all_upgrades();
```
ncloudioj, line 311: Do we need any post-upgrade checks here? Will `run_all_upgrades()` just fail automatically? The function signature suggests it returns nothing.

components/suggest/src/schema.rs
ncloudioj: Can you please help me understand the migration strategy here? I am a little confused as there appears to be two things we're handling here: 1). Switching from `cache_path` to `data_path`; 2). Adding a new `dismissed_suggestions` to the suggest DB.

Re: the path switch, are we planning to move the entire database to a new db under `data_path`? If so, shouldn't we just nuke the old database and create a new one, add the new dismissal table, and re-ingest everything?

As for yelp tables, unlike other suggestions, they don't follow the regular relational convention (`suggestions/keywords/icons`). We can just add all the yelp four tables to the drop list.

components/suggest/src/schema.rs
```diff
@@ -121,6 +123,12 @@ pub const SQL: &str = "
         description TEXT NOT NULL,
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
+
+    -- Just store the MD5 hash of the dismissed suggestion.  The collision rate is low and the
+    -- impact of a collision is not showing a suggestion, which is not that bad.
+    CREATE TABLE dismissed_suggestions (
+        url_hash INTEGER PRIMARY KEY
```
ncloudioj, line 130: This looks good, just a note that we might want to use `INSERT OR IGNORE` to avoid spurious insert failures due to hash collisions.

ncloudioj (approved):
LGTM, just a few clarifying questions.

components/suggest/src/schema.rs
bendk: The main goal of the PR is `1`, but I wanted to do `2` to make sure we had a good story on how to do migrations in the future.

> If so, shouldn't we just nuke the old database and create a new one, add the new dismissal table, and re-ingest everything?

I think that would work fine, but I'm not sure it's needed.  Adding the dismissal table shouldn't require re-ingestion AFAICT.  That said, in practice I think it will be the same since when a user upgrades their major version they'll hit one of the previous upgrades that nuked the DB.

I'm hoping that the `clear_database` function will be useful for future migrations.  There we are going to want to nuke most of the database, but keep the `dismissed_suggestions` and maybe other tables intact.

> As for yelp tables, unlike other suggestions, they don't follow the regular relational convention (suggestions/keywords/icons). We can just add all the yelp four tables to the drop list.

That makes sense, I'll update the `clear_database` to do that.

components/suggest/src/schema.rs
bendk: I think it should panic if an upgrade fails.  I updated the docs to mention that.

components/suggest/src/schema.rs
ncloudioj: Thanks for the clarification, it makes total sense to me now!

components/suggest/src/schema.rs
```diff
@@ -149,15 +157,165 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
...
+        ",
+    )
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use sql_support::open_database::test_utils::MigratedDatabaseFile;
+
+    const V16_SCHEMA: &str = r#"
```
ncloudioj, line 203: Is there a way to reuse the constant `SQL` to avoid a duplication? Looks like we wanted a snapshot of `SQL` for V16, would this be a one-off test just for V16, or will we also need the same for future versions?

ncloudioj (approved):
r+, thanks!

components/suggest/src/schema.rs
bendk: Yes it's a snapshot, so it needs to be duplicated.  I'm hoping that we can use the same sql as a basic test for all future migrations: take an empty v16 database, run all the migrations, and test that nothing fails.

PR #6147, opened 2024-02-27
Bug 1876208 - API for dismissing suggestions

bendk:
(This is on top of #5546, let's merge that one first)

Dismissed suggestion Urls are stored in the database and not returned again in subsequent queries.
    
This currently works with most providers, but not Yelp or Weather.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -252,6 +252,7 @@ impl<'a> SuggestDao<'a> {
             WHERE
               s.provider = :provider
               AND k.keyword = :keyword
+            AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
             "#,
             named_params! {
                 ":keyword": keyword_lowercased,
@@ -348,6 +349,7 @@ impl<'a> SuggestDao<'a> {
             WHERE
               s.provider = :provider
               AND k.keyword = :keyword
+              AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
             "#,
             named_params! {
                 ":keyword": keyword_lowercased,
@@ -430,6 +432,7 @@ impl<'a> SuggestDao<'a> {
                   k.keyword_prefix = :keyword_prefix
                   AND (k.keyword_suffix BETWEEN :keyword_suffix AND :keyword_suffix || x'FFFF')
                   AND s.provider = :provider
+                  AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
                 GROUP BY
                   s.id
                 ORDER BY
@@ -529,6 +532,7 @@ impl<'a> SuggestDao<'a> {
               k.keyword_prefix = :keyword_prefix
               AND (k.keyword_suffix BETWEEN :keyword_suffix AND :keyword_suffix || x'FFFF')
               AND s.provider = :provider
+              AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
             GROUP BY
               s.id,
               k.confidence
@@ -1189,6 +1193,22 @@ impl<'a> SuggestDao<'a> {
                 ":mimetype": mimetype,
             },
         )?;
+        Ok(())
+    }
+
+    pub fn insert_dismissal(&self, url: &str) -> Result<()> {
+        self.conn.execute(
+            "INSERT OR IGNORE INTO dismissed_suggestions(url)
+             VALUES(:url)",
+            named_params! {
+                ":url": url,
+            },
+        )?;
+        Ok(())
+    }
+
+    pub fn clear_dismissals(&self) -> Result<()> {
+        self.conn.execute("DELETE FROM dismissed_suggestions", ())?;
         Ok(())
     }
 
--- a/components/suggest/src/schema.rs
+++ b/components/suggest/src/schema.rs
@@ -15,7 +15,7 @@ use sql_support::open_database::{self, ConnectionInitializer};
 ///     [`SuggestConnectionInitializer::upgrade_from`].
 ///    a. If suggestions should be re-ingested after the migration, call `clear_database()` inside
 ///       the migration.
-pub const VERSION: u32 = 17;
+pub const VERSION: u32 = 18;
 
 /// The current Suggest database schema.
 pub const SQL: &str = "
@@ -124,10 +124,8 @@ pub const SQL: &str = "
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
 
-    -- Just store the MD5 hash of the dismissed suggestion.  The collision rate is low and the
-    -- impact of a collision is not showing a suggestion, which is not that bad.
     CREATE TABLE dismissed_suggestions (
-        url_hash INTEGER PRIMARY KEY
+        url TEXT PRIMARY KEY
     ) WITHOUT ROWID;
 ";
 
@@ -170,6 +168,17 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
                     "
                     CREATE TABLE dismissed_suggestions (
                         url_hash INTEGER PRIMARY KEY
+                    ) WITHOUT ROWID;",
+                    (),
+                )?;
+                Ok(())
+            }
+            17 => {
+                tx.execute(
+                    "
+                    DROP TABLE dismissed_suggestions;
+                    CREATE TABLE dismissed_suggestions (
+                        url TEXT PRIMARY KEY
                     ) WITHOUT ROWID;",
                     (),
                 )?;
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -189,6 +189,22 @@ impl SuggestStore {
         self.inner.query(query)
     }
 
+    /// Dismiss a suggestion
+    ///
+    /// Dismissed suggestions will not be returned again
+    ///
+    /// In the case of AMP suggestions this should be the raw URL.
+    #[handle_error(Error)]
+    pub fn dismiss_suggestion(&self, suggestion_url: String) -> SuggestApiResult<()> {
+        self.inner.dismiss_suggestion(suggestion_url)
+    }
+
+    /// Clear dismissed suggestions
+    #[handle_error(Error)]
+    pub fn clear_dismissed_suggestions(&self) -> SuggestApiResult<()> {
+        self.inner.clear_dismissed_suggestions()
+    }
+
     /// Interrupts any ongoing queries.
     ///
     /// This should be called when the user types new input into the address
@@ -273,6 +289,17 @@ impl<S> SuggestStoreInner<S> {
             return Ok(Vec::new());
         }
         self.dbs()?.reader.read(|dao| dao.fetch_suggestions(&query))
+    }
+
+    fn dismiss_suggestion(&self, suggestion_url: String) -> Result<()> {
+        self.dbs()?
+            .writer
+            .write(|dao| dao.insert_dismissal(&suggestion_url))
+    }
+
+    fn clear_dismissed_suggestions(&self) -> Result<()> {
+        self.dbs()?.writer.write(|dao| dao.clear_dismissals())?;
+        Ok(())
     }
 
     fn interrupt(&self) {
@@ -4945,10 +4972,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                         },
                     ),
@@ -5017,10 +5044,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                         },
                     ),
@@ -5216,10 +5243,10 @@ mod tests {
                     UnparsableRecords(
                         {
                             "clippy-2": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                             "fancy-new-suggestions-1": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                         },
                     ),
@@ -5305,7 +5332,7 @@ mod tests {
                     UnparsableRecords(
                         {
                             "invalid-attachment": UnparsableRecord {
-                                schema_version: 17,
+                                schema_version: 18,
                             },
                         },
                     ),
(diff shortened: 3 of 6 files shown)


bendk:
We've decided to stop storing unparsable records and simply always drop and recreate the suggestions DB whenever the schema is updated.

linabutler:
@bendk Hmmm, was that comment and close meant for #6107? 😅

bendk:
Yes it was, I was wondering where it went and why I had to close that issue twice :rofl:

components/suggest/src/db.rs
```diff
@@ -1213,6 +1218,22 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
+    pub fn insert_dismissal(&self, url: &str) -> Result<()> {
+        self.conn.execute(
+            "INSERT OR IGNORE INTO dismissed_suggestions(url_hash)
```
ncloudioj, line 1223: The raw AMP landing URLs might contain date time templates "%YYYYMMDDHH%", the Component populates those templates with proper values via `suggestion::cook_raw_suggestion_url()`. So, when Firefox records a dismissed URL, it should be the cooked URL rather than the raw one. We might need to process the URL before inserting it into the table otherwise it won't be sufficient to block future URLs generated from the same raw URL.

@0c0w3 Could you point us to code path where we handle this on Desktop? Maybe we can reuse it here.

components/suggest/src/schema.rs
```diff
@@ -149,15 +158,212 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
...
+                // value from actually being used. As a special case, we return
+                // null when the input is NULL. We return NULL instead of zero
+                // because the hash columns are NOT NULL, so attempting to
+                // actually use the null should fail.
+                if raw == ValueRef::Null {
+                    return Ok(None);
+                }
+                match raw.as_str() {
+                    Ok(s) => {
+                        let mut hasher = DefaultHasher::new();
```
ncloudioj, line 230: One potential issue of using the default hasher: if Rust changes the default hasher in the future, all the persisted hash digests in the db would no longer match against to the results using the new hasher. That said, we might want to pick a hasher explicitly here.

components/suggest/src/schema.rs
```diff
@@ -121,6 +123,12 @@ pub const SQL: &str = "
         description TEXT NOT NULL,
         FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
     );
+
+    -- Just store the MD5 hash of the dismissed suggestion.  The collision rate is low and the
```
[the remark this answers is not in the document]
bendk: Actually, maybe we should move to MD5.  The default hasher can change between Rust versions which would break the `dismissed_suggestions` table.

components/suggest/src/db.rs
bendk: Ahh interesting, I definitely need to address that.  I would have thought we should be storing the raw URL though.  Won't the timestamp will change each time the suggestion is displayed?

ncloudioj: Did some digging, looks like we're hashing the original URLs (the pre-cooked one) when recording for dismissals on Desktop (sources: [here](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarProviderQuickSuggest.sys.mjs#470-473) and [here](https://searchfox.org/mozilla-central/source/browser/components/urlbar/private/BlockedSuggestions.sys.mjs#38-60)), meaning that we could add a new field (e.g. `original_url: Option<String>`) to `Suggestion` in the Rust component, and populate it with the original URL if there is any template in it.

ncloudioj: Great!

components/suggest/src/schema.rs
```diff
@@ -149,15 +158,217 @@ impl ConnectionInitializer for SuggestConnectionInitializer {
...
+                    return Ok(None);
+                }
+                match raw.as_str() {
+                    Ok(s) => {
+                        // Hash using Md5, then return the first 64 bits since SQLite uses 64-bit
+                        // integers.
+                        let mut hasher = Md5::new();
+                        hasher.update(s.as_bytes());
+                        let hash_value = i128::from_le_bytes(hasher.finalize().into());
+                        Ok(Some(hash_value as i64))
```
ncloudioj, line 235: Nit: let's grab the leading 64 bits explicitly. Using `as` to cast from a larger integer to a smaller integer will truncate which could yield weird negative results.

```rs
    println!("{}", (i128::MAX >> 64) as i64);  // => 9223372036854775807

    println!("{}", i128::MAX as i64);               // => -1
    println!("{}", (i128::MAX - 1) as i64);       // => -2
    println!("{}", (i128::MAX - 2) as i64);      // => -3
    println!("{}", (i128::MAX - 3) as i64);      // => -4
```

 ```suggestion
                         Ok(Some((hash_value >> 64) as i64))
```

**Review by ncloudioj — APPROVED** [review]
LGTM, r=wc.

File: components/suggest/src/store.rs [hunk 5]
```diff
@@ -194,6 +189,20 @@ impl SuggestStore {
         self.inner.query(query)
     }
 
+    /// Dismiss a suggestion
+    ///
+    /// Dismissed suggestions will not be returned again
+    #[handle_error(Error)]
+    pub fn dismiss_suggestion(&self, suggestion: Suggestion) -> SuggestApiResult<()> {
```
ncloudioj, line 196: Hi @0c0w3, FYI, this is the dismissal API exposed by the component. It will take a `Suggestion` as the only argument. Any concerns about that?

components/suggest/src/schema.rs
bendk: I updated the code to do this, but I think there still could be negatives (for example `u128::MAX` gets converted to `-1`).  We could shift by 65 bits to avoid this, but I'd rather keep the extra hash bit and live with potentially negative numbers.  WDYT?

components/suggest/src/schema.rs
```diff
@@ -195,6 +196,56 @@ pub fn clear_database(db: &Connection) -> rusqlite::Result<()> {
...
+        "hash",
+        1,
+        FunctionFlags::SQLITE_UTF8 | FunctionFlags::SQLITE_DETERMINISTIC,
+        sql_fns::hash,
+    )?;
+    Ok(())
+}
+
+mod sql_fns {
+    use md5::{Digest, Md5};
```
linabutler, line 210: With my Places peer hat on 🧙🏼‍♀️ I should call out that the Desktop security engineering team has been working on [removing deprecated hashes from Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1871103#c1), including MD5. Places recently [switched to using SHA-256](http://bugzilla.mozilla.org/show_bug.cgi?id=1871103) for cases where it previously used MD5.

Even though we have a known and trusted set of URLs here, I think introducing a new use of a deprecated hash function is still going to raise a few eyebrows when we vendor.

linabutler: It looks like the average length of `suggestions.url` is ~64 bytes. SHA-256 digests are 32 bytes, so it looks like it's still a bit more space-efficient on average to use a hash instead of the full URL.

That said, though, I think storing the full URL should also be OK, especially if it turns out we need it because of the `%%YYYYMMDDHH%` replacement parameters.

In general, I think a hash is going to be more efficient than a long text primary key or index when there are a lot of rows, as in Places (tens of thousands). I don't think we expect dismissed suggestions to be anywhere close to that size, though? For dozens or hundreds of rows, storing the URL directly is A-OK.

bendk: Between moving to SHA-256 and using the full URL, I think we should just use the full one.  As you point out the space difference is small and we shouldn't expect many URLs to be stored in the table.  @ncloudioj how do you feel about it?

ncloudioj: Thanks for the heads up, Lina!

I am still leaning towards hashing those dismissed URLs to be consistent with what we are doing for other similar cases in the browser (e.g. we MD5 hash the dismissed URLs for New Tab; we SHA1 hash dismissed URLs for Firefox Suggest Desktop). We can pick SHA-256 to stay away from those insecure hashers. If we don't store the whole digest, we could grab a smaller portion of it (e.g. 16 bytes) which should offer enough entropy given the relatively smaller input size.

Does that make sense?

components/suggest/src/db.rs
```diff
@@ -1192,6 +1196,22 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
+    pub fn insert_dismissal(&self, url: &str) -> Result<()> {
```
linabutler, line 1199:
```suggestion
    pub fn insert_dismissal(&mut self, url: &str) -> Result<()> {
```

components/suggest/src/db.rs
```diff
@@ -1192,6 +1196,22 @@ impl<'a> SuggestDao<'a> {
...
+            "INSERT OR IGNORE INTO dismissed_suggestions(url_hash)
+             VALUES(hash(:url))",
+            named_params! {
+                ":url": url,
+            },
+        )?;
+        Ok(())
+    }
+
+    pub fn clear_dismissals(&self) -> Result<()> {
```
linabutler, line 1210:
```suggestion
    pub fn clear_dismissals(&mut self) -> Result<()> {
```

(Just a loose convention that we've been using for the other DAO methods!)

components/suggest/src/schema.rs
linabutler: > we could grab a smaller portion of it (e.g. 16 bytes) which should offer enough entropy given the relatively smaller input size.

That was my other question 😅—it looks like you're only keeping the first 8 bytes for the primary key, so any collisions will exclude all the colliding URLs. Is that right? That still sounds OK for our case, but definitely subtle.

Even if we used the full 32-byte hash, though, it's still going to be shorter than any URLs—the shortest URL we have is `https://en.wikipedia.org/wiki/LVM3`, which is 34 bytes.

For comparison, Places uses a [non-cryptographic 48-bit hash](https://github.com/mozilla/application-services/blob/9054db4bb5031881550ceab3448665ef6499a706/components/places/src/hash.rs), with a [non-unique index](https://github.com/mozilla/application-services/blob/9054db4bb5031881550ceab3448665ef6499a706/components/places/sql/create_shared_schema.sql#L38)—so if there are any collisions, it'll still compare URLs, but the index reduces the search space from tens of thousands to a handful. But that scheme pays off for larger tables—for smaller tables like ours, it can end up increasing the on-disk size.

Do you happen to know why we store hashes on Desktop, @ncloudioj and @0c0w3? Is it mostly for size benefits, or are there other reasons like privacy?

components/suggest/src/suggestion.rs
```diff
@@ -109,6 +109,37 @@ impl Ord for Suggestion {
     }
 }
 
+impl Suggestion {
+    /// Get the URL for this suggestion, if present
+    pub fn url(&self) -> Option<&str> {
```
linabutler, line 114:
```suggestion
    pub(crate) fn url(&self) -> Option<&str> {
```

Another loose convention I've been following—I'm curious what you think of it, actually, and if it's worth continuing? 😄—is using `pub(crate)` for crate-internal symbols, and `pub` for symbols that are exposed in the UDL.

I thought it might make it a little easier to see what's exposed to foreign callers, but I wonder if that's actually true. WDYT?

linabutler (commented):
The general shape of this looks great—thank you so much @bendk!

components/suggest/src/schema.rs
ncloudioj: To my understanding, we hash the dismissed URLs for both benefits: 1), we always store a fixed size hash digest regardless of the size of the original URL; 2), as both cases persist the dismissed URLs in preferences, the hashed digests don't expose anything in plain as they can be viewed / accessed by unauthorized individuals or addons.

I am not super worried about the size actually, we can store the entire 32-byte hash as is in the db. Alternatively, there are a whole lot of other non-cryptographic hashers available such as [smhasher](https://github.com/rurban/smhasher) with Rust implementations, also the one Lina mentioned above at our disposal.

I don't want to block us for this though, happy to move on with the option of storing raw URLs :)

components/suggest/src/db.rs
bendk: Good catch, I like those conventions.

components/suggest/src/schema.rs
bendk: How about we start with raw URLs and consider adding the hashes as a performance/privacy improvement later (although  I have some questions on how much of a privacy gain this is).

components/suggest/src/suggestion.rs
bendk: I wonder about this a lot myself and have gone back and forth on what's best.  My current practice is to label everything `pub` if it's not private to the module, then let the higher-level modules control the if they want to do `use vs pub use`.  It's less typing and when I need to see what's exposed I go to `lib.rs` anyways.  I'm not confident that this is the Right Way, but it's worked okay for me.

components/suggest/src/schema.rs
ncloudioj: Yup, let's revisit this later.

> although I have some questions on how much of a privacy gain this is.

Curious about your questions, can you share?

bendk:
Updated this PR to not hash the URLs for now at least.  AFAICT, these are the remaining open questions:
- Is the external API okay (passing a `Suggestion` object into `dismiss_suggestion`)?
- pub vs pub(crate)
- hashing only 63-bits to avoid signed integer issues (not currently relevant for this feature, but I figure we might as well check in the `hash` function in case we want to use it in the future).

@0c0w3 @linabutler @ncloudioj what do you think about the remaining questions?

components/suggest/src/schema.rs
bendk: The suggestion list is public and also fairly small.  So shouldn't an attacker be able to create a map of hashed URLs -> full URLs.  This won't be perfect if there's a hash collision, but it seems like this would get around most of the privacy protections.  Maybe adding a salt component would fix this though?  Not sure on that one.

components/suggest/src/schema.rs
```diff
@@ -127,7 +127,7 @@ pub const SQL: &str = "
     -- Just store the MD5 hash of the dismissed suggestion.  The collision rate is low and the
     -- impact of a collision is not showing a suggestion, which is not that bad.
     CREATE TABLE dismissed_suggestions (
-        url_hash INTEGER PRIMARY KEY
+        url TEXT PRIMARY KEY
```
ncloudioj, line 130: Now that we're storing raw URLs, let's the drop the above comment about hashing.

ncloudioj (approved):
LGTM!

> Is the external API okay (passing a Suggestion object into dismiss_suggestion)?

The reason I was asking is b/c those [comments](https://searchfox.org/mozilla-central/source/browser/components/urlbar/private/SuggestBackendRust.sys.mjs#385-400), and wanted to confirm with Drew to see if it's possible for the consumer to feed the Suggestion object back for dismissal.


> hashing only 63-bits to avoid signed integer issues (not currently relevant for this feature, but I figure we might as well check in the hash function in case we want to use it in the future).

Per Lina's comment, we'd like to avoid those legacy insecure hashers (e.g. MD5) in Places, shall we follow that here as well?

bendk:
> Per Lina's comment, we'd like to avoid those legacy insecure hashers (e.g. MD5) in Places, shall we follow that here as well?

Good point.  I just went ahead and removed that code, since it seems like we don't know exactly how we want to handle hashing in the future.

components/suggest/src/schema.rs
ncloudioj: Ah, I see. Your point makes total sense in this context.

I think for all those dismissal scenarios, we were not attempting to apply standard mechanisms (salting & encrypting) to secure the data or preserve users' privacy in the best effort. Rather, we meant to convert the data in a way such that the processed data is not obvious to read. Imagine one dismissed a URL from the New Tab page because they wanted to "hide" it, storing it in plain in another user visible place could cause some surprises & awkwardness later. Other benefits (such as smaller footprint & easier to index/query) are a plus so I think that's why decided to hash the raw URLs.

components/suggest/src/suggestion.rs
linabutler: That sounds perfect, thanks! I like your approach a lot more 😁

components/suggest/src/schema.rs
linabutler: > How about we start with raw URLs

That sounds great to me!

components/suggest/src/store.rs
linabutler: Thinking about it a bit more (sorry I didn't think of this earlier!), the one downside I can think of is that the caller needs to hold on to the entire `Suggestion` for as long as it's visible.

We can do that on Android and iOS—there's a [`metadata` map](https://github.com/mozilla-mobile/firefox-android/blob/8a69a9a83cbb86132d68f81e0e7f509a146c167c/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt#L105-L106) we can use to smuggle it on Android, and we can [add a field on iOS](https://github.com/mozilla-mobile/firefox-ios/blob/0194fca784ee190138615652439d9fb4ef9cee4e/firefox-ios/Storage/Rust/RustFirefoxSuggestion.swift#L24-L28)...

_But_ since the suggestion carries the raw `icon` data, we're duplicating the icon data in memory: once for the buffer in the `Suggestion`, and once for the decoded [`android.graphics.Bitmap`](https://github.com/mozilla-mobile/firefox-android/blob/8a69a9a83cbb86132d68f81e0e7f509a146c167c/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt#L126) / [`UIImage`](https://github.com/mozilla-mobile/firefox-ios/blob/0194fca784ee190138615652439d9fb4ef9cee4e/firefox-ios/Storage/Rust/RustFirefoxSuggestion.swift#L67) instance. I think we'll also incur a third copy of that icon data when we pass it back over the FFI to `dismiss_suggestion`.

This would be a backward-incompatible change, but I wonder if it's worth making `Url` a dictionary or an opaque `interface` type. Alternatively, maybe we could add a `dismissal_id` field—either a string, or an opaque `interface`—that carries the dismissal URL.

linabutler:
Thank you so much for working through all this, @bendk!

I did think of [one downside for passing the full `Suggestion` to `dismiss_suggestion`](https://github.com/mozilla/application-services/pull/6147/files#r1525312121) that I should've thought of earlier, but this review thread is getting pretty long, and the rest of the plumbing looks good. We can always land this PR as-is, and iterate on the API in a follow-up.

components/suggest/src/store.rs
ncloudioj: Right, I was initially thinking about the [special handling](https://searchfox.org/mozilla-central/source/browser/components/urlbar/private/SuggestBackendRust.sys.mjs#385-400) we did for desktop and wondered if it's possible to restore the `Suggestion` object and pass it back to the component with all needed information, especially, the enum variant type. Lina's comment makes this even more questionable.

How about we just take a `String` instead as the only input for this API? Or, if we expect to include other fields for a dismissal in the future, we can add an interface `Dismiss` with a string `url` field for now and use this to capture the context for this API.

bendk: I like the idea of having the consumer send us a URL rather than the full Suggestion.

I think now that https://bugzilla.mozilla.org/show_bug.cgi?id=1875848 has been implemented, we can do a type check to figure out the variant when we try to extract the URL.  This still means some duplicated code, but it's not that bad.

Going forward, I think we could probably improve the API shape.  Now that https://github.com/mozilla/uniffi-rs/pull/1869 is merged, it should be possible to define enums that store objects which gives us a tone of options.  Maybe the `icon` field could be an object.  I think that would allow us to pass Suggestion instances around efficiently (assuming that's the only field that's storing a large amount of data).

0c0w3 (commented):
Thanks Ben!

Would it be a bad idea to have a single check-for-dismissals step at the end of `fetch_suggestions()`? It seems a little onerous to have to add a subquery (or whatever logic) per suggestion type, especially as we start to scale up new types. Doing one more query at the end doesn't seem too bad especially since `dismissed_suggestions` is keyed on the URL.

This doesn't handle Yelp suggestions, which is fine for a follow-up IMO. We'll need to define what constitutes a raw URL for Yelp since the way Yelp builds URLs is relatively complex (location params, query modifiers).

bendk:
> Would it be a bad idea to have a single check-for-dismissals step at the end of `fetch_suggestions()`? It seems a little onerous to have to add a subquery (or whatever logic) per suggestion type, especially as we start to scale up new types. Doing one more query at the end doesn't seem too bad especially since `dismissed_suggestions` is keyed on the URL.

This seems like a good idea to me, but do we need to worry about the limits?  I'm slightly paranoid that the dismissed suggestions will push non-dismissed suggestions down the list and cause them to be filtered out.

bendk:
AFAICT, there's no major blockers left.  How do others feel about merging this one, then following up on:

- Simplifying the code by removing the sub-queries on each suggestion
- Better system to identify suggestions

Are there any other remaining questions that I'm missing?

ncloudioj:
Sounds good to me. Go ahead! :shipit:

PR #6151, opened 2024-03-01
Bug 1882954 - Expose icon mimetype in suggestions

dadaa:
Currently, if the blob content of the icon contained in suggestion from Rust is SVG, the image will be broken unless specify the mimetype. Especially for Yelp Suggestion, as it uses SVG icon, we specify the mimetype explicitly.
https://searchfox.org/mozilla-central/rev/da2c1e64cc7aa8718fb92eb602136da5d505d664/browser/components/urlbar/private/SuggestBackendRust.sys.mjs#141
However, we ideal to use mimeytpe specified in RemoteSettings to think in the future as well.
This PR exposes the mimetype to create the icon image.

And its JS code: https://phabricator.services.mozilla.com/D203237

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -285,9 +285,12 @@ impl<'a> SuggestDao<'a> {
                       amp.iab_category,
                       amp.impression_url,
                       amp.click_url,
-                      (SELECT i.data FROM icons i WHERE i.id = amp.icon_id) AS icon
+                      i.data AS icon,
+                      i.mimetype AS icon_mimetype
                     FROM
                       amp_custom_details amp
+                    LEFT JOIN
+                      icons i ON amp.icon_id = i.id
                     WHERE
                       amp.suggestion_id = :suggestion_id
                     "#,
@@ -308,6 +311,7 @@ impl<'a> SuggestDao<'a> {
                             raw_url,
                             full_keyword: full_keyword(keyword_lowercased, &keywords),
                             icon: row.get("icon")?,
+                            icon_mimetype: row.get("icon_mimetype")?,
                             impression_url: row.get("impression_url")?,
                             click_url: cooked_click_url,
                             raw_click_url,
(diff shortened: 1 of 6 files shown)


0c0w3 (approved):
Looks great, thanks! It looks like we have tests for Amp and Yelp suggestions without icons but not for Wikipedia. If it's not too much trouble could you add one please?

dadaa:
Thank you very much!
I will address your comments.

components/suggest/src/db.rs
```diff
@@ -350,21 +354,32 @@ impl<'a> SuggestDao<'a> {
...
+                        row.get::<_, Option<Vec<u8>>>(0)?,
+                        row.get::<_, Option<String>>(1)?,
+                    )) },
                 )?;
+                let (icon, icon_mimetype) = if let Some(result) = icons.pop() {
+                    result
+                } else {
+                    (None, None)
+                };
+
```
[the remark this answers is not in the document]
dadaa: Nice!!

components/suggest/src/db.rs
```diff
@@ -277,9 +277,12 @@ impl<'a> SuggestDao<'a> {
                       amp.iab_category,
                       amp.impression_url,
                       amp.click_url,
-                      (SELECT i.data FROM icons i WHERE i.id = amp.icon_id) AS icon
+                      i.data AS icon,
+                      i.mimetype AS icon_mimetype
                     FROM
                       amp_custom_details amp
+                    LEFT JOIN
+                      icons i ON amp.icon_id = i.id
```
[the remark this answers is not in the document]
linabutler: This makes sense to me, thanks for checking!

We had a super interesting bug ages ago in bookmark sync, where a left join on multiple columns for 40k rows caused a query to take over 4 minutes—but using a functionally equivalent correlated subquery took just 2 seconds. That edge case doesn't apply here, though! ☺️

components/suggest/src/store.rs
```diff
@@ -409,7 +409,7 @@ where
                     };
                     let data = self.settings_client.get_attachment(&attachment.location)?;
                     writer.write(|dao| {
-                        dao.put_icon(icon_id, &data)?;
+                        dao.put_icon(icon_id, &data, &attachment.mimetype)?;
```
linabutler, line 412: `attachment.mimetype` looks like it's a `String`, not an `Option<String>`, so I wonder if we should make the column `TEXT NOT NULL`—WDYT?

components/suggest/src/suggest.udl
```diff
@@ -40,6 +40,7 @@ interface Suggestion {
         string url,
         string raw_url,
         sequence<u8>? icon,
```
linabutler, line 42: It might be time to add a struct for the icon, like:

```webidl
dictionary SuggestionIcon {
    sequence<u8> icon;
    string mime_type;
}
```

And then in Rust:

```rs
struct SuggestionIcon {
    icon: Vec<u8>,
    mime_type: String,
}
````

...But only if you want to; the way you have it also looks good to me! 😊

**Review by linabutler — APPROVED** [review]
Thanks @dadaa!

File: components/suggest/src/suggest.udl (hunk 4 above)
Comment (dadaa) [reply components/suggest/src/suggest.udl:42]: Yeah, I also thought about it. Please let me do it in another PR!

File: components/suggest/src/db.rs [hunk 5]
```diff
@@ -350,21 +354,30 @@ impl<'a> SuggestDao<'a> {
                     },
                     |row| row.get(0),
                 )?;
-                let icon = self.conn.try_query_one(
-                    "SELECT i.data
+                let mut icons = self.conn.query_rows_and_then_cached(
```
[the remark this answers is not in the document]
dadaa: Oh, cool!! Thanks!

components/suggest/src/store.rs
dadaa: Yep, indeed! Thanks!

0c0w3:
> We had a super interesting bug ages ago in bookmark sync, where a left join on multiple columns for 40k rows caused a query to take over 4 minutes—but using a functionally equivalent correlated subquery took just 2 seconds. That edge case doesn't apply here, though! ☺️

@linabutler oh wow, do you happen to remember if the query plans were the same (apologies if that's what you mean by functionally equivalent)? Did you figure out the root cause? I seem to recall a similar problem and subquery fix when I worked on Places, but that was years ago and I don't remember the details. I'd like to refresh my knowledge on this, if you happen to remember.

PR #6152, opened 2024-03-01
Initial commit for the relevancy component

bendk:
Added boilerplate for UniFFI, error handling, databases, etc.  I tried to follow the best practices from other application-services components.

Created the initial types and methods for the API.  I think all of these should be considered works-in-progress.  It seems likely that some or all of them will change as we implement the functionality.

Added some code to read / write URL interest data from a test binary file currently stored in the repo.  This is also a WIP, at the very least it needs to change to download the data from remote settings.

  - Note:


bendk:
> Created the initial types and methods for the API. I think all of these should be considered works-in-progress. It seems likely that some or all of them will change as we implement the functionality.

So, I think this review should focus on "does this code seem good enough to start with?" rather than "is it correct?"

components/relevancy/src/db.rs
```diff
@@ -0,0 +1,118 @@
...
+impl RelevancyDb {
+    pub fn open(path: impl AsRef<Path>) -> Result<Self> {
+        let conn = open_database_with_flags(
+            path,
+            OpenFlags::SQLITE_OPEN_URI
+                | OpenFlags::SQLITE_OPEN_NO_MUTEX
+                | OpenFlags::SQLITE_OPEN_CREATE
+                | OpenFlags::SQLITE_OPEN_READ_WRITE,
+            &RelevancyConnectionInitializer,
+        )?;
```
bendk, line 30: We talked about waiting to handle persistence, but I'm wondering if we could just start it.  The `sql_support` crate makes it pretty easy to handle migrations, including migrations where we just throw away the entire database and start over.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "components/nimbus",
     "components/places",
     "components/push",
+    "components/relevancy",
```
ncloudioj, line 14: Nice pick for naming, we're calling for other candidates as well. Stay tuned :P

components/relevancy/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl RelevancyStore {
+    #[handle_error(Error)]
+    pub fn new(db_path: String) -> ApiResult<Self> {
+        Ok(Self {
+            db: RelevancyDb::open(db_path)?,
+        })
+    }
+
+    #[handle_error(Error)]
```
ncloudioj, line 37: Could you add some docstrings to those placeholder methods for future implementers?

components/relevancy/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+        todo!()
+    }
+
+    #[handle_error(Error)]
+    pub fn user_interest_vector(&self) -> ApiResult<InterestVector> {
+        todo!()
+    }
+}
+
+pub struct InterestMetrics {
```
ncloudioj, line 54: Can you also document this? Just describe what they are, no need to cover the calculation details.

ncloudioj: NVM, just say you put the docs in the UDL. Just a note to point readers to there would be fine.

components/relevancy/src/utils.rs
```diff
@@ -0,0 +1,63 @@
...
+use md5::{Digest, Md5};
+use url::{Host, Url};
+
+pub type UrlHash = [u8; 16];
+
+/// Given a URL, extract the part of it that we want to use to identify it.
+///
+/// We currently use the final 3 components of the URL domain.
+///
+/// TODO: decide if this should be 3 or 3 components.
```
ncloudioj, line 14: I think this should follow what components of the URL are used for hash generation from upstream, and that should be passed down from Remote Settings along with other remote assets.

Maybe we can define a mapping (following Github's [terminology](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/about-custom-domains-and-github-pages)) as part of the schema for RS assets:

```rs
enum InputTypes {
    /// `{ host }.{ top domain }`.
    /// Example: `https://www.example.com/` -> `example.com`
    UrlApexDomain,
    /// `{ subdomain }.{ host }.{ top domain }`.
    /// Example: `https://www/example.com` -> `www.example.com`
    UrlFullDomain,
   
    /// Other input types...
}
``` 

WDYT?

components/relevancy/src/interest.rs
```diff
@@ -0,0 +1,166 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// List of possible interests for a domain.  Each domain would have exactly one of these.  If we
```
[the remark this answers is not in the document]
bendk: Good point.  I'll update the docs.  I think I have have a case of multiple interests in the test data because I wanted to make sure that it worked.

components/relevancy/src/interest.rs
```diff
@@ -0,0 +1,166 @@
...
+/// can't calculate an interest based on the classifier data, then we put it in `Inconclusive`
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum Interest {
```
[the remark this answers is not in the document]
bendk: I kind of like the idea managing this via remote settings.

It seems like we could avoid breaking changes for the most part by just adding new interests rather than overwriting existing ones.  A versioning system still would be good though, maybe we would want to do something like add parent links and create a taxonomy.

This would impact some of the data structures in the rest of the component.   Maybe [InterestVector](https://github.com/mozilla/application-services/pull/6152/files#diff-967df1177b49cb7a2c410d8b0929e110ced3c88b9506ff4f618bf99452b9bc00R89-R110) would change to be a vec of `u32` values -- one for each interest.  Or maybe it should actually be a Vec of `(interest_code, count)` tuples.  That seems better if the interest vecs are sparse.

components/relevancy/src/utils.rs
bendk: That makes a lot of sense to me, but probably for future work.  I tried to implement it and realized it wasn't so simple to do.

components/relevancy/src/interest.rs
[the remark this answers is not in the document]
tiftran: +1 to remote settings.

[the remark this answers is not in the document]
ncloudioj: > It seems like we could avoid breaking changes for the most part by just adding new interests rather than overwriting existing ones.

Yup, that will be ideal. Though, given that we're still at the early stage and category def changes are in flux, it will take a while for us to land on a stable category list (or even better a taxonomy) to support append-only changes. Let's think about a stopgap for now, at least, future incompatible changes won't cause any weird issues for the old Fx users. What do you think?

> Maybe [InterestVector](https://github.com/mozilla/application-services/pull/6152/files#diff-967df1177b49cb7a2c410d8b0929e110ced3c88b9506ff4f618bf99452b9bc00R89-R110) would change to be a vec of u32 values -- one for each interest. Or maybe it should actually be a Vec of (interest_code, count) tuples. That seems better if the interest vecs are sparse.

Yup, that change looks reasonable. As in interests will come from RS, we would generate the interest vectors dynamically.

components/relevancy/src/utils.rs
ncloudioj: Yup, let's handle this as follow ups.

ncloudioj (approved):
LGTM 🚢

PR #6156, opened 2024-03-06
Bug 1876217 - Add full keyword support for AMP items

bendk:
Parse the `full_keyword` field for downloaded AMP / wikipedia suggestions and store it in the DB.

Use the field for AMP suggestions rather than the current hueristic-based code.

Don't use this field for Wikimedia suggestions.  This is because the full keyword field is just the full article title, which is already included in the suggestion.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -239,12 +239,16 @@ impl<'a> SuggestDao<'a> {
               s.title,
               s.url,
               s.provider,
-              s.score
+              s.score,
+              fk.full_keyword
             FROM
               suggestions s
             JOIN
               keywords k
               ON k.suggestion_id = s.id
+            LEFT JOIN
+              full_keywords fk
+              ON k.full_keyword_id = fk.id
             WHERE
               s.provider = :provider
               AND k.keyword = :keyword
@@ -256,8 +260,9 @@ impl<'a> SuggestDao<'a> {
             |row| -> Result<Suggestion> {
                 let suggestion_id: i64 = row.get("id")?;
                 let title = row.get("title")?;
-                let raw_url = row.get::<_, String>("url")?;
-                let score = row.get::<_, f64>("score")?;
+                let raw_url: String = row.get("url")?;
+                let score: f64 = row.get("score")?;
+                let full_keyword_from_db: Option<String> = row.get("full_keyword")?;
 
                 let keywords: Vec<String> = self.conn.query_rows_and_then_cached(
                     r#"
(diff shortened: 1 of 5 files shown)


components/suggest/src/store.rs
```diff
@@ -998,6 +998,211 @@ mod tests {
...
+                        impression_url: "https://example.com/impression_url",
+                        click_url: "https://example.com/click_url",
+                        raw_click_url: "https://example.com/click_url",
+                        score: 0.3,
+                    },
+                ]
+            "#]]
+            .assert_debug_eq(&dao.fetch_suggestions(&SuggestionQuery {
+                keyword: "los pollos h".into(),
+                providers: vec![SuggestionProvider::Amp],
```
bendk, line 1179: Should I also be handling `SuggestionProvider::AmpMobile`?  I'm not quite sure what the relationship between the two AMP providers is.

components/suggest/src/db.rs
```diff
@@ -853,22 +859,56 @@ impl<'a> SuggestDao<'a> {
...
+                                 VALUES(
+                                    :suggestion_id,
+                                    :keyword
+                                 )",
+                                named_params! {
+                                    ":keyword": full_keyword,
+                                    ":suggestion_id": suggestion_id,
+                                },
+                            )?;
+                            let full_keyword_id = self.conn.last_insert_rowid();
```
ncloudioj, line 886: We often use `RETURNING id` in the `Insert` SQL to get the last insert rowid, can we do that here as well?

ncloudioj (approved):
Looks great, only a few minor comments for your consideration.

components/suggest/src/store.rs
bendk: That makes sense, I added a test and refactored some code for this.  Can you review that the changes make sense?

components/suggest/src/rs.rs
```diff
@@ -281,6 +283,41 @@ impl DownloadedAmpWikipediaSuggestion {
...
+
+            AmpKeyword {
+                rank: i,
+                keyword,
+                full_keyword: common
+                    .full_keywords
+                    .get(full_keyword_index)
+                    .map(|(keyword, _)| keyword.as_str()),
+            }
+        })
```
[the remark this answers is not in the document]
bendk: Really slick, love it.

components/suggest/src/db.rs
bendk: TIL.  I updated the code to use that.

components/suggest/src/db.rs
```diff
@@ -1339,6 +1366,53 @@ impl<'a> SuggestDao<'a> {
...
+impl<'a> FullKeywordInserter<'a> {
+    fn new(conn: &'a Connection, suggestion_id: i64) -> Self {
+        Self {
+            conn,
+            suggestion_id,
+            last_inserted: None,
+        }
+    }
+
+    fn get_id(&mut self, full_keyword: &'a str) -> rusqlite::Result<i64> {
```
ncloudioj, line 1389: Suggestion: how about renaming this to `maybe_insert()`? `get_id()` kind of hides the insert part.

ncloudioj (approved):
Still r+!

PR #6172, opened 2024-03-21
Bug 1886272: Starting suggest benchmarks

bendk:
These can be run using `cargo suggest-bench`.
The only benchmark I added so far is the time for a full ingestion.

  - Note:


bendk:
On my machine:
 - Ingestion takes about 1.5 seconds
 - The time to download all the remote settings data is very similar.

I think this means that:
  - We should definitely be able to speed up ingestion
  - I'm not sure if this will fix the firefox-ios issue though, since the network requests themselves may cause the timeout.

components/suggest/benches/benchmark_all.rs
```diff
@@ -0,0 +1,19 @@
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
+use suggest::benchmarks::{ingest, BenchmarkWithInput};
+
+pub fn ingest_benchmarks(c: &mut Criterion) {
+    let mut group = c.benchmark_group("ingest");
+    viaduct_reqwest::use_reqwest_backend();
+    group.sample_size(10);
```
ncloudioj, line 7: 10 is the minimal sample size, any particular reason we want to pin it?

components/suggest/src/benchmarks/client.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::{rs::SuggestRemoteSettingsClient, Result};
+use parking_lot::Mutex;
+use remote_settings::{Client, GetItemsOptions, RemoteSettingsConfig, RemoteSettingsResponse};
+use std::collections::HashMap;
+
+/// Remotes settings client that runs during the benchmark warm-up phase.
+///
+/// This should be used to run a full ingestion.
+/// Then in can be converted into a [RemoteSettingsBenchmarkClient], which helps allows benchmark code to exclude the network request time.
```
ncloudioj, line 13:
```suggestion
/// Then iit can be converted into a [RemoteSettingsBenchmarkClient], which allows benchmark code to exclude the network request time.
```

ncloudioj (approved):
This looks great, thanks!

components/suggest/benches/benchmark_all.rs
bendk: The default of 100 makes the tests take a long time.  I think they would take around 150s, but I'm not totally sure because I got tired of waiting and killed them.  I'll add a note about this.

bendk:
I made a few changes based on our conversation and also from me trying to actually use these benchmarks:
  - Added a README.md file which describes the benchmarks and also lists some missing benchmarks that would be nice to have.
  - Split up the benchmark by record type.  Most record types take a small amount of time, it's really only AMP-wikipedia and AMP-mobile that are driving the total time.  I don't think we lose anything by splitting up the benchmarks, since when I add those 2 times together, I get approximately the same time as the total.

components/suggest/src/benchmarks/client.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::{rs::SuggestRemoteSettingsClient, Result};
+use parking_lot::Mutex;
+use remote_settings::{Client, GetItemsOptions, RemoteSettingsConfig, RemoteSettingsResponse};
+use std::collections::HashMap;
+
+/// Remotes settings client that runs during the benchmark warm-up phase.
+///
+/// This should be used to run a full ingestion.
+/// Then iit can be converted into a [RemoteSettingsBenchmarkClient], which allows benchmark code to exclude the network request time.
```
ncloudioj, line 13:
```suggestion
/// Then it can be converted into a [RemoteSettingsBenchmarkClient], which allows benchmark code to exclude the network request time.
```

ncloudioj (approved):
Like those changes, r+.

PR #6174, opened 2024-03-23
Speedup ingestion times

bendk:
Changed the ingestion code to prepare statements outside of any loops rather than creating them on each iteration.  Also, use tuples rather than named arguments.  This cuts ingestion time in half for the 2 record types that dominate the total time:

```
ingest/ingest-amp-wikipedia
                        time:   [446.26 ms 446.85 ms 447.53 ms]
                        change: [-51.727% -51.566% -51.408%] (p = 0.00 < 0.05)
                        Performance has improved.

ingest/ingest-amp-mobile
                        time:   [240.08 ms 240.25 ms 240.48 ms]
                        change: [-49.921% -49.801% -49.672%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Refactored the prepare and execute calls into their own struct.  This way it's easy to make sure that the arguments line up with correctly with the statement placeholders.  I also think it simplifies things to remove the specifics of each SQL statement from the general ingestion logic.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -23,7 +23,7 @@ use crate::{
     rs::{
         DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedAmpWikipediaSuggestion,
         DownloadedMdnSuggestion, DownloadedPocketSuggestion, DownloadedWeatherData,
-        SuggestRecordId,
+        DownloadedWikipediaSuggestion, SuggestRecordId,
     },
     schema::{clear_database, SuggestConnectionInitializer, VERSION},
     store::{UnparsableRecord, UnparsableRecords},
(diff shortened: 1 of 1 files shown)


bendk:
I saved the code I used to figure some of this stuff out in the [suggest-ingest-speedup-log](https://github.com/mozilla/application-services/compare/main...bendk:application-services:suggest-ingest-speedup-log) branch.  If we wanted to, we could also only go with speedup #1 in that branch, which would be a very simple fix.  However, I think the rest of the changes are good on their own anyways.

components/suggest/src/db.rs
```diff
@@ -1407,6 +1210,133 @@ impl<'a> FullKeywordInserter<'a> {
...
+        keyword: &str,
+        full_keyword_id: Option<i64>,
+        rank: usize,
+    ) -> Result<()> {
+        self.0
+            .execute((suggestion_id, keyword, full_keyword_id, rank))?;
+        Ok(())
+    }
+}
+
```
bendk, line 1339: We could define more of these statement types, for example to handle the inserting pocket/MDN/Weather data, or maybe even for the fetch/deletion code.  However, for this PR I only targetted the queries that I think would make a difference to the overall ingestion time for a fresh DB.

ncloudioj: Makes sense. Can you add a note for this to guide folks on when/how to adopt this treatment?

ncloudioj (approved):
Excellent! Thanks!

PR #6182, opened 2024-03-27
Bug 1892324 - Add the "close tabs" device command to the FxA client

linabutler:
This PR adds support for the "close tabs" command to our Rust FxA client. We'll use this new API in Firefox Android ([bug 1891157](https://bugzilla.mozilla.org/show_bug.cgi?id=1891157)) and iOS.

The implementation follows the same approach as our existing "send tab" command, down to the way we do encryption—I even kept the encryption key type names the same, to minimize churn.

  - Note:

### DIFF
--- a/components/fxa-client/src/device.rs
+++ b/components/fxa-client/src/device.rs
@@ -239,11 +239,10 @@ pub struct Device {
 /// executing these commands are encapsulated as part of the FxA Client component,
 /// so consumers simply need to select which ones they want to support, and can
 /// use the variants of this enum to do so.
-///
-/// In practice, the only currently-supported command is the ability to receive a tab.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub enum DeviceCapability {
     SendTab,
+    CloseTabs,
 }
 
 /// A client connected to the user's account.
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -528,7 +528,17 @@ interface FirefoxAccount {
   //
   [Throws=FxaError]
   void send_single_tab([ByRef] string target_device_id, [ByRef] string title, [ByRef] string url );
-  
+
+
+  /// Use device commands to close one or more tabs on another device.
+  ///
+  /// **💾 This method alters the persisted account state.**
+  ///
+  /// If a device on the account has registered the [`CloseTabs`](DeviceCapability::CloseTabs)
+  /// capability, this method can be used to close its tabs.
+  [Throws=FxaError]
+  void close_tabs([ByRef] string target_device_id, sequence<string> urls);
+
 
   // Get the URL at which to access the user's sync data.
   //
@@ -886,6 +896,14 @@ dictionary SendTabPayload {
   string stream_id = "";
 };
 
+/// The payload sent when invoking a "close tabs" command.
+//
+dictionary CloseTabsPayload {
+
+  /// The URLs of the tabs to close.
+  sequence<string> urls;
+};
+
 // An individual entry in the navigation history of a sent tab.
 //
 dictionary TabHistoryEntry {
@@ -981,10 +999,9 @@ enum FxaRustAuthState {
 // so consumers simply need to select which ones they want to support, and can
 // use the variants of this enum to do so.
 //
-// In practice, the only currently-supported command is the ability to receive a tab.
-//
 enum DeviceCapability {
   "SendTab",
+  "CloseTabs",
 };
 
 
@@ -1058,6 +1075,9 @@ interface IncomingDeviceCommand {
 
   // Indicates that a tab has been sent to this device.
   TabReceived(Device? sender, SendTabPayload payload );
+
+  /// Indicates that the sender wants to close one or more tabs on this device.
+  TabsClosed(Device? sender, CloseTabsPayload payload);
 };
 
 // Machinery for dry-run testing of FxaAuthStateMachine
--- a/components/fxa-client/src/internal/device.rs
+++ b/components/fxa-client/src/internal/device.rs
@@ -159,6 +159,7 @@ impl FirefoxAccount {
         command: &str,
         target: &Device,
         payload: &serde_json::Value,
+        ttl: Option<u64>,
     ) -> Result<()> {
         let refresh_token = self.get_refresh_token()?;
         self.client.invoke_command(
@@ -167,6 +168,7 @@ impl FirefoxAccount {
             command,
             &target.id,
             payload,
+            ttl,
         )
     }
 
--- a/components/fxa-client/src/internal/device.rs
+++ b/components/fxa-client/src/internal/device.rs
@@ -6,11 +6,15 @@ use std::collections::{HashMap, HashSet};
 
 pub use super::http_client::{GetDeviceResponse as Device, PushSubscription};
 use super::{
-    commands::{self, IncomingDeviceCommand},
+    commands::{
+        self,
+        send_tab::{PrivateSendTabKeys, PublicSendTabKeys},
+        IncomingDeviceCommand,
+    },
     http_client::{
         DeviceUpdateRequest, DeviceUpdateRequestBuilder, PendingCommand, UpdateDeviceResponse,
     },
-    telemetry, util, CachedResponse, FirefoxAccount,
+    scopes, telemetry, util, CachedResponse, FirefoxAccount,
 };
 use crate::{DeviceCapability, Error, LocalDevice, Result};
 use sync15::DeviceType;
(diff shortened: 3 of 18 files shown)


components/fxa-client/src/device.rs
```diff
@@ -239,11 +239,10 @@ pub struct Device {
 /// executing these commands are encapsulated as part of the FxA Client component,
 /// so consumers simply need to select which ones they want to support, and can
 /// use the variants of this enum to do so.
-///
-/// In practice, the only currently-supported command is the ability to receive a tab.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub enum DeviceCapability {
     SendTab,
+    CloseRemoteTabs,
```
linabutler, line 245: 🚲🎨🏡 time! Do we want to call this capability `CloseRemoteTabs`? `CloseTabs`? `CloseTab`? `CloseSynced{Tabs, Tab}`?

It doesn't affect the implementation, but might help to have a catchy, consistent name for our feature.

components/fxa-client/src/fxa_client.udl
```diff
@@ -886,6 +891,13 @@ dictionary SendTabPayload {
   string stream_id;
 };
 
+// ...
+//
+dictionary CloseTabsPayload {
+  // ...
+  sequence<string> urls;
```
linabutler, line 898: We talked for a bit about changing the payload to only close one tab at a time, and I wanted to capture that conversation here!

Our rationale for supporting just one tab is that we'll still need to:

* Raise the rate limit for FxA's `/v1/account/devices/invoke_command` endpoint, and...
* Collapse pushes that FxA sends to the receiving device's `pushCallback`, and...
* Coalesce the user-visible "N tabs closed" notifications on the receiving device.

...Regardless of whether we include one URL or multiple URLs in the payload.

Including multiple URLs means we need to batch closed tabs client-side, _and_ still handle the case where a batch might need to be split into multiple commands—if the user closes lots of tabs at once, or waits longer than the "undo" delay to close more, for example.

IOW, supporting multiple URLs complicates our implementation a bit, but doesn't avoid having to do the three things above.

components/fxa-client/src/internal/close_remote_tabs.rs
```diff
@@ -0,0 +1,98 @@
...
+                Ok(keys) => return Ok(keys),
+                Err(_) => {
+                    error_support::report_error!(
+                        "fxaclient-close-tabs-key-deserialize",
+                        "Could not deserialize Close Remote Tabs keys. Re-creating them."
+                    );
+                }
+            }
+        }
+        let keys = PrivateSendTabKeys::from_random()?;
```
linabutler, line 79: (Yep, the type name is still `PrivateSendTab` keys, even though these are the "close tabs" keys, because it uses the same encryption scheme 😅 I'd be happy to rename these to something generic in a separate commit, or a follow-up!)

components/fxa-client/src/internal/commands/close_remote_tabs.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::{Error, Result, ScopedKey};
+
+use super::{
+    super::device::Device,
+    send_tab::{PrivateSendTabKeysV1, PublicSendTabKeys, SendTabKeysPayload},
+};
+
+pub const COMMAND_NAME: &str = "https://identity.mozilla.com/cmd/close-uris";
+pub const COMMAND_TTL: u64 = 30 * 24 * 3600;
```
linabutler, line 17: FxA has a "default TTL" for [well-known commands](https://github.com/mozilla/fxa/blob/c89a3b0be02eb36f8db7661595845de31dc62983/packages/fxa-auth-server/lib/routes/devices-and-sessions.js#L27-L30), but I captured it here explicitly because:

1. `close-uri` isn't a well-known command yet.
2. We might want to tweak the value. Does 30 days make sense?
3. We'll want this value to match whatever TTL we use for our "pending closed tabs" entry in the tabs engine. If the TTL passes and the receiving device still hasn't synced its tabs, we'll want to remove the closed tab from the tabs engine's "pending closed tabs" table.

components/fxa-client/src/internal/commands/close_remote_tabs.rs
```diff
@@ -0,0 +1,97 @@
...
+use serde_derive::*;
+
+use crate::{Error, Result, ScopedKey};
+
+use super::{
+    super::device::Device,
+    send_tab::{PrivateSendTabKeysV1, PublicSendTabKeys, SendTabKeysPayload},
+};
+
+pub const COMMAND_NAME: &str = "https://identity.mozilla.com/cmd/close-uris";
```
linabutler, line 16:
```suggestion
pub const COMMAND_NAME: &str = "https://identity.mozilla.com/cmd/close-uri";
```

https://phabricator.services.mozilla.com/D204556 calls this `close-uri`, and it also makes sense for it to be singular if we only send one closed tab URL per command.

components/fxa-client/src/internal/device.rs
```diff
@@ -325,6 +341,26 @@ impl FirefoxAccount {
             None => Err(Error::NoCurrentDeviceId),
         }
     }
+
+    /// Generate the Send Tab command to be registered with the server.
```
linabutler, line 345:
```suggestion
    /// Generate the command to be registered with the server for the given capability.
```

components/fxa-client/src/device.rs
linabutler: I like `CloseTabs` (which is what I used at first! :joy:) or `CloseTab`, mirroring @mhammond's [comment](https://phabricator.services.mozilla.com/D204556?id=836796#inline-1137804) on the Desktop patch—it sidesteps the question of "which is the remote device and which is the local device".

mhammond: I hate naming :) In this case I think `CloseTabs` is correct for exactly the reason you say - but even then it's slightly ambigious - does the capability say "I can have my tabs closed" or "I can close other device's tab"? Regardless, I think `CloseTabs` is perfectly reasonable and better than the other options.

components/fxa-client/src/fxa_client.udl
```diff
@@ -528,7 +528,12 @@ interface FirefoxAccount {
   //
   [Throws=FxaError]
   void send_single_tab([ByRef] string target_device_id, [ByRef] string title, [ByRef] string url );
-  
+
+
+  // ...
```
mhammond, line 533: nano-nit from the future - triple-slashes in UDL will cause the comments to be emitted as docstrings. Nothing consumes them yet, and no one else does this yet, so not at all clear this should, but at some point we might as well start doing that :)

components/fxa-client/src/fxa_client.udl
mhammond: Yeah, while that's true, there's also no real harm in keeping this a sequence, right? (ie, I'm not sure if you are suggesting we should drop the sequence here or not - I'd say "no" just as risk mitigation, but could be convinced otherwise)

Also wonder if "Payload" is the correct term here? Payload implies "over the wire" but we would be free to transform the "payload" into this "client request". If it wasn't so verbose `IncomingCloseTabsRequest` captures everything the reader needs to know (ie, it's not used for outgoing close requests etc)

[eek, just noticed `Payload` is already used for SendTab, so what you wrote is consistent, so feel free to completely ignore the above :)]

components/fxa-client/src/internal/commands/close_remote_tabs.rs
mhammond: As above, it's not clear if we want to fully commit to this yet. In general, and IMO, the name should match the payload (ie, plural if the payload holds a vec, singular otherwise)

components/fxa-client/src/internal/commands/close_remote_tabs.rs
mhammond: The TTL issue is interesting, but 30 days does seem excessive purely from the user's POV - closing a tab they requested be closed 30 days ago seems very likely to surpise the user? But no less surprising than a tab coming in, so I think this is probably fine in practice.

components/fxa-client/src/device.rs
skhamis: I like `CloseTab(s)` as well! Keeps it consistent with SendTab!

components/fxa-client/src/internal/close_remote_tabs.rs
skhamis: A follow-up after the initial feature landed is definitely reasonable, otherwise I wonder if a comment here just mentioning we're using send tabs for convenience is good enough!

skhamis (commented):
This is looking great!

components/fxa-client/src/fxa_client.udl
linabutler: Closing the loop on this:

1. Yep, let's keep this a sequence, so that we can have the best of both worlds.
2. I like `Request` a lot, but I agree; I think we'd want to change Send Tab to use the same nomenclature. Maybe a good first bug for someone? 😉 I think the monorepofication will make these kinds of renames a lot easier, too.

components/fxa-client/src/device.rs
linabutler: `CloseTabs` it is! 😁

components/fxa-client/src/internal/commands/close_remote_tabs.rs
linabutler: The capability URL that Desktop uses is `https://identity.mozilla.com/cmd/close-uri/v1`, so I changed this one to match.

components/fxa-client/src/fxa_client.udl
linabutler: Done! Everything should have `///` docs now.

linabutler:
Alrighty, I think this is all ready for review!

mhammond (approved):
sorry for losing track of this!

PR #6183, opened 2024-03-28
Relevancy: implement ingestion

tiftran:
rebased/dependent on #6220

  - Note:

### DIFF
--- a/components/relevancy/src/lib.rs
+++ b/components/relevancy/src/lib.rs
@@ -55,8 +55,21 @@ impl RelevancyStore {
     ///
     ///  This method may execute for a long time and should only be called from a worker thread.
     #[handle_error(Error)]
-    pub fn ingest(&self, _top_urls_by_frecency: Vec<String>) -> ApiResult<()> {
-        ingest::ensure_interest_data_populated(&self.db)
+    pub fn ingest(&self, top_urls_by_frecency: Vec<String>) -> ApiResult<InterestVector> {
+        ingest::ensure_interest_data_populated(&self.db)?;
+        self.classify(top_urls_by_frecency)
+    }
+
+    pub fn classify(&self, top_urls_by_frecency: Vec<String>) -> Result<InterestVector> {
+        // For experimentation purposes we are going to return an interest vector.
+        // Eventually we would want to store this data in the DB and incrementally update it.
+        let mut interest_vector = InterestVector::default();
+        for url in top_urls_by_frecency {
+            let interest_count = self.db.read(|dao| dao.get_url_interest_vector(&url))?;
+            interest_vector = interest_vector + interest_count;
+        }
+
+        Ok(interest_vector)
     }
 
     /// Calculate metrics for the validation phase
@@ -86,3 +99,45 @@ pub struct InterestMetrics {
 }
 
 uniffi::include_scaffolding!("relevancy");
+
+#[cfg(test)]
+mod test {
+    use crate::url_hash::hash_url;
+
+    use super::*;
+
+    #[test]
+    fn test_ingest() {
+        let top_urls = vec![
+            "https://food.com/".to_string(),
+            "https://hello.com".to_string(),
+            "https://pasta.com".to_string(),
+            "https://dog.com".to_string(),
+        ];
+        let relevancy_store =
+            RelevancyStore::new("file:test_store_data?mode=memory&cache=shared".to_owned());
+        relevancy_store
+            .db
+            .read_write(|dao| {
+                dao.add_url_interest(hash_url("https://food.com").unwrap(), Interest::Food)?;
+                dao.add_url_interest(
+                    hash_url("https://hello.com").unwrap(),
+                    Interest::Inconclusive,
+                )?;
+                dao.add_url_interest(hash_url("https://pasta.com").unwrap(), Interest::Food)?;
+                dao.add_url_interest(hash_url("https://dog.com").unwrap(), Interest::Animals)?;
+                Ok(())
+            })
+            .expect("Insert should succeed");
+
+        assert_eq!(
+            relevancy_store.ingest(top_urls).unwrap(),
+            InterestVector {
+                inconclusive: 1,
+                animals: 1,
+                food: 2,
+                ..InterestVector::default()
+            }
+        );
+    }
+}
--- a/components/relevancy/src/relevancy.udl
+++ b/components/relevancy/src/relevancy.udl
@@ -22,7 +22,7 @@ interface RelevancyStore {
 
     // Ingest the top URLs by frequency to build up the user's interest vector
     [Throws=RelevancyApiError]
-    void ingest(sequence<string> top_urls);
+    InterestVector ingest(sequence<string> top_urls);
 
     // Calculate metrics for the user's interest vector in order to measure how strongly we're
     // identifying interests.  See the `InterestMetrics` struct for details.
--- a/components/relevancy/src/interest.rs
+++ b/components/relevancy/src/interest.rs
@@ -136,6 +136,34 @@ pub struct InterestVector {
     pub travel: u32,
 }
 
+impl std::ops::Add for InterestVector {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        Self {
+            inconclusive: self.inconclusive + other.inconclusive,
+            animals: self.animals + other.animals,
+            arts: self.arts + other.arts,
+            autos: self.autos + other.autos,
+            business: self.business + other.business,
+            career: self.career + other.career,
+            education: self.education + other.education,
+            fashion: self.fashion + other.fashion,
+            finance: self.finance + other.finance,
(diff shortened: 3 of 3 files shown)


components/relevancy/src/lib.rs
```diff
@@ -47,9 +57,53 @@ impl RelevancyStore {
     ///
     ///  This method may execute for a long time and should only be called from a worker thread.
     #[handle_error(Error)]
-    pub fn ingest(&self, _top_urls_by_frecency: Vec<String>) -> ApiResult<()> {
+    pub fn ingest(&self, top_urls_by_frecency: Vec<String>) -> ApiResult<HashMap<Interest, u32>> {
```
tiftran, line 60: For now, I just have the frequency of the number of interest occurrences, but we would probs start discussing which method we want to use to express interest?

components/relevancy/src/lib.rs
```diff
@@ -47,9 +57,53 @@ impl RelevancyStore {
...
+                        *interest_counter.entry(classified_url.interest).or_insert(0) += 1;
+                    }
+                    false => {
+                        *interest_counter.entry(Interest::Inconclusive).or_insert(0) += 1;
+                    }
+                }
+            } else {
+                *interest_counter.entry(Interest::Inconclusive).or_insert(0) += 1;
+            }
+        }
```
tiftran, line 83: I tried to use Inconclusive in the case we don't have the url classified or we can't hash it, but it looks goofy...

components/relevancy/src/lib.rs
```diff
@@ -47,9 +57,53 @@ impl RelevancyStore {
...
+
+        let hashed_top_urls: Vec<UrlHash> = top_urls_by_frecency
+            .into_iter()
+            .filter_map(|url| hash_url(&url))
+            .collect();
+
+        let classified_urls = RelevancyStore::get_url_interest_data();
+        for classified_url in classified_urls {
+            if let Some(url) = &classified_url.url {
+                match hashed_top_urls.contains(url) {
```
bendk, line 72: I feel like this should be reversed.  We should be looping over the top urls, and looking up them up in in the classified URLs hashmap (or database, see below).

I'm not totally sure of the logic though.  What would happen if you added `https://hello2.com` to the list of top URLs in the test below?  I think that should increase the inconclusive count by one, but the counts will stay the same.

components/relevancy/src/lib.rs
```diff
@@ -47,9 +57,53 @@ impl RelevancyStore {
...
+                }
+            } else {
+                *interest_counter.entry(Interest::Inconclusive).or_insert(0) += 1;
+            }
+        }
+        Ok(interest_counter)
+    }
+
+    /// Temp method until we actually retrieve data from RS
+    pub fn get_url_interest_data() -> Vec<URLInterest> {
```
bendk, line 88: Using a temporary method for this makes a lot of sense to me.

I'm my comment above was right, I think this should return a HashMap.  Alternatively, maybe this could insert rows into a database.  I think using a db would be closer to how things will work in the final product and there are some methods in `db.rs` that should work.  Maybe it's better to do that in a follow-up PR though.  WDYT?

components/relevancy/src/lib.rs
```diff
@@ -79,3 +133,26 @@ pub struct InterestMetrics {
...
+            "https://dog.com".to_string(),
+        ];
+        let relevancy_store =
+            RelevancyStore::new("file:test_store_data?mode=memory&cache=shared".to_owned())
+                .unwrap();
+        let mut result: HashMap<Interest, u32> = HashMap::new();
+        result.insert(Interest::Food, 2);
+        result.insert(Interest::Inconclusive, 1);
+        result.insert(Interest::Animals, 1);
+        assert_eq!(relevancy_store.ingest(top_urls).unwrap(), result);
```
bendk, line 156: My favorite way to construct a hashmap is by iterating over an array.  I think you could rewrite this something like:

```
 assert_eq!(relevancy_store.ingest(top_urls).unwrap(), HashMap::from([
    (Interest::Food, 2),
    (Interest::Inconclusive, 1),
    (Interest::Animals, 1),
]);
```

I don't think this code needs to change, I just wanted to share the idiom.

components/relevancy/src/lib.rs
bendk: That makes a lot of sense to me.

components/relevancy/src/lib.rs
bendk: Yeah, maybe we should just  skip the URL in this case.

bendk (changes requested):
I think it's right overall, but the URLs your looping over and the ones in the hash map should be reversed.  Or maybe confused and not picturing the flow right.  Updating the test should determine which one it is.

components/relevancy/src/lib.rs
tiftran: oooh i didn't know this could be done this way!

components/relevancy/src/lib.rs
ncloudioj: Agreed, we'd definitely want to do more things around the db for the final product instead of loading everything into memory for each classification.

For interest data ingestion, I think we'd periodically ingest the interest data from Remote Settings and persist it to the db, we will figure out a reasonable schema to facilitate the subsequent interest classifications.

For classification, we will query against the db for each URL hash to get its associated interest category if any. Note that as we've decided to store raw URL hashes (as opposed to other compact structures e.g. Bloom filters ) for each interest on Remote Settings, each input URL/hash would require a point lookups against the db, which should be fine for small/medium size inputs. There are ways to reduce the overhead for large inputs and we can look into those in the future.

Does that align with what's on your mind? Would love to hear your thoughts.

components/relevancy/src/lib.rs
```diff
@@ -47,9 +57,53 @@ impl RelevancyStore {
...
+                    }
+                    false => {
+                        *interest_counter.entry(Interest::Inconclusive).or_insert(0) += 1;
+                    }
+                }
+            } else {
+                *interest_counter.entry(Interest::Inconclusive).or_insert(0) += 1;
+            }
+        }
+        Ok(interest_counter)
```
ncloudioj, line 84: [Unrelated to this PR, just for discussion] While interest persistence is out-of-scope for the validation phase, it's still useful to think about how we might want to support that later? Questions like: should the component automatically store the derived user interests to the db and delete the old ones? Or we still want to return the result to the consumer and provide another API for persistence? I feel like the former makes more sense as it'll make consumer's life easier. But there could be other legit use cases for the latter.

components/relevancy/src/lib.rs
bendk: Yup, this is what I was thinking as well.

components/relevancy/src/lib.rs
bendk: I was definitely thinking the former, but it could also be interesting to have separate endpoints for calculating the interest vector and persisting it.  Maybe we could make the internal API work like that, then it would be easy to also expose that API to the consumers in the future if they wanted it.

components/relevancy/src/lib.rs
tiftran: Lets take in a Vec<Struct> instead of string

tiftran: lets move this logic to a separate module and call it `classify()`

components/relevancy/src/lib.rs
```diff
@@ -55,8 +54,19 @@ impl RelevancyStore {
     ///
     ///  This method may execute for a long time and should only be called from a worker thread.
     #[handle_error(Error)]
-    pub fn ingest(&self, _top_urls_by_frecency: Vec<String>) -> ApiResult<()> {
-        ingest::ensure_interest_data_populated(&self.db)
+    pub fn ingest(&self, _top_urls_by_frecency: Vec<String>) -> ApiResult<InterestVector> {
+        let _ = ingest::ensure_interest_data_populated(&self.db);
```
bendk, line 58: Does this need the `let _`?

components/relevancy/src/relevancy.udl
```diff
@@ -22,7 +22,7 @@ interface RelevancyStore {
 
     // Ingest the top URLs by frequency to build up the user's interest vector
     [Throws=RelevancyApiError]
-    void ingest(sequence<string> top_urls);
+    InterestVector ingest(sequence<string> top_urls);
```
bendk, line 25: For the experiment, I think it's fine to return an interest vector.  Eventually I think we want to store this data in the DB and incrementally update it when we get new URLs.  For now, how about keeping the code the same and adding a comment about our plans.

bendk (changes requested):
Looking good, just a couple minor suggestions.

components/relevancy/src/lib.rs
tiftran: Oops

bendk (approved):
Changes look good, thanks!

ncloudioj (approved):
r+wc.

PR #6188, opened 2024-04-03
[WIP] Remote Tabs: Filtering out pending closed tabs from getAll()

skhamis:
Initial draft of adding pending remote tabs to a new table in the tabs DB to and filtering them out in our main API call (getAll()). 

- One thing this doesn't have is removing tabs from the table upon either successful completion of those tabs being closed or removing it after X amount of time that other client synced. I anted to see if this is the route we do want to take before going too far down.

Thoughts from all welcome but cc'ing the people most likely to have more context: @linabutler @mhammond 

  - Note:


skhamis:
Ok decided to just add the removing method just so we can see how that looks (with a test to boot).
I also chose 48hrs as a starting point for removing rows in the new DB. So any pending tabs that are 48hrs old from the last time that other engine has synced will be removed. I came to 48hrs because:
1. Most of the time the pushes will happen instantaneously
2. Our current polling for any missed commands is 24hrs (potentially lowered too)
3. 48hrs seems enough padding in-case any serious delay to getting the delay

However totally open to any other number here!

skhamis:
Tested with https://phabricator.services.mozilla.com/D204556 and works pretty well! The one hiccup is reopening the same urls will still cause them not to show up in the synced menu (unless it's 48hrs later). So I'll need to drop that down to maybe an hour after that client has synced? Curious any thoughts!

mhammond (approved):
I'm leaving these comments in, but I'll approve and land this as-is, then make those changes in the next iteration Lina and I discussed.

PR #6191, opened 2024-04-04
Deprecate `RemoteSettingsConfig.server_url` in favor of a new, type-safe `RemoteSettingsConfig.server`

linabutler:
The `RemoteSettingsConfig.server_url` option has a couple of small pain points for consumers:

1. It forces them to hard-code the URLs for other environments (Fenix does this for stage [here](https://searchfox.org/mozilla-central/rev/e1e4a33e82ee1d278df238cf0896b7358a4bc359/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/Core.kt#631-632) and [here](https://searchfox.org/mozilla-central/rev/e1e4a33e82ee1d278df238cf0896b7358a4bc359/mobile/android/focus-android/app/src/main/java/org/mozilla/focus/telemetry/GleanMetricsService.kt#65-66), for example). We can make things a bit smoother for these callers by having the Remote Settings component hard-code these URLs instead.
2. It forces other consumers of Remote Settings, like Suggest, to override _all_ config options—including the bucket and collection name—even if they just wanted to override the server URL. For example, if Fenix wanted to ingest suggestions from the staging server, it would need to also specify `collectionName: "quicksuggest"`, even though the collection name is an implementation detail of the Suggest component.

This PR tries to make that situation better by:

1. Introducing a new type, `RemoteSettingsServer`, that allows specifying a well-known (`Prod`, `Stage`, `Dev`) or `Custom` server.
2. Adding `RemoteSettingsServer.server` and `SuggestStoreBuilder.remote_settings_server()`.
3. Deprecating `RemoteSettingsConfig.server_url` and `SuggestStoreBuilder.remote_settings_config()`. These APIs are still available, but can't be used together with the new APIs—consumers must specify one or the other, or neither (to use the default `Prod` server), but not both.

This is the first step to wiring up Fenix's "use Remote Settings {production / stage} server" switch to Suggest ([bug 1889495](https://bugzilla.mozilla.org/show_bug.cgi?id=1889495)).

  - Note:

### DIFF
--- a/components/nimbus/examples/experiment.rs
+++ b/components/nimbus/examples/experiment.rs
@@ -17,7 +17,7 @@ fn main() -> Result<()> {
             MetricsHandler,
         },
         AppContext, AvailableRandomizationUnits, EnrollmentStatus, NimbusClient,
-        NimbusTargetingHelper, RemoteSettingsConfig,
+        NimbusTargetingHelper, RemoteSettingsConfig, RemoteSettingsServer,
     };
     use std::collections::HashMap;
     use std::io::prelude::*;
@@ -218,7 +218,10 @@ fn main() -> Result<()> {
 
     // initiate the optional config
     let config = RemoteSettingsConfig {
-        server_url: Some(server_url.to_string()),
+        server: Some(RemoteSettingsServer::Custom {
+            url: server_url.to_string(),
+        }),
+        server_url: None,
         bucket_name: None,
         collection_name: collection_name.to_string(),
     };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Suggest
 - Improved full keyword display for AMP suggestions
 - `SuggestStoreBuilder.cache_path` is now deprecated because we no longer use the cache path.
+- `SuggestStoreBuilder.remote_settings_config` is deprecated in favor of `remote_settings_server`, because `remote_settings_config` forced consumers that wanted to override the Remote Settings server to also specify the bucket and collection names.
 
 ### Remote Settings
 - `RemoteSettingsConfig.server_url` is deprecated in favor of `server`, which is a `RemoteSettingsServer` instead of a string.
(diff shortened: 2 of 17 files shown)


components/nimbus/examples/experiment.rs
```diff
@@ -218,7 +218,10 @@ fn main() -> Result<()> {
 
     // initiate the optional config
     let config = RemoteSettingsConfig {
-        server_url: Some(server_url.to_string()),
+        server: Some(RemoteSettingsServer::Custom {
+            url: server_url.to_string(),
+        }),
+        server_url: None,
         bucket_name: None,
         collection_name: collection_name.to_string(),
```
bendk, line 226: What do you think about using `..RemoteSettingsConfig::default()` at the end of the struct initialization?  It could be a nice way to handle `server_url` being deprecated and eventually being removed.

bendk (approved):
This change is great.

What's the deprecation plan for `server_url`?  My preference would be to remove it in the near future if possible.

linabutler:
> What's the deprecation plan for `server_url`? My preference would be to remove it in the near future if possible.

Same! I think only Fenix and Desktop use it now, so we can migrate it as soon as this PR lands in a nightly, and then remove `server_url` in the next two weeks or so.I'm not sure how our breaking change flow works post-Android monorepofication, and this change felt relatively easy to make backward-compatible! 😅

components/nimbus/examples/experiment.rs
linabutler: Hmmm, I don't think that'll work, because `RemoteSettingsConfig` doesn't implement `Default`. We could have it derive `Default`, but we don't have a suitable default that we can use for `collection_name`, since that's the only setting that _is_ required.

I wonder if moving `RemoteSettingsConfig` to a builder-style API would be a good long-term direction, though?

```rs
let b = RemoteSettingsConfig::with_collection_name("quicksuggest".into());
b.server(RemoteSettingsServer::Stage)?.bucket_name("main".into()); // Optional.
let client = b.build();
```

linabutler: We could even do that in a backward-compatible way by renaming `RemoteSettingsConfig` to `RemoteSettingsBuilder`, deprecating `RemoteSettingsConfig`, and migrating callers over to `RemoteSettingsBuilder`.

I'm really liking the builder pattern you added for Suggest!

PR #6200, opened 2024-04-09
Bug 1889700 - interruption support for the Relevancy component

bendk:
Added the `sql_support::LazyDb` struct.  This handles various details of a lazily-opened DB.  The main issue is connecting the dbs the interrupt code as their are opened.

Switched the relevancy code to using `LazyDb`, switched the `close` method to interrupt any in-progress queries before closing the DB, and added a manual `interrupt` function.  Also, if we expose the `interrupt_support::shutdown`, then we can have shutdown support for it.

"interrupt/shutdown" support means being able to interrupt SQL queries and also integrating with a cooperative interrupt system.  Added some simple code to check for interruption when we're populating the interest data.

  - Note:

### DIFF
--- a/components/support/sql/src/lazy.rs
+++ b/components/support/sql/src/lazy.rs
@@ -1,0 +1,151 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::open_database::{open_database_with_flags, ConnectionInitializer, Error};
+use interrupt_support::{register_interrupt, SqlInterruptHandle, SqlInterruptScope};
+use parking_lot::{MappedMutexGuard, Mutex, MutexGuard};
+use rusqlite::{Connection, OpenFlags};
+use std::{
+    path::{Path, PathBuf},
+    sync::{Arc, Weak},
+};
+
+/// Lazily-loaded database with interruption support
+///
+/// In addition to the [Self::interrupt] method, LazyDb also calls
+/// [interrupt_support::register_interrupt] on any opened database.  This means that if
+/// [interrupt_support::shutdown] is called it will interrupt this database if it's open and
+/// in-use.
+pub struct LazyDb<CI> {
+    path: PathBuf,
+    open_flags: OpenFlags,
+    connection_initializer: CI,
+    // Note: if you're going to lock both mutexes at once, make sure to lock the connection mutex
+    // first.  Otherwise, you risk creating a deadlock where two threads each hold one of the locks
+    // and is waiting for the other.
+    connection: Mutex<Option<Connection>>,
+    // It's important to use a separate mutex for the interrupt handle, since the whole point is to
+    // interrupt while another thread holds the connection mutex.  Since the only mutation is
+    // setting/unsetting the Arc, maybe this could be sped up by using something like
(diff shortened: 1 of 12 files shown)


components/support/sql/src/lazy.rs
```diff
@@ -0,0 +1,141 @@
...
+            connection_initializer,
+            connection: Mutex::new(None),
+            interrupt_handle: Mutex::new(None),
+        }
+    }
+
+    /// Lock the database mutex and get a connection and interrupt scope.
+    ///
+    /// If the connection is closed, it will be opened.
+    pub fn lock(&self) -> Result<(MappedMutexGuard<'_, Connection>, SqlInterruptScope), Error> {
```
ncloudioj, line 50: Calling `lock()` twice (or any other method, i.e. `close()`, that requires a lock on `self.connection` ) from the same thread would cause a dead lock unless the `MappedMutexGuard` gets dropped before the 2nd call. We might want to note that in the docstring to avoid foot-guns.

components/support/sql/src/open_database.rs
```diff
@@ -356,17 +358,23 @@ pub mod test_utils {
...
 #[cfg(test)]
-mod test {
+pub mod test {
     use super::test_utils::MigratedDatabaseFile;
     use super::*;
     use std::cell::RefCell;
     use std::io::Write;
 
-    struct TestConnectionInitializer {
+    pub struct TestConnectionInitializer {
```
ncloudioj, line 367: Does it make sense to move this to the pub mod `test_utils`?

ncloudioj (changes requested):
Overall, this looks good. Dropped a few comments for your consideration.

components/support/sql/src/lazy.rs
ncloudioj: Looks good, thanks for adding that!

Assuming that the lock will eventually be released, I think calling `lock()` and `close()` from two separate threads should be fine, correct? Calling them from the same thread would definitely be problematic.

components/support/sql/src/open_database.rs
```diff
@@ -356,120 +470,11 @@ pub mod test_utils {
 }
 
 #[cfg(test)]
-mod test {
-    use super::test_utils::MigratedDatabaseFile;
+pub mod test {
```
ncloudioj, line 473: Can't it be private now?

components/support/sql/src/lazy.rs
```diff
@@ -0,0 +1,151 @@
...
+        // If we see an Interrupted error when beginning the scope, it means that we're in shutdown
+        // mode.
+        result.map_err(|_| Error::Shutdown)
+    }
+
+    /// Close the database if it's open
+    ///
+    /// Pass interrupt=true to interrupt any in-progress queries before closing the database.
+    ///
+    /// Do not call `close` if you already have a lock on the database, as this will cause a
```
ncloudioj, line 101: Like mentioned above, this might sound too scary to consumers. There are still a few cases that calling `close()` is OK while a lock is held. For example, from a separate thread, they could call `interrup()` followed by `close()` as the former will interrupt in-progress queries and the lock will be released short after. Even if they don't interrupt, the close call will proceed whenever the lock is released. It'll block, but won't cause a deadlock.

ncloudioj (approved):
LGTM, thanks!

bendk:
Good call on the deadlock doc strings.  I updated them to make it clear it's only an issue if you are performing the operations on the same thread.

PR #6208, opened 2024-04-23
Fixes tests that depend  map ordering

tarikeshaq:
On my new machine, I got a few failures running `cargo test` because some tests were depending on the ordering of `serde_json::Map` - this PR changes those tests do not depend on the ordering of keys in a JSON object as it's not guaranteed

  - Note:


components/remote_settings/src/client.rs
```diff
@@ -759,91 +758,56 @@ mod test {
             .sort("a", SortOrder::Ascending)
             .limit(3);
 
-        assert!(http_client.get_records_raw_with_options(&options).is_ok());
-        expect![[r#"
```
bendk, line 763: Is there any way to keep using the `expect!` macro?  It has the nice feature that if some of the record fields change, you can auto-update the tests, sort-of like TRYBUILD.  https://docs.rs/expect-test/latest/expect_test/.

If not, no big deal I've kind of wanted to refactor these anyways.

bendk (approved):

components/remote_settings/src/client.rs
tarikeshaq: Hmmm looking more onto why I was only getting those failures when running `cargo test` from application-services but not from the crate itself, I realized it was because in nimbus-fml, we add the `preserve_order` feature to serde_json, and because features are additive when running against the whole workspace all the crates get that feature. 

That feature uses an `IndexMap` instead of a `BTreeMap` within serde_json to preserve the order of serialization and deserialization, but it wouldn't be the same order a `BTreeMap` has (which is what this test would be verifying) - We might be able to disable that feature but I **assume** it was added for a good reason

Anywhoo, back to your question: Yes we can do that! I added a dev-dependency for those crates that depended on the ordering to also enable `preserve_order` so their result is consistent whether the test is being run through the workspace or the crate itself

components/nimbus/src/tests/test_enrollment.rs
```diff
@@ -1205,9 +1205,6 @@ fn test_evolver_feature_can_have_only_one_experiment() -> Result<()> {
         json!({ "text": "OK then", "number": 42})
     );
 
-    let string = serde_json::to_string(&enrolled_feature.feature.value).unwrap();
```
tarikeshaq, line 1208: This test was just redundant, as we test the same thing a few lines above without converting to a string

skhamis (approved):
Thanks for doing this -- I think we should land! LGTM!

PR #6211, opened 2024-04-24
Bug 1893086 - Port the potential impressions framework to the suggest component

0c0w3:
Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1893086 for details. Notes:

* I'm not sure who to request review from these days, so please forward the request if you'd like!
* This leaves a keyword-matching strategy as a to-do. I think we'll want to use fts5 but I'm not sure. I'd like to do that as a follow-up.
* This modifies `SuggestionQuery`, which breaks most existing tests in store.rs. To prevent that from happening again in the future as much as possible, I added a `SuggestionQueryBuilder` and a couple macros.
* `SuggestionQueryBuilder` isn't exposed to consumers. Should it be? I tried adding it to the udl but that requires `Arc`-ifying it, which makes it a pain to use in tests (I think?). Can we do that later if it turns out to be helpful?
* Since this does modify `SuggestionQuery`, do I need to do something to avoid breaking mobile? I read the breaking changes info but tbh I don't understand it.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -22,8 +22,8 @@ use crate::{
     provider::SuggestionProvider,
     rs::{
         DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedAmpWikipediaSuggestion,
-        DownloadedMdnSuggestion, DownloadedPocketSuggestion, DownloadedWeatherData,
-        DownloadedWikipediaSuggestion, SuggestRecordId,
+        DownloadedMdnSuggestion, DownloadedPhantomSuggestion, DownloadedPocketSuggestion,
+        DownloadedWeatherData, DownloadedWikipediaSuggestion, SuggestRecordId,
     },
     schema::{clear_database, SuggestConnectionInitializer, VERSION},
     store::{UnparsableRecord, UnparsableRecords},
@@ -207,6 +207,7 @@ impl<'a> SuggestDao<'a> {
                     SuggestionProvider::Yelp => self.fetch_yelp_suggestions(query),
                     SuggestionProvider::Mdn => self.fetch_mdn_suggestions(query),
                     SuggestionProvider::Weather => self.fetch_weather_suggestions(query),
+                    SuggestionProvider::Phantom => self.fetch_phantom_suggestions(query),
                 }?;
                 acc.extend(suggestions);
                 Ok(acc)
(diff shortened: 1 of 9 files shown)


components/suggest/src/store.rs
```diff
@@ -708,6 +716,60 @@ mod tests {
...
+            SuggestionQueryBuilder::new($keyword_str.into(), $providers_vec)
+        };
+    }
+
+    /// Example: query!("literal search string", vec![SuggestionProvider::Amp])
+    macro_rules! query {
+        ($keyword_str:literal, $providers_vec:expr $(,)?) => {
+            query_builder!($keyword_str, $providers_vec).build()?
+        };
+    }
```
bendk, line 771: Could this be a function?  If so, I think I'd rather use that than a macro to keep things simple.

components/suggest/src/suggestion.rs
```diff
@@ -91,20 +96,9 @@ impl PartialOrd for Suggestion {
...
-            Suggestion::Amp { score, .. }
-            | Suggestion::Pocket { score, .. }
-            | Suggestion::Amo { score, .. } => score,
-            _ => &DEFAULT_SUGGESTION_SCORE,
-        };
-        b_score
-            .partial_cmp(a_score)
+        other
+            .score()
+            .partial_cmp(&self.score())
```
bendk, line 101: Nice cleanup!

components/suggest/src/db.rs
```diff
@@ -663,6 +664,57 @@ impl<'a> SuggestDao<'a> {
...
+                  s.id
+                FROM
+                  suggestions s
+                JOIN
+                  phantom_custom_details d
+                  ON d.suggestion_id = s.id
+                WHERE
+                  s.provider = :provider
+                  AND d.type = :type
+                "#,
```
bendk, line 701: To double-check by understanding: Once we had the keyword strategy, then there would also be a JOIN and WHERE on the keywords table right?

components/suggest/src/store.rs
```diff
@@ -708,6 +716,60 @@ mod tests {
...
     use crate::SuggestionProvider;
 
+    /// Builder for `SuggestionQuery` to avoid breaking existing tests when it's
+    /// modified. See also the query macros below.
+    pub(crate) struct SuggestionQueryBuilder {
+        keyword: String,
+        providers: Vec<SuggestionProvider>,
+        limit: Option<i32>,
+        phantom_suggestion_type: Option<String>,
+    }
```
bendk, line 726: I like builders in Rust, but I notice that all of these fields have default values.  We could also use the struct update pattern:
```
SuggestionQuery {
  keyword: "foo",
  providers: vec![...],
  ..SuggestionQuery::default()
}
```

I'm not sure if this is better or worse, I'm happy with whichever one you like.  My one opinion is that I don't think we need to expose this in the UDL, since Kotlin/Swift have similar ways to do the same thing.

bendk (approved):
Looks right to me, I just had some minor suggestions that you can take or leave.

0c0w3:
Abandoning this for #6343

PR #6220, opened 2024-04-26
Relevancy: add fetch url interest data from RS

tiftran:

  - Note:


components/relevancy/src/populate_interests.rs
```diff
@@ -2,69 +2,119 @@
...
+
     db.read_write(move |dao| {
-        for chunk in interest_data.chunks(WRITE_CHUNK_SIZE) {
-            for (url_hash, interest) in chunk {
-                dao.add_url_interest(*url_hash, *interest)?;
+        match fetch_interest_data() {
+            Ok(data) => insert_interest_data(data, dao),
+            Err(e) => {
+                log::warn!("error fetching interest data: {e}");
+                return Err(Error::FetchInterestDataError);
```
bendk, line 25: Could you move this outside the `read_write()` call?  I think this way will keep a DB transaction open while the RS data is being fetched/parsed.

components/relevancy/src/populate_interests.rs
```diff
@@ -161,4 +220,76 @@ mod test {
...
+        let parsed_attachment_data: Vec<RelevancyAttachmentData> =
+            serde_json::from_str(attachment_string).unwrap();
+        let mut data: Vec<(Interest, UrlHash)> = vec![];
+        parse_record(record_fields, parsed_attachment_data, &mut data).unwrap();
+        assert_eq!(
+            data,
+            vec![
+                (
+                    Interest::Animals,
+                    [184, 160, 153, 251, 13, 250, 7, 82, 96, 170, 96, 31, 110, 46, 231, 250]
```
bendk, line 286: Nit: maybe replace this with `hash_url("the-url").unwrap()`.  I think that will be easier to understand later on.

components/relevancy/src/rs.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct RelevancyRecord {
+    #[serde(rename = "type")]
+    pub record_type: String,
+    pub record_custom_details: RecordCustomDetails,
+}
+
+// Custom details related to category of the record.
+#[derive(Clone, Debug, Deserialize)]
+pub struct RecordCustomDetails {
+    pub category_to_domains: CategoryToDomains,
```
bendk, line 22: This isn't a comment on the code, and I don't think we need to address this now, but this extra level of nesting seems unnecessary to me.  Couldn't we just store the `CategoryToDomains` fields in `RecordCustomDetails`?  IRC `CategoryToDomains` used to be in a list, but now that it's a single child object I don't see the point.

components/relevancy/src/error.rs
```diff
@@ -29,6 +29,15 @@ pub enum Error {
 
     #[error("Invalid interest code: {0}")]
     InvalidInterestCode(u32),
+
+    #[error("Remote Setting Error: {0}")]
+    RemoteSettingsError(#[from] remote_settings::RemoteSettingsError),
+
+    #[error("Serde Json Error: {0}")]
+    SerdeJsonError(#[from] serde_json::Error),
```
bendk, line 37: I really like using serde-path-to-error for this, so you see something like `record.record_custom_details.category_to_domains.version: value is null`.  Without that JSON path at the start, it's very hard to figure out what's actually going wrong.

Let's do that as a follow-up.  I made https://bugzilla.mozilla.org/show_bug.cgi?id=1894002 for it.

components/relevancy/src/populate_interests.rs
```diff
@@ -2,69 +2,119 @@
...
+        })?;
+        let interest = Interest::try_from(category_code as u32)?;
+        data.push((interest, url_hash));
+    }
+    Ok(())
+}
+
+/// Insert Interests into Db
+fn insert_interest_data(data: Vec<(Interest, UrlHash)>, dao: &mut RelevancyDao) {
+    for (interest, hash_url) in data {
```
bendk, line 82: Could you add the interruption support back here?  That means processing the data in chunks, and calling `dao.err_if_interrupted()?` so that we error out early if the operation is interrupted.

bendk (changes requested):
I had some minor suggestions, but the main code looks great.  Feel free to push back, I don't feel too strongly about any of them.

components/relevancy/src/populate_interests.rs
tiftran: something like this?
```
    match fetch_interest_data() {
        Ok(data) => {
            db.read_write(move |dao| {
                insert_interest_data(data, dao);
                Ok(())
            })?;
        }
        Err(e) => {
            log::warn!("error fetching interest data: {e}");
            return Err(Error::FetchInterestDataError);
        }
    }
    Ok(())
```

bendk: Looks basically right to me, maybe you could simplify the `read_write` lambda though.  If `insert_interest_data` checks for interruption, it will probably return `Result<()>`, and you could just return that value directly.

components/relevancy/src/rs.rs
tiftran: Yeah i agree this is clunky with the extra level of nesting.
This is what we have i our schema doc
```
// Example
{
  "id": "shopping-0001",
  "record_type": "category_to_domains",
  "record_custom_details": {
    "category_to_domains": {
      "category": "shopping",
      "category_code": 1,
      "version": 1
    }
  }
}

```

ncloudioj: To provide more context, we're reusing what we've done for Merino, see the ADR [here](https://github.com/mozilla-services/merino-py/blob/main/docs/adr/0002-merino-general-response.md). The version w/o nesting was also discussed in the ADR ([here](https://github.com/mozilla-services/merino-py/blob/main/docs/adr/0002-merino-general-response.md#b5-custom-details-field-without-the-provider-nesting)), feel free to take a read to see if that makes sense to you :)

components/relevancy/src/populate_interests.rs
```diff
@@ -2,69 +2,121 @@
...
+
+        let parsed_attachment_data =
+            serde_json::from_slice::<Vec<RelevancyAttachmentData>>(&attachment_data)?;
+        parse_record(record_fields, parsed_attachment_data, &mut data)?;
     }
     Ok(data)
 }
 
+/// Parse record and extract interest info
+fn parse_record(
```
ncloudioj, line 59: Looks like the record & attachment parsing was already done in `fetch_interest_data()`, what it really does here is processing them. What do you think about moving all that together? Doing so would make `fetch_interest_data()` slightly easier to follow as we don't need to pass that mutable `data` around any more.

```rs
/// Fetch the interest data
fn fetch_interest_data() -> Result<Vec<(Interest, UrlHash)>> {
    let rs = Client::new(RemoteSettingsConfig {
        collection_name: REMOTE_SETTINGS_COLLECTION.to_string(),
        server: Some(RemoteSettingsServer::Prod),
        server_url: None,
        bucket_name: None,
    })?;
    let remote_settings_response = rs.get_records()?;

    remote_settings_response
        .records
        .iter()
        .map(|record| parse_record(record, &rs))
        // This would turn a `Vec<Result<T, E>>` into a `Result<Vec<T>, E>`.
        // See more details https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect   
        .collect::<Result<Vec<_>>>()
        .map(|v| v.into_iter().flatten().collect())
}

/// Parse an RS record and its attachment to extract interest info
fn parse_record(
    record: &RemoteSettingsRecord,
    rs_client: &Client,
) -> Result<Vec<(Interest, UrlHash)>> {
    let record_fields: RelevancyRecord =
        serde_json::from_value(serde_json::Value::Object(record.fields.clone()))?;

    let attachment_data = match &record.attachment {
        None => return Err(Error::FetchInterestDataError),
        Some(a) => rs_client.get_attachment(&a.location)?,
    };

    let parsed_attachment_data =
        serde_json::from_slice::<Vec<RelevancyAttachmentData>>(&attachment_data)?;

    let custom_details = record_fields.record_custom_details;
    let category_code = custom_details.category_to_domains.category_code;

    let mut data = Vec::new();

    for attachment_data in parsed_attachment_data {
        let hash_url = STANDARD
            .decode(attachment_data.domain)
            .map_err(|_| Error::Base64DecodeError("Invalid base64 error".to_string()))?;
        let url_hash = hash_url.try_into().map_err(|_| {
            Error::Base64DecodeError("Base64 string has wrong number of bytes".to_string())
        })?;
        let interest = Interest::try_from(category_code as u32)?;
        data.push((interest, url_hash));
    }
    Ok(data)
}
```

components/relevancy/src/populate_interests.rs
```diff
@@ -161,4 +222,70 @@ mod test {
...
+        let record_fields: RelevancyRecord = serde_json::from_str(records_string).unwrap();
+        let parsed_attachment_data: Vec<RelevancyAttachmentData> =
+            serde_json::from_str(attachment_string).unwrap();
+        let mut data: Vec<(Interest, UrlHash)> = vec![];
+        parse_record(record_fields, parsed_attachment_data, &mut data)
+            .expect_err("Invalid base64 error");
+    }
+
+    #[test]
+
```
ncloudioj, line 257:
```suggestion
```

components/relevancy/src/populate_interests.rs
```diff
@@ -161,4 +222,70 @@ mod test {
         })
         .unwrap();
     }
+
+    #[test]
+
```
ncloudioj, line 227:
```suggestion
```

ncloudioj (changes requested):

components/relevancy/src/populate_interests.rs
tiftran: oooh we may or may not have done it this way to avoid mocking a remote settings client, for testing 🙃

components/relevancy/src/populate_interests.rs
```diff
@@ -2,69 +2,121 @@
...
+use crate::db::RelevancyDao;
+use crate::rs::{RelevancyAttachmentData, RelevancyRecord, REMOTE_SETTINGS_COLLECTION};
+use crate::url_hash::UrlHash;
+use crate::{Error, Interest, RelevancyDb, Result};
+use base64::{engine::general_purpose::STANDARD, Engine};
+use remote_settings::{Client, RemoteSettingsConfig, RemoteSettingsServer};
 
 // Number of rows to write when inserting interest data before checking for interruption
-const WRITE_CHUNK_SIZE: usize = 100;
+//const WRITE_CHUNK_SIZE: usize = 100;
```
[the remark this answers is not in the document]
bendk: I think it's still needed, see my comment in `insert_interest_data`.

components/relevancy/src/populate_interests.rs
bendk: In this version `rs_client` is only used to for `get_attachment`.  I think we could refactor so that `fetch_interest_data()` called `get_attachment` and `parse_record` inputted the `Vec<u8>` result.  In fact, I think that was essentially our 2nd pass at this one.

bendk (changes requested):
The changes look good to me.  There last detail from my review is adding interruption support to  `insert_interest_data`.

components/relevancy/src/populate_interests.rs
ncloudioj: Yup, I figure that was the reason. Though that still leaves the bulk of `fetch_interest_data()` untested, which is not ideal as future RS schema related errors would only be captured in production.

Generics is often used for this in Rust and [this](https://github.com/mozilla/application-services/blob/main/components/suggest/src/rs.rs#L67) is how we handle it in `suggest`. Feel free to file a follow-up ticket for this testing enhancement.

components/relevancy/src/populate_interests.rs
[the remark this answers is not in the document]
ncloudioj: Sorry, I thought we'd ingest record by record and do interruption checks accordingly. Looks like we will load all the attachments and do the insert afterwards. I think that's fine for now. Let's be mindful as the attachment payload may increase over time and loading everything into memory might get unwieldy.

components/relevancy/src/populate_interests.rs
ncloudioj: >  I think we could refactor so that fetch_interest_data() called get_attachment and parse_record inputted the Vec<u8> result.

Makes sense.

components/relevancy/src/populate_interests.rs
tiftran: i realized i totally read this wrong yesterday

components/relevancy/src/ingest.rs
```diff
@@ -0,0 +1,404 @@
...
+    let remote_settings_response = rs.get_records()?;
+    let mut result = vec![];
+
+    for record in remote_settings_response.records {
+        let attachment_data = match &record.attachment {
+            None => return Err(Error::FetchInterestDataError),
+            Some(a) => rs.get_attachment(&a.location)?,
+        };
+        for hash_url in get_hash_urls(attachment_data)? {
+            result.append(&mut parse_record(&record, hash_url)?);
```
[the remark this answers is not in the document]
ncloudioj: Also, looks like `parse_record()` would only return one `(Interest, UrlHash)`. Maybe we can call `parse_record()` once outside of this loop, something like this:

```rs
fn parse_record(record: &RemoteSettingsRecord) -> Result<Interest> {
...
}

fn fetch_interest_data_inner(...) -> ... {
    for record in remote_settings_response.records {
        ...
        let interest = parse_record(&record)?;
        for hash_url in get_hash_urls(attachment_data)? {
            result.append((interest, hash_url));
        }
    }
}
```

components/relevancy/src/lib.rs
```diff
@@ -55,8 +56,8 @@ impl RelevancyStore {
     ///  This method may execute for a long time and should only be called from a worker thread.
     #[handle_error(Error)]
     pub fn ingest(&self, _top_urls_by_frecency: Vec<String>) -> ApiResult<()> {
-        populate_interests::ensure_interest_data_populated(&self.db)?;
-        todo!()
+        ingest::ensure_interest_data_populated(&self.db)?;
+        Ok(())
```
ncloudioj, line 60:
```suggestion
        ingest::ensure_interest_data_populated(&self.db)
```

bendk (approved):
This all looks great to me.

PR #6236, opened 2024-05-10
Updated SuggestRemoteSettingsClient API

bendk:
Made the API more high-level and not just a copy of the remote settings client API.  It now inputs a `SuggestRemoteSettingsRecordRequest` which is specific to the kinds of requests we make.  It outputs a list of records linked to the attachment data.  This way:

  - It's simpler to mock up in the tests since the mocks can match the actual requests more directly.  I'm hoping to use this to simplify the store tests which have gotten really out of hand.
  - We abstract away some of the remote settings client details, which makes it easer to switch how we make requests in the future.

  - Note:

### DIFF
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -33,10 +33,10 @@
 
 use std::{borrow::Cow, fmt};
 
-use remote_settings::{GetItemsOptions, RemoteSettingsResponse};
+use remote_settings::{Attachment, GetItemsOptions, RemoteSettingsRecord, RsJsonObject, SortOrder};
 use serde::{Deserialize, Deserializer};
 
-use crate::{provider::SuggestionProvider, Result};
+use crate::{error::Error, provider::SuggestionProvider, Result};
 
 /// The Suggest Remote Settings collection name.
 pub(crate) const REMOTE_SETTINGS_COLLECTION: &str = "quicksuggest";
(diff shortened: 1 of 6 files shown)


components/suggest/src/store.rs
```diff
@@ -26,8 +23,8 @@ use crate::{
     provider::SuggestionProvider,
     rs::{
         SuggestAttachment, SuggestRecord, SuggestRecordId, SuggestRecordType,
-        SuggestRemoteSettingsClient, DEFAULT_RECORDS_TYPES, REMOTE_SETTINGS_COLLECTION,
-        SUGGESTIONS_PER_ATTACHMENT,
+        SuggestRemoteSettingsClient, SuggestRemoteSettingsRecord,
```
ncloudioj, line 26: Nit: those structs are getting harder to read, I feel like they can be simplified by dropping the prefixes or using some shorthanded replacements. All of them are defined in the `rs` mod of the `suggest` component, maybe both `Suggest` and `RemoteSettings` can be omitted from the struct IDs?

How about the following?
```rs
use rs::{Client, Record, RecordRequest};
// We can use aliases to disambiguate if needed

// or just follow what RS JS client's convention: use the prefix `Settings`
use rs::{SettingsClient, SettingsRecord, SettingsRecordRequest};
```
I don't have a strong preference, just wanted to hear your thoughts about this.

ncloudioj (approved):
LGTM.

components/suggest/src/rs.rs
```diff
@@ -65,27 +65,96 @@ pub(crate) const DEFAULT_RECORDS_TYPES: [SuggestRecordType; 9] = [
...
+            last_modified: record.last_modified,
+            attachment: record.attachment,
+            attachment_data,
+        }
+    }
+
+    pub fn get_attachment_data(&self) -> Result<&[u8]> {
+        self.attachment_data
+            .as_deref()
+            .ok_or_else(|| Error::MissingAttachment(self.id.clone()))
```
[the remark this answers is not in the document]
bendk: Good point.  I'm going to rename this `require_attachment_data` since it should only be used in cases where the it's an error for the attachment to be missing.  Maybe we could add a `get_attachment_data` later than returns an Option.

components/suggest/src/store.rs
bendk: I love the idea of shortening the names.  I'll open that as a follow-up PR.

linabutler:
I love the direction you took this, @bendk! 💯

PR #6240, opened 2024-05-13
[Tabs] Update the pending close schema so it's also a queue of things…

mhammond:
… to send.

* Abstracts the API towards being more general-purpose "tab command" rather than
  being purely about closing individual tabs, with the intention that it's
  easier to add, say, "close all inactive" (because with 100% certainty I
  know this will be a very early request)

* Updates the schema creation code to be less foot-gun-y.

* Renaming a few things so that we consistently use `device_id` for the fxa id
  and `client_id` for the Sync client ID.

* Fixes the case when syncing when the device ID != fxa id.

This is what I recall @linabutler and I chatted about although I probably got a few things wrong. There are other cleanups begging to be done, but one step at a time.

### DIFF
--- a/components/tabs/src/schema.rs
+++ b/components/tabs/src/schema.rs
@@ -7,16 +7,13 @@
 // syncChangeCounter/syncStatus nor a mirror etc.
 
 use rusqlite::{Connection, Transaction};
-use sql_support::{
-    open_database::{
-        ConnectionInitializer as MigrationLogic, Error as MigrationError, Result as MigrationResult,
-    },
-    ConnExt,
+use sql_support::open_database::{
+    ConnectionInitializer as MigrationLogic, Error as MigrationError, Result as MigrationResult,
 };
 
 // The record is the TabsRecord struct in json and this module doesn't need to deserialize, so we just
 // store each client as its own row.
-const CREATE_SCHEMA_SQL: &str = "
+const CREATE_TABS_TABLE_SQL: &str = "
     CREATE TABLE IF NOT EXISTS tabs (
         guid            TEXT NOT NULL PRIMARY KEY,
         record          TEXT NOT NULL,
@@ -32,12 +29,16 @@ const CREATE_META_TABLE_SQL: &str = "
 ";
 
 const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
-    CREATE TABLE IF NOT EXISTS pending_remote_tab_closures (
-        id                      INTEGER PRIMARY KEY AUTOINCREMENT,
-        client_id               TEXT NOT NULL,
-        url                     TEXT NOT NULL,
-        time_requested_close    INTEGER NOT NULL
+    CREATE TABLE IF NOT EXISTS remote_tab_commands (
+        id                      INTEGER PRIMARY KEY,
+        device_id               TEXT NOT NULL,
+        command                 INTEGER NOT NULL, -- a CommandKind value
+        url                     TEXT,
+        time_requested          INTEGER NOT NULL, -- local timestamp when this was initially written.
+        time_sent               INTEGER -- local timestamp, non-null == no longer pending.
     );
+
+    CREATE UNIQUE INDEX IF NOT EXISTS remote_tab_commands_index ON remote_tab_commands(device_id, command, url);
 ";
 
 pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
@@ -48,11 +49,18 @@ pub(crate) static COLLECTION_SYNCID_META_KEY: &str = "tabs_sync_id";
 // so we store it so we can translate from the tabs sync record ID to the FxA device id for the client
 pub(crate) static REMOTE_CLIENTS_KEY: &str = "remote_clients";
 
+fn init_schema(db: &Connection) -> rusqlite::Result<()> {
+    db.execute_batch(CREATE_TABS_TABLE_SQL)?;
+    db.execute_batch(CREATE_META_TABLE_SQL)?;
+    db.execute_batch(CREATE_PENDING_REMOTE_DELETE_TABLE_SQL)?;
+    Ok(())
+}
+
 pub struct TabsMigrationLogic;
 
 impl MigrationLogic for TabsMigrationLogic {
     const NAME: &'static str = "tabs storage db";
-    const END_VERSION: u32 = 3;
+    const END_VERSION: u32 = 5;
 
     fn prepare(&self, conn: &Connection, _db_empty: bool) -> MigrationResult<()> {
         let initial_pragmas = "
@@ -71,16 +79,13 @@ impl MigrationLogic for TabsMigrationLogic {
 
     fn init(&self, db: &Transaction<'_>) -> MigrationResult<()> {
         log::debug!("Creating schemas");
-        db.execute_all(&[
-            CREATE_SCHEMA_SQL,
-            CREATE_META_TABLE_SQL,
-            CREATE_PENDING_REMOTE_DELETE_TABLE_SQL,
-        ])?;
+        init_schema(db)?;
         Ok(())
     }
 
     fn upgrade_from(&self, db: &Transaction<'_>, version: u32) -> MigrationResult<()> {
         match version {
+            3 | 4 => upgrade_simple_commands_drop(db),
             2 => upgrade_from_v2(db),
             1 => upgrade_from_v1(db),
             _ => Err(MigrationError::IncompatibleVersion(version)),
@@ -88,13 +93,17 @@ impl MigrationLogic for TabsMigrationLogic {
     }
 }
 
+// while we can get away with this, we should :)
+fn upgrade_simple_commands_drop(db: &Connection) -> MigrationResult<()> {
+    // v3 changed the table schema. v5 changed the name.
+    db.execute_batch("DROP TABLE IF EXISTS pending_remote_tab_closures;")?;
+    db.execute_batch("DROP TABLE IF EXISTS remote_tab_commands;")?;
+    db.execute_batch(CREATE_PENDING_REMOTE_DELETE_TABLE_SQL)?;
+    Ok(())
+}
+
 fn upgrade_from_v2(db: &Connection) -> MigrationResult<()> {
-    // Recreate the world
-    db.execute_all(&[
-        CREATE_SCHEMA_SQL,
-        CREATE_META_TABLE_SQL,
-        CREATE_PENDING_REMOTE_DELETE_TABLE_SQL,
-    ])?;
+    db.execute_batch(CREATE_PENDING_REMOTE_DELETE_TABLE_SQL)?;
     Ok(())
 }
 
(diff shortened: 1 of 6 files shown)


components/tabs/src/schema.rs
```diff
@@ -34,10 +31,15 @@ const CREATE_META_TABLE_SQL: &str = "
...
-        url                     TEXT NOT NULL,
-        time_requested_close    INTEGER NOT NULL
+        device_id               TEXT NOT NULL,
+        command                 TEXT NOT NULL, -- probably 'close'
+        url                     TEXT,
+        time_requested          INTEGER NOT NULL, -- local timestamp when this was initially written.
+        time_sent               INTEGER -- local timestamp, non-null == no longer pending.
     );
+
+    -- Having an index on a URL isn't ideal, should just hash?
```
linabutler, line 41: Hmm, I think it's totally OK to use the URL directly here—a hash (specifically the hashing function that we use for URLs in Places) would prevent us from using a `UNIQUE` index directly; we'd need to make our queries a bit fancier to avoid duplicates.

I think the hash is more about [reducing the on-disk size](https://bugzilla.mozilla.org/show_bug.cgi?id=889561) for tables with lots of rows. With a few hundred or even thousand rows, though, I'm not sure it buys us a ton—counterintuitively, for smaller tables, it actually increases the storage size!

components/tabs/src/schema.rs
```diff
@@ -34,10 +31,15 @@ const CREATE_META_TABLE_SQL: &str = "
 const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
     CREATE TABLE IF NOT EXISTS pending_remote_tab_closures (
         id                      INTEGER PRIMARY KEY AUTOINCREMENT,
-        client_id               TEXT NOT NULL,
-        url                     TEXT NOT NULL,
-        time_requested_close    INTEGER NOT NULL
+        device_id               TEXT NOT NULL,
+        command                 TEXT NOT NULL, -- probably 'close'
```
linabutler, line 35: Or an `INTEGER` enum field, in the vein of `BookmarkType`, `SyncedBookmarkKind`, etc.? 😉

components/tabs/src/schema.rs
```diff
@@ -34,10 +31,15 @@ const CREATE_META_TABLE_SQL: &str = "
 const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
     CREATE TABLE IF NOT EXISTS pending_remote_tab_closures (
         id                      INTEGER PRIMARY KEY AUTOINCREMENT,
```
linabutler, line 33: I know this was already there, but might I suggest avoiding `AUTOINCREMENT`, if it's not too painful of a migration? It [has a cost](https://www.sqlite.org/autoinc.html), and it doesn't seem like we need its guarantees here.

components/tabs/src/schema.rs
```diff
@@ -34,10 +31,15 @@ const CREATE_META_TABLE_SQL: &str = "
 const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
     CREATE TABLE IF NOT EXISTS pending_remote_tab_closures (
```
linabutler, line 32: Shall we make this name more generic, since we're doing a migration—maybe `remote_tabs_commands`?

components/tabs/src/storage.rs
```diff
@@ -159,35 +137,128 @@ impl TabsStorage {
...
     /// will check against this table to filter out any urls
-    pub fn add_remote_tab_closures(
+    pub fn add_remote_tab_command(
         &mut self,
-        tabs_requested_closed: Vec<TabsRequestedClose>,
-    ) -> Result<()> {
+        device_id: &str,
+        command: &RemoteCommand,
+    ) -> Result<bool> {
+        self.add_remote_tab_command_at(device_id, command, Timestamp::now())
```
linabutler, line 145: Lovely!

components/tabs/src/storage.rs
```diff
@@ -509,22 +593,27 @@ impl TabsStorage {
          DELETE FROM pending_remote_tab_closures
          WHERE NOT EXISTS (
              SELECT 1 FROM new_remote_tabs
-             WHERE new_remote_tabs.client_id = pending_remote_tab_closures.client_id
+             WHERE new_remote_tabs.device_id = pending_remote_tab_closures.device_id
              AND new_remote_tabs.url = pending_remote_tab_closures.url
          )";
         conn.execute(delete_sql, [])?;
+        log::info!(
+            "deleted {} pending tab closures because they were not in the new tabs",
```
linabutler, line 601: This has worked super well in testing and the demos!

components/tabs/src/storage.rs
```diff
@@ -509,22 +593,27 @@ impl TabsStorage {
...
+            conn.changes()
+        );
 
         const TWENTY_FOUR_HRS_MS: u64 = 24 * 60 * 60 * 1000; // 24 hours in ms
 
         // Anything that couldn't be removed above and is older than 24 hours
         // is assumed not closeable and we can remove it from the list
```
linabutler, line 608: Note for myself (or if you'd like to do it—either is great!) to [make the command TTL in `fxa-client` match this](https://github.com/mozilla/application-services/blob/33ea53616a4d35989059bde593a68dc424a9b95e/components/fxa-client/src/internal/commands/close_tabs.rs#L17).

components/tabs/src/tabs.udl
```diff
@@ -1,6 +1,10 @@
 [Custom]
 typedef string TabsGuid;
 
+// a local timestamp, a custom type to avoid using the int directly.
+[Custom]
+typedef i64 Timestamp;
```
linabutler, line 6: I thought we had [a built-in `Timestamp` type](https://github.com/mozilla/uniffi-rs/tree/f15901cf54b6d68be8ee643cf52b3296d16c11b2/fixtures/uniffi-fixture-time) in UniFFI—would that work here?

linabutler (approved):
Thank you so much for doing this, @mhammond! I've adopted your API in https://bugzilla.mozilla.org/show_bug.cgi?id=1895171, and it's been working perfectly—both demos from this week were powered by it! 🚀

The biggest design suggestion I had was the shape of `PendingCommand` / `RemoteCommand`, so that the new store can do automatic batching and packing. Other than that, I left a few suggestions, and you're welcome to tackle here if you'd like, or save them for a follow-up!

components/tabs/src/tabs.udl
```diff
@@ -51,9 +56,37 @@ dictionary ClientRemoteTabs {
     sequence<RemoteTabRecord> remote_tabs;
 };
 
-dictionary TabsRequestedClose {
-    string client_id;
-    sequence<string> urls;
+[Enum]
+interface RemoteCommand {
+    CloseTab(string url);
```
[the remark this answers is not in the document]
mhammond: I was thinking a little more about this and wondering: if we are going to ask the component to understand the payload size limitations, that implies it too must understand the exact payload format - but it seems error prone for the component to do a kind-of "pretend" serialize just to ensure the payload fits, then return the data back to the app, so it does the "real" serialize and we hope they agree.

Can/should we come up with a system where the actual payload is returned as a string?

components/tabs/src/schema.rs
```diff
@@ -71,43 +79,40 @@ impl MigrationLogic for TabsMigrationLogic {
...
         match version {
+            3 | 4 => upgrade_simple_commands_drop(db),
             2 => upgrade_from_v2(db),
             1 => upgrade_from_v1(db),
             _ => Err(MigrationError::IncompatibleVersion(version)),
         }
     }
 }
 
+// while we can get away with this, we should :)
```
linabutler, line 96: I'm a fan! 😁

components/tabs/src/tabs.udl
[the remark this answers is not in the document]
linabutler: > it seems error prone for the component to do a kind-of "pretend" serialize just to ensure the payload fits, then return the data back to the app, so it does the "real" serialize and we hope they agree.

I agree, this sure feels error-prone! Returning the actual payload as a string doesn't sit quite right with me, either, though, for the same reason. I think the fundamental tension with both of these is that `tabs` has to know the command size limits—which is really an implementation detail of `fxa-client`.

Thinking about it a bit more, I actually wonder if the right place to do the batching / chunking is in `fxa-client`! So maybe we have the `tabs` component keep a list of all the tabs it's been asked to close for each device—that is, without any chunking; if you close 25 tabs on one device, you'll get a list with 25 URLs—and then the `fxa-client` will take care of splitting that into multiple commands.

components/tabs/src/tabs.udl
```diff
@@ -51,9 +56,37 @@ dictionary ClientRemoteTabs {
...
+    // CloseInactive(); ??
+};
+
+interface RemoteCommandStore {
+    // Add a new command, after which it will be pending. Returns false if the command is already active.
```
linabutler, line 66: I think you mentioned we can use `///` comments in UDLs now, too?

components/tabs/src/tabs.udl
```diff
@@ -51,9 +56,37 @@ dictionary ClientRemoteTabs {
...
+
+    [Throws=TabsApiError]
+    boolean remove_remote_command([ByRef] string device_id, [ByRef] RemoteCommand command);
+
+    // This is for the code sending the commands, result is sorted by time_requested.
+    [Throws=TabsApiError]
+    sequence<PendingCommand> get_unsent_commands();
+
+    [Throws=TabsApiError]
+    boolean set_remote_command_sent([ByRef] string device_id, [ByRef] RemoteCommand command);
```
linabutler, line 82: Design nit / question: I wonder if it makes sense to have `set_remote_command_sent` take a `PendingCommand`, instead of `(device_id, RemoteCommand)`—to make the expected flow of these commands a little clearer.

linabutler (approved):
Still looks good! ☺️

components/tabs/src/tabs.udl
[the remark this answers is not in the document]
mhammond: > Thinking about it a bit more, I actually wonder if the right place to do the batching / chunking is in `fxa-client`!

I think it probably is, and we can probably just move the "ensure tabs fit" helper function into a shared crate.

components/tabs/src/storage.rs
mhammond: I ended up making this match fxa - I believe 24 hours is maybe too short - I'm mildly concerned 48 is too - eg, sending a close command to my work computer on a Friday night will ttl out before I restart the PC Monday - but it's fine for now and I've left a breadcrumb pointing to the other in both places.

components/tabs/src/tabs.udl
mhammond: Not really - we don't use the underlying types backing that. TBH, I think UniFFI should deprecate those types now that we have custom types.

components/tabs/src/tabs.udl
mhammond: Yeah, that does make sense - but doing that means I hated the old name, so we now have `set_pending_command_sent`

mhammond:
> Thank you so much for doing this, @mhammond! I've adopted your API in https://bugzilla.mozilla.org/show_bug.cgi?id=1895171, and it's been working perfectly

Just to call out that it will now be broken, particularly due to the `set_pending_command_sent` rename/arg-change.

mhammond:
I'm ready to land this, but just another heads-up @skhamis and @linabutler about the `set_pending_command_sent` rename.

PR #6246, opened 2024-05-16
Bug 1897264 - interruption support for ingestion

bendk:
Added an `interrupt_everything` method that interrupts both the read and write connection.  This is intended for iOS to use to interrupt the suggest component during shutdown.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1897299 for a discussion on next steps here.

Created a new `WriteScope` type that stores an interrupt scope that can be used for multiple `write` calls.  This allows the ingestion code to catch all interruptions that happened after the ingestion started. With the old system, if `interrupt_everything` was called in-between ingestion two types then we might miss it.

  - Note:


components/suggest/src/store.rs
```diff
@@ -601,6 +639,8 @@ struct SuggestStoreDbs {
     writer: SuggestDb,
     /// A read-only connection used to query the database.
     reader: SuggestDb,
+    /// A read-write connection used specifically for ingestion
+    ingestion: SuggestDb,
```
bendk, line 643: I like the idea of adding a 3rd connection, but maybe we should put out a quick fix for the iOS issue and tackle this separately.  WDY2T?

linabutler: Oh, let's not have two read-write connections...having multiple writers is a bag of hurt in SQLite. We learned that the hard way with Places, where we saw contention between them—and, in Desktop, we could wedge one writer permanently, so that it always returned `SQLITE_BUSY` 😱

linabutler (changes requested):
The shape of the API looks great (thanks for doing this!), but having a second writer is inviting dragons 🐉

Would you mind dropping that in favor of having `interrupt_ingestion()` just interrupt the existing `writer`?

components/suggest/src/store.rs
bendk: Seems good for now, I updated the PR to do that.

Long-term, I still kind of like separating the connections since it allows you to cancel each separately.   The current `interrupt_ingestion` method could also interrupt a call to `dismiss_suggestion`.  I think there's a way to do it safely, although maybe it's just because I haven't been burned by SQLite yet.

components/suggest/src/db.rs
```diff
@@ -651,10 +659,11 @@ impl<'a> SuggestDao<'a> {
         &mut self,
         record_id: &SuggestRecordId,
         suggestions: &[DownloadedAmoSuggestion],
+        ingest_interrupt_scope: &SqlInterruptScope,
```
linabutler, line 662: Ahh, I had `SuggestDao` hold an interrupt scope specifically so that we could avoid passing it around 😄

The `Dao` is a transient object that only lives for the lifetime of a `read()` or `write()` call, so I was thinking that, if you wanted a separate interrupt scope for just ingestion, you could create a `SuggestDao` with a separate interrupt scope, and that `SuggestDao` would only live for the lifetime of the ingestion.

WDYT? This isn't blocking by any means; I just wanted to make sure I understood if there were other reasons for passing the scope around explicitly.

linabutler (approved):
Just a comment about passing the interrupt scope around explicitly now, but no need to block on that. Thanks again, @bendk!

components/suggest/src/store.rs
linabutler: Thanks, @bendk, and I'm truly sorry if my reaction came off as short in any way—I was running a little late for an appointment earlier, and you deserved a more thoughtful response with more time. I left some more noodlings on https://bugzilla.mozilla.org/show_bug.cgi?id=1897299 for your consideration 🧠

components/suggest/src/db.rs
bendk: The transaction gets committed at end of `write` and I didn't want to change that behavior.  But it does seem better to just pass around the `SuggestDao`.  I'll update the PR to create a single DAO for the entire ingestion and add a `SuggestDao::commit` method that we can call.  This seems better all-around, it seems better to see the explicit commit in the code.

bendk:
Updated the change so that instead of creating a new scope at the top of `ingest`, we create the `SuggestDao` and re-use it for the entire ingestion.  This feels a lot nicer to me, although there's a minor difference because before we would commit the transaction after each individual `write` call.  I don't think this is a big deal for now, we're not getting many reports of failures to ingest.

I'm thinking we can merge this, which will unblock Lina to fix things on iOS, then create a follow-up PR that adds back some extra commits.  I also think we could further simplify some code now that we only have one `SuggestDao`.

bendk:
Pushing one more change that adds back the transaction support here.  I did this because a) I realized it wasn't that hard to add it back and b) the code is not quite ready for the other cleanups I wanted to do.  Also, I renamed the method back to `interrupt_everything` since that's the most accurate.  Maybe we can add an `interrupt_ingestion` later on after we discuss it more.

Sorry for moving the target with this one, but I think it's ready for review again now.

components/suggest/src/db.rs
```diff
@@ -107,11 +107,45 @@ impl SuggestDb {
...
     }
+
+    /// Create a new write scope.
+    ///
+    /// This enables performing multiple `write()` calls with the same shared interrupt scope.
+    /// This is important for things like ingestion, where you want the operation to be interrupted
+    /// if [Self::interrupt_handle::interrupt] is called after the operation starts.  Calling
+    /// [Self::write] multiple times during the operation risks missing a call that happens after
+    /// between those calls.
+    pub fn write_scope(&self) -> Result<WriteScope> {
```
linabutler, line 123: Oooo, I really like this!

linabutler (approved):
Still LGTM, thanks!

ncloudioj (approved):
LGTM, could you also update the [README.md](https://github.com/mozilla/application-services/blob/main/components/suggest/README.md) to reflect the change? It only covers how to use `interrupt()` for `query()`.

ncloudioj:
> Also, I renamed the method back to `interrupt_everything` since that's the most accurate. Maybe we can add an `interrupt_ingestion` later on after we discuss it more.

I was gonna ask if it makes sense to consolidate those interrupt calls as consumers almost certainly need to come here to check the document to tell which one should be used. Would `interrupt(scope: InterruptScope)` sounds better? Where InterruptScope can be read, write, and read-write.

bendk:
> I was gonna ask if it makes sense to consolidate those interrupt calls as consumers almost certainly need to come here to check the document to tell which one should be used. Would `interrupt(scope: InterruptScope)` sounds better? Where InterruptScope can be read, write, and read-write.

I like this idea, but I feel like "scope" is already somewhat overloaded so I called it `InterruptKind`.  For backwards compatibility, I made this param optional.  Hopefully we can update consumers to pass it in and drop the optional part.

components/suggest/src/suggest.udl
```diff
@@ -127,6 +127,12 @@ interface SuggestProviderConfig {
     );
 };
 
+enum InterruptKind {
+    "Read",
+    "Write",
+    "Everything",
```
linabutler, line 133: Absolute nitpick: What do you think of `ReadWrite` instead of `Everything` as a name? 😅

This new API is so elegant!

linabutler:
@Mergifyio backport release-v127

PR #6248, opened 2024-05-16
Bug 1897319 - Stores icons for light/dark theme for Yelp suggestion

dadaa:
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1897319
Jira: https://mozilla-hub.atlassian.net/browse/SNG-1515

We need to update the UI for Yelp suggestion.
One of them is to chnage the logo. But it has to change by the light/dark theme. Thus, make it to store both icons.

This PR did two things.
1. Stores both icons and change the related codes.
2. Introduce icon struct as same as PR https://github.com/mozilla/application-services/pull/6162 (A warning that said paramters are too complex in the first commit, it was a good chance :))

  - Note:


components/suggest/src/schema.rs
0c0w3, line 1: You'll need to bump the schema `VERSION` const and add a migration step to `upgrade_from()`. See the comment above `VERSION`.

components/suggest/src/store.rs
0c0w3, line 1: Thanks for updating all these tests, I know it's pain.

It looks like we now have tests for Yelp suggestions with both light and dark icons, but no tests for only one or the other. Also we may want to keep including `icon` in the RS data in order to support older Firefox versions for some time, and I want to make sure that doesn't break newer versions.

So could you add cases for the following please, in whatever way is easiest? In all cases, ingest and fetch should complete successfully.

* Only `iconLightTheme` is present in the RS data
* Only `iconDarkTheme` is present
* `icon`, `iconLightTheme`, and `iconDarkTheme` are all present (`icon` should be ignored and ingest and fetch should still be successful)

components/suggest/src/schema.rs
```diff
@@ -113,9 +113,11 @@ CREATE TABLE yelp_location_signs(
 ) WITHOUT ROWID;
 
 CREATE TABLE yelp_custom_details(
-    icon_id TEXT PRIMARY KEY,
+    icon_light_theme_id TEXT NOT NULL,
+    icon_dark_theme_id TEXT NOT NULL,
```
0c0w3, line 117: Storing both light and dark icons in the same row is OK with me, and also I just wanted to mention another way to do it would be to have one icon per row plus an "icon type" column. If we need to support any more types of icons in the future -- and I don't know why we would! -- I think we should switch to that approach.

0c0w3 (approved):
Thanks Daisuke, r+ with comments -- especially, the schema version needs to be bumped and a migration step added.

components/suggest/src/schema.rs
dadaa: Yeah, if we do not know how many icons we should handle, then yeah, should introduce the mechanism.

components/suggest/src/schema.rs
```diff
@@ -193,6 +195,11 @@ CREATE TABLE IF NOT EXISTS dismissed_suggestions (
                 )?;
                 Ok(())
             }
+            19 => {
+                // Need to update new icons for Yelp especially in new schema.
+                let _ = clear_database(tx);
```
0c0w3, line 200: Since only the Yelp table changed, can we do something like this instead?

1. If `yelp_custom_details` exists, get the current record from `yelp_custom_details` and then `DROP TABLE yelp_custom_details`
2. `CREATE TABLE yelp_custom_details` just like in the main schema SQL
3. If we got the record from the old table in step 1, insert it into the new table, using the old `icon_id` as the new `icon_light_theme_id`

Starting with schema version 16, I believe the idea is to avoid clearing the entire DB on upgrade if possible.

Sorry to add more work but It would also be nice if you could add a test for this migration step too.

0c0w3 (commented):
Thanks, it looks like there are merge conflicts in lib.rs and store.rs, plus some failing tests. Could you fix those and make sure `cargo test -p suggest` passes locally for you?

components/suggest/src/schema.rs
dadaa: Okay, thanks!

dadaa:
Hi @0c0w3! As I changed more places than I expected, I'd like to request review from you again. Thanks!

0c0w3:
Hey @dadaa, Emilio's idea of choosing the appropriate image in the svg itself is really nice, and it means we don't need to modify the schema at all. I'm sorry for not thinking of that, and for all the work you've done in this PR that we might not need.

If you remove the schema change from this PR, it's basically the same as #6162, right?

dadaa:
> > Hey @dadaa, Emilio's idea of choosing the appropriate image in the svg itself is really nice, and it means we don't need to modify the schema at all.
> 
> Yeah, indeed. I'm also sorry to you that I took your time due to I could not come up with it.

But, I was a little concerned about whether it would be okay to change the official logo file, even though it would look the same.

0c0w3:
I think it's OK to create a new file that combines both official files as long as we don't modify any strokes or colors. My example file in Phabricator embeded only the `<path>`s from each file, but I think it's possible to embed the whole `<svg>` from each file inside a `<g>` in the combined file. That way, we're really not modifying anything at all, only combining.

dadaa:
We took another approach, let me close this PR.
https://phabricator.services.mozilla.com/D210745

PR #6254, opened 2024-05-27
Prefer non-optional UTF8 String <-> Data conversion.

mhammond:
Hopefully addresses the swift-lint errors.

Fix via https://github.com/realm/SwiftLint/issues/5263

### DIFF
--- a/components/nimbus/ios/Nimbus/Dictionary+.swift
+++ b/components/nimbus/ios/Nimbus/Dictionary+.swift
@@ -7,10 +7,7 @@ import Foundation
 extension Dictionary where Key == String, Value == Any {
     func stringify() throws -> String {
         let data = try JSONSerialization.data(withJSONObject: self)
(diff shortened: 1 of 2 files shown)


linabutler (approved):
Thanks!

linabutler:
Actually, it looks like SwiftLint is reverting this rule: https://github.com/realm/SwiftLint/pull/5601

While I think it might actually make sense for us in both cases—`JSONSerialization` and `JSONEncoder` should always return valid UTF-8 encoded byte sequences—I wouldn’t want to merge this, and then discover it’s warning again the other way 😅

Shall we give this a bit more time?

linabutler:
Superseded by #6269. It looks like the SwiftLint reversion is taking a bit of time to merge, so let's disable the rule for now. That way, we shouldn't need to do anything once the rule is changed back.

PR #6264, opened 2024-06-04
Persist relevancy data

bendk:

  - Note:

### DIFF
--- a/components/relevancy/src/schema.rs
+++ b/components/relevancy/src/schema.rs
@@ -13,7 +13,7 @@ use sql_support::open_database::{self, ConnectionInitializer};
 ///  1. Bump this version.
 ///  2. Add a migration from the old version to the new version in
 ///     [`RelevancyConnectionInitializer::upgrade_from`].
-pub const VERSION: u32 = 13;
+pub const VERSION: u32 = 14;
 
 /// The current database schema.
 pub const SQL: &str = "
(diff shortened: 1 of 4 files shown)


components/relevancy/src/schema.rs
```diff
@@ -47,7 +55,52 @@ impl ConnectionInitializer for RelevancyConnectionInitializer {
         Ok(db.execute_batch(SQL)?)
     }
 
-    fn upgrade_from(&self, _db: &Transaction<'_>, version: u32) -> open_database::Result<()> {
-        Err(open_database::Error::IncompatibleVersion(version))
+    fn upgrade_from(&self, tx: &Transaction<'_>, version: u32) -> open_database::Result<()> {
+        match version {
+            // Upgrades 1-12 are missing because we started with version 13, because of a
+            // copy-and-paste error.
```
bendk, line 61: Oops!  I think this was my fault when I was copying the code from `suggest`.  It's not that big of a deal though, we can just start with lucky number 13.

components/relevancy/src/schema.rs
```diff
@@ -22,6 +22,14 @@ pub const SQL: &str = "
         interest_code INTEGER NOT NULL,
         PRIMARY KEY (url_hash, interest_code)
     ) WITHOUT ROWID;
+
+    -- Stores user interest vectors.  The `kind` field stores the raw code from the `InterestVectorKind` enum.
+    CREATE TABLE user_interest(
+        kind INTEGER NOT NULL,
```
ncloudioj, line 28: Unrelated to this PR, just for discussion.

For debugging, what do you think about adding some dimension tables for `Kind` and `Interests`? We currently store those in their numeric forms in `url_interest` and `user_interest` which is fine as we can programmatically convert them into the more readable string representations. But when we query the db directly, we will have to manually do the conversion. With dimension tables, we can easily get the string forms via table joins.

ncloudioj: Sounds good, we can leave it for now.

A few cases I found it useful:

When I updated [this](https://mozilla-hub.atlassian.net/wiki/spaces/CLOUDSERVICES/pages/695173233/Suggest+Interaction+data+and+Implementation) to reflect the in-flight suggestion / keyword counts, I usually just pull data directly from `suggest.sqlite`. Since we only store provider IDs in the `suggestion` table, I have to open up the source code to manually map provider IDs to the actual strings. Likewise, when I query `relevancy.sqlite` to get the `interest -> url-counts` mapping, only interest codes are available and I have to look up the interest enum in the source to decode them.

PR #6267, opened 2024-06-06
Bug 1900837 - Speed up re-ingestion 

bendk:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -946,6 +946,18 @@ impl<'a> SuggestDao<'a> {
     /// the database.
     pub fn drop_suggestions(&mut self, record_id: &SuggestRecordId) -> Result<()> {
         self.conn.execute_cached(
+            "DELETE FROM keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
+            "DELETE FROM full_keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
+            "DELETE FROM prefix_keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
             "DELETE FROM suggestions WHERE record_id = :record_id",
             named_params! { ":record_id": record_id.as_str() },
         )?;
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -1000,6 +1000,18 @@ impl<'a> SuggestDao<'a> {
         Ok(())
     }
 
+    #[cfg(feature = "benchmark_api")]
+    /// Clears the value for a metadata key.
+    ///
+    /// This is currently only used for the benchmarks.
+    pub fn clear_meta(&mut self, key: &str) -> Result<()> {
+        self.conn.execute_cached(
+            "DELETE FROM meta WHERE key = :key",
+            named_params! { ":key": key },
+        )?;
+        Ok(())
+    }
+
     /// Updates the last ingest timestamp if the given last modified time is
     /// newer than the existing one recorded.
     pub fn put_last_ingest_if_newer(
--- a/components/suggest/src/schema.rs
+++ b/components/suggest/src/schema.rs
@@ -15,7 +15,7 @@ use sql_support::open_database::{self, ConnectionInitializer};
 ///     [`SuggestConnectionInitializer::upgrade_from`].
 ///    a. If suggestions should be re-ingested after the migration, call `clear_database()` inside
 ///       the migration.
-pub const VERSION: u32 = 19;
+pub const VERSION: u32 = 20;
 
 /// The current Suggest database schema.
 pub const SQL: &str = "
@@ -26,17 +26,16 @@ CREATE TABLE meta(
 
 CREATE TABLE keywords(
     keyword TEXT NOT NULL,
-    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id) ON DELETE CASCADE,
-    full_keyword_id INTEGER NULL REFERENCES full_keywords(id) ON DELETE SET NULL,
+    suggestion_id INTEGER NOT NULL,
+    full_keyword_id INTEGER NULL,
     rank INTEGER NOT NULL,
     PRIMARY KEY (keyword, suggestion_id)
 ) WITHOUT ROWID;
 
 -- full keywords are what we display to the user when a (partial) keyword matches
--- The FK to suggestion_id makes it so full keywords get deleted when the parent suggestion is deleted.
 CREATE TABLE full_keywords(
     id INTEGER PRIMARY KEY,
-    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id) ON DELETE CASCADE,
+    suggestion_id INTEGER NOT NULL,
     full_keyword TEXT NOT NULL
 );
 
@@ -45,7 +44,7 @@ CREATE TABLE prefix_keywords(
     keyword_suffix TEXT NOT NULL DEFAULT '',
     confidence INTEGER NOT NULL DEFAULT 0,
     rank INTEGER NOT NULL,
-    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id) ON DELETE CASCADE,
+    suggestion_id INTEGER NOT NULL,
     PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id)
 ) WITHOUT ROWID;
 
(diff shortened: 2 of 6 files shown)


components/suggest/src/schema.rs
```diff
@@ -193,6 +196,46 @@ CREATE TABLE IF NOT EXISTS dismissed_suggestions (
...
+                clear_database(tx)?;
+                tx.execute_batch(
+                    "
+-- Recreate the various keywords table to drop the `ON DELETE` clauses
+DROP TABLE keywords;
+DROP TABLE full_keywords;
+DROP TABLE prefix_keywords;
+CREATE TABLE keywords(
+    keyword TEXT NOT NULL,
+    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id),
```
ncloudioj, line 211: Alternatively, we could also add an `record_id` here with an index on it so that we can delete all the keywords by a record_id. That should save a sub-query in the delete statement above. Though the upside with your approach is that it'll be easier to support delta updates as it only needs to delete old keywords for updated suggestions instead of all the suggestions for a RS record.

components/suggest/src/schema.rs
```diff
@@ -193,6 +196,46 @@ CREATE TABLE IF NOT EXISTS dismissed_suggestions (
...
+    keyword_suffix TEXT NOT NULL DEFAULT '',
+    confidence INTEGER NOT NULL DEFAULT 0,
+    rank INTEGER NOT NULL,
+    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id),
+    PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id)
+) WITHOUT ROWID;
+CREATE UNIQUE INDEX keywords_suggestion_id_rank ON keywords(suggestion_id, rank);
+-- Create the new indexes
+CREATE INDEX prefix_keywords_suggestion_id ON prefix_keywords(suggestion_id);
+CREATE INDEX keywords_full_keyword_id ON keywords(full_keyword_id);
```
ncloudioj, line 232: What's this used for?

ncloudioj (approved):
LGTM, thanks!

components/suggest/src/schema.rs
```diff
@@ -45,11 +45,14 @@ CREATE TABLE prefix_keywords(
     keyword_suffix TEXT NOT NULL DEFAULT '',
     confidence INTEGER NOT NULL DEFAULT 0,
     rank INTEGER NOT NULL,
-    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id) ON DELETE CASCADE,
+    suggestion_id INTEGER NOT NULL REFERENCES suggestions(id),
     PRIMARY KEY (keyword_prefix, keyword_suffix, suggestion_id)
 ) WITHOUT ROWID;
 
 CREATE UNIQUE INDEX keywords_suggestion_id_rank ON keywords(suggestion_id, rank);
+CREATE INDEX prefix_keywords_suggestion_id ON prefix_keywords(suggestion_id);
```
linabutler, line 53: Thanks for adding these indexes here! I think we didn't add them earlier because the number of MDN, AMO, and Pocket suggestions was so small compared to AMP and Wikipedia, but it's wonderful to have them for symmetry!

I'm idly curious how well this would benchmark, and if it would have any impact one way or the other on the database size:

```suggestion
CREATE UNIQUE INDEX prefix_keywords_suggestion_id ON prefix_keywords(suggestion_id, confidence, rank);
```

(Pocket keywords have two separate lists, one with low-confidence and one with high-confidence prefixed keywords, and the `rank` is the index of the keyword within that list. For AMO and MDN, the `confidence` is always 0).

components/suggest/src/schema.rs
```diff
@@ -36,7 +36,7 @@ CREATE TABLE keywords(
 -- The FK to suggestion_id makes it so full keywords get deleted when the parent suggestion is deleted.
```
linabutler, line 36: Nit: Would you mind dropping this comment, since we aren't using `ON DELETE CASCADE` anymore?

components/suggest/src/db.rs
```diff
@@ -945,6 +945,18 @@ impl<'a> SuggestDao<'a> {
     /// Deletes all suggestions associated with a Remote Settings record from
     /// the database.
     pub fn drop_suggestions(&mut self, record_id: &SuggestRecordId) -> Result<()> {
+        self.conn.execute_cached(
```
linabutler, line 948: This seems totally reasonable to me, and SQLite could probably optimize this query plan much better than relying on `ON DELETE CASCADE` (which is implemented using row-level triggers under the hood).

linabutler (approved):
Looks great, thank you very much!

components/suggest/src/schema.rs
linabutler: Oooh, interesting! I'd be curious to see if this has any adverse impact on the file size.

Intuitively, I'm worried that duplicating (and indexing, which would duplicate it again) the `record_id` in the `keywords` table (for 180k rows), but maybe SQLite can actually store it efficiently.

ncloudioj: Good point. I didn't think of that overhead which indeed would bloat the db. It would only make sense for a numeric typed column.

components/suggest/src/db.rs
bendk: > SQLite could probably optimize these query plans much better than relying on ON DELETE CASCADE
Agreed and the benchmarks seem to verify this.

components/suggest/src/schema.rs
linabutler: I should say explicitly: this is totally non-blocking, and there's no need to do this now if you'd like to land the fix ASAP.

bendk:
Lina has been asking some great questions around DB size.  I added a measurement to the `cargo suggest-debug-ingestion-sizes` command and it looks like the indexes will increase the size from ~19M to ~26M.

components/suggest/src/schema.rs
bendk: I'm going to go with your index since it seems more correct and has a negligible affect on performance.  Disk size was exactly the same -- I guess the difference fit inside the remainder of the current page.  On my system the pocket benchmark finishes in 12ms and AMO is in the nano-second range, both before and after the change.

components/suggest/src/schema.rs
bendk: Okay, let's skip this for now since it doesn't seem like a clear win.

components/suggest/src/schema.rs
bendk: It speeds up checking the FK when rows are deleted in the `full_keywords` table.  AFAICT, it has a large impact on performance.  When I delete the index the benchmark times start to skyrocket again.

components/suggest/src/schema.rs
ncloudioj: Sounds good.

components/suggest/src/schema.rs
ncloudioj: Hmm, interesting.

```
DELETE FROM keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)
DELETE FROM full_keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)
```

The queries don't even use `full_keyword_id` anywhere, not sure how it'd affect the performance. Anyway, not a blocker, just curious.

bendk: According to the [SQLite docs](https://www.sqlite.org/foreignkeys.html), when a row is deleted when there is a foreign key, it will perform the equivalent of `SELECT rowid FROM <child-table> WHERE <child-key> = :parent_key_value` to check that the constraint is still valid.  I think that's `SELECT rowid FROM keywords WHERE full_keyword_id = :full_keyword_id` for the case of `full_keywords`

ncloudioj: Ahhh, I see. To enforce the FK invariants.

Now that we don't rely on foreign key cascade delete any more, what if we drop those foreign key constraints from the schema (e.g. drop the FKs on `suggestion_id` and `full_keyword_id` for `keywords`) so it doesn't need to check the FK invariant on deletes. Would that still be sufficiently safe and more performant?

bendk: I think that's a very reasonable idea.  We lose some safety, but it seems like it shouldn't matter too much in practice -- especially since ingestion in totally deterministic, so if we setup the relationships correctly for one user then it will work for all users.  Performance increases and we save a significant amount of disk space.

I guess my one concern would be the precedent going forward.  I could see us making the same call in a similar situation and it not being correct.  What's your feeling on this?

ncloudioj: No need to consider that change for this PR, we can continue discussing that offline and make changes later if needed.

Yeah, it's not uncommon to forfeit FK for performance gains in practice. Especially, if the table is mostly write-once-read-many and not storing safety-critical datasets, the safety could be enforced by application itself at the best effort (I think it applies in this case). In the case of things go wrong, we can always drop the whole table and create everything from scratch w/o any user impact. In return, we don't have to pay the FK taxes and it will run faster with lower footprint.

Anyway, we can think of this more offline.

linabutler:
> it looks like the indexes will increase the size from ~19M to ~26M.

Long tables really are something else, aren't they? 😅 Thank you _so_ much for validating this, and extending the benchmark suite to cover it! Considering the (two? three?) orders of magnitude improvement in the ingestion time, an increase of 7 MB seems fair.

bendk:
Soft-freeze is over and I squashed #6273 into this one.  If the CI is green I'll hit the merge button.

bendk:
@mergify backport release-v128

bendk:
Backporting this to the 128 branch so that we can uplift it to Android (maybe desktop as well https://bugzilla.mozilla.org/show_bug.cgi?id=1895110)

PR #6276, opened 2024-06-13
Bug 1901805 - Starting Fakespot ingestion code

bendk:
Defined the `fakespot` feature and put most of the changes behind that feature flag.

Updated ingest code to use 2 remote settings clients, one for fakespot and one for everything else.  I think I like this setup going forward, since it makes it easy for new suggestions to have their own colleciton if we decide it makes sense.

Implemented some very simple matching logic, the next step is for the SNG team to update this with the real matching logic.  I tried to leave `FAKESPOT-TODO` comment breadcrumbs where they need to update the code.

You can run `cargo suggest-debug-ingestion-sizes --features=fakespot` to do a simple real-world test of this code.   It should show 10,000 items in the fakespot_custom_details table, which corresponds to the 10,000 items we uploaded to remote settings.

  - Note:

### DIFF
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -38,15 +38,13 @@ use serde::{Deserialize, Deserializer};
 
 use crate::{error::Error, provider::SuggestionProvider, Result};
 
-/// The Suggest Remote Settings collection name.
-pub(crate) const REMOTE_SETTINGS_COLLECTION: &str = "quicksuggest";
-
 /// The maximum number of suggestions in a Suggest record's attachment.
 ///
 /// This should be the same as the `BUCKET_SIZE` constant in the
 /// `mozilla-services/quicksuggest-rs` repo.
 pub(crate) const SUGGESTIONS_PER_ATTACHMENT: u64 = 200;
 
+#[cfg(not(feature = "fakespot"))]
 /// A list of default record types to download if nothing is specified.
 /// This currently defaults to all of the record types.
 pub(crate) const DEFAULT_RECORDS_TYPES: [SuggestRecordType; 9] = [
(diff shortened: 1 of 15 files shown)


components/suggest/src/rs.rs
```diff
@@ -69,17 +83,58 @@ pub(crate) trait Client {
     fn get_records(&self, request: RecordRequest) -> Result<Vec<Record>>;
 }
 
-impl Client for remote_settings::Client {
+/// Implements the [Client] trait using a real remote settings client
+pub struct RemoteSettingsClient {
```
ncloudioj, line 87: Nit: maybe this should be plural as it has two clients now?

components/suggest/src/rs.rs
```diff
@@ -69,17 +83,58 @@ pub(crate) trait Client {
...
+                collection_name: "fakespot-suggest-products".to_owned(),
+                server_url,
+            })?,
+        })
+    }
+}
+
+impl Client for RemoteSettingsClient {
     fn get_records(&self, request: RecordRequest) -> Result<Vec<Record>> {
+        let client = match request.record_type.as_str() {
```
ncloudioj, line 122: Or we can make this as a helper for RemoteSettingsClient, e.g. `RemoteSettingsClient.get_client(...)`?

components/suggest/src/rs.rs
bendk: I kind of like keeping it singular, since from the store's perspective it's a single client that's used to fetch data from remote settings.  Maybe this is just laziness though, since if I changed the name of the type, I'd also want to change the name of the field, maybe the name of the trait, etc.  I'm not that tied to this opinion though, happy to discuss it more in another issue/PR.

ncloudioj: It's really a nitpick, singular is also fine.

PR #6288, opened 2024-06-27
Desktop 128.1 release

bendk:
I want to uplift the Suggest performance fix to Desktop, but it's complicated by the lack of a nightly process.

The 128 desktop beta is using app-services https://github.com/mozilla/application-services/commit/3ca067683c7546e34bc8cdbe820fbe8e4089939a.  This is about a dozen [commits](https://github.com/mozilla/application-services/commits/release-v128/?since=2024-05-30&until=2024-06-07) before our 128 release.  This means that if we just vendor in 128.1 to Desktop, we're going to be taking a bunch of commits that haven't been tested at all there, which seems pretty scary for an uplift.

Instead of that, I'm thinking we can make a branch that starts on https://github.com/mozilla/application-services/commit/3ca067683c7546e34bc8cdbe820fbe8e4089939a and has https://github.com/mozilla/application-services/commit/b57868178993306c3fb8eacc5447facab572a4f6 (the performance fix) cherry-picked in.  We can then vendor that commit on Desktop.  I called this branch `release-v128-desktop`, this PR contains the cherry-pick commit.

So my questions with this PR are:
  - Does this plan seem good?
  - Does the cherry-pick seem correct?

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -1027,6 +1027,18 @@ impl<'a> SuggestDao<'a> {
     /// the database.
     pub fn drop_suggestions(&mut self, record_id: &SuggestRecordId) -> Result<()> {
         self.conn.execute_cached(
+            "DELETE FROM keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
+            named_params! { ":record_id": record_id.as_str() },
+        )?;
+        self.conn.execute_cached(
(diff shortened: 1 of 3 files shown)


bendk:
The swiftlint error is expected, we've been dealing with that for awhile.  I think it was just fixed this week in main.

I believe the android-build-pr is also expected, but this is only going to be consumed on desktop anyways.

lougeniaC64 (approved):
A bit of a rubber stamp, but approving to unblock the uplift. 🚢

PR #6289, opened 2024-06-27
Teach the tabs engine about the ability to close inactive tabs.

mhammond:
This also includes #6281 - you might prefer to review both together or review that by itself, then review this (ie, just the tabs engine changes) once that is merged.

  - Note:

### DIFF
--- a/components/tabs/src/schema.rs
+++ b/components/tabs/src/schema.rs
@@ -38,7 +38,7 @@ const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
         time_sent               INTEGER -- local timestamp, non-null == no longer pending.
     );
 
-    CREATE UNIQUE INDEX IF NOT EXISTS remote_tab_commands_index ON remote_tab_commands(device_id, command, url);
+    CREATE UNIQUE INDEX IF NOT EXISTS remote_tab_commands_index ON remote_tab_commands(device_id, command, IFNULL(url, ''));
 ";
 
 pub(crate) static LAST_SYNC_META_KEY: &str = "last_sync_time";
@@ -60,7 +60,7 @@ pub struct TabsMigrationLogic;
 
 impl MigrationLogic for TabsMigrationLogic {
     const NAME: &'static str = "tabs storage db";
-    const END_VERSION: u32 = 5;
+    const END_VERSION: u32 = 6;
 
     fn prepare(&self, conn: &Connection, _db_empty: bool) -> MigrationResult<()> {
         let initial_pragmas = "
@@ -85,12 +85,20 @@ impl MigrationLogic for TabsMigrationLogic {
 
     fn upgrade_from(&self, db: &Transaction<'_>, version: u32) -> MigrationResult<()> {
         match version {
+            5 => upgrade_remotetabs_index(db),
             3 | 4 => upgrade_simple_commands_drop(db),
             2 => upgrade_from_v2(db),
             1 => upgrade_from_v1(db),
             _ => Err(MigrationError::IncompatibleVersion(version)),
         }
     }
+}
+
+fn upgrade_remotetabs_index(db: &Connection) -> MigrationResult<()> {
+    // We changed the index to take nulls into account.
+    db.execute_batch("DROP INDEX IF EXISTS remote_tab_commands_index;")?;
+    db.execute_batch(CREATE_PENDING_REMOTE_DELETE_TABLE_SQL)?;
+    Ok(())
 }
 
 // while we can get away with this, we should :)
@@ -199,5 +207,19 @@ mod tests {
             [],
         )
         .expect_err("identical command should fail");
+        conn.execute(
+            "INSERT INTO remote_tab_commands
+                (device_id, command, url, time_requested, time_sent)
+                VALUES ('d', 'other', null, 1, null)",
+            [],
+        )
+        .unwrap();
+        conn.execute(
+            "INSERT INTO remote_tab_commands
+                (device_id, command, url, time_requested, time_sent)
+                VALUES ('d', 'other', null, 1, null)",
+            [],
+        )
+        .expect_err("identical command with null URL should fail");
     }
 }
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -270,8 +270,10 @@ impl TabsStorage {
             Ok(Some(conn)) => conn,
         };
         let pending_tabs_result: Result<Vec<(String, String)>> = conn.query_rows_and_then_cached(
-            "SELECT device_id, url FROM remote_tab_commands",
-            [],
+            "SELECT device_id, url
+             FROM remote_tab_commands
+             WHERE command = :command_close_tab",
+            rusqlite::named_params! { ":command_close_tab": CommandKind::CloseTab },
             |row| {
                 Ok((
                     row.get::<_, String>(0)?, // device_id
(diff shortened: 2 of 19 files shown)


components/tabs/src/schema.rs
```diff
@@ -38,7 +38,7 @@ const CREATE_PENDING_REMOTE_DELETE_TABLE_SQL: &str = "
         time_sent               INTEGER -- local timestamp, non-null == no longer pending.
     );
 
-    CREATE UNIQUE INDEX IF NOT EXISTS remote_tab_commands_index ON remote_tab_commands(device_id, command, url);
+    CREATE UNIQUE INDEX IF NOT EXISTS remote_tab_commands_index ON remote_tab_commands(device_id, command, IFNULL(url, ''));
```
linabutler, line 41: Hmmm, would you be open to modeling this a little differently? It's a more invasive change, but it got me thinking about a more general way to handle commands that aren't tied to URLs, especially since we haven't shipped closing tabs yet:

```sql
CREATE TABLE IF NOT EXISTS remote_tab_commands (
    id                      INTEGER PRIMARY KEY,
    device_id               TEXT NOT NULL,
    command                 INTEGER NOT NULL, -- a CommandKind value
    time_requested          INTEGER NOT NULL, -- local timestamp when this was initially written.
    time_sent               INTEGER -- local timestamp, non-null == no longer pending.
);

CREATE TABLE IF NOT EXISTS remote_tab_command_urls (
    command_id INTEGER NOT NULL REFERENCES remote_tab_commands(id) ON DELETE CASCADE,
    url TEXT NOT NULL,
    PRIMARY KEY (command_id, url)
) WITHOUT ROWID;
```

The `remote_tab_command_urls` table takes the place of our `remote_tab_commands_index` index, and lets us send multiple URL-less commands to the same device.

components/tabs/src/storage.rs
```diff
@@ -270,8 +270,10 @@ impl TabsStorage {
             Ok(Some(conn)) => conn,
         };
         let pending_tabs_result: Result<Vec<(String, String)>> = conn.query_rows_and_then_cached(
```
linabutler, line 272: Nit: Let's unify this and the inactive query you added into the same `SELECT`; it would avoid scanning the table twice, since `command = :command` can't use an index 😅

components/tabs/src/storage.rs
```diff
@@ -293,18 +295,40 @@ impl TabsStorage {
                 return crts;
             }
         };
+        // building this in to a hashset seems a little over the top?
```
linabutler, line 298: I think so, too! I'd almost suggest a sorted `Vec` (with `ORDER BY device_id`, or sorting in Rust) that you can `binary_search`, but even that feels a bit excessive 😅

components/tabs/src/storage.rs
```diff
@@ -293,18 +295,40 @@ impl TabsStorage {
...
+                    "Failed to read database for remote inactive closes: {}",
+                    e
+                );
+                Default::default()
+            }
+        };
+
         // Check if any of the client records that were passed in have urls that the user closed
         // This means that they requested to close those tabs but those devices have not yet got
         // actually closed the tabs
```
linabutler, line 320: I do wonder if there's a way we could generalize this kind of "commands subsuming other commands" in the future—but no need to do that in this PR!

components/tabs/src/storage.rs
```diff
@@ -558,7 +598,8 @@ impl TabsStorage {
     }
 
     // Remove any pending tabs that are 24hrs older than the last time that client has synced
-    // Or that client's incoming tabs does not have those tabs anymore
+    // Or that client's incoming tabs reflects that the command was acted on (eg, for tab close
```
linabutler, line 601: Zepto-nit: Sentence fragment 👓 and missing triple-slash comment. Let's do:

```rs 
/// Removes any pending "close tabs" commands that are either:
///
/// 1. 24hrs older than the last time the target device has synced, or...
/// 2. The target device's incoming tabs reflect that the command was acted on
///    (eg, for tab close commands; does not have those tabs anymore;
///    for close inactive commands, has no inactive tabs).
```

(I know this was already there before; sorry to ask you to fix it!)

linabutler (changes requested):
This looks just wonderful to me, @mhammond; thank you so much, and sorry for the review delay!

Requesting changes just to hear your thoughts on the schema change I suggested, and a few other small things; but no hard blockers!

components/tabs/src/schema.rs
mhammond: > this design would let us send multiple URL-less commands to the same device.

I don't quite get this part - I don't think we want to queue multiple "CloseAllInactiveTabs" commands any more than we want to queue multiple CloseTab commands for the same URL. I don't see how this world would let us enforce that without significant changes that might seem overkill at face value?

mhammond:
I think I got most of the review comments, but still a little unclear on how to proceed re the schema

components/tabs/src/schema.rs
skhamis: I definitely like the idea of potentially having just a queue of any kind of "remote tab management" into the queue. However I wonder if we should move that to a new issue while we can discuss the implications of that! I agree that there's nothing yet that would indicate we'd have more than "CloseAllInactive" (move tabs maybe?? heh). Don't have a strong opinion either way but thought I'd give me 2c!

skhamis (approved):
Sorry I let this sit in queue for awhile -- Looks like Lina got most of the major concerns. Looks good to me! Thanks for doing this! Definitely gonna be useful from a technical and UX perspective.

components/tabs/src/schema.rs
mhammond: Lina and I chatted about this and agreed to not do this at the current time and instead to wait an see if a future command presents another opportunity to model this differently.

components/tabs/src/storage.rs
```diff
@@ -268,44 +268,61 @@ impl TabsStorage {
...
-                        .get(&crt.client_id)
-                        // The top level in the url_history is the "active" tab, which we should use
-                        // TODO: probably not the best way to url check
-                        .map_or(false, |urls| urls.contains(&tab.url_history[0]))
+                    (!hide_inactive || !tab.inactive)
+                        && !pending_closures
+                            .get(&crt.client_id)
+                            // The top level in the url_history is the "active" tab, which we should use
+                            // TODO: probably not the best way to url check
+                            .map_or(false, |urls| urls.contains(&tab.url_history[0]))
```
linabutler, line 325: Woof, this entire expression took me a couple of tries to parse! IIUC, we're trying to exclude tabs that are either:

* Inactive, _only if_ there's a "close all inactive tabs" command pending for that device.
* Closed from this device, using the "close tabs" command.

Is that right?

In a follow-up—no need to do that here 😅—I wonder if we can do some refactoring, to make this a little easier to follow. Conceptually, it makes perfect sense, but I think the parallel data structures and conditionals obscure the flow a little.

components/tabs/src/storage.rs
```diff
@@ -578,50 +615,76 @@ impl TabsStorage {
...
 
         // Insert new remote tabs into a temporary table
-        conn.execute(
-            "CREATE TEMP TABLE if not exists new_remote_tabs (device_id TEXT, url TEXT)",
-            [],
+        conn.execute_batch(
+            "
+            CREATE TEMP TABLE if not exists new_remote_tabs (device_id TEXT, url TEXT);
+            DELETE FROM new_remote_tabs;
+            CREATE TEMP TABLE if not exists new_remote_tabs_without_inactive (device_id TEXT);
```
linabutler, line 622:
```suggestion
            CREATE TEMP TABLE IF NOT EXISTS new_remote_tabs (device_id TEXT, url TEXT);
            DELETE FROM new_remote_tabs;
            CREATE TEMP TABLE IF NOT EXISTS new_remote_tabs_without_inactive (device_id TEXT);
```

Nit: Let's capitalize keywords (I know the `if not exists` was lowercased before).

Follow-up / food for thought: You could turn this into a single table with an `ifempty(url, '')` and a `command` type—basically, mirroring how `remote_tab_commands` is set up! That might make it a little easier to do the `DELETE FROM remote_tab_commands` below?—but that's more of my own style preference; functionally, the way you have this now is also good!

linabutler (approved):
This looks lovely; thank you for putting it together and taking it over the finish line, @mhammond! I had a few nits and follow-up suggestions, but nothing blocking—ship it!

PR #6292, opened 2024-07-03
Bug 1905748 - Fakespot ingestion and matching

bendk:
Create a FTS table for fakespot and populate it during ingestion. Updated the `SuggestQuery` code to generate a string to use with the FTS MATCH operator. I couldn't get DELETE triggers working for this, so I made it so we delete the FTS data in `drop_suggestions`.

Also:
- Make the `clear_database` function delete the FTS data
- Added some more tests for the fakespot data
- Updated the `assert_schema_matches_new_database()` function to also check the triggers.  Followed the rule of 3 and refactored things now we're checking tables, indexes, and triggers.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -686,7 +686,6 @@ impl<'a> SuggestDao<'a> {
     #[cfg(feature = "fakespot")]
     /// Fetches fakespot suggestions
     pub fn fetch_fakespot_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
-        // FAKESPOT-TODO: The SNG will update this based on the results of their FTS experimentation
         self.conn.query_rows_and_then_cached(
             r#"
             SELECT
@@ -700,14 +699,17 @@ impl<'a> SuggestDao<'a> {
             FROM
                 suggestions s
             JOIN
+                fakespot_fts fts
+                ON fts.rowid = s.id
+            JOIN
                 fakespot_custom_details f
                 ON f.suggestion_id = s.id
             WHERE
-                s.title LIKE '%' || ? || '%'
+                fakespot_fts MATCH ?
             ORDER BY
                 s.score DESC
             "#,
-            (&query.keyword,),
+            (&query.fts_query(),),
             |row| {
                 Ok(Suggestion::Fakespot {
                     title: row.get(0)?,
@@ -935,7 +937,6 @@ impl<'a> SuggestDao<'a> {
         record_id: &SuggestRecordId,
         suggestions: &[DownloadedFakespotSuggestion],
     ) -> Result<()> {
-        // FAKESPOT-TODO: The SNG will update this based on the results of their FTS experimentation
         let mut suggestion_insert = SuggestionInsertStatement::new(self.conn)?;
         let mut fakespot_insert = FakespotInsertStatement::new(self.conn)?;
         for suggestion in suggestions {
@@ -1036,6 +1037,17 @@ impl<'a> SuggestDao<'a> {
             "DELETE FROM prefix_keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
             named_params! { ":record_id": record_id.as_str() },
         )?;
+        #[cfg(feature = "fakespot")]
+        {
+            self.scope.err_if_interrupted()?;
+            self.conn.execute_cached(
+                "
+                DELETE FROM fakespot_fts
+                WHERE rowid IN (SELECT id from suggestions WHERE record_id = :record_id)
+                ",
+                named_params! { ":record_id": record_id.as_str() },
+            )?;
+        }
         self.scope.err_if_interrupted()?;
         self.conn.execute_cached(
             "DELETE FROM suggestions WHERE record_id = :record_id",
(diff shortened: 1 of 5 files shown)


components/suggest/src/query.rs
```diff
@@ -118,4 +120,122 @@ impl SuggestionQuery {
...
+                fts_query.push_str(last_keyword);
+                fts_query.push('"');
+                fts_query.push('*');
+            } else {
+                fts_query.push('"');
+                fts_query.push_str(last_keyword);
+                fts_query.push('"');
+            }
+        }
+        fts_query
```
ncloudioj, line 185: Can be simplified a bit. (I personally prefer raw string literals to avoid quote escaping)
```
        // Quote and append each term from `query`
        let keywords = self.parse_keywords();

        if keywords.is_empty() {
            return String::from(r#""""#);
        }

        let mut fts_query = keywords
            .iter()
            .map(|keyword| format!(r#""{keyword}""#))
            .collect::<Vec<_>>()
            .join(" ");


        // If the input is > 3 characters, and there's no whitespace at the end.
        // We want to append a `*` char to the end to do a prefix match on it.
        if !self.keyword.ends_with(' ') && keywords.iter().fold(0, |count, s| count + s.len()) > 3 {
            fts_query.push('*');
        }

        fts_query
```

components/suggest/src/db.rs
```diff
@@ -1022,6 +1023,22 @@ impl<'a> SuggestDao<'a> {
             "DELETE FROM prefix_keywords WHERE suggestion_id IN (SELECT id from suggestions WHERE record_id = :record_id)",
             named_params! { ":record_id": record_id.as_str() },
         )?;
+        #[cfg(feature = "fakespot")]
+        {
+            self.scope.err_if_interrupted()?;
+            // Delete FTS data for this record, which counter-intuitively requires an INSERT
```
ncloudioj, line 1029: o.O That's really odd. Regardless, we might want to benchmark the FTS deletion bits to better understand its performance characteristics.

ncloudioj (approved):
Looks great, just a few nits.

components/suggest/src/schema.rs
```diff
@@ -170,6 +169,22 @@ CREATE TABLE fakespot_custom_details(
     total_reviews INTEGER NOT NULL,
     FOREIGN KEY(suggestion_id) REFERENCES suggestions(id) ON DELETE CASCADE
 );
+CREATE VIRTUAL TABLE IF NOT EXISTS fakespot_fts USING FTS5(
+  title,
+  prefix='4 5 6 7 8 9 10 11',
+  content='',
```
ncloudioj, line 175: Per SQLite's [doc](https://www.sqlite.org/fts5.html#contentless_delete_tables), contentless-delete tables should be preferred over contentless table. Shall we enable that?

components/suggest/src/schema.rs
```diff
@@ -170,6 +169,22 @@ CREATE TABLE fakespot_custom_details(
...
+  tokenize=\"porter unicode61 remove_diacritics 2 tokenchars '''-'\"
```
ncloudioj, line 176: `'''-'\"` reads a bit cryptic, maybe we can add a comment clarifying that both apostrophe and hyphen are treated as valid tokenchars by the tokenizer.

components/suggest/src/query.rs
```diff
@@ -118,4 +120,122 @@ impl SuggestionQuery {
...
+        // that we don't need re-allocations
+        let mut fts_query = String::with_capacity(self.keyword.len() * 2);
+        // Quote and append each term from `query`
+        let keywords = self.parse_keywords();
+
+        if keywords.is_empty() {
+            return String::from("\"\"");
+        }
+        for keyword in keywords[..keywords.len() - 1].iter() {
+            fts_query.push('"');
```
ncloudioj, line 163: Per [doc](https://www.sqlite.org/fts5.html#fts5_strings), an FTS string can be enclosed by double quotes as a whole, e.g. `match '"foo bar baz"'`, no need to wrap each token by double quotes, e.g `match '"foo" "bar" "baz"'`, correct?

ncloudioj (commented):
Added a few more comments after taking another look.

components/suggest/src/schema.rs
bendk: Yes please, that makes the code much nicer.

components/suggest/src/query.rs
bendk: Those docs are very confusing to me, but I think that makes the entire string a phrase, which needs to be matched together, without any tokens in the middle.  If I change the code to quote things differently, then [this test](https://github.com/mozilla/application-services/blob/a3c093f7afc72df86116097a649f0195bdf300e7/components/suggest/src/store.rs#L2114-L2117) fails.

bendk: I'm still not totally sure what's going on, but the code appears to work when I quote each term separately so I'm going to leave it like that.   I'm happy to continue to iterate on this one, I definitely think there's room for improvement here.

components/suggest/src/schema.rs
quiiver: i think i missed the utility of the contentless-delete tables. looks much better

PR #6294, opened 2024-07-03
Update AndroidX Test to the latest releases

rvandermeulen:
Changelogs: https://developer.android.com/jetpack/androidx/releases/test


skhamis (approved):
LGTM!

PR #6295, opened 2024-07-03
Remote settings caching support

bendk:
I'm creating a PR get feedback from anyone interested, but I'm mainly pushing this branch so that Nish can test it out on iOS.

Added a `get_cached_records` method that gets all record for a collection, using a cache make it efficient.  I'm hoping that this will help out Nish who is experimenting with this in iOS and also that we can use something like this for remote settings.

  - Note:

### DIFF
--- a/components/remote_settings/src/cache.rs
+++ b/components/remote_settings/src/cache.rs
@@ -1,0 +1,152 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::RemoteSettingsResponse;
+use std::collections::HashSet;
+
+/// Merge a cached RemoteSettingsResponse and a newly downloaded one to get a merged response
+///
+/// cached is a previously downloaded remote settings response (possibly run through merge_cache_and_response).
+/// new is a newly downloaded remote settings response (with `_expected` set to the last_modified
+/// time of the cached response).
+///
+/// This will merge the records from both responses, handle deletions/tombstones, and return a
+/// response that has:
+///   - The newest `last_modified_date`
+///   - A record list containing the newest version of all live records.  Deleted records will not
+///     be present in this list.
+///
+/// If everything is working properly, the returned value will exactly match what the server would
+/// have returned if there was no `_expected` param.
+pub fn merge_cache_and_response(
+    cached: RemoteSettingsResponse,
+    new: RemoteSettingsResponse,
+) -> RemoteSettingsResponse {
+    let new_record_ids = new
+        .records
+        .iter()
+        .map(|r| r.id.as_str())
+        .collect::<HashSet<&str>>();
+    // Start with any cached records that don't appear in new.
+    let mut records = cached
+        .records
+        .into_iter()
+        .filter(|r| !new_record_ids.contains(r.id.as_str()))
+        // deleted should always be false, check it just in case
+        .filter(|r| !r.deleted)
+        .collect::<Vec<_>>();
+    // Add all (non-deleted) records from new
+    records.extend(new.records.into_iter().filter(|r| !r.deleted));
+
+    RemoteSettingsResponse {
+        last_modified: new.last_modified,
+        records,
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{RemoteSettingsRecord, RsJsonObject};
+
+    // Quick way to generate the fields data for our mock records
+    fn fields(data: &str) -> RsJsonObject {
+        let mut map = serde_json::Map::new();
+        map.insert("data".into(), data.into());
+        map
+    }
+
+    #[test]
+    fn test_combine_cache_and_response() {
+        let cached_response = RemoteSettingsResponse {
+            last_modified: 1000,
+            records: vec![
+                RemoteSettingsRecord {
+                    id: "a".into(),
+                    last_modified: 100,
+                    deleted: false,
+                    attachment: None,
+                    fields: fields("a"),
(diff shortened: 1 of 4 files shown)


bendk:
The use case here is a consumer that just wants to get some settings from RS.

Writing that out, I realize that I should add another method that only checks the cache, never hits the network, and returns `Option<RemoteSettingsResponse>`.  Consumers will probably want to use this version to load settings on startup.

components/remote_settings/src/cache.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Interface for caching remote settings data
+///
+/// This is implemented by the remote settings consumer, sometimes with foreign code via UniFFI.
+///
+/// This interface has no error handling.  Consumers should report any errors themselves.  If there
+/// is an error in `get()` simply return `None` and the remote settings client will re-fetch any
+/// data.
+pub trait RemoteSettingsCache: Send + Sync {
+    /// Update the cached value
+    fn store(&self, value: String);
```
nbhasin2, line 17: Note: Does iOS need to provide a path for the cache or is this something that is already handled by remote settings?

bendk: My thought was that iOS would implement this interface and pass that to the remote settings client.  Behind the scenes you could use whatever backing store made sense for the data.

nbhasin2: Yup, thats what my plan is. I got confused by the naming but what I will do is have a quick chat/call with you when this gets merged. Can show you the caching and if we are missing an API or need to make changes we can do that.

bendk:
@linabutler what do you think about merging this one?  I've been using the API from Rust-land to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1908802 and it's been working pretty well there, so I think it makes sense to push out for consumers to test out.

bendk:
> [...] I realize that I should add another method that only checks the cache, never hits the network, and returns `Option<RemoteSettingsResponse>`. Consumers will probably want to use this version to load settings on startup.

This is now how the function works by default.  Pass in `options.sync = true` to make it always hit the network and sync any new records.  There's no option for the old behavior -- to hit the network only if the cache was empty, I can't see a use-case for that.

bendk:
We discussed this in the app-services engineering call today and there were some questions on exactly how we should expose this to the consumer.  I reworked this PR to only contain what I need to make use of this in Suggest and not expose any new functionality yet.  I'll open a new PR soon with an ADR with different options for exposing this to consumers.

linabutler:
I actually really liked the look and general design of your `RemoteSettingsCache` interface! I think it addresses iOS's current needs, and sets us up for a future where we can provide a default SQLite-backed cache implementation, that clients without specific storage needs ("not Suggest" 😅) could use in a transparent, and backward-compatible way.

I know you reverted that addition, but I wonder if you're open to making just a couple of changes to it?

1. Make `RemoteSettingsCache.store()` take a `RemoteSettingsRecord` instead of a string, and make `RemoteSettingsCache.get()` return a `RemoteSettingsRecord?`. That would let the cache implementation control how to serialize, store, and re-inflate records.
2. Add a `cache: RemoteSettingsCache?` field to `RemoteSettingsConfig`—or maybe even introduce a `RemoteSettingsClientBuilder`. That way, the `RemoteSettings` client could call into the cache at any time to store and request cached records.
3. Drop the separate `RemoteSettings.get_cached_records()` method, and have `RemoteSettings` automatically call into the `cache` if one is supplied in `RemoteSettingsConfig`. If one is supplied, the client will transparently use caching for all requests; if it's not, it'll just download and fetch records every time, as it does now.

WDYT? I'd be happy to re-raise this comment on an ADR, of course, if you'd still like to do that! 🚀

bendk:
> I know you reverted that addition, but I wonder if you're open to making just a couple of changes to it?

Could we merge this now and discuss in a separate PR.  Maybe we don't need an ADR, but I think there are some decisions to be made and I don't want https://bugzilla.mozilla.org/show_bug.cgi?id=1908802 to pressure us into a quick decision.

> Make RemoteSettingsCache.store() take a RemoteSettingsRecord instead of a string, and make RemoteSettingsCache.get() return a RemoteSettingsRecord?. That would let the cache implementation control how to serialize, store, and re-inflate records.

This makes a lot of sense, but I also really like the idea of the consumer passing us a path and giving us control.  This is simpler for the consumer and matches the way users construct the various store types.   Maybe we could give users the option when they construct `RemoteSettings`.  This could work very nicely with your builder suggestion.
 
>   Add a cache: RemoteSettingsCache? field to RemoteSettingsConfig—or maybe even introduce a RemoteSettingsClientBuilder. That way, the RemoteSettings client could call into the cache at any time to store and request cached records.
>
> Drop the separate RemoteSettings.get_cached_records() method, and have RemoteSettings automatically call into the cache if one is supplied in RemoteSettingsConfig. If one is supplied, the client will transparently use caching for all requests; if it's not, it'll just download and fetch records every time, as it does now.

I like this but I think we're going to want to give users more control over the network requests.  There are 2 major use cases that I know of:
  - Suggest: we always want to fetch the latest records and just use the cache to optimize the requests
  - Settings: we probably don't want to make a network request just to get the value of a setting -- especially in startup.  Instead, we want to get the setting from the cache, fallback to a default if the cache is empty, and sync the records on some schedule.

To handle those cases, maybe we could do something like this:
  - Update construction and `get_records` like you're suggesting.
  - Keep `get_cached_records` for the settings use-case.
  - Add `sync_records`, also for the settings use-case.
  - If you don't supply a cache when constructing the client, then `get_cached_records` and `sync_records` will either return an error or be no-ops.

linabutler:
> Could we merge this now and discuss in a separate PR?

Of course!

linabutler (approved):
Thanks!

PR #6298, opened 2024-07-09
support: Extract `payload_support::shrink_to_fit`.

linabutler:
The tabs and clients engines use the same logic to truncate a list of
items to fit within a size limit when serialized to JSON, though
they handle failures differently: the tabs engine sends the complete
list of tabs; the clients engine clears the list of commands.

We'll reuse the same core logic to pack URLs for closed synced tabs
into multiple commands, so let's factor it out into a shared
support crate.

  - Note:

### DIFF
--- a/components/support/payload/Cargo.toml
+++ b/components/support/payload/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "payload-support"
+version = "0.1.0"
+authors = ["Sync Team <sync-team@mozilla.com>"]
+edition = "2021"
+license = "MPL-2.0"
+
+[dependencies]
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
--- a/components/support/payload/src/lib.rs
+++ b/components/support/payload/src/lib.rs
@@ -1,0 +1,173 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use serde::Serialize;
+use std::{
+    io::{self, Write},
+    num::NonZeroUsize,
+};
+
+/// A quantity of items that can fit into a payload, accounting for
+/// serialization, encryption, and Base64-encoding overhead.
+pub enum Fit {
+    /// All items can fit into the payload.
+    All,
+
+    /// Some, but not all, items can fit into the payload without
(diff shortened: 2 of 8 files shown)


components/support/payload/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+use std::{
+    io::{self, Write},
+    num::NonZeroUsize,
+};
+
+/// A quantity of items that can fit into a payload, accounting for
+/// serialization, encryption, and Base64-encoding overhead.
+pub enum Fit {
+    /// All items in the list can fit into the payload.
+    All,
```
mhammond, line 15: I'm not really sure about the value of All vs Some tbh - I can't see how real code would actually differ in how they are treated.  None is a bit borderline but can see this being special.

components/support/payload/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+    None,
+
+    /// The serialized size of the items couldn't be determined because of
+    /// an error.
+    Err(serde_json::Error),
+}
+
+impl Fit {
+    #[inline]
+    pub fn truncate_to(&self) -> Option<NonZeroUsize> {
```
mhammond, line 32: `truncate_to` seems like it's actually taking action, but really isn't returning an indicator that the caller should take action. I wonder if `needs_truncation` might be better? I really don't care though!

components/support/payload/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "payload-support"
```
mhammond, line 2: totally unrelated to this patch, but something I've noticed in the past is that the names of our packages aren't great for being in a repo like m-c. I wonder if we should start using a longer `package` name, but add a `lib` section where the name used by the code is specified.

(This sounds fine until I try and think of what a good longer name would be, and I come up with something like `app-services-payload-support`, then realize that's just as bad - if not more - in a moz-central context.

Anyway, leaving this comment in even though I don't think it's worth doing anything about :)

mhammond (approved):
This looks great. Just yesterday I discovered https://bugzilla.mozilla.org/show_bug.cgi?id=1879003, which almost implies this isn't working quite correctly for Fenix - I wonder if you would be able to manually reassure us that this does indeed work the same as desktop?

components/support/payload/src/lib.rs
mhammond: (oops, meant to delete this comment, I think what you have is great!)

PR #6300, opened 2024-07-10
Bug 1907004: Chunk oversized "close tabs" commands into multiple device commands.

linabutler:
This PR:

* Introduces a new `CloseTabsResult` return type.
* Adds a `CloseTabsResult::TabsNotClosed` variant, which includes the URLs of any tabs that couldn't be sent in a "close tabs" command.
* Reworks `FirefoxAccount::close_tabs` to pack URLs that exceed the maximum payload size (16 KB by default) into multiple device commands.

  - Note:

### DIFF
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -91,6 +91,20 @@ enum FxaError {
 };
 
 
+/// An error type that carries additional information for
+/// device command-related [`FirefoxAccount`] operations.
+[Error]
+interface DeviceCommandError {
+  /// An error thrown if a device command to close one or more tabs
+  /// couldn't be sent. The caller should try to invoke
+  /// [`FirefoxAccount::close_tabs`] with the returned `urls` later.
+  TabsNotClosed(sequence<string> urls);
+
+  /// An error thrown if the operation failed for an
+  /// account-related reason.
+  Account(FxaError error);
+};
+
 
 // Object representing the signed-in state of an application.
 //
@@ -536,7 +550,7 @@ interface FirefoxAccount {
   ///
   /// If a device on the account has registered the [`CloseTabs`](DeviceCapability::CloseTabs)
   /// capability, this method can be used to close its tabs.
-  [Throws=FxaError]
+  [Throws=DeviceCommandError]
   void close_tabs([ByRef] string target_device_id, sequence<string> urls);
 
 
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -91,18 +91,30 @@ enum FxaError {
 };
 
 
-/// An error type that carries additional information for
-/// device command-related [`FirefoxAccount`] operations.
-[Error]
-interface DeviceCommandError {
-  /// An error thrown if a device command to close one or more tabs
-  /// couldn't be sent. The caller should try to invoke
-  /// [`FirefoxAccount::close_tabs`] with the returned `urls` later.
-  TabsNotClosed(sequence<string> urls);
+/// The result of invoking a "close tabs" command.
+///
+/// If [`FirefoxAccount::close_tabs`] is called with more URLs than can fit
+/// into a single command payload, the URLs will be chunked and sent in
+/// multiple commands.
+///
+/// Chunking breaks the atomicity of a "close tabs" command, but
+/// reduces the number of these commands that FxA sends to other devices.
+/// This is critical for platforms like iOS, where every command triggers a
+/// push message that must show a user-visible notification.
+[Enum]
+interface CloseTabsResult {
+  /// All URLs passed to [`FirefoxAccount::close_tabs`] were chunked and sent
+  /// in one or more device commands.
+  Ok();
 
-  /// An error thrown if the operation failed for an
-  /// account-related reason.
-  Account(FxaError error);
+  /// One or more URLs passed to [`FirefoxAccount::close_tabs`] couldn't be sent
+  /// in a device command. The caller can assume that:
+  ///
+  /// 1. Any URL in the returned list of `urls` was not sent, and
+  ///    should be retried.
+  /// 2. All other URLs that were passed to [`FirefoxAccount::close_tabs`], and
+  ///    that are _not_ in the list of `urls`, were chunked and sent.
+  TabsNotClosed(sequence<string> urls);
 };
 
 
@@ -550,8 +562,8 @@ interface FirefoxAccount {
   ///
   /// If a device on the account has registered the [`CloseTabs`](DeviceCapability::CloseTabs)
   /// capability, this method can be used to close its tabs.
-  [Throws=DeviceCommandError]
-  void close_tabs([ByRef] string target_device_id, sequence<string> urls);
+  [Throws=FxaError]
+  CloseTabsResult close_tabs([ByRef] string target_device_id, sequence<string> urls);
 
 
   // Get the URL at which to access the user's sync data.
--- a/components/fxa-client/src/internal/close_tabs.rs
+++ b/components/fxa-client/src/internal/close_tabs.rs
@@ -2,36 +2,100 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::mem;
+
+use payload_support::Fit;
+
 use super::{
     commands::{
         close_tabs::{self, CloseTabsPayload},
         decrypt_command, encrypt_command, IncomingDeviceCommand, PrivateCommandKeys,
     },
+    device::COMMAND_MAX_PAYLOAD_SIZE,
     http_client::GetDeviceResponse,
     scopes, telemetry, FirefoxAccount,
 };
 use crate::{Error, Result};
 
 impl FirefoxAccount {
-    pub fn close_tabs<T: AsRef<str>>(&mut self, target_device_id: &str, urls: &[T]) -> Result<()> {
+    pub fn close_tabs<T>(&mut self, target_device_id: &str, urls: Vec<T>) -> Result<()>
+    where
+        T: Into<String>,
+    {
         let devices = self.get_devices(false)?;
         let target = devices
             .iter()
             .find(|d| d.id == target_device_id)
             .ok_or_else(|| Error::UnknownTargetDevice(target_device_id.to_owned()))?;
-        let (payload, sent_telemetry) =
-            CloseTabsPayload::with_urls(urls.iter().map(|url| url.as_ref().to_owned()).collect());
-        let oldsync_key = self.get_scoped_key(scopes::OLD_SYNC)?;
-        let command_payload =
-            encrypt_command(oldsync_key, target, close_tabs::COMMAND_NAME, &payload)?;
-        self.invoke_command(
-            close_tabs::COMMAND_NAME,
-            target,
-            &command_payload,
-            Some(close_tabs::COMMAND_TTL),
-        )?;
-        self.telemetry.record_command_sent(sent_telemetry);
-        Ok(())
+
+        let sent_telemetry = telemetry::SentCommand::for_close_tabs();
+        let mut urls_to_retry = Vec::new();
+
+        // Sort the URLs shortest to longest, so that we can at least make
+        // some forward progress, even if there's an oversized URL at the
+        // end that won't fit into a single command.
+        let mut urls: Vec<_> = urls.into_iter().map(Into::into).collect();
+        urls.sort_unstable_by_key(String::len);
+
+        while !urls.is_empty() {
+            // If we were asked to close more URLs than will fit in a
+            // single command, chunk the URLs into multiple commands,
+            // packing as many as we can into each. Do this until we've either
+            // drained and packed all the URLs, or we see an oversized URL
+            // that won't fit into a single command.
+            let chunk = match payload_support::try_fit_items(&urls, COMMAND_MAX_PAYLOAD_SIZE.get())
+            {
+                Fit::All => mem::take(&mut urls),
+                Fit::Some(count) => urls.drain(..count.get()).collect(),
+                Fit::None | Fit::Err(_) => {
+                    // Oversized URLs that won't fit into a single command, and
+                    // serialization errors, are permanent; retrying to send
+                    // these URLs won't help. But we want our consumers to keep
+                    // any pending closed URLs hidden from the user's synced
+                    // tabs list, until they're eventually sent (for temporary
+                    // errors; see below), or expire after some time
+                    // (for oversized URLs that can't ever be sent).
+                    urls_to_retry.append(&mut urls);
+                    break;
+                }
+            };
+
+            let sent_telemetry = sent_telemetry.clone_with_new_stream_id();
+            let payload = CloseTabsPayload::with_telemetry(&sent_telemetry, chunk);
+
+            let oldsync_key = self.get_scoped_key(scopes::OLD_SYNC)?;
+            let command_payload =
+                encrypt_command(oldsync_key, target, close_tabs::COMMAND_NAME, &payload)?;
+            let result = self.invoke_command(
+                close_tabs::COMMAND_NAME,
+                target,
+                &command_payload,
+                Some(close_tabs::COMMAND_TTL),
+            );
+            match result {
+                Ok(()) => {
+                    self.telemetry.record_command_sent(sent_telemetry);
+                }
+                Err(e) => {
+                    error_support::report_error!(
+                        "fxaclient-close-tabs-invoke",
+                        "Failed to send bulk Close Tabs command: {}",
+                        e
+                    );
+                    // Temporary error; if the consumer retries, we expect that
+                    // we _will_ eventually send these URLs.
+                    urls_to_retry.extend(payload.urls);
+                }
+            }
+        }
+
+        if urls_to_retry.is_empty() {
+            Ok(())
+        } else {
+            Err(Error::TabsNotClosed {
+                urls: urls_to_retry,
+            })
+        }
     }
 
     pub(crate) fn handle_close_tabs_command(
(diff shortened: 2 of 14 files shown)


components/fxa-client/src/internal/close_tabs.rs
```diff
@@ -92,3 +144,403 @@ impl FirefoxAccount {
         self.state.clear_commands_data(close_tabs::COMMAND_NAME);
     }
 }
+
+#[cfg(test)]
+mod tests {
```
linabutler, line 149: The tests are the biggest part of this patch, by quite a margin! 😅

components/fxa-client/src/internal/close_tabs.rs
```diff
@@ -19,19 +24,70 @@ impl FirefoxAccount {
...
+            // If we were asked to close more URLs than will fit in a
+            // single command, chunk the URLs into multiple commands,
+            // packing as many as we can into each. Do this until we've either
+            // drained and packed all the URLs, or we see an oversized URL
+            // that won't fit into a single command.
+            let chunk = match payload_support::try_fit_items(
+                &urls,
+                // This can be simplified to `COMMAND_MAX_PAYLOAD_SIZE.get()`
+                // once we bump our MSRV to 1.73.0.
+                COMMAND_MAX_PAYLOAD_SIZE.with(Cell::get),
```
linabutler, line 47: The `LocalKey<Cell<T>>` convenience methods were [stabilized in 1.73.0](https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html#thread-local-initialization).

m-c current MSRV is [1.76.0](https://searchfox.org/mozilla-central/rev/0b1d02b2cb5736511139cf0e40b318273e825899/python/mozboot/mozboot/util.py#14); ours is [1.70.0](https://github.com/mozilla/application-services/blob/16fdc3b6fee4c5404e77f7034822f5b5d9c464e2/.circleci/config.yml#L16). I wonder if we could bump ours?

components/fxa-client/src/fxa_client.udl
```diff
@@ -91,6 +91,20 @@ enum FxaError {
...
 
 
+/// An error type that carries additional information for
+/// device command-related [`FirefoxAccount`] operations.
+[Error]
+interface DeviceCommandError {
+  /// An error thrown if a device command to close one or more tabs
+  /// couldn't be sent. The caller should try to invoke
+  /// [`FirefoxAccount::close_tabs`] with the returned `urls` later.
+  TabsNotClosed(sequence<string> urls);
```
mhammond, line 101: Shouldn't we be able to differentiate transient vs permanent errors here? Naively, it looks like an oversized URL is going to be retried until it expires, which doesn't seem ideal?

Also, it seems mildly odd that an error may actually mean "partially successful" - I wonder if this might be better represented directly as a return value or similar (eg, maybe a struct that holds a vec of the commands which failed or similar, which might allow it to be reused for all commands)

This is tricky though: eg, what to return when "we tried to make 2 actual requests where the first worked but second failed transiently", and how to differentiate that from "we tried to make 2 actual requests where the first worked but we declined to even try and make the second due to the payload size limitations"

We can think that through a little more if you think that's work pursuing - I'm adding this comment without thinking it through too far to see how you react :)

components/fxa-client/src/internal/close_tabs.rs
```diff
@@ -2,36 +2,91 @@
...
+                }
+            };
+
+            let sent_telemetry = sent_telemetry.clone_with_new_stream_id();
+            let payload = CloseTabsPayload::with_telemetry(&sent_telemetry, chunk);
+
+            let oldsync_key = self.get_scoped_key(scopes::OLD_SYNC)?;
+            let command_payload =
+                encrypt_command(oldsync_key, target, close_tabs::COMMAND_NAME, &payload)?;
+            let result = self.invoke_command(
```
mhammond, line 62: (eg, here could be a transient network error, whole the `Fit::None` case is a permanent error, right?)

mhammond (commented):
This looks great, although I'm not sure about the new return/error semantics (not sure about both what is here and what I think *should* be here!)

components/fxa-client/src/fxa_client.udl
linabutler: It's a great question! ☺️

> Naively, it looks like an oversized URL is going to be retried until it expires, which doesn't seem ideal?

You're totally right; an oversized URL is going to be retried until it expires. My thought was, since we know we aren't going to be able to send it, anyway [1], this would be a simple way to give the user a "cool-off" period.

Android will still reschedule the worker to retry flushing the command for that oversized URL [2], but we won't ever make a request to the FxA server with the oversized URL, and, crucially, the tab for that URL will remain hidden from the list of synced tabs until the command expires.

If we treated a permanent error as a signal to clear the command from the queue, I think that would (ironically) look like a bug to the user: they tapped the "close" button, the tab disappeared, they saw the "undo" snackbar, the snackbar disappeared, and then, a few seconds later, the URL they just closed reappears again. If they've since scrolled down to close more tabs, the tab reappearing could also throw off their scroll position.

I don't think we'd _need_ to treat a permanent error as a signal to remove the command. We could still keep the command in the queue, but without trying to resend it, until it expires. But I think that would complicate our error handling path, because we'd now need to be aware of special commands that we want to keep in the queue, but not send.

Does that make sense, or does my thinking feel off?

I think the TL;DR summary of my thoughts is: we already gave the user visual feedback that we closed the tab, and they've likely already moved on. If we know that we're never going to be able to send the command, then keeping the command in the queue—and the tab-that-can-never-be-closed hidden from the list of synced tabs—until it expires a day later, feels a bit smoother than giving them confusing feedback when the tab reappears in a few seconds.

> Also, it seems mildly odd that an error may actually mean "partially successful" - I wonder if this might be better represented directly as a return value or similar

I definitely felt that tension!

I also think the throwing semantics of Kotlin and Swift make this interesting. In Rust, a `Result::Err` actually feels like the perfect way to model the difference between "everything worked" and "something didn't work; inspect the result to see what went wrong." Rust's `Result<T, U>` is generic over `T` and `U`, so both variants can have meaningful semantics, while Swift encourages type erasure for errors [3], and Kotlin has JVM exceptions.

Given that tension—and also what you mentioned in the next paragraph, about communicating what a "partially successful request" means—I opted to model `DeviceCommandError::TabsNotClosed` as a recoverable error, with context for the caller. (I think `TabsNotClosed` would make even more sense as a variant of `FxaError`, but `FxaError` is currently a field-less enum, and UDL doesn't let us mix variants with and without fields).

[1]: ...Unless FxA happens to do a deploy that bumps the payload limit within the 24 hours of the command being queued 😆
[2]: These retries will use [exponential backoff](https://searchfox.org/mozilla-central/rev/f7b41fc41c5505db507d076c16961a8da67dd318/mobile/android/android-components/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/commands/SyncedTabsCommandsFlushScheduler.kt#56), and the worker will run and return almost instantly. When the user closes another tab, we'll [interrupt the retry loop](https://searchfox.org/mozilla-central/rev/f7b41fc41c5505db507d076c16961a8da67dd318/mobile/android/android-components/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/commands/SyncedTabsCommandsFlushScheduler.kt#36), and schedule a flush to run within 10 seconds. So, even if there's an oversized URL in the queue, it won't block us from sending other commands, if the user wants to close more tabs.
[3]: In Swift, you can still inspect propagated errors with `catch` patterns, but the type of the error doesn't feel as semantically significant as in Rust. Swift also has a [`Result` type](https://developer.apple.com/documentation/swift/result) analogous to Rust's, but it's uncommon to see it used as a return type. I think Swift's `Result` was mostly used for fallible async APIs, and now that Swift has async functions, `async throws` has replaced that use case.

components/fxa-client/src/internal/close_tabs.rs
linabutler: Our MSRV is now 1.76.0, as of #6301, so we can use those convenience methods now! 💃🏻

components/fxa-client/src/internal/close_tabs.rs
skhamis: definitely love to see it though!!

components/fxa-client/src/fxa_client.udl
skhamis: > I think the TL;DR summary of my thoughts is: we already gave the user visual feedback that we closed the tab, and they've likely already moved on. If we know that we're never going to be able to send the command, then keeping the command in the queue—and the tab-that-can-never-be-closed hidden from the list of synced tabs—until it expires a day later, feels a bit smoother than giving them confusing feedback when the tab reappears in a few seconds.

I agree with @linabutler here and this is also the way desktop is going to go as well. If we get a high amount of people hitting oversized URLs we can maybe revisit this?

skhamis (commented):
LGTM and don't see anything that would give me pause landing this (thanks for getting us up to speed on Rust with m-c!) -- last thing seems to be settling on the error handling conversation.

components/fxa-client/src/fxa_client.udl
mhammond: > If we treated a permanent error as a signal to clear the command from the queue, I think that would (ironically) look like a bug to the user: they tapped the "close" button, the tab disappeared, they saw the "undo" snackbar, the snackbar disappeared, and then, a few seconds later, the URL they just closed reappears again. If they've since scrolled down to close more tabs, the tab reappearing could also throw off their scroll position.

I'm perfectly fine with that - however, it's worth being explicit that it's still a bug from the user's POV - that oversized URL is not closed, and it is going to come back. But there's no general solution here and it's an edge-case that this seems fine to me. It is work spelling this out in the comments though, otherwise it risks looking like an oversight.

> > Also, it seems mildly odd that an error may actually mean "partially successful" - I wonder if this might be better represented directly as a return value or similar

I guess I don't quite understand why we need this at all then? As noted above, we expect clients to ignore "permanent" failure as we expect it to act as though the command was actually send (ie, we will hide the tab either way). A transient failure is also something we don't expect the client to deal with because we assume it will work eventually.

So what's the point of this new error result? What will we expect the callers to do with it?

linabutler: > So what's the point of this new error result? What will we expect the callers to do with it?

We'll expect callers to inspect the returned URLs, and skip calling `setPendingCommandSent()` for them. Here's a quick sketch of how that'll look on Android, with some Android Components machinery elided:


```kt
val requestedUrls = outgoingCommand.urls
val closedUrls = try {
    account.closeTabs(targetDeviceId, requestedUrls)
    requestedUrls
} catch (e: DeviceCommandException.TabsNotClosed) {
    // If the user wanted to close multiple tabs with the same URL, and
    // more than one of those tabs couldn't be closed, the count for
    // that tab's URL will be > 1.
    val countsByNotClosedUrl = e.urls.groupingBy { it }.eachCountTo(mutableMapOf())
    requestedUrls.filter { url ->
        val notClosedCount = countsByNotClosedUrl.getOrDefault(url, 0)
        if (notClosedCount > 0) {
            // A tab that the user tried to close couldn't be closed,
            // so we omit it from the list of closed URLs.
            countsByNotClosedUrl[url] = notClosedCount - 1
            false
        } else {
            true
        }
    }
}

// Iterate over the pending tabs commands; only call
// `setPendingCommandSent()` for commands with URLs in
// `closedUrls`.
```

components/fxa-client/src/internal/close_tabs.rs
mhammond: We discussed the reasoning behind this, so can you please add a comment summarizing that here, at least enough so it's clear the behaviour is intentional.

components/fxa-client/src/fxa_client.udl
linabutler: > How is this model better than simply returning the Urls which were closed? That seems to offer the same general semantics and makes the above code snippet far clearer?

I pushed a follow-up commit to switch us to the model you're describing, and I actually prefer it to what I had! Thank you for pushing back on that a bit! 🚀

I'm realizing I didn't explain this (sorry!), but one other thing I had in mind with the new `DeviceCommandError` type is how we'd consume it on Android. Android Components' wrapper around our Rust client has a single API, [`sendCommandToDevice`](https://searchfox.org/mozilla-central/rev/87630817796aae3cbceb369a1b642da3e4741745/mobile/android/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt#181-218), that's generic for all types of device commands.

I think that abstraction makes sense for them, _but_ it complicates the handling of return types a little, because "close tabs" now has a partial success case that "send tab" doesn't.

I spent a bit of time figuring out how to express that in Kotlin, and settled on bubbling up a `TabsNotClosed` exception as being the clearest way. I rejected having an enum like this:

```kt
sealed interface SendCommandToDeviceResult {
    data object Ok : SendCommandToDeviceResult
    data object Error : SendCommandToDeviceResult
    data class TabsNotClosed(val urls: List<String>) : SendCommandToDeviceResult
}

/* partial */ interface DeviceConstellation {
    suspend fun sendCommandToDevice(targetDeviceId: String, outgoingCommand: DeviceCommandOutgoing): SendCommandToDeviceResult
}
```

Because `SendCommandToDeviceResult.TabsNotClosed` only applies to `DeviceCommandOutgoing.CloseTabs`. A `TabsNotClosed` result for `SendTab` is impossible, but an exhaustive `when` over `SendCommandToDeviceResult` would still need an arm for it.

An exception felt cleaner to me, because a caller would only need to be aware of (and catch) the `TabsNotClosed` error if it passed a `DeviceCommandOutgoing.CloseTabs`.

_But_, I think I got too many steps ahead in thinking about that, and missed the forest for the trees a little. There's no reason why Android Components' wrapper has to influence whether our Rust API returns a `CloseTabsResult`. We can still do the latter here, then have the Android `DeviceConstellation` wrapper inspect that result, and turn it into a `SendCommandToDeviceException.TabsNotClosed`, to bubble up to the `RemoteTabsCommandQueue` layer.

(All these different abstractions and wrappers sure complicate how we communicate state! 😅)

linabutler: My last comment was getting way too long, but I wanted to also say something about:

> simply returning the Urls which _were_ closed

I think that's a judgment call! In the common case, we expect all the tabs passed to `close_tabs` to be closed; I didn't think it was worth the serialization overhead of passing the same URLs back over the FFI.

But they are analogous, and handling both—especially when duplicates are in the picture; more on that in a bit—is equally easy.

That is, if we return `SomeTabsClosed` with the tabs that we _did_ close, figuring out which commands we sent is a subset operation; if we return `TabsNotClosed` with the URLs that we _didn't_ close, figuring out which ones we sent is a set subtraction.

components/fxa-client/src/internal/close_tabs.rs
linabutler: > Isn't that impossible in our model? The command+url must be unique.

I think that's a bug in our current model, though?

IIRC, synced tabs don't filter out duplicates, so I thought we agreed on the heuristic that ["if we see N duplicates of a URL in the command, we'll close N instances of that tab"](https://searchfox.org/mozilla-central/rev/87630817796aae3cbceb369a1b642da3e4741745/mobile/android/android-components/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/CloseTabsCommandReceiver.kt#60-64). There's no way to know _which_ N it is, though.

Did we decide to change those semantics? (It's OK if we did!) I was assuming we'd come back to that, but if we actually think that the "get N-close N" semantics aren't helpful, I can yank that out of Android Components.

components/fxa-client/src/fxa_client.udl
linabutler: > Isn't that impossible in our model? The command+url must be unique.

I think that's a bug in our current model, though?

IIRC, synced tabs don't filter out duplicates, so I thought we agreed on the heuristic that ["if we see N duplicates of a URL in the command, we'll close N instances of that tab"](https://searchfox.org/mozilla-central/rev/87630817796aae3cbceb369a1b642da3e4741745/mobile/android/android-components/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/CloseTabsCommandReceiver.kt#60-64). There's no way to know _which_ N it is, though.

Did we decide to change those semantics? (It's OK if we did!) I was assuming we'd come back to that, but if we actually think that the "get N-close N" semantics aren't helpful, I can yank that out of Android Components.

components/fxa-client/src/internal/close_tabs.rs
linabutler: Done and done!

components/fxa-client/src/fxa_client.udl
linabutler: To close the loop, here's what the Android side looks like with the latest changes in this PR:

* A change to make `DeviceConstellation.sendCommandToDevice()` support a different return type for close tabs: https://phabricator.services.mozilla.com/D216647
* Adopting `CloseTabsResult` in Android Components, and resending commands for any URLs that fail to send: https://phabricator.services.mozilla.com/D216649

mhammond: > Did we decide to change those semantics? (It's OK if we did!) 

I think we just ignored this hoping it would go away :) I think the main decision here is actually about showing duplicate tabs on the surfaces, and, of course, desktop does it differently in different surfaces. Assuming we make it consistent: if we do not show dupe tabs, then the current behaviour seems correct and we'd need to close *all* matching. If we do show dupes things get much messier - eg, we'd also need to change the tabs logic to count the matching tabs when deciding to filter too, which seems error prone.

For desktop, the surfaces with the most recent UX love are View and Sidebar and they decided to drop duplicates. The fact that the toolbar menus did not is, IMO, more an oversight. It will be very difficult to justify having 2 different behaviours. So I think we should just declare that we expect all surfaces to drop duplicate tabs and run that past UX (who I believe will agree given the precedent of view/sidebar), meaning this restriction can remain, which in turn means the implementation of the actual closing will need to close all.

mhammond (approved):
Sorry for all the back-and-forth, but I think this is great!

PR #6302, opened 2024-07-11
Remove fakespot feature flag

bendk:
The plan is to use the current code for the experiment, so we don't need a feature flag anymore.  The main point of the flag was to avoid creating multiple migrations as the schema changed, but it looks like the schema will be stable for a while.

  - Note:


ncloudioj (approved):
Looks good.

One clarifying questions: On desktop, the component will ingest all the suggestion types including FS, will mobile ingest the FS data as well? I can't recall if we specify the suggestion types on mobile. If so, would that be a concern?

bendk:
> Looks good.
> 
> One clarifying questions: On desktop, the component will ingest all the suggestion types including FS, will mobile ingest the FS data as well? I can't recall if we specify the suggestion types on mobile. If so, would that be a concern?

Great point.  Looks like both [Android](https://searchfox.org/mozilla-central/source/mobile/android/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt#67) and [iOS](https://github.com/mozilla-mobile/firefox-ios/blob/dc9a11db3765513cc7fcd84ad3bc8b2d8985d99f/firefox-ios/Storage/Rust/RustFirefoxSuggest.swift#L62) don't specify a list of providers.  I'm just updated the code so that fakespot is not listed in the default providers so that we don't change their behavior.  Going forward, I think we can update consumers to specify the list of providers they want if it's different than the current list.

PR #6303, opened 2024-07-12
Bug 1907577 - Fakespot icon support

bendk:
Fetch the icons using a separate query than manually joined them with the suggestion data.  This has the advantage of not needing to update the schema.

Also updated the suggest CLI a bit so that I could use it to test that the code was working in practice.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -690,7 +690,9 @@ impl<'a> SuggestDao<'a> {
                 f.fakespot_grade,
                 f.product_id,
                 f.rating,
-                f.total_reviews
+                f.total_reviews,
+                i.data,
+                i.mimetype
             FROM
                 suggestions s
             JOIN
@@ -699,6 +701,9 @@ impl<'a> SuggestDao<'a> {
             JOIN
                 fakespot_custom_details f
                 ON f.suggestion_id = s.id
+            LEFT JOIN
+                icons i
+                ON i.id = f.icon_id
             WHERE
                 fakespot_fts MATCH ?
             ORDER BY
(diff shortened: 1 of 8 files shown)


components/suggest/src/db.rs
```diff
@@ -681,6 +685,8 @@ impl<'a> SuggestDao<'a> {
 
     /// Fetches fakespot suggestions
     pub fn fetch_fakespot_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        let icon_map = IconMap::fakespot(self.conn)?;
```
ncloudioj, line 688: What do you think about caching the icon map in memory so that it doesn't need to query the icon table for all FS icons and create the map for every single FS query? The cache can be refreshed upon re-ingestion.

Alternatively, we can follow other providers by storing the icon ID in the `fakespot_custom_details` table during ingestion, then use a table join to fetch the icon.

bendk: I like the idea of a DB cache a lot, but it's actually not so easy to implement.  The most obvious way to do it would be to store the cache inside the mutex with `rusqlite::Connection`, then have `SuggestDao` hold a reference to it.  The issue with that is that all the reader methods `SuggestDao` take a shared ref, so you can't re-populate the cache when it's empty.  You could put the cache inside an inner mutex, but it's pretty wasteful and unintuitive to do that when you've just locked the outer mutex.  The root of the issue here is that `SuggestDao` uses a shared reference to indicate some methods are reader methods, even though we have exclusive access to connection because we just locked the mutex.  I'm not sure what the best solution here is.  We could create 2 structs `SuggestDaoWriter` and `SuggestDaoReader` and implement some form of forwarding where you can call the reader methods from the reader.   Alternatively, we could keep 1 struct and use an UnsafeCell since we know that we actually do have exclusive access.  I don't really love either option and can't think of a better one.

I went with the second approach, my main reason not to do it was to avoid the schema migration, but I think it should be fine.  Almost all users will run migrations 20 and 21 at the same time so we shouldn't need to delete and re-ingesting the fakespot data.

ncloudioj: Sounds good! The second approach is also consistent with other providers, so let's do that.

ncloudioj (approved):

PR #6313, opened 2024-07-24
FTS prefix configuration ADR

bendk:
[Rendered version](https://github.com/bendk/application-services/blob/adr-fts-prefix/docs/adr/0008-fts-prefix-config.md)

The DISCO team discussed the FTS prefix setting and decided that we should do some benchmarking to see what the differences were.  This was actually not so simple and I fell down a bit of a rabbit hole, but I'm back with some benchmarks and a recommendation that I wrote up as an ADR.  The decision is still pending so please suggest other options if you think one of them would be better, it doesn't feel like a clear decision to me at all.

CC @clouserw

  - Note:

### DIFF
--- a/docs/adr/0008-fts-prefix-config.md
+++ b/docs/adr/0008-fts-prefix-config.md
@@ -1,0 +1,162 @@
+# FTS prefix configuration
+
+* Status: Accepted
+* Deciders: Ben Dean-Kawamura, Will Stuckey, Drew Willcoxon, Nan Jiang
+* Date: 2024-07-25
+
+
+## Context and Problem Statement
+
+The Suggest component will use the [SQLite FTS5 extension](https://sqlite.org/fts5.html) to index suggestions for the Fakespot project.
+FTS queries will usually be prefix queries -- if the user types in "abcd" we will match any word that begins with "abcd".
+The FTS extension supports configurable [prefix indexes](https://sqlite.org/fts5.html#prefix_indexes) to optimize these queries.
+This ADR will explore different parameters for the prefix index configuration and decide on a value for the Fakespot experiment.
+
+This decision was based on benchmarks from the current code and data set.
+We should revisit this in the future if the Fakespot data set or query strategy changes or for new FTS-based suggestion providers.
+Hopefully it can serve as a starting point for these future decisions.
+
+
+## Considered Options
+
+In general, prefix indexes can optimize up certain queries at the cost of database size and ingestion time.
+Increased database size will often slow down the non-optimize queries.
+This ADR will benchmark and compare a large range of options.
+
+### A) no prefix index
+### B) prefix='1'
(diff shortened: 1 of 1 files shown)


quiiver:
My take away from this is that it doesn't make a material difference in user experience if we use a prefix index at all. Really any query under 100ms will appear instantaneous from a user's perspective. Probably as the data in the fts index scales query optimization will come into play, however I would recommend option A as it results in the smallest db size.

bendk:
I should also mention that there's a ton more benchmarking that could be done and I'm happy to test other suggestions.

0c0w3:
Thanks for putting this together. Overall it seems like a low-stakes decision since on one hand, as Wil said, so far these queries seem plenty fast enough even without indexes (I'm on a fast machine though), and on the other hand it doesn't appear that adding new prefix indexes increases the DB size all that much. Earlier I [found](https://docs.google.com/document/d/1vl-kPh2JSrA9ANlqzrE1XA1qpwXm0poXL-TE0s6EAQs/edit?usp=sharing) that each new prefix index even in a DB with 300k suggestions increased its size by only 3 MB. It might also be worth noting that desktop Firefox doesn't use indexes when it matches queries against history and bookmark titles in the urlbar. Not a reason Suggest shouldn't, just a comparison.

Wdyt about implementing [query telemetry](https://bugzilla.mozilla.org/show_bug.cgi?id=1908397) and using that data to make a decision? It could record word and char counts, whether a prefix query was triggered, etc.

ncloudioj:
Thanks for putting things up together, Ben!

I am also a fan of starting off w/o prefix indices and instrumenting query metrics for future decision making.

Just curious, I assume the test queries are all positive queries (query hits), correct? Did you test any negative queries (query misses)?

bendk:
I just pushed an update that adds benchmarks for ingestion time.  I completely forgot that I also ran those yesterday.  I also updated the decision to a) since that was the general consensus.  A sounds good to me as well, my one concern is when we move to mobile.  A 1ms benchmark result on my machine might be much more on an old phone with a cold cache.  But that's pretty speculative and it's very possible it won't be that much more and/or user's won't notice.

> Wdyt about implementing [query telemetry](https://bugzilla.mozilla.org/show_bug.cgi?id=1908397) and using that data to make a decision? It could record word and char counts, whether a prefix query was triggered, etc.

That would be very nice.  I'm not sure exactly how to measure and present all of that, but I'll try to include this data somehow.

> Just curious, I assume the test queries are all positive queries (query hits), correct? Did you test any negative queries (query misses)?

Yes, all queries were hits.

bendk:
Seems like we have a decision.  What do you all think about updating the header to say approved and merging this one?

ncloudioj (approved):
Looks great. Option A is the winner for now.

PR #6319, opened 2024-07-26
Suggest ingestion refactor

bendk:
This is prep-work for https://bugzilla.mozilla.org/show_bug.cgi?id=1908802.  The main goal is to simplify ingestion in certain places.

Removed the `SuggestIngestionConstraints::max_suggestions` field. AFAICT, it was not being used by any consumers and doesn't seem to be implemented correctly.  We apply the limit to each record type, but a single ingestion request will have many record types.  It also doesn't seem clear to me how this should work, for example should the config/icons count?  I want to update this code, but I don't want to worry about this field.

Updated ingestion code so that it loops over providers rather than record types.  This is needed to handle deletion.  When we see a deleted record, we don't see any of the old fields.  This makes it hard to know if the record matches the ingestion constraints and if we should actually delete it. The suggestions table does have a `provider` field, so I think we can add that to the WHERE clause in `drop_suggestions`, effectively making it a no-op if the provider didn't match. However, that only works if we loop over providers rather than record types.o

A side effect of this is that we always ingest the icons/config, but I don't think there will be any real performance hit for this.

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -4,7 +4,6 @@
  */
 
 use std::{
-    collections::BTreeSet,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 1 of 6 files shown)


components/suggest/src/store.rs
```diff
@@ -367,41 +360,45 @@ where
...
         breadcrumb!("Ingestion complete");
 
         Ok(())
     }
 
-    fn ingest_records_by_type(
+    fn fetch_and_ingest_records(
         &self,
         ingest_record_type: SuggestRecordType,
+        _provider: Option<SuggestionProvider>,
```
bendk, line 394: My WIP commit to fix 1908802 will use this.

bendk:
The more I think about this one, the more I want to refactor some more things.  I'm going to close this one and re-open once I have those in place.

PR #6322, opened 2024-07-30
update Rust for recorded context to handle event store queries

jeddai:

  - Note:

### DIFF
--- a/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
+++ b/components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
@@ -24,6 +24,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
+import org.junit.Assert.assertThrows
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Ignore
@@ -37,7 +38,9 @@ import org.mozilla.experiments.nimbus.GleanMetrics.NimbusHealth
 import org.mozilla.experiments.nimbus.internal.EnrollmentChangeEvent
 import org.mozilla.experiments.nimbus.internal.EnrollmentChangeEventType
 import org.mozilla.experiments.nimbus.internal.JsonObject
+import org.mozilla.experiments.nimbus.internal.NimbusException
 import org.mozilla.experiments.nimbus.internal.RecordedContext
+import org.mozilla.experiments.nimbus.internal.validateEventQueries
 import org.robolectric.RobolectricTestRunner
 import java.util.Calendar
 import java.util.concurrent.Executors
@@ -71,6 +74,7 @@ class NimbusTests {
     private fun createNimbus(
         coenrollingFeatureIds: List<String> = listOf(),
         recordedContext: RecordedContext? = null,
+        block: Nimbus.() -> Unit = {},
     ) = Nimbus(
         context = context,
         appInfo = appInfo,
@@ -80,7 +84,7 @@ class NimbusTests {
         observer = null,
         delegate = nimbusDelegate,
         recordedContext = recordedContext,
-    )
+    ).also(block)
 
     @get:Rule
     val gleanRule = GleanTestRule(context)
@@ -734,21 +738,34 @@ class NimbusTests {
         assertEquals("Event count must match", isReadyEvents.count(), 3)
     }
 
+    class TestRecordedContext(
+        private val eventQueries: Map<String, String>? = null,
+        private var eventQueryValues: Map<String, Double>? = null,
+    ) : RecordedContext {
+        var recorded = mutableListOf<JSONObject>()
+
+        override fun getEventQueries(): Map<String, String> {
+            return eventQueries?.toMap() ?: mapOf()
+        }
+
+        override fun setEventQueryValues(eventQueryValues: Map<String, Double>) {
+            this.eventQueryValues = eventQueryValues
+        }
+
+        override fun record() {
+            recorded.add(this.toJson())
+        }
+
+        override fun toJson(): JsonObject {
+            val contextToRecord = JSONObject()
+            contextToRecord.put("enabled", true)
+            contextToRecord.put("events", JSONObject(eventQueryValues ?: mapOf<String, Double>()))
+            return contextToRecord
+        }
+    }
+
     @Test
     fun `Nimbus records context if it's passed in`() {
-        class TestRecordedContext : RecordedContext {
-            var recordCount = 0
-
-            override fun record() {
-                recordCount++
-            }
-
-            override fun toJson(): JsonObject {
-                val contextToRecord = JSONObject()
-                contextToRecord.put("enabled", true)
-                return contextToRecord
-            }
-        }
         val context = TestRecordedContext()
         val nimbus = createNimbus(recordedContext = context)
 
@@ -761,7 +778,42 @@ class NimbusTests {
             job.join()
         }
 
-        assertEquals(context.recordCount, 1)
+        assertEquals(context.recorded.size, 1)
+    }
+
+    @Test
+    fun `Nimbus recorded context event queries are run and the value is written back into the object`() {
+        val context = TestRecordedContext(
+            mapOf(
+                "TEST_QUERY" to "'event'|eventSum('Days', 1, 0)",
+            ),
+        )
+        val nimbus = createNimbus(recordedContext = context) {
+            recordEvent("event")
+        }
+
+        suspend fun getString(): String {
+            return testExperimentsJsonString(appInfo, packageName)
+        }
+
+        val job = nimbus.applyLocalExperiments(::getString)
+        runBlocking {
+            job.join()
+        }
+
+        assertEquals(context.recorded.size, 1)
+        assertEquals(context.recorded[0].getJSONObject("events").getDouble("TEST_QUERY"), 1.0, 0.0)
+    }
+
+    @Test
+    fun `Nimbus recorded context event queries are validated`() {
+        val context = TestRecordedContext(
+            mapOf(
+                "FAILING_QUERY" to "'event'|eventSumThisWillFail('Days', 1, 0)",
+            ),
+        )
+
+        assertThrows("Expected an error to be thrown", NimbusException::class.java, { validateEventQueries(context) })
     }
 }
 
--- a/components/nimbus/src/stateful/behavior.rs
+++ b/components/nimbus/src/stateful/behavior.rs
@@ -62,6 +62,7 @@ impl PartialEq for Interval {
         self.to_string() == other.to_string()
     }
 }
+
 impl Eq for Interval {}
 
 impl Hash for Interval {
@@ -84,7 +85,7 @@ impl FromStr for Interval {
             _ => {
                 return Err(NimbusError::BehaviorError(
                     BehaviorError::IntervalParseError(input.to_string()),
-                ))
+                ));
             }
         })
     }
@@ -365,7 +366,7 @@ impl EventQueryType {
                 return Err(NimbusError::TransformParameterError(format!(
                     "event transform {} requires a positive number as the second parameter",
                     self
-                )))
+                )));
             }
         } as usize;
         let zero = &Value::from(0);
@@ -375,7 +376,7 @@ impl EventQueryType {
                 return Err(NimbusError::TransformParameterError(format!(
                     "event transform {} requires a positive number as the third parameter",
                     self
-                )))
+                )));
             }
         } as usize;
 
@@ -402,7 +403,7 @@ impl EventQueryType {
                 return Err(NimbusError::TransformParameterError(format!(
                     "event transform {} requires a positive number as the second parameter",
                     self
-                )))
+                )));
             }
         } as usize;
 
@@ -425,6 +426,13 @@ impl EventQueryType {
             | Self::AveragePerNonZeroInterval => self.validate_counting_arguments(args)?,
             Self::LastSeen => self.validate_last_seen_arguments(args)?,
         })
+    }
+
+    pub fn validate_query(maybe_query: &str) -> Result<bool> {
+        let regex = regex::Regex::new(
+            r#"^(?:"[^"']+"|'[^"']+')\|event(?:Sum|LastSeen|CountNonZero|Average|AveragePerNonZeroInterval)\(["'](?:Years|Months|Weeks|Days|Hours|Minutes)["'],\s*\d+\s*(?:,\s*\d+\s*)?\)$"#,
+        )?;
+        Ok(regex.is_match(maybe_query))
     }
 
     fn error_value(&self) -> f64 {
--- a/components/nimbus/src/stateful/nimbus_client.rs
+++ b/components/nimbus/src/stateful/nimbus_client.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+#[cfg(test)]
+use crate::tests::helpers::{TestMetrics, TestRecordedContext};
 use crate::{
     defaults::Defaults,
     enrollment::{
@@ -26,10 +28,10 @@ use crate::{
         },
         matcher::AppContext,
         persistence::{Database, StoreId, Writer},
+        targeting::{validate_event_queries, RecordedContext},
         updating::{read_and_remove_pending_experiments, write_pending_experiments},
     },
     strings::fmt_with_map,
-    targeting::RecordedContext,
     AvailableExperiment, AvailableRandomizationUnits, EnrolledExperiment, Experiment,
     ExperimentBranch, NimbusError, NimbusTargetingHelper, Result,
 };
@@ -42,9 +44,6 @@ use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, MutexGuard};
 use uuid::Uuid;
-
-#[cfg(test)]
-use crate::tests::helpers::{TestMetrics, TestRecordedContext};
 
 const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
 pub const DB_KEY_INSTALLATION_DATE: &str = "installation-date";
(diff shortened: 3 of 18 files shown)


components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +35,71 @@ impl NimbusTargetingHelper {
...
+}
+
+impl dyn RecordedContext {
+    pub fn execute_queries(
+        &self,
+        nimbus_targeting_helper: &NimbusTargetingHelper,
+    ) -> Result<Map<String, Value>> {
+        let results = Map::from_iter(self.get_event_queries().iter().filter_map(
+            |(key, value): (&String, &Value)| match value.as_str() {
+                Some(v) => match EventQueryType::validate_query(v) {
```
brennie, line 69: I was under the impression we were only going to do this for debug builds and that we were going to cause a hard error e.g., crash, so that invalid queries couldn't land?

components/nimbus/src/stateful/behavior.rs
```diff
@@ -427,6 +429,13 @@ impl EventQueryType {
         })
     }
 
+    pub fn validate_query(maybe_query: &str) -> Result<bool> {
+        let regex = Regex::new(
+            r#"^["'][^"']+["']\|event(?:Sum|LastSeen|CountNonZero|Average|AveragePerNonZeroInterval)\(["'](?:Years|Months|Weeks|Days|Hours|Minutes)["'], \d+(?:, \d+)?\)$"#,
+        )?;
```
brennie, line 435: > Note that regex compilation tends to be a somewhat expensive process, and unlike higher level environments, compilation is not automatically cached for you. One should endeavor to compile a regex once and then reuse it. For example, it’s a bad idea to compile the same regex repeatedly in a loop.

If we want to cache this we should use https://doc.rust-lang.org/stable/std/cell/struct.OnceCell.html or https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html (threadsafe).

OR if we decide to only enable this check in debug builds, then we can take the hit, it doesn't matter

brennie (changes requested):

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+    ) -> Result<Map<String, Value>> {
+        let results = Map::from_iter(self.get_event_queries().iter().filter_map(
+            |(key, value): (&String, &Value)| match value.as_str() {
+                Some(v) => match nimbus_targeting_helper.evaluate_jexl_raw_value(v.into()) {
+                    Ok(result) => Some((key.to_string(), result)),
+                    Err(err) => {
+                        log::info!(
+                            "error during jexl evaluation for query {} — {}",
+                            value,
+                            err.to_string()
```
brennie, line 76: You dont need to .to_string() this. the formatting in info! will do it for you.

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+                        );
+                        None
+                    }
+                },
+                None => {
+                    log::info!("value {} for key {} is not a string", key, value);
+                    Some((key.to_string(), json!(value)))
+                }
```
brennie, line 84: imo if they add non-string queries we should omit them.

Also will this logging ever be seen anywhere?

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+
+impl dyn RecordedContext {
+    pub fn execute_queries(
+        &self,
+        nimbus_targeting_helper: &NimbusTargetingHelper,
+    ) -> Result<Map<String, Value>> {
+        let results = Map::from_iter(self.get_event_queries().iter().filter_map(
+            |(key, value): (&String, &Value)| match value.as_str() {
+                Some(v) => match nimbus_targeting_helper.evaluate_jexl_raw_value(v.into()) {
+                    Ok(result) => Some((key.to_string(), result)),
```
brennie, line 71: key is a &String, we can .clone() to coerce to String

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+    fn record(&self);
+}
+
+impl dyn RecordedContext {
+    pub fn execute_queries(
+        &self,
+        nimbus_targeting_helper: &NimbusTargetingHelper,
+    ) -> Result<Map<String, Value>> {
+        let results = Map::from_iter(self.get_event_queries().iter().filter_map(
+            |(key, value): (&String, &Value)| match value.as_str() {
```
brennie, line 69: You don't need to hint the types here. The compiler is smart enough to figure it out.

components/nimbus/src/targeting.rs
```diff
@@ -61,6 +48,16 @@ impl NimbusTargetingHelper {
         }
     }
 
+    pub fn evaluate_jexl_raw_value(&self, expr: String) -> Result<Value> {
```
brennie, line 51: This doesn't need to take a String. You're immediately taking a ref to pass it to jexl_eval_raw. It should take a &str

```suggestion
    pub fn evaluate_jexl_raw_value(&self, expr: &str) -> Result<Value> {
```

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+                Some(v) => match nimbus_targeting_helper.evaluate_jexl_raw_value(v.into()) {
```
brennie, line 70: if evaluate_jexl_raw_value doesnt take String, we dont need into() here.

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+                        );
+                        None
+                    }
+                },
+                None => {
+                    log::info!("value {} for key {} is not a string", key, value);
+                    Some((key.to_string(), json!(value)))
+                }
+            },
+        ));
```
brennie, line 86: Its more idiomatic to use Iterator::collect<T> to make a T when T: FromIterator. See [docs](https://doc.rust-lang.org/nightly/core/iter/trait.FromIterator.html)

We can also clean this up to use a single match:

```suggestion
    let results: Map<_, _> = m
        .iter()        
        .filter_map(|(key, value)|
            match value.as_str().map(|v| nimbus_targeting_helper.evaluate_jexl_raw_value(v)) {
                Some(Ok(v)) => Some((key.clone(), v)),
                Some(Err(e)) => {
                    log::info!(
                        "error during jexl evalaution for query {}={} - {}",
                        key,
                        value,
                        e,
                    );
                    None
                },
                None => {
                    log::info!("query {}={} does not map to string", key, value);
                    None
                },
            }      
        )
        .collect();

       
```

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+        self.set_event_query_values(results.clone());
+        Ok(results)
+    }
+
+    pub fn validate_queries(&self) -> Result<()> {
+        for (_, value) in self.get_event_queries().iter() {
```
brennie, line 92: You don't need to call .iter()

```rust
        for (_, value) in &self.get_event_queries() {
```

However, you're only using the values, so you can do:

```suggestion
        for value in self.get_event_queries().values() {
```

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+                            return Err(NimbusError::BehaviorError(
+                                BehaviorError::EventQueryParseError(v.to_string()),
+                            ))
+                        }
+                    },
+                    Err(e) => return Err(e),
+                },
+                None => {
+                    return Err(NimbusError::BehaviorError(BehaviorError::TypeError(
+                        value.to_string(),
```
brennie, line 109: I find .into() more idiomatic than to_string()

components/nimbus/src/stateful/targeting.rs
```diff
@@ -27,3 +36,82 @@ impl NimbusTargetingHelper {
...
+                        "string".into(),
+                    )))
+                }
```
brennie, line 112: `match` on bool is less idiomatic than if/else

```rust
Ok(result) => if result {
    continue;
} else {
    return Err(NimbusError::BehaviorError(BehaviorError::EventQueryParseError(v.to_string())));
}                       }
```

but its even more idiomatic to match on the value

```rust
Ok(true) => continue,
Ok(false) => return Err(NimbusError::BehaviorError(BehaviorError::EventQueryParseError(v.to_string()))),
```

but we can also clean this up into a single match
```rust
            match value.as_str().map(EventQueryType::validate_query) {
                Some(Ok(true)) => continue,
                Some(Ok(false)) => return Err(NimbusError::BehaviorError(
                                       BehaviorError::EventQueryParseError(v.to_string()),
                                   )),
                Some(Err(e)) => return Err(e),
                None => {
                    return Err(NimbusError::BehaviorError(BehaviorError::TypeError(
                        value.to_string(),
                        "string".into(),
                    )))
                }
```

and finally we can use some Option and result methods (plus ?) to turn this into a linear chain of calls:

```suggestion
            let query = value
                .as_str()
                .ok_or_else(|| NimbusError::BehaviorError(BehaviorError::TypeError(
                    value.into(),
                    "string".into(),
                ))?;
            
            if !EventQueryType::validate_query(query)? {
                return Err(NimbusError::BehaviorError(
                    BehaviorError::EventQueryParseError(v.into()),
                ));
            }
```

components/nimbus/src/targeting.rs
```diff
@@ -131,7 +124,27 @@ pub fn jexl_eval<Context: serde::Serialize>(
         })
         .with_transform("bucketSample", bucket_sample);
 
-    let res = evaluator.eval_in_context(expression_statement, context)?;
+    match evaluator.eval_in_context(expression_statement, context) {
+        Ok(v) => Ok(v),
+        Err(err) => Err(NimbusError::EvaluationError(err.to_string())),
+    }
```
brennie, line 130: You want [`Result::map_err`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.map_err)

```suggestion
    evaluator
        .eval_in_context(expression_statement, context)
        .map_err(|err| NimbusError::EvaluationError(err.to_string()))
```

components/nimbus/src/targeting.rs
```diff
@@ -131,7 +124,27 @@ pub fn jexl_eval<Context: serde::Serialize>(
         })
         .with_transform("bucketSample", bucket_sample);
 
-    let res = evaluator.eval_in_context(expression_statement, context)?;
+    match evaluator.eval_in_context(expression_statement, context) {
+        Ok(v) => Ok(v),
+        Err(err) => Err(NimbusError::EvaluationError(err.to_string())),
```
brennie, line 129: is err a &str? if so, into()

components/nimbus/src/tests/helpers.rs
```diff
@@ -14,21 +14,41 @@ cfg_if::cfg_if! {
...
+use log::{Level, LevelFilter, Metadata, Record};
 use serde::Serialize;
 use serde_json::{json, Value};
 use std::collections::HashSet;
 use std::sync::{Arc, Mutex};
 
-cfg_if::cfg_if! {
-    if #[cfg(feature = "stateful")] {
-        use crate::stateful::behavior::EventStore;
+struct SimpleLogger;
```
brennie, line 29: TestLogger

components/nimbus/src/tests/stateful/test_nimbus.rs
```diff
@@ -1686,3 +1687,59 @@ fn test_recorded_context_recorded() -> Result<()> {
...
+    recorded_context.set_context(json!({
+        "app_version": "125.0.0",
+        "other": "stuff",
+    }));
+    recorded_context.set_event_queries(Map::from_iter(vec![(
+        "TEST_QUERY".to_string(),
+        serde_json::to_value("'event'|eventSum('Days', 1, 0)")?,
+    )]));
+    let client = NimbusClient::new(
+        app_context.clone(),
```
brennie, line 1714: You don't need to cloen this. It can take ownership since we're not using app_context below here.

brennie (changes requested):

components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
```diff
@@ -71,6 +74,7 @@ class NimbusTests {
     private fun createNimbus(
         coenrollingFeatureIds: List<String> = listOf(),
         recordedContext: RecordedContext? = null,
+        block: Nimbus.() -> Unit = {},
```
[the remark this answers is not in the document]
brennie: Couldn't you do `createNimbus(recordedContext=context).also { recordEvent("event") }` or smth?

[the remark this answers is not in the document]
jeddai: Yes to all of your questions, but I would argue it's a very common pattern to see used in kotlin libraries, e.g: [mockk](https://mockk.io/#constructor-mocks). What I added is very Kotlin-like, while your suggested change is more commonly seen in Java.

components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
```diff
@@ -734,21 +738,49 @@ class NimbusTests {
...
+        }
 
-            override fun toJson(): JsonObject {
-                val contextToRecord = JSONObject()
-                contextToRecord.put("enabled", true)
-                return contextToRecord
+        override fun setEventQueryValues(json: JsonObject) {
+            for (key in json.keys()) {
+                try {
+                    eventQueries?.put(key, json.getDouble(key))
```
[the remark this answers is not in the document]
jeddai: For the purpose of testing we can assume that they do

components/nimbus/src/stateful/nimbus_client.rs
```diff
@@ -745,6 +768,17 @@ impl NimbusClient {
                 })
             .expect("failed to unwrap RecordedContext object")
     }
+
+    pub fn validate_event_queries(&self) -> Result<()> {
```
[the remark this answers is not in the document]
jeddai: It is called through Kotlin/Swift

components/nimbus/src/stateful/behavior.rs
```diff
@@ -427,6 +428,13 @@ impl EventQueryType {
         })
     }
 
+    pub fn validate_query(maybe_query: &str) -> Result<bool> {
+        let regex = regex::Regex::new(
+            r#"^(?:"[^"']+"|'[^"']+')\|event(?:Sum|LastSeen|CountNonZero|Average|AveragePerNonZeroInterval)\(["'](?:Years|Months|Weeks|Days|Hours|Minutes)["'],\s*\d+\s*(?:,\s*\d+\s*)?\)$"#,
+        )?;
+        Ok(regex.is_match(maybe_query))
```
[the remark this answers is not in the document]
jeddai: yep yep! I need to add a comment on that method noting that

components/nimbus/src/targeting.rs
jeddai: err is a `jexl_eval::error::EvaluationError<'a>`, which doesn't support `.into()` on String/&str, but does have a `to_string` method

components/nimbus/src/stateful/targeting.rs
jeddai: `value` is a `serde_json::Value` and does not support `.into()`

components/nimbus/src/stateful/targeting.rs
jeddai: Yes — it'll be seen in test runs now thanks to the TestLogger, and will be visible in Kotlin logs as well

components/nimbus/src/stateful/targeting.rs
jeddai: Ah at some point during writing this it was mad I wasn't hinting the types, good call tho

components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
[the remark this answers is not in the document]
brennie: ok if this is more idiomatic in kotlin then its good with me :)

components/nimbus/src/stateful/nimbus_client.rs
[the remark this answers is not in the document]
brennie: yep!

components/nimbus/src/stateful/targeting.rs
jeddai: Actually, on the "None" section, we do need to return the value there because if its values are set more than once then it'll keep the value as it was already applied rather than removing it. That happens in tests and I suspect it'll happen in fenix because of how the messaging infra is set up

components/nimbus/src/stateful/nimbus_client.rs
```diff
@@ -161,7 +158,25 @@ impl NimbusClient {
...
-        self.event_store.lock().unwrap().read_from_db(db)?;
+        self.event_store
+            .lock()
+            .expect("unable to lock event_store mutex")
+            .read_from_db(db)?;
+
+        if let Some(recorded_context) = &self.recorded_context {
+            let targeting_helper = self.create_targeting_helper_with_context(serde_json::to_value(
+                state.targeting_attributes.clone(),
+            )?);
```
brennie, line 169: You don't need to clone, since there is a blanket impl for [`&T where T: Serialize`](https://docs.rs/serde/1.0.209/serde/ser/trait.Serialize.html#impl-Serialize-for-%26T)

```suggestion
            let targeting_helper = self.create_targeting_helper_with_context(serde_json::to_value(
                &state.targeting_attributest,
            )?);
```

brennie (approved):

components/nimbus/src/stateful/nimbus_client.rs
jeddai: It can't move `state`, so I do have to

components/nimbus/android/src/test/java/org/mozilla/experiments/nimbus/NimbusTests.kt
```diff
@@ -734,21 +738,34 @@ class NimbusTests {
...
+        }
 
-            override fun toJson(): JsonObject {
-                val contextToRecord = JSONObject()
-                contextToRecord.put("enabled", true)
-                return contextToRecord
-            }
+        override fun setEventQueryValues(eventQueryValues: Map<String, Double>) {
+            this.eventQueryValues = eventQueryValues
+        }
```
brennie, line 753: 👍

components/nimbus/src/stateful/nimbus_client.rs
brennie: You don't have to move state, `&state.targeting_attributes` borrows.

brennie (approved):

PR #6325, opened 2024-08-01
(DISCUSS) DISCO-2938 - return highlighted search matches

bendk:
In the last meeting we discussed the issue of results not being highlighted correctly because of stemming.  This is my attempt to try to fix that issue using the `highlight` function that FTS5 gives you.  I'm not totally sure if this is a good idea or not, I'd love feedback on it.

Some notes:
  - `highlight` doesn't work with a contentless table, so I disabled that.  This brings the size of the fakespot DB from ~3.6mb to 4mb.  The DB size for the other providers is ~20mb.  Seems like an acceptable tradeoff to me.
    - Maybe we could eventually use external content tables but that's harder to make work.  When I tried to do that I got SQLite errors when updating the data.  I haven't looked in to this, but I'm pretty sure it's related to deletion order.
  - I used `<b>` and `</b>` as the markers for the highlight, but we can switch that to whatever works for you or make it configurable.
  - My initial suggestion was to run the input through the stemming algorithm and return it to the component to use with its highlighting logic.  But, AFAICT, there's no way to do that with FTS5.

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -2144,25 +2144,36 @@ mod tests {
         store.ingest(SuggestIngestionConstraints::all_providers());
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::fakespot("globe")),
-            vec![snowglobe_suggestion()],
+            vec![snowglobe_suggestion(
+                "Make Your Own Glitter Snow <b>Globes</b>"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::fakespot("simpsons")),
-            vec![simpsons_suggestion()],
+            vec![simpsons_suggestion(
+                "The <b>Simpsons</b>: Skinner's Sense of Snow (DVD)"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::fakespot("snow")),
-            vec![simpsons_suggestion(), snowglobe_suggestion()],
+            vec![
+                simpsons_suggestion("The Simpsons: Skinner's Sense of <b>Snow</b> (DVD)"),
+                snowglobe_suggestion("Make Your Own Glitter <b>Snow</b> Globes"),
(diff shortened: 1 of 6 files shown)


components/suggest/src/store.rs
```diff
@@ -2184,15 +2195,21 @@ mod tests {
         store.ingest(SuggestIngestionConstraints::all_providers());
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::fakespot("simp")),
-            vec![simpsons_suggestion()],
+            vec![simpsons_suggestion(
+                "The <b>Simpsons</b>: Skinner's Sense of Snow (DVD)"
+            )],
```
bendk, line 2200: One interesting aspect of the feature is that it highlights the entire term, not just the part that was matched.  So here it highlights "Simpsons", even though the user input was "simp".

0c0w3:
My first thought is, it seems like stemming is the problem. "Maps" shouldn't match "maple" (per that Slack thread). It seems like people agree on that. I don't think stemming makes sense for this suggestion type, and maybe not at all for Suggest or any type of similar autocomplete? Can we just remove `porter` from the schema?

Are there examples where we're currently returning a Fakespot suggestion that we think we should be returning but desktop isn't properly highlighting it?

Aside from that, the desktop front-end already does highlighting, so I'm kind of reluctant to introduce an alternate code path for that.

bendk:
Disabling stemming seems very reasonable to me and if we did that, then I don't think we would need this code.  That said, some stemming feels like it's useful for some queries to me (i.e. "shoes" matching "shoe").  Maybe we can discuss this in the meeting today.

bendk:
Discussed this on the call and the decision was to not implement this.  We couldn't choose a clear leader between the several options available: not stemming, stemming, stemming with updated highlighting.  Because of that, we'll just go with what we have and not try to make changes late in the game.

PR #6327, opened 2024-08-02
Switch relevancy component from UDL to proc-macros [ci full]

gruberb:
Deleting the `relevancy.udl` file and use proc-macros to export directly from the Rust codebase.


bendk (commented):
Looks right to me.  Let's keep this in a branch for a while though, because Desktop doesn't support proc-macros yet.  I'm working on a PR that I hope to merge in a couple weeks.

PR #6328, opened 2024-08-02
Remote settings caching support

bendk:
Continued from #6295.  This is an attempt to expose the caching API to consumers.

Added a builder API for constructing RemoteSettings instances.  Gave the builder a couple ways to enable caching support:
  - Simple method: specify a cache path
  - Advance method: implement the RemoteSettingsCache callback interface

If a cache is supplied then:
  - get_records() will use the cache for efficient network requests
  - Consumers can use the `get_cached_records` and `sync_cached_records` for extra control over when we hit the network.  This is useful for settings where typically only want to get the setting from cache and not hit the network.

  - Note:

### DIFF
--- a/components/remote_settings/src/cache.rs
+++ b/components/remote_settings/src/cache.rs
@@ -1,0 +1,259 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::{
+    collections::HashSet,
+    fs::{File, OpenOptions},
+    io::prelude::*,
+};
+
+use parking_lot::Mutex;
+
+use crate::{RemoteSettingsResponse, Result};
+
+/// Interface for caching remote settings data
+///
+/// This is implemented by the remote settings consumer, sometimes with foreign code via UniFFI.
+///
+/// This interface has no error handling.  Consumers should report any errors themselves.  If there
+/// is an error in `get()` simply return `None` and the remote settings client will re-fetch any
+/// data.
+pub trait RemoteSettingsCache: Send + Sync {
+    /// Update the cached value
+    fn store(&self, value: RemoteSettingsResponse);
+
+    /// Get the last value passed to `set`.
+    fn get(&self) -> Option<RemoteSettingsResponse>;
+}
+
+/// Merge a cached RemoteSettingsResponse and a newly downloaded one to get a merged reesponse
+///
+/// cached is a previously downloaded remote settings response (possibly run through merge_cache_and_response).
+/// new is a newly downloaded remote settings response (with `_expected` set to the last_modified
+/// time of the cached response).
+///
+/// This will merge the records from both responses, handle deletions/tombstones, and return a
+/// response that has:
+///   - The newest `last_modified_date`
+///   - A record list containing the newest version of all live records.  Deleted records will not
+///     be present in this list.
+///
+/// If everything is working properly, the returned value will exactly match what the server would
+/// have returned if there was no `_expected` param.
+pub fn merge_cache_and_response(
+    cached: RemoteSettingsResponse,
+    new: RemoteSettingsResponse,
+) -> RemoteSettingsResponse {
+    let new_record_ids = new
+        .records
+        .iter()
+        .map(|r| r.id.as_str())
+        .collect::<HashSet<&str>>();
+    // Start with any cached records that don't appear in new.
+    let mut records = cached
+        .records
+        .into_iter()
+        .filter(|r| !new_record_ids.contains(r.id.as_str()))
+        // deleted should always be false, check it just in case
+        .filter(|r| !r.deleted)
+        .collect::<Vec<_>>();
+    // Add all (non-deleted) records from new
+    records.extend(new.records.into_iter().filter(|r| !r.deleted));
+
+    RemoteSettingsResponse {
+        last_modified: new.last_modified,
+        records,
+    }
+}
+
+/// Implements RemoteSettingsCache by serializing data and writing it to a file
+pub struct RemoteSettingsCacheFile {
+    file: Mutex<File>,
+}
+
+impl RemoteSettingsCacheFile {
+    pub fn new(path: String) -> Result<Self> {
+        let file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .truncate(false)
+            .open(path)?;
+        Ok(Self {
+            file: Mutex::new(file),
+        })
+    }
+
+    fn try_store(&self, value: RemoteSettingsResponse) -> Result<()> {
+        let mut file = self.file.lock();
+        file.rewind()?;
+        serde_json::to_writer(&mut *file, &value)?;
+        Ok(())
+    }
+
+    fn try_get(&self) -> Result<Option<RemoteSettingsResponse>> {
+        let mut file = self.file.lock();
+        if file.metadata()?.len() == 0 {
+            return Ok(None);
+        }
+        file.rewind()?;
+        Ok(serde_json::from_reader(&mut *file)?)
+    }
+}
+
+impl RemoteSettingsCache for RemoteSettingsCacheFile {
+    fn store(&self, value: RemoteSettingsResponse) {
+        if let Err(e) = self.try_store(value) {
+            log::warn!("Error writing remote settings cache: {e}");
+        }
+    }
+
+    fn get(&self) -> Option<RemoteSettingsResponse> {
+        match self.try_get() {
+            Ok(r) => r,
+            Err(e) => {
+                log::warn!("Error reading remote settings cache: {e}");
+                None
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{RemoteSettingsRecord, RsJsonObject};
+
+    // Quick way to generate the fields data for our mock records
+    fn fields(data: &str) -> RsJsonObject {
+        let mut map = serde_json::Map::new();
+        map.insert("data".into(), data.into());
+        map
+    }
+
+    #[test]
+    fn test_combine_cache_and_response() {
+        let cached_response = RemoteSettingsResponse {
+            last_modified: 1000,
+            records: vec![
+                RemoteSettingsRecord {
+                    id: "a".into(),
+                    last_modified: 100,
+                    deleted: false,
+                    attachment: None,
+                    fields: fields("a"),
+                },
+                RemoteSettingsRecord {
+                    id: "b".into(),
+                    last_modified: 200,
+                    deleted: false,
+                    attachment: None,
+                    fields: fields("b"),
+                },
+                RemoteSettingsRecord {
+                    id: "c".into(),
+                    last_modified: 300,
+                    deleted: false,
+                    attachment: None,
+                    fields: fields("c"),
+                },
+            ],
+        };
+        let new_response = RemoteSettingsResponse {
+            last_modified: 2000,
+            records: vec![
+                // d is new
+                RemoteSettingsRecord {
+                    id: "d".into(),
+                    last_modified: 1300,
+                    deleted: false,
+                    attachment: None,
+                    fields: fields("d"),
+                },
+                // b was deleted
+                RemoteSettingsRecord {
+                    id: "b".into(),
+                    last_modified: 1200,
+                    deleted: true,
+                    attachment: None,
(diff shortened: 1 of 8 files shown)


components/remote_settings/src/client.rs
```diff
@@ -48,14 +59,59 @@ impl Client {
...
+            }
+        };
+        if cached_last_modified != response.last_modified {
+            cache.store(response.clone());
+        }
+        Ok(response)
+    }
+
+    pub fn get_cached_records(&self) -> Option<RemoteSettingsResponse> {
+        self.cache.as_deref().and_then(|cache| cache.get())
```
mhammond, line 101: None here seems mildly odd - at face value I'd have thought calling this (and below) with self.cache is a logic error of some kind, and subtly different from "cache exists but does not have it".

Or maybe not! I'm fine with this if you think it really is fine and I can see how some parts of the code really don't care about the distinction, but thought I'd throw it out anyway.

mhammond (approved):
This seems fine to me

components/remote_settings/src/client.rs
bendk: I can go either way as well.  The thing I like about this is that it presents a simpler API for consumers who do specify a cache (not throwing at all is simpler than telling users that throwing won't happen in practice).

components/remote_settings/src/client.rs
```diff
@@ -48,14 +59,59 @@ impl Client {
...
+    }
+
+    pub fn sync_cached_records(&self) -> Result<()> {
```
gruberb, line 104: I think a `if let` would be more idiomatic, and even improve readability a bit:

```rust
pub fn sync_cached_records(&self) -> Result<()> {
    if let Some(cache) = self.cache.as_deref() {
        // Use get_records_with_cache to update the cache, then throw out the result.
        // This avoids sending the response over the FFI in cases where the consumer
        // doesn't need it.
        self.get_records_with_cache(cache)?;
    }
    Ok(())
}
```

components/remote_settings/src/client.rs
```diff
@@ -48,14 +59,59 @@ impl Client {
...
+        &self,
+        cache: &dyn RemoteSettingsCache,
+    ) -> Result<RemoteSettingsResponse> {
+        let cached_response = cache.get();
+        let cached_last_modified = cached_response
+            .as_ref()
+            .map(|r| r.last_modified)
+            .unwrap_or(0);
+
+        let response = match cached_response {
```
gruberb, line 87: I would also do a `if let` here, but maybe it's just my way of finding it easier to read in this case.

```rust
    let response = if let Some(cached) = cached_response {
        let new = self.get_records_since(cached.last_modified)?;
        cache::merge_cache_and_response(cached, new)
    } else {
        self.get_records()?
    };
```

components/remote_settings/src/lib.rs
```diff
@@ -50,6 +60,80 @@ impl RemoteSettings {
...
+        }
+    }
+}
+
+impl RemoteSettingsBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn collection_name(self: Arc<Self>, collection_name: String) -> Arc<Self> {
```
gruberb, line 88: I wonder why we return `Arc` instead of `Self` directly?

Typically in a Builder Pattern, one might expect a `Self`.

```rust
  pub fn collection_name(mut self, collection_name: String) -> Result<Self, RemoteSettingsError> {
        let mut inner = self.inner.lock().map_err(|_| RemoteSettingsError::LockError)?;
        inner.collection_name = Some(collection_name);
        Ok(self)
    }
```

But not sure how devs might expect to use it, so just throwing it out there.

bendk: It's a UniFFI thing.  Right now the only way to express this is an an interface and interfaces always need to be wrapped in an Arc, since that's how objects work in all the foreign languages.

I think we could avoid this if we supported methods on records/enums.  That would also remove the need for the inner mutex.

components/remote_settings/src/cache.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Interface for caching remote settings data
+///
+/// This is implemented by the remote settings consumer, sometimes with foreign code via UniFFI.
+///
+/// This interface has no error handling.  Consumers should report any errors themselves.  If there
+/// is an error in `get()` simply return `None` and the remote settings client will re-fetch any
+/// data.
+pub trait RemoteSettingsCache: Send + Sync {
+    /// Update the cached value
+    fn store(&self, value: RemoteSettingsResponse);
```
ncloudioj, line 24: Wouldn't it be useful to provide `clear()`? Two potential uses I can think of:
- To clear out the cache if the user disables the feature
- For cache busting, sometimes we want the component to ignore the cache and re-download everything

components/remote_settings/src/cache.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    /// Get the last value passed to `set`.
```
ncloudioj, line 26:
```suggestion
    /// Get the last value passed to `store`.
```

components/remote_settings/src/remote_settings.udl
```diff
@@ -69,7 +71,53 @@ interface RemoteSettings {
...
+    [Self=ByArc]
+    RemoteSettingsBuilder bucket_name(string bucket_name);
+
+    // Setup a cache that writes to a file.
+    // This is the simplest way to enable caching support.
+    [Self=ByArc, Throws=RemoteSettingsError]
+    RemoteSettingsBuilder cache_file(string path);
+
+    // Setup cache using a callback interface that implements RemoteSettingsCache
+    [Self=ByArc]
```
ncloudioj, line 111: Can you add a comment noting that `cache_file` and `cache` should be called exclusively? Otherwise they will override the previous cache setting.

components/remote_settings/src/remote_settings.udl
```diff
@@ -69,7 +71,53 @@ interface RemoteSettings {
...
+    RemoteSettingsBuilder server(RemoteSettingsServer server);
+
+    [Self=ByArc]
+    RemoteSettingsBuilder bucket_name(string bucket_name);
+
+    // Setup a cache that writes to a file.
+    // This is the simplest way to enable caching support.
+    [Self=ByArc, Throws=RemoteSettingsError]
+    RemoteSettingsBuilder cache_file(string path);
```
ncloudioj, line 108: Let's also note the consumers that it's up to them to ensure the path is valid and writable, otherwise, the `cache.store()` will fail and it doesn't support error handling at that point.

leplatrem (commented):
I would suggest to align the philosophy of this client with the one we have on desktop 🙏 

Instead of having a new method `get_cached_records()` etc., we would just have:
- `client.get(syncIfEmpty=false)` --> returns local cache, and calls sync() if local is empty
- `client.sync(timestamp)` --> refreshes the local cache (if timestamp is None or differs from local cache).

Clients that need default data when network is off and local cache empty would provide just hardcoded values (`.get().unwrap_or(...)`).

As we talked about elsewhere about this, eventually (and ideally) `sync()` won't be called by consumers, but by a background job that keeps local data in sync for every  instantiated clients (ie. on push notification, or scheduled). The general idea is to keep some control over the network activity, instead of letting every consumer code initiate ad-hoc requests to the server.

nbhasin2 (approved):

components/remote_settings/src/client.rs
nbhasin2, line 1: @bendk thanks for this, finally get back to the remote settings work so gave you an r+

let us know when this lands and I will add it to the Firefox iOS 

Q1. Is there a chance in future that full data models will be supplied instead of the clients (iOS / Android) having to parse the json for the record? (I doubt but asking anyways) 

Q2. When I am getting a record is there a way to know the cache's timestamp? I know we have `since` but this is a different ask

bendk:
> I would suggest to align the philosophy of this client with the one we have on desktop 🙏
> 
> Instead of having a new method `get_cached_records()` etc., we would just have:
> 
>     * `client.get(syncIfEmpty=false)` --> returns local cache, and calls sync() if local is empty
> 
>     * `client.sync(timestamp)` --> refreshes the local cache (if timestamp is None or differs from local cache).
> 
> 
> Clients that need default data when network is off and local cache empty would provide just hardcoded values (`.get().unwrap_or(...)`).

+1 to this plan.

What do you think about having the clients specify default data in the constructor?  That way `get()` wouldn't need to return an Option?

components/remote_settings/src/client.rs
bendk: Q1.  It could be possible, that seems pretty compatible with @leplatrem's suggestion of having the client control more things.  My one concern would be how do we update the schema definitions?

Q2. That should be possible, but can I understand the use-case better.

How well would Matt's proposed API work for you?  I like the idea of creating an API that feels like Desktop's, but I want to make sure it's going to work for consumers.

nbhasin2: @bendk The below api would work for us, just a few clarifications. We do send raw json files for use case c) 

Q. for a) what will happen if there is no local cache and no internet, is there a timeout or straight up return nothing? I wonder if its easier to simply return nothing and revert to option c) and that we can do on the client side while A~S fetches and write cache for future? 

Q. in terms of b) is this something that we maintain as in to refresh local cache?

I guess you stumbled upon the slack thread so you have an idea on what I am going for in terms of cache. 

```
    a) client.get(syncIfEmpty=false) --> returns local cache, and calls sync() if local is empty
    b) client.sync(timestamp) --> refreshes the local cache (if timestamp is None or differs from local cache).
    c) Clients that need default data when network is off and local cache empty would provide just hardcoded values (.get().unwrap_or(...)).
```

bendk: We're kind of mixing this discussion over slack and GH.  I'll answer here too for visibility.

>Q. for a) what will happen if there is no local cache and no internet, is there a timeout or straight up return nothing? I wonder if its easier to simply return nothing and revert to option c) and that we can do on the client side while A~S fetches and write cache for future?

Assuming `syncIfEmpty=false` then there should be no timeout.  I like your idea of combining this with option C.  In fact, I'd rather that's how the A-S client handled it:
  - When you construct a client, we load the set of defaults.  There's a few ways that could happen:
     - Constructor parameter / builder method.
     - Application-services stores a recent dump of remote settings data in our repo.
     - Maybe we'll define a global service that stores things like which server to use and have that be the thing that creates the clients.  If we did that then it could also be the place where default data is stored.
  - When you call `get`, if there's no cached data, we fallback to returning the default data.

bendk: > Q. in terms of b) is this something that we maintain as in to refresh local cache?

I think it would be on you to periodically call sync.  I'm not sure if you need to store the timestamp yourself, maybe we could handle that internally.

This is also another reason why it could be good to define the global service.  It seems better to call sync on that and update all collections at once.  If you need to sync each client and two clients use the same collection, then maybe we'll end up syncing the same collection twice.

Standard8:
> What do you think about having the clients specify default data in the constructor? That way `get()` wouldn't need to return an Option?

In the draft specs document, @leplatrem was suggesting t to move the dumps into the component here, so that they could be read directly. This might be easier done once a-s moves into m-c though. However, this would avoid the potentially expensive operation of passing long strings across the uniffi interface.

leplatrem:
> What do you think about having the clients specify default data in the constructor? That way get() wouldn't need to return an Option?

That would work for me, but as Mark said we have to clarify how we combine this with packaged dumps.
For example, if a collection has a packaged dump, then it will be its default data. 
If it doesn't, the default data specified during client build is used.
If no default data is provided during client build, what do we return?

I suggest the Option, because in desktop we made the "mistake" to return an empty list by default. With no way for the consumer to distinguish between an empty list (eg. no ongoing experiment), a fresh profile, a faulty sync, etc.

bendk:
> That would work for me, but as Mark said we have to clarify how we combine this with packaged dumps. For example, if a collection has a packaged dump, then it will be its default data. If it doesn't, the default data specified during client build is used. If no default data is provided during client build, what do we return?
> 
> I suggest the Option, because in desktop we made the "mistake" to return an empty list by default. With no way for the consumer to distinguish between an empty list (eg. no ongoing experiment), a fresh profile, a faulty sync, etc.

You've convinced me that an option is better.  I was hoping that we could guarantee that there would be fallback data, but it seems like we can't.  It's means the clients will have to add some extra comments explaining how they know a value isn't null and maybe add some `!!` expressions, but that doesn't seem so bad.

bendk:
I'm going to close this one since we want to update this API to bring it closer to the one used by the official client and this code is now pretty stale.

https://github.com/mozilla/application-services/pull/6378 contains a proposal for the new API, let's continue the discussion there.

PR #6334, opened 2024-08-05
Bug 1908802 - Handle deleted suggest records

bendk:
The issue is that once a record is deleted, it no longer has a type field.  Reworked the ingestion client to deal with this.  It now always fetches all records, then filters them client side.  Client-side filtering is more flexible and allows us to handle the deleted records. Updated the code to use caching make the requests more efficient.

Rather than store last modified times for record types, store details on each record that we've ingested.  This means that if we fail to process a record for any reason, on the next ingest we have another chance to process it and recover.

Updated the `rs::Record` type to store the parsed `SuggestRecord`.  This simplifies a lot of the filtering/caching code.

  - Note:

### DIFF
--- a/components/nimbus/src/tests/stateful/test_evaluator.rs
+++ b/components/nimbus/src/tests/stateful/test_evaluator.rs
@@ -155,7 +155,7 @@ fn test_event_transform_sum_cnz_avg_avgnz_parameters() {
             &th,
         ),
         Some(EnrollmentStatus::Error {
-            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1`, expected a string"
+            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1.0`, expected a string"
                 .to_string()
         })
     );
@@ -165,7 +165,7 @@ fn test_event_transform_sum_cnz_avg_avgnz_parameters() {
             &th,
         ),
         Some(EnrollmentStatus::Error {
-            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1`, expected a string"
+            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1.0`, expected a string"
                 .to_string()
         })
     );
@@ -231,7 +231,7 @@ fn test_event_transform_last_seen_parameters() {
             &th,
         ),
         Some(EnrollmentStatus::Error {
-            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1`, expected a string"
+            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1.0`, expected a string"
                 .to_string()
         })
     );
@@ -241,7 +241,7 @@ fn test_event_transform_last_seen_parameters() {
             &th,
         ),
         Some(EnrollmentStatus::Error {
-            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1`, expected a string"
+            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1.0`, expected a string"
                 .to_string()
         })
     );
--- a/components/suggest/src/rs.rs
+++ b/components/suggest/src/rs.rs
@@ -31,12 +31,12 @@
 //!     the new suggestion in their results, and return `Suggestion::T` variants
 //!     as needed.
 
-use std::{borrow::Cow, fmt};
+use std::fmt;
 
-use remote_settings::{Attachment, GetItemsOptions, RemoteSettingsRecord, RsJsonObject, SortOrder};
+use remote_settings::{Attachment, RemoteSettingsRecord};
 use serde::{Deserialize, Deserializer};
 
-use crate::{error::Error, provider::SuggestionProvider, Result};
+use crate::{db::SuggestDao, error::Error, provider::SuggestionProvider, Result};
 
 /// A list of default record types to download if nothing is specified.
 /// This defaults to all record types available as-of Fx128.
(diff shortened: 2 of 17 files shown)


bendk:
I tested this using the dev remote settings server and the suggest CLI and it seemed to work to me.  I tested the following cases, using the `-v` output to verify that the correct records were processed:
- Ingesting everything from scratch
- Ingesting some providers, then others
- Ingesting after a record was deleted
- Ingesting after a record was updated

components/suggest/src/rs.rs
```diff
@@ -93,90 +103,85 @@ impl RemoteSettingsClient {
...
-                let attachment_data = record
-                    .attachment
-                    .as_ref()
-                    .map(|a| client.get_attachment(&a.location))
-                    .transpose()?;
-                Ok(Record::new(record, attachment_data))
-            })
-            .collect()
+    fn get_records(&self, db: &SuggestDb, fetch_fakespot_records: bool) -> Result<Vec<Record>> {
+        let records = get_records_for_collection(&self.quicksuggest_client, db, "quicksuggest")?;
```
ncloudioj, line 116: Non-issue; just out of curiosity (your code is absolutely fine :P). Since iterators are lazy, does the following do the same (assuming `get_records_for_collection` always succeeds)?

```rs
        let records = get_records_for_collection(..., "quicksuggest")?;
        let fsrecords = get_records_for_collection(..., "fakespot-suggest-products")?;

        Ok(if fetch_fakespot_records {
            records.chain(fsrecords).collect()
        } else {
            records.collect()
        })
```

components/suggest/src/rs.rs
```diff
@@ -93,90 +103,85 @@ impl RemoteSettingsClient {
...
-        }
-        options
+fn get_records_for_collection(
+    client: &remote_settings::Client,
+    db: &SuggestDb,
+    collection_name: &'static str,
+) -> Result<impl Iterator<Item = Record>> {
+    // For now, handle the cache manually.  Once 6328 is merged, we should be able to delegate
+    // this to remote_settings.
+    let cache = db.read(|dao| dao.read_cached_rs_data(collection_name))?;
```
ncloudioj, line 141: As noted earlier, loading the cache over here feels a bit strange to me. Imagine if we want to handle cache failures later, that would leak more cache handling details into this module. Maybe we can move that to another place and and just consumer the cache data here.

ncloudioj (commented):
It's a big patch and I just reviewed a few modules. Added some early feedback to kick off the discussion.

components/suggest/src/schema.rs
```diff
@@ -28,6 +28,17 @@ CREATE TABLE meta(
     value NOT NULL
 ) WITHOUT ROWID;
 
+CREATE TABLE rs_cache(
+    collection TEXT PRIMARY KEY,
+    data TEXT NOT NULL
+) WITHOUT ROWID;
+
+CREATE TABLE ingested_records(
```
ncloudioj, line 36: Should we add a field for `collection` here? I am a little concerned that things will mess up if we pull `Fakespot` out of the enums of `SuggestRecord` and `SuggestRecordType` in the future. If you disagree, please add a comment here to note that.

components/suggest/src/store.rs
```diff
@@ -389,124 +384,115 @@ where
             DEFAULT_RECORDS_TYPES.to_vec()
         };
 
+        let fetch_fakespot_records = ingest_record_types
+            .iter()
+            .any(|r| *r == SuggestRecordType::Fakespot);
+        let records = self
+            .settings_client
+            .get_records(writer, fetch_fakespot_records)?;
```
ncloudioj, line 392: What's your thought on cache corruption handling? Let's say if the db is corrupted and `get_records_for_collection()` always fails and returns early from `let cache = db.read(|dao| dao.read_cached_rs_data(collection_name));`. It will never try to fetch records from RS due to the absence of the cache, correct?

components/suggest/src/rs.rs
bendk: Good question!  I *think* this would do some work, since `get_records_for_collection` makes network connections, etc.  It's not like the `async fn` -> `Future` translation which doesn't do anything until the future is first polled.

But even simpler than either of these options is to go with your suggestion to add a collection argument and not special-case fakespot so much.

components/suggest/src/rs.rs
```diff
@@ -58,8 +58,18 @@ pub(crate) const DEFAULT_RECORDS_TYPES: [SuggestRecordType; 9] = [
...
+    /// Get all records from the server
+    ///
+    /// We use this plus client-side filtering rather than any server-side filtering, as
+    /// recommended by the remote settings docs
+    /// (https://remote-settings.readthedocs.io/en/stable/client-specifications.html). This is
+    /// relatively inexpensive since we use a cache and don't fetch attachments until after the
+    /// client-side filtering.
+    ///
+    /// Records that can't be parsed as [SuggestRecord] are ignored.
+    fn get_records(&self, db: &SuggestDb, fetch_fakespot_records: bool) -> Result<Vec<Record>>;
```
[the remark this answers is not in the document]
bendk: I tried out letting the consumer specify the collection and not special-casing fakespot in general and I like the new codebetter.

It's not so easy to remove the dependency on a DB (now a `SuggestDao`), since it also sometimes writes to the DB.  It's definitely possible, but I'm kind of hoping to avoid it since all of this code is just a stop-gap until #6328 is implemented.  Once that happens, then the remote settings client will handle caching internally and we can drop the argument.

components/suggest/src/store.rs
bendk: We should definitely be handling this differently.  I changed it so the main methods never fail, they just log warnings and return `None`.  I think that's the best way to handle things, it was definitely nice that ingestion continued to work when I switched the format from JSON to msgpack.

components/suggest/src/schema.rs
```diff
@@ -28,6 +28,17 @@ CREATE TABLE meta(
     value NOT NULL
 ) WITHOUT ROWID;
 
+CREATE TABLE rs_cache(
+    collection TEXT PRIMARY KEY,
+    data TEXT NOT NULL
```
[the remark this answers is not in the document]
bendk: Storing record id and mtime seems interesting, I'll consider that for #6328.  I switched the serialization format to msgpack, which seems decent enough.  I tried bincode and postcard, but neither of them can handle the `serde_json::Value` field.  I think that's because it can contain heterogeneous lists.

ncloudioj (approved):
Looks great. Thanks!

components/nimbus/src/tests/stateful/test_evaluator.rs
```diff
@@ -241,7 +241,7 @@ fn test_event_transform_last_seen_parameters() {
             &th,
         ),
         Some(EnrollmentStatus::Error {
-            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1`, expected a string"
+            reason: "EvaluationError: Custom error: JSON Error: invalid type: floating point `1.0`, expected a string"
```
bendk, line 244: @jeddai Looks like the fallout was just the error strings that `serde` sent out.  Any concerns with these test changes?

jeddai: Nope!

PR #6341, opened 2024-08-08
Switch to UniFFI library mode [ci full]

bendk:
I'm reviving #5592 and giving it another shot.  If successful, this can be the basis for @gruberb's proc-macro work.

  - Note:


bendk:
In that PR, Tarik said "Looks like this ended up being a breaking change for firefox-ios because of a swift name conflict".  I completely forget what that means, but can someone test this code with a local ios build and see if it works?

gruberb (approved):
I ran the PR locally against an firefox-ios and it worked. So I approve, but not sure if we want another pair of eyes.

bendk:
I tested a local Fenix build and it worked for me.  @gruberb was able to get iOS to successfully compile.  This seems like good enough evidence to me, but please tell me if we should test more.

@jeddai do you think we need to do any extra testing for cirrus?

components/autofill/uniffi.toml
```diff
@@ -1,6 +1,5 @@
 [bindings.kotlin]
 package_name = "mozilla.appservices.autofill"
-cdylib_name = "megazord"
```
gruberb, line 3: I am curious: Why can we remove the `cdylib_name`?

bendk: That's one perk of library mode, since we're passing the library path to the bindings generator we can calculate the cdylib name from that.

docs/howtos/adding-a-new-component.md
```diff
@@ -162,8 +159,6 @@ The result should look something like this:
 
 Click on the top-level "MozillaTestServices" project in the navigator, then go to "Build Phases".
 
-> Double-check that `<your_crate_name>.udl` does **not** appear in the "Copy Bundle Resources" section.
-
 Add `<your_crate_name>.udl` to the list of "Compile Sources". This will trigger an Xcode Build Rule that generates
```
bendk, line 162: I tried to update this as best I could, but I was unsure of some Swift things.
- Do my current changes make sense?
- What should we do with the above line?  In a proc-macro world, we won't have a UDL file to add to compile sources.  Do we need to add some other file?  Or is this not needed anymore?

linabutler: I think your current changes make sense!

For the above line...I'm not sure just yet! How do we invoke `uniffi-bindgen` to generate the Swift bindings, and what does it take for the input, in the proc-macro world?

The Build Rule that we have now takes all `.*udl` files as inputs, and runs `uniffi-bindgen generate` on each one; but we can change it to whatever it needs!

bendk: The new command is `cargo uniffi-bindgen generate --library [path-to-megazord-ios-a] -l swift -o [out-dir]`.  The input is the megazord library, which is awkward to specify in the compile sources.  Can we add a build step that unconditionally runs that after the megazord is built?  Or is my change to `build-xcframework.sh` already enough?

megazords/ios-rust/build-xcframework.sh
```diff
@@ -139,37 +139,22 @@ cp "$WORKING_DIR/DEPENDENCIES.md" "$COMMON/DEPENDENCIES.md"
 
 mkdir -p "$COMMON/Headers"
 
-# First we move the files that are common between both
-# firefox-ios and Focus
+# Library to generate the UniFFI bindings with.  We use an arbitrary target, since that doesn't
```
linabutler, line 142: "Arbitrary target" here means that it doesn't matter whether we use `aarch64-apple-ios`, `aarch64-apple-ios-sim`, or `x86_64-apple-ios` (because we build all three, and the metadata for all three should be identical), right?

publish.gradle
```diff
@@ -185,25 +185,47 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
...
+            // Re-generate when the megazord is rebuilt, which should happen if any components
+            // change
+            it.inputs.files megazordPathCandidates
             // Re-generate if our uniffi-bindgen tooling changes.
             inputs.dir "${project.rootDir}/tools/embedded-uniffi-bindgen/"
             // Re-generate if our uniffi-bindgen version changes.
             inputs.file "${project.rootDir}/Cargo.lock"
         }
         variant.registerJavaGeneratingTask(t.get(), new File(buildDir, uniffiGeneratedPath))
+        tasks[taskName].dependsOn(":full-megazord:assembleRelease")
```
linabutler, line 228: `t.get()` (and `tasks[taskName]`) will force evaluation of the task graph—which can have unintended consequences down the line—and there's new lazy configuration APIs that we can use here, for both Gradle and the Android Gradle plug-in. Not at all a blocker for this PR, since we were already doing this, but I can get a follow-up bug on file for that!

linabutler (approved):
This looks terrific, @bendk! :shipit:

megazords/ios-rust/build-xcframework.sh
bendk: Yes that's exactly right.  In theory, with proc-macros, you could have different definitions depending on the target, but we don't do that yet.

PR #6343, opened 2024-08-14
Bug 1893086 - Port the potential impressions framework to the suggest component

0c0w3:
This is a redo of #6211 and has some substantial changes from that, but overall it's very similar. It implements a new type of suggestion called exposure suggestions. In #6211 they were called phantom suggestions, but I think "exposure" is a better name.

In summary, exposure suggestions can be matched on a subtype in addition to being matched on keyword. (The subtype is called `exposure_suggestion_type` throughout this commit.) Exposure suggestions in remote settings define a subtype, and suggest consumers pass in a subtype at query time to match suggestions only of that subtype. No code changes are required to support new subtypes. Exposure suggestions are intended to be used with telemetry implemented in the consumer to quickly and easily test potential exposures of hypothetical suggestion types.

Differences between this commit and #6211:

* As mentioned, rename "phantom" to "exposure"
* Implement a keyword-matching strategy instead of leaving it as a to-do (more on this below)
* Add `exposure_suggestion_type` to ingest constraints so consumers don't have to ingest all exposure suggestions, only the subtype they're interested in. This is parallel to how consumers pass in `exposure_suggestion_type` at query time. The way this works is, RS records for exposure suggestions are expected to have a `exposure_suggestion_type` inline field, and we pass the type in the RS request.
* Don't include a `matched_keyword` in the suggestion returned to the consumer. I found we don't actually need it.

### Keyword-matching strategy

The strategy is informed by the first two use cases for potential exposures on desktop, translation and VPN suggestions. In both cases, we have a large list of keywords and prefixes of those keywords. Prefix keywords don't necessarily start with the first word, unlike AMO, Pocket, and other suggestion types that use the `prefix_keywords` table. So at query time, we don't know how to break down a query into a prefix and a rest-of-the-query in order to do the lookup.

This commit expects all keywords for an exposure suggestion, including prefix keywords, to be defined in the remote settings data, and it ingests them into the `keywords` table. At query time, it does a simple lookup into that table, like how AMP suggestions work. It doesn't keep track of "full keywords" like AMP does (in the `full_keywords` table) since they don't matter here. For exposure suggestions, keywords are just keywords, prefix or not.

The format of the keywords in the RS data is a little novel. The keywords list can contain full keyword strings for keywords that don't have prefixes. It can also contain tuples, which are used as a compact way to represent a starting prefix keyword and all the other keywords associated with that prefix. There's a code comment in rs.rs above `FullOrPrefixKeywords` that explains, and `test_exposure_keywords` in that file has examples.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -13,7 +13,7 @@ use rusqlite::{
     types::{FromSql, ToSql},
     Connection, OpenFlags, OptionalExtension,
 };
-use sql_support::{open_database::open_database_with_flags, ConnExt};
+use sql_support::{open_database::open_database_with_flags, repeat_sql_vars, ConnExt};
 
 use crate::{
     config::{SuggestGlobalConfig, SuggestProviderConfig},
@@ -24,8 +24,9 @@ use crate::{
     provider::SuggestionProvider,
     rs::{
         DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedAmpWikipediaSuggestion,
-        DownloadedFakespotSuggestion, DownloadedMdnSuggestion, DownloadedPocketSuggestion,
-        DownloadedWeatherData, DownloadedWikipediaSuggestion, Record, SuggestRecordId,
+        DownloadedExposureSuggestion, DownloadedFakespotSuggestion, DownloadedMdnSuggestion,
+        DownloadedPocketSuggestion, DownloadedWeatherData, DownloadedWikipediaSuggestion, Record,
+        SuggestRecordId,
     },
     schema::{clear_database, SuggestConnectionInitializer},
     suggestion::{cook_raw_suggestion_url, AmpSuggestionType, Suggestion},
@@ -805,6 +806,67 @@ impl<'a> SuggestDao<'a> {
         )
     }
 
+    /// Fetches exposure suggestions
+    pub fn fetch_exposure_suggestions(&self, query: &SuggestionQuery) -> Result<Vec<Suggestion>> {
+        // A single exposure suggestion can be spread across multiple remote
+        // settings records, for example if it has very many keywords. On ingest
+        // we will insert one row in `exposure_custom_details` and one row in
+        // `suggestions` per record, but that's only an implementation detail.
+        // Logically, and for consumers, there's only ever at most one exposure
+        // suggestion with a given exposure suggestion type.
+        //
+        // Why do insertions this way? It's how other suggestions work, and it
+        // lets us perform relational operations on suggestions, records, and
+        // keywords. For example, when a record is deleted we can look up its ID
+        // in `suggestions`, join the keywords table on the suggestion ID, and
+        // delete the keywords that were added by that record.
+
+        let Some(suggestion_types) = query
+            .provider_constraints
+            .as_ref()
+            .and_then(|c| c.exposure_suggestion_types.as_ref())
+        else {
+            return Ok(vec![]);
+        };
+
+        let keyword = query.keyword.to_lowercase();
+        let params = rusqlite::params_from_iter(
+            std::iter::once(&SuggestionProvider::Exposure as &dyn ToSql)
+                .chain(std::iter::once(&keyword as &dyn ToSql))
+                .chain(suggestion_types.iter().map(|t| t as &dyn ToSql)),
+        );
+        self.conn.query_rows_and_then_cached(
+            &format!(
+                r#"
+                    SELECT DISTINCT
+                      d.type
+                    FROM
+                      suggestions s
+                    JOIN
+                      exposure_custom_details d
+                      ON d.suggestion_id = s.id
+                    JOIN
+                      keywords k
+                      ON k.suggestion_id = s.id
+                    WHERE
+                      s.provider = ?
+                      AND k.keyword = ?
+                      AND d.type IN ({})
+                    ORDER BY
+                      d.type
+                    "#,
+                repeat_sql_vars(suggestion_types.len())
+            ),
+            params,
+            |row| -> Result<Suggestion> {
+                Ok(Suggestion::Exposure {
+                    suggestion_type: row.get("type")?,
+                    score: 1.0,
+                })
+            },
+        )
+    }
+
     /// Inserts all suggestions from a downloaded AMO attachment into
     /// the database.
     pub fn insert_amo_suggestions(
@@ -1055,6 +1117,39 @@ impl<'a> SuggestDao<'a> {
             SuggestionProvider::Weather,
             &SuggestProviderConfig::from(data),
         )?;
+        Ok(())
+    }
+
+    /// Inserts exposure suggestion records data into the database.
+    pub fn insert_exposure_suggestions(
+        &mut self,
+        record_id: &SuggestRecordId,
+        suggestion_type: &str,
+        suggestions: &[DownloadedExposureSuggestion],
+    ) -> Result<()> {
+        // `suggestion.keywords()` can yield duplicates for exposure
+        // suggestions, so ignore failures on insert in the uniqueness
+        // constraint on `(suggestion_id, keyword)`.
+        let mut keyword_insert = KeywordInsertStatement::new_with_or_ignore(self.conn)?;
+        let mut suggestion_insert = SuggestionInsertStatement::new(self.conn)?;
+        let mut exposure_insert = ExposureInsertStatement::new(self.conn)?;
+        for suggestion in suggestions {
+            self.scope.err_if_interrupted()?;
+            let suggestion_id = suggestion_insert.execute(
+                record_id,
+                "", // title, not used by exposure suggestions
+                "", // url, not used by exposure suggestions
+                DEFAULT_SUGGESTION_SCORE,
+                SuggestionProvider::Exposure,
+            )?;
+            exposure_insert.execute(suggestion_id, suggestion_type)?;
+
+            // Exposure suggestions don't use `rank` but `(suggestion_id, rank)`
+            // must be unique since there's an index on that tuple.
+            for (rank, keyword) in suggestion.keywords().enumerate() {
+                keyword_insert.execute(suggestion_id, &keyword, None, rank)?;
+            }
+        }
         Ok(())
     }
 
@@ -1498,12 +1593,47 @@ impl<'conn> FakespotInsertStatement<'conn> {
     }
 }
 
+struct ExposureInsertStatement<'conn>(rusqlite::Statement<'conn>);
+
+impl<'conn> ExposureInsertStatement<'conn> {
+    fn new(conn: &'conn Connection) -> Result<Self> {
+        Ok(Self(conn.prepare(
+            "INSERT INTO exposure_custom_details(
+                 suggestion_id,
+                 type
+             )
+             VALUES(?, ?)
+             ",
+        )?))
+    }
+
+    fn execute(&mut self, suggestion_id: i64, suggestion_type: &str) -> Result<()> {
+        self.0
+            .execute((suggestion_id, suggestion_type))
+            .with_context("exposure insert")?;
+        Ok(())
+    }
+}
+
 struct KeywordInsertStatement<'conn>(rusqlite::Statement<'conn>);
 
 impl<'conn> KeywordInsertStatement<'conn> {
     fn new(conn: &'conn Connection) -> Result<Self> {
         Ok(Self(conn.prepare(
             "INSERT INTO keywords(
+                 suggestion_id,
+                 keyword,
+                 full_keyword_id,
+                 rank
+             )
+             VALUES(?, ?, ?, ?)
+             ",
+        )?))
+    }
+
+    fn new_with_or_ignore(conn: &'conn Connection) -> Result<Self> {
+        Ok(Self(conn.prepare(
+            "INSERT OR IGNORE INTO keywords(
                  suggestion_id,
                  keyword,
                  full_keyword_id,
--- a/components/suggest/src/provider.rs
+++ b/components/suggest/src/provider.rs
@@ -12,6 +12,17 @@ use rusqlite::{
 
 use crate::rs::SuggestRecordType;
 
+/// Record types from these providers will be ingested when consumers do not
+/// specify providers in `SuggestIngestionConstraints`.
+pub(crate) const DEFAULT_INGEST_PROVIDERS: [SuggestionProvider; 6] = [
+    SuggestionProvider::Amp,
+    SuggestionProvider::Wikipedia,
+    SuggestionProvider::Amo,
+    SuggestionProvider::Yelp,
+    SuggestionProvider::Mdn,
+    SuggestionProvider::AmpMobile,
+];
+
 /// A provider is a source of search suggestions.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 #[repr(u8)]
@@ -25,6 +36,7 @@ pub enum SuggestionProvider {
     Weather = 7,
     AmpMobile = 8,
     Fakespot = 9,
+    Exposure = 10,
 }
 
 impl fmt::Display for SuggestionProvider {
@@ -39,6 +51,7 @@ impl fmt::Display for SuggestionProvider {
             Self::Weather => write!(f, "weather"),
             Self::AmpMobile => write!(f, "ampmobile"),
             Self::Fakespot => write!(f, "fakespot"),
+            Self::Exposure => write!(f, "exposure"),
         }
     }
 }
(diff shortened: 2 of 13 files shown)


components/suggest/src/suggestion.rs
```diff
@@ -175,6 +166,20 @@ impl Suggestion {
...
+        match self {
+            Self::Amp { score, .. }
+            | Self::Pocket { score, .. }
+            | Self::Amo { score, .. }
+            | Self::Yelp { score, .. }
+            | Self::Mdn { score, .. }
+            | Self::Weather { score, .. }
+            | Self::Fakespot { score, .. }
+            | Self::Exposure { score, .. } => *score,
+            Self::Wikipedia { .. } => DEFAULT_SUGGESTION_SCORE,
```
bendk, line 180: I really like how this doesn't have a default branch.  It makes it a lot more difficult to forget to update it when we add a new suggestion.

components/suggest/src/rs.rs
```diff
@@ -520,6 +530,81 @@ pub(crate) struct DownloadedFakespotSuggestion {
...
+///    "sugg"
+///    "sugge"
+///    "sugges"
+///    "suggest"
+///    "suga"
+///    "sugar"
+///    "sugarp"
+///    "sugarpl"
+///    "sugarplu"
+///    "sugarplum"
```
bendk, line 571: I wonder if this should be using FTS search rather than explicit keywords, since we might be using FTS more in the future.  I really don't know though, what's your thoughts?

components/suggest/src/rs.rs
```diff
@@ -141,6 +141,11 @@ impl From<RecordRequest> for GetItemsOptions {
         if let Some(last_modified) = value.last_modified {
             options.filter_gt("last_modified", last_modified.to_string());
         }
+
+        if let Some(exposure_suggestion_type) = value.exposure_suggestion_type {
+            options.filter_eq("exposure_suggestion_type", exposure_suggestion_type);
```
bendk, line 146: I think this is going to conflict with some of my recent PR -- sorry about that.  Also, the general advice we've gotten from the RS team is to not use any filters when querying the data and do all filtering client-side.

components/suggest/src/suggest.udl
```diff
@@ -125,6 +133,9 @@ dictionary SuggestIngestionConstraints {
     // there is a schema upgrade that causes the database to be thrown away, then the
     // `empty_only=true` ingestion that runs on startup will repopulate it.
     boolean empty_only = false;
+    // If `providers` contains `Exposure`, this must be set to an exposure
+    // suggestion type to ingest.
+    string? exposure_suggestion_type;
```
bendk, line 138: I don't love the way this field is coupled with the exposure type (this also goes for `SuggestionQuery`).  I'd rather one of these solutions:

 - Changing `SuggestionProvider` from a simple enum to one with fields.  The only variant that would have a field is `SuggestionProvider::Exposure` and it would get a `suggestion_type` field.
 - Creating a separate provider for each exposure type (`SuggestionProvider::ExposureFakespot`, `SuggestionProvider::ExposureAmp`, etc).

I think this would work better with the newly merged code.  Also, it would let consumers query for multiple exposure types at once.

bendk (changes requested):
I like this idea a lot, but I have some questions/concerns with how this `exposure_suggestion_type` field is going to work -- especially with the recent move to client-side filtering of the remote settings data.

components/suggest/src/rs.rs
0c0w3: We need to have pretty tight control over keywords and matching, mostly due to legal reasons. Every time we run a potential exposure experiment, we have to clear the keywords with legal. There may be a little wiggle room for fuzzy matching, but things like stemming and arbitrary prefix matching can alter the set of "actual" keywords too dramatically. Other than that legal reason, there's also a lot of value in having all experiment stakeholders on the same page w/r/t the keywords being tested, and it becomes harder to explain to PMs, QA, and external teams (who come up with the keywords) what exactly we're testing if the matching is too fuzzy.

That said, I do think there's probably a nice middle ground that uses FTS, and ideally we wouldn't require external teams to come up with huge lists of keyword phrases, many of which are variations on each other. I'd like to start with the simpler strategy in this commit and think about FTS for a follow-up.

components/suggest/src/suggest.udl
0c0w3: Yeah, I don't like it much either. The second solution won't work because the whole point of this is to not require any code changes to run new exposure experiments. I like your first solution so I'll try that.

0c0w3: My latest commit converts `SuggestionProvider` from a unit-only enum to a non-unit-only enum. There weren't as many resulting changes as I thought there would need to be, which I guess means it's not a very consequential change. I do think it makes it a little harder to understand what "provider" means. To me, each use of a `SuggestionProvider` is now more like an "instance" than a simple static const value, but that might just be my Rust newbie-ness showing.

components/suggest/src/provider.rs
```diff
@@ -12,19 +12,30 @@ use rusqlite::{
...
+/// Record types from these providers will be ingested when consumers do not
+/// specify providers in `SuggestIngestionConstraints`.
+pub(crate) const DEFAULT_INGEST_PROVIDERS: [SuggestionProvider; 6] = [
+    SuggestionProvider::Amp,
+    SuggestionProvider::Wikipedia,
+    SuggestionProvider::Amo,
+    SuggestionProvider::Yelp,
+    SuggestionProvider::Mdn,
+    SuggestionProvider::AmpMobile,
+];
```
0c0w3, line 24: IMO consumers should always have to pass in providers they want to ingest. I don't think it makes sense for desktop to ingest AMP mobile by default, or vice versa. But that's orthogonal to this PR and here I'm just replacing `DEFAULT_RECORDS_TYPES` with `DEFAULT_INGEST_PROVIDERS`. It would be nice remove the concept of default ingest providers/record types in a follow-up.

components/suggest/src/suggest.udl
bendk: IMO this is the main thing we need to get right.  I don't think this is your "Rust newbie-ness", the current patch changes the meaning of "provider" and the result is confusing.  For example, the `suggestions` table has a provider field which is an int, which is no longer really correct.  Maybe we could make this work by treating providers like records by having both a `SuggestionProvider` and `SuggestionProviderType` types, but I'm not sure I like that.

Would it be possible to implement this by adding a `SuggestionProvider::Exposure` variant without the `suggestion_type` field?  The suggestion type is already stored an a per-suggestion basis so this should be doable:
  - Whenever a query includes `SuggestionProvider::Exposure` we'd return all exposure suggestions that matched, and the client could filter them based on the `Suggestion::Exposure::type` field and the enabled experiment features.   
 - When ingesting, we'd ingest all exposure suggestions together if `SuggestionProvider::Exposure` was specified.  That might be slightly wasteful, but would it be too bad?

0c0w3: I don't really like the idea of ingesting and fetching all exposure types. With translations suggestions for example, we're going to have en-US, fr, and de versions. It's not a good design to have any user enrolled in a translations experiment download all localizations.

Tbh I was thinking simply renaming `Provider` to something like `ProviderConfig`, `ProviderOptions`, or `ProviderInstance` might make it feel right.

The other thought I had was introducing a separate parallel `ProviderOptions` enum, and then you would pass them on query and ingest. But I don't know how to force consumers at compile-time to pass correct options given the providers they also passed. It would have to be a runtime error if the options don't match the providers, or we could just silently ignore it I guess.

0c0w3: > The other thought I had was introducing a separate parallel `ProviderOptions` enum, and then you would pass them on query and ingest. 

This would be a generalization of my first approach in this PR. Instead of exposure-specific fields in `SuggestionQuery` and `SuggestIngestionConstraints`, we'd have a generalized `provider_options` field -- a `sequence` I guess. Is there a map type in udl/uniffi? If the options could be a map from `SuggestionProvider` (as a plain `enum` variant) to `SuggestionProviderOptions`, that would be nice. But I don't know how to statically force consumers to use the `SuggestionProviderOptions` that matches the `SuggestionProvider`.

I do wonder whether exposure suggestions will always be the one special case though. If so, the simplest thing is my first approach IMO. Maybe we could go with that for now and cross the generalized bridge if/when we get to it?

bendk: > The other thought I had was introducing a separate parallel ProviderOptions enum, and then you would pass them on query and ingest.

I like this idea.

It's possible to create a map in UDL using `record<KeyType, ValueType>`.  "record" is kind of counter-intuitive here, but it's inherited form IDL.  The keys used to have to be string, but I think you can now use enum values nowadays.

However, I think I prefer using a struct and having each suggestion type ignore any fields that didn't apply to them.  Initially, it could just have a single  `exposure_suggestion_types` field.  Later on, if we added a `locale` field, that couple apply to exposure suggestions, AMP, and any other suggestion type that had localization support.

0c0w3: That sounds good to me. I've come around to thinking we should go back to the first approach or something like it to keep things simple given the use case and not over-engineer or over-generalize. A simple provider options struct seems like a good compromise between the first approach and a more generalized approach. We can always re-evaluate if more providers end up needing ingest- or query-time options.

0c0w3: I called it `SuggestionProviderConstraints` since it's analogous to ingest constraints and I didn't want to cause confusion with the existing `SuggestProviderConfig`, which is something else entirely. Let me know what you think.

components/suggest/src/db.rs
```diff
@@ -805,6 +806,62 @@ impl<'a> SuggestDao<'a> {
...
+        // in `suggestions`, join the keywords table on the suggestion ID, and
+        // delete the keywords that were added by that record.
+
+        let mut suggestions = Vec::new();
+        if let Some(suggestion_types) = query
+            .provider_constraints
+            .as_ref()
+            .and_then(|c| c.exposure_suggestion_types.as_ref())
+        {
+            for suggestion_type in suggestion_types {
```
bendk, line 830: How many types to you expect to be enabled at once?  If it's more than a couple, maybe we could optimize this to select all the types in one query.

components/suggest/src/db.rs
```diff
@@ -1058,6 +1115,51 @@ impl<'a> SuggestDao<'a> {
...
+        // suggestions, so ignore failures on insert in the uniqueness
+        // constraint on `(suggestion_id, keyword)`.
+        let mut keyword_insert = KeywordInsertStatement::new_with_or_ignore(self.conn)?;
+        let mut suggestion_insert = SuggestionInsertStatement::new(self.conn)?;
+        for suggestion in suggestions {
+            self.scope.err_if_interrupted()?;
+            let suggestion_id = suggestion_insert.execute(
+                record_id,
+                "",
+                "",
```
bendk, line 1135: Can you add a note explaining the blank strings here?

components/suggest/src/db.rs
```diff
@@ -1058,6 +1115,51 @@ impl<'a> SuggestDao<'a> {
...
+                 )
+                 VALUES(
+                     :suggestion_id,
+                     :type
+                 )",
+                named_params! {
+                    ":suggestion_id": suggestion_id,
+                    ":type": suggestion_type,
+                },
+            )?;
```
bendk, line 1152: Please add a  `.with_context` call before checking the error type.  That way if we get SQL errors, we'll see which statement caused them.  Secondarily, I think it would faster to define an `ExposureInsertStatement` type like the others.  I think `execute_cached` does a map lookup each iteration.  This is much less important though.

bendk (approved):
This looks good to me, I just had some minor suggestions.

@ncloudioj how do you feel about the addition of `SuggestionProviderConstraints`?  Drew and I went back and forth about how to design this, does it make sense to you?

components/suggest/src/db.rs
0c0w3: Only one at a time actually, but I don't think there's a good reason to enforce that in the API since supporting more than one is easy, and who knows, we might want to do more than one at a time someday. Thanks for pointing this out, it should be done in one query.

0c0w3: @bendk Could you give my changes to this method (`fetch_exposure_suggestions`) another look please? It's changed substantially and I want to make sure it's OK.

ncloudioj:
> @ncloudioj how do you feel about the addition of SuggestionProviderConstraints? Drew and I went back and forth about how to design this, does it make sense to you?

I think it looks good for now, we can give it a run and revisit if more providers want to use `ProviderConstraints` in the future.

I think the main concern of keeping `providers` and `providers_constraints` separate is that it'd be hard to enforce/validate query parameters as the consumer can initiate them with arbitrary values. IMO the root cause of that is b/c all `dao.fetch_*(query: &SuggestionQuery)` take the whole struct rather than a provider specific struct for contexts/options.

On a separate note, @0c0w3 so I was thinking about alternative matching strategies for cases like exposure suggestions. Especially, those probabilistic solutions such as Bloom filters. They're much more efficient to deliver/store/query than the keyword-matching-based counterparts. The only downside is that it comes with false positives, but I guess instrumentation errors (still can be controlled) are totally acceptable in this case since there is little user impact, correct? The benefits are really compelling.

As an example, let's say we have an exposure suggestion type with 100k keywords and the average keyword length is 10 bytes. Then we would  send a 1MB payload to the client and ingest it and store that many keywords in the db. Whereas if we load those keywords into a Bloom filter with 1% false positive ratio, each keyword would only take approx. 10 bits (would be even smaller as the error ratio increases) regardless of its length. That's 8x smaller than the raw payload. Also, only one record needs to be stored (deleted later) in the db as we can serialize/deserialize the filter to facilitate delivery. Once deserialized, the query runs at the constant time. (we'd keep them in memory, so we'd be mindful about the number of BFs)   

This perhaps belongs to another thread, but would like to hear your early thoughts here.

components/suggest/src/db.rs
bendk: These changes look good to me and more importantly the tests in `store.rs` look very complete which is great.

0c0w3:
> I think the main concern of keeping `providers` and `providers_constraints` separate is that it'd be hard to enforce/validate query parameters as the consumer can initiate them with arbitrary values. IMO the root cause of that is b/c all `dao.fetch_*(query: &SuggestionQuery)` take the whole struct rather than a provider specific struct for contexts/options.

I had that concern too. To your second point (about the root cause), in fact [an earlier commit](https://github.com/mozilla/application-services/pull/6343/commits/fbacbee8d07033cd2567afa5cedc158aca1aef51) in this PR changed `SuggestionProvider` from a unit-only enum to a non-unit-only enum where the `Exposure` variant carried the extra data (`suggestion_type`) in a field. We decided that made it harder to understand what a "provider" is, so we ditched it.

We could add provider-specific options/constraints structs (separate from `SuggestionProvider`), but IMO we should cross that bridge if/when we come to it and not over-generalize now. I kinda think that exposure suggestions will end up being the only provider that needs query- and ingest-time options. Ben mentioned locales as another possibility for a constraint (e.g., ingest/query only en-US suggestions for a given provider) and I agree we'll want locale support at some point, but locales in particular seem like an option that applies to all providers, not something that belongs in provider-specific structs, but maybe!

> On a separate note, @0c0w3 so I was thinking about alternative matching strategies for cases like exposure suggestions. Especially, those probabilistic solutions such as Bloom filters. They're much more efficient to deliver/store/query than the keyword-matching-based counterparts. The only downside is that it comes with false positives, but I guess instrumentation errors (still can be controlled) are totally acceptable in this case since there is little user impact, correct? The benefits are really compelling.

I'd definitely like to explore better matching strategies, not only for exposure suggestions but for Suggest in general. It's way past time for us to move on from simplistic keyword matching. FTS with Fakespot moves us down that road a little but there's still plenty to go.

For exposures in particular, there are some legal constraints around what we're allowed to test and there's a lot of value in having a simple matching strategy so that all experiment stakeholders are on the same page -- see [this comment](https://github.com/mozilla/application-services/pull/6343#discussion_r1717568749) if you're interested.

ncloudioj:
> We could add provider-specific options/constraints structs (separate from SuggestionProvider), but IMO we should cross that bridge if/when we come to it and not over-generalize now. I kinda think that exposure suggestions will end up being the only provider that needs query- and ingest-time options.

Makes sense. We can revisit this if needed in the future.

> For exposures in particular, there are some legal constraints around what we're allowed to test and there's a lot of value in having a simple matching strategy so that all experiment stakeholders are on the same page -- see https://github.com/mozilla/application-services/pull/6343#discussion_r1717568749 if you're interested.

Ah, I see. Some non-functional constraints are at play here. Thanks for the context!

0c0w3:
As we discussed on Slack, I'll force-push a squashed commit and merge that since we went through so many revisions and I don't want to touch blame for lines that ended up not changing in the end. (Mostly commenting about this since there's a real possibility I mess it up lol)

0c0w3:
Ahh it blew away all the previous commits linked from this PR, which makes sense, but I was hoping the PR would keep them archived or something so we could refer back to them and not orphan the review comments. That's too bad. I still have them locally though.

PR #6346, opened 2024-08-19
Cherry-pick changes for Desktop 1.0

bendk:
I created the release-v130-desktop branch from 8fd08c6f2f8acd38579bd3142fecda9272957b72 (the current application-services rev in Desktop v130).  I then cherry-picked all the fakespot commits on top of that.  I'm hoping that this can be the code we vendor/uplift into Desktop 130.

Normally, I'd try to avoid bringing in this many commits however they basically only touch the Fakespot code, which is currently behind a feature flag.  This means that we're relying on manual QA testing to verify this code rather than nightly users.  It also means that we can't break the feature for anyone since no one is using it yet.

The only non-fakespot commits were https://github.com/mozilla/application-services/commit/ab675ffbefa4f356296a424e6c61b3c033f04bc0, which was pretty trivial, and https://github.com/mozilla/application-services/commit/61839befb2811dbb3d2f04cf096a2a6099c83ef4 which only touched the benchmark code.

I didn't include c3774b262 since it depends on bac73bb44.  I feel like pulling in both of those commits to the uplift is too risky -- especially since we have a workaround for the deleted records issue (make sure you update records rather than delete them and create new ones).  However, I could definitely be convinced to bring those in as well.

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Suggest
 - Removed the deprecated `remote_settings_config` method.  No consumers were using this.
+- Removed the `SuggestIngestionConstraints::max_suggestions` field.  No consumers were using this.
 
 ## ✨ What's New ✨
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1424,6 +1424,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "clap 4.2.2",
+ "env_logger",
  "log",
  "remote_settings",
  "suggest",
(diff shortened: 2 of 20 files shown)


bendk:
After our call, I realized that the current 130 desktop commit was missing many changes from the previous cycle.   That means this PR is bigger than expected, but like I said in the description I don't think this is so bad since it only touches the Fakespot code.

bendk:
The only conflict I saw was trivial.  It came from [this comment](https://github.com/mozilla/application-services/commit/016192965#diff-6e354ed7e89505584e48bdf26e00b4f7b61403795fa37d0a181dbc32311f4896R344-R347), which needs to be in a different place without https://github.com/mozilla/application-services/commit/bac73bb442f82a93fa1bfb46dc08da840b52e842.  I just deleted the comment.

ncloudioj (approved):
I've manually checked the picked commits against the 130-release branch, looks good.

PR #6348, opened 2024-08-21
Update to Glean v61.0.0

badboy:
@bendk is coordinating the update.


bendk (approved):
Changes look good to me.  I also have https://github.com/mozilla/application-services/pull/6326 for upgrading UniFFI in app-services and we have some patches in moz-central.  What's the right order of operations?  Can we merge this PR, then https://github.com/mozilla/application-services/pull/6326, then do the moz-central vendor?

badboy:
This can land as is. a-s is not depending on the Rust parts, so there's no duplicated UniFFI.
If #6326 then lands before the next nightly it can land on m-c (you might need to do some additional stuff for the uniffi upgrade?)

bendk:
> a-s gets nightlies and that auto-opens a bug against m-c.

This is the Android nightly right?  I don't think there's something like this for Desktop.  Does the Android nightly bump update what's vendored in?

In any case, I'll rebase my PR on top of this one and try to get a desktop patch that vendors in that commit.  Once those are all approved, I think we should be able to land everything together.

badboy:
> > a-s gets nightlies and that auto-opens a bug against m-c.
> 
> This is the Android nightly right? I don't think there's something like this for Desktop. Does the Android nightly bump update what's vendored in?

Yes, Android. Everything ends up in the same repo so we need that a-s nightly in Android + manual updates for the Desktop parts.

> In any case, I'll rebase my PR on top of this one and try to get a desktop patch that vendors in that commit. Once those are all approved, I think we should be able to land everything together.

PR #6349, opened 2024-08-21
Bug 1914145 - Create initial search component stub.

Standard8:
This adds a skeleton search component and mechanics, which we will build on in subsequent PRs. I don't think it needs a changelog entry, as this isn't really adding functionality, but can add one if needed.

  - Note:


components/search/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Documentation for `pub` symbols is written with application developers in mind: even if you're a Desktop, Android or iOS developer, the Rust documentation is meant to give you an understanding of how the Gecko, Kotlin and Swift bindings work.
+
+You can see the documentation for all public symbols by running from the command line:
+
+```shell
+cargo doc --open
+```
+
+By convention, symbols that are exported as part of the component's foreign language interface have `pub` visibility, and symbols that are only used within the component have `pub(crate)` or private visibility. As an exception to this convention, symbols with [doctests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) are also `pub`, because doctests [can only link against public symbols](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests).
```
linabutler, line 33: I'm feeling a little on the fence about this convention. I experimented with it in Suggest, but not sure if the awkwardness of typing `pub(crate)` vs. `pub` each time is worth it. I think @bendk had some good thoughts about it also!

When I was experimenting with this convention, I thought that having Rust docs would be better than nothing for application developers. Even if the names and syntax didn't quite match up, at least the `cargo doc` output could be somewhat helpful for explaining a component's core concepts and abstractions to Desktop, Android, and iOS devs.

I was worried that using `pub` would clutter up the `cargo doc` output with Rust-only symbols, which are super helpful if you're working on the component itself, but not so much if you just want to consume it.

I think I overestimated how many Android / iOS devs would be open to installing Rust and running `cargo doc`, though 😅 I get the feeling that our consumers (completely understandably!) just want to see docs in Android Studio and Xcode, and "run `cargo doc` then figure out how to translate its output to the Kotlin / Swift API you're using" might not be super interesting to them.

components/search/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you're working on the component, you can see the documentation for `pub(crate)` and private symbols using:
+
+```shell
+cargo doc --document-private-items --open
+```
+
+💡 If you're adding a new suggestion type, the documentation for the `rs` module is a great place to start.
```
linabutler, line 41:
```suggestion
```

components/search/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+This component follows the architecture of the other [Application Services Rust components](https://mozilla.github.io/application-services/book/index.html): a cross-platform Rust core, and platform-specific bindings for Firefox Desktop, Android, and iOS. These bindings are generated automatically using the [UniFFI](https://mozilla.github.io/uniffi-rs/) tool.
+
+### For contributors
+
+This section is a primer for engineers contributing code to the Search Rust component.
+
+`search.udl` describes the component's interface for foreign language consumers. UniFFI uses this file to generate the language bindings for each platform.
+
+`store.rs` contains the implementation of `SuggestStore` and most of the Suggest component's tests.
```
linabutler, line 19:
```suggestion
```

😉

components/search/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```shell
+cargo test -p search
+```
+
+## ⚠️ Breaking Changes ⚠️
+
+A "breaking change" is any code change that breaks the build, tests, or behavior of a consuming application.
+
+These are some common changes that can break consumers:
```
linabutler, line 57: I wonder if we should nix this list here (and in the Suggest readme), and expand on it in our ["breaking changes"](https://mozilla.github.io/application-services/book/howtos/breaking-changes.html) page?

I think I remember a message from @0c0w3 (please forgive me if I'm misremembering!) that this list was extremely confusing, and I can definitely see that rereading it a few months later!

components/search/src/error.rs
```diff
@@ -0,0 +1,31 @@
+/// The error type for all Search component operations. These errors are
```
linabutler, line 1:
```suggestion
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

/// The error type for all Search component operations. These errors are
```

components/search/src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+/// A list of errors that are internal to the component. This is the error
+/// type for private and crate-internal methods, and is never returned to the
+/// application.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Error {
+    #[error("Not Implemented")]
+    NotImplemented(),
+}
+
+// #[non_exhaustive]
```
linabutler, line 14: It does look like we can have `[NonExhaustive]` [in UDL](https://github.com/mozilla/uniffi-rs/pull/1924) now!

components/search/src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+/// exposed to your application, which should handle them as needed.
+use error_support::{ErrorHandling, GetErrorHandling};
+
+/// A list of errors that are internal to the component. This is the error
+/// type for private and crate-internal methods, and is never returned to the
+/// application.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Error {
+    #[error("Not Implemented")]
+    NotImplemented(),
```
linabutler, line 11:
```suggestion
    NotImplemented,
```

components/search/src/selector.rs
```diff
@@ -0,0 +1,29 @@
...
+ */
+use crate::{error::Error, SearchApiResult};
+use error_support::handle_error;
+
+pub struct SearchEngineSelector();
+
+impl SearchEngineSelector {
+    #[handle_error(Error)]
+    pub fn filter_engine_configuration(&self) -> SearchApiResult<()> {
+        Err(Error::NotImplemented())
```
linabutler, line 13:
```suggestion
        Err(Error::NotImplemented)
```

linabutler (approved):
This initial skeleton looks great to me, @Standard8! 🎉

We can add the component to the Android and iOS Megazords once it's ready, but I think this is great to land as-is, if the Rust component is the direction you and your team decide on. (A little selfishly, though, it's exciting to see you and your team's interest in Rust, and we'd love to support you however we can! 😁)

One question, more for @bendk and @gruberb: would you recommend @Standard8 start with UniFFI's proc macros (I think you're still working on the Gecko JS backend's support for them?), or is UDL still the better way to go for now?

components/search/README.md
bendk: I'm not sure if this qualifies as a good thought, but another way of controlling visibility is to make everything `pub` if it should be visible outside the module, and use `pub use` vs `use`  in the top-level module to distinguish what gets exposed to consumer crates.  I kind of like that model, since there's one place that controls what's exposed.  This has worked out okay for me, but I'm not totally sold on it.  I keep feeling like I'll find a corner case that I hadn't though out.

bendk:
> One question, more for @bendk and @gruberb: would you recommend @Standard8 start with UniFFI's proc macros (I think you're still working on the Gecko JS backend's support for them?), or is UDL still the better way to go for now?

The proc-macros are great, but I believe that https://github.com/mozilla/application-services/pull/6341/ needs to be merged first.  I hope to do that next week or the first week in September.  I guess this decision depends on how fast you need to get this code in app-services.

components/search/README.md
Standard8: Thanks, I'll think I'll drop that text for now, and go with the `pub use` option for the time being.

components/search/README.md
Standard8: I've dropped it from here for now. I'll let you manage the other page, since I think you understand more about it.

components/search/src/error.rs
Standard8: I think that only works with Enums at the moment.

Standard8:
> > One question, more for @bendk and @gruberb: would you recommend @Standard8 start with UniFFI's proc macros (I think you're still working on the Gecko JS backend's support for them?), or is UDL still the better way to go for now?
> 
> The proc-macros are great, but I believe that #6341 needs to be merged first. I hope to do that next week or the first week in September. I guess this decision depends on how fast you need to get this code in app-services.

Looks like it isn't landed yet. I'm going to go ahead and land this, so that I can start playing around with the next stages. Hopefully it won't be too much work to switch afterwards.

components/search/src/error.rs
linabutler: Oops, I didn't look carefully enough! 😬

linabutler (approved):
Still LGTM! :shipit:

PR #6351, opened 2024-08-28
Bug 1915299 - Download attachments of unchanged exposure suggestion r…

0c0w3:
…ecords if the suggestion type hasn't been ingested

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -867,6 +867,22 @@ impl<'a> SuggestDao<'a> {
         )
     }
 
+    pub fn is_exposure_suggestion_ingested(&self, record_id: &SuggestRecordId) -> Result<bool> {
+        Ok(self.conn.exists(
+            r#"
+            SELECT
+              id
+            FROM
+              suggestions
+            WHERE
+              record_id = :record_id
+            "#,
+            named_params! {
+                ":record_id": record_id.as_str(),
+            },
+        )?)
+    }
+
     /// Inserts all suggestions from a downloaded AMO attachment into
     /// the database.
     pub fn insert_amo_suggestions(
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -503,9 +503,15 @@ where
             // Suggestions in particular don't have a stable identifier, and
             // determining which suggestions in the record actually changed is
             // more complicated than dropping and re-ingesting all of them.
-            log::trace!("Reingesting record ID: {}", record.id.as_str());
+            log::trace!("Reingesting updated record ID: {}", record.id.as_str());
             dao.delete_record_data(&record.id)?;
             self.ingest_record(dao, record, constraints, download_timer)?;
+        }
+        for record in &changes.unchanged {
+            if self.should_reingest_record(dao, record)? {
+                log::trace!("Reingesting unchanged record ID: {}", record.id.as_str());
+                self.ingest_record(dao, record, constraints, download_timer)?;
+            }
         }
         for record in &changes.deleted {
             log::trace!("Deleting record ID: {:?}", record.id);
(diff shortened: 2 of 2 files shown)


components/suggest/src/db.rs
```diff
@@ -867,6 +867,23 @@ impl<'a> SuggestDao<'a> {
...
+    pub fn is_exposure_suggestion_type_ingested(&self, suggestion_type: &str) -> Result<bool> {
+        let type_exists = self.conn.exists(
+            r#"
+            SELECT
+              type
+            FROM
+              exposure_custom_details
+            WHERE
+              type = :type
+            "#,
```
bendk, line 879: I'm not sure this will handle all the corner cases.  The main one is that if an exposure type is not enabled, I think this will always be true and we'll always try to process the record.  Maybe that's not a big deal though, since we'll quickly skip over it in `ingest_record()`.  I worry that there could be other, more tricky corner cases though that end up affecting the user.  For example:
- Fx version 150 ingests `exposure-record-foo-100`
- Fx version 151 disables `foo` and therefore skips `exposure-record-foo-200`
- Fx version 152 re-enables `foo`, but then doesn't ingest `exposure-record-foo-200` since it's unchanged and we have an suggestions from `exposure-record-foo-100`.

I'm not sure how realistic that one is, but I'm a bit worried that something like that might happen.  What if you added a field to `ingest_records` that tracks if it matched the constraints when we last processed it?  If that's false but the record now does match the constraints, then we process it as a new record.

bendk (changes requested):
My only concern is the DB check, which may fail in some corner cases.  Everything else looks great.

components/suggest/src/db.rs
0c0w3: > The main one is that if an exposure type is not enabled, I think this will always be true and we'll always try to process the record. Maybe that's not a big deal though, since we'll quickly skip over it in `ingest_record()`.

Right, that's my opinion since otherwise `should_reingest_record()` would need to duplicate the logic inside `ingest_record()` that determines whether the attachment should be ingested. It would need to take on part of the role that `ingest_record()` already does.

I think the names of these methods might be misleading in this context. I almost want to say `process_record()` or `handle_record()` instead of `ingest_record()`. Ingest encompasses processing, downloading, and updating the DB. If `should_reingest_record()` returns true, it doesn't mean all those things need to happen, just that the record should be re-processed and then maybe they should happen.

> I'm not sure how realistic that one is, but I'm a bit worried that something like that might happen. What if you added a field to `ingest_records` that tracks if it matched the constraints when we last processed it?

I had a similar thought but was hoping we could get away without something like that. :-) My thought was, we could avoid those problems either by updating the record on RS or by using a new exposure suggestion type altogether. That would require some manual work on our part, which we might forget to do. I think for the use case we have in mind for exposure suggestions -- running one-off experiments -- we're fine, but you're probably right it should be addressed in the code, so I'll take look.

0c0w3: This turned out to be pretty simple to fix: If there's no row in `suggestions` with the `record_id` of the unchanged record, then its attachment was not ingested, and so the record should be re-processed. (And then if the ingest provider constraints match the record's exposure type, `ingest_record()` should go on to ingest its attachment.) I don't think I'm overlooking anything?

components/suggest/src/store.rs
```diff
@@ -667,13 +673,21 @@ where
             Err(_) => Ok(()),
         }
     }
+
+    fn should_reingest_record(&self, dao: &mut SuggestDao, record: &Record) -> Result<bool> {
+        match &record.payload {
+            SuggestRecord::Exposure(_) => Ok(!dao.is_exposure_suggestion_ingested(&record.id)?),
```
bendk, line 679: Can you add a comment explaining that this can happen when the constraints change?

bendk (approved):
Changes look good to me, it's nice that it turned out to be so easy.

components/suggest/src/db.rs
0c0w3: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1915715

PR #6354, opened 2024-08-29
Generate static website docs from swift components

gruberb:
This PR is a first step in building static-website docs for the Swift components produced by `application-services`. 

It adds a new folder inside `automation`. Inside, we build a new Swift project, with all the `.swift` and `.h` files from the output of `uniffi-bindgen`. We have to manipulate a few things to make it work, and provide a fixed `module.modulemap` and `Package.swift`.

In the end, we use `jazzy` to generate the documentation.

## Feedback

I decided to move this to this repo instead of `rust-components-swift`, because it feels closer to the source, and we really just need the output of `uniffi-bindgen` to build this. No `Xcframework` or `Xcodeproject` needed.

## Outcome (local)
Screenshot 2024-08-29 at 09 26 25

### DIFF
--- a/.github/workflows/swift-docs.yaml
+++ b/.github/workflows/swift-docs.yaml
@@ -1,0 +1,49 @@
+name: Generate Swift Docs
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-and-deploy:
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout repository with submodules
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install Rust
+        run: |
+          RUSTUP_PLATFORM=x86_64-apple-darwin
+          RUSTUP_VERSION=1.27.0
+          curl -sfSL --retry 5 --retry-delay 10 -O "https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/${RUSTUP_PLATFORM}/rustup-init"
+          chmod +x rustup-init
+          ./rustup-init -y --no-modify-path
+          rm rustup-init
+          echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $GITHUB_ENV
+          source $HOME/.cargo/env
+
+      - name: Set up Ruby environment
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.3'
+
+      - name: Install Jazzy
+        run: gem install jazzy
+
+      - name: Run build script
+        run: |
+          cd ./automation/swift-components-docs
+          ./build.sh
+
+      - name: Deploy
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout --orphan swift-docs
+          git --work-tree automation/swift-components-docs/docs add --all
+          git --work-tree automation/swift-components-docs/docs commit -m "Deploy Swift docs"
+          git push origin HEAD:swift-docs --force
--- a/automation/swift-components-docs/Cargo.toml
+++ b/automation/swift-components-docs/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "swift_components_docs"
+edition = "2021"
+version = "0.1.0"
+authors = ["Sync Team <sync-team@mozilla.com>"]
+license = "MPL-2.0"
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+megazord_ios = { path = "../../megazords/ios-rust" }
--- a/megazords/ios-rust/Cargo.toml
+++ b/megazords/ios-rust/Cargo.toml
@@ -6,7 +6,7 @@ authors = ["Sync Team <sync-team@mozilla.com>"]
 license = "MPL-2.0"
 
 [lib]
-crate-type = ["staticlib"]
+crate-type = ["staticlib", "rlib"]
 
 [dependencies]
 rust-log-forwarder = { path = "../../components/support/rust-log-forwarder" }
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,11 @@ cirrus.py
 fml.py
 *.dylib
 *.so
+
+# Build static website files
+automation/swift-components-docs/.build
+automation/swift-components-docs/docs
+automation/swift-components-docs/Sources/SwiftComponents/*.swift
+automation/swift-components-docs/Sources/SwiftComponents/include/*.h
+
+
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4160,6 +4160,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "swift_components_docs"
+version = "0.1.0"
+dependencies = [
+ "megazord_ios",
+]
+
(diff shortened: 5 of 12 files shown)


mhammond:
We briefly discussed UDL yesterday, so it's worth pointing out that the fxa-client crate has [extensive docs in UDL](https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/fxa_client.udl) which we should make sure are rendered. I believe some edits will be needed though as their is a mix of `///` and `//` comments there and IIRC, only triple slash docs are extracted.

automation/swift-components-docs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+autofill = { path = "../../components/autofill" }
+push = { path = "../../components/push" }
+tabs = { path = "../../components/tabs" }
+places = {path = "../../components/places" }
+remote_settings = { path = "../../components/remote_settings" }
+suggest = { path = "../../components/suggest" }
+sync15 = {path = "../../components/sync15"}
+error-support = { path = "../../components/support/error" }
+sync_manager = { path = "../../components/sync_manager" }
+as-ohttp-client = { path = "../../components/as-ohttp-client" }
```
bendk, line 28: Could you use a dependency on `megazord_ios` instead of this list?  That way the crate list will always be kept in sync.

automation/swift-components-docs/Sources/MozillaRustComponents/include/module.modulemap
```diff
@@ -0,0 +1,20 @@
...
+    header "fxa_clientFFI.h"
+    header "loginsFFI.h"
+    header "nimbusFFI.h"
+    header "placesFFI.h"
+    header "pushFFI.h"
+    header "remote_settingsFFI.h"
+    header "rustlogforwarderFFI.h"
+    header "sync15FFI.h"
+    header "syncmanagerFFI.h"
+    header "tabsFFI.h"
```
bendk, line 16: We really should create a way for UniFFI to help here.   Maybe if you use library mode and generate swift files, there could be an option for creating a single modulemap with everything in it.  In the meantime, I'm totally fine with hacking it with a hard-coded list of headers.

automation/swift-components-docs/replace-imports.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# To build a Swift project, we cannot import "MozillaRustComponents" from the modulemap.
+# For it to work, the module needs another name. This script replaces
+# import MozillaRustcomponents
+# with
+# import MyCMap
+# 'MyCMap` is the name of the module the `module.modulemap` is exporting.
```
bendk, line 8: I'm don't totally understand this one, but could this be avoided if UniFFI had support for generating a modulemap for all the components?

.github/workflows/swift-docs.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Set up Ruby environment
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.3'
+
+      - name: Install Jazzy
+        run: gem install jazzy
```
bendk, line 22: +1 to Jazzy.  If Glean is using it, then I love the choice.

bendk (commented):
Nice!  I guess the main thing is figuring out the CI.  Is it possible to commit the HTML to the `gh-pages` branch?  I think that's where we eventually want this to live.

If possible, it would be nice to make this a step in the GH actions in that branch.  I guess we need to enable the Ruby setup action though?

For now, it's not necessary to integrate it with the rest of the docs.  It would just be great to have it somewhere on the static site.  Later on we can re-organize those docs and start linking to the API reference.

gruberb:
> We briefly discussed UDL yesterday, so it's worth pointing out that the fxa-client crate has [extensive docs in UDL](https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/fxa_client.udl) which we should make sure are rendered. I believe some edits will be needed though as their is a mix of /// and // comments there and IIRC, only triple slash docs are extracted.

I didn't want to include an upgrade to `uniffi = "0.28.1`, because I think that's a different issue. I can though if we want. I quickly tested it locally, and yes, the doc comments (`///`) are extracted and shown in the output docs:

Screenshot 2024-08-29 at 12 29 39

We can open a PR to upgrade `uniffi` and adjust all the UDL comments to be `///` instead of `//`? Can also do it in this PR.

automation/swift-components-docs/replace-imports.sh
gruberb: Yeah that's a weird one. At this point, I surrender to the Swift gods. I tried what feels like a hundred different setups until `spm` finally included the header files.

The issue, I believe, is, that the project is named `MozillaRustComponents`, and the `.swift` files try to import `MozillaRustComponents` as well, because that's the name of the generated `module.modulemap` file. It works for `Xcframework`, but it didn't work for this setup (pure Swift). 

I can play around with the namings and see if we can leave the imports as is...

automation/swift-components-docs/Sources/MozillaRustComponents/include/module.modulemap
gruberb: I was "fine" with hardcoding it, because we currently have to hardcode any new component in the `generate-swift-project.sh` as well. But as soon as your PR lands with the library mode, we should auto-generate this `modulemap` file as well. I can put together a small script which does it as well as of now, so we don't hardcode anything here.

gruberb:
I requested to use `ruby/setup-ruby` for this repository:
https://bugzilla.mozilla.org/show_bug.cgi?id=1915656

gruberb:
@bendk I simplified the `Cargo.toml` a lot, and we directly use the `megazords/ios-rust` dependencies directly.

We have one last challenge: Where do we host this? Different options.

* We can use subfolders in the `docs` branch. Which means we have to integrate this workflow with the `generate-docs` workflow. We then create a subfolder there called `swift-docs`, which we then can link.
* We create a new repository, called `uniffi-docs`, where we also host the topic guides, and push it there. 

The problem with both is: We cannot host multiple GitHub pages domains for a single repository.

megazords/ios-rust/Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Sync Team <sync-team@mozilla.com>"]
 license = "MPL-2.0"
 
 [lib]
-crate-type = ["staticlib"]
+crate-type = ["staticlib", "rlib"]
```
gruberb, line 9: Had to add `rlib` so I can reference it inside `swift-components-docs`. This will increase the compile times.

bendk:
>   We can use subfolders in the `docs` branch. Which means we have to integrate this workflow with the `generate-docs` workflow. We then create a subfolder there called `swift-docs`, which we then can link.

I like this option.  We only want one static site in the end, so we may as well put everything in one branch.  You could either adjust this PR or do that in a follow up, either way is fine with me.  I think the branch is called `gh-pages` though right?

bendk (approved):
Looks great to me, as you mention the only remaining issue is where to put the docs.  I'll let you decide if you want to adjust this PR or create a follow-up.

PR #6362, opened 2024-09-03
WIP: NSS managed logins key storage

jo:
:warning: **Work in Progress - Please Do Not Merge**

_This pull request is a proof of concept, representing an initial approach to integrating key management for the logins component in Firefox’s desktop password manager. The implementation is still in its early stages and serves as a foundation for further discussion and refinement._

### Objective
To integrate the logins component into the desktop environment and replace the existing logins backend with the Rust implementation, key management must be incorporated. Currently, NSS handles this, ensuring that the application code doesn't directly interact with key management. Instead, it simply uses the encrypt and decrypt endpoints provided by NSS, which internally access the corresponding keys stored in the slot.

Another approach would be to handle this integration on the JavaScript side, but it would present similar challenges due to the absence of the necessary bindings. Here, I have pursued the Rust path.

### Contents of the Pull Request
This pull request includes the following changes:

1.  **Feature Flag `keydb`**:
    -   A new feature flag `keydb` has been added, which provides the additional functionalities listed below.
2.  **New Struct `ManagedLoginStore`**:
    -   Implements a new struct `ManagedLoginStore`, responsible for generating, persisting, and retrieving the key, providing the same API as `LoginStore`, but with transparent key management.
3.  **Extended NSS Bindings**:
    -   New NSS bindings have been added, necessary for the tasks mentioned above:
        -   pub fn PK11\_GetInternalKeySlot() -> \*mut PK11SlotInfo;
        -   pub fn PK11\_SetSymKeyNickname(key: \*mut PK11SymKey, name: \*const c\_uchar);
        -   pub fn PK11\_ImportSymKeyWithFlags(...) -> \*mut PK11SymKey;
            pub fn PK11\_ListFixedKeysInSlot(...) -> \*mut PK11SymKey;
4.  **New Function `nss::pk11::sym_key::retrieve_or_create_and_import_and_persist_aes256_key_data(name: &str) -> Result<Vec<u8>>`**:
    -   This function checks if an existing key can be found.
    -   If no key is found, it generates and stores a new one.
    -   Finally, it returns the key data for use with Logins component.
5.  **Adjustment of NSS Initialization**:
    -   The initialization of NSS via `ensure_nss_initialized` has been adjusted based on the new feature flag, allowing keys to be persisted.

### Open Questions/Problems/TODOs

- **Binding Signatures**:
   -  We need to double check whether binding signatures are written correctly, especially in regards to mutability.
-   **Profile Directory Path**:
    -   When initializing NSS via `ensure_nss_initialized`, this patch reads the path to the profile (where the NSS key databases are located) from the `PROFILE_DIR` environment variable based on the feature flag. This approach might not be optimal. We need to explore better ways to pass the path.
-   **Feature Flag and `UDL` Integration**:
    -   This version just compiles a different UDL, depending on the `keydb` feature flag. I bet there is a better way?
-   **Further Implementation in `ManagedLoginStore`**:
    -   Functions still need to be implemented (this concerns `Arc` handling):
        -   `reset`
        -   `register_with_sync_manager`
        -   `create_logins_sync_engine`
-   **Naming of `ManagedStore`**:
    -   We might want to consider a better name for the `ManagedStore` struct.
-   **Architecture Review**:
    -   The architecture and code structure implemented here should be reviewed and discussed.
-   **Abstraction of Key Management**:
    -   The `LoginStore` should be modified to accept a `LoginManager` trait implementation, allowing key management to be abstracted

---

  - Note:

### DIFF
--- a/components/logins/src/logins-managed-store.udl
+++ b/components/logins/src/logins-managed-store.udl
@@ -106,18 +106,18 @@ interface LoginsApiError {
     UnexpectedLoginsApiError(string reason);
 };
 
-interface LoginStore {
+interface ManagedLoginStore {
     [Throws=LoginsApiError]
     constructor(string path);
 
     [Throws=LoginsApiError]
-    EncryptedLogin add(LoginEntry login, [ByRef]string encryption_key);
+    Login add(LoginEntry login);
 
     [Throws=LoginsApiError]
-    EncryptedLogin update([ByRef] string id, LoginEntry login, [ByRef]string encryption_key);
+    Login update([ByRef] string id, LoginEntry login);
 
     [Throws=LoginsApiError]
-    EncryptedLogin add_or_update(LoginEntry login, [ByRef]string encryption_key);
+    Login add_or_update(LoginEntry login);
 
     [Throws=LoginsApiError]
     boolean delete([ByRef] string id);
@@ -132,16 +132,16 @@ interface LoginStore {
     void touch([ByRef] string id);
 
     [Throws=LoginsApiError]
-    sequence<EncryptedLogin> list();
+    sequence<Login> list();
 
     [Throws=LoginsApiError]
-    sequence<EncryptedLogin> get_by_base_domain([ByRef] string base_domain);
+    sequence<Login> get_by_base_domain([ByRef] string base_domain);
 
     [Throws=LoginsApiError]
-    Login? find_login_to_update(LoginEntry look, [ByRef]string encryption_key);
+    Login? find_login_to_update(LoginEntry look);
 
     [Throws=LoginsApiError]
-    EncryptedLogin? get([ByRef] string id);
+    Login? get([ByRef] string id);
 
     [Self=ByArc]
     void register_with_sync_manager();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1734,7 +1734,7 @@ checksum = "78cc372d058dcf6d5ecd98510e7fbc9e5aec4d21de70f65fea8fecebcd881bd4"
 
 [[package]]
 name = "glean-build"
-version = "14.5.1"
+version = "14.0.1"
 dependencies = [
  "xshell-venv",
 ]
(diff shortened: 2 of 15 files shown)


components/logins/src/logins-managed-store.udl
```diff
@@ -0,0 +1,148 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+namespace logins {
```
mhammond, line 5: we are very very close to being able to use procmacros, which would remove the need for all this duplication :)

mhammond (commented):
This is great - maybe we could lean into it further :) 

Note that only makes sense once you are actually moving ahead rather than exploring.

The existing sig for mobile is, eg:
> pub fn update(&self, id: &str, entry: LoginEntry, enc_key: &str)

where you have removed that key. Which is much better for a public api.

Alternatively, we could have, say

```trait KeyStore { ... }``` - there would be a few options for what this should do, but it could start with just `get_key()`.

The public API could use `Arc<dyn KeyStore>` and maybe could be passed to the constructors. You could impl that trait like done here and mobile could impl that trait wrapping their string key (or however they like, we just don't want to hold it ourselves)

We could probably change many rust signatures here but avoid a breaking change for consumers by updating the wrappers.

There would be a few hiccups, but we might be able to end up with a great api that we can use everywhere. But as above, that's more for once you lean into this :)

bendk:
> `trait KeyStore { ... }` - there would be a few options for what this should do, but it could start with just `get_key()`.
> 
> The public API could use `Arc<dyn KeyStore>` and maybe could be passed to the constructors. You could impl that trait like done here and mobile could impl that trait wrapping their string key (or however they like, we just don't want to hold it ourselves)

+1 to this design.  I really like the pattern of using traits for functionality that we can't always handle in Rust and giving consumers multiple options for constructing an implementation.

jo:
This has been a spike, which is completed. Actual implementation work continues in https://github.com/mozilla/application-services/pull/6469 (introduction of EncryptorDecryptor trait)

PR #6363, opened 2024-09-04
Add Swift documentation to GitHub pages

gruberb:
Integrating the `swift-docs` for the Rust components into the `md-book` for `application-services`. We want to have language specific API docs accessible through the browser. This PR adds a folder in the `gh-pages` branch for the `swift-docs`, which allows us to access it through the GitHub pages domain. 

The link will not show up on the `md-books` itself. 

I did a test deploy to `gh-pages` and it all works:
https://mozilla.github.io/application-services/swift/Classes/FirefoxAccount.html

The `gh-pages` docs are also not impacted by the change:
https://mozilla.github.io/application-services/book/index.html


.github/workflows/build-docs.yaml
```diff
@@ -1,56 +1,107 @@
-name: Build and Deploy
-on:   
-  schedule:
-    - cron: '0 1 * * *'
```
gruberb, line 4: Not sure why this runs on a schedule. We can leave it on this cron job, or we build whenever we push to `main`. I am a fan of "building when pushing to `main`", but there might be a reason we do it currently on a scheduled basis.

.github/workflows/build-docs.yaml
```diff
@@ -1,56 +1,106 @@
...
+      - name: Prepare documentation
+        run: |
+          mkdir -p gh-pages
+          mv docs-ubuntu-latest/* gh-pages/
+          mkdir -p gh-pages/swift
+          mv docs-macos-latest/* gh-pages/swift/
+
+      - name: Deploy to GitHub Pages
         run: |
-          git config --local user.email "${{ github.actor }}@users.noreply.github.com"
```
gruberb, line 45: Not sure if we want to leave it as a local config, or set it to a github actions bot.

.github/workflows/build-docs.yaml
bendk: +1 to changes this to trigger on pushes to main.

.github/workflows/build-docs.yaml
```diff
@@ -1,56 +1,106 @@
...
+
 jobs:
-  build-and-deploy-docs:
-    runs-on: ubuntu-latest
-    steps: 
-      - name: Get repository and install Rust dependencies
+  build-docs:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
bendk, line 12: I'm slightly confused about this matrix.  It seems steps either run on one platform or the other without much shared behavior.  What do you think about splitting this into 2 jobs?

bendk (changes requested):
Seeing the docs in action is great.  I think I would prefer splitting this into mac and ubuntu jobs, but maybe I'm just missing something.

.github/workflows/build-docs.yaml
bendk: I don't know GitHub actions well enough to have an opinion.   I'd say go with whatever you think is the best practice.

.github/workflows/build-docs.yaml
gruberb: Very fair point! Will do 👍

bendk (approved):

PR #6371, opened 2024-09-10
fix: add proper Rust doc comments to UDL files

gruberb:
So that the docs are being rendered properly, UDL files have to adhere to the Rust commenting style (`//` for internal comments, `///` for Rust doc comments, which will be rendered when creating documentation).

We use `uniffi` to create language documentation out of the UDL file, and put it here for:
* Swift: https://mozilla.github.io/application-services/swift/index.html
* Kotlin: https://mozilla.github.io/application-services/kotlin/index.html

### DIFF
--- a/components/as-ohttp-client/src/as_ohttp_client.udl
+++ b/components/as-ohttp-client/src/as_ohttp_client.udl
@@ -13,23 +13,23 @@ enum OhttpError {
     "DuplicateHeaders",
 };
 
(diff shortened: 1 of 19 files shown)


bendk (changes requested):
The changes look fine, but one of the changes broke the build.  I think the issue is that we don't support "UDL-level" docstrings -- i.e. triple slashes at the top of the UDL file.

PR #6372, opened 2024-09-12
Add ability for the tabs engine to close db connection

skhamis:
This is mainly a fix for the issue that came up in trying to land https://bugzilla.mozilla.org/show_bug.cgi?id=1911626. Which is happening because we don't properly close the tabs DB connection in the browser mochitest thus causing the error in the bug.

This exposes the ability to call `close` on the tab store during shutdown and ensure that we don't hit this issue.

  - Note:

### DIFF
--- a/components/tabs/src/storage.rs
+++ b/components/tabs/src/storage.rs
@@ -57,6 +57,12 @@ pub struct ClientRemoteTabs {
     pub remote_tabs: Vec<RemoteTab>,
 }
 
+pub(crate) enum DbConnection {
+    Created,
+    Open(Connection),
+    Closed,
+}
+
 // Tabs has unique requirements for storage:
 // * The "local_tabs" exist only so we can sync them out. There's no facility to
 //   query "local tabs", so there's no need to store these persistently - ie, they
@@ -72,7 +78,7 @@ pub struct ClientRemoteTabs {
 pub struct TabsStorage {
     local_tabs: RefCell<Option<Vec<RemoteTab>>>,
     db_path: PathBuf,
-    db_connection: Option<Connection>,
+    db_connection: DbConnection,
 }
 
 impl TabsStorage {
@@ -80,7 +86,18 @@ impl TabsStorage {
         Self {
             local_tabs: RefCell::default(),
             db_path: db_path.as_ref().to_path_buf(),
-            db_connection: None,
+            db_connection: DbConnection::Created,
+        }
+    }
+
+    pub fn close(&mut self) {
+        if let DbConnection::Open(conn) =
+            std::mem::replace(&mut self.db_connection, DbConnection::Closed)
+        {
+            if let Err(err) = conn.close() {
+                // Log the error, but continue with shutdown
+                log::error!("Failed to close the connection: {:?}", err);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


components/tabs/src/storage.rs
```diff
@@ -131,8 +155,11 @@ impl TabsStorage {
             flags,
             &crate::schema::TabsMigrationLogic,
         )?;
-        self.db_connection = Some(conn);
-        Ok(self.db_connection.as_ref().unwrap())
+        self.db_connection = DbConnection::Open(conn);
+        match self.db_connection {
+            DbConnection::Open(ref conn) => Ok(conn),
+            _ => Err(Error::UnexpectedConnectionState),
```
linabutler, line 161: This might be a good place for `unreachable!()`. Since we just moved the connection into `self.db_connection` above, and we know we have exclusive access to the `&mut self`, any other variant here is impossible.

components/tabs/src/storage.rs
```diff
@@ -105,8 +124,11 @@ impl TabsStorage {
             &crate::schema::TabsMigrationLogic,
         ) {
             Ok(conn) => {
-                self.db_connection = Some(conn);
-                Ok(self.db_connection.as_ref())
+                self.db_connection = DbConnection::Open(conn);
+                match self.db_connection {
+                    DbConnection::Open(ref conn) => Ok(Some(conn)),
+                    _ => Err(Error::UnexpectedConnectionState),
```
linabutler, line 130:
```suggestion
                    _ => unreachable!("impossible value"),
```

Ditto, this is a great place for an `unreachable!()`.

components/tabs/src/storage.rs
```diff
@@ -72,15 +78,26 @@ pub struct ClientRemoteTabs {
...
     pub fn new(db_path: impl AsRef<Path>) -> Self {
         Self {
             local_tabs: RefCell::default(),
             db_path: db_path.as_ref().to_path_buf(),
-            db_connection: None,
+            db_connection: DbConnection::Created,
+        }
+    }
+
+    pub fn close(&mut self) -> Result<bool> {
```
linabutler, line 93:
```suggestion
    pub fn close(&mut self) -> Result<()> {
```

Minor nit: It's a bit more idiomatic to use `()` when the result isn't meaningful.

components/tabs/src/storage.rs
```diff
@@ -72,15 +78,26 @@ pub struct ClientRemoteTabs {
...
+        match std::mem::replace(&mut self.db_connection, DbConnection::Closed) {
+            DbConnection::Open(conn) => {
+                // Now we own the connection, so we can close it.
+                conn.close().map_err(|(_conn, err)| err)?;
```
linabutler, line 97: Thought: I wonder if it's worth swallowing `close()` errors, and having this method return nothing. On the one hand, it does lose context if closing fails; on the other...I'm not super sure how a caller _would_ recover from a `close()` failure.

Thinking a bit more about it, if `close()` fails inside an `AsyncShutdown` blocker, it's just going to print a message to the Firefox JS console, and continue with shutdown, so I wonder if we can simplify that even more, and make this non-throwing.

WDYT?

linabutler (approved):
Nice work, @skhamis! I left a few comments for you to think about, but I trust you to land this whenever you're happy with it!

components/tabs/src/storage.rs
skhamis: decided to go with your below approach and remove this completely!

components/tabs/src/storage.rs
skhamis: I'm inclined to agree with you here! I decided to keep the error logged just so anyone potentially running into issues can see what happened but agree that bubbling the error doesn't really help in this scenario. Certainly simplifies the code!

PR #6373, opened 2024-09-12
Switch Android tests from junit 4 to junit-vintage [ci full]

rvandermeulen:
This is a useful first step to migrating to junit 5 as it allows you to switch to version 5 without having to rewrite every test en masse at the same time.


skhamis (approved):
:shipit:

PR #6374, opened 2024-09-12
Convert AutofillTest to JUnit 5

rvandermeulen:


components/autofill/android/src/test/java/mozilla/appservices/autofill/AutofillTest.kt
```diff
@@ -1,37 +1,33 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
 import org.robolectric.annotation.Config
+import java.io.File
 
-@RunWith(RobolectricTestRunner::class)
 @Config(manifest = Config.NONE)
 class AutofillTest {
-    @Rule
-    @JvmField
```
bendk, line 16: Stackoverflow says that we need to keep the `@JvmField`: https://stackoverflow.com/questions/68202376/how-to-use-junit-5s-tempdir-with-kotlin

bendk (changes requested):
Looks pretty much right, but I don't think this test is actually running.  When I change one of the `assertFalse` lines to `assertTrue` and run `./gradlew autofill:test` it still passes.

rvandermeulen:
Turns out robolectric doesn't support JUnit 5, so this is a dead-end at the moment.

PR #6377, opened 2024-09-13
Disco 2821 - Add a score API for relevancy

ncloudioj:

  - Note:

This adds a score API to the relevancy component.

Note:
- For simplicity, I opted to implement this as a stateless function instead of adding it as a method to `RelevancyStore`. Meaning that the consumer needs to provide both the user interest vector and the content categories for scoring. The former should be accessible through either `RelevancyStore::ingest()` or `RelevancyStore::user_interest_vector()`. The latter can be specified directly by the content provider.
- The score is implemented using a simple sigmoid function for now. See the code for more details.
- This is slightly different from the original design, in particular, it doesn't provide the ranking directly. The consumer needs to score all the content one by one (optionally, merge the score result with other scores) and rank all the candidates by themselves. I don't think that's too bad for them for the experimentation phase. We can add more ranking capabilities later as we go.

### DIFF
--- a/components/relevancy/src/lib.rs
+++ b/components/relevancy/src/lib.rs
@@ -13,6 +13,7 @@ mod db;
 mod error;
 mod ingest;
 mod interest;
+mod ranker;
 mod rs;
 mod schema;
 pub mod url_hash;
@@ -20,6 +21,7 @@ pub mod url_hash;
 pub use db::RelevancyDb;
 pub use error::{ApiResult, Error, RelevancyApiError, Result};
 pub use interest::{Interest, InterestVector};
+pub use ranker::score;
 
 use error_support::handle_error;
 
--- a/components/relevancy/src/lib.rs
+++ b/components/relevancy/src/lib.rs
@@ -96,7 +96,7 @@ impl RelevancyStore {
     /// consumer can show it in an `about:` page.
     #[handle_error(Error)]
     pub fn user_interest_vector(&self) -> ApiResult<InterestVector> {
-        todo!()
+        self.db.read(|dao| dao.get_frecency_user_interest_vector())
     }
 }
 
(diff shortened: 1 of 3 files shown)


bendk (approved):
This seems like a good step to me, although I'm not sure how it fits into a larger plan of re-ordering suggestions.  A consumer could call this for each suggestion we return, but then it's not clear how they should reorder those suggestions based on the results.

ncloudioj:
>  A consumer could call this for each suggestion we return, but then it's not clear how they should reorder those suggestions based on the results.

True, the consumer would need to do some additional work here. My thinking is that doing so actually allows consumers to get some more flexibility (e.g. how to consolidate the score with other score(s)) for now. Otherwise, it'll be much harder for the component to offer the same IMO.

Here is an example of how the consumer uses this to simply sum the scores up and rank in the descending order in JS.   

```js
const items = [
  { id: 0, score: 0.2, categories: ["food"] },
  { id: 1, score: 0.25, categories: ["animals"] },
  { id: 2, score: 0.2, categories: ["food", "animal"] },
]

items.map(item => {
  ..item,
  // The consumer will decide the score consolidation strategy: sum/max/weighted sum etc.
  score: item.score + score(interestVec, item.categories),  
}).sort((l, r) => {  // sort in descending order by score
  if (l.score < r.score) {
    return 1;
  } else if (l.score > r.score) {
    return -1;
  } else {
    return 0;
  }
});
```

I think this would meet the MVP for experimentation. Longer term, once we have a better understanding about scoring/ranking strategies, then we can backport those into the component later.

Does that make sense?

bendk:
> I think this would meet the MVP for experimentation. Longer term, once we have a better understanding about scoring/ranking strategies, then we can backport those into the component later.

That sounds like a great plan to me.

PR #6378, opened 2024-09-16
Starting new Remote Settings API

bendk:
Added the `RemoteSettingsService` and `RemoteSettingsClient` types. These provide the new remote settings API that we're hoping to implement. The goal is to match the API of the official client more closely.  The main changes are:
  - Downloaded data is stored locally
  - Getting/syncing records are split up into different operations to avoid hitting the network when getting settings values during startup.
  - Application-wide configuration is managed by `RemoteSettingsService`, which uses that to construct the clients.

This code is still a WIP.  The hope is that this outlines the general shape of the new API and the work that needs to be done.

  - Note:

### DIFF
--- a/components/remote_settings/src/lib.rs
+++ b/components/remote_settings/src/lib.rs
@@ -2,17 +2,23 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::{fs::File, io::prelude::Write, sync::Arc};
+
+use camino::Utf8Path;
+
 pub mod cache;
 pub mod error;
 pub use error::{RemoteSettingsError, Result};
-use std::{fs::File, io::prelude::Write};
 pub mod client;
 pub use client::{
     Attachment, Client, GetItemsOptions, RemoteSettingsRecord, RemoteSettingsResponse,
     RsJsonObject, SortOrder,
 };
 pub mod config;
-pub use config::{RemoteSettingsConfig, RemoteSettingsServer};
+pub mod storage;
+pub use config::{RemoteSettingsConfig, RemoteSettingsConfig2, RemoteSettingsServer};
+
+use storage::Storage;
 
 uniffi::include_scaffolding!("remote_settings");
 
@@ -47,6 +53,78 @@ impl RemoteSettings {
         let resp = self.client.get_attachment(&attachment_location)?;
         let mut file = File::create(path)?;
         file.write_all(&resp)?;
+        Ok(())
+    }
+}
+
+pub struct RemoteSettingsService {
+    config: RemoteSettingsConfig2,
+}
+
+impl RemoteSettingsService {
+    pub fn new(config: RemoteSettingsConfig2) -> Self {
+        RemoteSettingsService { config }
+    }
+
+    pub fn make_client(&self, collection_name: String) -> Result<Arc<RemoteSettingsClient>> {
+        let storage = Storage::new(
+            &Utf8Path::new(&self.config.storage_dir).join(format!("{}.sqlite", collection_name)),
+        )?;
+        let client = Client::new_for_remote_settings_service(
+            self.config.server.clone(),
+            self.config
+                .bucket_name
+                .clone()
+                .unwrap_or_else(|| "main".to_string()),
+            collection_name,
+        )?;
+        Ok(Arc::new(RemoteSettingsClient { client, storage }))
+    }
+}
+
+pub struct RemoteSettingsClient {
+    client: Client,
+    storage: Storage,
+}
+
+impl RemoteSettingsClient {
+    pub fn get(&self) -> Result<Option<Vec<RemoteSettingsRecord>>> {
+        self.storage.get_records()
+    }
+
+    pub fn sync(&self) -> Result<()> {
+        let resp = match self.storage.last_modified_time()? {
+            Some(last_modified) => self.client.get_records_since(last_modified)?,
+            None => self.client.get_records()?,
+        };
+        self.storage
+            .update_records(resp.records, resp.last_modified)?;
+        Ok(())
+    }
+
+    pub fn sync_if_empty(&self) -> Result<()> {
+        if self.storage.last_modified_time()?.is_none() {
+            self.sync()?;
+        }
+        Ok(())
+    }
+
+    pub fn download_attachment_to_path(
+        &self,
+        attachment_location: String,
+        path: String,
+    ) -> Result<()> {
+        let data = match self.storage.get_attachment_data(&attachment_location)? {
+            Some(data) => data,
+            None => {
+                let data = self.client.get_attachment(&attachment_location)?;
+                self.storage
+                    .store_attachment_data(&attachment_location, &data)?;
+                data
+            }
+        };
+        let mut file = File::create(path)?;
+        file.write_all(&data)?;
         Ok(())
     }
 }
--- a/components/remote_settings/src/remote_settings.udl
+++ b/components/remote_settings/src/remote_settings.udl
@@ -55,6 +55,10 @@ enum RemoteSettingsError {
     "ConfigError",
 };
 
+/// Remote settings client
+///
+/// Soft-deprecated.  We plan to move clients over to the new RemoteSettingsService and
+/// RemoteSettingsClient interfaces.
 interface RemoteSettings {
     /// Construct a new Remote Settings client with the given configuration.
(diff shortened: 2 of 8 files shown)


components/remote_settings/src/remote_settings.udl
```diff
@@ -73,3 +77,50 @@ interface RemoteSettings {
...
 };
+
+/// Stores application-level configuration data and constructs clients
+interface RemoteSettingsService {
+    constructor(RemoteSettingsConfig2 config);
+
+    /// Construct a new client with the configuration passed to the constructor
+    [Throws=RemoteSettingsError]
+    RemoteSettingsClient make_client(string collection_name);
+};
```
bendk, line 88: This seems like a nice idea to me, but it's not totally clear that we need it.  How do others feel about this?

components/remote_settings/src/lib.rs
```diff
@@ -51,6 +57,78 @@ impl RemoteSettings {
...
+                .clone()
+                .unwrap_or_else(|| "main".to_string()),
+            collection_name,
+        )?;
+        Ok(Arc::new(RemoteSettingsClient { client, storage }))
+    }
+}
+
+pub struct RemoteSettingsClient {
+    client: Client,
```
leplatrem, line 86: nit: it could be called `httpClient`

leplatrem (commented):
This is great 💯 Thank you Ben 👌

components/remote_settings/src/remote_settings.udl
Standard8: I'm not quite sure I understand the discussion here. Would this fit with the following use case?

* iOS (or Android) code (swift/Kotlin) sets up the remote settings server to use stage + preview.
* Rust based search client calls `RemoteSettingsClient("search-config-v2").get()`.

Here I'd want the client to return the configuration according to the stage+preview server.

In other words, I would like the Rust based search client to be able to get the collection without having to worry about passing through the remote settings configuration.

I think we want the search client to be able to go directly to remote settings client as it'll avoid having to pass the configuration back and forth across the Uniffi interfaces.

Standard8: And further clarification, I think the Remote Settings server settings should be app-global rather than per-collection, hence I don't really want the search client to have to deal with those as well.

components/remote_settings/src/remote_settings.udl
```diff
@@ -73,3 +77,50 @@ interface RemoteSettings {
...
+///
+/// This is a WIP matches the semantics of the official client more closely.  Once this is complete,
+/// we may decide to deprecate and remote the RemoteSettings class.
+interface RemoteSettingsClient {
+    /// Get the current set of records.
+    ///
+    /// If no records have been downloaded, this will return null.  Use `sync` to fetch the current
+    /// records from the server.
+    [Throws=RemoteSettingsError]
+    sequence<RemoteSettingsRecord>? get();
```
Standard8, line 113: Are we going to include the dumps in remote settings as well?

This would avoid the application having to load them separately, or pass them in. We can also include some of the logic that the desktop client uses to determine when the local records should be updated from the dump or not.

If so, then I'd expect this to be more `If no records have been downloaded, this will load from the local dumps, if found. Otherwise this will return null`.

components/remote_settings/src/remote_settings.udl
```diff
@@ -73,3 +77,50 @@ interface RemoteSettings {
...
+
+    /// Sync records with the server.
+    [Throws=RemoteSettingsError]
+    void sync();
```
Standard8, line 117: Are we going to also periodically check with the server for updates? If so, it might be worth making clear that this is only needed in the force-sync case.

components/remote_settings/src/remote_settings.udl
```diff
@@ -73,3 +77,50 @@ interface RemoteSettings {
...
+
+    /// Sync records with the server if there are no records currently persisted.
+    [Throws=RemoteSettingsError]
+    void sync_if_empty();
+
+    /// Download an attachment with the provided id to the provided path.
+    [Throws=RemoteSettingsError]
+    void download_attachment_to_path(string attachment_id, string path);
```
Standard8, line 125: Desktop currently has a slightly different interface. It downloads the attachment to the database (if not included in dumps locally), and then either allows loading from the dump, the database cache or to actually download it. 

It also passes in the attachment record, which I believe means it can check hashes and out-of-date.

What's the advantage of downloading this to application specified file? Would it be better to have remote settings handle the file paths/cache, and then maybe tell the app where the path of the file is?

The desktop api currently returns blobs, but maybe storing the file under storage_path and returning the path within that would be better?

components/remote_settings/src/storage.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// Get the last_modified time from the most recent call to [Self::update_records].
+    pub fn last_modified_time(&self) -> Result<Option<u64>> {
+        Ok(None)
+    }
+
+    /// Get attachment data from storage
+    pub fn get_attachment_data(&self, _location: &str) -> Result<Option<Vec<u8>>> {
+        // Q: should we store attachment data directly in SQLite or put it in the filesystem?
+        todo!()
```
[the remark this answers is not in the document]
Standard8: Presumably that data would then have to be returned across the Uniffi interface. Are we likely to hit performance issues with serialising that data vs saving it to a file?

components/remote_settings/src/remote_settings.udl
leplatrem: On Desktop we deprecated the file-base API, because it was tedious to purge old files etc. 
If there are uniffi concerns, maybe we could have a hybrid approach, where downloaded files are tracked in a db and referenced as paths?

components/remote_settings/src/remote_settings.udl
bendk: Thanks for pointing this out.  My proposal was that the consumer app would create a `RemoteSettingsService` and use that as a factory for all the clients/collections.  `RemoteSettingsService` stores the server config and avoids the need for a global variable.

However, I can definitely see why that would be annoying for your use-case.  In order for you to create remote settings clients that used the correct server, the consumer app would need to pass the `RemoteSettingsService` to you.

A global variable could simplify things, but then you have to worry about startup order.  What happens if you constructed your client before the consumer app configured the server?  I'm not sure if this is a big deal or not, but I like the fact that the `RemoteSettingsService` makes this more explicit.

I don't have strong feelings either way, what do others think?

components/remote_settings/src/storage.rs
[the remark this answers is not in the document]
bendk: Passing data across the FFI requires a couple extra copies, but so would saving/reading it to a file.  I think it should be fine for the use-cases we're talking about, but should probably double-check that.

components/remote_settings/src/remote_settings.udl
bendk: Yes, but I plan to add this later on.  I'm wondering if that should be a different class though, since if you know there's a local dump then `get()` doesn't need to return a nullable value.

components/remote_settings/src/remote_settings.udl
bendk: This is a great question.  I think the periodic syncing should be driven by the embedding application, since they have a closer integration with the OS and can create background workers, etc.  However, it seems annoying to make each remote settings client consumer have to call `sync()` for their client.

How about this:
- We rename this method to `force_sync()` and documentation about that.
- We add a `RemoteSettingsService::sync()` method, which is what the embedding application will call.  That updates the remote settings data for all the clients that it knows about (or something similar, I'm not sure exactly how this should work).

components/remote_settings/src/remote_settings.udl
bendk: Let's just remove this method for now, I don't know of a real use-case for it.

components/remote_settings/src/lib.rs
bendk: I'm going to go with `api_client`, since it does a bit more than just straight HTTP.

bendk:
I'm going to close this one in favor of https://bugzilla.mozilla.org/show_bug.cgi?id=1919135.  I like that one better for a couple reasons:
 - It takes a docs-first approach and just describes what the general API should look like.  I think that's a lot simpler to review than this one.
 - It has several updates based on the feedback I got here.

bendk:
Oops, I got the bug ID wrong in the last comment when I first posted it.  I updated that link.  Also, here's the phabricator patch: https://phabricator.services.mozilla.com/D225266

PR #6380, opened 2024-09-18
Switch suggest component from UDL to proc-macros

gruberb:
Deleting the `suggest.udl` file and use proc-macros to export directly from the Rust codebase.


testing/separated/places-bench/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 publish = false
 
 [[bench]]
-name = "places"
+name = "places-bench"
```
gruberb, line 10: Had to rename the target to something else than `places`, because building the project sometimes caused an error when two crates have the same name (there is a `places` crate as well).

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
```diff
@@ -68,7 +68,6 @@
 		39F5D766295616E3004E2384 /* NimbusBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39F5D765295616E3004E2384 /* NimbusBuilder.swift */; };
 		45CC574A28AD9C86006D55AA /* errorsupport.udl in Sources */ = {isa = PBXBuildFile; fileRef = 45CC574828AD9C31006D55AA /* errorsupport.udl */; };
 		AB9C392E2A0DA61900AF5ADE /* remote_settings.udl in Sources */ = {isa = PBXBuildFile; fileRef = AB65933B2A0C524E00DBF059 /* remote_settings.udl */; };
-		BCFF93E82AA7E1ED005B5B71 /* suggest.udl in Sources */ = {isa = PBXBuildFile; fileRef = 3082A3E24B1A68B03DB4EEA3 /* suggest.udl */; };
```
gruberb, line 71: Since we don't have a `suggest.udl` file anymore, I am not sure how this has to change. Maybe @linabutler knows?

automation/swift-components-docs/generate-swift-project.sh
```diff
@@ -1,78 +1,54 @@
...
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/suggest/src/suggest.udl" -l swift -o "$SWIFT_DIR"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/sync_manager/src/syncmanager.udl" -l swift -o "$SWIFT_DIR"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/sync15/src/sync15.udl" -l swift -o "$SWIFT_DIR"
-$CARGO uniffi-bindgen generate "$REPO_ROOT/components/as-ohttp-client/src/as_ohttp_client.udl" -l swift -o "$SWIFT_DIR"
+# Move generated header files to the include directory
+echo "Moving header files to include directory..."
+mv "$SWIFT_DIR"/*.h "$INCLUDE_DIR" || {
+  echo "Error: Failed to move header files."
+  exit 1
+}
```
bendk, line 39: I just put up a PR to do a similar cleanup.  We can merge one first and rebase the other on top, I don't really have a preference on which order to do it in -- do you?

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
bendk: Lina definitely knows best, but remove the UDL file from the sources seems right to me.  AFAIK, the only reason it's there is to trigger a recompile if it changes.

bendk (approved):
Looks great.  To me this seems perfect, but we should wait for @linabutler to confirm that `project.pbxproj` change.

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
linabutler: This looks totally fine to me—I don't think we have any Swift-specific tests, so I wonder if we can remove the entire `suggest` group (including the `Generated/` subfolder), in addition to the UDL?

linabutler (approved):
So cool! This looks just wonderful to me, too; thanks!

automation/swift-components-docs/generate-swift-project.sh
gruberb: Hm, this one is open for so long, I would love to see it done :D

megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj
gruberb: Hm, I might need a direct pointer to that! Do you mean this part? https://github.com/mozilla/application-services/blob/main/megazords/ios-rust/MozillaTestServices/MozillaTestServices.xcodeproj/project.pbxproj#L669-L677

linabutler: Yep!

Let's try deleting lines 71 (✅), 190 (✅), 356, 571-579, 669-677, 814 (✅).

I wish GitHub could link to non-contiguous ranges.

PR #6381, opened 2024-09-18
Define the initial SearchEngineSelector interface

Standard8:
This is a PR to define the initial SearchEngineSelector interface, subsequent PRs will work on the actual implementation & full tests.

No changelog should be necessary as this is code is not yet used.

  - Note:


components/search/src/search.udl
```diff
@@ -10,8 +10,200 @@ interface SearchApiError {
...
+    string locale;
+
+    /// The home region that the user is currently identified as being within.
+    /// On desktop & android there is a 14 day lag after detecting a region
+    /// change before the home region changes. TBD: iOS?
+    string region;
+
+    /// The current distribution channel.
+    /// Use `default` for a self-build or an unknown channel.
+    SearchDistributionChannel channel;
```
rocketsroger, line 158: Will this allow us to support different search engine list for distribution partnerships? Or should we use distribution_id below?

example here is a manufacturer that preloads/preinstalls Fenix in their Android OS but want to add a search engine (default or not) as a part of the partnership?

components/search/src/types.rs
```diff
@@ -0,0 +1,179 @@
...
+/// Use `default` for a self-build or an unknown channel.
+#[derive(Debug, uniffi::Enum)]
+pub enum SearchUpdateChannel {
+    Default,
+    Nightly,
+    Aurora,
+    Beta,
+    Release,
+    ESR,
+}
```
nbhasin2, line 37: Do they need to correspond to exact names or these are generic?

Ex. Fennec is the dev build for Firefox iOS

nbhasin2: So in theory I can locally associate Fennec to fetch Release / Beta or Nightly right?

nbhasin2: Oh ignore this, I understand

Standard8: Just to confirm, for a dev build we would ideally use `default`. For the official builds it depends on if they were nightly/beta/release etc.

components/search/src/selector.rs
```diff
@@ -1,27 +1,60 @@
...
+    /// Clears the search configuration from memory if it is known that it is
+    /// not required for a time.
+    pub fn clear_search_config(&self) {}
+
+    /// Filters the search configuration with the user's given environment,
+    /// and returns the set of engines and parameters that should be presented
+    /// to the user.
     #[handle_error(Error)]
-    pub fn filter_engine_configuration(&self) -> SearchApiResult<()> {
+    pub fn filter_engine_configuration(
```
sarah541, line 32: Did we finalize if this would be an ordered set of search engines or not?

components/search/src/types.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// The update channel of the user's build.
+    pub update_channel: SearchUpdateChannel,
+
+    /// The distribution id for the user's build.
+    pub distribution_id: String,
+
+    /// The search related experiment id that the user is included within. On
+    /// desktop this is the `searchConfiguration.experiment` variable.
+    pub experiment: String,
```
sarah541, line 71: I am not sure what this is for Android today. Will look for it

components/search/src/types.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug, uniffi::Record)]
+pub struct SearchEngineUrls {
+    /// The URL to use for searches.
+    pub search: SearchEngineUrl,
+
+    /// The URL to use for suggestions.
+    pub suggestions: Option<SearchEngineUrl>,
+
+    /// The URL to use for trending suggestions.
+    pub trending: Option<SearchEngineUrl>,
```
sarah541, line 127: Just curious, do we have this for Android?

components/search/src/types.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// (e.g. general, shopping, travel, dictionary). Currently, only marking as
+    /// a general search engine is supported.
+    /// On Android, only general search engines may be selected as "default"
+    /// search engines.
+    pub classification: String,
+
+    /// The identifier of the search engine. This is used as an internal
+    /// identifier, e.g. for saving the user's settings for the engine. It is
+    /// also used to form the base telemetry id and may be extended by telemetrySuffix.
+    pub identifier: String,
```
sarah541, line 146: This is the one that will also map to identifier in search-config-icons-v2?

components/search/src/types.rs
Standard8: I don't think Android has anything similar to this. We introduced it on desktop with the previous version of the configuration because it was that configuration was flexible enough to allow the experimentation with the different parameters & deployments.

components/search/src/types.rs
Standard8: Not currently, this is a relatively new feature on desktop that we introduced earlier in the year.

components/search/src/types.rs
Standard8: Yes.

components/search/src/selector.rs
Standard8: At the moment it does, but I'm in two minds about this. On desktop, we don't use the order of the engines from the selector because when we order the engines, we take into account the user installed engines. So the default order on desktop is roughly:

* Default Engine
* Engines defined in the `orders` section of the configuration
* All other engines (application and user-installed) in alphabetical order

Obviously the engine selector can't manage that last point as it doesn't know about user installed engines.

I think mobile currently does:

* Default Engine
* All other application engines in alphabetical order
* All user installed engines in alphabetical order

Desktop also allows reordering of the engines by the user.

Overall, I'm not sure if the selector should actually handle the order, since that feels more of a UI function than back-end functionality, however, I'm open to thoughts.

onemahon: Hmm, this is indeed a bit of a weird case. IMO, generally speaking, I think this kind of comes down to "does the search engine order 'matter' to the Search team?" because, if we wind up having to make 3x the changes every time we decide to change the search engine order, that feels like the type of problem this consolidation project is meant to fix.

But if it's more-or-less "fine" to let each of the clients make different business decisions about search engine ordering, then it'd be fine to leave this unsorted.

IMO, perhaps a middle ground would be a "loose" API contract about Search engine order, which I imagine covers the majority of user scenarios - e.g., perhaps this *should* promise to return them in order, but with a clarifying bit of API documentation saying "list is sorted in default search engine order, but clients might override that (e.g. to accommodate user-entered search engines, or to allow users to manually change the order)."

components/search/src/types.rs
onemahon: Oh! This would actually be pretty helpful, as one of the open questions I have about the Trending Searches request that we've been asked to roadmap is "what API is providing trending search suggestions for us?" Sounds like this might be a simple opportunity to carry that functionality over from Desktop into Android without having to re-think it 😄 

I'm going to link this comment in that ticket, just for transparency! No clue if it'll wind up useful.

components/search/src/selector.rs
Standard8: I've just updated the PR with the new comments about the search order.

I also added an `orderHint` field to `SearchEngineDefinition` as I realised that it was missing, and desktop would need it.

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+/// The list of possible update channels for a user's build.
+/// Use `default` for a self-build or an unknown channel.
+#[derive(Debug, uniffi::Enum)]
+pub enum SearchUpdateChannel {
+    Default,
+    Nightly,
+    Aurora,
+    Beta,
+    Release,
+    ESR,
```
linabutler, line 36:
```suggestion
    Default = 1,
    Nightly = 2,
    Aurora = 3,
    Beta = 4,
    Release = 5,
    ESR = 6,
```

Let's give these variants fixed discriminants, too—unless you wanted to specifically avoid them for `SearchUpdateChannel` but not `SearchApplicationName`?

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+pub struct SearchEngineDefinition {
+    /// An optional list of aliases for this engine.
+    pub aliases: Option<Vec<String>>,
+
+    /// The classification of search engine according to the main search types
+    /// (e.g. general, shopping, travel, dictionary). Currently, only marking as
+    /// a general search engine is supported.
+    /// On Android, only general search engines may be selected as "default"
+    /// search engines.
+    pub classification: String,
```
linabutler, line 141: Do you think this will become an enum eventually, or is the idea to keep it as flexible as possible?

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// The URL to use for trending suggestions.
+    pub trending: Option<SearchEngineUrl>,
+}
+
+/// A definition for an individual search engine to be presented to the user.
+#[derive(Debug, uniffi::Record)]
+pub struct SearchEngineDefinition {
+    /// An optional list of aliases for this engine.
+    pub aliases: Option<Vec<String>>,
```
linabutler, line 134:
```suggestion
    pub aliases: Vec<String>,
```

I _have_ seen `Option<Vec<...>>` in the wild, but I'm wondering if an empty `Vec` to mean "no aliases" would be equivalent, and easier to work with.

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+#[derive(Debug, uniffi::Record)]
+pub struct SearchEngineUrl {
+    /// The PrePath and FilePath of the URL. May include variables for engines
+    /// which have a variable FilePath, e.g. `{searchTerm}` for when a search
+    /// term is within the path of the url.
+    pub base: String,
+
+    /// The HTTP method to use to send the request (`GET` or `POST`).
+    /// If not specified, defaults to GET.
+    pub method: Option<String>,
```
linabutler, line 106:
```suggestion
    pub method: String,
```

Idle thought: IIUC, this would leave it up to the JS / Kotlin / Swift consumer to null-check, and fall back to `GET` if the method is unspecified—is that right? I wonder if we can / should have the Rust logic pass back `GET` automatically?

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/// Parameter definitions for search engine URLs. The name property is always
+/// specified, along with one of value, experiment_config or search_access_point.
+#[derive(Debug, uniffi::Record)]
+pub struct SearchUrlParam {
+    /// The name of the parameter in the url.
+    pub name: String,
+    /// The parameter value, this may be a static value, or additionally contain
+    /// a parameter replacement, e.g. `{inputEncoding}`. For the partner code
```
linabutler, line 87: Idle thought: Would you see us moving parameter replacements into the Rust layer eventually, too (and having it return the "cooked" URLs, with replacements interpolated), or leaving that up to the consuming Desktop / Android / iOS code?

linabutler (approved):
This looks terrific! I left a few non-blocking comments and thoughts, but please feel free to land when ready!

components/search/src/selector.rs
```diff
@@ -1,27 +1,65 @@
...
+        Self::default()
+    }
+
+    /// Sets the search configuration from the given string. This allows for
+    /// reprocessing of the configuration if it has not changed since the
+    /// previous update, e.g. to optimise test running.
+    #[handle_error(Error)]
+    pub fn set_search_config(&self, _configuration: String) -> SearchApiResult<()> {
+        Err(Error::NotImplemented)
+    }
```
mandysGit, line 27: What I'm understanding from your comment here is that we will cache the config, reuse the cached config until the string passed in is updated, and if the string is updated, we update our cached config. 

I think pointing out that the config is cached and the benefits of caching it will make it the comment more clear.
Maybe something like this or parts of this comment I wrote? 

```
/// Sets the search configuration from the given string. If the configuration string 
/// is unchanged since the last update, the cached configuration is reused to avoid
/// unnecessary reprocessing. This helps optimize performance, particularly during
/// test runs where the same configuration may be used repeatedly.
```

components/search/src/selector.rs
```diff
@@ -1,27 +1,65 @@
...
+
+    /// Clears the search configuration from memory if it is known that it is
+    /// not required for a time.
+    pub fn clear_search_config(&self) {}
```
mandysGit, line 31: How or when does it know that it is not required? What's the conditions for that?

components/search/src/selector.rs
```diff
@@ -1,27 +1,65 @@
...
+
+    /// Filters the search configuration with the user's given environment,
+    /// and returns the set of engines and parameters that should be presented
+    /// to the user.
     #[handle_error(Error)]
-    pub fn filter_engine_configuration(&self) -> SearchApiResult<()> {
+    pub fn filter_engine_configuration(
```
mandysGit, line 37: I assume this function is the equivalent for [`fetchEngineConfiguration` ](https://searchfox.org/mozilla-central/rev/7e0ae4372c52b8183d1178132dd6493edb576738/toolkit/components/search/SearchEngineSelector.sys.mjs#233) ? 

I'm wondering if we should keep the same function names for consistency?

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+
+/// The list of possible application names that are currently supported.
+#[derive(Debug, uniffi::Enum)]
+pub enum SearchApplicationName {
+    Firefox = 1,
+    FirefoxAndroid = 2,
+    FocusAndroid = 3,
+    FirefoxIOS = 4,
+    FocusIOS = 5,
+}
```
mandysGit, line 13: Question 1: 
Are these constants in Rust? I wonder if you're writing constants variables here, and if they should be all caps? Or maybe that's just a javascript thing?

Question 2: 
I'm not sure if ordering matters? The list is in a different order than the [applications section for search config v2 schema ](https://searchfox.org/mozilla-central/rev/7e0ae4372c52b8183d1178132dd6493edb576738/toolkit/components/search/schema/search-config-v2-schema.json#110-116)

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+            SearchApplicationName::FirefoxIOS => "firefox-ios",
+            SearchApplicationName::FocusIOS => "focus-ios",
+        }
+    }
+}
+
+/// The list of possible update channels for a user's build.
+/// Use `default` for a self-build or an unknown channel.
+#[derive(Debug, uniffi::Enum)]
+pub enum SearchUpdateChannel {
```
mandysGit, line 30: I noticed this is named `SearchUpdateChannel`, and I was wondering about the reasoning behind that. To maintain consistency with the naming conventions in Desktop, do you think it might make sense to rename it to just channel?

components/search/src/types.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// have other requirements.
+    ///
+    /// The application default engines should not be assumed from this order in
+    /// case of future changes.
+    ///
+    /// The sort order is:
+    ///
+    /// * Application Default Engine
+    /// * Application Default Engine for Private Mode (if specified & different)
+    /// * Engines sorted by descending `orderHint`
```
mandysGit, line 185: I believe we deprecated `orderHint`, and `engineOrders` is now used for the same purpose. To be clear,  in the code we still use `orderHint` but it's really from the ordering of `engineOrders`. 

I remember we decided to not add `orderHint` property in the new config. I'm wondering if it would be clearer to reference `engineOrders` directly instead of `orderHint`, since the `orderHint` property does not exist in search-config-v2? Thoughts?

mandysGit (changes requested):

components/search/src/types.rs
linabutler: > Are these constants in Rust? 

They are! Specifically, they’re enum variants with explicit discriminants, so they’ll have stable raw values when exposed to foreign languages. This makes it easier to convert them [to and from raw values in Swift and Kotlin](https://mozilla.github.io/uniffi-rs/0.27/proc_macro/index.html#variant-discriminants)—but I don’t think the Gecko JS bindings don’t do anything special with them just yet.

> I wonder if you're writing constants variables here, and if they should be all caps? Or maybe that's just a javascript thing?

Yep, it’s a JavaScript thing—Rust convention is to write enum variants in `PascalCase`. UniFFI will take care of turning them into `SHOUTY_SNAKE_CASE` when it generates the JavaScript and Kotlin bindings (and `camelCase` when it generates the Swift bindings).

linabutler: > I'm not sure if ordering matters?

With explicit discriminants, order doesn’t matter.

components/search/src/selector.rs
Standard8: They're not quite equivalent. `fetchEngineConfiguration` gets the configuration from remote settings and filters it.

This function only filters the configuration.

components/search/src/types.rs
Standard8: I certainly have a background thought for the long term that we could potentially completely rewrite the search service in Rust. I think it is also possible that we could have `SearchEngineUrl` be a more featured URL that returns the filled-in URLs. I'm also trying to avoid a bit of scope creep 😉, though we'll keep it in mind.

components/search/src/types.rs
Standard8: I'll make it an enum, to reflect what we currently have in the schema. It might become interesting for what we do when processing the search configuration, but I'll be looking at that in the next patch.

components/search/src/types.rs
Standard8: @linabutler tells me that Swift code doesn't support namespaces, so everything gets flattened into a single "module". Hence, if two different components have a type like `ApplicationName` or `Channel` then it'll refuse to build on Swift.

It doesn't always need to be prefixed with `Search` but it should be fairly guaranteed to be unique (e.g. I'm going to change `RefinedConfig` to `RefinedSearchConfig`).

components/search/src/types.rs
Standard8: Note that I've updated the comment to reference `SearchEngineDefinition.orderHint`, which should hopefully make this a bit clearer.

components/search/src/selector.rs
mandysGit: Right, that makes sense because mobile is fetching the config from remote settings themselves. After that, it's going to call this.

components/search/src/types.rs
mandysGit: LGTM :)

mandysGit (approved):
Thank you for the updates! This looks really good! It's very exciting to see this Rust SearchEngineSelector coming together.

components/search/src/selector.rs
Standard8: I think we might add a fetch + filter, but at the moment I'm waiting to see what the remote settings APIs look like before we decide that.

PR #6384, opened 2024-09-20
Pass the right symbols directory to upload_android_symbols.sh

jcristau:
The command had a stray double quote after `crashreporter-symbols`, and wasn't pointing at $MOZ_FETCHES_DIR, so AFAICT it was not actually uploading anything.


bendk (approved):
Looks good, thanks for doing this.

rvandermeulen:
@mergify backport release-v131

PR #6387, opened 2024-09-20
Fix: add default values for suggest component

gruberb:
Fix https://bugzilla.mozilla.org/show_bug.cgi?id=1919998


bendk (approved):
:ship:

PR #6391, opened 2024-09-23
fix: use library mode for kotlin docs

gruberb:
Using the `uniffi-bindgen` library mode to generate components. We updated `suggest` and `relevancy` and removed the `UDL` file. Therefore, both components are missing in the current published docs.

### DIFF
--- a/automation/swift-components-docs/generate-swift-project.sh
+++ b/automation/swift-components-docs/generate-swift-project.sh
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
-#
+
 # This script builds the Rust crate in its directory and generates Swift bindings,
 # headers, and a module map using UniFFI in library mode.
 
-set -euo pipefail  # Ensure script exits on errors or unset variables
+# Ensure script exits on errors or unset variables
+set -euo pipefail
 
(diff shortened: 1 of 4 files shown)


automation/swift-components-docs/generate-swift-project.sh
```diff
@@ -3,8 +3,6 @@
 # This script builds the Rust crate in its directory and generates Swift bindings,
 # headers, and a module map using UniFFI in library mode.
 
-set -euo pipefail  # Ensure script exits on errors or unset variables
-
```
bendk, line 7: 1) Why remove this one?

2) This is slightly separate, but can you change the megazord below from `megazord` to `megazord_ios`?  That's the megazord we use on iOS.  Nowadays, I think it contains the same components as the Android megazord, but this will probably change at some point.

bendk (changes requested):
Looks good to me, just a couple minor comments.

automation/swift-components-docs/generate-swift-project.sh
gruberb: Fixed both 👍

bendk (approved):

PR #6392, opened 2024-09-23
fix: remove empty variable

gruberb:
Script breaks with unset variable.


automation/kotlin-components-docs/generate_docs.sh
```diff
@@ -31,4 +31,4 @@ $CARGO uniffi-bindgen generate --library "$LIBRARY_FILE" --language kotlin --out
 
 # Generate documentation with increased memory
 (cd "$WORKING_DIR" && "$REPO_ROOT/gradlew" --max-workers=2 dokkaHtml -Dorg.gradle.vfs.watch=false)
-echo "Documentation generated in $DOC_DIR/build/dokka directory"
+echo "Documentation successfullu generated"
```
bendk, line 34:
```suggestion
echo "Documentation successfully generated"
```

bendk (approved):

PR #6394, opened 2024-09-24
fix: add relevancy and suggest to docs (and workspace members)

gruberb:
The `relevancy` and `suggest` component were not part of the `default` workspace members, and therefore also not part of the `rust-docs`.


bendk (approved):
:ship:

PR #6395, opened 2024-09-24
Add topic guide for the suggest component

gruberb:
Add a new sub-section in the `md-book` called `topic-guides`. The idea is to explain how the Rust components are used for each language, and best practices on how to use them.


gruberb:
It's very bare bones. One of the reason is that I never implemented the `SuggestStore` in Swift. So any feedback from @linabutler and @bendk is highly appreciated :)

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
+# Suggest
+
+The API for the `SuggestStore` can be found in the [MozillaComponents Kotlin documentation](https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.suggest/-suggest-store/index.html).
+
+> The `SuggestStore` is a synchronous, which needs to be wrapped in the asynchronous primitive of the target language you are using it in.
+
+## Setting up the store
+
+You need to import the following primitives to work with the `SuggestStore` (these come from the generated `suggest.swift` file, produced by `uniffi`):
```
bendk, line 9: Maybe "you need to import one or more of the following primitives".  In some cases consumers won't need to import them all.

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+import mozilla.appservices.remotesettings.RemoteSettingsServer
+import mozilla.appservices.suggest.SuggestApiException
+import mozilla.appservices.suggest.SuggestIngestionConstraints
+import mozilla.appservices.suggest.SuggestStore
+import mozilla.appservices.suggest.SuggestStoreBuilder
+import mozilla.appservices.suggest.Suggestion
+import mozilla.appservices.suggest.SuggestionQuery
+```
+
+On start up of your application, you create a `SuggestStore` (as a singleton). You do this via the `SuggestStoreBuilder`, which returns a `SuggestStore`:
```
bendk, line 21:
```suggestion
Create a `SuggestStore` as a singleton. You do this via the `SuggestStoreBuilder`, which returns a `SuggestStore`.   No IO or network requests are performed during construction, which makes this safe to do at any point in the application startup:
```

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+
+```kotlin
+internal val store: Lazy<SuggestStore> = lazy {
```
bendk, line 24: There's no need for the `lazy` on the consumer size because we lazily load the database nowadays (https://github.com/mozilla/application-services/blob/b10c1319e74041b2777e07657fc83dc879f841f4/components/suggest/src/store.rs#L317).

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+    SuggestStoreBuilder()
+        .dataPath(context.getDatabasePath(DATABASE_NAME).absolutePath)
+        .remoteSettingsServer(remoteSettingsServer)
+        .build()
+```
+
+In addition to the `dataPath`, which is the path (the SQLite location) where you store your suggestions, you need a `remoteSettingsServer`, which is the URL to the instance of `RemoteSettings` you will be using (e.g. `https://remote-settings-dev.allizom.org`)
```
bendk, line 31: The `remoteSettingsServer` is only needed if you want to use something other than `prod`.  They pass in a `RemoteSettingsServer` object rather than a raw URL (https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.remotesettings/-remote-settings-server/index.html).

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Ingesting suggestions
+
+We have to handle three different phases of the ingesting suggestions: The initial one on starting up the application, periodically in the background, and aborting an ongoing ingestion.
```
bendk, line 35: I personally don't consider aborting a phase, maybe you could leave that out here and just discuss it at the bottom.

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+
+Ingest with `SuggestIngestionConstraints(emptyOnly=false)` on regular schedule (like once a day).
+
+Example:
+
+```kotlin
+store.value.ingest(SuggestIngestionConstraints(emptyOnly = false))
+```
+
+The defintion of `SuggestIngestionConstraints`:
```
bendk, line 54: I'd rather this be more example-focused.   How about saying something like:
  - Starting with "SuggestIngestionConstraints" is used to control what gets ingested (link to the API docs)
  - Use the `providers` field to limit ingestion by provider type (show an example where you're only ingesting a couple providers.   `AMP_MOBILE`, `WIKIPEDIA`, and `WEATHER` are a reasonable combination.
  - Use the `providerConstraints` field to add additional constraints, currently this is only used for exposure suggestions.  (I don't have a great example for this field. @0c0w3 do you have any suggestions?)

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
...
+     */
+    var `emptyOnly`: kotlin.Boolean,
+) {
+    companion object
+}
+```
+
+## Querying Suggestions
+
+Call `SuggestStore::query` to fetch suggestions for the suggest bar:
```
bendk, line 75: `providerConstraints` should be the same value that gets passed to `ingest()`

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,97 @@
+# Suggest
+
+The API for the `SuggestStore` can be found in the [MozillaComponents Kotlin documentation](https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.suggest/-suggest-store/index.html).
+
+> The `SuggestStore` is a synchronous, which needs to be wrapped in the asynchronous primitive of the target language you are using it in.
```
bendk, line 5: I'd normally want this to be expanded, but I think it's okay for now since we plan to update our async handling in the near future.

docs/topic-guides/swift/suggest.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Ingesting suggestions
+
+We ensure that the Rust networking stack has been initialized via:
+
+```Swift
+import class MozillaAppServices.Viaduct
+
+Viaduct.shared.useReqwestBackend()
+```
```
bendk, line 46: Viaduct initialization is required for all platforms and is required for multiple components.  How about:
 - Creating a viaduct section for this.
 - Any components that require viaduct can say "make sure you initialize viaduct" and link to the section.
 - Expanding this description to say that there are 3 supported options:
    - Any `libxul` based can ignore initialization, since it's handled by `libxul`
    - Using the reqwest backend, which uses the `reqwest` library and a reqwest-managed thread.
    - Implementing the C FFI like `libxul` does (https://searchfox.org/mozilla-central/source/toolkit/components/viaduct).

Does this seem correct @linabutler ?

docs/topic-guides/swift/suggest.md
```diff
@@ -0,0 +1,109 @@
...
+## Querying Suggestions
+
+Call `SuggestStore::query` to fetch suggestions for the suggest bar:
+
+```swift
+try self.store.query(query: SuggestionQuery(
+    keyword: keyword,
+    providers: providers,
+    limit: limit
+)).compactMap(RustFirefoxSuggestion.init)
```
bendk, line 96: I don't think the `compactMap` is required it's something firefox-ios does in order to use its own class for suggestions.  I'm not sure if this is something we should recommend or not so maybe leave it out for now.

bendk (changes requested):
This is a good start.  I added most of my suggestions suggestions to the Kotlin section, some of those also apply to Swift.

docs/topic-guides/kotlin/suggest.md
gruberb: > The remoteSettingsServer is only needed if you want to use something other than prod.

I don't quite understand. I assumed we always pass the object, just with a different Url?

bendk: I think it's optional and you could skip the `.remoteSettingsServer(remoteSettingsServer)` call.  In practice though, consumers will probably always pass it so maybe it's not worth mentioning.

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,85 @@
...
+* The `remoteSettingsServer` is only needed if you want to set the server to anything else but `prod`. If so, you pass a `RemoteSettingsServer` object. 
+
+## Ingesting suggestions
+
+We have to handle different phases of the ingesting suggestions: The initial one on starting up the application, and then, periodically, in the background.
+
+* [`SuggestIngestionConstraints`](https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.suggest/-suggest-ingestion-constraints/index.html?query=data%20class%20SuggestIngestionConstraints(var%20providers:%20List%3CSuggestionProvider%3E?%20=%20null,%20var%20providerConstraints:%20SuggestionProviderConstraints?%20=%20null,%20var%20emptyOnly:%20Boolean%20=%20false)) is used to control what gets ingested.
+* Use the `providers` field to limit ingestion by provider type (show an example where you're only ingesting a couple providers. `AMP_MOBILE`, `WIKIPEDIA`, and `WEATHER` are a reasonable combination.
+* Use the `providerConstraints` field to add additional constraints, currently this is only used for exposing suggestions.
```
bendk, line 42:
```suggestion
* Use the `providerConstraints` field to add additional constraints, currently this is only used for exposure suggestions.
```

"exposure suggestions" are a new suggestion type that the SnG team is introducing.  They're going to be used to measure if a user would be exposed to a suggestion type if it was introduced with a given set of keywords.  The user actually won't be exposed to them, it's all handled internally.

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,85 @@
...
+        .remoteSettingsServer(remoteSettingsServer)
+        .build()
+```
+
+* You need to set the `dataPath`, which is the path (the SQLite location) where you store your suggestions.
+* The `remoteSettingsServer` is only needed if you want to set the server to anything else but `prod`. If so, you pass a `RemoteSettingsServer` object. 
+
+## Ingesting suggestions
+
+We have to handle different phases of the ingesting suggestions: The initial one on starting up the application, and then, periodically, in the background.
```
bendk, line 38:
```suggestion
Ingesting suggestions happens in two different ways: On startup, and then, periodically, in the background.
```

Here's my attempt to wordsmith this one, I'm open to other wordings as well.  The main things I'm concerned about are that "we" could be a bit confusing since there are multiple teams involved and also "phases" could imply that the application only needs to do one initial `emptyOnly=true` ingestion, but it should do it every startup to handle upgrades.  The reason for this is that a lot of times when we change the DB schema, we also throw away all the suggestions because it's easier than doing a real migration (and maybe we can't migrate the suggestions without data from remote settings).

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,85 @@
...
+
+* [`SuggestIngestionConstraints`](https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.suggest/-suggest-ingestion-constraints/index.html?query=data%20class%20SuggestIngestionConstraints(var%20providers:%20List%3CSuggestionProvider%3E?%20=%20null,%20var%20providerConstraints:%20SuggestionProviderConstraints?%20=%20null,%20var%20emptyOnly:%20Boolean%20=%20false)) is used to control what gets ingested.
+* Use the `providers` field to limit ingestion by provider type (show an example where you're only ingesting a couple providers. `AMP_MOBILE`, `WIKIPEDIA`, and `WEATHER` are a reasonable combination.
```
bendk, line 41: the "show an example" part was for you :)  How about updating the examples below to add the parameter: `providers=listOf(SuggestionProvider.AMP_MOBILE, SuggestionProvider.WIKIPEDIA, SuggestionProvider.WEATHER)`

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,85 @@
...
+
+Example:
+
+```kotlin
+store.value.ingest(SuggestIngestionConstraints(emptyOnly = false))
+```
+
+## Querying Suggestions
+
+Call `SuggestStore::query` to fetch suggestions for the suggest bar. The `providerConstraints` should be the same value that got passed to `ingest()`.
```
bendk, line 63:
```suggestion
Call `SuggestStore::query` to fetch suggestions for the suggest bar. The `providers` parameter should be the same value that got passed to `ingest()`.
```

bendk (changes requested):
Couple more changes needed, but this is looking great overall.  Again, I only focused on the Kotlin section, but some of the same comments apply to Swift.

bendk:
I've been trying to think through when we should say "Kotlin/Swift" vs "Android/iOS".  I'm still not totally sure, but my current feeling is that these guides should say "Kotlin/Swift", since there's nothing specificic to Android/iOS.

However, when we talk about the megazords, we should say "Android/iOS", since those are built for those particular platforms.

Does this make sense?

docs/topic-guides/kotlin/suggest.md
gruberb: Oh my, this was a copy/paste mistake!!

docs/topic-guides/kotlin/suggest.md
```diff
@@ -0,0 +1,85 @@
...
+* [`SuggestIngestionConstraints`](https://mozilla.github.io/application-services/kotlin/kotlin-components-docs/mozilla.appservices.suggest/-suggest-ingestion-constraints/index.html?query=data%20class%20SuggestIngestionConstraints(var%20providers:%20List%3CSuggestionProvider%3E?%20=%20null,%20var%20providerConstraints:%20SuggestionProviderConstraints?%20=%20null,%20var%20emptyOnly:%20Boolean%20=%20false)) is used to control what gets ingested.
+* Use the `providers` field to limit ingestion by provider type.
+* Use the `providerConstraints` field to add additional constraints, currently this is only used for exposure suggestions.
+
+### On Start Up
+Ingest with `SuggestIngestionConstraints(emptyOnly=true)` shortly after each startup. This ensures we have something in the DB on the first run and also after upgrades where we often will clear the DB to start from scratch.
+
+```kotlin
+store.value.ingest(SuggestIngestionConstraints(emptyOnly = true))
+```
```
bendk, line 49: Is it worth it to duplicate the `providers` field from below here?  That could reinforce the idea that they should be the same for both calls.

bendk (approved):
Looks great, I just had a minor comment.  @linabutler did you want to take a pass at this?

docs/topic-guides/kotlin/suggest.md
gruberb: Added the `providers` field to both Swift and Kotlin here as well 👍

gruberb:
> I've been trying to think through when we should say "Kotlin/Swift" vs "Android/iOS". I'm still not totally sure, but my current feeling is that these guides should say "Kotlin/Swift", since there's nothing specificic to Android/iOS.

> However, when we talk about the megazords, we should say "Android/iOS", since those are built for those particular platforms.

I agree. The guides are currently written very language specific, and not mentioning any platform. I think once we mention `megazord`, we refer to the platforms.

PR #744, opened 2019-03-11
Publish Android artifacts on Maven

eoger:
Fixes https://github.com/mozilla/application-services/issues/252

PR #6400, opened 2024-09-27
Bug 1921532 - Update symbol upload scripts [ci full]

bendk:
Copied the latest script from moz-central:
https://searchfox.org/mozilla-central/rev/837f3a1ff2622c8303750c35d84bdc41a5cd079c/toolkit/crashreporter/tools/symbolstore.py

  - Note:

### DIFF
--- a/automation/symbols-generation/symbolstore.py
+++ b/automation/symbols-generation/symbolstore.py
@@ -116,6 +116,13 @@ def read_output(*args):
     (stdout, _) = subprocess.Popen(args=args, stdout=subprocess.PIPE).communicate()
     return stdout.decode("utf-8").rstrip()
 
+def get_version(srcdirs):
+    for dir in srcdirs:
+        path_try = os.path.join(dir, "version.txt")
+        if os.path.exists(path_try):
+            return open(path_try).read().strip()
+    return "Unknown"
+
 class GitHubRepoInfo:
     """
     Info about a locally cloned Git repository that has its "origin" remote on GitHub.
@@ -384,7 +391,15 @@ class Dumper:
         Get the commandline used to invoke dump_syms.
         '''
         # The Mac dumper overrides this.
-        return [self.dump_syms, file]
+        return [self.dump_syms, "--inlines", file] + self.dump_syms_extra_info()
+
+    def dump_syms_extra_info(self):
+        return [
+            "--extra-info",
+            "VERSION {}".format(get_version(self.srcdirs)),
+            "--extra-info",
+            "PRODUCTNAME ApplicationServices",
+        ]
 
     def ProcessFileWork(self, file, arch_num, arch, vcs_root, dsymbundle=None):
         t_start = time.time()
@@ -403,7 +418,7 @@ class Dumper:
                 # MODULE os cpu guid debug_file
                 (guid, debug_file) = (module_line.split())[3:5]
                 # strip off .pdb extensions, and append .sym
-                sym_file = re.sub("\.pdb$", "", debug_file) + ".sym"
+                sym_file = re.sub(r"\.pdb$", "", debug_file) + ".sym"
                 # we do want forward slashes here
                 rel_path = os.path.join(debug_file,
                                         guid,
@@ -639,7 +654,16 @@ class Dumper_Mac(Dumper):
         # in order to dump all the symbols.
         if dsymbundle:
             # This is the .dSYM bundle.
-            return [self.dump_syms] + arch.split() + ['-g', dsymbundle, file]
+            cmdline = [self.dump_syms]
+            cmdline.extend(arch.split())
+            cmdline.extend([
+                "--inlines",
+                "-j",
+                "2",
+            ])
+            cmdline.extend(self.dump_syms_extra_info())
+            cmdline.extend([dsymbundle, file])
+            return cmdline
         return Dumper.dump_syms_cmdline(self, file, arch)
 
     def GenerateDSYM(self, file):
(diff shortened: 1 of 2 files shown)


bendk:
I didn't update `dump_syms` because when I search [tooltool](https://tooltool.mozilla-releng.net/static/index.html) this seems to be the latest version.  Is there a newer version we can use?

rvandermeulen:
dump_syms is a taskcluster toolchain artifact these days:
https://searchfox.org/mozilla-central/source/taskcluster/kinds/toolchain/dump-syms.yml

mstange:
Thank you, that was fast!

There have been two commits to this file on the application-services side, from PR #744 and PR #1453:
https://github.com/mozilla/application-services/commit/84e077d1534dc287bbd472da658ce22eea5af032#diff-787f740bf8b1eb70e517c7c6441fb96645d3db9359f0017532ceaecf8cf92c3c
https://github.com/mozilla/application-services/commit/0db58fd40547abfe2f957437adc7930d0181a220

Do we need to reapply those? Probably hard to say without doing a full test run.

I do believe we need to update dump_syms, because the old version will probably complain if it's called with `--inlines`.

mstange:
Sorry Ben, I don't actually know much about how this code and the python modules all work. That's unfortunate that copying the file wasn't enough to resolve this :(

dsymutil+otool are only used when compiling for macOS.
What platforms is libmegazord compiled for? Is it just Android? Or maybe also iOS?

bendk:
> dsymutil+otool are only used when compiling for macOS. What platforms is libmegazord compiled for? Is it just Android? Or maybe also iOS?

`libmegazord` is compiled for Android.  `libmegazord_ios` and `libmegazord_focus` are used on iOS.  It would be good to get symbols for those as well, but that can be a second pass.

mstange:
So I think there are two options:

 1. Either we can treat application-services' copy of symbolstore.py as a true fork, and manually apply the changes that happened to the m-c version of the file, or at least a subset of the changes that we care about.
 2. Or we do the necessary work so that the files can stay in sync. I don't have concrete proposals for how to do that.

Maybe we can do 1 for now, and leave 2 for the future. If there's a future where application-services becomes part of mozilla-central, then 2 would become easier at that point.

Is there anyone who might have opinions on the approach?

bendk:
1. makes sense to me as well.  What are the changes we care about?  I see [this commit](https://hg.mozilla.org/mozilla-central/rev/ff3a983aeae1) to add the `--inlines` argument, are there others?

mstange:
The `--inlines` change is the one I care the most about. And it requires an updated dump_syms version.

Additionally, the following would be nice to have, but are much less urgent:

 - `INFO VENDOR` and `INFO VERSION` fields (like those added by [bug 1870823](https://bugzilla.mozilla.org/show_bug.cgi?id=1870823))
 - More reliable error reporting, i.e. the equivalent of [bug 1576748](https://bugzilla.mozilla.org/show_bug.cgi?id=1576748) and [bug 1759833](https://bugzilla.mozilla.org/show_bug.cgi?id=1759833).

bendk:
I tried pulling in the `--inlines` change as well as the one that adds `PRODUCTNAME` and `VERSION`.  I figure the error reporting is probably least important.  Let's see what happens when this runs in CI.

bendk:
@mstange Looks like the new script worked.  Here's the command line it ran:
```
[task 2024-10-07T20:33:43.612Z] /builds/worker/checkouts/vcs/automation/symbols-generation/bin/dump_syms --inlines megazords/full/android/build/rustJniLibs/android/armeabi-v7a/libmegazord.so --extra-info VERSION 133.0a1 --extra-info PRODUCTNAME ApplicationServices
```

Here's a [link to the tarball](https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/JcQwRhJjTzy6nY9UZ5gT8w/runs/0/artifacts/public%2Fbuild%2Fcrashreporter-symbols.tar.gz).  It seems like it worked because I see files there, not sure if there's a better check.

mstange:
I looked at the tarball and it looks like everything worked perfectly. Thank you!

automation/symbols-generation/symbolstore.py
```diff
@@ -403,7 +418,7 @@ def ProcessFileWork(self, file, arch_num, arch, vcs_root, dsymbundle=None):
                 # MODULE os cpu guid debug_file
                 (guid, debug_file) = (module_line.split())[3:5]
                 # strip off .pdb extensions, and append .sym
-                sym_file = re.sub("\.pdb$", "", debug_file) + ".sym"
+                sym_file = re.sub(".pdb$", "", debug_file) + ".sym"
```
mstange, line 421: What's the purpose of this change? I think the previous line was correct; we want to turn "hello.pdb" into "hello", but leave "helloxpdb" as-is. In any case, it makes no difference to us here because libmegazord isn't used on Windows, and pdb files are only used on Windows.

bendk: My Python linter reported "unsupported escape character in string".

After thinking about this more, I went the wrong way.  I should have added an extra `\` or maybe better changed this to a raw string.  Once this is merged, I'll open a followup to fix this.

bendk: Oops, didn't realize that this wasn't approved yet.  I made that fix in this PR instead.

mstange (approved):
Looks great to me! Thanks again!

mhammond (approved):

rvandermeulen:
@mergify backport release-v132

PR #6405, opened 2024-10-02
Add topic guide for relevancy

gruberb:
Adding a topic guide for Swift and Kotlin for the relevancy component. I also fixed a linking issue for the `viaduct` component.

### DIFF
--- a/docs/topic-guides/kotlin/relevancy.md
+++ b/docs/topic-guides/kotlin/relevancy.md
@@ -1,0 +1,75 @@
+# Relevancy
+
+The `relevancy` component tracks the user's interests locally, without sharing any data over the network. The component currently supports building an interest vector based on the URLs they visit.
+
+## Setting up the store
+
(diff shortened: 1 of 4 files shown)


docs/topic-guides/kotlin/relevancy.md
```diff
@@ -0,0 +1,75 @@
+# Relevancy
+
+The `relevancy` component enables tracking the user's interests and building an interest vector based on the URLs they visit.
```
bendk, line 3: Couple thoughts here:
  - How about something like "tracks the user's interests locally, without sharing any data over the network".  This privacy feature is a key part of the component.
  - Maybe prefix "building an interest vector", with "the component currently supports".  The goal is to add more ways to track interests in the future.

bendk (approved):
I just had a couple suggestions for the intro section, other than that it all look good.  This is a new component so there's not that much to say about it at this point.  I'm excited to see how this doc evolves over time.

PR #6406, opened 2024-10-06
Bug 1921124 - Implement city-based weather suggestions and GeoNames

0c0w3:
This does a number of things to support weather queries that include cities and regions. Sorry for how big it is. I could split it up but I think it's more helpful to see everything in context.

### GeoNames

I added `geoname.rs` to support ingesting and querying [GeoNames](https://www.geonames.org/) ([docs here](https://download.geonames.org/export/dump/readme.txt)). This is how we detect cities and regions, and how we map cities to regions when a query contains only a city. Potentially we could use it for other suggestion types.

I wanted to mirror the raw geonames tables (the ones in the geonames dataset, not the ones in our DB) pretty much exactly rather than trying to overfit them into something specific to weather suggestions. I think that will serve us better long term.

The code makes some assumptions about the geonames data we store in remote settings. I tried to document them. And as long as weather and Suggest are US only, we'll only upload US geonames to RS. We can also scale up or down the number of geonames in RS. For example, we might use some significantly large population as a threshold.

### Weather suggestion matching

Weather suggestions are matched on any of the following:

* A weather keyword. `min_keyword_length` in the weather RS data controls whether prefix matching is allowed. This is pretty much the same as before except we were relying on desktop to do the prefix threshold check, and I moved that here.
* A full city name (or abbreviation like "nyc" -- anything that's exactly equal to a geoname alternate name)
* A full city name + any prefix of its region

After any of the above are typed, you can continue to type any prefix of any of the others and that will match too. Once you add a word that isn't recognized, matching stops.

### `util.rs`

I added a query parsing helper called `filter_map_chunks()`. I renamed `keyword.rs` to `util.rs` and added it there. Hopefully its doc explains what it does, and it would be nice if it could be useful for other providers in the future.

### Caching

I added a couple of cache structs, `WeatherCache` and `GeonameCache`, as an optimization to avoid fetching common values from the DB that are used to handle every weather query.

### Keyword metrics

I added some tables to store keyword and geoname metrics: max keyword length and max word count. Weather query handling uses them to quickly discard search strings that are too long so we don't end up with any performance problems trying to parse huge strings that people sometimes paste into the urlbar. We could simply hardcode a const (like `yelp.rs` does), but I think it's nice to do this.

### Providers and record types

Providers can now map to more than one record type. This lets us ingest geonames when weather is ingested.

  - Note:


components/suggest/src/util.rs
```diff
@@ -0,0 +1,1578 @@
...
+///     vec!["a1", "b c1"],
+///     vec!["a b0", "c0"],
+///     vec!["a b0", "c1"],
+///     vec!["a b1", "c0"],
+///     vec!["a b1", "c1"],
+///     vec!["a b c0"],
+///     vec!["a b c1"]
+/// ]);
+/// ```
+pub fn filter_map_chunks<T: Clone>(
```
bendk, line 152: I'm going to need some time to wrap my head around this one, but I have some initial thoughts

1. Can you document what "chunk" means in this context?

2. Let me propose a simpler solution, which is probably wrong, but hopefully knowing why will help me understand this better.  It seems like the requirement is to match a optional city, region, and/or weather keyword out of the input query, and the tricky part is that the keyword might be multiple-words long and you don't know which order the city/region/weather keyword will come.  Could this helper function iterate through the words and pass the string from joining the all the words that haven't been mapped to a Some value yes?  So for the "a b c" case it would:

 
- Call `mapper("a")`
  - If mapper returns `Some(_)`, then call `mapper("b")`
      - If mapper returns `Some(_)`, then call `mapper("c")`
      - If the mapper returns `None`, then call `mapper("b c")`
   - If mapper returns `None`, then call `mapper("a b")`
      - If mapper returns `Some(_)`, then call `mapper("c")`
      - If the mapper returns `None`, then call `mapper("a b c")`

For weather, the the mapper function would try to match the string to a `Token` value.

This seems simpler to understand and less complexity.  Which weather queries would it miss?

components/suggest/src/util.rs
```diff
@@ -0,0 +1,1578 @@
...
+
+    for chunk_size in 1..=max_chunk_size {
+        if words.len() < chunk_size {
+            // `chunk_size` and the later chunk sizes in this for-loop are too
+            // big to visit the remaining words. We already visited them earlier
+            // in the loop when the chunk size was small enough.
+            break;
+        }
+
+        let chunk = words[..chunk_size].join(" ");
```
bendk, line 180: There's probably a way to do this with using `&str` objects to avoid allocating a string for each join.

bendk (commented):
Most of this seems very straightforward and looks fine to me.  Then I get to the query parsing/matching part and my head explodes :)

components/suggest/src/schema.rs
```diff
@@ -444,6 +480,48 @@ CREATE INDEX exposure_custom_details_type ON exposure_custom_details(type);
...
+            26 => {
+                // Create tables related to city-based weather.
+                tx.execute_batch(
+                    "
+CREATE TABLE keywords_metrics(
+    record_id TEXT NOT NULL PRIMARY KEY,
+    provider INTEGER NOT NULL,
+    max_length INTEGER NOT NULL,
+    max_word_count INTEGER NOT NULL
+) WITHOUT ROWID;
```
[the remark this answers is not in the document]
0c0w3: For now, I only need metrics for weather keywords in this table, and we'll upload new weather keywords before this feature goes live, which will populate the table. The weather provider is the only one that even updates the table right now, although any provider that uses the `keywords` table could, if metrics would be helpful to it. Would it be all right to leave it as is and add a comment about all that?

components/suggest/src/util.rs
```diff
@@ -0,0 +1,1578 @@
...
+        .unwrap_or_else(|| {
+            keywords
+                .iter()
+                .map(AsRef::as_ref)
+                .filter(|phrase| phrase.starts_with(query) && query.len() < phrase.len())
+                .max_by_key(|phrase| phrase.trim().len())
+                .unwrap_or(query)
+                .to_owned()
+        })
+}
```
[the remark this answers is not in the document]
0c0w3: Sorry, I did `git mv keyword.rs util.rs` to keep blame for `full_keyword()`, but in the end I guess the file ended up changing so much git wasn't able to track the rename, even though `full_keyword()` didn't change at all. (I think I did `git mv` and not just `mv`, maybe I should double check. Not sure it makes a difference.)

components/suggest/src/util.rs
0c0w3: > 1. Can you document what "chunk" means in this context?

Sure. To explain here too, it's "chunk" as in a contiguous non-overlapping sequence of items in a slice, as used in a bunch of slice and `Iterator` method names, especially https://doc.rust-lang.org/std/vec/struct.Vec.html#method.chunks

I kind of assumed "chunk" is a well known term in Rust at least w/r/t to strings and slices, but I'm not sure.

> 2. Let me propose a simpler solution

That's exactly what this function does (recursively), except it has the added complexity of allowing `mapper()` to return multiple matches and then it recurses over those matches. That's what the inner `for value in mapped_values` does. Without that for-loop, the function would behave like your example.

The reason that complexity is valuable is that a chunk can match more than one token. "New York" can be the city or the state. "Waterloo" can be Waterloo, IA or Waterloo, AL. Further, not all combinations of tokens are valid. If the query string is "waterloo al", then `[<Waterloo IA>, <AL>]` is not a valid token combination. (I'm using `<foo>` to mean a token, and `[<foo>, <bar>]` to mean a list of tokens.)

So regardless of the particular implementation, there are two steps inherent in this parsing process: (1) Map chunks to tokens, and (2) reject invalid token combinations. There's more than one way to do it, but the implementation I've chosen here is elegant IMO.

Here's an example parsing the query string "waterloo al" where the mapper only returns an `Option<Token>`, similar to your example:

1. Call `mapper("waterloo")`
2. `mapper()` returns `Some(<Waterloo IA>)`
3. Call `mapper("al")`
4. `mapper()` returns `Some(<AL>)`

We end up with tokens `[<Waterloo IA>, <AL>]`. We need a post-processing step that figures out that this combination is invalid, and then we need to start again and somehow tell the mapper to return different tokens if it can.

This example builds on that and lets the mapper return multiple tokens, but it still doesn't recurse on each mapped token:

1. Call `mapper("waterloo")`
2. `mapper()` returns `[<Waterloo IA>, <Waterloo AL>]`
3. Call `mapper("al")`
4. `mapper()` returns `[<AL>]`

We end up with `[[<Waterloo IA>, <Waterloo AL>], <AL>]`. This is better because now we can figure out valid combinations and we don't need to start again. But it still requires a post-processing step that's separate from the token parsing itself. I actually implemented this at one point, but due to this post-processing step, it was uglier than what I ended up with.

Here's what my patch does:

1. Call `mapper("waterloo", [])`
2. `mapper()` returns `Some([<Waterloo IA>, <Waterloo AL>])`
3. Call `mapper("al", [<Waterloo IA>])`
4. `mapper()` finds that "al" maps to `<AL>` but it returns `None` because `<AL>` doesn't match `<Waterloo IA>`
5. Call `mapper("al", [<Waterloo AL>])`
6. `mapper()` returns `Some([<Waterloo AL>, <AL>])`

And that's it, we're done.

components/suggest/src/util.rs
0c0w3: Hmm that might be doable. We could pass around the original query string instead of or probably in addition to the `words` slice, and then get the candidate chunk as a slice of the original string. I'll try that.

components/suggest/src/geoname.rs
```diff
@@ -0,0 +1,1069 @@
...
+    Other,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Geoname {
+    pub geoname_id: i64,
+    pub geoname_type: GeonameType,
+    pub name: String,
+    pub country_code: String,
+    pub admin1_code: String,
```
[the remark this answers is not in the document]
0c0w3: I copied/moved comments from `DownloadedGeoname` to here in `Geoname`.

components/suggest/src/schema.rs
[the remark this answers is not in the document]
0c0w3: I added a comment above this table.

components/suggest/src/util.rs
0c0w3: I document what "chunk" means and also added some comments to the recursive helper.

components/suggest/src/util.rs
0c0w3: That worked. I was also able to avoid the `path.clone()` by passing around `&mut path` and popping mapped values from it after recursing.

0c0w3:
The commit I just pushed expects alternate names to be stored inline in the main geonames attachments instead of in separate records/attachments. I'm working on the uploader script and most geonames only have one or three alternate names. There's no reason not to store them inline like we do for AMP keywords for example. The uploader script needs to access the data it keeps for a geoname when it processes its alternate names anyway, so might as well add them directly to that data.

I modified the `geonames_alternates` schema accordingly: We can generate our own alternate-names IDs instead of using the ones from the original GeoNames data. This table is pretty similar to the `keywords` table now. I also removed `record_id` from the table since nothing used it. I must have added it thinking it would be necessary to remove rows whose records were deleted, but the table accomplishes that through the `ON DELETE`.

components/suggest/src/util.rs
```diff
@@ -0,0 +1,1517 @@
...
+            // in the loop when the chunk size was small enough.
+            break;
+        }
+
+        // Get the current chunk within the remaining query. Its char length is
+        // the sum of the lengths of the words in the chunk + `chunk_size - 1`
+        // spaces between the words.
+        let chunk_char_len = remaining_words[..chunk_size]
+            .iter()
+            .fold(chunk_size - 1, |memo, w| memo + w.len());
```
bendk, line 207: I think this would this break if there were multiple spaces between words, but it's not since you normalize the whitespace in the parent function.  Couple thoughts on this:

- Add a docstring saying why this is safe.
- Add a test for multiple spaces below (if it's not there already, sorry if I missed it)
- Maybe `remaining_words` could be replaced with `remaining_word_positions: &[(usize, usize)]` where each item is the start/end position of the word in the original query.  That would avoid issues even if you didn't normalize.  If this adds complexity, it's not with it though.

components/suggest/src/weather.rs
```diff
@@ -132,94 +439,416 @@ mod tests {
...
+                        region: Some("AL".to_string()),
+                        score: 0.24,
+                    },
+                ],
+            ),
+            ("waterloo foo", vec![]),
+            ("waterloo weather foo", vec![]),
+            ("foo waterloo", vec![]),
+            ("foo waterloo weather", vec![]),
+        ];
```
bendk, line 759: I love these tests, can you add some with weird whitespace?
```
"new     york weather"
"    new york weather"
"new york weather    "
```

components/suggest/src/util.rs
```diff
@@ -0,0 +1,1517 @@
...
+                        chunk_index + chunk_size,
+                        max_chunk_size,
+                        f,
+                    )?;
+                    path.pop();
+                    for mut p in subtree_paths {
+                        p.insert(0, value.clone());
+                        this_step_paths.push(p);
+                    }
+                }
```
bendk, line 234: Suggestion: use `get()` to let Rust do the bounds check for you and remove the comment about it.
```
match remaining_words.get(chunk_size..) {
   None => {
      // final chunk
   }
   Some(remaining_words) => {
       // Recurse
   }
}
```

This could also be an `if let`.

bendk (approved):
Still not sure what the best way is to express your chunked string search, but we can figure that out later.  The paranoid side of me says that some of the bounds checking could be improved, but I'll approve and let you decide how best to handle it.

components/suggest/src/weather.rs
```diff
@@ -1,53 +1,235 @@
...
+                    }
+                }
+            }
+            // This chunk doesn't match. Return `None` to filter out the path.
+            Ok(None)
+        })?;
+
+        // `paths` contains all matching paths, and if a path has both a city
+        // and region, the city is in the region since we filtered out city-
+        // region mismatches above via `match_weather_tokens()`.
```
bendk, line 114: Okay, I do have one suggestion.  What if the `filter_map_chunks` only allowed the mapper to return an `Option` rather than a `Vec`?  I think that makes that function much simpler to describe and understand, although though the implementation basically stays the same.

That pushes the complexity out to this code, but I don't think that's too hard to handle, we just need a second dose of recursion.  The rest of this function could be replaced with something like this:
```
    // no changes before here
    let mut suggestions = vec![];
    collect_suggestions_recurse(&mut suggestions, &paths, &mut Vec::new());
    Ok(suggestions);
}

fn collect_suggestions_recurse(
    suggestions: &mut Vec<Suggestion>,
    remaining: &[Token],
    current: &mut Vec<Token>)
{
    // check if the previous code generated an mismatched series of tokens (City:NewYorkCity +
    // Region:CA)
    if tokens_mismatched(tokens) {
        return;
    }
    match remaining.split_first() {
        None => {
            if let Some(s) = tokens_to_suggestion(current) {
                suggestions.push(s)
            }
        }
        Some((first, remaining)) => {
            current.push(first);
            collect_suggestions_recurse(suggestions, remaining, current);
            current.pop();
            // Token::alternate_tokens generates a Vec of other ways to interpret the token.  For
            // example, when we parse "New York" in `filter_map_chunks` we would parse it as
            // `Token::City("New York)`, but then alternative_tokens would generate
            // `Token::Region("New York")`.  If there's only 1 possible alternate, then it
            // could return an option instead of a vec.
            for alt in first.alternate_tokens() {
                current.push(alt);
                collect_suggestions_recurse(suggestions, remaining, current);
                current.pop();
            }
        }
    }
}
```


I think my main issue conceptually is there are 2 sources of branching: different chunk combinations and the different ways to parse strings into tokens.  IMO, it would be simpler if those 2 things were split out into separate processes.

bendk: However, I really meant it when I said don't block this PR on this issue.  If you disagree then feel free to merge first and we can discuss if this refactoring makes sense or not in a separate one.

components/suggest/src/weather.rs
```diff
@@ -1,53 +1,235 @@
...
+                TokenType::City,
+                TokenType::Region,
+                TokenType::WeatherKeyword,
+            ] {
+                if !path.iter().any(|t| t.token_type() == tt) {
+                    let tokens = self.match_weather_tokens(tt, path, chunk, chunk_index == 0)?;
+                    if !tokens.is_empty() {
+                        // This chunk matches, so return the tokens to continue
+                        // down this path.
+                        return Ok(Some(tokens));
```
bendk, line 104: I just realized that this short-circuits the logic and doesn't allow for matching other token types.  This means that `rochester ny weather` will work, but `ny rochester weather` won't because `ny` is always matched as a city only and not a region.

bendk:
Like I mentioned in Slack, this PR has slightly broken my brain and forced me to come up with the nicest way to implement this depth-first search that I can.  [Here's what I've got](https://github.com/mozilla/application-services/compare/adw/suggest-geonames-and-city-weather...bendk:application-services:push-vyvtqzxukrnx).  I don't think we need a generic depth-first-search implementation, because all the generic code is very easy to implement.  Almost all the complexity is in the specifics.  If we want to re-use this approach for another suggestion type, it wouldn't be hard to start from scratch and follow the same pattern.

I like this approach better because the code flow is simple and direct.  WDYT?

components/suggest/src/util.rs
```diff
@@ -0,0 +1,1517 @@
...
+/// ]);
+/// ```
+pub fn filter_map_chunks<T: Clone>(
+    query: &str,
+    max_chunk_size: usize,
+    f: impl Fn(&str, usize, &[T]) -> Result<Option<Vec<T>>>,
+) -> Result<Vec<Vec<T>>> {
+    let words: Vec<_> = query.split_whitespace().filter(|s| !s.is_empty()).collect();
+    let normalized_query = words.join(" ");
+    filter_map_chunks_recurse(
```
ncloudioj, line 164: We might want to check the input before recursively processing it.

I tested the following with a few inputs (w/ debug build), looks like the run time increase exponentially and eventually overflew the stack for a large N (e.g. N=250):
```rs
        const N: usize = 5;
        let paths = filter_map_chunks(
            "a b c d e".repeat(N).as_ref(),
            3,
            |chunk, chunk_index, _| Ok(Some(vec![(chunk.to_string(), chunk_index)])),
        )?;
``` 
N = 5; run time = 0.8 secs
N = 6; run time = 7.6 secs
N = 7; run time = 108 secs
N = 8; run time >= 27 mins
...
N = 260; stack overflow

If the input query has too many tokens, we could return early.

0c0w3: Thanks Nan, I appreciate it. The docstring talks about that, about returning `None` to halt traversal: "This is important for keeping traversal depth reasonably bounded." It should probably be called out more prominently, and it should mention branching factor too. The exponential behavior is inherent to this kind of tree traversal (as you know), and I'm assuming we'll be careful about short-circuiting it. Bugs and mistakes happen, of course.

I don't think the function should necessarily bail if the input is arbitrarily large. Searching for a token that's N words long requires only O(N) steps. This example uses a query string with 10k words with a max chunk size of 10k and searches for the entire query string. It takes ~450ms on my machine (debug, just using `time` on the command line when running the test):

```rust
const WORD_COUNT: usize = 10_000;
let s = (0..WORD_COUNT).map(|i| i.to_string()).collect::<Vec<_>>().join(" ");
// Sanity check the query len and word count
assert_eq!(s.len(), (1 * 10) + (2 * 90) + (3 * 900) + (4 * 9000) + (WORD_COUNT - 1));
assert_eq!(s.split_whitespace().collect::<Vec<_>>().len(), WORD_COUNT);
println!("Calling");
let paths = filter_map_chunks(&s, WORD_COUNT, |chunk, _, _| {
    if chunk == s {
        Ok(Some(vec![chunk.to_string()]))
    } else {
        Ok(None)
    }
})?;
println!("Done calling");
assert_eq!(paths, vec![vec![s]]);
```

Changing the example to search for "0" (the first word in the query) instead of the whole query string causes *more* steps to be traversed: 2N steps, since the function will traverse the remainder of the string after the "0". Searching for both "0" and "1" takes ~3N steps, and so on. Still O(N), but getting worse [1]. But keep in mind that N = the number of words in the query string. In practice, we would never use a max chunk size that arbitrarily large. In reality, the number of calls is O(T * M), where T = the number of tokens being searched for and M = a reasonable max chunk size. The number of steps is not a function of the query string's length or word count at all. At least it shouldn't be. If it is, we goofed. For weather suggestions, T = 3 and M = ~7 (I don't remember off hand, something like 7).

We might say we shouldn't add a function that has the kind of danger you pointed out. We might not use it right, and mistakes happen. That's true, but I don't know how else we can do the kind of matching we want to do without placing restrictions on what the query string can look like.

[1] This could be fixed by giving the mapper function a way to short-circuit (break out of) the outer for-loop in the recursive function. I'm not sure it's worth the added complexity, especially after all the discussion in this PR, but I'm definitely not against it.

components/suggest/src/weather.rs
0c0w3: You're right, really nice catch! It's an error. I'm working on a fix.

components/suggest/src/util.rs
0c0w3: This doesn't quite work because at this point the chunk isn't ever out of bounds, due to the earlier `break` in the outer for-loop.

ncloudioj (approved):
This looks good to me. r=nanj.

I agree that this is a challenging and interesting problem to tackle and the current implementation does require some extra effort to comprehend. We can try simplifying it as followups later on.

0c0w3:
Summary of latest commit:

* Fix the Rochester problem
* Add tests with long queries/city names
* Address review comments as discussed

components/suggest/src/weather.rs
0c0w3: This is fixed in the latest commit.

components/suggest/src/util.rs
0c0w3: `filter_map_chunks_2()` had a query with extra spaces, but I moved it to its own test function so it's more obvious. I also added a comment.

ncloudioj:
My last comment :)

I was trying to mirror what we're doing here via Elasticsearch, I think it can be roughly reformulated as follows:

```json
{
  "analysis": {
    "tokenizer": "whitespace",
    "filter": [
      {
        "type": "shingle",
        "min_shingle_size": 1,
        "max_shigle_size": MAX_CHUNK_SIZE
      },
      {
        "type": "synonym",
        "synonyms": [
          "nyc => new york city NY",
          "la => los angeles CA",
          ...
          "waterloo => waterloo AL",
          "waterloo => waterloo IL",
          ...
        ]
      }
    ]
  }
}
```

- First, we use the "whitespace" tokenizer to split the query input into words (i.e. token stream)
- Then we apply two token filters to the result token stream:
  - A "shingle" token filter [1], which concatenates adjacent tokens into "shingles" (or word n-grams)
  - Then for each shingled token, a "synonym" token filter checks if there are synonyms associated with it, if so expand it

Use "nyc weather" as an example:
- The "whitespace" tokenizer will produce a token stream of ["nyc", "weather"]
- The "shingle" token filter will produce a new token stream of ["nyc weather", "nyc", "weather"]
- The "synonym" token filter will further expand it to ["nyv weather", "new york city NY", "weather"]

While this is an over-simplified rewrite compared to the original one, I think it helps me understand things and also the terminology is worth borrowing as they're well known in the world of ES/Lucene.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-shingle-tokenfilter.html
[2]: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html

bendk (commented):
The changes look great, let's merge this one.

0c0w3:
Thanks @bendk and @ncloudioj for all your help.

PR #6408, opened 2024-10-08
Update remote settings errors

bendk:
Use fields inside the error enum, rather than the UniFFI "flat" errors which ignore the inner data.  This is prep for moving to proc-macros.

Switched to using error_support and define both internal and public error types.

Switch app-services consumers to using the `RemoteSettings` type rather than `Client`.  This way they're using the same public API as everyone else.  This required adding a couple more pub RemoteSettings methods that I didn't expose via UniFFI.  I didn't feel like adding new features to this client, because I'm hoping to replace it (#6378).

  - Note:

### DIFF
--- a/components/nimbus/src/stateful/client/http_client.rs
+++ b/components/nimbus/src/stateful/client/http_client.rs
@@ -16,9 +16,9 @@
 use crate::error::Result;
 use crate::schema::parse_experiments;
 use crate::stateful::client::{Experiment, SettingsClient};
-use remote_settings::Client;
+use remote_settings::RemoteSettings;
 
-impl SettingsClient for Client {
+impl SettingsClient for RemoteSettings {
     fn get_experiments_metadata(&self) -> Result<String> {
         unimplemented!();
     }
--- a/components/remote_settings/src/config.rs
+++ b/components/remote_settings/src/config.rs
@@ -10,7 +10,7 @@
 
 use url::Url;
 
-use crate::Result;
+use crate::{ApiResult, Error, Result};
 
 /// Custom configuration for the client.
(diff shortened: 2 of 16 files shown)


components/nimbus/src/stateful/client/http_client.rs
```diff
@@ -16,9 +16,9 @@
 use crate::error::Result;
 use crate::schema::parse_experiments;
 use crate::stateful::client::{Experiment, SettingsClient};
-use remote_settings::Client;
+use remote_settings::RemoteSettings;
 
-impl SettingsClient for Client {
+impl SettingsClient for RemoteSettings {
```
bendk, line 21: These nimbus changes seem pretty harmless, but please double-check my work here.

components/remote_settings/src/config.rs
```diff
@@ -36,7 +36,17 @@ pub enum RemoteSettingsServer {
...
+    /// Get the [url::Url] for this server
+    #[error_support::handle_error(Error)]
+    pub fn url(&self) -> ApiResult<Url> {
+        self.get_url()
+    }
+
+    /// Internal version of `url()`.
+    ///
+    /// The difference is that it uses `Error` instead of `ApiError`.  This is what we need to use
+    /// inside the crate.
```
bendk, line 48: This is one annoying part about splitting up the errors into internal vs external errors.  `url()` needs to return an `ApiResult` since it's pub and used by consumers (nimbus), but then `Client::new` can't use it since it returns a `Result<>` type.

ncloudioj (approved):
LGTM, r=nanj.

PR #6411, opened 2024-10-09
Bug 1923689 - Implement basic configuration processing on SearchEngineSelector.

Standard8:
This implements processing the main parts of the search configuration JSON into the Rust structure using serde_json. Currently the processing applies to:

* The identifier and base properties of the engine records.
* The default engine records.

As variant and environment handling is not yet implemented, the filter function will return all engines defined in the configuration.

These functions are not currently used, hence no breaking API changes, similarly with change log.

  - Note:

### DIFF
--- a/components/search/src/configuration_types.rs
+++ b/components/search/src/configuration_types.rs
@@ -1,0 +1,135 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! This module defines the structures that we use for serde_json to parse
+//! the search configuration.
+
+use crate::{SearchEngineClassification, SearchUrlParam};
+use serde::Deserialize;
+
+/// The list of possible submission methods for search engine urls.
+#[derive(Debug, uniffi::Enum, PartialEq, Deserialize, Clone, Default)]
+#[serde(rename_all = "UPPERCASE")]
+pub(crate) enum JSONEngineMethod {
+    Post = 2,
+    #[serde(other)]
+    #[default]
+    Get = 1,
+}
+
+impl JSONEngineMethod {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            JSONEngineMethod::Get => "GET",
+            JSONEngineMethod::Post => "POST",
+        }
+    }
+}
+
+/// Defines an individual search engine URL. This is defined separately to
+/// `types::SearchEngineUrl` as various fields may be optional in the supplied
+/// configuration.
+#[derive(Debug, uniffi::Record, PartialEq, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONEngineUrl {
+    /// The PrePath and FilePath of the URL. May include variables for engines
+    /// which have a variable FilePath, e.g. `{searchTerm}` for when a search
+    /// term is within the path of the url.
+    pub base: String,
+
+    /// The HTTP method to use to send the request (`GET` or `POST`).
+    /// If the engine definition has not specified the method, it defaults to GET.
+    pub method: Option<JSONEngineMethod>,
+
+    /// The parameters for this URL.
+    pub params: Option<Vec<SearchUrlParam>>,
+
+    /// The name of the query parameter for the search term. Automatically
+    /// appended to the end of the query. This may be skipped if `{searchTerm}`
+    /// is included in the base.
+    pub search_term_param_name: Option<String>,
+}
+
+/// Reflects `types::SearchEngineUrls`, but using `EngineUrl`.
+#[derive(Debug, uniffi::Record, PartialEq, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONEngineUrls {
+    /// The URL to use for searches.
+    pub search: JSONEngineUrl,
+
+    /// The URL to use for suggestions.
+    pub suggestions: Option<JSONEngineUrl>,
+
+    /// The URL to use for trending suggestions.
+    pub trending: Option<JSONEngineUrl>,
+}
+
+/// Represents the engine base section of the configuration.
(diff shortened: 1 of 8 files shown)


components/search/src/filter.rs
```diff
@@ -0,0 +1,205 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::{
+    error::Error, JSONEngineBase, JSONEngineRecord, JSONEngineUrl, JSONEngineUrls,
+    JSONSearchConfigurationRecords, RefinedSearchConfig, SearchEngineDefinition, SearchEngineUrl,
+    SearchEngineUrls, SearchUserEnvironment,
+};
+
+impl SearchEngineUrl {
+    pub(crate) fn from_engine_url(url: &JSONEngineUrl) -> Self {
```
Standard8, line 12: I'm not sure if this is the best way of handling these conversions. There appears to be multiple ways, and this seemed to make sense?

components/search/src/types.rs
```diff
@@ -165,8 +174,8 @@ pub struct SearchEngineDefinition {
     pub name: String,
 
     /// The partner code for the engine. This will be inserted into parameters
-    /// which include `{partnerCode}`.
-    pub partner_code: Option<String>,
+    /// which include `{partnerCode}`. May be the empty string.
+    pub partner_code: String,
```
Standard8, line 178: I made this change, as I think we'd generally replace `{partnerCode}` with the string, whether it is empty or not.

linabutler (changes requested):
This looks wonderful, @Standard8; it's great to see this coming together!

I left a lot of comments—some are style nits; some are more substantive thoughts on the implementation—but they're all meant to start conversations, not a prescriptive "let's do it this way".

I'm happy to answer more questions, and drop links to the Rust book and examples if you'd like; and also, please feel free to push back on any suggestions that don't make sense.

I'm also really excited that we're adding another component to our collection. Rust! Rust! Rust! 🦀

components/search/src/configuration_types.rs
```diff
@@ -0,0 +1,134 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::{SearchEngineClassification, SearchUrlParam};
+use serde::Deserialize;
+
+/// This module defines the structures that we use for serde_json to parse
+/// the search configuration.
+
+/// The list of possible submission methods for search engine urls.
+#[derive(Debug, uniffi::Enum, PartialEq, Deserialize, Clone)]
```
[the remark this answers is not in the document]
Standard8: I don't think I need this, it seems to work without that.

Standard8:
@linabutler Thank you for all the excellent comments. I definitely like the idea of cloning up-front. Although it feels like cloning a large amount when we might use just a small amount, I agree it is better that way, so that we can reduce the individual allocations & make the code simpler.

components/search/src/configuration_types.rs
```diff
@@ -0,0 +1,134 @@
...
+pub(crate) struct JSONEngineOrdersRecord {
+    // TODO: Implementation.
+}
+
+/// Represents an individual record in the raw search configuration.
+#[derive(Debug, Deserialize)]
+#[serde(tag = "recordType", rename_all = "camelCase")]
+pub(crate) enum JSONSearchConfigurationRecords {
+    DefaultEngines(JSONDefaultEnginesRecord),
+    Engine(Box<JSONEngineRecord>),
```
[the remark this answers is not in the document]
linabutler: Oh, I see! Nope, this is perfect; thanks! I didn't realize that the difference in sizes of the variants was so large.

components/search/src/filter.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            JSONSearchConfigurationRecords::Unknown => {
+                // Prevents panics if a new record type is added in future.
+            }
+        }
+    }
+
+    Ok(RefinedSearchConfig {
+        engines,
+        app_default_engine_id: default_engine_id.unwrap(),
```
linabutler, line 80: Is this `.unwrap()` temporary? If not, I'm wondering if we should think about returning an error here instead—`.unwrap()` will immediately `abort()` the main process on Desktop.

components/search/src/filter.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    for record in configuration {
+        match record {
+            JSONSearchConfigurationRecords::Engine(engine) => {
+                let result = extract_engine_config(&user_environment, engine);
+                engines.extend(result);
+            }
+            JSONSearchConfigurationRecords::DefaultEngines(default_engines) => {
+                default_engine_id = Some(default_engines.global_default);
+                default_private_engine_id.clone_from(&default_engines.global_default_private);
```
linabutler, line 67: Oooh, I haven't seen a use of `.clone_from()` in the wild before! Did Clippy recommend this?

components/search/src/selector.rs
```diff
@@ -2,20 +2,31 @@
...
 #[derive(Default, uniffi::Object)]
-pub struct SearchEngineSelector {}
+pub struct SearchEngineSelector(Mutex<SearchEngineSelectorInner>);
 
 #[uniffi::export]
 impl SearchEngineSelector {
     #[uniffi::constructor]
     pub fn new() -> Self {
-        Self::default()
+        Self(Mutex::new(SearchEngineSelectorInner::default()))
```
linabutler, line 29:
```suggestion
        Self(Mutex::default())
```

(`Mutex<T>` implements `Default` if `T` implements `Default`—but your code looks good as-is, too, if you'd like to keep it explicit!)

linabutler (approved):
This looks great to me, thank you so much for folding in all my feedback! :shipit:

components/search/src/filter.rs
Standard8: Yeah, it is temporary. I'm actually working on it in a patch for full handling of default engines, and changing this to be an `Option<String>` - in the worst case that we can't determine a default engine, we'll let the app decide how to handle it. As that's in a patch already, I won't change that here.

PR #6417, opened 2024-10-18
Starting new Remote Settings API

bendk:
Implemented client functionality and started the general structure for the types.  The big item left todo is storage.

Added a CLI to test it, you can run it using `cargo rs`.

  - Note:


bendk:
This new API is still a WIP, but I think we still could merge this since no consumers are using it yet.  I tested it with the `cargo rs` CLI and it seemed to work.

@leplatrem I added you as a reviewer mostly to make sure that the client code is hitting the right endpoints and following the client specs correctly.

components/remote_settings/src/client.rs
```diff
@@ -18,6 +19,205 @@ const HEADER_BACKOFF: &str = "Backoff";
...
 
+/// Internal Remote settings client API
+pub struct RemoteSettingsClient {
+    // This is immutable, so it can be outside the mutex
+    collection_name: String,
+    inner: Mutex<RemoteSettingsClientInner>,
+}
+
+struct RemoteSettingsClientInner {
+    storage: Storage,
```
bendk, line 30: I believe that `Storage` will want to use `&mut` in a lot of its methods, since that's what `rusqlite::Connection::transaction` uses and in general we want exclusive access to the connection.  But I'm not totally sure this is right and maybe we should move it outside the mutex.

components/remote_settings/src/client.rs
```diff
@@ -18,6 +19,205 @@ const HEADER_BACKOFF: &str = "Backoff";
...
+
+    pub fn update_config(&self, server_url: Url, bucket_name: String) -> Result<()> {
+        let mut inner = self.inner.lock();
+        inner.api_client = ApiClient::new(server_url, &bucket_name, &self.collection_name)?;
+        inner.storage.empty()
+    }
+}
+
+/// Client for Remote settings API requests
+struct ApiClient {
```
bendk, line 94: I'm never sure if I should write unit tests for these kinds of classes.  On the one hand, unit tests are great.  On the other hand, testing this would involve a bunch of mocking and I often feel like the tests end up mirroring the implementation in a way that doesn't actually test anything.  What do others think?

Another option would be e2e tests.  The `cargo rs` client provides a very rudimentary way to kind of do that.  Maybe we could extend that into a real e2e test suite.

components/remote_settings/src/storage.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// Set the list of records stored in the database, clearing out any previously stored records
+    pub fn set_records(&self, _base_url: &str, records: &[RemoteSettingsRecord]) -> Result<()> {
+        for record in records {
+            println!("Should store record: {record:?}");
+        }
+        Ok(())
+    }
+
+    pub fn store_attachment(
```
gruberb, line 70: Adding doc comments here as well?

components/remote_settings/src/storage.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// Get cached records for this collection
+    ///
+    /// Returns None if no records are stored or if `base_url` does not match the `base_url` passed
+    /// to `set_records`.
+    pub fn get_records(&self, _base_url: &str) -> Result<Option<Vec<RemoteSettingsRecord>>> {
+        Ok(None)
+    }
+
+    pub fn get_attachment(
```
gruberb, line 54: Adding doc comments here as well?

components/remote_settings/src/storage.rs
```diff
@@ -0,0 +1,86 @@
...
+///   - Still, there might be contention if there are multiple clients for the same collection, or
+///     if RemoteSettingsService::sync() and RemoteSettingsClient::get_records(true) are called at
+///     the same time.  Maybe we should create a single write connection and put it behind a mutex
+///     to avoid the possibility of SQLITE_BUSY.  Or maybe not, the writes seem like they should be
+///     very fast.
+///   - Maybe we should refactor this to use the DAO pattern like suggest does.
+pub struct Storage {}
+
+impl Storage {
+    pub fn new(_path: Utf8PathBuf) -> Result<Self> {
```
gruberb, line 34: Adding doc comments here as well?

components/remote_settings/src/service.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        for client in self.inner.lock().active_clients() {
+            if synced_collections.insert(client.collection_name()) {
+                client.internal.sync()?;
+            }
+        }
+        Ok(synced_collections.into_iter().collect())
+    }
+
+    pub fn update_config(&self, config: RemoteSettingsConfig2) -> Result<()> {
```
gruberb, line 84: Adding doc comments here as well?

components/remote_settings/src/client.rs
gruberb: I think we could spin up a basic test server which returns a JSON, and then test if we can store and fetch records? I am not sure if we even need a e2e test suite for this (could spin up a server, and shut it down via a `oneshot` channel after the test finish for example. But maybe we want something more...fancy.

components/remote_settings/src/client.rs
```diff
@@ -18,6 +19,205 @@ const HEADER_BACKOFF: &str = "Backoff";
...
+                api_client,
+            }),
+        })
+    }
+
+    pub fn collection_name(&self) -> &str {
+        &self.collection_name
+    }
+
+    pub fn get_records(&self, sync_if_empty: bool) -> Result<Option<Vec<RemoteSettingsRecord>>> {
```
gruberb, line 55: Would be good to comment here as well. Especially to explain `sync_if_empty`.

gruberb (approved):
Just left a few comments (to add doc comments :)). But looks very good to me 👍

components/remote_settings/src/storage.rs
bendk: I'm adding comments for the other methods, but I'm not sure what to say here other than "this creates a new storage".  I'll leave it to you to decide the details on what's happening.

components/remote_settings/src/client.rs
bendk: I guess I disagree with myself, because I ended up writing some mock tests.  I'm still not sure how useful they are but it's not all that much code, so I figured it was worth it.

components/remote_settings/src/client.rs
```diff
@@ -18,6 +19,205 @@ const HEADER_BACKOFF: &str = "Backoff";
...
+    /// This is something like
+    /// `https://[server-url]/v1/buckets/[bucket-name]/collections/[collection-name]/"
+    base_url: Url,
+    remote_state: RemoteState,
+}
+
+impl ApiClient {
+    fn new(server_url: Url, bucket_name: &str, collection_name: &str) -> Result<Self> {
+        let base_url = server_url.join(&format!(
+            "v1/buckets/{bucket_name}/collections/{collection_name}/"
```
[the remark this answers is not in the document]
bendk: I wanted something that would change when the bucket changed, so that we could use this URL to test if the cached data was still valid.  I should probably use a different name for this though, maybe `collection_url`?  What's your feeling on this one?

.cargo/config.toml
```diff
@@ -9,4 +9,5 @@ suggest-bench = ["bench", "-p", "suggest", "--features", "benchmark_api"]
 suggest-debug-ingestion-sizes = ["run", "-p", "suggest", "--bin", "debug_ingestion_sizes", "--features", "benchmark_api"]
 relevancy = ["run", "-p", "examples-relevancy-cli", "--"]
 suggest = ["run", "-p", "examples-suggest-cli", "--"]
+rs = ["run", "-p", "examples-remote-settings-cli", "--"]
```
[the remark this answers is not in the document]
bendk: I was being a bit lazy there, let's spell it out.

components/remote_settings/src/client.rs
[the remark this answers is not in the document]
bendk: I went ahead and renamed this to `collection_url`.  Hopefully that makes sense, but I'm open to just storing the `base_url` to match how other implementations do it.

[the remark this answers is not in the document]
leplatrem: `collection_url` works.
However, please put the `v1` as part of the server URL.
(eg. https://github.com/mozilla-extensions/remote-settings-devtools/blob/15245fdb8e4fb5875a7866c7328ded6671ab466c/extension/experiments/remotesettings/api.js#L8-L11)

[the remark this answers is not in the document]
bendk: Sounds good, but let me do this in a followup.  It's a small change, but not trivial since the URL is part of the public API.

PR #6429, opened 2024-10-24
Rework remote settings server URL

bendk:
Let's update these to match the JS client:
https://github.com/mozilla-extensions/remote-settings-devtools/blob/15245fdb8e4fb5875a7866c7328ded6671ab466c/extension/experiments/remotesettings/api.js#L8-L11

This is a breaking change to the public API, but this API is not exposed via UniFFI and it's only used in 1 place:
https://github.com/mozilla/application-services/blob/50bf235b9a3a62a6b6749cd0bd7358d7b96a42e4/components/nimbus/src/stateful/client/mod.rs#L24-L35. AFAICT, the only affect this will have is that the error message will change slightly.

Also removed the `.unwrap()` calls.  It shouldn't matter because there's no way for `.join()` to fail, but it feels silly calling `unwrap()` from a function that returns a result.

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -122,25 +122,14 @@ pub trait ApiClient {
 
 /// Client for Remote settings API requests
 pub struct ViaductApiClient {
-    /// Base URL for requests to a collections endpoint
-    ///
-    /// This is something like
-    /// `https://[server-url]/v1/buckets/[bucket-name]/collections/[collection-name]/"
-    ///
-    /// Note: this is different than the `base_url` used for other client implementations (
-    /// (`https://[server-url]/v1).  The main reason to use the collection_url is that we can use
-    /// it to check if we need to invalidate the cached data stored in the [Storage] layer.
-    collection_url: Url,
+    endpoints: RemoteSettingsEndpoints,
     remote_state: RemoteState,
 }
 
 impl ViaductApiClient {
-    fn new(server_url: Url, bucket_name: &str, collection_name: &str) -> Result<Self> {
-        let collection_url = server_url.join(&format!(
-            "v1/buckets/{bucket_name}/collections/{collection_name}/"
-        ))?;
+    fn new(base_url: Url, bucket_name: &str, collection_name: &str) -> Result<Self> {
         Ok(Self {
-            collection_url,
+            endpoints: RemoteSettingsEndpoints::new(&base_url, bucket_name, collection_name)?,
             remote_state: RemoteState::default(),
         })
     }
@@ -207,11 +196,11 @@ impl ViaductApiClient {
 
 impl ApiClient for ViaductApiClient {
     fn collection_url(&self) -> String {
-        self.collection_url.to_string()
+        self.endpoints.collection_url.to_string()
     }
 
     fn get_records(&mut self, timestamp: Option<u64>) -> Result<Vec<RemoteSettingsRecord>> {
-        let mut url = self.collection_url.join("changeset")?;
+        let mut url = self.endpoints.changeset_url.clone();
         // 0 is used as an arbitrary value for `_expected` because the current implementation does
         // not leverage push timestamps or polling from the monitor/changes endpoint. More
         // details:
@@ -239,8 +228,9 @@ impl ApiClient for ViaductApiClient {
         let attachments_base_url = match &self.remote_state.attachments_base_url {
             Some(attachments_base_url) => attachments_base_url.to_owned(),
             None => {
-                let collection_url = self.collection_url.clone();
-                let server_info = self.make_request(collection_url)?.json::<ServerInfo>()?;
+                let server_info = self
+                    .make_request(self.endpoints.root_url.clone())?
+                    .json::<ServerInfo>()?;
                 let attachments_base_url = match server_info.capabilities.attachments {
                     Some(capability) => Url::parse(&capability.base_url)?,
                     None => Err(Error::AttachmentsUnsupportedError)?,
@@ -257,11 +247,9 @@ impl ApiClient for ViaductApiClient {
 
 /// A simple HTTP client that can retrieve Remote Settings data using the properties by [ClientConfig].
 /// Methods defined on this will fetch data from
-/// <base_url>/v1/buckets/<bucket_name>/collections/<collection_name>/
+/// <base_url>/buckets/<bucket_name>/collections/<collection_name>/
 pub struct Client {
-    pub(crate) base_url: Url,
-    pub(crate) bucket_name: String,
-    pub(crate) collection_name: String,
+    endpoints: RemoteSettingsEndpoints,
     pub(crate) remote_state: Mutex<RemoteState>,
 }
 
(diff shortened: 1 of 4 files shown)


components/remote_settings/src/client.rs
```diff
@@ -122,24 +122,29 @@ pub trait ApiClient {
 
 /// Client for Remote settings API requests
 pub struct ViaductApiClient {
+    /// Server base URL
+    ///
+    /// This is something like `https://[domain]/v1/`.  It's what's normally called the `base_url`
```
leplatrem, line 127: Our convention is not put the trailing `/`

components/remote_settings/src/client.rs
```diff
@@ -122,24 +122,29 @@ pub trait ApiClient {
...
+    /// in other client implementations.
+    server_url: Url,
     /// Base URL for requests to a collections endpoint
     ///
     /// This is something like
-    /// `https://[server-url]/v1/buckets/[bucket-name]/collections/[collection-name]/"
+    /// `https://[server-url]/buckets/[bucket-name]/collections/[collection-name]/`
```
leplatrem, line 133: This is correct (only if no trailing slash in `server-url`)

components/remote_settings/src/client.rs
```diff
@@ -122,24 +122,29 @@ pub trait ApiClient {
...
+    /// the cached data stored in the [Storage] layer.
     collection_url: Url,
     remote_state: RemoteState,
 }
 
 impl ViaductApiClient {
     fn new(server_url: Url, bucket_name: &str, collection_name: &str) -> Result<Self> {
         let collection_url = server_url.join(&format!(
-            "v1/buckets/{bucket_name}/collections/{collection_name}/"
+            "buckets/{bucket_name}/collections/{collection_name}/"
```
leplatrem, line 144:
```suggestion
            "/buckets/{bucket_name}/collections/{collection_name}/"
```

leplatrem (changes requested):
I'm sorry if it wasn't ultra clear in other messages, but our convention is to not put the trailing `/` in the server url and build paths with `{server-url}/buckets/...`

(Note: I didn't put comments everywhere this has to be adjusted)

components/remote_settings/src/client.rs
bendk: I avoided this because it's going to break the join.  Without the trailing slash, then join treats the last component as a file rather than a directory and removes it.  However, there is the `path_segments_mut()` method.  Let me try that and see how it comes out.

components/remote_settings/src/client.rs
```diff
@@ -450,6 +436,83 @@ impl Client {
...
+
+impl RemoteSettingsEndpoints {
+    /// Construct a new RemoteSettingsEndpoints
+    ///
+    /// `base_url` should have the form `https://[domain]/v1` (no trailing slash).
+    fn new(base_url: &Url, bucket_name: &str, collection_name: &str) -> Result<Self> {
+        let mut server_url = base_url.clone();
+        // Push the empty string to add a trailing slash.  This avoids a redirect when we hit
+        // the server.
+        Self::path_segments_mut(&mut server_url)?.push("");
```
bendk, line 481: Using `https://firefox.settings.services.mozilla.com/v1/` instead of `https://firefox.settings.services.mozilla.com/v1` seems right to me.  I always try to avoid hitting a redirect when I make an API call.  But tell me if that's a bad idea.

bendk:
Your `v1/` vs `v1` suggestion pushed me to make a bunch of changes and I'm liking the result.  Encapsulating the URLs into a single class seems like the nicest way to handle this.

components/remote_settings/src/client.rs
```diff
@@ -239,8 +228,9 @@ impl ApiClient for ViaductApiClient {
         let attachments_base_url = match &self.remote_state.attachments_base_url {
             Some(attachments_base_url) => attachments_base_url.to_owned(),
             None => {
-                let collection_url = self.collection_url.clone();
-                let server_info = self.make_request(collection_url)?.json::<ServerInfo>()?;
+                let server_info = self
+                    .make_request(self.endpoints.server_url.clone())?
```
leplatrem, line 232: I would call it `root_url`

components/remote_settings/src/client.rs
```diff
@@ -450,6 +436,83 @@ impl Client {
...
 }
 
+/// Stores all the endpoints for a Remote Settings server
+///
+/// There's actually not to many of these, so we can just pack them all into a struct
+struct RemoteSettingsEndpoints {
+    /// Top-level URL for Remote Settings server
+    ///
+    /// This has the form `https://[domain]/v1/`. It's where we get the attachment base url from.
+    server_url: Url,
```
leplatrem, line 446: In other implementations, we call this the root URL and it has the form of `{server_url}/`

components/remote_settings/src/client.rs
leplatrem: There should not be any redirect, we don't call the `server_url` directly (without trailing slash)

leplatrem (approved):
This looks good, and approve to avoid blocking you longer.

Nevertheless, in order to avoid confusion between the different implementations, I would like to insist that we shouldn't put any trailing slash to the server URL.

I agree with you, it would be ugly to hit a redirect, but we don't because we consider `/` as the root endpoint. 

So I think that with your PR, if we just add a `root` endpoint to your struct, we should be aligned with the other implementations 😊 

Examples in [kinto-http.py](https://github.com/Kinto/kinto-http.py/blob/c5aeb26f61b84740768f731ea42dbf28d86d4903/src/kinto_http/endpoints.py#L10-L23):

```py

class Endpoints(object):
    endpoints = {
        "root": "{root}/",
        "batch": "{root}/batch",
        "buckets": "{root}/buckets",
        "bucket": "{root}/buckets/{bucket}",
        "history": "{root}/buckets/{bucket}/history",
        "groups": "{root}/buckets/{bucket}/groups",
        "group": "{root}/buckets/{bucket}/groups/{group}",
        "collections": "{root}/buckets/{bucket}/collections",
        "collection": "{root}/buckets/{bucket}/collections/{collection}",
        "records": "{root}/buckets/{bucket}/collections/{collection}/records",  # NOQA
        "record": "{root}/buckets/{bucket}/collections/{collection}/records/{id}",  # NOQA
    }
```

or  [kinto.js](https://github.com/Kinto/kinto.js/blob/70500a1a0024872b804ca16b5ba4fc72c00f236e/src/http/endpoints.ts#L6-L19)

```js
const ENDPOINTS = {
  root: () => "/",
  bucket: (bucket?: string) => "/buckets" + (bucket ? `/${bucket}` : ""),
  collection: (bucket: string, coll?: string) =>
    `${ENDPOINTS.bucket(bucket)}/collections` + (coll ? `/${coll}` : ""),
  record: (bucket: string, coll: string, id?: string) =>
    `${ENDPOINTS.collection(bucket, coll)}/records` + (id ? `/${id}` : ""),
    // ...
    // ...
};
```

bendk:
That makes sense.  I updated the terminology to be `root_url`.  That actually removed all references to `server_url`.

PR #6430, opened 2024-10-24
feat: add storage layer to remote-settings component

gruberb:
Adding a local storage (`sqlite` via `rusqlite`) to the `remote-settings` component.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,10 @@ commands:
       - run:
           name: Verify the build environment
           command: ./libs/verify-desktop-environment.sh
+      # Install SQLite development libraries
+      - run:
+          name: Install SQLite development libraries
+          command: sudo apt-get install libsqlite3-dev
   run-tests:
     steps:
       - run: automation/tests.py rust-tests
--- a/components/remote_settings/src/storage.rs
+++ b/components/remote_settings/src/storage.rs
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use camino::Utf8PathBuf;
+use rusqlite::{params, Connection, OptionalExtension};
+use serde_json;
 
 use crate::{Attachment, RemoteSettingsRecord, Result};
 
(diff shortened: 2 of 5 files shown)


components/remote_settings/src/storage.rs
```diff
@@ -19,37 +22,75 @@ use crate::{Attachment, RemoteSettingsRecord, Result};
...
-    pub fn get_records(&self, _collection_url: &str) -> Result<Option<Vec<RemoteSettingsRecord>>> {
-        Ok(None)
+    /// Returns None if no records are stored or if `collection_url` does not match the `collection_url` passed
+    /// to `set_records`.
+    pub fn get_records(&self, collection_url: &str) -> Result<Option<Vec<RemoteSettingsRecord>>> {
+        let conn = self.conn.lock().unwrap();
+        let mut stmt = conn.prepare("SELECT data FROM records WHERE collection_url = ?")?;
+        let mut rows = stmt.query(params![collection_url])?;
+        let mut records = Vec::new();
+        while let Some(row) = rows.next()? {
```
bendk, line 84: I usually prefer [query_and_then](https://github.com/mozilla/application-services/blob/d160a87f30a8c1c752d8323f7b42f8ad46a6b450/components/suggest/src/db.rs#L265-L270) for this kind of iteration.  I'm not sure if it's better or worse in this case, I just wanted to point it out to you.

components/remote_settings/src/storage.rs
```diff
@@ -19,37 +22,75 @@ use crate::{Attachment, RemoteSettingsRecord, Result};
...
+            let data: Vec<u8> = row.get(0)?;
+            let record: RemoteSettingsRecord = serde_json::from_slice(&data)?;
+            records.push(record);
+        }
+        if records.is_empty() {
+            Ok(None)
```
bendk, line 90: Like I mentioned above, this one feels a bit risky to me since there could be a collection with 0 records.

components/remote_settings/src/storage.rs
```diff
@@ -60,39 +101,373 @@ impl Storage {
...
+        collection_url: &str,
         records: &[RemoteSettingsRecord],
     ) -> Result<()> {
-        for record in records {
-            println!("Should store record: {record:?}");
+        let mut conn = self.conn.lock().unwrap();
+
+        let tx = conn.transaction()?;
+
+        // Delete existing records for this collection_url
```
bendk, line 131: Maybe this should delete records, regardless of collection_url.   If that changes, then I don't think there's any use in keeping around the old records.  I'm not totally sure on this one though, what's your opinion?

components/remote_settings/src/storage.rs
```diff
@@ -60,39 +101,373 @@ impl Storage {
...
+            params![collection_url],
+        )?;
+
+        // Insert new records
+        {
+            let mut stmt =
+                tx.prepare("INSERT INTO records (id, collection_url, data) VALUES (?, ?, ?)")?;
+            for record in records {
+                let data = serde_json::to_vec(record)?;
+                stmt.execute(params![record.id, collection_url, data])?;
```
bendk, line 143: Nit: I believe you could pass a tuple for this.  The params! macro is only needed for weird cases.

components/remote_settings/src/storage.rs
```diff
@@ -60,39 +101,373 @@ impl Storage {
...
+        // Get records when none are set
+        let fetched_records = storage.get_records(collection_url)?;
+        assert!(fetched_records.is_none());
+
+        // Get last modified timestamp when no records
+        let last_modified = storage.get_last_modified_timestamp(collection_url)?;
+        assert!(last_modified.is_none());
+
+        Ok(())
+    }
```
bendk, line 255: Let's add one test for storing 0 records and then calling `get_records`.  Whatever the behavior we end up deciding on, there should be a test for it.

bendk (changes requested):
Looking good.  A lot of my comments were unimportant nits based on my experience with `rusqlite`, if you don't like them feel free to keep the code how it is.

The important issues are figuring the behavior of `get_records()` with 0 records and what our mutex strategy should be.

.circleci/config.yml
```diff
@@ -145,6 +145,10 @@ commands:
       - run:
           name: Verify the build environment
           command: ./libs/verify-desktop-environment.sh
+      # Install SQLite development libraries
+      - run:
+          name: Install SQLite development libraries
+          command: sudo apt-get install libsqlite3-dev
```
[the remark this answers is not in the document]
gruberb: Also seemed weird to me. This commit was just to test if the tests would run through (and they did). I'll do some more digging to see why. It's just that `nimbus-sdk` tests fail because of missing `sqlite-dev` in the CI.

components/remote_settings/src/storage.rs
```diff
@@ -60,39 +101,373 @@ impl Storage {
...
-        _attachment: Attachment,
+        attachment: Attachment,
     ) -> Result<()> {
-        println!("Should store attachment: {attachment_id}");
+        let conn = self.conn.lock().unwrap();
+        let data = serde_json::to_vec(&attachment)?;
+        conn.execute(
+            "INSERT OR REPLACE INTO attachments (id, collection_url, data) VALUES (?, ?, ?)",
+            params![attachment_id, collection_url, data],
+        )?;
```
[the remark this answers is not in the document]
gruberb: Hm, maybe I am missing something, but where does this delete attachments for other collection URLs?

components/remote_settings/src/storage.rs
gruberb: I am not 100% sure about when this is used. Based on the signature of the function, I would assume we insert records just for this one particular `collection_url`, and therefore we decide to clear out old entries. If we clear out all other `collection_urls` as well, then... what happens if we call this method 5 times for example with different `collection_urls`? Or can one instance just have one `collection_url`?

components/remote_settings/src/storage.rs
bendk: Ahh, I'm pretty sure you're right about that.

components/remote_settings/src/storage.rs
bendk: ..and because it's kind of a hack, I'm hoping to prevent corner cases as best as we can.  It feels to me like there would be less if we deleted the old data, but I could definitely be wrong.

gruberb: Ok! What happens to the metadata though? Should we delete all previous metadata as well? Scenario (maybe unlikely but have to account for that anyway)

* Store records for `url1`
* Afterwards, store records for `url2`

State of the `db`: Just regords for `url2` because we delete all other entries.

What happens to a) the `attachements`? I assume they stay. And I introduced a `collection_metadata` where we store the `last_modified` and `fetched true/false`. Should we delete all entries for `url1` there as well?

gruberb: I updated the PR to reflect that. We delete all `records` and `collection_metadata` entries when setting new `records`.

components/remote_settings/src/storage.rs
gruberb: Done!

components/remote_settings/src/storage.rs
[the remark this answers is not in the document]
gruberb: Got it. I could create a primary key which combines the `url` and `id`. The question is: The behaviour throughout the storage now is that we delete all `collection_urls` when inserting new ones, so we cannot have the case of `url1` and `url2` both exist. Is this the case? I still wonder how this affects `attachements`.

components/remote_settings/src/storage.rs
```diff
@@ -19,37 +21,96 @@ use crate::{Attachment, RemoteSettingsRecord, Result};
...
+    /// Returns None if no records are stored or if `collection_url` does not match the `collection_url` passed
+    /// to `set_records`.
+    pub fn get_records(&self, collection_url: &str) -> Result<Option<Vec<RemoteSettingsRecord>>> {
+        // Check if the collection has been fetched before
+        let mut stmt = self
+            .conn
+            .prepare("SELECT fetched FROM collection_metadata WHERE collection_url = ?")?;
+        let fetched: Option<bool> = stmt
+            .query_row(params![collection_url], |row| row.get(0))
+            .optional()?;
```
bendk, line 86: How about using a transaction for this?  I think that means you'll get a consistent read even if another thread is calling `set_records` at the same time.  Maybe none if this matters because of the mutex, but I don't love relying on that.

components/remote_settings/src/storage.rs
```diff
@@ -19,37 +21,96 @@ use crate::{Attachment, RemoteSettingsRecord, Result};
...
+                        Ok(serde_json::from_slice(&data)?)
+                    })?
+                    .collect::<Result<Vec<_>>>()?;
+
+                if records.is_empty() {
+                    // collection_url was fetched before, but has no entries
+                    Ok(Some(Vec::new()))
+                } else {
+                    Ok(Some(records))
+                }
```
bendk, line 108: I think this `if` can be reduced to `Ok(Some(records))` now.

components/remote_settings/src/storage.rs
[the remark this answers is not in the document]
bendk: What if the above statement was changed to `DELETE FROM attachments WHERE collection_url != ?`

That way, when collection_url changes, we throw away the old attachment data on the next `set_attachment` call.  That more-or-less matches the behavior of `set_records`.  I feel like I might be missing some subtleties here, but this seems good enough for the initial implementation at least.

bendk (changes requested):
Looking really good, I think the last thing we need to figure out is the behavior of `set_attachments` when the collection URL changes.

PR #6433, opened 2024-10-25
fix: bundle sqlite [ci full]

gruberb:
Fixes https://firefox-ci-tc.services.mozilla.com/tasks/SiXHEcqgR2OrEimRQXMADQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,10 +145,6 @@ commands:
       - run:
           name: Verify the build environment
           command: ./libs/verify-desktop-environment.sh
-      # Install SQLite development libraries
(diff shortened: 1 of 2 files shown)


bendk (approved):
If the CI tests pass, I think this is ready to go.

bendk:
Wait, did the full CI tasks run?  You need to push again after the title is updated for that to happen.  I usually just do `git commit --amend` to get a new commit ID then force push it.

bendk:
When `[full ci]` takes affect, you'll see a couple dozen tasks under the  firefoxci-taskcluster heading.

PR #6439, opened 2024-10-28
Bug 1927526 - Add geographic coordinates to weather suggestions

0c0w3:
I had to remove some of the derived implementations on `Geoname` (`Eq`, `Hash`, `PartialEq`) due to the addition of these `f64`s. That had a couple of consequences: I also had to remove `Geoname::geoname_type` since nothing uses it anymore, and that meant I also had to remove `GeonameType::Other`. Equality comparison and hashing are now based on `Geoname::geoname_id`, which is probably how it should have worked all along.

  - Note:


0c0w3:
Also the coordinates are assumed to be strings in the RS data rather than floats. That seems safer to me but I'm open to changes around that.

0c0w3:
An alternative to returning coordinates in the suggestion is to allow the consumer to pass in an optional geolocation as part of the query, probably as a provider constraint, and have Rust do the distance calculation. I'm not sure which is better and I'm open to that approach too.

components/suggest/src/rs.rs
```diff
@@ -648,6 +648,22 @@ pub(crate) struct DownloadedGlobalConfigInner {
...
+    String::deserialize(deserializer).map(|s| s.parse().ok())
+}
+
+pub(crate) fn deserialize_f64_or_default<'de, D>(
+    deserializer: D,
+) -> std::result::Result<f64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(deserialize_f64(deserializer)?.unwrap_or_default())
```
bendk, line 664: Could you use `Option<f64>::deserialize(deserializer)` instead of `deserialize_f64`?  If not, maybe that function should be renamed to `deserialize_opt_f64`.  My concern is that if someone tried to use `deserialize_f64` with a regular `f64` field, it won't work (I think -- maybe I'm not understanding this right).

bendk (approved):
Looks good to me.  I like the move to use more shared data for the tests.  I'll let you decide how to handle the deserialization function.

components/suggest/src/rs.rs
0c0w3: That compiles (with a slight change to `Option::<f64>`) but fails at runtime with:

```
Err(Error("invalid type: string \"34.91814\", expected f64", line: 0, column: 0))
```

Instead of having two functions, I think I'll just combine everything into `deserialize_f64_or_default()`. I copied `deserialize_f64()` from [an older revision](https://github.com/mozilla/application-services/blob/04a9195927f3e4f7df5cc8eba1d1598bc599a5d4/components/suggest/src/rs.rs#L321) back when weather data was stored inline in the RS record and we needed to parse the optional score float. I thought it might be useful to bring back, but on second thought there's probably no need.

0c0w3:
@mergify backport release-v133

PR #6442, opened 2024-10-29
Bug 1927637 - Ignore punctuation when matching weather suggestions

0c0w3:
We should ignore punctuation when matching weather suggestions, e.g., the comma in: "weather waterloo, ia".

This modifies `fetch_weather_suggestions()` so it breaks the query string into words, removes punctuation, and passes the words into `filter_map_chunks()`. I modified `filter_map_chunks()` accordingly so it takes a slice directly instead of taking a string and then breaking it into words.

We do a fair amount of lowercasing the query string in a bunch of places, the Yelp functions also break the query into words, and we may want to remove punctuation for other suggestion types too, so I started down the path of creating a `NormalizedQuery` struct to cache all those things, but the patch ended up being much larger of course, so I might try that in a separate PR.

  - Note:

### DIFF
--- a/components/suggest/src/weather.rs
+++ b/components/suggest/src/weather.rs
@@ -76,10 +76,10 @@ impl SuggestDao<'_> {
         // into words. We want to avoid that work for strings that are so long
         // they can't possibly match. The longest possible weather query is two
         // geonames + one weather keyword + at least two spaces between those
-        // three components, say, 10 spaces total for some wiggle room. There's
-        // not much point in an analogous min length check since weather
-        // suggestions can be matched on city alone and many city names are only
-        // a few characters long ("nyc").
+        // three components, say, 10 extra characters total for spaces and
+        // punctuation. There's no point in an analogous min length check since
+        // weather suggestions can be matched on city alone and many city names
+        // are only a few characters long ("nyc").
         let g_cache = self.geoname_cache();
         let w_cache = self.weather_cache();
         let max_query_len = 2 * g_cache.max_name_length + w_cache.max_keyword_length + 10;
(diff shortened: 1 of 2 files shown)


components/suggest/src/weather.rs
```diff
@@ -89,10 +89,19 @@ impl SuggestDao<'_> {
 
         let max_chunk_size =
             std::cmp::max(g_cache.max_name_word_count, w_cache.max_keyword_word_count);
-        let kw = query.keyword.to_lowercase();
+
+        // Lowercase, strip punctuation, and split the query into words.
+        let kw_lower = query.keyword.to_lowercase();
```
ncloudioj, line 94: Nit: this seems slightly easier to read to me.

```rs
    let normalized = query
        .keyword
        .to_lowercase()
        .replace(|c| c.is_ascii_punctuation(), "");
    let words: Vec<_> = normalized.split_whitespace().collect();
```

ncloudioj (approved):
Looks good, thanks!

components/suggest/src/weather.rs
0c0w3: Thanks!

This doesn't quite work because we want (IMO) to split on punctuation as well as spaces, not collapse it. e.g., "rochester,ny" should be split into "rochester" and "ny". It does work if we do `replace(|c: char| c.is_ascii_punctuation(), " ")`, so no big deal.

I do agree my version is a little harder to understand, but it has the benefit of not allocating a new string, so I think I'd prefer to leave it if that's OK.

ncloudioj: Ah right, I guess we could replace with space, but I am fine to keep yours as well.

0c0w3:
@mergify backport release-v133

PR #6443, opened 2024-10-29
Bug 1927543 - Add an `if_page_missing` option for history metadata.

linabutler:
This commit:

* Introduces an options argument to the Places `noteHistoryMetadataObservation()` functions.
* Adds an `if_page_missing` option that specifies what to do if the observed page doesn't exist in the Places database, defaulting to ignoring the observation.

We'll use this option to only insert new pages into Places for the initial `DocumentType` observation, and ignore all other observations if the page isn't in Places.

  - Note:

### DIFF
--- a/components/places/src/places.udl
+++ b/components/places/src/places.udl
@@ -76,7 +76,7 @@ interface PlacesConnection {
     sequence<HistoryHighlight> get_history_highlights(HistoryHighlightWeights weights, i32 limit);
 
     [Throws=PlacesApiError]
-    void note_history_metadata_observation(HistoryMetadataObservation data);
+    void note_history_metadata_observation(HistoryMetadataObservation data, NoteHistoryMetadataObservationOptions options);
 
     [Throws=PlacesApiError]
     void metadata_delete(Url url, Url? referrer_url, string? search_term);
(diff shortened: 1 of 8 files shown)


linabutler:
This isn't an API-breaking change, thanks to the optional arguments, but it _is_ a breaking behavioral change. I still liked this option a little more than special-casing `DocumentTypeObservation`, but I'm open to feedback either way!

components/places/src/places.udl
```diff
@@ -76,7 +76,7 @@ interface PlacesConnection {
     sequence<HistoryHighlight> get_history_highlights(HistoryHighlightWeights weights, i32 limit);
 
     [Throws=PlacesApiError]
-    void note_history_metadata_observation(HistoryMetadataObservation data);
+    void note_history_metadata_observation(HistoryMetadataObservation data, NoteHistoryMetadataObservationOptions options);
```
linabutler, line 79: I tried to put the default argument value here, but:

1. `optional NoteHistoryMetadataObservationOptions options` didn't work—it generated source code identical to `NoteHistoryMetadataObservationOptions options` for both Swift and Kotlin.
2. `optional NoteHistoryMetadataObservationOptions options = {}` raised a "No support for literal" error.

So they're in the Kotlin and Swift wrappers instead!

mhammond:
What I don't quite understand here is why we give the apps this option? Couldn't we make this new behaviour unconditional and make the check of the doc type ourself? *Someone* is going to just hard-code this behaviour right?

linabutler:
We could, definitely! I'm thinking it would be easier for iOS and Android to have visibility into the behavior by being more explicit about it, though. If we special-case `DocumentType` observations ourselves, it's easier for it to feel out of sight, out of mind.

mhammond:
I'm thinking that *all* of this is already out-of-sight-out-of-mind for them. I'd still love to see this metadata api die entirely after we work out the (small) value it does have for mobile clients and integrate that value more seamlessly.

@csadilek, wdyt?

mhammond (approved):
Still not convinced the mobile clients actually want or need options for this, but it also isn't going to hurt, so let's just land it!

csadilek:
> @csadilek, wdyt?

My apologies, I missed the initial notifications here, but I agree it's fine to land this to fix the immediate problem. 

I still believe the metadata API has the right shape to be useful. Capturing browsing metadata locally allows us to implement many useful features in privacy-neutral ways. However, I also acknowledge that it's been 3+ years since we landed this and that the functionality is heavily underutilized. It also diverged even further from Desktop, and it doesn't appear likely that the mobile team will continue to build new features with it. It's worth discussing this with them, and if true, deprecate and ultimately remove the API.

PR #6445, opened 2024-10-29
[DISCO-3043] Relevancy: Multi Armed Bandit API

misaniwere:
Implementing init, select & update methods for multi armed bandit API (Thompson Sampling)

  - Note:

### DIFF
--- a/components/relevancy/src/db.rs
+++ b/components/relevancy/src/db.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+use crate::Error::BanditNotFound;
 use crate::{
     interest::InterestVectorKind,
     schema::RelevancyConnectionInitializer,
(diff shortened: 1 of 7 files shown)


components/relevancy/src/lib.rs
```diff
@@ -95,6 +97,101 @@ impl RelevancyStore {
...
+    /// Initializes probability distributions for any uninitialized items (arms) within a bandit model.
+    ///
+    /// This method takes a `bandit` identifier and a list of `arms` (items) and ensures that each arm
+    /// in the list has an initialized probability distribution in the database. For each arm, if the
+    /// probability distribution does not already exist, it will be created, using Beta(1,1) as default,
+    /// which represents uniform distribution
+    ///
+    /// # Parameters
+    /// - `bandit` (String): The identifier for the bandit model to which the arms belong.
+    /// - `arms` (&[String]): A list of arms (items) that need to be checked for initialization.
```
bendk, line 110: Love the detailed docstrings, but I would leave out the type annotations here since you need to list them in the function signature.

components/relevancy/src/lib.rs
```diff
@@ -95,6 +97,101 @@ impl RelevancyStore {
...
+    ///
+    /// # Returns
+    /// - `ApiResult<()>`: An empty result if all arms are successfully initialized, or an error if
+    /// any database operation fails.
+    #[handle_error(Error)]
+    pub fn bandit_init(&self, bandit: String, arms: &[String]) -> ApiResult<()> {
+        for arm in arms {
+            self.db
+                .read_write(|dao| dao.initialize_multi_armed_bandit(&bandit, arm))?;
```
bendk, line 119: How about moving the for loop inside the `read_write` closure?  I think that means we'll use a single transaction for the entire operation.

components/relevancy/src/lib.rs
```diff
@@ -95,6 +97,101 @@ impl RelevancyStore {
...
+    #[handle_error(Error)]
+    pub fn bandit_update(&self, bandit: String, arm: String, selected: bool) -> ApiResult<()> {
+        if selected {
+            self.db
+                .read_write(|dao| dao.update_bandit_arm_data(&bandit, &arm, selected))?;
+        } else {
+            // update beta (failure) distribution (+1) for item
+            self.db
+                .read_write(|dao| dao.update_bandit_arm_data(&bandit, &arm, selected))?;
+        }
```
bendk, line 191: Now that `update_bandit_arm_data()` inputs `selected`, I don't think there's any need for this `if`.

components/relevancy/src/schema.rs
```diff
@@ -30,6 +30,15 @@ pub const SQL: &str = "
         count INTEGER NOT NULL,
         PRIMARY KEY (kind, interest_code)
     ) WITHOUT ROWID;
+    CREATE TABLE multi_armed_bandit(
+        bandit TEXT NOT NULL,
+        arm TEXT NOT NULL,
+        alpha INTEGER NOT NULL,
+        beta INTEGER NOT NULL,
+        clicks INTEGER NOT NULL,
+        impressions INTEGER NOT NULL,
```
bendk, line 39: Right now, I think  `clicks` and `impressions` can always be calculated from `alpha` and `beta`, but I still like storing these in a separate column since they could diverge in the future.  For example, I'm still hoping that there could be a way to pass CTR guesses to the init function that would initialize `alpha` and `beta` to something other than 1, 1.  In that case, these columns wouldn't be redundant and it would be nice to track the actual CTR separately.

bendk (changes requested):
This looks great, I love that we're getting ahead of schedule for the 2025 R2D2 work  Basically all my comments were about style issues.  Feel free to ignore any that you disagree with.

Couple more details:
  - Please add a `CHANGELOG` entry under a `What's new` header.
  - Can you rebase/squash this to 1 commit before merging?  It's not that big of a deal, but it makes the git history easier to understand.

components/relevancy/src/lib.rs
misaniwere: good catch!

components/relevancy/src/db.rs
```diff
@@ -172,11 +173,91 @@ impl<'a> RelevancyDao<'a> {
...
+
+    /// Updates the Beta distribution parameters and counters for a specific arm in a bandit model based on user interaction.
+    ///
+    /// This method updates the `alpha` or `beta` parameters in the `multi_armed_bandit` table for the specified
+    /// `bandit` and `arm` based on whether the arm was selected by the user. If `selected` is true, it increments
+    /// both the `alpha` (indicating success) and the `clicks` and `impressions` counters. If `selected` is false,
+    /// it increments `beta` (indicating failure) and only the `impressions` counter. This approach adjusts the
+    /// distribution parameters to reflect the arm's performance over time.
+    pub fn update_bandit_arm_data(&self, bandit: &str, arm: &str, selected: bool) -> Result<()> {
+        // update the beta distribution for given bandit and arm with alpha & beta
```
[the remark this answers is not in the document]
misaniwere: oops missed that

components/relevancy/src/db.rs
```diff
@@ -172,11 +173,91 @@ impl<'a> RelevancyDao<'a> {
...
     }
+
+    /// Initializes a multi-armed bandit record in the database for a specific bandit and arm.
+    ///
+    /// This method inserts a new record into the `multi_armed_bandit` table with default probability
+    /// distribution parameters (`alpha` and `beta` set to 1) and usage counters (`impressions` and
+    /// `clicks` set to 0) for the specified `bandit` and `arm`. If a record for this bandit-arm pair
+    /// already exists, the insertion is ignored, preserving the existing data.
+    pub fn initialize_multi_armed_bandit(&mut self, bandit: &str, arm: &str) -> Result<()> {
+        let bandit_data = (bandit, arm, 1, 1, 0, 0);
```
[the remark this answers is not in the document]
misaniwere: yeah i agree

ncloudioj:
Just skimmed through the patch, overall it looks great! First patch in Rust, right? Hope you've enjoyed it :)

One note I'd like to share for future consideration, in most of our use cases, the bandit updates would come influx, for instance, multiple suggestion candidates for Suggest for each keystroke in the URL bar; Multiple recommended stories on the New Tab page. Updating those individually would require a large number of DB writes.

Alternatively, we could do it in batch to mitigate that. Specifically, updates can be stashed in memory first and flush back into DB in bulk later. Prior to that, all the bandit arm selection will be determined by the DB counters and the in-memory updates (or only by the former if combining the two is not easy).

This is not a blocker for this PR and can be considered for future iterations.

misaniwere:
> Just skimmed through the patch, overall it looks great! First patch in Rust, right? Hope you've enjoyed it :)
> 
> One note I'd like to share for future consideration, in most of our use cases, the bandit updates would come influx, for instance, multiple suggestion candidates for Suggest for each keystroke in the URL bar; Multiple recommended stories on the New Tab page. Updating those individually would require a large number of DB writes.
> 
> Alternatively, we could do it in batch to mitigate that. Specifically, updates can be stashed in memory first and flush back into DB in bulk later. Prior to that, all the bandit arm selection will be determined by the DB counters and the in-memory updates (or only by the former if combining the two is not easy).
> 
> This is not a blocker for this PR and can be considered for future iterations.

haha yes it is, this was a fun one! I totally agree and there's still some followup work to be done for performance. caching and performing batch updates was along the lines of what I was also considering

bendk (approved):
This looks great, thanks for all the changes and let's land this!

PR #6448, opened 2024-10-30
Refactored `sendToDevice` function to return the result of the close tab call

lougeniaC64:
Refactored the iOS `sendToDevice` function to return `CloseTabsResult` when a close tab command is called which will assist in setting the right commands as sent.

  - Note:

### DIFF
--- a/components/fxa-client/ios/FxAClient/FxAccountDeviceConstellation.swift
+++ b/components/fxa-client/ios/FxAClient/FxAccountDeviceConstellation.swift
@@ -13,6 +13,11 @@ public struct ConstellationState {
     public let remoteDevices: [Device]
 }
 
+public enum SendEventError: Error {
+    case tabsNotClosed(urls: [String])
+    case other(Error)
+}
+
 public class DeviceConstellation {
     var constellationState: ConstellationState?
     let account: PersistedFirefoxAccount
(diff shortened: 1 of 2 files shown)


components/fxa-client/ios/FxAClient/FxAccountDeviceConstellation.swift
```diff
@@ -87,18 +92,29 @@ public class DeviceConstellation {
     }
 
     /// Send an event to another device such as Send Tab.
-    public func sendEventToDevice(targetDeviceId: String, e: DeviceEventOutgoing) {
+    public func sendEventToDevice(targetDeviceId: String,
+                                  _ e: DeviceEventOutgoing,
```
linabutler, line 96:
```suggestion
                                  e: DeviceEventOutgoing,
```

I'm so sorry, @lougeniaC64; I realized this would be a breaking change 🤦🏼‍♀️ I thought it would be OK because we'd have to take a breaking change for a non-optional completion handler, anyway; but now that the completion handler is optional, removing the label would be the only breaking change.

linabutler (approved):
Thank you so much for making those changes, @lougeniaC64; this looks great! (And so sorry for the churn about the `e` label; I didn't think that one all the way through!)

lougeniaC64:
> Thank you so much for making those changes, @lougeniaC64; this looks great! (And so sorry for the churn about the `e` label; I didn't think that one all the way through!)

No worries, thanks for the quick feedback! Definitely think this was a worthwhile improvement.

components/fxa-client/ios/FxAClient/FxAccountDeviceConstellation.swift
lougeniaC64: Yeah, I was wondering about that but I kinda deferred to you 😬 🫠 . I think I'm going to revert this label change just to be safe.

PR #6449, opened 2024-10-30
feat: add jexl to remote-settings component

gruberb:
Adds `jexl_filtering.rs` to the `remote-settings` component. 

For this to work, we need to make the `Evaluator` `Sync + Send`. Waiting for this PR to get merged https://github.com/mozilla/jexl-rs/pull/33 so we can merge this one.

The advantage is that we can instantiate the `Evaluator` once, and share it across threads.


components/remote_settings/src/storage.rs
```diff
@@ -88,14 +91,48 @@ impl Storage {
...
+                    match evaluate_filter_expression(&filter_expr, &env_context) {
+                        Ok(true) => {
+                            // Include the record
+                            records.push(record);
+                        }
+                        Ok(false) => {
+                            // Exclude the record
+                            continue;
+                        }
+                        Err(e) => {
```
gruberb, line 126: Not sure what should happen in an error case here. Abort or just logging?

leplatrem (commented):
I'm sorry but this is not exactly the right approach.

The JEXL filtering has to remain internal, and should not be exposed to the API consumers.

The filter expressions are part of the record data. Records that don't have the `filter_expression` field are always included. Records which a `filter_expression` field will be passed through the JEXL filter. This filter consists in executing the expression against a context passed as `env` (eg. `env.version > 112`). If the expression returns false or fails to be executed, we exclude the record (same behaviour as desktop, see [here](https://searchfox.org/mozilla-central/rev/783f3fca1dda58353f7d3075744dd48b66e00e5e/services/settings/RemoteSettingsClient.sys.mjs#1272) and [here](https://searchfox.org/mozilla-central/rev/783f3fca1dda58353f7d3075744dd48b66e00e5e/services/settings/remote-settings.sys.mjs#88))

leplatrem:
> Thanks @leplatrem. What I struggle understanding in the context of a cross-platform component: Is it the components job to read and parse ENV variables, or is it the job of the implementing client (Swift, Kotlin etc.) to pass down the ENV variable to this component?

Note: The context is not exactly env variables (as Unix environment vars), but a struct with variables that are provided by the target platform. 

I believe we should share the same approach (ideally even code) as Nimbus. For example, the context used as JEXL input is built in:
- Swift here https://github.com/mozilla/application-services/blob/608b930b02af156cc7c91fa0d8bc865bb4772dba/components/nimbus/ios/Nimbus/NimbusCreate.swift#L135-L152
- Kotlin here https://github.com/mozilla/application-services/blob/79df7fe34aebd9a929c3b98ee7cd2dcc4e799ad8/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt#L578-L595

For desktop, the environment is implement with a class and static properties https://searchfox.org/mozilla-central/source/toolkit/components/utils/ClientEnvironment.sys.mjs

> I think the filtering behavior currently works as you stated though

👍 

> Another question: I see that Nimbus is transforming the JEXL values.

Nimbus seems to have specific transform functions (`eventAveragePerInterval`, `event*`...).

Desktop supports these transforms: https://searchfox.org/mozilla-central/rev/783f3fca1dda58353f7d3075744dd48b66e00e5e/toolkit/components/utils/FilterExpressions.sys.mjs#33-47
So eventually we will have to provide them too for a full cross-platform compatibility. You can start with basic ones like `versionCompare` and then we can implement more transforms in follow-ups.

See [docs](https://docs.rs/jexl-eval/latest/jexl_eval/struct.Evaluator.html)
And https://remote-settings.readthedocs.io/en/latest/target-filters.html#transforms


Examples of PROD collections with filter expressions:
* https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/search-config-icons/records
* https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/addons-manager-settings/records
* https://firefox.settings.services.mozilla.com/v1/buckets/security-state/collections/cert-revocations/records
* https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/translations-models/records (seem all empty)
* https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/translations-wasm/records (seem all empty)

For example in `search-config-icons`, you can see transform at use: `filter_expression: 'env.appinfo.ID == "{ec8030f7-c20a-464f-9b0e-13a3a9e97384}" && env.version|versionCompare("128.0a1") > 0'`

components/remote_settings/src/storage.rs
leplatrem: Because they are target filters, not exclude filters, if the expressions fails to be evaluated then we consider the records has to be excluded.

gruberb:
@leplatrem I updated the PR and we accept now a `AppContext`. We don't want to write custom Swift/Kotlin code for a multitude of reasons (we don't know the product which will implement the `remote-settings` client, and therefore cannot be too opinionated how it will be used).

For the `transforms`: I have to dig into the `JS` code how they are actually being written. If this PR is a first pass, I can add `transforms` in a follow up. The datasets you linked are super helpful, thanks!!

components/remote_settings/src/jexl_filtering.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let result = compare_versions(version_a, version_b);
+    Ok(Value::from(result))
+}
+
+/// Helper function to compare two version strings.
+fn compare_versions(version_a: &str, version_b: &str) -> i8 {
+    let parse_version = |v: &str| {
+        v.split('.')
+            .map(|s| s.parse::<i32>().unwrap_or(0))
```
leplatrem, line 42: what about `a1`, `pre0`?

components/remote_settings/src/lib.rs
```diff
@@ -24,6 +28,48 @@ use storage::Storage;
 
 uniffi::setup_scaffolding!("remote_settings");
 
+/// The `AppContext` object represents the parameters and characteristics of the
+/// consuming application. For `remote-settings`, it is used to filter locally stored `records`.
```
leplatrem, line 32: Is `remote-settings` referring to something specific here? Otherwise I would suggest we spell Remote Settings everywhere we refer to the service/feature

components/remote_settings/src/lib.rs
```diff
@@ -24,6 +28,48 @@ use storage::Storage;
...
+/// We always fetch all `records` from the remote-settings storage. Some records could have a `filter_expression`
+/// attached to them, which will be matched against the `AppContext`. So just `records` for this particualar instance
+/// will be used if set.
```
leplatrem, line 35: > So just `records` for this particualar instance will be used if set.

I would rephrase into something simpler, for example: When set, only records where the expression is true will be returned.

components/remote_settings/src/storage.rs
```diff
@@ -82,21 +84,33 @@ impl Storage {
...
+                let mut stmt = tx.prepare("SELECT data FROM records WHERE collection_url = ?")?;
+                let mut rows = stmt.query(params![collection_url])?;
+
+                let mut records = Vec::new();
+                while let Some(row) = rows.next()? {
+                    let data: Vec<u8> = row.get(0)?;
+                    let record: RemoteSettingsRecord = serde_json::from_slice(&data)?;
+
+                    if let Some(filter_expr) = &record.filter_expression {
+                        let eval_result = evaluate_filter_expression(filter_expr, &jexl_context);
```
leplatrem, line 105: Are we sure we want to execute the filtering at the storage level? 
That means every type of storage will have to duplicate this code. Why not do it at a higher level?

components/remote_settings/src/client.rs
```diff
@@ -541,6 +541,8 @@ pub struct RemoteSettingsRecord {
     pub attachment: Option<Attachment>,
     #[serde(flatten)]
     pub fields: RsJsonObject,
+    #[serde(default)]
+    pub filter_expression: Option<String>,
```
leplatrem, line 545: Are there really performance involved if `filter_expression` remains a field reflecting what's on the server?

components/remote_settings/src/lib.rs
gruberb: I wasn't sure either. When spelling `remote-settings`, I mean the component.

components/remote_settings/src/storage.rs
gruberb: Yeah that's an interesting question. I don't see that we have different types of storage, parallel to each other. I decided to keep the `client` as clean and unchanged as possible, and we might want to store the `context` in the database as well? But maybe not. 

But I agree also with the other points: Nimubs has already implemented the filters. We could create a shared `jexl` crate and share it between components?

@jeddai what do you think? We could move the `jexl` code from Nimbus out into it's own crate, and share it across components?

components/remote_settings/src/jexl_filtering.rs
```diff
@@ -0,0 +1,158 @@
...
+            .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
gruberb: Thanks for pointing this out. I missed the `pre` and `a` cases etc. It seems like if we want to support `jexl` in our components, it makes sense to have a shared crate which we can import. Duplicating this doesn't make much sense. I will get in touch with the Nimbus folks to find a solution.

components/remote_settings/src/storage.rs
leplatrem: > we might want to store the context in the database as well? But maybe not.

I don't think so. Those are like build-time constants

bendk (changes requested):
The code looks mostly correct to me, the main issue is that `version_compare` function.  The other question is how to organize things.  I'm currently in the "move filtering out of storage camp", but I could be convinced otherwise pretty easily.

components/remote_settings/src/jexl_filtering.rs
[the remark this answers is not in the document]
gruberb: I talked with the Nimbus folks and we decided to create a `support` crate for this:
https://github.com/mozilla/application-services/pull/6454

What do you think?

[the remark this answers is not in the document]
leplatrem: Excellent!

components/remote_settings/src/client.rs
leplatrem: `filter_expression` is already part of `fields`. 

Having it as a separate column is a sort of optimization to avoid doing:

```rust
if records.fields.contains_key("filter_expression")
```

components/remote_settings/src/storage.rs
gruberb: Moved it out!

bendk (approved):
Thanks for updating this, the new version looks great to me.

components/remote_settings/src/client.rs
leplatrem: How is this resolved? 
What advantage is there in having a dedicated storage column for the `filter_expression`?

components/remote_settings/src/jexl_filter.rs
```diff
@@ -0,0 +1,52 @@
...
+        let env_context = match context {
+            Some(ctx) => {
+                serde_json::to_value(ctx).expect("Failed to serialize RemoteSettingsContext")
+            }
+            None => json!({}),
+        };
+
+        Self {
+            evaluator: Evaluator::new()
+                .with_transform("versionCompare", |args| Ok(version_compare(args)?)),
```
leplatrem, line 31: Please a comment here (or open a follow-up issue) that eventually we want to support the same transports as desktop https://remote-settings.readthedocs.io/en/latest/target-filters.html#transforms

components/remote_settings/src/client.rs
```diff
@@ -1529,11 +1588,111 @@ mod test_new_client {
...
+    #[test]
+    fn test_get_records_filtered_app_version_too_low() {
+        let mut api_client = MockApiClient::new();
+        let records = vec![RemoteSettingsRecord {
+            id: "record-0001".into(),
+            last_modified: 100,
+            deleted: false,
+            attachment: None,
+            fields: json!({"foo": "bar"}).as_object().unwrap().clone(),
+            filter_expression: Some("app_version|versionCompare('4.0') >= 0".to_string()),
```
leplatrem, line 1663: This is what I mean when I say that it could reflect the server content:
```suggestion
            fields: json!({"foo": "bar", "filter_expression": "app_version|versionCompare('4.0') >= 0".}).as_object().unwrap().clone(),
```

components/remote_settings/src/client.rs
```diff
@@ -48,6 +56,19 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         &self.collection_name
     }
 
+    /// Filters records based on the presence and evaluation of `filter_expression`.
+    fn filter_records(&self, records: Vec<RemoteSettingsRecord>) -> Vec<RemoteSettingsRecord> {
+        records
+            .into_iter()
+            .filter(|record| {
+                match &record.filter_expression {
```
leplatrem, line 64: When and where do you set this attribute? 

Why not just have `match record.fields.get("filter_expession")` here? I don't get it...

components/remote_settings/src/client.rs
```diff
@@ -1529,11 +1588,111 @@ mod test_new_client {
...
+
+    #[test]
+    fn test_get_records_filtered_app_version_pass() {
+        let mut api_client = MockApiClient::new();
+        let records = vec![RemoteSettingsRecord {
+            id: "record-0001".into(),
+            last_modified: 100,
+            deleted: false,
+            attachment: None,
+            fields: json!({"foo": "bar"}).as_object().unwrap().clone(),
```
leplatrem, line 1617: Driveby question: in your record representation, do you explicitly exclude `id`, `last_modified`, `deleted`, `attachment` from `fields`?

leplatrem (commented):
I'm still a bit doubtful about the dedicated attribute for the `filter_expression`, I don't see the value

components/remote_settings/src/client.rs
gruberb: Yeah, because we just test for `filter_expression` here. But maybe we should always have a "real" record representation so it's clear what format the `record` has.

components/remote_settings/src/client.rs
gruberb: I resolved it! Makes total sense to me now and you are right, we need to get the `filter_expression` from the `fields`. Or at least, it makes more sense that way. I updated the PR.

leplatrem:
> I resolved it! Makes total sense to me now and you are right, we need to get the filter_expression from the fields. Or at least, it makes more sense that way. I updated the PR.

I don't see it. Maybe you changed your mind? It's a bit hard to follow the latest changes with forced pushes :/ 

> I don't have a strong preference on where filter_expression ends up, but I want to discuss this more to make sure we have a common understanding on how RemoteSettingsRecord works. The way I see it, the top-level RemoteSettingsRecord attributes store the common/standard data and the fields attribute stores all the other/user-defined data. This is essentially the same way we handle sync records. filter_expression feels like a standard field so to me it feels better as a top-level attribute. Why do you like it better as part of the fields data?
>


I understand your point totally. You are right to insist 😉 
[Hence my question here](https://github.com/mozilla/application-services/pull/6449#discussion_r1832701918).
If I had a magic wand and could easily redesign the shape of the API and srever payloads, I would distinguish user fields from built-in fields (we had [a conversation in 2015](https://github.com/Kinto/kinto/issues/256) and I closed it because I was alone maintaining the stack in 2020, and definitely not ready to handle API breaking changes with millions of clients running different versions just for this).

Anyway, you may have noticed that all resources in remote settings look the same (buckets, collections, groups, records), but `filter_expression` is a bit different than `id` or `last_modified` timestamp, since it is something that only applies to records, and it is specific to the client application, namely gecko/a-s (for example, we don't filter the records when we pull the collection via the Python client).

However, that being said, I don't think it has much to do with having a dedicated storage column. 

I would suggest that we keep `RemoteSettignsRecord` high level attributes for `id`, `last_modified`, `deleted`.
And, when needed, for the Rust counterpart of certain fields. For example, `record.attachment` has type `Optional<Attachment>`, and `record.fields["attachment"]` is the raw JSON data.  

In this particular case of `filter_expression`, I don't see much value of having a dedicated storage column, and a dedicated attribute an optional string type. Seeing all these `filter_expression: None` also kind of confirms it I think.

> I don't feel that strongly either way and I definitely don't want block the PR on this. I'm just want us to be on the same page.

It's not a hill I want to die on either, but you're very right, we need to be aligned on these sort of implicit standard, because if we have a good consistency, then the developer experience is way better for those consuming the API :)

leplatrem:
In our Rust implementation, this is how I did it:

```rust
    // Return the record id.
    pub fn id(&self) -> &str {
        // `id` field is always present as a string.
        self.fields["id"].as_str().unwrap()
    }
```

https://github.com/mozilla-services/remote-settings-client/blob/12cea355b414458875d2b72a119d6f1b9cca702a/src/client.rs#L132-L150

bendk:
> Anyway, you may have noticed that all resources in remote settings look the same (buckets, collections, groups, records), but filter_expression is a bit different than id or last_modified timestamp, since it is something that only applies to records, and it is specific to the client application, namely gecko/a-s (for example, we don't filter the records when we pull the collection via the Python client).

Ahh, I understand better now and this makes sense to me.  Thanks for the explanation.

gruberb:
> I don't see it. Maybe you changed your mind? It's a bit hard to follow the latest changes with forced pushes :/

Sorry @leplatrem. I chatted with Ben back and forth and changed my mind a few times. I tried out different versions and now see how it's basically a 50/50 split where we see these fields.

I removed the `filter_expression` field (again) and use your suggested solution of keeping it in the `fields` and get it from there.

leplatrem:
🎉 Thank you for your patience on this one! The result is great!

PR #6451, opened 2024-11-01
Bug 1928544 - For weather suggestions, don't match on airport codes or abbreviations when they're the only term in the query

0c0w3:
In the original GeoNames data, each alternate name has an associated `iso_language` that indicates what kind of name it is. In addition to actual language codes, it can also be `"abbr"` for an abbreviation ("NY") and a handful of other values that all mean an airport code ("PDX"). We need to keep track of the `iso_language` per alternate so that we don't match on an abbreviation or airport code when it's the only term in the query. (Or, we could store an `is_abbreviation` in the RS data, but that's worse in the long term, and it's also good to be consistent with the original data where feasible IMO.)

I modified the schema and RS structs accordingly. In `DownloadedGeoname`, I changed the name of `alternate_names` to `alternate_names_2` because, while I'd like to uplift this change, it's not clear there's enough time. So we may need to support the old `alternate_names` on 133.

The other main change is to `filter_map_chunks()`: The filter-map function is now passed the chunk size, and weather uses it to tell whether the chunk is the last one in the query string.

  - Note:

### DIFF
--- a/components/suggest/src/weather.rs
+++ b/components/suggest/src/weather.rs
@@ -15,7 +15,7 @@ use crate::{
         KeywordInsertStatement, KeywordMetricsInsertStatement, SuggestDao,
         SuggestionInsertStatement, DEFAULT_SUGGESTION_SCORE,
     },
-    geoname::{Geoname, GeonameType},
+    geoname::{GeonameMatch, GeonameType},
     metrics::MetricsContext,
     provider::SuggestionProvider,
     rs::{Client, Record, SuggestRecordId},
@@ -101,7 +101,8 @@ impl SuggestDao<'_> {
             .collect();
 
         let mut matches =
-            filter_map_chunks::<Token>(&words, max_chunk_size, |chunk, c_i, c_size, path| {
+            // Step 2: Parse the query words into a list of token paths.
+            filter_map_chunks::<Token>(&words, max_chunk_size, |chunk, chunk_i, path| {
                 // Match the chunk to token types that haven't already been matched
                 // in this path. `all_tokens` will remain `None` until a token is
                 // matched.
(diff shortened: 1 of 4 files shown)


components/suggest/src/weather.rs
```diff
@@ -197,17 +197,21 @@ impl SuggestDao<'_> {
         path: &[Token],
         candidate: &str,
         is_first_chunk: bool,
+        is_last_chunk: bool,
```
ncloudioj, line 200: It took me a bit to sort out the combination of `is_first_chunk` and `is_last_chunk`. What do you think about merging those two booleans into an enum so that both `fetch_geonames()` and `match_weather_tokens()` would take an enum instead of two booleans?

Something like:

```rs
enum ChunkPosition {
  Singleton,
  HeadOfMany,
  TailOfMany, // If useful
  Other,
}
```

ncloudioj (approved):
LGTM!

components/suggest/src/weather.rs
0c0w3: Thanks for the quick review! I think things would be clearer if the param names of `match_weather_tokens()` were changed to `match_prefixes` and `match_abbreviations` and some comments were added to the call site, where `is_first_chunk` and `is_last_chunk` are defined. Even if we had an enum, it would likely be created by first computing `is_first_chunk` and `is_last_chunk` (or something like them), so we'd still need them.

Re: `fetch_geonames()`, I actually started making a `GeonameFetchOptions` struct since that method's param list is getting kind of big. In the end I thought it might be too unnecessary, and with the `match_prefixes` and `match_abbreviations` change I mentioned above, I think it's much clearer now (in the new commit I'll push soon).

0c0w3:
I canceled the merge because on second thought the matching might still be too aggressive. This commit prevents "act" (airport code for Waco, TX) from triggering a suggestion, but it'll still be triggered as soon as you start typing another word that matches (keyword or state name), e.g. "act w" (for "act weather") or "act t" (for "act tx"). Maybe the `min_keyword_length` threshold from the config should apply to the first term following an abbreviation/airport code, or we could just hardcode a threshold like 3.

0c0w3:
Requesting review again since I made some significant changes. The matching is now much less aggressive and much better. City abbreviations and airport codes now must be typed in full, and they require other terms to be present in the query. City abbreviations require a weather keyword, state abbreviation, or state name. IOW we want to confirm either a weather intent or an intent for a specific city. Airport codes now require a weather intent. Airport codes need stricter requirements than city abbreviations because they're often not well known and might have nothing in common with the corresponding city name, which can make them surprising.

Here's a summary of the kinds queries that match:

* Full city name ("new york")
* Full city abbreviation + full state name ("ny new york")
* Full city abbreviation + full state abbreviation ("ny ny", "nyc ny")
* Weather keyword
* Weather keyword + city prefix ("weather w", desktop will pick the city that's closest to you from the matching cities returned by Rust)
* Weather keyword + full city abbreviation ("weather ny", "ny weather")
* Weather keyword + full airport code ("weather pdx", "pdx weather")

Summary of code changes:

* Remove the abbreviation-matching logic in `fetch_geonames()` that I added in this PR's previous commit.
* Make `fetch_geonames()` return a new `GeonameMatch` struct instead of a `Geoname`. The match struct tells us what kind of thing was matched (abbreviation, airport code, other name) and whether a prefix was matched.
* Make `match_weather_tokens()` return a new `WeatherkeywordMatch` instead of a string for matched keywords. It tells us whether the match met the min keyword length in the config.
* Change the main weather-matching logic to use the new info from the previous two points.
* Add test cases.

ncloudioj (approved):
Yup, the matching logic makes more sense now!

PR #6454, opened 2024-11-04
feat: move versioning into its own support crate

gruberb:
Multiple components in application-services need to filter for the exact Firefox version. Instead of duplicating the code, we move it into its own support crate, which other components can now make use of.

Moves out of `Nimbus`:
- [x] `versioning.rs`
- [x] `tests/test_versioning.rs` 

Adds:
- [x] `components/support/versioning`

### DIFF
--- a/components/nimbus/src/error.rs
+++ b/components/nimbus/src/error.rs
@@ -8,6 +8,8 @@
 //! TODO: Implement proper error handling, this would include defining the error enum,
 //! impl std::error::Error using `thiserror` and ensuring all errors are handled appropriately
 
+#[cfg(feature = "stateful")]
+use firefox_versioning::error::VersionParsingError;
 use std::num::{ParseIntError, TryFromIntError};
 
 #[derive(Debug, thiserror::Error)]
@@ -104,6 +106,13 @@ pub enum CirrusClientError {
 impl<'a> From<jexl_eval::error::EvaluationError<'a>> for NimbusError {
     fn from(eval_error: jexl_eval::error::EvaluationError<'a>) -> Self {
         NimbusError::EvaluationError(eval_error.to_string())
+    }
+}
+
+#[cfg(feature = "stateful")]
+impl From<VersionParsingError> for NimbusError {
+    fn from(eval_error: VersionParsingError) -> Self {
+        NimbusError::VersionParsingError(eval_error.to_string())
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1523,6 +1523,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "firefox-versioning"
+version = "0.1.0"
+dependencies = [
+ "serde_json",
+ "thiserror",
+]
+
+[[package]]
 name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2697,6 +2705,7 @@ dependencies = [
  "ctor 0.2.8",
  "env_logger",
  "error-support",
+ "firefox-versioning",
  "glean-build",
  "hex",
  "jexl-eval",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ members = [
 
     "examples/*/",
     "testing/separated/*/",
+    "components/support/firefox-versioning",
 ]
 
 exclude = [
--- a/components/nimbus/Cargo.toml
+++ b/components/nimbus/Cargo.toml
@@ -17,7 +17,7 @@ name = "nimbus"
 default=["stateful"]
 rkv-safe-mode = ["dep:rkv"]
 stateful-uniffi-bindings = []
-stateful = ["rkv-safe-mode", "stateful-uniffi-bindings", "dep:remote_settings", "dep:regex"]
+stateful = ["rkv-safe-mode", "stateful-uniffi-bindings", "dep:remote_settings", "dep:regex", "dep:firefox-versioning"]
 
 [dependencies]
 anyhow = "1"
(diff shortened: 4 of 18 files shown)


components/support/versioning/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "versioning"
+version = "0.1.0"
+authors = ["Bastian Gruber <bgruber@mozilla.com>"]
```
brennie, line 4: This should probably include the author credit of whomever added this module to nimbus first, or `project-nimbus@` at the very least (the credit for nimbus isnt right either but thats for another day)

components/support/versioning/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "versioning"
```
brennie, line 2: This is specific to Firefox versioning so the title should probably reflect that.

components/support/versioning/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod error;
```
brennie, line 1: This is missing the license header.

components/support/versioning/src/error.rs
```diff
@@ -0,0 +1,24 @@
+use std::fmt;
```
brennie, line 1: This is missing the license header.

components/support/versioning/src/error.rs
```diff
@@ -0,0 +1,24 @@
...
+
+/// The `VersioningError` indicates that a Firefox Version, being passed down as a `String`,
+/// cannot be parsed into a `Version`.
+///
+/// It can either be caused by a non-ASCII character or a integer overflow.
+#[derive(Debug, PartialEq)]
+pub enum VersioningError {
```
brennie, line 8: The old enum variant was VersionParsingError. Lets keep that name, so its very clear what the error is about.

components/support/versioning/src/error.rs
```diff
@@ -0,0 +1,24 @@
...
+    Overflow(String),
+    /// Indicates a general parsing error.
+    ParseError(String),
+}
+
+impl fmt::Display for VersioningError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            VersioningError::Overflow(msg) => write!(f, "Versioning overflow error: {}", msg),
+            VersioningError::ParseError(msg) => write!(f, "Versioning parser error: {}", msg),
```
brennie, line 19: Maybe we want to keep the Display impl equivalent, e.g., just write `msg` ?

brennie:
If you want to add the jexl stuff to this library specifically, you can move [`version_compare`](https://github.com/mozilla/application-services/blob/79df7fe34aebd9a929c3b98ee7cd2dcc4e799ad8/components/nimbus/src/targeting.rs#L153-L175) into `src/jexl.rs` with a `#[cfg(feature = "jexl")]`. You'll need to add conditional deps on `serde_json`.

jeddai (commented):
Some thoughts — if you would like a walkthrough of how to add the version transformer we could hop on a quick pairing call? Or I can write a short explanation out

gruberb:
Thanks! I updated the PR. I am not 100% sure how you use your feature flags, so please double check if it's ok to import `firefox-versioning` with the new `jexl` feature when using `stateful` in `Nimbus. 

Happy to jump on a call @jeddai if it's quicker :)

components/support/firefox-versioning/src/jexl.rs
```diff
@@ -0,0 +1,40 @@
...
+    let min_version = Version::try_from(min_version)?;
+    let curr_version = Version::try_from(curr_version)?;
+    Ok(json!(if curr_version > min_version {
+        1
+    } else if curr_version < min_version {
+        -1
+    } else {
+        0
+    }))
+}
```
bendk, line 40: This would be a good time to introduce a bunch of tests with various obscure versions.

bendk (approved):
Bastian and I discussed this on a call and creating a support crate for this makes sense to me.  I haven't reviewed the code thoroughly since it's just a cut-and-paste job.

components/nimbus/src/error.rs
```diff
@@ -107,4 +109,11 @@ impl<'a> From<jexl_eval::error::EvaluationError<'a>> for NimbusError {
     }
 }
 
+#[cfg(feature = "stateful")]
+impl From<VersionParsingError> for NimbusError {
+    fn from(eval_error: VersionParsingError) -> Self {
+        NimbusError::VersionParsingError(eval_error.to_string())
+    }
+}
```
brennie, line 117: The NimbusError should wrap this error isntead of coercing it to a string. Remove this and edit the VersionParsingError variant above:

```rust
#[cfg(feature="stateful")]
#[error("Error parsing a string into a version {0}")]
VersionParsingError(#[from] VersionParsingError),
```

brennie: Oh this merged. Whelp nevermind.

gruberb: https://github.com/mozilla/application-services/pull/6459

gruberb: Thanks for having another look! Sorry for being too fast. I have to merge a bunch of PRs to unblock myself for the current PR I am working on. I fixed it!

gruberb: Ah, now I remember why I added it like this. 

```
rror[E0220]: associated type `VersionParsingError` not found for `Self`
  --> /home/circleci/project/target/debug/build/nimbus-sdk-29671292901466d4/out/cirrus.uniffi.rs:75:5
   |
75 |     r#VersionParsingError {
   |     ^^^^^^^^^^^^^^^^^^^^^ help: `Self` has the following associated type: `FfiType`
```

I already had to make a few changes because of the feature flag. `compare_version` is now behind the `stateful` feature flag, which has a larger impact on the codebase. I wasn't sure if you want to keep `VersionParsingError` in the `currius.uniffi` file or not.

PR #6458, opened 2024-11-06
feat: remove codecov

gruberb:
Up for discussion! I talked with @bendk about our code coverage reports, and the question if people are looking at the stats, or if they just ignore it anyway.  `codecov` is also breaking the CI at the moment.

This PR is here to discuss and if people feel strongly one way or another!


mhammond:
I'm fine with that and it's one less thing to ponder for the monorepo.

linabutler (approved):
Agreed; let's drop it!

PR #6465, opened 2024-11-08
fix: increase timeout for nimbus ios test for CI 

gruberb:
This test is flaky, and fails every now and then. My assumption is that it's because of the timeout. Locally, it runs through, and on some PRs as well. Just every now and then, the CI takes longer than expected.


freshstrangemusic:
@gruberb sorry I missed this in my review queue. Lgtm

PR #6469, opened 2024-11-12
EncryptorDecryptor Trait for Logins Component

jo:
The encryption and decryption of credentials is outsourced to a [Foreign Trait](https://mozilla.github.io/uniffi-rs/0.27/foreign_traits.html) so that Android, desktop and iOS can each bring their own implementations of encryption, including key management. This makes the Logins API way simpler and cleaner. It is the first step in making AS-Logins desktop-ready.

The new `EncryptorDecryptor` trait replaces the current `EncryptorDecryptor` struct. Instead of using the `decrypt_struct` method, which involves serializing and deserializing the string to be encrypted, this trait uses only byte-based operations and serialization is up to the consumer.

We do not Uniffi-expose the crypto primitives `encrypt`, `decrypt`, `encrypt_struct` and `decrypt_struct` anymore. Also `EncryptedLogin` will not be exposed anymore.

A `ManagedEncryptorDecryptor` will provide an `EncryptorDecryptor` implementation which uses the currently used crypto methods, given a `KeyManager` implementation to ease adaption for mobile.

### BREAKING CHANGE

This commit introduces breaking changes to the Logins component:

During initialization, it receives an additional argument, a `EncryptorDecryptor` implementation. In addition, several `LoginsStore` API methods have been changed to not require an encryption key argument anymore, and return `Login` objects instead of `EncryptedLogin`.

Additionally, a new API method has been added to the LoginsStore, `has_logins_by_base_domain(&self, base_domain: &str)`, which can be used to check for the existence of a login for a given base domain.

### EncryptorDecryptor

With the introduction of the `EncryptorDecryptor` trait, encryption becomes transparent. That means, the LoginStore API receives some breaking changes as outlined above.  A `ManagedEncryptorDecryptor` will provide an `EncryptorDecryptor` implementation which uses the currently used crypto methods, given a `KeyManager` implementation. This eases adaption for mobile.  Furthermore, we provide a `StaticKeyManager` implementation, which can be used in tests and in cases where the key is - you name it - static.  Constructors Now an implementation of the above property must be passed to the constructors. To do this, the signatures are extended as follows:

```
pub fn new(path: impl AsRef<Path>, encdec: Arc<dyn EncryptorDecryptor>) -> ApiResult<Self>
pub fn new_from_db(db: LoginDb, encdec: Arc<dyn EncryptorDecryptor>) -> Self
pub fn new_in_memory(encdec: Arc<dyn EncryptorDecryptor>) -> ApiResult<Self>
```

### LoginStore API Methods
This allows the `LoginStore` API to be simplified as follows, making encryption transparent by eliminating the need to pass the key and allowing the methods to return decrypted login objects.

```
pub fn list(&self) -> ApiResult<Vec<Login>>
pub fn get(&self, id: &str) -> ApiResult<Option<Login>>
pub fn get_by_base_domain(&self, base_domain: &str) -> ApiResult<Vec<Login>>
pub fn find_login_to_update(&self, entry: LoginEntry) -> ApiResult<Option<Login>>
pub fn update(&self, id: &str, entry: LoginEntry) -> ApiResult<Login>
pub fn add(&self, entry: LoginEntry) -> ApiResult<Login>
pub fn add_or_update(&self, entry: LoginEntry) -> ApiResult<Login>
```

New LoginsStore methods:
```
// Checking whether the database contains logins (does not utilize the `EncryptorDecryptor`):
is_empty(&self) -> ApiResult<bool>
// Checking for the Existence of Logins for a given base domain (also does not utilize the `EncryptorDecryptor`):
has_logins_by_base_domain(&self, base_domain: &str) -> ApiResult<bool>
```

We will stop Uniffi-exposing the crypto primitives `encrypt`, `decrypt`, `encrypt_struct` and `decrypt_struct`. Also `EncryptedLogin` will not be exposed anymore. 

### SyncEngine
The logins sync engine has been adapted for above EncryptorDecryptor trait and therefore does not support `set_local_encryption_key` anymore.

### Flattened Login Struct

The current Login structure is nested:

```
Login {
  RecordFields record;
  LoginFields fields;
  SecureLoginFields sec_fields;
}
```

and thus exposes internal data structuring to the consumer. Since we make the encryption transparent for the consumer (Android, iOS, desktop) here, such a separation no longer makes sense here and above can be simplified to

```
Login {
    // record fields
    string id;
    i64 times_used;
    i64 time_created;
    i64 time_last_used;
    i64 time_password_changed;

    // login fields
    string origin;
    string? http_realm;
    string? form_action_origin;
    string username_field;
    string password_field;

    // secure login fields
    string password;
    string username;
}
```

The advantage of eliminating this separation lies, on the one hand, in the simplification of the API and the resulting easier use of the component, and on the other hand, in the easier changeability of the internal data structure. If, for example, we decide later to encrypt additional fields, such a change is possible without having to adapt the consumers.

So in addition to the changes mentioned above we will stop Uniffi-exposing the structs `RecordFields`, `LoginFields` and `SecureLoginFields`.

### Corresponding PR & Patch addressing breaking changes:

* **Android:** https://phabricator.services.mozilla.com/D232391
* **iOS:** https://github.com/mozilla-mobile/firefox-ios/pull/24108

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -223,13 +223,9 @@ impl LoginDb {
         Ok(logins
             // First, try to match the username
             .iter()
-            .find(|login| login.sec_fields.username == look.sec_fields.username)
+            .find(|login| login.username == look.username)
             // Fall back on a blank username
-            .or_else(|| {
-                logins
-                    .iter()
-                    .find(|login| login.sec_fields.username.is_empty())
-            })
+            .or_else(|| logins.iter().find(|login| login.username.is_empty()))
             // Clone the login to avoid ref issues when returning across the FFI
             .cloned())
     }
@@ -364,6 +360,10 @@ impl LoginDb {
         let now_ms = util::system_time_ms_i64(SystemTime::now());
 
         let new_entry = self.fixup_and_check_for_dupes(&guid, entry, encdec)?;
+        let sec_fields = SecureLoginFields {
+            username: new_entry.username,
+            password: new_entry.password,
+        };
         let result = EncryptedLogin {
             record: RecordFields {
                 id: guid.to_string(),
@@ -372,8 +372,14 @@ impl LoginDb {
                 time_last_used: now_ms,
                 times_used: 1,
             },
-            fields: new_entry.fields,
-            sec_fields: new_entry.sec_fields.encrypt(encdec)?,
+            fields: LoginFields {
+                origin: new_entry.origin,
+                form_action_origin: new_entry.form_action_origin,
+                http_realm: new_entry.http_realm,
+                username_field: new_entry.username_field,
+                password_field: new_entry.password_field,
+            },
+            sec_fields: sec_fields.encrypt(encdec)?,
         };
         let tx = self.unchecked_transaction()?;
         self.insert_new_login(&result)?;
@@ -402,8 +408,8 @@ impl LoginDb {
             // Try to detect if sync is enabled by checking if there are any mirror logins
             let has_mirror_row: bool =
                 self.db.query_one("SELECT EXISTS (SELECT 1 FROM loginsM)")?;
-            let has_http_realm = entry.fields.http_realm.is_some();
-            let has_form_action_origin = entry.fields.form_action_origin.is_some();
+            let has_http_realm = entry.http_realm.is_some();
+            let has_form_action_origin = entry.form_action_origin.is_some();
             report_error!(
                 "logins-duplicate-in-update",
                 "(mirror: {has_mirror_row}, realm: {has_http_realm}, form_origin: {has_form_action_origin})");
@@ -415,27 +421,36 @@ impl LoginDb {
 
         // We must read the existing record so we can correctly manage timePasswordChanged.
         let existing = match self.get_by_id(sguid)? {
-            Some(e) => e,
+            Some(e) => e.decrypt(encdec)?,
             None => return Err(Error::NoSuchRecord(sguid.to_owned())),
         };
-        let time_password_changed =
-            if existing.decrypt_fields(encdec)?.password == entry.sec_fields.password {
-                existing.record.time_password_changed
-            } else {
-                now_ms
-            };
+        let time_password_changed = if existing.password == entry.password {
+            existing.time_password_changed
+        } else {
+            now_ms
+        };
 
         // Make the final object here - every column will be updated.
+        let sec_fields = SecureLoginFields {
+            username: entry.username,
+            password: entry.password,
+        };
         let result = EncryptedLogin {
             record: RecordFields {
-                id: existing.record.id,
-                time_created: existing.record.time_created,
+                id: existing.id,
+                time_created: existing.time_created,
                 time_password_changed,
                 time_last_used: now_ms,
-                times_used: existing.record.times_used + 1,
+                times_used: existing.times_used + 1,
             },
-            fields: entry.fields,
-            sec_fields: entry.sec_fields.encrypt(encdec)?,
+            fields: LoginFields {
+                origin: entry.origin,
+                form_action_origin: entry.form_action_origin,
+                http_realm: entry.http_realm,
+                username_field: entry.username_field,
+                password_field: entry.password_field,
+            },
+            sec_fields: sec_fields.encrypt(encdec)?,
         };
 
         self.update_existing_login(&result)?;
@@ -451,7 +466,7 @@ impl LoginDb {
         // Make sure to fixup the entry first, in case that changes the username
         let entry = entry.fixup()?;
         match self.find_login_to_update(entry.clone(), encdec)? {
-            Some(login) => self.update(&login.record.id, entry, encdec),
+            Some(login) => self.update(&login.id, entry, encdec),
             None => self.add(entry, encdec),
         }
     }
@@ -497,7 +512,7 @@ impl LoginDb {
         for possible in self.get_by_entry_target(entry)? {
             if possible.guid() != *guid {
                 let pos_sec_fields = possible.decrypt_fields(encdec)?;
-                if pos_sec_fields.username == entry.sec_fields.username {
+                if pos_sec_fields.username == entry.username {
                     return Ok(Some(possible.guid()));
                 }
             }
@@ -546,13 +561,10 @@ impl LoginDb {
                 common_cols = schema::COMMON_COLS
             );
         }
-        match (
-            entry.fields.form_action_origin.as_ref(),
-            entry.fields.http_realm.as_ref(),
-        ) {
+        match (entry.form_action_origin.as_ref(), entry.http_realm.as_ref()) {
             (Some(form_action_origin), None) => {
                 let params = named_params! {
-                    ":origin": &entry.fields.origin,
+                    ":origin": &entry.origin,
                     ":form_action_origin": form_action_origin,
                 };
                 self.db
@@ -562,7 +574,7 @@ impl LoginDb {
             }
             (None, Some(http_realm)) => {
                 let params = named_params! {
-                    ":origin": &entry.fields.origin,
+                    ":origin": &entry.origin,
                     ":http_realm": http_realm,
                 };
                 self.db
@@ -885,21 +897,16 @@ mod tests {
     use super::*;
     use crate::encryption::test_utils::TEST_ENCDEC;
     use crate::sync::merge::LocalLogin;
-    use crate::SecureLoginFields;
     use std::{thread, time};
 
     #[test]
     fn test_username_dupe_semantics() {
         let mut login = LoginEntry {
-            fields: LoginFields {
-                origin: "https://www.example.com".into(),
-                http_realm: Some("https://www.example.com".into()),
-                ..LoginFields::default()
-            },
-            sec_fields: SecureLoginFields {
-                username: "test".into(),
-                password: "sekret".into(),
-            },
+            origin: "https://www.example.com".into(),
+            http_realm: Some("https://www.example.com".into()),
+            username: "test".into(),
+            password: "sekret".into(),
+            ..LoginEntry::default()
         };
 
         let db = LoginDb::open_in_memory().unwrap();
@@ -916,7 +923,7 @@ mod tests {
         );
 
         // Add one with an empty username - not a dupe.
-        login.sec_fields.username = "".to_string();
+        login.username = "".to_string();
         db.add(login.clone(), &*TEST_ENCDEC)
             .expect("empty login isn't a dupe");
 
@@ -935,17 +942,13 @@ mod tests {
         let added = db
             .add(
                 LoginEntry {
-                    fields: LoginFields {
-                        form_action_origin: Some("http://😍.com".into()),
-                        origin: "http://😍.com".into(),
-                        http_realm: None,
-                        username_field: "😍".into(),
-                        password_field: "😍".into(),
-                    },
-                    sec_fields: SecureLoginFields {
-                        username: "😍".into(),
-                        password: "😍".into(),
-                    },
+                    form_action_origin: Some("http://😍.com".into()),
+                    origin: "http://😍.com".into(),
+                    http_realm: None,
+                    username_field: "😍".into(),
+                    password_field: "😍".into(),
+                    username: "😍".into(),
+                    password: "😍".into(),
                 },
                 &*TEST_ENCDEC,
             )
@@ -973,16 +976,12 @@ mod tests {
         let added = db
             .add(
                 LoginEntry {
-                    fields: LoginFields {
-                        form_action_origin: None,
-                        origin: "http://😍.com".into(),
-                        http_realm: Some("😍😍".into()),
-                        ..Default::default()
-                    },
-                    sec_fields: SecureLoginFields {
-                        username: "😍".into(),
-                        password: "😍".into(),
-                    },
+                    form_action_origin: None,
+                    origin: "http://😍.com".into(),
+                    http_realm: Some("😍😍".into()),
+                    username: "😍".into(),
+                    password: "😍".into(),
+                    ..Default::default()
                 },
                 &*TEST_ENCDEC,
             )
@@ -1019,15 +1018,10 @@ mod tests {
         for h in good.iter().chain(bad.iter()) {
             db.add(
                 LoginEntry {
-                    fields: LoginFields {
-                        origin: (*h).into(),
-                        http_realm: Some((*h).into()),
-                        ..Default::default()
-                    },
-                    sec_fields: SecureLoginFields {
-                        password: "test".into(),
-                        ..Default::default()
-                    },
+                    origin: (*h).into(),
+                    http_realm: Some((*h).into()),
+                    password: "test".into(),
+                    ..Default::default()
                 },
                 &*TEST_ENCDEC,
             )
@@ -1108,15 +1102,11 @@ mod tests {
     fn test_add() {
         let db = LoginDb::open_in_memory().unwrap();
         let to_add = LoginEntry {
-            fields: LoginFields {
-                origin: "https://www.example.com".into(),
-                http_realm: Some("https://www.example.com".into()),
-                ..Default::default()
-            },
-            sec_fields: SecureLoginFields {
-                username: "test_user".into(),
-                password: "test_password".into(),
-            },
+            origin: "https://www.example.com".into(),
+            http_realm: Some("https://www.example.com".into()),
+            username: "test_user".into(),
+            password: "test_password".into(),
+            ..Default::default()
         };
         let login = db.add(to_add, &*TEST_ENCDEC).unwrap();
         let login2 = db.get_by_id(&login.record.id).unwrap().unwrap();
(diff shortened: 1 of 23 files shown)


mhammond (commented):
This is fantastic and a massive improvement, thanks! Dealing with Android and iOS will be a challenge though.

components/logins/src/db.rs
```diff
@@ -212,13 +212,13 @@ impl LoginDb {
...
         look: LoginEntry,
-        encdec: &EncryptorDecryptor,
+        encdec: Arc<dyn EncryptorDecryptor>,
     ) -> Result<Option<Login>> {
         let look = look.fixup()?;
         let logins = self
             .get_by_entry_target(&look)?
             .into_iter()
-            .map(|enc_login| enc_login.decrypt(encdec))
+            .map(|enc_login| enc_login.decrypt(encdec.clone()))
```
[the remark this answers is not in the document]
jo: Yes, I agree. I also tried to pass around internal references at first. Unfortunately, for reasons I can't remember, that wasn't so easy, so I went the way of least resistance and passed the Arcs around internally as well. I'll look into it again – if it's not too much work, that would indeed be cleaner.

components/logins/src/encryption.rs
```diff
@@ -28,22 +28,82 @@
...
+    pub fn new(key: String) -> Self {
+        Self { key }
+    }
+}
+impl KeyManager for StaticKeyManager {
+    #[handle_error(Error)]
+    fn get_key(&self) -> ApiResult<Vec<u8>> {
+        Ok(self.key.as_bytes().into())
+    }
+}
```
bendk, line 92: IIUC consumers have 3 options for this:
  - Implement `EncryptorDecryptor` directly
  - Implement `KeyManager`
  - Generate a single key and create a `StaticKeyManager`

If so, I like the design a lot.  We should probably document it at some point, but that can be a follow-up.

components/logins/src/logins.udl
```diff
@@ -107,19 +82,41 @@ interface LoginsApiError {
...
 };
 
+[Trait, WithForeign]
+interface EncryptorDecryptor {
+    [Throws=LoginsApiError]
+    bytes encrypt(bytes cleartext, string description);
+
+    [Throws=LoginsApiError]
+    bytes decrypt(bytes ciphertext, string description);
+};
```
bendk, line 92: What do you think about using an enum for the description?  I'm thinking there's only going to be a small number of cases that we need to distinguish and using an enum will reduce the amount of bytes going across the FFI.

bendk (dismissed):
Looks really great to me, I like the overall approach.  I commented on some of the details -- treat them as suggestions only.  I don't have strong feelings either way.

Hopefully the consumer app updates aren't too bad.

components/logins/src/logins.udl
```diff
@@ -134,16 +131,16 @@ interface LoginStore {
     void touch([ByRef] string id);
 
     [Throws=LoginsApiError]
-    sequence<EncryptedLogin> list();
+    sequence<Login> list();
 
     [Throws=LoginsApiError]
-    sequence<EncryptedLogin> get_by_base_domain([ByRef] string base_domain);
+    sequence<Login> get_by_base_domain([ByRef] string base_domain);
```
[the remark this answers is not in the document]
jo: very good point, will do

components/logins/src/encryption.rs
jo: Yes, thats how it's intended.

```rust
// Implement EncryptorDecryptor directly:
LoginStore::new(MyEncryptorDecryptor)
// Implement KeyManager:
LoginStore::new(ManagedEncryptorDecryptor::new(MyKeyManager))
// Generate a single key and create a StaticKeyManager:
LoginStore::new(ManagedEncryptorDecryptor::new(StaticKeyManager { key: myKey }))
```
(simplified, w/o the use of Arcs)

Is this the design you expected?

I will add and update the documentation within this PR.

components/logins/src/logins.udl
jo: Yes, I very much like that. Maybe we can even go without an enum entirely; afaik we currently only use two strings, and one of them is for the tests. We could use a single static one, wdyt?

bendk: IIRC, the description ends up in the error message when we report decryption failures to Sentry.

Maybe it's not needed anymore.  If it is, then it probably doesn't need to be part of the trait.  You could a wrapper function that calls decrypt/encrypt and then adds the message to the error type on failure.

Speaking of that, should `decrypt` return a `LoginsApiError` or a more specific error type?  I could see this working either way, but I wanted to raise this for discussion.

jo: Thats a good question. I improved error handling a little and added a `LoginsApiError::MissingKey`, which is used internally by `ManagedEncryptorDecryptor` and can also be used by the external implementation along with `LoginsApiError::IncorrectKey` - do you think thats specific enough?

bendk: That sounds fine to me, I was mostly just curious on your thoughts.  When I considered similar ideas for other crates I wasn't sure exactly how I would handle it.

components/logins/src/logins.udl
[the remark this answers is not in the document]
jo: done

components/logins/src/encryption.rs
jo: done

components/logins/src/logins.udl
jo: Yes, I think less is more, the simpler the better. If there comes up a concrete need for more detailed bugs at some point, we can build them in :)

components/logins/src/sync/engine.rs
```diff
@@ -30,26 +29,15 @@ pub struct LoginsSyncEngine {
     pub store: Arc<LoginStore>,
     pub scope: SqlInterruptScope,
     pub staged: RefCell<Vec<IncomingBso>>,
-    // It's unfortunate this is an Option<>, but tricky to change because sometimes we construct
-    // an engine for, say, a `reset()` where this isn't needed or known.
-    encdec: Option<EncryptorDecryptor>,
```
bendk, line 35: Very happy to lose this  :clinking_glasses:

components/logins/src/error.rs
```diff
@@ -58,12 +61,21 @@ pub enum Error {
...
+    DecryptionFailed,
+
     #[error("Error synchronizing: {0}")]
     SyncAdapterError(#[from] sync15::Error),
 
     #[error("Error parsing JSON data: {0}")]
     JsonError(#[from] serde_json::Error),
 
+    #[error("Error constructing UTF8 string: {0}")]
+    Utf8Error(#[from] std::str::Utf8Error),
```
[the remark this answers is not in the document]
jo: yeah, agree!

components/logins/src/store.rs
```diff
@@ -48,48 +48,76 @@ fn create_sync_engine(
 
 pub struct LoginStore {
     pub db: Mutex<LoginDb>,
+    pub encdec: Arc<dyn EncryptorDecryptor>,
```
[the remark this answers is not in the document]
jo: good catch!

components/logins/src/sync/engine.rs
jo: \o/

components/logins/src/error.rs
```diff
@@ -21,6 +21,9 @@ pub enum LoginsApiError {
     #[error("No record with guid exists (when one was required): {reason:?}")]
     NoSuchRecord { reason: String },
 
+    #[error("Encryption key is missing.")]
+    MissingKey,
```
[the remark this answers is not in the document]
jo: yesss!

components/logins/src/error.rs
```diff
@@ -58,12 +61,21 @@ pub enum Error {
     #[error("local encryption key not set")]
     EncryptionKeyMissing,
 
+    #[error("encryption failed")]
+    EncryptionFailed,
```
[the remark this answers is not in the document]
jo: for sure!

jo:
> This looks great to me. A lot of code is changing, but in general it seems like the same system as before only better. I mostly just had suggestions around the error handling.

Thanks for the good points! Incorporated in [9a5f309](https://github.com/mozilla/application-services/pull/6469/commits/9a5f30982713c49c8acd9b26f89b1bb8751e43f4), will squash later

bendk (approved):
The changes look great to me, I'm all for merging this.  I think we should get a second review if possible though.

components/logins/src/logins.udl
```diff
@@ -72,15 +68,6 @@ dictionary Login {
     SecureLoginFields sec_fields;
 };
 
-/// An encrypted version of [Login].  This is what we return for all the "read"
-/// APIs - we never return the cleartext of encrypted fields.
-dictionary EncryptedLogin {
```
mhammond, line 77: I'm really happy to see this die. Back in the day, we were torn about whether the logins store should be able to function while in a "locked" state - ie, we were picturing a world where the app knows a login exists for a site, but the app must force the user to unlock via biometrics etc before the actual details were available, which then allows the app to "re-lock" the store at its convenience.

It looks like you've decided this is a non-goal, and your team is the correct team to make that decision, and the extra simplicity that comes with that is a nice bonus!

You might be able to take this further though - it seems that without `EncryptedLogin` you might also be able to flatten out  the `Login` record so avoid the nested records (ie, can RecordFields, and SecureLoginFields now die?) While that might be able to be done as a followup, the cost of re-breaking the mobile platforms might mean the opportunity to do so might not arise. IMO, the current splitting of these records makes the use of this component in the mobile code quite complicated to understand so would likely benefit from a simplification here.

(alternatively, a different split might make sense - eg, split out "editable" fields from "computed" fields, but even that might not have enough value)

components/logins/src/encryption.rs
```diff
@@ -26,24 +26,140 @@
 // multiple layers, from the app saying "sync now" all the way down to the
 // low level sync code.
 // To make life a little easier, we do that via a struct.
+//
+// Consumers of the Login component have 3 options for setting up encryption:
```
mhammond, line 30: It might make sense to offer some guidance about which of the 3 options consumers might choose, even if it requires speculation about the future (eg, you might mention how you picture desktop using it even though that doesn't exist etc)

components/logins/src/encryption.rs
```diff
@@ -26,24 +26,140 @@
...
+//       eg `LoginStore::new(ManagedEncryptorDecryptor::new(StaticKeyManager { key: myKey }))`
 
 use crate::error::*;
+use std::sync::Arc;
+
+/// This is the generic EncryptorDecryptor trait, as handed over to the Store during initialization.
+/// Consumers can implement either this generic trait and bring in their own crypto, or leverage the
+/// ManagedEncryptorDecryptor below, which provides encryption algorithms out of the box.
+///
+/// Note that EncryptorDecryptor must not call the db mutex of the LoginStore to prevent a
```
mhammond, line 46: I don't quite understand why this comment is relevant here - *nothing* probably should be taking that mutex other than the store (and, sadly, the sync engine)? If not for the sync engine that mutex could be private - there's probably ways to fix that which might be worthwhile in the future?

(I don't really care about the comment being here, I'm just trying to make sure I understand what's going on :)

components/logins/src/encryption.rs
```diff
@@ -26,24 +26,140 @@
...
+
+impl KeyManager for StaticKeyManager {
+    #[handle_error(Error)]
+    fn get_key(&self) -> ApiResult<Vec<u8>> {
+        Ok(self.key.as_bytes().into())
+    }
+}
 
 #[handle_error(Error)]
 pub fn create_canary(text: &str, key: &str) -> ApiResult<String> {
```
mhammond, line 151: The canary story can probably be better too - the intent there is to better work with keys which are our of our control. IIUC there's no equivalent thing on desktop. Maybe that concept could be optionally added to the traits? This might be good fodder for a followup.

components/logins/src/error.rs
```diff
@@ -21,9 +21,21 @@ pub enum LoginsApiError {
     #[error("No record with guid exists (when one was required): {reason:?}")]
     NoSuchRecord { reason: String },
 
+    #[error("Encryption key is missing.")]
+    MissingKey,
+
+    #[error("Encryption key is not valid.")]
+    InvalidKey,
+
     #[error("Encryption key is in the correct format, but is not the correct key.")]
```
mhammond, line 30: It's not immediately clear how "incorrect key" works with the 2 new "Failed" error cases? I wonder if that can just be replaced with the new error values?

mhammond (approved):
This is fantastic!

components/logins/src/logins.udl
jo: Yes, I think the advantages of a simplified way of working for LoginsStore outweigh the disadvantages. The idea of a locked state further complicates things and has a knock-on effect at lower levels. We can still perform certain operations without the store unlocked – for example, I added (thanks @bendk!) the method `has_logins_by_base_domain(&self, base_domain: &str)` so that the application can still check whether logins exist for a specific domain.

I like that you suggest further simplifying the login structure. I hadn't even dared to tackle that :) I'll discuss it with the mobile teams. On Android, the code is already abstracted, so the changes would be very simple here. On iOS, I'm not sure. At the latest now, a decoupling on the iOS side seems sensible to me in order to be able to make adjustments more agilely.

Furthermore, we would actually like to encrypt more fields, especially the domain. This information could already be interesting for certain attackers and some users could therefore support such changes. Although this would complicate or even prevent the lookup by domain in the locked state.

components/logins/src/encryption.rs
jo: @bendk has addressed this problem – I think that if a consumer were to implement it (we all know that this shouldn't be the case, but hey, what's possible will be implemented at some point) and use the (pub) `db` property of the `LoginStore` in the `EncyptorDecryptor` implementation, then this could result in a deadlock. Did I understand that right, @bendk? And how can we rephrase the comment to be more precise?

components/logins/src/encryption.rs
jo: Yes, very good idea!

components/logins/src/error.rs
jo: I think its used only for the `check_canary` function. To me it sounds a little strange that it throws if the key does not match - isn't it intended to just return false in that case?

mhammond: > I think its used only for the `check_canary` function. To me it sounds a little strange that it throws if the key does not match - isn't it intended to just return false in that case?

Yeah, I'd expect a valid but incorrect key to just return false there instead of an err (or *only* return the Err without any bool)

components/logins/src/encryption.rs
bendk: I raised the point since I think there's a subtle risk here.  Maybe the comment could focus more what EncryptorDecryptor implementers should not do and make the part about the mutex secondary.

Note that EncryptorDecryptor must not call any LoginStore methods.  The login store can call out to the EncryptorDecryptor when it's internal mutex is held so calling back in to the LoginStore may deadlock.

components/logins/src/error.rs
jo: ok cool, will make this change, since we're breaking anyways :)

components/logins/src/logins.udl
mhammond: Yeah! I'm a little torn by `RecordFields` though - roughly speaking, they are fields which the consumer should never *provide* - eg, when creating a login you never specify the `id` or the number of times it has been used. This is a tension we often face - use different structs for "get" vs "set" operations, which makes consumers a little more painful in some ways, but makes things clearer in others.

But certainly the flattening of `LoginFields` looks like a big improvement.

jo: Yes, I know, this is an old problem. And then, of course, there is something appealing about grouping all these meta fields, which only play a role when reading, into a separate _record_ struct. And on the other hand, there is this simplicity and clarity ¯\_(ツ)_/¯

mhammond: exactly :) It sounds like we agree there's no objectively correct answer here, so I can get behind whatever you decide!

components/logins/src/encryption.rs
jo: We would have to change our udl so that it is automatically generated with [procmacros](https://mozilla.github.io/uniffi-rs/0.27/proc_macro/index.html), which I would very much like, right?

components/logins/src/logins.udl
jo: So, I've implemented this here. It makes perfect sense to me, it will be much easier and cleaner for consumers, which can be nicely seen from the sync-pass example. Thanks again for intervening here!

components/logins/src/encryption.rs
mhammond: I don't quite understand the question, but as a meta comment, I think we are in the position to use procmacros (see #6413), so if you'd prefer killing the UDL entirely I think it would work fine.

mhammond (approved):
This is great!

jo:
> I added you as an admin I think, so hopefully you will not need re-approval for all new pushes! Note however you have an iOS test failure

that iOS test failure is a flaky one. This time, it passed. It seems as it has nothing to do with the changes of in PR.

PR #6471, opened 2024-11-13
feat: Download collections from remote-settings, sync with local packaged data.

gruberb:
Adding the functionality to be able to read local, packaged data for a `collection`. This allows us to provide configurations without being able to connect to the internet on a device, or on first start-up, where the sync is not finished yet. 

This PR also adds two commands to the cargo remote-settings CLI, so developers can work with and integrate `collections` more seamlessly.
 
### `dump-sync`
This command checks the local `components/remote_settings/dumps/{bucket}/{collection_name}.json` files and updates them with the latest remote copy if they are outdated.

The command scans the `components/remote_settings/dumps` folder, and goes through each `bucket` (represented via a folder-name), and each `{collection_name}.json`. It checks the `timestamp` of the collection and compares it to the one from the `changes` endpoint. 

* `cargo remote-settings dump-sync --create-pr` will create a local branch and push it to the repo.
* `cargo remote-settings dump-sync --dry-run` is to test a possible run, without writing the results to the file(s).

```bash
> cargo remote-settings dump-sync --dry-run
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.30s
Running `target/debug/examples-remote-settings-cli dump-sync --dry-run`

Collection main/search-telemetry-v2 is up to date

Dry run summary:
- Would update 0 collections
- 1 collections already up to date
- 0 collections not found on remote
```

Or without `--dry-run`:

```bash
> cargo remote-settings dump-sync
Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
Running `target/debug/examples-remote-settings-cli dump-sync`

Collection main/search-telemetry-v2 is up to date

Execution summary:
Collections already up to date:
  - main/search-telemetry-v2
```

### `dump-get --bucket BUCKETNAME --collection-name COLLECTION`. 
This will download a copy of a single collection from remote, into the following local folder: `components/remote_settings/dumps/{bucket}/{collection_name}.json`.

Get a copy of a single collection locally:
```bash
> cargo remote-settings dump-get --bucket main --collection-name search-config-v2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.67s

[00:00:00] ██████████████████████████████ Downloaded search-config-v2

Successfully downloaded collection to ./components/remote_settings/dumps/main/search-config-v2.json
```

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -23,6 +23,12 @@ const HEADER_BACKOFF: &str = "Backoff";
 const HEADER_ETAG: &str = "ETag";
 const HEADER_RETRY_AFTER: &str = "Retry-After";
 
+#[derive(Debug, Clone, Deserialize)]
+struct CollectionData {
+    data: Vec<RemoteSettingsRecord>,
+    timestamp: u64,
+}
+
 /// Internal Remote settings client API
 ///
 /// This stores an ApiClient implementation.  In the real-world, this is always ViaductApiClient,
@@ -57,8 +63,27 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
             }),
         }
     }
+
     pub fn collection_name(&self) -> &str {
         &self.collection_name
+    }
+
+    fn get_packaged_data(collection_name: &str) -> Option<&'static str> {
+        match collection_name {
+            // Add entries for each locally dumped collection in the `dumps/` folder.
+            // This is also the place where we want to think about a macro! and feature-gating
+            // different platforms.
+            "search-telemetry-v2" => Some(include_str!(concat!(
+                env!("CARGO_MANIFEST_DIR"),
+                "/dumps/main/search-telemetry-v2.json"
+            ))),
+            _ => None,
+        }
+    }
+
+    fn load_packaged_data(&self) -> Option<CollectionData> {
+        Self::get_packaged_data(&self.collection_name)
+            .and_then(|data| serde_json::from_str(data).ok())
     }
 
     /// Filters records based on the presence and evaluation of `filter_expression`.
@@ -87,30 +112,58 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
     pub fn get_records(&self, sync_if_empty: bool) -> Result<Option<Vec<RemoteSettingsRecord>>> {
         let mut inner = self.inner.lock();
         let collection_url = inner.api_client.collection_url();
-
-        // Try to retrieve and filter cached records first
         let cached_records = inner.storage.get_records(&collection_url)?;
+        let is_prod = inner.api_client.is_prod_server()?;
+        let packaged_data = if is_prod {
+            self.load_packaged_data()
+        } else {
+            None
+        };
 
-        match cached_records {
-            Some(records) if !records.is_empty() || !sync_if_empty => {
-                // Filter and return cached records if they're present or if we don't need to sync
-                let filtered_records = self.filter_records(records);
-                Ok(Some(filtered_records))
+        // Case 1: We have no cached records
+        if cached_records.is_none() {
+            // Case 1a: Use packaged data if available (prod only)
+            if let Some(collection) = packaged_data {
+                inner
+                    .storage
+                    .set_records(&collection_url, &collection.data)?;
+                return Ok(Some(self.filter_records(collection.data)));
             }
-            None if !sync_if_empty => {
-                // No cached records and sync_if_empty is false, so we return None
-                Ok(None)
-            }
-            _ => {
-                // Fetch new records if no cached records or if sync is required
+            // Case 1b: No packaged data - fetch from remote if sync_if_empty
+            if sync_if_empty {
                 let records = inner.api_client.get_records(None)?;
                 inner.storage.set_records(&collection_url, &records)?;
+                return Ok(Some(self.filter_records(records)));
+            }
+            return Ok(None);
+        }
 
-                // Apply filtering to the newly fetched records
-                let filtered_records = self.filter_records(records);
-                Ok(Some(filtered_records))
+        // Now we know we have cached records
+        let cached_records = cached_records.unwrap();
+        let cached_timestamp = inner.storage.get_last_modified_timestamp(&collection_url)?;
+
+        // Case 2: We have packaged data and are in prod
+        if let Some(packaged_data) = packaged_data {
+            if packaged_data.timestamp > cached_timestamp.unwrap_or(0) {
+                // Packaged data is newer
+                inner
+                    .storage
+                    .set_records(&collection_url, &packaged_data.data)?;
+                return Ok(Some(self.filter_records(packaged_data.data)));
             }
         }
+
+        // Case 3: Return cached data if we have it and either:
+        // - it's not empty
+        // - or we're not allowed to sync
+        if !cached_records.is_empty() || !sync_if_empty {
+            return Ok(Some(self.filter_records(cached_records)));
+        }
+
+        // Case 4: Cache is empty and we're allowed to sync
+        let records = inner.api_client.get_records(None)?;
+        inner.storage.set_records(&collection_url, &records)?;
+        Ok(Some(self.filter_records(records)))
     }
 
     pub fn sync(&self) -> Result<()> {
@@ -175,6 +228,9 @@ pub trait ApiClient {
 
     /// Fetch an attachment from the server
     fn get_attachment(&mut self, attachment_location: &str) -> Result<Vec<u8>>;
+
+    /// Check if this client is pointing to the production server
+    fn is_prod_server(&self) -> Result<bool>;
 }
 
 /// Client for Remote settings API requests
@@ -299,6 +355,14 @@ impl ApiClient for ViaductApiClient {
 
         let resp = self.make_request(attachments_base_url.join(attachment_location)?)?;
         Ok(resp.body)
+    }
+
+    fn is_prod_server(&self) -> Result<bool> {
+        Ok(self
+            .endpoints
+            .root_url
+            .as_str()
+            .starts_with(RemoteSettingsServer::Prod.get_url()?.as_str()))
     }
 }
 
@@ -1556,9 +1620,12 @@ mod test_new_client {
         api_client.expect_collection_url().returning(|| {
             "http://rs.example.com/v1/buckets/main/collections/test-collection".into()
         });
+        api_client.expect_is_prod_server().returning(|| Ok(false));
+
         // Note, don't make any api_client.expect_*() calls, the RemoteSettingsClient should not
         // attempt to make any requests for this scenario
         let storage = Storage::new(":memory:".into()).expect("Error creating storage");
+
         let rs_client =
             RemoteSettingsClient::new_from_parts("test-collection".into(), storage, api_client);
         assert_eq!(
@@ -1588,9 +1655,12 @@ mod test_new_client {
                 Ok(records.clone())
             }
         });
+        api_client.expect_is_prod_server().returning(|| Ok(false));
         let storage = Storage::new(":memory:".into()).expect("Error creating storage");
+
         let rs_client =
             RemoteSettingsClient::new_from_parts("test-collection".into(), storage, api_client);
+
         assert_eq!(
             rs_client.get_records(true).expect("Error getting records"),
             Some(records)
@@ -1628,6 +1698,7 @@ mod jexl_tests {
                 Ok(records.clone())
             }
         });
+        api_client.expect_is_prod_server().returning(|| Ok(false));
 
         let context = RemoteSettingsContext {
             app_version: Some("129.0.0".to_string()),
@@ -1646,6 +1717,7 @@ mod jexl_tests {
             JexlFilter::new(Some(context)),
             api_client,
         );
+
         assert_eq!(
             rs_client.get_records(false).expect("Error getting records"),
             Some(records)
@@ -1677,6 +1749,7 @@ mod jexl_tests {
                 Ok(records.clone())
             }
         });
+        api_client.expect_is_prod_server().returning(|| Ok(false));
 
         let context = RemoteSettingsContext {
             app_version: Some("127.0.0.".to_string()),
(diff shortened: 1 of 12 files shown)


components/remote_settings/src/client.rs
```diff
@@ -20,6 +21,28 @@ const HEADER_BACKOFF: &str = "Backoff";
...
+impl PackagedData {
+    pub fn from_json_str(json: &str) -> Result<Self> {
+        Ok(serde_json::from_str(json)?)
+    }
+
+    pub fn empty() -> Self {
+        Self {
+            collections: HashMap::new(),
+        }
+    }
```
bendk, line 38: How about deriving `Default` instead?

components/remote_settings/src/client.rs
```diff
@@ -20,6 +21,28 @@ const HEADER_BACKOFF: &str = "Backoff";
...
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct CollectionData {
+    records: Vec<RemoteSettingsRecord>,
+}
```
bendk, line 44: Could we also load the `last_modified` timestamp?  I think this would simplify checking if this is newer or older than the cached records.

components/remote_settings/src/jexl_filter.rs
```diff
@@ -26,9 +26,11 @@ impl JexlFilter {
     pub(crate) fn new(context: Option<RemoteSettingsContext>) -> Self {
         let env_context = match context {
             Some(ctx) => {
-                serde_json::to_value(ctx).expect("Failed to serialize RemoteSettingsContext")
+                let serialized_context =
+                    serde_json::to_value(ctx).expect("Failed to serialize RemoteSettingsContext");
+                json!({ "env": serialized_context })
             }
-            None => json!({}),
+            None => json!({ "env": {} }),
```
bendk, line 33: I think this is spillover from the other PR.

examples/remote-settings-cli/Cargo.toml
```diff
@@ -12,3 +12,10 @@ log = "0.4"
 clap = {version = "4.2", features = ["derive"]}
 anyhow = "1.0"
 env_logger = { version = "0.10", default-features = false, features = ["humantime"] }
+reqwest = { version = "0.12", features = ["json"] }
+serde_json = "1"
+futures = "0.3"
+indicatif = "0.17"
+tokio = { version = "1", features = ["full"] }
+serde = { version = "1.0.205", features = ["derive"] }
+chrono = "0.4.26"
```
bendk, line 21: I'm slightly worried about moz-central vendoring with all these new packages.  Can you test vendoring and see if there are version mismatches and/or missing audits.

components/remote_settings/src/client.rs
```diff
@@ -80,6 +107,52 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
+                if let Some(packaged_records) = self.load_packaged_data()? {
+                    if let Some(cached) = &cached_records {
+                        // Compare timestamps
+                        let cached_timestamp =
+                            cached.iter().map(|r| r.last_modified).max().unwrap_or(0);
+                        let packaged_timestamp = packaged_records
+                            .iter()
+                            .map(|r| r.last_modified)
+                            .max()
+                            .unwrap_or(0);
```
leplatrem, line 141: Don't we store the collection timestamp? 

Using the max of records timestamps won't work for deletions, because tombstones are not stored.

components/remote_settings/src/client.rs
```diff
@@ -287,6 +373,13 @@ impl ApiClient for ViaductApiClient {
         let resp = self.make_request(attachments_base_url.join(attachment_location)?)?;
         Ok(resp.body)
     }
+
+    fn is_prod_server(&self) -> bool {
+        self.endpoints
+            .root_url
+            .as_str()
+            .starts_with("https://firefox.settings.services.mozilla.com")
```
leplatrem, line 381: Can't we have constants instead of repeating this?

https://github.com/mozilla/application-services/blob/13908171a87a48c2d52c5f5d73f0a5dda55fb4b0/components/remote_settings/src/config.rs#L69

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+            anyhow::bail!("Cannot use --dry-run with --create-pr");
+        }
+
+        let collections = self.download_all().await?;
+
+        let packaged_data = PackagedData {
+            collections: collections.clone(),
+        };
+
+        let output_file = self.output_dir.join("defaults.json");
```
leplatrem, line 251: Why not having a folder with a file per collection? 
This file will end up being huge. Sometimes developers will want to open it just to double check its content compared to the server, etc.

components/remote_settings/src/client.rs
```diff
@@ -80,6 +107,52 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
         let cached_records = inner.storage.get_records(&collection_url)?;
 
+        // Check if we should load packaged data
+        let should_load_packaged = match &cached_records {
+            None => true,                                // No local data, try packaged data
+            Some(records) if records.is_empty() => true, // Empty local data, try packaged data
+            _ => false,
+        };
+
+        if should_load_packaged {
```
bendk, line 117: I don't like this `if` because you're loading the packaged data in both branches.  How about something like:

- if we're on the prod server:
   - load the data
   - if the last modified for the  packaged data is newer than the cached records:
      - use the packaged data instead and store the packaged data in the DB.

bendk (changes requested):
This is looking really good, just a couple things to work out.

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+            .as_array()
+            .context("No data array in monitor response")?
+            .iter()
+            .find(|record| {
+                record["collection"].as_str() == Some(&collection)
+                    && record["bucket"].as_str() == Some("main")
+            })
+            .context("Collection not found in monitor")?["last_modified"]
+            .as_u64()
+            .context("No last_modified in record")?;
```
leplatrem, line 92: I would move this to `download_all()`: fetch the timestamps, and pass them to `fetch_collection()` here by adding a `last_modified` parameter, in order to avoid issuing one request per collection.

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+    "search-telemetry-v2",
+    "sites-classification",
+    "top-sites",
+    "translations-models",
+    "translations-wasm",
+    "url-classifier-skip-urls",
+    "url-parser-default-unknown-schemes-interventions",
+    "urlbar-persisted-search-terms",
+    "websites-with-shared-credential-backends",
+];
```
leplatrem, line 33: Instead of having a hardcoded list, on desktop we have a folder and we use the file presents in that folder as the list to fetch.
Touching an empty and running the cargo command would update it.

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+        let monitor_response: Value = self.client.get(&monitor_url).send().await?.json().await?;
+
+        // Find the last_modified for our collection
+        let last_modified = monitor_response["data"]
+            .as_array()
+            .context("No data array in monitor response")?
+            .iter()
+            .find(|record| {
+                record["collection"].as_str() == Some(&collection)
+                    && record["bucket"].as_str() == Some("main")
```
leplatrem, line 88: You will probably have to take into account that we have other buckets: `blocklists` and `security-state`.

We usually refer to collection as `{bucket_id}/{collection_id}` (eg. `blocklists/addons-bloomfilters`, `main/cfr`, `security-state/onecrl`, ...) and then split it into bucket and collection names

leplatrem (commented):
Nice! Lots of work done already! 

Things that raised my attention:
- using the highest records timestamp is a bit fragile, I would use the timestamp value that we obtain from the changeset instead (maybe you saw that somewhere in the desktop code? ...before we didn't have the changeset endpoint, there could be a few places that would benefit from it)
- why combine all data in one file?
- dumps can be for other buckets than `main`
- the PROD url is repeated in a few places (not a major issue)

Also, as a side note, I'm working on a `./mach` command that will allow developers to edit the binary dumps, and upload them to the server. I called it `./mach remote-settings dump-sync`. Even if yours only does server to source tree, we could align on naming :)

examples/remote-settings-cli/src/downloader.rs
gruberb: It's easier to go through one file and check if a collection exist. If we have multiple, I can check if the file exists, but the name could change etc. 

I am not against it, and also don't like the "one large file", but having multiple seem to have too many downsides?

leplatrem: You seem convinced 😊 Maybe we can ask @Standard8 since his team seems interested to be able to modify the dump manually ([source](https://docs.google.com/document/d/1IIjopsfRLFlems-qDMDFO1HvSTeIOtPDHQolcuSTFvQ/edit#heading=h.t7qv62q3ujwt))

Note: The name cannot change, we cannot rename collections in Remote Settings.

components/remote_settings/src/client.rs
```diff
@@ -29,6 +51,7 @@ pub struct RemoteSettingsClient<C = ViaductApiClient> {
     collection_name: String,
     jexl_filter: JexlFilter,
     inner: Mutex<RemoteSettingsClientInner<C>>,
+    packaged_data: PackagedData,
```
Standard8, line 54: Is this going to mean that we have multiple copies (or even a single copy) of the full packaged data in memory for the lifetime of the app?

Given the `defaults.json` file is 1Mb, I'm concerned about the overhead here. On desktop, if the cache is either empty, invalid or out of date, then we'll read the dump files, otherwise, we skip loading the dumps completely.

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+                    (a_type, a_id).cmp(&(b_type, b_id))
+                });
+            }
+        }
+
+        // // Handle attachments if needed
+        // if collection == "search-config-icons" {
+        //     self.download_attachments("main", &collection, &changeset["changes"])
+        //         .await?;
+        // }
```
Standard8, line 131: We probably will want this enabled for mobile to begin with. I'm not sure why it is commented out here.

Something to note, is that ideally we should only package the attachments relevant to the platform / app to avoid unnecessary package bloat - the search-config-icons collection includes JEXL filters and at the moment we only have icons for desktop in there, but we'll soon be adding them for Android & iOS as well.

examples/remote-settings-cli/src/downloader.rs
Standard8: Desktop's `last_modified.json` file (generated with [gen_last_modified.py](https://searchfox.org/mozilla-central/source/services/settings/dumps/gen_last_modified.py)) has a list of the collections downloaded plus their timestamps. This allows you to look up in that file if the collection was downloaded, and then also check the timestamp to see if it is more recent than the local cache (e.g. for updating from local dumps when an update is released).

It is a little more work, but I think it would allow us to avoid having to load and parse the whole collection in one go, which might also have benefits especially for the invalid cache / update only what's needed on update cases.

From a developer perspective on the search team on desktop, we have a few scenarios that I think would lean towards separate files:

* We sometimes look back at the history of a collection to check roughly when it was updated. Although the date of the dump isn't necessarily the "live" date, if can give us a timeframe.
  * Similarly, if we're looking at changesets that have landed, it is easier to see that a file has changed, rather than having to dig through a large file diff for specific changes.
* For various collections, we will edit the collection locally and test it, before pushing changes to remote settings.
  * This allows us to test locally, and also run automated tests.
  * search-config-v2 and search-telemetry-v2 are quite complex for remote settings to display, hence we end up manually editing and uploading just the edited file.
* We quite often want to check or refer to something in the collection, this is easier to do in our local code editors. It is probably viable with a larger file, but they do tend to slow editors down / not be supported as well.

examples/remote-settings-cli/src/downloader.rs
```diff
@@ -0,0 +1,289 @@
...
+            if let Some(changes) = changeset["changes"].as_array_mut() {
+                changes.sort_by(|a, b| {
+                    let a_type = a["recordType"].as_str().unwrap_or("");
+                    let b_type = b["recordType"].as_str().unwrap_or("");
+                    let a_id = a["identifier"].as_str().unwrap_or("");
+                    let b_id = b["identifier"].as_str().unwrap_or("");
+                    (a_type, a_id).cmp(&(b_type, b_id))
+                });
+            }
+        }
```
Standard8, line 125: Is this also sorting the properties within the records by name? (i.e. going through all the sub-objects within the records & sorting those properties as well). I can't see that it isn't, but would like to make sure. I believe `jq` does that for us in the mozilla-central output.

As a note, I'm thinking about expanding sorting to more of the search team's collections - the sorting helps with creating clean diffs in the repository. It'll probably be my next patch for `periodic_file_updates.sh`.

examples/remote-settings-cli/src/downloader.rs
gruberb: Understand now! Thanks! Which collections should we fetch here? For example this one `anti-tracking-url-decoration` exists in the `main` and `main-preview` bucket. 

Can you pass over a list of `collections` we want to have?

examples/remote-settings-cli/src/downloader.rs
gruberb: Can you explain "touching an empty and running the cargo command"? What happens if the files don't previously exist? It's probably not going to happen because than we cannot build, but I would like these collection names at least somewhere referred to in the codebase.

examples/remote-settings-cli/src/downloader.rs
gruberb: Thanks for pointing it out. The PR is not done yet but I wanted an early review. Reading and understanding the bash script from Desktop gave me a headache, so I tried to cover as much as I can. I commented it out because I was not happy how the attachements are currently stored in-tree and want to find a better way. 

Can you point me to something more tangible how these attachments "work"? How do I know which attachements are for which platform? 

I am not 100% sure how we would feature gate attachments during build. Maybe @bendk has an idea?

components/remote_settings/src/client.rs
gruberb: I have a patch ready to split the files up. But I don't think we have anything in place to if/else load the attachments. We currently had in mind to always package them, and then during runtime decide to read from them.

bendk: I think the main difference is that as a component, we don't have the ability to package files and load them at runtime.  So the solution here is to load the file at compile time and store it in the binary.

One alternative would be that the application hands us the packaged data in some way.  Maybe Desktop could pass in the directory of the dumps.  This would allow us to determine at runtime what we want to load.  This seems a lot more complicated to me though.

A micro optimization could be to store the data in an `Arc` to limit the number of copies to 1.  I think this won't matter so much when the data is split up since each client will only get the data for their collection, but it's still possible to create 2 clients for the same collection.

examples/remote-settings-cli/src/downloader.rs
Standard8: Sorry, I meant to include [a link](https://searchfox.org/mozilla-central/rev/30afd6631333bf83d5c62ca48c313d96a21b4bbd/services/settings/dumps/main/search-config-icons.json#16).

Specifically, the `env.appinfo.ID == \"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}\"` part has the identifier for desktop.

bendk: Feature gating seems tricky, especially when the data is split up.  Maybe something like this could work?

```
/// Get a map of collection name -> packaged data
fn get_packaged_data(collection: &str) -> HashMap<String, PackageData> {
   let mut data = HashMap::new();
   #[cfg(feature = "android-package-data")]
   {
         data.insert("foo", Package_data::from_str(include_str!("../data/foo.json)"))
         data.insert("bar", Package_data::from_str(include_str!("../data/bar.json)"));
   }
   #[cfg(feature = "ios-package-data")]
   {
       ... 
   }
   data
}
```

The combination of this feature and splitting up the data adds some complexity maintaining the list of collections, but it doesn't seem too bad to me.

gruberb:
@Standard8 @leplatrem Is there a spec somewhere or a clear definition about fetching which collection, for which platform? I struggle to have a full picture. All I did so far was reverse engineer this [bash script ](https://searchfox.org/mozilla-central/source/taskcluster/docker/periodic-updates/scripts/periodic_file_updates.sh) into Rust.

But I am missing sort-of a spec of which collections we need to made available, and which bucket of this collection we need.

components/remote_settings/src/client.rs
gruberb: Where would we store these JSON files then? If we want to read them on demand, we need to have access to them. Which means we need a cross-platform way of shipping and storing these files. 

Do we do this for any other component before?

examples/remote-settings-cli/src/downloader.rs
gruberb: I removed the list and simply download all available collections. This might be too much. We can also manually download them, and just update them based on file name.

components/remote_settings/src/client.rs
bendk: We've never stored files on our own before, I don't think that's something we should be doing in this PR.

IIUC, one suggestion from @Standard8 would be to still use `include_str!` for the JSON source, but not store the parsed JSON objects in memory.  That way we have one less copy of the data in RAM.  That seems reasonable to me, since you store the data in the database when you realize the packaged data is newer.

examples/remote-settings-cli/src/downloader.rs
gruberb: Then we need to know which collections to load for each platform. Where do we get this list from? @Standard8 @leplatrem

components/remote_settings/src/client.rs
gruberb: Now we need to know which `collections` we need and want, and from which `bucket`. Can someone shed a light on this?

examples/remote-settings-cli/src/downloader.rs
leplatrem: > Can you explain "touching an empty and running the cargo command"?

Sorry if it was not clear.

My idea was, in order to package a new collection, to do this:

```
$ touch components/remote-settings/dumps/main/password-rules.json
$ cargo remote-settings download --create-pr
```

> I removed the list and simply download all available collections. This might be too much. 

Yes, maybe, unless the developers/packagers can pick at build time which collections to include. But that would be equivalent to having a hardcoded list.

> We can also manually download them, and just update them based on file name.

The idea on desktop is that we use the list of files in `services/settings/dumps/*/*.json` to determine which collections to download

examples/remote-settings-cli/src/downloader.rs
leplatrem: > Which collections should we fetch here? For example this one anti-tracking-url-decoration exists in the main and main-preview bucket.

We don't download collections from `*-preview` buckets. Only published data ends up in dumps. [More info](https://remote-settings.readthedocs.io/en/latest/tutorial-multi-signoff.html#introduction)

>  How do we know which bucket we want to get the collection from?

You could use the list returned from `/buckets/monitor/collections/changes` to know in which bucket is which collection.
But, on desktop, we scan the dump folder: https://searchfox.org/mozilla-central/source/services/settings/dumps, and get the bucket name from the path: `services/settings/dumps/{bucket}/{collection}.json`

examples/remote-settings-cli/src/downloader.rs
leplatrem: > Then we need to know which collections to load for each platform. Where do we get this list from? @Standard8 @leplatrem

This is not really our decision to make in advance. 

The RS consumers will know whether their use-case will need default data or not. Sometimes an empty list works well until the first synchronization happens.
In some use-cases, it's crucial, for example, if their code need data before first paint on new profiles. Or if the UI elements rely on records and will look bad if there was no network before the first sync.
If a collection changes often (eg. Nimbus), it makes less sense, but that's the developers' decision.

What's important is that when a developer wants to package a new collection, the procedure is crystal clear with explicit points on considerations to keep in mind (binary size, etc.).

Our first concrete use-case here is the Search config AFAIU. So let's start with this one first.

components/remote_settings/src/client.rs
```diff
@@ -57,10 +63,26 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
+
     pub fn collection_name(&self) -> &str {
         &self.collection_name
     }
 
+    fn get_packaged_data(collection_name: &str) -> Option<&'static str> {
+        match collection_name {
+            "search-telemetry-v2" => Some(include_str!(concat!(
+                env!("CARGO_MANIFEST_DIR"),
+                "/dumps/main/search-telemetry-v2.json"
```
gruberb, line 75: I still don't quite understand how this should work in the `app-services` environment. I think we create and package a Rust binary, and ship it to the target platform. Which means these files have to be shipped with it as well?

examples/remote-settings-cli/src/downloader.rs
gruberb: I would like to pick this one up once we deal with `search-config-v2`. Based on the workflow, we might add collections as needed.

components/remote_settings/src/client.rs
```diff
@@ -1695,9 +1769,325 @@ mod jexl_tests {
...
 }
+
+#[cfg(not(feature = "jexl"))]
+#[cfg(test)]
+mod cached_data_tests {
+    use super::*;
+
+    #[test]
+    fn test_no_cached_data_use_packaged_data() -> Result<()> {
+        let collection_name = "search-telemetry-v2"; // Changed to match actual file
```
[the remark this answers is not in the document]
gruberb: Oh, this is a left over comment when I wasn't sure if I used the correct file. The test ran through and I didn't clean up afterwards :)

components/remote_settings/src/client.rs
```diff
@@ -87,30 +109,62 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
-                // Apply filtering to the newly fetched records
-                let filtered_records = self.filter_records(records);
-                Ok(Some(filtered_records))
+        // Now we know we have cached records
+        let cached_records = cached_records.unwrap();
+        let cached_timestamp = cached_records
+            .iter()
+            .map(|r| r.last_modified)
+            .max()
+            .unwrap_or(0);
```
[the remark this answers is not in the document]
gruberb: It's true, I basically replicate what already know from the storage layer. But even there, we take the `max` timestamp:

```rust
for record in records {
    max_last_modified = max_last_modified.max(record.last_modified);
    let data = serde_json::to_vec(record)?;
   ...
```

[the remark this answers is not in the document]
leplatrem: You don't have this? https://github.com/mozilla/application-services/blob/ab36d7adceab85bcc344d117d6bcb7a3e0b7097d/components/remote_settings/src/storage.rs#L60-L72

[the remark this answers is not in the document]
gruberb: I do. But we set the `last_modified` based on the `records` we store. I guess we could adjust the `set_records` functionality and add a timestamp to it.

bendk (approved):
Looks great to me, thanks everyone for working through this one.

PR #6472, opened 2024-11-13
fix: set proper filter_expression format

gruberb:
The previous version didn't work against production records. These expressions look like this:
```json
filter_expression: 'env.appinfo.ID == "{ec8030f7-c20a-464f-9b0e-13a3a9e97384}" && env.version|versionCompare("128.0a1") > 0'
```

Two things were wrong:
1. We have to add `env` to our `jexl_filter` JSON
2. We have to transform `app_version` to `version`.

I like to keep our Rust struct proper (`app_version` is more accurate than just `version`).


leplatrem (approved):
LGTM :) 
I agree, `app_version` is more explicit. Alternatively, we could have both, but renaming like you did is good enough I think.

bendk (approved):
Looks great to me.

PR #6474, opened 2024-11-14
Bug 1919540 - Implement basic environment handling for the search engine selector.

Standard8:
This adds most of the environment handling necessary for the search engine selector. There are some remaining parts that will be handled in [bug 1930969](https://bugzilla.mozilla.org/show_bug.cgi?id=1930969).

This code is not currently used anywhere so no API / changelog actions are necessary.

  - Note:

### DIFF
--- a/components/search/src/configuration_types.rs
+++ b/components/search/src/configuration_types.rs
@@ -5,7 +5,7 @@
 //! This module defines the structures that we use for serde_json to parse
 //! the search configuration.
 
-use crate::{SearchEngineClassification, SearchUrlParam};
+use crate::{SearchApplicationName, SearchEngineClassification, SearchUrlParam};
 use serde::Deserialize;
 
 /// The list of possible submission methods for search engine urls.
@@ -91,12 +91,79 @@ pub(crate) struct JSONEngineBase {
     pub urls: JSONEngineUrls,
 }
 
+/// Specifies details of possible user environments that the engine or variant
+/// applies to.
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONVariantEnvironment {
+    /// Indicates that this section applies to all regions and locales. May be
+    /// modified by excluded_regions/excluded_locales.
+    #[serde(default)]
+    pub all_regions_and_locales: bool,
+
+    /// A vector of locales that this section should be excluded from. 'default'
+    /// will apply to situations where we have not been able to detect the user's
+    /// locale.
+    #[serde(default)]
+    pub excluded_locales: Vec<String>,
+
+    /// A vector of regions that this section should be excluded from. 'default'
+    /// will apply to situations where we have not been able to detect the user's
+    /// region.
+    #[serde(default)]
+    pub excluded_regions: Vec<String>,
+
+    /// A vector of locales that this section applies to. 'default' will apply
+    /// to situations where we have not been able to detect the user's locale.
+    #[serde(default)]
+    pub locales: Vec<String>,
+
+    /// A vector of regions that this section applies to. 'default' will apply
+    /// to situations where we have not been able to detect the user's region.
+    #[serde(default)]
+    pub regions: Vec<String>,
+
+    /// A vector of distribution identifiers that this section applies to.
+    #[serde(default)]
+    pub distributions: Vec<String>,
+
+    /// A vector of distributions that this section should be excluded from.
+    #[serde(default)]
+    pub excluded_distributions: Vec<String>,
+
+    /// A vector of applications that this applies to.
+    #[serde(default)]
+    pub applications: Vec<SearchApplicationName>,
+    // TODO: Implement these.
+    // pub channels: Option<Vec<String>,
+    // pub experiment: Option<String>,
+    // pub min_version: Option<String>,
+    // pub max_version: Option<String>,
+    // pub device_type: Option<String>,
+}
+
+/// Describes an individual variant of a search engine.
+#[derive(Debug, Deserialize, Clone)]
+pub(crate) struct JSONEngineVariant {
+    /// Details of the possible user environments that this variant applies to.
+    pub environment: JSONVariantEnvironment,
+}
+
 /// Represents an individual engine record in the configuration.
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub(crate) struct JSONEngineRecord {
+    /// The identiifer for the search engine.
     pub identifier: String,
+
+    /// The base information of the search engine, may be extended by the
+    /// variants.
     pub base: JSONEngineBase,
+
+    /// Describes variations of this search engine that may occur depending on
+    /// the user's environment. The last variant that matches the user's
+    /// environment will be applied to the engine, subvariants may also be applied.
+    pub variants: Vec<JSONEngineVariant>,
 }
 
 /// Represents the default engines record.
--- a/components/search/src/environment_matching.rs
+++ b/components/search/src/environment_matching.rs
@@ -1,0 +1,844 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! This module defines functions for testing if an environment from the
+//! configuration matches the user environment.
+
+use crate::{JSONVariantEnvironment, SearchApplicationName, SearchUserEnvironment};
+
+/// Matches the user's environment against the given environment from the
+/// configuration.
+///
+/// This function expects the locale, region and app version in the environment
+/// to be lower case.
+pub(crate) fn matches_user_environment(
+    environment: &JSONVariantEnvironment,
+    user_environment: &SearchUserEnvironment,
+) -> bool {
+    if !environment.excluded_distributions.is_empty()
+        && environment
+            .excluded_distributions
+            .contains(&user_environment.distribution_id)
+    {
+        return false;
+    }
+
+    matches_region_and_locale(
+        &user_environment.region,
+        &user_environment.locale,
+        environment,
+    ) && matches_distribution(
+        &user_environment.distribution_id,
+        &environment.distributions,
+    ) && matches_application(&user_environment.app_name, &environment.applications)
+}
+
+/// Determines whether the region and locale constraints in the supplied
+/// environment apply to a user given the region and locale they are using.
+fn matches_region_and_locale(
+    user_region: &str,
+    user_locale: &str,
+    environment: &JSONVariantEnvironment,
+) -> bool {
+    if does_array_include(&environment.excluded_regions, user_region)
+        || does_array_include(&environment.excluded_locales, user_locale)
+    {
+        return false;
+    }
+
+    // This is a special case, if all_regions_and_locales is false (default value)
+    // and region and locales are not set, then we assume that for the purposes of
+    // matching region & locale, we do match everywhere. This allows us to specify
+    // none of these options but match against other items such as distribution or
+    // application name.
+    if !environment.all_regions_and_locales
+        && environment.regions.is_empty()
+        && environment.locales.is_empty()
+    {
+        return true;
+    }
+
+    if does_array_include(&environment.regions, user_region)
+        && does_array_include(&environment.locales, user_locale)
+    // && environment.all_regions_and_locales
+    {
+        return true;
+    }
+
+    if environment.regions.is_empty() && does_array_include(&environment.locales, user_locale) {
+        return true;
+    }
+
+    if environment.locales.is_empty() && does_array_include(&environment.regions, user_region) {
+        return true;
+    }
+
+    if environment.all_regions_and_locales {
+        return true;
+    }
+
+    false
+}
+
+fn matches_distribution(user_distribution_id: &str, environment_distributions: &[String]) -> bool {
+    environment_distributions.is_empty()
+        || environment_distributions.contains(&user_distribution_id.to_string())
+}
+
+fn matches_application(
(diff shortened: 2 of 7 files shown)


Standard8:
Asking @mandysGit for the main review please, and @linabutler to take a look at the rust usage for us please.

components/search/src/environment_matching.rs
```diff
@@ -0,0 +1,854 @@
...
+        || does_array_include(&environment.excluded_locales, user_locale)
+    {
+        return false;
+    }
+
+    if environment.all_regions_and_locales.unwrap_or_default() {
+        return true;
+    }
+
+    // When none of the regions and locales are set. This implies its available
```
linabutler, line 56:
```suggestion
    // When none of the regions and locales are set. This implies it's available
```

(Sorry!)

components/search/src/environment_matching.rs
```diff
@@ -0,0 +1,854 @@
...
+        &user_environment.locale,
+        environment,
+    ) && matches_distribution(
+        &user_environment.distribution_id,
+        &environment.distributions,
+    ) && matches_application(&user_environment.app_name, &environment.applications)
+}
+
+/// Determines whether the region and locale constraints in the supplied
+/// environment applies to a user given the region and locale they are using.
```
linabutler, line 40:
```suggestion
/// environment apply to a user given the region and locale they are using.
```

components/search/src/types.rs
```diff
@@ -7,13 +7,14 @@
...
 
 /// The list of possible application names that are currently supported.
-#[derive(Debug, uniffi::Enum)]
+#[derive(Clone, Debug, Deserialize, PartialEq, uniffi::Enum)]
+#[serde(rename_all = "kebab-case")]
 pub enum SearchApplicationName {
     Firefox = 1,
     FirefoxAndroid = 2,
-    FirefoxIOS = 3,
+    FirefoxIos = 3,
```
linabutler, line 15: Yeah, [Rust's convention for acronyms](https://rust-lang.github.io/api-guidelines/naming.html) leads to them looking a little funky...

linabutler (approved):
The Rust parts look terrific; thanks, @Standard8!

components/search/src/selector.rs
```diff
@@ -317,4 +344,248 @@ mod tests {
...
+                  "recordType": "defaultEngines",
+                  "globalDefault": "test1",
+                }
+              ]
+            })
+            .to_string(),
+        );
+        assert!(
+            config_result.is_ok(),
+            "Should not have errored: `{config_result:?}`"
```
mandysGit, line 425: Could we make the test statement more descriptive? For example, something like, "Should verify that the correct search config is set."

Similarly with the other test statements. For filter_engine_configuration, we could have something like, "Should have filtered engine without error".

components/search/src/selector.rs
```diff
@@ -317,4 +344,248 @@ mod tests {
...
+                                search_term_param_name: Some("search".to_string())
+                            },
+                            suggestions: None,
+                            trending: None
+                        }
+                    },
+                ),
+                app_default_engine_id: "test1".to_string(),
+                app_default_private_engine_id: None
+            }, "Should have selected the single engine, as the environments do not match for the other two"
```
mandysGit, line 527: This comment doesn't seem to match the selection. It looks like there's two selected `searchEngineDefintions` that are being asserted for. 

What I'm understanding is: 
- It selected engine `test1` because that's applied to all applications.
- It also selected engine `test2` because of "applications": ["firefox-android", "focus-ios"], which matches `app_name: SearchApplicationName::FocusIos`

components/search/src/selector.rs
```diff
@@ -317,4 +344,248 @@ mod tests {
...
+                                search_term_param_name: Some("trek".to_string())
+                            },
+                            suggestions: None,
+                            trending: None
+                        }
+                    },
+                ),
+                app_default_engine_id: "test1".to_string(),
+                app_default_private_engine_id: None
+            }, "Should have selected the single engine, as the environments do not match for the other two"
```
mandysGit, line 588: Could we have more description test statements for this as well? 
Maybe naming the specific properties we're filtering for, "Should have selected engine with distro: starship, and firefox application" ?

mandysGit (changes requested):
Thank you! I learned so much Rust reviewing this patch. 

I left some suggestions and comments around code clarity, and few questions I had about things I didn't quite understand. No major blocking comments though.

components/search/src/configuration_types.rs
```diff
@@ -91,12 +91,71 @@ pub(crate) struct JSONEngineBase {
...
     pub identifier: String,
+
+    /// The base information of the search engine, may be extended by the
+    /// variants.
     pub base: JSONEngineBase,
+
+    /// Describes variations of this search engine that may occur depending on
+    /// the user's environment. The last variant that matches the user's
+    /// environment will be applied to the engine, subvariants may also be applied.
+    pub variants: Vec<JSONEngineVariant>,
```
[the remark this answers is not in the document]
Standard8: My plan was to add that in a later bug - this one is for handling of filtering for most of the environment.

components/search/src/filter.rs
```diff
@@ -56,10 +59,15 @@ pub(crate) fn filter_engine_configuration(
     let mut default_engine_id: Option<String> = None;
     let mut default_private_engine_id: Option<String> = None;
 
+    let mut user_environment = user_environment.clone();
+    user_environment.locale = user_environment.locale.to_lowercase();
+    user_environment.region = user_environment.region.to_lowercase();
+    user_environment.version = user_environment.version.to_lowercase();
+
```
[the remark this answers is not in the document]
Standard8: The application name is passed in [as an enumeration](https://github.com/mozilla/application-services/blob/ab36d7adceab85bcc344d117d6bcb7a3e0b7097d/components/search/src/types.rs#L9-L29) rather than a string, hence we don't need to lower case it here.

components/search/src/filter.rs
```diff
@@ -82,21 +90,26 @@ pub(crate) fn filter_engine_configuration(
...
-    ))
+    let base = record.base;
+    record
+        .variants
+        .iter()
+        .rev()
+        .find(|r| matches_user_environment(&r.environment, user_environment))
+        .map(|_variant| {
+            SearchEngineDefinition::from_configuration_details(&record.identifier, base)
+        })
```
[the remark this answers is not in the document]
Standard8: I left `_variant` in there because we will need to pass `variant` to `from_configuration_details` when I do the next step of handling other items in the variants (partner codes etc). The initial `_` denotes it is unused at the moment. Though I could maybe add a TODO in there.

> I had a thought to not create the base variable, since it's only access once. We could directly do record.base ?

The separate `let base = record.base;` is required because of Rust's borrowing rules / memory management and avoids the clone. I'll forward you a copy of the discussion I had about this. Splitting into two statements doesn't help because we're still hitting the memory issue.

Regarding separating those, I was confused about `.find().map()` until I found out how `map` worked in Rust. So I'm not sure separating those buys us much.

components/search/src/selector.rs
```diff
@@ -127,6 +134,11 @@ mod tests {
                     },
                     "extraField2": "123"
```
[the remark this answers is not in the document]
Standard8: These are testing that if add extra properties to the configuration in future then these will not cause existing clients to fail. The test description mentions these should be allowed, but I'll improve the assertion as well.

components/search/src/selector.rs
Standard8: Oh this was a copy & paste error. I'll improve it.

components/search/src/filter.rs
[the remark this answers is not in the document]
linabutler: > I also learned that if we're not using a parameter, we don't need to name it. So `|_variant|` can become `||`.

Not quite: we can use `|_|` if we don't want to name a parameter, but we can't use `||`. `|_|`, `|_variant|`, and `|variant|` all mean "a closure that takes one argument", but `||` means "a closure that takes zero arguments."

[`Option::map`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) takes a [closure of type](https://doc.rust-lang.org/reference/types/closure.html) `FnOnce(T) -> U`, but `||` has a type of `FnOnce() -> U`. If we tried to pass `||` to `Option::map`, we'll get a compile error:

```
error[E0593]: closure is expected to take 1 argument, but it takes 0 arguments
  --> src/main.rs:19:10
   |
19 |         .map(|| SearchEngineDefinition::from_configuration_details(&record.identifier, base))
   |          ^^^ -- takes 0 arguments
   |          |
   |          expected closure that takes 1 argument
   |
help: consider changing the closure to take and ignore the expected argument
   |
19 |         .map(|_| SearchEngineDefinition::from_configuration_details(&record.identifier, base))
   |              ~~~
```

> I had a thought to not create the `base` variable, since it's only access once. We could directly do `record.base`?

This is a very subtle one, but directly using `record.base` won't compile:

```
error[E0505]: cannot move out of `record` because it is borrowed
  --> src/main.rs:18:14
   |
10 |       record: Box<JSONEngineRecord>,
   |       ------ binding `record` declared here
11 |   ) -> Option<SearchEngineDefinition> {
12 |       let matching_variant = record
   |  ____________________________-
13 | |         .variants
   | |_________________- borrow of `record.variants` occurs here
...
18 |           .map(|_| SearchEngineDefinition::from_configuration_details(&record.identifier, record.base))
   |            --- ^^^ move out of `record` occurs here                                       ----------- move occurs due to use in closure
   |            |
   |            borrow later used by call
```

@Standard8 asked me about that on Slack, and I actually gave him a partly wrong explanation (it has nothing to do with whether the iterator is consumed or not, but it does have to do with moving `record` into the `.map()` closure while we're borrowing its data. Sorry, Mark! 😬).

I made a [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=472dc7613212da22ddf4b409fd29472c) with a few stubs, to make sure I also understood the behavior. Here's what's happening:

* `record.variants.iter()` creates an iterator of type `Iterator<Item = &JSONEngineVariant>`, meaning the iterator borrows `record.variants`, and yields `&JSONEngineVariant`s. The `&JSONEngineVariant`s are borrowed references to data from `record.variants`, so `record.variants` can't be moved or dropped until all `&JSONEngineVariant` references are dropped.

* `SearchEngineDefinition::from_configuration_details()` takes an owned `JSONRecordBase`, meaning `base` needs to be moved out of `record`. For that to work, the compiler implicitly moves `record` into the closure that we pass to `.map()`. But...

* 💥 `matching_variant` is an `Option<&JSONEngineVariant>`—it's borrowing data from `record.variants`—so `record.variants` must be valid until _after_ `.map()` returns. But the compiler implicitly moves `record` into the closure, which happens _before_ `.map()` is called. This would invalidate the option's `&JSONEngineVariant` reference, so the borrow checker rejects the code.

We can work around that in a few ways:

1. Move `base` out of `record`, before borrowing `record.variants`. That's what the `base` variable does.
2. Clone the `Option<&JSONEngineVariant>`, so that it becomes an `Option<JSONEngineVariant>`, and avoids the borrow before move: `.rev().find(...).cloned()`.
3. Clone `record.variants`, and turn it into an iterator of owned variants (`Iterator<Item = JSONEngineVariant>`): `record.variants.clone().into_iter().rev()...`.
4. Clone `record.base`, which @Standard8 had originally: `SearchEngineDefinition::from_configuration_details(&record.identifier, record.base.clone())`.

All of these are OK, but (1) avoids an extra allocation, even though it looks a little weird. Another version of (1) is to explicitly move all the fields we need out of `record`, instead of just `base`:

```rust
let JSONEngineRecord {
    identifier,
    variants,
    base,
} = *record;
let matching_variant = variants
    .into_iter()
    .rev()
    .find(|variant| matches_user_environment(&variant.environment, user_environment));
matching_variant.map(|_| SearchEngineDefinition::from_configuration_details(&identifier, base))
```

I included the other options in the Playground link also! 😊

components/search/src/configuration_types.rs
```diff
@@ -91,12 +91,71 @@ pub(crate) struct JSONEngineBase {
...
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONVariantEnvironment {
+    /// Indicates that this section applies to all regions and locales. May be
+    /// modified by excludedRegions/excludedLocales.
+    pub all_regions_and_locales: Option<bool>,
+
+    /// A vector of locales that this section should be excluded from. 'default'
+    /// will apply to situations where we have not been able to detect the user's
+    /// locale.
+    pub excluded_locales: Option<Vec<String>>,
```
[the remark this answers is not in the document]
Standard8: It is indeed simpler, and actually highlights a bit where the search config probably isn't as good as it could be (the all regions & locales boolean treated as tri-state). However, I've got it all working & it does simplify it down, so going without the options.

We might be able to change more of the configuration_types to be option-less, however, I'll only do the environment ones at the moment, since that's what this patch is about.

Standard8:
Thank you both for the reviews, I've updated for the comments and believe I've addressed everything @mandysGit

components/search/src/selector.rs
```diff
@@ -317,4 +356,262 @@ mod tests {
...
+                                search_term_param_name: Some("q".to_string())
+                            },
+                            suggestions: None,
+                            trending: None
+                        }
+                    },
+                ),
+                app_default_engine_id: "test1".to_string(),
+                app_default_private_engine_id: None
+            }, "Should have selected the single engine, as the environments do not match for the other two"
```
mandysGit, line 483: I might have missed this one in my earlier review. But I think this can be revised to more specific wording: 

"Should have selected test1 for all matching locales, as the environments do not match for the other two"

mandysGit (approved):
Thank you! This looks great!

One last thing - could you change the comment for one of the assertions? I've left a suggestion below. 

It looks good after that. Approved.

PR #6475, opened 2024-11-15
feat: feature gate jexl

gruberb:
Putting `jexl` module behind a feature flag.


gruberb:
> looks great to me, but I'm mildly surprised to not see anyone actually enable that feature?

Could be that we just added `JEXL` a week ago, so no-one is making use of it yet?

PR #6476, opened 2024-11-15
Update AGP to version 8.7.2

rvandermeulen:


bendk:
I think I know what's happening, but I need some gradle help to figure out a solution.

Before the AGP upgrade, I see gradle copying `libmegazord.so` to `components/crashtest/android/build/intermediates/java_res/debugUnitTest/out/linux-x86-64/libmegazord.so`.  After the upgrade I don't see that file there anymore.  That means when the app-services code tries to load the megazord library, it results in an `UnsatisfiedLinkError`.

I think the issue is with [this line](https://github.com/mozilla/application-services/blob/4ccf7de4292dbe0a0580121baaf64fc9094043e4/publish.gradle#L258).  This line was working for AGP 8.0, but stops working after 8.1.

That code seems pretty hacky/brittle to me and I think there are better ways to do this. Maybe something to do with a configuration or artifact?  However, this is where my gradle knowledge ends.  Can someone point me in the right direction?

bendk:
After more investigation I think the issue is simply that AGP is excluding the .so file when it copies it.

I changed that line to point to an absolute/static directory on my machine and copied the expected contents there -- still no libmegazord.so.  Then I tried adding foo.txt file in the directory and found that gradle did copy it over.  Interestingly, libmegazord.dylib also worked, maybe that's why this is working on macs?

One guess I had was that the abiFilters property which was causing gradle to skip the library, but it doesn't seem so.  I tried adjusting that property and also changing the subdirectory name to x86_64 which I don't think should be filtered by default, but still no success.
Any ideas on what I can try next?  It's really feeling like I'm fighting against gradle here.

After I did this, I noticed that @badboy reported almost exactly the same issue when trying to upgrade Glean: https://github.com/mozilla/glean/pull/2680#issuecomment-2056627683

PR #6479, opened 2024-11-18
Bug 1931812 - Don't reingest the DB on each query benchmark

0c0w3:
This addresses the bug and also makes one other change: I added a new method to `BenchmarkWithInput` called `global_input()` that returns some data that's then passed by reference to the benchmark's iterations. That's a little better than creating global data in benchmark constructors because it means only the global data for the current benchmark is in memory instead of the global data for all benchmarks in the group. In the case of query benchmarks, the global data is the store, so we avoid having N stores in memory at once, N = the number of benchmarks in the group. (I'm adding quite a few weather benchmarks.) It also made sense to rename `generate_input()` to `iteration_input()` to emphasize that that method is called on each iteration.

I'm open to reverting this other change though if you disagree.

The speedup between this and the main branch when running `time cargo suggest-bench -- query/query` isn't dramatic but it's a little better:

```
main branch: ~83s
this branch: ~68s (~18% improvement)
```

  - Note:


bendk (approved):
Looks great to me, I like this change.

Re-using the store means that each iteration will benefit from caching on the last run.  However, this isn't much different than before, since even if you create a new store the OS disk cache will still be warmed up.  If anything, it's an improvement because it's more consistent.

PR #6483, opened 2024-11-20
Bug 1931963 - Add UniFFI bindings for fetch_geonames() in suggest

0c0w3:
I didn't expect to be adding bindings for this, and there might be better ways to support the use case described in bug than exposing `fetch_geonames()` as is, but this works for now at least, and we need a fix for Firefox 134. Desktop will call this to validate cities and states returned by the Yelp ML model. Not only to validate but also to do prefix matching, since we can get that for free.

I considered adding a new function that matches both cities and regions at once so that desktop doesn't need to call into suggest twice and ideally so that we run only one SQL query. But I don't really like the idea of having a completely different code path for this. Or, it might be possible to rewrite the current function so it works that way. But that ends up getting a little messy pretty quickly, and the weather use case doesn't need it. So for now I did the simplest thing.

I also considered making geonames a full-fledged `SuggestionProvider` and `Suggestion`, but then we'd have to add more options to `SuggestionProviderConstraints` to support the `fetch_geonames()` params, and these aren't really standalone suggestions anyway.

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -20,6 +20,7 @@ use crate::{
     config::{SuggestGlobalConfig, SuggestProviderConfig},
     db::{ConnectionType, IngestedRecord, Sqlite3Extension, SuggestDao, SuggestDb},
     error::Error,
+    geoname::{Geoname, GeonameMatch, GeonameType},
     metrics::{MetricsContext, SuggestIngestionMetrics, SuggestQueryMetrics},
     provider::{SuggestionProvider, SuggestionProviderConstraints, DEFAULT_INGEST_PROVIDERS},
     rs::{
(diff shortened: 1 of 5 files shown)


components/suggest/src/store.rs
```diff
@@ -261,6 +259,18 @@ impl SuggestStore {
     ) -> SuggestApiResult<Option<SuggestProviderConfig>> {
         self.inner.fetch_provider_config(provider)
     }
+
+    #[handle_error(Error)]
+    pub fn fetch_geonames(
```
ncloudioj, line 264: Can we get a docstring for this API since this would be the only place for documentation?

components/suggest/src/store.rs
```diff
@@ -470,6 +469,27 @@ impl<S> SuggestStoreInner<S> {
...
+
+    fn fetch_geonames(
+        &self,
+        query: &str,
+        match_name_prefix: bool,
+        geoname_type: Option<GeonameType>,
+        filter: Option<Vec<Geoname>>,
+    ) -> Result<Vec<GeonameMatch>> {
+        self.dbs()?.reader.read(|dao| {
+            if let Some(f) = filter {
```
ncloudioj, line 481: You can do:

```rs
            dao.fetch_geonames(
                query,
                match_name_prefix,
                geoname_type,
                filter.as_ref().map(|f| f.iter().collect())
            )
```

0c0w3: Ah thanks!

PR #6486, opened 2024-11-21
NDK 28.0.12674087 [ci full]

rvandermeulen:
Nothing to see here, just a test to see if NDK 28 (currently in prerelease) does indeed give us 16 KB page sizes for free...


rvandermeulen:
```
llvm-objdump.exe -p libmegazord.so | Select-String -Pattern "LOAD"

    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**14
    LOAD off    0x000000000038df20 vaddr 0x0000000000391f20 paddr 0x0000000000391f20 align 2**14
    LOAD off    0x00000000009e7fa0 vaddr 0x00000000009effa0 paddr 0x00000000009effa0 align 2**14
    LOAD off    0x0000000000a639b0 vaddr 0x0000000000a6f9b0 paddr 0x0000000000a6f9b0 align 2**14
```

Huzzah! Per the [docs](https://developer.android.com/guide/practices/page-sizes#alignment-use-tools), it looks like we indeed get the needed changes for free with NDK 28. As expected, 133.0.1 says `align 2**12` instead when I look at that libmegazord.so.

PR #6490, opened 2024-11-22
feat: store remote-settings attachments in local storage

gruberb:
Solves: https://bugzilla.mozilla.org/show_bug.cgi?id=1929659

Hooking up `attachments` to the `sqlite` storage in the remote-settings component. 

I also changed the name of the `Attachment` struct to `AttachmentMetadata`. I tried to do the same for the `RemoteSettingsRecord`, but we had to do some serialziation magic, and it seemed to cumbersome just for renaming things.

The idea is to be intentional that what's part of the `RemoteSettingsRecord` is not the `attachment` itself, but just the metadata for an `attachment`. The Remote Settings server response is calling it `attachment`, and therefore we adopt the name here as well.

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -177,10 +177,27 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
     /// Downloads an attachment from [attachment_location]. NOTE: there are no guarantees about a
     /// maximum size, so use care when fetching potentially large attachments.
     pub fn get_attachment(&self, attachment_location: &str) -> Result<Vec<u8>> {
-        self.inner
+        let mut inner = self.inner.lock();
+        let collection_url = inner.api_client.collection_url();
+
+        if let Some(attachment) = inner
+            .storage
+            .get_attachment(&collection_url, attachment_location)?
+        {
+            return Ok(attachment);
+        }
+
+        let attachment = self
+            .inner
             .lock()
             .api_client
-            .get_attachment(attachment_location)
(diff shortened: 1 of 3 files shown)


components/remote_settings/src/client.rs
```diff
@@ -659,15 +676,15 @@ pub struct RemoteSettingsRecord {
...
+    pub attachment: Option<AttachmentMetadata>,
     #[serde(flatten)]
     pub fields: RsJsonObject,
 }
 
 /// Attachment metadata that can be optionally attached to a [Record]. The [location] should
 /// included in calls to [Client::get_attachment].
 #[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, uniffi::Record)]
-pub struct Attachment {
+pub struct AttachmentMetadata {
```
bendk, line 687: I like this name better, but technically it's a breaking change for our consumers.   Can you do a couple things for this?
- Call out the change in `CHANEGLOG`
- Test building firefox-ios and firefox-android with these changes and see if anything breaks

bendk (approved):
Looks great, the only potential issue is the name change.  I'm going to approve, but let's test building the consumer apps before merging this one.

components/remote_settings/src/client.rs
gruberb: * Added a `CHANGELOG`
* Tested it on `firefox-ios` and could built it with the local `app-services`

gruberb: Reverted the change.

PR #6494, opened 2024-11-25
Bug 1933217 - Implement suggestion and trending URL handling on the search engine selector.

Standard8:
I didn't add these initially to keep the original patch a bit simpler, but they're easy to add now.

  - Note:

No changelog necessary as search engine selector isn't being used yet.

### DIFF
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -26,8 +26,8 @@ impl From<JSONEngineUrls> for SearchEngineUrls {
     fn from(urls: JSONEngineUrls) -> Self {
         Self {
             search: urls.search.into(),
-            suggestions: None,
-            trending: None,
+            suggestions: urls.suggestions.map(|suggestions| suggestions.into()),
+            trending: urls.trending.map(|trending| trending.into()),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


components/search/src/filter.rs
```diff
@@ -183,8 +183,26 @@ mod tests {
...
+                            experiment_config: Some("lang-param".to_string()),
+                        }]),
+                        search_term_param_name: Some("suggest".to_string()),
+                    }),
+                    trending: Some(JSONEngineUrl {
+                        base: "https://example.com/trending".to_string(),
+                        method: Some(crate::JSONEngineMethod::Get),
+                        params: Some(vec![SearchUrlParam {
+                            name: "trend".to_string(),
+                            value: Some("now".to_string()),
```
mandysGit, line 201: How about changing `trend` to `trend-name`and `now` to `trend-value` might be more clear?

components/search/src/filter.rs
```diff
@@ -183,8 +183,26 @@ mod tests {
...
                     },
-                    suggestions: None,
-                    trending: None,
+                    suggestions: Some(JSONEngineUrl {
+                        base: "https://example.com/suggestions".to_string(),
+                        method: Some(crate::JSONEngineMethod::Get),
+                        params: Some(vec![SearchUrlParam {
+                            name: "lang".to_string(),
+                            value: None,
+                            experiment_config: Some("lang-param".to_string()),
```
mandysGit, line 192: How about changing the name and experiment_config? 
`lang` to `suggest-name`and `now` to `suggest-experiment-value` might be more clear?

components/search/src/selector.rs
```diff
@@ -247,6 +247,23 @@ mod tests {
                         "base": "https://example.com/1",
                         "method": "GET",
                         "searchTermParamName": "q"
+                      },
+                      "suggestions": {
+                        "base": "https://example.com/suggestions",
+                        "method": "POST",
+                        "params": [{
+                          "name": "type",
+                          "value": "space",
```
mandysGit, line 256: Wondering if we could we name these to `suggestion-name` and `suggestion-value`?

components/search/src/selector.rs
```diff
@@ -247,6 +247,23 @@ mod tests {
...
+                        }],
+                        "searchTermParamName": "suggest"
+                      },
+                      "trending": {
+                        "base": "https://example.com/trending",
+                        "method": "GET",
+                        "params": [{
+                          "name": "area",
+                          "experimentConfig": "area-param",
```
mandysGit, line 265: Wondering if we could we rename these to `trending-name` and `tending-experiment-value`?

mandysGit (approved):
Thank you - I have a few suggestions in renaming the trending / suggestions names, values, and experiment-config. Rather than having some random value could we be more explicit. What do you think? :)

PR #6495, opened 2024-11-26
Bug 1930984 - Implement default search engine selection on the search engine selector.

Standard8:
This adds handling of the default search engines from the search configuration.

This changes the object returned by `filter_engine_configuration` to use an `Option` for `app_default_engine_id` as that allows us to highlight there may have been an issue with the configuration - in theory, this should never happen, but just in case, I'd rather the application had potential to be aware of it.

No API/changelog necessary as this isn't currently used.

  - Note:

### DIFF
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -6,9 +6,9 @@
 
 use crate::environment_matching::matches_user_environment;
 use crate::{
-    error::Error, JSONEngineBase, JSONEngineRecord, JSONEngineUrl, JSONEngineUrls,
-    JSONSearchConfigurationRecords, RefinedSearchConfig, SearchEngineDefinition, SearchEngineUrl,
-    SearchEngineUrls, SearchUserEnvironment,
+    error::Error, JSONDefaultEnginesRecord, JSONEngineBase, JSONEngineRecord, JSONEngineUrl,
+    JSONEngineUrls, JSONSearchConfigurationRecords, RefinedSearchConfig, SearchEngineDefinition,
+    SearchEngineUrl, SearchEngineUrls, SearchUserEnvironment,
 };
 
 impl From<JSONEngineUrl> for SearchEngineUrl {
@@ -56,13 +56,13 @@ pub(crate) fn filter_engine_configuration(
     configuration: Vec<JSONSearchConfigurationRecords>,
 ) -> Result<RefinedSearchConfig, Error> {
     let mut engines = Vec::new();
-    let mut default_engine_id: Option<String> = None;
-    let mut default_private_engine_id: Option<String> = None;
 
     let mut user_environment = user_environment.clone();
     user_environment.locale = user_environment.locale.to_lowercase();
     user_environment.region = user_environment.region.to_lowercase();
     user_environment.version = user_environment.version.to_lowercase();
+
+    let mut default_engines_record = None;
 
     for record in configuration {
         match record {
@@ -71,8 +71,7 @@ pub(crate) fn filter_engine_configuration(
                 engines.extend(result);
             }
             JSONSearchConfigurationRecords::DefaultEngines(default_engines) => {
-                default_engine_id = Some(default_engines.global_default);
-                default_private_engine_id.clone_from(&default_engines.global_default_private);
+                default_engines_record = Some(default_engines);
             }
             JSONSearchConfigurationRecords::EngineOrders(_engine_orders) => {
                 // TODO: Implementation.
@@ -83,10 +82,13 @@ pub(crate) fn filter_engine_configuration(
         }
     }
 
+    let (default_engine_id, default_private_engine_id) =
+        determine_default_engines(&engines, default_engines_record, &user_environment);
+
     Ok(RefinedSearchConfig {
         engines,
-        app_default_engine_id: default_engine_id.unwrap(),
-        app_default_private_engine_id: default_private_engine_id,
+        app_default_engine_id: default_engine_id,
+        app_private_default_engine_id: default_private_engine_id,
     })
 }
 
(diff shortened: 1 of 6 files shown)


components/search/src/filter.rs
```diff
@@ -217,4 +295,401 @@ mod tests {
             }
         )
     }
+
+    lazy_static! {
+        static ref ENGINES_LIST: Vec<SearchEngineDefinition> = vec![
```
Standard8, line 300: Using lazy_static seemed to be the simplest way of creating this as a global so that it could be used across multiple tests - `SearchEngineDefinition` properties all need to be `String`s rather than `str`s, so need to have dynamic allocation afaict.

@linabutler If there's a better way of doing this, I'm open to improvements.

linabutler: You're totally right; we can't avoid the dynamic allocation here. `lazy_static` is ["soft-deprecated"](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214); the standardized replacement for it is [`std::sync::LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html), but that's only stable in 1.80+.

In our newer components, we've been using [`once_cell`](https://docs.rs/once_cell/latest/once_cell). [`once_cell::sync::Lazy`](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) should do the trick; `std::sync::LazyLock` is based on it!

```rs
static ENGINES_LIST: Lazy<Vec<SearchEngineDefinition>> = Lazy::new(|| {
    vec![
        SearchEngineDefinition {
            // ...
        },
        SearchEngineDefinition
            // ...
        },
    ]
});
```

components/search/src/filter.rs
```diff
@@ -108,11 +110,87 @@ fn maybe_extract_engine_config(
...
+                    let engine_id = find_engine_with_match(engines, data.default);
+                    if engine_id.is_some() {
+                        default_engine_id = engine_id;
+                    }
+                    let private_engine_id = find_engine_with_match(engines, data.default_private);
+                    if private_engine_id.is_some() {
+                        default_engine_private_id = private_engine_id;
+                    }
+                }
+            }
```
linabutler, line 137:
```suggestion
            let (default_engine_id, default_engine_private_id) = record.specific_defaults
                .into_iter()
                .filter(|data| matches_user_environment(&data.environment, user_environment))
                .fold((None, None), |(default_engine_id, default_engine_private_id), data| {
                    (
                        // Check the engine is present in the list of engines before
                        // we return it as default.
                        find_engine_with_match(engines, data.default).or(default_engine_id),
                        find_engine_with_match(engines, data.default_private).or(default_engine_private_id),
                    )
                });
```

I like using `fold()` to extract multiple bits of data like this, but either way is great with me—whichever you like more!

linabutler (changes requested):
Looks great to me, thank you! Just a couple of idiomatic Rust suggestions, plus switching over to `once_cell`.

linabutler (approved):
Terrific, thank you!

components/search/src/filter.rs
```diff
@@ -108,11 +110,91 @@ fn maybe_extract_engine_config(
...
+                // we return it as default.
+                let engine_id = find_engine_with_match(engines, specific_default.default);
+                if engine_id.is_some() {
+                    default_engine_id = engine_id;
+                }
+                let private_engine_id =
+                    find_engine_with_match(engines, specific_default.default_private);
+                if private_engine_id.is_some() {
+                    default_engine_private_id = private_engine_id;
+                }
```
mandysGit, line 141: I noticed we're calling `is_some` to check if we've found a matching engine. Could we do the same as line `130` and use an `if let Some` here instead? I was curious and read: https://doc.rust-lang.org/stable/book/ch06-03-if-let.html maybe that could work and make the code more concise / clear? Thoughts? 
```
if let Some(engine_id) = find_engine_with_match(engines, specific_default.default) {
    default_engine_id = engine_id;
}
if let Some(private_engine_id) = find_engine_with_match(engines, specific_default.default_private) {
    default_engine_private_id = private_engine_id;
}

```

components/search/src/filter.rs
```diff
@@ -108,11 +110,91 @@ fn maybe_extract_engine_config(
...
+                // then fall back to the global default engine - but only if that
+                // exists in the engine list.
+                //
+                // For the normal mode engine (`default_engine_id`), this would
+                // effectively be considered an error. However, we can't do anything
+                // sensible here, so we will return `None` to the application, and
+                // that can handle it.
+                default_engine_id.or_else(|| engine_id_if_found(engines, record.global_default)),
+                default_engine_private_id
+                    .or_else(|| engine_id_if_found(engines, record.global_default_private)),
```
mandysGit, line 155: Two questions: 

1. I don't quite understand the second paragraph -- what is "normal mode engine"?  If it uses the global default as the default engine in this situation, why is that considered an error?  

2. Remind me again, how does the application layer handle `None` when we're in the situation with no default engine found?

components/search/src/filter.rs
```diff
@@ -108,11 +110,91 @@ fn maybe_extract_engine_config(
...
+            )
+        }
+    }
+}
+
+fn engine_id_if_found(engines: &[SearchEngineDefinition], engine_id: String) -> Option<String> {
```
mandysGit, line 161: What are your thoughts on renaming this function to `get_engine_id` or `find_engine_id`? 

I find the `engine_id_if_found` , specifically the `if_found` part repetitive because the `-> Option<String>` indicates it could return a String or None if not found.

components/search/src/filter.rs
```diff
@@ -217,4 +299,458 @@ mod tests {
...
+        );
+
+        let (default_engine_id, default_engine_private_id) = determine_default_engines(
+            &ENGINES_LIST,
+            Some(JSONDefaultEnginesRecord {
+                global_default: "engine2".to_string(),
+                global_default_private: String::new(),
+                specific_defaults: vec![JSONSpecificDefaultRecord {
+                    default: "engine1".to_string(),
+                    default_private: String::new(),
```
mandysGit, line 425: Why this default_private is `String::new()` and the `global_default_private` above on line 395 is `"".to_string()`?

Could we keep these consistent and use one way to define the string, otherwise if there's a reason to use different methods in Rust -- I'm fine with that too. I wasn't sure why they're different. I'm wondering if it's because they are different types or something else?

components/search/src/filter.rs
```diff
@@ -217,4 +299,458 @@ mod tests {
...
+                experiment: String::new(),
+                app_name: SearchApplicationName::Firefox,
+                version: String::new(),
+            },
+        );
+
+        assert_eq!(
+            default_engine_id.unwrap(),
+            "engine2",
+            "Should have returned the global default engine when specific environments match, but do not override the global"
```
mandysGit, line 706: I understand we're testing the default_engine_id is "engine2" but how does this test make sure global default `engine2` is not overridden? 

Same comment for the test below.

components/search/src/filter.rs
```diff
@@ -217,4 +299,458 @@ mod tests {
...
+        );
+        assert_eq!(
+            default_engine_private_id.unwrap(),
+            "engine1",
+            "Should have returned the specific default engine for private mode when environments match"
```
mandysGit, line 711: Would it be helpful to add comment specifying the matched environments?

For example, 
"Should have returned the specific default engine for private mode for matched locale Fi and region FR"

I noticed we do that in some of the test in Desktop. I think it helps with making the assertion more explicit. That way, readers can quickly see the matching environment properties without needing to scan the code. Thoughts?

mandysGit (changes requested):

mandysGit:
Thank you for writing this. Interesting to see how the filtering is done in Rust for default engines. 

I was reviewing [test_engine_selector_defaults.js#311-321](https://searchfox.org/mozilla-central/rev/db4c501abb679a81c921510d7e38a8028dd6024a/toolkit/components/search/tests/xpcshell/test_engine_selector_defaults.js#311-321) and wanted to ask if you think we should add a test for matching distro? So we can have the same tests are that file?

components/search/src/filter.rs
linabutler: `default_engine_id` and `default_engine_private_id` are options, so it would be a little wordier to assign to them directly:

```rust
if let Some(engine_id) = find_engine_with_match(engines, specific_default.default) {
    default_engine_id = Some(engine_id);
}
```

But [`Option::replace`](https://doc.rust-lang.org/std/option/enum.Option.html#method.insert) can help!

```rust
if let Some(engine_id) = find_engine_with_match(engines, specific_default.default) {
    default_engine_id.replace(engine_id);
}
```

components/search/src/filter.rs
linabutler: >  if there's a reason to use different methods in Rust -- I'm fine with that too

They're equivalent! 😊 And so are all these:

```
String::new()
"".to_string()
"".to_owned()
"".into() // When assigning to a `String`.
String::from("")
```

components/search/src/filter.rs
Standard8: 1. Normal mode = not in private browsing mode. It is considered an error if we don't have a default engine, as we should always have an application default engine specified.

2. It is up to the application layer. On desktop we fall back to the first "general" search engine found (iirc according to alphabetical order).

components/search/src/filter.rs
Standard8: I think there's a balance between making the comment too long vs being explicit. The focus point for the message is that we should be doing something when the environments match - that's the point of the test. The details of the actual matching for the environments are doesn't actually matter for the test.

I think it'll help to switch these tests to use `..Default::default()` as that'll help highlight the bits we are matching on. However I discovered that after writing this patch and I'm planning on addressing it in a follow-up.

Standard8:
> I was reviewing [test_engine_selector_defaults.js#311-321](https://searchfox.org/mozilla-central/rev/db4c501abb679a81c921510d7e38a8028dd6024a/toolkit/components/search/tests/xpcshell/test_engine_selector_defaults.js#311-321) and wanted to ask if you think we should add a test for matching distro? So we can have the same tests are that file?

From a unit testing perspective, I don't think that's necessary. `environment_matching.rs` has the tests for checking that the environments match in appropriate conditions (e.g. when distributions are specified).

We're using the exact same `matches_user_environment` for matching the specific defaults. As that's the case, I'm not convinced there's much additional value another test here would add - except for checking that you can indeed match on distribution and someone hasn't re-implemented `matches_user_environment` separately - which I think would be unlikely.

I do notice, however, that I don't have an integration test involving `specificDefaults` in the selector, which is what I was intended the tests in that file to be - so I'll add something there.

components/search/src/filter.rs
mandysGit: Oh I see. Yeah for no.1 if we fall back to the global default, it means the user is in some kind of environment where it didn't match any engines in the config or something else went wrong, and no default appProvided search engine was found for them.

components/search/src/filter.rs
mandysGit: Thank you both!

components/search/src/filter.rs
mandysGit: The other patch does make it more clear with `..Default::default()`. Thanks for adding that.

components/search/src/filter.rs
mandysGit: Actually, I'm still not understanding this assertion statement because I don't understand why or how the global engine would ever become overridden?

mandysGit (approved):
Thank you. Changes look good to me 👍🏼 

There's one assertion statement I don't quite understand - I left it as unresolved. It might be I'm misinterpreting it, if you could help me clarify what you mean there?

components/search/src/filter.rs
Standard8: We're trying to test that if the specific environment is overriding the private default, then it is not overriding the global default as well.

I've updated the wording for that.

PR #6497, opened 2024-11-27
feat: read attachments from tree

gruberb:
* Add ability to read local attachments stored in tree, and check against hash and size if they are current. If not, fetch from remote.
* Adding `regions` collections plus attachments in-tree.
* Change storage layer to store metadata for attachments.
* Moved read_from_local_* functions into macros! to easier add future collections
* Add two tests for the local attachments


components/remote_settings/src/macros.rs
```diff
@@ -0,0 +1,41 @@
...
+               ))),)*
+               _ => None,
+           }
+       }
+   };
+}
+
+#[macro_export]
+macro_rules! packaged_attachments {
+    () => {
```
gruberb, line 20: I left this in, since I am not sure if we keep the `regions` collections + attachments in tree. Can update the PR once we decide.

CHANGELOG.md
```diff
@@ -40,6 +40,9 @@
   the page before recording the observation, set the new `if_page_missing`
   option to `HistoryMetadataPageMissingBehavior::InsertPage`.
 
+### Remote Settings
+- Changed the name of the `Attachment` struct (used inside `RemoteSettingsRecord`) to `AttachmentMetadata` to better reflect the purpose.
```
gruberb, line 44: Checked against `firefox-android` and it breaks. Don't exactly know how to notify the teams (slacked them earlier).

bendk (changes requested):
Looks great overall, I just had a couple requests.

bendk:
FWIW, it might have been good to split out the `Attachment` -> `AttachmentMetadata` changes into it's own PR.   A colleague shared [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/) with me last year and I've been a big fan of it.

gruberb:
> FWIW, it might have been good to split out the Attachment -> AttachmentMetadata changes into it's own PR

Very good point! I want to generally clean up the `remote-settings` component and split out the `client.rs` into smaller modules.

components/remote_settings/src/client.rs
```diff
@@ -46,6 +49,27 @@ struct RemoteSettingsClientInner<C> {
...
 
+// Add your local packaged data you want to work with here
+impl<C: ApiClient> RemoteSettingsClient<C> {
+    // One line per bucket + collection
+    packaged_collections! {
+        ("main", "search-telemetry-v2"),
+        ("main", "regions"),
+    }
+
+    // You have to speciffy
```
leplatrem: typo :) 
```suggestion
    // You have to specify
```

components/remote_settings/src/error.rs
```diff
@@ -47,6 +47,8 @@ pub enum Error {
     ConfigError(String),
     #[error("Database error: {0}")]
     DatabaseError(#[from] rusqlite::Error),
+    #[error("No attachement in given record: {0}")]
```
leplatrem, line 50: typo :) 
```suggestion
    #[error("No attachment in given record: {0}")]
```

leplatrem, line 50: This error covers a bit more than that. Should we have a distinct error like `RecordAttachmentMismatchError` for when downloaded data doesn't match the server?

components/remote_settings/src/macros.rs
leplatrem: This example data is great because it almost never changes, but it's also kind of big. 
Search config icons are smaller in size, and will eventually be needed AFAIU, but I admit that the list is kind of long and not so elegant to serve as an example

leplatrem (approved):
Nice 💯

bendk (approved):
Changes look great!

PR #6500, opened 2024-11-28
Bug 1930969 - Implement remaining environment matching in the search engine selector

Standard8:
This implements the remaining environment matching in the search engine selector. This also gets the desktop `test_engine_selector_environment.js` passing.

Note that we do not add support for `optional` here, as that isn't really an environment variable and should be handled by the application itself according to what it does.

  - Note:

### DIFF
--- a/components/search/src/environment_matching.rs
+++ b/components/search/src/environment_matching.rs
@@ -5,7 +5,8 @@
 //! This module defines functions for testing if an environment from the
 //! configuration matches the user environment.
 
-use crate::{JSONVariantEnvironment, SearchApplicationName, SearchUserEnvironment};
+use crate::{JSONVariantEnvironment, SearchUserEnvironment};
+use firefox_versioning::version::Version;
 
 /// Matches the user's environment against the given environment from the
 /// configuration.
@@ -16,6 +17,10 @@ pub(crate) fn matches_user_environment(
     environment: &JSONVariantEnvironment,
     user_environment: &SearchUserEnvironment,
 ) -> bool {
+    if !environment.experiment.is_empty() && user_environment.experiment != environment.experiment {
+        return false;
+    }
+
     if !environment.excluded_distributions.is_empty()
         && environment
             .excluded_distributions
@@ -28,10 +33,16 @@ pub(crate) fn matches_user_environment(
         &user_environment.region,
         &user_environment.locale,
         environment,
-    ) && matches_distribution(
-        &user_environment.distribution_id,
+    ) && is_empty_or_contains(
         &environment.distributions,
-    ) && matches_application(&user_environment.app_name, &environment.applications)
+        &user_environment.distribution_id,
+    ) && matches_version(
+        &user_environment.version,
+        &environment.min_version,
+        &environment.max_version,
+    ) && is_empty_or_contains(&environment.channels, &user_environment.update_channel)
+        && is_empty_or_contains(&environment.applications, &user_environment.app_name)
+        && is_empty_or_contains(&environment.device_type, &user_environment.device_type)
 }
 
 /// Determines whether the region and locale constraints in the supplied
@@ -61,7 +72,6 @@ fn matches_region_and_locale(
 
     if does_array_include(&environment.regions, user_region)
         && does_array_include(&environment.locales, user_locale)
-    // && environment.all_regions_and_locales
     {
         return true;
     }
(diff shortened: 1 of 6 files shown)


components/search/src/environment_matching.rs
```diff
@@ -81,16 +92,44 @@ fn matches_region_and_locale(
...
+    true
 }
 
-fn matches_application(
-    user_application_name: &SearchApplicationName,
-    environment_applications: &[SearchApplicationName],
-) -> bool {
-    environment_applications.is_empty() || environment_applications.contains(user_application_name)
+fn is_empty_or_contains<T: std::cmp::PartialEq>(env_value: &[T], user_value: &T) -> bool {
+    env_value.is_empty() || env_value.contains(user_value)
```
linabutler, line 132: Nice use of the `PartialEq` trait bound here!

If you want to get a little fancier—or if you find yourself wanting more generic slice methods like this in the future—you can make an [extension trait](http://xion.io/post/code/rust-extension-traits.html), and have all slices implement it:

```rust
use std::cmp::PartialEq;

trait SliceExt<T: ?Sized> {
    fn is_empty_or_contains(&self, value: &T) -> bool where T: PartialEq;
}

impl<T> SliceExt<T> for [T] {
    fn is_empty_or_contains(&self, value: &T) -> bool where T: PartialEq {
        self.is_empty() || self.contains(value)
    }
}
```

Then, you could use it like this:

```rust
environment.channels.is_empty_or_contains(&user_environment.update_channel)
    && environment.applications.is_empty_or_contains(&user_environment.app_name)
    && environment.device_type.is_empty_or_contains(&user_environment.device_type)
```

And, if you wanted to use your extension methods in multiple modules, you could put the trait inside a separate module—let's say `ext.rs`—and import it into your other modules like any other item:

```rust
use crate::ext::SliceExt;
```

components/search/src/environment_matching.rs
```diff
@@ -117,18 +156,12 @@ mod tests {
                     excluded_regions: vec![],
                     locales: vec![],
                     regions: vec![],
-                    distributions: vec![],
-                    excluded_distributions: vec![],
-                    applications: vec![],
+                    ..Default::default()
```
linabutler, line 159: Great!

linabutler (approved):
Looks great to me; thank you!

components/search/src/environment_matching.rs
```diff
@@ -81,16 +92,44 @@ fn matches_region_and_locale(
     false
 }
 
-fn matches_distribution(user_distribution_id: &str, environment_distributions: &[String]) -> bool {
-    environment_distributions.is_empty()
-        || environment_distributions.contains(&user_distribution_id.to_string())
+fn matches_version(
```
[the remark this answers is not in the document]
linabutler: I keep forgetting about `.map_or()` on [`Option`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or) and [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_or)—updated to use that!

components/search/src/environment_matching.rs
```diff
@@ -788,26 +635,15 @@ mod tests {
...
-                    region: "fr".into(),
-                    update_channel: SearchUpdateChannel::Default,
-                    distribution_id: "distro-4".into(),
-                    experiment: String::new(),
                     app_name: SearchApplicationName::FirefoxIos,
-                    version: String::new(),
+                    ..Default::default()
                 }
             ),
             "Should return false when the distribution does not match any in the environment"
```
mandysGit, line 649: If I'm reading this correctly, this one seems to be testing applications and not distributions.

components/search/src/environment_matching.rs
```diff
@@ -816,29 +652,359 @@ mod tests {
...
+                    ..Default::default()
+                },
+                &SearchUserEnvironment {
+                    locale: "fi".into(),
+                    region: "fr".into(),
+                    app_name: SearchApplicationName::Firefox,
+                    ..Default::default()
+                }
+            ),
+            "Should return true when the distribution and region matches the environment"
```
mandysGit, line 669: This one also is not testing distribution. 

Suggesting to change it to:
"Should return true when the application and region matches the environment"

components/search/src/environment_matching.rs
```diff
@@ -816,29 +652,359 @@ mod tests {
...
+                    ..Default::default()
+                }
+            ),
+            "Should return true when the experiment matches the one in the environment"
+        );
+
+        assert!(
+            !matches_user_environment(
+                &crate::JSONVariantEnvironment {
+                    experiment: "warp-drive".to_string(),
```
mandysGit, line 760: TIL - Warp drive was a technology that allowed space travel at faster-than-light speeds. 😄

mandysGit (changes requested):
Thanks for adding these in. It looks great! 
I really love the use of `..Default::default()`. It reads so nicely! 🤩 

I left comments for two of the tests that may need to be changed. The assertion doesn't seem to match the test statement.

mandysGit (approved):
Thank you.

PR #6517, opened 2024-12-05
Remote Settings: store full changesets instead of just records

leplatrem:
In order to prepare the work of signature verification, where we want to be able to assess data integrity/authenticity, we need to store collection metadata and timestamps.

Also, this PR fixes an issue where the timestamp was wrong in certain cases (empty collections, tombstones, etc.)

It is an internal change that does not require a CHANGELOG entry.

* [x] Refactoring
* [x] Storing
* [x] Retrieving 
* [x] Storage DB migration

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -151,9 +151,15 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
                 .get_last_modified_timestamp(&collection_url)?
                 .unwrap_or(0);
             if packaged_data.timestamp > cached_timestamp {
-                inner
-                    .storage
-                    .set_records(&collection_url, &packaged_data.data)?;
+                // Remove previously cached data (packaged data does not have tombstones like diff responses do).
+                inner.storage.empty()?;
+                // Insert new packaged data.
+                inner.storage.insert_collection_content(
+                    &collection_url,
+                    &packaged_data.data,
+                    packaged_data.timestamp,
+                    CollectionMetadata::default(),
+                )?;
                 return Ok(Some(self.filter_records(packaged_data.data)));
             }
         }
@@ -168,9 +174,14 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
             (Some(cached_records), _) => Some(self.filter_records(cached_records)),
             // Case 3: sync_if_empty=true
             (None, true) => {
-                let records = inner.api_client.get_records(None)?;
-                inner.storage.set_records(&collection_url, &records)?;
-                Some(self.filter_records(records))
+                let changeset = inner.api_client.fetch_changeset(None)?;
+                inner.storage.insert_collection_content(
+                    &collection_url,
+                    &changeset.changes,
+                    changeset.timestamp,
+                    changeset.metadata,
+                )?;
+                Some(self.filter_records(changeset.changes))
             }
             // Case 4: Nothing to return
             (None, false) => None,
@@ -181,8 +192,13 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         let mut inner = self.inner.lock();
         let collection_url = inner.api_client.collection_url();
         let mtime = inner.storage.get_last_modified_timestamp(&collection_url)?;
-        let records = inner.api_client.get_records(mtime)?;
-        inner.storage.merge_records(&collection_url, &records)
+        let changeset = inner.api_client.fetch_changeset(mtime)?;
+        inner.storage.insert_collection_content(
+            &collection_url,
+            &changeset.changes,
+            changeset.timestamp,
+            changeset.metadata,
+        )
     }
 
     /// Downloads an attachment from [attachment_location]. NOTE: there are no guarantees about a
@@ -221,7 +237,7 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         }
 
         // Try to download the attachment because neither the storage nor the local data had it
-        let attachment = inner.api_client.get_attachment(&metadata.location)?;
+        let attachment = inner.api_client.fetch_attachment(&metadata.location)?;
 
         // Verify downloaded data
         if attachment.len() as u64 != metadata.size {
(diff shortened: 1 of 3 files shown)


components/remote_settings/src/storage.rs
```diff
@@ -51,6 +51,7 @@ impl Storage {
         CREATE TABLE IF NOT EXISTS collection_metadata (
             collection_url TEXT PRIMARY KEY,
             last_modified INTEGER,
+            metadata BLOB NOT NULL,
```
bendk, line 54: I'd rather storing each field metadata separately rather than serializing it into a single field.  I don't feel very strongly about this one, though, feel free to push back on it.

components/remote_settings/src/storage.rs
```diff
@@ -140,38 +141,31 @@ impl Storage {
...
 
-    /// Set the list of records stored in the database, clearing out any previously stored records
-    pub fn set_records(
+    // TODO
+    pub fn set_collection_content(
         &mut self,
         collection_url: &str,
         records: &Vec<RemoteSettingsRecord>,
+        timestamp: u64,
+        metadata: CollectionMetadata,
```
bendk, line 150: As you pointed out, we're not doing the right thing when syncing with both new and old records.  Should we handle that here?  Based on the other changes, I think `records` could be renamed `changes` and we could implement the merging here while deleting/inserting things into the DB.

bendk (commented):
Looks like the right direction to me, thanks for working on this one.

components/remote_settings/src/storage.rs
leplatrem: I also hesitated there. 

The reason behind my choice is that eventually we will potentially look at other metadata fields, like `attachment` to know whether there's a bundle to pull, or `flags` to know whether a collection is required to have records during startup. 
But, we can perfectly decide to put this aside, and add the additional columns only when we implement these features.

It depends a bit how confident we are in database migration for the future

components/remote_settings/src/storage.rs
leplatrem: Let's work on this in a separate PR, and I can rebase this work on top 👌

leplatrem: See #6526

components/remote_settings/src/storage.rs
leplatrem: That would definitely help yes please 🙏  Maybe a candidate would be the `fetched` column

components/remote_settings/src/client.rs
```diff
@@ -151,9 +151,15 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
                 .get_last_modified_timestamp(&collection_url)?
                 .unwrap_or(0);
             if packaged_data.timestamp > cached_timestamp {
-                inner
-                    .storage
-                    .set_records(&collection_url, &packaged_data.data)?;
+                // Remove previously cached data (packaged data does not have tombstones).
```
gruberb, line 154: Can we explain here what a tombstone is?

components/remote_settings/src/storage.rs
```diff
@@ -50,7 +53,10 @@ impl Storage {
         );
         CREATE TABLE IF NOT EXISTS collection_metadata (
             collection_url TEXT PRIMARY KEY,
+            bucket TEXT,
             last_modified INTEGER,
+            signature TEXT,
+            x5u TEXT,
```
gruberb, line 59: Can we add a comment somewhere what x5u is?

gruberb (approved):
Just some comments, looks good!

components/remote_settings/src/storage.rs
```diff
@@ -140,30 +176,13 @@ impl Storage {
...
-    ///
-    /// Records with `deleted=false` will be inserted into the DB, replacing any previously stored
-    /// records with the same ID. Records with `deleted=true` will be removed.
-    pub fn merge_records(
+    /// Set cached content for this collection.
+    pub fn insert_collection_content(
         &mut self,
         collection_url: &str,
         records: &[RemoteSettingsRecord],
+        timestamp: u64,
```
bendk, line 184: Nit: how about `last_modified` or `last_modified_timestamp`?  `timestamp` could also mean the timestamp when we fetched the data.

bendk (changes requested):
This looks great to me.  The one remaining issue is the database migration.  I'll try to put up a PR one soon so that you can use it to base your work off.

components/remote_settings/src/storage.rs
```diff
@@ -776,6 +851,42 @@ mod tests {
         // Verify last modified timestamp
         let last_modified = storage.get_last_modified_timestamp(collection_url)?;
         assert_eq!(last_modified, Some(1300));
+        Ok(())
+    }
+    #[test]
+    fn test_storage_get_collection_content() -> Result<()> {
```
leplatrem, line 857:
```suggestion
    fn test_storage_get_collection_metadata() -> Result<()> {
```

components/remote_settings/src/storage.rs
```diff
@@ -105,6 +111,36 @@ impl Storage {
         result
     }
 
+    /// Get cached content for this collection
```
leplatrem, line 114:
```suggestion
    /// Get cached metadata for this collection
```

components/remote_settings/src/storage.rs
bendk: I got a bit distracted last week, but it's finally up: https://github.com/mozilla/application-services/pull/6537.

components/remote_settings/src/schema.rs
```diff
@@ -27,7 +27,7 @@ CREATE TABLE IF NOT EXISTS attachments (
     data BLOB NOT NULL);
 CREATE TABLE IF NOT EXISTS collection_metadata (
     collection_url TEXT PRIMARY KEY,
-    last_modified INTEGER);
+    last_modified INTEGER, bucket TEXT, signature TEXT, x5u TEXT);
```
bendk, line 30: Sorry you needed to format it like this to make the migration test work.

bendk (approved):
This looks great to me, are we ready to merge?

leplatrem:
> This looks great to me, are we ready to merge?

Yes! I'll squash the commits and merge 👌

PR #6518, opened 2024-12-07
Bug 1935797 - Replace CHECK and parent FOREIGN KEY constraints on `moz_bookmarks` with triggers.

linabutler:

  - Note:

### DIFF
--- a/components/places/sql/create_shared_triggers.sql
+++ b/components/places/sql/create_shared_triggers.sql
@@ -13,6 +13,60 @@ BEGIN
     DELETE FROM moz_places_tombstones WHERE guid = NEW.guid;
 END;
 
+-- Validate new rows. These were CHECK and FOREIGN KEY constraints
+-- in schemas <= 17; a trigger lets us provide custom error messages.
+CREATE TEMP TRIGGER moz_bookmarks_beforeinsert_trigger
+BEFORE INSERT ON moz_bookmarks
+BEGIN
+    -- SQLite < 3.47.0 only supports string literals for error messages,
+    -- so we use our own `throw(...)` function.
+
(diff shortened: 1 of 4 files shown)


linabutler:
@mhammond Welcome to my latest SQL shenanigans! I'm hoping that landing this will help us learn:

1. Which connection is causing the problem? If merging has gone wrong, we'd expect to see all constraint errors on the Sync connection—is that actually true, or do we see data integrity issues on the main connection, too?
2. What's the type of the bookmark that's causing it to complain?
3. Are we failing to insert a (potentially malformed) new bookmark, or trying to update an existing bookmark with bad data?
4. Is Sync trying to change the type of an existing bookmark item? This is a new constraint that I added on a hunch.

components/places/sql/create_shared_triggers.sql
```diff
@@ -260,6 +314,14 @@ BEGIN
         AND last_visit_date_remote = 0;
 END;
 
+-- Equivalent to a CASCADE foreign key action on `moz_bookmarks.parent`.
+CREATE TEMP TRIGGER moz_bookmarks_parent_afterdelete_trigger
```
mhammond, line 318: I don't understand why this is preferred - maybe a short comment to elaborate?

mhammond (approved):
this seems like a great idea! We should probably document this somewhere, but there's no need to block on those docs.

components/places/sql/create_shared_triggers.sql
linabutler: Will do! (It's because the `insert: item without parent` check already checks if `NEW.parent` exists in `moz_bookmarks`, so a `FOREIGN KEY(parent)` definition will do extra work to validate that constraint again).

PR #6522, opened 2024-12-11
[DISCO-3065] relevancy: caching for multi armed bandit API

misaniwere:
Updating select method to pull beta distribution values from a hashmap.
Testing is covered in the thompson sampling convergence test.

  - Note:

### DIFF
--- a/components/relevancy/src/lib.rs
+++ b/components/relevancy/src/lib.rs
@@ -23,15 +23,19 @@ use rand_distr::{Beta, Distribution};
 pub use db::RelevancyDb;
 pub use error::{ApiResult, Error, RelevancyApiError, Result};
 pub use interest::{Interest, InterestVector};
+use parking_lot::Mutex;
 pub use ranker::score;
 
 use error_support::handle_error;
+
+use std::collections::HashMap;
 
 uniffi::setup_scaffolding!();
 
 #[derive(uniffi::Object)]
 pub struct RelevancyStore {
     db: RelevancyDb,
+    cache: Mutex<BanditCache>,
 }
 
 /// Top-level API for the Relevancy component
@@ -45,6 +49,7 @@ impl RelevancyStore {
     pub fn new(db_path: String) -> Self {
         Self {
             db: RelevancyDb::new(db_path),
+            cache: Mutex::new(BanditCache::new()),
         }
     }
 
@@ -125,15 +130,12 @@ impl RelevancyStore {
     /// of success. The arm with the highest sampled probability is selected and returned.
     #[handle_error(Error)]
     pub fn bandit_select(&self, bandit: String, arms: &[String]) -> ApiResult<String> {
-        // we should cache the distribution so we don't retrieve each time
-
+        let mut cache = self.cache.lock();
         let mut best_sample = f64::MIN;
         let mut selected_arm = String::new();
 
         for arm in arms {
-            let (alpha, beta) = self
-                .db
-                .read(|dao| dao.retrieve_bandit_arm_beta_distribution(&bandit, arm))?;
+            let (alpha, beta) = cache.get_beta_distribution(&bandit, arm, &self.db)?;
             // this creates a Beta distribution for an alpha & beta pair
             let beta_dist = Beta::new(alpha as f64, beta as f64)
                 .expect("computing betas dist unexpectedly failed");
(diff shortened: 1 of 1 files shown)


components/relevancy/src/lib.rs
```diff
@@ -125,15 +130,12 @@ impl RelevancyStore {
     /// of success. The arm with the highest sampled probability is selected and returned.
     #[handle_error(Error)]
     pub fn bandit_select(&self, bandit: String, arms: &[String]) -> ApiResult<String> {
-        // we should cache the distribution so we don't retrieve each time
-
+        let mut cache = self.cache.lock();
```
gruberb, line 133: Can we move the `lock` inside the `for arm in arms`? This way we hold the lock of the Mutex not for the whole function. Maybe I am missing something, but it looks like we can request the `lock` in the `for` loop.

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
     /// its likelihood of a negative outcome.
     #[handle_error(Error)]
     pub fn bandit_update(&self, bandit: String, arm: String, selected: bool) -> ApiResult<()> {
+        let mut cache = self.cache.lock();
+
+        cache
+            .clear(&bandit, &arm)
+            .expect("failed to clear cached distribution");
```
gruberb, line 168: I commented below. `cache.clear()` doesn't fail, so no reason to return a `Result` and use `expect`. Also, expect means the application panics. We should return a `Error::CacheError` and let the calling application figure out how to handle an error case (logging or something else).

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
...
+
```
bendk, line 169: I was thinking about this case yesterday and I think we can do better here.  If the bandit/arm is cached, then rather than clear the entry we can update the entry based on the `selected` value.  I'm not sure if this is worth the complexity or not though.

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
...
         self.db
             .read_write(|dao| dao.update_bandit_arm_data(&bandit, &arm, selected))?;
+
+        Ok(())
+    }
+}
+
+pub struct BanditCache {
```
gruberb, line 177: Let's do this instead if `impl Default` ourselves:

`#[derive(Default)]
pub struct BanditCache`

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
...
+    }
+}
+
+impl BanditCache {
+    /// Creates a new, empty `BanditCache`.
+    ///
+    /// The cache is initialized as an empty `HashMap` and is used to store
+    /// precomputed Beta distribution parameters for faster access during
+    /// Thompson Sampling operations.
+    pub fn new() -> Self {
```
gruberb, line 193: We can return `BanditCache::default()` now

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
...
+        self.cache.insert(key, params);
+
+        Ok(params)
+    }
+
+    /// Clears the cached Beta distribution parameters for a given bandit and arm.
+    ///
+    /// This removes the cached values for the specified `bandit` and `arm` from the cache.
+    /// Use this method if the cached parameters are no longer valid or need to be refreshed.
+    pub fn clear(&mut self, bandit: &str, arm: &str) -> Result<()> {
```
gruberb, line 229: This function never fails, so we can remove the `Result` here and return nothing.

gruberb (changes requested):
Looks great! Just a few things.

components/relevancy/src/lib.rs
```diff
@@ -159,8 +161,76 @@ impl RelevancyStore {
...
+    /// If the parameters for the specified `bandit` and `arm` are already cached,
+    /// they are returned directly. Otherwise, the parameters are fetched from
+    /// the database, added to the cache, and then returned.
+    pub fn get_beta_distribution(
+        &mut self,
+        bandit: &str,
+        arm: &str,
+        db: &RelevancyDb,
+    ) -> Result<(usize, usize)> {
+        let key = (bandit.to_string(), arm.to_string());
```
bendk, line 210: We talked a bunch about how to best handle these cache keys and this seems like the best solution for now.  I would really love to see these `to_string` calls eliminated, but I can't think of a good way to do it without introducing a lot of extra complexity.

bendk (approved):
Looking good.  I had one suggestion that should result in more cache hits, but I'll let you make the call on if it's worth the complexity or not.

components/relevancy/src/lib.rs
bendk: Is this an improvement though?  There's overhead to locking the mutex multiple times and I don't think we're really giving other threads an opportunity to take the lock themselves.  Almost all of the work will be in `cache.get_beta_distribution` where we need to hold the lock.

gruberb: Yeah, that's true!

components/relevancy/src/lib.rs
misaniwere: yeah this makes sense. if we're updating the cache with the new values, there's no need for the clear method then. If we update after writing to the DB, is that enough of a fail safe? What happens if the DB write fails?

bendk: I think it's safe as long as the cache is updated after the DB operation completes.  If it works, then I think you can safely update the cache.  If not, then the transaction should be rolled back and it's safe to keep the older cached value.

misaniwere: after some thought, I think clearing it and forcing a re-fetch is the best way to handle this. If not we'd have to make another call to retrieve the current values from the cache before updating, or make changes to the DB method to return the new values. It becomes a little complex to keep the cache up to date & consistent in cases of failures in the DB operation. We're plan on updating this logic to allow for bulk updates so maybe we leave it for now and revisit?

gruberb (dismissed):

bendk (approved):
That logic about invalidating the cache makes sense to me, let's merge this.

gruberb (approved):

bendk (approved):

PR #6524, opened 2024-12-11
Upgrading AGP to 8.7.3 [ci full]

bendk:
The AGP upgrade part is very boring, the 2 commits before that are the ones where the real changes are and the ones I'd love some review on.

Tested by:
 - Running `./gradlew test` from the terminal
 - Running tests from android studio (I'd love someone to double-check this one)
 - Running a local fenix build and installing on an emulator (I'd also love a double check here, although I don't know if we have a great way now that fenix is in m-c and our branch build scripts don't work anymore).

  - Note:

### DIFF
--- a/build-scripts/component-common.gradle
+++ b/build-scripts/component-common.gradle
@@ -50,3 +50,115 @@ dependencies {
     androidTestImplementation libs.test.espresso.core
     androidTestImplementation libs.test.runner
 }
+
+// Shared logic for projects that depend on libmegazord
+//
+// This ensures that libmegazord will be in the library path so that it can be loaded.  It also adds
+// the transitive JNA dependency.
+ext.dependsOnTheMegazord = {
+    dependencies {
+        api project(":full-megazord")
+
+        // Add a JNA dependency, which is required by UniFFI.
+        //
+        // The tricky thing is getting the correct libjnidispatch library. Use the AAR version for
+        // normal builds, since that ensures the various Android libraries get packaged correctly when
+        // we build our own AAR.  Use the JAR version for tests, since that ensures the tests can load
+        // the native version.
+        implementation(libs.jna) {
+            artifact {
+                type = "aar"
+            }
+        }
+        testImplementation(libs.jna) {
+            artifact {
+                type = "jar"
+            }
+        }
+    }
+
+    // Configurations are a somewhat mysterious Gradle concept.  For our purposes, we can treat them
+    // sets of files produced by one component and consumed by another.
+    configurations {
+        megazordNative {
+            canBeConsumed = false
+        }
+    }
+
+    // Wire up the megazordNative configuration to the output produced from the `full-megazord` project.
+    dependencies {
+        megazordNative(project("path": ":full-megazord", "configuration": "megazordNative"))
+    }
+
+    afterEvaluate {
+        android.libraryVariants.all { variant ->
+            def variantName = variant.name.capitalize();
+            def testTask = tasks["test${variantName}UnitTest"]
+            def processTestResTask = tasks["process${variantName}UnitTestJavaRes"]
+
+            // Copy the native libmegazord to the resource dir so that it can be loaded by JNA.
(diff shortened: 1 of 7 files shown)


build-scripts/component-common.gradle
```diff
@@ -50,3 +50,115 @@ dependencies {
...
+        implementation(libs.jna) {
+            artifact {
+                type = "aar"
+            }
+        }
+        testImplementation(libs.jna) {
+            artifact {
+                type = "jar"
+            }
+        }
```
bendk, line 77: I'm pretty sure the fact this works, means we don't need the `jnaForTest` configuration.  Tell me if I'm missing anything.

build-scripts/component-common.gradle
```diff
@@ -50,3 +50,115 @@ dependencies {
...
+            //
+            // * https://github.com/mozilla/application-services/pull/6476#issuecomment-2537227576
+            // * https://github.com/mozilla/glean/pull/2680#issuecomment-2056627683
+            def copyNativeMegazord = tasks.register("copy${variantName}NativeMegazord", Copy) {
+                from configurations.getByName("megazordNative")
+                into processTestResTask.destinationDir
+                // Make sure to run after the process java res task, otherwise that one may
+                // overwrite our work.
+                dependsOn(processTestResTask)
+            }
```
bendk, line 112: I think this task is the key to getting the AGP upgrade to work.  Something similar should also work for Glean @badboy

tools/nimbus-gradle-plugin/build.gradle
```diff
@@ -7,7 +7,6 @@ plugins {
     id 'groovy'
     id 'maven-publish'
     id 'groovy-gradle-plugin'
-    alias libs.plugins.kotlin.dsl
```
bendk, line 10: Seems like this deletion doesn't hurt anything, but I don't have much context.  Tell me if it's needed.

rvandermeulen (approved):
Not sure I have much to add to this review, but this is freaking awesome. Thanks so much for finding a path forward!

tools/nimbus-gradle-plugin/build.gradle
rvandermeulen: @jeddai Any thoughts on this? I'm inclined to agree that it can be removed if everything works without it, but you might have more context.

jeddai: I actually keep having to delete it locally because it won't resolve. Not sure when it got added but I don't see a problem with removing it

bendk:
@ncalexan can you give this code a quick check?  It seems like it's working, I just want to make sure I'm not doing anything too wrong in the new code.

build-scripts/component-common.gradle
```diff
@@ -50,3 +50,115 @@ dependencies {
...
+ext.configureUniFFIBindgen = { crateName ->
+    // This will store the uniffi-bindgen generated files for our component
+    def uniffiOutDir = layout.buildDirectory.dir("generated/uniffi/")
+
+    android {
+        sourceSets.main.kotlin.srcDirs += uniffiOutDir
+    }
+
+    // Call `uniffi-bindgen` to generate the Kotlin bindings
+    def generateUniffiBindings = tasks.register("generateUniffiBindings") {
```
ncalexan, line 131: Future: this would be better as an actual `Exec` task, since the `doFirst` will also break configuration caching, I think.

ncalexan (approved):
Nothing here seems worse than the pile of hacks that came before, so the proof is in the eating.  If it's good for you, it's good for me.

build-scripts/component-common.gradle
bendk: Yeah, that does feel like it's likely to be an issue later on.

Your idea could work.  Another idea I had was to actually build a `.jar` file with the native libmegazord and depend on that -- that's how we're loading `libjnidispatch` after all.  Maybe we can explore these ideas more in the future.

PR #6525, opened 2024-12-12
Use the Default trait and the pretty_assertions in search selector tests.

Standard8:
This should make it easier to see the important part of tests, and to have a better output when tests fail.

Test-only so no changelog needed.

  - Note:

### DIFF
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -194,9 +194,13 @@ mod tests {
     use super::*;
     use crate::*;
     use once_cell::sync::Lazy;
+    use pretty_assertions::assert_eq;
 
     #[test]
     fn test_from_configuration_details_fallsback_to_defaults() {
+        // This test doesn't use `..Default::default()` as we want to
+        // be explicit about `JSONEngineBase` and handling `None`
+        // options/default values.
         let result = SearchEngineDefinition::from_configuration_details(
             "test",
             JSONEngineBase {
@@ -338,14 +342,8 @@ mod tests {
     static ENGINES_LIST: Lazy<Vec<SearchEngineDefinition>> = Lazy::new(|| {
         vec![
             SearchEngineDefinition {
-                aliases: Vec::new(),
-                charset: String::from("UTF-8"),
-                classification: SearchEngineClassification::General,
                 identifier: "engine1".to_string(),
-                partner_code: String::new(),
                 name: "Test".to_string(),
-                order_hint: None,
-                telemetry_suffix: None,
                 urls: SearchEngineUrls {
                     search: SearchEngineUrl {
                         base: "https://example.com".to_string(),
@@ -353,19 +351,13 @@ mod tests {
                         params: Vec::new(),
                         search_term_param_name: None,
                     },
-                    suggestions: None,
-                    trending: None,
+                    ..Default::default()
                 },
+                ..Default::default()
             },
             SearchEngineDefinition {
-                aliases: Vec::new(),
-                charset: "UTF-8".to_string(),
-                classification: SearchEngineClassification::General,
                 identifier: "engine2".to_string(),
-                partner_code: String::new(),
                 name: "Test 2".to_string(),
-                order_hint: None,
-                telemetry_suffix: None,
                 urls: SearchEngineUrls {
                     search: SearchEngineUrl {
                         base: "https://example.com/2".to_string(),
@@ -373,19 +365,13 @@ mod tests {
                         params: Vec::new(),
                         search_term_param_name: None,
                     },
-                    suggestions: None,
-                    trending: None,
+                    ..Default::default()
                 },
+                ..Default::default()
             },
             SearchEngineDefinition {
-                aliases: Vec::new(),
-                charset: "UTF-8".to_string(),
-                classification: SearchEngineClassification::General,
                 identifier: "engine3".to_string(),
-                partner_code: String::new(),
                 name: "Test 3".to_string(),
-                order_hint: None,
-                telemetry_suffix: None,
                 urls: SearchEngineUrls {
                     search: SearchEngineUrl {
                         base: "https://example.com/3".to_string(),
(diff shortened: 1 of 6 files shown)


mhammond:
Just a drive-by, but I wonder if a builder pattern might make more sense for, eg, `SearchEngineDefinition` where it might be possible to provide a convenient pattern while avoiding "invalid" values - eg, `Default` lets you construct with empty names, identifiers and charsets, but I expect in practice these aren't actually allowed to be empty.

Of course, the existing code *already* allows that because all the elements are `pub`, so I agree this is a strict improvement over the status quo - but wondering if over the longer term this might bite?

(I don't have enough context to know if this makes sense - feel free to value it at what it cost you :)

components/search/src/filter.rs
```diff
@@ -338,84 +342,61 @@ mod tests {
...
+                    ..Default::default()
                 },
+                ..Default::default()
             },
             SearchEngineDefinition {
-                aliases: Vec::new(),
-                charset: "UTF-8".to_string(),
-                classification: SearchEngineClassification::General,
                 identifier: "engine2".to_string(),
                 partner_code: String::new(),
```
mandysGit, line 360: I noticed partnerCode isn't assigned a specific value. Could partner code be removed here?

components/search/src/selector.rs
```diff
@@ -478,14 +462,9 @@ mod tests {
         );
 
         let mut result = Arc::clone(&selector).filter_engine_configuration(SearchUserEnvironment {
-            locale: "fi".into(),
-            region: "FR".into(),
-            update_channel: SearchUpdateChannel::Default,
             distribution_id: String::new(),
```
mandysGit, line 465: Should `distribution_id` be removed? I noticed there no specific value set for it. 

It similar to `version: String::new()`  which has been removed.

mandysGit (approved):
Thanks for tidying this up. I left a few comments on some properties I think should be removed too.

mandysGit:
> Just a drive-by, but I wonder if a builder pattern might make more sense for, eg, `SearchEngineDefinition` where it might be possible to provide a convenient pattern while avoiding "invalid" values - eg, `Default` lets you construct with empty names, identifiers and charsets, but I expect in practice these aren't actually allowed to be empty.
> 
> Of course, the existing code _already_ allows that because all the elements are `pub`, so I agree this is a strict improvement over the status quo - but wondering if over the longer term this might bite?
> 
> (I don't have enough context to know if this makes sense - feel free to value it at what it cost you :)

Hi @mhammond - Thank you for your thoughts on this. You're right on that in practice we don't allow some of these properties to be empty. 

From my perspective, allowing empty values for SearchEngineDefinition gives us the flexibility we need for testing how the filtering logic behaves with specific properties. Making the required properties populated each time would make the code more robust, but it might add unnecessary complexity and reduce readability. With the builder pattern, I am assuming it will add more boilerplate to the the code by having the required properties each time. The way we've been testing it on Desktop, we aren't filling in the engine definitions with all the required properties, we only add the properties we need for testing and leaving the rest empty. 

As long as we have our schema tested regularly—[search-config-v2 schema](https://searchfox.org/mozilla-central/source/toolkit/components/search/schema/search-config-v2-schema.json)—and all properties meet their requirements, I believe that's where we should focus on validating the presence or absence of values.

That's my initial thoughts on it, but I'm not sure if @Standard8 might feel differently or have other suggestions I haven't thought about. Do you think we should be stricter here or is the schema validation enough?

Standard8:
> Just a drive-by, but I wonder if a builder pattern might make more sense for, eg, `SearchEngineDefinition` where it might be possible to provide a convenient pattern while avoiding "invalid" values - eg, `Default` lets you construct with empty names, identifiers and charsets, but I expect in practice these aren't actually allowed to be empty.

When you refer to builder pattern, I assume you referring to this [type of pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html)?

If so, then to check I understand correctly - the pattern allows to specify what the default values should be? (btw I think Rust might be getting better [default values specifications](https://rust-lang.github.io/rfcs/3681-default-field-values.html) at some stage).

> Of course, the existing code _already_ allows that because all the elements are `pub`, so I agree this is a strict improvement over the status quo - but wondering if over the longer term this might bite?

For items like `SearchEngineDefinition` I'm not too worried - these are effectively "out" parameters that we're returning to the application and are filled out from processing the search configuration, and it is the component's responsibility to set those correctly. We do that in a [central place](https://github.com/mozilla/application-services/blob/f1118093260fa9921d58e2193ab301724e3279c6/components/search/src/filter.rs#L35-L52).

The one I was a bit wary of, [is `SearchUserEnvironment`](https://github.com/mozilla/application-services/blob/f1118093260fa9921d58e2193ab301724e3279c6/components/search/src/types.rs#L57-L86) because that's passed into the component, and I'd kinda like to make sure the application specifies everything in that object.

I think you are saying that even without the `Default` the `pub` will mean that the values are optional for the application? So I think we wouldn't be able to ensure everything is passed?

Alternately, if there's a better way to do `Default` trait for debug-only, or something like that, then I think I might consider that.

mhammond:
Thanks for the extra context. Re the "out" params, that does make sense. Even for in params, obviously there's nothing wrong with checking the values explicitly too, but I do like exploring ways for the compiler to make things like that redundant. But even when you can it's sometimes not worth it :)

Eg, by not having attributes be `pub`, you are effectively *forcing* consumers to go through one of your functions or methods to create the object. However, the cost of this is that you are then forced to create `pub` methods for every simple access to these elements. The juice is often not worth the squeeze as in practice, you've done the same thing but with extra steps.

And to be clear, I wasn't suggesting you don't use `Default` in your current world - if you are going to require consumers to specify each field because they create the object directly, then making their lives easier is always worthwhile, so `Default` makes sense.

All this is moot when stuff is exposed via uniffi anyway - a struct that is exposed as a uniffi record *must* have public fields as it is the fields which are exposed to the ffi.

So yeah, it was a throw away comment that was made without any context of your component - happy to have spawned a discussion though!

Standard8:
> So yeah, it was a throw away comment that was made without any context of your component - happy to have spawned a discussion though!

Thank you for commenting. I'm still very new to this and potentially better ways of working are most welcome.

components/search/src/selector.rs
Standard8: In the tests in this section, we're using `distribution_id` and `app_name` for the environment filtering parameters. Hence I wanted to keep them both there explicitly so that it was clearer that the `distribution_id` wasn't being set and that environment section wouldn't apply.

PR #6526, opened 2024-12-12
Bug 1936053: RemoteSettingsClient.sync() should merge records

bendk:
Merge new records with the old ones, rather than unconditionally storing the new records.  This fixes buggy behavior when sync does an incremental update.

Also added a refactoring commit that I think cleans some stuff up.

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -144,7 +144,7 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         // Case 1: The packaged data is more recent than the cache
         //
         // This happens when there's no cached data or when we get new packaged data because of a
-        // Firefox update
+        // product update
         if let Some(packaged_data) = packaged_data {
             let cached_timestamp = inner
                 .storage
@@ -162,6 +162,9 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
 
         Ok(match (cached_records, sync_if_empty) {
             // Case 2: We have cached records
+            //
+            // Note: we should return these even if it's an empty list and `sync_if_empty=true`.
+            // The "if empty" part refers to the cache being empty, not the list.
             (Some(cached_records), _) => Some(self.filter_records(cached_records)),
             // Case 3: sync_if_empty=true
             (None, true) => {
@@ -179,7 +182,7 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         let collection_url = inner.api_client.collection_url();
         let mtime = inner.storage.get_last_modified_timestamp(&collection_url)?;
         let records = inner.api_client.get_records(mtime)?;
-        inner.storage.set_records(&collection_url, &records)
+        inner.storage.merge_records(&collection_url, &records)
     }
 
     /// Downloads an attachment from [attachment_location]. NOTE: there are no guarantees about a
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -134,7 +134,6 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
     pub fn get_records(&self, sync_if_empty: bool) -> Result<Option<Vec<RemoteSettingsRecord>>> {
         let mut inner = self.inner.lock();
         let collection_url = inner.api_client.collection_url();
-        let cached_records = inner.storage.get_records(&collection_url)?;
         let is_prod = inner.api_client.is_prod_server()?;
         let packaged_data = if is_prod {
             self.load_packaged_data()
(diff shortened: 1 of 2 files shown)


components/remote_settings/src/client.rs
```diff
@@ -134,66 +134,52 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
-        }
+        let cached_records = inner.storage.get_records(&collection_url)?;
 
-        // Case 4: Cache is empty and we're allowed to sync
-        let records = inner.api_client.get_records(None)?;
-        inner.storage.set_records(&collection_url, &records)?;
-        Ok(Some(self.filter_records(records)))
+        Ok(match (cached_records, sync_if_empty) {
+            // Case 2: We have cached records
+            (Some(cached_records), _) => Some(self.filter_records(cached_records)),
```
bendk, line 165: This is a small behavioral change in that we will return the cached records even if it's an empty list and sync_if_empty is true (i.e. we previously fetched the collection, but for whatever reason there were 0 records in it).  The way I interpret that flag is that we should sync if the cache is empty, we shouldn't check the record count.  @leplatrem can you confirm or deny if this is the correct behavior?

components/remote_settings/src/storage.rs
```diff
@@ -148,32 +148,83 @@ impl Storage {
...
+    }
+
+    /// Insert/remove/update rows in the records table based on a records list
+    ///
+    /// Returns the max last modified record from the list
+    fn update_record_rows(
+        tx: &Transaction<'_>,
+        collection_url: &str,
+        records: &[RemoteSettingsRecord],
+    ) -> Result<u64> {
```
bendk, line 195: It's slightly awkward that this has to be an associated function, rather than a `&self` method.  It currently has to be like this, since there's already a mutable borrow to get the transaction.  I think it would be easier to split up operations into multiple methods if we moved to the `Dao` pattern, which stores the transaction in an internal field.

components/remote_settings/src/client.rs
```diff
@@ -134,66 +134,52 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
+            // Case 3: sync_if_empty=true
+            (None, true) => {
+                let records = inner.api_client.get_records(None)?;
+                inner.storage.set_records(&collection_url, &records)?;
+                Some(self.filter_records(records))
+            }
+            // Case 4: Nothing to return
+            (None, false) => None,
+        })
```
bendk, line 174: I thought I was going to be touching this code more to implement the feature, but it turns out I didn't really need to.  I still think this refactoring is an improvement, but it's not really related to the main change.  If anyone prefers the previous version I can revert.

components/remote_settings/src/storage.rs
```diff
@@ -148,32 +148,83 @@ impl Storage {
...
+        tx: &Transaction<'_>,
+        collection_url: &str,
+        last_modified: u64,
+    ) -> Result<()> {
         // Update the metadata
         let fetched = true;
         tx.execute(
             "INSERT OR REPLACE INTO collection_metadata (collection_url, last_modified, fetched) VALUES (?, ?, ?)",
-            (collection_url, max_last_modified, fetched),
+            (collection_url, last_modified, fetched),
```
bendk, line 226: @gruberb it seems like `fetched` is set to true when the row is inserted, do you think we could remove it and instead just check the presence of the row?  It's not really a big deal either way, the main reason I ask is that I want to start off some migrations for this component.

leplatrem: I was also surprised with this `fetched` column. A count would do it

components/remote_settings/src/client.rs
```diff
@@ -134,66 +134,52 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
-
-        // Now we know we have cached records
-        let cached_records = cached_records.unwrap();
-        let cached_timestamp = inner.storage.get_last_modified_timestamp(&collection_url)?;
-
-        // Case 2: We have packaged data and are in prod
+        // Case 1: The packaged data is more recent than the cache
+        //
+        // This happens when there's no cached data or when we get new packaged data because of a
+        // Firefox update
```
leplatrem, line 147: *Product* update?

components/remote_settings/src/client.rs
leplatrem: This is way simpler, I prefer this 👌

leplatrem (approved):

leplatrem:
Now that I think of it more. If we want to implement signature verification, we'll have to do these steps:

- `.merge_records()`
- `.get_records()`
- verify signature
- clear storage if invalid

It's valid with this approach https://remote-settings.readthedocs.io/en/latest/client-specifications.html#signature-verification

If we would do the merge in memory:

- `.get_records()`
- merge in memory
- verify signature
- `set_records()` if signature is valid, else if previous content has valid signature leave it as is, otherwise clear storage
 
It's one method less in storage, and also, it gives us the ability to keep the previous valid dataset
Like we do on desktop https://firefox-source-docs.mozilla.org/services/settings/index.html#synchronization-process

components/remote_settings/src/client.rs
gruberb: Can we add a comment stating this? I am still confused by what we do when...

gruberb (approved):

PR #6530, opened 2024-12-18
Add appServicesRootDir attribute gradle should generally use instead …

mhammond:
…of rootDir.

The intent is that this will be the same as `rootDir` when in app-services, but magically turn into the root of the app-services tree once in mozilla-central.

The intent here is to avoid needing to touch these files after "vendoring" into m-c. @rvandermeulen does this make sense to you?

### DIFF
--- a/components/places/android/build.gradle
+++ b/components/places/android/build.gradle
@@ -2,9 +2,9 @@ plugins {
     alias libs.plugins.gradle.python.envs
 }
 
-apply from: "$rootDir/build-scripts/component-common.gradle"
-apply from: "$rootDir/build-scripts/protobuf-common.gradle"
-apply from: "$rootDir/publish.gradle"
+apply from: "$appServicesRootDir/build-scripts/component-common.gradle"
+apply from: "$appServicesRootDir/build-scripts/protobuf-common.gradle"
+apply from: "$appServicesRootDir/publish.gradle"
 
 ext.gleanNamespace = "mozilla.telemetry.glean"
 apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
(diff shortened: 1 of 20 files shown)


settings.gradle
```diff
@@ -3,7 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import org.yaml.snakeyaml.Yaml
 
-includeBuild('tools/nimbus-gradle-plugin') {
+// We prefer `appServicesRootDir` over `rootDir` to help us on the path to the monorepo.
+// (They are the same in app-services but different in moz-central)
+def maybeAppServicesRootDir = new File(rootDir, "services/app-services")
```
mhammond, line 8: Some of this probably highlights my poor groovy/gradle skills, but I couldn't work out a sane way to share this variable both out and inside `setupProject`, so this is what I came up with.

settings.gradle
```diff
@@ -66,14 +73,20 @@ if (file('local.properties').canRead()) {
 
 def calcVersion(buildconfig) {
     def local = gradle.rootProject.findProperty("local")
-    def version = new File(rootDir, 'version.txt').getText().trim()
 
-    if (gradle.rootProject.hasProperty("nightlyVersion")) {
+    if (gradle.hasProperty("mozconfig")) {
```
mhammond, line 77: Get something sane from a build in m-c - not clear how (or even if) we need a dedicated version string once landed.

components/places/android/build.gradle
```diff
@@ -2,9 +2,8 @@ plugins {
     alias libs.plugins.gradle.python.envs
 }
 
-apply from: "$rootDir/build-scripts/component-common.gradle"
-apply from: "$rootDir/build-scripts/protobuf-common.gradle"
```
mhammond, line 6: places no longer needs protobuf.

PR #6534, opened 2024-12-19
Remote settings: add signature verification support

leplatrem:
* [x] Basic signature verification feature
* [x] Investigate why hard-coded root hash is necessary with rc_crypto https://bugzilla.mozilla.org/show_bug.cgi?id=1940903
* [ ] Investigate whether x5u certificates are cached
* [x] Test plan for project features combinations testing (eg. jexl alone, signature alone, both together) --> both together
* [x] Add a command to the remote settings example that will verify collection signatures
* [x] Create follow-up issue to add a method `.verify_and_get()` which will verify signature before returning cache content --> https://bugzilla.mozilla.org/show_bug.cgi?id=1940908

  - Note:

### DIFF
--- a/components/remote_settings/Cargo.toml
+++ b/components/remote_settings/Cargo.toml
@@ -11,6 +11,7 @@ exclude = ["/android", "/ios"]
 [features]
 default = []
 jexl = ["dep:jexl-eval"]
+signatures = ["dep:canonical_json", "dep:rc_crypto"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
@@ -32,6 +33,8 @@ regex = "1.9"
 anyhow = "1.0"
 firefox-versioning = { path = "../support/firefox-versioning" }
 sha2 = "^0.10"
+canonical_json = { version = "0.5", optional = true }
+rc_crypto = { path = "../support/rc_crypto", optional = true }
 
 [build-dependencies]
 uniffi = { version = "0.28.2", features = ["build"] }
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -6,6 +6,8 @@ use crate::config::RemoteSettingsConfig;
 use crate::error::{Error, Result};
 #[cfg(feature = "jexl")]
 use crate::jexl_filter::JexlFilter;
+#[cfg(feature = "signatures")]
+use crate::signatures;
 use crate::storage::Storage;
 #[cfg(feature = "jexl")]
 use crate::RemoteSettingsContext;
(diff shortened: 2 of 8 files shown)


components/remote_settings/src/client.rs
```diff
@@ -189,16 +214,126 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
-            &changeset.changes,
-            changeset.timestamp,
-            changeset.metadata,
-        )
+        inner.storage.empty()?;
+        let is_prod = inner.api_client.is_prod_server()?;
+        let packaged_data = if is_prod {
+            self.load_packaged_data()
+        } else {
+            None
```
gruberb, line 267: AFAIK, we can return `Ok(())` here right away? And do a bit of cleanup:

```rust
fn reset_storage(&self) -> Result<()> {
    let mut inner = self.inner.lock();
    let collection_url = inner.api_client.collection_url();
    
    // Clear existing storage
    inner.storage.empty()?;

    // Load packaged data only for production
    if inner.api_client.is_prod_server()? {
        if let Some(packaged_data) = self.load_packaged_data() {
            inner.storage.insert_collection_content(
                &collection_url,
                &packaged_data.data,
                packaged_data.timestamp,
                CollectionMetadata::default(),
            )?;
        }
    }

    Ok(())
}
```

components/remote_settings/src/client.rs
```diff
@@ -189,16 +214,126 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
     }
 
     pub fn sync(&self) -> Result<()> {
```
gruberb, line 216: How about moving the duplicated code out into its own function, and clean it up like this (haven't test run this - just a rough idea).

```rust
fn perform_sync_operation(inner: &Mutex<Inner>) -> Result<()> {
    let mut inner = inner.lock();
    let collection_url = inner.api_client.collection_url();
    let timestamp = inner.storage.get_last_modified_timestamp(&collection_url)?;
    let changeset = inner.api_client.fetch_changeset(timestamp)?;
    
    inner.storage.insert_collection_content(
        &collection_url,
        &changeset.changes,
        changeset.timestamp,
        changeset.metadata,
    )
}

pub fn sync(&self) -> Result<()> {
    // First attempt
    perform_sync_operation(&self.inner)?;
    
    if let Err(err) = self.verify_signature() {
        // Retry with packaged dataset
        self.reset_storage()?;
        perform_sync_operation(&self.inner)?;
        
        // Final verification
        self.verify_signature().map_err(|verification_err| {
            // Reset and return the error if second attempt fails
            self.reset_storage().expect("Failed to reset storage after verification failure");
            verification_err
        })?;
    }

    Ok(())
}
```

gruberb (approved):
LGTM! A few cleanup comments. 

I tried to spot if we can have a `Signature` struct, and `impl` the functions inside `signatures.rs`, but maybe that's not  needed? I would then maybe change the module/file name to `signature_utils.rs`. But just a suggestion.

gruberb (approved):

components/remote_settings/Cargo.toml
```diff
@@ -44,3 +47,4 @@ mockito = "0.31"
 # We add the perserve_order feature to guarantee ordering of the keys in our
 # JSON objects as they get serialized/deserialized.
 serde_json = { version = "1", features = ["preserve_order"] }
+mock_instant = "0.5.1"
```
bendk, line 50: I'm wondering if we can implement this with less dependencies.  When this gets merged to moz-central, all the new dependencies will require an audit, so we should only add them if they're bringing in a lot of functionality.  Is there another way to get a mock time for the unit tests?

OTOH, canonical_json is easy to audit since you're the one publishing it.

Side question: how does the epoch time get used in the signatures?  How far off would a user's clock have to be to get a false negative?

components/remote_settings/src/client.rs
```diff
@@ -188,11 +213,20 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
...
         let collection_url = inner.api_client.collection_url();
-        let mtime = inner.storage.get_last_modified_timestamp(&collection_url)?;
-        let changeset = inner.api_client.fetch_changeset(mtime)?;
+        let timestamp = inner.storage.get_last_modified_timestamp(&collection_url)?;
+        let changeset = inner.api_client.fetch_changeset(timestamp)?;
+        log::debug!(
+            "{0}: apply {1} change(s) locally.",
+            self.collection_name,
+            changeset.changes.len()
+        );
```
bendk, line 229: FWIW, there's also `error_reporting::breadcrumb!` which will log, but also set a breadcrumb on any sentry errors.  I don't know if this would be useful there, but I just wanted to point it out.

components/remote_settings/src/signatures.rs
```diff
@@ -0,0 +1,107 @@
...
+    if let Value::Object(map) = value {
+        let new_map = map
+            .iter()
+            .filter_map(|(key, v)| {
+                if key == "deleted" || key == "attachment" && v.is_null() {
+                    None
+                } else {
+                    Some((key.clone(), v.clone()))
+                }
+            })
```
bendk, line 18: Maybe splitting this out into a call to `filter()` and a call to `cloned()` would be simpler, but however you want to spell this out is fine with me.

bendk (approved):
Looks good to me.  I don't understand all the signature stuff, but I like how it's organized.

I left some suggestions that you can take or leave.  My only real concern is with the `mock_instant` dependency.  If that's helping a lot in the tests, let's keep it, but if not I'd rather remove it and not have to go through the auditing process.

components/remote_settings/Cargo.toml
leplatrem: > Is there another way to get a mock time for the unit tests?

I'll try using a static/lazy

> Side question: how does the epoch time get used in the signatures? How far off would a user's clock have to be to get a false negative?

The epoch is used to validate the certificate. It is valid for several weeks and we refresh the signatures regularly to pick the latest ones. Client clock would have to be days/weeks off for a false negative

components/remote_settings/src/signatures.rs
leplatrem: Could not use `clone()` because the compiler wouldn't let me clone the values of type `(&String, &Value)`. I left `.map(|(key, v)| (key.clone(), v.clone()))`

PR #6535, opened 2024-12-20
Revert "Expose webext storage"

lougeniaC64:
Reverts mozilla/application-services#6473


linabutler:
[The landing of bug 1888472 stuck](https://bugzilla.mozilla.org/show_bug.cgi?id=1888472#c20), so there's no need to revert this!

PR #6548, opened 2025-01-07
NSS keydb

jo:
Add NSS functionality for key management, enabled via feature flag.

New feature flag `keydb` in `rc_crypto/nss`, which enables NSS key persistence: `ensure_initialized_with_profile_dir(path: impl AsRef<Path>)` initializes NSS with a profile directory and appropriate flags to persist keys (and certificates) in its internal PKCS11 software implementation. 

New methods for dealing with primary password and key persistence, available within the `keydb` feature:
  * `authorization_with_primary_password_is_needed()`: checks weather a primary password is set and needs to be authorized
  * `authorize_with_primary_password(primary_password: &str)`: method for authorizing NSS key store against a user-provided primary password
  * `get_or_create_aes256_key(name: &str)`: retrieve a key by `name` from the internal NSS key store. If none exists, create one, persist, and return.

  - Note:

### DIFF
--- a/components/support/rc_crypto/nss/Cargo.toml
+++ b/components/support/rc_crypto/nss/Cargo.toml
@@ -15,7 +15,9 @@ error-support = { path = "../../error" }
 nss_sys = { path = "nss_sys" }
 serde = "1"
 serde_derive = "1"
+once_cell = { version = "1.20.2", optional = true }
 
 [features]
 default = []
 gecko = ["nss_sys/gecko"]
+keydb = ["dep:once_cell"]
--- a/components/support/rc_crypto/nss/src/pk11/sym_key.rs
+++ b/components/support/rc_crypto/nss/src/pk11/sym_key.rs
@@ -2,11 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[cfg(feature = "keydb")]
+use crate::util::get_last_error;
 use crate::{
     error::*,
     pk11::{context::HashAlgorithm, slot, types::SymKey},
     util::{ensure_nss_initialized, map_nss_secstatus, sec_item_as_slice, ScopedPtr},
 };
+#[cfg(feature = "keydb")]
+use std::ffi::{c_char, CString};
 use std::{
     mem,
     os::raw::{c_uchar, c_uint, c_ulong},
@@ -54,8 +58,8 @@ pub fn hkdf_expand(
         )?
     };
     map_nss_secstatus(|| unsafe { nss_sys::PK11_ExtractKeyValue(sym_key.as_mut_ptr()) })?;
-    // This doesn't leak, because the SECItem* returned by PK11_GetKeyData
-    // just refers to a buffer managed by `sym_key` which we copy into `out`.
+    // SAFETY: This doesn't leak, because the SECItem* returned by PK11_GetKeyData just refers to a
+    // buffer managed by `sym_key` which we copy into `out`.
     let mut key_data = unsafe { *nss_sys::PK11_GetKeyData(sym_key.as_mut_ptr()) };
     if u32::try_from(len)? > key_data.len {
         return Err(ErrorKind::InternalError.into());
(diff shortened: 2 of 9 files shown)


jo:
@mhammond , thanks for the review!

I think you have raised important points here, I will think again in detail about the error handling. I have mostly imitated the behavior of `ensure_nss_initialized()` here, but I absolutely agree, in some code paths we should not panic, because in particular file corruption can occur, or the directory may have disappeared.

@bendk , please wait a bit until I have incorporated Mark's feedback.

jo:
I have now revised the implementation. The new method `ensure_nss_initialized_with_profile_dir` now returns a result containing any errors. It also checks, as @bendk  mentioned recently, whether any previous calls have taken place with the same path.
Thanks again for your valuable feedback, @mhammond . Can you take another look at this and give us your opinion?

jo:
btw I added the [once_cell](https://github.com/mozilla/application-services/pull/6548/files#diff-7ba97d05aea837c9a2f3818ab511138b116296c82bba61b6b6c7394d8fa8fe44R18) dependency because we already use it heavily throughout AS. Do I have to take any additional steps here?

components/support/rc_crypto/nss/src/pk11/sym_key.rs
```diff
@@ -90,3 +94,180 @@ pub(crate) fn import_sym_key(
...
+                    nss_sys::CKM_AES_KEY_WRAP_KWP,
+                    ptr::null_mut(),
+                    wrapping_key.as_mut_ptr(),
+                    sym_key.as_mut_ptr(),
+                    &mut wrap_len,
+                )
+            })
+            .map_err(|_| get_last_error())?;
+            let wrapped_key = unsafe {
+                nss_sys::SECITEM_AllocItem(ptr::null_mut(), ptr::null_mut(), wrap_len.len + 8)
```
mhammond, line 191: the 8 here seems a bit magic - maybe a comment?

components/support/rc_crypto/nss/Cargo.toml
```diff
@@ -15,7 +15,9 @@ error-support = { path = "../../error" }
 nss_sys = { path = "nss_sys" }
 serde = "1"
 serde_derive = "1"
+once_cell = "1.20.2"
```
mhammond, line 18: I think our min versions are such that we can now use `std::cell::OnceCell` and don't actually need this? (But if we did, it looks like it should be an optional dep gated on your new feature)

mhammond (commented):
this looks great!

components/support/rc_crypto/nss/Cargo.toml
jo: I tried to use  `std::cell::OnceCell` but rustc yelled

```
23 | static NSS_PROFILE_PATH: OnceCell<String> = OnceCell::new();
   |                          ^^^^^^^^^^^^^^^^ `std::cell::OnceCell<std::string::String>` cannot be shared between threads safely
```

jo: changed the dependency to be optional, and controlled by the `keydb` feature

bendk (approved):
Looks great from my POV as well, I just had one documentation request.

components/support/rc_crypto/nss/src/pk11/sym_key.rs
jo: done

components/support/rc_crypto/nss/src/pk11/sym_key.rs
```diff
@@ -90,3 +92,182 @@ pub(crate) fn import_sym_key(
...
+        }
+    };
+
+    // This doesn't leak, because the SECItem* returned by PK11_GetKeyData
+    // just refers to a buffer managed by `sym_key` which we copy into `out`.
+    let mut key_data = unsafe { *nss_sys::PK11_GetKeyData(sym_key.as_mut_ptr()) };
+    if key_data.len != nss_sys::AES_256_KEY_LENGTH {
+        return Err(ErrorKind::InternalError.into());
+    }
+    let buf = unsafe { sec_item_as_slice(&mut key_data)? };
```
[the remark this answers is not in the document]
jo: Is that what you mean? I think we're missing a great many safety instructions, but I don't feel qualified to add them all.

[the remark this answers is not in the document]
bendk: This is very good, but IMO it could be made slightly better by framing things in terms of ownership and contracts.  Also, the standard formatting is slightly different for the unsafe function vs the usage of that function.

Maybe something like this?  

```
/// # Safety
///
/// The returned reference must not outlive the `sym_key`, since that owns the `SecItem` buffer.
pub(crate) unsafe fn sec_item_as_slice(sec_item: &mut SECItem) -> Result<&mut [u8]> {
...
}
```

```
// SAFETY: `to_vec` copies the data out before there's any chance for `sym_key` to be destroyed.
let buf = unsafe { sec_item_as_slice(&mut key_data)? };
```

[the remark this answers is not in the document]
bendk: I'm not sure if I got the specifics right, my main point was the general framing of the docstrings.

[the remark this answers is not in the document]
jo: ok thanks! Incorporated.

PR #6549, opened 2025-01-07
Async Rust ADR

bendk:
This is my new version of https://github.com/mozilla/application-services/pull/5910, which is more focused and more concrete.  Maybe we're finally ready to actually make a decision here :crossed_fingers: 

  - Note:

### DIFF
--- a/docs/adr/files/0009/option-b-android-components.diff
+++ b/docs/adr/files/0009/option-b-android-components.diff
@@ -1,0 +1,256 @@
+diff --git a/mobile/android/android-components/components/feature/fxsuggest/build.gradle b/mobile/android/android-components/components/feature/fxsuggest/build.gradle
+index d5b717c0ef..28d9c82eb2 100644
+--- a/mobile/android/android-components/components/feature/fxsuggest/build.gradle
++++ b/mobile/android/android-components/components/feature/fxsuggest/build.gradle
+@@ -42,6 +42,7 @@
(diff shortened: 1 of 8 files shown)


docs/adr/files/0009/option-b-android-components.diff
```diff
@@ -0,0 +1,256 @@
+diff --git a/mobile/android/android-components/components/feature/fxsuggest/build.gradle b/mobile/android/android-components/components/feature/fxsuggest/build.gradle
+index d5b717c0ef..28d9c82eb2 100644
```
bendk, line 2: I wasn't sure the best way to communicate these changes, but I went with checking in diffs because I wanted to make sure they never expired.  I can put these into a github PR or phab patch if people want, maybe those would look nicer.

bendk:
I updated this after our meeting yesterday:

- Added a list of issues this is trying to address to the problem statement section
- Changed option A from "Do nothing" to "Address as many issues as we can without moving to async Rust".

With this change, I could get behind option A.  I think we could solve the worst issue, which is how the functionality is split across different repos/languages and easy to forget about when you're writing Rust code.  The impact on the docs is unfortunate, but not really a deal-breaker.

That said, I still like option C the best. I like that it makes the code we write match more closely with how consumers are using it and it opens up new possibilities for the future.  I also don't see much downside/risk personally.

PR #6550, opened 2025-01-07
Update gradle files to work in both app-services and mozilla-central

mhammond:
Newer version of #6530

There are 3 commits here, mainly as some might be more controversial than others:

* first commit is use `appServicesRootDir` rather than `rootDir`; no real functionality changes just a way so gradle works roughly the same in both places.
* second commit moves app-services to use the same shaped [`Config` object as m-c](https://searchfox.org/mozilla-central/rev/f53c09a22edc700ab1a9eaaf4da0f0dd9f11bff3/mobile/android/shared-settings.gradle#39-63). When app-services is in m-c, we end up using the `Config` object from m-c - the app-services version of this object is just for use in app-services so the same gradle code works in both places. Given some comments in the source it appears we maybe used to do that?
* Last commit is where some behaviour changes are implemented depending on whether we are in m-c or app-services. In particular, the code checks `gradle.hasProperty("mozconfig")` if it needs to do something different based on the location. For now, that means things like disabling the cargo-gradle plugin and changing how uniffi-bindgen is run, etc. The implemented behaviour should be correct in app-services, but might be less so in m-c for now (eg, currently we just disable uniffi-bindgen, whereas ultimately we obviously still need to run it, once I work out how!)

### DIFF
--- a/build-scripts/component-common.gradle
+++ b/build-scripts/component-common.gradle
@@ -12,22 +12,22 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    ndkVersion rootProject.ext.build.ndkVersion
-    compileSdkVersion rootProject.ext.build.compileSdkVersion
-
     defaultConfig {
-        minSdkVersion rootProject.ext.build['minSdkVersion']
-        targetSdkVersion rootProject.ext.build['targetSdkVersion']
+        ndkVersion config.ndkVersion
+        compileSdkVersion config.compileSdkVersion
+
+        minSdkVersion config.minSdkVersion
+        targetSdkVersion config.targetSdkVersion
 
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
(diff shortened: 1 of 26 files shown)


ncalexan:
Nothing here looks particularly likely to cause pain in the future.

I'm generally concerned that we don't have a lot of high-level oversight of the Gradle build system, and that's causing slow builds and awkward development experiences (that look to get worse before they get much better), but that's the world we're in right now and nothing here is likely to make that more than linearly worse.

mhammond:
Awesome, thanks Nick! I guess I might as well open this for review then!

bendk (approved):
I like the approach in general and the changes seem correct.

rvandermeulen (approved):
Approach seems reasonable to me, deferring to others for the more specific Gradle changes. Note that there's some docs which probably need updating for these changes (like NDK updates).

mhammond:
I updated the docs and a couple of other gradle tweaks around uniffi, so I'm going to land this!

PR #6558, opened 2025-01-14
AMP FTS experiment logic

bendk:
This sets it up so we always ingest the FTS data and use the `SuggestionProviderConstraints` passed to the query to determine how to perform the query.  This seems like the simplest approach and it doesn't increase ingestion time that much.  The benchmarks on my machine went from 339.29 ms to 465.60 ms.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -21,7 +21,7 @@ use crate::{
     fakespot,
     geoname::GeonameCache,
     pocket::{split_keyword, KeywordConfidence},
-    provider::SuggestionProvider,
+    provider::{AmpMatchingStrategy, SuggestionProvider},
     query::FtsQuery,
     rs::{
         DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedAmpWikipediaSuggestion,
@@ -322,34 +322,66 @@ impl<'a> SuggestDao<'a> {
         query: &SuggestionQuery,
         suggestion_type: AmpSuggestionType,
     ) -> Result<Vec<Suggestion>> {
+        let strategy = query
+            .provider_constraints
+            .as_ref()
+            .and_then(|c| c.amp_alternative_matching.as_ref());
+        match strategy {
+            None => self.fetch_amp_suggestions_using_keywords(query, suggestion_type, true),
+            Some(AmpMatchingStrategy::NoKeywordExpansion) => {
+                self.fetch_amp_suggestions_using_keywords(query, suggestion_type, false)
+            }
+            Some(AmpMatchingStrategy::FtsAgainstFullKeywords) => {
+                self.fetch_amp_suggestions_using_fts(query, suggestion_type, "full_keywords")
+            }
+            Some(AmpMatchingStrategy::FtsAgainstTitle) => {
+                self.fetch_amp_suggestions_using_fts(query, suggestion_type, "title")
+            }
+        }
+    }
+
+    pub fn fetch_amp_suggestions_using_keywords(
+        &self,
+        query: &SuggestionQuery,
+        suggestion_type: AmpSuggestionType,
+        allow_keyword_expansion: bool,
+    ) -> Result<Vec<Suggestion>> {
         let keyword_lowercased = &query.keyword.to_lowercase();
         let provider = match suggestion_type {
             AmpSuggestionType::Mobile => SuggestionProvider::AmpMobile,
             AmpSuggestionType::Desktop => SuggestionProvider::Amp,
         };
+        let where_extra = if allow_keyword_expansion {
+            ""
+        } else {
+            "AND INSTR(CONCAT(fk.full_keyword, ' '), k.keyword) != 0"
+        };
         let suggestions = self.conn.query_rows_and_then_cached(
-            r#"
-            SELECT
-              s.id,
-              k.rank,
-              s.title,
-              s.url,
-              s.provider,
-              s.score,
-              fk.full_keyword
-            FROM
-              suggestions s
-            JOIN
-              keywords k
-              ON k.suggestion_id = s.id
-            LEFT JOIN
-              full_keywords fk
-              ON k.full_keyword_id = fk.id
-            WHERE
-              s.provider = :provider
-              AND k.keyword = :keyword
-            AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
-            "#,
+            &format!(
+                r#"
+                SELECT
+                  s.id,
+                  k.rank,
+                  s.title,
+                  s.url,
+                  s.provider,
+                  s.score,
+                  fk.full_keyword
+                FROM
+                  suggestions s
+                JOIN
+                  keywords k
+                  ON k.suggestion_id = s.id
+                LEFT JOIN
+                  full_keywords fk
+                  ON k.full_keyword_id = fk.id
+                WHERE
+                  s.provider = :provider
+                  AND k.keyword = :keyword
+                  {where_extra}
+                AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
+                "#
+            ),
             named_params! {
                 ":keyword": keyword_lowercased,
                 ":provider": provider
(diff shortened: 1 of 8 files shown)


0c0w3 (approved):
Thanks, I think there are some open questions but I'll go ahead and r+ to unblock you. Also, would you be able to add any tests?

components/suggest/src/db.rs
```diff
@@ -427,6 +477,96 @@ impl<'a> SuggestDao<'a> {
...
+                ":provider": provider
+            },
+            |row| -> Result<Suggestion> {
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let raw_url: String = row.get("url")?;
+                let score: f64 = row.get("score")?;
+
+                // We don't have a good way to determine the full keyword, let's just leave that
+                // blank for now.
```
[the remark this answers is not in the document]
bendk: That behavior sounds right to me.  I thought that desktop would use the search string when we sent back a blank string, but in hindsight I'm not sure why I would think that.

I don't think we can do better without investing a lot more into FTS. We can get a full keyword string, but how do we know it's the right one?  In the case if title matching, there might not even be a full keyword to display.

components/suggest/src/db.rs
```diff
@@ -427,6 +477,96 @@ impl<'a> SuggestDao<'a> {
...
+                FROM
+                  suggestions s
+                JOIN
+                  amp_fts fts
+                  ON fts.rowid = s.id
+                WHERE
+                  s.provider = :provider
+                  AND amp_fts match '{fts_column}: {keyword_lowercased}'
+                AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
+                "#
```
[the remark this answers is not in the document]
bendk: Great point.  I'm thinking we can `rank` and de-dupe the keywords before inserting them.  I'm not sure if that will work well or not, but it seems like it's worth a shot.

components/suggest/src/rs.rs
```diff
@@ -436,6 +436,12 @@ impl DownloadedSuggestionCommonDetails {
             },
         )
     }
+
+    /// Get the full keywords as a single string
+    pub fn full_keywords_string(&self) -> String {
```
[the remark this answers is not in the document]
bendk: Good point, I went `full_keywords_joined` because it's easier to spell :).

components/suggest/src/db.rs
```diff
@@ -321,35 +321,85 @@ impl<'a> SuggestDao<'a> {
...
-              ON k.suggestion_id = s.id
-            LEFT JOIN
-              full_keywords fk
-              ON k.full_keyword_id = fk.id
-            WHERE
-              s.provider = :provider
-              AND k.keyword = :keyword
-            AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
-            "#,
+            if allow_keyword_expansion {
```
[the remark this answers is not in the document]
bendk: I heard that term from Nan.  My understanding is that there's a phase where they add keywords for related terms and misspellings and we can detect those keywords by the substring test.

components/suggest/src/db.rs
```diff
@@ -427,6 +477,96 @@ impl<'a> SuggestDao<'a> {
...
+            |row| -> Result<Suggestion> {
+                let suggestion_id: i64 = row.get("id")?;
+                let title = row.get("title")?;
+                let raw_url: String = row.get("url")?;
+                let score: f64 = row.get("score")?;
+
+                // We don't have a good way to determine the full keyword, let's just leave that
+                // blank for now.
+
+                self.conn.query_row_and_then(
```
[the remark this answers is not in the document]
bendk: I tried factoring this out, but there's actually some differences because the keywords are different.  Let's leave the duplicate code in for now, if we end up supporting both exact keywords and AMP, we can refactor later.

bendk:
Thanks for the review, I made most of the suggested changes and left notes on the others.  Do you want to review again?

components/suggest/src/rs.rs
```diff
@@ -436,6 +436,19 @@ impl DownloadedSuggestionCommonDetails {
             },
         )
     }
+
+    /// Get the full keywords as a single string
+    pub fn full_keywords_joined(&self) -> String {
+        let parts: HashSet<_> = self.full_keywords.iter().map(|(s, _)| s.as_str()).collect();
+        let mut result = String::new();
+        for (i, part) in parts.into_iter().enumerate() {
+            if i == 0 {
```
0c0w3: This should be `i > 0`, right? And IIUC you're using a hash set to discard duplicate full keywords, but that's not the problem, and I'd be surprised if AMP included duplicate full keywords for one suggestion. The problem is that many of a suggestion's full keywords can contain the same substring, e.g. "amazon". So you'd need to `split_whitespace()` on each full keyword string and then collect each resulting word into a hash set. That should work.

You know, it might be worth preprocessing this and including the joined full keywords string in the data itself. I bet it wouldn't be too hard to modify our AMP upload script, and it probably wouldn't increase the dataset size by too much.

components/suggest/src/db.rs
```diff
@@ -974,6 +1097,11 @@ impl<'a> SuggestDao<'a> {
                     wiki_insert.execute(suggestion_id, wikipedia)?;
                 }
             }
+            fts_insert.execute(
+                suggestion_id,
+                &common_details.full_keywords_joined(),
+                &common_details.title,
+            )?;
```
0c0w3, line 1104: Sorry, just thought of something: Should this be gated on provider constraints in the ingest constraints? Most people won't be in the experiment, and ideally they wouldn't pay the cost of ingesting into the FTS table. Desktop can easily pass in some ingest constraints for users in the experiment. Maybe it's not a big deal though.

components/suggest/src/rs.rs
0c0w3: Ah my `i > 0` comment is outdated, thanks.

components/suggest/src/db.rs
bendk: My concern with that is that we won't re-ingest records if they've already been ingested, and I couldn't figure out a good way to handle this in `should_reprocess_record` with the data we currently have.

It does add some time to ingestion, but not all that much.  I put some metrics in the title.

Maybe we could store the provider constraints in a column in `ingested_records`, but that seems more like a long-term improvement to me.

0c0w3: Seems doable: If the constraints are `FtsAgainstFullKeywords` or `FtsAgainstTitle` and the FTS table is empty, then reprocess? I'm assuming it's simple to tell whether the table is empty like it is for non-virtual/non-FTS tables. I don't want to block on that though, and thanks for doing the benchmarks.

> Maybe we could store the provider constraints in a column in `ingested_records`, but that seems more like a long-term improvement to me.

I had that thought too when I added `should_reprocess_record`. That would be a general, sure-fire way of reingesting, so it might be worth pursuing.

components/suggest/src/provider.rs
```diff
@@ -130,4 +130,21 @@ pub struct SuggestionProviderConstraints {
     /// settings record(s).
     #[uniffi(default = None)]
     pub exposure_suggestion_types: Option<Vec<String>>,
+    /// Which experiment branch should we use for the AMP provider while the FTS experiment is
+    /// ongoing?
+    /// `0` means the treatment branch.
+    #[uniffi(default = None)]
+    pub amp_alternative_matching: Option<AmpMatchingStrategy>,
```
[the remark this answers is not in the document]
0c0w3: Nit: Sorry, in the same spirit of renaming this struct member, would you mind rewording the doc comment so it doesn't talk about experiments? And now that it's an enum, I think it would be fine for the comment to only mention what `None` means (i.e., the usual exact keyword matching), if you'd like.

0c0w3 (approved):
Thanks, lgtm!

components/suggest/src/store.rs
```diff
@@ -1147,6 +1149,128 @@ pub(crate) mod tests {
...
+                // the space key.
+                ..SuggestionQuery::amp("los pollos ")
+            }),
+            vec![los_pollos_suggestion("los pollos")],
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn amp_fts() -> anyhow::Result<()> {
```
0c0w3, line 1203: Naming nit: `amp_fts_against_full_keywords`

components/suggest/src/db.rs
bendk: Good point, I think we can do it.  My last push added this.

bendk:
This PR got a bit messy because it's been approved, but I keep adding more code to it.  How are you feeling?  Merge it or do you want to review the code to conditionally ingest the FTS data?

0c0w3:
I'll take a quick look now.

PR #6560, opened 2025-01-15
AMP FTS metrics

bendk:
These seem like they could be useful for the experiment.  I'm not sure we'll be able to hook them up but we might as well try.

Refactored the logic to get the full keywords and added tests for it since we're now using it in two places.

PR note: This one's on top of the other since I think that one's pretty close to landing and this one I'm less sure of.  The second commit has changes specific to this one.

  - Note:

### DIFF
--- a/components/suggest/src/query.rs
+++ b/components/suggest/src/query.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::collections::HashSet;
+
 use crate::{LabeledTimingSample, Suggestion, SuggestionProvider, SuggestionProviderConstraints};
 
 /// A query for suggestions to show in the address bar.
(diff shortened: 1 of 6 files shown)


components/suggest/src/query.rs
```diff
@@ -297,4 +321,25 @@ mod test {
...
+        let mut expected_counts = HashMap::<&str, usize>::new();
+        let mut actual_counts = HashMap::<&str, usize>::new();
+        for term in expected.split_whitespace() {
+            *expected_counts.entry(term).or_default() += 1;
+        }
+        let fts_content = full_keywords_to_fts_content(input);
+        for term in fts_content.split_whitespace() {
+            *actual_counts.entry(term).or_default() += 1;
+        }
+        assert_eq!(actual_counts, expected_counts);
```
0c0w3, line 343: Why not simply `assert_eq!(fts_content, expected)`? That would simplify this test logic, making bugs less likely, and I think it's OK to make an assumption about the order of words in `fts_content`.

0c0w3 (approved):
Thanks, lgtm.

components/suggest/src/query.rs
0c0w3: I see, I didn't realize that.

PR #6563, opened 2025-01-18
Bug 1930985 - Implement search engine order handling on the Rust base…

mandysGit:
…d SearchEngineSelector.

  - Note:

### DIFF
--- a/components/search/src/configuration_types.rs
+++ b/components/search/src/configuration_types.rs
@@ -240,11 +240,28 @@ pub(crate) struct JSONDefaultEnginesRecord {
     pub specific_defaults: Vec<JSONSpecificDefaultRecord>,
 }
 
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONEngineOrder {
+    /// The specific environment to match for this record.
+    pub environment: JSONVariantEnvironment,
+
+    /// The order of the engines for the associated environment. If engines are
+    /// present for the user but not included in this list, they will follow
+    /// after the ones in this list in alphabetical order. If an individual entry
+    /// is suffixed with a star, matching is applied on a \"starts with\" basis.
+    #[serde(default)]
+    pub order: Vec<String>,
+}
+
 /// Represents the engine orders record.
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub(crate) struct JSONEngineOrdersRecord {
-    // TODO: Implementation.
+    /// When a user's instance matches the defined environments, the associated
+    /// engine order will be applied. The array is ordered, when multiple entries
+    /// match on environments, the later entry will override earlier entries.
+    pub orders: Vec<JSONEngineOrder>,
 }
 
 /// Represents an individual record in the raw search configuration.
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -116,6 +116,7 @@ pub(crate) fn filter_engine_configuration(
     user_environment.version = user_environment.version.to_lowercase();
 
     let mut default_engines_record = None;
+    let mut engine_orders_record = None;
 
     for record in configuration {
         match record {
@@ -126,8 +127,8 @@ pub(crate) fn filter_engine_configuration(
             JSONSearchConfigurationRecords::DefaultEngines(default_engines) => {
                 default_engines_record = Some(default_engines);
             }
-            JSONSearchConfigurationRecords::EngineOrders(_engine_orders) => {
-                // TODO: Implementation.
+            JSONSearchConfigurationRecords::EngineOrders(engine_orders) => {
+                engine_orders_record = Some(engine_orders)
             }
             JSONSearchConfigurationRecords::Unknown => {
                 // Prevents panics if a new record type is added in future.
@@ -137,6 +138,19 @@ pub(crate) fn filter_engine_configuration(
 
     let (default_engine_id, default_private_engine_id) =
         determine_default_engines(&engines, default_engines_record, &user_environment);
+
+
+    if let Some(orders_record) = engine_orders_record {
+        for order_data in &orders_record.orders {
+            if matches_user_environment(&order_data.environment, &user_environment) {
+                set_engine_order(&mut engines, &order_data.order);
+            }
+        }
+    }
+
+    engines.sort_by(|a, b| {
+        sort(default_engine_id.as_ref(), default_private_engine_id.as_ref(), a, b)
+    });
 
     Ok(RefinedSearchConfig {
         engines,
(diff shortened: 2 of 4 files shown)


mandysGit:
This PR is still work in progress and there will be more changes soon. 
I need a first pass review, please. 

I've added the implemention for `orderHint` property. I included setting the sort priority and sorting the engines orders. 

The part I need some feedback on right now is the first test I wrote, and of course everything else if there's any suggestions. 

I converted this test into Rust in the `selector.rs`: 

https://searchfox.org/mozilla-central/source/toolkit/components/search/tests/xpcshell/test_engine_selector_engine_orders.js#163-173

And I'm copying over this engine config from the top of the file:

https://searchfox.org/mozilla-central/source/toolkit/components/search/tests/xpcshell/test_engine_selector_engine_orders.js#15-43

As I was following the test structure in `selector.rs` I noticed the config and expected_engines parts are duplicated and very long. The way we've done it in Desktop is that we hadn't need to write all of the properties in. I wanted some feedback if I should continue the tests in this format even though they're quite long and has lots of duplication? 

My aim is to continue writing [tests for starts_with and defaults engines](https://searchfox.org/mozilla-central/source/toolkit/components/search/tests/xpcshell/test_engine_selector_engine_orders.js#208-240). But that would mean I'll need to create another json config and another expected_engines list -- which I think is fine to do but there's a lot of duplication. 



Curious to know what your thoughts on this are?

components/search/src/configuration_types.rs
```diff
@@ -240,11 +240,28 @@ pub(crate) struct JSONDefaultEnginesRecord {
...
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JSONEngineOrder {
+    /// The specific environment to match for this record.
+    pub environment: JSONVariantEnvironment,
+
+    /// The order of the engines for the associated environment. If engines are
+    /// present for the user but not included in this list, they will follow
+    /// after the ones in this list in alphabetical order. If an individual entry
+    /// is suffixed with a star, matching is applied on a \"starts with\" basis.
```
Standard8, line 252: nit: Shouldn't need the backslashes here.

components/search/src/filter.rs
```diff
@@ -241,6 +266,49 @@ fn find_engine_with_match(
...
+
+fn sort(
+    default_engine_id: Option<&String>,
+    default_private_engine_id: Option<&String>,
+    a: &SearchEngineDefinition,
+    b: &SearchEngineDefinition
+) -> std::cmp::Ordering {
+    let b_index = get_priority(b, default_engine_id, default_private_engine_id);
+    let a_index = get_priority(a, default_engine_id, default_private_engine_id);
+    b_index.cmp(&a_index)
```
Standard8, line 309: There's a piece missing here, which the engine selector on desktop doesn't do - the items which aren't default/have an order hint set, need to be sorted alphabetically by their display name.

This is the full function we use: https://searchfox.org/mozilla-central/rev/5b061cdc4d40d44988dc61aa941cfbd98e31791f/toolkit/components/search/SearchUtils.sys.mjs#362-440

One thing I'm not sure about is how we'll handle the locale-based sorting like we do on desktop. I haven't looked into the options for sort functions on Rust.

components/search/src/filter.rs
```diff
@@ -221,7 +235,18 @@ fn find_engine_id(engines: &[SearchEngineDefinition], engine_id: String) -> Opti
     }
 }
 
-fn find_engine_with_match(
+fn set_engine_order(engines: &mut [SearchEngineDefinition], ordered_engines: &[String]) {
```
Standard8, line 238: I'm tempted to say we could put the sort related functions into a separate module, which might make it easier to write unit tests for them (and smaller module sizes), what do you think?

Standard8 (changes requested):

mandysGit:
I addressed the review comments above. Please resolve them if you think they're done. 

I added unit tests for sort function and I kept the larger tests in `selector.rs`. The tests in `selector.rs` have been refactored so it's a bit easier to read because I extract out only the relevant pieces which are the engine ids for comparing. 

I'm thinking maybe we can refactor the rest of the tests with a similar helper function and in combination with the builder pattern you've been working on. 

Lastly, I'm a bit worried the git checks won't pass since it didn't pass last time. I'm not too sure how to run the lint locally or fix them.

components/search/src/sort_helpers.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    fn create_engine(engine_id: &str, order_hint: Option<u32>) -> SearchEngineDefinition {
+        SearchEngineDefinition {
+            identifier: engine_id.to_string(),
+            order_hint,
+            ..Default::default()
+        }
+    }
+
+    #[test]
```
Standard8, line 78: We should add a unit test here for the "starts with" match. We then wouldn't necessarily need the test in selector.rs, since we know that part is being tested here.

Standard8 (changes requested):
Thank you for the updates, this is looking a lot nicer. I've just left a few comments in-line to consider.

components/search/src/sort_helpers.rs
```diff
@@ -0,0 +1,249 @@
...
+        ];
+        set_engine_order(
+            &mut engines,
+            &[
+                "c-engine".to_string(),
+                "b-engine".to_string(),
+                "a-engine".to_string(),
+            ],
+        );
+        let expected_order = vec![
```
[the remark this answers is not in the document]
mandysGit: I agree. This was confusing. 
I made some changes to both the tests for set engine orders: 

- I used `expected_order_hints` and `actual_order_hints`
- changed the engine ids to not use `a`, `b,`, `c`. Instead I used: `last-engine`, `secondary-engine`, `primary-engine` and `no-order-hint-engine` to make it more clear we're not expecting alphabetical sorting here
- Changed the test descriptions 
- extracted out the ordered_engines_lists as variable to pass into `set_engine_order`. I felt like, when it was passed in as a vector literal, it wasn't clear that it's the ordered list.

components/search/src/sort_helpers.rs
mandysGit: I've added a unit test in `sort_helpers.rs` to test for the "starts with" match. 

I'm hesitant to remove the test in selectors.rs because it does more than just verify the "starts with" match. It also ensures the correct filtering based on environment, verifying that it finds the correct `wiki-uk` or `wiki-ca` engines based on locale and region when given `wiki*` as the engine_id to match. 

Whereas the "starts with" match in sort_helpers, I can only test it finds the first engine that starts with `wiki` so it will always match `wiki-ca` and and ignore `wiki-uk`.

Standard8 (approved):
Thank you, this looks good. Just the headers in the new file to sort out.

PR #6565, opened 2025-01-22
Relax chrono version requirement for nimbus-cli.

mhammond:
mozilla-central is on 0.4.19 and we'd like nimbus to work there without forcing m-c to upgrade.

All other chrono deps here use this loose version.


mhammond:
welp, apparently we do rely on some version greater than 0.4.19

PR #6571, opened 2025-01-27
NSSKeyManager for logins component

jo:
Introduce a new trait `NSSKeyManager`, which provides an NSS-backed key manager for the logins component. Given a `PrimaryPasswordAuthenticator` implementation, the NSS keystore is used to store and retrieve the login encryption key. These features are only available when the Logins component is compiled with the `keydb` feature.

The `sync-pass` example has been adapted to use the NSSKeyManager. The example program can be called with an FX profile path in which the key is stored in the file key4.db and secured with a possibly set primary password.

Also add some more docs to `init_rust_components` and add it to book sidebar, and improve the nss init assertion error message.

  - Note:


components/logins/src/encryption.rs
```diff
@@ -160,6 +168,133 @@ impl KeyManager for StaticKeyManager {
...
 }
 
+/// `PrimaryPasswordAuthenticator` is used in conjunction with `NSSKeyManager` to provide the
+/// primary password and the success or failure actions of the authentication process.
+#[cfg(feature = "keydb")]
+#[uniffi::export(with_foreign)]
+pub trait PrimaryPasswordAuthenticator: Send + Sync {
+    /// Get a primary password for authentication, otherwise return the
+    /// AuthenticationCancelled error to cancel the authentication process.
+    fn get_primary_password(&self) -> ApiResult<String>;
```
bendk, line 178: I've been thinking about this method a lot lately and I think it should be `async` since asking the user to unlock the primary password is essentially an async process.  On Swift/Kotlin, maybe this could still be a sync method that blocks until a user response comes in, but AFAIK there's no way to do that on JS since it has a single threaded execution model.  Of course, adding an async callback method is tricky considering the logins code is all sync right now, but I think there are ways to handle it.

Here's all the options that I can see and my opinion of them:

1. Make the method async and:
     1. Use `block_on` from [pollster](https://lib.rs/crates/pollster), which will block the thread the sync function is running on until the async method completes.  This is my preferred option.
     2. Rewrite the logins code to be async.  I don't think this is realistic at all in the short-term, but maybe it's worth doing in the long-term.
2. Keep the method sync and:
    1. Pass a UniFFI interface that represents a promise/completion callback.  I think the promise object would use  [oneshot](https://docs.rs/oneshot/latest/oneshot/) channel under the hood, and the sync code would block waiting for it to complete.  I don't really like this solution, it seems like a more complicated version of `1.i` without any upside.
    2. Figure out a way to suspend a sync JS function while it awaits a promise.  I don't know of any way to do this, but I'm not a JS/SpiderMonkey expert
    3. Add something like the reverse of `wrapped-async` to uniffi-bindgen-gecko-js.  This would wrap an async JS function to work like a sync Rust function.  I think under the hood it would generate code that implemented `2.i`.  However, I'm not sure that we really want to add this code to the `uniffi-bindgen-gecko-js` and I don't think you want your project to depend on us doing that.

components/logins/src/encryption.rs
```diff
@@ -160,6 +168,133 @@ impl KeyManager for StaticKeyManager {
...
+                        .primary_password_authenticator
+                        .get_primary_password()
+                        .map_err(|e| LoginsApiError::AuthenticationError {
+                            reason: e.to_string(),
+                        })?;
+
+                    result = authenticate_with_primary_password(&primary_password).map_err(
+                        |e: nss::Error| LoginsApiError::NSSAuthenticationError {
+                            reason: e.to_string(),
+                        },
```
bendk, line 279: How do you feel about adding functions that wrap these NSS methods and map the error?  I think that could make the logic of this function more clear, but it's not a big deal either way.

examples/sync-pass/src/sync-pass.rs
```diff
@@ -274,32 +272,27 @@ fn prompt_record_id(s: &LoginStore, action: &str) -> Result<Option<String>> {
...
-        StaticKeyManager::new(encryption_key),
-    )));
-    let store = LoginStore::new(db_path, encdec)?;
-    Ok(store)
-}
+struct MyPrimaryPasswordAuthenticator {}
+impl PrimaryPasswordAuthenticator for MyPrimaryPasswordAuthenticator {
+    fn get_primary_password(&self) -> Result<String, LoginsApiError> {
+        let password = prompt_string("primary password").unwrap_or_default();
+        Ok(password.to_owned())
```
bendk, line 279: a) Thanks for updating `sync-pass`, it's great to keep the example code up-to-date.
b) In this case, a sync method is actually perfect.  However, I think this could be could be converted to be a "fake-async" method without too much work.  Given that there's no await points, maybe pollster could even poll it to completion by itself.

bendk (changes requested):
This looks great!

The only thing that's needed is to figure out the async questions surrounding `get_primary_password()`.  There's quite a bit of complexity there, but IMO `pollster::block_on` is a simple solution to that.

components/logins/src/encryption.rs
jo: Thats a great suggestion, did not know about pollster, which is just awesome!

examples/sync-pass/src/sync-pass.rs
jo: yes, it works with the async trait just the same way

components/logins/README.md
```diff
@@ -91,7 +91,28 @@ To effectively work on the Logins component, you will need to be familiar with:
...
+- Checking if primary password authentication is needed.
+- Coordinating with `PrimaryPasswordAuthenticator` to obtain the primary password when required.
+- Retrieving or creating an AES-256 key that is wrapped with the primary password if one is set.
+- Serializing the key into a JSON format for use by encryption and decryption routines.
+
+The `PrimaryPasswordAuthenticator` is a foreign trait that is used to supply the primary password necessary for unlocking the NSS key database.
+
+See the header comment in [`src/encryption.rs`](./src/encryption.rs) for a more detailed explanation of encryption options and key management in the logins component.
+
+The storage schema is based on the one
```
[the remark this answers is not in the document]
bendk: Absolutely, that would be great.

components/logins/src/encryption.rs
```diff
@@ -160,6 +174,136 @@ impl KeyManager for StaticKeyManager {
...
+            if result {
+                self.primary_password_authenticator
+                    .on_authentication_success();
+            } else {
+                while !result {
+                    self.primary_password_authenticator
+                        .on_authentication_failure();
+
+                    let primary_password =
+                        block_on(self.primary_password_authenticator.get_primary_password())?;
```
mhammond, line 288: I wonder if later we want this to be async too and move the `block_on` to the caller, but this is fine for now.

components/logins/src/logins.udl
```diff
@@ -69,6 +69,18 @@ dictionary Login {
...
+    NSSUninitialized();
+
+    /// NSS error during authentication
+    NSSAuthenticationError(string reason);
+
+    /// error during authentication (in PrimaryPasswordAuthenticator)
+    AuthenticationError(string reason);
+
+    /// authentication has been cancelled.
+    AuthenticationCancelled();
```
mhammond, line 82: fwiw, `Canceled` appears more common in american engine, so maybe change this. OTOH, neither you or I are American, so 🤷 :)

mhammond (approved):

components/logins/src/encryption.rs
jo: yes, good point. Would this be a breaking change for the key managers currently in use?

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 [Full Changelog](In progress)
 
+### Logins
+New trait `NSSKeyManager`, which provides an NSS-backed key manager. Given a
```
bendk, line 6:
```suggestion
New trait `KeyManager`, which provides an NSS-backed key manager. Given a
```

components/logins/src/encryption.rs
bendk: Yes it would be.  However, are there any foreign implementations of `KeyManager` yet?  If not, then I don't think it will actually matter in practice and this a good time to sneak in this breaking change before there are any.

If there are, then you'll need to be a bit more careful.  If the only difference is that we're going to be changing `func` to `async func` then I think it's okay to introduce the breaking change and fix it when the next a-s nightly comes out.  We can work through that one together.

Either way, it's seems easier to make this change now than later, but I'll let you make that call.  We could also land this now, then make just that change in a follow-up.

bendk (approved):
Looks good to me.  That `get_key` method should probably be async like Mark suggested, but it doesn't need to happen in this PR.

CHANGELOG.md
jo: oh good catch, `NSSKeyManager` is not a trait, but an actual implementation of the `KeyManager` trait. Fixed this

components/logins/src/encryption.rs
jo: `KeyManager` is currently implemented in `StaticKeyManager`, which is used for Android and iOS. I absolutely agree that this should be asynchronous, but I'd like to make the change in a follow-up later this week to keep things short and simple.

bendk: A follow-up is great.

`StaticKeyManager` is defined in the Rust code right?  I think that means you can make the change without having to change any of the consumer code, unless they're calling `get_key` themselves for some reason.  If all they're doing is creating a `StaticKeyManager` then passing it to the `LoginStore` constructor, then changing the method to be async should be no problem.

PR #6573, opened 2025-01-30
sql-support updates

bendk:
While working on an example component for new devs, I noticed a few warts with this component that I'd like to fix.

While working on an example component for new devs, I noticed a few warts with this component that I'd like to fix.

Added some utility function and started using them in `suggest`.

Better sql schema upgrade checking:
* Normalize whitespace.  This way we don't need to write awkward upgrade   code to ensurue the whitespace exactly matches.  Updated the tests to reflect this.  IMO the new formating looks nicer, but the important part is that the check function is more forgiving and either way would work.
* Better error messages on failure

  - Note:

### DIFF
--- a/components/support/sql/src/lib.rs
+++ b/components/support/sql/src/lib.rs
@@ -38,6 +38,33 @@ pub fn escape_string_for_pragma(s: &str) -> String {
     s.replace('\'', "''")
 }
 
+/// Default SQLite pragmas
+///
+/// Most components should just stick to these defaults.
+pub fn setup_sqlite_defaults(conn: &rusqlite::Connection) -> rusqlite::Result<()> {
+    conn.execute_batch(
+        "
+        PRAGMA temp_store = 2;
+        PRAGMA journal_mode = WAL;
+        ",
+    )?;
+    let page_size: usize = conn.query_row("PRAGMA page_size", (), |row| row.get(0))?;
+    // Aim to checkpoint at 512Kb
+    let target_checkpoint_size = 2usize.pow(19);
+    // Truncate the journal if it more than 3x larger than the target size
+    let journal_size_limit = target_checkpoint_size * 3;
+    conn.execute_batch(&format!(
+        "
+        PRAGMA wal_autocheckpoint = {};
+        PRAGMA journal_size_limit = {};
+        ",
+        target_checkpoint_size / page_size,
+        journal_size_limit,
+    ))?;
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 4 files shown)


components/autofill/src/db/schema.rs
```diff
@@ -98,13 +98,13 @@ impl ConnectionInitializer for AutofillConnectionInitializer {
         Ok(db.execute_batch(CREATE_SHARED_SCHEMA_SQL)?)
     }
 
-    fn upgrade_from(&self, db: &Transaction<'_>, version: u32) -> Result<()> {
```
mhammond, line 101: tbh I don't think this is an improvement as the new name is vague. If you feel strongly about it, I think it should be named `upgrade_to`

mhammond (changes requested):
I'm fairly underwhelmed by this as the choice of "to" vs "from" is very subjective and this is adding risk (how confident can we be that we didn't miss a case here?). I do think the new names have made things worse though, so we'd need to fix that at least.

(A problem I have with the "to_" names is that it seems less clear to me that we are upgrading incrementally - that "update_to_v3" kinda gives the impression you can pass a v1 schema in, where an "upgrade_from_v3" doesn't have the same implication - "upgrade_from_v3" obviously can not upgrade to versions which don't exist yet, so a simple increment is the only possible semantics it could have).

mhammond:
TBH I'm fine with your judgement now I've expressed my concerns. I do feel the change to plain "upgrade" is a step backwards though.

bendk:
Let's be conservative about this one.  I removed the changes to `upgrade_from` and left the rest.

components/support/sql/src/lib.rs
```diff
@@ -38,6 +38,33 @@ pub fn escape_string_for_pragma(s: &str) -> String {
     s.replace('\'', "''")
 }
 
+/// Default SQLite pragmas
+///
+/// Most components should just stick to these defaults.
+pub fn setup_sqlite_defaults(conn: &rusqlite::Connection) -> rusqlite::Result<()> {
```
mhammond, line 44: this appears to only be used by tests - is that intentional?

bendk: The intention was to just use these for this example component that I'm putting together and future components.  However, we should probably eat our dogfood.  I updated suggest to start using this.  I could switch over more components too if you think that makes sense.

components/support/sql/src/open_database.rs
```diff
@@ -118,6 +120,38 @@ pub fn open_database_with_flags<CI: ConnectionInitializer, P: AsRef<Path>>(
...
+) -> Result<Connection> {
+    open_database_with_flags(
+        path,
+        OpenFlags::SQLITE_OPEN_URI
+            | OpenFlags::SQLITE_OPEN_NO_MUTEX
+            | OpenFlags::SQLITE_OPEN_CREATE
+            | OpenFlags::SQLITE_OPEN_READ_WRITE,
+        connection_initializer,
+    )
+}
```
bendk, line 136: I also think this function and the next are nice to add.  I'd rather handle the SQLite flags in this crate than in a component.

mhammond: I wonder if later we should just return the flags, so consumers with a slightly different requirement can still levergage this?

mhammond (approved):
Thanks, I just wanted to make sure I understood! We can iterate on this if we find it inadequate.

components/support/sql/src/open_database.rs
bendk: I like that idea, just updated the code to match.

PR #6574, opened 2025-01-30
Move CLIs to component subdirectories

bendk:
It makes sense to me to put these next to the component they're for. Also moved cli-support to components/support/cli.

Tried to normalize the names a bit, so `sync-pass` becomes `logins-cli` and can be executed with `cargo logins`.

Left behind the push examples.  I didn't look into these much, but they seem more like actual examples to me than the other ones.

  - Note:


bendk:
This layout makes sense to me, but I'm curious to know how other devs feel about it.

mhammond:
What's the motivation here? I mildly prefer the status quo TBH (mainly because, eg, I'd prefer `rg foo components` to not include this non-prod code) and it probably makes less sense once this repo is in m-c (eg, searchfox should eventually treat them differently too?)

bendk:
For me, it lessens the friction to creating and maintaining these.  This code feels similar to the unit tests to me, it's not what's executing in prod, but they can an important part of development.  That said, it's not a big deal either way.  The status quo is fine.

The thing that bothers me most is calling all of these "examples".  They seem much closer to binaries to me.  If we don't move them here, how would you feel about changing the `Cargo.toml` to reflect that and maybe moving them into a top-level `cli` or `bin` directory?

mhammond:
I'm a little torn between "bin" and "example" - "bin" implies something useful and something you might ship whereas example implies something which demonstrates how the API is to be used, which actually is more accurate for the ones I'm familiar with. IOW, I think something like `nimbus-cli` probably is a "bin", but "places-utils" isn't really something we really expect anyone else to *actually* run. In m-c terms, having a "bin" implies you need to choose between saying it runs on the "compile target" (ie, may be cross-compiled), or run on the "host" because it is used to *build* Firefox - but neither of these really make sense.

I guess I don't feel too strongly about this, and do agree that having them closer to the actual code has appeal - but is there a reason to do this before we move to m-c which will just force us to revisit this once again with a different set of constraints? I guess I don't really object if you feel the problems the status quo causes is worth it, but tbh I don't really see them.

bendk:
That makes sense.  I'm going to just close this one.  Let's not make changes now and we can revisit this once we move repos.

PR #6578, opened 2025-02-03
Bug 1945621 - Re‐enable missing‐parent checks and fix ON CONFLICT par…

skhamis:
…ent logic

[Bug 1935797](https://bugzilla.mozilla.org/show_bug.cgi?id=1935797) added

    Instead of using CHECK constraints, we can create BEFORE INSERT and BEFORE UPDATE triggers on moz_bookmarks to enforce the invariants that we want. Unlike a CHECK constraint, a trigger gives us access to all values in the new row, and lets us fail an insert or update with a custom error message.

Which allowed issues like [bug 1876320](https://bugzilla.mozilla.org/show_bug.cgi?id=1876320) to have more specific debugging messages and allowed us to see more clearly in what exact constraint is failing. While it was technically "correct" in what the code was doing, it caused a spike of new issues like in [bug 1941655](https://bugzilla.mozilla.org/show_bug.cgi?id=1941655).

From :markh's comment in [bug 1941655](https://bugzilla.mozilla.org/show_bug.cgi?id=1941655)

    My best guess is that this is caused by this "ON CONFLICT" clause - note the lack of "parent" in that clause.
    I can reproduce something similar to this with these STR:
    Run our places-utils example to setup a DB with bookmarks via sync.
    Use a sqlite editor and locate a folder which has bookmarks.
    Delete the parent item (thus making the bookmarks tree invalid as items will reference a non-existing parent)
    Run places-utils again to sync.
    When doing this I get an error from the above-linked SQL about a conflict on guid - however, note the above clause is only for id. So if we modify the SQL such that guid is also handled (ie, copy the ON CONFLICT clause changing id to guid in the copy), and then we can get the same error we see.
    Naively, I tried changing the ON CONFLICT clause to include parent and set it to the root ID as described in the comments - this does allow the sync to complete without error, but causes tests to fail as the final tree is not what's expected - ie, the omission of parent appears intentional and that in this scenario, dogear has created a valid tree. IOW, it does appear that the condition here can be seen for real and does not imply the final bookmarks tree will be invalid.

So I think making a patch that deals with the parent in the ON CONFLICT as mentioned above, however with one slight difference.

What we should do is on conflict, we only update the parent if the new parent GUID actually exists in moz_bookmarks. This avoids referencing a non-existent parent, which would trigger an error. If the incoming parent is NULL or invalid, we keep the existing parent. That preserves any valid local parent info, and avoids forcing a fallback to the root prematurely. For newly inserted items (the "INSERT" path), we already set the parent to the root as a placeholder, and a later "fix up" step reattaches them to the correct parent once it's guaranteed to exist.

  - Note:

### DIFF
--- a/components/places/sql/create_shared_triggers.sql
+++ b/components/places/sql/create_shared_triggers.sql
@@ -61,12 +61,15 @@ BEGIN
     WHERE NEW.guid = 'root________'
       AND NEW.parent IS NOT NULL;
 
-    SELECT throw('update: item without parent')
+    SELECT throw(format(
+        'update: nonexistent parent: old_parent=%d, new_parent=%d, operation=%q',
+        OLD.parent,
+        NEW.parent,
+        CASE WHEN NEW.syncChangeCounter > 0 THEN 'sync' ELSE 'user' END
+    ))
     WHERE NEW.guid <> 'root________'
-      AND NEW.parent IS NULL;
--- bug 1941655, this seemingly more correct check causes obscure problems.
---      AND NOT EXISTS(
---          SELECT 1 FROM moz_bookmarks WHERE id = NEW.parent);
+    AND NOT EXISTS(
+        SELECT 1 FROM moz_bookmarks WHERE id = NEW.parent);
 
     SELECT throw(format('update: old type=%d; new=%d', OLD.type, NEW.type))
     WHERE OLD.type <> NEW.type;
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -498,6 +498,14 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp) 
     // when we apply the new local structure. This `INSERT` is a full table
     // scan on `itemsToApply`. The no-op `WHERE` clause is necessary to
     // avoid a parsing ambiguity.
+    // On conflict, we only update the parent if the new parent GUID actually
+    // exists in moz_bookmarks. This avoids referencing a non-existent parent,
+    // which would trigger an error. If the incoming parent is NULL or invalid,
+    // we keep the existing parent. That preserves any valid local parent info,
+    // and avoids forcing a fallback to the root prematurely. For newly inserted
+    // items (the "INSERT" path), we already set the parent to the root as a
+    // placeholder, and a later "fix up" step reattaches them to the correct
+    // parent once it's guaranteed to exist.
     log::debug!("Upserting new items");
     scope.err_if_interrupted()?;
     db.execute_batch(&format!(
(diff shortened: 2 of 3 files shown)


components/places/sql/create_shared_triggers.sql
```diff
@@ -61,12 +61,15 @@ BEGIN
...
 
-    SELECT throw('update: item without parent')
+    SELECT throw(format(
+        'update: nonexistent parent: old_parent=%d, new_parent=%d, operation=%q',
+        OLD.parent,
+        NEW.parent,
+        CASE WHEN NEW.syncChangeCounter > 0 THEN 'sync' ELSE 'user' END
+    ))
     WHERE NEW.guid <> 'root________'
-      AND NEW.parent IS NULL;
```
lougeniaC64, line 66: This might be worth spelling out in the PR description, but why are we rolling back this change? I assumed we were keeping it and continuing to research why it's problematic. I'm not necessarily opposed to making this change, I just think an explanation should be provided.

skhamis: ah yeah I was going to detail a little of why would should add back in and added the relevant comment in the `ON CONFLICT` section but i'll also detail it in the description too!

components/places/src/bookmark_sync/engine.rs
```diff
@@ -519,6 +527,14 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
          FROM itemsToApply
          WHERE 1
          ON CONFLICT(id) DO UPDATE SET
+         parent = CASE
```
mhammond, line 530: I don't think we understand the situation correctly yet. If this change is removed it seems the tests all continue to pass, because dogear already ensures what the comment describes. 

I think we need to understand exactly when this "ON CONFLICT" clause is entered, which I don't think we do (right?)

skhamis: I believe you're correct in the sense that this might not be the best way to go about it. I will put up a new patch that goes a more logging/telemetry way.

mhammond: I really do think we need to understand that `ON CONFLICT` clause - our tests hit it, so it should be possible to deduce exactly why it exists

skhamis: I believe we use the `ON CONFLICT(id)` in that insert statement because Dogear may decide that a supposedly "new" item from the server actually corresponds to an existing local row in moz_bookmarks. Instead of failing when we "insert" that row a second time, we update it in place. https://github.com/mozilla/dogear/blob/f83e6db9640332bb4d8b9689740a7db68ba087f3/src/tree.rs#L208-L239

At least from what I understand this ON CONFLICT. I don't know necessarily if this is the right "spot" that the problem is happening. My chief guess right now is that there are some old bookmarks left over with missing parentIds and when we switched to triggers, they got caught with the more stricter checks (which is why we don't see an increasing amount of the previous error over time). My next PR will be to run an orphan check on the local DBs and see if this theory is true

PR #6580, opened 2025-02-04
[DISCO-3199] Merino - Initial start to Component

tiftran:

  - Note:


tiftran:
So this is my initial attempt at starting the Merino component.

components/merino/src/merino.rs
```diff
@@ -0,0 +1,844 @@
...
+    De,
+    #[serde(rename = "de-DE")]
+    DeDe,
+    #[serde(rename = "de-AT")]
+    DeAt,
+    #[serde(rename = "de-CH")]
+    DeCh,
+}
+// Configuration settings for a Section
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
tiftran, line 42: gut feeling told me that i should be adding something like `uniffi::Record` but it wasn't too happy with me....

components/merino/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod merino;
```
ncloudioj, line 1: Suggestion: Merino has many endpoints (suggest, curated_recommendation, manifest, etc.), how about naming the mods accordingly? E.g. `curation` for now, `suggest` and `manifest` can be added in the future.

components/merino/src/merino.rs
ncloudioj: It should just work. What errors are you getting?

components/merino/src/merino.rs
```diff
@@ -0,0 +1,844 @@
...
+    pub region: Option<String>,
+    pub count: Option<i32>,
+    pub topics: Option<Vec<String>>,
+    pub feeds: Option<Vec<String>>,
+    pub sections: Option<Vec<SectionSettings>>,
+    #[serde(rename = "experimentName")]
+    pub experiment_name: Option<String>,
+    #[serde(rename = "experimentBranch")]
+    pub experiment_branch: Option<String>,
+}
```
bendk, line 65: Is this going to be exposed to the consumer.  You might want to make this derive `uniffi::Record` and also add ` #[uniffi(default = None]` for the optional fields.  (https://mozilla.github.io/uniffi-rs/next/types/records.html).  This could be done in a second pass though.

components/merino/src/merino.rs
```diff
@@ -0,0 +1,844 @@
...
+    pub has_excerpt: bool,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json;
+
+    #[test]
+    fn test_curated_reccomendations_request_deserialization() {
```
bendk, line 210: I hesitate to say this, but do we need these unit tests?  It feels like it's testing the serde code more than anything.    I could be missing something though, I'll let you make the call on if you think they're needed or not.

components/merino/src/merino.rs
bendk: Maybe you need `uniffi::setup_scaffolding!()` in `lib.rs`?

bendk (approved):
Looks like a good start to me.

adudenamedruby:
This looks great. And I'm learning about serde, which I've never used!

Side-note: I'm always still caught off guard by `.rs` files with A TON OF LINES OF CODE... only to realize that most of it is actually just tests. I saw this one file in a repo that was, like, nearly 3k lines, of which only 300 lines were "actual code" and the rest was tests. I'm so used to separated test files from Xcode & Swift, but I do like this style more because it's much easier for test context.

components/merino/src/curation.rs
```diff
@@ -0,0 +1,200 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+use serde::{Deserialize, Serialize};
+
+// Locales supported by Merino Curated Reccomendations
+#[derive(Debug, Serialize, Deserialize, PartialEq, uniffi::Enum)]
```
ncloudioj, line 8: Nit: I assume `Locale` would only be used by consumers to specify the `locale` field in the POST request, then we can just derive `Serialize` for it w/o deriving `Deserialize` to reduce the compiling overhead as the compiler won't generate code for deserialization. 

In general, I think for structs or enums defined here, we can just pick either `Serialize` or `Deserialize` to derive unless it's used for both in some rare cases. Deriving `Deserialize` for structs/enums so we can load Merino API JSON response into Rust structs. Likewise, `Serialize` can be derived to serialize Rust structs into API requests.

ncloudioj (approved):
Looks good, r+wc.

ncloudioj (approved):

PR #6589, opened 2025-02-06
Bug 1945621 - Run check for users to re-parent orphaned bookmarks and…

skhamis:
… log outcome

[Bug 1935797](https://bugzilla.mozilla.org/show_bug.cgi?id=1935797) added

Instead of using CHECK constraints, we can create BEFORE INSERT and BEFORE UPDATE triggers on moz_bookmarks to enforce the invariants that we want. Unlike a CHECK constraint, a trigger gives us access to all values in the new row, and lets us fail an insert or update with a custom error message.

Which allowed issues like [bug 1876320](https://bugzilla.mozilla.org/show_bug.cgi?id=1876320) to have more specific debugging messages and allowed us to see more clearly in what exact constraint is failing. While it was technically "correct" in what the code was doing, it caused a spike of new issues like in [bug 1941655](https://bugzilla.mozilla.org/show_bug.cgi?id=1941655).

From :markh's comment in [bug 1941655](https://bugzilla.mozilla.org/show_bug.cgi?id=1941655)

> My best guess is that this is caused by this "ON CONFLICT" clause - note the lack of "parent" in that clause.
> I can reproduce something similar to this with these STR:
Run our places-utils example to setup a DB with bookmarks via sync.
Use a sqlite editor and locate a folder which has bookmarks.
Delete the parent item (thus making the bookmarks tree invalid as items will reference a non-existing parent)
Run places-utils again to sync.
When doing this I get an error from the above-linked SQL about a conflict on guid - however, note the above clause is only for id. So if we modify the SQL such that guid is also handled (ie, copy the ON CONFLICT clause changing id to guid in the copy), and then we can get the same error we see.
Naively, I tried changing the ON CONFLICT clause to include parent and set it to the root ID as described in the comments - this does allow the sync to complete without error, but causes tests to fail as the final tree is not what's expected - ie, the omission of parent appears intentional and that in this scenario, dogear has created a valid tree. IOW, it does appear that the condition here can be seen for real and does not imply the final bookmarks tree will be invalid.

To continue testing the above theory. I think adding better logging and telemetry might be the next small step to figuring out what is going on. Dogear technically still does the right thing but mainly operates on syncing bookmarks. This patch hopes to do a couple of things:

1. Add better logging around the `ON CONFLICT` part of the code, since we've relaxed the trigger, we've seen an uptick of `UNIQUE CONSTRAINT: moz_bookmarks.guid`. Thus more logging around the guid should help here.
2. Check for any orphaned bookmarks that somehow already exist in the tree, we expect this to be relatively low to (1) but a potential hit.

  - Note:

### DIFF
--- a/components/places/sql/create_shared_triggers.sql
+++ b/components/places/sql/create_shared_triggers.sql
@@ -62,9 +62,7 @@ BEGIN
       AND NEW.parent IS NOT NULL;
 
     SELECT throw(format(
-        'update: nonexistent parent: old_parent=%d, new_parent=%d, operation=%q',
-        OLD.parent,
-        NEW.parent,
+        'update: item without parent: operation=%q',
         CASE WHEN NEW.syncChangeCounter > 0 THEN 'sync' ELSE 'user' END
     ))
     WHERE NEW.guid <> 'root________'
--- a/components/places/sql/create_shared_triggers.sql
+++ b/components/places/sql/create_shared_triggers.sql
@@ -61,7 +61,12 @@ BEGIN
     WHERE NEW.guid = 'root________'
       AND NEW.parent IS NOT NULL;
 
-    SELECT throw('update: item without parent')
+    SELECT throw(format(
+        'update: nonexistent parent: old_parent=%d, new_parent=%d, operation=%q',
+        OLD.parent,
+        NEW.parent,
+        CASE WHEN NEW.syncChangeCounter > 0 THEN 'sync' ELSE 'user' END
+    ))
     WHERE NEW.guid <> 'root________'
       AND NEW.parent IS NULL;
 -- bug 1941655, this seemingly more correct check causes obscure problems.
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -364,7 +364,18 @@ fn update_local_items_in_places(
 
     log::debug!("Changing GUIDs");
     scope.err_if_interrupted()?;
-    db.execute_batch("DELETE FROM changeGuidOps")?;
+    let result = db.execute_batch("DELETE FROM changeGuidOps");
+    // Due to bug 1935797, guid collisions are happening in some step of bookmarks
+    // but not clear if it's the "fixup" step, so we catch and log to sentry
+    if let Err(rusqlite::Error::SqliteFailure(e, _)) = &result {
+        if e.code == ErrorCode::ConstraintViolation {
+            error_support::report_error!(
+                "places-sync-bookmarks-guid-collision",
+                "changeGuidOps ran into guid collision violation"
+            );
+        }
+    }
+    result?;
 
     log::debug!("Applying remote items");
     apply_remote_items(db, scope, now)?;
@@ -373,6 +384,35 @@ fn update_local_items_in_places(
     log::debug!("Applying new local structure");
     scope.err_if_interrupted()?;
     db.execute_batch("DELETE FROM applyNewLocalStructureOps")?;
+
+    // Similar to the check in apply_remote_items, however we do a post check
+    // to see if dogear was unable to fix up the issue
+    let orphaned_count: i64 = db.query_row(
+        "WITH RECURSIVE orphans(id) AS (
+           SELECT b.id
+           FROM moz_bookmarks b
+           WHERE b.parent IS NOT NULL
+             AND NOT EXISTS (
+               SELECT 1 FROM moz_bookmarks p WHERE p.id = b.parent
+             )
+           UNION
+           SELECT c.id
+           FROM moz_bookmarks c
+           JOIN orphans o ON c.parent = o.id
+         )
+         SELECT COUNT(*) FROM orphans;",
+        [],
+        |row| row.get(0),
+    )?;
+
+    if orphaned_count > 0 {
+        log::warn!("Found {} orphaned bookmarks after sync", orphaned_count);
+        error_support::report_error!(
+            "places-sync-bookmarks-orphaned",
+            "found local orphaned bookmarks after we applied new local structure ops: {}",
+            orphaned_count,
+        );
+    }
 
     log::debug!("Resetting change counters for items that shouldn't be uploaded");
     sql_support::each_chunk_mapped(
@@ -491,8 +531,8 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp) 
         }
     ).ok(); // We ignore the error because it's expected, users should not have entries with the same id that have different types
 
-    // Same as above but we check if any users have any undetected orphaned bookmarks and
-    // report them appropiately
+    // Due to bug 1935797, we need to check if any users have any
+    // undetected orphaned bookmarks and report them
     let orphaned_count: i64 = db.query_row(
         "WITH RECURSIVE orphans(id) AS (
            SELECT b.id
@@ -513,9 +553,8 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp) 
 
     if orphaned_count > 0 {
         log::warn!("Found {} orphaned bookmarks during sync", orphaned_count);
-        error_support::report_error!(
-            "places-sync-bookmarks-orphaned",
-            "found local orphaned bookmarks {}",
+        error_support::breadcrumb!(
+            "places-sync-bookmarks-orphaned: found local orphans before upsert {}",
             orphaned_count
         );
     }
@@ -559,7 +598,7 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp) 
     let result = db.execute_batch(&upsert_sql);
 
     // In trying to debug bug 1935797 - relaxing the trigger caused a spike on
-    // guid collisions, we want to report on this error before the actual upsert to see
+    // guid collisions, we want to report on this during the upsert to see
     // if we can discern any obvious signs
     if let Err(rusqlite::Error::SqliteFailure(e, _)) = &result {
         if e.code == ErrorCode::ConstraintViolation {
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -24,6 +24,7 @@ use dogear::{
     UploadTombstone,
 };
 use interrupt_support::SqlInterruptScope;
+use rusqlite::ErrorCode;
 use rusqlite::Row;
 use sql_support::ConnExt;
 use std::cell::RefCell;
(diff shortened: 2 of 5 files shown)


components/places/src/bookmark_sync/engine.rs
```diff
@@ -493,6 +494,32 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
...
+         )
+         SELECT COUNT(*) FROM orphans;",
+        [],
+        |row| row.get(0),
+    )?;
+
+    if orphaned_count > 0 {
+        log::warn!("Found {} orphaned bookmarks during sync", orphaned_count);
+        breadcrumb!("bookmarks-sync: found local orphaned bookmarks");
+        error_support::report_error!("bookmarks-sync", "orphaned bookmarks {}", orphaned_count);
```
skhamis, line 520: @lougeniaC64 I couldn't find an easy way to add telemetry to this output here so I kinda use "sentry" for this. Is there potentially anything wrong here? I know you did some work for this reporting stuff so not sure if i'm missing something.

components/places/src/db/schema.rs
```diff
@@ -325,6 +325,39 @@ pub fn upgrade_from(db: &Connection, from: u32) -> rusqlite::Result<()> {
                 rusqlite::params![NEW_SQL],
             )?;
         }
+        18 => {
+            // Fix any multi-level orphans by reparenting them to "unfiled_____". We'll do this in
+            // one shot using a recursive CTE.
+            db.execute_batch(
```
skhamis, line 331: @mhammond I was going to do this before the `ON CONFLICT` statement and log but rather not have this run on every sync (even if we do a one shot and add some kinda flag that we've done it). However I believe this would _prove_ the current theory that somehow the DB got into a bad state and dogear never ended fixing it up later. If true, we would start seeing the bookmark engine going back up and then the trigger would not end up breaking for migrated users moving forward

components/places/src/bookmark_sync/engine.rs
lougeniaC64: @skhamis That looks right to me but you can also test this by using the [test a-s sentry project](https://mozilla.sentry.io/settings/projects/application-services-test/keys/) to be sure. If you change the sentry DSN in either of the mobile apps it should report to the test project (though I've only done this in iOS). This can be especially useful for double-checking the breadcrumbs.

components/places/src/db/schema.rs
mhammond: I'm not sure what you mean. I'm very skeptical that 20% of the databases out there are corrupt, which is what this would imply given the ~20% of failures with the new check we backed out?

mhammond: shouldn't we do this only after we have evidence of a problem?

skhamis: You're right I think this is a little heavy-handed here. I moved this to the `apply_remote_items` part of the code since it'll help us give a better idea of the state of the users DB. I don't expect to fully keep the logging in there and this is only temporary.

components/places/src/bookmark_sync/engine.rs
skhamis: Thanks! I also decided to add a bit of test scaffolding for the `report_error` so we can actually use them in tests for checking things!

components/places/src/bookmark_sync/engine.rs
```diff
@@ -524,7 +553,25 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
...
+
+    // In trying to debug bug 1935797 - relaxing the trigger caused a spike on
+    // guid collisions, we want to report on this error before the actual upsert to see
+    // if we can discern any obvious signs
+    if let Err(rusqlite::Error::SqliteFailure(e, _)) = &result {
+        if e.code == ErrorCode::ConstraintViolation {
+            // This is a real collision that the sync code couldn't unify or fix.
+            error_support::report_error!(
+                "places-sync-bookmarks-guid-collision",
+                "GUID collision during bookmark sync upsert"
```
skhamis, line 569: I'm trying to think if there is anything valuable we can add in this reporting error statement but guids would be all but worthless to us in terms of understanding what's going on (unless we potentially have what it's colliding with?)

components/places/sql/create_shared_triggers.sql
```diff
@@ -61,7 +61,12 @@ BEGIN
     WHERE NEW.guid = 'root________'
       AND NEW.parent IS NOT NULL;
 
-    SELECT throw('update: item without parent')
+    SELECT throw(format(
```
skhamis, line 64: So this probably won't help us _right now_ as we don't hit this constraint anymore since we've relaxed the trigger, however I think we should keep the stronger debugging here incase we decide to risk it and try change back to the more "correct" trigger

components/places/src/bookmark_sync/engine.rs
```diff
@@ -490,14 +491,42 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
         }
     ).ok(); // We ignore the error because it's expected, users should not have entries with the same id that have different types
 
+    // Same as above but we check if any users have any undetected orphaned bookmarks and
```
lougeniaC64, line 494: nit: This comment is a bit confusing because I don't know what "same as above" is referring to. Is it the select statement on `itemsToApply`?

components/places/src/bookmark_sync/engine.rs
```diff
@@ -490,14 +491,42 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
...
+    // report them appropiately
+    let orphaned_count: i64 = db.query_row(
+        "WITH RECURSIVE orphans(id) AS (
```
lougeniaC64, line 497: I just want to be sure I understand what's happening here. It looks like you're querying for any bookmarks without parents and any dupes of that orphan. Is that correct? If so, LGTM.

components/places/src/bookmark_sync/engine.rs
lougeniaC64: Everything I can think of would most likely be captured in pre-existing breadcrumbs/errors. Is there any error reporting around this issue already captured in Sentry?

components/places/sql/create_shared_triggers.sql
mhammond: I'm not sure about the new string - IIUC, it will mean each user with this problem will be reporting a different string (so counting it will be tricky?), and the specific unique string isn't going to help them or us? Even if we got a user who captured their exact string, how would it help us or that specific user?

(the constant operation does make sense though!)

components/places/src/bookmark_sync/engine.rs
```diff
@@ -490,14 +491,42 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
...
+           SELECT c.id
+           FROM moz_bookmarks c
+           JOIN orphans o ON c.parent = o.id
+         )
+         SELECT COUNT(*) FROM orphans;",
+        [],
+        |row| row.get(0),
+    )?;
+
+    if orphaned_count > 0 {
```
mhammond, line 514: We also talked a little about doing this both before (like here) and after (somewhere below) a sync so we can get a better idea whether dogear really does fix all cases up - I think this would be useful to know because it might inform next steps.

(I'm also very very slightly worried about perf here for users with many thousands of deep bookmarks. I wonder if it might be worth counting bookmarks first and only doing these expensive recursive checks if there are less than, say 500. Really happy for you to decide this though!)

components/places/src/bookmark_sync/engine.rs
mhammond, line 569: (ie, see comment above, maybe here is also a good place to check if items remained orphaned?)

If you do decide to do this, maybe we just want a single log/report_error, reporting both the "before" and "after" values?

mhammond (commented):
This is looking good, only real concern is the "random" IDs in our messages?

components/places/sql/create_shared_triggers.sql
skhamis: Thinking about this more, I'm inclined to agree -- I was thinking if there was a possibility that we'd see potentially some static guids like `root__..., menu____` but those would be captured by other triggers. I removed them for now to keep things simpler as we'll have better logging in the actual code.

components/places/src/bookmark_sync/engine.rs
skhamis: Yes! Basically just checking if there is a possibility of existing orphans (which should be impossible but yeknow...), since the tree could be deeply nested, the recursive helps ensure it's not happening at any level

components/places/src/bookmark_sync/engine.rs
skhamis: Great idea! I decided to add a breadcrumb here instead that will be added to the report later if dogear ends up _not_ fixing it. This will prevent any noise/red herrings that were orphans here but fixed up later.

I wrote a quick test on testing that query with 9k bookmarks (but not deeply nested) and it ran about 7ms. Which doesn't tell us much on a real android or iOS device but since this is something we'll be closely watching in nightly, I'd like to cast as wide a net as possible here and we can quickly limit/eliminate this depending on what the data will tell us in the next few days

components/places/src/bookmark_sync/engine.rs
skhamis: Nothing in sentry, Tarik tried to add some `report_error` on the `fk` part but that didn't report anything.

skhamis: I did end up adding an "after" for this but that had to be in the method higher up `update_local_items_in_places` where the triggers:
```
db.execute_batch("DELETE FROM changeGuidOps");
...
db.execute_batch("DELETE FROM applyNewLocalStructureOps")?;
```
run. That way we definitely know at that point dogear has fixed up what it could.

components/places/sql/create_shared_triggers.sql
mhammond: that would have kinda made sense, but I think they were logging `parent` which is the rowid.

components/places/sql/create_shared_triggers.sql
```diff
@@ -61,7 +61,10 @@ BEGIN
...
+        'update: item without parent: operation=%q',
+        CASE WHEN NEW.syncChangeCounter > 0 THEN 'sync' ELSE 'user' END
```
mhammond, line 66: I'm actually not sure we can entirely trust this `operation` - the sync change counter might be > 0 even for user operations (eg, on a second operation done to the bookmark before syncing). But this seems fine, so long as we understand this.

components/places/src/bookmark_sync/engine.rs
```diff
@@ -373,6 +385,35 @@ fn update_local_items_in_places(
...
```
mhammond, line 408: It might be tricky to use sentry to tie together how many users see one or both of these, but that's fine for now.

mhammond (approved):
Thanks, let's do it!

PR #6592, opened 2025-02-07
Bug 1947694 - Add an API to the SearchEngineSelector to allow using the RemoteSettingService directly.

Standard8:

  - Note:

### DIFF
--- a/components/remote_settings/src/service.rs
+++ b/components/remote_settings/src/service.rs
@@ -63,7 +63,12 @@ impl RemoteSettingsService {
         context: Option<RemoteSettingsContext>,
     ) -> Result<Arc<RemoteSettingsClient>> {
         let mut inner = self.inner.lock();
-        let storage = Storage::new(inner.storage_dir.join(format!("{collection_name}.sql")))?;
+        // Allow using in-memory databases for testing of external crates.
+        let storage = if inner.storage_dir == ":memory:" {
+            Storage::new(inner.storage_dir.clone())?
+        } else {
+            Storage::new(inner.storage_dir.join(format!("{collection_name}.sql")))?
+        };
 
         let client = Arc::new(RemoteSettingsClient::new(
             inner.base_url.clone(),
@@ -83,7 +88,14 @@ impl RemoteSettingsService {
         #[allow(unused_variables)] context: Option<RemoteSettingsContext>,
     ) -> Result<Arc<RemoteSettingsClient>> {
         let mut inner = self.inner.lock();
-        let storage = Storage::new(inner.storage_dir.join(format!("{collection_name}.sql")))?;
+        // Allow using in-memory databases for testing of external crates.
+        let storage = if inner.storage_dir == ":memory:" {
+            Storage::new(Utf8PathBuf::from(
+                "file:remote-settings-{collection-name}?mode=memory&cache=shared".to_string(),
+            ))?
+        } else {
+            Storage::new(inner.storage_dir.join(format!("{collection_name}.sql")))?
+        };
         let client = Arc::new(RemoteSettingsClient::new(
             inner.base_url.clone(),
             inner.bucket_name.clone(),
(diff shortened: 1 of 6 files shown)


Standard8:
@bendk please can you check the remote settings changes? They're to help the tests work for us. You're welcome to look through the rest as well.

components/remote_settings/src/service.rs
```diff
@@ -63,7 +63,12 @@ impl RemoteSettingsService {
         context: Option<RemoteSettingsContext>,
     ) -> Result<Arc<RemoteSettingsClient>> {
         let mut inner = self.inner.lock();
-        let storage = Storage::new(inner.storage_dir.join(format!("{collection_name}.sql")))?;
+        // Allow using in-memory databases for testing of external crates.
+        let storage = if inner.storage_dir == ":memory:" {
+            Storage::new(inner.storage_dir.clone())?
```
bendk, line 68: Maybe this should be Something like `file:remote-settings-{collection-name}?mode=memory&cache=shared`.  That way clients for the same collection will get the same in-memory db.

bendk (approved):
Looks good to me.  I had a minor suggestion about the in-memory db path.  I think it's an improvement, but if it breaks your tests just ignore it.

components/search/src/filter.rs
```diff
@@ -121,62 +122,140 @@ impl SearchEngineDefinition {
...
+            default_engines_record: default_engines_record.cloned(),
+            engine_orders_record: engine_orders_record.cloned(),
+        })
+    }
+}
+
+pub(crate) fn filter_engine_configuration_impl(
     user_environment: SearchUserEnvironment,
-    configuration: Vec<JSONSearchConfigurationRecords>,
+    configuration: &impl FilterRecords,
```
mandysGit, line 219: I have a few suggestions on making this more clear. But I'll explain first the parts I found confusing. 

The function `filter_engine_configuration_impl` calls `filter_records()` to perform the actual filtering. However, the argument `configuration`, which is a type that implements the `FilterRecords` trait, is actually an unfiltered record when passed in. Because `configuration` is a type that implements `FilterRecords`, it makes it sound like it's a record that's already filtered. 

The trait actually just provides a filtering method,`filter_records`, but the name `FilterRecords` seem to suggest it might store filtered records. I think that's why it was unclear to me.

I suggest renaming the trait `FilterRecords` to `Records` since it represents a collection of unprocessed records. The filtering can then remain as a method, `filter_records`, within `Records`. This way, it's more intuitive that `filter_records` is an operation performed on the records.

mandysGit (changes requested):
Thank you for the update on this. The whole remote settings piece is still fuzzy to me. I'm still learning and left some inline comments for discussion, particularly around:  `filter_engine_configuration` part.

components/search/src/filter.rs
```diff
@@ -121,62 +122,140 @@ impl SearchEngineDefinition {
     }
 }
 
-pub(crate) fn filter_engine_configuration(
+pub(crate) struct FilterRecordsResult {
+    reduced_engines: Vec<SearchEngineDefinition>,
```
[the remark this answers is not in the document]
Standard8: I think htat's fine in this case, though I do worry we're getting too many `engines` about.

components/search/src/selector.rs
```diff
@@ -58,19 +82,33 @@ impl SearchEngineSelector {
...
+            } else {
+                return Err(Error::SearchConfigNoRecords);
+            }
+        }
         let data = match &self.0.lock().configuration {
             None => return Err(Error::SearchConfigNotSpecified),
             Some(configuration) => configuration.data.clone(),
         };
-        filter_engine_configuration(user_environment, data)
+        return filter_engine_configuration_impl(user_environment, &data);
```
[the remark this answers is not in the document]
mandysGit: Thank you. That clarifies it for me.

components/search/src/filter.rs
mandysGit: > The way I understand the trait documentation, is that the trait is defining the functionality a type might have, not what the type is.
 
Ahh I see. That makes more sense to me.

mandysGit: I think `Filter` sounds good.

mandysGit (approved):
Thank you for the updates.

PR #6593, opened 2025-02-07
Example Component

bendk:
Created an example component.  I'm hoping this eases the work required to create a new component and serves as a starting point for engineers that are new to cross-platform component development.

I'd love feedback on if the docs make sense or not from the POV of someone new to Rust and/or cross-platform development.

  - Note:

### DIFF
--- a/components/example/src/lib.rs
+++ b/components/example/src/lib.rs
@@ -1,0 +1,224 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Top-level module, see README.md for an overview of what's going on in this component.
+//!
+//! `lib.rs` defines the main public API for the component.
+
+mod db;
+mod error;
+mod http;
+mod schema;
+
+// Components use UniFFI to export Swift/Kotlin/JS bindings for their public API.
+//
+// * The [user guide](https://mozilla.github.io/uniffi-rs/latest/) gives an overview of UniFFI.
+// * App-services is currently in the process to switching to the [proc-macro based approach](https://mozilla.github.io/uniffi-rs/latest/proc_macro/index.html), away from the UDL-based approach.
+//
+// If you have questions, ask on the `#application-services-eng` channel in Slack.
+uniffi::setup_scaffolding!("example");
+
+pub use db::{SavedTodoItem, TodoItem};
+pub use error::{ApiError, ApiResult, Error, Result};
+
+use error_support::handle_error;
+
+/// Top-level API for the example component
+#[derive(uniffi::Object)]
+pub struct ExampleComponent {
+    // This example splits `ExampleComponent` into a public and internal part.
+    //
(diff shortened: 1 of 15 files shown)


components/example/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+    //
+    // `ExampleComponent` provides the public API, while `ExampleComponentInner` implements the
+    // internal API.  The two APIs are exactly the same, except `ExampleComponent` uses an
+    // `ApiResult`/`ApiError` and `ExampleComponentInner` uses `Result`/`Error`. See `error.rs` for
+    // a discussion of internal vs public results and errors.
+    //
+    //  The reason for this split is that `ExampleComponent` methods can't be called by other
+    //  functions, since they all expect a `Result` rather than an `ApiResult`.  This extra layer
+    //  requires writing extra boilerplate code, but it's usually worth it in the long-run.
+    inner: ExampleComponentInner,
```
bendk, line 40: This split was the thing I was least sure of when writing this code.  I've worked on a handful of components at this point and I think they all do something slightly different, so I wasn't sure what the example component should have.

If I was starting a new component today, this is how I would organize things.  However, is this what we should be recommending to other devs?  I'm not sure.  But then again, I also wanted to recommend something because I've run into the situation several times where the public/internal split wasn't there and it made writing the code awkward -- for example because I wanted to call some method, but couldn't because it returned the wrong result type.

What do others think?

bendk:
Once we merge this, we should link to it from some other documentation.  That could be the [app-services book](https://mozilla.github.io/application-services/), but that's gotten a bit out of date and I'm not sure it's worth updating that with the monorepo move coming up.  My current plan is to link to it from the [Rust Components section of firefox-source-docs](https://firefox-source-docs.mozilla.org/rust-components/index.html).

components/example/src/lib.rs
ncloudioj: This seems reasonable to me as a tutorial example.

ncloudioj (approved):
This is so good. My only suggestion would be advertising this in the [project doc](https://mozilla.github.io/application-services/book/index.html) and the [repo doc](https://github.com/mozilla/application-services?tab=readme-ov-file#rust-components).

components/example/src/schema.rs
```diff
@@ -0,0 +1,128 @@
...
+    const NAME: &'static str = "example-component";
+    const END_VERSION: u32 = VERSION;
+
+    /// Setup PRAGMAs for a new SQLite connection
+    ///
+    /// In general, just use `setup_sqlite_defaults` and maybe turn on foreign keys. It's
+    /// recommended to start with foreign key support on, but consider turning them off if you
+    /// notice performance issues with your component.
+    ///
+    /// If you have questions, ask the `#app-storage` channel in Slack.
```
tiftran, line 64: I really like these references. Thank you!

tiftran (approved):
Thank you for doing this. While reading through this I learned a lot and will probably referring to this often 🙂

bendk:
> This is so good. My only suggestion would be advertising this in the [project doc](https://mozilla.github.io/application-services/book/index.html) and the [repo doc](https://github.com/mozilla/application-services?tab=readme-ov-file#rust-components).

Agreed.  Doc links are going to be my next step.

PR #6596, opened 2025-02-14
WIP: Initialize nss explicitly

jo:
BREAKING CHANGE:
you now need to call `rc_crypto::ensure_initialized()` before any calls to functions which depend on NSS.

  - Note:


components/logins/src/logins.udl
```diff
@@ -5,6 +5,9 @@
 namespace logins {
     /// We expose the crypto primitives on the namespace
 
+    /// Initialize NSS
+    void ensure_initialized();
```
bendk, line 9: My one concern with this change is that applications are initializing NSS globally, through one of the components that depends on it.  This seems like small footgun -- if an application stops using `logins`, this could break the `autofill` integration.

How do you feel about adding UniFFI support to `rc_crypto` and making applications call `rc_crypto::ensure_initialized`?

bendk (commented):
I like this idea in general, the only question for me is how applications should initialize NSS.

components/logins/src/logins.udl
jo: Yes, this is also possible, also already (see the comment). I feel like this is a

components/support/rc_crypto/src/lib.rs
```diff
@@ -52,8 +52,8 @@ pub use ece;
 
 pub use crate::error::{Error, ErrorKind, Result};
 
-/// Only required to be called if you intend to use this library in conjunction
-/// with the `hawk` or the `ece` crate.
+/// Only required to be called if you intend to either use this library in conjunction with the
+/// `hawk` or the `ece` cratem, or use any functionality based on nss.
```
mhammond, line 56: nit: typo "cratem" - but it seems these docs could be tightened up - IIUC, the only time people use this crate is specifically for nss, so in practice I think it's now "you must call this"?

components/support/rc_crypto/nss/src/util.rs
```diff
@@ -21,6 +22,24 @@ static NSS_INIT: Once = Once::new();
...
 static NSS_PROFILE_PATH: OnceCell<String> = OnceCell::new();
 
+#[cfg(not(feature = "keydb"))]
+fn is_nss_initialized() -> bool {
+    NSS_INIT.is_completed()
+}
+
+#[cfg(feature = "keydb")]
+fn is_nss_initialized() -> bool {
+    NSS_INIT.is_completed() || NSS_PROFILE_PATH.get().is_some()
```
mhammond, line 32: this seems surprising to me - I guess I'd expect an `&&` here? Or possible an assertion if we expect both or neither to be true? Or even just a comment.

mhammond (commented):
Part of what I don't understand here is what initialization looks like when the "keydb" feature is enabled? Something like `logins` is eventually going to be used in both contexts, so how would that look?  (Sorry if I missed this!)

megazords/full/src/lib.rs
```diff
@@ -25,6 +27,17 @@ pub use viaduct;
...
+
+// TODO: would be nice if we could write
+//  #[uniffi::export]
+//  pub use nss::ensure_initialized as ensure_nss_initialized;
+// instead.
+#[uniffi::export]
+pub fn ensure_nss_initialized() {
+    ensure_initialized()
+}
+
```
jo, line 40: @bendk is there a way to do something like

```
#[uniffi::export]
pub use nss::ensure_initialized as ensure_nss_initialized;
```

or a way to avoid wrapping the function?

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
```diff
@@ -5,13 +5,15 @@
 package mozilla.appservices
 
 import org.mozilla.appservices.full_megazord.BuildConfig
+import org.mozilla.appservices.full_megazord.ensureNssInitialized
```
jo, line 8: @bendk

Any ideas what I am missing here?

It fails with
```
[task 2025-02-21T14:10:27.148Z] e: file:///builds/worker/checkouts/vcs/megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt:8:46 Unresolved reference: ensureNssInitialized
[task 2025-02-21T14:10:27.148Z] e: file:///builds/worker/checkouts/vcs/megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt:16:13 Unresolved reference: ensureNssInitialized
```
in [the tests](https://github.com/mozilla/application-services/pull/6596/checks?check_run_id=37607415245)

megazords/full/src/lib.rs
mhammond: There is https://mozilla.github.io/uniffi-rs/latest/proc_macro/functions.html#renaming-functions-methods-and-constructors, but I don't think that will actually work here :( So I don't think there is a good way, but it seems like a reasonable feature request!

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
mhammond: I'm expecting to see a gradle change somewhere with a line like `ext.configureUniFFIBindgen("full_megazord")`, which would end up generating the .kt file that needs to end up included. Eg, you will find that line in build.gradle in "components/logins", and you will find the generated file buried under "components/logins/android/build"

megazords/full/src/lib.rs
jo: doesn't look like there's already an issue filed for this, will do that

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
bendk: What happens if you add that `ext.configureUniFFIBindgen("full_megazord")` line to https://github.com/mozilla/application-services/blob/main/megazords/full/android/build.gradle?

If we're lucky that should fix things.  If not, I can take a look at that gradle code

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
```diff
@@ -5,13 +5,15 @@
...
 
 class Megazord {
     companion object {
         @JvmStatic
         fun init() {
             System.setProperty("mozilla.appservices.megazord.library", "megazord")
             System.setProperty("mozilla.appservices.megazord.version", BuildConfig.LIBRARY_VERSION)
+            ensureNssInitialized()
```
bendk, line 16: In the future, maybe we could initialize viaduct and other dependencies here too.

megazords/full/src/lib.rs
bendk: Thanks for filing that.  IMO, we could extend the concept of "external types" to also support "external functions".

megazords/ios-rust/src/lib.rs
```diff
@@ -21,3 +23,14 @@ pub use sync15;
...
+use nss::ensure_initialized;
+
+// TODO: would be nice if we could write
+//  #[uniffi::export]
+//  pub use nss::ensure_initialized as ensure_nss_initialized;
+// instead.
+#[uniffi::export]
+pub fn ensure_nss_initialized() {
+    ensure_initialized()
+}
```
bendk, line 36: a) I wonder if this should just be `initialize` and it could handle any initialization we needed (like `Megazord.init()` on Android).

b) Whatever change you make, it could be good to add a prep PR that adds the function, but doesn't do anything.  Then we could get the consumer applications to call this function before merging the actual breaking changes.  That removes the stress of having to resolve the breaking change the day the nightly is released.

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
jo: I did add this: https://github.com/mozilla/application-services/pull/6596/files#diff-4f7e37e066a724665e1f913003204ed1395ed7e4702cdfe4f9b2dbb3806f4366R75 but it fails with

```
[task 2025-02-21T15:00:29.190Z] A problem occurred evaluating project ':full-megazord'.
[task 2025-02-21T15:00:29.190Z] > No signature of method: org.gradle.internal.extensibility.DefaultExtraPropertiesExtension.configureUniFFIBindgen() is applicable for argument types: (String) values: [full_megazord]
```
see [the CI run](https://github.com/mozilla/application-services/pull/6596/checks?check_run_id=37610573091)

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
 ### Glean
 - Updated to v63.1.0 ([#6584](https://github.com/mozilla/application-services/pull/6584))
 
+### `nss`
+- Initialize nss explicitly ([#6596](https://github.com/mozilla/application-services/pull/6596))
+
+#### BREAKING CHANGE:
+you now need to call `nss::ensure_initialized()` before any NSS calls
```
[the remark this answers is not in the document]
jo: done

megazords/ios-rust/src/lib.rs
jo: I really like the idea :100: will do

jo: https://github.com/mozilla/application-services/pull/6607 has initialize function stubs

megazords/full/android/src/main/java/mozilla/appservices/Megazord.kt
jo: goes on in https://github.com/mozilla/application-services/pull/6607

megazords/full/src/lib.rs
jo: https://github.com/mozilla/uniffi-rs/issues/2461

components/support/rc_crypto/src/lib.rs
```diff
@@ -55,7 +55,6 @@ pub use crate::error::{Error, ErrorKind, Result};
 /// Only required to be called if you intend to use this library in conjunction
 /// with the `hawk` or the `ece` crate.
 pub fn ensure_initialized() {
-    nss::ensure_initialized();
```
bendk, line 58: Should this become `expect_nss_initialized()?`

components/init_rust_components/src/lib.rs
```diff
@@ -1,7 +1,32 @@
...
+///
+/// Receives the path to the profile directory.
+///
+/// For adding additional initialization code: Note that this function is called very early in the
+/// app lifetime and therefore affects the startup time. Only the most necessary things should be
+/// done here.
+#[cfg(feature = "keydb")]
+#[uniffi::export]
+pub fn initialize(profile_path: String) {
+    ensure_nss_initialized_with_profile_dir(profile_path).expect("could not initialize NSS")
```
bendk, line 31: Maybe this should be one function with the `#[cfg]` attributes wrapping different blocks inside or a separate private function.

I think that would make for better API docs, which we autogenerate for each exported function.  That code will only use one of these docstrings and I'm not sure which.

bendk (approved):
This looks great to me, I'm approving.  I left a couple minor comments -- however you want to deal with them is fine with me.

components/support/rc_crypto/src/lib.rs
jo: Good catch!

If we'd change the signature `pub fn ensure_initialized() -> Result<()> {` this result would propagate all over the place, for example to [HawkRequestBuilder::new](https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/internal/http_client.rs#L608-L616). I think that's fine, I'd go ahead and make those changes if you agree

components/init_rust_components/src/lib.rs
jo: How can I make one function accepting different arguments depending on `#[cfg]`?

jo: When I write
```
#[uniffi::export]
pub fn initialize(#[cfg(feature = "keydb")] profile_path: String) {
    #[cfg(not(feature = "keydb"))]
    ensure_nss_initialized();
    #[cfg(feature = "keydb")]
    ensure_nss_initialized_with_profile_dir(profile_path).expect("could not initialize NSS")
}
```
clippy says:
```
$ cargo clippy
    Checking init_rust_components v0.1.0 (/home/jo/moz/application-services/components/init_rust_components)
error[E0061]: this function takes 0 arguments but 1 argument was supplied
  --> components/init_rust_components/src/lib.rs:17:8
   |
16 | #[uniffi::export]
   | ----------------- unexpected argument of type `std::string::String`
17 | pub fn initialize(#[cfg(feature = "keydb")] profile_path: String) {
   |        ^^^^^^^^^^
   |
note: function defined here
  --> components/init_rust_components/src/lib.rs:17:8
   |
17 | pub fn initialize(#[cfg(feature = "keydb")] profile_path: String) {
   |        ^^^^^^^^^^

For more information about this error, try `rustc --explain E0061`.
error: could not compile `init_rust_components` (lib) due to 1 previous error
```

components/support/rc_crypto/src/lib.rs
jo: done

PR #6597, opened 2025-02-18
[DISCO-3200] Merino RUST Client - Curated Recommendations

misaniwere:
Adding a client for the merino curated recommendations endpoint

  - Note:

### DIFF
--- a/components/merino/src/curated_recommendations/error.rs
+++ b/components/merino/src/curated_recommendations/error.rs
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use error_support::{ErrorHandling, GetErrorHandling};
+pub type Result<T> = std::result::Result<T, Error>;
+pub type ApiResult<T> = std::result::Result<T, CuratedRecommendationsApiError>;
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum CuratedRecommendationsApiError {
+    #[error("Curated recommendations network error: {reason}")]
+    Network { reason: String },
+
+    #[error("Curated recommendations error: code {code:?}, reason: {reason}")]
+    Other { code: Option<u16>, reason: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("URL parse error: {0}")]
+    UrlParse(#[from] url::ParseError),
+
+    #[error("Error sending request: {0}")]
+    Request(#[from] viaduct::Error),
+
+    #[error("JSON error: {0}")]
+    Json(#[from] serde_json::Error),
+
+    #[error("Validation error ({code}): {message}")]
+    Validation { code: u16, message: String },
+
+    #[error("Bad request ({code}): {message}")]
+    BadRequest { code: u16, message: String },
+
+    #[error("Server error ({code}): {message}")]
+    Server { code: u16, message: String },
+
+    #[error("Unexpected error ({code}): {message}")]
+    Unexpected { code: u16, message: String },
+}
+
+impl GetErrorHandling for Error {
+    type ExternalError = CuratedRecommendationsApiError;
+
+    fn get_error_handling(&self) -> ErrorHandling<Self::ExternalError> {
+        match self {
+            Self::Request { .. } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Network {
+                    reason: self.to_string(),
+                })
+                .log_warning()
+            }
+            Self::Validation { code, message } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                    code: Some(*code),
+                    reason: format!("Validation error: {}", message),
+                })
+            }
+            Self::Server { code, message } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                    code: Some(*code),
+                    reason: format!("Server error: {}", message),
+                })
+            }
+            Self::Unexpected { code, message } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                    code: Some(*code),
+                    reason: format!("Unexpected error: {}", message),
+                })
+            }
+            Self::BadRequest { code, message } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                    code: Some(*code),
+                    reason: format!("Bad request: {}", message),
+                })
+            }
+            _ => ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                code: None,
+                reason: self.to_string(),
+            }),
+        }
+    }
+}
--- a/components/merino/src/curated_recommendations/http.rs
+++ b/components/merino/src/curated_recommendations/http.rs
@@ -10,17 +10,12 @@ use viaduct::{header_names, Request, Response};
 pub struct HttpClient;
 
 impl HttpClient {
-    pub fn new() -> Self {
-        Self {}
-    }
-
     pub fn make_curated_recommendation_request(
         &self,
         request: &CuratedRecommendationsRequest,
         user_agent_header: &str,
-        url: &str,
+        url: Url,
     ) -> Result<CuratedRecommendationsResponse, Error> {
-        let url = Url::parse(&url)?;
         log::trace!("making request: {url}");
         let response: Response = Request::post(url)
             .header(header_names::ACCEPT, "application/json")?
@@ -64,7 +59,7 @@ pub trait HttpClientTrait {
         &self,
         request: &CuratedRecommendationsRequest,
         user_agent_header: &str,
-        url: &str,
+        url: Url,
     ) -> super::error::Result<CuratedRecommendationsResponse>;
 }
 
@@ -73,7 +68,7 @@ impl HttpClientTrait for HttpClient {
         &self,
         request: &CuratedRecommendationsRequest,
         user_agent_header: &str,
-        url: &str,
+        url: Url,
     ) -> super::error::Result<CuratedRecommendationsResponse> {
         self.make_curated_recommendation_request(request, user_agent_header, url)
     }
--- a/components/merino/src/curated_recommendations/http.rs
+++ b/components/merino/src/curated_recommendations/http.rs
@@ -7,7 +7,7 @@ use super::models::{CuratedRecommendationsRequest, CuratedRecommendationsRespons
 use url::Url;
 use viaduct::{header_names, Request, Response};
 
-pub struct HttpClient {}
+pub struct HttpClient;
 
 impl HttpClient {
     pub fn new() -> Self {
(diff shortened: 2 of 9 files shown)


components/merino/curated_recommendations/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use error_support::{ErrorHandling, GetErrorHandling};
+pub type Result<T> = std::result::Result<T, Error>;
+pub type ApiResult<T> = std::result::Result<T, CuratedRecommendationsApiError>;
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum CuratedRecommendationsApiError {
+    #[error("Curated recommendations unexpected error: {reason}")]
```
bendk, line 11: Nit: "network error"

components/merino/curated_recommendations/src/http.rs
```diff
@@ -0,0 +1,75 @@
...
+            .json(request)
+            .send()?;
+
+        let status = response.status;
+        if status >= 400 {
+            let error_message = response.text();
+            let error = match status {
+                400 => Error::BadRequest {
+                    code: status,
+                    message: error_message.parse().unwrap(),
```
bendk, line 37: Do you need the `parse().unwrap()`?

 I think `error_message` should be a more-or-less a string -- maybe a `Cow`.  Something like `.to_string()`, `.to_owned`, or `.into()` should convert it to a `String`.

components/merino/curated_recommendations/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "curated_recommendations"
```
bendk, line 2: Why move/rename the crate?  Is the plan to create a separate client for the other endpoints?

bendk (changes requested):
My only question is about the crate name it's scope.  I don't think what you did was necessarily wrong, but I want to understand it a bit better before hitting approve.

Everything else was great, I liked the way you split up the error enums.  The tests are nice -- how much of a pain were they to write?

components/merino/curated_recommendations/src/http.rs
```diff
@@ -0,0 +1,75 @@
...
+        &self,
+        request: &CuratedRecommendationsRequest,
+    ) -> Result<CuratedRecommendationsResponse, Error> {
+        let url = Url::parse("https://merino.services.mozilla.com/api/v1/curated-recommendations")?;
+        log::trace!("making request: {url}");
+        let response: Response = Request::post(url)
+            .header(header_names::ACCEPT, "application/json")?
+            .header(
+                header_names::USER_AGENT,
+                "application-services merino client",
```
tiftran, line 26: just want to ask for some clarification for myself. 
This would give us the ability to figure out whether a merino request was done by the client, is that all we care about? or should we tact on platform/fx version, that gets passed in by the consumers?

misaniwere: yeah i'm not too sure what's required yet. i'm hoping the ios team can let us know what additional information they want in the header but i thought to start with this for now

components/merino/curated_recommendations/Cargo.toml
ncloudioj: I think we can keep the merino crate as is for the merino component and implement endpoints in different modules. For example:

```
components/merino
    |-- src
        |-- lib.rs
        |-- curated_recommendations
            |-- mod.rs
            |-- error.rs
            |-- http.rs
            |-- models.rs
        |-- suggest   => can be added in future iterations
            |-- mod.rs
            |-- errors.rs
            |-- models.rs
            |-- http.rs 
```

WDYT?

components/merino/curated_recommendations/src/http.rs
```diff
@@ -0,0 +1,75 @@
...
+impl HttpClient {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn make_curated_recommendation_request(
+        &self,
+        request: &CuratedRecommendationsRequest,
+    ) -> Result<CuratedRecommendationsResponse, Error> {
+        let url = Url::parse("https://merino.services.mozilla.com/api/v1/curated-recommendations")?;
```
ncloudioj, line 20: We might want to allow the client consumer to override the top-level URL (`https://merino.services.mozilla.com/`) so that they can use stage/local URLs if they want. They can do so by providing an optional configuration when they create the client.

components/merino/curated_recommendations/src/http.rs
ncloudioj: I think that the `User-Agent` header should be provided by the client consumer as a required configuration when they create the Merino client. Merino needs this to derive the platform/form factor (desktop/mobile) for each request it receives.

components/merino/curated_recommendations/Cargo.toml
misaniwere: ahh yes, that was what I was trying to go for. I guess my understanding of crate vs module in RUST is limited. is a crate related to the location of the `Cargo.toml` file?

components/merino/curated_recommendations/src/http.rs
misaniwere: okay makes sense @ncloudioj. I'll make that update

components/merino/curated_recommendations/src/http.rs
misaniwere: good point.

components/merino/curated_recommendations/Cargo.toml
bendk: Yes, the crate is where the `Cargo.toml` file is.

Zooming out, a crate is similar to a package in Python terms.  It's a collection of modules that:
  - Get distributed together as a single package on crates.io.  This also means when you add a dependency to `Cargo.toml`, you specify the crate and get all the modules.
  - Share some permissions.  For example, you can always implement traits defined in your own crate on any type, but implementing traits form other crates has some restrictions.
  - Compiled together.  This means you can have circular dependencies with modules, but not with crates.

misaniwere: that makes so much sense, thanks! @bendk

components/merino/src/curated_recommendations/http.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self {}
+    }
+
+    pub fn make_curated_recommendation_request(
+        &self,
+        request: &CuratedRecommendationsRequest,
+        user_agent_header: &str,
+        base_host: &str, // e.g. "https://merino.services.mozilla.com"
+    ) -> Result<CuratedRecommendationsResponse, Error> {
+        let full_url = format!("{}/api/v1/curated-recommendations", base_host);
```
ncloudioj, line 23: Nit: the `full_url` can be finalized and parsed once in `CuratedRecommendationsClient::new()` when the user specifies `base_host`. So this method can take a `Url` as the endpoint URL rather than `base_host`.

components/merino/src/curated_recommendations/http.rs
```diff
@@ -0,0 +1,80 @@
...
+        let url = Url::parse(&full_url)?;
+        log::trace!("making request: {url}");
+        let response: Response = Request::post(url)
+            .header(header_names::ACCEPT, "application/json")?
```
ncloudioj, line 27: If it's supported in viaduct, we can also do gzip.
```suggestion
            .header(header_names::ACCEPT, "application/json")?
            .header(header_names:ACCEPT_ENCODING, "gzip")?
```

ncloudioj (approved):
Looks good, r+wc.

components/merino/src/curated_recommendations/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+struct CuratedRecommendationsClientInner<T: http::HttpClientTrait> {
+    http_client: T,
+}
+
+#[uniffi::export]
+impl CuratedRecommendationsClient {
+    #[uniffi::constructor]
+    #[handle_error(Error)]
+    pub fn new(base_host: String, user_agent_header: String) -> ApiResult<Self> {
```
ncloudioj, line 29: I just realized that while `user_agent_header` should be required, `base_host` is better left as optional with the default being the production host. It's very likely that we will add more arguments like these in the future, so let's use the builder pattern for the client creation here. You can [reference](https://github.com/mozilla/application-services/blob/main/components/suggest/src/store.rs#L38-L128) this as an example.

components/merino/src/curated_recommendations/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+mod models;
+
+pub use error::{ApiResult, Error, Result};
+use error_support::handle_error;
+pub use models::{CuratedRecommendationsRequest, CuratedRecommendationsResponse};
+
+#[derive(uniffi::Object)]
+pub struct CuratedRecommendationsClient {
+    inner: CuratedRecommendationsClientInner<http::HttpClient>,
+    base_host: String,
```
ncloudioj, line 17: I think we can define an `endpoint_url` and leave `base_host` as a parameter.

```rs
pub struct CuratedRecommendationClient {
    inner: ...,
    endpoint_url: Url,
    user_agent_header: String,
}

impl CuratedRecommendationsClient {
    pub fn new(base_host: String, user_agent_header: String) -> ApiResult<Self> {
        let url = format!("{}/api/v1/curated-recommendations", base_host);
        let endpoint_url = Url::parse(&url)?;
        Ok(self {
             inner: CuratedRecommendationsClientInner::new()?,
             endpoint_url,  // Already a valid URL, no more parsing needed
             user_agent_header,
        })
    }
}
```

If we want to refactor this into a builder as suggested below, we can do something similar.

components/merino/src/curated_recommendations/http.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use super::error::Error;
+use super::models::{CuratedRecommendationsRequest, CuratedRecommendationsResponse};
+use url::Url;
+use viaduct::{header_names, Request, Response};
+
+pub struct HttpClient;
+
+impl HttpClient {
+    pub fn new() -> Self {
```
ncloudioj, line 13: `HttpClient` is a unit struct, it doesn't need a constructor anymore.

components/merino/src/curated_recommendations/error.rs
```diff
@@ -0,0 +1,83 @@
...
+                    code: Some(*code),
+                    reason: format!("Unexpected error: {}", message),
+                })
+            }
+            Self::BadRequest { code, message } => {
+                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
+                    code: Some(*code),
+                    reason: format!("Bad request: {}", message),
+                })
+            }
```
bendk, line 76: FWIW, I think you could consolidate these somewhat by combining the patterns and relying on the `Display` implementation that `thiserror::Error` generates:
```
            Self::Validation { code, .. }
           | Self::Server { code, ... }
           | Self::Unexpected { code, .. } 
           | Self::BadRequest { code, .. } => {
                ErrorHandling::convert(CuratedRecommendationsApiError::Other {
                    code: Some(*code),
                    reason: self.to_string(),
              })
          }
```

It's not a big deal either way.  I'm mostly just sharing in case the Rust knowledge is useful.

bendk (dismissed):
Looks great to me, thanks for working through this one!

components/merino/src/curated_recommendations/mod.rs
misaniwere: this work is blocking other tickets so i think i'll push this in for now and use the builder pattern later

components/merino/src/curated_recommendations/error.rs
misaniwere: thanks for the suggestion. I'll implement this as a follow up :)

ncloudioj (approved):
LGTM, thanks! :shipit:

PR #6598, opened 2025-02-19
[DISCO-3211] Suggest: move to new remote settings API

tiftran:

  - Note:


components/suggest/src/rs.rs
```diff
@@ -181,11 +157,23 @@ impl Record {
     }
 }
 
+impl From<Record> for RemoteSettingsRecord {
```
tiftran, line 160: cargo clippy said impl `From` is preferred over `Into`

components/suggest/src/rs.rs
```diff
@@ -214,6 +202,15 @@ pub(crate) enum SuggestRecord {
     Geonames,
 }
 
+impl SuggestRecord {
```
tiftran, line 205: This feels a little goofy just so that we can use `Record` to `get_attachment`, but setting `fields` to `Map::new( )` also felt weird

components/suggest/src/store.rs
```diff
@@ -98,19 +98,14 @@ impl SuggestStoreBuilder {
     }
 
     #[handle_error(Error)]
-    pub fn build(&self) -> SuggestApiResult<Arc<SuggestStore>> {
+    pub fn build(&self, rs_service: &RemoteSettingsService) -> SuggestApiResult<Arc<SuggestStore>> {
```
bendk, line 101: One thing that we haven't talked about is that this is going to be a breaking change for all applications, and it might not be so simple to fix.  I'm not sure they all have started using the new remote settings API and have a `RemoteSettingsService` ready to give us.

To make the transition easy, could you open a second PR that changes this signature to something like this:
```
#[uniffi::method(default(rs_service=None))]
pub fn build(&self, rs_service: Option<Arc<RemoteSettingsService>>)
```

That PR doesn't need to do actually do anything with `rs_service`.  The point is to give us a decent way to merge all the changes:
  - Merge the above, non-breaking, change. 
  - Get all consumers to start passing in the `rs_service`
  - Merge your actual PR, which technically will be a breaking change since `rs_service` is no longer optional, but it shouldn't be much work to fix the consumer apps.  Maybe they won't need any changes at all.

components/suggest/src/benchmarks/mod.rs
```diff
@@ -86,8 +95,15 @@ fn new_store() -> SuggestStore {
...
+    let rs_config = RemoteSettingsConfig2 {
+        bucket_name: None,
+        server: None,
+    };
+    let remote_settings_service =
+        Arc::new(RemoteSettingsService::new("".to_string(), rs_config).unwrap());
     std::fs::copy(starter_db_path, &db_path).expect("Error copying starter DB file");
-    SuggestStore::new(&db_path.to_string_lossy(), None).expect("Error building store")
+    SuggestStore::new(&db_path.to_string_lossy(), remote_settings_service)
+        .expect("Error building store")
```
bendk, line 106: Nice!  Thanks for keeping this code up to date.

components/suggest/src/rs.rs
```diff
@@ -69,85 +72,58 @@ pub(crate) trait Client {
...
-                },
-            )?,
+            quicksuggest_client: rs_service.make_client("quicksuggest".to_owned(), None)?,
+            fakespot_client: rs_service
+                .make_client("fakespot-suggest-products".to_owned(), None)?,
         })
     }
 
-    fn client_for_collection(&self, collection: Collection) -> &remote_settings::RemoteSettings {
+    fn client_for_collection(&self, collection: Collection) -> &Arc<RemoteSettingsClient> {
```
bendk, line 96: Nit: this could just be a `&RemoteSettingsClient`.  The `Arc` makes me think it might be cloned to create a new reference, but I don't think you ever do that.

components/suggest/src/rs.rs
```diff
@@ -69,85 +72,58 @@ pub(crate) trait Client {
...
-        let response = match cache {
-            None => client.get_records()?,
-            Some(cache) => remote_settings::cache::merge_cache_and_response(
-                cache,
-                client.get_records_since(last_modified)?,
-            ),
-        };
-        if last_modified != response.last_modified {
-            dao.write_cached_rs_data(collection.name(), &response);
+        let _ = client.sync();
```
bendk, line 109: This should be `client.sync()?`, that way errors get propagated up.

components/suggest/src/rs.rs
bendk: Makes sense, Rust will automatically generate the `Into` if you define a `From`.

components/suggest/src/rs.rs
bendk: Yes, slightly goofy but this makes sense to me.

components/suggest/src/rs.rs
```diff
@@ -214,6 +202,15 @@ pub(crate) enum SuggestRecord {
     Geonames,
 }
 
+impl SuggestRecord {
+    fn to_json_map(&self) -> Map<String, Value> {
+        match serde_json::to_value(self) {
+            Ok(Value::Object(map)) => map,
+            _ => Map::new(),
```
bendk, line 209: Maybe this branch could be `unreachable!()`.  I think we can safely say that `serde` will always be able to to deserialize to `Value::Object`.

bendk (changes requested):
This looks great.  I had some minor suggestions, but the only real issue is how we land this.  I'm marking this `Request Changes` just because I think we should land that preparatory PR I mentioned to make dealing with the breaking changes easier.

tiftran:
I see that there are a couple of PRs that will slightly change this one, I will wait and see when those merge and change accordingly

components/suggest/src/store.rs
0c0w3: `remote_settings_server` and `remote_settings_bucket_name` can be removed from `SuggestStoreBuilder` too.

0c0w3: Actually why not make `rs_service` a method on `SuggestStoreBuilder` instead of an arg to `build`?

tiftran:
I see that the other PRs are merged, so i rebased and made changes 🙂

0c0w3 (commented):
IIUC we no longer need our RS cache in `suggest`, right? And that's why this removes the RS-cache-related methods from `SuggestDao`. It looks like `rs_cache` is still referenced in `schema.rs`, so we'll also need to:

* remove those references
* add a schema migration that removes the `rs_cache` table

I think we need to do this before vendoring, so either in this PR or in a fast follow-up.

components/suggest/src/rs.rs
```diff
@@ -69,85 +72,58 @@ pub(crate) trait Client {
...
         }
     }
 }
 
-impl Client for RemoteSettingsClient {
-    fn get_records(&self, collection: Collection, dao: &mut SuggestDao) -> Result<Vec<Record>> {
+impl Client for SuggestRemoteSettingsClient {
+    fn get_records(&self, collection: Collection) -> Result<Vec<Record>> {
         // For now, handle the cache manually.  Once 6328 is merged, we should be able to delegate
         // this to remote_settings.
```
0c0w3, line 106: This comment can be removed.

components/suggest/src/rs.rs
```diff
@@ -69,85 +72,58 @@ pub(crate) trait Client {
...
-            None => client.get_records()?,
-            Some(cache) => remote_settings::cache::merge_cache_and_response(
-                cache,
-                client.get_records_since(last_modified)?,
-            ),
-        };
-        if last_modified != response.last_modified {
-            dao.write_cached_rs_data(collection.name(), &response);
+        client.sync()?;
+        let response = client.get_records(false);
```
0c0w3, line 109: Just for my own information -- does this new RS client periodically sync on its own? (No, right?) Say we didn't call `sync()` here and instead passed true for `sync_if_empty`. That would mean we would do an initial sync but AFAICT nothing would ever be synced after that, right?

components/suggest/src/rs.rs
```diff
@@ -69,85 +72,58 @@ pub(crate) trait Client {
...
-
-        Ok(response
-            .records
-            .into_iter()
-            .filter_map(|r| Record::new(r, collection).ok())
-            .collect())
     }
 
-    fn download_attachment(&self, record: &Record) -> Result<Vec<u8>> {
+    fn download_attachment(&self, record: Record) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
tiftran: oh i ran into trouble yesterday and this is my rust newbie solution. `get_attachment` expects a `&RemoteSettingRecord` and `.into()` wasn't helping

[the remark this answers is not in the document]
0c0w3: If you add back the `&` in both places in `rs.rs` and then fix the errors in the other files, it should work (it does for me). I do think it's worth fixing so we're not unnecessarily creating clones.

components/suggest/src/rs.rs
bendk: Yes that's right.  I think the system we should be eventually moving towards is the application calls `RemoteSettingsService::sync` on some periodic schedule.  However, that's not happening yet so clients have to call `sync()` themselves.

components/suggest/src/rs.rs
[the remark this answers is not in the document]
tiftran: let me try that again

components/suggest/src/store.rs
```diff
@@ -115,15 +117,16 @@ impl SuggestStoreBuilder {
...
-            inner: SuggestStoreInner::new(data_path, extensions_to_load, client),
-        }))
+        if let Some(rs_service) = &inner.remote_settings_service {
+            let client = SuggestRemoteSettingsClient::new(rs_service)?;
+            Ok(Arc::new(SuggestStore {
+                inner: SuggestStoreInner::new(data_path, extensions_to_load, client),
+            }))
+        } else {
+            Err(Error::RemoteSettings(RemoteSettingsError::Other {
+                reason: "Unable to create client".to_string(),
```
tiftran, line 127: Im unsure what's the best way to handle this

tiftran:
okie doke, i think i've addressed the latest feedback

components/suggest/src/store.rs
0c0w3: Yeah this makes sense to me and it's similar to how we return an error if `data_path` is none, right above. You could use that as an example to rewrite this without an `if` statement:

```rust
        let rs_service = inner.remote_settings_service.clone().ok_or_else(|| {
            Error::RemoteSettings(RemoteSettingsError::Other {
                reason: "Unable to create client".to_string(),
            })
        })?;
        Ok(Arc::new(SuggestStore {
            inner: SuggestStoreInner::new(
                data_path,
                extensions_to_load,
                SuggestRemoteSettingsClient::new(&rs_service)?,
            ),
        }))
```

I'll need to defer to @bendk if all this is kosher or not w/r/t `Arc`. `SuggestRemoteSettingsClient` should probably take an `Arc<RemoteSettingsService>` instead of `&RemoteSettingsService`, right? Like how `SuggestStore` does.

0c0w3: The error message should be something like `"remote_settings_service not specified"` (similar to the `data_path` error message)

components/suggest/src/store.rs
```diff
@@ -82,10 +83,11 @@ impl SuggestStoreBuilder {
 
     pub fn remote_settings_service(
         self: Arc<Self>,
-        _rs_service: Arc<RemoteSettingsService>,
+        rs_service: Arc<RemoteSettingsService>,
     ) -> Arc<Self> {
         // When #6607 lands, this will set the remote settings service.
         // For now, it just exists so we can move consumers over to the new API ahead of time.
```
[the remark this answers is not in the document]
bendk: Yeah, this comment can be removed.  Assuming we get mobile consumers using the new API from #6607 ahead of time, then there's no breakage.

components/suggest/src/rs.rs
```diff
@@ -181,11 +157,23 @@ impl Record {
     }
 }
 
+impl From<Record> for RemoteSettingsRecord {
+    fn from(record: Record) -> Self {
```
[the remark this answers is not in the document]
bendk: The key difference is the `payload` field. `RemoteSettingsRecord` stores a JSON value there, since it doesn't know the specifics of how it's going to be used.  `suggest::rs::Record` stores the deserialized `SuggestRecord` there.  So, if a function inputs `suggest::rs::Record` it can access the fields more conveniently and it doesn't have to deal with deserialization errors.

It's still a good idea and I'm pretty sure there's a way to remove some code duplication.  Maybe `RemoteSettingsRecord` could be generic over its `payload` field.  It's just not going to be a trivial change.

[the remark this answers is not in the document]
tiftran: created a [jira](https://mozilla-hub.atlassian.net/browse/DISCO-3296) in our maintenance epic for this

components/suggest/src/store.rs
```diff
@@ -578,8 +569,7 @@ where
         // For each collection, fetch all records
         for (collection, record_types) in record_types_by_collection {
             breadcrumb!("Ingesting collection {}", collection.name());
-            let records =
-                write_scope.write(|dao| self.settings_client.get_records(collection, dao))?;
+            let records = write_scope.write(|_| self.settings_client.get_records(collection))?;
```
0c0w3, line 572:
```rust
            let records = self.settings_client.get_records(collection)?;
```

No need for the write scope anymore since we don't need to write to the DB when getting records. (Sorry for not catching that before)

components/suggest/src/store.rs
```diff
@@ -852,9 +842,9 @@ where
         let mut context = MetricsContext::default();
         let ingested_records = writer.read(|dao| dao.get_ingested_records()).unwrap();
         let records = writer
-            .write(|dao| {
+            .write(|_| {
                 self.settings_client
-                    .get_records(ingest_record_type.collection(), dao)
+                    .get_records(ingest_record_type.collection())
             })
             .unwrap();
```
0c0w3, line 849: Here too:

```rust
        let records = self
            .settings_client
            .get_records(ingest_record_type.collection())
            .unwrap();
```

0c0w3 (approved):
Thanks Tif!

bendk (approved):
These changes look good to me.   Let's hold this back and not merge until phabricator.services.mozilla.com/D239603 is merged.  I'll be following that one and try to keep you updated on it.

PR #6600, opened 2025-02-19
Move app_context to RemoteSettingsService

bendk:
It seems to me that the app context should belong to `RemoteSettingsService`.  This way all clients share the same context.  I'm not really sure though, so I'm opening this PR to ask you all the question.

  - Note:

### DIFF
--- a/components/remote_settings/src/service.rs
+++ b/components/remote_settings/src/service.rs
@@ -25,6 +25,7 @@ struct RemoteSettingsServiceInner {
     storage_dir: Utf8PathBuf,
     base_url: Url,
     bucket_name: String,
+    app_context: Option<RemoteSettingsContext>,
     /// Weakrefs for all clients that we've created.  Note: this stores the
     /// top-level/public `RemoteSettingsClient` structs rather than `client::RemoteSettingsClient`.
     /// The reason for this is that we return Arcs to the public struct to the foreign code, so we
@@ -50,18 +51,13 @@ impl RemoteSettingsService {
                 storage_dir,
                 base_url,
                 bucket_name,
+                app_context: config.app_context,
                 clients: vec![],
             }),
         })
     }
 
-    /// Create a new Remote Settings client
-    #[cfg(feature = "jexl")]
-    pub fn make_client(
-        &self,
-        collection_name: String,
-        context: Option<RemoteSettingsContext>,
-    ) -> Result<Arc<RemoteSettingsClient>> {
+    pub fn make_client(&self, collection_name: String) -> Result<Arc<RemoteSettingsClient>> {
         let mut inner = self.inner.lock();
         // Allow using in-memory databases for testing of external crates.
         let storage = if inner.storage_dir == ":memory:" {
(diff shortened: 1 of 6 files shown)


components/search/src/selector.rs
```diff
@@ -47,11 +47,10 @@ impl SearchEngineSelector {
     pub fn use_remote_settings_server(
         self: Arc<Self>,
         service: &Arc<RemoteSettingsService>,
-        options: Option<RemoteSettingsContext>,
```
bendk, line 50: Is this being used by consumers yet?  If so this will be a breaking change and I'll have to be more careful about it.

leplatrem:
On Desktop, we don't allow customization of the app context.
I think it is safe to assume that it is shared by all clients, so I'm +1 on this change 👌

components/remote_settings/src/service.rs
```diff
@@ -81,31 +79,6 @@ impl RemoteSettingsService {
         Ok(client)
     }
 
-    #[cfg(not(feature = "jexl"))]
```
leplatrem, line 84: I'm not sure `jexl` has to be an optional feature. But leaving it to your appreciation

leplatrem (approved):
👌 
Related https://bugzilla.mozilla.org/show_bug.cgi?id=1944609

components/search/src/selector.rs
Standard8: I only just landed it on Wednesday, but they are looking at using it already. I don't know if they've landed code for it yet.

components/remote_settings/src/service.rs
Standard8: I think we should probably enable jexl for Android/iOS at least, as they may need it for filtering some collections - even if we don't use it filtering up-front, we might need it later, so this would avoid some work when adding it & having to remember that older clients wouldn't necessarily have support for it.

For desktop, I'm not sure if we're using this or not yet, but it would seem a good idea to get it enabled.

Do I remember correctly that we needed to do a bit of extra work to enable it on desktop - some sort of crate review or something?

Standard8 (commented):
I'm fine with the general idea of this, however do we have a way to update the context if necessary?

I know we don't use it today, but for example, `locale` can change dynamically on desktop. Do we need to allow that here?

components/search/src/selector.rs
bendk: From my POV this should be safe to land.  Searching `useRemoteSettingsServer` on searchfox returns 0 results and I don't see any usage of this method in [SearchEngineSelector.sys.mjs](https://searchfox.org/mozilla-central/source/toolkit/components/search/SearchEngineSelector.sys.mjs).

bendk:
> I'm fine with the general idea of this, however do we have a way to update the context if necessary?
>
> I know we don't use it today, but for example, locale can change dynamically on desktop. Do we need to allow that here?

Good point.  I changed it so the app context is now in `RemoteSettingsConfig2`.  This means the existing `update_config` method can update the app context too.

components/remote_settings/src/service.rs
bendk: I think it's probably just auditing the crates.  I'll check on this and see how much work it is.

Standard8 (approved):
Thank you, I think moving it to RemoteSettingsConfig2 is the right thing to do and as you say, makes it easier for update.

Hence looks good to me. For the JEXL comments, I'm happy for those to be managed in a follow-up.

PR #6610, opened 2025-02-24
Initialization stubs for the Android megazord

bendk:
I started with Jo's work in https://github.com/mozilla/application-services/pull/6607/ and added gradle code so that this actually compiles.

  - Note:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2924,6 +2924,7 @@ dependencies = [
  "suggest",
  "sync_manager",
  "tabs",
+ "uniffi",
  "viaduct",
 ]
 
--- a/build-scripts/component-common.gradle
+++ b/build-scripts/component-common.gradle
@@ -91,7 +91,10 @@ ext.dependsOnTheMegazord = {
 // Shared logic for projects that use UniFFI-generated bindings
 //
 // Make sure to also call dependsOnTheMegazord()
-ext.configureUniFFIBindgen = { crateName ->
+//
+// If present, libraryBuildDir should be a directory provider that contains the megazord library
+// files.
+ext.configureUniFFIBindgen = { crateName, libraryBuildDir = null ->
     // This will store the uniffi-bindgen generated files for our component
     def uniffiOutDir = layout.buildDirectory.dir("generated/uniffi/")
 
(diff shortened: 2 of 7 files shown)


mhammond:
This seems OK but I wonder if it is the correct approach longer term - eg, imagine a world where these components are in libxul for Android, whereas a different set of components are in libxul for desktop. In this scenario there's not really any such thing as a "megazord" which makes sense to own this. In that world I guess we'd end up with a component dedicated to this initialization and which targets one single "configuration". I'm wondering if we should just do that here?

jo:
Can we have a Megazord for Desktop and one for Android? I think that's not too much code duplication, isn't it?

bendk:
> This seems OK but I wonder if it is the correct approach longer term - eg, imagine a world where these components are in libxul for Android, whereas a different set of components are in libxul for desktop. In this scenario there's not really any such thing as a "megazord" which makes sense to own this. In that world I guess we'd end up with a component dedicated to this initialization and which targets one single "configuration". I'm wondering if we should just do that here?

I figured in that world the megazords would be the configuration.  They change from being the crate that builds the dylib to the libxul dependency that bundles the Rust components.  I haven't thought this through deeply at all though, do you forsee any issues here?

mhammond:
> Can we have a Megazord for Desktop and one for Android? I think that's not too much code duplication, isn't it?

We should maybe try and move away from the term "megazord"? In practice, it just means "the library in which the components ship". For Android, `megazords/full` doesn't really have any code at all, in the same way that when these components are in libxul for desktop there's also no (runtime) code involved in that. Thus, there's no real scope for *any* code duplication here - other than possibly this new capability. (This ignores all the build related code though)

> I figured in that world the megazords would be the configuration. They change from being the crate that builds the dylib to the libxul dependency that bundles the Rust components. I haven't thought this through deeply at all though, do you forsee any issues here?

I'm not quite sure tbh. In oak, the build environment is slightly different from how it's done in app-services. Specifically, each crate is built, then `megazords/full` is built into a stand-alone static-lib `libmegazord.a`, then there's a *new* `moz.build` which builds `libmegazord.so`, taking only `libmegazord.a` as input. This is kinda forced on us by the m-c build process and is how `libxul` is built.

So I guess that while this *might* work fine, there also seems a risk that it will not - that we will still want this new "megazord config" code in its own crate so it ends up in `libmegazord.a` - ie, that the process for building `libmegazord.a` might not know how to do this (it currently doesn't compile anything IIUC, just links). But maybe that *new* project could make it work - that *does* have "code", even though in-practice it is just "dummy" code to ensure the correct symbols are pulled in (eg, see [this monstrosity](https://hg.mozilla.org/projects/oak/file/6f6ba580929aaca2fc5003516599e8a73412b5db/services/app-services/megazord_stub.c))

To put it another way, conceptually this is great, I'm just uncertain how great it will actually be due to build requirements I don't fully understand.

TBH, I'm also not too bothered by this, because it just means we need to adjust things for m-c, which is going to be true for many things.

mhammond:
another thing I don't understand and I'm not sure if we experimented with, but why did we choose to have many megazords, rather than a single megazord with features to select the components?

jo:
init component: https://github.com/mozilla/application-services/pull/6618

bendk:
I don't think this is needed anymore since @jo created a separate component for initialization.

PR #6614, opened 2025-02-26
 Bug 1938924 - Add support for search-config-overrides-v2 to the Rust…

mandysGit:
… based search engine selector.

  - Note:

### DIFF
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -4,6 +4,7 @@
 
 //! This module defines the functions for managing the filtering of the configuration.
 
+use crate::configuration_overrides_types::JSONOverridesRecord;
 use crate::environment_matching::matches_user_environment;
 use crate::{
     error::Error, JSONDefaultEnginesRecord, JSONEngineBase, JSONEngineRecord, JSONEngineUrl,
@@ -94,6 +95,16 @@ impl SearchEngineDefinition {
         }
     }
 
+    fn merge_override(&mut self, override_record: &JSONOverridesRecord) {
+        self.partner_code = override_record.partner_code.clone();
+        self.urls.merge(&override_record.urls);
+        self.click_url = Some(override_record.click_url.clone());
+
+        if let Some(telemetry_suffix) = &override_record.telemetry_suffix {
+            self.telemetry_suffix = telemetry_suffix.clone();
+        }
+    }
+
     pub(crate) fn from_configuration_details(
         identifier: &str,
         base: JSONEngineBase,
@@ -111,6 +122,7 @@ impl SearchEngineDefinition {
             partner_code: base.partner_code.unwrap_or_default(),
             telemetry_suffix: String::new(),
             urls: base.urls.into(),
+            click_url: None,
         };
 
         engine_definition.merge_variant(variant);
@@ -132,13 +144,25 @@ pub(crate) trait Filter {
     fn filter_records(
         &self,
         user_environment: &SearchUserEnvironment,
+        overrides: Option<Vec<JSONOverridesRecord>>,
     ) -> Result<FilterRecordsResult, Error>;
+}
+
+fn apply_overrides(engines: &mut [SearchEngineDefinition], overrides: &[JSONOverridesRecord]) {
+    for override_record in overrides {
+        for engine in engines.iter_mut() {
+            if engine.identifier == override_record.identifier {
+                engine.merge_override(override_record);
+            }
+        }
+    }
 }
 
 impl Filter for Vec<RemoteSettingsRecord> {
     fn filter_records(
         &self,
         user_environment: &SearchUserEnvironment,
+        overrides: Option<Vec<JSONOverridesRecord>>,
     ) -> Result<FilterRecordsResult, Error> {
         let mut engines = Vec::new();
         let mut default_engines_record = None;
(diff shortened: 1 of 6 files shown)


mandysGit:
I updated the patch to fix the failing tests. I didn't notice them when I first push this up.

components/search/src/filter.rs
```diff
@@ -191,7 +207,10 @@ impl Filter for Vec<JSONSearchConfigurationRecords> {
         for record in self {
             match record {
                 JSONSearchConfigurationRecords::Engine(engine) => {
-                    let result = maybe_extract_engine_config(user_environment, engine.clone());
+                    let mut result = maybe_extract_engine_config(user_environment, engine.clone());
+                    if let Some(overrides_data) = &overrides {
+                        apply_overrides(&mut result, overrides_data);
+                    }
```
Standard8, line 213: Currently, I think this is being called every time we add a new engine to the result - each time around the loop.

I think it should be called after the loop so that we only loop over the engines once for each overrides record.

components/search/src/filter.rs
```diff
@@ -214,16 +233,30 @@ impl Filter for Vec<JSONSearchConfigurationRecords> {
     }
 }
 
+fn apply_overrides(
```
Standard8, line 236: I think this could possibly be moved to before it is used, as I think that helps readability.

components/search/src/filter.rs
```diff
@@ -94,6 +95,17 @@ impl SearchEngineDefinition {
         }
     }
 
+    fn merge_override(&mut self, override_record: &JSONOverridesRecord) {
```
Standard8, line 98: It'd be nice to add a unit test for this - something fairly simple that checks we merge the parameters correctly (and don't change the telemetry suffix if none is supplied).

Standard8 (changes requested):
Thank you, this is a good start and looks like it will fix some more of the desktop tests 🎉 

A few things to consider that I've added in-line.

components/search/src/selector.rs
```diff
@@ -92,16 +103,21 @@ impl SearchEngineSelector {
                 if records.is_empty() {
                     return Err(Error::SearchConfigNoRecords);
                 }
-                return filter_engine_configuration_impl(user_environment, &records);
+                return filter_engine_configuration_impl(user_environment, &records, None);
```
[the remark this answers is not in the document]
mandysGit: If I'm understanding correctly, and I pass the `search-config-overrides-v2` client to `filter_engine_configuration_impl()` Then I have to convert the overrides record to strings and then deserialize them again to the records we can use, similar to the code below. 
I want to clarify is this what you're suggesting or am I over complicating it and there's a simpler way? 


```rust
    fn filter_records(
        &self,
        user_environment: &SearchUserEnvironment,
        _overrides: Option<Vec<JSONOverridesRecord>>,
    ) -> Result<FilterRecordsResult, Error> {
        let mut engines = Vec::new();
        let mut default_engines_record = None;
        let mut engine_orders_record = None;

        for record in self {
            // TODO: Bug 1947241 - Find a way to avoid having to serialise the records
            // back to strings and then deserilise them into the records that we want.
            let stringified = serde_json::to_string(&record.fields)?;
            match record.fields.get("recordType") {
                Some(val) if *val == "engine" => {
                    let engine_config: Option<JSONEngineRecord> =
                        serde_json::from_str(&stringified)?;
                    if let Some(engine_config) = engine_config {
                        let result =
                            maybe_extract_engine_config(user_environment, Box::new(engine_config));
                        engines.extend(result);
                    }
                }
                Some(val) if *val == "defaultEngines" => {
                    default_engines_record = serde_json::from_str(&stringified)?;
                }
                Some(val) if *val == "engineOrders" => {
                    engine_orders_record = serde_json::from_str(&stringified)?;
                }
                // These cases are acceptable - we expect the potential for new
                // record types/options so that we can be flexible.
                Some(_val) => {}
                None => {}
            }
        }

        Ok(FilterRecordsResult {
            engines,
            default_engines_record,
            engine_orders_record,
        })
    }

```

components/search/src/filter.rs
```diff
@@ -94,6 +95,17 @@ impl SearchEngineDefinition {
...
+        self.partner_code = override_record.partner_code.clone();
+        self.urls.merge(&override_record.urls);
+        if let Some(telemetry_suffix) = &override_record.telemetry_suffix {
+            self.telemetry_suffix = telemetry_suffix.clone();
+        }
+        if self.click_url.is_none() {
+            self.click_url = Some(override_record.click_url.clone());
+        }
```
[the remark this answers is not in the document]
mandysGit: I think I forgot to remove that. I was trying to fix a build error. What I really needed to add was ` Some(override_record.click_url.clone())` and not the `is_none()`

components/search/src/selector.rs
```diff
@@ -1621,7 +1870,15 @@ mod tests {
...
 
-    fn setup_remote_settings_test() -> Arc<SearchEngineSelector> {
+    const APPLY_OVERRIDES: bool = true;
+    const DO_NOT_APPLY_OVERRIDES: bool = false;
+    const RECORDS_MISSING: bool = false;
+    const RECORDS_PRESENT: bool = true;
+
+    fn setup_remote_settings_test(
+        should_apply_overrides: bool,
+        received_records: bool,
```
Standard8, line 1880: I think something like `expect_sync_successful` might be slightly clearer.

Standard8 (approved):
Thank you, this looks good. One comment to consider. AFAIK mobile hasn't landed anything using this yet, so the types.rs change shouldn't need to count as breaking API.

PR #6616, opened 2025-02-27
Expose Search component APIs to iOS

skhamis:

  - Note:

### DIFF
--- a/components/search/Cargo.toml
+++ b/components/search/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 description = "Library for search service related components."
 readme = "README.md"
 license = "MPL-2.0"
+exclude = ["/ios"]
 
 [dependencies]
 env_logger = { version = "0.10.0", default-features = false, optional = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2954,6 +2954,7 @@ dependencies = [
  "push",
  "remote_settings",
  "rust-log-forwarder",
+ "search",
  "suggest",
  "sync15",
  "sync_manager",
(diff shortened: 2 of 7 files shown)


mattreaganmozilla:
Thanks for your help with this @skhamis!

mhammond:
cc @bendk, it looks like you are helping getting this going for Android too

lougeniaC64:
@skhamis @mattreaganmozilla Assuming this doesn't need to be exposed in focus? I'm approving under the assumption that it doesn't.

lougeniaC64 (approved):

components/search/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 description = "Library for search service related components."
 readme = "README.md"
 license = "MPL-2.0"
+exclude = ["/ios"]
```
bendk, line 8: I know you're just following what the other crates are doing here, but I'm wondering why this is needed.  We don't publish our crates to crates.io and if we did, I think it's as likely as not that we would want these directories.

This shouldn't impact this PR, I'm just thinking out loud.  Maybe we could do something in a follow-up.

bendk (approved):
Looks good to me, for what it's worth.

mattreaganmozilla:
> Assuming this doesn't need to be exposed to focus? I'm approving under the assumption that it doesn't.

@lougeniaC64 Thanks for calling this out. I'm still attempting to clarify what we'll be doing for Focus for Search Engine Consolidation (if anything). But right now I don't believe it's required.

Is it problematic for us to expose these also to Focus at some future point?

components/search/Cargo.toml
skhamis: That's a good point! Yeah pretty much decided to stay inline with other crates. I think that's a good idea to simply remove that exclusion for all crates to reduce confusion in the future -- I'll maybe make a follow up bug for discussion on that!

lougeniaC64:
> Is it problematic for us to expose these also to Focus at some future point?

@mattreaganmozilla Not at all! I just wanted to make sure that wasn't neglected. 😊

PR #6638, opened 2025-03-04
Add pre-guid-collision check for bookmarks

skhamis:
and potentially remove them (see comments in the PR)

So now that we're receiving reports of users actually hitting the collision in the `apply_remote_items`. It's still unclear _why_ or _how_ users are running into this.

Link to sentry: https://mozilla.sentry.io/issues/6329806367/?project=6295546&query=is%3Aunresolved%20places-sync&referrer=issue-stream&statsPeriod=7d&stream_index=0

and if we look at our top issue in android bookmarks: https://sql.telemetry.mozilla.org/queries/86634/source?p_Months=1&p_channel=org_mozilla_fenix&p_engine=bookmarks&p_minimum%20error%20count=20#214559

`UNIQUE constraint failed: moz_bookmarks.guid`. Which, thanks to the new reporting, we know is happening exactly here https://github.com/mozilla/application-services/blob/d867ed13261afb5141e8004c75293e35f26aa552/components/places/src/bookmark_sync/engine.rs#L574

which means we're somehow getting items inserted in `itemsToApply` where the `mergedGuid` already exists but not the `id` (otherwise the ON CONFLICT would hit). So we add another pre-check to identify these bookmarks and report them with more information (Just incase we can catch any weirdness). I then also go a step further and **delete them** since it's blocking the user from syncing and given that its low user count but high events. The issue doesn't seem to be widespread?

### DIFF
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -364,18 +364,7 @@ fn update_local_items_in_places(
 
     log::debug!("Changing GUIDs");
     scope.err_if_interrupted()?;
-    let result = db.execute_batch("DELETE FROM changeGuidOps");
-    // Due to bug 1935797, guid collisions are happening in some step of bookmarks
-    // but not clear if it's the "fixup" step, so we catch and log to sentry
-    if let Err(rusqlite::Error::SqliteFailure(e, _)) = &result {
-        if e.code == ErrorCode::ConstraintViolation {
-            error_support::report_error!(
-                "places-sync-bookmarks-guid-collision",
-                "changeGuidOps ran into guid collision violation"
-            );
-        }
-    }
-    result?;
+    db.execute_batch("DELETE FROM changeGuidOps")?;
 
     log::debug!("Applying remote items");
     apply_remote_items(db, scope, now)?;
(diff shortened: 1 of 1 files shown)


components/places/src/bookmark_sync/engine.rs
```diff
@@ -501,35 +490,70 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp)
...
-    ).ok(); // We ignore the error because it's expected, users should not have entries with the same id that have different types
+
+        // Get list of colliding GUIDs
+        let colliding_guids: Vec<&str> = potential_collisions
+            .iter()
+            .map(|(_, guid, _, _, _)| guid.as_str())
+            .collect();
+
+        // Remove the colliding items from itemsToApply
+        if !colliding_guids.is_empty() {
```
skhamis, line 542: So this might be the only contentious part of this PR. The difference between desktop and mobile code is desktop will actually _skip_ any issues with bookmarks while mobile will halt the syncing completely. IMO, we should delete the items that have a weird scenario here (which is, the same `mergedGuid` but different `moz_bookmarks.id` somehow).

I have the reporting of this above so that we know exactly how many we're deleting and we can potentially revert but I think this will potentially fix up users who got into this issue and we should see a downtrend in this error

mhammond:
I'd like to better understand the theory here, and see tests which demonstrate it. I doubt the theory is "the guid generator generates dupe IDs", but not sure what it is. If the theory is desktop is broken and causing this I think a patch should be targeting desktop (possibly just emitting telemetry in the first instance).

Let's consider, say, bookmark resurrection - where we delete a bookmark, sync it, but the remote device sees a conflict, so resurrects it.

IOW, this seems like a band-aid, and without understanding the underlying problem, a band-aid that may end up actually causing more problems than it solves?

Happy to get in zoom to chat about this if you want (but sadly I've still got no theories for the underlying problem either)

skhamis:
> I'd like to better understand the theory here, and see tests which demonstrate it. I doubt the theory is "the guid generator generates dupe IDs", but not sure what it is. If the theory is desktop is broken and causing this I think a patch should be targeting desktop (possibly just emitting telemetry in the first instance).
> 
> Let's consider, say, bookmark resurrection - where we delete a bookmark, sync it, but the remote device sees a conflict, so resurrects it.
> 
> IOW, this seems like a band-aid, and without understanding the underlying problem, a band-aid that may end up actually causing more problems than it solves?
> 
> Happy to get in zoom to chat about this if you want (but sadly I've still got no theories for the underlying problem either)

Yeah definitely a band-aid, which _may_ be all we need as this seems to be a low-user high-error type of bug. However I understand we should get more logging here, so the PR is strictly adding more breadcrumbs around the issue in hopes that we can see a little more in how this is actually happening. The `test_db_level_unique_guid_violation` shows at least _one_ scenario that the user can run into a guid collision but not clear on the how.

mhammond (approved):

PR #6640, opened 2025-03-05
Update Kotlin to version 2.1.10 [ci full]

rvandermeulen:
We're almost ready to land this in mozilla-central, so let's do the same for A-S when ready.


mhammond (approved):
Approving now so you can land it when ready

PR #6641, opened 2025-03-05
Create the storage dir in the remote settings client

bendk:
The iOS devs have noticed errors because this directory doesn't exist.

Removed code in `cli-support` to auto-create the directory.  In general, our components should be doing that.

  - Note:

### DIFF
--- a/components/remote_settings/src/service.rs
+++ b/components/remote_settings/src/service.rs
@@ -40,6 +40,9 @@ impl RemoteSettingsService {
     /// This is typically done early in the application-startup process
     pub fn new(storage_dir: String, config: RemoteSettingsConfig2) -> Result<Self> {
         let storage_dir = storage_dir.into();
+        if storage_dir != ":memory:" {
+            std::fs::create_dir(&storage_dir)?;
+        }
         let base_url = config
             .server
             .unwrap_or(RemoteSettingsServer::Prod)
--- a/examples/cli-support/src/lib.rs
+++ b/examples/cli-support/src/lib.rs
@@ -44,12 +44,11 @@ pub fn cli_data_path(filename: &str) -> String {
 
 fn data_path(relative_path: Option<&str>) -> PathBuf {
     let dir = workspace_root_dir().join(".cli-data");
-    let dir = match relative_path {
+
+    match relative_path {
         None => dir,
         Some(relative_path) => dir.join(relative_path),
-    };
-    std::fs::create_dir_all(&dir).expect("Error creating dir {dir:?}");
-    dir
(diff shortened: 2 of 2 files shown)


mhammond:
Interested to know why you think the component should create the directory which is passed in from foreign code - I can see why we'd create children but not the top-level. IIUC, no other component does this.

mhammond:
(eg, what if that dir needed app-specific permissions?)

bendk:
> Interested to know why you think the component should create the directory which is passed in from foreign code - I can see why we'd create children but not the top-level. IIUC, no other component does this.

That's a good point, I haven't thought this through all the way.   The difference is that remote settings inputs a directory in it's constructor, so I figured it should create that directory.  I'm not sure if that's right though.

I updated the commit to use `create_dir` rather than `create_dir_all`.  I definitely don't think we should be creating any parent directories.

mhammond:
This would also be a 1-line fix in consumers, right? I'm still not sure why remote settings is special (as in "different" from any other component). Don't *all* components that use storage take a directory as input at creation time?

bendk:
> This would also be a 1-line fix in consumers, right? I'm still not sure why remote settings is special (as in "different" from any other component). Don't all components that use storage take a directory as input at creation time?

Other components input a file path AFAIK.   Remote settings inputs a directory, then creates multiple SQLite Dbs inside that directory.  So there's definitely a difference, but I'm not sure if it's significant or not.  You're right that it's just a 1-line fix for consumers so I'd be fine just closing this one.

mhammond:
TBH I don't care *that* much, although I do think all components should be consistent (even if a path is specified it could makes sense to ensure the directory exists). As mentioned above, apps might have special requirements for how this directory is created (imagine, eg, some attribute was required to make this able to be backed up or something), and if that's true, then this change could actually cause more harm than good - the app behaviour would be different depending on whether the app already existed or not, which would be very difficult to track down.

All that said though, I think my biggest reaction was to:

> The iOS devs have noticed errors because this directory doesn't exist.

And wondering why they reached out instead of just ensuring the path *they* ask us to create the DBs in exists when it's probably 1 line directly next to the code that passes it to us 🤷

bendk:
> And wondering why they reached out instead of just ensuring the path they ask us to create the DBs in exists when it's probably 1 line directly next to the code that passes it to us 🤷

Their first reaction was to fix it in their codebase.  I felt like we could make an improvement on our side as well.  However, it seems like it's not clear if this is better or worse so let's not try to land it.  It feels better to err on the side of not doing anything and make changes when they're really needed.

PR #6645, opened 2025-03-12
Handle GUID collisions for users with orphaned bookmarks

skhamis:
Landed extra logging and sentry reports for these GUID collisions to understand what’s going on before the constraint hits
https://github.com/mozilla/application-services/pull/6638
https://github.com/mozilla/application-services/pull/6589

Based on data collected from the 2nd PR. There is a correlation between having orphaned bookmarks (bookmarks with no parent id, which should be “impossible” in most scenarios) and hitting the GUID collisions.

Take for example this sentry report https://mozilla.sentry.io/issues/6381267805/events/2bb60d08f5524309b83db8cdf096d685/?project=6295546
`Hit a constraint violation Err(SqliteFailure(Error { code: ConstraintViolation, extended_code: 2067 }, Some("UNIQUE constraint failed: moz_bookmarks.guid")))`
Which is happening here: https://github.com/mozilla/application-services/blob/main/components/places/src/bookmark_sync/engine.rs#L572-L598
Breadcrumbs show:
```
places::bookmark_sync::engine[559]: places-sync-bookmarks-orphaned: found local orphans before upsert 1
places::bookmark_sync::engine[526]: Found GUID collision: ia_localId=None, ia_mergedGuid={redacted-guid-1},ia_remoteGuid={redacted-guid-1}, mb_id=8, mb_guid={redacted-guid-1}
```
**Which means the ON CONFLICT is not triggering because it’s only on ID conflict and not GUID**
Which also proves dogear doesn’t think there is a bookmark with that GUID locally thus not running into the conflict, thus running into the current error we’re seeing. However dogear is _really_ good at ensuring we should never run into these situations. Given that this is a low user, high event type of issue -- we can reliably assume it's not a symptom of a wider issue. 

I decided to go with this extra `ON CONFLICT(guid)` approach, for multiple reasons:
- these users were not continuing with their bookmarks sync because there was no localid found and it failed to insert due to the guid already being in the DB. This will only update the existing bookmark -- thus no dupes could be created
- It's the least invasive, since we are only updating when it's conflicting, we're not running this on any user who has a normal bookmark state -- this will only run for users who hit the conflict. No potential for loss of data for existing users
- This will allow it to leave it up to dogear to fix any downstream issues with reconciliation
  - Which by local testing, Mark and I found dogear eventually fixes things up

### DIFF
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -591,6 +591,19 @@ fn apply_remote_items(db: &PlacesDb, scope: &SqlInterruptScope, now: Timestamp) 
            dateAdded = excluded.dateAdded,
            lastModified = excluded.lastModified,
            fk = excluded.fk,
+           syncStatus = {sync_status}
+       /* Due to bug 1935797, we found scenarios where users had bookmarks with GUIDs that matched
+        * incoming records BUT for one reason or another dogear doesn't believe it exists locally
+        * This handles the case where we try to insert a new bookmark with a GUID that already exists,
+        * updating the existing record instead of failing with a constraint violation.
+        * Usually the above conflict will catch most of these scenarios and there's no issue of
+        * any dupes being added here since users that hit this before would've just failed the bookmark sync
+        */
+        ON CONFLICT(guid) DO UPDATE SET
+           title = excluded.title,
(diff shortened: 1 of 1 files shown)


mhammond (approved):
thanks Sammy! fingers-crossed :)

PR #6649, opened 2025-03-18
Update to Glean v64.0.0 [ci full]

badboy:
As before this needs to land in m-c together with https://bugzilla.mozilla.org/show_bug.cgi?id=1953191 (and  #6615 should also get merged, and https://bugzilla.mozilla.org/show_bug.cgi?id=1950665 land as part of that whole stack)


rvandermeulen:
Please bump the JNA version to 5.17.0 as well.

badboy:
> Please bump the JNA version to 5.17.0 as well.

Oh right. I wanted to include that!

bendk (approved):
Looking good to me and I think the other PRs are more-or-less ready to go.   Should we merge this one?

PR #6656, opened 2025-03-21
Bug 1953667: Make location sign skippable

dadaa:

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -1792,8 +1792,20 @@ pub(crate) mod tests {
             ),],
         );
         assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_delivery")),
+            vec![ramen_suggestion(
+                "ramen invalid_delivery",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_delivery"
+            )
+            .has_location_sign(false),],
+        );
+        assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_delivery in tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen invalid_delivery in tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_delivery+in+tokyo"
+            )
+            .has_location_sign(false),],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen in tokyo")),
@@ -1811,7 +1823,11 @@ pub(crate) mod tests {
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_in tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen invalid_in tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_in+tokyo"
+            )
+            .has_location_sign(false),],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen in San Francisco")),
@@ -1831,21 +1847,22 @@ pub(crate) mod tests {
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near by")),
             vec![ramen_suggestion(
                 "ramen near by",
-                "https://www.yelp.com/search?find_desc=ramen+near+by"
-            )
-            .has_location_sign(false),],
+                "https://www.yelp.com/search?find_desc=ramen"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near me")),
             vec![ramen_suggestion(
                 "ramen near me",
-                "https://www.yelp.com/search?find_desc=ramen+near+me"
-            )
-            .has_location_sign(false),],
+                "https://www.yelp.com/search?find_desc=ramen"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near by tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen near by tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=tokyo"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen")),
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -30,6 +30,12 @@ impl ToSql for Modifier {
     }
 }
 
+#[derive(Eq, PartialEq)]
+enum FindFrom {
+    First,
+    Last,
+}
+
 /// This module assumes like following query.
 /// "Yelp-modifier? Pre-modifier? Subject Post-modifier? (Location-modifier | Location-sign Location?)? Yelp-modifier?"
 /// For example, the query below is valid.
@@ -46,6 +52,11 @@ const MAX_QUERY_LENGTH: usize = 150;
 /// The max number of words consisting the modifier. To improve the SQL performance by matching with
 /// "keyword=:modifier" (please see is_modifier()), define this how many words we should check.
 const MAX_MODIFIER_WORDS_NUMBER: usize = 2;
+
+/// The max number of words consisting the location sign. To improve the SQL performance by matching
+/// with "keyword=:modifier" (please see is_location_sign()), define this how many words we should
+/// check.
+const MAX_LOCATION_SIGN_WORDS_NUMBER: usize = 2;
 
 /// At least this many characters must be typed for a subject to be matched.
 const SUBJECT_PREFIX_MATCH_THRESHOLD: usize = 2;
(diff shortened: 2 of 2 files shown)


components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,187 @@ impl<'a> SuggestDao<'a> {
...
+    /// )
+    fn find_subject(&self, query: &str) -> Result<Option<(String, String)>> {
+        if query.is_empty() {
+            return Ok(None);
+        }
+
+        if let Ok(keyword_lowercase) = self.conn.query_row_and_then_cachable(
+            "SELECT keyword
+             FROM yelp_subjects
+             WHERE :query || ' ' LIKE keyword || ' %'
```
0c0w3, line 270: Can you do `:query BETWEEN keyword AND keyword || x'FFFF'`? IIRC that's more efficient than `LIKE`.

Here you're checking if `keyword` is a substring of `query`, right? And below it's the opposite?

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,187 @@ impl<'a> SuggestDao<'a> {
...
+        };
+
+        if query.len() < SUBJECT_PREFIX_MATCH_THRESHOLD {
+            return Ok(None);
+        }
+
+        if let Ok(keyword_lowercase) = self.conn.query_row_and_then_cachable(
+            "SELECT keyword
+             FROM yelp_subjects
+             WHERE keyword LIKE :query || '%'
```
0c0w3, line 290: Similar here, `keyword BETWEEN :query AND :query || x'FFFF'`

0c0w3 (commented):
Thanks Daisuke, sorry for the slow review. Were you looking for a full review right now, or just some general feedback? Overall this approach looks good to me, and it's pretty much what we talked about on Slack. I have some initial comments inline. I haven't yet gone over everything with a fine-toothed comb.

dadaa:
Thank you very much, Drew!
Ya, if this approach works for you, please do full review!
And, I will address your review comments, of course.

components/suggest/src/yelp.rs
dadaa: Yes, thanks!

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,187 @@ impl<'a> SuggestDao<'a> {
...
-                location_sign: None,
-                location: None,
-                need_location: false,
-                icon,
-                icon_mimetype,
-                score,
-            };
-            return Ok(vec![builder.into()]);
-        }
+        let mut query_string = query.keyword.trim();
```
dadaa, line 156: Most of the functions that parse queries split with white space and process each one. It might be a waste if we split every time. So I want to try to split first, then parse it in each function.

0c0w3:
OK, I'll do a full review. It looks like there are clippy failures, so would you mind running it and then updating your PR?

```
cargo clippy --all-targets --manifest-path components/suggest/Cargo.toml --all-features -- -D warnings
```

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+                true,
+                keyword.split_whitespace().count(),
+            )));
+        };
+
+        if query_string.len() < SUBJECT_PREFIX_MATCH_THRESHOLD {
+            return Ok(None);
+        }
+
+        if let Ok(keyword_lowercase) = self.conn.query_row_and_then_cachable(
```
0c0w3, line 308: Same here about a comment and using `?`

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+             WHERE keyword BETWEEN :query AND :query || x'FFFF'
+             ORDER BY LENGTH(keyword) ASC, keyword ASC
+             LIMIT 1",
+            named_params! {
+                ":query": query_string.to_lowercase(),
+            },
+            |row| row.get::<_, String>(0),
+            true,
+        ) {
+            let keyword = format!(
```
0c0w3, line 320: Please add a comment here about why we're using `format!`, i.e. we want to preserve the query as the user typed it including its case.

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+                "{}{}",
+                query_string,
+                &keyword_lowercase[query_string.len()..]
+            );
+            return Ok(Some((
+                keyword.to_string(),
```
0c0w3, line 326: `format!` returns a `String` so `to_string()` shouldn't be necessary.

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+    ///          Maximum number is MAX_LOCATION_SIGN_WORDS_NUMBER.
+    /// )
+    fn find_location_sign(&self, query_words: &[&str]) -> Result<Option<(String, usize)>> {
+        if query_words.is_empty() {
+            return Ok(None);
+        }
+
+        for n in (1..=std::cmp::min(MAX_LOCATION_SIGN_WORDS_NUMBER, query_words.len())).rev() {
+            let mut candidate_chunks = query_words.chunks(n);
+            let candidate = candidate_chunks.next().unwrap_or(&[""]).join(" ");
```
0c0w3, line 349: Same style nit here as above re: `unwrap_or_default()`

components/suggest/src/store.rs
```diff
@@ -1333,8 +1333,12 @@ mod tests {
             ),],
         );
         assert_eq!(
-            store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_delivery in tokyo")),
```
0c0w3, line 1336: Instead of replacing this query, could you keep it and then add the new one?

0c0w3 (changes requested):

dadaa:
Thank you very much for your kind review!
I will address all of your comments.

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
-            subject_exact_match,
-            pre_modifier,
-            post_modifier,
-            location_sign,
+            subject: &subject.0,
+            subject_exact_match: subject.1,
+            pre_modifier: match &pre_modifier_tuple {
+                Some((words, _)) => Some(words.to_string()),
+                _ => None,
+            },
```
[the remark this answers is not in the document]
dadaa: Oh, I see.

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+        if query_words.is_empty() {
+            return Ok(None);
+        }
+
+        for n in (1..=std::cmp::min(MAX_MODIFIER_WORDS_NUMBER, query_words.len())).rev() {
+            let mut candidate_chunks: Box<dyn Iterator<Item = &[&str]>> = match find_from {
+                FindFrom::First => Box::new(query_words.chunks(n)),
+                _ => Box::new(query_words.rchunks(n)),
+            };
+            let candidate = candidate_chunks.next().unwrap_or(&[""]).join(" ");
```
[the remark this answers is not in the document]
dadaa: `map` is interesting.

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,219 @@ impl<'a> SuggestDao<'a> {
...
+    ///   usize: Number of words in query_words that match the keyword.
+    /// )
+    fn find_subject(&self, query_words: &[&str]) -> Result<Option<(String, bool, usize)>> {
+        if query_words.is_empty() {
+            return Ok(None);
+        }
+
+        let query_string = query_words.join(" ");
+
+        if let Ok(keyword_lowercase) = self.conn.query_row_and_then_cachable(
```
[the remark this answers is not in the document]
dadaa: If there is no record for the SQL, `query_row_and_then_cachable()` throws an error `QueryReturnedNoRows`.
https://searchfox.org/mozilla-central/rev/a920f0f657f8f2d81ef53c581433942059863dd5/third_party/rust/sql-support/src/conn_ext.rs#124
So, I use Ok() here. I will investiage another better way, but if you know idea, please let me know!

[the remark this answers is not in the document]
0c0w3: Oh right. I've hit that too, and in cases like this I use a method that does not return an error, like [try_query_one](https://github.com/mozilla/application-services/blob/cf164f976d42ebac5ba0ae330b6f3b3168e6a73c/components/support/sql/src/conn_ext.rs#L85-L87) or [try_query_row](https://github.com/mozilla/application-services/blob/cf164f976d42ebac5ba0ae330b6f3b3168e6a73c/components/support/sql/src/conn_ext.rs#L201-L204). That way, if there actually is an error (and not just no rows), then it will be propagated up the stack and you'll see it.

There are two places where `Connection` methods are defined and it's been really helpful for me to refer to them when writing SQL queries:

* [rusqlite](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html)
* [ConnExt](https://github.com/mozilla/application-services/blob/main/components/support/sql/src/conn_ext.rs) in application-services, which defines a bunch of helpers on `Connection` like those two methods I mentioned above

0c0w3:
Were you able to run clippy? It looks like your latest commit still has clippy errors, and it should have caught some of the things I mentioned, including using `Option::map`.

dadaa:
> Were you able to run clippy? It looks like your latest commit still has clippy errors

Yes, there is no failure on local.
And, the error seems happen from another reason??
https://app.circleci.com/pipelines/github/mozilla/application-services/40479/workflows/b2320ff5-b85e-42d1-8be3-d1d7c0eedb94/jobs/214438?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

> and it should have caught some of the things I mentioned, including using `Option::map`.
I'm sorry, I did not update yet including it.

components/suggest/src/yelp.rs
dadaa: If no to_string(), the compiler said:
```
move occurs because `keyword` has type `std::string::String`, which does not implement the `Copy` trait
```
Said also:
```
help: consider cloning the value if the performance cost is acceptable
```

I do something wrong??

0c0w3: I think the move error is because you use `keyword` again in the third element of the returned tuple. By removing the `to_string()` and returning `keyword` directly, you're moving it, and after that you can't access it again. I bet the error message also pointed out "value borrowed here after move" and "value moved here".

If instead you save the count before moving `keyword`, that should work I think:

```rust
            let keyword = format!(
                "{}{}",
                query_string,
                &keyword_lowercase[query_string.len()..]
            );
            let word_count = keyword.split_whitespace().count();
            return Ok(Some((
                keyword,
                false,
                word_count,
            )));
```

components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,211 @@ impl<'a> SuggestDao<'a> {
...
+        }
 
-        if let (Some(_), false) = (&location, need_location) {
-            // The location sign does not need the specific location, but user is setting something.
+        let subject_tuple = self.find_subject(query_words)?;
+        if let Some((_, _, n)) = subject_tuple {
+            query_words = &query_words[n..];
+        } else {
             return Ok(vec![]);
         }
```
0c0w3, line 175: You can use [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) to avoid having to unwrap `subject_tuple` below (on line 201):

```suggestion
        if let Some(subject_tuple) = self.find_subject(query_words)? else {
            return Ok(vec![]);
        };
        // Now you can use subject_tuple without unwrapping it
        query_words = &query_words[subject_tuple.2..]
```

0c0w3 (approved):
Looks good, thanks Daisuke!

0c0w3:
> Yes, there is no failure on local. And, the error seems happen from another reason?? https://app.circleci.com/pipelines/github/mozilla/application-services/40479/workflows/b2320ff5-b85e-42d1-8be3-d1d7c0eedb94/jobs/214438?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

You're right, these errors look entirely unrelated to your PR. I guess we can ignore them. I've seen unrelated errors in my own PRs too.

components/suggest/src/yelp.rs
dadaa: Oh, nice, it is more readable. Thank you so much!

PR #6657, opened 2025-03-21
fix multithreaded NSS initialization

jo:

  - Note:

### DIFF
--- a/components/support/rc_crypto/nss/src/util.rs
+++ b/components/support/rc_crypto/nss/src/util.rs
@@ -3,83 +3,131 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::error::*;
-use crate::ErrorKind::NSSUninitialized;
 use nss_sys::*;
-use std::{ffi::CString, os::raw::c_char, sync::Once};
+use std::{ffi::CString, os::raw::c_char};
+
+use std::path::PathBuf;
+use std::sync::OnceLock;
 
 #[cfg(feature = "keydb")]
 use crate::pk11::slot;
-#[cfg(feature = "keydb")]
-use once_cell::sync::OnceCell;
-#[cfg(feature = "keydb")]
-use std::{fs, path::Path};
 
 // This is the NSS version that this crate is claiming to be compatible with.
 // We check it at runtime using `NSS_VersionCheck`.
 pub const COMPATIBLE_NSS_VERSION: &str = "3.26";
 
-static NSS_INIT: Once = Once::new();
-#[cfg(feature = "keydb")]
-static NSS_PROFILE_PATH: OnceCell<String> = OnceCell::new();
-
-// Check whether either thr Rust-internal NSS initialization has run, or NSS has been initialized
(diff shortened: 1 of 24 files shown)


components/support/rc_crypto/nss/src/util.rs
```diff
@@ -3,184 +3,165 @@
...
-
-// Check whether either thr Rust-internal NSS initialization has run, or NSS has been initialized
-// in the outside.
-#[cfg(not(feature = "keydb"))]
-fn is_nss_initialized() -> bool {
-    NSS_INIT.is_completed() || unsafe { NSS_IsInitialized() == PR_TRUE }
+pub fn expect_nss_initialized() -> Result<()> {
+    INITIALIZED
+        .get()
+        .map(|_| {})
```
bendk, line 22: nit: `.map(|_| ())` is more idiomatic IMO.

components/support/rc_crypto/nss/src/util.rs
```diff
@@ -3,184 +3,165 @@
...
+    #[cfg(feature = "keydb")]
+    Db,
 }
 
-pub fn expect_nss_initialized() -> Result<()> {
-    if is_nss_initialized() {
-        Ok(())
-    } else {
-        Err(NSSUninitialized.into())
+static INITIALIZED: OnceLock<Result<NssLoaded>> = OnceLock::new();
```
bendk, line 35: I kind of think this should be `OnceLock<NssLoaded>`, without the `Result`.

This means `init_once` would just return `NssLoaded` and panic on failure, like the previous code did.  There's nothing we can realistically do to recover from that.  Also, it means less noise in the PR, since you don't need to change all the call sites.

bendk (changes requested):
I mostly focused on the high-level flow here and just took a quick look at the refactoring of `ensure_nss_initialized()` -> `init_once()`.  It seemed right to me and I figure the tests will pick up any issues  that we missed.

Overall it looks good, my only suggestion is that this could maybe be easier if you paniced instead of returning a Result.  What do you think about that?  I could easily be convinced to go with your version.

components/support/rc_crypto/nss/src/util.rs
jo: yes!

I'd say that then also `expect_nss_initialized` should become `assert_nss_initialized` and panic the same way, wdyt?

mhammond: I think `expect_foo` implies a panic on failure - ie, `expect_foo` and `assert_foo` seem almost synonyms?

components/support/rc_crypto/nss/src/util.rs
jo: thanks!

components/support/rc_crypto/nss/src/util.rs
jo: ok I see (already renamed ¯\_(ツ)_/¯)

bendk (approved):
The updates look good to me!

PR #6662, opened 2025-03-24
Move to new RS api

tiftran:

  - Note:


bendk (commented):
Looking good overall, I left a few suggestions mostly just to try to reduce the amount of changes needed.  The biggest question in my mind is how easy it is for consumers to pass the `RemoteSettingsService` that they're passing to the suggest/relavency constructors.  In theory it should be pretty easy, but I know consumers often want to initialize nimbus very early on in the process which could complicate things.

components/nimbus/src/schema.rs
```diff
@@ -102,39 +103,25 @@ impl ExperimentMetadata for Experiment {
...
+                    Ok(exp) => res.push(exp),
+                    Err(e) => {
+                        log::trace!("Malformed experiment data: {:#?}", record.fields);
+                        log::warn!(
+                            "Malformed experiment found! Experiment {},  Error: {}",
+                            record.id,
+                            e
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
tiftran: todo: tif this is about the array, ref 122

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -68,7 +67,7 @@ open class Nimbus(
     override val prefs: SharedPreferences? = null,
     appInfo: NimbusAppInfo,
     coenrollingFeatureIds: List<String>,
-    server: NimbusServerSettings?,
+    remoteSettingsService: RemoteSettingsService,
```
jeddai, line 70: This will require an associated change in the constructor use here https://github.com/mozilla/application-services/blob/b56ecd38f1e76c1fc36f2fe94ac9a1035b68bf87/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/NimbusBuilder.kt#L221-L232
As well as in a number of tests

components/nimbus/ios/Nimbus/NimbusCreate.swift
```diff
@@ -77,34 +77,31 @@ public extension Nimbus {
...
         coenrollingFeatureIds: [String] = [],
         dbPath: String,
         resourceBundles: [Bundle] = [Bundle.main],
         enabled: Bool = true,
         userDefaults: UserDefaults? = nil,
         errorReporter: @escaping NimbusErrorReporter = defaultErrorReporter,
-        recordedContext: RecordedContext? = nil
+        recordedContext: RecordedContext? = nil,
+        remoteSettingsService: RemoteSettingsService? = nil,
+        collectionName: String? = nil
```
jeddai, line 89: This will require an associated change here https://github.com/mozilla/application-services/blob/b56ecd38f1e76c1fc36f2fe94ac9a1035b68bf87/components/nimbus/ios/Nimbus/NimbusBuilder.swift#L262-L271

components/nimbus/src/nimbus.udl
```diff
@@ -147,8 +150,9 @@ interface NimbusClient {
         RecordedContext? recorded_context,
         sequence<string> coenrolling_feature_ids,
         string dbpath,
-        RemoteSettingsConfig? remote_settings_config,
-        MetricsHandler metrics_handler
+        MetricsHandler metrics_handler,
+        string? collection_name,
```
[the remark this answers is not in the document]
jeddai: My suggestion would be that it lives in the client code, actually.

components/nimbus/src/stateful/evaluator.rs
```diff
@@ -3,12 +3,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::stateful::persistence::{Database, StoreId};
+use crate::AppContext;
```
jeddai, line 6: Was this change necessary?

jeddai (commented):
I've done a cursory review, and this is a great start! There's quite a bit more that will need to be added for the Kotlin/Swift side of things. If you'd like any further guidance on that I'm more than happy to help — you're welcome to throw something on my calendar or I can provide answers to any questions you have here.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
bendk: What's you're feeling on how hard it's going to be to get a `RemoteSettingsService` to pass to the constructor on the firefox-android side?

If I traced through the code correctly, I think it just means moving the [RemoteSettingsService val](https://searchfox.org/mozilla-central/rev/601256c3cc6f397b018995810fd3f586570f50ee/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/Components.kt#268`) above the [Nimbus components val](https://searchfox.org/mozilla-central/rev/601256c3cc6f397b018995810fd3f586570f50ee/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/Components.kt#178).  That seems easy enough, but I'm slightly worried that there are special requirements for Nimbus startup that could cause problems.

Similarly, do you foresee any issues on focus or ios?

components/nimbus/src/stateful/client/http_client.rs
```diff
@@ -13,18 +13,33 @@
...
-impl SettingsClient for RemoteSettings {
+impl SettingsClient for Arc<RemoteSettingsClient> {
     fn get_experiments_metadata(&self) -> Result<String> {
         unimplemented!();
     }
 
     fn fetch_experiments(&self) -> Result<Vec<Experiment>> {
-        let resp = self.get_records_raw()?;
-        parse_experiments(&resp.text())
+        //let resp = self.get_records_raw()?;
```
jeddai: Let's remove this comment if it's no longer necessary please

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
jeddai: Hi I'm sorry — I don't _think_ there should be any issues, no. As long as everything is implemented in such a way that it doesn't result in breaking changes, it should be fine!

I'd recommend trying to build iOS & Android with these changes and seeing if there's any issues.

jeddai (commented):
I think this is in a good spot such that you should try building iOS & Android using this iteration of the repo!

components/nimbus/src/stateful/nimbus_client.rs
```diff
@@ -105,11 +105,12 @@ impl NimbusClient {
         recorded_context: Option<Arc<dyn RecordedContext>>,
         coenrolling_feature_ids: Vec<String>,
         db_path: P,
-        config: Option<RemoteSettingsConfig>,
         metrics_handler: Box<dyn MetricsHandler>,
         gecko_pref_handler: Option<Box<dyn GeckoPrefHandler>>,
+        collection_name: Option<String>,
+        remote_settings_service: Option<Arc<RemoteSettingsService>>,
```
bendk, line 112: You can use ` #[allow(clippy::too_many_arguments)] ` to get rid of the clippy error.

...but maybe you could fix it by getting rid of the `collection_name` argument.  I forget if we decided that it was needed or not.

bendk (changes requested):
This looks pretty good to me, I think the only issues left are the leftover comment that Charlie flagged and the clippy errors.

components/nimbus/src/stateful/client/mod.rs
```diff
@@ -5,39 +5,40 @@
...
 use null_client::NullClient;
-use remote_settings::RemoteSettings;
-use remote_settings::RemoteSettingsConfig;
+use remote_settings::RemoteSettingsService;
+use url::Url;
 
 pub(crate) fn create_client(
-    config: Option<RemoteSettingsConfig>,
+    collection_name: Option<String>,
+    rs_service: Option<Arc<RemoteSettingsService>>,
```
bendk, line 18: I realize now that `collection_name` will be ignored when `rs_service` is None.  Maybe a cleaner API would be to put both in a struct Kotlin's NimbusServerSettings.

That said, I don't think it's a big deal.  We can merge this as-is and maybe do a follow-up or maybe just leave it for a while.  I think a comment would be good though.

components/nimbus/src/nimbus.udl
```diff
@@ -184,9 +187,10 @@ interface NimbusClient {
         RecordedContext? recorded_context,
         sequence<string> coenrolling_feature_ids,
         string dbpath,
-        RemoteSettingsConfig? remote_settings_config,
         MetricsHandler metrics_handler,
-        GeckoPrefHandler? gecko_pref_handler
+        GeckoPrefHandler? gecko_pref_handler,
+        string? collection_name,
```
freshstrangemusic: nit: swap RemoteSettingsSerivce and collection name

components/nimbus/src/stateful/nimbus_client.rs
freshstrangemusic: We still need the collection_name because that comes from the embedding application. It is differs depending on application and whether or not the app is using a preview collection.

components/nimbus/src/stateful/client/mod.rs
```diff
@@ -5,39 +5,38 @@
...
```
freshstrangemusic: nit: Swap these. The collection only makes sense if the service is present and IMO dependant arguments should go left to right.

components/nimbus/src/stateful/client/mod.rs
```diff
@@ -5,39 +5,38 @@
...
+                    Box::new(FileSystemClient::new(path)?)
+                }
+                _ => Box::new(rs_service.make_client(collection_name)),
             }
         }
-        // If no server is provided, then we still want Nimbus to work, but serving
-        // an empty list of experiments.
-        None => Box::new(NullClient::new()),
+        (Some(_), None) => return Err(NimbusError::InternalError("collection name required")),
+        (None, _) => Box::new(NullClient::new()),
```
freshstrangemusic, line 39: i was not aware that nimbus was doing this and it makes me sad (not for you to fix)

components/nimbus/tests/common/mod.rs
```diff
@@ -57,35 +58,44 @@ fn new_test_client_internal(
...
+    };
+
+    let config = RemoteSettingsConfig2 {
+        server: Some(RemoteSettingsServer::Custom {
+            url: url.as_str().to_string(),
+        }),
+        bucket_name: None,
+        app_context: Some(rs_ctx),
+    };
+    let remote_settings_service = RemoteSettingsService::new("".to_string(), config);
```
freshstrangemusic: should the storage dir be /tmp or something? Should we mkdtemp here?

components/nimbus/src/stateful/nimbus_client.rs
```diff
@@ -101,16 +101,18 @@ pub struct NimbusClient {
...
     pub fn new<P: Into<PathBuf>>(
         app_context: AppContext,
         recorded_context: Option<Arc<dyn RecordedContext>>,
         coenrolling_feature_ids: Vec<String>,
         db_path: P,
-        config: Option<RemoteSettingsConfig>,
         metrics_handler: Box<dyn MetricsHandler>,
         gecko_pref_handler: Option<Box<dyn GeckoPrefHandler>>,
+        collection_name: Option<String>,
+        remote_settings_service: Option<Arc<RemoteSettingsService>>,
```
freshstrangemusic: Actually since these two are tied together, it should be `Option<Arc<RemoteSettingsService>, String>` and then we dont need to handle the case where only one is provided!

components/nimbus/src/stateful/client/http_client.rs
```diff
@@ -13,18 +13,32 @@
...
-        let resp = self.get_records_raw()?;
-        parse_experiments(&resp.text())
+        let records = self.get_records(false).ok_or(RemoteSettingsError::Other {
+            reason: "Unable to fetch experiment records".to_owned(),
+        })?;
+        let data_vec: Vec<_> = records
+            .into_iter()
+            .map(|r| r.fields.get("data").cloned())
+            .collect();
+        let wrapped_data = json!({ "data": data_vec });
```
freshstrangemusic, line 38: I'm pretty sure we don't need to call .cloned() here -- it should be perfectly fine to borrow r.fields.data

components/nimbus/src/stateful/client/http_client.rs
```diff
@@ -13,18 +13,32 @@
...
+        let resp = serde_json::to_string(&wrapped_data).map_err(|e| {
+            NimbusError::JSONError("resp = serde_json::to_string".to_owned(), e.to_string())
```
freshstrangemusic: This should mention that its in SettingsClient::fetch_experiments so we can guarantee its a unique string

components/nimbus/src/stateful/client/http_client.rs
tiftran: I tried and it didn't compile, `cargo` wasn't too happy with me.

components/nimbus/src/stateful/nimbus_client.rs
tiftran: after discussing with @freshstrangemusic, will do as a follow up

PR #6663, opened 2025-03-24
Bug 1944609 - Normalize JEXL context

bendk:
There's two commits here and a choice of how much to change at once

The first commit implements https://bugzilla.mozilla.org/show_bug.cgi?id=1944609 by normalizing the `env` we create for our JEXL filter.  It keeps the `RemoteSettingsContext` API exactly the same, so there's no breaking changes to the consumers.

The second commit implements https://bugzilla.mozilla.org/show_bug.cgi?id=1955266 by removing `RemoteSettingsContext` fields that we aren't using and adding UniFFI default values for the remaining fields.  This is a breaking change for consumers, but it's pretty mechanical.  I considered using a Builder or something for this, but it felt like overkill vs a dict with `None` defaults.

Questions:
 * Are there any other `env` fields that I should be sending?
 * How do folks feel about this breaking consumer change?
 
  - Note:


components/remote_settings/src/context.rs
```diff
@@ -0,0 +1,103 @@
...
+        if let Some(os) = self.os {
+            v.insert(
+                "appinfo".to_string(),
+                json!({
+                    "OS": os,
+                }),
+            );
+        }
+        if let Some(form_factor) = self.form_factor {
+            v.insert("form_factor".to_string(), form_factor.into());
```
leplatrem, line 59: Can we use `formFactor`? Otherwise it would be the only field that is not camelcase.
@cbguder changed it too on Mars https://github.com/mozilla-services/mars/commit/fe8a6cceb207c6bc2bc553e589e4394ad64b9393
([Slack](https://mozilla.enterprise.slack.com/team/U0614EVEZCM))

components/remote_settings/src/context.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// The delivery channel of the application (e.g "nightly")
+    #[uniffi(default = None)]
+    pub channel: Option<String>,
+    /// User visible version string (e.g. "1.0.3")
+    #[uniffi(default = None)]
+    pub app_version: Option<String>,
+    /// The locale of the application during initialization (e.g. "es-ES")
+    #[uniffi(default = None)]
+    pub locale: Option<String>,
+    /// The name of the operating system (e.g. "Android", "iOS", "Darwin", "Windows")
```
leplatrem, line 26: nit: I think it's `WINNT` in Gecko

components/remote_settings/src/context.rs
```diff
@@ -0,0 +1,103 @@
...
+    #[uniffi(default = None)]
+    pub os: Option<String>,
+    /// Form-factor of the device ("phone", "tablet", or "desktop")
+    #[uniffi(default = None)]
+    pub form_factor: Option<String>,
+    /// Country of the user
```
leplatrem, line 32: How is this supposed to be populated by clients? 
I think that in quicksuggest uses something from Gecko that relies on a geojson file (`Region.sys.mjs`)
https://searchfox.org/mozilla-central/rev/88e49196740d9448f44506c2239a475c6c9eadc7/browser/components/urlbar/private/SuggestBackendRust.sys.mjs#444-447

leplatrem (approved):
Nice, thank you for tackling this so quickly, that's very appreciated 🙏 

> Are there any other env fields that I should be sending?

I don't see any obvious ones. But I just asked on `#delivery`, in case someone has an opinion.

I know we noticed that `env.appinfo.ID` led to too many different values. But shouldn't we at least keep it in case `application-services` is used in a different product that *Firefox*? ([see list of potential products](https://mozilla-hub.atlassian.net/wiki/spaces/LPT/pages/201655396/Legal+Product#Who-to-Talk-To%3A))

I like the clean-up. Historically, the reason why there were so many fields in the Gecko client is Normandy. It used to be our platform for experimentation, live repair, etc. Advanced targeting, bucket sampling etc. were necessary for progressive rollouts.
I like the idea to start small and iterate the two client implementation together. I will add `country` to the Gecko client (https://bugzilla.mozilla.org/show_bug.cgi?id=1956190)

> How do folks feel about this breaking consumer change?

I'm not familiar with the a-s ecosystem. But it is a pretty simple upgrading change. Maybe you could add a link to the final struct?


> [!NOTE]
> Instead of Desktop client, let's call it Gecko client. Because the application-services client is also used on Desktop (quicksuggest), and the Gecko client is also used on Android (Fenix)

bendk:
> I know we noticed that env.appinfo.ID led to too many different values. But shouldn't we at least keep it in case application-services is used in a different product that Firefox? ([see list of potential products](https://mozilla-hub.atlassian.net/wiki/spaces/LPT/pages/201655396/Legal+Product#Who-to-Talk-To%3A))

Yes that makes sense.  When I made this, I just looked at the fields in the `Context` section and skipped the Desktop-only ones.  Are there any other Desktop-only ones I should be including?

components/remote_settings/src/context.rs
bendk: It's not totally clear in my mind, but I added some pointers based on what I know.

components/remote_settings/src/context.rs
bendk: Good question.  @0c0w3 What do you think?  What's currently on remote-settings for the geo-expansion project?

components/remote_settings/src/lib.rs
```diff
@@ -32,51 +32,6 @@ use storage::Storage;
...
-    /// The manufacturer of the device the application is running on
-    pub device_manufacturer: Option<String>,
-    /// The model of the device the application is running on
-    pub device_model: Option<String>,
-    /// The locale of the application during initialization (e.g. "es-ES")
-    pub locale: Option<String>,
-    /// The name of the operating system (e.g. "Android", "iOS", "Darwin", "Windows")
-    pub os: Option<String>,
-    /// The user-visible version of the operating system (e.g. "1.2.3")
-    pub os_version: Option<String>,
```
leplatrem, line 66: I'm having doubt about removing `osVersion`. 
(Imagine a new iOS version offers a new feature or hardware capabilities that we would want to target in our products)
> Note: AFAIK we don't have it in Gecko, so it would have to be added there 👌

leplatrem:
> Are there any other Desktop-only ones I should be including?

I don't any other obvious one right now. 

We can wait a bit for some users to provide feedback on this list. Or start with this first version, and document well how to add new fields

Standard8 (approved):
This looks good to me, thank you!

I can't think of any other filters that search service would need at the moment.

components/remote_settings/src/context.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[uniffi(default = None)]
+    pub os_version: Option<String>,
+    /// Form-factor of the device ("phone", "tablet", or "desktop")
+    #[uniffi(default = None)]
+    pub form_factor: Option<String>,
+    /// Country of the user.
+    ///
+    /// This can be populated in two ways:
+    ///   - The second component of the locale
+    ///   - By using the Mozilla geolocation service, which determines country via the user's IP.
```
0c0w3, line 42: Nit: These doc comments are useful for consumers (like me), and for that reason this particular comment shouldn't be so prescriptive IMO. Desktop uses the `Region` module to get the client's country, and a desktop engineer might not know how that works under the hood. If they see that `Region` isn't included in this list, they might be confused. So I would make it clear that the two ways listed here are only examples, and this isn't an exhaustive list of how `country` can be populated.

(Also, IIRC desktop only uses the Mozilla location service on Nightly, and it uses a Google API in Release.)

components/remote_settings/src/context.rs
0c0w3: Right -- Mathieu, Can, and I [just discussed this](https://mozilla.slack.com/archives/CBGR8AW6R/p1742551041620309), so we're all on the same page here. I think right now the data in RS still uses snake case but Can said he would update it sometime this week. Desktop still passes in snake case too -- we'll need to make a desktop patch for that and this PR (as part of the vendor that includes this PR probably).

0c0w3 (approved):
Thanks!

0c0w3:
Oh one more thing -- do we really want to remove `custom_targeting_attributes`? That was very helpful for me in doing Suggest geo expansion because I didn't have to make any changes to this crate to do what I needed to do. It's still useful, if only for rapid prototyping and as part of the process for introducing new fields IMO, so I would leave it if it's not too hard.

components/remote_settings/src/context.rs
```diff
@@ -0,0 +1,123 @@
...
+        if self.app_id.is_some() || self.os.is_some() || self.os_version.is_some() {
+            let mut appinfo = Map::default();
+            if let Some(os) = self.os {
+                appinfo.insert("OS".to_string(), os.into());
+            }
+            if let Some(app_id) = self.app_id {
+                appinfo.insert("ID".to_string(), app_id.into());
+            }
+            if let Some(os_version) = self.os_version {
+                appinfo.insert("osVersion".to_string(), os_version.into());
```
leplatrem, line 70: Here we will have `env.appinfo.OS` and `env.appinfo.osVersion`.

In Gecko, `env.os` returns:
```
{
  "isWindows": false,
  "isMac": true,
  "isLinux": false,
  "windowsVersion": null,
  "windowsBuildNumber": null,
  "macVersion": 12,
  "darwinVersion": 24.2
}
``` 
which I don't find elegant at all. I would have preferred:
```
{
  "name": "iOS",
  "version": "1.2.3",
  "buildNumber": "44"
}
```

I will add `env.appinfo.osVersion` and these `env.os` fields in Gecko, and we can decide later whether we add `env.os` to A-S too.

leplatrem:
> Oh one more thing -- do we really want to remove custom_targeting_attributes?

I let you decide for this if the use-case is prototyping/debugging. But generally speaking, it is better if we make new fields official in all clients, especially if they are generic.
Or alternatively we can leave a generic `string` field in all platforms (eg. `env.extra`) that you initialize the way you want in your component, and put expressions in records that leverage it like `'"feat:2;" in env.extra'`

bendk:
Pushed some more changes:
 - Added back `custom_targetting_attributes`, but tried to make it clear that it should only be used for prototyping and in the short-term.
 - Renamed the key to `formFactor`

My plan is to merge this once the next nightly cycle starts, tell me if there's any changes that need to be made before then.

PR #6667, opened 2025-03-25
Add function to locally remove corrupted logins

lougeniaC64:
This PR adds a function to locally delete logins that cannot be decrypted so the deletion won't be propagated to the sync server. It additionally includes the following changes:
- Adds NSS initialization for the sync integration tests.
- Adds unit and integration tests for the new function

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -645,22 +645,50 @@ impl LoginDb {
         Ok(exists)
     }
 
-    pub fn delete_local_for_remote_replacement(&self, id: &str) -> Result<bool> {
+    pub fn verify_logins(&self, encdec: &dyn EncryptorDecryptor) -> Result<bool> {
+        // Retrieve a list of guids for logins that cannot be decrypted
+        let corrupted_logins = self
+            .get_all()?
+            .into_iter()
+            .filter(|login| login.clone().decrypt(encdec).is_err())
+            .collect::<Vec<_>>();
+        let ids = corrupted_logins
+            .iter()
+            .map(|login| login.guid_str())
+            .collect::<Vec<_>>();
+
+        Ok(self
+            .delete_local_records_for_remote_replacement(ids)
+            .is_ok())
+    }
+
+    pub fn delete_local_records_for_remote_replacement(&self, ids: Vec<&str>) -> Result<()> {
         let tx = self.unchecked_transaction_imm()?;
-        let exists = self.exists(id)?;
 
-        self.execute(
-            "DELETE FROM loginsL WHERE guid = :guid",
-            named_params! { ":guid": id },
-        )?;
+        sql_support::each_chunk(&ids, |chunk, _| -> Result<()> {
+            self.execute(
+                &format!(
+                    "DELETE FROM loginsL WHERE guid IN ({guids})",
+                    guids = sql_support::repeat_sql_values(chunk.len())
+                ),
+                rusqlite::params_from_iter(chunk),
+            )?;
+            Ok(())
+        })?;
 
-        self.execute(
-            "DELETE FROM loginsM WHERE guid = :guid",
-            named_params! { ":guid": id },
-        )?;
+        sql_support::each_chunk(&ids, |chunk, _| -> Result<()> {
+            self.execute(
+                &format!(
+                    "DELETE FROM loginsM WHERE guid IN ({guids})",
+                    guids = sql_support::repeat_sql_values(chunk.len())
+                ),
+                rusqlite::params_from_iter(chunk),
+            )?;
+            Ok(())
+        })?;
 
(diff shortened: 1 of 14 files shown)


lougeniaC64:
@mhammond @skhamis This is me introducing the local delete functionality for re-introducing the logins verification logic that caused the incident a while back. The new `delete_local` function will replace  the `delete` usage thereby preventing the deletions in the verification logic from being sent to the sync server.

components/logins/src/logins.udl
```diff
@@ -140,12 +143,18 @@ interface LoginStore {
     [Throws=LoginsApiError]
     boolean delete([ByRef] string id);
 
+    [Throws=LoginsApiError]
+    boolean delete_local([ByRef] string id);
```
mhammond, line 147: I think we should come up with a better name, because I can imagine someone accidentally calling this rather than `delete`. I'm not too sure what a good name is though, but I think a long and ugly name is fine given it's just for this special edge-case - eg, something like `delete_local_for_remote_replacement()` or something? And maybe a comment to explain why it shouldn't actually be used in normal cases.

components/logins/src/db.rs
```diff
@@ -645,6 +645,24 @@ impl LoginDb {
         Ok(exists)
     }
 
+    pub fn delete_local(&self, id: &str) -> Result<bool> {
```
mhammond, line 648: I think a comment here explaining why these 2 deletions are the "right thing" to do in this scenario and how it causes us to re-sync the remote item (which IIUC, also requires the engine to be "reset" - which I guess is done in the swift layer?)

lougeniaC64: @mhammond Yeah, I just convinced myself those two functions shouldn't be exposed. Let me know if you think `reset` is prefered over my `set_last_sync` usage.

mhammond: `set_last_sync` seems fine here! I agree a single `verify_logins` function seems better than exposing `delete_local` and `set_last_sync` too!

skhamis (commented):
Looks great! Only really had one question but for the most part looks solid!

components/logins/src/db.rs
```diff
@@ -645,6 +645,52 @@ impl LoginDb {
...
+            .get_all()?
+            .into_iter()
+            .filter(|login| login.clone().decrypt(encdec).is_err())
+            .collect::<Vec<_>>();
+        let ids = corrupted_logins
+            .iter()
+            .map(|login| login.guid_str())
+            .collect::<Vec<_>>();
+
+        Ok(self
```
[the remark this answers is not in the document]
lougeniaC64: This is a very good call out. I was rewriting this function from the reverted iOS PR and didn't think to reconsider the signature 🫠.

Anyway, I updated the `verify_logins` function so that if the local delete call fails it will be returned. I also updated the signature of the function as there's no need to return a boolean here as the API wrapper function in iOS can handle that and log any errors `verify_logins` returns.

skhamis (approved):
Thanks for making those changes -- looks good to me!

components/logins/src/db.rs
```diff
@@ -645,6 +645,50 @@ impl LoginDb {
         Ok(exists)
     }
 
+    pub fn verify_logins(&self, encdec: &dyn EncryptorDecryptor) -> Result<()> {
```
mhammond, line 648: I hate to nitpick the names too much, but "verify_foo" doesn't imply stuff might be deleted. Maybe `delete_undecryptable_records_for_remote_replacement()` just to keep the theme going :)

I honestly don't care too much though, whatever :)

components/logins/src/store.rs
```diff
@@ -151,6 +165,16 @@ impl LoginStore {
         Ok(())
     }
 
+    #[handle_error(Error)]
+    pub fn set_last_sync(self: Arc<Self>, last_sync: ServerTimestamp) -> ApiResult<()> {
```
mhammond, line 169: is this needed now `verify_logins` calls it?

mhammond (approved):
this is great, thanks!

components/logins/src/store.rs
lougeniaC64: I left a comment about this below but this function only exists for the `test_login_local_delete` integration test. I'll delete both of these functions now as I created that test to validate my overall approach.

lougeniaC64:
@mhammond @skhamis  FYI I made changes to [dependency_summary.py](https://github.com/mozilla/application-services/pull/6667/files#diff-ba3d4fcb31f0b23262bfd1e745403bb42173b104ee36ff9dfa1509c0945c7269R263) to resolve the Dependency Check failure after you both approved. Feel free to check my math. If I don't hear back in a few hours I'll merge.

PR #6672, opened 2025-03-28
Remote settings updates

bendk:
3 updates to the remote settings code.  I'm putting them all in 1 PR since they're fairly small and feel related to me, but I could split them out if any of them need extra time for review/consideration.  It's probably easier to review this by going through the commits 1 by 1.

Bugs links:
- [1956908: Defer IO until we need to get records](https://bugzilla.mozilla.org/show_bug.cgi?id=1956908)
- [1956803: Make relevancy close it's remote settings client](https://bugzilla.mozilla.org/show_bug.cgi?id=1956803)
- [1956527: Make the remote settings constructor infallible](https://bugzilla.mozilla.org/show_bug.cgi?id=1956527)

  - Note:

### DIFF
--- a/components/remote_settings/src/lib.rs
+++ b/components/remote_settings/src/lib.rs
@@ -48,11 +48,10 @@ impl RemoteSettingsService {
     ///
     /// This is typically done early in the application-startup process
     #[uniffi::constructor]
-    #[handle_error(Error)]
-    pub fn new(storage_dir: String, config: RemoteSettingsConfig2) -> ApiResult<Self> {
-        Ok(Self {
-            internal: service::RemoteSettingsService::new(storage_dir, config)?,
-        })
+    pub fn new(storage_dir: String, config: RemoteSettingsConfig2) -> Self {
+        Self {
+            internal: service::RemoteSettingsService::new(storage_dir, config),
+        }
     }
 
     /// Create a new Remote Settings client
--- a/components/remote_settings/src/lib.rs
+++ b/components/remote_settings/src/lib.rs
@@ -46,7 +46,14 @@ pub struct RemoteSettingsService {
 impl RemoteSettingsService {
     /// Construct a [RemoteSettingsService]
     ///
-    /// This is typically done early in the application-startup process
+    /// This is typically done early in the application-startup process.
+    ///
+    /// This method performs no IO or network requests and is safe to run in a main thread that
+    /// can't be blocked.
+    ///
+    /// `storage_dir` is a directory to store SQLite files in -- one per collection. If the
+    /// directory does not exist, it will be created when the storage is first used. Only the
+    /// directory and the SQLite files will be created, any parent directories must already exist.
     #[uniffi::constructor]
     pub fn new(storage_dir: String, config: RemoteSettingsConfig2) -> Self {
         Self {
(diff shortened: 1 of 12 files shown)


components/remote_settings/src/storage.rs
```diff
@@ -27,27 +27,51 @@ use sql_support::{open_database::open_database_with_flags, ConnExt};
...
-        Ok(Self { conn })
+            conn: ConnectionCell::Uninitialized,
+        })
+    }
+
+    fn transaction(&mut self) -> Result<Transaction<'_>> {
+        match &self.conn {
+            ConnectionCell::Uninitialized => {
+                if self.path != ":memory:" && std::fs::exists(&self.path)? {
+                    std::fs::create_dir(&self.path)?;
```
bendk, line 46: @mhammond I think there's some more motivation for this `create_dir` call now.  The issue I'm trying to work around is that creating a directory in JS requires an async call and we're currently creating `RemoteSettingsService` outside an async context.  I think it would be possible to update the JS code to do that, but it would mean wrapping `RemoteSettingsService` in some sort of async/lazy wrapper -- which would then mean all the suggest/relevancy components also need this wrapping.

components/remote_settings/src/storage.rs
```diff
@@ -270,6 +295,13 @@ impl Storage {
     }
 }
 
+/// Stores the SQLite connection, which is lazily constructed and can be closed/shutdown.
+enum ConnectionCell {
+    Uninitialized,
+    Initialized(Connection),
+    Closed,
+}
```
bendk, line 303: Lately I've been feeling like this should be the standard way we handle SQLite connections: make them lazily created and give the consumers a way to close them during shutdown.  This simplifies things on the consumer side:
* No need to wrap our components in a lazy wrapper.  You can just construct them with confidence that there's not going to be any IO happening.
* JS code can shutdown components with a `close` method and be sure that all resources are released.  Hopefully this will avoid the late-write errors.

components/remote_settings/src/lib.rs
```diff
@@ -46,16 +46,24 @@ pub struct RemoteSettingsService {
 impl RemoteSettingsService {
     /// Construct a [RemoteSettingsService]
     ///
-    /// This is typically done early in the application-startup process
+    /// This is typically done early in the application-startup process.
+    ///
+    /// This method performs no IO or network requests and is safe to run in a main thread that
```
mhammond, line 51: This part of the comment is wrong now, right? It *does* (well, "may") perform IO and thus is *not* safe to call on the main thread.

components/remote_settings/src/storage.rs
mhammond: that works ok in simple cases, but I'm not sure how that translates to connections which want custom flags? Also slightly concerned this will cause people to create brand new connections for every requests etc rather than managing them well. But the fact this is private for now makes me not worry until it spreads :)

components/remote_settings/src/lib.rs
bendk: I think it's right.   There's no IO preformed until we actually need to use the DB, which doesn't happen until something like `get_records` is called (https://github.com/mozilla/application-services/blob/42a75dec389e0a0673bc2836fdaee9d8d64a7a5e/components/remote_settings/src/storage.rs#L44).

components/remote_settings/src/storage.rs
bendk: Good points.  I won't try to spread this more than it needs to be and we can revisit in the future when we have a concrete use-case.

gruberb (approved):

PR #6680, opened 2025-04-01
Bug 1955387 - Add locale fallback handling to the Rust based search engine selector.

mandysGit:
- Introduce a new record type for a list of availablelocales in the search configuration
- Implement logic to use the user's locale directly, or strip the regional part for fallback
- Map unsupported English locales to en-US

  - Note:

### DIFF
--- a/components/search/src/filter.rs
+++ b/components/search/src/filter.rs
@@ -11,8 +11,9 @@ use crate::{
     JSONEngineUrls, JSONEngineVariant, JSONSearchConfigurationRecords, RefinedSearchConfig,
     SearchEngineDefinition, SearchEngineUrl, SearchEngineUrls, SearchUserEnvironment,
 };
-use crate::{sort_helpers, JSONEngineOrdersRecord};
+use crate::{sort_helpers, JSONAvailableLocalesRecord, JSONEngineOrdersRecord};
 use remote_settings::RemoteSettingsRecord;
+use std::collections::HashSet;
 
 impl From<JSONEngineUrl> for SearchEngineUrl {
     fn from(url: JSONEngineUrl) -> Self {
@@ -143,7 +144,7 @@ pub(crate) struct FilterRecordsResult {
 pub(crate) trait Filter {
     fn filter_records(
         &self,
-        user_environment: &SearchUserEnvironment,
+        user_environment: &mut SearchUserEnvironment,
         overrides: Option<Vec<JSONOverridesRecord>>,
     ) -> Result<FilterRecordsResult, Error>;
 }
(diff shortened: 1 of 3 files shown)


components/search/src/filter.rs
```diff
@@ -158,12 +160,52 @@ fn apply_overrides(engines: &mut [SearchEngineDefinition], overrides: &[JSONOver
...
+        .map(|locale| locale.to_lowercase())
+        .collect();
+
+    if available_locales_set.contains(&user_locale) {
+        return;
+    }
+    if locales_map_to_en_us.contains(&user_locale) {
+        user_environment.locale = "en-us".to_string();
+        return;
+    }
```
Standard8, line 180: I think we could simplify this to if user_locale starts with "en-" then make it "en-US".

The `en-*` locales we care about will presumably have already been listed in `available_locales_set`, so any other version of `en-` should just fall back to `en-US`. This would also avoid hard-coding the list of `en-*` locales which could easily be incomplete or be missed later.

components/search/src/filter.rs
```diff
@@ -213,14 +258,23 @@ impl Filter for Vec<JSONSearchConfigurationRecords> {
         user_environment: &SearchUserEnvironment,
         overrides: Option<Vec<JSONOverridesRecord>>,
     ) -> Result<FilterRecordsResult, Error> {
+        let mut available_locales = Vec::new();
+        for record in self {
+            if let JSONSearchConfigurationRecords::AvailableLocales(locales_record) = record {
+                available_locales = locales_record.locales.clone();
+            }
+        }
+        let mut user_environment = user_environment.clone();
```
Standard8, line 267: I think we should pass in a cloned, mutable user_environment here - currently I think `filter_engine_configuration_impl` would not get the updated locale, and hence it might determine the sort order/default engines incorrectly.

Standard8 (commented):
I've started looking at this, but haven’t had much time, will finish looking tomorrow but thought I'd submit the two comments in-line anyway in case you have time to have a look.

components/search/src/filter.rs
```diff
@@ -158,12 +160,52 @@ fn apply_overrides(engines: &mut [SearchEngineDefinition], overrides: &[JSONOver
...
         user_environment: &SearchUserEnvironment,
         overrides: Option<Vec<JSONOverridesRecord>>,
     ) -> Result<FilterRecordsResult, Error> {
+        let mut available_locales = Vec::new();
+        for record in self {
+            let stringified = serde_json::to_string(&record.fields)?;
+            if let Some(val) = record.fields.get("recordType") {
+                if *val == "availableLocales" {
+                    let locales_record: Option<JSONAvailableLocalesRecord> =
+                        serde_json::from_str(&stringified)?;
```
Standard8, line 201: Please can you put the `serde_json::to_string` statement within the second if statement to avoid it being called unnecessarily.

Standard8 (changes requested):
Thanks, one additional comment to address. The tests look good.

mandysGit:
Applied review comment changes and rebased ontop latest change on main.

Standard8 (approved):

PR #6681, opened 2025-04-02
Bug 1957805 - Replace exposure suggestions with dynamic suggestions.

0c0w3:
This replaces the exposure-suggestion type with a more general dynamic-suggestion type. The key differences between the two are:

* Dynamic suggestions are intended to be shown to the user unlike exposure suggestions -- but they don't have to be
* They can have an arbitrary JSON payload
* They can have an opaque dismissal key so they can be dismissed
* They can have a non-default score

Dynamic suggestions can be used to do exposure suggestions, so they're a superset basically.

Not sure if "dynamic" is a great name, I'm open to alternatives. I'm trying to convey that they can carry arbitrary payloads and don't require code changes.

My motivation is that we have a bunch of new suggestion types planned, and I don't want to have to keep making similar additions over and over. And more generally I'd like to move offline Suggest more in the direction of an Elasticsearch or generalized search engine that doesn't need to know every detail about the suggestions it serves.

This PR keeps the keyword-matching strategy of exposure suggestions: exact keyword matching, which should be sufficient for now. It's not hard to imagine supporting different strategies in the future and letting each dynamic suggestion type choose one (as we talked about for [exposure suggestions][1], e.g.: FTS, first word + prefix of remaining words, etc.) In the future I want to explore more sophisticated query parsing and matching like a real search engine would do (as we kind of talked about in [the weather-suggestions PR][2]).

[1]: https://github.com/mozilla/application-services/pull/6343
[2]: https://github.com/mozilla/application-services/pull/6406

  - Note:

### DIFF
--- a/components/suggest/src/suggestion.rs
+++ b/components/suggest/src/suggestion.rs
@@ -97,8 +97,14 @@ pub enum Suggestion {
         // and therefore the only one we'll collect metrics for.
         match_info: Option<FtsMatchInfo>,
     },
-    Exposure {
+    Dynamic {
         suggestion_type: String,
+        data: Option<serde_json::Value>,
+        /// This value is optionally defined in the suggestion's remote settings
+        /// data and is an opaque token used for dismissing the suggestion in
+        /// lieu of a URL. If `Some`, the suggestion can be dismissed by passing
+        /// the wrapped string to [crate::SuggestStore::dismiss_suggestion].
+        dismissal_key: Option<String>,
         score: f64,
     },
 }
@@ -190,7 +196,7 @@ impl Suggestion {
             | Self::Mdn { score, .. }
             | Self::Weather { score, .. }
             | Self::Fakespot { score, .. }
-            | Self::Exposure { score, .. } => *score,
+            | Self::Dynamic { score, .. } => *score,
             Self::Wikipedia { .. } => DEFAULT_SUGGESTION_SCORE,
         }
     }
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -23,9 +23,9 @@ use crate::{
     provider::{AmpMatchingStrategy, SuggestionProvider},
     query::{full_keywords_to_fts_content, FtsQuery},
     rs::{
-        DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedExposureSuggestion,
-        DownloadedFakespotSuggestion, DownloadedMdnSuggestion, DownloadedPocketSuggestion,
-        DownloadedWikipediaSuggestion, Record, SuggestRecordId,
+        DownloadedAmoSuggestion, DownloadedAmpSuggestion, DownloadedDynamicRecord,
+        DownloadedDynamicSuggestion, DownloadedFakespotSuggestion, DownloadedMdnSuggestion,
+        DownloadedPocketSuggestion, DownloadedWikipediaSuggestion, Record, SuggestRecordId,
     },
     schema::{clear_database, SuggestConnectionInitializer},
     suggestion::{cook_raw_suggestion_url, FtsMatchInfo, Suggestion},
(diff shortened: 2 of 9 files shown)


components/suggest/src/suggestion.rs
```diff
@@ -97,8 +97,14 @@ pub enum Suggestion {
         // and therefore the only one we'll collect metrics for.
         match_info: Option<FtsMatchInfo>,
     },
-    Exposure {
+    Dynamic {
         suggestion_type: String,
+        data: Option<serde_json::Value>,
```
bendk, line 102: Instead of storing a `Value` here, could this store an enum so that consumers don't need to deserialize the raw JSON?  That means that adding a new dynamic suggestion type would require app-services changes, but it would just be adding a new variant to the enum.

I think that would mean using `#[serde(tag = "suggestion_type")]` on the enum rather than storing the `suggestion_type` as a separate field.  I don't think there would need to that many other changes though.

bendk (changes requested):
This code looks correct.  My one concern is returning raw JSON to the consumer.  My feeling is that it's going to add a lot of friction on the mobile side of things and it's easier to do the deserialization in Rust.

components/suggest/src/suggestion.rs
bendk: I'm also confused at how UniFFI is letting you use `serde_json::Value` like that.  Normally there needs to be some sort of custom type definition.

ncloudioj (approved):
Looks good to me, r=nanj.

I like this generalization. This reminds me of the transition from the pref-based experimentation at the Normandy age to the JSON-based experimentation for Nimbus.

0c0w3:
Thanks!

> Instead of storing a `Value` here, could this store an enum so that consumers don't need to deserialize the raw JSON? That means that adding a new dynamic suggestion type would require app-services changes, but it would just be adding a new variant to the enum.

I'd really like to avoid that. The whole point is not to require any code changes at all, in general. In specific cases, where the product spec requires special query parsing or something like that, we can do business as usual and add a bunch of Rust code. Yelp and weather are two examples, although we could do a better job of generalizing them too, to @ncloudioj's point in the weather PR about Elasticsearch. But for the general case going forward, the benefits of this approach are compelling.

> My feeling is that it's going to add a lot of friction on the mobile side of things and it's easier to do the deserialization in Rust.

Is that true? I definitely don't intend to cause headaches for mobile. I'm not a mobile developer but I can't imagine it's hard nowadays in either Android or iOS to handle a non-typed JSON object? It's similar to what they'd need to do to handle suggestions from Merino.

I mean desktop will need to handle that too, and that's part of the point -- there will always need to be some consumer knowledge of the suggestion schemas so they can build UIs from them, visit URLs when clicked, etc. (Unless/until we develop some sort of codified UI framework...)

Also -- most of these suggestion types haven't even been used by mobile at all. There are like half a dozen new suggestion types coming up as part of the "carrot" approach to enticing users to opt in to online, and I don't think any of them are planned for mobile. If we view this as primarily a way for desktop to ship new suggestions, does that help?

0c0w3:
> It's similar to what they'd need to do to handle suggestions from Merino.

Although that's only true if mobile were to talk to Merino directly, like desktop has always done, instead of us building Merino support into `suggest`, which is probably what we'd do if/when we bring Merino to mobile.

bendk:
> > My feeling is that it's going to add a lot of friction on the mobile side of things and it's easier to do the deserialization in Rust.
> 
> Is that true? I definitely don't intend to cause headaches for mobile. I'm not a mobile developer but I can't imagine it's hard nowadays in either Android or iOS to handle a non-typed JSON object? It's similar to what they'd need to do to handle suggestions from Merino.

My understanding is that Swift will need to add a struct that reflects the JSON schema.  I'm not sure how it's handled on Kotlin, but the same work might be needed.  You might end up avoiding a change in application-services, but having to make an equivalent change twice on mobile.  OTOH, it'll be alongside a PR with the other changes needed, so maybe that's actually less work overall.

> Also -- most of these suggestion types haven't even been used by mobile at all. There are like half a dozen new suggestion types coming up as part of the "carrot" approach to enticing users to opt in to online, and I don't think any of them are planned for mobile. If we view this as primarily a way for desktop to ship new suggestions, does that help?

Yes that definitely does.  I think it's a decision between the Rust devs and the application devs.  If these are only landing on Desktop, then that's basically just you.

I'm going to change my feedback to approve.  I personally would like more resources devoted to Suggest on Mobile, but that's a very different discussion and shouldn't block this.

bendk (approved):

0c0w3:
OK, thanks for the discussion. I'll go ahead and merge this after addressing Nan's comment.

PR #6688, opened 2025-04-07
More gradle tweaks so our gradle files work unchanged in mozilla-cent…

mhammond:
…ral.


bendk (approved):
Mostly a rubber stamp, but this seems good to me.

mhammond:
cc @rvandermeulen, but I'll take Ben's review as all I need for now.

PR #6707, opened 2025-04-16
Bug 1960786 - Dismiss AMP suggestions by full keyword instead of URL

0c0w3:
This makes one other substantive change: If an AMP suggestion doesn't have `full_keywords` defined in its RS data, then `Suggestion.full_keyword` will now be an empty string instead of a string computed at query time from its keywords.

The problem with using such a computed "full keyword" as the dismissal key is that it's not stable. It's just the prefix of one of the suggestion's keywords up to the next full word after the query string, so it changes the more you type.

When the RS data doesn't have a `full_keywords`, I set `Suggestion.full_keyword` to an empty string and then fall back to the suggestion's URL as the dismissal key. In practice, all AMP suggestions have full keywords in the RS data (I checked), and going forward with the new AMP API that won't change, so we shouldn't ever hit this to begin with.

Other possible ways to address this:

* Change the query-time full keyword calculation so that the longest matching keyword is used as the full keyword. That might actually be good for UX too, maybe not in all cases.
* Add a `Suggestion` member called `is_full_keyword_from_db` or something (like the similarly named variable in the AMP fetch function), and then we'd use the URL as the dismissal key if `is_full_keyword_from_db` is false.

These are both more complex than setting `full_keywords` to an empty string and using that to tell if the dismissal key should be the URL, and since we shouldn't ever hit this anyway, I didn't bother.

  - Note:


0c0w3:
Or:

* `dismiss_suggestion` could take a `Suggestion` object (although we'd probably want to add a new function and deprecate the old one)
* Callers could just know they should call `store.dismiss_suggestion(suggestion.full_keyword)` for AMP suggestions

components/suggest/src/store.rs
```diff
@@ -205,12 +205,11 @@ impl SuggestStore {
...
     ///
-    /// Dismissed suggestions will not be returned again
-    ///
-    /// In the case of AMP suggestions this should be the raw URL.
+    /// To dismiss a suggestion, pass in its [Suggestion::dismissal_key].
+    /// Dismissed suggestions will not be returned again.
     #[handle_error(Error)]
-    pub fn dismiss_suggestion(&self, suggestion_url: String) -> SuggestApiResult<()> {
-        self.inner.dismiss_suggestion(suggestion_url)
+    pub fn dismiss_suggestion(&self, dismissal_key: String) -> SuggestApiResult<()> {
```
ncloudioj, line 211: Yup, I like the idea of taking a `Suggestion` enum as the argument so that `dismissal_key()` can be called without crossing the FFI boundary.

components/suggest/src/testing/mod.rs
```diff
@@ -33,6 +40,19 @@ impl JsonExt for JsonValue {
         self_map.append(other_map);
         self
     }
+
+    fn remove<Q>(mut self, key: &Q) -> JsonValue
+    where
+        String: Borrow<Q>,
+        Q: ?Sized + Ord + Eq + Hash,
+    {
+        let self_map = match &mut self {
```
ncloudioj, line 49: Just an FYI, there is another handy way to do "match this pattern and bind the matched, otherwise panic".

```rs
    let JsonValue::Object(obj) = &mut self else {
        panic!("remove called on non-object: {self:?}");
    };
    obj.remove(key);
```

ncloudioj (approved):
LGTM, Thanks!

Re: the difficulty of generating the dismissal_key, I like the idea of passing a `Suggestion` enum to `dismiss_suggestion()`, but it's up to you or Ben perhaps can provide a better solution.

components/suggest/src/db.rs
```diff
@@ -308,7 +308,9 @@ impl<'a> SuggestDao<'a> {
                   s.provider = :provider
                   AND k.keyword = :keyword
                   {where_extra}
-                AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url=s.url)
+                  AND NOT EXISTS (
```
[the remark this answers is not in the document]
0c0w3: Good idea -- in light of my commit that added `SuggestStore::dismiss` and deprecated `dismiss_suggestion`, I phrased this so it's clear URLs were used in the old API and full keywords are used in the new API.

components/suggest/src/testing/mod.rs
0c0w3: Thanks, I updated `JsonExt::merge` too.

ncloudioj (approved):
r+!

PR #6710, opened 2025-04-18
Revert "Bug 1957309: Handle location signs as one of Yelp modifier"

dadaa:
Reverts mozilla/application-services#6678


0c0w3 (approved):

0c0w3:
FTR, we're backing out the recent Yelp commits as a short-term fix to https://bugzilla.mozilla.org/show_bug.cgi?id=1961308. That will give us time to figure out the cause of the crash and make the necessary fixes.

PR #6712, opened 2025-04-18
Revert "Bug 1953667: Make location sign skippable"

dadaa:
Reverts mozilla/application-services#6656


0c0w3 (approved):
FTR, we're backing out the recent Yelp commits as a short-term fix to https://bugzilla.mozilla.org/show_bug.cgi?id=1961308. That will give us time to figure out the cause of the crash and make the necessary fixes.

PR #6715, opened 2025-04-18
Remove xcodeproj [ci full]

skhamis:
There is a fair amount of happening in the code but the concept is pretty straightforward:

## What is happening in this PR
1. Move all swift wrappers & tests to one directory (our current iOS megazord directory)
2. Remove the xcodeproject from our repository and use a simple SPM `Package.swift` file to handle our wrappers and run our tests

## Why are we doing this
1. XCode tends to be un-intuitive for most people not using it as a daily driver, this project doesn't really _need_ it as this is mostly the rust/api layer.
2. application-services is eventually moving into mozilla-central and where desktop/android lives and thus swift will be the odd one out. So we want to ensure that the development experience is repo-agnostic.
3. SPM is easier to work with than XCodeproject (see Package.swift) and we _already_ use SPM for https://github.com/mozilla/rust-components-swift so there's not much changes to the fundamentals, just in the way we test.


skhamis:
@lougeniaC64 @mhammond now that this is in the green, I think it's in good shape to get reviewed. The biggest discussion I see is moving all the wrappers out of their own components and instead goes into one space `megazords/ios-rust/Sources/MozillaRustComponentsWrapper` (recommended https://www.swift.org/documentation/package-manager/ in Creating a Package Library). I think it should simplify things (write your rust, anything swift related is megazords/ios-rust)

libs/verify-ios-environment.sh
```diff
@@ -18,13 +18,11 @@ fi
...
-echo "- Run the tests via the XCode project:"
-echo "    open megazords/ios-rust/MozillaTestServices.xcodeproj"
+echo "Looks good! You can either:"
 echo ""
 echo "- Run the iOS tests via command line:"
 echo "    ./automation/run_ios_tests.sh"
+echo ""
+echo " If you want to just generate the rust binaries"
+echo "- Build the XCFramework:"
+echo "    ./megazords/ios-rust/build-xcframework.sh"
```
lougeniaC64, line 28: nit: Should the build script be run before the tests? If so it might make sense to move this addition above the test instructions.

megazords/ios-rust/generate-files.sh
```diff
@@ -25,6 +25,12 @@ CARGO="$HOME/.cargo/bin/cargo"
 "$CARGO" uniffi-bindgen-library-mode -l "$UNIFFI_BINDGEN_LIBRARY" swift --headers "$COMMON/Headers"
 "$CARGO" uniffi-bindgen-library-mode -l "$UNIFFI_BINDGEN_LIBRARY" swift --modulemap "$COMMON/Modules" --xcframework --modulemap-filename module.modulemap
 
+## Tests will need the generated swift files from uniffi
+# TODO: Should we wrap this around an argument? we'd only need this for tests
```
lougeniaC64, line 29: What other processes need to use this script? Is it (just) rust-components-swift? It might be worth conditionally executing this script only when needed so that the processes that don't require it can skip it and bypass any future bustages.

lougeniaC64:
FYI: I attempted to execute the `run_ios_tests.sh` script (after running the iOS and desktop verification and build scripts) and it failed with the following error:

```
2025-04-24 15:24:40.075 xcodebuild[94280:33935168] Writing error result bundle to /var/folders/7_/l17tfv7d1bj07xz5jjzq8n980000gn/T/ResultBundle_2025-24-04_15-24-0040.xcresult
xcodebuild: error: Unable to find a device matching the provided destination specifier:
                { platform:iOS Simulator, OS:latest, name:iPhone 15 }
```

I do have the targets below installed so I'm assuming the issue is with `OS:latest` but don't mind me.
```
 { platform:iOS Simulator, arch:arm64, id:84F47380-4F73-467A-A3CD-98804C977908, OS:17.0, name:iPhone 15 }
                { platform:iOS Simulator, arch:x86_64, id:84F47380-4F73-467A-A3CD-98804C977908, OS:17.0, name:iPhone 15 }
                { platform:iOS Simulator, arch:arm64, id:E4E76DC8-72A8-4030-9DA2-821C49481FAA, OS:17.2, name:iPhone 15 }
                { platform:iOS Simulator, arch:x86_64, id:E4E76DC8-72A8-4030-9DA2-821C49481FAA, OS:17.2, name:iPhone 15 }
                { platform:iOS Simulator, arch:arm64, id:860B1BF6-E39C-4A29-9B90-82CD044D678E, OS:17.5, name:iPhone 15 }
                { platform:iOS Simulator, arch:x86_64, id:860B1BF6-E39C-4A29-9B90-82CD044D678E, OS:17.5, name:iPhone 15 }
```

lougeniaC64:
nit: Also, when I run the `run_ios_tests.sh` script it creates the `.venv/` directory which isn't git ignored. It might be that if I executed the script successfully that directory would have been removed but just want to note this in the event that's not the case and this directory should be git ignored.

mhammond:
Sammy walked me though this and it looks great to me at a high level - so @lougeniaC64 and a review from the ios team is all you need here!

libs/verify-ios-environment.sh
skhamis: The test scripts have builds baked in before running the tests. This is just if the user only needed the binary for some reason. Happy to make this more clear though!

megazords/ios-rust/generate-files.sh
skhamis: generate-files is mainly just for building the xcframework, so in actual processes it would take a couple of actual seconds but do nothing with those sources. However I totally forgot to add a nice wrapper to ensure we only do it if we pass something like `--generate-sources`, which I did now -- thanks for catching that!

lougeniaC64 (approved):
Finally got the iOS tests to execute successfully. Outside of the Focus fix we discussed LGTM! 🚀

PR #6717, opened 2025-04-21
Bug 1953667: Make location sign skippable (again)

dadaa:
The reason for the crash (bug 1961308) was that if the number of words taken from DB is larger than the number of remaining words of the user's query, it failed to change the rest of the query.
e.g.
```
let query_words = ["spi"];
let subject_tuple = ("spicy ramen" /* due to prefix match */, false, 2 /* matched words */);
query_words = &query_words[subject_tuple.2...]; 
// means,
query_words = &query_words[2...]; 
// then crash. Because query_words.len(*) is 1.
```

So, I changed as below:
1. find_subject() returns the number of shorter words in the rest of the query or in the keywords from DB.
2. Modify the remaining query array with more safe number.

  - Note:

### DIFF
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -1851,8 +1851,20 @@ pub(crate) mod tests {
             ),],
         );
         assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_delivery")),
+            vec![ramen_suggestion(
+                "ramen invalid_delivery",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_delivery"
+            )
+            .has_location_sign(false),],
+        );
+        assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_delivery in tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen invalid_delivery in tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_delivery+in+tokyo"
+            )
+            .has_location_sign(false),],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen in tokyo")),
@@ -1870,7 +1882,11 @@ pub(crate) mod tests {
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen invalid_in tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen invalid_in tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=invalid_in+tokyo"
+            )
+            .has_location_sign(false),],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen in San Francisco")),
@@ -1890,21 +1906,22 @@ pub(crate) mod tests {
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near by")),
             vec![ramen_suggestion(
                 "ramen near by",
-                "https://www.yelp.com/search?find_desc=ramen+near+by"
-            )
-            .has_location_sign(false),],
+                "https://www.yelp.com/search?find_desc=ramen"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near me")),
             vec![ramen_suggestion(
                 "ramen near me",
-                "https://www.yelp.com/search?find_desc=ramen+near+me"
-            )
-            .has_location_sign(false),],
+                "https://www.yelp.com/search?find_desc=ramen"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen near by tokyo")),
-            vec![],
+            vec![ramen_suggestion(
+                "ramen near by tokyo",
+                "https://www.yelp.com/search?find_desc=ramen&find_loc=tokyo"
+            )],
         );
         assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("ramen")),
@@ -1993,6 +2010,15 @@ pub(crate) mod tests {
             .subject_exact_match(false)],
         );
         assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("spi")),
+            vec![ramen_suggestion(
+                "spicy ramen",
+                "https://www.yelp.com/search?find_desc=spicy+ramen"
+            )
+            .has_location_sign(false)
+            .subject_exact_match(false)],
+        );
+        assert_eq!(
             store.fetch_suggestions(SuggestionQuery::yelp("BeSt             Ramen")),
             vec![ramen_suggestion(
                 "BeSt Ramen",
@@ -2050,6 +2076,51 @@ pub(crate) mod tests {
             )
             .has_location_sign(false)
             .subject_exact_match(false)],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("best sp")),
+            vec![ramen_suggestion(
+                "best spicy ramen",
+                "https://www.yelp.com/search?find_desc=best+spicy+ramen"
+            )
+            .has_location_sign(false)
+            .subject_exact_match(false)],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("ramenabc")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("ramenabc xyz")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("best ramenabc")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("bestabc ra")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("bestabc ramen")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("bestabc ramen xyz")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("best spi ram")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("bes ram")),
+            vec![],
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("bes ramen")),
+            vec![],
         );
 
         Ok(())
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -30,6 +30,12 @@ impl ToSql for Modifier {
     }
 }
 
+#[derive(Clone, Copy, Eq, PartialEq)]
+enum FindFrom {
+    First,
+    Last,
+}
+
 /// This module assumes like following query.
 /// "Yelp-modifier? Pre-modifier? Subject Post-modifier? (Location-modifier | Location-sign Location?)? Yelp-modifier?"
 /// For example, the query below is valid.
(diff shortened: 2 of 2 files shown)


components/suggest/src/yelp.rs
```diff
@@ -142,69 +153,209 @@ impl SuggestDao<'_> {
...
         Ok(vec![builder.into()])
     }
 
+    /// Find the modifier for given query and modifier type.
+    /// Find from last word, if set FindFrom::Last to find_from.
+    /// It returns Option<tuple> as follows:
+    /// (
+    ///   String: The keyword in DB (but the case is inherited by query).
+    ///   usize: Number of words in query_words that match the keyword.
+    ///          Maximum number is MAX_MODIFIER_WORDS_NUMBER.
```
0c0w3, line 220: The maximum number is actually `min(MAX_MODIFIER_WORDS_NUMBER, query_words.len())`

0c0w3 (commented):
Thanks! This should be safer but it still uses direct indexes into slices -- I know that was my idea, sorry! I think this episode has taught us that we should avoid that.

How about this instead. Given `n`, `find_modifier` and the other `find` methods return two subslices: `0..n` and `n..`. Actually they would return an optional two subslices in case `n` is out of bounds. There's a slice method that does exactly that: [`split_at_checked`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_at_checked). And it does it safely, i.e. it won't panic if `n` is out of bounds. The tradeoff is that it returns an `Option` so we need to always check the return value.

So as a minimal example:

```rust
fn find_foo<'a>(n: usize, slice: &'a [&'a str]) -> Option<(&'a [&'a str], &'a [&'a str])> {
    slice.split_at_checked(n)
}
```

The caller would need to join the first returned subslice to produce a substring.

Could you try something like that?

If `split_at_checked` isn't practical in this case for some reason, there are other similar safe methods for getting subslices, for example: [`get`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get) (can be passed a range to return a subslice), [`first_chunk`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first_chunk).

components/suggest/src/yelp.rs
dadaa: Indeed, I will update this comment. Thanks!

dadaa:
> Thanks! This should be safer but it still uses direct indexes into slices -- I know that was my idea, sorry! I think this episode has taught us that we should avoid that.
> 
> How about this instead. Given `n`, `find_modifier` and the other `find` methods return two subslices: `0..n` and `n..`. Actually they would return an optional two subslices in case `n` is out of bounds. There's a slice method that does exactly that: [`split_at_checked`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_at_checked). And it does it safely, i.e. it won't panic if `n` is out of bounds. The tradeoff is that it returns an `Option` so we need to always check the return value.
> 
> So as a minimal example:
> 
> ```rust
> fn find_foo<'a>(n: usize, slice: &'a [&'a str]) -> Option<(&'a [&'a str], &'a [&'a str])> {
>     slice.split_at_checked(n)
> }
> ```
> 
> The caller would need to join the first returned subslice to produce a substring.
> 
> Could you try something like that?
> 
> If `split_at_checked` isn't practical in this case for some reason, there are other similar safe methods for getting subslices, for example: [`get`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get) (can be passed a range to return a subslice), [`first_chunk`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first_chunk).

Thank you very much, @0c0w3!
Please let me use `get`. For the modifiers, if we use `split_at_checked`, we can use the latter one as remaining query, but the former is not used as is since we keep the cases. For the subject, we get it by SQL with prefix-matching, can not use the former/latter as it is. So, I want to try `get` first.

0c0w3:
There's still an unnecessary step here because the `find` methods return an `n` that the callers do not assume is in bounds, so they're required to safely check that it is in bounds by using `get` (or another safe function). Instead, the `find` methods can/should return a slice containing the rest of the query words, and then callers don't need to do anything else. Here's what I mean:

```rust
    // Returns Option<(candidate, rest)>
    fn find_modifier2<'a>(
        &self,
        query_words: &'a [&'a str],
        modifier_type: Modifier,
        find_from: FindFrom,
    ) -> Result<Option<(String, &'a [&'a str])>> {
        // The `query_words.is_empty()` early return isn't necessary since the for-loop
        // below won't be entered in that case.

        for n in (1..=std::cmp::min(MAX_MODIFIER_WORDS_NUMBER, query_words.len())).rev() {
            let Some((candidate_chunk, rest)) = (match find_from {
                FindFrom::First => query_words
                    .split_at_checked(n)
                    .map(|(front, back)| (front, back)),
                FindFrom::Last => query_words
                    .split_at_checked(query_words.len() - n)
                    .map(|(front, back)| (back, front)),
            }) else {
                continue;
            };

            assert!(candidate_chunk.len() > 0); // debugging only
            let candidate = candidate_chunk.join(" ");

            if self.conn.query_row_and_then_cachable(
                "
                SELECT EXISTS (
                    SELECT 1 FROM yelp_modifiers WHERE type = :type AND keyword = :word LIMIT 1
                )
                ",
                named_params! {
                    ":type": modifier_type,
                    ":word": candidate.to_lowercase(),
                },
                |row| row.get::<_, bool>(0),
                true,
            )? {
                return Ok(Some((candidate, rest)));
            }
        }

        Ok(None)
    }
```

And an example caller:

```rust
        if let Some((_, rest)) = self.find_modifier2(query_words, Modifier::Yelp, FindFrom::First)? {
            query_words = rest;
        }
```

Also I noticed that in `find_modifier`, `candidate` can be an empty string (I think). So we end up querying the DB with an empty candidate string sometimes. That's why I have that `assert!` in the code above, just to confirm that the candidate is never empty.

Could you add some tests at the bottom of `yelp.rs` just for `find_modifier` and `find_subject` please? We should test all reasonable combinations of query strings, modifier types, and `FindFrom` values. We should make sure to get this right. Example:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    use crate::{
        testing::*,
        store::tests::TestStore,
        SuggestIngestionConstraints,
    };

    #[test]
    fn t1() -> anyhow::Result<()> {
        before_each();

        let store = TestStore::new(
            MockRemoteSettingsClient::default()
                .with_record(SuggestionProvider::Yelp.record("data-4", json!([ramen_yelp()])))
                .with_record(SuggestionProvider::Yelp.icon(yelp_favicon()))
        );

        store.ingest(SuggestIngestionConstraints {
            providers: Some(vec![SuggestionProvider::Yelp]),
            ..SuggestIngestionConstraints::all_providers()
        });

        store.read(|dao| {
            assert_eq!(
                dao.find_modifier2(
                    &"".split_whitespace().collect::<Vec<_>>(),
                    Modifier::Pre,
                    FindFrom::First
                )?,
                None
            );

            // Example query strings to test:
            //
            // "foo"
            // "foo bar"
            // "best"
            // "best foo"
            // "best foo bar"
            // "foo best" (FindFrom::First shouldn't parse this)
            // etc.
        })?;

        Ok(())
    }
}
```

You can write the test cases like we do in JS tests where there's an array of test objects/tuples with query strings to test + expected return values from `find_modifier`, etc. [Here's a geonames example](https://github.com/mozilla/application-services/blob/main/components/suggest/src/geoname.rs#L827-L846). You could use tuples instead of a new struct type if that's simpler.

dadaa:
Hi @0c0w3! I'm sorry for my delay. Please review when you have time. Thanks!

components/suggest/src/yelp.rs
```diff
@@ -509,3 +483,227 @@ impl<'a> From<SuggestionBuilder<'a>> for Suggestion {
...
+        );
+
+        store.ingest(SuggestIngestionConstraints {
+            providers: Some(vec![SuggestionProvider::Yelp]),
+            ..SuggestIngestionConstraints::all_providers()
+        });
+
+        store.read(|dao| {
+            test_find_modifier(dao, "", Modifier::Pre, FindFrom::First, None)?;
+            test_find_modifier(
```
0c0w3, line 510: Please also test these modifier queries (neither should match):

```
"bestabc"
"bestabc ramen"
```

Also IMO there's no need to test strings with leading/trailing spaces because both `find_modifier` and `find_subject` take a vec slice, not a string, and I think it's safe to assume that they will be called with slices that were created with `split_whitespace().collect::<Vec<_>>()`, which does not create vec items for leading/trailing spaces. e.g., `"   foo    ".split_whitespace().collect::<Vec<_>>()` is just `["foo"]`, not `["", "foo", ""]`.

If you do want to actually test slices with empty strings, that would be fine with me, but you'll need to modify `test_find_modifier`/`test_find_subject`.

components/suggest/src/yelp.rs
```diff
@@ -142,299 +153,262 @@ impl SuggestDao<'_> {
...
-        )?;
+    fn find_modifier<'a>(
+        &self,
+        query_words: &'a [&'a str],
+        modifier_type: Modifier,
+        find_from: FindFrom,
+    ) -> Result<Option<(String, &'a [&'a str])>> {
+        if query_words.is_empty() {
+            return Ok(None);
+        }
```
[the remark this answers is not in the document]
dadaa: Thanks! As the other find functions also do the same check first, I want to keep it for now.

components/suggest/src/yelp.rs
```diff
@@ -142,299 +153,262 @@ impl SuggestDao<'_> {
...
+            true,
+        )? {
+            // Preserve the query as the user typed it including its case.
+            let keyword = &query_string[0..keyword_lowercase.len()];
+            let count = keyword.split_whitespace().count();
+            return Ok(Some((
+                keyword.to_string(),
+                true,
+                query_words.get(count..).unwrap_or_default(),
+            )));
```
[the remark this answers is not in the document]
dadaa: Oh, indeed, thank you very much! I will fix.

components/suggest/src/yelp.rs
```diff
@@ -142,299 +153,260 @@ impl SuggestDao<'_> {
...
+                query_string,
+                &keyword_lowercase[query_string.len()..]
+            );
+
+            let count = std::cmp::min(query_words.len(), keyword.split_whitespace().count());
+            return Ok(Some((
+                keyword,
+                false,
+                query_words.get(count..).unwrap_or_default(),
+            )));
```
0c0w3, line 330:
```suggestion
            return Ok(keyword_lowercase
                .get(query_string.len()..)
                .map(|keyword_rest| {
                    query_string.push_str(keyword_rest);
                    let count =
                        std::cmp::min(query_words.len(), query_string.split_whitespace().count());
                    (
                        query_string,
                        false,
                        query_words.get(count..).unwrap_or_default(),
                    )
                }));
```

Similar here. Also, `format!` is probably slower than just appending the rest of the keyword to the query string. With my suggested change, you'll need to declare `query_string` as `mut`.

components/suggest/src/yelp.rs
```diff
@@ -509,3 +481,238 @@ impl<'a> From<SuggestionBuilder<'a>> for Suggestion {
...
+                .with_record(SuggestionProvider::Yelp.icon(yelp_favicon())),
+        );
+
+        store.ingest(SuggestIngestionConstraints {
+            providers: Some(vec![SuggestionProvider::Yelp]),
+            ..SuggestIngestionConstraints::all_providers()
+        });
+
+        store.read(|dao| {
+            test_find_modifier(dao, "", Modifier::Pre, FindFrom::First, None)?;
```
0c0w3, line 507: It doesn't look like there are any tests for queries that are substrings of modifiers, like "be", "bes", "chea", etc. Could you add some please, using the different combinations of `Modifier` and `FindFrom`? (In contrast there are such tests for subjects, which is great.)

components/suggest/src/store.rs
```diff
@@ -2051,6 +2077,15 @@ pub(crate) mod tests {
...
         );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::yelp("best sp")),
+            vec![ramen_suggestion(
+                "best spicy ramen",
+                "https://www.yelp.com/search?find_desc=best+spicy+ramen"
+            )
+            .has_location_sign(false)
+            .subject_exact_match(false)],
+        );
```
0c0w3, line 2088: Could you add some queries like these please, where a modifier/subject is a substring of a query word?

```
"ramenabc"
"ramenabc xyz"
"best ramenabc"
"bestabc ra"
"bestabc ramen"
"bestabc ramen xyz"
```

Also different combinations of prefixes like these:

```
"best spi ram"
"bes ram"
"bes ramen"
```

0c0w3 (approved):
Thanks for all your work on this! I have some comments but nothing major.

components/suggest/src/yelp.rs
```diff
@@ -142,299 +153,260 @@ impl SuggestDao<'_> {
...
+    /// It returns Option<tuple> as follows:
     /// (
-    ///   String: Subject.
-    ///   bool: Whether the subject matched exactly with the parameter.
+    ///   String: The keyword in DB (but the case is inherited by query).
+    ///   &[&str]: Words after removed matching location sign.
     /// )
-    fn find_subject(&self, candidate: &str) -> Result<Option<(String, bool)>> {
-        if candidate.is_empty() {
+    fn find_location_sign<'a>(
```
0c0w3, line 342: Oh also I forgot about `find_location_sign` -- could you add similar tests in `yelp.rs` for it too please?

components/suggest/src/yelp.rs
dadaa: I see, thanks!

PR #6724, opened 2025-04-23
Fixup localpublish workflows.

mhammond:
I updated the localpublish docs to reflect the fact the gradle cache must be disabled, and also to try and help clarify the fact we are talking about 2 different `local.properties` files and clarify which ones we are asking be modified.

 I also updated the script to delete possibly "stale" .so files, which caused re-publishes to not work.

Flagging @gruberb as I hope to walk him through this, and @badboy as I suspect all these changes apply to glean, so he may or may not want to apply this to the glean repo.

### DIFF
--- a/docs/howtos/locally-published-components-in-fenix.md
+++ b/docs/howtos/locally-published-components-in-fenix.md
@@ -8,44 +8,69 @@ from there.
 With support from the upstream project, it's possible to do this in a single step using
 our auto-publishing workflow.
 
-## rust.targets
-
-Both the auto-publishing and manual workflows can be sped up significantly by
-using the `rust.targets` property which limits which architectures the Rust
-code gets build against.  You can set this property by creating/editing the
-`local.properties` file in the repository root and adding a line like
-`rust.targets=x86,linux-x86-64`.  The trick is knowing which targets to put in
-that comma separated list:
-
-  - Use `x86` for running the app on most emulators (in rare cases, when you have a 64-bit emulator, you'll want `x86_64`)
-  - If you're running the `android-components` or `fenix` unit tests, then you'll need the architecture of your machine:
-    - OSX running Intel chips: `darwin-x86-64`
-    - OSX running M1 chips: `darwin-aarch64`
-    - Linux: `linux-x86-64`
-
-## Using the auto-publishing workflow
+# Using the auto-publishing workflow
 
 mozilla-central has support for automatically publishing and including a local development version of application-services in the build.
 This is supported for most of the Android targets available in mozilla-central including Fenix - this
 doc will focus on Fenix, but the same general process is used for all. The workflow is:
 
-1. Ensure you have a regular build of Fenix working from mozilla-central and that you've done a `./mach build`
+## Pre-requisites:
 1. Ensure you have a regular [build of application-services working](../building.md).
-1. Edit (or create) the file `local.properties` - this can be in the root of the mozilla-central checkout,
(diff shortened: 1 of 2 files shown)


badboy:
Thanks! I'll double-check the Glean docs on this.
Indeed the `org.gradle.configuration-cache=true` was causing issues for me too, but I also had other issues where it didn't pick up the local version at all (but that's unrelated, I need to re-test that)

docs/howtos/locally-published-components-in-fenix.md
```diff
@@ -8,44 +8,63 @@ from there.
...
+## Pre-requisites:
+1. Ensure you have a regular [build of application-services working](../building.md).
+1. Disable the gradle cache in mozilla-central - edit `./gradle.properties`, comment out `org.gradle.configuration-cache=true`
+1. Ensure you have a regular build of Fenix from mozilla-central testable in Android Studio or an emulator.
+
+## Setup 2 x `local.properties`
+
+Edit (or create) the file `local.properties` in each of the repos:
+
+### app-services
```
gruberb, line 26: I am not sure if we want to add this here, but I had to make sure that:

- add the `ndk` and `sdk` path to your `local.properties`:
```
# EXAMPLE PATH
ndk.dir=/Users/USERNAME/.mozbuild/android-ndk-r28b
sdk.dir=/Users/USERNAME/.mozbuild/android-sdk-macosx
```

docs/howtos/locally-published-components-in-fenix.md
```diff
@@ -8,44 +8,63 @@ from there.
...
+find your local checkout of application-services by adding a line like:
+
+`autoPublish.application-services.dir=path/to/your/checkout/of/application-services`
+
+Note that the path can be absolute or relative from `local.properties`. For example, if `application-services`
+and `mozilla-central` are at the same level, and you are using a `local.properties` in the root of mozilla-central,
+the relative path would be `../application-services`
 
-   `autoPublish.application-services.dir=path/to/your/checkout/of/application-services`
+## Build and test your Fenix again.
```
gruberb, line 59: Not sure, but should we add that:

If you have trouble building `m-c` inside Android Studio, try:

1. Have a virtual device running inside Android Studio
2. Do `./mach build` 
3. And then `./mach run --app=fenix`, which installs and runs Fenix inside the running emulator.

docs/howtos/locally-published-components-in-fenix.md
mhammond: I've never really worked those out - I *used* to have `sdk.dir` but it's been commented out for the last few months. I *do* have `ANDROID_HOME` set, so I suspect that's why I don't need it.

This is also kinda covered in https://github.com/mozilla/application-services/blob/main/docs/building.md#building-for-fenix - maybe we should just link to that?

docs/howtos/locally-published-components-in-fenix.md
mhammond: huh - I didn't know that worked :) I thought that `run` command always started a stand-alone emulator. Ideally though we'd find canonical docs for building Fenix and link to that, but adding this here for now sgtm

gruberb (approved):
I left a few comments. Looks good, although I still run into issues for fresh builds

- Not be able to run `m-c` with the local Application Services inside Android Studio
- Not be able to run `./gradlew assembleDebug` on macOS

These don't seem to be issues, I wonder however if we should collect all these common issues into another trouble shooting doc. 

Also, should we implement a sane check if we run with local app-services in a follow up PR? A log somewhere?

PR #6733, opened 2025-04-28
More gradle tweaks for mozilla-central

mhammond:


mhammond:
So this patch seems to be working fine on both a-s and m-c. There are a few stacked commits to try and capture the fixes Nick already started here, but I'll land everything squashed. I can't see a good reason not to get this landed to make the monorepo work easier. cc @moztcampbell

.buildconfig-android.yml
```diff
@@ -130,13 +130,6 @@ projects:
     - name: full-megazord-libsForTests
       type: jar
     description: Megazord containing all features
-  tooling-nimbus-gradle:
```
ncalexan: Won't this stop publishing `tooling-nimbus-gradle`?  Is that not needed any longer?  It's certainly referenced in-tree: https://searchfox.org/mozilla-central/search?q=tooling-nimbus-gradle&path=

build-scripts/component-common.gradle
```diff
@@ -99,18 +108,29 @@ ext.configureUniFFIBindgen = { crateName ->
...
-                def libraryPath = "${mozconfig.topobjdir}/dist/bin/libmegazord.so"
-                megazordNative = "${mozconfig.topobjdir}/dist/host/bin/embedded-uniffi-bindgen"
-                exec {
-                    workingDir project.rootDir
-                    commandLine "${mozconfig.topobjdir}/dist/host/bin/embedded-uniffi-bindgen", 'generate', '--library', libraryPath, "--crate", crateName, '--language', 'kotlin', '--out-dir', uniffiOutDir.get(), '--no-format'
-                }
-            } else {
+    if (gradle.hasProperty("mozconfig")) {
+        // in moz-central we can use an `Exec` task because we can assume the bindgen tool has already been built.
+        generateUniffiBindings = tasks.register("generateUniffiBindings", Exec) {
```
ncalexan, line 118: This will need adjusting to accommodate the toolchain task under `a-g-d`, but that can follow.

build-scripts/component-common.gradle
```diff
@@ -99,18 +108,29 @@ ext.configureUniFFIBindgen = { crateName ->
...
+            commandLine "${gradle.mozconfig.topobjdir}/dist/host/bin/embedded-uniffi-bindgen"
+            args 'generate', '--library', libraryPath, "--crate", crateName, '--language', 'kotlin', '--out-dir', uniffiOutDir.get(), '--no-format'
+
+            outputs.dir uniffiOutDir
+            // Re-generate when the native megazord library is rebuilt
+            inputs.files megazordNative
+            // Re-generate if our uniffi-bindgen tooling changes.
+            inputs.dir "${project.appServicesRootDir}/tools/embedded-uniffi-bindgen/"
+        }
+    } else {
```
ncalexan, line 132: I don't think you need to block on this, but this doesn't make much sense to me.  You have a task that runs `exec` in `doFirst` -- just in a configuration-caching unfriendly way.  That's what `Exec` tasks are good at.

build-scripts/component-common.gradle
```diff
@@ -138,9 +159,12 @@ ext.configureUniFFIBindgen = { crateName ->
             def compileTask = tasks["compile${variantName}Kotlin"]
 
             compileTask.dependsOn(generateUniffiBindings)
-            if (!gradle.hasProperty("mozconfig")) { // XXX - needs uniffi-bindgen love in m-c
+            if (!gradle.hasProperty("mozconfig")) {
```
ncalexan, line 163: If we're generating Kotlin, don't we need to register the generating task regardless of in-tree or out?

ncalexan (approved):
Some things here are still not obvious to me, but perfect is the enemy of the good -- this certainly takes some steps forward.

build-scripts/component-common.gradle
mhammond: yeah, that's the existing code we can kill when we land this.

.buildconfig-android.yml
mhammond: seems like this all works either way now :) So yeah, I left it in.

build-scripts/component-common.gradle
mhammond: in tree it works due to the blocks above.

PR #6743, opened 2025-04-30
More lint fixups from mozilla-central.

mhammond:
I got some pushback from the linting team about excluding these from linting in m-c, so here we are.

All these were auto-fixed by the linter *except for script.js* which has some hand-edits.


freshstrangemusic (approved):
rs+

PR #6746, opened 2025-05-05
Add first version of context_id component.

mikeconley:

  - Note:

### DIFF
--- a/components/context_id/src/callback.rs
+++ b/components/context_id/src/callback.rs
@@ -1,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::ApiResult;
+
+#[uniffi::export(callback_interface)]
+pub trait ContextIdCallback: Sync + Send {
+    fn persist(&self, context_id: String, creation_date: i64) -> ApiResult<()>;
+    fn rotated(&self, old_context_id: String) -> ApiResult<()>;
+}
+
+pub struct DefaultContextIdCallback;
+impl ContextIdCallback for DefaultContextIdCallback {
+    fn persist(&self, _context_id: String, _creation_date: i64) -> ApiResult<()> {
+        Ok(())
+    }
+    fn rotated(&self, _old_context_id: String) -> ApiResult<()> {
+        Ok(())
+    }
+}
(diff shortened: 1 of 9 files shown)


components/context_id/src/callback.rs
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[uniffi::export(callback_interface)]
+pub trait ContextIdCallback: Sync + Send {
+    fn persist(&self, context_id: String, creation_date: i64);
+    fn rotated(&self, old_context_id: String);
```
bendk, line 8: I think these methods should return a `Result` and the `E` side should implement `From<uniffi::UnexpectedUniFFICallbackError>`.  Otherwise, if the foreign side throws for some unexpected reason the Rust code will panic.  The bottom part of https://mozilla.github.io/uniffi-rs/latest/foreign_traits.html has more details.

bendk (changes requested):
This looks great to me.  I really like the API choice of not exposing errors to the consumers, since they don't have a great way to handle them.  I've been moving more and more in that direction myself.

The only blocker I have is having the callback return a `Result` and the `E` side implement `From<uniffi::UnexpectedUniFFICallbackError>`.

components/context_id/src/callback.rs
mikeconley: I've gone ahead and tried to log errors when these callbacks fail. Does that seem like the right move?

components/context_id/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    InvalidTimestamp { timestamp: i64 },
+
+    #[error("URL parse error: {0}")]
+    UrlParseError(#[from] url::ParseError),
+
+    #[error("Viaduct error: {0}")]
+    ViaductError(#[from] viaduct::Error),
+
+    #[error("UniFFI callback error: {0}")]
+    UniFFICallbackError(#[from] uniffi::UnexpectedUniFFICallbackError),
```
bendk, line 28: This looks good to me, although it's a bit weird to think through the flow.  I feel like we could have a better story here, maybe callbacks like yours could return `Result<(), uniffi::UnexpectedUniFFICallbackError`>` and not have to do any of this conversion.

In any case, this is just me thinking out loud how this could be improved from the UniFFI side of things.  I don't think you need to change any of this code.

bendk (approved):
This all looks great, thanks for the updates.

components/context_id/src/error.rs
mikeconley: Amazing, thanks!

PR #6753, opened 2025-05-11
Bug 1965166: Introduce subject type for Yelp suggestion

dadaa:

  - Note:

### DIFF
--- a/components/suggest/src/yelp.rs
+++ b/components/suggest/src/yelp.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-use rusqlite::types::ToSqlOutput;
+use rusqlite::types::{FromSql, FromSqlResult, ToSqlOutput, ValueRef};
 use rusqlite::{named_params, Result as RusqliteResult, ToSql};
 use sql_support::ConnExt;
 use url::form_urlencoded;
@@ -13,6 +13,7 @@ use crate::{
     provider::SuggestionProvider,
     rs::{DownloadedYelpSuggestion, SuggestRecordId},
     suggestion::Suggestion,
+    suggestion::YelpSubjectType,
     Result, SuggestionQuery,
 };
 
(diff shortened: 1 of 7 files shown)


components/suggest/src/yelp.rs
```diff
@@ -270,36 +301,36 @@ impl SuggestDao<'_> {
     /// (
     ///   String: The keyword in DB (but the case is inherited by query).
     ///   bool: Whether or not the keyword is exact match.
+    ///   SubjectType: The subject type.
     ///   &[&str]: Words after removed matching subject.
     /// )
```
0c0w3, line 306: And then please update this comment for the new struct type. I would probably move these comments up to the struct definition itself.

0c0w3 (changes requested):
Thanks, looks good! I'd like to do one more review pass.

General comment: `Specific` is a little unclear. It could mean different things. How about `BusinessName`?

dadaa:
Thank you very much! I will address your comments.

dadaa:
> General comment: `Specific` is a little unclear. It could mean different things. How about `BusinessName`?

Ah, I see. Thanks. If we rename to `BusinessName`, can I rename `Service` to `ServiceName`?
Ah, but these variables are `SubjectType`. So the sound of `XxxName` might seem a little strange?
So, how about using just `Business` and `Service`? What do you think?

0c0w3:
> So, how about using just `Business` and `Service`? What do you think?

Sounds good!

dadaa:
Hm, I am getting same warning of clippy, but if I remove the line, can't compile with following error.
```
error[E0433]: failed to resolve: use of undeclared type `YelpSubjectType
--> components/suggest/src/store.rs:2259:27
```
I don't understand the reason yet..

0c0w3:
It's because you import `suggestion::YelpSubjectType` in the main part of the file but only use it in the tests, specifically inside `pub(crate) mod tests`. Inside that `tests` module, we do `use super::*`, which means the `tests` module can also use the imports in the main part of the file, including `suggestion::YelpSubjectType`. That's why everything compiles and runs OK. But clippy sees that you don't use `suggestion::YelpSubjectType` in the main part of the file, and that's why it warns.

If you move the import of `suggestion::YelpSubjectType` into the `tests` module, that will fix it.

dadaa:
> It's because you import `suggestion::YelpSubjectType` in the main part of the file but only use it in the tests, specifically inside `pub(crate) mod tests`. Inside that `tests` module, we do `use super::*`, which means the `tests` module can also use the imports in the main part of the file, including `suggestion::YelpSubjectType`. That's why everything compiles and runs OK. But clippy sees that you don't use `suggestion::YelpSubjectType` in the main part of the file, and that's why it warns.

Oh, thanks!! It works!

PR #6754, opened 2025-05-14
add logins store api method `add_with_record`

jo:
The new method
```
add_with_record(&self, entry: LoginEntry, record: RecordFields)
```
is intended to be used during migration on desktop.

  - Note:


components/logins/src/store.rs
```diff
@@ -49,6 +57,23 @@ fn create_sync_engine(
...
+    match enc_login {
+        Ok(enc_login) => match enc_login.decrypt(encdec) {
+            Ok(login) => BulkResultEntry::Success { login },
+            Err(error) => BulkResultEntry::Error {
+                message: error.to_string(),
+            },
+        },
+        Err(error) => BulkResultEntry::Error {
+            message: error.to_string(),
+        },
```
bendk, line 73: This branch indicates that we can't decrypt the data we just encrypted right?  I think you should log a warning -- it seems like it's an indication of a bigger issue.

components/logins/src/login.rs
```diff
@@ -342,6 +342,19 @@ pub struct RecordFields {
     pub times_used: i64,
 }
 
+/// A login together with record fields, handed over to the store API; ie a login persisted
+/// elsewhere, useful for migrations
+pub struct LoginEntryWithRecordFields {
+    pub entry: LoginEntry,
+    pub record: RecordFields,
```
bendk, line 349: "RecordFields" is a pre-existing term, but I've never liked it and now it's becoming part of the public API.  If you have a better name, I'd be happy to switch to that.  If not, how about renaming it to `LoginRecordFields`.  That avoids some issues on Swift which doesn't have any modules / name spaces.

bendk (changes requested):
This looks great to me.  The only thing I'm really concerned about is finding a better name for `RecordFields`.  At the very least it should be `LoginRecordFields`, otherwise there's a high chance for confusion or name conflicts on Swift.

components/logins/src/db.rs
```diff
@@ -351,6 +351,74 @@ impl LoginDb {
...
+                        times_used: 1,
+                    },
+                }
+            })
+            .collect();
+
+        self.add_many_with_records(entries_with_records, encdec)
+    }
+
+    /// Adds multiple logins with records within a single transaction and returns the successfully saved logins.
```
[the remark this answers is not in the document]
jo: I improved the documentation. It's not absolutely necessary that it was saved beforehand. In tests, for example, you can simply compile the metadata yourself.

components/logins/src/store.rs
jo: Changed the logging to a `warn!` after rebasing

PR #6758, opened 2025-05-20
Bug 1967617 - Replace individual `Geoname::admin{level}_code` properties with one `admin_division_codes` map

0c0w3:
This builds on #6745 and replaces the individual `Geoname::admin{level}_code` properties with a single `admin_division_codes` map. A few reasons:

* I'm working on another patch that adds a way for consumers to fetch localized/alternate names for geonames, including alternates for a geoname's admin divisions. I'm introducing another struct for that, and I don't want to repeat `admin{level}_name` for each level in that struct.
* I like the idea of not baking the max number of levels into the public API.
* Not a big deal, but a map is similar to how AccuWeather's location API handles admin divisions, although it uses an array of objects instead of one object.

  - Note:

### DIFF
--- a/components/suggest/src/benchmarks/geoname.rs
+++ b/components/suggest/src/benchmarks/geoname.rs
@@ -66,15 +66,12 @@ impl BenchmarkWithInput for GeonameBenchmark {
 pub fn all_benchmarks() -> Vec<(&'static str, GeonameBenchmark)> {
     let ny_state = Geoname {
         geoname_id: 5128638,
-        geoname_type: GeonameType::Admin1,
+        geoname_type: GeonameType::AdminDivision { level: 1 },
         name: "New York".to_string(),
         feature_class: "A".to_string(),
         feature_code: "ADM1".to_string(),
         country_code: "US".to_string(),
-        admin1_code: Some("NY".to_string()),
-        admin2_code: None,
-        admin3_code: None,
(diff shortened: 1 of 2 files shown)


0c0w3:
This doesn't work with UniFFI because the `HashMap` is converted to a JS object, and the keys of JS objects are strings. So when UniFFI checks that the keys are ints, it throws an error.

Not sure what to do. Either use an array of objects like AccuWeather does -- `[{ level: 1, code: "ENG" }]` -- or make a UniFFI patch that converts `HashMap` to JS `Map` instead of plain objects, or something.

0c0w3:
I'm going to submit a UniFFI patch ASAP. This PR aside, using a plain JS object for a map is a straight up bug. So let's go forward with this PR assuming that will be fixed, but let me know if you think using a map for the admin divisions is a bad idea aside from that.

ncloudioj (approved):
:shipit:

0c0w3:
> This doesn't work with UniFFI because the `HashMap` is converted to a JS object, and the keys of JS objects are strings. So when UniFFI checks that the keys are ints, it throws an error.

There was a bug already on file: [Bug 1809459 - Use Map instances for UniFFI maps to allow non-string keys.](https://bugzilla.mozilla.org/show_bug.cgi?id=1809459)

PR #6760, opened 2025-05-22
Bug 1967899 - Add a way for consumers to fetch localized names and other alternates for geonames

0c0w3:
This adds a `fetch_geoname_alternates` function that returns a struct containing localized names and other alternates for a geoname, its country, and its admin divisions.

It also changes the RS schema for alternates records: Each alternate can now be a string (as before) or an object with `is_preferred` and `is_short` booleans. These bools come straight from the geonames data, and we use them to help decide which alternates to return from `fetch_geoname_alternates`.

  - Note:

### DIFF
--- a/components/suggest/src/geoname.rs
+++ b/components/suggest/src/geoname.rs
@@ -17,6 +17,7 @@ use serde::Deserialize;
 use sql_support::ConnExt;
 use std::{
     borrow::Cow,
+    cell::OnceCell,
     collections::HashMap,
     hash::{Hash, Hasher},
 };
@@ -61,11 +62,7 @@ pub struct Geoname {
     /// The geoname type. This is derived from `feature_class` and
     /// `feature_code` as a more convenient representation of the type.
     pub geoname_type: GeonameType,
-    /// This is pretty much the place's canonical name. Usually there will be a
-    /// row in the alternates table with the same name, but not always. When
-    /// there is such a row, it doesn't always have `is_preferred_name` set, and
-    /// in fact fact there may be another row with a different name with
-    /// `is_preferred_name` set.
+    /// The place's primary name.
     pub name: String,
     /// ISO-3166 two-letter uppercase country code, e.g., "US".
     pub country_code: String,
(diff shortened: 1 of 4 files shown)


0c0w3:
It would be nice from an API POV if `fetch_geonames` returned the alternates instead of the consumer having to call a separate function. The reason I didn't do it that way is performance. Weather, for example, calls `fetch_geonames` a ton as it's parsing the query. Most of those calls don't need the alternates. For external `suggest` consumers, it's less of an issue, so we could have a separate version of `fetch_geonames` that also fetches the alternates and returns them too, and then consumers would only use it. But even then, the desktop Yelp implementation calls `fetch_geonames` up to two times per query, once to fetch the city and once to fetch the region, and there's no need to pay the extra cost of fetching alternates for both fetches.

If you have thoughts on that I'd be happy to hear them!

That said, I think `Suggestion::Weather` probably should include the alternates since it already includes the `Geoname`, and there are no other geonames fetches the consumer could conceivably make after that for the same query. So I'll probably make another PR on top of this one, assuming this one is OK.

0c0w3:
> That said, I think `Suggestion::Weather` probably should include the alternates since it already includes the `Geoname`, and there are no other geonames fetches the consumer could conceivably make after that for the same query. So I'll probably make another PR on top of this one, assuming this one is OK.

Oops, no, from a perf perspective it doesn't make sense for weather to return alternate names either, since it can return all cities that match a small prefix, and that can be a big list. That's what I get for commenting late at night.

components/suggest/src/geoname.rs
```diff
@@ -207,8 +232,42 @@ struct DownloadedGeonamesAlternatesAttachment {
...
     language: String,
-    /// Tuples of geoname IDs and their localized names.
-    names_by_geoname_id: Vec<(GeonameId, Vec<String>)>,
+    /// Tuples of geoname IDs and their alternate names.
+    alternates_by_geoname_id: Vec<(GeonameId, Vec<DownloadedGeonamesAlternate<String>>)>,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(untagged)]
+enum DownloadedGeonamesAlternate<S: Deref<Target = str>> {
```
ncloudioj, line 241: In general, `<S: AsRef<str>>` is more flexible than `<S: Deref<Target = str>>`. As a result, the former trait constraint would cover more types than the latter. This [comment](https://users.rust-lang.org/t/understanding-the-perils-of-deref/47958/7) summarizes the differences and subtleties quite well.  
```suggestion
enum DownloadedGeonamesAlternate<S: AsRef<str>> {
```

components/suggest/src/geoname.rs
```diff
@@ -207,8 +232,42 @@ struct DownloadedGeonamesAlternatesAttachment {
...
+        is_preferred: Option<bool>,
+        is_short: Option<bool>,
+    },
+}
+
+impl<S: Deref<Target = str>> DownloadedGeonamesAlternate<S> {
+    fn name(&self) -> &str {
+        match self {
+            Self::Name(name) => name,
+            Self::Full { name, .. } => name,
```
ncloudioj, line 254:
```suggestion
impl<S: AsRef<str>> DownloadedGeonamesAlternate<S> {
    fn name(&self) -> &str {
        match self {
            Self::Name(name) => name.as_ref(),
            Self::Full { name, .. } => name.as_ref(),
```

components/suggest/src/geoname.rs
```diff
@@ -562,26 +763,36 @@ struct GeonameAlternateInsertStatement<'conn>(rusqlite::Statement<'conn>);
...
+                is_preferred,
+                is_short
+            )
+            VALUES(?, ?, ?, ?, ?, ?)
+            "#,
         )?))
     }
 
-    fn execute(
+    fn execute<S: Deref<Target = str>>(
```
ncloudioj, line 780:
```suggestion
    fn execute<S: AsRef<str>>(
```

ncloudioj (approved):
Looks good, just a minor suggestion.

I also agreed that it makes more sense to not include alternates in the `fetch_geonames()`.

0c0w3:
Thanks Nan!

PR #6770, opened 2025-05-26
Remove Pocket suggest

kpatenio:
This PR is part of an effort to decouple Pocket-related code from Firefox Desktop; see meta bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1959497. Although these changes only apply to application services, they're needed so that we can eventually update the vendor; see https://bugzilla.mozilla.org/show_bug.cgi?id=1968325. cc @0c0w3 

Summary:
- removed all mentions of pocket from code and documentation
- removed tests relating to pocket
- removed metrics relating to pocket
- repurposed `pocket::split_keyword` to `util::split_keyword`

Testing:
- Verified using `./automation/tests.py changes`
- Added a test in `util.rs` to ensure `split_keyboard` works as expected

I admit that Rust, Suggest and Application Services are new to me, so please let me know if I'm missing something important.

  - Note:


0c0w3 (commented):
Thanks, this looks great! I only have a couple of requests.

components/suggest/src/provider.rs
```diff
@@ -38,12 +38,11 @@ pub enum SuggestionProvider {
...
-    Yelp = 5,
-    Mdn = 6,
-    Weather = 7,
-    Fakespot = 8,
-    Dynamic = 9,
+    Yelp = 4,
+    Mdn = 5,
+    Weather = 6,
+    Fakespot = 7,
+    Dynamic = 8,
```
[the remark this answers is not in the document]
kpatenio: Sounds good. I reverted the value changes and added a comment to preserve them whenever we add/remove a provider.

components/suggest/src/db.rs
```diff
@@ -583,6 +581,12 @@ impl<'a> SuggestDao<'a> {
         Ok(suggestions)
     }
 
+    /// Split the keyword by the first whitespace into the prefix and the suffix.
+    /// Return an empty string as the suffix if there is no whitespace.
+    fn split_keyword(keyword: &str) -> (&str, &str) {
+        keyword.split_once(' ').unwrap_or((keyword, ""))
+    }
+
```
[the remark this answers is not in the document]
kpatenio: Thanks for the suggestion (no pun intended). I moved the function to utils.rs as requested.

components/suggest/src/provider.rs
```diff
@@ -32,13 +32,13 @@ pub(crate) const DEFAULT_INGEST_PROVIDERS: [SuggestionProvider; 5] = [
 ];
 
 /// A provider is a source of search suggestions.
+/// Please preserve values after removing or adding providers.
```
0c0w3: Thanks for adding this! Would you mind briefly mentioning the reason? (Because these ints are stored in the DB as part of the keys for provider configs)

components/suggest/src/provider.rs
```diff
@@ -93,12 +91,11 @@ impl SuggestionProvider {
...
-            5 => Some(Self::Yelp),
-            6 => Some(Self::Mdn),
-            7 => Some(Self::Weather),
-            8 => Some(Self::Fakespot),
-            9 => Some(Self::Dynamic),
+            4 => Some(Self::Yelp),
+            5 => Some(Self::Mdn),
+            6 => Some(Self::Weather),
+            7 => Some(Self::Fakespot),
+            8 => Some(Self::Dynamic),
```
0c0w3: These need to stay the same too, for the same reason. IMO no comment here is necessary since it should be implied by the one you added above. Or if you'd like, a comment that they should match the similar `enum` values above would be great.

0c0w3 (dismissed):
Thanks! I have a couple comments but I'll go ahead and approve. Also I just bitrotted your `util.rs` change in #6774, sorry about that, but it shouldn't be too hard to resolve.

components/suggest/src/provider.rs
kpatenio: Good point. Updated to undo the value changes

components/suggest/src/provider.rs
kpatenio: Oops, yes thanks for the reminder. I just added an additional comment to clarify why this is needed.

kpatenio:
@0c0w3 the mergify bot reverted your review since I modified the pull request. Would you mind taking a look again?

0c0w3 (dismissed):
Thanks again!

0c0w3:
@kpatenio Looks like there's still a merge conflict, would you be able to resolve it?

kpatenio:
> @kpatenio Looks like there's still a merge conflict, would you be able to resolve it?

Sorry about that. Turns out I'm quite rusty with Git, so please let me know if there's another issue with merging the changes. (And apologies again, this mean another review!)

PR #6771, opened 2025-05-28
Add filter_adult module

mikeconley:

  - Note:


components/filter_adult/src/adult_set.rs
```diff
@@ -0,0 +1,8767 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+pub static ADULT_SET: [[u8; 16]; 2920] = [
```
mhammond, line 6: this seems, um, alot :) I assume it's generated - should we describe how it is generated?

(Also not sure it makes sense here, but https://doc.rust-lang.org/std/macro.include_str.html might be useful as an alternative - this would mean the Rust code ends up getting a very long string to parse at runtime, but the format of the string therefore becomes more flexibile (eg, you could imagine each line has a single base64 entry or similar)

mikeconley: For some reason, it wouldn't accept HashSet::from(adult_set.chunks_exact(16)), so I had to iterate the chunks and insert them manually.

mikeconley: Yes, I can turn this into adult_set.bin, that's no problem.

components/filter_adult/src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum ApiError {
+    #[error("Other error: {reason}")]
+    Other { reason: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Failed to construct")]
+    FailedToConstruct(String),
```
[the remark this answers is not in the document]
mikeconley: I got rid of the unused errors.

components/filter_adult/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+    hashes: HashSet<[u8; 16]>,
+}
+
+impl FilterAdultInner {
+    pub fn new() -> Result<Self> {
+        let adult_set_bytes = include_bytes!("adult_set.bin");
+        let mut hashes = HashSet::new();
+        for chunk in adult_set_bytes.chunks_exact(16) {
+            let mut arr = [0; 16];
+            arr.copy_from_slice(chunk);
```
bendk, line 45: I think this is needed to convert the reference to an owned array.  You might be able to get away without it by using `hashes: HashSet<&'static [u8; 16]>`.  Either way is fine with me though.

bendk (approved):
Looks great to me.  The `.bin` feels way nicer.

PR #6773, opened 2025-05-30
for MPP-4194: wip: start uniffi code for Relay component

groovecoder:

- [x] **Breaking changes**:  This PR follows our [breaking change policy](https://github.com/mozilla/application-services/blob/main/docs/howtos/breaking-changes.md)
     - [x] This PR has no breaking API changes
- [x] **Tests**: This PR includes thorough tests ~or an explanation of why it does not~
- [x] **Changelog**: This PR includes ~a changelog entry in [CHANGELOG.md](../CHANGELOG.md) or~ an explanation of why it does not need one
     - [x] No Swift or Kotlin consumers yet.
     - [x] No new dependencies

### DIFF
--- a/components/relay/Cargo.toml
+++ b/components/relay/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+error-support = { path = "../support/error" }
 log = "0.4"
 serde = { version = "1", features=["derive"] }
 serde_json = "1"
--- a/components/relay/Cargo.toml
+++ b/components/relay/Cargo.toml
@@ -10,6 +10,7 @@ serde_json = "1"
 thiserror = "1.0"
 url = "2"
 viaduct = { path = "../viaduct" }
+uniffi = "0.29.0"
 
 [dev-dependencies]
 env_logger = { version = "0.10", default-features = false, features = ["humantime"] }
@@ -17,3 +18,6 @@ expect-test = "1.4"
 mockall = "0.11"
 mockito = "0.31"
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
+
+[build-dependencies]
+uniffi = { version = "0.29.0", features=["build"]}
--- a/components/relay/src/error.rs
+++ b/components/relay/src/error.rs
@@ -4,23 +4,41 @@
 
 pub type Result<T> = std::result::Result<T, Error>;
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, thiserror::Error, uniffi::Error)]
 pub enum Error {
     #[error("Network error: {0}")]
-    Network(#[from] viaduct::Error),
+    Network(String),
 
     #[error("UTF-8 error: {0}")]
-    Utf8(#[from] std::string::FromUtf8Error),
+    Utf8(String),
 
     #[error("URL parse error: {0}")]
-    UrlParse(#[from] url::ParseError),
+    UrlParse(String),
 
     #[error("JSON error: {0}")]
-    Json(#[from] serde_json::Error),
+    Json(String),
 
     #[error("Status error: {0}")]
-    Status(#[from] viaduct::UnexpectedStatus),
+    Status(String),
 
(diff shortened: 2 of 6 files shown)


components/relay/src/error.rs
```diff
@@ -4,23 +4,41 @@
...
+    fn from(err: viaduct::Error) -> Self {
+        Error::RelayApi(format!("Viaduct error: {}", err))
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(err: serde_json::Error) -> Self {
+        Error::RelayApi(format!("JSON error: {}", err))
+    }
+}
```
bendk, line 44: I'd recommend keeping the old error code, which uses separate variants for each of these errors.  It's often useful to distinguish the cases when you're handling them internally.

OTOH, consumers of your component might not care about these details.  In that case, you can define a second error enum, for the public API along with logic to convert between them.  The other thing we typically do is report errors to sentry during the internal -> public error conversion, if we see an internal error that we don't expect.  Check out [errors.rs](https://github.com/mozilla/application-services/blob/main/components/example/src/error.rs) from the example crate for details.

components/relay/Cargo.toml
```diff
@@ -10,10 +10,18 @@ serde_json = "1"
...
 mockall = "0.11"
 mockito = "0.31"
 viaduct-reqwest = { path = "../support/viaduct-reqwest" }
+
+[build-dependencies]
+uniffi = { version = "0.29.0", features=["build"]}
+
+[lib]
+name = "relay"
+crate-type = ["cdylib", "staticlib"]
```
bendk, line 27: You probably also want `rlib` here if you're going to list the crate types.  That way the example can still use it.

Stepping back, how are you going to distribute this component?  In general, we bundle all our components into one of the megazord libraries and ship those to FF android/ios/desktop.  If you're doing the same thing, you can leave `crate-type` out, since the megazord's crate type is what matters in those cases.

bendk (approved):
Looks like a good start to me.  I think you might want update your error handling strategy, but I'll leave that decision to you.  I definitely recommend checking out the example crate for inspiration there.

Maybe you could feed the example component to cursor/claude and see how it generates the boilerplate.

There's definitely more work to do here, but I'm approving because this isn't being shipped yet and this seems like a step forward to me.

bendk:
See https://github.com/mozilla/application-services/blob/main/docs/howtos/adding-a-new-component.md for how to start generating the Kotlin/Swift bindings.  You shouldn't have to check the bindings code in, that's all autogenerated.  However, there is some boilerplate that it will generate -- some of that you already have.

components/relay/Cargo.toml
groovecoder: I'll remove this `[lib]` section completely - most other components don't include it in their `Cargo.toml`.

components/relay/src/error.rs
groovecoder: I added https://github.com/mozilla/application-services/pull/6773/commits/8a2d05d098ab252a4fda1ad2b5b8b766d0cde573 defining the second error enum for the public API and logic to convert between them using the code from the example errors. Check that I did it alright?

components/relay/src/error.rs
```diff
@@ -2,25 +2,56 @@
...
+                .log_warning()
+            }
+            Error::RelayApi(detail) => ErrorHandling::convert(ApiError::RelayApi {
+                detail: detail.clone(),
+            })
+            .report_error("relay-api-error"),
+            _ => ErrorHandling::convert(ApiError::Other {
+                reason: self.to_string(),
+            })
+            .report_error("relay-unexpected-error"),
```
bendk, line 54: You're going to be reporting an error to Sentry here and in the `relay-api-error` case above.  That seems fine to me, just want double with you that that's your intention.  The stakes are pretty low, it's easy to iterate on these.

groovecoder: Yeah, I think having these show up in Sentry will be helpful. Will they all show up in the Sentry project for the mobile browser? Or does A-S have its own sentry project too?

PR #6774, opened 2025-05-30
Bug 1969522 - Make weather suggestions do prefix matching only for candidates at the end of the query string

0c0w3:

  - Note:

### DIFF
--- a/components/suggest/src/util.rs
+++ b/components/suggest/src/util.rs
@@ -197,14 +197,14 @@ fn filter_map_chunks_recurse<T: Clone>(
             break;
         }
 
-        // Get the current chunk within the remaining query. Its char length is
+        // Get the current chunk within the remaining query. Its byte length is
         // the sum of the lengths of the words in the chunk + `chunk_size - 1`
         // spaces between the words. There will only be one space between each
         // word in `remaining_query` because `remaining_query` is normalized.
-        let chunk_char_len = remaining_words[..chunk_size]
+        let chunk_byte_len = remaining_words[..chunk_size]
             .iter()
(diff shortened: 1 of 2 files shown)


bendk (approved):
I don't have context on the intention, but the code change looks correct to me.  The tests suite is very extensive.

components/suggest/src/util.rs
```diff
@@ -204,12 +205,13 @@ fn filter_map_chunks_recurse<T: Clone>(
             .iter()
             .fold(chunk_size - 1, |memo, w| memo + w.len());
         let chunk = &remaining_query[..chunk_char_len];
```
ncloudioj: Not related to this PR, but I found `chunk_char_len` and the comment "char length" a little bit misleading during the review because char(acter) length and byte length mean different things to the utf-8 based strings in [Rust](https://doc.rust-lang.org/nightly/book/ch08-02-strings.html#slicing-strings).

Specifically, since `w.len()` returns the byte length of `w` (v.s. the char lengh `w.chars().count()`), calling it `chunk_byte_len` is perhaps more precise than `chunk_char_len` in this case and slicing it via `&remaining_query[..chunk_byte_len]` would be perfectly sound than `&remaining_query[..query_char_len]` b/c the latter could panic if `query_char_len` is a char length.

ncloudioj (approved):
Looks good, just a minor suggestion on naming.

components/suggest/src/util.rs
0c0w3: Great point, thanks Nan, I'll fix this now.

PR #6775, opened 2025-06-04
Document that `wipe_local` is a no-op for empty databases

bendk:
I'm planning on making use of that behavior on Android, so I wanted to document/test this.

  - Note:


components/logins/src/logins.udl
```diff
@@ -188,6 +188,17 @@ interface LoginStore {
     [Throws=LoginsApiError]
     boolean delete([ByRef] string id);
 
+    /// Clear out locally stored logins data
+    ///
+    /// If sync is enabled, then we will try to recover the data on the next sync.
+    ///
+    /// The main reason to call this is when regenerating a new encryption key.
+    /// In that case, there's no reason to keep around the local data sinc it can't be decrypted.
```
lougeniaC64: nit: since

components/logins/src/logins.udl
```diff
@@ -188,6 +188,17 @@ interface LoginStore {
...
+    /// Calling `wipe_local` is better than keeping around these un-decryptable logins, since we
+    /// might be able to recover the data via sync.
+    /// 
+    /// This is a no-op for freshly created databases, so it's safe to call this whenever a key is
+    /// generated.
```
lougeniaC64, line 201: Do you need to change the return type on the function from `void` below?

lougeniaC64 (approved):
Other than the two comments I left, LGTM!

components/logins/src/logins.udl
bendk: This is actually still `void`.  I only changed the signature for `LoginsDb.wipe_local`.  The higher-level `wipe_locals` don't return the row count.

lougeniaC64: This has little to no reflection on the approval but . . . what are you planning to do with the return value of the lower-level function? I might have missed something.

bendk: Right now it's just for the unit tests.  It's kind of silly, but when implementing this I got very concerned that we'd change things so that calling `wipe_local` right after creating a new logins store had some unintended effect.  I can't really think how that would happen, but in any case I tried to verify the no-op behavior into a test.

lougeniaC64: I can definitely understand taking baby steps with this. Thanks!

PR #6783, opened 2025-06-10
expose Logins `delete_local_records_for_remote_replacement(ids)`

jo:
to be used during desktop's rolling migration.

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Logins
 - add checkpoint API: `set_checkpoint(checkpoint)` and `get_checkpoint()` for desktop's rolling migration
+- expose `delete_local_records_for_remote_replacement(ids)` to be used during desktop's rolling migration
 
 [Full Changelog](In progress)
 
--- a/components/logins/src/logins.udl
+++ b/components/logins/src/logins.udl
@@ -214,6 +214,10 @@ interface LoginStore {
     [Throws=LoginsApiError, Self=ByArc]
     void delete_undecryptable_records_for_remote_replacement();
 
+    /// Locally delete many records identified by id. These deletion will not propagate via sync.
+    [Throws=LoginsApiError, Self=ByArc]
+    void delete_local_records_for_remote_replacement(sequence<string> ids);
+
     [Throws=LoginsApiError]
     void touch([ByRef] string id);
 
--- a/components/logins/src/store.rs
+++ b/components/logins/src/store.rs
@@ -178,6 +178,20 @@ impl LoginStore {
         Ok(())
     }
 
+    // Note we need to receive a vector of String here because `Vec<&str>` is not supported
+    // with UDL.
+    #[handle_error(Error)]
+    pub fn delete_local_records_for_remote_replacement(
+        self: Arc<Self>,
+        ids: Vec<String>,
+    ) -> ApiResult<()> {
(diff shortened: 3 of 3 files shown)


bendk:
Can you share some more details about the intent here?  How does this fit into the rolling migration strategy?  My main concern is what that sync could kick in right after calling this and the data could be re-synced.  I don't know if this would be okay or not.

bendk:
Also, let's pull in @lougeniaC64 who wrote the original code.

jo:
> Can you share some more details about the intent here? How does this fit into the rolling migration strategy? My main concern is what that sync could kick in right after calling this and the data could be re-synced. I don't know if this would be okay or not.

In the rolling migration discussed here, the Rust store is initially operated as a write-only mirror without synchronization so that telemetry data collection can be used to check the consistency of both stores in the field. Only in the second phase, when the Rust store becomes the primary store, is synchronization enabled in the Rust store and the rolling migration disabled.

bendk:
> > Can you share some more details about the intent here? How does this fit into the rolling migration strategy? My main concern is what that sync could kick in right after calling this and the data could be re-synced. I don't know if this would be okay or not.
> 
> In the rolling migration discussed here, the Rust store is initially operated as a write-only mirror without synchronization so that telemetry data collection can be used to check the consistency of both stores in the field. Only in the second phase, when the Rust store becomes the primary store, is synchronization enabled in the Rust store and the rolling migration disabled.

Ahh, I'm understanding better now.  I still wonder about the initial scenario, but on a longer timeline.  Could this be an issue:

Steps:
* User syncs all their logins
* User disconnects from FxA
* The rolling migration starts
* User deletes a login
* Rolling migration ends and the user is switched to the Rust store
* User connects to FxA

At this point, could the deleted login be re-synced and re-created?

I'm coming in late to this one, so please tell me if this isn't helpful.  The change will certainly expose `delete_local_records_for_remote_replacement`. I'd could just focus on that and approve this one.

lougeniaC64:
> Also, let's pull in @lougeniaC64 who wrote the original code.

I marked `delete_local_records_for_remote_replacement` as a public function for testing reasons, which is why I didn't expose it. I exposed `delete_undecryptable_records_for_remote_replacement` which retrieves the undecryptable records and then calls `delete_local_records_for_remote_replacement`. I think calling `delete_undecryptable_records_for_remote_replacement` would be better in that it keeps this logic consolidated in A-S and it prevents us from deviating on how we're identifying undecryptable records. But I'm also aware that some departure from the iOS approach may be necessary.

In regards to not wanting the deletions to interfere with syncing, on iOS this function is called right before a logins sync. If the verification fails, the sync is skipped (see [here](https://github.com/mozilla-mobile/firefox-ios/blob/bdb64824bc30300a63ce275b0762f7ac3250a5a0/firefox-ios/Providers/RustSyncManager.swift#L402) for details). It's worth noting that there's also logic to limit this call to once per user.

jo:
Thanks for your thoughts!

Yes, Ben, you're absolutely right, that's indeed a realistic scenario, which also highlights further problems: I think we'll have to migrate the existing deletion stubs as well. I'll see how that can be done and possibly extend the API accordingly.

Furthermore, I will implement a `delete_many(ids: Vec<String>)` method that deletes multiple records within a single transaction: https://github.com/mozilla/application-services/pull/6784

PR #6784, opened 2025-06-11
add Logins `delete_many(ids)`

jo:
for batch deletion within a single transaction

  - Note:


components/logins/src/db.rs
```diff
@@ -674,45 +674,61 @@ impl LoginDb {
...
+                         sync_status = {status_changed},
+                         is_deleted = 1,
+                         secFields = '',
+                         origin = '',
+                         httpRealm = NULL,
+                         formActionOrigin = NULL
+                     WHERE guid = :guid",
+                    status_changed = SyncStatus::Changed as u8
+                ),
+                named_params! { ":now_ms": now_ms, ":guid": id },
```
bendk: Suggestion (non-blocking): use `prepare` or `prepare_cached` to prepare the SQL statements outside the loop.  This will give a large speedup when there are many ids.

components/logins/src/db.rs
```diff
@@ -674,45 +674,61 @@ impl LoginDb {
...
-        )?;
+        let mut result = vec![];
 
-        // Mark the mirror as overridden
-        self.execute(
-            "UPDATE loginsM SET is_overridden = 1 WHERE guid = :guid",
-            named_params! { ":guid": id },
-        )?;
+        for id in ids {
+            let exists = self.exists(id)?;
```
bendk: Suggestion (non-blocking): `execute()` returns the number of rows inserted/changed/deleted.  You can probably use that and avoid this query.

components/logins/src/db.rs
jo: Great suggestion, thank you!

components/logins/src/db.rs
jo: I fear we need this query, because `execute()` seem to not give us enough information - at least according to [the docs](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.execute)
> On success, returns the number of rows that were changed or inserted or deleted (via sqlite3_changes).

Here the API assumes to
> Returns a list of Boolean values indicating whether the respective records already existed.

Don't know if we're actually using this, and if we could change the API?

bendk: I'm pretty sure, if the `UPDATE` query returns more than 0 rows changed, then would be a `true` for the respective record.  However, this is a side issue and you're just re-using the code that was already there.  If this is important, we can do it in a follow-up.

jo: you are right! How awesome. Updated.

jo: small note: had to add `is_deleted IS FALSE` to the update condition to preserve behavior with deleted records

PR #6787, opened 2025-06-12
Make `ContextIDComponent` constructor not throw

issammani:

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 ### Logins
 - add checkpoint API: `set_checkpoint(checkpoint)` and `get_checkpoint()` for desktop's rolling migration
 
+### Context ID
+- The `ContextIDComponent` constructor no longer throws an error. This is not a breaking change since neither iOS or Android implement this yet.
+- `rotate_context_id` is no longer public since consumers can use `force_rotation` instead.
+
 [Full Changelog](In progress)
 
 # v140.0 (_2025-05-23_)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Context ID
 - The `ContextIDComponent` constructor no longer throws an error. This is not a breaking change since neither iOS or Android implement this yet.
+- The `ContextIDComponent` can now synchronously invoke the rotation callback when it receives an invalid timestamp from callers; in such cases, it falls back to the current timestamp and forces an ID rotation.
 - `rotate_context_id` is no longer public since consumers can use `force_rotation` instead.
 
 [Full Changelog](In progress)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Context ID
 - The `ContextIDComponent` constructor no longer throws an error. This is not a breaking change since neither iOS or Android implement this yet.
-- The `ContextIDComponent` can now synchronously invoke the rotation callback when it receives an invalid timestamp from callers; in such cases, it falls back to the current timestamp and forces an ID rotation.
+- The `ContextIDComponent` constructor can now synchronously invoke the rotation callback when it receives an invalid timestamp from callers; in such cases, it falls back to the current timestamp and forces an ID rotation.
 - `rotate_context_id` is no longer public since consumers can use `force_rotation` instead.
 
 [Full Changelog](In progress)
--- a/components/context_id/src/lib.rs
+++ b/components/context_id/src/lib.rs
@@ -159,7 +159,7 @@ impl ContextIDComponentInner {
         Ok(self.context_id.clone())
     }
 
-    pub fn rotate_context_id(&mut self, now: DateTime<Utc>) {
+    fn rotate_context_id(&mut self, now: DateTime<Utc>) {
         let original_context_id = self.context_id.clone();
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 ### Logins
 - add checkpoint API: `set_checkpoint(checkpoint)` and `get_checkpoint()` for desktop's rolling migration
 
+### Context ID
+- The `ContextIDComponent` constructor no longer throws an error. This is not a breaking change since neither iOS or Android implement this yet.
```
mikeconley, line 11: We should also mention that it's possible for the constructor to call the rotation callback synchronously now.

bendk (commented):
The UniFFI-related changes look good to me.  I'll let mconley review the behavior changes.

components/context_id/src/lib.rs
```diff
@@ -109,34 +109,40 @@ impl ContextIDComponentInner {
...
+                (false, secs) if secs > 0 => (
+                    DateTime::<Utc>::from_timestamp(secs, 0).unwrap_or(now),
+                    false,
                     false,
                 ),
-                _ => (now, true),
-            }
-        };
+                // Pre-existing context ID with zero timestamp then use current time, no rotation needed
+                (false, 0) => (now, false, false),
```
mikeconley: Ah - so we're generating the timestamp here, so this second return value should be `true`

mikeconley: So we only caught this because I have an automated test on the Desktop side that happened to trip this - but it's something we should probably add coverage for... namely, that passing in a `0` for the creation timestamp means regenerating / persisting.

components/context_id/src/lib.rs
```diff
@@ -109,34 +109,40 @@ impl ContextIDComponentInner {
...
             callback_handle: RwLock::new(callback),
             mars_client,
             running_in_test_automation,
         };
 
+        if force_rotation {
+            instance.force_rotation(creation_timestamp)
+        }
+
         // We only need to persist these if we just generated one.
```
mikeconley: Since `force_rotation` causes a persist, we should probably have the second condition be an `else if` instead.

mikeconley (dismissed):
Thanks, @issammani! I think there's a small bug here, which I think has also revealed some missing test coverage. Please let me know if you have any questions, or tips on how to write that test (although I myself am a bit of a Rust noob).

components/context_id/src/lib.rs
issammani: Yes good catch. I think too many boolean values confused me here. I can add a test that fails now and we would pass by changing that to true.

components/context_id/src/lib.rs
issammani: Updated in [d1d5cbb](https://github.com/mozilla/application-services/pull/6787/commits/d1d5cbb54bed523aaa93070fccf48a38fa26e929)

CHANGELOG.md
issammani: Updated in [56d8703](https://github.com/mozilla/application-services/pull/6787/commits/56d8703060ef662600c7a414005ed936d72995fc)

components/context_id/src/lib.rs
issammani: Added more checks and a utility for spying on callbacks in [9aa4dc7](https://github.com/mozilla/application-services/pull/6787/commits/9aa4dc790d295c034f5563738cf30121f132a112)

issammani: @mikeconley let me know if we need additional test cases. Happy to add more.

components/context_id/src/lib.rs
```diff
@@ -109,34 +109,39 @@ impl ContextIDComponentInner {
...
-                secs if secs > 0 => (
-                    DateTime::<Utc>::from_timestamp(secs, 0)
-                        .ok_or(Error::InvalidTimestamp { timestamp: secs })?,
+        let (creation_timestamp, generated_creation_timestamp, force_rotation) =
+            match (generated_context_id, creation_timestamp_s) {
+                // We generated a new context ID then we need a fresh timestamp and no rotation needed
+                (true, _) => (now, true, false),
+                // Pre-existing context ID with valid timestamp then use timestamp and no rotation needed
+                (false, secs) if secs > 0 => (
+                    DateTime::<Utc>::from_timestamp(secs, 0).unwrap_or(now),
```
mikeconley: If we fail to unwrap this, then we use `now`, and then we _should_ force rotation though, right? We have no idea how old the context ID is, so we should probably just generate a new one and invalidate the old one... or is there no way that `from_timestamp(secs, 0)` can fail to unwrap with a positive i64?

issammani: So according to [the docs](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#panics) it will only use fail if we exceed a limit of 260, 000 years from CE; effectively never for any real UNIX epoch. With that being said, this is the stuff that one thinks would never happen but it does 😂 Updated in [7ab66bb](https://github.com/mozilla/application-services/pull/6787/commits/7ab66bb6cd260763aaeca65d55c13fffbb3ead3a) and added a test for that.

mikeconley (approved):
Okay, this works for me! Thanks!

PR #6790, opened 2025-06-13
Tracing speedup

bendk:
Added a callsite filter for the simple event layer.  This allows us to avoid the lock + map lookup for non-app-services crates.

I think this is about as good as we can do performance-wise:
* Non app-services crates are statically disabled, no lock/map lookup needed
* App-services crates are dynamically handled.  We need to do the lock/map lookup each time, since we don't know when outside crates will call `register_event_sink`.

  - Note:

### DIFF
--- a/components/support/tracing/src/lib.rs
+++ b/components/support/tracing/src/lib.rs
@@ -6,8 +6,91 @@ mod testing;
 #[cfg(feature = "testing")]
 pub use testing::{init_for_tests, init_for_tests_with_level};
 
-pub use layer::{register_event_sink, unregister_event_sink, SimpleEventLayer};
-pub use tracing::{debug, error, info, trace, warn};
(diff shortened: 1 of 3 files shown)


components/support/tracing/src/lib.rs
```diff
@@ -6,8 +6,91 @@ mod testing;
...
+// This will statically disable the layer for events from outside crates that use tracing.
+// This way we don't pay a performance penalty for those events.
+// See `SimpleEventFilter` for details.
+
+#[macro_export]
+macro_rules! trace {
+    (target: $target:expr, $($tt:tt)*) => {
+        $crate::tracing::trace!(
+        target: $target,
+        app_services = true,
```
mhammond: How do you feel about making this just `tracing_support`? I fear `app_services` will not age well after we are in the monorepo and these crates end up spread around the tree.

mhammond (approved):
huh, that's clever, thanks!

PR #6797, opened 2025-06-19
expose Logins keydb structs constructors

jo:

  - Note:

### DIFF
--- a/components/logins/src/encryption.rs
+++ b/components/logins/src/encryption.rs
@@ -94,6 +94,7 @@ pub struct ManagedEncryptorDecryptor {
 }
 
 impl ManagedEncryptorDecryptor {
+    #[uniffi::constructor()]
     pub fn new(key_manager: Arc<dyn KeyManager>) -> Self {
         Self { key_manager }
     }
@@ -232,15 +233,21 @@ pub struct NSSKeyManager {
 }
 
 #[cfg(feature = "keydb")]
+#[uniffi::export]
 impl NSSKeyManager {
     /// Initialize new `NSSKeyManager` with a given `PrimaryPasswordAuthenticator`.
     /// There must be a previous initializiation of NSS before initializing
     /// `NSSKeyManager`, otherwise this panics.
+    #[uniffi::constructor()]
     pub fn new(primary_password_authenticator: Arc<dyn PrimaryPasswordAuthenticator>) -> Self {
         assert_nss_initialized();
         Self {
             primary_password_authenticator,
         }
+    }
+
+    pub fn into_dyn_key_manager(self: Arc<Self>) -> Arc<dyn KeyManager> {
+        self
     }
 }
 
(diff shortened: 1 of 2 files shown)


jo:
Is there anything else I need to consider to get the following code to work in the JavaScript Login Storage Adapter? Or do I still need factory methods? And does this already work with the Foreign Trait in JavaScript, and in this way or else?

```
class LoginStorageAuthenticator extends PrimaryPasswordAuthenticator {
  // init () {
  //   super()
  //   return new LoginStorageAuthenticator()
  // }

  async getPrimaryPassword() {
    console.log("LoginStorageAuthenticator: get primary password");
    console.log("TBD")
    return "secure"
  }

  onAuthenticationSuccess() {
    console.log("LoginStorageAuthenticator: authentication success!");
  }
  
  onAuthenticationFailure() {
    console.error("LoginStorageAuthenticator: authentication failure!");
  } 
} 

// then, later
const authenticator = new LoginStorageAuthenticator;
const keyManager = NssKeyManager.init(authenticator);
const encdec = ManagedEncryptorDecryptor.init(keyManager);
const path = `${Services.dirsvc.get("ProfD", Ci.nsIFile).path}/logins.db`;
const store = LoginStore.init(path, encdec);
```

mhammond:
Is that not working, or are you just checking if it should work? I'd think it should, but that's a naive opinion as I've not done that before, partly because the capability has only recently been added (until recently we only supported callback interfaces)

jo:
Currently it is not working, mostly due to the missing constructor exports I think.

jo:
So at least this patch is needed. What I'm curious is especially the Foreign Trait handling - can we just extend a class with it?

components/logins/src/encryption.rs
```diff
@@ -236,6 +237,7 @@ impl NSSKeyManager {
     /// Initialize new `NSSKeyManager` with a given `PrimaryPasswordAuthenticator`.
     /// There must be a previous initializiation of NSS before initializing
     /// `NSSKeyManager`, otherwise this panics.
+    #[uniffi::constructor()]
```
bendk, line 241: Does this compile with the keydb feature?  I think you might need to wrap the entire `impl` block with `#[uniffi::export]`.

bendk:
> const encdec = ManagedEncryptorDecryptor.init(keyManager);

I think this line may not work.  NssKeyManager implements `KeyManager` in Rust, but UniFFI doesn't yet have the capabilities to auto convert it into a `dyn KeyManager` like you can do in Rust.

I think you may need to add another `NssKeyManager` method to do that.  I think you just need to get the signature right and Rust will make the conversion itself.  Something like this should work:

```
pub fn into_dyn_key_manage(self: Arc<Self>) -> Arc<dyn KeyManager> {
   self
}
```

I could easily be missing something and you might need a couple tries to get everything right.  I'd recommend taking the git commit from this PR, vendoring it into moz-central / generating the UniFFI bindings, and testing yourself if the JS works or not.

bendk (approved):
I thought you were missing the `uniffi::export`, but I'm not totally sure since CI is passing and I believe that runs with `--all-features`.  I guess try running `cargo build logins --features=keydb` and see if it works or not.

Approving, since the code changes seem good and I trust you to figure out the build issue.

components/logins/src/encryption.rs
jo: yes it compiles also without the #[uniffi::export] impl.

jo:
Thank you @bendk for your hints! I will link mc against this pr and try locally before continuing.

PR #6802, opened 2025-06-20
Add tracing_support::TracingEvent.target_prefix

mhammond:
Consumers in JS will need to know what this prefix is. Even though it's trivially derived on the foreign side, given Rust already knows this it seems to make sense to supply it.

@bendk I'm kinda torn here - JS doing a trivial string split might actually be faster than the extra `RustBuffer` this will cuase uniffi to use, so maybe insisting foreign code does that split is fine. I'm almost leaning against it, but my JS patch had to do this twice, so I'm really not sure - wdyt?

  - Note:

### DIFF
--- a/components/support/tracing/src/layer.rs
+++ b/components/support/tracing/src/layer.rs
@@ -79,6 +79,7 @@ where
                 let event = crate::Event {
                     level: level.into(),
                     target: target.to_string(),
+                    target_prefix: prefix.to_string(),
                     name: event.metadata().name().to_string(),
                     message,
                     fields: serde_json::to_value(&fields).unwrap_or_default(),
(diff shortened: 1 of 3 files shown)


bendk:
> @bendk I'm kinda torn here - JS doing a trivial string split might actually be faster than the extra RustBuffer this will cuase uniffi to use, so maybe insisting foreign code does that split is fine. I'm almost leaning against it, but my JS patch had to do this twice, so I'm really not sure - wdyt?

I tend to agree.  I think it's fine to have consumers do the splitting.  Some other ideas would be to include the index to split on in the struct or actually split target into `target_prefix` and `target_suffix` parts.  I think those would perform pretty well, but I'm still not sure it's worth the extra code complexity.

PR #6812, opened 2025-06-25
Bug 1973876 - Use the geonames SQL collation for weather keywords too

0c0w3:
This creates a new `keywords_i18n` table that's the same as `keywords` except the main column uses the collation from the geonames table. Weather keywords are now added to this table instead of the old one.

I think we should use the new table in the future especially if/as we expand to non-English locales, but it's useful for English too since it ignores some punctuation.

Since the collation is now used for other things besides geonames, I renamed it `i18n_collation`.

Let me know if you have ideas on better names for the table and collation. I used "i18n" because I wanted to capture the idea they're useful for locale-specific user-facing strings. "i18n" seemed more appropriate than "l10n" because these things are structures/facilities that enable l10n and aren't the l10n artifacts themselves. I thought about using "unicode" but that's pretty ambiguous I think.

This also makes some improvements to keywords metrics.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -28,7 +28,7 @@ use crate::{
     },
     schema::{clear_database, SuggestConnectionInitializer},
     suggestion::{cook_raw_suggestion_url, FtsMatchInfo, Suggestion},
-    util::{full_keyword, split_keyword},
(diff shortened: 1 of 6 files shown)


ncloudioj (approved):
This looks great, thanks!
Merino could take lots of inspirations from here to handle i18n in the future.  👍

PR #6818, opened 2025-06-26
use an Option<LoginDB> in the mutex

jo:
to be able to take ownership for `db.close()`

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -850,6 +850,10 @@ impl LoginDb {
         tx.commit()?;
         Ok(row_count)
     }
+
+    pub fn shutdown(self) -> Result<()> {
+        self.db.close().map_err(|(_, e)| Error::SqlError(e))
+    }
 }
 
 lazy_static! {
--- a/components/logins/src/error.rs
+++ b/components/logins/src/error.rs
@@ -62,6 +62,9 @@ pub enum LoginsApiError {
 /// is never returned to the consumer.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("Database is closed")]
+    DatabaseClosed,
+
     #[error("Malformed incoming record")]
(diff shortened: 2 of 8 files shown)


components/logins/src/sync/engine.rs
```diff
@@ -343,6 +344,7 @@ impl LoginsSyncEngine {
             self.scope.err_if_interrupted()?;
             Ok(())
         })?;
+        let db = self.store.lock_db()?;
```
bendk: I think this is causing the deadlocks.  The DB is already locked and you should be able to just reuse the `db` variable.

components/logins/src/db.rs
```diff
@@ -850,6 +850,10 @@ impl LoginDb {
         tx.commit()?;
         Ok(row_count)
     }
+
+    pub fn shutdown(self) -> Result<()> {
+        self.db.close().map_err(|_| Error::DatabaseClosed)
```
bendk: I don't think this should be a database closed error.  It seems like it would be some sort of IO error while closing the files.  Could you leave off the `map_err()` part and use the normal error conversions (`Ok(self.db.close()?)`)?

components/logins/src/error.rs
```diff
@@ -93,6 +96,9 @@ pub enum Error {
     #[error("Error executing SQL: {0}")]
     SqlError(#[from] rusqlite::Error),
 
+    #[error("SQL Connection Error")]
+    SqlConnectionError,
+
```
bendk: Nit: I think this doesn't need to be added.

bendk (approved):
Looks good.  I had a minor change, but I don't need to re-review.

components/logins/src/error.rs
jo: cood catch! Removed the leftover.

PR #6821, opened 2025-06-30
Logins: unset encdec on shutdown

jo:

  - Note:

### DIFF
--- a/components/logins/src/store.rs
+++ b/components/logins/src/store.rs
@@ -79,37 +79,45 @@ fn map_bulk_result_entry(
 
 pub struct LoginStore {
     pub db: Mutex<Option<LoginDb>>,
-    pub encdec: Arc<dyn EncryptorDecryptor>,
+    pub encdec: Mutex<Option<Arc<dyn EncryptorDecryptor>>>,
 }
 
 impl LoginStore {
     #[handle_error(Error)]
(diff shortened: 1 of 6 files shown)


components/logins/src/store.rs
```diff
@@ -297,15 +312,29 @@ impl LoginStore {
...
         } else {
             Ok(())
         }
     }
 
+    fn shutdown_encdec(&self) -> Result<()> {
+        self.encdec
+            .lock()
+            .take()
+            .ok_or(Error::EncryptorDecryptorClosed)?;
```
bendk: I don't think this needs to be an error.  If `shutdown()` is called twice, that seems okay.  It's a close call though, however you decide is fine with me.

bendk (approved):
Looks good.

The 2 mutexes feel a bit scary to me.  Maybe I'll try to make a follow-up to merge them into one.  I don't think this should block you on your work though.

components/logins/src/store.rs
jo: I agree, good point!

jo:
> The 2 mutexes feel a bit scary to me. Maybe I'll try to make a follow-up to merge them into one. I don't think this should block you on your work though.

Thank you, that would be great!

PR #6822, opened 2025-06-30
Move EncryptorDecryptor into `LoginsDb`

bendk:
The main reason to do this is that `LoginStore` now only needs to store a single mutex.  This improves performance, since in most cases we were locking both mutexes together.  It also avoids potential deadlocks when the mutexes were locked in different orders.

It feels logical for `LoginsDb` to store the EncryptorDecryptor, since we're encrypt/decrypt things as they move in/out of the database.

The first commit is just some refactoring/preparation.

  - Note:


components/logins/src/store.rs
```diff
@@ -232,66 +220,59 @@ impl LoginStore {
...
 
     /// This method is intended to preserve metadata (LoginMeta) during a migration.
     /// In normal operation, this method should not be used; instead,
     /// use `add(entry)`, which manages the corresponding fields itself.
     #[handle_error(Error)]
     pub fn add_with_meta(&self, entry_with_meta: LoginEntryWithMeta) -> ApiResult<Login> {
-        let encdec = self.lock_encdec()?;
-        self.lock_db()?
-            .add_with_meta(entry_with_meta, encdec.as_ref())
-            .and_then(|enc_login| enc_login.decrypt(encdec.as_ref()))
```
bendk, line 273: I didn't realize it when I approved the PR, but I think the mutex order is a significant issue.  Here's we're locking `self.encdec` first, but in other functions we lock `self.db` first.  This could definitely lead to deadlocks.

jo: That's right, thankfully you noticed that!

components/logins/src/sync/engine.rs
```diff
@@ -26,16 +27,21 @@ use sync_guid::Guid;
 
 // The sync engine.
 pub struct LoginsSyncEngine {
-    pub store: Arc<LoginStore>,
-    pub scope: SqlInterruptScope,
-    pub staged: RefCell<Vec<IncomingBso>>,
+    store: Arc<LoginStore>,
+    scope: SqlInterruptScope,
```
jo: we need scope to be public in our sync-pass example

components/logins/src/db.rs
```diff
@@ -71,12 +73,15 @@ impl LoginDb {
...
 
-    pub fn open(path: impl AsRef<Path>) -> Result<Self> {
-        Self::with_connection(Connection::open(path)?)
+    pub fn open(path: impl AsRef<Path>, encdec: Arc<dyn EncryptorDecryptor>) -> Result<Self> {
+        Self::with_connection(Connection::open(path)?, encdec)
     }
 
-    pub fn open_in_memory() -> Result<Self> {
-        Self::with_connection(Connection::open_in_memory()?)
+    #[cfg(test)]
```
jo, line 85: I think you need to remove this because its used in `LoginStore::open_in_memory` which is used in `sync-test`

jo:
Since you moved encdec into db, you need to init nss in the schema tests, too:

```diff
diff --git a/components/logins/src/schema.rs b/components/logins/src/schema.rs
index 5dfb9fcba..c923f776e 100644
--- a/components/logins/src/schema.rs
+++ b/components/logins/src/schema.rs
@@ -259,9 +259,11 @@ mod tests {
     use crate::encryption::test_utils::TEST_ENCDEC;
     use crate::LoginDb;
     use rusqlite::Connection;
+    use nss::ensure_initialized;
 
     #[test]
     fn test_create_schema() {
+        ensure_initialized();
         let db = LoginDb::open_in_memory();
         // should be VERSION.
         let version = db.query_one::<i64>("PRAGMA user_version").unwrap();
@@ -270,6 +272,7 @@ mod tests {
 
     #[test]
     fn test_upgrade_v1() {
+        ensure_initialized();
         // manually setup a V1 schema.
         let connection = Connection::open_in_memory().unwrap();
         connection
```

components/logins/src/db.rs
bendk: I ended up keeping the `open_in_memory` functions private.  It feels a bit weird for these to be exposed as the public API and it's simple enough for the `sync-test` code to use `:memory:` as the path to get the same effect.

PR #6828, opened 2025-07-10
Fixed a deadlock in delete_undecryptable_records_for_remote_replacement

bendk:

  - Note:

### DIFF
--- a/components/logins/src/store.rs
+++ b/components/logins/src/store.rs
@@ -198,10 +198,13 @@ impl LoginStore {
         // This function was created for the iOS logins verification logic that will
         // remove records that prevent logins syncing. Once the verification logic is
         // removed from iOS, this function can be removed from the store.
+
+        // Creating an engine requires locking the DB, so make sure to do this first
+        let engine = LoginsSyncEngine::new(Arc::clone(&self))?;
+
         let db = self.lock_db()?;
         let deletion_stats =
             db.delete_undecryptable_records_for_remote_replacement(db.encdec.as_ref())?;
-        let engine = LoginsSyncEngine::new(Arc::clone(&self))?;
         engine.set_last_sync(&db, ServerTimestamp(0))?;
         Ok(deletion_stats)
     }
@@ -526,6 +529,16 @@ mod test {
             Err(LoginsApiError::UnexpectedLoginsApiError { reason: _ })
         ));
         assert!(store.db.lock().is_none());
+    }
+
+    #[test]
+    fn test_delete_undecryptable_records_for_remote_replacement() {
+        ensure_initialized();
+        let store = Arc::new(LoginStore::new_in_memory());
+        // Not much of a test, but let's make sure this doesn't deadlock at least.
+        store
+            .delete_undecryptable_records_for_remote_replacement()
+            .unwrap();
     }
 }
 


lougeniaC64:
I wasn't able to reproduce the reported issue via the given STRs with these changes OR without them. Waiting to get clarification from the iOS team so I can verify.

mhammond:
That's a real footgun - I can verify that the old code does deadlock unconditionally every time it is called! It should be trivially reproducible in xcode with a breakpoint on the call.

lougeniaC64:
@mhammond I'm a bit perplexed here because I have successfully tested the logins verification in the simulator and it was also successfully QA'd. I was just attempting to verify this fixed the iOS bug based on their STRs but I'm not blocking merging these changes.

mhammond:
I'm a bit confused too - @bendk is it your understanding that since b23713d55c101e3e327611f9ed5894727fce5e84,  `delete_undecryptable_records_for_remote_replacement()` unconditionally deadlocks every time it is called? Or is there something I'm missing here?

bendk:
> @bendk is it your understanding that since https://github.com/mozilla/application-services/commit/b23713d55c101e3e327611f9ed5894727fce5e84, delete_undecryptable_records_for_remote_replacement() unconditionally deadlocks every time it is called? Or is there something I'm missing here?

That's my understanding too.

> I'm a bit perplexed here because I have successfully tested the logins verification in the simulator and it was also successfully QA'd.

Was the testing before https://github.com/mozilla/application-services/commit/b23713d55c101e3e327611f9ed5894727fce5e84?

>  I was just attempting to verify this fixed the iOS bug based on their STRs but I'm not blocking merging these changes.

This is the only reason I'm not hitting the merge button right now.  If this doesn't fix the iOS bug and we decide to backout   b23713d55c101e3e327611f9ed5894727fce5e84, then we'll also need to backout this commit.  It's not a big deal, but I figured we could get a yes or no on if it fixed the bug before merging.

That said, if @lougeniaC64  can't repro the bug, then maybe the best way to test if this fixes it to merge now and ask them to test the next nightly?

lougeniaC64:
> Was the testing before [b23713d](https://github.com/mozilla/application-services/commit/b23713d55c101e3e327611f9ed5894727fce5e84)? 

Yes, it was. 

> That said, if @lougeniaC64 can't repro the bug, then maybe the best way to test if this fixes it to merge now and ask them to test the next nightly?

Sounds good to me. This change makes sense if the other changes remain. Worse case scenario we need to revert this. 🤷🏾‍♀️

mhammond:
> That said, if @lougeniaC64 can't repro the bug, then maybe the best way to test if this fixes it to merge now and ask them to test the next nightly?

My problem with this is that if there aren't reliable steps to hit the deadlock, then there aren't reliable steps which call this function. Without reliable steps known to call this function, I don't see how a new Nightly can be effectively tested?

mhammond:
To clarify: it's not clear that *this* deadlock is exactly the same as the iOS UI deadlock - ie, it might be that sometimes, even though this will definitely deadlock a thread, it will not definitely deadlock the UI.

That said though, it seems literally impossible that this function could have fixed any logins since the broken commit. This in turn implies we do not have manual or automated tests here which require that function to have completed to pass?

lougeniaC64:
> That said though, it seems literally impossible that this function could have fixed any logins since the broken commit. This in turn implies we do not have manual or automated tests here which require that function to have completed to pass?

Saying this for context not at to oppose, the iOS deadlock was reproducible in the iOS dev environment but not in nightly, beta, or release. There is [this test](https://github.com/mozilla/application-services/blob/50c79b1165a618ebf80068f7a2cb98404b065233/testing/sync-test/src/logins.rs#L341) but to your point it is not automated.

bendk:
Looks like there's general agreement to merge this, even though it's not clear if it's causing the iOS issue.

lougeniaC64:
> Looks like there's general agreement to merge this, even though it's not clear if it's causing the iOS issue.

@bendk It was a good call. I got the updated STRs and was able to reproduce the issue and confirm that the changes in this PR resolves it.

PR #6837, opened 2025-07-18
Upgrade rusqlite to 0.37.

mhammond:
* They added a `query_one` method which conflicts with our method
extension of the same name. So `conn_ext_query_one` added.

* We were depending on the order of triggers in places, which seems to have changed.

* We were trying to execute 2 statements in a single
command in suggest, which theoretically should never have worked.

### DIFF
--- a/components/support/sql/src/conn_ext.rs
+++ b/components/support/sql/src/conn_ext.rs
@@ -70,7 +70,10 @@ pub trait ConnExt {
     }
 
     /// Execute a query that returns a single result column, and return that result.
-    fn query_one<T: FromSql>(&self, sql: &str) -> SqlResult<T> {
+    /// NOTE: rusqlite now has a builtin `query_one` (with not quite these semantics)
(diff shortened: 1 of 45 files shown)


components/support/sql/src/conn_ext.rs
```diff
@@ -70,7 +70,10 @@ pub trait ConnExt {
     }
 
     /// Execute a query that returns a single result column, and return that result.
-    fn query_one<T: FromSql>(&self, sql: &str) -> SqlResult<T> {
+    /// NOTE: rusqlite now has a builtin `query_one` (with not quite these semantics)
+    /// and a `one_column` (with these semantics but subtly different args) which should
+    /// generally be preferred. We've kept this to make upgrading easier.
+    fn conn_ext_query_one<T: FromSql>(&self, sql: &str) -> SqlResult<T> {
```
bendk, line 76: This makes total sense to me.  Let's continue using this one for now, but try to move to one of the builtin methods.

PR #6840, opened 2025-07-22
Bug 1978637 - Bug 1976993 - Change `display_name` in the search config to `display_name_map`

0c0w3:

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@
 
 - `SearchEngineUrls` now has an optional `visual_search` field, supporting
   visual search endpoints in engine configs.
-- `SearchEngineUrl` now has an optional `display_name` field, which is useful if
-  a URL corresponds to a brand name distinct from the engine's brand name.
+- `SearchEngineUrl` now has an optional `display_name_map` field, which is useful
+  if a URL corresponds to a brand name distinct from the engine's brand name.
 
 [Full Changelog](https://github.com/mozilla/application-services/compare/v141.0...v142.0)
 
--- a/components/search/src/types.rs
+++ b/components/search/src/types.rs
@@ -5,6 +5,7 @@
 //! This module defines the types that we export across the UNIFFI interface.
 
 use serde::Deserialize;
+use std::collections::HashMap;
 
 /// The list of possible application names that are currently supported.
 #[derive(Clone, Debug, Default, Deserialize, PartialEq, uniffi::Enum)]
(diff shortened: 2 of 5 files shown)


components/search/src/types.rs
```diff
@@ -127,9 +128,13 @@ pub struct SearchEngineUrl {
...
-    /// The display name of the URL, if any. This is useful if the URL
-    /// corresponds to a brand name distinct from the engine's brand name.
-    pub display_name: Option<String>,
+    /// A map from locale codes to display names of the URL. This is useful if
+    /// the URL corresponds to a brand name distinct from the engine's brand
+    /// name. Since brand names can be localized, this is a map rather than a
+    /// URL. The client will fall back to the special locale code "default" when
+    /// its locale is not present in the map.
+    #[uniffi(default = None)]
+    pub display_name_map: Option<HashMap<String, String>>,
```
Standard8: Thought: Since we already do processing based on the locale within the component, I'm wondering if we could should simplify this, and simply return the `display_name`. This would simplify the application implementation.

There's `user_environment.locale` which could be used and passed into `from_configuration_details`, however... this would complicate the `From<JSONEngineUrls>` functions...

Standard8 (approved):
There's one comment in-line, which is optional. I'm not sure if it is worth it, but I thought I'd see what your views are.

If you do change, I'd like to take another look, but if not, I'm happy for this to be merged as-is, hence approving.

CHANGELOG.md
```diff
@@ -23,8 +23,8 @@
 
 - `SearchEngineUrls` now has an optional `visual_search` field, supporting
   visual search endpoints in engine configs.
-- `SearchEngineUrl` now has an optional `display_name` field, which is useful if
-  a URL corresponds to a brand name distinct from the engine's brand name.
+- `SearchEngineUrl` now has an optional `display_name_map` field, which is useful
+  if a URL corresponds to a brand name distinct from the engine's brand name.
```
Standard8: I just realised - this probably needs updating now that 143.0 is in progress.

components/search/src/types.rs
0c0w3: Yeah, that makes sense. I'll give it a shot.

0c0w3:
I think it worked out OK. I removed the `From` impls on `SearchEngineUrl` and `SearchEngineUrls` and instead take advantage of the fact we can make defaults and then `merge` the relevant JSON counterparts. Otherwise the logic that converts from `display_name_map` to `display_name` would have to be duplicated, once in the `From` impl and once in the `merge` impl, or at least it would have to be factored out and called from both places.

There's a little messiness where the default value for `SearchEngineUrl::method` comes from `JSONEngineMethod::default()`. That dependency isn't super obvious, but since `SearchEngineUrl::method` is a non-optional string and `JSONEngineUrl` is an optional `JSONEngineMethod`, I don't think there's a better way without making `SearchEngineUrl::method` also optional. As it stands, `SearchEngineUrl::default()` creates a value with an empty string as the method, which doesn't seem right and might bite us in the future, and I think this patch is a small improvement in that regard.

For these functions that now take a `&SearchUserEnvironment`, I made it the first param since existing functions that take one mostly seem to take it as the first param, but not all. I'm happy to move it to the end or somewhere else if that would make more sense.

0c0w3:
Thanks, I'll squash these commits and merge.

CHANGELOG.md
0c0w3: The final commit ended up not touching the changelog.

PR #6842, opened 2025-07-25
Fixup monorepo script for new a-s crates

mhammond:
Fixing up a few issues with recent component changes.


mhammond:
Still haven't thought about the `sed` problem :(

PR #6843, opened 2025-07-28
Pull in the latest cctools and clang-dist toolchains rather than pinning to expired toolchains [ci full]

rvandermeulen:


rvandermeulen:
Switching to `latest` carries the risk of things breaking out from under us, but I don't think that's a super big concern in the big picture, especially with the upcoming migration. That said, we can always take the more conservative option here and just bump to a newer fixed date instead of people are overly worried about it.

rvandermeulen:
@mergify backport release-v142 release-v141

PR #6844, opened 2025-07-28
Bump UniFFI to 0.29.4

bendk:
We fixed a bug where objects were dropped to early in some cases.  I highly doubt we're hitting it currently, but it seems good to bump the version just to be sure in the future.

  - Note:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,54 +82,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
-dependencies = [
- "anstyle",
- "anstyle-parse",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
-[[package]]
 name = "anstyle"
 version = "1.0.10"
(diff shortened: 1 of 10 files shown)


bendk:
I think this only affects the megazords.  I'll try to put a moz-central PR as well.

bendk:
> The client.rs changes seem a little odd?

I think the `serde_json` upgrade caused the output to change slightly.  I got the changes by running the tests with  `UPDATE_EXPECT=1`

bendk:
> Also, do you think we should bump askama for nimbus-fml to avoid duplicating that crate?

Yes that makes sense to me.  @freshstrangemusic Do you forsee any issues from bumping the askama version?

freshstrangemusic:
> > Also, do you think we should bump askama for nimbus-fml to avoid duplicating that crate?
> 
> Yes that makes sense to me. @freshstrangemusic Do you forsee any issues from bumping the askama version?

This is used to generate Swift and Kotlin that is consumed in Firefox for iOS and Fenix. If this breaks it, we should expect to see CI build failures in the next nightly build. We could potentially just diff the generated code between both versions of the fml application, but I'm not really an expert on how we do that. @jeddai could probably tell you.

jeddai:
At first glance, I am not seeing any breaking changes that would affect Nimbus between versions 0.12 and 0.14 of Askama.

I think it's worth attempting the upgrade and seeing what happens.

We have tests that run kotlin/swift compilation on the generated code in Nimbus FML, so we should see errors in whatever PR includes the upgrade.

mhammond (approved):

PR #6847, opened 2025-07-28
Align Android lint configs with firefox-main

rvandermeulen:
This updates ktlint to the latest 1.7.0 version with the same rules that are currently in place for firefox-main and also updates our detekt configuration to be aligned with firefox-main.


rvandermeulen:
In general, I aimed to suppress the warnings that didn't seem like we'd ever be wanting to fix while leaving others in the baseline that might want to get fixed eventually.

PR #6849, opened 2025-07-29
Fix sed usage in import-app-services-to-central.sh

mhammond:
As we discussed in slack, might be worth double-checking it actually works for you


bendk:
This worked great for me.

Side issue: I think some of those phab patches need upgrading now that the 2nd tracing support Pr landed.

PR #6851, opened 2025-07-29
logins: allow re-adding of previously deleted entries

jo:

  - Note:


components/logins/src/db.rs
```diff
@@ -291,7 +291,7 @@ impl LoginDb {
     // just the SQL - no validation or anything.
     fn insert_new_login(&self, login: &EncryptedLogin) -> Result<()> {
         let sql = format!(
-            "INSERT INTO loginsL (
+            "INSERT OR REPLACE INTO loginsL (
```
jo, line 294: is the sync status `new` correct in the case of a re-add?

mhammond: that's a great question - I *think* so. The main thing "new" does is that when an item with "new" is deleted, we assume it's never made it to the server, so don't upload a tombstone - but in this scenario there is likely to already be a tombstone on the server.

(Note I haven't yet looked at the patch, just responding to this specific question :)

jo: ok, but couldn't it be that
1. user makes a new entry, sync status `new`
2. it gets synced to the server, sync status is `synced`
3. user deletes it, status is `changed`
4. re-adds it, sync status is again `new`
5. user deletes it again, status is `changed`
6. the tombstone never makes it to the server?

I don't quite understand [the logic](https://github.com/mozilla/application-services/blob/main/components/logins/src/db.rs#L717-L768) - are we even looking at the sync status here?

bendk (approved):
This looks more-or-less right to me, other than the syncing question.  I'll let Mark handle that, since he knows the system better.  My only suggestion is that you could try adding some code to https://github.com/mozilla/application-services/blob/main/testing/sync-test/src/logins.rs to test the corner cases you're talking about.

components/logins/src/db.rs
mhammond: We are looking at sync status when fetching outgoing [here](https://github.com/mozilla/application-services/blob/main/components/logins/src/sync/engine.rs#L239-L250) - but that too isn't excluding items with "new". So it does look like maybe we are uploading a tombstone for an item that is created and deleted before being synced?

mhammond:
Yeah, I agree re tests - https://github.com/mozilla/application-services/blob/main/components/logins/src/sync/engine.rs#L523 might also be fine and is going to be run more often. That code is roughly setup to assume a "local" state and to apply incoming records, which is an important part of this, and the test would want to simulate that it previously saw the incoming tombstone, so has the tombstone in `LoginsM`, no record in `LoginsL` and the login is never returned to the app, but after the sync it functions correctly (which also mean no record in `LoginsL`, but the now non-tombstone version in `LoginsM` *is* returned to the app (there will still be no `LoginsL` because it's not locally modified). (I think - that's how it all works, right? It's all becoming a bit vague to me!)

That sync-test code is a more "realistic" test scenario which sets up a "nearly real" multi-device environment, and it would be *ideal* if a test could be there, but I think I'd be fine with one or the other for now :)

components/logins/src/db.rs
jo: Yes, it looks like, see test case `test_sync_local_delete` below

jo:
I added some tests to the engine, and it looks like everything is working as expected.

Also added a test case for push synchronization of a deletion of a login that only exists locally, and indeed, the tombstone is sent even if it only existed locally. This would be an optimization, but I find this behaviour simple and solid.

PR #6877, opened 2025-08-05
Bug 1925091 - [ci full] Make Android FxaAccountManager supply comma-separate list of values for service=sync

AadityaDhingra:

  - Note:

### DIFF
--- a/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
+++ b/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
@@ -113,9 +113,10 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
     fun beginOAuthFlow(
         scopes: Array<String>,
         entrypoint: String,
+        services: List<String> = listOf("sync"),
     ): String {
         return withMetrics {
-            this.inner.beginOauthFlow(scopes.toList(), entrypoint)
+            this.inner.beginOauthFlow(scopes.toList(), entrypoint, services)
         }
     }
 
@@ -133,9 +134,10 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
         pairingUrl: String,
         scopes: Array<String>,
         entrypoint: String,
+        services: List<String> = listOf("sync"),
     ): String {
         return withMetrics {
-            this.inner.beginPairingFlow(pairingUrl, scopes.toList(), entrypoint)
+            this.inner.beginPairingFlow(pairingUrl, scopes.toList(), entrypoint, services)
         }
     }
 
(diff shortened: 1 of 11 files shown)


components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
```diff
@@ -115,7 +115,7 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
         entrypoint: String,
```
jonalmeida, line 115: We should expose the service param here and use the default value from here since there is this thin wrapper around the rust generated FFI code.

```suggestion
        entrypoint: String,
        services: List<String> = listOf("sync"),
```

components/fxa-client/src/fxa_client.udl
```diff
@@ -189,13 +189,12 @@ interface FirefoxAccount {
...
   ///         For example, the application toolbar, on the onboarding flow.
+  ///   - `service` - list of services to request.
   ///   - `metrics` - optionally, additional metrics tracking parameters.
   ///       - These will be included as query parameters in the resulting URL.
   ///
   [Throws=FxaError]
-  string begin_oauth_flow([ByRef] sequence<string> scopes, [ByRef] string entrypoint);
-  
-
+  string begin_oauth_flow([ByRef] sequence<string> scopes, [ByRef] string entrypoint, [ByRef] sequence<string> service);
```
jonalmeida: We can use default values for the udl so that the iOS tests don't fail if that's a preference: https://mozilla.github.io/uniffi-rs/latest/proc_macro/functions.html#default-values

Alternatively, we can also fix the breaking tests too. :)

jonalmeida (commented):
Self note: Leaving my offline feedback here to come back to it later.

components/fxa-client/src/fxa_client.udl
bendk: Defaults in UDL are done slightly differently: https://mozilla.github.io/uniffi-rs/latest/udl/functions.html.  I think a list of strings should be supported, but I'm not totally sure.

components/fxa-client/src/auth.rs
```diff
@@ -76,17 +76,23 @@ impl FirefoxAccount {
...
     ///   - `metrics` - optionally, additional metrics tracking parameters.
     ///       - These will be included as query parameters in the resulting URL.
     #[handle_error(Error)]
+    #[uniffi::method(default(service = &["sync"]))]
     pub fn begin_oauth_flow<T: AsRef<str>>(
         &self,
         // Allow both &[String] and &[&str] since UniFFI can't represent `&[&str]` yet,
         scopes: &[T],
         entrypoint: &str,
+        service: &[String],
```
bendk: This could be `&[T]` as well, using the same hack as `scopes`.  I think that makes some of the Rust code a bit more idiomatic, but it's not a big deal.  Both `scopes` and `service` would need to be the same type (either `&str` or `&String`) but that seems okay to me.

components/fxa-client/src/fxa_client.udl
jonalmeida: Yeah, sorry I realized too late that I shared the wrong doc link. 🤦

jonalmeida:
Closing because we have this ability already.

PR #6878, opened 2025-08-05
DRAFT - feat: Adding new Mozilla Ads Client component

luc-lisi:
#### Overview
This PR creates a new component in the Application-services repo, `MAC` (Mozilla Ads Client). MAC acts as an SDK for Mozilla's internal privacy preserving ads routing service (MARS) and allows for simple integration with MARS' Unified API (UAPI). This component is able to fetch ads and handle logging ad interaction events such as clicks and impressions. In the future, it will also support caching. It is primarily intended for use on mobile surfaces but is extendable to any surface able to integrate with this repo.

This PR only includes the "core" of the library. Additional Swift and Kotlin wrappers still need to be created to handle displaying of ads fetched through `MAC` to the respective surfaces.

#### What was done

- A new component `MAC` was created.
- Functionality was adding which allows for fetching of ads, handling click/impression/report callbacks, and error logging.
- Unit tests were added.

#### Component Feature list

- Fetch ads from MARS given a set of user defined `placement_ids`.
- Provides type definitions for ads, placements, and placement definitions.
- Ability to report back to MARS that a particular placement had a click or impression occur.
- Ability to report back to MARS that a particular placement was reported by a user.
- If errors occur during any of these processes, an attempt is made to emit an event to internal telemetry dashboards for monitoring purposes.

#### Final Notes:

This PR contains no breaking changes and does not affect any components other than the one being created.
This PR includes tests for the new component that pass.
This PR adds no new dependencies to application-services.

Like MARS, MAC is privacy-first. It does not track user information and it does not send clearly identifiable information to Mozilla. Of the information Mozilla does receive, anything shared with advertisers is aggregated and/or de-identified to preserve user privacy.

  - Note:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "components/fxa-client",
     "components/init_rust_components",
     "components/logins",
+    "components/mac",
     "components/merino",
     "components/nimbus",
     "components/places",
--- a/components/mac/src/error.rs
+++ b/components/mac/src/error.rs
@@ -3,7 +3,7 @@
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 */
 
-use error_support::{ErrorHandling, GetErrorHandling, error};
+use error_support::{error, ErrorHandling, GetErrorHandling};
 use viaduct::Response;
 
 pub type ApiResult<T> = std::result::Result<T, ApiError>;
--- a/components/mac/src/error.rs
+++ b/components/mac/src/error.rs
@@ -1,0 +1,175 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+use error_support::{ErrorHandling, GetErrorHandling, error};
+use viaduct::Response;
+
+pub type ApiResult<T> = std::result::Result<T, ApiError>;
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum ApiError {
+    #[error("Something unexpected occurred.")]
+    Other { reason: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ComponentError {
+    #[error("Error requesting ads: {0}")]
+    RequestAds(#[from] RequestAdsError),
+
+    #[error("Error recording a click for a placement: {0}")]
+    RecordClick(#[from] RecordClickError),
+
+    #[error("Error recording an impressions for a placement: {0}")]
+    RecordImpression(#[from] RecordImpressionError),
+
+    #[error("Error reporting an ad: {0}")]
+    ReportAd(#[from] ReportAdError),
+}
+
+impl GetErrorHandling for ComponentError {
+    type ExternalError = ApiError;
+
+    fn get_error_handling(&self) -> ErrorHandling<Self::ExternalError> {
+        ErrorHandling::convert(ApiError::Other {
+            reason: self.to_string(),
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RequestAdsError {
+    #[error("Error building ad requests from configs: {0}")]
+    BuildRequest(#[from] BuildRequestError),
+
+    #[error("Error requesting ads from MARS: {0}")]
+    FetchAds(#[from] FetchAdsError),
+
+    #[error("Error building placements from ad response: {0}")]
+    BuildPlacements(#[from] BuildPlacementsError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum BuildRequestError {
+    #[error("Could not build request with empty placement configs")]
+    EmptyConfig,
+
+    #[error("Duplicate placement_id found: {placement_id}. Placement_ids must be unique.")]
+    DuplicatePlacementId { placement_id: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum BuildPlacementsError {
+    #[error("Duplicate placement_id found: {placement_id}. Placement_ids must be unique.")]
+    DuplicatePlacementId { placement_id: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum FetchAdsError {
+    #[error("URL parse error: {0}")]
+    UrlParse(#[from] url::ParseError),
+
+    #[error("Error sending request: {0}")]
+    Request(#[from] viaduct::Error),
+
+    #[error("JSON error: {0}")]
+    Json(#[from] serde_json::Error),
+
+    #[error("Could not fetch ads, MARS responded with: {0}")]
+    HTTPError(#[from] HTTPError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum EmitTelemetryError {
+    #[error("URL parse error: {0}")]
+    UrlParse(#[from] url::ParseError),
+
+    #[error("Error sending request: {0}")]
+    Request(#[from] viaduct::Error),
+
+    #[error("JSON error: {0}")]
+    Json(#[from] serde_json::Error),
+
+    #[error("Could not fetch ads, MARS responded with: {0}")]
+    HTTPError(#[from] HTTPError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CallbackRequestError {
+    #[error("URL parse error: {0}")]
+    UrlParse(#[from] url::ParseError),
+
+    #[error("Error sending request: {0}")]
+    Request(#[from] viaduct::Error),
+
+    #[error("JSON error: {0}")]
+    Json(#[from] serde_json::Error),
+
+    #[error("Could not fetch ads, MARS responded with: {0}")]
+    HTTPError(#[from] HTTPError),
+
+    #[error("Callback URL missing: {message}")]
+    MissingCallback { message: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RecordImpressionError {
+    #[error("Callback request to MARS failed: {0}")]
+    CallbackRequest(#[from] CallbackRequestError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RecordClickError {
+    #[error("Callback request to MARS failed: {0}")]
+    CallbackRequest(#[from] CallbackRequestError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ReportAdError {
+    #[error("Callback request to MARS failed: {0}")]
+    CallbackRequest(#[from] CallbackRequestError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum HTTPError {
+    #[error("Validation error ({code}): {message}")]
+    Validation { code: u16, message: String },
+
+    #[error("Bad request ({code}): {message}")]
+    BadRequest { code: u16, message: String },
+
+    #[error("Server error ({code}): {message}")]
+    Server { code: u16, message: String },
+
+    #[error("Unexpected error ({code}): {message}")]
+    Unexpected { code: u16, message: String },
+}
+
+pub fn check_http_status_for_error(response: &Response) -> Result<(), HTTPError> {
+    let status = response.status;
+    if status >= 400 {
+        let error_message = response.text();
+        let error = match status {
+            400 => HTTPError::BadRequest {
+                code: status,
+                message: error_message.to_string(),
+            },
+            422 => HTTPError::Validation {
+                code: status,
+                message: error_message.to_string(),
+            },
+            500..=599 => HTTPError::Server {
+                code: status,
+                message: error_message.to_string(),
+            },
+            _ => HTTPError::Unexpected {
+                code: status,
+                message: error_message.to_string(),
+            },
+        };
+        return Err(error);
+    }
+    Ok(())
+}
--- a/components/mac/src/error.rs
+++ b/components/mac/src/error.rs
@@ -173,3 +173,65 @@ pub fn check_http_status_for_error(response: &Response) -> Result<(), HTTPError>
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use url::Url;
+
+    fn mock_response(status: u16, body: &str) -> Response {
+        Response {
+            request_method: viaduct::Method::Get,
+            url: Url::parse("https://example.com").unwrap(),
+            status,
+            headers: viaduct::Headers::new(),
+            body: body.as_bytes().to_vec(),
+        }
+    }
+
+    #[test]
+    fn test_ok_status_returns_ok() {
+        let response = mock_response(200, "OK");
+        let result = check_http_status_for_error(&response);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_bad_request_returns_http_error() {
+        let response = mock_response(400, "Bad input");
+        let result = check_http_status_for_error(&response);
+        match result {
+            Err(HTTPError::BadRequest { code, ref message }) => {
+                assert_eq!(code, 400);
+                assert_eq!(message, "Bad input");
+            }
+            other => panic!("Unexpected result: {:?}", other),
+        }
+    }
+
+    #[test]
+    fn test_validation_failure_returns_http_error() {
+        let response = mock_response(422, "Invalid data");
+        let result = check_http_status_for_error(&response);
+        match result {
+            Err(HTTPError::Validation { code, ref message }) => {
+                assert_eq!(code, 422);
+                assert_eq!(message, "Invalid data");
+            }
+            other => panic!("Unexpected result: {:?}", other),
+        }
+    }
+
+    #[test]
+    fn test_server_error_500() {
+        let response = mock_response(500, "Something broke");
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
     "components/fxa-client",
     "components/init_rust_components",
     "components/logins",
+    "components/mac",
```
mhammond: gotta admit I'm not a huge fan of the component name, which will probably seem even worse once in mozilla-central - "ads-client" or even "mars-client" do seem better.

components/mac/src/error.rs
```diff
@@ -0,0 +1,237 @@
...
+pub type ApiResult<T> = std::result::Result<T, ApiError>;
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum ApiError {
+    #[error("Something unexpected occurred.")]
+    Other { reason: String },
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ComponentError {
```
mhammond: there are a lot of errors here, which seems like a liability in the longer term. I haven't dug into exactly how they are all used, but most aren't reported to the consumer, so there might be scope for simplifying them.

components/mac/src/models.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[serde(rename = "IAB-2.0")]
+    IAB2_0,
+
+    #[serde(rename = "IAB-2.1")]
+    IAB2_1,
+
+    #[serde(rename = "IAB-2.2")]
+    IAB2_2,
+
+    #[serde(rename = "IAB-3.0")]
```
mhammond: surprised that this works - I'd expect renaming to a symbol with a dash in it might break foreign bindings, where names typically can't have dashes.

components/mac/src/test_utils.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use std::collections::HashMap;
+
+use crate::{
+    mars::DefaultMARSClient,
+    models::{AdCallbacks, AdResponse, IABContentTaxonomy, MozAd},
+    IABContent, MozAdsPlacement, MozAdsPlacementConfig,
+};
+
+#[allow(dead_code)]
```
mhammond: I guess these are all just placeholders while you work out your testing story?

components/mac/src/instrument.rs
```diff
@@ -0,0 +1,179 @@
...
+    Init,
+    RenderError,
+    AdLoadError,
+    FetchError,
+    InvalidUrlError,
+}
+
+#[allow(dead_code)]
+pub trait TrackError<T, ComponentError> {
+    fn emit_telemetry_if_error(self) -> Self;
```
mhammond: Some of the error_support work @bendk did also attempts to handle this use-case - although that's via glean rather than a dedicated end-point. We should talk with him to see about how to make this simpler.

Cargo.toml
luc-lisi: Yeah that's very fair. We've had some internal struggles with agreeing on the name but I'd agree `mac` doesn't feel particularly great or descriptive.

I'll see if we can get agreement on `ads-client`.

components/mac/src/models.rs
luc-lisi: Oh that's interesting. I'll have to play around with this to verify nothing is going on that is unexpected.

My understanding of `serde rename` might also be incorrect. I understood this as "tell serde that during JSON serialization/deserialization, that these symbols should convert to the given string." But not that the enum symbol itself is changing across the FFI.

mhammond: oops, I misread that as a *uniffi* rename - I think serde is fine, sorry for the noise.

components/mac/src/error.rs
luc-lisi: Yeah great mention. I wasn't entirely sure the best way to "lift" errors through function calls in Rust so I opted for this approach. That said, it certainly might be the wrong way to go about it.

Ideally, we want as much traceability as possible if there is a failure somewhere in the SDK, so I essentially gave each function its own error type to make things incredibly explicit. Then each method's error type is lifted through to it's calling method. Or at least that is the intent.

Let me know if you have better recommendation on how to do this kind of error propagation though.

components/mac/src/test_utils.rs
luc-lisi: Oh this `dead_code` annotation is just because it seems like functions used inside of tests seem to still be marked as unused by clippy. So these functions aren't actually dead code, but just not used in non-test code.

Is there a better way to organize these kinds of "test helper functions" which you might want to share across different tests? I could also just put them explicitly in the test mods I suppose.

components/mac/src/instrument.rs
luc-lisi: Oh yeah let's definitely talk that over once he is back! We have our own custom endpoints we are using to track errors in the SDK, but it is definitely possible we could consider moving them to glean.

We may also include some custom logging in the future though so it may just be that a small custom solution like this is worth while. I'm not particularly opinionated here though.

components/mac/src/models.rs
luc-lisi: Let me know if my understanding of `serde rename` is incorrect though. Being new to the language and all these crates I absolutely could be missing things!

components/mac/src/test_utils.rs
mhammond: > So these functions aren't actually dead code, but just not used in non-test code.

So typically these are just wrapped in `#[cfg(test)]` - that's also why it's common to see test-specific modules - eg:
```
#[cfg(test)]
mod tests {
  fn my_test_helper() { ... }

  #[test]
  fn actual_test_function() { ... }
}
```

components/mac/src/models.rs
mhammond: Your understanding is correct. But I guess I'm still mildly surprised you want `IAB-3.0` in the JSON. This does however imply you want a test for exactly this - ie, a test that JSON converts to the enum you want, and that an enum converts to the JSON you want.

Cargo.toml
luc-lisi: Confirmed `ads-client` will work for the component name in a-s 😄 

Will swap that over before we merge. Holding off on doing that immediately just so the current comments don't go away as it will mean moving all these files to a new directory.

components/mac/src/models.rs
luc-lisi: Ah yeah, we just don't have control over what the API expects. I agree though that it's a bit funny to have characters like `-` or `.` in the JSON.

Great point about the test though. I think we indirectly test our serialization but I'll add an explicit test as well.

components/mac/src/test_utils.rs
luc-lisi: Oh that makes much more sense. I've gone ahead and wrapped the deadcode using those attributes 👍

Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "2"
 
 # Note: Any additions here should be repeated in default-members below.
 members = [
+    "components/ads-client",
```
mhammond, line 6: It probably means sense to add this to default-members too?

luc-lisi:
@mhammond I'm noticing the `android-build-pr` CI step is failing, but I'm not entirely sure what is causing this.

Is this a known issue, or is this something these changes are breaking?

components/ads-client/src/error.rs
```diff
@@ -0,0 +1,237 @@
...
+    #[error("Server error ({code}): {message}")]
+    Server { code: u16, message: String },
+
+    #[error("Unexpected error ({code}): {message}")]
+    Unexpected { code: u16, message: String },
+}
+
+pub fn check_http_status_for_error(response: &Response) -> Result<(), HTTPError> {
+    let status = response.status;
+    if status >= 400 {
```
mhammond: this seems a little risky wrt other codes - eg, 3xx, even 1xx, if seen, probably shouldn't be treated as success? Is anything other than 200 really expected here?

components/ads-client/src/error.rs
```diff
@@ -0,0 +1,237 @@
...
+        let response = mock_response(200, "OK");
+        let result = check_http_status_for_error(&response);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_bad_request_returns_http_error() {
+        let response = mock_response(400, "Bad input");
+        let result = check_http_status_for_error(&response);
+        match result {
```
mhammond: it doesn't really matter, but you could probably replace these lines with something like `assert!(matches!(result, Err(HTTPError::BadRequest { code, message }) if code == 400 && message == "Bad input"));`

or even:
```
  let err = check_http_status_for_error(&response).expect_err();
  assert!(matches!(err, HTTPError::BadRequest { code, message } if code == 400 && message == "Bad input"));
```

components/ads-client/src/error.rs
```diff
@@ -0,0 +1,237 @@
...
+    RecordClick(#[from] RecordClickError),
+
+    #[error("Error recording an impressions for a placement: {0}")]
+    RecordImpression(#[from] RecordImpressionError),
+
+    #[error("Error reporting an ad: {0}")]
+    ReportAd(#[from] ReportAdError),
+}
+
+impl GetErrorHandling for ComponentError {
```
mhammond, line 32: there's a very extensive set of errors here that seems like they are all just going to be collapsed into something the consumer really can't do anything with. I think you might eventually want to log many of these errors to, say, sentry (which is supported by the error-support crate), and for your own sanity seem likely to end up squashing some of these together - but there's no harm here, so it seems fine to land with this.

(see https://github.com/mozilla/application-services/blob/main/components/places/src/error.rs for how you might want to report errors to sentry - also note that component is quite a bit more complex than this one and has significantly fewer error types)

components/ads-client/src/instrument.rs
```diff
@@ -0,0 +1,179 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum TelemetryEvent {
+    Init,
+    RenderError,
+    AdLoadError,
+    FetchError,
+    InvalidUrlError,
+}
+
+#[allow(dead_code)]
```
mhammond: we should try and avoid allowing dead code if we can - how is this used? I can't see this used anywhere in this PR. Maybe related to the mocks?

components/ads-client/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+            inner: Mutex::new(MozAdsClientInner::new()),
+        }
+    }
+}
+
+#[uniffi::export]
+impl MozAdsClient {
+    #[uniffi::constructor]
+    pub fn new() -> Self {
+        Self {
```
mhammond: might as well call `Self::default()` here? (aside: we should probably teach uniffi how to export a constructor just using `Default::default`!)

components/ads-client/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    #[handle_error(ComponentError)]
+    pub fn request_ads(
+        &self,
+        moz_ad_configs: Vec<MozAdsPlacementConfig>,
+    ) -> ApiResult<HashMap<String, MozAdsPlacement>> {
+        let inner = self.inner.lock();
+        let placements = inner
+            .request_ads(&moz_ad_configs)
```
mhammond, line 57: I'm not sure how your component is actually going to be used, but the locking here seems like it might not be ideal given the mutex is held while doing network IO. For example, on a very slow network, I could imagine subsequent calls to `record_impression` could end up starving the app of worker threads.

(There's actually no great solution here - not holding the lock while doing the network request just moves the bottle-neck somewhere else.I suspect a queue of requests might be where you end up needing to be?) Or ultimately with an async API?

components/ads-client/src/mars.rs
```diff
@@ -0,0 +1,240 @@
...
+    models::{AdRequest, AdResponse},
+};
+use url::Url;
+use uuid::Uuid;
+use viaduct::Request;
+
+const DEFAULT_MARS_API_ENDPOINT: &str = "https://ads.mozilla.org/v1";
+
+#[cfg_attr(test, mockall::automock)]
+pub trait MARSClient: Sync + Send {
```
mhammond, line 20: This is exactly the same shape as `MozAdsClient`? It might make sense later to define your API in terms of this trait rather than the struct. That would probably end up allowing you to run "mocked" tests directly from the mobile client test code.

components/ads-client/src/test_utils.rs
```diff
@@ -0,0 +1,146 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+#[cfg(test)]
```
mhammond: you can probably just wrap the "mod test_utils" statement in lib.rs to make the entire module conditional on tests, meaning you can remove these individual cfg checks?

mhammond (approved):
there are a few concerns I have here, but there's nothing I see which should prevent this landing to help more people lean in to a polishing and integration effort.

mhammond:
> @mhammond I'm noticing the `android-build-pr` CI step is failing, but I'm not entirely sure what is causing this.

me either - I'm fairly sure that's unrelated to this, especially as there's actually no android changes here yet :)

components/ads-client/src/mars.rs
luc-lisi: Unless I'm misunderstanding your comment, the function names are almost identical but the signature types and response types are pretty different. Our `MARSClient` uses types that are native to the UAPI, but that are very unwieldy for developers to actually use. So `MozAdsClient` exposes similarly named functions but their signatures and return types are changed so developer can work with them without implementing a ton of request/response translation logic.

It's a really interesting point though about mocking this client directly in mobile code though. Maybe there could be a clean way to allow us to build the `MozAdsClient` with a custom defined `MARSClient` so mobile code could create their own custom mocks? But I might leave including that to a different ticket just in the interest in getting this first pass landed if that seems okay.

components/ads-client/src/test_utils.rs
luc-lisi: Oh great call! Making that update.

components/ads-client/src/instrument.rs
luc-lisi: Ah yeah this isn't needed! thanks for catching.

components/ads-client/src/lib.rs
luc-lisi: Yeah really polishing the networking is something we have for a future milestone. If async is coming down the pipe then we would love to play around with that though as we get closer to a finished state!

components/ads-client/src/error.rs
luc-lisi: Oh I like this a lot more! Updating 😄

components/ads-client/src/lib.rs
luc-lisi: Oh nice call, making that change.

components/ads-client/src/mars.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn get_mars_endpoint(&self) -> &str {
+        &self.endpoint
+    }
+
+    /// Updates the client's context_id to the passed value and returns the previous context_id
+    fn cycle_context_id(&mut self) -> String {
+        let old_context_id = self.context_id.clone();
+        self.context_id = Uuid::new_v4().to_string();
+        old_context_id
+    }
```
mikeconley: Sorry for the unsolicited review comment, but I was just perusing this, and it looks like this component is attempting to take ownership of creating the context_id.

That's actually more the purview of the pre-existing context_id component: https://github.com/mozilla/application-services/tree/main/components/context_id

Or is this method purely for testing or something?

components/ads-client/src/error.rs
luc-lisi: There are some 400s we might expect but otherwise you're totally right that it probably is just safest to explicitly match 200, and treat the rest as errors. Updated the code accordingly!

components/ads-client/src/error.rs
luc-lisi: Yeah eventually the plan is that we will need to implement our own custom logging through MARS. Due to privacy concerns surrounding ads, Sentry might not be viable solution although this is still yet to be fully decided on. But for context, a lot of this error scaffolding is put in place so in the future we can develop what would end up being totally custom telemetry/logging that we emit to MARS directly.

But it certainly is overkill right now and may absolutely be something we tear out/re-work later.

components/ads-client/src/mars.rs
luc-lisi: No worries, glad you commented! Sorry, I should have made this clear in the code and will add a TODO to make it explicit. This super simple context_id implementation is just for testing while we validate the basic functionality of this component. We absolutely plan to hook up to your existing context_id component as we iterate. "Storage" in general for context_id/caching, etc is just a future milestone 😄

luc-lisi: If you feel strongly though, I'm more than happy to hook things up to your component before merging this!

Cargo.toml
luc-lisi: Yup good catch! Added.

components/ads-client/src/mars.rs
mikeconley: Ah, okay, just making sure you were aware that context_id existed. I won't scope creep you into adding support for it yet - let's have you focus on getting this first version of the component landed. Thanks!

PR #6882, opened 2025-08-08
[DISCO-3621]: Suggest: add serp categories logic

tiftran:

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -522,17 +522,21 @@ impl<'a> SuggestDao<'a> {
             "#,
         )?;
 
-        let category_rows = category_stmt.query_map(
-            named_params! {
-                ":suggestion_id": suggestion_id
-            },
-            |row| {
-                let category: Option<i32> = row.get(0)?;
-                Ok(category)
-            },
-        )?;
+        let categories: Option<Vec<i32>> = category_stmt
+            .query_map(
+                named_params! {
+                    ":suggestion_id": suggestion_id
+                },
+                |row| {
+                    let category: Option<i32> = row.get(0)?;
+                    Ok(category)
+                },
+            )?
+            .collect::<std::result::Result<Vec<_>, _>>()?
+            .into_iter()
+            .collect();
 
-        Ok(category_rows.filter_map(|res| res.ok().flatten()).collect())
+        Ok(categories.unwrap_or_default())
     }
 
     /// Fetches Suggestions of type Wikipedia provider that match the given query


components/suggest/src/schema.rs
```diff
@@ -806,6 +812,19 @@ impl ConnectionInitializer for SuggestConnectionInitializer<'_> {
                 )?;
                 Ok(())
             }
+            42 => {
+                clear_database(tx)?;
```
ncloudioj, line 816: Don't forget to add the new table to the conditional tables in `clear_database()` so it will be cleared properly for future migrations. Likewise, let's update `SuggestDao::drop_suggestions()` to delete things when the associated RS records get deleted.

You can reference this [PR](https://github.com/mozilla/application-services/pull/6815) for details.

components/suggest/src/db.rs
```diff
@@ -1007,6 +1020,13 @@ impl<'a> SuggestDao<'a> {
                     keyword.rank,
                 )?;
             }
+
+            if let Some(categories) = &suggestion.serp_categories {
+                // for AMP we only want one category
```
ncloudioj: What do you think about dropping this assumption so that it could insert more than 1 category into the table? Doing that would save us another change when we handle multiple categories for AMP.

If we do that, we will need to change the queries above (those `LEFT JOIN` ones), specifically, it needs to run a separate query to load all the categories for a given suggestion similar to how we query `amp_custom_details` after querying the `suggestions` table.

components/suggest/src/schema.rs
```diff
@@ -85,6 +85,12 @@ CREATE TABLE prefix_keywords(
 
 CREATE UNIQUE INDEX keywords_suggestion_id_rank ON keywords(suggestion_id, rank);
 
+CREATE TABLE suggestion_serp_categories(
```
ncloudioj: Nit: all the tables are for suggestion in the suggest DB, so we can also call this `serp_categories` for short.

components/suggest/src/suggestion.rs
```diff
@@ -39,6 +39,7 @@ pub enum Suggestion {
         block_id: i64,
         advertiser: String,
         iab_category: String,
+        categories: Vec<i32>,
```
ncloudioj, line 42:
```suggestion
        serp_categories: Vec<i32>,
```

components/suggest/src/db.rs
```diff
@@ -509,6 +515,26 @@ impl<'a> SuggestDao<'a> {
...
+            named_params! {
+                ":suggestion_id": suggestion_id
+            },
+            |row| {
+                let category: Option<i32> = row.get(0)?;
+                Ok(category)
+            },
+        )?;
+
+        Ok(category_rows.filter_map(|res| res.ok().flatten()).collect())
```
ncloudioj: A subtle issue with `res.ok().flatten()` as it will ignore all the `Err(_)` cases for `res` if any. What we usually prefer is to return an error if anything of `query_map` fails, otherwise return all the query_map results.

Maybe we can do a double-collect trick for that:

```rs
let categories: Option<Vec<i32>> = category_stmt.query_map(
    named_params! {
        ":suggestion_id": suggestion_id
    },
    |row| {
        let category: Option<i32> = row.get(0)?;
        Ok(category)
    },
)?
.collect::<std::result::Result<Vec<_>, _>>()?  // this will error out if any of the iterator entries is `Err(_)`
.into_iter()
.collect();  // this will turn an `Iter<Option<i32>>` into an `Option<Vec<i32>>`

Ok(categories.unwrap_or_default())
```

PR #6885, opened 2025-08-08
Go back to a callback interface for EventSink

bendk:
I realized there was an issue with moving to a trait interface: we want to use `FireAndForget` on JS for the methods, but that only works on a callback interface.

I think a callback interface should be fine for this.  We can still implement these traits in Rust and use them -- that's what the tests do. The only thing we can't do is pass a trait impl into JS, but I don't think we have a reason to do that (BTW, this use-case is where trying to implement `FireAndForget` for a trait interface gets tricky).

  - Note:

### DIFF
--- a/components/support/tracing/src/layer.rs
+++ b/components/support/tracing/src/layer.rs
@@ -24,7 +24,6 @@ static SINKS_BY_TARGET: LazyLock<RwLock<HashMap<String, LogEntry>>> =
 
 static MIN_LEVEL_SINK: RwLock<Option<LogEntry>> = RwLock::new(None);
 
-#[uniffi::export]
 pub fn register_event_sink(target: &str, level: crate::Level, sink: Arc<dyn EventSink>) {
     SINKS_BY_TARGET.write().insert(
         target.to_string(),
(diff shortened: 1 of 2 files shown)


bendk:
IMO, the other main option would be to switch the method to `async` and create a way to run it as a fire-and-forget detached task in Rust.  Maybe we could investigate that in the future, but it doesn't seem worth it at this point.

components/support/tracing/src/layer.rs
```diff
@@ -16,7 +16,7 @@ use tracing::field::{Field, Visit};
 
 struct LogEntry {
     level: tracing::Level,
-    sink: Arc<dyn EventSink>,
```
mhammond: In many ways I preferred how it was before - we still used `Arc<>` everywhere, and only the uniffi entry-point needed to take the `Box<>` - so it just re-wrapped the box in an Arc. I'm not too bothered though.

bendk: That sounds good to me, I changed it to work like that again.  I didn't realize how many Rust consumers of the API we had.

mhammond (approved):

PR #6889, opened 2025-08-12
OAuth related fixes

bendk:
Added a couple small fixes for the Fxa state machine based on @jonalmeida's testing.

The main goal is to add some defensive programming so we don't fail if an app sends us the `BeginOAuthFlow` event while the state machine is already in the middle of an oauth flow (i.e. it's in the `Authenticating` state).  In this case, we should cancel the existing oauth flow and start a new one.

  - Note:


jonalmeida (approved):
Logically, this lgtm and should solve our problem - I haven't tested against this patch yet, but I also don't want to block it.

We did speak offline about an orthogonal issue where we may want to clear only the oauth flow for the service we've started. This also isn't blocking because it isn't yet a supported feature, but would be worth keeping in mind for the future.

PR #6897, opened 2025-08-15
New viaduct backend

bendk:
Added a new viaduct backend trait.  The new backend is UniFFI and async-based and handles settings slightly differently.  See the `new_backend` module for the transition plan.

Added the `hyper` backend to `viaduct` directly.  I don't think we really need to have different crates for each backend, we can use feature flags instead.

Made some tweaks to `Error` so that it's UniFFI compatible.

  - Note:


components/viaduct/src/backend/hyper.rs
```diff
@@ -0,0 +1,166 @@
...
+/// initialize the backend, which is expected during the tests.  The first one will take effect and
+/// all others can be safely ignored.
+#[uniffi::export]
+pub fn init_backend_hyper() -> Result<(), Error> {
+    info!("initializing hyper backend");
+    // Create a multi-threaded runtime, with 1 worker thread.
+    //
+    // This creates and manages a single worker thread.
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
```
bendk: This starts a thread for the viaduct backend, which is exactly what `viaduct-reqwest` was doing behind the scenes.  I thought we might be able to avoid this, but probably not since `hyper` can only run inside a tokio runtime.

bendk:
I think this one is ready to merge.  Here's the migration plan:

- Merging this will immediately upgrade Android to the new backend, which is just a new way to wrap the `Fetch` library.  I tested this and it's working for me.  I'm even more confident because I saw errors in logcat at first because I confused ms with us.  Once I fixed that the errors when away.
- Other applications won't see any change until they opt-in by calling `init_backend()` with a new backend implementation (or something equivalent like `init_backend_hyper()`.  Once this is merged we can work to move swift and Desktop over to the new backend.
   - Swift will switch from the `reqwest` backend to the `hyper` backend.
   - Desktop will replace it's protobuf-based Necko bridge with a new Necko bridge.  I'm not sure exactly how this will work, but there's definitely a way.  I'm hoping that we can call into Necko from Rust.  There's no official Rust bindings, but maybe we could declare `extern "C"` Necko functions and use them.  If not, [we can define extern "C" functions in the Rust code](https://github.com/mozilla/application-services/pull/6018/files#diff-2fddf3239e291de11e4ae3165cda07027b6bb218b24aa0e36916368039729c78) and hook them up in C++-land.
- CLI apps have been moved from the old `dev` backend to the new `hyper` backend.  The code is almost exactly the same and I don't expect to see any issues here.  If we do see issues, there shouldn't be any user impact just things like broken CI jobs that we can fix.
- Once all applications using the new backend API, we can move components over to the new Client API
- Finally, we can remove the old backend and the protobuf dependency.

bendk:
Mark pointed out that this would cause dependency issues in moz-central, since `viaduct-hyper` depends on `hyper-tls` which brings in crates like `openssl` that we don't want (we want to use NSS instead).  I think it also brings in duplicate `windows-sys` dependencies.

To avoid this, I made 2 different hyper-based backends:
  - The `dev` backend, which is a builtin to `viaduct` behind a feature flag.  This depends on `hyper`, but not `hyper-tls`.  It doesn't bring in unwanted dependencies in moz-central, but it also doesn't support HTTPS.  This is what we want to use for tests.
  - The `hyper` backend, which is in a separate `viaduct-hyper` crate.  It does depend on `hyper-tls`, does support HTTPS, and should not be used in a crate that's part of the moz-central workspace.  We can use this for `megazord_ios` and the binaries like the example CLIs.  There are downsides to not including these crates the moz-central workspace, like we'll create a second `target` directory when running the CLIs, but we can accept that for now.
  
Note, there are a couple crates that still depend on `viaduct-hyper` that shouldn't:
  - `suggest` needs it for the benchmark binary.  We can fix this by making the benchmarks a separate crate.
  - `nimbus` needs this for `examples/experiments.rs`.  I'll work with that team to figure out how to handle this.  I'm guessing we can also make this a separate crate.

components/fxa-client/src/internal/config.rs
```diff
@@ -76,6 +76,22 @@ impl Config {
         result
     }
 
+    #[cfg(test)]
+    pub fn set_test_remove_config(&self) {
```
mhammond: I assume this is a typo - s/remove/remote/? Given this is for tests, I do wonder if we should be using something other than our prod urls?

And now I better understand what this is for, a quick comment that this is returning what `.well-known/fxa-client-configuration` returns would be helpful.

components/fxa-client/src/internal/oauth.rs
```diff
@@ -612,14 +612,13 @@ mod tests {
...
         nss::ensure_initialized();
-        // FIXME: this test shouldn't make network requests.
-        viaduct_dev::use_dev_backend();
         let config = Config::new(
             "https://accounts.firefox.com",
             "12345678",
             "https://foo.bar",
         );
         let mut fxa = FirefoxAccount::with_config(config);
+        fxa.state.set_test_remote_config();
```
mhammond: We spoke a bit about this in slack as I found it confusing. Renaming `set_test_remote_config` to something like `mock_well_known_fxa_client_configuration` (maybe `override_*` or `set_well_known_fxa_client_configuration_for_tests` or something - but you said you are thinking fo refactoring this further, so I'll let you think more about that.

components/remote_settings/src/client.rs
```diff
@@ -1694,7 +1694,7 @@ mod test {
 
     #[test]
     fn test_invalid_etag() {
-        viaduct_dev::use_dev_backend();
```
mhammond, line 1697: tbh I'm not "init_backend_dev" is a better name than "use_dev_backend" - the "init_x" pattern is used for all other backends, and this is a kinda "real" backend IIUC (ie, an expectation that *all* viaduct code insists that some "init_*_backend" method has been called seems easy to understand)

(to be clear, I don't care too much about this, just thought I'd mention it)

components/support/viaduct-hyper/src/lib.rs
```diff
@@ -0,0 +1,174 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Viaduct hyper backend
+//!
+//! This implements a backend using `hyper`.
+//! It's essentially the same as the builtin `dev` backend, except it also supports HTTPS.
```
mhammond: I found this comment slightly misleading - it first sounds like this *is* a dev/test backend, but then it sounds like a very real backend that will be used for real crates.

IIUC, this is really just saying "you should never use this when a gecko backend is available"?

[edit: *sob* - even "necko backend" isn't accurate, given there's a fetch backend which is also necko. So yeah, whatever you decide is an improvement here is fine with me]

components/support/viaduct-hyper/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+//! The reason it exists is that so we can avoid the `hyper-tls` dependency in our main crates,
+//! which causes dependency issues when vendored into `moz-central`.
+//!
+//! Because of those dependency issues, only use this for crates that aren't members of the moz-central workspace.
+//! For example, CLI binaries or the ios megazord.
```
mhammond: (aside: I thought we were actually hoping to not need to exclude all these crates eventually? Regardless, I think not vendoring the deps of the ios-megazord and the cli tools is eventually going to be seen as bad - but future us can deal with that)

components/support/viaduct-hyper/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+type Client = hyper::client::Client<Connector, hyper::Body>;
+
+struct HyperBackend {
+    runtime: tokio::runtime::Runtime,
+    client: Client,
+}
+
+/// Set the viaduct backend to the `hyper`-based one.
+///
+/// This is mainly used for iOS and tests.
```
mhammond: as discussed in slack, this is better described as "for iOS and CLI programs" (or like above, maybe even more accurately as "for non-test code which doesn't have the necko backend"? Whatever you decide is fine.

Ditto a few comments below - eg, "during tests" is wrong etc. And it doesn't look like we are ignoring errors here - maybe the entire comment block below should be removed?

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
```diff
@@ -63,9 +52,7 @@ object RustHttpConfig {
      * are sure you are running on an emulator.
      */
     fun allowAndroidEmulatorLoopback() {
-        lock.read {
-            LibViaduct.INSTANCE.viaduct_allow_android_emulator_loopback()
-        }
+        // This is a no-op now that we use the new backend, which doesn't prevent `http` requests.
```
mhammond: Are we sure this is the correct thing to do? The existing https-only mode was intentionally added and I don't think we just want to allow Rust components in Fenix to "accidentally" use http.

Further, all the `addn_allowed_insecure_url` code remains in place, which IIUC, exists only for this.

[edit: slack convo agrees we probably do not want to relax the https-only config for "real" code, unclear what this means to the dev backend in m-c, where we need to avoid vendoring ssl. Maybe `validate_request` moves to the trait?]

components/viaduct/src/backend.rs
```diff
@@ -24,27 +26,27 @@ pub fn note_backend(which: &str) {
...
+pub fn send(request: crate::Request) -> Result<crate::Response, crate::ViaductError> {
     validate_request(&request)?;
     get_backend().send(request)
 }
 
-pub fn validate_request(request: &crate::Request) -> Result<(), crate::Error> {
+pub fn validate_request(request: &crate::Request) -> Result<(), crate::ViaductError> {
     if request.url.scheme() != "https"
         && match request.url.host() {
             Some(url::Host::Domain(d)) => d != "localhost",
```
mhammond: Another comment says we no longer restrict http but this code remains. It also still checks `addn_allowed_insecure_url` which is for that Android thing.

ah - after chatting on slack, the new "client" doesn't make this check - and I think we agreed that for this cut, we should keep all https restrictions even if we later relax them for the dev backend

components/viaduct/src/client.rs
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::{new_backend::get_backend, Request, Response, Result};
+
+/// HTTP Client
+///
+/// This represents the "new" API.  It only works if the application has initialized a new-style
+/// backend.  Don't use this for real components until all applications have switched over.
```
mhammond: what's the plan here? The comment says it should not be used by components - is the intention that there is a "fast follow" to move all components over?)

edit: I see some comments around this in new_backend.rs - I wonder if the README should cover some of this, then both these files can point to that?

components/fxa-client/src/internal/oauth.rs
bendk: I ended up making a `new_with_mock_well_known_fxa_client_configuration` method which does essentially the same thing but in 1 call.

components/viaduct/android/src/main/java/mozilla/appservices/httpconfig/HttpConfig.kt
bendk: Yeah, let's not lose `validate_request`.  I moved that function to `settings.rs` and we're now calling it from both the old and new code paths (`Request.send` and `Client.send`).

components/remote_settings/src/client.rs
bendk: Added a comment in the README about the naming.

components/support/viaduct-hyper/src/lib.rs
bendk: Yeah, I totally agree with this.  I think it'll come down to a discussion with the build engineers that we have once the dust clears from the monorepo work.

components/support/viaduct-hyper/src/lib.rs
```diff
@@ -0,0 +1,166 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Viaduct hyper backend
+//!
+//! This implements a backend using `hyper`.
+//! Unlike the `dev` backend it also supports HTTPS.
+//!
+//! In order to support HTTPS it depends on `hyper-tls`, which we don't want to vendor into moz-central at this time.
```
mhammond, line 10: still not quite sure how we can ban https given `validate_request`, but if this is an actual problem it can be a future problem.

mhammond:
looks like the lint failure is real.

PR #6902, opened 2025-08-18
Bug 1983587 - Namespace dismissals of dynamic suggestions by suggestion_type

moritzbeier:
We want to namespace dismissals to avoid clashes between dismissal keys. We do this by creating a new table for dismissed dynamic suggestions that also has a column for the `suggestion_type`.

Since the only live dynamic suggestion is the VPN one, it is still easy to migrate existing dismissals. Ideally, we want to land this before important dates suggestions go live to avoid complicating the migration.

  - Note:

### DIFF
--- a/components/suggest/src/db.rs
+++ b/components/suggest/src/db.rs
@@ -905,7 +905,10 @@ impl<'a> SuggestDao<'a> {
                   s.provider = ?
                   AND k.keyword = ?
                   AND d.suggestion_type IN ({})
-                  AND NOT EXISTS (SELECT 1 FROM dismissed_dynamic_suggestions WHERE dismissal_key = s.url AND suggestion_type = d.suggestion_type)
+                  AND NOT EXISTS (
+                    SELECT 1 FROM dismissed_dynamic_suggestions 
+                    WHERE dismissal_key = s.url AND suggestion_type = d.suggestion_type
+                  )
                 ORDER BY
                   s.score ASC, d.suggestion_type ASC, s.id ASC
                 "#,
--- a/components/suggest/src/store.rs
+++ b/components/suggest/src/store.rs
@@ -3887,6 +3887,7 @@ pub(crate) mod tests {
         });
 
         // Make sure the suggestions are initially fetchable.
+        assert!(!store.inner.any_dismissed_suggestions()?);
         let suggestions_0: Vec<Suggestion> =
             store.fetch_suggestions(SuggestionQuery::dynamic("aaa", &["aaa"]));
         let suggestions_1: Vec<Suggestion> =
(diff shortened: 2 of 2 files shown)


components/suggest/src/store.rs
```diff
@@ -3883,8 +3916,8 @@ pub(crate) mod tests {
         );
 
         // Dismiss the first suggestion.
-        assert_eq!(suggestions[0].dismissal_key(), Some("dk0"));
-        store.inner.dismiss_by_suggestion(&suggestions[0])?;
+        assert_eq!(suggestions_0[0].dismissal_key(), Some("dk0"));
+        store.inner.dismiss_by_suggestion(&suggestions_0[0])?;
```
0c0w3, line 3921: Could you assert the following here please?

* `store.inner.is_dismissed_by_suggestion` returns true for this suggestion
* `store.inner.any_dismissed_suggestions` returns true

components/suggest/src/store.rs
```diff
@@ -3904,8 +3937,8 @@ pub(crate) mod tests {
         );
 
         // Dismiss the second suggestion.
-        assert_eq!(suggestions[1].dismissal_key(), Some("dk1"));
-        store.inner.dismiss_by_suggestion(&suggestions[1])?;
+        assert_eq!(suggestions_0[1].dismissal_key(), Some("dk1"));
+        store.inner.dismiss_by_suggestion(&suggestions_0[1])?;
```
0c0w3, line 3945: Here too

components/suggest/src/store.rs
```diff
@@ -3916,6 +3949,22 @@ pub(crate) mod tests {
...
+        );
+        assert_eq!(
+            store.fetch_suggestions(SuggestionQuery::dynamic("bbb", &["bbb"])),
+            vec![Suggestion::Dynamic {
+                suggestion_type: "bbb".to_string(),
+                data: None,
+                dismissal_key: Some("dk0".to_string()),
+                score: DEFAULT_SUGGESTION_SCORE,
+            },],
+        );
```
0c0w3, line 3973: The opposite here please, i.e., `store.inner.is_dismissed_by_suggestion` returns false for the bbb suggestion

components/suggest/src/db.rs
```diff
@@ -905,7 +905,7 @@ impl<'a> SuggestDao<'a> {
                   s.provider = ?
                   AND k.keyword = ?
                   AND d.suggestion_type IN ({})
-                  AND NOT EXISTS (SELECT 1 FROM dismissed_suggestions WHERE url = s.url)
+                  AND NOT EXISTS (SELECT 1 FROM dismissed_dynamic_suggestions WHERE dismissal_key = s.url AND suggestion_type = d.suggestion_type)
```
0c0w3: Nit: This is a little long, so could you wrap it somehow please? We don't have a totally consistent style though. I would do something like:

```
                  AND NOT EXISTS (
                      SELECT 1 FROM dismissed_dynamic_suggestions
                      WHERE dismissal_key = s.url AND suggestion_type = d.suggestion_type
                  )
```

0c0w3 (approved):
This looks great, thanks!

0c0w3 (approved):

PR #6903, opened 2025-08-20
Ensure find_places_db() is XDG aware.

mhammond:
In mozilla-central, this crate fails the lint warning at https://searchfox.org/firefox-main/source/tools/lint/dot-mozilla-reference.yml

This patch just takes the approach in https://searchfox.org/firefox-main/rev/60308bc3792ef201b82377682de068a5a1c72575/toolkit/crashreporter/crash_helper_server/src/logging/env.rs#34, but without support for `MOZ_LEGACY_HOME`, because this code is only ever run interactively by people who are unlikely to ever have that set - that's an improvement we could make later if needed.

cc @standard8 because with this change I intend adding this file to dot-mozilla-reference.yml in the section where "Files which reference .mozilla and have XDG handling in place" items are listed to https://phabricator.services.mozilla.com/D246875

### DIFF
--- a/components/support/find-places-db/src/lib.rs
+++ b/components/support/find-places-db/src/lib.rs
@@ -8,7 +8,7 @@
 // If we ever need to split it out again we should contact Thom, who I'm sure
 // would be happy to give us ownership of the crate.
 
-use anyhow::{bail, format_err, Result};
+use anyhow::{format_err, Result};
 use error_support::{debug, info, trace, warn};
 use std::{fs, path::PathBuf, process};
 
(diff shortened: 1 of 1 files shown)


lissyx:
> but without support for `MOZ_LEGACY_HOME`, because this code is only ever run interactively by people who are unlikely to ever have that set - that's an improvement we could make later if needed.

I dont get the point here. You're assuming you're going to break the fallback-to-previous case case?

It's hard to judge not knowing what this crate is for.

lissyx:
> > but without support for `MOZ_LEGACY_HOME`, because this code is only ever run interactively by people who are unlikely to ever have that set - that's an improvement we could make later if needed.
> 
> I dont get the point here. You're assuming you're going to break the fallback-to-previous case case?
> 
> It's hard to judge not knowing what this crate is for.

At least Snap builds will force `MOZ_LEGACY_HOME` for some time, and it's not unlikely many users will want it as well.

gcp:
>The application-services library primary consumers are Fenix (Firefox on Android) and Firefox iOS.

This is in the `README.md` (I had the same question as @lissyx about what this is for), but if the above is true I'm confused why `.mozilla` would even be a thing here.

mhammond:
The primary use-case here is engineers who work on the application-services repo and want to make use of some of the "import from desktop places" capabilities we have so they have a local database they can experiment with (eg, performance analysis etc). It doesn't ship with app-services. It wouldn't be too bad to just remove this entirely and force the developer to specify a full path (which they can do regardless), but I don't see how keeping this convenience hurts.

lissyx:
> The primary use-case here is engineers who work on the application-services repo and want to make use of some of the "import from desktop places" capabilities we have so they have a local database they can experiment with (eg, performance analysis etc). It doesn't ship with app-services. It wouldn't be too bad to just remove this entirely and force the developer to specify a full path (which they can do regardless), but I don't see how keeping this convenience hurts.

so a developper running with `MOZ_LEGACY_HOME` may fail there? Handing the legacy path is not super complicated and it would be better in case anybody tries to make use of this for something else in the future.

components/support/find-places-db/src/lib.rs
```diff
@@ -40,23 +40,29 @@ impl PlacesLocation {
...
+            let mut path =
+                dirs::home_dir().ok_or_else(|| anyhow::anyhow!("No home directory found!"))?;
             // ~/Library/Application Support/Firefox/Profiles
             path.extend(&["Library", "Application Support", "Firefox", "Profiles"]);
+            path
         } else {
             // I'm not actually sure if this is true for all non-macos unix likes.
+            // Using `dirs::config_dir()` gives us XDG support.
+            let mut path =
+                dirs::config_dir().ok_or_else(|| anyhow::anyhow!("No config directory found!"))?;
```
lissyx: So here we end up with `$XDG_CONFIG_HOME/.mozilla/firefox` ? That's incorrect, it should be `$XDG_CONFIG_HOME/mozilla/firefox`

mhammond: huh - I was taking this from https://searchfox.org/firefox-main/rev/60308bc3792ef201b82377682de068a5a1c72575/toolkit/crashreporter/crash_helper_server/src/logging/env.rs#38-42 which also seems to use `.mozilla` in that case. That other code also has a strange `||` which seems odd - it will use the home directory if it exists even if `MOZ_LEGACY_HOME` isn't set - I didn't copy those semantics.

I'll update this patch as you note though, thanks!

lissyx: Looks like it's my mistake :(

lissyx: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1984557

bendk (approved):
Looks good to me.  I don't totally understand all the cases here, but it matches the other code so I'm happy to ship this.

PR #6904, opened 2025-08-20
gradle tweaks for the monorepo - updated build.gradle for glean.

mhammond:
Because there's no "global" build.gradle in mozilla-firefox where glean plugin
references can live, we use the same approach already used there - each component's
build.gradle duplicates that configuration.

cc @rvandermeulen in case this is of interest for you.

### DIFF
--- a/components/fxa-client/android/build.gradle
+++ b/components/fxa-client/android/build.gradle
@@ -1,3 +1,23 @@
+buildscript {
+    // in mozilla-central there's no build.gradle suitable for the maven/glean config, so each component gets it.
+    if (gradle.hasProperty("mozconfig")) {
+        repositories {
+            gradle.mozconfig.substs.GRADLE_MAVEN_REPOSITORIES.each { repository ->
+                maven {
+                    url = repository
+                    if (gradle.mozconfig.substs.ALLOW_INSECURE_GRADLE_REPOSITORIES) {
+                        allowInsecureProtocol = true
+                    }
+                }
+            }
+        }
+
+        dependencies {
+            classpath libs.glean.gradle.plugin
+        }
+    }
+}
+
 plugins {
     alias libs.plugins.python.envs.plugin
 }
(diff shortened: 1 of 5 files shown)


components/autofill/android/lint-baseline.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 8.12.1" type="baseline" client="gradle" dependencies="false" name="AGP (8.12.1)" variant="all" version="8.12.1">
+
+    <issue
```
ncalexan: This worries me, but surely I'm missing something.  Is this lint saying that we're going to regress our 16kb alignment (needed in November, IIRC) by continuing to use JNA 5.14.0?

ncalexan: Generally, I'd rather see us disable this lint in these projects rather than boil all of this version specific nonsense into these files.  Is that possible?

components/nimbus/android/lint-baseline.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 8.12.1" type="baseline" client="gradle" dependencies="false" name="AGP (8.12.1)" variant="all" version="8.12.1">
+
+    <issue
```
ncalexan: Doesn't block this patch, but surely we can get a ticket to fix these trivial lint issues?

ncalexan (approved):
Sure.  Centralizing all the Maven repository handling goo is A Thing We Care About that I'm hoping Alex Hochheiden makes progress on soon.

components/autofill/android/lint-baseline.xml
rvandermeulen: Agreed that we should just suppress this lint for now so we don't have to baseline it in all the places. We're already taking steps to avoid this problem with the apps that actually need to pass Play Store review at the end of the day, so in practice this isn't telling us anything useful that we didn't already know.

mhammond: sgtm, I just assumed this was somehow "normal" - I'll work out how to disable the lints.

mhammond:
I just dropped all the linter related changes here and I'll tackle that later.

PR #6906, opened 2025-08-20
Bug 1979764: Use partial index to speed up slower places queries

skhamis:
[Bug 1979764](https://bugzilla.mozilla.org/show_bug.cgi?id=1979764) ran a profiler on their android device and identified a really slow sync query. 

I spun up a new sync performance dash (sorry not public): https://sql.telemetry.mozilla.org/dashboard/sync-engine-performance so that we can better track some of these things on a broader level.

This PR speeds up that query by:
1. Adding a [partial index](https://sqlite.org/partialindex.html?utm_source=chatgpt.com) for outgoing syncs `(changed + visible)` 
2. Make the queries index-friendly by allowing the SQLite planner can actually use the index (change from `NOT hidden` to `hidden = 0`

I also included some new benchmark tests to prove that we do indeed see a speedup. Below is from running the nifty places benches we had:

we go from `~124.20 µs` to `41.810 µs` which is about a ~67% speedup increase!
```
Populating test database...
Done populating test db
Populating test database...
Done populating test db
outgoing_candidates: original WHERE, NO partial index
                        time:   [**124.20 µs 124.62 µs 125.04 µs**]
                        change: [−0.9627% −0.5078% −0.0667%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  8 (8.00%) high mild
  1 (1.00%) high severe

outgoing_candidates: index-friendly WHERE + partial index
                        time:   [**41.748 µs 41.810 µs 41.888 µs**]
                        change: [−0.9751% −0.5086% −0.0790%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  8 (8.00%) high mild
  1 (1.00%) high severe

outgoing_candidates: original WHERE + partial index
                        time:   [124.14 µs 124.40 µs 124.71 µs]
                        change: [−1.1175% −0.6187% −0.1143%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 13 outliers among 100 measurements (13.00%)
  8 (8.00%) high mild
  5 (5.00%) high severe
```

  - Note:


components/places/src/db/schema.rs
```diff
@@ -326,6 +326,13 @@ pub fn upgrade_from(db: &Connection, from: u32) -> rusqlite::Result<()> {
                 rusqlite::params![NEW_SQL],
             )?;
         }
+        18 => {
+            // Create the new indexes by just calling the shared schema file
+            // idx_places_outgoing_by_frecency and idx_places_visible_by_frecency
+            db.execute_batch(CREATE_SHARED_SCHEMA_SQL)?;
+            // Manually call analyze so the planner can start using the indexes immediately
+            db.execute("ANALYZE", [])?;
```
mhammond: Given this is going to execute close to startup and might block some features until the indexes have been built, I wonder if we should limit the analyze to, say, just `moz_places` (the docs say that specifying a table will analyze all indexes associated with it, which sound like what we want?)

components/places/src/storage/history.rs
```diff
@@ -1393,8 +1393,8 @@ pub fn get_visit_infos(
          JOIN moz_historyvisits v
            ON h.id = v.place_id
          WHERE v.visit_date BETWEEN :start AND :end
-           AND ((1 << visit_type) & :allowed_types) != 0 AND
```
mhammond: I guess these changes are trying to make use of one of the new indexes, probably `idx_places_visible_by_frecency`, but my experiments show it is not used - I see:
```
SCAN h USING INDEX frecencyindex
CORRELATED SCALAR SUBQUERY 1
SEARCH v USING INDEX placedateindex (place_id=?)
```

components/places/sql/create_shared_schema.sql
```diff
@@ -44,6 +44,16 @@ CREATE INDEX IF NOT EXISTS lastvisitdateremoteindex ON moz_places(last_visit_dat
 CREATE UNIQUE INDEX IF NOT EXISTS guid_uniqueindex ON moz_places(guid);
 CREATE INDEX IF NOT EXISTS originidindex ON moz_places(origin_id);
 
+-- Some partial indexes to help speed up expensive outgoing calcs
```
mhammond: given these really are targetted at very specific queries, I think we should be explicit - eg, maybe name the exact function name with the query(s?) we are targeting?

testing/separated/places-bench/src/database.rs
```diff
@@ -130,3 +130,184 @@ pub fn bench_match_url(c: &mut Criterion) {
         match_url(db, "https://hg.mozilla.org/mozilla-central").unwrap()
     });
 }
+
+// Outgoing sync query benches
+
+// A helper function that benches can use to validate what queries are doing
+fn explain(name: String, conn: &PlacesDb, sql_template: &str, limit: i64) {
```
mhammond, line 139: this is great, but is only exercising one of the indexes?

components/places/src/storage/history.rs
skhamis: The index is actually for the one above `get_top_frecent_site_infos` since it has the
`AND (h.last_visit_date_local + h.last_visit_date_remote) != 0`

I changed hidden to be  = 0 instead of `NOT` but I realize maybe I shouldn't do that for anything not taking advantage of the indexes to reduce confusion

components/places/src/db/schema.rs
skhamis: Great idea!

testing/separated/places-bench/src/database.rs
```diff
@@ -130,3 +130,284 @@ pub fn bench_match_url(c: &mut Criterion) {
...
+        );
+    }
+}
+
+/*
+ * Benchmarking speeding up get_top_frecent_site_infos by adding a partial index to help speed up the query
+*/
+
+const VISIBLE_BY_FRECENCY_SQL: &str = r#"
+    SELECT h.frecency, h.title, h.url
```
mhammond: this isn't the exact query used. When I tried the exact query used, I didn't see the new index - I did however see the `frecencyindex` used in both "old" and "new" cases. Once we get this right, should we remove that other index? Regardless, it looks like this should maybe copy both the old and new exact sql like you did above?

PR #6917, opened 2025-08-29
Replace rinja with askama in start-bindings

mhammond:
@moztcampbell hit this in some obscure way


moztcampbell:
"some obscure way" was a version conflict on `rustc-hash` which seemed happy enough with `cargo -p rustc-hash` to fix it

PR #6921, opened 2025-09-02
Clean up version catalog dependency naming and Gradle dependency sorting

rvandermeulen:
This is keeping things in sync with bug 1986451 to hopefully avoid adding any new complications to the repo migration work. It isn't making any functional changes to any dependencies.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,15 +5,15 @@
 
 [versions]
 # AGP
-android-plugin = "8.12.0"
+android-gradle-plugin = "8.12.0"
 
 # Google
 protobuf = "4.31.1"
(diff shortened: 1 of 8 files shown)


gradle/libs.versions.toml
```diff
@@ -50,15 +50,15 @@ python-envs-plugin = "0.0.31"
 
 [libraries]
 # AGP
-tools-android-plugin = { group = "com.android.tools.build", name = "gradle", version.ref = "android-plugin" }
```
mhammond, line 53: some of these appear unused in the repo - I wonder if we can actually just remove them? (regardless though, this still seems like an improvement - unused and consistent with m-c is better than unused and inconsistent). I might also be very wrong about how this actually ties together.

PR #6922, opened 2025-09-03
Use ICU4X instead of unicode-normalization and unicase

hsivonen:

  - Note:

Following the docs didn't work on Ubuntu due to NSS integration failing to build, which is why I was unable to run the tests.

This change is not supposed to change observable behavior.

This is all about reducing dependencies (both Firefox and Application Services already depended on `icu_normalizer`, so this removes the duplicative dependency `unicode-normalization`) and about using dependencies that Firefox is trying to converge towards (`icu_casemap` is part of ICU4X despite Firefox not already having it vendored).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Added request caching mechanism using SQLite with configurable TTL and max size.
 - Added configuration options for the cache.
 - Deserialize callbacks with `url::Url`
+- Support for multiple ads request (with count)
 
 ### Relay
 - **⚠️ Breaking Change:** The error handling for the Relay component has been refactored for stronger forward compatibility and more transparent error reporting in Swift and Kotlin via UniFFI.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,7 @@ dependencies = [
  "uuid",
  "viaduct",
  "viaduct-dev",
+ "viaduct-reqwest",
 ]
 
 [[package]]
@@ -2280,15 +2281,6 @@ dependencies = [
 
 [[package]]
 name = "itertools"
-version = "0.10.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a9a9d19fa1e79b6215ff29b9d6880b706147f16e9b1dbb1e4e5947b5b02bc5e3"
-dependencies = [
- "either",
-]
-
-[[package]]
-name = "itertools"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "413ee7dfc52ee1a4949ceeb7dbc8a33f2d6c088194d9f922fb8318faf1f01186"
@@ -3410,9 +3402,9 @@ dependencies = [
 
 [[package]]
 name = "prost"
-version = "0.12.0"
+version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa8473a65b88506c106c28ae905ca4a2b83a2993640467a41bb3080627ddfd2c"
+checksum = "2796faa41db3ec313a31f7624d9286acf277b52de526150b7e69f3debf891ee5"
 dependencies = [
  "bytes",
  "prost-derive",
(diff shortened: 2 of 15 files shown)


hsivonen:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1986265

hsivonen:
AFAICT, the cargo-audit failures are unrelated to this PR and also fail on `main`.

hsivonen:
The other failures are real failures to remove the combining marks, though.

hsivonen:
Looks like local debugging is needed to understand what's going wrong. `is_combining_mark` seems to work when testing on the Rust Playground.

mhammond:
I'll leave the review of the actual changes to the suggest team, but I think these new crates will block updating app-services in m-c until they are "audited" there - so we probably can't land this until they are.

mhammond:
hrm, I just noticed https://bugzilla.mozilla.org/show_bug.cgi?id=1986265, which is implying that some of these libs would already be in m-c. I'm not sure if it's just the rust implementations which are not, or that I'm simply wrong :) But I had a quick look at Cargo.lock in m-c and noticed some of the crates aren't already listed there, which implies an audit might be necessary.

hsivonen:
`icu_casemap` and `icu_casemap_data` are not already in Firefox. The rest of the crates are.

I was expecting `icu_casemap` and `icu_casemap_data` to be audited as part of vendoring this to Firefox. I'm not sure if we have a workflow for auditing them ahead of time, but I'll look into it.

hsivonen:
There isn't much to see in the audit delta, but I posted https://phabricator.services.mozilla.com/D263566 in case it helps the process along.

hsivonen:
The iOS failure looks unrelated.

hsivonen:
Looks like GitHub doesn't allow me to request review. CC @0c0w3

mhammond:
> I was expecting `icu_casemap` and `icu_casemap_data` to be audited as part of vendoring this to Firefox. I'm not sure if we have a workflow for auditing them ahead of time, but I'll look into it.

Thanks! It should be fine to request review on just the audit bits and land them ahead of time - Nika has explicitly said this is OK and we've done that in the past for things like this

megazords/ios-rust/DEPENDENCIES.md
```diff
@@ -510,6 +510,7 @@ The following text applies to code linked from these dependencies:
 [lalrpop-util](https://github.com/lalrpop/lalrpop),
 [lazy_static](https://github.com/rust-lang-nursery/lazy-static.rs),
 [libc](https://github.com/rust-lang/libc),
+[linux-raw-sys](https://github.com/sunfishcode/linux-raw-sys),
```
hsivonen: I don't know why this line wasn't here before. AFAICT, I don't add any new way to depend on this, and this dependency is pre-existing.

0c0w3:
I'm out of my depth on the unicode details but the main things I care about are that the suggest tests still pass and that merging this PR does not indefinitely block vendoring into m-c. We don't have any planned upcoming changes to suggest, so we don't have any plans to vendor in the near future, but I don't want to be blocked for a long time if something comes up. I don't think we have any planned changes to search either but I'm not as clear on that.

hsivonen:
The `icu_casemap` audit has now landed: https://bugzilla.mozilla.org/show_bug.cgi?id=1986721

hsivonen:
> I'm out of my depth on the unicode details but the main things I care about are that the suggest tests still pass

The tests pass now.

> and that merging this PR does not indefinitely block vendoring into m-c. We don't have any planned upcoming changes to suggest, so we don't have any plans to vendor in the near future, but I don't want to be blocked for a long time if something comes up.

It would be really nice to vendor this change to m-c sooner than later so that the vendoring script could be updated not to allow further dependencies on non-ICU4X crates that duplicate ICU4X functionality and, more importantly, data.

`icu_casemap` is now pre-audited, so vendoring this shouldn't require new audits.

hsivonen:
The failing test says:

test encryption::keydb_test::test_nss_key_manager ... FAILED

which looks unrelated to the changes made here.

0c0w3 (approved):
I'm very sorry for the slow review. I've been swamped for the past few months and forgot about this.

0c0w3 (dismissed):

PR #6923, opened 2025-09-03
Monorepo build and version tweaks for nimbus-fml.

mhammond:
* Update Cargo.toml for m-c.
* The gradle files are modified so they work correctly in m-c.
* Upgraded "heck" to 0.5, which required some code changes.
* A couple other version bumps to match m-c which didn't require code changes.
* Includes a bonus bump of rkv in nimbus itself.
* Hacky work-arounds for the fact that the swift and kotlin tests have been broken for years
  via features. This avoids needing to vendor `jsonschema` into m-c, but keeps the door open
  to fixing these tests (and requiring a vendor of jsonschema) at that time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,20 +47,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.8.7"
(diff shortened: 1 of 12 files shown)


freshstrangemusic (approved):
rs+

PR #6938, opened 2025-09-09
Stop building Android x86

rvandermeulen:
To be landed after bug 1987789 does.


skhamis (approved):
Bit of a rs but everything here LGTM!

PR #6970, opened 2025-09-23
docs(relay): add API documentation for RelayClient and its methods

groovecoder:
Add detailed Rust doc comments to `RelayClient`, `RelayAddress`, and all public methods to clarify usage, parameters, authentication, and known server limitations for the Relay API client.

     - This PR has no breaking API changes
- [x] **Quality**: This PR builds and tests run cleanly.
- [x] **Tests**: This PR only change code comments.
- [x] **Changelog**: This PR only changes code comments.
  - ~Any new dependencies are accompanied by a summary of the due diligence applied in selecting them.~

### DIFF
--- a/components/relay/src/lib.rs
+++ b/components/relay/src/lib.rs
@@ -13,12 +13,33 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 use viaduct::{header_names, Method, Request};
 
+/// Represents a client for the Relay API.
+///
+/// Use this struct to connect and authenticate with a Relay server,
+/// managing authorization to call protected endpoints.
+///
+/// # Authorization
+/// - Clients should use the [fxa_client::FirefoxAccount::getAccessToken()] function
+///   to obtain a relay-scoped access token (scope: `https://identity.mozilla.com/apps/relay`).
+/// - Then, construct the [`RelayClient`] with the access token.
+///   All requests will then be authenticated to the Relay server via `Authorization: Bearer {fxa-access-token}`.
+/// - The Relay server verifies this token with the FxA OAuth `/verify` endpoint.
+/// - Clients are responsible for getting a new access token when needed.
 #[derive(uniffi::Object)]
 pub struct RelayClient {
+    /// Base URL for the Relay server.
     server_url: String,
+    /// Optional authentication token for API requests.
     auth_token: Option<String>,
 }
 
+/// Represents a Relay email address object returned by the Relay API.
+///
+/// Includes metadata and statistics for an alias, such as its status,
+/// usage stats, and identifying information.
+///
+/// See:
+/// https://mozilla.github.io/fx-private-relay/api_docs.html
 #[derive(Debug, Deserialize, uniffi::Record)]
 pub struct RelayAddress {
     pub mask_type: String,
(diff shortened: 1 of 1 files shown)


components/relay/src/lib.rs
```diff
@@ -13,12 +13,33 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 use viaduct::{header_names, Method, Request};
 
+/// Represents a client for the Relay API.
+///
+/// Use this struct to connect and authenticate with a Relay server,
+/// managing authorization to call protected endpoints.
+///
+/// # Authorization
+/// - Clients should user the `FirefoxAccount.getAccessToken()` function
```
mattreaganmozilla:
```suggestion
/// - Clients should use the `FirefoxAccount.getAccessToken()` function
```

mattreaganmozilla:
@groovecoder Thank you for adding these docs 🚀

components/relay/src/lib.rs
```diff
@@ -13,12 +13,33 @@ use serde::{Deserialize, Serialize};
...
     auth_token: Option<String>,
 }
 
+/// Represents a Relay email address object returned by the Relay API.
+///
+/// Includes metadata and statistics for an alias, such as its status,
+/// usage stats, and identifying information.
+///
+/// See:
+/// https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/api/v1/docs/redoc/#tag/emails/operation/relayaddresses_retrieve
```
mhammond: should this link to a .md file or similar in some git repo? This url smells fragile.

components/relay/src/lib.rs
mhammond: I have no idea if/how any of this works in terms of rustdocs etc 😅 - but maybe  `fxa_client::FirefoxAccount::getAccessToken()`? I doubt either of these will render a workable link though.

mhammond (approved):
gotta love better docs, thanks!

components/relay/src/lib.rs
bendk: I think that's the path to use, but also with brackets: `[fxa_client::FirefoxAccount::getAccessToken()]`.  That should render a correctlink in rustdoc.  It won't work for the generated Kotlin/Swift/JS docs, but it's something we could eventually get working.

Also: the parens are optional and backticks are allowed, but ignored.
https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

components/relay/src/lib.rs
groovecoder: Good idea. I've opened https://github.com/mozilla/fx-private-relay/pull/5903 to publish the API docs to the Relay repo's GitHub Pages site. When we get that url working I'll update this.

components/relay/src/lib.rs
groovecoder: Yeah, I thought about trying to figure it out, but then even if I got the rust docs to link together, I doubt it would carry thru to the Swift or Kotlin docs too, which is the main goal here.

components/relay/src/lib.rs
groovecoder: Updated to https://mozilla.github.io/fx-private-relay/api_docs.html.

PR #6976, opened 2025-09-26
Bug 1887958 - call `disconnect` from the AuthIssues state

bendk:
This works because a side effect of moving to the internal `Disconnect` state is calling the `disconnect()` method.

  - Note:


bendk:
> This works because a side effect of moving to the internal Disconnect state is calling the disconnect() method.

This feels a bit weird to me at this point.  It made sense when the state machine checker was actively checking our behavior against the Fenix behavior, but that's not the case anymore.  I wonder if we should try to ditch some abstractions and just call the methods directly.

Related to the above point -- the unit tests don't seem like they're giving us much value at this point.  The test code is a direct reflection of the actual code, I'm not sure what errors they would actually catch.

PR #6981, opened 2025-09-30
FxA: Track location when parsing URLs

bendk:
I'm hoping this will help us track down https://bugzilla.mozilla.org/show_bug.cgi?id=1991756

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/util.rs
+++ b/components/fxa-client/src/internal/util.rs
@@ -52,3 +52,52 @@ impl Xorable for [u8] {
         }
     }
 }
+
+pub fn parse_url(url: &str, when: impl Into<String>) -> Result<url::Url> {
+    url::Url::parse(url).map_err(|_| Error::MalformedUrl {
+        sanitized_url: sanitized_url(url),
+        when: when.into(),
+    })
+}
+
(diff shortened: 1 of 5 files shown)


components/fxa-client/src/internal/util.rs
```diff
@@ -52,3 +52,27 @@ impl Xorable for [u8] {
...
+    url.join(path).map_err(|_| Error::MalformedUrl {
+        sanitized_url: sanitized_url(url.as_str()),
+        when: when.into(),
+    })
+}
+
+fn sanitized_url(url: &str) -> String {
+    // Remove everything after the `?` char, this is the URL params where all the auth data
+    // goes.
+    match url.split_once("?") {
```
skhamis: This is non-blocking since i'm not entirely sure, but is it possible for fragments to occur before the query? if so we can maybe include `split_once(['?','#'])` but if not ignore.

skhamis (approved):
Nice! One minor nit/question but otherwise this looks great and will be helpful for identifying the root cause (at least area) where these issues are occurring.

components/fxa-client/src/internal/util.rs
bendk: Great point!  I think they should go after the query, but there's also the possibility of only a fragment.  I'm going to update the code to match your suggestion.

PR #6992, opened 2025-10-03
MPP-4337 - feat(relay): improve API error handling to extract error c…

groovecoder:
…ode and details from JSON

Update Relay error handling logic to parse both `error_code` and `detail` fields from API error responses. This enables propagating the exact error code (if present) alongside the detail message when reporting API errors.

Introduces a helper to extract JSON API errors and updates tests to cover new parsing logic for various error payload shapes.

  - Note:


components/relay/src/error.rs
```diff
@@ -44,10 +51,21 @@ impl GetErrorHandling for Error {
...
-            })
-            .report_error("relay-api-error"),
+            Error::RelayApi(detail) => {
+                // Try to parse error_code and detail from the provided string
+                // Accept both {"error_code", "detail"} or {"detail"} JSON.
+                let parsed: Option<ApiErrorJson> = serde_json::from_str(detail).ok();
+                let code = parsed
+                    .as_ref()
+                    .and_then(|j| j.error_code.clone())
+                    .unwrap_or_else(|| "unknown".to_string());
```
mhammond, line 64: this all seems a little odd to me in terms of how consumers are expected to treat this. Splitting `code` out implies you expect consumers to explicitly check it for some purpose - I assume via a http response - but the types and values seem wrong there - eg, some particular 4XX value, but it would be checked as a string, and somehow they'd need to learn "unknown" is a valid "code"?  IOW, I'm wondering why this isn't an explicit status code, with a default value of some 5XX value, or some other error representing a network error, etc.

OTOH, maybe it's purely for diagnostics etc, in which case you might as well return the original error without any massaging (ie, just keep details about the "code", along with any error message from the server in the same string.

That said, this is your API - there's nothing *wrong* here, just seems odd, so I'll approve it and leave any decisions for you.

mhammond (approved):

components/relay/src/error.rs
groovecoder: Good points. We do expect consumers to check the code in order to adjust the client logic or show a different error message. I think I'll try to find a way to surface up both the HTTP status code AND a Relay-specific error "sub-code".

PR #6996, opened 2025-10-05
Remove sync15's "standalone-sync" feature, remove UDL for a single enum.

mhammond:
* Remove sync15's "standalone-sync" feature, remove UDL for a single enum.
* Remove the udl for sync15 - it just had one enum.
* "Bonus" fxa-client, logins, places and sync15 all updated to 2024, which is a bigger diff than I anticipated. Seems ok though?
* 2024 update caused rustfmt to go crazy - better reviewed as 2 commits, 2nd is just "cargo fmt"


mhammond:
somehow a dupe of #7152

PR #6997, opened 2025-10-06
Ensure wrapped key is freed in `get_aes256_key`

jo:
This 

https://bugzilla.mozilla.org/show_bug.cgi?id=1966327#c7

  - Note:


components/support/rc_crypto/nss/src/pk11/sym_key.rs
```diff
@@ -217,6 +217,9 @@ fn get_aes256_key(name: &str) -> Result<SymKey> {
             }
             .map_err(|_| get_last_error())?;
 
+            // free wrapped key
+            unsafe { nss_sys::SECITEM_FreeItem(wrapped_key, nss_sys::PR_TRUE) }
```
bendk, line 226: This won't run if one of the previous lines errors out and causes an early return.  I think a RAII-style wrapper class would be the best way to free the item to avoid that possibility.

That said, if one of the previous lines errors out, we're probably going to fail to startup anyways so leaking some memory is not a big deal.  This is also an improvement over the previous code, so I'm fine merging this now and maybe fixing in the future.

The one thing I think is needed is a comment explaining this, but I don't need to review the wording there.

bendk (approved):
Approving this one, but see my inline comment for details.

components/support/rc_crypto/nss/src/pk11/sym_key.rs
bendk: Maybe `scoped_ptr!` is all you need for the RAII.

PR #7001, opened 2025-10-07
feat: add caching to ads-client

Almaju:
- Auto-caches successful responses in SQLite to skip redundant API calls for ads data
- Configurable settings: DB path, max size (10MB default), TTL (5min default), auto-cleanup of expired entries
- Honors Cache-Control headers: Respects no-store, max-age, etc.
- LRU eviction: Drops oldest entries when size limit hit
- Drop-in replacement: Just use send() instead of direct HTTP calls and caching happens transparently

  - Note:

### DIFF
--- a/components/ads-client/src/http_cache.rs
+++ b/components/ads-client/src/http_cache.rs
@@ -1,0 +1,121 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+mod builder;
+mod bytesize;
+mod cache_control;
+mod connection_initializer;
+mod request_hash;
+mod store;
+
+use self::{
+    builder::HttpCacheBuilder, cache_control::CacheControl, request_hash::RequestHash,
+    store::HttpCacheStore,
+};
+
+use viaduct::{Request, Response};
+
+pub use self::bytesize::ByteSize;
+use std::path::Path;
+use std::time::Duration;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Could not build cache: {0}")]
+    Builder(#[from] builder::Error),
+
+    #[error("SQLite operation failed: {0}")]
+    Sqlite(#[from] rusqlite::Error),
+
+    #[error("Error sending request: {0}")]
+    Viaduct(#[from] viaduct::ViaductError),
+}
+
+pub struct HttpCache {
+    max_size: ByteSize,
+    store: HttpCacheStore,
+    ttl: Duration,
+}
+
+impl HttpCache {
+    pub fn builder<P: AsRef<Path>>(db_path: P) -> HttpCacheBuilder {
+        HttpCacheBuilder::new(db_path.as_ref())
+    }
+
+    pub fn clear(&self) -> Result<(), Error> {
+        self.store.clear_all().map_err(Error::from)?;
+        Ok(())
+    }
+
+    pub fn send(&self, request: Request) -> Result<(Response, Option<RequestHash>), Error> {
+        let cutoff = chrono::Utc::now().timestamp() - self.ttl.as_secs() as i64;
+        self.store.delete_expired_entries(cutoff)?;
+
+        if let Some((response, request_hash)) = self.store.lookup(&request)? {
+            return Ok((response, Some(request_hash)));
+        }
+
+        let response = request.clone().send()?;
+        let cache_control = CacheControl::from(&response);
+        if cache_control.should_cache() {
+            self.store.store(&request, &response)?;
+            self.store.trim_to_max_size(self.max_size.as_u64() as i64)?;
+        }
+        Ok((response, None))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_http_cache_creation() {
+        // Test that HttpCache can be created successfully with test config
+        let cache = HttpCache::builder("test_cache.db").build();
+        assert!(cache.is_ok());
+
+        // Test with custom config
+        let cache_with_config = HttpCache::builder("custom_test.db")
+            .max_size(ByteSize::mib(1))
+            .ttl(Duration::from_secs(60))
+            .build();
+        assert!(cache_with_config.is_ok());
+    }
+
+    #[test]
+    fn test_clear_cache() {
+        let cache = HttpCache::builder("test_clear.db").build().unwrap();
+
+        // Create a test request and response
+        let request = viaduct::Request {
+            method: viaduct::Method::Get,
+            url: "https://example.com/test".parse().unwrap(),
+            headers: viaduct::Headers::new(),
+            body: None,
+        };
+
+        let response = viaduct::Response {
+            request_method: viaduct::Method::Get,
+            url: "https://example.com/test".parse().unwrap(),
+            status: 200,
+            headers: viaduct::Headers::new(),
+            body: b"test response".to_vec(),
+        };
+
+        // Store something in the cache
+        cache.store.store(&request, &response).unwrap();
+
+        // Verify it's cached
+        let retrieved = cache.store.lookup(&request).unwrap();
+        assert!(retrieved.is_some());
+
+        // Clear the cache
+        cache.clear().unwrap();
+
+        // Verify it's cleared
+        let retrieved_after_clear = cache.store.lookup(&request).unwrap();
+        assert!(retrieved_after_clear.is_none());
+    }
+}
--- a/components/ads-client/src/http_cache/builder.rs
+++ b/components/ads-client/src/http_cache/builder.rs
@@ -1,0 +1,232 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::http_cache::HttpCache;
+
+use super::bytesize::ByteSize;
+use super::connection_initializer::HttpCacheConnectionInitializer;
+use super::store::HttpCacheStore;
+use sql_support::open_database;
+use std::path::PathBuf;
+use std::time::Duration;
+
+const DEFAULT_MAX_SIZE: ByteSize = ByteSize::mib(10);
+const DEFAULT_TTL: Duration = Duration::from_secs(300);
+
+const MIN_CACHE_SIZE: ByteSize = ByteSize::kib(1);
+const MAX_CACHE_SIZE: ByteSize = ByteSize::mib(100);
+const MIN_TTL: Duration = Duration::from_secs(1);
+const MAX_TTL: Duration = Duration::from_secs(60 * 60 * 24 * 7); // 7 days
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Database path cannot be empty")]
+    EmptyDbPath,
+    #[error("Database error: {0}")]
+    Database(#[from] open_database::Error),
+    #[error(
+        "Maximum cache size must be between {min_size} and {max_size}, got {size_bytes} bytes"
+    )]
+    InvalidMaxSize {
+        max_size: String,
+        min_size: String,
+        size_bytes: u64,
+    },
+    #[error("TTL must be between {min_ttl} and {max_ttl}, got {ttl} seconds")]
+    InvalidTtl {
+        max_ttl: String,
+        min_ttl: String,
+        ttl: u64,
+    },
+}
+
+#[derive(Debug)]
+pub struct HttpCacheBuilder {
+    db_path: PathBuf,
+    max_size: Option<ByteSize>,
+    ttl: Option<Duration>,
+}
+
+impl HttpCacheBuilder {
+    pub fn new(db_path: impl Into<PathBuf>) -> Self {
+        Self {
+            db_path: db_path.into(),
+            max_size: None,
+            ttl: None,
+        }
+    }
+
+    pub fn max_size(mut self, max_size: ByteSize) -> Self {
+        self.max_size = Some(max_size);
+        self
+    }
+
+    pub fn ttl(mut self, ttl: Duration) -> Self {
+        self.ttl = Some(ttl);
+        self
+    }
+
+    fn validate(&self) -> Result<(), Error> {
+        if self.db_path.to_string_lossy().trim().is_empty() {
+            return Err(Error::EmptyDbPath);
+        }
+
+        if let Some(max_size) = self.max_size {
+            if max_size < MIN_CACHE_SIZE || max_size > MAX_CACHE_SIZE {
+                return Err(Error::InvalidMaxSize {
+                    size_bytes: max_size.as_u64(),
+                    min_size: MIN_CACHE_SIZE.to_string(),
+                    max_size: MAX_CACHE_SIZE.to_string(),
+                });
+            }
+        }
+
+        if let Some(ttl) = self.ttl {
+            if !(MIN_TTL..=MAX_TTL).contains(&ttl) {
+                return Err(Error::InvalidTtl {
+                    ttl: ttl.as_secs(),
+                    min_ttl: format!("{} seconds", MIN_TTL.as_secs()),
+                    max_ttl: format!("{} seconds", MAX_TTL.as_secs()),
+                });
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn build(self) -> Result<HttpCache, Error> {
+        self.validate()?;
+
+        let initializer = HttpCacheConnectionInitializer {};
+        let conn = if cfg!(test) {
+            open_database::open_memory_database(&initializer)?
+        } else {
+            open_database::open_database(&self.db_path, &initializer)?
+        };
+
+        let max_size = self.max_size.unwrap_or(DEFAULT_MAX_SIZE);
+        let store = HttpCacheStore::new(conn);
+        let ttl = self.ttl.unwrap_or(DEFAULT_TTL);
+
+        Ok(HttpCache {
+            max_size,
+            store,
+            ttl,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_cache_builder_with_defaults() {
+        let builder = HttpCacheBuilder::new("test.db".to_string());
+        assert_eq!(builder.db_path, PathBuf::from("test.db"));
+        assert_eq!(builder.max_size, None);
+        assert_eq!(builder.ttl, None);
+        assert!(builder.build().is_ok());
+    }
+
+    #[test]
+    fn test_cache_builder_valid_custom() {
+        let builder = HttpCacheBuilder::new("custom.db".to_string())
+            .max_size(ByteSize::b(1024))
+            .ttl(Duration::from_secs(60));
+
+        assert_eq!(builder.db_path, PathBuf::from("custom.db"));
+        assert_eq!(builder.max_size, Some(ByteSize::b(1024)));
+        assert_eq!(builder.ttl, Some(Duration::from_secs(60)));
+        assert!(builder.build().is_ok());
+    }
+
+    #[test]
+    fn test_validation_empty_db_path() {
+        let builder = HttpCacheBuilder::new("   ".to_string());
+
+        let result = builder.build();
+        assert!(matches!(result, Err(Error::EmptyDbPath)));
+    }
+
+    #[test]
+    fn test_validation_max_size_too_small() {
+        let builder = HttpCacheBuilder::new("test.db".to_string()).max_size(ByteSize::b(512));
+
+        let result = builder.build();
+        assert!(matches!(
+            result,
+            Err(Error::InvalidMaxSize {
+                size_bytes: 512,
+                min_size: _,
+                max_size: _,
+            })
+        ));
+    }
+
+    #[test]
+    fn test_validation_max_size_too_large() {
+        let builder = HttpCacheBuilder::new("test.db".to_string())
+            .max_size(ByteSize::b(2 * 1024 * 1024 * 1024));
+
+        let result = builder.build();
+        assert!(matches!(
+            result,
+            Err(Error::InvalidMaxSize {
+                size_bytes: 2147483648,
+                min_size: _,
+                max_size: _,
+            })
+        ));
+    }
+
+    #[test]
+    fn test_validation_max_size_boundaries() {
+        let builder_min = HttpCacheBuilder::new("test.db".to_string()).max_size(MIN_CACHE_SIZE);
+        assert!(builder_min.build().is_ok());
+
+        let builder_max = HttpCacheBuilder::new("test.db".to_string()).max_size(MAX_CACHE_SIZE);
+        assert!(builder_max.build().is_ok());
+    }
+
+    #[test]
+    fn test_validation_ttl_too_small() {
+        let builder = HttpCacheBuilder::new("test.db".to_string()).ttl(Duration::from_secs(0));
+
+        let result = builder.build();
+        assert!(matches!(
+            result,
+            Err(Error::InvalidTtl {
+                ttl: 0,
+                min_ttl: _,
+                max_ttl: _,
+            })
+        ));
+    }
+
+    #[test]
+    fn test_validation_ttl_too_large() {
+        let builder =
+            HttpCacheBuilder::new("test.db".to_string()).ttl(Duration::from_secs(8 * 24 * 60 * 60));
+
+        let result = builder.build();
+        assert!(matches!(
+            result,
+            Err(Error::InvalidTtl {
+                ttl: 691200,
+                min_ttl: _,
+                max_ttl: _,
+            })
+        ));
+    }
+
+    #[test]
+    fn test_validation_ttl_boundaries() {
+        let builder_min = HttpCacheBuilder::new("test.db".to_string()).ttl(MIN_TTL);
+        assert!(builder_min.build().is_ok());
+
+        let builder_max = HttpCacheBuilder::new("test.db".to_string()).ttl(MAX_TTL);
+        assert!(builder_max.build().is_ok());
+    }
+}
--- a/components/ads-client/src/http_cache/connection_initializer.rs
+++ b/components/ads-client/src/http_cache/connection_initializer.rs
@@ -1,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use rusqlite::Connection;
+use sql_support::open_database;
+use std::time::Duration;
+
+pub struct HttpCacheConnectionInitializer {}
+
+impl open_database::ConnectionInitializer for HttpCacheConnectionInitializer {
+    const NAME: &'static str = "http_cache";
+    const END_VERSION: u32 = 1;
+
+    fn prepare(&self, conn: &Connection, _db_empty: bool) -> open_database::Result<()> {
+        conn.execute_batch("PRAGMA journal_mode=wal;")?;
+        conn.busy_timeout(Duration::from_secs(5))?;
+        Ok(())
+    }
+
+    fn init(&self, tx: &rusqlite::Transaction<'_>) -> open_database::Result<()> {
+        const SCHEMA: &str = "
+            CREATE TABLE IF NOT EXISTS http_cache (
+                cached_at INTEGER NOT NULL,
+                request_hash TEXT NOT NULL,
+                response_body BLOB NOT NULL,
+                response_headers BLOB,
+                response_status INTEGER NOT NULL,
+                size INTEGER NOT NULL,
+                PRIMARY KEY (request_hash)
+            );
+            CREATE INDEX IF NOT EXISTS idx_http_cache_cached_at ON http_cache(cached_at);
+        ";
+        tx.execute_batch(SCHEMA)?;
+        Ok(())
+    }
+
+    fn upgrade_from(
+        &self,
+        conn: &rusqlite::Transaction<'_>,
+        version: u32,
+    ) -> open_database::Result<()> {
+        match version {
+            0 => {
+                // Version 0 means we need to create the initial schema
+                self.init(conn)
+            }
+            _ => Err(open_database::Error::IncompatibleVersion(version)),
+        }
+    }
+}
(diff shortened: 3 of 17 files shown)


luc-lisi:
@Almaju taking a look at this PR now! One thing I did notice though is that we might be adding net-new dependencies to A-S with the way this is built right now. `bytesize` doesn't look like it's used elsewhere in the repo and so we would need to make sure we follow the process outlined here: [dependency management guidelines](https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md) if we wanted to add it.

Personally, I'm not sure the crate gives us enough utility it is worth the extra external dependency, but let me know if you feel strongly! Otherwise we should just be able to represent the size as a normal `u64`.

And even better, we can just make our own converter if we really like the readability:
```rust
const fn mib(n: u64) -> u64 { n * 1024 * 1024 }
```

Also I think you might have accidentally bumped a bunch of versions in the top level cargo-lock. @mhammond might be able to speak more about the correct way to manage that top-level lock file but in general, my experience is that things tend to break when changes are made to that file. Specifically, it looks like the `Dependency Check` CI step is failing. I normally just err on the side of not bumping any versions and letting the A-S team handle that part for us as it can affect all the components

I'll continue to take a look at the rest of the PR though but just wanted to quickly point these out!

luc-lisi:
Nit: `check-formatting` looks upset. I think you should be able to fix it with a local run of:
```sh
cargo fmt --all
```
from our component's directory

Almaju:
@luc-lisi Thanks for the feedback! I fixed the dependencies and formatting issues ✅

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+use sql_support::open_database;
+use viaduct::{Request, Response, Result};
+
+#[derive(uniffi::Object)]
+pub struct HttpCache {
+    store: HttpCacheStore,
+}
+
+#[uniffi::export]
+impl HttpCache {
```
luc-lisi, line 41: Because of the naming collision issue in Swift, we may want to consider naming this something more explicit like `AdsClientHttpCache`. That way we don't risk someone exposing a generic `HttpCache` that collides with ours.

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub fn new(config: Option<HttpCacheConfig>) -> Self {
+        let config = config.unwrap_or_default();
+        let internal_config = HttpCacheConfigInner::from(config);
+        let initializer = HttpCacheConnectionInitializer {};
+        let conn = if cfg!(test) {
+            open_database::open_memory_database(&initializer)
+                .expect("failed to open memory cache db")
+        } else {
+            open_database::open_database(&internal_config.db_path, &initializer)
+                .expect("failed to open cache db")
```
luc-lisi: I think there might be a panic risk with the `.expect`. It might be worth making some new named errors for different failure modes of the `HttpCache` just so we can bubble them up meaningfully 🙂

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+#[derive(uniffi::Object)]
+pub struct HttpCache {
+    store: HttpCacheStore,
+}
+
+#[uniffi::export]
+impl HttpCache {
+    #[uniffi::constructor]
+    pub fn new(config: Option<HttpCacheConfig>) -> Self {
+        let config = config.unwrap_or_default();
```
luc-lisi: There might be a risk here that `unwrap_or_default` could mask an invalid config provided by the user. Definitely double check me on the behavior but I think if you provided the constructor an invalid config it would silently fallback to the default without letting the user know. So maybe some explicit error handling would be nice here?

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+        if response.is_success() {
+            self.store.store(&request, &response).ok();
+            self.store.trim_to_max_size().ok();
+        }
+        Ok(response)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
luc-lisi, line 70: In the spirit of the above comments, some tests around invalid configs and sad paths of the constructor could be nice!

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+        };
+        let store = HttpCacheStore::new(internal_config, conn);
+        Self { store }
+    }
+}
+
+impl HttpCache {
+    pub fn send(&self, request: Request) -> Result<Response> {
+        self.store.cleanup_expired().ok();
+        if let Some(resp) = self.store.lookup(&request).ok().flatten() {
```
luc-lisi: I think we risk masking an error here where we might not be able to tell if `store.lookup()` returned some `Err(e)` vs. `Ok(None)` which might make debugging tricky down the line.

Similar risks I believe with:

```rust
self.store.cleanup_expired().ok();
self.store.store(&request, &response).ok();
self.store.trim_to_max_size().ok();
```

If there was some kind of failure I believe we fail silently. So we may want to make sure we catch and raise these errors should they occur. Similar to above, I think adding some Error types in `error.rs` could help make this easy to bubble up!

components/ads-client/src/http_cache/store.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    pub fn store(&self, request: &Request, response: &Response) -> SqliteResult<()> {
+        let cached = HttpCacheRow::from_request_response(request, response);
+
+        if !cached.cache_control.should_cache() {
+            return Ok(());
+        }
+
+        let cache_control_str = serde_json::to_string(&cached.cache_control).ok();
```
luc-lisi: Similar to previous comments. We should try and avoid `.ok()` where possible and have explicit error handling if we can! If somehow serde fails to parse I think this will end up writing `NULL` (or whatever the binding is for `None` here).

components/ads-client/src/http_cache/store.rs
```diff
@@ -0,0 +1,272 @@
...
+            },
+        )
+        .optional()
+        .map(|result| result.flatten())
+    }
+
+    pub fn store(&self, request: &Request, response: &Response) -> SqliteResult<()> {
+        let cached = HttpCacheRow::from_request_response(request, response);
+
+        if !cached.cache_control.should_cache() {
```
luc-lisi: nit: we kind of have a "double negative" here. It might make it easier to read if we define:

```rust
pub fn should_not_cache(&self) -> bool {
        self.no_store
    }
```

As the default behavior here is to always cache unless explicitly told not to vs the other way around.

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+            return Ok(resp);
+        }
+        let response = request.clone().send()?;
+        if response.is_success() {
```
luc-lisi: This is a funny quirk, but I think viaduct's `is_success` includes _any_ 2xx response. We don't really run the risk here for this particular endpoint we are hitting, but because this code should be agnostic of the endpoints we are hitting, it might be worth it to explicitly check for a `200` instead.

Another thought here (not sure if you agree!) is that we could move the logic for parsing the headers here as well just so that the store/no-store decision is all made in one place. Curious your thoughts though if you feel that is cleaner or not 🙂

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl From<HttpCacheConfig> for HttpCacheConfigInner {
+    fn from(config: HttpCacheConfig) -> Self {
+        Self {
+            db_path: config
+                .db_path
+                .unwrap_or_else(|| DEFAULT_DB_PATH.to_string()),
+            max_size: config
+                .max_size_bytes
+                .map(ByteSize::b)
```
luc-lisi: This is a goofy guard rail but maybe worth considering: right now someone could in theory mis-configure the cache to have 0 `maz_size_bytes` or `ttl_seconds`

I don't think we want to allow this behavior? So maybe some kind of minimum value here might be sensible with some kind of misconfiguration error thrown if the cache size or TTL is too small.

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ByteSize(u64);
+
+impl ByteSize {
+    pub const fn b(value: u64) -> Self {
+        Self(value)
+    }
+
+    pub const fn mib(value: u64) -> Self {
+        Self(value * 1024 * 1024)
```
luc-lisi: Oh shoot, I know I suggested this code but seeing it written out made me realize we have an overflow risk here. We probably want:

```rust
pub const fn mib(value: u64) -> Self {
    Self(value.saturating_mul(1024).saturating_mul(1024))
}
```

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,80 @@
...
+            ttl: config
+                .ttl_seconds
+                .map(Duration::from_secs)
+                .unwrap_or(DEFAULT_TTL),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
luc-lisi: In the spirit of the above comments, some tests around bad configs probably would be good!

components/ads-client/src/http_cache/connection_initializer.rs
```diff
@@ -0,0 +1,44 @@
...
+                etag TEXT,
+                request_hash TEXT NOT NULL,
+                response_body BLOB NOT NULL,
+                response_headers BLOB,
+                response_status INTEGER NOT NULL,
+                size INTEGER NOT NULL,
+                PRIMARY KEY (request_hash)
+            );
+            CREATE INDEX IF NOT EXISTS idx_http_cache_cached_at ON http_cache(cached_at);
+            CREATE INDEX IF NOT EXISTS idx_http_cache_size ON http_cache(size);
```
luc-lisi: I might be totally missing something but is there a reason we need this index? Eviction looks like it is using `cached_at` and I don't see any queries by size, but let me know if I totally glossed over something!

components/ads-client/src/http_cache/row.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    pub fn hash_request(request: &Request) -> String {
+        use std::collections::hash_map::DefaultHasher;
+        let mut hasher = DefaultHasher::new();
+        request.method.hash(&mut hasher);
+        request.url.hash(&mut hasher);
+
+        for header in request.headers.clone().into_iter() {
```
luc-lisi: I'm not sure your take here, but I think this _could_ cause an issue as HTTP headers don't have a deterministic order. `HeaderMap` _does_ seem to have a deterministic order but it is ordered based on insertion, so I'm not sure there is a guarantee that two identical requests would have the same same order of headers when we iterate through them.

So something like:

```rust
let mut headers: Vec<_> = request.headers.clone().into_iter().collect();
headers.sort_by_key(|header| header.name.to_ascii_lowercase());
for header in headers {
        header.name.hash(&mut hasher);
        header.value.hash(&mut hasher);
    }
```

I think solves this potential issue

components/ads-client/src/http_cache/row.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub request_hash: String,
+    pub response_body: Vec<u8>,
+    pub response_headers: Vec<u8>,
+    pub response_status: i64,
+    pub size: i64,
+}
+
+impl HttpCacheRow {
+    pub fn from_request_response(request: &Request, response: &Response) -> Self {
+        let headers_map: HashMap<String, String> = response.headers.clone().into();
```
luc-lisi: I think we could have an issue here with duplicate headers collapsing arbitrarily. I don't think it is common... but it looks like it is technically permissible to have something like:

```
HTTP/1.1 200 OK
Cache-Control: max-age=60
Cache-Control: must-revalidate
...
```

In which case the last `Cache-Control` would be the only one written.

So we might want something like:

```rust
let headers_vec: Vec<(String, String)> = response.headers.clone().into_iter()
    .map(|h| (h.name, h.value))
    .collect();
```

luc-lisi (changes requested):
This is awesome work!! I left a few comments for some minor things I noticed but by no means anything major. Really appreciate all this!! 😄 

I will also wait from additional eyes from one of the A-S maintainers before approving just because I'm definitely not as much of a Rust expert as they are. 

Awesome job!!

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,80 @@
+use std::time::Duration;
```
mhammond: all files will need a license header please - mozilla-central insists and we are trying to move there.

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,80 @@
...
+pub struct HttpCacheConfigInner {
+    pub db_path: String,
+    pub max_size: ByteSize,
+    pub ttl: Duration,
+}
+
+impl From<HttpCacheConfig> for HttpCacheConfigInner {
+    fn from(config: HttpCacheConfig) -> Self {
+        Self {
+            db_path: config
```
mhammond: in practice all clients will need to pass a path, so maybe a default doesn't make sense?

components/ads-client/src/http_cache.rs
Almaju: Not exactly, this would fallback to default config if the user passes `None` so that allows the user to create an `HttpCache` directly in Swift without passing any config at all. 
But I could not get my head around this, maybe we don't want to have default values for the cache and force the user to provide values? 🤔
This is the signature of Swift's URLCache:
```swift
let memoryCapacity = 4 * 1024 * 1024 // 4 MB
let diskCapacity = 100 * 1024 * 1024 // 100 MB
let urlCache = URLCache(memoryCapacity: memoryCapacity, diskCapacity: diskCapacity, diskPath: "myCache")
URLCache.shared = urlCache // Set as the shared cache
```

components/ads-client/src/http_cache.rs
Almaju: I was anticipating that maybe HttpCache would have it's own crate at some point because it does not rely on "ads-client" at all but rather on viaduct's Request/Response.

components/ads-client/src/http_cache.rs
Almaju: This is kind of intended because we have a fallback here in case there is a problem with the cache (just doing the normal request) so I don't think we should send an error in that case. Or if it sends an error, I guess it should send it with the data like `Result<(Data, Option<Err>), Err>`?

components/ads-client/src/http_cache.rs
Almaju: I am not sure why "200" would be better than "2xx", do you mean that for example it would not make sense to store a "204" since there is no content?
Agreed for the parsing header logic, I will move it!

components/ads-client/src/http_cache/connection_initializer.rs
Almaju: Good catch! I really thought we would need it but it looks like we do not indeed 😮

components/ads-client/src/http_cache/row.rs
Almaju: Oh good point! 👏

components/ads-client/src/http_cache.rs
luc-lisi: Ah yeah, what I mean is we should only cache responses with an explicit `200` rather than _all_ 2xx responses.  MARS doesn't really have other codes it can respond with using the endpoint we care about, but in the spirit of it being generalized, we probably should keep it from caching something like a `204` or even a `206`.

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+            open_database::open_database(&internal_config.db_path, &initializer)
+                .expect("failed to open cache db")
+        };
+        let store = HttpCacheStore::new(internal_config, conn);
+        Self { store }
+    }
+}
+
+impl HttpCache {
+    pub fn send(&self, request: Request) -> Result<Response> {
```
luc-lisi: Let's modify this to return a tuple with the Response and Hash for later invalidation logic.

components/ads-client/src/http_cache/row.rs
Almaju: So I digged up a bit about that one and it looks like this is not handled by viaduct itself. Viaduct uses protobuf and headers are typed as map<string, string> which are converted to HashMap<String, String>.
I made a quick test because I wanted to be really sure about that: https://gist.github.com/Almaju/5519594d8460e5c63fa85d77eba7c457
And the output is:
```
=== Viaduct Duplicate Headers Demo ===

Initializing network backend...
✓ Network backend initialized successfully

=== REAL HTTP REQUEST TEST ===

1. Making real HTTP request to: https://httpbin.org/response-headers?Cache-Control=max-age%3D60&Cache-Control=must-revalidate
   This server will return duplicate Cache-Control headers

2. Response received:
   Status: 200
   URL: https://httpbin.org/response-headers?Cache-Control=max-age%3D60&Cache-Control=must-revalidate

3. Raw headers from viaduct:
   date: Wed, 08 Oct 2025 20:46:45 GMT
   content-type: application/json
   content-length: 137
   connection: keep-alive
   server: gunicorn/19.9.0
   cache-control: max-age=60
   cache-control: must-revalidate
   access-control-allow-origin: *
   access-control-allow-credentials: true

4. Cache-Control header analysis:
   Found Cache-Control: 'max-age=60'
   ✗ Only 'max-age=60' found - 'must-revalidate' was lost!

5. What should happen according to RFC 7234:
   Multiple Cache-Control headers should be combined with commas:
   Cache-Control: max-age=60, must-revalidate

   Current result: 'max-age=60'

==================================================
```

So not much we can do about it, sounds like this should be supported at viaduct level. 🤔

components/ads-client/src/http_cache/config.rs
Almaju: Done https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache/connection_initializer.rs
Almaju: Done https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache/row.rs
Almaju: Done https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache/config.rs
Almaju: Done https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache/config.rs
Almaju: You are right! Done here https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache.rs
Almaju: Refactored: config is no longer optional because we need the db_path anyway. https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache.rs
Almaju: Done https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache.rs
Almaju: I tried with a callback approach which feels cleaner than a weird return type but I did not test it yet, I am open to feedbacks: https://github.com/mozilla/application-services/pull/7001/commits/39e71dcda83420d0f48178f2a37cb1caa1149fd9

components/ads-client/src/http_cache.rs
mhammond: Does it need to be exposed at all? ISTM that clients aren't going to interact with it directly and really only need a way to configure it (and even then, in practice seem likely to use hard-coded values)

(That would still leave the same question for the config object though, but that seems like much lower stakes)

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,253 @@
...
+
+pub use config::{HttpCacheConfig, HttpCacheConfigError};
+use sql_support::open_database;
+use std::sync::Arc;
+use viaduct::{Request, Response, Result};
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum HttpCacheError {
+    #[error("Could not open SQLite database: {0}")]
+    OpenDatabase(String),
```
mhammond: `OpenDatabase` seems the wrong name as it's regularly used for any sqlite error? Maybe it should be a generic `DatabaseError` with a `#[from]` a sqlite error? Or maybe a new variant if `OpenDatabase` has specific value.

components/ads-client/src/http_cache/row.rs
mhammond: viaduct is moving away from the protobuf backend, but I suspect the new backend has the same issue and it should indeed be fixed there.

components/ads-client/src/http_cache/connection_initializer.rs
```diff
@@ -0,0 +1,47 @@
...
+        tx.execute_batch(SCHEMA)?;
+        Ok(())
+    }
+
+    fn upgrade_from(
+        &self,
+        _conn: &rusqlite::Transaction<'_>,
+        _version: u32,
+    ) -> open_database::Result<()> {
+        Ok(())
```
mhammond: I think we should do something here for unexpected version numbers - a panic would probably be fine. That would probably mean you could create the schema for version 0?

components/ads-client/src/http_cache/config.rs
```diff
@@ -0,0 +1,251 @@
...
+use super::bytesize::ByteSize;
+use std::time::Duration;
+
+const DEFAULT_MAX_SIZE: ByteSize = ByteSize::mib(10);
+const DEFAULT_TTL: Duration = Duration::from_secs(300);
+
+const MIN_CACHE_SIZE_BYTES: ByteSize = ByteSize::kib(1);
+const MAX_CACHE_SIZE_BYTES: ByteSize = ByteSize::mib(100);
+const MIN_TTL_SECONDS: u64 = 1;
+const MAX_TTL_SECONDS: u64 = 24 * 60 * 60;
```
mhammond: this seems short?

luc-lisi: In general for ads, we actually have an incredibly short TTL requirement before we can technically no longer use the ad anymore. Although I agree that for the sake of a "max" value, we can maybe make this a week as there could in theory be ads that are allowed to live roughly on that order. But even a day in this case would be quite long for an ad.

mhammond: Right - but that would probably come from the body of the response somehow rather than from the cache ttl? If the server says the response can be cached for a week but we evict it anyway after a day, just to hit the server again the next day, just to discover the response body tells us the ad itself can no longer be used, things would seem a little strange. I must admit I've no idea if that actually makes sense or not in your scenarios :)

components/ads-client/src/http_cache.rs
Almaju: The cache is not exposed anymore via UniFFI 
https://github.com/mozilla/application-services/pull/7001/commits/51a9b7ade61795b4baeebacc64b1f738f8b94c22

components/ads-client/src/http_cache.rs
Almaju: I added a bunch of tests for inside `builder.rs`

components/ads-client/src/http_cache.rs
Almaju: I ended up doing some simpler logic with `Result<Result<...>>` and let the client do what it needs to do

components/ads-client/src/http_cache/config.rs
Almaju: A lot of tests in `builder.rs`, feel free to tell me if you spot a missing scenario!

components/ads-client/src/http_cache/row.rs
Almaju: @luc-lisi I think this cannot be supported then and should probably be handled by the new backend, is it okay if I close this issue?

components/ads-client/src/http_cache.rs
Almaju: Done! https://github.com/mozilla/application-services/pull/7001/commits/51a9b7ade61795b4baeebacc64b1f738f8b94c22

components/ads-client/src/http_cache.rs
Almaju: Done! I was wondering, where do you stand on using `#[error(transparent)]` instead of a custom message?

components/ads-client/src/http_cache/connection_initializer.rs
Almaju: Done!
```rs
    fn upgrade_from(
        &self,
        conn: &rusqlite::Transaction<'_>,
        version: u32,
    ) -> open_database::Result<()> {
        match version {
            0 => {
                // Version 0 means we need to create the initial schema
                self.init(conn)
            }
            _ => Err(open_database::Error::IncompatibleVersion(version)),
        }
    }
```

components/ads-client/src/http_cache/config.rs
Almaju: Do we need a max TTL at all? 🤔 
Or maybe we just put max 24h then?

components/ads-client/src/http_cache/row.rs
luc-lisi: Oh okay good to know this is just a deeper issue. Yeah fair to close this then. Is there an active issue open to fix this in the new backend? If so it would be good to link to it!

components/ads-client/src/http_cache/config.rs
luc-lisi: @mhammond Ah okay I believe I see your point here. There might be some confusion on the implementation and how we are wording things in the config. (Or it's totally possible I'm misunderstanding something as well)

In our scenario the way things will work is that the server will most often return no cache-control headers. The server takes a strong stance that ads should really _never_ be be cached, but for performance reasons that isn't always the case. So, we have some guidelines on how long they could possibly ever live before being invalid. But because we don't often send cache-control info in the response, the cache will be using whatever the surface sets this to be assuming they override the default. This `MAX_TTL_SECONDS` is (if I'm reading the implementation correctly) not a strict limit on the cache, but just a limit on what the surface can set the default TTL for objects without a cache-control header in a response. This is not a limit though on what the server could set a TTL to given proper headers.

Maybe we could rename this to `MAX_CONFIGURABLE_DEFAULT_TTL_SECONDS` or something to be really explicit about what we mean here?

components/ads-client/src/http_cache.rs
```diff
@@ -0,0 +1,121 @@
...
+    Sqlite(#[from] rusqlite::Error),
+
+    #[error("Error sending request: {0}")]
+    Viaduct(#[from] viaduct::ViaductError),
+}
+
+pub struct HttpCache {
+    max_size: ByteSize,
+    store: HttpCacheStore,
+    ttl: Duration,
```
luc-lisi, line 38: In writing the response to Mark's comment I realize this is maybe confusing. Could we make this `default_ttl` rather than just `ttl`?

If I'm understanding correctly, any cache-control headers that define a different TTL will always override this value. So this is really not the `ttl` in general but more of the  "default" duration used in the absence of any other information in the response I think. Let me know if my understanding of the implementation is incorrect though.

components/ads-client/src/http_cache/builder.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    pub fn max_size(self, max_size: ByteSize) -> Self {
+        Self {
+            max_size: Some(max_size),
+            ..self
+        }
+    }
+
+    pub fn ttl(self, ttl: Duration) -> Self {
```
luc-lisi: Same comment as above. Maybe we define this as `default_ttl` to make things explicit.

luc-lisi (commented):
This looks great! I left one comment about renaming the TTL but otherwise I think this is fantastic and good to start testing with!

components/ads-client/src/http_cache/config.rs
mhammond: I see, that makes sense, thanks.

Almaju: Not sure I understand what's the issue here. Currently we only look at the "no-store" value from the cache control header. We could potentially consider "max-age=0" as a no-store by the way. If in the future we want to really support max-age, for me the best way would be to have a strategy to say if we want to use the cache TTL as the upper limit or if we want the trust the max-age. Something like `enum TtlStrategy { Bounded, CacheControlled }`. I don't see why there would be a confusion over the name "ttl" for the cache itself though or how renaming it `default_ttl` makes it clearer but once again maybe I am not completely understanding the issue there.

luc-lisi (approved):

components/ads-client/src/http_cache/config.rs
Almaju: After talking with @luc-lisi , I think I understand the issue better now! `MAX_TTL` is a property of the cache builder (which caps the maximum TTL you can set when building a cache), rather than a property of the cache itself. In that context, I think MAX_TTL works well and is pretty clear.
I'm not sure a CONFIGURABLE_ prefix adds much value here since it's already in a builder pattern. And DEFAULT_ might be confusing since we already have the concept of a default TTL (when you pass None).
I'll go ahead and merge it like this, but feel free to slack me or reply here if you have a better idea and I'll do a quick refactor PR!

mhammond: I'm really not too bothered, but if we were to consider this a "general" http cache I would have thought both "no-store" and "max-age=0" would mean it should not be stored - ie, that a "default ttl" probably doesn't make sense in that case unless you are willing to have it return stale content, which a general purpose cache probably isn't wanting to do.

I really only thought if this as being a potential "general" cache because of an earlier throw-away comment - but this PR isn't making this a general cache but is instead specific to your ads requirement - so I'm perfectly fine with whatever makes sense here, and we could have a discussion about what a "general" cache should do if/when someone proposes using it in that way.

luc-lisi: Yeah @mhammond this particular module is just for the ads-client specific cache. I would love to chat though about a general cache that could be built off a modified version of this though 😄

Almaju: Yes I think we could potentially turn it into a general cache with some tweaks, that was my intention at least when writing the code to make it as much agnostic as I could! But this would indeed mean we need a better support of the cache-control header. Would love to chat about that and move forward with a generic cache though!

PR #7002, opened 2025-10-07
Add convenience `get_most_recent` method to fetch newest history metadata with a limit

issammani:

This patch adds a new convenience method `get_most_recent` to retrieve the newest history metadata entries ordered by updated_at DESC with a configurable limit. Without this method, we have to use `get_since(start: min_i64)` and pop first n elements that we need from the vec which can be wasteful at times. Using `get_most_recent` should give back exactly the number of items we need ordered correctly.

  - Note:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,15 +78,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anstyle"
 version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -186,17 +177,6 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.89",
-]
-
-[[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


issammani:
cc @mhammond @bendk

components/places/src/storage/history_metadata.rs
```diff
@@ -1297,6 +1315,102 @@ mod tests {
...
+        assert!(rows.is_empty());
+    }
+
+    #[test]
+    fn test_get_most_recent_orders_and_limits() {
+        let conn = PlacesDb::open_in_memory(ConnectionType::ReadWrite).expect("memory db");
+
+        note_observation!(&conn,
+            url "https://example.com/1",
+            view_time Some(10),
```
mhammond, line 1392: we should probably have this identical for all visits to make it clear it doesn't matter. Related, this might be racey in that it probably assumes incrementing timestamps between observations, which might not be true, so the order back is different. I'm not immediately sure what to do about that :( Maybe ignore it if other tests rely on it, sqlite doing a full transaction before the timestamp advances seems unlikely to be a problem for a while?

mhammond (commented):
This looks great! I think you will need an addition to the UDL too? And I'm wondering if we should we maybe just replace `get_since()` - it sounds like you are saying it's not useful so it seems likely Android might want the same thing? We wont block on Android changing, but it might be nice to leave a breadcrumb here for future us.

(It's unfortunate this "history-metadata" stuff still exists - it's basically "normal" history with some other "view time" stuff that's not synced - but it is what is it is :)

issammani:
> I think you will need an addition to the UDL too?

True. I got so used to the macros now hahah. Added in https://github.com/mozilla/application-services/pull/7002/commits/50eee6bb8cadaf85f4fdd44496410adb35c11e6d

> it sounds like you are saying it's not useful so it seems likely Android might want the same thing? We wont block on Android changing, but it might be nice to leave a breadcrumb here for future us.

Yeah that's the plan. I will also change Android after this lands and we can remove it after. For some reason, I always end up breaking android when I do breaking changes 🙈 So this seems safer. Added ticket and comment on top of method to track this in  https://github.com/mozilla/application-services/pull/7002/commits/d7eaab98d8dbee7ae1b4f28775a41655f1349e06 😄

components/places/src/storage/history_metadata.rs
issammani: Good callout. I added both with same observation and different ones in https://github.com/mozilla/application-services/pull/7002/commits/77cca7fd5e273ed2c1049ad641107cc40e38da2e. Let me know if you think we need to test more variations.

issammani: > Related, this might be racey in that it probably assumes incrementing timestamps between observations, 

Yeah I can see how that can be a problem. I don't see anything in the other tests that attempts to solve it other than maybe some `thread::sleep` that might have been added for this ?

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## ✨ What's New ✨
 
 ### Swift
-- Added `@unchecked Sendable` to classes that conform to `FeatureManifestInterface`. ([#6963](https://github.com/mozilla/application-services/pull/6963)
+- Added `@unchecked Sendable` to classes that conform to `FeatureManifestInterface`. ([#6963](https://github.com/mozilla/application-services/pull/6963))
```
issammani, line 6: Was missing a `)` at the end and it bothered me 😂

components/places/src/storage/history_metadata.rs
mhammond: yeah, maybe a comment and a super-short sleep between reecords will do for now

mhammond (dismissed):
this is great, thanks!

components/places/src/storage/history_metadata.rs
issammani: Sweet ! Done in https://github.com/mozilla/application-services/pull/7002/commits/f8acb29202c6e93b7ca5f15f9cb9ac4ec268e228. I added a tiny helper with a comment. @mhammond if it looks good can you please merge after CI is green. I don't have merge access. Thanks !

PR #7281, opened 2026-03-19
Add breadcrumb when rate limiting fails to read the clock

bendk:
This rate limit check doesn't seem to be working in same cases, for example for https://bugzilla.mozilla.org/show_bug.cgi?id=2004954 I saw many reports per minute from the same user.  I'm wondering if the cause is this check failing.

  - Note:

### DIFF
--- a/components/support/error/src/error_tracing.rs
+++ b/components/support/error/src/error_tracing.rs
@@ -9,6 +9,8 @@ use std::{
 
 use parking_lot::Mutex;
 
+use crate::breadcrumb;
+
 static GLOBALS: Mutex<Globals> = Mutex::new(Globals::new());
 
 pub fn report_error_to_app(type_name: String, message: String) {
@@ -139,7 +141,9 @@ impl RateLimiter {
                 // with NTP, if users manually adjust their clocks, etc.  Letting an extra event
                 // through seems okay in this case. We should get back into a good state soon
                 // after.
-                _ => (),
+                _ => {
+                    breadcrumb!("error_support: checked_duration_since failed");
+                }
             }
         }
         last_report.insert(component.to_string(), now);


components/support/error/src/error_tracing.rs
```diff
@@ -139,7 +141,9 @@ impl RateLimiter {
                 // with NTP, if users manually adjust their clocks, etc.  Letting an extra event
                 // through seems okay in this case. We should get back into a good state soon
                 // after.
-                _ => (),
+                _ => {
```
mhammond, line 144: I'm a little skeptical of this - I can see how extra single events might get through as the clock changes, but not how that could cause many reports per minute. What's your theory about how that actually happens in practice?

bendk: It does seem really weird.

Here's what I'm looking at in [grafana](https://yardstick.mozilla.org/d/sync-errors/sync-error-list?orgId=1&from=2026-01-01T10:04:10.931Z&to=2026-01-03T15:34:16.797Z&timezone=browser&var-application=firefox_android&var-channel=nightly&var-error_type=places-bookmark-corruption&var-version=&var-build_date=&var-architecture=&var-device_manufacturer=&var-device_model=XQ-AT52&var-locale=zh-CN&var-os=&var-os_version=&refresh=1h).  One client is generating multiple error pings per second (I need to add a client filter, but for now I just added enough filters other that I'm pretty sure I'm only capturing one person).

I don't really understand what's going on.  This is one wild theory, maybe there's something very wrong at the system level.  My other theory was that the FF was restarting and clearing out the rate limiting data, but it seems impossible for FF to restart that much in practice.  My only other theory was that it was part of some automation.  This was really just a shot in the dark.  Do you have any ideas on what could be happening?

(BTW, the unit in that graph is "errors / day" which is not right for unique clients.  I think it's just one unique client, but since the interval is 2 hours it's multiplying to get 12 unique clients / day.  I'm going to try to fix that today.)

bendk:
I let this one drop and I can't see anything weird anymore on the graphs.  Let's just close this one.

PR #7306, opened 2026-04-07
Add the ability to start a new OAuth flow while authenticated, so you can authorize additional scopes [ci full]

mhammond:
As @jonalmeida and I were chatting about. @bendk does this look correct to you? @LZoog and @skhamis for some context for our slack messages.

This supports desktop-like flows, where a user is already signed in to (say) Sync, but not wants to authorize "vpn" or "smartwindow".

I used the term "Authorization" to try and reflect we aren't expected to "authenticate" here - we are already authenticated - we typically expect just a confirmation button. I'm not entirely convinced the distinction between "authorize" and "authenticate" is enough, so the names might be confusing - I started with something like "UpgradeScopes" etc. Regardless, better names welcome here.

I've tried to keep the names consistent, which made them verbose. Better names welcome here too.

  - Note:

### DIFF
--- a/components/fxa-client/src/fxa_client.udl
+++ b/components/fxa-client/src/fxa_client.udl
@@ -195,12 +195,10 @@ interface FirefoxAccount {
   ///       - This parameter is used for metrics purposes, to identify the
   ///         UX entrypoint from which the user triggered the signin request.
   ///         For example, the application toolbar, on the onboarding flow.
-  ///   - `metrics` - optionally, additional metrics tracking parameters.
-  ///       - These will be included as query parameters in the resulting URL.
-  ///
+  ///   - `service` - The service being signed up for.
+  //        - The combination of service and entrypoint may impact the UI shown.  
   [Throws=FxaError]
-  string begin_oauth_flow([ByRef] sequence<string> scopes, [ByRef] string entrypoint);
-  
+  string begin_oauth_flow([ByRef] sequence<string> scopes, [ByRef] string entrypoint, [ByRef] optional string service = "");
 
   /// Get the URL at which to begin a device-pairing signin flow.
   ///
@@ -234,11 +232,10 @@ interface FirefoxAccount {
   ///       - This parameter is used for metrics purposes, to identify the
   ///         UX entrypoint from which the user triggered the signin request.
   ///         For example, the application toolbar, on the onboarding flow.
-  ///   - `metrics` - optionally, additional metrics tracking parameters.
-  ///       - These will be included as query parameters in the resulting URL.
-  ///
+  ///   - `service` - The service being signed up for.
+  //        - The combination of service and entrypoint may impact the UI shown.  
   [Throws=FxaError]
-  string begin_pairing_flow([ByRef] string pairing_url, [ByRef] sequence<string> scopes, [ByRef] string entrypoint);
+  string begin_pairing_flow([ByRef] string pairing_url, [ByRef] sequence<string> scopes, [ByRef] string entrypoint, [ByRef] optional string service = "");
   
 
   /// Complete an OAuth flow.
@@ -258,7 +255,6 @@ interface FirefoxAccount {
   [Throws=FxaError]
   void complete_oauth_flow([ByRef] string code, [ByRef] string state );
   
-
   /// Check authorization status for this application.
   ///
   /// **💾 This method alters the persisted account state.**
@@ -1003,7 +999,7 @@ dictionary Profile {
 interface FxaState {
   Uninitialized();
   Disconnected();
-  Authenticating(string oauth_url);
+  Authenticating(string oauth_url, FxaRustAuthState initial_state);
   Connected();
   AuthIssues();
 };
@@ -1011,8 +1007,8 @@ interface FxaState {
 [Enum]
 interface FxaEvent {
   Initialize(DeviceConfig device_config);
-  BeginOAuthFlow(sequence<string> scopes, string entrypoint);
-  BeginPairingFlow(string pairing_url, sequence<string> scopes, string entrypoint);
+  BeginOAuthFlow(string service, sequence<string> scopes, string entrypoint);
+  BeginPairingFlow(string pairing_url, string service, sequence<string> scopes, string entrypoint);
   CompleteOAuthFlow(string code, string state);
   CancelOAuthFlow();
   CheckAuthorizationStatus();
--- a/components/fxa-client/src/internal/oauth.rs
+++ b/components/fxa-client/src/internal/oauth.rs
@@ -12,7 +12,9 @@ use super::{
     util, FirefoxAccount,
 };
 use crate::auth::UserData;
-use crate::{error, warn, AuthorizationParameters, Error, FxaServer, Result, ScopedKey};
+use crate::{
+    debug, error, info, warn, AuthorizationParameters, Error, FxaServer, Result, ScopedKey,
+};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use jwcrypto::{EncryptionAlgorithm, EncryptionParameters};
 use rate_limiter::RateLimiter;
@@ -20,7 +22,6 @@ use rc_crypto::digest;
 use serde_derive::*;
 use std::{
     collections::{HashMap, HashSet},
-    iter::FromIterator,
     time::{SystemTime, UNIX_EPOCH},
 };
 use url::Url;
@@ -165,11 +166,15 @@ impl FirefoxAccount {
     pub fn begin_pairing_flow(
         &mut self,
         pairing_url: &str,
+        service: &str,
         scopes: &[&str],
         entrypoint: &str,
     ) -> Result<String> {
         let mut url = self.state.config().pair_supp_url()?;
         url.query_pairs_mut().append_pair("entrypoint", entrypoint);
+        if !service.is_empty() {
+            url.query_pairs_mut().append_pair("service", service);
+        }
         let pairing_url = util::parse_url(pairing_url, "begin_pairing_flow")?;
         if url.host_str() != pairing_url.host_str() {
             let fxa_server = FxaServer::from(&url);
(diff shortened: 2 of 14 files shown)


components/fxa-client/src/state_machine/internal_machines/connected.rs
```diff
@@ -18,6 +18,9 @@ impl InternalStateMachine for ConnectedStateMachine {
             FxaEvent::Disconnect => Ok(Disconnect),
             FxaEvent::CheckAuthorizationStatus => Ok(CheckAuthorizationStatus),
             FxaEvent::CallGetProfile => Ok(GetProfile),
+            FxaEvent::BeginOAuthScopeAuthorizationFlow { scopes, entrypoint } => {
+                Ok(BeginOAuthFlow { scopes, entrypoint })
```
bendk: I think you'll need a new internal state for this since it should call `begin_oauth_scope_authorization_flow()` instead of `begin_oauth_flow()`.  Ditto for the complete flow event.

`state_machine/internal_machines/mod.rs` should get some extra cases to map the new states to the right calls.

bendk (commented):
This looks pretty much correct to me, except I think there's one more step to do.

My main takeaway is that these state machines feel overly complicated.  It seems like you should be able to add an extra event, add code in a corresponding switch case, and that's it.  When we were doing the Android migration and using the state machine tester to verify the code, we were getting some value in exchange.  Now, more and more I'm feeling like we could rework this and simplify it.

mhammond:
https://phabricator.services.mozilla.com/D293447 is the fenix side of this. I *think* I got the semantics correct for the new flow, but it's hard to know if it doesn't work for that reason or for an fxa reason - I know Lauren mentioned some work needs to be done on the fxa side for this to work. This PR however doesn't seem like a breaking change and should be safe even if it need tweaks - wdyt?

mhammond:
> This looks pretty much correct to me, except I think there's one more step to do.

I agree, but not sure it should happen here? wdyt? Otherwise I think this seems ready to roll!

components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt
```diff
@@ -160,6 +160,30 @@ class FxaClient(inner: FirefoxAccount, persistCallback: PersistCallback?) : Auto
...
+    /**
+     * Completes an OAuth scope authorization flow initiated by [beginOAuthScopeAuthorizationFlow].
+     *
+     * This performs network requests, and should not be used on the main thread.
+     */
+    fun completeOAuthScopeAuthorizationFlow(code: String, state: String) {
+        this.inner.completeOauthScopeAuthorizationFlow(code, state)
+        this.tryPersistState()
+    }
+
```
bendk, line 186: Should we add these or is it better to force Android to use the state machine?  I don't have strong feelings either way, just wanted to ask the question.

components/fxa-client/src/state_machine/internal_machines/authorizing.rs
```diff
@@ -0,0 +1,136 @@
...
+            e => Err(Error::InvalidStateTransition(format!("Authorizing -> {e}"))),
+        }
+    }
+
+    fn next_state(&self, state: State, event: Event) -> Result<State> {
+        Ok(match (state, event) {
+            // On success, go directly to Connected — no InitializeDevice step needed since
+            // the device is already initialized.
+            (CompleteOAuthScopeAuthorizationFlow { .. }, CompleteOAuthFlowSuccess) => {
+                Complete(FxaState::Connected)
```
bendk, line 36: This is for another PR, but I think we should include the list of scopes as associated data for the `Connect` state.  It's kind of weird that you went through a "state change" and ended up in the exact same state as before.

bendk (approved):
Agreed, this looks ready to roll to me.  I had a couple questions/suggestions but I'm happy to merge as-is.

mhammond:
I decided to stay with the same entry-points, and by adding a little extra data to a couple of the states I also avoided needing any more states. It also seems to correctly implement the merging of scopes and managing the device record. @bendk do you mind having another look?

mhammond:
also has the quite cool ability to test most of this from the commandline - first log in with vpn (ie, not sync), then request access tokens for vpn and sync
```
cargo run --example fxa-client disconnect
...
cargo run --example fxa-client login --scope profile --scope https://identity.mozilla.com/apps/vpn --scope https://identity.mozilla.com/tokens/session
...
> Account is logged in and authorized by the server
cargo run --example fxa-client get-access-token https://identity.mozilla.com/apps/vpn
...
> Requesting access token with scope: https://identity.mozilla.com/apps/vpn
> Success: AccessTokenInfo { scope: "https://identity.mozilla.com/apps/vpn", token: "...", key: None, expires_at: 1776918561 }
cargo run --example fxa-client get-access-token https://identity.mozilla.com/apps/oldsync
...
> Requesting access token with scope: https://identity.mozilla.com/apps/oldsync
> 2026-04-22T04:30:10.705901Z  WARN error_support::handling: Remote server error: '403' '112' 'Forbidden' 'Forbidden' 'https://mozilla.github.io/ecosystem-platform/api#section/Response-format' tracing_support=true
> Error: forbidden
```
now upgrade to sync:
```
cargo run --example fxa-client login --scope profile --scope https://identity.mozilla.com/apps/oldsync
...
> Requesting access token with scope: https://identity.mozilla.com/apps/oldsync
> Success: AccessTokenInfo { scope: "https://identity.mozilla.com/apps/oldsync", token: "1fa52...", key: Some(ScopedKey { kty: "oct", scope: "https://identity.mozilla.com/apps/oldsync", kid: "1747756258989-9SHtUl0VJosxLZ_oUPu6rw" }), expires_at: 1776918743 }
cargo run --example fxa-client get-access-token https://identity.mozilla.com/apps/vpn
...
Requesting access token with scope: https://identity.mozilla.com/apps/vpn
Success: AccessTokenInfo { scope: "https://identity.mozilla.com/apps/vpn", token: "df1a8...", key: None, expires_at: 1776918839 }
> 
```

mhammond:
this looks like it's not a breaking change for iOS. But probably will be after @skhamis's patch lands (or if this lands first, it will break that patch)

components/fxa-client/src/internal/oauth.rs
```diff
@@ -427,16 +444,73 @@ impl FirefoxAccount {
             },
             None => None,
         };
-        // In order to keep 1 and only 1 refresh token alive per client instance,
-        // we also destroy the existing refresh token.
-        if let Some(ref refresh_token) = old_refresh_token {
+
+        if let Some(ref old_refresh_token) = old_refresh_token {
+            // We now have a new refresh token, but only with new scopes we explicitly requested.
+            // We possibly an old refresh token with only the scopes we had before.
```
bendk:
```suggestion
            // We possibly had an old refresh token with only the scopes we had before.
```

components/fxa-client/src/internal/oauth.rs
```diff
@@ -427,16 +444,73 @@ impl FirefoxAccount {
...
+            if all_scopes != new_refresh_token.scopes {
+                if let Some(session_token) = self.state.session_token() {
+                    info!("New refresh token is missing some of our old scopes, upgrading");
+                    let scopes_slice = all_scopes.iter().map(|s| s.as_ref()).collect::<Vec<&str>>();
+                    let merged_refresh_token_resp =
+                        self.client.create_refresh_token_using_session_token(
+                            self.state.config(),
+                            session_token,
+                            &scopes_slice,
+                        )?;
```
bendk, line 480: IIUC, this is a workaround to the fact that we don't want to re-request all the scopes when fetching the refresh token.  In particular, re-requesting the sync scope creates a UI issue.

Are there downsides to this approach or is this okay to keep for the long term?  The one downside I can see is that this maybe prevents us from revoking access to a scope, since the client can always get it back via the session token.  Is that right?

I think the code is fine, but can you document what the current status and plan is?

components/fxa-client/src/state_machine/internal_machines/authenticating.rs
```diff
@@ -6,7 +6,9 @@ use super::{invalid_transition, Event, InternalStateMachine, State};
 use crate::{Error, FxaEvent, FxaState, Result};
 use error_support::report_error;
 
-pub struct AuthenticatingStateMachine;
+pub struct AuthenticatingStateMachine {
+    pub is_connected: bool,
```
bendk: How about using an enum to allow for more possibilities?

In fact, maybe we should add a variant for coming from the `AuthIssues` state so that we can go back to that state on cancellation.  This would be a UX change though so that doesn't need to be in this PR, but I do like an enum in any case.

components/fxa-client/src/fxa_client.udl
```diff
@@ -1003,16 +999,16 @@ dictionary Profile {
...
   AuthIssues();
 };
 
 [Enum]
 interface FxaEvent {
   Initialize(DeviceConfig device_config);
-  BeginOAuthFlow(sequence<string> scopes, string entrypoint);
-  BeginPairingFlow(string pairing_url, sequence<string> scopes, string entrypoint);
+  BeginOAuthFlow(string service, sequence<string> scopes, string entrypoint);
+  BeginPairingFlow(string pairing_url, string service, sequence<string> scopes, string entrypoint);
```
bendk, line 1011: This is a breaking change since the clients are going to need to send us the new field.  I'm fine with that, but let's call it out in the changelog.

bendk (approved):
This makes sense to me.  I had a couple suggestions, but nothing needs re-review.

components/fxa-client/src/internal/oauth.rs
mhammond: >  In particular, re-requesting the sync scope creates a UI issue.

In particular, requesting the sync scope forces you to reenter your password. I'm not 100% sure why the server can't handle that better if it knows you already have that scope - Lauren indicated the server doesn't actually know that, which doesn't make much sense to me, but I don't know their code at all.

> Are there downsides to this approach or is this okay to keep for the long term? The one downside I can see is that this maybe prevents us from revoking access to a scope, since the client can always get it back via the session token. Is that right?

It is still making a request to the server for the new refresh token, so I guess the server could reject it still - but like desktop, it seems a little like the server doesn't verify much in that regard.

But yeah, the longer term intent is to call `exchange_token_for_scope` instead of `create_refresh_token_using_session_token`, but that's not quite ready to be used in that way - I added a note about that too.

components/fxa-client/src/state_machine/internal_machines/authenticating.rs
mhammond: That `AuthIssues` point is a good one - that's a bug, so I figured I'd fix it. I ended up using `FxaRustAuthStatus` as the enum, but some comments made me wonder if that's actually a good choice - it has the exact variants we need, but it's not clear if that's going to remain true. wdyt?

bendk: That's funny, I was just looking at the PR to remove it.  However, this seems like a totally valid use for the enum to me and I'm for it.  Maybe we should rename it at some point and/or make it an internal type.

jonalmeida (approved):
Tested out the latest patches with the updated phab link too and nothing alarming there in my manual testing.

PR #7312, opened 2026-04-08
Breach Alerts component

jo:

  - Note:


components/breach-alerts/src/api.rs
```diff
@@ -0,0 +1,221 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::error::*;
+use rusqlite::Connection;
+use sql_support::ConnExt;
+
+#[derive(Debug, Clone, PartialEq, uniffi::Record)]
+pub struct BreachAlertDismissal {
+    pub breach_id: String,
```
Vinnl: Nit: One incongruence I noticed while working with this API is that HIBP refers to this as a breach name, so I kept passing `{ breachId: breach.Name }`. That's a little bit awkward, so we could consider using `breach_name` here as well?

(But I can go with ID too, since they are unique identifiers of the breaches. Ideally HIBP had called them IDs.)

components/breach-alerts/src/api.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub dismissed_at: i64,
```
Vinnl: Nit: Could be useful to indicate a unit here, e.g.


```suggestion
    pub dismissed_at_unixtime: i64,
```

(Because I just passed in a `Date.now()` on the JS side, but I wasn't actually confident that that's what was expected.)

components/breach-alerts/src/api.rs
```diff
@@ -0,0 +1,221 @@
...
+        .map_err(Error::from)
+}
+
+pub fn set_breach_alert_dismissals(
+    conn: &Connection,
+    dismissals: &[BreachAlertDismissal],
+) -> Result<()> {
+    if dismissals.is_empty() {
+        return Ok(());
+    }
```
Vinnl, line 51: I'm wondering if we should just return an error in this case? As it's almost certainly a mistake if this code path gets hit?

(I was wondering the same for the equivalent in `get_breach_alert_dismissals`, but that makes sense when visiting an unbreached website.)

Vinnl (approved):
I can't comment on the Rust code and AS conventions, but I integrated this into Firefox using [your patch](https://phabricator.services.mozilla.com/D293480), and it did what it says on the tin :)

Added some inline suggestions, but none of them are blockers, and for each of them one could also make a valid argument that they should be ignored, so 🤷

components/breach-alerts/src/api.rs
jo: Yes good point! Changed.

components/breach-alerts/src/api.rs
jo: Good hint - I looked at all the other components and the general theme is `time_` prefix, so I adjusted it here.

components/breach-alerts/src/api.rs
jo: I think we also don't throw in other similar places and this is a quiet minimalist approach, so I'd leave it as it is.

bendk (approved):
The code looks good to me.

My only slight concern is that we should make sure we're tracking any errors that come from this component and fixing them.  At some point soon I'd love to set up a dashboard so that credential management can do that.  I've been working on the dashboards for a while and I think they're close to being ready for wider use.

jo:
> My only slight concern is that we should make sure we're tracking any errors that come from this component and fixing them. At some point soon I'd love to set up a dashboard so that credential management can do that. I've been working on the dashboards for a while and I think they're close to being ready for wider use.

Yes, this sounds good!

PR #7317, opened 2026-04-13
fxa-client: Don't expose the session token directly to consumers.

mhammond:
This was exposed primarily for use with the web channel, but the fact it is exposed means that consumers are free to use it in ways we'd like to better control. We do this by adding a couple of new methods which are explicitly used only for webchannels, and the data moved over those APIs are abstracted away - the consumers just pass the raw JSON data from the webchannel message and the component knows what format it is in and decodes it appropriately.

This is a breaking change for Android and iOS:
* Android - https://phabricator.services.mozilla.com/D293782
* iOS - https://github.com/mozilla-mobile/firefox-ios/pull/33041

@jonalmeida @bendk @skhamis @LZoog - I think this is a good improvement, wdyt?

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/oauth.rs
+++ b/components/fxa-client/src/internal/oauth.rs
@@ -11,7 +11,6 @@ use super::{
     scoped_keys::ScopedKeysFlow,
     util, FirefoxAccount,
 };
-use crate::auth::UserData;
 use crate::{
     debug, error, info, warn, AuthorizationParameters, Error, FxaServer, Result, ScopedKey,
 };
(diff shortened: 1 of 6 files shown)


bendk (approved):
I like this change.  The more we can move the logic into the component the better, IMO.  That way if there's a bug or change we only need to change the Rust code.

components/fxa-client/src/internal/oauth.rs
```diff
@@ -114,12 +113,25 @@ impl FirefoxAccount {
...
-    pub fn set_user_data(&mut self, user_data: UserData) {
-        // for now, we only have use for the session token
-        // if we'd like to implement a "Signed in but not verified" state
-        // we would also consume the other parts of the user data
-        self.state.set_session_token(user_data.session_token)
+    /// Extracts and stores the session token from a WebChannel login JSON payload.
+    /// The JSON payload is the `data` object from the `fxaccounts:login` WebChannel command.
+    pub fn handle_web_channel_login(&mut self, json_payload: &str) -> Result<()> {
+        let data: serde_json::Value = serde_json::from_str(json_payload)?;
+        if let Some(token) = data.get("sessionToken").and_then(|v| v.as_str()) {
```
skhamis: Is there ever a scenario where we wouldn't get the sessionToken? I see `NoSessionToken` below for pw change but here we still return success even if no `sessionToken`?

mhammond: yeah, good spot thanks, this should probably return the same error.

skhamis (approved):
I agree I think this is an awesome improvement! Love to see it!

jonalmeida:
This is a good step! :shipit: 

I think protecting against _using_ the session token incorrectly is more valuable than protecting against holding it - the `handleWebChannelLogin` doesn't prevent the application from grabbing the session token during the channel message nor during state persistence.

PR #7320, opened 2026-04-15
Add a new "Forbidden" FxA error.

mhammond:
This is distinct from Authentication in that there's no implication the account state isn't good, it's just that you aren't allowed to do what you tried. An example of this is asking for an access token you don't have the scopes for.

This has the side effect of fixing a bug in `get_access_token()` - when that code calls `exchange_token_for_scope()`, the server responds with a http 403 response - which ends up getting re-thrown as "other error".

It arranges for `NoCachedToken` errors to turn into this (that was previously `Authentication`) and for `NoSessionToken` to be `Authentication`. It also tweaks `get_access_token()` to explicitly throw `NoSessionToken` when there's no session token. That way `get_access_token()` continues to return `Authentication` when not logged in, but otherwise will generally return `Forbidden`.

Technically a breaking change, but both Android and iOS seem fine with new variants here.

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/oauth.rs
+++ b/components/fxa-client/src/internal/oauth.rs
@@ -12,7 +12,7 @@ use super::{
     util, FirefoxAccount,
 };
 use crate::auth::UserData;
-use crate::{warn, AuthorizationParameters, Error, FxaServer, Result, ScopedKey};
+use crate::{error, warn, AuthorizationParameters, Error, FxaServer, Result, ScopedKey};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use jwcrypto::{EncryptionAlgorithm, EncryptionParameters};
 use rate_limiter::RateLimiter;
@@ -71,10 +71,16 @@ impl FirefoxAccount {
                             new_refresh_token,
                             exchange_resp.scope,
                         ));
+                    } else {
+                        // A request for a new token succeeding but without a new token is unexpected.
+                        error!("successful response for a new refresh token with additional scopes, but no token was delivered");
+                        // at this stage we are almost certainly still going to fail to get a token...
                     }
                     // Get the updated refresh token from state.
                     refresh_token = match self.state.refresh_token() {
-                        None => return Err(Error::NoCachedToken(scope.to_string())),
+                        // We had a refresh token, we must either still have the original or maybe a new one,
+                        // but it's impossible for us to not have one at this point.
+                        None => unreachable!("lost the refresh token"),
                         Some(token) => token,
                     };
                 }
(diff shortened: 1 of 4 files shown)


bendk (approved):
This change looks good to me.  My only suggestion is maybe we could get rid of `NoCachedToken` entirely and replace it with more specific errors.  Looking through the code I see 2 remaining uses:

- [We don't have a refresh token immediately after exchanging it and setting a new one](https://github.com/mozilla/application-services/blob/ed086e7165364aabe4d7a999d83d95575297da0a/components/fxa-client/src/internal/oauth.rs#L77).  This doesn't seem possible to me.  Maybe that branch should just be `unreachable!()` or the code could be refactored so it's not needed.
- [We've exchanged the refresh token after asking for a new scope, but the new token doesn't have the scope either](https://github.com/mozilla/application-services/blob/ed086e7165364aabe4d7a999d83d95575297da0a/components/fxa-client/src/internal/oauth.rs#L89).  I think this is possible, but it represents a server error.  Maybe we could replace that with a `ScopedKeyMissingInServerResponse` error which generalizes the current `SyncScopedKeyMissingInServerResponse` error.

If it's easy and makes sense to you, then I'm for adding those changes to this PR.  Otherwise, we can land it as-is and keep iterating.

mhammond:
> This change looks good to me. My only suggestion is maybe we could get rid of NoCachedToken entirely and replace it with more specific errors.

Yeah, good idea. I agree the first one is literally impossible, so I made that `unreachable!()`. For the second case I didn't generalize `ScopedKeyMissingInServerResponse` because that has unfortunately been exposed to the mobile clients, wihch I think is unfortunate because there's really nothing they can do about it and it does imply an expected server response. So I added a new `UnexpectedServerResponse` for cases like this - eg, where we ask for a token and get a 200 response without a token, etc. I also added a note that in the future we should be able to replace `ScopedKeyMissingInServerResponse` with this, but that's a larger refactor, both in the clients and in the rust code (ie, we should probably be throwing this error at the time we expected to have got the key, not at a later time when someone finally asks for an access token?

(A problem with `UnexpectedServerResponse`, which I also added to the comments, is that in many cases this would actually show up as a JSON error - eg, when the shape of the payload is wrong. But that seems OK, and both of them end up as "OtherError" to the consumer.)

mhammond:
(`ScopedKeyMissingInServerResponse` also overlaps with `NoScopedKey`!)

components/fxa-client/src/internal/oauth.rs
```diff
@@ -86,7 +92,11 @@ impl FirefoxAccount {
                         &[scope],
                     )?
                 } else {
-                    return Err(Error::NoCachedToken(scope.to_string()));
```
jonalmeida, line 89: @mhammond I might be reading this incorrectly, but I think this else is unnecessary? Line 61 checks if the refresh token doesn't contain the scopes, and line 87 then acts as the else for it. So in that case, we shouldn't need this extra else?

mhammond: Line 87 isn't just the else, it's also handling that the first block might yet finish without the scope - which would mean we somehow got a response from the server without that scope - hence that error below. IOW, we have that error below to handle line 70 not actually checking whether we got the new scope.

This could probably be refactored to handle that case better (ie, return that error around line 70), but before I do that I want to check that makes sense?

jonalmeida: Oh I see, so it's sort of like a retry that happens in this order of operations. I wouldn't recommend refactoring this bit right now for my sake, but only if it makes sense to move it around with some more value.

Thanks!

PR #7326, opened 2026-04-17
Perform Additional Origin Fixups

jo:
Add `perform_additional_origin_fixups` feature flag that repairs `origin` values which fail URL parsing (bare domains, FireFTP quirks, bare `https:` / `https://`, etc.) into parseable URLs instead of rejecting them. Scoped to the `origin` field only; `form_action_origin` keeps the stricter behavior, but can be disabled via `ignore_form_action_origin_validation_errors`.

Intended to be enabled on desktop during migration to salvage legacy/addon-generated entries.

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -167,7 +167,7 @@ impl LoginDb {
     }
 
     pub fn count_by_form_action_origin(&self, form_action_origin: &str) -> Result<i64> {
-        match LoginEntry::validate_and_fixup_origin(form_action_origin) {
+        match LoginEntry::validate_and_normalize_form_action_origin(form_action_origin) {
             Ok(result) => {
                 let form_action_origin = result.unwrap_or(form_action_origin.to_string());
                 let mut stmt = self.db.prepare_cached(&COUNT_BY_FORM_ACTION_ORIGIN_SQL)?;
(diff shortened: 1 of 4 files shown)


components/logins/src/db.rs
```diff
@@ -167,7 +167,7 @@ impl LoginDb {
     }
 
     pub fn count_by_form_action_origin(&self, form_action_origin: &str) -> Result<i64> {
-        match LoginEntry::validate_and_fixup_origin(form_action_origin) {
+        match LoginEntry::validate_and_fixup_form_action_origin(form_action_origin) {
```
bendk: Maybe this should just be `validate_form_action_origin` at this point.  I think all the fixup code has moved out of it.

components/logins/src/login.rs
```diff
@@ -382,6 +382,71 @@ pub struct LoginEntry {
     pub password: String,
 }
 
+#[cfg(feature = "perform_additional_origin_fixups")]
+fn looks_like_bare_ipv4(s: &str) -> bool {
```
bendk: When I first saw this I thought it was operating on the entire URL, but I think it's just the origin.  Maybe the names could be updated to make it more clear.

What about creating a module called something like `origin_fixup`?  That way it's obvious what all the functions are operating over.  Also, you'd could have one `#[cfg]` attr around the entire module.  Either an inline module or a separate file could work.

bendk (approved):
Looks very nice to me.  I'm picturing you going through some huge log of invalid URLs to get this code.  I just had a couple naming/code organization suggestions, but nothing blocking.

components/logins/src/login.rs
jo: Yes, good idea, thanks!

components/logins/src/db.rs
jo: Yeah that's right. I added the term `normalize` to hint that this methods returns something useful. For the origin I think `fixup` includes normalization, so I think that's fine.

PR #7342, opened 2026-04-28
[RS] Run maintenance after each sync

issammani:

  - Note:

### DIFF
--- a/components/remote_settings/src/client.rs
+++ b/components/remote_settings/src/client.rs
@@ -399,6 +399,11 @@ impl<C: ApiClient> RemoteSettingsClient<C> {
         Ok(())
     }
 
+    pub fn run_maintenance(&self) -> Result<()> {
+        let mut inner = self.lock_inner()?;
+        inner.storage.run_maintenance()
+    }
+
     pub fn reset_storage(&self) -> Result<()> {
         trace!("{0}: reset local storage.", self.collection_name);
         let mut inner = self.lock_inner()?;
--- a/components/remote_settings/src/service.rs
+++ b/components/remote_settings/src/service.rs
@@ -253,9 +253,8 @@ struct ChangesCollection {
 mod test {
     use super::*;
     use crate::telemetry::UptakeEventExtras;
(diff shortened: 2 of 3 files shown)


bendk (approved):
Changes look great, thanks for fixing this.

jonalmeida (approved):
r+ we will need client-side patches for iOS and Android to execute this I think, but can get there easily since we've done this before and we don't have breaking API changes here.

bendk:
> r+ we will need client-side patches for iOS and Android to execute this I think, but can get there easily since we've done this before and we don't have breaking API changes here.

I think the mobile clients are already calling `RemoteSettingsService::sync()` which means they'll get this automatically.  So we should double check that for sure, but maybe we don't need to land any other patches.

issammani:
> > r+ we will need client-side patches for iOS and Android to execute this I think, but can get there easily since we've done this before and we don't have breaking API changes here.
> 
> I think the mobile clients are already calling `RemoteSettingsService::sync()` which means they'll get this automatically. So we should double check that for sure, but maybe we don't need to land any other patches.

In the worst case scenario, we will run the `sync` one day after the fix lands. This should be enough.

jonalmeida:
> > > r+ we will need client-side patches for iOS and Android to execute this I think, but can get there easily since we've done this before and we don't have breaking API changes here.
> > 
> > 
> > I think the mobile clients are already calling `RemoteSettingsService::sync()` which means they'll get this automatically. So we should double check that for sure, but maybe we don't need to land any other patches.
> 
> In the worst case scenario, we will run the `sync` one day after the fix lands. This should be enough.

Thanks, good catch. We are running sync [every 24 hours](https://searchfox.org/firefox-main/rev/298f495e520d43a74d22d83ce39b78eb47d89563/mobile/android/android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsSyncScheduler.kt#43), but that is going to be increased to [every 2 hours](https://phabricator.services.mozilla.com/D296172).

issammani:

PR #7345, opened 2026-04-30
[DISCO-4151] Merino: add client for wcs

tiftran:
This PR adds a new module to the merino crate that calls the Merino /api/v1/wcs endpoint
Also added a suggest subcommand to merino-cli for local testing.

  - Note:

### DIFF
--- a/components/merino/src/worldcup/http.rs
+++ b/components/merino/src/worldcup/http.rs
@@ -1,0 +1,138 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use url::Url;
+use viaduct::{Request, Response};
+
+use super::error::{Error, Result};
+
+pub struct HttpClient;
+
+#[derive(Default)]
+pub struct WorldCupQueryParams {
+    pub limit: Option<u32>,
+    pub teams: Option<String>,
+}
+
+pub trait HttpClientTrait {
+    fn make_request(&self, url: Url, params: WorldCupQueryParams) -> Result<Option<Response>>;
+}
+
+impl HttpClientTrait for HttpClient {
+    fn make_request(&self, url: Url, params: WorldCupQueryParams) -> Result<Option<Response>> {
+        send_get(build_url(url, &params))
+    }
+}
+
+pub fn build_url(endpoint_url: Url, params: &WorldCupQueryParams) -> Url {
+    if params.limit.is_none() && params.teams.is_none() {
+        return endpoint_url;
+    }
+    let mut url = endpoint_url;
+    {
+        let mut pairs = url.query_pairs_mut();
+        if let Some(v) = params.limit {
+            pairs.append_pair("limit", &v.to_string());
+        }
+        if let Some(v) = &params.teams {
+            pairs.append_pair("teams", v);
+        }
+    }
+    url
+}
+
+fn send_get(url: Url) -> Result<Option<Response>> {
+    let response = Request::get(url)
+        .header("accept", "application/json")?
+        .send()?;
+    let status = response.status;
+    match status {
+        200 => Ok(Some(response)),
+        204 => Ok(None),
+        400 => Err(Error::BadRequest {
+            code: status,
+            message: response.text().to_string(),
+        }),
+        422 => Err(Error::Validation {
+            code: status,
+            message: response.text().to_string(),
+        }),
+        500..=599 => Err(Error::Server {
+            code: status,
+            message: response.text().to_string(),
+        }),
+        _ => Err(Error::Unexpected {
+            code: status,
+            message: response.text().to_string(),
+        }),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const BASE_URL: &str = "https://merino.services.mozilla.com/api/v1/wcs/teams";
+
+    fn base_url() -> Url {
+        Url::parse(BASE_URL).unwrap()
+    }
+
+    fn has_param(url: &Url, key: &str, value: &str) -> bool {
+        url.query_pairs().any(|(k, v)| k == key && v == value)
+    }
+
+    #[test]
+    fn test_build_url_with_params() {
+        let options = WorldCupQueryParams {
+            limit: Some(10),
+            ..WorldCupQueryParams::default()
+        };
+        let url = build_url(base_url(), &options);
+        assert!(has_param(&url, "limit", "10"));
+    }
+
+    #[test]
+    fn test_build_url_with_teams() {
+        let options = WorldCupQueryParams {
+            teams: Some("FRA,ENG".to_string()),
+            ..WorldCupQueryParams::default()
+        };
+        let url = build_url(base_url(), &options);
+        assert!(has_param(&url, "teams", "FRA,ENG"));
+    }
+
+    #[test]
+    fn test_build_url_with_all_options() {
+        let options = WorldCupQueryParams {
+            limit: Some(5),
+            teams: Some("FRA,ENG".to_string()),
+        };
+        let url = build_url(base_url(), &options);
+        assert!(has_param(&url, "limit", "5"));
+        assert!(has_param(&url, "teams", "FRA,ENG"));
+    }
+
+    #[test]
+    fn test_build_url_omits_none_options() {
+        let url = build_url(base_url(), &WorldCupQueryParams::default());
+        let keys: Vec<_> = url.query_pairs().map(|(k, _)| k.into_owned()).collect();
+        assert_eq!(keys.len(), 0);
+    }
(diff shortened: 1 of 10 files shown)


components/merino/src/worldcup/http.rs
```diff
@@ -0,0 +1,176 @@
...
+            ..WorldCupQueryParams::default()
+        };
+        let url = build_url(base_url(), &options);
+        assert!(has_param(&url, "teams", "FRA,ENG"));
+    }
+
+    #[test]
+    fn test_build_url_with_date() {
+        let options = WorldCupQueryParams {
+            date: Some("2026-06-15".to_string()),
```
annasob: non blocking but lets update these dates to use something like Duration::from_secs

examples/merino-cli/src/main.rs
```diff
@@ -85,6 +86,41 @@ enum Commands {
         #[arg(long)]
         accept_language: Option<String>,
     },
+    /// Fetch World Cup data
+    WorldCup {
+        /// OHTTP relay URL
+        #[arg(long, default_value = "https://ohttp-merino.mozilla.fastly-edge.com")]
```
annasob: Nan said no need for OHTTP since no search data is being passed.

annasob (changes requested):

components/merino/src/worldcup/http.rs
tiftran: are you asking to have the time in the date string?

examples/merino-cli/src/main.rs
tiftran: right

components/merino/src/worldcup/http.rs
tiftran: ~nvm, removing param, as it was removed from teh api contract~

tiftran: yup removing the param

components/merino/src/worldcup/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.base_host = Some(base_host);
+        self
+    }
+
+    pub fn build(self) -> Result<WorldCupClient> {
+        let base_host = self
+            .base_host
+            .unwrap_or_else(|| DEFAULT_BASE_HOST.to_string());
+
+        let base_url = Url::parse(&format!("{}/api/v1/wcs/", base_host))?;
```
gruberb: We  could add the base url as a const (in addition to the `DEFAULT_BASE_HOST`? Or just do `DEFAULT_BASE_URL`? 

And then we can use this in the tests too (import it there) instead of having another hardcoded url.

components/merino/src/worldcup/tests.rs
```diff
@@ -0,0 +1,713 @@
...
+      "region": "BRA",
+      "colors": ["Yellow", "Green", "Blue"],
+      "icon": "https://example.com/logos/brazil.png",
+      "group": "Group A",
+      "eliminated": false
+    }]
+}"#;
+
+const LIVE_RESPONSE: &str = r#"{
+  "current": [
```
gruberb: Another little thing is to use a `const ICON_URL_BASE: &'static str = "https://storage.googleapis.com/merino-images-prod/logos/"` maybe for the logos?

components/merino/src/worldcup/http.rs
```diff
@@ -0,0 +1,158 @@
...
+
+pub struct HttpClient;
+
+#[derive(Default)]
+pub struct WorldCupQueryParams {
+    pub limit: Option<u32>,
+    pub teams: Option<String>,
+}
+
+pub trait HttpClientTrait {
```
gruberb, line 19: They are all the same. Can we collapse it into one `make_request`?

```rust
pub trait HttpClientTrait {
    fn make_request(&self, url: Url, params: WorldCupQueryParams) -> Result<Option<Response>>;
}
```

components/merino/src/worldcup/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+            .as_ref()
+            .filter(|v| !v.is_empty())
+            .map(|v| v.join(","));
+        http::WorldCupQueryParams {
+            limit: options.limit,
+            teams,
+        }
+    }
+
+    pub fn get_teams(
```
gruberb: Could all just be `make_request` and then `self.http_client.make_request`?

components/merino/src/worldcup/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        })
+    }
+}
+
+#[uniffi::export]
+impl WorldCupClient {
+    /// Creates a new `WorldCupClient` from the given configuration.
+    #[uniffi::constructor]
+    #[handle_error(Error)]
+    pub fn new(base_host: Option<String>) -> ApiResult<Self> {
```
gruberb: We should pass a `WorldCupConfig` here instead of a String. And I don't think it can be optional.

components/merino/src/worldcup/tests.rs
```diff
@@ -0,0 +1,713 @@
...
+      "name": "Brazil",
+      "region": "BRA",
+      "colors": ["Yellow", "Green", "Blue"],
+      "icon": "https://example.com/logos/brazil.png",
+      "group": "Group A",
+      "eliminated": false
+    }]
+}"#;
+
+const LIVE_RESPONSE: &str = r#"{
```
gruberb: We could move this to `components/merino/src/worldcup/fixtures/{teams,matches,live}.json` and then load it here via `include_str!`

components/merino/src/worldcup/tests.rs
```diff
@@ -0,0 +1,713 @@
...
+        limit: None,
+        teams: None,
+    }
+}
+
+fn base_url() -> Url {
+    WorldCupClientBuilder::new().build().unwrap().base_url
+}
+
+struct FakeHttpClientSuccess;
```
gruberb: `FakeHttpClientSuccess`, `FakeHttpClientNoContent`, `FakeHttpClientServerError` differ only in their return values. With the trait collapsed to just one method (from the comment earlier), they reduce to one parameterized fake:

```rust
struct FakeHttpClient(fn() -> Result<Option<Response>>);
impl http::HttpClientTrait for FakeHttpClient { ... }
```

components/merino/src/worldcup/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    inner: WorldCupClientInner<http::HttpClient>,
+    base_url: Url,
+}
+
+struct WorldCupClientInner<T: http::HttpClientTrait> {
+    http_client: T,
+}
+
+#[derive(Default)]
+pub struct WorldCupClientBuilder {
```
gruberb: Nit: Doesn't need to be `pub`

gruberb (commented):
I would recommend these changes. Nothing functional, but would be cleaner. Knowing that we likely remove all of that again after the event.

annasob (approved):
Approving to clear my block

components/merino/src/worldcup/mod.rs
gruberb: I mean, if we want to hardcode the endpoint to prevent typos etc, we have to use a fixed `url` here for each of the requests.

components/merino/src/worldcup/tests.rs
tiftran: actually i guess if we're storing those responses in json files, we can't do this right?

components/merino/src/worldcup/http.rs
tiftran: makes sense

components/merino/src/worldcup/mod.rs
tiftran: i kinda do want to leave it, so that the endpoint is hardcoded here. Yes this leaves to some duplication, but maybe okay since it won't be sticking around for long?

components/merino/src/worldcup/mod.rs
tiftran: Suggest and CuratedRecommendations both have `Option<String>` and it defaults to the merino prod if it's not present, should I follow convention?

components/merino/src/worldcup/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+#[cfg(test)]
+mod tests;
+
+use error_support::handle_error;
+use url::Url;
+
+pub use error::{ApiResult, Error, MerinoWorldCupApiError, Result};
+pub use schema::{WorldCupConfig, WorldCupOptions};
+
+pub(crate) const DEFAULT_BASE_URL: &str =
```
tiftran: not sure if this is the best way to make it accessible for `test.rs`?

tiftran:
~Added the accept language header and a "language" param, will remove/edit once there's a decision on how we're passing locale~

gruberb (approved):

ncloudioj (approved):

tiftran:
@Mergifyio backport release-v151

PR #7347, opened 2026-05-04
test(ads-client): assert mockito expectations to prevent silent false-positives

Almaju:
## Summary

Some ads-client tests configured mockito mocks with `.expect(N)` to declare an expected request count, but never called `.assert()` to verify it. Without the assert, mockito does not enforce the expectation, so tests pass even when the underlying behavior is wrong.

This PR pairs every interaction-verifying mock setup with an explicit `.assert()` call across `mars.rs`, `http_cache.rs`, and `client.rs` (including the ignored `test_record_click_invalidates_cache`).

While doing this, the missing `.assert()` surfaced a real silent false-positive: `mars::tests::test_fetch_ads_cache_hit_skips_network` was constructing the test client with `None` for the cache, so both calls hit the network and the cache-hit path the test claimed to verify was never actually exercised. The test now constructs a real `HttpCache` so the second call serves from cache, and the `.assert()` enforces that only one network request was made.

A couple of duplicate-mock setups in `http_cache.rs` (`test_refresh_policy_always_uses_network_then_caches`, `test_expired_entry_is_a_miss_on_next_send`) were collapsed into a single mock with `expect(2)`. The bodies were never inspected, and a single mock makes the network call count directly verifiable. `test_not_cacheable_no_store` was tightened to assert both calls hit the network as `MissNotCacheable`.

## Test plan

- [x] `cargo test -p ads-client --lib` — all 77 tests pass (1 ignored, unchanged)
- [x] Confirmed the previously silent test failure is now caught: reverting the `make_test_client(Some(cache))` change in `test_fetch_ads_cache_hit_skips_network` causes the test to fail with "Expected 1 request(s) ... but received 2"

### DIFF
--- a/components/ads-client/src/client.rs
+++ b/components/ads-client/src/client.rs
@@ -280,7 +280,7 @@ mod tests {
         viaduct_dev::init_backend_dev();
 
         let expected_response = get_example_happy_image_response();
-        let _m = mockito::mock("POST", "/ads")
+        let m = mockito::mock("POST", "/ads")
             .with_status(200)
             .with_header("content-type", "application/json")
             .with_body(serde_json::to_string(&expected_response.data).unwrap())
@@ -291,6 +291,7 @@ mod tests {
 
         let result = ads_client.request_image_ads(make_happy_placement_requests(), None, false);
         assert!(result.is_ok());
+        m.assert();
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


copyrighthero (approved):
Verified locally that the test ran successfully!

copyrighthero:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #7354, opened 2026-05-07
Bug 2037747 - Add per-feature update notifications to the Android Nimbus client [ci full]

freshstrangemusic:

  - Note:

### DIFF
--- a/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
+++ b/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
@@ -24,7 +24,6 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import mozilla.appservices.remotesettings.RemoteSettingsService
 import mozilla.telemetry.glean.Glean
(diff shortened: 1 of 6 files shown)


components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -377,11 +389,22 @@ open class Nimbus(
         }
 
     @WorkerThread
-    private fun postEnrolmentCalculation() {
-        nimbusClient.getActiveExperiments().let {
-            recordExperimentTelemetry(it)
+    private fun postEnrolmentCalculation(initial: Boolean) {
```
jeddai: I'd recommend defaulting this to `false` here

```suggestion
    private fun postEnrolmentCalculation(initial: Boolean = false) {
```

Then when using it you only have to pass it for the initial call.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
```diff
@@ -470,18 +496,25 @@ open class Nimbus(
...
+                registerPreviousGeckoPrefStatesOnThisThread(geckoPrefStates)
             }
         }
     }
 
+    @WorkerThread
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun registerPreviousGeckoPrefStatesOnThisThread(geckoPrefStates: List<GeckoPrefState>) {
+        nimbusClient.registerPreviousGeckoPrefStates(geckoPrefStates)
+    }
```
jeddai, line 515: in other places your `OnThisThread` methods handle the `withCatchAll` inside them, I think this should be updated to be the same.

jeddai (approved):
r+wc so there isn't another review pass required, but I do think the comments should be addressed.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
freshstrangemusic: Whoops! Good catch.

Long term I think we should revisit this decision to have withCatchAll() on the '*OnThisThread* methods because it results in the no errors being thrown in tests.

components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt
freshstrangemusic: 👍

PR #7357, opened 2026-05-08
Allow get_access_token to return tokens with multiple scopes

mhammond:
This brings mobile in line with desktop. Multiple scopes can be specified. These scopes are normalized and used as the cache key.

Split into 2 commits for ease of review:
* first patch with the changes
* second patch splits the access token code and tests into its own module, so oauth.rs stays a sane size. This commit has no actual functionality change.

  - Note:

### DIFF
--- a/components/fxa-client/src/error.rs
+++ b/components/fxa-client/src/error.rs
@@ -73,9 +73,6 @@ pub enum Error {
     #[error("Unknown OAuth State")]
     UnknownOAuthState,
 
-    #[error("Multiple OAuth scopes requested")]
-    MultipleScopesRequested,
-
     #[error("No cached scoped keys for scope {0}")]
     NoScopedKey(String),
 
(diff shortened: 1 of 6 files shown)


components/fxa-client/src/internal/oauth.rs
```diff
@@ -56,14 +71,21 @@ impl FirefoxAccount {
                 }
             }
         }
+        let mut requested_scopes: Vec<&str> = requested_set.iter().copied().collect();
+        requested_scopes.sort_unstable();
```
bendk, line 75: Nit: Now that we have `normalize_scopes`, I think this could just be `let requested_scopes: Vec<&str> = requested.split_ascii_whitespace().collect()`.

bendk (approved):
Looks good to me.  I had a small suggestion, but no need for re-review.

components/fxa-client/src/error.rs
```diff
@@ -73,9 +73,6 @@ pub enum Error {
     #[error("Unknown OAuth State")]
     UnknownOAuthState,
 
-    #[error("Multiple OAuth scopes requested")]
```
jonalmeida, line 76: Good riddance!

jonalmeida (approved):
I can verify this works for me! 🙌

jonalmeida:
I'd like to merge this since it's important for VPN and I'm relying on it - with my manual testing, I feel confident with the changes and we can address any comments in a follow-up.

PR #7358, opened 2026-05-08
Put the new telemetry submission capability behind a feature

mhammond:
This is slightly problematic on desktop. There's a good chance a callback interface would work, but that's a little problematic in other ways (as that forces you to use a `Box<dyn T>` instead of an `Arc<>`, which is particularly problematic for tests. This seems like a reasonable compromise while we work out a better option.

  - Note:

### DIFF
--- a/components/remote_settings/Cargo.toml
+++ b/components/remote_settings/Cargo.toml
@@ -11,6 +11,8 @@ exclude = ["/android", "/ios"]
 [features]
 default = []
 signatures = ["dep:canonical_json", "dep:rc_crypto"]
+# submitting glean telemetry is tricky on desktop for now.
+telemetry-submission = []
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/components/remote_settings/src/lib.rs
+++ b/components/remote_settings/src/lib.rs
@@ -61,14 +61,6 @@ impl RemoteSettingsService {
         }
     }
 
-    /// Set the telemetry implementation used to record Glean metrics.
-    /// This should be set to a real implementation (eg. Kotlin, Swift).
-    /// If not set, all metric recording is a no-op.
-    pub fn set_telemetry(&self, telemetry: Arc<dyn RemoteSettingsTelemetry>) {
-        self.internal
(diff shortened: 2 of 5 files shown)


bendk (approved):
This seems like a good workaround for now until we find a better solution for trait interfaces on desktop.

leplatrem (approved):
If that helps, let's do it, sure! 

Our original goal is to have some client telemetry for when we'll gradually migrate to the v2 API. And currently we only have glean on our desktop client. We would want the A-S client to be able to report sync uptake too, so that we can detect issue if any.

I'm not too familiar with the architecture of the megazords and the implications of disabling the feature by default. But if that's possible, I would suggest enabling the feature by default, and only disabling it where it causes issues.

Also, I chose this pattern because it seems to be a possible way to go, but I'm not familiar enough with FFi and runtime dyn pointers, etc. And if there's a better way to plug Glean into A-S, now would be a good time to redesign it because we haven't landed the iOS and Android parts yet...  I believe we could mimic what the Ads or Nimbus client does, no? Would one of you have some time for some hand holding? 

Thank you 🙏

mhammond:
> And currently we only have glean on our desktop client.

Ouch - it's only desktop which has this problem. And the problem is the threading model. In a mobile context, the callback fully completes on the thread it is called from. However, on Desktop, these callbacks are called from some thread other than the main thread, but the callback can only be actually called from the main thread as that's the only place JS code can run. We possibly could arrange for xpcom code to block the calling thread and marshal things to the main thread, but that's complicated and not clear it would actually meet expectations. In this particular scenario, it could be "fire and forget" as there's no return value needed, but that's not always the case.

I also think Ads is likely to have the exact same problem once they target desktop. I'm not that familiar with nimbus but AFAIK, that's not used on desktop either. There are some glean changes coming down the line which should help here - but while these specific use-cases are glean, not all of them are. So we need a better story.

> But if that's possible, I would suggest enabling the feature by default, and only disabling it where it causes issues.

Ironically, the only place it is causing issues is the only place it is used. The only place where it doesn't cause issues are the places where you haven't landed the bits yet.

I think I'll just land what I have and prioritize removing the feature flag entirely.

leplatrem:
It makes sense, thank you Mark.

PR #7359, opened 2026-05-08
Refactor FxA state machine to simplify transitions

skhamis:
Now that FxiOS has the new state machine, we're able to now re-look at the state machine system as a whole and we realized we can simplify it now that we're not merging with existing systems on the consumers. My main goals were:

1. Make it easy to reason about how clients move through the FxA flows
2. Make it simple to add/modify existing states
3. Try limit any API churn (for now), and simplify the internal parts before any public-facing API changes (if any)

@bendk had to manage two state machines in android which was the reason for the initial system, but now that they've migrated fully over, the idea came that we can remove the internal state completely and move to a flat-state-machine style. It's in draft now as I think this is a strong denature and want to get thoughts on this before moving on.

  - Note:


components/fxa-client/src/state_machine/helpers.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// `target` as the new public state.
+    Handled { cause: Box<Error>, target: FxaState },
+    /// Programming error. Driver returns `Err(*cause)`; public state unchanged.
+    Fatal(Box<Error>),
+}
+
+impl StateMachineErr {
+    /// Programming errors (logic / invalid-transition) become `Fatal` and
+    /// ignore `target_if_handled`; everything else becomes `Handled`.
+    pub fn from_cause(cause: Error, target_if_handled: FxaState) -> Self {
```
bendk: Nit: `StateMachineErr:from_cause()` and `Result<T, StateMachineError>::err_state()` do pretty much the same thing but have very different names.  I can't think of a good single name, but maybe we could get a pair of names that seem related like  `StateMachineErr::new()` and `Result<T, StateMachineError>::map_err_to_state_machine_err()` .  I don't really love those names either though, so whatever you want to go with is good with me.

components/fxa-client/src/state_machine/transitions.rs
```diff
@@ -0,0 +1,305 @@
...
+            FxaEvent::BeginOAuthFlow {
+                service,
+                scopes,
+                entrypoint,
+            },
+        ) => {
+            let scope_refs: Vec<&str> = scopes.iter().map(String::as_str).collect();
+            let oauth_url = account
+                .begin_oauth_flow(&service, &scope_refs, &entrypoint)
+                .err_state(|| S::Disconnected)?;
```
bendk: I think there could be a default handler that would keep the user on their current state.  Would that be better than having to write out the `err_state` for every call?  I'm not sure, but I think I like having to write out the error handling logic explicitly.

bendk (commented):
I like this approach a lot, it feels way more direct than having the internal states implicitly trigger a FirefoxAccount method call.   I just left some comments/questions about the details.

components/fxa-client/src/state_machine/transitions.rs
skhamis: Yeah this is a tough one, I do agree with writing out the error handling logic explicitly cause you kinda need to think about "what does this mean and where to do we go", but I think keeping the user in the current state as default might bite us later and I don't think we'll be adding _that_ many states for FxA that we'd need to be ergonomic about it? Happy to still continue if we think so.

components/fxa-client/src/state_machine/helpers.rs
skhamis: ah good callouts, I'll go to `::new()` as that feels the most "rust-y". I feel like for chaining on the result maybe going `to_state_machine_err` has the strongest readability imo.

components/fxa-client/src/state_machine/transitions.rs
bendk: Nope, I was just thinking out loud and wondering what you thought.  The current design is good for me, let's keep using it.

skhamis:
@mhammond since you also recently used the FxA state machine/authorization I'd love to also get your thoughts to ensure we're moving in the right direction here.

components/fxa-client/src/state_machine/helpers.rs
```diff
@@ -0,0 +1,456 @@
...
+                    crate::warn!("handling error: {e}");
+                    match e.get_error_handling().err {
+                        FxaError::Network if network_retries < self.policy.network_retries => {
+                            network_retries += 1;
+                            continue;
+                        }
+                        FxaError::Authentication
+                            if self.policy.auth_retry_with_cache_clear && !auth_retried =>
+                        {
+                            self.inner.clear_access_token_cache();
```
mhammond, line 235: I'm actually skeptical this is needed here (I don't think access tokens are used by the state machine?) - but I see it already existed, so I wont get too distracted by it :)

components/fxa-client/src/state_machine/README.md
```diff
@@ -1,85 +1,58 @@
...
-    InitializeDevice --> |Error| Cancel:::terminal
+`Authenticating { initial_state }` tracks where the user came from. Error and cancel paths from `Authenticating` return to `initial_state.into()` (not always `Disconnected`) — so a re-auth attempt from `AuthIssues` that the user cancels lands back at `AuthIssues`, and a re-auth from `Connected` that errors out keeps the user at `Connected`. The exception is `InitializeDevice` errors, which always land at `Disconnected`. A `CompleteOAuthFlow` success from `Authenticating { initial_state: Connected }` skips `InitializeDevice` because the device is already initialized.
 
-    classDef default fill:#0af, color:black, stroke:black
-    classDef terminal fill:#FC766A, stroke: black;
-```
+## Retry behavior
 
-## Uninitialized
+`RetryingAccount` applies the same policy the old `CallErrorHandler` did:
```
mhammond: mentioning the "old `CallErrorHandler`" doesn't really make sense for future readers of this.

components/fxa-client/src/state_machine/README.md
```diff
@@ -1,85 +1,58 @@
...
+2. **`helpers.rs`** — the supporting types:
+   - [`RetryingAccount`] wraps a `&mut FirefoxAccount` and exposes only the methods the FSM uses, with retry policy applied automatically. Holding a `&mut RetryingAccount` instead of a `&mut FirefoxAccount` makes it hard to call a network method without retry.
+   - [`StateMachineErr`] + [`ResultExt::to_state_machine_err()`] extension trait give the `?` ergonomics for "on error, transition to this state".
+   - [`RetryPolicy`] holds the network-retry count and auth-recovery flag.
 
-For each public state, we also define an internal state machine that represents the process of transitioning out of that state.
-Internal state machine states correspond to `FirefoxAccount` method calls and events correspond to call results.
-Unlike the public state machine, the internal state machines are deterministic meaning that each `(state, event)` pair always results in the same next state.
+The driver in `mod.rs` is now thin: validate `Initialize` invariant, build a `RetryingAccount`, call `transition()` once, route the error (if any) through `convert_log_report_error` for logging/Sentry, commit the new state, fire `on_auth_issues()` if applicable.
```
mhammond: similarly, "now thin" would make sense in a changelog, but not really here.

components/fxa-client/src/state_machine/transitions.rs
```diff
@@ -0,0 +1,307 @@
...
+                        Ok(_) => Ok(S::Connected),
+                        Err(e) if is_auth_error(&e) => {
+                            // Report inline since we don't propagate this error to the driver.
+                            let _: FxaError = convert_log_report_error(e);
+                            let active = account
+                                .check_authorization_status()
+                                .to_state_machine_err(|| S::AuthIssues)?;
+                            Ok(if active { S::Connected } else { S::AuthIssues })
+                        }
+                        Err(cause) => Err(StateMachineErr::new(cause, S::Disconnected)),
```
mhammond, line 44: this seems a little wrong (and may also be wrong currently) - if `ensure_capabilities()` fails due to a non-auth error (eg, a network error), the state becomes disconnected? Shouldn't this use something like `with_auth_recovery` where there's a bit of retry logic (even though, as I mentioned, clearing the access token cache and retrying isn't going to help in practice because `ensure_capabilities()` doesn't use an access token).

And for the `is_auth_error` case, it also seems wrong that failure to ensure capabilities due to an auth error ends up in the `Connected` state when we haven't set the device capabilities?

components/fxa-client/src/state_machine/transitions.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+        (
+            S::Connected,
+            FxaEvent::BeginOAuthFlow {
+                service,
+                scopes,
+                entrypoint,
+            },
+        ) => {
+            // Re-auth: lands back at Connected after CompleteOAuthFlow because
```
mhammond: I tend to think of the term "reauth" meaning "recover from auth issues" - this is "authorize new scopes"

mhammond (approved):
this is a nice improvement, thanks!

components/fxa-client/src/state_machine/transitions.rs
skhamis: I definitely agree with this, I think this makes sense to prioritize in the follow-up where we're removing some of the public APIs and then we can start changing what we're actually doing in the state machine. Right now I'm leaning towards just keeping the patch for simplifying and consolidating just incase we need to think through what actually happens here (and also maybe removing clearing the access token too)

components/fxa-client/src/state_machine/mod.rs
```diff
@@ -31,102 +27,199 @@ impl FirefoxAccount {
...
 
-    /// Handles initialization before we process an event
-    ///
-    /// This checks that the first event we see is `FxaEvent::Initialize` and it returns the
-    /// `DeviceConfig` from that event.
-    fn handle_state_machine_initialization(&mut self, event: &FxaEvent) -> Result<DeviceConfig> {
-        match &event {
+    /// Seeds `device_config` from the first `Initialize`; rejects other events
+    /// before that and a second `Initialize` afterwards.
+    fn handle_state_machine_initialization(&mut self, event: &FxaEvent) -> Result<()> {
```
bendk, line 64: Nit: Maybe this could be moved to the top of `process_event` since it's so coupled to that logic.  I don't have a strong feeling either way though.

bendk (approved):
This still looks great to me, I just have a very small suggestion.

components/fxa-client/src/state_machine/mod.rs
skhamis: I was initially leaning towards just doing that but I think if we ever expand the init of the state machine I feel it might read easier as separate (how you had it before), we'll try keep it and can always fold it in later!

PR #7370, opened 2026-05-14
Update various Gradle dependencies

rvandermeulen:
It's that time again.


mhammond (approved):
Not just gradle, but glean too, including on iOS - so maybe adjust the title? Doesn't really matter though...

mhammond:
(doesn't matter enough to unqueue to me :)

PR #7375, opened 2026-05-18
Bug 2039791 - Filter orphaned and non-folder-parent structure rows from fetch_remote_tree

mhammond:
moz_bookmarks_synced_structure rows can accumulate with tombstoned/absent or non-folder parent endpoints. The strict dogear by_children call in pass 3 of fetch_remote_tree would fail on these with MissingParentForUnknownChild or InvalidParent respectively.

Workaround: join against moz_bookmarks_synced to both report and skip rows whose endpoints are tombstoned/absent or whose parent is not a folder.
* report via `report_error!` so we can learn more about what's wrong.
* skip so dogear doesn't see the errors, with the expectation being that either the skip causes no problems (eg, tombstoned items) or are reparented (eg, when parent is missing entirely)

Note this is on top of #7372.

### DIFF
--- a/components/places/src/bookmark_sync/engine.rs
+++ b/components/places/src/bookmark_sync/engine.rs
@@ -1641,11 +1641,88 @@ impl dogear::Store for Merger<'_> {
             }
         }
 
+        // An "endpoint" here means the guid/parent in a `moz_bookmarks_synced_structure` row.
+        // There are *possibly* schema changes here which would make some of these states unrepresentable.
+        // eg, no FK relationship between `moz_bookmarks_synced_structure` and `moz_bookmarks_synced(guid)`,
+        // and a trigger to reject non-parent items. However, other existing code should already be
+        // enforcing these invariants.
+        // Working out which shape(s) are occurring in the wild might help us narrow down where the
+        // defect is.
+        //
+        // So we check for and report structure rows with bad endpoints before skipping them.
+        let folder_kind = SyncedBookmarkKind::Folder as u8;
+        let root_guid_str = BookmarkRootGuid::Root.as_guid();
+        let (child_tombstoned, child_absent, parent_tombstoned, parent_absent, non_folder_parent):
+            (bool, bool, bool, bool, bool) = self.db.query_row(
+            &format!(
+                "SELECT
+                    MAX(child.isDeleted IS 1),
+                    MAX(child.guid IS NULL),
+                    MAX(parent.isDeleted IS 1),
+                    MAX(parent.guid IS NULL),
+                    MAX(parent.isDeleted IS 0 AND parent.kind <> {folder_kind})
+                    FROM moz_bookmarks_synced_structure s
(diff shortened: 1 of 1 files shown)


components/places/src/bookmark_sync/engine.rs
```diff
@@ -1641,11 +1641,99 @@ impl dogear::Store for Merger<'_> {
...
+        // enforcing these invariants.
+        // Working out which shape(s) are occurring in the wild might help us narrow down where the
+        // defect is.
+        //
+        // So we check for and report structure rows with bad endpoints before skipping them.
+        let folder_kind = SyncedBookmarkKind::Folder as u8;
+        let root_guid_str = BookmarkRootGuid::Root.as_guid();
+        let (child_tombstoned, child_absent, parent_tombstoned, parent_absent, non_folder_parent):
+            (bool, bool, bool, bool, bool) = self.db.query_row(
+            &format!(
```
skhamis, line 1657: Instead of the exists here we could potentially use `MAX` that way we don't need the 5 exists. This would reduce the amount of scans this query would be doing.

```
SELECT
  MAX(child.isDeleted IS 1),
  MAX(child.guid IS NULL),
  MAX(parent.isDeleted IS 1),
  MAX(parent.guid IS NULL),
  MAX(parent.isDeleted IS 0 AND parent.kind <> ?)
FROM moz_bookmarks_synced_structure s
LEFT JOIN moz_bookmarks_synced child  ON child.guid  = s.guid
LEFT JOIN moz_bookmarks_synced parent ON parent.guid = s.parentGuid
WHERE s.guid <> ?
```

skhamis: Actually while thinking about it more we could also just do:
```
SELECT
  SUM(child.isDeleted IS 1),
  SUM(child.guid IS NULL),
  SUM(parent.isDeleted IS 1),
  SUM(parent.guid IS NULL),
  SUM(parent.isDeleted IS 0 AND parent.kind <> ?)
FROM moz_bookmarks_synced_structure s
LEFT JOIN moz_bookmarks_synced child  ON child.guid  = s.guid
LEFT JOIN moz_bookmarks_synced parent ON parent.guid = s.parentGuid
WHERE s.guid <> ?
```
Then we can know exactly how many rows are tombstoned? Idk if we really care about that though in regards to what you're trying to actually do.

skhamis (approved):
This looks great! Excited to see if we can nab some of these persistent bookmark issues! Had one comment about exists vs max but don't feel strongly if we don't want to go that route.

lougeniaC64 (approved):
LGTM!

components/places/src/bookmark_sync/engine.rs
mhammond: yeah, that's a good idea. I took the first one because we don't really need to know counts.

mhammond:

mhammond:
@mergify backport release-v152

PR #7432, opened 2026-06-18
feat(fxa-client): use typed-Bearer scheme instead of Hawk for session token

vbudhram:
This migrates session-token authentication from Hawk to the auth-server's typed-Bearer scheme: requests now send Authorization: `Bearer fxs_<token_id>` instead of a Hawk signature. BearerRequestBuilder replaces HawkRequestBuilder across all session-token calld, the hawk feature is dropped from rc_crypto, and the unused HawkError variant is removed. 

The format matches the auth-server's authentication-schemes reference (https://mozilla.github.io/ecosystem-platform/reference/authentication-schemes); production routes accept both schemes, so this is internal-only with no consumer-facing API impact.

▎ Note: Update the #PRNUM placeholder in CHANGELOG.md with this PR's number before merge.

Pull Request checklist

- Breaking changes:  This PR follows our breaking change policy (https://github.com/mozilla/application-services/blob/main/docs/howtos/breaking-changes.md)
  
- [x] Quality: This PR builds and tests run cleanly
- [x] Tests: This PR includes thorough tests or an explanation of why it does not
  - Adds four unit tests for BearerRequestBuilder covering the Bearer fxs_<token_id> header format
- [x] Changelog: This PR includes a changelog entry in CHANGELOG.md (https://github.com/mozilla/application-services/blob/main/CHANGELOG.md) or an explanation of why it does not need one
  - Added under v154.0 → Firefox Accounts → "What's Changed".
- [x] Dependencies: This PR follows our dependency management guidelines (https://github.com/mozilla/application-services/blob/main/docs/dependency-management.md)
  - No new dependencies. This PR removes the hawk feature from the rc_crypto dependency.

### DIFF
--- a/components/fxa-client/src/internal/http_client.rs
+++ b/components/fxa-client/src/internal/http_client.rs
@@ -12,11 +12,7 @@ use super::{config::Config, util};
 use crate::{trace, Error, Result};
 use error_support::breadcrumb;
 use parking_lot::Mutex;
-use rc_crypto::{
-    digest,
-    hawk::{Credentials, Key, PayloadHasher, RequestBuilder, SHA256},
-    hkdf, hmac,
-};
+use rc_crypto::{digest, hkdf, hmac};
 use serde_derive::{Deserialize, Serialize};
 use serde_json::json;
 use std::{
@@ -28,8 +24,12 @@ use sync15::DeviceType;
 use url::Url;
 use viaduct::{header_names, status_codes, Method, Request, Response};
 
-const HAWK_HKDF_SALT: [u8; 32] = [0b0; 32];
-const HAWK_KEY_LENGTH: usize = 32;
+const AUTH_HKDF_SALT: [u8; 32] = [0b0; 32];
+const AUTH_KEY_LENGTH: usize = 32;
+// Auth-server typed-Bearer prefix for sessionToken (matches KIND_PREFIXES in
+// the auth-server's bearer-fxa-token scheme). See:
+// https://mozilla.github.io/ecosystem-platform/reference/authentication-schemes
+const BEARER_PREFIX_SESSION_TOKEN: &str = "fxs";
 const RETRY_AFTER_DEFAULT_SECONDS: u64 = 10;
 // Devices older than this many days will not appear in the devices list
 const DEVICES_FILTER_DAYS: u64 = 21;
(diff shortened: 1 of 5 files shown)


bendk (dismissed):
The code change looks fine to me.  My one question is if there are could be any issues with clients that are already logged in, but I can't see how there would be.  I see you modified fxa-client, did you test logging in from the current main, moving to this commit,  then running the `attached-clients` command?

The crypto parts of this are above my head, but since you have an approved ADR I assume that's all worked out.

And yes, this will only affect mobile.  You'll need to do something similar with desktop to get the change there.

components/fxa-client/src/internal/http_client.rs
```diff
@@ -633,21 +641,22 @@ pub struct AuthorizationRequestParameters {
...
     body: Option<String>,
     hkdf_sha256_key: &'a [u8],
+    prefix: &'static str,
 }
 
-impl<'a> HawkRequestBuilder<'a> {
-    pub fn new(method: Method, url: Url, hkdf_sha256_key: &'a [u8]) -> Self {
-        rc_crypto::ensure_initialized();
-        HawkRequestBuilder {
+impl<'a> BearerRequestBuilder<'a> {
```
mhammond, line 651: afact, every user of this passes `BEARER_PREFIX_SESSION_TOKEN` except for one test. Is there a future where this might be different? And every single user of the builder is calling `derive_auth_key_from_session_token()`.  The few callers who don't use a session token then end up calling `bearer_token()` and manually setting up the header themselves.

I wonder if maybe we should replace `derive_auth_key_from_session_token()` with, say, `derive_bearer_token_from_session_token()`, which returns the complete header with the prefix already applied. The existing key param just gets renamed as it is now the full bearer token ready to use as-is You could also imagine a `derive_bearer_token_from_refresh_token()` and those requests could maybe start using this constructor too.

mhammond (commented):
this looks fine to me. As Ben mentioned, exercising that example would be useful too.

mhammond:
I also don't see how this will break for existing users - I don't think we are changing the format of the session token or anything like that, just making a small change to how it's used with the auth header - right? But I agree it's worth testing.

components/fxa-client/src/internal/http_client.rs
vbudhram: Updated to `derive_bearer_token_from_session_token` 👍🏽, I didn't make any changes to refresh token since this pr is just scoped to session tokens though.

vbudhram:
@bendk and @mhammond Thanks for the review! I just double checked and session token from main work in this branch and can return connected devices. The session token format has not changed, just how it is used in the auth header.

I also made recommened changes.

mhammond (approved):
This seems fine to me

PR #7440, opened 2026-06-22
Bug 2049279 - Add passport record type to autofill

DimiDL:
Add a passport autofill record type (name, country, passport number, and issue/expiry dates) with full CRUD, modeled on the existing addresses type.

The schema is laid out sync-ready: passports_data carries a sync_change_counter alongside passports_mirror / passports_tombstones and the sync staging tables, so a sync engine can be added later without a schema migration. No sync engine is registered yet, so passports are not synced. Fields follow the autofill convention - stored as plaintext, with empty/0 meaning "not provided".

  - Note:

### DIFF
--- a/components/autofill/sql/create_shared_schema.sql
+++ b/components/autofill/sql/create_shared_schema.sql
@@ -90,6 +90,43 @@ CREATE TABLE IF NOT EXISTS credit_cards_tombstones (
     time_deleted    INTEGER NOT NULL
 ) WITHOUT ROWID;
 
+-- Passport records. The passport number is stored as plaintext (no field-level
+-- encryption).
+CREATE TABLE IF NOT EXISTS passports_data (
+    guid                TEXT NOT NULL PRIMARY KEY CHECK(length(guid) != 0),
+    name                TEXT NOT NULL,              -- full name on passport
+    country             TEXT NOT NULL,              -- ISO 3166 code
+    passport_number     TEXT NOT NULL,
+    issue_date_month    INTEGER,
+    issue_date_day      INTEGER,
+    issue_date_year     INTEGER,
+    expiry_date_month   INTEGER,
+    expiry_date_day     INTEGER,
+    expiry_date_year    INTEGER,
+
+    time_created        INTEGER NOT NULL,
+    time_last_used      INTEGER,
+    time_last_modified  INTEGER NOT NULL,
+    times_used          INTEGER NOT NULL,
+
+    /* Same "sync change counter" strategy used by other components. */
+    sync_change_counter INTEGER NOT NULL
+);
+
+-- What's on the server as the JSON payload.
+CREATE TABLE IF NOT EXISTS passports_mirror (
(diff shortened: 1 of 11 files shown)


components/autofill/src/db/schema.rs
```diff
@@ -238,6 +271,12 @@ fn upgrade_from_v3(db: &Connection) -> Result<()> {
     Ok(())
 }
 
+fn upgrade_from_v4(db: &Connection) -> Result<()> {
+    let migration_string: &str = include_str!("../../sql/migrations/v5_migration.sql");
```
mhammond: Do we need this migration, or can we just re-execute create_shared_schema? That would avoid needing the test for them drifting too. I'm not sure why the others weren't done that way, but that's the approach most other components use

components/autofill/sql/create_shared_schema.sql
```diff
@@ -90,6 +90,43 @@ CREATE TABLE IF NOT EXISTS credit_cards_tombstones (
...
+    -- The plain-text sync15 payload.
+    payload             TEXT NOT NULL CHECK(length(payload) != 0)
+);
+
+-- Tombstones are items deleted locally but not deleted in the mirror (ie, ones
+-- we are yet to upload)
+CREATE TABLE IF NOT EXISTS passports_tombstones (
+    guid            TEXT PRIMARY KEY CHECK(length(guid) != 0),
+    time_deleted    INTEGER NOT NULL
+) WITHOUT ROWID;
```
mhammond, line 128: you don't have the shared triggers for this yet. That seems fine I guess until you sync though.

components/autofill/sql/create_sync_temp_tables.sql
```diff
@@ -27,3 +27,16 @@ CREATE TEMP TABLE credit_cards_sync_outgoing_staging (
     payload             TEXT NOT NULL CHECK(length(payload) != 0),
     sync_change_counter INTEGER NOT NULL
 );
+
+DROP TABLE IF EXISTS passports_sync_staging;
```
mhammond: because these are temp tables you don't actually need them yet - they can be added later without needing a migration. I guess there's no harm adding them though, but also no real benefit.

mhammond (commented):
This looks like I expected it to :)

components/autofill/sql/create_shared_schema.sql
DimiDL: I'd defer the shared triggers until implementing sync, to keep this patch simpler. I'm keeping the tombstones table in this patch so we don't need a migration when sync is implemented.

components/autofill/src/db/schema.rs
DimiDL: Nice catch, thanks! I updated the patch!

components/autofill/sql/create_sync_temp_tables.sql
DimiDL: ok, removed

mhammond (approved):
👍

PR #7442, opened 2026-06-22
feat(ads-client): roundtrip cap key for spoc impression callbacks (AC-125)

jonesetc:
Add roundtripping of `cap_key` in sposored content impression callback urls. This value is currently just discarded, but in the future will be used to keep a rolling count of impressions in the last 24 hours on device to make sure sponsored content is not being shown too often.

This does not change any internal or external APIs, and does not require an entry in the changelog

  - Note:

### DIFF
--- a/components/ads-client/src/client.rs
+++ b/components/ads-client/src/client.rs
@@ -121,7 +121,7 @@ where
 
     pub fn record_impression(
         &self,
-        impression_url: Url,
+        mut impression_url: Url,
         ohttp: bool,
     ) -> Result<(), RecordImpressionError> {
         // TODO: Re-enable cache invalidation behind a Nimbus experiment.
@@ -130,6 +130,26 @@ where
         // if let Some(request_hash) = pop_request_hash_from_url(&mut impression_url) {
         //     let _ = self.client.invalidate_cache_by_hash(&request_hash);
         // }
+
+        // TODO: Add count call with _cap_key for impression capping logic
+        if let Some((_, _cap_key)) = impression_url
+            .query_pairs()
+            .find(|(key, _)| key == "cap_key")
+        {
+            let original_url = impression_url.clone();
+            impression_url
+                .query_pairs_mut()
+                .clear()
+                .extend_pairs(
+                    original_url
+                        .query_pairs()
+                        .collect::<Vec<_>>()
+                        .iter()
+                        .filter(|(key, _)| key != "cap_key"),
+                )
+                .finish();
+        }
+
         self.client
             .record_impression(impression_url, ohttp)
             .inspect_err(|e| {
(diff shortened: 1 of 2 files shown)


components/ads-client/src/mars/ad_response.rs
```diff
@@ -173,6 +178,10 @@ impl AdResponseValue for AdSpoc {
     fn callbacks_mut(&mut self) -> &mut AdCallbacks {
         &mut self.callbacks
     }
+
+    fn impression_enriching_params(&self) -> Vec<(String, String)> {
+        vec![("cap_key".to_owned(), self.caps.cap_key.clone())]
+    }
```
Almaju, line 184: I think you can do `fn impression_enriching_params(&self) -> [(&str, &str); 1]` to have zero allocation (no Vec, no String)

components/ads-client/src/client.rs
```diff
@@ -130,6 +130,26 @@ where
         // if let Some(request_hash) = pop_request_hash_from_url(&mut impression_url) {
         //     let _ = self.client.invalidate_cache_by_hash(&request_hash);
         // }
+
+        // TODO: Add count call with _cap_key for impression capping logic
+        if let Some((_, _cap_key)) = impression_url
+            .query_pairs()
+            .find(|(key, _)| key == "cap_key")
+        {
+            let original_url = impression_url.clone();
```
Almaju: I don't think we need `mut` if we are cloning the url anyway?

You could remove the mut and do something like:
```rs
let impression_url =  if let Some((_, _cap_key)) { 
  ... // clone here then mut 
} else { 
  impression_url 
};
```

components/ads-client/src/mars/ad_response.rs
```diff
@@ -55,7 +56,8 @@ impl<A: AdResponseValue> AdResponse<A> {
                 callbacks
                     .impression
                     .query_pairs_mut()
-                    .append_pair("request_hash", &hash_str);
+                    .append_pair("request_hash", &hash_str)
+                    .extend_pairs(impression_enriching_params);
```
Almaju: Since we already have the `request_hash` acting as a unique identifier up top, could storing it in the SQLite table "impression_log" be a cleaner alternative here?

Almaju (changes requested):

components/ads-client/src/mars/ad_response.rs
jonesetc: A request hash maps to an an ad response, each of which contains 0-* spocs. Each spoc has its own cap key. So going just from the request hash to a single cap key doesn't work.

A much messier thing that could be done would be to just map the impression url itself to the block key, but that is assuming that the mapping between spoc:impression url:cap key is always 1:1:1. Actually getting back to the cap key directly makes the fewest assumptions and allows the cap keys to do more complex things from the server like "all of these spocs have cap key A, and these have cap key B" to have group limits instead of just per individual spoc.

components/ads-client/src/client.rs
jonesetc: Little bit of an out of it with rust moment, I forgot `let impression_url =  ...` and `impression_url =  ...` are different here and you are actually free to shadow the name of an argument without making it mutable!

components/ads-client/src/mars/ad_response.rs
jonesetc: The problem with the `&str` (or any reference returned) in there is that this result is used during the mutable borrow of `callbacks_mut`, so those would have to be cloned anyway. I think returning something like `Option<IntoIterator<(String, String)>>` should be possible and avoid allocation in the non-spoc case and just a single small array in the spoc case.

jonesetc: I just went ahead and simplified it instead of worrying about iterators or refs. Now there is just a method that returns an optional cap_key, by default None, and spoc returns a string. The string would need to be owned eventually like i mentioned because of the mutable borrow for callbacks. so I figured just returning the string made more sense than making the caller do something like:

```rust
let cap_key = ad.cap_key().map(String::from)
```

Almaju (approved):
Thanks for the changes, lgtm!
We might be able refactor this later with async model because we will need some `AdStore` that can hold stateful data such as request hash or cap key but this looks great for now and the current sync stateless model we have!

PR #7449, opened 2026-06-24
Bug 2049856 - Remove old viaduct backend code

bendk:
Changed the default timeout for `viaduct::Client` to 7 seconds for iOS and 10 seconds otherwise.  This matches the behavior of we had for `Request::send()`.  Should we make these longer?

Also removed viaduct-reqwest which was no longer used

  - Note:


components/viaduct/src/client.rs
```diff
@@ -106,7 +103,10 @@ impl ClientSettings {
 impl Default for ClientSettings {
     fn default() -> Self {
         Self {
-            timeout: 60000,
+            #[cfg(target_os = "ios")]
+            timeout: 7000,
+            #[cfg(not(target_os = "ios"))]
+            timeout: 10000,
```
bendk, line 109: Does this seem right for default timeouts?  I really don't want to mess them up again as part of this process.

I believe these match the defaults for the old backend and so this shouldn't change the behavior for anyone.  This does lower the default timeout for the few consumers using the newer so I tried to find all uses of it and set the timeout to 60s for them.

bendk:
I can't figure out these weird linker errors.  Going to take this down and try again tomorrow.

components/viaduct/src/client.rs
mhammond: I don't think these make sense - I've lost some of the history, but I don't think the very first cuts of viaduct had such short timeouts. The problem is that none of us will ever see a problem as we are on good networks, but things just randomly break in the field - ie, the 10s sync timeout was introduced (I'm quite sure we never started with that timeout), but it took some time for us to finally notice in user-generated bug reports that it was a problem.

I'd rather invert this - have sane defaults and individual clients who really do need insanely short timeouts get to override?

PR #7450, opened 2026-06-25
Bug 2049856 - Remove old viaduct backend code [ci full]

bendk:
* Removed the old `backend` and `backend::ffi` modules
* Renamed `new_backend` to `backend`.

Changed the default timeout for `viaduct::Client` to 7 seconds for iOS and 10 seconds otherwise.  This matches the behavior of we had for `Request::send()`.  Should we make these longer?

Also: removed viaduct-reqwest which was no longer used.  Removed the `rusqlite::bundled` feature from `ads-client`.  It's not set by other crates and this fixed a build issue for me.

  - Note:


components/viaduct/src/client.rs
```diff
@@ -106,7 +103,10 @@ impl ClientSettings {
 impl Default for ClientSettings {
     fn default() -> Self {
         Self {
-            timeout: 60000,
+            #[cfg(target_os = "ios")]
+            timeout: 7000,
+            #[cfg(not(target_os = "ios"))]
+            timeout: 10000,
```
bendk: From markh in an earlier version: 
> I don't think these make sense - I've lost some of the history, but I don't think the very first cuts of viaduct had such short timeouts. The problem is that none of us will ever see a problem as we are on good networks, but things just randomly break in the field - ie, the 10s sync timeout was introduced (I'm quite sure we never started with that timeout), but it took some time for us to finally notice in user-generated bug reports that it was a problem.
> I'd rather invert this - have sane defaults and individual clients who really do need insanely short timeouts get to override?

bendk: Thanks for catching this, I bumped the timeout back up to 60s.

I got confused because this was the defaults for the old backend that I was removing: https://github.com/mozilla/application-services/pull/7449/changes#diff-f05535f2df59d33408a4bec415bf621aa648501e7f6f5279424e7d74684636c7L39.  However, I guess that wasn't actually taking effect because all platforms are using the new backend now, which is why we can open this PR in the first place.

The thing I still don't understand why we weren't seeing timeouts before we switched to the new backend.  With the old backend the defaults did seem to be 7s and 10s and I don't see how they were overridden.

mhammond: I'm not sure, I've lost context on when we switched exactly - but the dashboards certainly show timeout errors dropping significantly since that change.

components/merino/src/suggest/http.rs
```diff
@@ -70,7 +70,11 @@ impl HttpClientTrait for HttpClient {
     ) -> Result<Option<Response>> {
         let url = build_suggest_url(query, &options, endpoint_url);
 
-        let client = Client::with_ohttp_channel("merino", ClientSettings::default())?;
+        let settings = ClientSettings {
```
mhammond: it looks like `ClientSettings::default()` will use a timeout of 60000, so these overrides don't seem necessary. (It also looks like uniffi will supply different defaults, which seems a little like a footgun - is it worth a followup?)

I'm sorry if my other previous comments around this caused confusion?)

components/viaduct/android/build.gradle
```diff
@@ -1,40 +1,11 @@
-plugins {
```
mhammond, line 1: woohoo - it looks like this patch means `prost` can be dropped as a dependency from this crate, meaning `prost` gets removed from the repo - so we are officially protobuf free - iirc, @rvandermeulen will be excited by this (but I can't remember exactly why 😅 )

mhammond (approved):
Great to see this finally happen! Probably worth an announcement on #app-services-engineering once you land this as a heads up!

components/merino/src/suggest/http.rs
bendk: No, I think I was just confused and juggling two backends adds confusion.  I removed this code and updated the UniFFI default.

components/viaduct/android/build.gradle
bendk: At first I thought this was just some cleanup, but halfway through I realized we could finally drop the protobuf dependency.  :champagne: 

```
 > cargo tree -i prost
error: package ID specification `prost` did not match any packages
```

PR #7451, opened 2026-06-25
Bug 2050097 - Add AI engine classification

mandysGit:

  - Note:

### DIFF
--- a/components/search/src/configuration_types.rs
+++ b/components/search/src/configuration_types.rs
@@ -110,8 +110,7 @@ pub(crate) struct JSONEngineBase {
     pub charset: Option<String>,
 
     /// The classification of search engine according to the main search types
-    /// (e.g. general, shopping, travel, dictionary). Currently, only marking as
-    /// a general search engine is supported.
+    /// (e.g. general, ai, and unknown).
     #[serde(default)]
     pub classification: SearchEngineClassification,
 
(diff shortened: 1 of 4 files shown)


components/search/src/configuration_types.rs
```diff
@@ -110,8 +110,7 @@ pub(crate) struct JSONEngineBase {
     pub charset: Option<String>,
 
     /// The classification of search engine according to the main search types
-    /// (e.g. general, shopping, travel, dictionary). Currently, only marking as
-    /// a general search engine is supported.
+    /// (e.g. general, ai, and unknown).
```
Standard8, line 113: The types that I put in previously were intended as examples, rather than exact definitions. I don't think we want to put ourselves in a situation of having to update documentation that's defined elsewhere, so how about something like:

```rust
/// The classification of search engine according to the main search types
/// (e.g. general, shopping, travel, dictionary; for exact definitions see `SearchEngineClassification`).
```

Standard8 (changes requested):
As discussed elsewhere, the fix here is correct, however due to our lack of fallback (`#[serde(other)]`), this won't work for existing clients that have already been shipped.

I think we should delay this until we need search-config-v3. In the meantime the relevant engine could be switched to "unknown".

Setting to "request changes" to clear from review queue.

PR #7457, opened 2026-06-30
Bug 1868418 - Rerun ensure capabilities after logging in

bendk:
Moved the code to RetryingAccount, which should give us this functionality for free.

  - Note:


bendk:
I think I got the state machine stuff right, but please double-check my work here.

components/fxa-client/src/state_machine/transitions.rs
```diff
@@ -27,20 +26,8 @@ pub fn transition(
...
-                        Err(e) if is_auth_error(&e) => {
-                            // Report inline since we don't propagate this error to the driver.
-                            let _: FxaError = convert_log_report_error(e);
-                            let active = account
-                                .check_authorization_status()
-                                .to_state_machine_err(|| S::AuthIssues)?;
-                            Ok(if active { S::Connected } else { S::AuthIssues })
-                        }
+                    match account.finish_initialize(&device_config.capabilities) {
+                        Ok(()) => Ok(S::Connected),
```
skhamis, line 30: One thing worth checking: in the old code we land at S::AuthIssues when `check_authorization_status` comes back `false`. The new code propagates the original FxaError::Authentication to the Err(cause) arm, which sends us to `S::Disconnected`?. For someone who is signed-in and whose refresh token is server-invalidated (password-changed elsewhere), I think our AuthIssues preserves the account state and lets consumers show the "re-auth" flow. Disconnected will put them back to full sign-in right? 
Should we do something like:
```      
 match account.finish_initialize(&device_config.capabilities) {
                         Ok(()) => Ok(S::Connected),
                        Err(cause) => {
                            let target = if matches!(cause.get_error_handling().err, FxaError::Authentication) {
                                S::AuthIssues
                            } else {
                                S::Disconnected
                            };
                            Err(StateMachineErr::new(cause, target))
                        }
                    }
```
or something similar? However if i'm missing something here please let me know!

skhamis (changes requested):
Nice fix! I think we're close but just one question about the error part

components/fxa-client/src/state_machine/transitions.rs
bendk: That seems like an improvement, but maybe it should always be `S::AuthIssues`.  I think that's what the current code does and it seems more correct to me.   WDYT?

skhamis (approved):
Thanks for making that change. I agree with the new direction and think its "more correct". LGTM!

bendk:
Dependency checks seem unrelated to this PR, I didn't change `Cargo.lock` at all.  I'm going to hit the merge button anyways.

PR #7459, opened 2026-07-01
fxa-client: add has_scope() method.

mhammond:

  - Note:

### DIFF
--- a/components/fxa-client/src/internal/oauth.rs
+++ b/components/fxa-client/src/internal/oauth.rs
@@ -26,6 +26,18 @@ use url::Url;
 pub const OAUTH_WEBCHANNEL_REDIRECT: &str = "urn:ietf:wg:oauth:2.0:oob:oauth-redirect-webchannel";
 
 impl FirefoxAccount {
+    /// Check whether every requested scope has been granted to the account's refresh token.
+    pub fn has_scope(&self, scope: &str) -> bool {
+        let mut requested = scope.split_ascii_whitespace().peekable();
+        if requested.peek().is_none() {
+            return false;
+        }
+        match self.state.refresh_token() {
+            Some(refresh_token) => requested.all(|s| refresh_token.scopes.contains(s)),
+            None => false,
+        }
+    }
+
     /// Extracts and stores the session token from a WebChannel login JSON payload.
     /// The JSON payload is the `data` object from the `fxaccounts:login` WebChannel command.
     pub fn handle_web_channel_login(&mut self, json_payload: &str) -> Result<()> {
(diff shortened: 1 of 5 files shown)


components/fxa-client/src/internal/oauth.rs
```diff
@@ -651,6 +663,24 @@ mod tests {
...
+        let mut fxa =
+            FirefoxAccount::with_config(Config::stable_dev("12345678", "https://foo.bar"));
+        // No refresh token -> false.
+        assert!(!fxa.has_scope("profile"));
+        fxa.state.force_refresh_token(RefreshToken {
+            token: "rt".to_owned(),
+            scopes: ["profile", "sync"].iter().map(|s| s.to_string()).collect(),
+        });
+        assert!(fxa.has_scope("profile"));
+        assert!(fxa.has_scope("sync profile"));
```
jonalmeida, line 678: Is this an accurate expectation? In the persisted state, we have an array of individual scopes, so a space-delimited list of scopes wouldn't work there.

Image

Our access tokens are based on the grouped scopes, so that tracks if we were looking at that.

jonalmeida (approved):
Ha, nice thank you for this! I left a question on usage expectations, but it's a non-blocker.

components/fxa-client/src/internal/oauth.rs
mhammond: I don't fully understand your question. The API using space sep'd scopes is odd, but I went with that to match how `get_access_token()` works - the impl splits it. So roughly, this is saying "if I pass this exact string of scopes to `get_access_token()`, is it going to think the refresh token can handle this without an exchange".

So my expectation is that passing all of ".../oldsync profile", "profile .../oldsync" and "profile" would return true for the token in your screenshot.

I'm open to changing the api to be a vec, and that's probably what should have been done earlier - but IMO the consistency with `get_access_token()` seems the lesser of 2 evils. Or maybe that's not what you are referring to?

jonalmeida: Why does the scopes need to be a vec or a space-delimited list? Aren't we checking the scopes against the refresh token which checks that in the vec of scopes?

I guess I'm [not] concerned with how it's implemented as long as `has_scope("A") && has_scope("B") == has_scope("A B")` is still true.

EDIT: added negation.

mhammond: Right, I see. But wouldn't taking a string which *does not* support multiple scopes make the api even more confusing than it already is? That would break `has_scope("A") && has_scope("B") == has_scope("A B")` because `has_scope("A B")` would return false.

jonalmeida: I think you've confirmed that my expectation on how the `has_scope` works is accurate and I probably confused you along the way - sorry! If we're looking at the same data source that we currently have I'd be happy to replace the current code with this.

PR #7460, opened 2026-07-01
Don't package the glean_sym files

badboy:
Not sure that's really the best solution.
Right now I manually removed that file in https://github.com/mozilla-mobile/firefox-ios/pull/34478
I could also remove it in the script in fx-ios that updates appservices, but that feels even more wrong.


bendk (approved):
Seems like a reasonable way to handle this to me.

PR #7467, opened 2026-07-09
Bug 2053557 - remove all logins without decrypting

jo:
https://bugzilla.mozilla.org/show_bug.cgi?id=2053557

  - Note:

### DIFF
--- a/components/logins/src/db.rs
+++ b/components/logins/src/db.rs
@@ -874,6 +874,33 @@ impl LoginDb {
         Ok(results.pop().expect("there should be a single result"))
     }
 
+    // Delete all records. Return an array with the ids of the deleted logins
+    pub fn delete_all(&self) -> Result<Vec<String>> {
+        let ids: Vec<String> = self.db.query_rows_and_then_cached(
+            "SELECT guid FROM loginsL WHERE is_deleted = 0
+             UNION ALL
+             SELECT guid FROM loginsM WHERE is_overridden = 0",
+            [],
+            |row| row.get(0),
+        )?;
+        self.delete_many(ids.iter().map(String::as_str).collect())?;
+        Ok(ids)
+    }
+
+    // Delete all records, except the FxA login. Return an array with the ids of
+    // the deleted logins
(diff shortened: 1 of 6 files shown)


components/logins/src/db.rs
```diff
@@ -874,6 +874,33 @@ impl LoginDb {
         Ok(results.pop().expect("there should be a single result"))
     }
 
+    // Delete all records. Return an array with the ids of the deleted logins
+    pub fn delete_all(&self) -> Result<Vec<String>> {
+        let ids: Vec<String> = self.db.query_rows_and_then_cached(
+            "SELECT guid FROM loginsL WHERE is_deleted = 0
+             UNION ALL
+             SELECT guid FROM loginsM WHERE is_overridden = 0",
```
bendk, line 882: Why do you need the second query?  I would have thought the `loginsL` one would be enough.  I'm guessing I'm missing something, could you add a comment explaining.  The same applies to `delete_all_except_fxa`.

bendk (approved):
This looks good to me, I just had a question about one of the queries

components/logins/src/db.rs
jo: This is a very good question which I have to admin I cannot answer. I looked at the `get_all` [query](https://github.com/mozilla/application-services/blob/main/components/logins/src/db.rs#L1042-L1047) and replicated it for the needs here. Throughout the code we union against both tables. 🤔

jo: I think that's the way logins work: https://github.com/mozilla/application-services/blob/main/components/logins/src/db.rs#L7-L24
> When we want to fetch a record, we need to look in both loginsL and loginsM for the data.

bendk: Ahh, so if a record isn't updated then there won't be a row in `loginsL`.  I always forget that and picture it working the opposite way.  This all makes sense and seems good to merge to me.

mhammond: yeah, this bites me regularly too. It's "clever", but borders on "too clever" 😅

PR #7468, opened 2026-07-10
Bug 2053207 - Use glean-sym for places telemetry [ci full]

bendk:
Use glean-sym for real, rather than just for test metrics. Added support for Desktop/FOG. Removed the metrics meant to test glean-sym.

Simplified `run_maintenance` now that we can record metrics in Rust.

Updated chrono to `0.4.45` for compatibility with `glean`

  - Note:


components/places/android/metrics.yaml
```diff
@@ -2,124 +2,8 @@
...
-# specific to the *syncing* of places are defined in a separate "sync_ping"
-# package.
-#
-# Changes to these metrics require data review, which should take into
-# consideration
-# the following known consumers of the places component Android bindings:
-#
-#  * Fenix for Android
+# This file defines the metrics that are recorded by android for the places component.
+# It's no longer used, but we still need to keep it around for probe-scraper reasons.
```
bendk, line 6: I believe this is true, is there a better way to deal with this?  If not, it's not that bad and I believe the issue goes away when we move into moz-central?

components/places/Cargo.toml
```diff
@@ -34,7 +36,8 @@ sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random"
 thiserror = "2"
 anyhow = "1.0"
 uniffi = { version = "0.31" }
-glean-sym = { git = "https://github.com/mozilla/glean", tag = "v68.0.0", optional = true }
+glean = { version = "68.0.1", optional = true }
+glean-sym = { git = "https://github.com/mozilla/glean", tag = "v68.0.1", optional = true }
```
bendk: Would it be worth it to publish `glean-sym` so that we get all the dependencies from the same source?

components/places/src/ffi.rs
```diff
@@ -384,55 +380,55 @@ impl PlacesConnection {
...
-        places_manager::run_maintenance_prune_time_temp.stop_and_accumulate(timer_id);
+        Ok(())
+    }
 
+    fn run_maintenance_prune(&self, db_size_limit: u32, prune_limit: u32) -> Result<()> {
+        #[cfg(any(feature = "glean-sym", feature = "glean-fog"))]
+        let timer_id = glean_metrics::places_manager::run_maintenance_prune_time.start();
+        let res =
+            self.with_conn(|conn| storage::run_maintenance_prune(conn, db_size_limit, prune_limit));
+        #[cfg(any(feature = "glean-sym", feature = "glean-fog"))]
```
bendk: Having to feature flag everything like this isn't great.  One idea I had would is adding support for a `rust-noop` format to `glean_parser`.  That would generate the same code as `rust`/`rust-sym`, but with the function bodies wouldn't do anything.  @badboy what do you think about this?  I could try my hand at implementing it if the glean team was busy with other things.

bendk:
This seems to be working and I even did a quick check that the FOG stuff works on desktop.  However, I'm not sure about a couple things and I left comments for them.

bendk:
The tests are failing because pip3 can't find the right glean_parser version.  I'm not sure what's happening there, any ideas?

components/places/Cargo.toml
badboy: yes! https://bugzilla.mozilla.org/show_bug.cgi?id=2054472

components/places/src/ffi.rs
badboy: Basically https://bugzilla.mozilla.org/show_bug.cgi?id=1754020
We can probably generate that part, now that we have something to generate glean-sym, taking that and generating a no-op API would be easy.

badboy: Though would you even need this here, now that it's going to support all platforms it should not need to be behind a flag?

bendk: Yes, but the dependencies get weird.  On Android/iOS we depend on `glean-sym` and on Desktop we depend on `glean`.  So what should happen when you run `cargo build` or `cargo test` from the crate?

- We could add `glean` as a non-optional dependency and that would make the build/tests work.  However, that means that Android/iOS get that dependency when they don't need it.
- We could not depend on any glean crates by default and require the user to pass in one of the `glean-*` feature flags.  However, that means builds without a feature flag would fail.
- We could put the metric-generation code behind feature flags, which is what I did for this PR.  It's not terrible, but I don't love it either.

badboy: Ah, ok. Yeah, that seems like a use case for a glean-nop

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=2055209

bendk: Nice, I think that will work great.

badboy:
> The tests are failing because pip3 can't find the right glean_parser version. I'm not sure what's happening there, any ideas?

```
❯ docker run -it --rm --platform linux/amd64 cimg/rust:1.53.0 python3 --version
Python 3.8.5
```

Pretty sure that's because the image in use is way too old to have a recent Python

Edit: #7505

bendk:
Pushing a new version of this based on all the nice Glean changes in `main` (thanks Jan-Erik).  Do not merge this until these are in a Glean release and we've updated the Cargo.toml files to point to that release.

components/places/Cargo.toml
```diff
@@ -34,16 +34,15 @@ sync-guid = { path = "../support/guid", features = ["rusqlite_support", "random"
 thiserror = "2"
 anyhow = "1.0"
 uniffi = { version = "0.31" }
-
-# glean-sym is only used on Android and iOS.
-[target.'cfg(any(target_os = "android", target_os = "ios"))'.dependencies]
-glean-sym = { git = "https://github.com/mozilla/glean", tag = "v68.0.0", optional = true }
+glean = { git = "https://github.com/mozilla/glean", branch = "main", optional = true }
+glean-sym = { git = "https://github.com/mozilla/glean", branch = "main" }
```
bendk: With the introduction of the `glean-sym/noop` feature this gets much nicer:

- For Android, we build without any feature flags and use `glean-sym` for the metrics
- For Desktop, we build with the `glean-fog` feature flag and use that for the metrics. TODO: double check that the APIs are identical and everything compiles cleanly.  
- For testing, we build with the `glean-sym/noop` feature.

This means that can always just record metrics and never need `#[cfg]` attributes to conditionally enable them.

The main downside I see is that it's a bit weird for the "default" build, i.e. if I just run `cargo build -p places`.  In that case, we use `glean-sym` and everything will fail unless it ends up linked to another library like `libxul` that provides the symbols.  This is because we can't have a feature flag that disables `glean-sym/noop`, only one that enables it.  I don't think this is that big of a deal, but I wonder if we could swap the way `glean-sym` works so that it's `noop` by default and there's a feature flag to enable the metrics recording.  Alternatively, maybe `glean-sym` could always go into no-op mode if the library failed to load.

The other downside is that we're depending on `glean-sym` for the Desktop case, when we don't really need it.  AFAIK, `glean-sym` is very lightweight though so I don't care about this at all.

badboy: "Disable it when library couldn't be loaded" -- I initially tried writing the code that way but it was very unwieldy (and that was before we thought about a "real noop" mode). It either requires null-routing all the function pointers or constant checks if function pointers are non-null. Not impossible, but annoying to put together.

No-op by default however seems possible I guess. Cargo features are indeed not perfect for toggling between these modes, so I'm sure we can also find problems with that, so I need to think about it.

badboy: filed https://bugzilla.mozilla.org/show_bug.cgi?id=2062178 for tracking

badboy: filed https://bugzilla.mozilla.org/show_bug.cgi?id=2062178 for tracking

bendk: With glean v70 this is much nicer.  `glean-sym` is always enabled, but by default it doesn't do anything.  `megazord` and `megazord-ios` enable the `glean-sym/active` feature and that turns on the actual recording.

bendk:
I think this could be ready to merge once moz-central has updated to glean v70.

bendk:
Glean 70 is in.  I rebased this to the current main.  Let's wait a bit just so that the current Glean changes have some time to settle, but I plan to merge this soon.

bendk:
Actually, I'm going to hold this back until we figure out the perf issue in https://bugzilla.mozilla.org/show_bug.cgi?id=2047566

components/places/Cargo.toml
```diff
@@ -7,9 +7,9 @@ license = "MPL-2.0"
 exclude = ["/android", "/ios"]
 
 [features]
-default = ["glean-sym"]
-# Enable metric recording through glean-sym (Android only)
-glean-sym = ["dep:glean-sym"]
+# Use Firefox-on-Glean to publish metrics (Desktop)
+# Note: if set, we will not use `glean-sym`
+glean-fog = ["dep:glean"]
```
badboy, line 12: Is this actually tested in FOG?

bendk: Not really.  I believe I tried doing a local build with this change vendored in and the Rust code worked, but the ping submission failed because of probe-scraper issues.  That was a while ago though and the code has changed since then.  I also might just be misremembering this.

I left the code in since I figure it's a good starting point, but it shouldn't have any affect since neither `glean-sym` nor `glean-fog` is enabled on desktop.

badboy: Alright, as long as we're not accidentally enabling it right now, this seems good to go.

lougeniaC64 (approved):
A bit of a rubber stamp but LGTM

PR #7474, opened 2026-07-14
Bug 2054982 - Finer grained locking for RemoteSettingService

bendk:
Split up `RemoteSettingServiceInner` into multiple structs, each with their own Mutex and add a bit of policy around locking them.  Right now this is just enforced by putting all the methods that lock them in one impl block.

The issue we're trying to avoid is having `RemoteSettingService::sync` block `RemoteSettingService::make_client`.

Also, start using `viaduct::Client` which should be possible now since all applications are using the new backend.

  - Note:

### DIFF
--- a/components/remote_settings/src/service.rs
+++ b/components/remote_settings/src/service.rs
@@ -7,12 +7,12 @@ use std::{
     sync::{Arc, Weak},
 };
 
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use error_support::trace;
 use parking_lot::Mutex;
 use serde::Deserialize;
 use url::Url;
-use viaduct::Request;
+use viaduct::{Client, ClientSettings, Request};
 
 use crate::{
     client::RemoteState, config::BaseUrl, error::Error, storage::Storage,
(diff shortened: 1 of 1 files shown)


components/remote_settings/src/service.rs
```diff
@@ -53,39 +62,38 @@ impl RemoteSettingsService {
...
-            inner.bucket_name.clone(),
+            config.base_url,
+            config.bucket_name,
             collection_name.clone(),
-            inner.app_context.clone(),
+            config.app_context.clone(),
             storage,
         ));
-        inner.clients.push(Arc::downgrade(&client));
+        self.push_client(Arc::downgrade(&client));
```
gruberb: Could `make_client()` race with `update_config()` here? `clone_config()` releases the config lock before this client is added to clients. If `update_config()` runs between these calls, `active_clients()` won’t include the new client, but the service config will be replaced. The client can then be registered with the old config and never receive the update.

Could config access and client registration/update share one short-lived mutex-protected state? `sync_client` could remain separate, so this would not reintroduce the original network-blocking issue.

gruberb (commented):
I think we have to(please double check) combine creating the config and registering the client into one operation not to get out of sync. We can keep `sync_client` separate not to be blocking.

```rust
struct RemoteSettingsService {
    storage_dir: Utf8PathBuf,

    // Short-lived lock: configuration and client registration.
    state: Mutex<RemoteSettingsServiceState>,

    // Long-lived lock: held across the monitoring network request.
    sync_client: Mutex<RemoteSettingsServiceSyncClient>,

    telemetry: Mutex<RemoteSettingsTelemetryWrapper>,
}

struct RemoteSettingsServiceState {
    config: RemoteSettingsServiceConfig,
    clients: Vec<Weak<RemoteSettingsClient>>,
}
```

components/remote_settings/src/service.rs
bendk: Oh yes it could indeed, nice catch.

bendk: I followed your suggestion and put everything in one mutex and changed `make_client` to hold the mutex for the entire operation (which should be quick).

gruberb (commented):
I am not 100% sure if that's the case, but I think the race technically still exists? The `update_config()` still acquires the lock twice? Could we have a helper function like

```rust
fn update_config_and_clients(&self, config: RemoteSettingsServiceConfig) {
    let mut state = self.client_state.lock();

    let active_clients = state.active_clients();
    for client in active_clients {
        client.internal.update_config(
            config.base_url.clone(),
            config.bucket_name.clone(),
            config.app_context.clone(),
        );
    }

    state.config = config;
}
```

gruberb (approved):

PR #7477, opened 2026-07-15
doc: Minor details to documentation for building A-S

thesuzerain:
This PR makes no changes aside from minor documentation additions for setup, so doesn't need changelog updates or any breaking changes (or changes at all for that matter).

  - Note:
- [X] **Changelog**: This PR includes a changelog entry in [CHANGELOG.md](https://github.com/mozilla/application-services/blob/main/CHANGELOG.md) or an explanation of why it does not need one

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -54,6 +54,12 @@ a number of hours to complete.
                ```shell
                PYPATH=$(which python3); ln -s $PYPATH `dirname $PYPATH`/python
                ```
+                1. Recent brew installations of Python do not contain a `python`/`pip` symlink. You may be able to run the following to create them:
+                    ```shell
+                    brew uninstall python3
+                    brew install python3
+                    ```
+               
         1. Install gyp:
             ```shell
             wget https://bootstrap.pypa.io/ez_setup.py -O - | python3 -
@@ -66,6 +72,14 @@ a number of hours to complete.
                export PATH="~/tools/gyp:$PATH"
                ```
             1. If you have additional questions, consult [this guide](https://github.com/mogemimi/pomdog/wiki/How-to-Install-GYP).
+            1. More recent version of Python (3.12+) have deprecated `distutils`, a dependency of the above script. If you encounter a ModuleNotFoundError, you can run first:
+                ```shell
+               brew install python-setuptools
+               ```
+            1. Similarly, recent brew installations may result in **externally-managed-environment** errors. This can be addressed with `brew install pipx` and replacing the last line with `pipx install .`
+
+
+
         1. Make sure your homebrew python's bin folder is on your path by updating your bash/zsh profile with the following:
             ```shell
             export PATH="$PATH:$(brew --prefix)/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/bin"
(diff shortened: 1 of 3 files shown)


docs/building.md
```diff
@@ -107,8 +121,13 @@ The following instructions assume that you are building `application-services` f
...
       1. Android SDK Command-line Tools (latest)
    1. Set environment variables based on the boostrapped code and the downloaded NDK. Add it to your rc file (either `.zshrc` or `.bashrc` depending on your shell) to make it permanent
-      1. Set `JAVA_HOME` to point to the bootstraped JDK 17 installation directory. Ex: `export JAVA_HOME=~/.mozbuild/jdk/jdk-17.0.18+8/`
-      1. Set `ANDROID_HOME` to the bootstraped Android SDK. Ex: `export ANDROID_HOME=~/.mozbuild/android-sdk-linux`
+      1. Set `JAVA_HOME` to point to the bootstraped JDK 17 installation directory. Ex: `export JAVA_HOME=~/.mozbuild/jdk/jdk-17.0.18+8/` (or `export JAVA_HOME=~/.mozbuild/jdk/jdk-17.0.18+8/Contents/Home` on Mac).
+      1. Set `ANDROID_HOME` and `ANDROID_SDK_ROOT` to the bootstraped Android SDK. As an example:
+            ```shell
+            export ANDROID_HOME=~/.mozbuild/android-sdk-linux
+            export ANDROID_SDK_ROOT=~/.mozbuild/android-sdk-linux
+            ```
```
thesuzerain, line 127: I'm not actually totally sure if the `ANDROID_SDK_ROOT` set is necessary here, but a later step [here](https://github.com/mozilla/application-services/blob/main/docs/howtos/locally-published-components-in-fenix.md) describes it being set here and it's not mentioned elsewhere on the page.

In the codebase I don't see any other uses of ANDROID_SDK_ROOT (aside from documentation or CI setting it) so I assume its deprecated/unused. If so, I can instead adjust the other step to remove the reference to this.

docs/building.md
```diff
@@ -54,6 +54,12 @@ a number of hours to complete.
                ```shell
                PYPATH=$(which python3); ln -s $PYPATH `dirname $PYPATH`/python
                ```
+                1. Recent brew installations of Python do not contain a `python`/`pip` symlink. You may be able to run the following to create them:
+                    ```shell
+                    brew uninstall python3
+                    brew install python3
```
bendk: If the recent installations don't contain `pip` then how does uninstalling and re-installing fix it?  I don't doubt that this works, I'm just curious why since I'm not a brew user.

docs/building.md
```diff
@@ -66,6 +72,14 @@ a number of hours to complete.
                export PATH="~/tools/gyp:$PATH"
                ```
             1. If you have additional questions, consult [this guide](https://github.com/mogemimi/pomdog/wiki/How-to-Install-GYP).
+            1. More recent version of Python (3.12+) have deprecated `distutils`, a dependency of the above script. If you encounter a ModuleNotFoundError, you can run first:
+                ```shell
+               brew install python-setuptools
+               ```
```
bendk, line 77: I'm pretty sure I've seen this same issue, how about making this a bit more generic?  I'm thinking something like "If you encounter a ModuleNotFoundError, then install setuptools on your system (for example `brew install python-setuptools)`"

docs/building.md
bendk: I never set `ANDROID_SDK_ROOT` just `ANDROID_HOME`.  How about removing the reference from that other doc rather than adding one here?

bendk (changes requested):
This is great, thanks for keeping these fresh.  I had a couple suggestions, but feel free to push back on any/all of them.

docs/building.md
thesuzerain: I'm actually a little bit surprised it worked as well. I had assumed it was because of `brew install python` vs `brew install python3`, but on checking now, these are aliases. The reinstallation step in particular is noted [here](https://stackoverflow.com/questions/60711023/after-installing-python3-with-homebrew-homebrew-doesnt-create-symlink-of-pytho).  it worked perfectly for me, but on reflection here it's unclear why- I have a couple thoughts but I'm unsure so I'll make a note and examine them later.

The source [issue](https://redirect.github.com/orgs/Homebrew/discussions/3596) about the symlinks is pretty clear, though, and `brew info python` shows the intended symlinks, so for now I've changed my comment to give a better + more idiomatic solution more in line with what `brew` intends.

docs/howtos/locally-published-components-in-firefox-ios.md
```diff
@@ -86,6 +86,12 @@ for example, from the root of your application-services directory:
 cp ./megazords/ios-rust/Sources/MozillaRustComponentsWrapper/Generated/*.swift ../firefox-ios/MozillaRustComponents/Sources/MozillaRustComponentsWrapper/Generated/.
 ```
 
+4. Remove the copied over `glean_sym.swift` file. This exists to make glean-sym work in Rust, but is not needed in linking. From the same directory:
+
+```bash
+rm ../firefox-ios/MozillaRustComponents/Sources/MozillaRustComponentsWrapper/Generated/glean_sym.swift
+```
+
```
thesuzerain, line 94: I've added this step per recent discussion. We could also the `cp` and `rm` in one command with something like:

```find /some/folder -maxdepth 1 -name '*.swift' ! -name 'glean_sym.swift' -exec cp {} /other/folder/ \;``` but I like the clarity of separating them. Let me know what you think, I'm not tied to it.

Comment (bendk) [reply docs/howtos/locally-published-components-in-firefox-ios.md:94]: Agreed, I like this way of presenting it better as well.

File: docs/building.md [hunk 5]
```diff
@@ -66,6 +71,14 @@ a number of hours to complete.
                export PATH="~/tools/gyp:$PATH"
                ```
             1. If you have additional questions, consult [this guide](https://github.com/mogemimi/pomdog/wiki/How-to-Install-GYP).
+            1. If you encounter a `ModuleNotFoundError`, try installing setup tools on your system. For example:
```
bendk:
```suggestion
            1. If you encounter a `ModuleNotFoundError`, try installing setuptools on your system. For example:
```

bendk (approved):
Looks great, I think we're ready to merge this.

PR #7483, opened 2026-07-19
Bug 2056071 - remove BridgedEngine, app-services side.

mhammond:
Removes lots of code and combined desktop and mobile syncing significantly. Tagging Bastian too, sharing the love 😆 ❤️ 

This uniffis (ha) the desktop and mobile sync engine implementations. Desktop still uses the macro generated "bridge", and that's a duck-type JS model - which all of desktop sync already is so, yeah.

Breaking change, will need to land https://phabricator.services.mozilla.com/D312903 and vendor this.

  - Note:


components/sync15/src/engine/bridged_engine.rs
```diff
@@ -359,27 +173,29 @@ impl BridgedEngineWrapper {
...
+/// The macro always emits `set_clients`; a crate whose UDL doesn't declare it
+/// (logins, webext-storage) simply leaves that inherent method unbound, which is
+/// harmless.
 #[macro_export]
 macro_rules! uniffi_bridged_engine {
     ($name:ident, $guid:ty) => {
         // This is what UniFFI exposes; it does nothing other than delegate to
-        // the shared `BridgedEngineWrapper`. See
-        // services/interfaces/mozIBridgedSyncEngine.idl for the Desktop contract.
+        // the shared `BridgedEngineWrapper`, which adapts our `SyncEngine`.
```
bendk, line 176: What's stopping us from removing this macro and the various `[Component]BridgedEngine` interfaces and replacing them with a single thing (maybe a trait interface that's defined in `sync15`)?  The differences I can see are the GUID handling and the error type, which don't seem that big?  I'm not suggesting you do anything different in this PR, I'm just wondering where we are in the larger process.

components/sync15/src/engine/sync_engine.rs
```diff
@@ -232,6 +226,19 @@ pub trait SyncEngine {
     fn wipe(&self) -> Result<()> {
         unimplemented!("The engine does not implement wipe, no wipe should be requested")
     }
+
+    /// A couple of desktop specific "bridged engine" helpers, where the
+    /// last-modified timestamps for collections are handled differently;
+    /// who does the `get_collection_request()` etc impacts the owner of the
+    /// timestamp.
+    /// Engines should do both or neither, longer term it should be absorbed.
```
bendk: This is connected to `last_sync`, but it's also for `reset_last_sync`.  What about changing it to double-slashes and adding a newline before the method?  Alternatively, it could be reworded to be about `last_sync` with a note at the end that says something like "this also applies to "reset_last_sync".

bendk (approved):
Looks great to me, I love seeing all the deleted lines.

jo (commented):
Looks good to me

components/sync15/src/engine/bridged_engine.rs
mhammond: Most recent version changes the guid handling too. Not entirely sure what you mean otherwise though - SyncEngine already is a trait in sync15, and bridged engine *was* a trait, but this is trying to roll them up so that only 1 engine need be implemented. But I'm totally up for future improvements!

mhammond:
hey @DimiDL, this now also needed to touch addresses - I think it looks fine but want your review and also serves as a heads-up that I'm looking into landing this.

components/sync15/src/engine/bridged_engine.rs
bendk: I think I was wondering if we could have a `sync15::BridgedEngine` trait and a blanket impl for everything that implements `SyncEngine`.  I'm not sure that's better than the this PR, but doing away with the macro seems good to me.

DimiDL:
Thank you for the heads-up. This looks good to me!

PR #7487, opened 2026-07-20
test: Updating older cross-build smoke tests (iOS, Android)

thesuzerain:
This is a _draft_ PR bringing in smoke tests scripts for building application-services against various targets (Android and iOS here, HNT to come, see below). It deprecates some older similar tests that are very out of date, and remakes them with up-to-date building methods (and more test flexibility).

I used existing helpers from the existing smoke tests where possible.

See: [build instructions for iOS](https://github.com/mozilla/application-services/blob/main/docs/howtos/locally-published-components-in-firefox-ios.md) and [build instructions for Fenix](https://github.com/mozilla/application-services/blob/main/docs/howtos/locally-published-components-in-fenix.md) for context about the steps.

This currently makes no breaking changes. It is also thought that the deprecated smoke tests have long fallen out of use.

### Remaining work:

We want to build against HNT if possible as well, but currently, HNT doesn’t have a system for building against a local a-s like the other two do. There are a couple possibilities I've explored (eg: redirecting cargo paths) but none that have been trivial. As it may require non-test-script changes, I would propose we do that in another PR.

There is also a good amount of iOS and Android tests that fail naturally (seemingly even in clean unrelated builds), and most possible regressions in a-s don’t cause a build or test failure, so it may be prudent in the future to make to extend iOS/Android tests to cover a-s behaviour specifically now that we have these scripts.

  - Note:

### DIFF
--- a/automation/build_against_all.py
+++ b/automation/build_against_all.py
@@ -14,15 +14,14 @@
 #       --action            => Can be either `run-tests` (default) or `build-without-testing`
 #       --use-local-repo    => Use a local firefox-ios repository instead (at the provided path). Exclusive with `remote-repo-url`.
 #       --verbose           => Includes the stdout of subprocesses (like the xcodebuild output, or other bootstrapping scripts)
-#       --branch            => What branch of the firefox-ios repo to use
-# TODO: args
+#       --allow-clears      => Clear existing uniffi bindings, swift files, and so on during the various build processes.
+#       --action            => Either 'run-tests' or 'build-without-testing
 
 import argparse
 import time
 from shared import err_msg, step_msg
 from build_against_fenix import build_against_fenix
 from build_against_ios import build_against_ios
-from build_against_hnt import build_against_hnt
 
 parser = argparse.ArgumentParser(
     description="Run groups of tests against this application-services working tree."
@@ -42,7 +41,7 @@ parser.add_argument(
     "--action",
     required=True,
     choices=["run-tests", "build-without-testing"],
-    help="Run the following action once firefox-ios is set up.",
+    help="Run the following action for target's test",
 )
 
 args = parser.parse_args()
@@ -52,9 +51,9 @@ allow_clears = args.allow_clears
 action = args.action
 
 # Build against iOS
-# TODO: We *may* be able to run this one concurrently to the other two?
 start_time_ios = time.time()
-success_ios = build_against_ios(None, None, clear_previous_bindings=allow_clears, clean_ios_caches=allow_clears, verbose=verbose, action=action)
+success_ios = build_against_ios(None, None, scheme="Fennec", test_plan="Smoketest", 
+                                clear_previous_bindings=allow_clears, clean_ios_caches=allow_clears, verbose=verbose, action=action)
 time_diff_ios = time.time() - start_time_ios
 
 # Build against Fenix
--- a/automation/build_against_all.py
+++ b/automation/build_against_all.py
@@ -4,7 +4,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 # Purpose: Run various smoke tests against this application-services working tree.
-# Requirements: 
+# Requirements:
 # - python
 # - application-services built and working.
 # For the mac builds:
@@ -33,10 +33,16 @@ parser.add_argument(
     required=True,
     help="Path to existing mozilla-central directory.",
 )
-parser.add_argument("--verbose", help="Includes subprocess running logs.", action=argparse.BooleanOptionalAction)
-parser.add_argument('--allow-clears', 
-                    help="Clear existing uniffi bindings, swift files, and so on during the various build processes.",
-                    action=argparse.BooleanOptionalAction)
+parser.add_argument(
+    "--verbose",
+    help="Includes subprocess running logs.",
+    action=argparse.BooleanOptionalAction,
+)
+parser.add_argument(
+    "--allow-clears",
+    help="Clear existing uniffi bindings, swift files, and so on during the various build processes.",
+    action=argparse.BooleanOptionalAction,
+)
 parser.add_argument(
     "--action",
     required=True,
@@ -52,13 +58,29 @@ action = args.action
 
 # Build against iOS
 start_time_ios = time.time()
-success_ios = build_against_ios(None, None, scheme="Fennec", test_plan="Smoketest", 
-                                clear_previous_bindings=allow_clears, clean_ios_caches=allow_clears, verbose=verbose, action=action)
+success_ios = build_against_ios(
+    None,
+    None,
+    scheme="Fennec",
+    test_plan="Smoketest",
+    clear_previous_bindings=allow_clears,
+    clean_ios_caches=allow_clears,
+    verbose=verbose,
+    action=action,
+)
 time_diff_ios = time.time() - start_time_ios
 
 # Build against Fenix
 start_time_fenix = time.time()
-success_fenix = build_against_fenix(firefox_dir, None, prefix_ff="fenix", prefix_as=None, clear_bindings=allow_clears, verbose=verbose, action=action)
+success_fenix = build_against_fenix(
+    firefox_dir,
+    None,
+    prefix_ff="fenix",
+    prefix_as=None,
+    clear_bindings=allow_clears,
+    verbose=verbose,
+    action=action,
+)
 time_diff_fenix = time.time() - start_time_fenix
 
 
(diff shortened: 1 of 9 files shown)


automation/build_against_ios.py
```diff
@@ -0,0 +1,388 @@
...
+    run_cmd_is_successful,
+    dir_file_sanity_check,
+)
+
+DEFAULT_REMOTE_REPO_URL = "https://github.com/mozilla-mobile/firefox-ios.git"
+MOZILLA_RUST_COMPONENTS_IOS_PATH = "MozillaRustComponents/Package.swift"
+MOZILLA_RUST_COMPONENTS_AS_PATH = "megazords/ios-rust/MozillaRustComponents.xcframework"
+
+
+def replace_swift_package_artifact(ios_repo_path, as_repo_path):
```
thesuzerain, line 48: I don't love using regex here but given how often this artifact file is updated (with hashes, etc) it doesn't really seem possible to use a fixture for it.

automation/build_against_all.py
```diff
@@ -0,0 +1,104 @@
...
+    prefix_ff="fenix",
+    prefix_as=None,
+    clear_bindings=allow_clears,
+    verbose=verbose,
+    action=action,
+)
+time_diff_fenix = time.time() - start_time_fenix
+
+
+did_tests_string = "" if action != "run-tests" else " (and tested)"
```
thesuzerain, line 141: HNT is not included yet/in this PR because we don't currently seemingly have a way to build against HNT yet that works beyond:
- writing in the a-s repository itself
- committing then vendoring
- redirecting cargo paths (which doesn't work 100% yet). 
I'm going to be exploring the third option but it may require code changes outside of these scripts, so to keep this PR smaller, this includes test for the two existing ones.

automation/build_against_all.py
```diff
@@ -0,0 +1,104 @@
...
+# Purpose: Run various smoke tests against this application-services working tree.
+# Requirements:
+# - python
+# - application-services built and working.
+# For the mac builds:
+# - xcpretty (`gem install xcpretty`)
+# - xcode + xcodebuild + xcodetools setup and running (a successful build of the firefox-ios repository)
+# Arguments:
+#       --action            => Can be either `run-tests` (default) or `build-without-testing`
+#       --use-local-repo    => Use a local firefox-ios repository instead (at the provided path). Exclusive with `remote-repo-url`.
```
jonesetc: Looks outdated? There is a `--firefox-dir FIREFOX_DIR`, but no `--use-local-repo` or `--remote-repo-url`

automation/build_against_fenix.py
```diff
@@ -0,0 +1,283 @@
...
+    step_msg(f"Using `mozconfig` path: `{moz_config_location}`. Displaying:")
+    with open(moz_config_location) as f:
+        print(f.read())
+
+    # Basic sanity check here. Not remotely exhaustive, just to make sure the wrong directory wasn't passed.
+    step_msg("Checking for sanity of firefox repository...")
+    if not dir_file_sanity_check(
+        firefox_repo_path,
+        "mozilla-central",
+        ["mach", "CLOBBER", "gradlew", "Cargo.toml", "local.properties"],
```
jonesetc, line 135: @thesuzerain I have a functioning local android build, but I don't have a `local.properties` in there at all. I'll continue on trying this with just an empty file there, but seems like there's a chance to be here without one

thesuzerain: Got it- if it works it should be fine. It might be necessary though for [this step](https://github.com/mozilla/application-services/blob/main/docs/howtos/locally-published-components-in-fenix.md#mozilla-central) though, to ensure it's being built against the local a-s directory instead of the m-c tree.
```shell
autoPublish.application-services.dir=../application-services
```

If you put in an intentional error in the external a-s folder does it fail to build?

jonesetc:
Was able to force an android regression in testing with this patch:
```patch
diff --git a/components/ads-client/src/lib.rs b/components/ads-client/src/lib.rs
index 2088ee16c..1279423e9 100644
--- a/components/ads-client/src/lib.rs
+++ b/components/ads-client/src/lib.rs
@@ -72,10 +72,10 @@ impl MozAdsClient {
     #[uniffi::method(default(options = None))]
     pub fn record_impression(
         &self,
-        impression_url: String,
+        impression_url_foo: String,
         options: Option<MozAdsCallbackOptions>,
     ) -> AdsClientApiResult<()> {
-        let url = AdsClientUrl::parse(&impression_url).map_err(|e| {
+        let url = AdsClientUrl::parse(&impression_url_foo).map_err(|e| {
             ComponentError::RecordImpression(CallbackRequestError::InvalidUrl(e).into())
         })?;
         let ohttp = options.map(|o| o.ohttp).unwrap_or(false);
```

and running `automation/build_against_fenix.py --prefix-ff fenix --prefix-as ads-client --firefox-dir ../firefox` produced:
```
...
> Compiling firefox with mozconfig with `./gradlew fenix:assembleDebug` (mozconfig=`/var/folders/lr/zvwpm2dn5n133j0st4z6n6c00000gn/T/tmphrsr28ca-test-fenix/mozconfig_android`)...
e: file:///.../firefox/mobile/android/android-components/components/service/mars/src/main/java/mozilla/components/service/mars/MozAdsUseCases.kt:50:70 No parameter with name 'impressionUrl' found.
e: file:///.../firefox/mobile/android/android-components/components/service/mars/src/main/java/mozilla/components/service/mars/MozAdsUseCases.kt:50:70 No value passed for parameter 'impressionUrlFoo'.
```

automation/build_against_fenix.py
jonesetc: Yep, it was needed, and so long as it's documented there seems right to check for!

automation/build_against_all.py
thesuzerain: Closing because HNT will exist after discussion with maintainers

docs/howtos/locally-published-components-in-firefox-hnt.md
```diff
@@ -0,0 +1,88 @@
...
+Now that `components` will read from your local build, you can build, run, and test. From your local m-c checkout, run:
+
+```bash
+./mach build
+```
+
+And if so desired:
+
+```bash
+./mach run
```
thesuzerain, line 74: I tried this strategy this morning and it's having a strange new error with `link_nss_rustlib` from I think recent commits- will update or delete based on trying it out, but wnted to flag this.

thesuzerain: I believe this is related to some breaking changes in some of the monorepo patches this morning (things out of sync, etc) and not related to this. Will wait for some patches and I'll ensure that these are all OK before merging.

automation/build_against_hnt.py
```diff
@@ -0,0 +1,298 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# Purpose: Run Firefox fenix tests against this application-services working tree.
```
mhammond: I have no idea what HNT stands for and google isn't helping. It looks like it's somehow related to desktop, but this all says Fenix.

docs/howtos/locally-published-components-in-firefox-hnt.md
```diff
@@ -0,0 +1,88 @@
+# How to locally test application-services components on HNT / Desktop
```
mhammond, line 1: ditto here, no idea what a "hnt desktop" is?

docs/howtos/locally-published-components-in-firefox-hnt.md
```diff
@@ -0,0 +1,88 @@
...
+
+---
+
+## At a glance
+
+**Goal:** Build a local Firefox Desktop against a local Application Services.
+
+**Current workflow (recommended):**
+
+1. Verify the local build of `A-S` is ready for a desktop build.
```
mhammond, line 13: You could consider documenting the strategy of having your a-s changes on a branch and vendoring from that, avoiding symlinks etc. For many use-cases, the re-vendor is *needed*, so it isn't going to be wasted work but instead will be working on what needs to land anyway.

(IOW, using this process when you need to actually re-vendor anyway is a waste of time, might as well just work directly on the vendor)

docs/howtos/locally-published-components-in-firefox-hnt.md
```diff
@@ -0,0 +1,88 @@
...
+2. Move the A-S components folder in M-C/Firefox to a temporary rename.
+3. Create a symlink between the A-S and M-C components folders.
+4. Generate uniffi bindings.
+5. Run and build.
```
mhammond, line 17: Should probably include "./mach vendor rust" and "./mach cargo vet" - only necessary if there have been dependency changes, but that would block (or at least slow) landing.

mhammond (approved):
This all seems very elaborate and unlikely to be widely used, but you aren't forcing anyone to use this and if it scratches your itch I've no objection. I think "hnt" is very unclear though.

docs/howtos/locally-published-components-in-firefox-hnt.md
thesuzerain: Ah, HNT to my understanding is referring to the home newtab which I've been using synonymously with desktop here. I'll add a bit more clarification.

thesuzerain:
> This all seems very elaborate and unlikely to be widely used, but you aren't forcing anyone to use this and if it scratches your itch I've no objection. I think "hnt" is very unclear though.

Fixed the HNT stuff. 

Point well taken- you are correct that vendoring is a better solution for this, I mostly wanted to avoid committing in an automated test.  I'll discuss and maybe rewrite this test in the future depending on our needs.

PR #7490, opened 2026-07-21
Bug 2050036 - Add Store::shutdown() to close the autofill DB connection

DimiDL:
The connection is otherwise only closed when the Store is dropped, which on Firefox Desktop happens during GC at shutdown -- past the late-write barrier, so the flush crashes debug builds.

Wrap the connection in Mutex<Option<AutofillDb>> behind a fallible lock_db() helper (mirroring LoginStore); shutdown() closes it early, and later operations return DatabaseClosed.

  - Note:

### DIFF
--- a/components/autofill/src/db/store.rs
+++ b/components/autofill/src/db/store.rs
@@ -10,13 +10,14 @@ use crate::db::{
 };
 use crate::error::*;
 use error_support::handle_error;
+use parking_lot::{MappedMutexGuard, Mutex, MutexGuard};
 use rusqlite::{
     types::{FromSql, ToSql},
     Connection,
 };
 use sql_support::{self, run_maintenance, ConnExt};
 use std::path::Path;
-use std::sync::{Arc, Mutex, Weak};
+use std::sync::{Arc, Weak};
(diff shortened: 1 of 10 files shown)


components/autofill/src/db/store.rs
```diff
@@ -48,49 +49,53 @@ pub fn get_registered_sync_engine(engine_id: &SyncEngineId) -> Option<Box<dyn Sy
 
 // This is the type that uniffi exposes.
 pub struct Store {
-    pub(crate) db: Mutex<AutofillDb>,
+    pub(crate) db: parking_lot::Mutex<Option<AutofillDb>>,
```
mhammond: fwiw, many of our crates use `parking_lot::Mutex` and typically just `use` it directly rather than fully qualifying it in each use. I'm not against doing it this way, but fyi

mhammond (dismissed):

components/autofill/src/db/store.rs
DimiDL: thank you for the review! Updated to follow the convention.

mhammond (approved):

PR #7491, opened 2026-07-21
Bug 2056116 - Don't crash sync on logins with invalid guids

jo:
https://bugzilla.mozilla.org/show_bug.cgi?id=2056116

A login guid that's invalid for the sync server panicked the uploader ("record's ID is invalid").

Fixing the problem on 3 levels:
- harden login sync to not panic on an invalid login
- harden sync to not block entire sync on single invalid login
- validate guid in `add_many_with_meta`
- fix existing invalid records with a migration

  - Note:

### DIFF
--- a/components/logins/src/sync/engine.rs
+++ b/components/logins/src/sync/engine.rs
@@ -258,11 +258,22 @@ impl LoginsSyncEngine {
         ))?;
         let bsos = stmt.query_and_then(
             named_params! { ":fxa_origin": FXA_CREDENTIALS_ORIGIN },
-            |row| {
+            |row| -> Result<Option<OutgoingBso>> {
                 self.scope.err_if_interrupted()?;
-                Ok(if row.get::<_, bool>("is_deleted")? {
+                let guid: Guid = row.get::<_, String>("guid")?.into();
+                // A guid that is invalid for the sync server would panic the uploader
+                // (bug 2056116). Such a record can never have reached the server, so
+                // skip it instead of blocking the whole sync; the schema migration
+                // repairs these locally so they can sync on a later run.
+                if !guid.is_valid_for_sync_server() {
+                    warn!(
+                        "skipping outgoing login with a guid that is invalid for the sync server"
+                    );
+                    return Ok(None);
+                }
+                Ok(Some(if row.get::<_, bool>("is_deleted")? {
                     let envelope = OutgoingEnvelope {
-                        id: row.get::<_, String>("guid")?.into(),
+                        id: guid,
                         sortindex: Some(TOMBSTONE_SORTINDEX),
                         ..Default::default()
                     };
@@ -273,10 +284,10 @@ impl LoginsSyncEngine {
                         EncryptedLogin::from_row(row)?.into_bso(db.encdec.as_ref(), unknown)?;
                     bso.envelope.sortindex = Some(DEFAULT_SORTINDEX);
                     bso
-                })
+                }))
             },
         )?;
-        bsos.collect::<Result<_>>()
+        bsos.filter_map(|r| r.transpose()).collect::<Result<_>>()
     }
 
     fn do_apply_incoming(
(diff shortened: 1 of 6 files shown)


components/logins/src/sync/engine.rs
```diff
@@ -258,11 +258,22 @@ impl LoginsSyncEngine {
         ))?;
         let bsos = stmt.query_and_then(
             named_params! { ":fxa_origin": FXA_CREDENTIALS_ORIGIN },
-            |row| {
+            |row| -> Result<Option<OutgoingBso>> {
                 self.scope.err_if_interrupted()?;
-                Ok(if row.get::<_, bool>("is_deleted")? {
+                let guid: Guid = row.get::<_, String>("guid")?.into();
+                // A guid that is invalid for the sync server would panic the uploader
+                // (bug 2056116). Such a record can never have reached the server, so
```
mhammond, line 265: sorry you are hitting this. I'm not sure of the entire context here, but this assertion was really more a sanity check for our code - ie, the server *will* accept anything as a GUID. This means that if this record somehow ended up coming from the old passwords engine, then the record may well already be on the server.

It's not clear to me how much of your strategy here relies on "Such a record can never have reached the server". Another strategy (which I'm not sure I like entirely, but which I think would work) would be to just remove the assertion. That would leave us without this safely check, but I don't recall anyone ever hitting it in the past and if it turns out this ID was being used in the past then maybe that's not such a crazy idea.

mhammond: (oops, but I also meant to say this patch looks fine to me if this is the route we should take)

jo: Ah, cool, it’s definitely possible that this ID is already on our servers. In that case, I’ll take the check out of here but still fix the `add_with_meta`. And do you think I can migrate the IDs that are already in the store (the ones that were just migrated via `add_with_meta` on the desktop) here?

bendk: If the GUID hit the server, then wouldn't a mobile logins engine download it and re-upload a record with the same GUID and hit the same panic?  I'm not against either solution (removing the assertion or skipping the record), but I'd like to understand what's happening better before we commit to either one of them.

jo: Good point - maybe it was Desktop sync only, no mobile involved?

I think this happened:
1. I added `add_with_meta` without validating guid
2. this was used on Desktop to import logins
3. now these got synced, and for one user it crashed the entire fx because this local-only login had an invalid id

Before Rust logins, this login could have been synced with the old logins sync engine, but maybe only to other Desktop's?

That's why I thought we need to fix this on three levels:

1. prevent `add_with_meta` to add invalid ids (or fix them, guarded by a flag)
2. fix this login on the users desktop
3. somehow harden sync to not break on a single invalid login

mhammond: I think this only makes sense if it came from desktop originally - this error should mean it was impossible for it to originate on mobile. Even then it seems odd though - "`add_with_meta` without validating guid" doesn't fully explain it because *someone* generated a guid (ie, it's never a free-form text field), and the desktop password engine has always generated "valid" GUIDs up to now.

Fixing it is also complicated in theory. It would mean you need to leave a tombstone behind so it doesn't get duplicated on other devices. And it still leaves the (improbable!) possibility it gets edited on another device without the migration implemented yet, so the tombstone is resurrected.

This all seems so edge-casey - if we can't identify how this ID might have been created legitimately, then the best thing probably is to just skip the record entirely, both reading and writing - it seems unlikely that's actual data loss. That strategy couldn't work for bookmarks though - you can't sanely just skip a single parent folder. So it seems unfortunate that the problem looks like it needs to be solved in every engine which sucks as ever engine is theoretically susceptible.

If it helps or hurts, you can almost certainly reproduce this with about:sync - it lets you edit/add records via plain JSON - that's how I verified the server doesn't reject crazy guids. Indeed, if it seems like there's exactly 1 user with this problem, someone doing silly things with about:sync is the most likely explanation I can think of.

jo: Yeah, that leaves me pretty stumped, too.

Just a thought: It could be a very old login, from a time when it might have been possible to create strange ids, eg via import or whatever. There were also times when extensions could write logins directly.

How exactly would you handle this now? Should we monitor the error for now to see if it occurs more often?

mhammond: > eg via import or whatever.
yeah! Meant to suggest this as a theory.

If literally one device, then yeah, I'd say ignore it. I don't really know how to read those stats in terms of client vs crash counts - but raw crash counts are very low, and I *think* the check makes sense rather than loosening here? So somehow making noise and ignoring this rather than crashing or otherwise blocking other items seems ideal? Sorry I haven't dug into the actual impl here to know if that actually makes sense

mhammond: > and I think the check makes sense

Does it? This is a genuine question? 😅 I'd be easily convinced we just drop the assert - love opinions from everyone :)

jo: The [`is_valid_for_sync_server` check](https://github.com/mozilla/application-services/blob/main/components/support/guid/src/lib.rs#L213-L219) behaves differently from the [server-side Python implementation referenced in the comment above the function](https://github.com/mozilla-services/server-syncstorage/blob/d92ef07877aebd05b92f87f6ade341d6a55bffc8/syncstorage/bso.py#L24) as well as in the [Rust successor's validation regexp](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncserver/src/server/mod.rs#L41): the client-side validation also checks for a comma in the ID. This difference in validation actually argues in favor of removing the client-side validation entirely.

At the very least, as a minimal fix, we definitely shouldn’t panic about this. What do you think - the release is in a week, and I think we might want to include a fix for 2056116?

jo: So I've broken this down into two parts:

1. Validate IDs during `add_with_meta` (and optionally fix them via a feature) https://github.com/mozilla/application-services/pull/7524
2. The towel: https://github.com/mozilla/application-services/pull/7536

PR #7496, opened 2026-07-26
Add opt-in `mozbuild-rustlib` Cargo feature chain and `link_nss` implementation for NSS crates

ahochheiden:
This necessary for the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D295090

In `mozilla-central`, `application-services` crates need to find NSS in the
objdir rather than through the standalone build's `NSS_DIR`. When
`MOZ_TOPOBJDIR` is set, `link_nss` emits cargo link directives for the NSS
dylibs in the build tree through the `mozbuild` crate. That is the default
path. The opt-in `mozbuild-rustlib` feature switches to `link_nss_rustlib`,
which also links `mozpkix` and `pure_virtual` statically. Megazord and
rusttests need that static linkage, but libxul already links them itself, so
libxul leaves the feature off.


mhammond (approved):
nice!

PR #7498, opened 2026-07-27
Reference `UNIFFI_META_*` symbols in the megazord stub so the linker keeps them

ahochheiden:
This necessary for the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D313317


mhammond (commented):
Does the existing pattern work here? All this code is solving the exact same problem, and what I hate about this version is that it looks like every public api change is going to need an entry here which is going to regularly shoot people's toes off.

ahochheiden:
> Does the existing pattern work here? All this code is solving the exact same problem, and what I hate about this version is that it looks like every public api change is going to need an entry here which is going to regularly shoot people's toes off.

I agree this is not optimal, but the checksum trick doesn't work for most of the entries I added because they're data only. 

We could maybe use `--whole-archive` or `codegen-units=1` to get around this approach entirely, but I haven't thought through all of the consequences with those approaches (We do `codegen-units=1` for Release on Firefox, and it's a huge (in the 20x range) build time regression.

ahochheiden:
Hmm, I seem to have irreparably merged these unintentionally by trying to change the order...

mhammond (approved):
I hate it but have no better ideas :)

PR #7500, opened 2026-07-27
Use android-components publish.gradle for app-services components

ahochheiden:
This necessary for the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D292602

In `mozilla-central`, a-s subprojects should publish through the same
publish.gradle as android-components, not a-s's own. This routes them
there when mozconfig is set, while the standalone a-s build is
unchanged. `configurePublish` takes explicit (groupId, artifactId,
description) args because m-c doesn't have a-s's `rootProject.config`.

We also add a `publishAppServicesAndroid` task that aggregates every
app-services component's publish task, so the geckoview archive build
publishes the full set to local maven in one step.

### DIFF
--- a/components/ads-client/android/build.gradle
+++ b/components/ads-client/android/build.gradle
@@ -22,7 +22,7 @@ plugins {
 }
 
 apply from: "$appServicesRootDir/build-scripts/component-common.gradle"
-apply from: "$appServicesRootDir/publish.gradle"
+apply from: "$publishDir/publish.gradle"
 
 ext {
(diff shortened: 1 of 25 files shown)


mhammond (approved):
also a rubber stamp, not sure if you care about having someone more knowledgeable having a look.

ahochheiden:
> also a rubber stamp, not sure if you care about having someone more knowledgeable having a look.

Nick has reviewed this on Phabricator, so I think it's fine.

PR #7501, opened 2026-07-27
Package `libmegazord.so` into full-megazord AAR with NSS dependencies and `libsForTests` publication

ahochheiden:
This necessary for the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D292604

The full-megazord AAR ships libmegazord.so for Android consumers. Apps
that don't pull GeckoView still need NSS at runtime, so the AAR also
includes libnss3, libfreebl3, libsoftokn3, and libmozglue. The
libsForTests publication exposes the desktop libraries (megazord plus
the same NSS set) for JVM unit tests.

The `:machStagePackage` dependency only applies when GeckoView is part
of the Gradle build, so it is guarded on `findProject(":geckoview")`
for the standalone application-services build where that task does not
exist.

### DIFF
--- a/megazords/full/android/build.gradle
+++ b/megazords/full/android/build.gradle
@@ -22,7 +22,7 @@ android {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
-            consumerProguardFiles "$rootDir/proguard-rules-consumer-jna.pro"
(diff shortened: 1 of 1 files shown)


mhammond (approved):
rubber stamping this one as it is a bit over my head :)

ahochheiden:
> rubber stamping this one as it is a bit over my head :)

Nick has also reviewed this one on Phabricator, so it should be fine.

PR #7505, opened 2026-07-28
CI: Update the Rust image in use

badboy:
Note: rustup will still be used to install the right version as defined in rust-toolchain.toml or the minimum version defined in .circleci/config.yml

However newer Rust images come with newer tooling as well (such as a not outdated Python version)


bendk (approved):
🚀

PR #7513, opened 2026-07-29
Add address metadata APIs and a bridged sync engine to autofill

DimiDL:
Firefox Desktop is moving its address storage to this component and needs two things that are not currently exposed.

Storage APIs for importing records already persisted elsewhere - add_address_with_meta, add_many_addresses_with_meta, update_address_with_meta and add_many_address_tombstones - taking a caller-supplied guid, timestamps and change counter. The bulk variants run each record in a savepoint so one bad record neither aborts the migration nor leaves a row behind. Carrying the change counter means update_internal_address now takes a CounterUpdate of Increment, Leave or Set(i64).

A bridged sync engine, so Desktop's Sync framework can drive address sync: Store::addresses_bridged_engine() exposes the address engine the crate already has, which only implemented sync15::SyncEngine and so was unreachable from Desktop's mozIBridgedSyncEngine.

Nothing on mobile calls any of this; it is additive for the Desktop migration.

  - Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Autofill
 
 - Add `Store::shutdown()`, which closes the database connection early so it happens before Firefox Desktop's late-write shutdown barrier rather than during GC. Operations after shutdown return `DatabaseClosed`. ([Bug 2050036](https://bugzilla.mozilla.org/show_bug.cgi?id=2050036))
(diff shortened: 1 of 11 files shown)


jo (approved):
This looks good, I esp. like `with_savepoint`.

I think we could go without support for `sync_change_counter` update via `AddressMeta` because when this is used during import (where I think it is designed for) the local mirror will be always empty and the record is [transferred anyway](https://github.com/mozilla/application-services/blob/81cecb5ddbaf07760aef62b75f656e216c488875/components/autofill/src/sync/address/outgoing.rs#L28-L42).

Note that this will race with https://github.com/mozilla/application-services/pull/7483 which greatly simplifies the sync bridge code.

PR #7516, opened 2026-07-30
SyncEngine::wipe no longer has a default panicing implementation

mhammond:
I might as well not make this mistake twice 😅


jo (approved):
:)

